Issue #14, opened 2015-09-30
Handle missing width and height attributes for images


mjangda:
Not all `<img>` tags have width/height attributes which are required for the matching `<amp-img>` tag. We should find a way to include that somehow.

- [ ] If `srcset` or `sizes` are included, try to pull from them, if possible.
- [x] Pull from attachment metadata.
- [x] Fetch the remote image and calculate/save the size (`getimagesize` and wpcom's `getimagesize`).

Related #13

PR #40, opened 2015-10-08
Disables Photon.

### LINKED ISSUE #32: Images are compressed vertically
In this blog post: http://www.chrisfinke.com/2015/09/28/and-some-benches/amp/ the images are squished to about 60px tall but are still the full width of the column.

kraftbj:
Photon strips height/width attr from the img tag, causing many problems with AMP.

Fixes #32. See #14

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -35,6 +35,10 @@ function amp_add_actions() {
 		return;
 	}
 
+	if ( class_exists( 'Jetpack' ) ) {
+		require_once( dirname( __FILE__ ) . '/jetpack-helper.php' );
+	}
+
 	if ( false !== get_query_var( AMP_QUERY_VAR, false ) ) {
 		// TODO: check if post_type supports amp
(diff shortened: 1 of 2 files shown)


mjangda:
Sweet. Still might be good for us to run the image itself through Photon when we convert to elements `amp-img` but let's run with this for now.

kraftbj:
@mjangda Agreed. Right now, there isn't an easy way to give Photon a full WP url and return a fully Photon URL (e.g. example.com/wp-content/uploads/whatever-200x200.jpg and have it spit out i1.wp.com/example.com/wp-content/uploads/whatever.jpg?w=200&h=200 )

The closest function would convert the above example to i1.wp.com/example.com/wp-content/uploads/whatever-200x200.jpg which both eliminates the advantages of the single-cached-dynamically-resized image bit of Photon so not terribly crazy about that.

tl;dr: Totally see this as a bandaid to make AMP functional under Photon until a better solution is vetted and flushed out.

Issue #15, opened 2015-09-30
Jetpack support for wpcom file


mjangda:
Need to remove PHP 5.3-isms like anonymous functions.

Need to also test against Jetpack :)

kraftbj:
Need to `remove_filter` for likes, sharedaddy, and related posts.
Untested on if using this for `remove_filter` works, but function/priority for the add_filters
`array( 'Jetpack_Likes', 'post_likes' ), 30`
`'sharing_display`, 19`
`array( 'Jetpack_RelatedPosts', 'filter_add_target_to_dom' ), 40 )`

PR #44, opened 2015-10-08
Jetpack: remove Related Posts' placeholder.

jeherve:
Related Posts are not outputted on AMP pages.
No use to display a "Related" headline at the end of the posts.

This PR removes that headline.

screen shot 2015-10-08 at 5 00 48 pm

### DIFF
--- a/jetpack-helper.php
+++ b/jetpack-helper.php
@@ -2,8 +2,16 @@
 
 // Jetpack bits.
 
-add_action( 'pre_amp_render', 'amp_jetpack_disable_photon' );
-add_action( 'pre_amp_render', 'amp_jetpack_disable_related_posts' );
+add_action( 'pre_amp_render', 'amp_jetpack_mods' );
+
+/**
+ * Disable Jetpack features that are not compatible with AMP.
+ *
+ **/
+function amp_jetpack_mods() {
+	amp_jetpack_disable_photon();
+	amp_jetpack_disable_related_posts();
+}
 
 /**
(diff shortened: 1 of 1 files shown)


mjangda:
Maybe we should put all Jetpack mods in the same callback for `amp_pre_render`? 

```
add_action( 'pre_amp_render', 'amp_jetpack_mods' );
function amp_jetpack_mods() {
    amp_jetpack_disable_photon();
    amp_jetpack_disable_related_posts();
}
```

/cc @kraftbj

jeherve:
@mjangda That makes sense! See 5a5e6f31dbfcbe559eff96720f34bbf463c2d7d8

kraftbj:
:y: I'll merge this in and roll into the branch that I'm working on to unify the WP.com/Jetpack helpers. We'll need to do the same thing for Likes and Sharedaddy.

#15

luckycabarlo:
What if i just created my own plugin for the related post and share/like to be outputted on post. Is there a way that we can just get the raw result from "the_content" code from wordpress?

kraftbj:
@luckycabarlo `get_the_content()` gets the content directly from the database with zero filtering (meaning shortcodes aren't parsed, paragraphs aren't auto-generated, videos aren't automatically embedded, etc).

luckycabarlo:
@kraftbj what I mean is why don't we just get the default filters coming from the wordpress core. With this approach we can eliminate all added filters coming from <code>the_content()</code>

PR #45, opened 2015-10-08
Unify wpcom and Jetpack helpers

### LINKED ISSUE #15: Jetpack support for wpcom file

kraftbj:
This PR does the following:
- Make wpcom-helper.php 5.2 friendly.
- Make wpcom-helper.php Jetpack friendly through `IS_WPCOM` and `function_exists` checks.
- Moves the Related Posts and Photon Jetpack helpers into wpcom-helper.php
- Add compat for Sharedaddy/Likes (by removing them).

A second set of eyes appreciated. Tested on my personal site, but didn't test against a 5.2.x environment directly.

Hopefully resolves #15


wpcom-helper.php
```diff
@@ -67,29 +85,55 @@ function wpcom_amp_get_stats_extras_url() {
...
+
+function wpcom_helper_pre_amp_render() {
+	if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) { // WP.com only.
+		add_filter( 'post_flair_disable', '__return_true', 99 );
+
+		remove_filter( 'the_title', 'widont' );
+
+		remove_filter( 'pre_kses', array( 'Filter_Embedded_HTML_Objects', 'filter' ), 11 );
+		remove_filter( 'pre_kses', array( 'Filter_Embedded_HTML_Objects', 'maybe_create_links' ), 100 );
+	} else { // Jetpack only.
```
mjangda: I think we should have an explicit Jetpack check here rather than rely on the `else`. Also, can we split up the wpcom and Jetpack bits into two separate functions?

kraftbj: I relied on the `else` since this file was only included when `JETPACK__VERSION` is defined, which is commonly shared between both platforms (WP.com has the `JETPACK__VERSION` set to `wpcom`). As far as I could tell, there was no usage if checking that specifically, so defaulted to the typical wp.com check to reduce likelihood of a change breaking things (as parts of the shared codebase assumes the constant is set).

For the check to determine if it is Jetpack, we can check for `JETPACK__VERSION != 'wpcom'. Initially, I was going to opt for `class_exists( 'Jetpack' )`; however, that would return true on wp.com as well. Any particular thoughts or ideas?

westonruter:
This should rather be added to the Jetpack plugin itself. See #1021.

PR #50, opened 2015-10-14
Update code to improve readability

mitogh:
Adding some code changes in order to improve readability, in the end is the same result but in my personal opinion much easier to read.

### DIFF
--- a/class-amp-img.php
+++ b/class-amp-img.php
@@ -29,12 +29,13 @@ class AMP_Img_Converter extends AMP_Converter {
 
 				// Workaround for https://github.com/Automattic/amp-wp/issues/20
 				// responsive + float don't mix
-				if ( isset( $attributes['class'] )
-					&& (
-						false !== strpos( $attributes['class'], 'alignleft' )
-						|| false !== strpos( $attributes['class'], 'alignright' )
-					)
-				) {
+				$img_class = isset( $attributes['class'] ) ? $attributes['class'] : '';
+				// alignleft or alignright
+				$pattern = "/alignleft|alignright/";
+				preg_match_all( $pattern, $img_class, $matches );
+				// test if there are any match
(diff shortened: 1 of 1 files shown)


class-amp-img.php
```diff
@@ -55,16 +56,9 @@ private function filter_attributes( $attributes ) {
...
-			switch ( $name ) {
-				case 'src':
-				case 'alt':
-				case 'width':
-				case 'height':
-				case 'class':
-					$out[ $name ] = $value;
-					break;
-				default;
-					break;
```
mjangda: The switch is an explicit choice here because it makes it easy to skim through the attributes we're whitelisting and let's us handle attribute special attributes more easily if needed (e.g. if `alt` needed to be manipulated in some way we could just break it out into its own `case`).

class-amp-img.php
```diff
@@ -29,12 +29,13 @@ public function convert( $amp_attributes = array() ) {
...
-						|| false !== strpos( $attributes['class'], 'alignright' )
-					)
-				) {
+				$img_class = isset( $attributes['class'] ) ? $attributes['class'] : '';
+				// alignleft or alignright
+				$pattern = "/alignleft|alignright/";
+				preg_match_all( $pattern, $img_class, $matches );
+				// test if there are any match
+				$is_aligned = count( $matches ) >= 1 && count( $matches[0] ) > 1;
+				if ( $is_aligned ) {
```
mjangda: I agree that the `if` isn't ideal, but there's small performance tradeoff with running `preg_match_all` every time even when the class attribute is not set.

mjangda:
Thanks for the PR @mitogh. I'm going to close this though as code readability is a bit subjective and we've made explicit choices here that I noted in comments.

mitogh:
Thank you :+1:

PR #52, opened 2015-10-15
Provide a method to allow the ID to be accessed

whyisjake:


mjangda:
PR seems fine, but just so we have context around the PR, how will the `get_id` method be used?

whyisjake:
Was using it in some of the templating that I was doing where I needed the post ID.

whyisjake:
I guess that get_the_id() would also work in this context, but it seems like using the properties/methods of the class would be helpful too.

PR #68, opened 2015-10-29
loadHTML instead of loadXML

mjangda:
Because the latter doesn't like HTML entities. Also don't fatal if the
parse fails.


danielbachhuber:
I thought DOMDocument didn't work with HTML5 tags?

mjangda:
It throws warnings, but otherwise seems to work fine.

Issue #88, opened 2015-12-24
Post content is not parsed on UTF-8


j3j5:
The dummy tags used when loading the HTML on the DOMDocument ](https://github.com/Automattic/amp-wp/blob/master/class-amp-sanitizer.php#L27) make the encoding of the document to be ignored, causing issues on UTF-8 posts. I have a fix on [my fork](https://github.com/j3j5/amp-wp/commit/c15c1e454cc66cd9ffeafed4604a8662e485e820) which loads UTF-8 by default, although I'm not sure if more encodings should be supported. I guess the best solution would be to try and read the original encoding of the page and if not, use UTF-8.

besttechie:
Thanks for the fix! Just implemented it and it works well. The encoding issues were driving me nuts.

j3j5:
You are welcome @besttechie, let's see what @mjangda thinks so I can open a PR.

mjangda:
Good catch; can you start a PR and I'll run with that to start?

j3j5:
Sure!

PR #89, opened 2015-12-24
AMP_Sanitizer

j3j5:
Load content from posts on UTF8 instead of default encoding.


j3j5:
PR to fix #88

j3j5:
Travis fails with `The command "bash bin/install-wp-tests.sh wordpress_test root '' localhost $WP_VERSION" failed and exited with 1 during .` but I don't think my commit breaks that script, any idea what could that be?

mjangda:
Thanks @j3j5! Can you share some sample text/content that this fixes?

j3j5:
Cool, I was going to reply now, anything UTF-8 using, for example `'` breaks into `â`.

PR #90, opened 2015-12-29
Add/amp anim for gif

### LINKED ISSUE #72: gifs should be wrapped in <amp-anim>
https://github.com/ampproject/amphtml/tree/master/extensions/amp-anim

j3j5:
Parse `.gif` images into `<amp-ani>` tags and add the proper script script to the scripts content array.
It should fix #72

### DIFF
--- a/class-amp-img.php
+++ b/class-amp-img.php
@@ -5,7 +5,7 @@ require_once( dirname( __FILE__ ) . '/class-amp-converter.php' );
 class AMP_Img_Converter extends AMP_Converter {
 	public static $tag = 'img';
 
-	private $anim_extensions = '/.(gif|gifv)/';
+	private $anim_extensions = '/.(gif)/';
 
 	private static $script_slug = 'amp-anim';
 	private static $script_src = 'https://cdn.ampproject.org/v0/amp-anim-0.1.js';
--- a/class-amp-img.php
+++ b/class-amp-img.php
@@ -5,6 +5,11 @@ require_once( dirname( __FILE__ ) . '/class-amp-converter.php' );
 class AMP_Img_Converter extends AMP_Converter {
 	public static $tag = 'img';
 
+	private $anim_extensions = '/.(gif|gifv)/';
+
+	private static $script_slug = 'amp-anim';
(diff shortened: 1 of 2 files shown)


class-amp-content.php
```diff
@@ -32,12 +32,15 @@ public function transform() {
...
 
+		$img_converter = new AMP_Img_Converter( $content );
 		// Convert HTML to AMP
 		// see https://github.com/ampproject/amphtml/blob/master/spec/amp-html-format.md#html-tags)
-		$content = $this->convert( new AMP_Img_Converter( $content ), array(
+		$content = $this->convert( $img_converter, array(
 			'layout' => 'responsive',
 		) );
 
+		$this->add_scripts( $img_converter->get_scripts() );
```
mjangda: The `convert` method calls `add_scripts` for us so we shouldn't need to call it here.

class-amp-img.php
```diff
@@ -5,6 +5,11 @@
 class AMP_Img_Converter extends AMP_Converter {
 	public static $tag = 'img';
 
+	private $anim_extensions = '/.(gif|gifv)/';
```
mjangda: Does `gifv` really makes sense here? (in most cases, it serves up an HTML file)

mjangda:
Thanks for this PR @j3j5! Left two comments inline.

class-amp-img.php
j3j5: Yeah, I wasn't really sure, I actually asked in [SO](https://stackoverflow.com/questions/34501955/what-are-the-supported-formats-for-amp-anim) about what file formats should `<amp-anim>` be used with, but to no avail so far. I can remove it for `.gif` only and we can add more formats if necessary later.

class-amp-content.php
j3j5: True, Idk how I looked at it I thought they weren't being added.

j3j5:
I've fixed the changes from your comments.

mjangda:
Made a few small tweaks and merged in manually. Thanks again @j3j5!

PR #91, opened 2016-01-05
img converter should handle GIF urls with query

mjangda:
Parse the URL path and use that to check instead of just checking the end of the URL.

h/t @j3j5 (see https://github.com/Automattic/amp-wp/commit/73f4cd55e17f21f6f9970770a11f77b271eb58e0#commitcomment-15264564)


j3j5:
Cool, `parse_url` it is indeed much faster than `preg_match`.

Issue #85, opened 2015-12-15
Improve Embeds


mjangda:
Right now, the implementation is pretty messy. Lots of unnecessary repeated code. Plus it's not that easy to add your own embed handlers.

- [ ] Move shortcode and embed registration into separate method (likely an `abstract` method in `AMP_Embed_Handler`).
- [ ] Make it possible to add/remove extra embed types.

Issue #81, opened 2015-12-02
Convert Vine oEmbed URL and shortcode to <amp-vine>


niallkennedy:
A Vine video is described in an AMP template as `<amp-vine vineid="">`

A Vine video may appear in WordPress content as an oEmbed-compatible URL. Vine is a supported oEmbed provider as of WordPress 4.1.

A Vine video may be specified in a shortcode as `[vine url=""]` or `[vine id=""]`

* [WordPress.com Vine docs](https://en.support.wordpress.com/videos/vine/)
* [AMP template component](https://github.com/ampproject/amphtml/blob/master/extensions/amp-vine/amp-vine.md)

mjangda:
Are there any other URL formats for vine than `https://vine.co/v/MdKjXez002d`?

niallkennedy:
A video link appears as:
https://vine.co/v/MdKjXez002d

It may have additional query parameters such as ref=email, which do not affect display and which do not need to be interpreted for <amp-vine>. There is currently no autoplay audio or postcard template option for the AMP Vine component.

PR #96, opened 2016-01-07
Add an instagram embed handler

### LINKED ISSUE #7: Convert Instagram oEmbed and Shortcode to <amp-instagram>
See https://github.com/ampproject/amphtml/blob/master/extensions/amp-instagram/amp-instagram.md

j3j5:
This should close #7

### DIFF
--- a/class-amp-instagram-embed.php
+++ b/class-amp-instagram-embed.php
@@ -1,0 +1,91 @@
+<?php
+
+require_once( dirname( __FILE__ ) . '/class-amp-embed-handler.php' );
+
+// Much of this class is borrowed from Jetpack embeds
+class AMP_Instagram_Embed_Handler extends AMP_Embed_Handler {
+	const SHORT_URL_HOST = 'instagr.am';
+	const URL_PATTERN = '#http(s?)://(www\.)?instagr(\.am|am\.com)/p/([^/]*)#i';
+	const DEFAULT_WIDTH = 600;
+	const DEFAULT_HEIGHT = 480;
+
+	private static $script_slug = 'amp-instagram';
+	private static $script_src = 'https://cdn.ampproject.org/v0/amp-instagram-0.1.js';
+
+	private $args;
+
+	function __construct( $args = array() ) {
+		$this->args = wp_parse_args( $args, array(
+			'width' => self::DEFAULT_WIDTH,
+			'height' => self::DEFAULT_HEIGHT,
+		) );
+
+		wp_embed_register_handler( 'amp-instagram', self::URL_PATTERN, array( $this, 'oembed' ), -1 );
+		add_shortcode( 'instagram', array( $this, 'shortcode' ) );
+	}
+
+	public function get_scripts() {
+		if ( ! $this->did_convert_elements ) {
+			return array();
+		}
+
+		return array( self::$script_slug => self::$script_src );
+	}
+
+	public function shortcode( $attr ) {
+		$url = false;
+
+		$instagram_id = false;
+		if ( isset( $attr[0] ) ) {
+			$url = ltrim( $attr[0] , '=' );
+		} elseif ( function_exists ( 'shortcode_new_to_old_params' ) ) {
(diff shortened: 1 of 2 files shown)


j3j5:
Before going for #81...is #85 going to be part of the v1 milestone or is this something that's going to be left for later?

j3j5:
The branch is still missing the unit tests and I have to leave now, but I'd thought I'll submit it in case there's any feedback. I'll be AFK till Monday so if somebody wants to make them in the mean time...otherwise, I'll probably push them on Monday.

class-amp-instagram-embed.php
```diff
@@ -0,0 +1,91 @@
+<?php
+
+require_once( dirname( __FILE__ ) . '/class-amp-embed-handler.php' );
+
+// Much of this class is borrowed from Jetpack embeds
+class AMP_Instagram_Embed_Handler extends AMP_Embed_Handler {
+	const SHORT_URL_HOST = 'instagr.am';
+	const URL_PATTERN = '#http(s?)://(www\.)?instagr(\.am|am\.com)/p/([^/]*)#i';
```
mjangda: We can update the last selector (`([^/]*)`) to ignore `?` as well using `([^/?]+)` to select the full id.

class-amp-instagram-embed.php
```diff
@@ -0,0 +1,91 @@
...
+		$instagram_id = $this->get_instagram_id_from_url( $url );
+
+		return $this->render( array(
+			'url' => $url,
+			'instagram_id' => $instagram_id,
+		) );
+	}
+
+	public function oembed( $matches, $attr, $url, $rawattr ) {
+		return $this->shortcode( array( $url ) );
```
mjangda: With the updated regex mentioned above, we can bypass the `shortcode` method and just call `render` directly.

class-amp-instagram-embed.php
```diff
@@ -0,0 +1,91 @@
...
+			'amp-instagram',
+			wp_parse_args( array(
+				'data-shortcode' => $args['instagram_id'],
+				'layout' => 'responsive',
+			), $this->args )
+		);
+	}
+
+	private function get_instagram_id_from_url( $url ) {
+		$parsed_url = parse_url( $url );
```
mjangda: We can just pass `PHP_URL_PATH` as the second param and just get back the path given that's the only value we need.

class-amp-instagram-embed.php
```diff
@@ -0,0 +1,91 @@
...
+		return array( self::$script_slug => self::$script_src );
+	}
+
+	public function shortcode( $attr ) {
+		$url = false;
+
+		$instagram_id = false;
+		if ( isset( $attr[0] ) ) {
+			$url = ltrim( $attr[0] , '=' );
+		} elseif ( function_exists ( 'shortcode_new_to_old_params' ) ) {
```
mjangda: The `elseif` isn't needed for this shortcode.

class-amp-instagram-embed.php
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		return array( self::$script_slug => self::$script_src );
+	}
+
+	public function shortcode( $attr ) {
+		$url = false;
+
+		$instagram_id = false;
+		if ( isset( $attr[0] ) ) {
```
mjangda: I believe the Instagram shortcode expects <a href="https://en.support.wordpress.com/instagram/instagram-images/">an `url` attribute</a>.

mjangda:
Thanks for this @j3j5; looking pretty good! Left a few notes. And unit tests would be great! I've been meaning to write some for the other embeds, but keep putting it off :)

> is #85 going to be part of the v1 milestone or is this something that's going to be left for later?

Yep, will be tackled soon. There are some changes we need to make to all the embed classes (biggest one being moving the shortcode/embed registration out of the constructor and into a method instead).

mjangda:
Thanks again @j3j5. Made a few small changes and merged!

Issue #105, opened 2016-01-20
Google AMP Pages Test


kkranzo:
I just received emails from Google this morning after they tested several of my sites that have the amp-wp plugin installed. Below are the errors that were received for pretty much all the AMP pages on my sites. So the wp-amp plugin must not be setting things up correctly.

- Error in schema.org NewsArticle
- Invalid attribute value (type=text/javascript, allowfullscreen=allowfullscreen, etc.)
- Prohibited tag (img)
- Stylesheet CDATA attribute missing or invalid (Disallowed @import in CSS)
- Incorrectly nested tag (body > script)

I also received this error on a few pages on all my sites as well

- Required attribute missing (height)

mjangda:
Are you using the latest version of the plugin from Github? Or the version currently in the WordPress plugins directory?

kkranzo:
I am using the one in the plugins directory (0.1).

rudygalfi:
@mjangda Are you planning any release updates to the plugins directory soon?

mjangda:
Yep, hopefully a release next week.

benbowler:
Perfect! I had the same issue so good to see it's been resolved

DaleeCro:
Hey,

i had the same problems in webmastertools:

Error/not permitted Atttribute:

content_max_width

layout="fill"

allowfullscreen=true

i use the plugin from github directory and have downloaded 1 week ago. When you plan the update to WordPress Directory?

Best wishes

kkranzo:
Looks like all these validation errors are still present in the Google Amp Report even after updating to version 0.2 from the repository.

HammyHavoc:
I'm also getting errors from Google's Search Console validation tests for https://www.previousmagazine.com/five-ways-to-keep-elderly-loved-ones-comfortable-at-home/amp

The errors are as follows:

- `Invalid attribute values: type=text/javascript`
- `Incorrectly nested tags: body > script`

mjangda:
@HammyHavoc Looks like this cloudflare(?) script is breaking things:

```
<body><script type="text/javascript">
//<![CDATA[
try{(function(a){var b="https://",c="www.previousmagazine.com",d="/cdn-cgi/cl/",e="img.gif",f=new a;f.src=[b,c,d,e].join("")})(Image)}catch(e){}
//]]>
</script>
```

Issue #104, opened 2016-01-20
Structured Data Testing Tool errors


RecensioniDroni:
Hi everybody,
plugin needs some fixes.

- BlogPosting

logo:	 missing and required

image: missing and required

dateModified: missing and recommended

Otto42:
Same here. The publisher.logo appears to be required. Could use the site icon if available.

The image is also required. Suggest featured image and a default. Reference issue #98

cfinke:
<blockquote>Could use the site icon if available.</blockquote>

Site icons are always (?) square, but publisher.logo is supposed to be rectangular, with a max height of 60px and max width of 600px, according to https://developers.google.com/structured-data/rich-snippets/articles. I supposed it could be 60x60.

mjangda:
> dateModified: missing and recommended

dateModified was added in dc69efff26a531a7b5fb777c812a36b5452a8667

mjangda:
Closing this as we have other issues tracking this.

Issue #98, opened 2016-01-12
Enforce `image` in json metadata


mjangda:
AMP metadata examples suggest including an `image` object. We should attempt the following order:

- [x] Featured Image
- [x] First attachment on post
- [ ] Site Icon
- [ ] Fallback image (blank or WP logo?)

Spec: https://developers.google.com/structured-data/rich-snippets/articles

Otto42:
Schema.org specs actually require the image; it's not optional. So adding in a default blank image would be a good idea as well.

mjangda:
No point in forcing an image if one is not available. This is something Webmaster Tools needs to change.

PR #106, opened 2016-01-20
Add the required 'image' attribute to the Article.

cfinke:
According to https://developers.google.com/structured-data/rich-snippets/articles, the 'image' property is required, so we should do our best to find an image associated with the post.

### DIFF
--- a/class-amp-post.php
+++ b/class-amp-post.php
@@ -71,6 +71,34 @@ class AMP_Post {
 				'name' => get_bloginfo( 'name' ),
 			),
 		);
+		
+		$post_thumbnail_id = false;
+
+		// Include a reference to either a featured image or the first attached image.
+		if ( has_post_thumbnail( $this->ID ) ) {
+			$post_thumbnail_id = get_post_thumbnail_id( $this->ID );
+		}
+		else {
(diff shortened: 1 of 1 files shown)


class-amp-post.php
```diff
@@ -71,6 +71,34 @@ private function build_metadata() {
...
 		);
+		
+		$post_thumbnail_id = false;
+
+		// Include a reference to either a featured image or the first attached image.
+		if ( has_post_thumbnail( $this->ID ) ) {
+			$post_thumbnail_id = get_post_thumbnail_id( $this->ID );
+		}
+		else {
+			$attached_media = get_attached_media( 'image', $this->ID );
```
mjangda: We can optimize a bit with a direct `get_posts` lookup here:

```
$attached_image_ids = get_posts( array(
    'post_parent' => $this->ID,
    'post_type' => 'attachment',
    'post_mime_type' => 'image',
    'posts_per_page' => 1,
    'orderby' => 'menu_order',
    'order' => 'ASC',
    'fields' => 'ids',
) );

if ( ! empty( $attached_image_ids ) ) {
    $post_thumbnail_id = array_shift( $attached_image_ids );
}
```

mjangda:
Related #98 #104 #105

PR #107, opened 2016-01-21
Reactivate Photon

### LINKED ISSUE #102: Restore Photon
Previously disabled in #32, we should bring Photon back because it's cool.

kraftbj:
Photon in JP 3.8.2 resolved any issues with AMP previously. I'd suggest we just activate it. I've had second-thoughts on the version check—this is in dev and not released, I think it is fine to assume current version of other plugins, at least at time of release.

Fixes #102


mjangda:
:+1:

mjangda:
I think tests are stuck. Merging.

PR #120, opened 2016-01-24
Bugfix for iframe / blacklist sanitizer

kantan2015:
Hi, these are  bug fixes for iframe / blacklist sanitizer.
Please review and marge.

**fix for iframe sanitizer**
There is a bug when several iframe tags in p tag.
`<p><iframe></iframe><iframe></iframe></p>`
In this case, as it is replacing parent p tag with the new amp-iframe tag,  it will crash by referring invalid node in the next loop. 
I changed to append amp-iframe after p tag, and remove p tag only when there is nothing in it.

**fix for blocklist sanitizer**
Currently the only first blacklist attribute is removed as it is removing the attribute in foreach loop.
 This is a minor as the blacklist attribute is only "style" for now.

Regards


mjangda:
Continued at #122

Issue #100, opened 2016-01-13
Whitelisted values for "rel" attribute on <a> tags


mjangda:
AMP has a whitelist of values that can be used for the `rel` attribute on `<a>` tags: https://github.com/ampproject/amphtml/issues/596#issuecomment-147803480

WP adds the following to linked attachments which throws validation errors: `rel="attachment wp-att-1686"`

mjangda:
Quick fix might be to just strip out `rel` values for `a` tags that are `attachment` or `wp-att-(\d)+`.

mjangda:
I think stripping that WP-specific `rel` values is good enough for now.

PR #151, opened 2016-01-29
Strip WordPress's `wp-att-` rel from links

mjangda:
AMP validator does not like it very much.

See #100


jattsaini:
How to use fix with amp plugin?

mjangda:
You can download the latest development version from here: https://github.com/Automattic/amp-wp/archive/master.zip

jattsaini:
Have just replaced the latest files available from that link in the plugin.

PR #167, opened 2016-02-02
Move `style` tag out of style file

mjangda:
To ensure wider compatibility. If you replace the stylesheet, you shouldn't have to add the action. This way, plugins and such can still work nicely with folks with custom stylesheets.


mjangda:
Failures are from svn issues in trunk.

PR #178, opened 2016-02-10
Adding backwards compatibility for older versions of WordPress

bcampeau:
Upon testing this plugin on version 4.1.6, I discovered that two functions were in use that caused fatal errors:

`get_site_icon_url`: only exists since 4.3
`get_avatar_url`: only exists since 4.2

The site icon already has a nice SVG default and I've made it filterable to allow you to provide your own in absence of the core function.

There isn't really a great fallback for the avatar URL, but simply removing it does not significantly degrade the experience.

Both of these seem like small compromises to allow wider usage of AMP on older installs.

### DIFF
--- a/templates/single.php
+++ b/templates/single.php
@@ -16,7 +16,7 @@
 <nav class="title-bar">
 	<div>
 		<a href="<?php echo esc_url( $this->get( 'home_url' ) ); ?>">
-			<?php $site_icon_url = $this->get( 'site_icon_url' ); ?>
+			<?php $site_icon_url = apply_filters( 'amp_site_icon_url', $this->get( 'site_icon_url' ) ); ?>
 			<?php if ( $site_icon_url ) : ?>
 				<amp-img src="<?php echo esc_url( $site_icon_url ); ?>" width="32" height="32" class="site-icon"></amp-img>
 			<?php else : ?>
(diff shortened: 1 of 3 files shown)


mjangda:
As mentioned in https://github.com/Automattic/amp-wp/pull/175#issuecomment-181575935, the fix is easy, yes, but it's going to be a pain to keep maintaining compatibility for older versions moving forward. Any reason you can't just add shims for the missing functions?

In any case, probably okay to merge but I think it'd be better to keep the min requirement to a newer version like 4.3/4.4 so we're not committing to supporting those older versions.

The site_icon filter does seem like a good addition, though.

templates/single.php
```diff
@@ -16,7 +16,7 @@
 <nav class="title-bar">
 	<div>
 		<a href="<?php echo esc_url( $this->get( 'home_url' ) ); ?>">
-			<?php $site_icon_url = $this->get( 'site_icon_url' ); ?>
+			<?php $site_icon_url = apply_filters( 'amp_site_icon_url', $this->get( 'site_icon_url' ) ); ?>
```
mjangda: This filter should probably fire in the `AMP_Post_Template` class?

bcampeau:
I can't see anything else in here that would break in older versions, so I'd probably advocate for just merging it. I'm fine with keeping the advertised minimum version higher though.

templates/single.php
bcampeau: Fair. I'll move this now.

bcampeau:
Filter has been moved.

PR #191, opened 2016-02-17
WIP: getimagesize

mjangda:
Calculate dimensions for remote images by fetching them.

TODOs:

- [x] Disable extraction for sanitizer tests
- [x] Implement fastimage class
- [x] Add caching
- [x] Add tests
- [ ] <strike>Add fallback for images that end up with no dimensions (force to 1x1?)</strike>

See #14

### DIFF
--- a/includes/lib/class-fastimage.php
+++ b/includes/lib/class-fastimage.php
@@ -1,0 +1,243 @@
+<?php
+
+/**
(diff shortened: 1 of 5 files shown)


mjangda:
Asked about a fallback in https://github.com/ampproject/amphtml/issues/2039#issuecomment-185429721 -- will follow-up in separate PR based on response.

Issue #174, opened 2016-02-05
Branding PR


tlovett1:
Hey All,

10up is seeking an AMP plugin solution for clients. We love the simplicity and extensibility we are seeing in this project so far. However, we want whatever we recommend to clients and use to be branding agnostic and powered by the customizer so clients aren't forced into making customizations.

Would Automattic be open to PRs that does the following? If so, 10up would love to contribute to the project.
* Remove default WordPress.com styles and branding.
* Add customizer functionality for choosing basic site colors and logo for AMP views.

mjangda:
> Remove default WordPress.com styles and branding.

Sure, open to suggestions on how we could make it more generic. We mainly wanted something in place that looks good across most blogs. We will likely end up removing the WordPress logo since that seems to be a pretty contentious thing :)

> Add customizer functionality for choosing basic site colors and logo for AMP views.

That'd be pretty great! We've been trying to avoid adding extra options and pulling from existing functionality wherever possible (e.g. Site Icon) but will likely need a few additional options. It'd be good to discuss specifically what options we're adding to make sure we have the right balance before a PR.

Let's split this off into two separate issues to discuss?

tlovett1:
Will do! Thanks Mo.

mjangda:
:raised_hands:

Issue #83, opened 2015-12-10
Customize the navbar at the top


justinferrell:
Replace the SVG logo with the site icon if one is available and allow the color of the bar to be specify in the appearance customizer.
- [x] Show <a href="https://make.wordpress.org/core/2015/07/27/site-icon/">Site Icon</a>, if set (instead of WP logo)
- [ ] Show <a href="https://codex.wordpress.org/get_header_image">header image</a>, if set

mjangda:
Plus, header image if set.

westonruter:
This is obsolete now that the plugin encourages use of AMP with your active theme's templates (as of 1.0).

PR #199, opened 2016-02-18
Add AMP template/style editor using the Customizer

DrewAPicture:
AMP Navigation Bar section in the Customizer

Following up on #83 and #174, I propose that we introduce an interface for customizing AMP templates.

In this PR I've added a new Customizer panel that will only be loaded via a submenu link under Appearance (or when any AMP-generated page is loaded in the preview pane). It implements three controls under a "Navigation Bar" section:

1. Header Text Color – text color in the navbar
2. Header Background Color – background color in the navbar
3. Site Icon – site icon in the navbar

The code is structured in such a way that if we later decide to add more sections and controls, that's fairly straightforward. We could easily see further extending this new panel for different types of templates, including post types or pages, that might be autodiscovered and added as new sections/controls or something.

This PR also indirectly addresses #187, by hiding/showing the site icon image tags rather than outputting nothing in the event no icon is set – we need the tag to be in place to facilitate swapping out images sources and still getting AMP-ready markup in the preview.

Finally, after some finagling, the AMP panel in the Customizer supports the postMessage transport, allowing users to live preview changes. Works great.

TODO: Readme updates if any, 10up attribution.

Thanks for your consideration, I look forward to your feedback.


DrewAPicture:
With no site icon:

No site icon

DrewAPicture:
Updated from master including the new class names. I also opted to prefix the new class for consistency.

octalmage:
I spent some time testing this and it seems to work well.

templates/style.php
```diff
@@ -138,17 +138,17 @@
...
 
 nav.amp-wp-title-bar div {
 	line-height: 54px;
-	color: #fff;
+	color: <?php echo get_theme_mod( 'amp_navbar_color' ); ?>;
 }
 
 nav.amp-wp-title-bar a {
-	color: #fff;
+	color: <?php echo get_theme_mod( 'amp_navbar_color' ); ?>;
```
mjangda: We should probably pass these through template data instead (to allow for more easier filtering). Also, I'm not sure if `get_theme_mod` already sanitizes the values, but if not, we should probably escape them on output.

includes/admin/amp-backend-actions.php
```diff
@@ -0,0 +1,46 @@
...
+/** AMP_Template_Customizer class */
+require_once( AMP__DIR__ . '/includes/admin/class-amp-customizer.php' );
+
+/**
+ * Instantiates the AMP template editor for the Customizer.
+ *
+ * @param WP_Customize_Manager $wp_customize Customizer instance.
+ */
+function init_amp_template_customizer( $wp_customize ) {
+	AMP_Template_Customizer::init( $wp_customize );
```
mjangda: It'd be nice to have a filter to not enable the AMP Customizer since it won't be relevant for all sites.

includes/admin/amp-backend-actions.php
```diff
@@ -0,0 +1,46 @@
...
+add_action( 'customize_register', 'init_amp_template_customizer', 500 );
+
+/**
+ * Registers a submenu page to access the AMP template editor panel in the Customizer.
+ */
+function amp_customizer_editor_link() {
+	$post_id = get_posts( array(
+		'post_status'      => 'publish',
+		'post_type'        => 'post',
+		'orderby'          => 'rand',
```
mjangda: I think we should avoid the `rand` and just pick the last published post.

includes/admin/amp-backend-actions.php
```diff
@@ -0,0 +1,46 @@
...
+		'posts_per_page'   => 1,
+		'fields'           => 'ids',
+		'suppress_filters' => false
+	) );
+
+	// Teensy little hack on menu_slug, but it works. No redirect!
+	$menu_slug = add_query_arg( array(
+		'autofocus[panel]' => 'amp_template_editor',
+		'url'              => rawurlencode( amp_get_permalink( $post_id ) ),
```
mjangda: Two things here
- We should make sure the post supports AMP before using it.
- Bit of an edge case but probably worth handling the case where we don't have any posts?

mjangda:
Overall, this is awesome. I love how it works and feels. And the code, unsurprisingly, is solid :)

A big concern is forward compatibility because of all the hackery we're doing. It's most definitely going to break sites doing specialized things with the customizer and high likelihood that it could break with future core updates?

Can you remind me why we need to do the drop/restore dance with the core's Customizer panels?

rubenwoudsma:
Great stuff this. Why not use the styling options as Yoast has built this into the [Glue for Yoast SEO & AMP plugin](https://github.com/Yoast/yoastseo-amp). I think the way Yoast implemented it, is good. This should be part of the WordPress AMP plugin and not of the Glue-plugin.

templates/style.php
DrewAPicture: > We should probably pass these through template data instead (to allow for more easier filtering)
Not sure I'm following. Customizer settings are stored in the options table. Can you clarify what you're talking about?

> Also, I'm not sure if get_theme_mod already sanitizes the values, but if not, we should probably escape them on output.
The value is sanitized on save via the sanitization callback. In the case of the hex codes, that's `sanitize_hex_color()`. We could probably escape, though it's worth mentioning that even the default themes in core don't both escaping the values on output.

includes/admin/amp-backend-actions.php
DrewAPicture: This is covered in 257146b12, where the logic for dropping core panels was reworked to include a filter.

includes/admin/amp-backend-actions.php
DrewAPicture: This is covered in 0af694acf where I've removed `rand` and made the post type filterable.

includes/admin/amp-backend-actions.php
DrewAPicture: The first point is also covered in 0af694acf where we bail if the post type doesn't support AMP.

DrewAPicture:
@mjangda I've addressed all of your above code feedback inline.

> A big concern is forward compatibility because of all the hackery we're doing. It's most definitely going to break sites doing specialized things with the customizer and high likelihood that it could break with future core updates?

I would disagree with an expectation of breaking sites doing specialized things with the Customizer; Everything we're doing is above board and falls in-line with how core intended the Customizer to operate. In our case, we're doing something very specific that falls within the confines of intended behavior of the Customizer APIs.

> Can you remind me why we need to do the drop/restore dance with the core's Customizer panels?

The "hackery" in terms of dropping the core panels is essentially the only way to correctly accomplish what we're doing – that is to say dropping all of the registered controls and panels and replacing them with our own to provide a "standalone" customizer instance effect.

The `customize_loaded_components` filter is intentionally evaluated by core in code that runs on 'plugins_loaded' at priority 10. This unfortunately means that we have to hook in earlier than priority 10 to disable the core panels properly. This even if we're doing a bunch of operations on hooks that fire much later in the loading order. The best way I've found to accomplish this "standalone" customizer instance effect is to default to disabling the panels early, then adding them back if the user isn't attempting to customize an AMP preview. 

Just as there's a very specific hook for disabling the admin toolbar, the same goes for unhooking core panels. It's not great, but it's the only official option core offers.

jdevalk:
If this is merged we'll happily work on top of it to move the stuff we have in our Glue plugin to this plugin. This is where it should be.

DrewAPicture:
@mjangda e467d7da8 switches to running Customizer settings through template data. See the changes to the `$this->data` array in `AMP_Post_Template` for more.

DrewAPicture:
@mjangda What's the status here?

mjangda:
This was moved to #406

PR #202, opened 2016-02-18
Removing 'on' attribute from blacklist for image and button tags.

sevatt:
Removing 'on' attribute from blacklist for image and button tags.
Needed to get example in amp-lightbox to work.
https://www.ampproject.org/docs/reference/extended/amp-lightbox.html

### DIFF
--- a/includes/sanitizers/class-amp-blacklist-sanitizer.php
+++ b/includes/sanitizers/class-amp-blacklist-sanitizer.php
@@ -39,7 +39,14 @@ class AMP_Blacklist_Sanitizer extends AMP_Base_Sanitizer {
 
 				// on* attributes (like onclick) are a special case
 				if ( 0 === stripos( $attribute_name, 'on' ) ) {
-					$node->removeAttribute( $attribute_name );
+					switch ($node->tagName) {
+						case 'button':
+						case 'img':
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-blacklist-sanitizer.php
```diff
@@ -39,7 +39,14 @@ private function strip_attributes_recursive( $node, $bad_attributes, $bad_protoc
 
 				// on* attributes (like onclick) are a special case
 				if ( 0 === stripos( $attribute_name, 'on' ) ) {
```
mjangda: It would be better to change the check to skip when we explicitly match an attribute_name of `on` since attributes like `onclick` are not allowed:

> Attribute names starting with on (such as onclick or onmouseover) are disallowed in AMP HTML. The attribute with the literal name on (no suffix) is allowed.

via https://www.ampproject.org/docs/reference/spec.html

sevatt: I see what you are saying.  I'll make that change.

Thanks

Steven

sevatt:
Agree the problem was approached the wrong way.  Reworking solution.

PR #224, opened 2016-02-21
Remove datetime attribute from del tag

kantan2015:
datetime attribute in del tag cause validation error.


mjangda:
`datetime` is an allowed element for `del` so this might be a bug in the AMP validator: https://www.w3.org/TR/html5/edits.html#attr-mod-datetime

kantan2015:
True. Latest validator does not cause the error.

PR #226, opened 2016-02-22
Update `remove_filter()` call to match its `add_filter()` call

ethitter:
Currently this `remove_filter()` instance doesn't work due to the mismatch. If `fopen()` doesn't enable the `https` wrapper, warnings result because the unsupported image extraction still runs:

> PHP Warning:  fopen(): https:// wrapper is disabled in the server configuration by allow_url_fopen=0 in /.../wp-content/plugins/amp/includes/lib/class-fastimage.php on line 31

See https://github.com/ethitter/amp-wp/blob/2de5eb39b92bfc43b342ce87d278beb1784bed45/includes/utils/class-amp-image-dimension-extractor.php#L51.


mjangda:
Good catch!

PR #231, opened 2016-02-23
Blacklists rev attribute

kraftbj:
Per the validator, `rev` is not a valid attribute. I don't see reference to this in the docs. As seen by using Jetpack's Markdown to create a footnote, which creates a link like `<a href="#fnref-3468-1" rev="footnote">↩</a>`

### DIFF
--- a/includes/sanitizers/class-amp-blacklist-sanitizer.php
+++ b/includes/sanitizers/class-amp-blacklist-sanitizer.php
@@ -55,7 +55,9 @@ class AMP_Blacklist_Sanitizer extends AMP_Base_Sanitizer {
 					} elseif ( $old_value !== $new_value ) {
 						$node->setAttribute( $attribute_name, $new_value );
 					}
-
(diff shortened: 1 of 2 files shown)


mjangda:
Aside: should Jetpack switch to using `rel=footnote` instead?

mjangda:
Also, is the `rev` only used with `a` tags? If so, might be better to only look for and strip from those tags, instead of all.

kraftbj:
Looking closer, `rev` is only on `a` and now obsolete. HTML5 removed it from the spec. I'll dig in on the Jetpack side to determine the best attribute to use (I assume `rel`).

kraftbj:
@mjangda What do you think about closing this since, technically, the problem is invalid HTML being provided by the site, not sanitizing valid HTML into AMP?

kraftbj:
Via discussion on Slack, even if corrected in Jetpack, Jetpack's markdown would have saved the older HTML4 to the database until the next time the post is saved. It's would be better of us to remove it.

mjangda:
:boom:

PR #232, opened 2016-02-23
Invalid amphtml by using plain permalink structure

rakuishi:
I was faced with an invalid amp url by using plain permalink structure. In this case, `amphtml` is like this:

```
<link rel="amphtml" href="http://192.168.99.100?amp=123" />
```

This url just redirect to homepage.

I fixed it by appending `?amp=1` behind it's permalink.

```
<link rel="amphtml" href="http://192.168.99.100/?p=123&#038;amp=1" />
```


mjangda:
Good catch! Thank you!

PR #246, opened 2016-02-24
Jetpack Stats

mjangda:
Needs testing.


kraftbj:
Minor variable issue, but otherwise looks good. Tested logged out and logged in, confirmed that the stats pixel loads while respecting Jetpack's setting for which roles are counted. Viewed wp.com/stats to confirm they're being recorded for the proper post. :shipit:

PR #256, opened 2016-02-24
Use https:// URL for amp-iframe src instead of protocol relative URL

mister-ben:
amp-iframe rejects a src with a protocol relative URL: `Invalid <amp-iframe> src. Must start with https://.`

This PR uses an https:// URL where the original src is protocol relative.

### DIFF
--- a/includes/sanitizers/class-amp-iframe-sanitizer.php
+++ b/includes/sanitizers/class-amp-iframe-sanitizer.php
@@ -91,7 +91,7 @@ class AMP_Iframe_Sanitizer extends AMP_Base_Sanitizer {
 					break;
 
 				case 'src':
-					$out[ $name ] = preg_replace('/^\/\//', 'https://', $value);
+					$out[ $name ] = set_url_scheme( $value, 'https' );
 					break;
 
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -91,6 +90,10 @@ private function filter_attributes( $attributes ) {
 					$out[ $name ] = $value;
 					break;
 
+				case 'src':
+					$out[ $name ] = preg_replace('/^\/\//', 'https://', $value);
```
mjangda: Can we use `set_url_scheme` instead here?

mister-ben: OK, changed

mjangda:
Nice, thanks!

joshlevinson:
@mjangda Thanks for the quick attention. Note that this is breaking Youtube oEmbeds on non-SSL VIP sites that use AMP.

mjangda:
@joshlevinson This was merged in to WP.com yesterday.

PR #268, opened 2016-02-25
fixed typo

Pcosta88:
Quote marks made required into a code snippet, which seemed odd. I removed them.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -28,7 +28,7 @@ If you're using an off-the-shelf theme (like from the WordPress.org Theme Direct
 
 If you're using a custom theme:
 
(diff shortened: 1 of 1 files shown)


mjangda:
That's intentional :) The point is to put your code snippets in a file that is loaded from `functions.php` via a `require` call. Open to suggestions on rewording if that's not clear.

Pcosta88:
Haha! :-)
Maybe something like, (or via a `require` call to files that load from `functions.php`).

Not sure if you want me to close this PR request or open a new with a new suggestion.

mjangda:
Your suggestion is much better! :) You can just update this PR or create a new one; either is fine.

PR #280, opened 2016-02-28
Master

oqm4:
style.php missing <style> … </style> tags, which caused output errors
on the front end.


studiohypersetadmin:
style.php missing < style > ... < / style > tags, which caused output errors on the front end.

mjangda:
Thanks for the PR. However, this is intentional. The `<style>` tag is added in `single.php`.

studiohypersetadmin:
When we launched the latest version of the plugin, with a custom template, we saw the following:

pasted image at 2016_02_26 12_47 pm

mjangda:
Yes, sorry about that. The change was necessary to allow for easier integrations with plugins.

We noted this in the "Upgrade Notice" section for 0.3 (https://github.com/Automattic/amp-wp/blob/master/readme.txt#L74) which I know can be easy to miss so we'll try to be careful with future breaking changes.

PR #281, opened 2016-02-28
Clarify custom post type instructions

kevinlisota:
The previous documentation referred to using "custom post_type slug". A custom post type has both a post_type and a rewrite slug, and both are sometimes unique values. The function `add_post_type_support` only works with the post_type parameter used in `register_post_type`. The inclusion of "slug" can lead to folks (like me) trying to use this function with the rewrite slug, which doesn't work.


mjangda:
Thanks!

PR #297, opened 2016-03-01
Modifying class-amp-blacklist-sanitizer to handle tag specific attrib…

sevatt:
Modifying class-amp-blacklist-sanitizer to strip tag specific attributes. The removed attributes are moved to the class of the tag in the form attribute-name_value.

The function get_blacklisted_attributes_for_tags()  is designed to allow for a different set of attributes to be stripped per tag.  For the first iteration, it is using at the same set of attributes for each of the tags.


mjangda:
Closing as we've moved to a whitelist-based approach #600

PR #305, opened 2016-03-03
Make sure image exists before trying to fetch via fastimage

ChrisWiegman:
Due to the fastimage implementation, if it is used PHP will throw a warning if the image doesn't exist. As it is a fallback anyway this takes the simpler approach of verifying the file exists by simply checking the header (much easier than a full pull) before attempting to open it with fastimage.


mjangda:
This would end up making two remote requests for valid images, which isn't great. If the goal is to get rid of the warning, it'd be better just to suppress it with `@`.

PR #307, opened 2016-03-04
Update typography to match Calypso.

mattmiklic:
This PR makes two changes:

- Replaces the Google Fonts-hosted version of Merriweather with the one we host on WP.com (related: https://github.com/Automattic/wp-calypso/pull/2766)
- Replaces Open Sans with system UI fonts for interface text, and removes the link to Open Sans on Google Fonts. (related: https://github.com/Automattic/wp-calypso/pull/3016)

Before:
screen shot 2016-03-03 at 8 32 22 pm

After:
screen shot 2016-03-03 at 8 31 09 pm

### DIFF
--- a/templates/single.php
+++ b/templates/single.php
@@ -3,7 +3,7 @@
 <head>
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=1,user-scalable=no">
-	<link href="https://fonts.googleapis.com/css?family=Merriweather:400,400italic,700,700italic|Open+Sans:400,700,400italic,700italic" rel="stylesheet" type="text/css">
+	<link href="https://s1.wp.com/i/fonts/merriweather/merriweather.css?v=20160212" rel="stylesheet" type="text/css">
 	<?php do_action( 'amp_post_template_head', $this ); ?>
 
 	<style amp-custom>
--- a/templates/single.php
+++ b/templates/single.php
@@ -3,7 +3,6 @@
 <head>
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=1,user-scalable=no">
-	<link href="https://s1.wp.com/i/fonts/merriweather/merriweather.css?v=20160212" rel="stylesheet" type="text/css">
 	<?php do_action( 'amp_post_template_head', $this ); ?>
 
 	<style amp-custom>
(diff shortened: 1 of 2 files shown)


templates/single.php
```diff
@@ -3,7 +3,7 @@
 <head>
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=1,user-scalable=no">
-	<link href="https://fonts.googleapis.com/css?family=Merriweather:400,400italic,700,700italic|Open+Sans:400,700,400italic,700italic" rel="stylesheet" type="text/css">
+	<link href="https://s1.wp.com/i/fonts/merriweather/merriweather.css?v=20160212" rel="stylesheet" type="text/css">
```
mjangda: Currently linked fonts can only be hosted at `fast.fonts.net` or `fonts.googleapis.com` so we'll have to use `@font-face` instead.

https://github.com/ampproject/amphtml/blob/master/spec/amp-html-format.md#custom-fonts

templates/style.php
```diff
@@ -1,3 +1,5 @@
+@import "https://s1.wp.com/i/fonts/merriweather/merriweather.css?v=20160212";
```
mattmiklic: Thanks for looking @mjangda -- can we link to the fonts using an `@import` like this? If not, I can include the actual `@font-face` rules from merriweather.css here.

mjangda: Sadly `@import` is a no-go with AMP too:

```
AMP validation had errors.
CSS syntax error in tag 'style amp-custom' - saw invalid at rule '@import'.
```

templates/style.php
```diff
@@ -1,3 +1,44 @@
...
+@font-face {
+	font-family:'Merriweather';
+	src:url('https://s1.wp.com/i/fonts/merriweather/merriweather-bolditalic-webfont.woff2') format('woff2'),
+		url('https://s1.wp.com/i/fonts/merriweather/merriweather-bolditalic-webfont.woff') format('woff'),
+		url('https://s1.wp.com/i/fonts/merriweather/merriweather-bolditalic-webfont.ttf') format('truetype'),
+		url('https://s1.wp.com/i/fonts/merriweather/merriweather-bolditalic-webfont.svg#merriweatherbold_italic') format('svg');
+	font-weight:700;
+	font-style:italic;
+}
+
```
mattmiklic: Thanks again @mjangda -- I've included all the `@font-face` rules from merriweather.css, except for the EOT font included for older versions of Internet Explorer, which I assumed that AMP pages probably don't need. Will this work?

mjangda:
Beautiful! Thanks @mattmiklic!

PR #310, opened 2016-03-04
Handle additional validation errors

bcampeau:
- Removes `<a>` tags that have invalid href attributes, but preserves child elements.
- Removes `<font>` tags, which aren't allowed, but preserves child elements.
- Slightly improved handling of `target` attributes which must be lowercase for validation.
- Removes embeds during sanitization that weren't converted during embed handling.

### DIFF
--- a/includes/sanitizers/class-amp-audio-sanitizer.php
+++ b/includes/sanitizers/class-amp-audio-sanitizer.php
@@ -31,15 +31,28 @@ class AMP_Audio_Sanitizer extends AMP_Base_Sanitizer {
 
 			$new_node = AMP_DOM_Utils::create_node( $this->dom, 'amp-audio', $new_attributes );
 
-			// TODO: limit child nodes too (only allowed: `source`; move rest to div+fallback)
 			// TODO: `source` does not have closing tag, and DOMDocument doesn't handle it well.
-			// TODO: add a filter to kill the audio if the src or any children are not https
 			foreach ( $node->childNodes as $child_node ) {
 				$new_child_node = $child_node->cloneNode( true );
-				$new_node->appendChild( $new_child_node );
+				$old_child_attributes = AMP_DOM_Utils::get_node_attributes_as_assoc_array( $new_child_node );
+				$new_child_attributes = $this->filter_attributes( $old_child_attributes );
+
+				// Only append source tags with a valid src attribute
+				if ( ! empty( $new_child_attributes['src'] ) && 'source' === $new_child_node->tagName ) {
+					$new_node->appendChild( $new_child_node );
+				}
 			}
 
-			$node->parentNode->replaceChild( $new_node, $node );
+			// If the node has at least one valid source, replace the old node with it.
+			// Otherwise, just remove the node.
+			//
+			// TODO: Add a fallback handler.
+			// See: https://github.com/ampproject/amphtml/issues/2261
+			if ( 0 === $new_node->childNodes->length && empty( $new_attributes['src'] ) ) {
+				$node->parentNode->removeChild( $node );
+			} else {
+				$node->parentNode->replaceChild( $new_node, $node );
+			}
 
 			$this->did_convert_elements = true;
 		}
@@ -51,6 +64,8 @@ class AMP_Audio_Sanitizer extends AMP_Base_Sanitizer {
 		foreach ( $attributes as $name => $value ) {
 			switch ( $name ) {
 				case 'src':
+					$out[ $name ] = $this->validate_src_attribute( $value );
+					break;
 				case 'width':
 				case 'height':
 				case 'class':
--- a/includes/sanitizers/class-amp-audio-sanitizer.php
+++ b/includes/sanitizers/class-amp-audio-sanitizer.php
@@ -64,7 +64,7 @@ class AMP_Audio_Sanitizer extends AMP_Base_Sanitizer {
 		foreach ( $attributes as $name => $value ) {
 			switch ( $name ) {
 				case 'src':
-					$out[ $name ] = $this->validate_src_attribute( $value );
+					$out[ $name ] = $this->maybe_enforce_https_src( $value );
 					break;
 				case 'width':
 				case 'height':
--- a/includes/sanitizers/class-amp-audio-sanitizer.php
+++ b/includes/sanitizers/class-amp-audio-sanitizer.php
@@ -33,6 +33,7 @@ class AMP_Audio_Sanitizer extends AMP_Base_Sanitizer {
 
 			// TODO: limit child nodes too (only allowed: `source`; move rest to div+fallback)
 			// TODO: `source` does not have closing tag, and DOMDocument doesn't handle it well.
+			// TODO: add a filter to kill the audio if the src or any children are not https
 			foreach ( $node->childNodes as $child_node ) {
 				$new_child_node = $child_node->cloneNode( true );
 				$new_node->appendChild( $new_child_node );
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -63,7 +63,7 @@ abstract class AMP_Base_Sanitizer {
 	 * @param boolean $force_https
 	 * @return string
 	 */
-	public function validate_src_attribute( $src, $force_https = false ) {
+	public function maybe_enforce_https_src( $src, $force_https = false ) {
 		$protocol = strtok( $src, ':' );
 		if ( 'https' !== $protocol ) {
 			// Check if https is required
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -67,11 +67,11 @@ abstract class AMP_Base_Sanitizer {
 		$protocol = strtok( $src, ':' );
 		if ( 'https' !== $protocol ) {
 			// Check if https is required
-			$https_required = apply_filters( 'amp_require_https_src', false );
-			if ( true === $https_required ) {
+			if ( isset( $this->args['require_https_src'] ) && true === $this->args['require_https_src'] ) {
 				// Remove the src. Let the implementing class decide what do from here.
 				$src = '';
-			} elseif ( false === $https_required && true === $force_https ) {
+			} elseif ( ( ! isset( $this->args['require_https_src'] ) || false === $this->args['require_https_src'] )
+				&& true === $force_https ) {
 				// Don't remove the src, but force https instead
 				$src = set_url_scheme( $src, 'https' );
 			}
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -54,4 +54,29 @@ abstract class AMP_Base_Sanitizer {
 			$attributes[ $key ] = $value;
 		}
 	}
+
+	/**
+	 * Decide if we should remove a src attribute if https is required.
+	 * If not required, the implementing class may want to try and force https instead.
+	 *
+	 * @param string $src
+	 * @param boolean $force_https
+	 * @return string
+	 */
+	public function validate_src_attribute( $src, $force_https = false ) {
+		$protocol = strtok( $src, ':' );
+		if ( 'https' !== $protocol ) {
+			// Check if https is required
+			$https_required = apply_filters( 'amp_require_https_src', false );
+			if ( true === $https_required ) {
+				// Remove the src. Let the implementing class decide what do from here.
+				$src = '';
+			} elseif ( false === $https_required && true === $force_https ) {
+				// Don't remove the src, but force https instead
+				$src = set_url_scheme( $src, 'https' );
+			}
+		}
+
+		return $src;
+	}
 }
--- a/includes/sanitizers/class-amp-blacklist-sanitizer.php
+++ b/includes/sanitizers/class-amp-blacklist-sanitizer.php
@@ -106,13 +106,13 @@ class AMP_Blacklist_Sanitizer extends AMP_Base_Sanitizer {
 	}
 
 	private function get_blacklisted_protocols() {
-		return array(
+		return apply_filters( 'amp_blacklisted_protocols', array(
 			'javascript',
-		);
+		) );
 	}
 
 	private function get_blacklisted_tags() {
-		return array(
+		return apply_filters( 'amp_blacklisted_tags', array(
 			'script',
 			'noscript',
 			'style',
@@ -141,14 +141,14 @@ class AMP_Blacklist_Sanitizer extends AMP_Base_Sanitizer {
 			//'video',
 			//'audio',
 			//'iframe',
-		);
+		) );
 	}
 
 	private function get_blacklisted_attributes() {
-		return array(
+		return apply_filters( 'amp_blacklisted_attributes', array(
 			'style',
 			'srcset',
 			'size',
-		);
+		) );
 	}
 }
--- a/includes/sanitizers/class-amp-blacklist-sanitizer.php
+++ b/includes/sanitizers/class-amp-blacklist-sanitizer.php
@@ -11,6 +11,12 @@ require_once( AMP__DIR__ . '/includes/sanitizers/class-amp-base-sanitizer.php' )
 class AMP_Blacklist_Sanitizer extends AMP_Base_Sanitizer {
 	const PATTERN_REL_WP_ATTACHMENT = '#wp-att-([\d]+)#';
 
+	protected $DEFAULT_ARGS = array(
+		'add_blacklisted_protocols' => array(),
+		'add_blacklisted_tags' => array(),
+		'add_blacklisted_attributes' => array(),
+	);
+
 	public function sanitize() {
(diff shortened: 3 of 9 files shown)


tests/test-amp-blacklist-sanitizer.php
```diff
@@ -102,6 +107,26 @@ public function get_data() {
 				'<a href="http://example.com" target="boom">Link</a>',
 				'<a href="http://example.com">Link</a>',
 			),
+
+			'a_with_href_mailto' => array(
+				'<a href="mailto:email@domain.com">Link</a>',
```
mjangda: `mailto` should be okay (it's included in an example here: https://github.com/ampproject/amphtml/blob/eb641b052d44f78b8b83fc91bc64394b7263ee5e/examples/article.amp.html#L203)

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -42,6 +42,8 @@ public function sanitize() {
 				continue;
 			}
 
+			// TODO: add a filter to kill the iframe if not https
```
mjangda: For this, it'd be nice to use a `fallback` element instead that links out to the embed so it's still accessible. I've opened https://github.com/ampproject/amphtml/issues/2433 to see if it's supported for `amp-iframe`.

includes/sanitizers/class-amp-blacklist-sanitizer.php
```diff
@@ -93,24 +104,52 @@ private function sanitize_a_attribute( $node, $attribute ) {
...
+				$node->setAttribute( $attribute_name, '_blank' );
+			} else {
+				// only _blank is allowed
+				$node->removeAttribute( $attribute_name );
+			}
+		} elseif ( 'href' === $attribute_name ) {
+			$valid_protocols = array( 'http', 'https' );
+			$protocol = strtok( $attribute->value, ':' );
+			if ( false === filter_var( $attribute->value, FILTER_VALIDATE_URL )
+				|| ! in_array( $protocol, $valid_protocols ) ) {
```
mjangda: This would end up rejecting relative URLs like `href="/"` which are valid.

bcampeau: That is very fair. Will address that.

includes/sanitizers/class-amp-iframe-sanitizer.php
bcampeau: Looks like they've moved this here: https://github.com/ampproject/amphtml/issues/2261

Let me see if I can work out an architecture to support both. I can very easily see getting requests for a fallback or just removing it altogether.

tests/test-amp-blacklist-sanitizer.php
bcampeau: This was throwing validation errors in Webmaster Tools as recently as last week. It looks like they've just updated the spec to support it and a few others:


      allowed_protocol: "http"
      allowed_protocol: "https"
      allowed_protocol: "mailto"
      # Whitelisting additional commonly observed third party
      # protocols which should be safe
      allowed_protocol: "sms"
      allowed_protocol: "tel"
      allowed_protocol: "viber"
      allowed_protocol: "whatsapp"

Will update accordingly.

bcampeau:
With regards to the comment about moving the new filters `amp_blacklisted_protocols`, `amp_blacklisted_tags` and `amp_blacklisted_attributes` to use args like `add_placeholder` on the iframe class, I dug deep into this logic. 

The initial problem is that the base class is using `array_merge` to replace default args with the supplied args for the class. Following the same pattern, you'd end up needing to supply the entire list of current blacklisted protocols/tags/attributes to add to or modify the list if they were directly customizable. A filter works better in this use case since it doesn't require you to keep up with changes to the plugin just to add a few additional tags/protocols/attributes required by your theme's content.

However, I can't think of a scenario where you'd ever want to allow someone to modify the built in since they will always be invalid. In essence, you'd be letting someone intentionally break validation which seems pointless. So, the args I've added are only for additional tags/protocols/attributes and do not allow any modification of the built-in list. I think this addresses following the pattern established with the iframe class and also protects anyone using this plugin against doing something intentionally or accidentally invalid.

bcampeau:
This is ready for final consideration for merging and is no longer a work in progress. I decided to add a new function to `AMP_Base_Sanitizer` called `validate_src_attribute` that allows the implementing function (iframe, video or audio) to decide whether or not the src should be forced to https (iframes do this currently, audio and video don't) or allow the theme to decide via the `amp_require_https_src` filter whether https should be required. If this filter returns true, the src is removed altogether.

If an iframe, audio or video node is found to have no valid src, it is then altogether removed. This only happens if the `amp_require_https_src` is used so without that, the functionality is identical to how it is currently. We could consider making the removal more automated once the determination is made on the proper fallback methodology. However, I advocate for merging this before then as this filter will help many sites resolve a large batch of validation errors.

I've also added a bunch of new unit tests for these scenarios.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -54,4 +54,29 @@ public function add_or_append_attribute( &$attributes, $key, $value, $separator
...
+	 *
+	 * @param string $src
+	 * @param boolean $force_https
+	 * @return string
+	 */
+	public function validate_src_attribute( $src, $force_https = false ) {
+		$protocol = strtok( $src, ':' );
+		if ( 'https' !== $protocol ) {
+			// Check if https is required
+			$https_required = apply_filters( 'amp_require_https_src', false );
```
mjangda: I think we should move this to an arg.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -54,4 +54,29 @@ public function add_or_append_attribute( &$attributes, $key, $value, $separator
...
+
+	/**
+	 * Decide if we should remove a src attribute if https is required.
+	 * If not required, the implementing class may want to try and force https instead.
+	 *
+	 * @param string $src
+	 * @param boolean $force_https
+	 * @return string
+	 */
+	public function validate_src_attribute( $src, $force_https = false ) {
```
mjangda: This might be better named `maybe_enforce_https_src` or something like that, since we're not actually validating anything.

includes/sanitizers/class-amp-blacklist-sanitizer.php
```diff
@@ -41,18 +48,28 @@ private function strip_attributes_recursive( $node, $bad_attributes, $bad_protoc
 				if ( 0 === stripos( $attribute_name, 'on' ) && $attribute_name != 'on' ) {
 					$node->removeAttribute( $attribute_name );
 					continue;
-				} elseif ( 'href' === $attribute_name ) {
+				} elseif ( 'href' === $attribute_name && 'a' !== $node_name ) {
```
mjangda: What tags other than `<a>` have an `href` attribute?

includes/sanitizers/class-amp-blacklist-sanitizer.php
```diff
@@ -93,24 +110,68 @@ private function sanitize_a_attribute( $node, $attribute ) {
...
+			$old_value = strtolower( $attribute->value );
+			if ( '_blank' === $old_value || '_new' === $old_value ) {
+				// _new is not allowed; swap with _blank
+				$node->setAttribute( $attribute_name, '_blank' );
+			} else {
+				// only _blank is allowed
+				$node->removeAttribute( $attribute_name );
+			}
+		} elseif ( 'href' === $attribute_name ) {
+			// If the href starts with a '/', append the home_url to it for validation purposes.
```
mjangda: Relative URLs aren't allowed? That seems like a bug in AMP.

includes/sanitizers/class-amp-blacklist-sanitizer.php
```diff
@@ -93,24 +110,68 @@ private function sanitize_a_attribute( $node, $attribute ) {
...
+			$href = $attribute->value;
+			if ( 0 === stripos( $attribute->value, '/' ) ) {
+				$href = untrailingslashit( get_home_url() ) . $href;
+			}
+
+			$valid_protocols = array( 'http', 'https', 'mailto', 'sms', 'tel', 'viber', 'whatsapp' );
+			$protocol = strtok( $href, ':' );
+			if ( false === filter_var( $href, FILTER_VALIDATE_URL )
```
mjangda: We probably need to allow for anchor links as well, i.e. `href="#section-2"`

includes/sanitizers/class-amp-blacklist-sanitizer.php
```diff
@@ -41,18 +48,28 @@ private function strip_attributes_recursive( $node, $bad_attributes, $bad_protoc
...
 					$protocol = strtok( $attribute->value, ':' );
 					if ( in_array( $protocol, $bad_protocols ) ) {
 						$node->removeAttribute( $attribute_name );
 						continue;
 					}
 				} elseif ( 'a' === $node_name ) {
-					$this->sanitize_a_attribute( $node, $attribute );
+					// Sanitize the tag, but remove it entirely if the href is invalid.
+					// Children will be preserved as part of the parent.
+					if ( false === $this->sanitize_a_attribute( $node, $attribute ) ) {
```
mjangda: We should separate out the validation and sanitization. No point in sanitizing the attributes if we're just going to drop the link. May have to do the validation before we process any attributes.

includes/sanitizers/class-amp-blacklist-sanitizer.php
```diff
@@ -93,24 +110,68 @@ private function sanitize_a_attribute( $node, $attribute ) {
...
+		// clone and re-add all the children just before current node.
+		if ( $node->hasChildNodes() && $node->parentNode ) {
+			foreach ( $node->childNodes as $child_node ) {
+				$new_child = $child_node->cloneNode( true );
+				$node->parentNode->insertBefore( $new_child, $node );
+			}
+		}
+
+		// Remove the node from the parent, if defined.
+		if ( $node->parentNode ) {
```
mjangda: Do we need the parentNode check when doing the child re-assignment too?

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -37,15 +37,21 @@ public function sanitize() {
 			$node = $nodes->item( $i );
 			$old_attributes = AMP_DOM_Utils::get_node_attributes_as_assoc_array( $node );
 
-			if ( ! array_key_exists( 'src', $old_attributes ) ) {
-				$node->parentNode->removeChild( $node );
-				continue;
-			}
-
 			$this->did_convert_elements = true;
```
mjangda: This should probably be moved after the empty `src` check.

includes/sanitizers/class-amp-blacklist-sanitizer.php
bcampeau: Only `<link>` and that's not valid in the body anyway. This is fine as-is in my opinion.

includes/sanitizers/class-amp-base-sanitizer.php
bcampeau: Fair, updated that.

includes/sanitizers/class-amp-base-sanitizer.php
bcampeau: Fine with me. Rewrote this and the corresponding unit tests.

includes/sanitizers/class-amp-blacklist-sanitizer.php
bcampeau: It's not a bug, Google does not want you to use `//` URLs. They want explicit protocols and I don't consider that a bug.

However, this isn't that. Relative URLs are totally allowed. To validate them, I'm just appending the home url so they can be tested. There's a corresponding unit test for this that's passing called 'a_with_href_relative'. I think we're OK here.

includes/sanitizers/class-amp-blacklist-sanitizer.php
bcampeau: Very good point. I've added that and a corresponding unit test.

includes/sanitizers/class-amp-blacklist-sanitizer.php
bcampeau: Oh, heh, I see your point. Yeah this block is pointless, isn't it?

includes/sanitizers/class-amp-blacklist-sanitizer.php
bcampeau: Yeah, this is an important sanity check in my opinion. Some weird, invalid HTML could in theory cause a fatal error here. There's no harm in having this check and I think we should leave it.

includes/sanitizers/class-amp-iframe-sanitizer.php
bcampeau: Agreed. Moved it.

includes/sanitizers/class-amp-blacklist-sanitizer.php
bcampeau: Fair point indeed. I've handled this and also added the scenario of `<a name="section2"></a>` both to the validation and unit tests.

bcampeau:
This is ready for another review.

includes/sanitizers/class-amp-blacklist-sanitizer.php
```diff
@@ -93,24 +102,87 @@ private function sanitize_a_attribute( $node, $attribute ) {
...
+		}
+	}
+
+	private function validate_a_node( $node ) {
+		// Get the href attribute
+		$href = $node->getAttribute( 'href' );
+
+		// If no href is set and this isn't an anchor, it's invalid
+		if ( empty( $href ) ) {
+			if ( ! empty( $node->getAttribute( 'name' ) ) ) {
```
mjangda:
```Fatal error: Can't use method return value in write context in /home/travis/build/Automattic/amp-wp/includes/sanitizers/class-amp-blacklist-sanitizer.php on line 124

Comment (bcampeau) [inline includes/sanitizers/class-amp-blacklist-sanitizer.php]: Fixed.

**Comment by bcampeau** [comment]
This is ready for another review. Just got back from a few days off and saw/fixed the failing test.

File: includes/sanitizers/class-amp-audio-sanitizer.php [hunk 13]
```diff
@@ -31,14 +31,28 @@ public function sanitize() {
 
 			$new_node = AMP_DOM_Utils::create_node( $this->dom, 'amp-audio', $new_attributes );
 
-			// TODO: limit child nodes too (only allowed: `source`; move rest to div+fallback)
 			// TODO: `source` does not have closing tag, and DOMDocument doesn't handle it well.
 			foreach ( $node->childNodes as $child_node ) {
 				$new_child_node = $child_node->cloneNode( true );
-				$new_node->appendChild( $new_child_node );
+				$old_child_attributes = AMP_DOM_Utils::get_node_attributes_as_assoc_array( $new_child_node );
+				$new_child_attributes = $this->filter_attributes( $old_child_attributes );
```
mjangda: Not going to block merging (I will open a new issue for this) but we should probably filter attributes for the `source` elements separately (since the general attributes don't apply for it).

includes/sanitizers/class-amp-blacklist-sanitizer.php
```diff
@@ -93,24 +102,88 @@ private function sanitize_a_attribute( $node, $attribute ) {
...
+		}
+
+		// If this is an anchor link, just return true
+		if ( 0 === strpos( $href, '#' ) ) {
+			return true;
+		}
+
+		// If the href starts with a '/', append the home_url to it for validation purposes.
+		if ( 0 === stripos( $href, '/' ) ) {
+			$href = untrailingslashit( get_home_url() ) . $href;
```
mjangda: I think we can just return `true` here?

includes/sanitizers/class-amp-blacklist-sanitizer.php
```diff
@@ -126,18 +199,25 @@ private function get_blacklisted_tags() {
...
-		);
+		) );
 	}
 
 	private function get_blacklisted_attributes() {
-		return array(
+		return $this->merge_defaults_with_args( 'add_blacklisted_attributes', array(
 			'style',
-		);
+			'srcset',
```
mjangda: Why are we stripping `srcset`?

mjangda:
Spotted two things but otherwise this is good to go. Thanks again for working on this!

includes/sanitizers/class-amp-audio-sanitizer.php
bcampeau: Fair point and agreed on a new issue

includes/sanitizers/class-amp-blacklist-sanitizer.php
bcampeau: Yes, in theory but I still want to check that the relative URL is in a valid format. I think this bit of extra validation is fine.

includes/sanitizers/class-amp-blacklist-sanitizer.php
bcampeau: I swear it wasn't supported when I started this, but looks like it was added on 2/11 to the spec. I'll remove this.

bcampeau:
Fixed or commented on those items, when you're ready to take another look.

PR #313, opened 2016-03-04
Add support for Archive / Authors / Tags and pagination

gabrielperezs:
Some URL structure examples:

Categories:
http://www.mysite.com/amp/category/linux/
http://www.mysite.com/amp/category/linux/page/2/

Tags:
http://www.mysite.com/amp/tag/linux/
http://www.mysite.com/amp/tag/linux/page/2/

Authors:
http://www.mysite.com/amp/author/gabriel/
http://www.mysite.com/amp/author/gabriel/page/2/

Posts:
http://www.mysite.com/my-nice-post-title/amp/

At the same time, make find and replace of the URL in the menu, the visitor can surf in the AMP version.

Also a new template for archive. And the object to handle. I think could be good to have a "common" class (abstract) and from this extends to archive, posts. In the future home-page, or other types.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -90,7 +90,7 @@ function amp_maybe_add_actions() {
 	if ( ( !is_home() && !is_singular() && !is_category() && !is_author()) && !is_tag() || is_feed() ) {
 		return;
 	}
-	
+
 	$is_amp_endpoint = is_amp_endpoint();
 
 	if ( !$is_amp_endpoint ) return;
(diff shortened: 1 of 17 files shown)


davidnaviaweb:
Is someone looking into this issue?

westonruter:
Support for archives, authors, tags, pagination and other templates is implemented in 0.7-beta via adding `amp` theme support.

PR #316, opened 2016-03-05
Remove iframe if protocol is http

kantan2015:
Probably removing iframe is the better than replacing the protocol to https.

### DIFF
--- a/includes/sanitizers/class-amp-iframe-sanitizer.php
+++ b/includes/sanitizers/class-amp-iframe-sanitizer.php
@@ -41,6 +41,10 @@ class AMP_Iframe_Sanitizer extends AMP_Base_Sanitizer {
 				$node->parentNode->removeChild( $node );
 				continue;
 			}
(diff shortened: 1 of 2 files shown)


jery61007:
No keeping the iframe replacing the protocol is better, because major sites that you may be using iframes to embed their videos and other things have their https version of their content or will gradually adapt it sooner as it is becoming one of the most important part of SEO strategy....  and with Google decision to make https protocol necessary to embed iframe to other sites , those websites with such sharable content will definitely have https protocol soon....   almost all major websites in the near future will have their https versions .....

kantan2015:
Hi, jery61007
I agree with your point, most of the site will support https in the futures.
However, I still think it better to remove http in the plugin because
i) URL of https version may differ from http
ii) Plugin user can replace the URL if the site support https
iii) Unfortunately, there is many sites which don't support https now
 Thanks

mjangda:
Yep, instead of stripping out, a better option would be to provide a `fallback` element instead which links to the `http` URL.

mjangda:
Issue for fallback #354

PR #319, opened 2016-03-07
Allow filtering of title bar

jdevalk:
In the new version of my Yoast SEO glue plugin I'm allowing people to change some colors etc for their AMP output. For a lot of reasons I'd like to stay as close as possible to the default template, however, I'd like to be able to change the title bar. A filter seems more appropriate for this then creating my own template file, so I was wondering if you'd consider this change: no change in functionality for the AMP plugin, but the option to replace the title bar content with simply a larger image and no blog name.


jdevalk:
@mjangda well actually, you could just do an echo on the filter and return false, and it'd work.

PR #323, opened 2016-03-08
some more blacklisted attributes

LDAV:
_noshade_ is blacklisted for **hr** tags
_align_ is blacklisted in **p** tags

in alternative, you can add an apply_filters:

```
private function get_blacklisted_attributes() {
  $args = array('style');
  return apply_filters('amp_blacklisted_attributes', $args);
}
```


mjangda:
Closing as we've moved to a whitelist-based approach #600

PR #326, opened 2016-03-09
Move nav bar to its own template

jdevalk:
This would make it easily overwritten, instead of having to overwrite all of `single.php`.

If this is merged, closes #319 too.


templates/single.php
```diff
@@ -11,17 +11,7 @@
...
-		<a href="<?php echo esc_url( $this->get( 'home_url' ) ); ?>">
-			<?php $site_icon_url = $this->get( 'site_icon_url' ); ?>
-			<?php if ( $site_icon_url ) : ?>
-				<amp-img src="<?php echo esc_url( $site_icon_url ); ?>" width="32" height="32" class="amp-wp-site-icon"></amp-img>
-			<?php endif; ?>
-			<?php echo esc_html( $this->get( 'blog_name' ) ); ?>
-		</a>
-	</div>
-</nav>
+<?php $this->load_parts( apply_filters( 'amp_post_template_meta_parts', array( 'nav' ) ) ); ?>
```
mjangda: Can we drop the filter here? It's only relevant for the `meta` section of the template.

templates/nav.php
```diff
@@ -0,0 +1,11 @@
+<nav class="amp-wp-title-bar">
```
mjangda: I think `title-bar` or `header-bar` might be a better name?

jdevalk:
@mjangda thanks for the comments, agreed on both counts, updated the code :)

mjangda:
Perfect, thank you!

PR #341, opened 2016-03-18
Update readme - callback error

trepmal:
Make function match callback for custom post type template example


mjangda:
Good catch!

PR #364, opened 2016-04-06
filter `amp_template_dir` added

mustafauysal:
this filter might help to use completely different template files.

### DIFF
--- a/includes/class-amp-post-template.php
+++ b/includes/class-amp-post-template.php
@@ -26,7 +26,7 @@ class AMP_Post_Template {
 	private $data;
 
 	public function __construct( $post_id ) {
-		$this->template_dir = AMP__DIR__ . '/templates';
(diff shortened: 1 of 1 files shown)


includes/class-amp-post-template.php
```diff
@@ -26,7 +26,7 @@ class AMP_Post_Template {
 	private $data;
 
 	public function __construct( $post_id ) {
-		$this->template_dir = AMP__DIR__ . '/templates';
+		$this->template_dir = apply_filters( 'amp_template_dir', AMP__DIR__ . '/templates' );
```
mjangda: For consistency, we should name this `amp_post_template_dir`.

mustafauysal: sure thing

mjangda:
Thanks!

PR #378, opened 2016-04-28
Adds a filter for canonical URL.

harvitronix:
Allows implementor to filter the canonical. 

In our case, we will use this to remove `/amp` that's appended to the canonical so it does not appear to search engines as duplicated content.


mjangda:
Why is `/amp` being appended to the canonical URL? That sounds like something else is wrong.

harvitronix:
At first I thought this was just how the plugin worked (that it was by design). But I just installed a fresh copy of WP and enabled the plugin and I don't see `/amp` appended to the canonical on the posts. So I'll close this PR and investigate further. Thanks!

PR #379, opened 2016-04-28
Google has requested we move the amphtml link rel higher up on the page.

harvitronix:
From Adam Greenberg @ Google: "Even though the page has a rel=amphtml tag when I fetch the page, it looks like Google's view of the page doesn't have any rel=amphtml tag present which makes it seem like they are cloaking in a way .  It might have to do with where they are putting their rel=amphtml tag.  **They should try to have it at the beginning of the header section**."

This pull request moves the rel="amphtml" tag near the top of the header section.

### DIFF
--- a/includes/amp-frontend-actions.php
+++ b/includes/amp-frontend-actions.php
@@ -1,7 +1,7 @@
 <?php
 // Callbacks for adding AMP-related things to the main theme
 
-add_action( 'wp_head', 'amp_frontend_add_canonical' );
+add_action( 'wp_head', 'amp_frontend_add_canonical', 1 );
 
 function amp_frontend_add_canonical() {
 	if ( false === apply_filters( 'amp_frontend_show_canonical', true ) ) {


mjangda:
From a code perspective, this seems fine.

But I find it weird that the position of the `link` tag would have an impact on things. Why would a machine/parser care at what point in a page the `link` tag is declared?

> it looks like Google's view of the page doesn't have any rel=amphtml tag present

Have you looked into whether this is actually the case and why it is happening?

harvitronix:
Hey Moe, I've taken a look at their "[amp bench](https://ampbench.appspot.com/validate?url=http://ftw.usatoday.com/2016/04/adam-scott-just-showed-why-golf-shouldnt-be-an-olympic-sport/amp)" and it doesn't trigger any issues, so I'm not sure if they're looking somewhere else. Agreed, it seems weird that their bot would care where on the page it is.

I'll follow up with Google on that.

PR #387, opened 2016-05-13
Fix image tags with `px` in width/height attrs.

STRML:
It's safe to strip these. Prevents WMT errors like:

"The attribute 'width' in tag 'amp-img' is set to the invalid value '450px;'."

### DIFF
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -42,6 +42,12 @@ abstract class AMP_Base_Sanitizer {
 
 		$attributes['sizes'] = sprintf( '(min-width: %1$dpx) %1$dpx, 100vw', absint( $max_width ) );
 
+		// Convert attributes like width="300px" to width="300"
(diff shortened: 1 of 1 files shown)


includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -42,6 +42,12 @@ public function enforce_sizes_attribute( $attributes ) {
 
 		$attributes['sizes'] = sprintf( '(min-width: %1$dpx) %1$dpx, 100vw', absint( $max_width ) );
 
+		// Convert attributes like width="300px" to width="300"
+		$attributes['width'] = absint( $attributes['width'] );
```
mjangda: Should we do this higher up in the method? We use `width` as the `max_width` value earlier and that could benefit from the `absint`.

One other issue is percentage values like `80%`. Passing that through `absint` converts the value to `80` which is not actually what we want.

mjangda:
Thanks for the PR. I've pushed #402 which adds a bit more like handling `%` dimensions.

Issue #391, opened 2016-05-18
AMP query_var being removed in build_query_string() – class-wp.php


mkuplens:
Although the AMP query_var is being parsed correctly on most sites, I've run into a situation on one site where the query_var is stripped unless a value is supplied.

This condition occurs at line 546 of build_query_string() function of class-wp.php (WordPress 4.5.2), caused by the null value (variable omitted from query string because of line 528 - skipped for empty value).

mkuplens:
I wonder whether adding something along these lines might do the job:

if ( has_filter( 'query_string' ) ) { add_filter( 'query_string', 'inject_amp_query' ); }
function inject_amp_query( $query_string ) {
    global $wp_query;
    if ( $wp_query->get( 'amp' ) ) {
        $query_string .= '&amp';
    }
    return $query_string;
}

mkuplens:
Per #392:

Activating a plugin that hooks into the 'query_string' filter will disable AMP output unless the amp parameter is given a value (?amp=1 or /amp/1).

Reproducible with this single-file plugin: http://cl.ly/gDiM (simply adds a hook and returns the value unchanged)

PR #392, opened 2016-05-19
Prevent query_string filters from blocking AMP display (Issue #391)

mkuplens:
Test for presence of 'query_string' filters and append AMP parameter to output to ensure the query_var is not lost.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -43,6 +43,7 @@ function amp_init() {
 	add_post_type_support( 'post', AMP_QUERY_VAR );
 
 	add_action( 'wp', 'amp_maybe_add_actions' );
+	add_action( 'parse_request', 'amp_maybe_preserve_query_var' );
 
 	if ( class_exists( 'Jetpack' ) && ! ( defined( 'IS_WPCOM' ) && IS_WPCOM ) ) {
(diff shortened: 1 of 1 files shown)


mjangda:
Thanks @mkuplens! Can you provide reproduction steps for the issue so I can confirm this does fix things? (I didn't see any in #391)

mkuplens:
Activating a plugin that hooks into the 'query_string' filter will disable AMP output unless the amp parameter is given a value (?amp=1 or /amp/1).

Reproduce with this single-file plugin: http://cl.ly/gDiM

mjangda:
Thanks. Confirmed the issue, although, how about this as an alternate fix (feels like this could prevent other future issues that might crop up with `query_vars`)?

```
// Make sure the `amp` query var has an explicit value.
// Avoids issues when filtering `query_string`.
add_filter( 'request', 'amp_force_query_var_value' );

function amp_force_query_var_value( $query_vars ) {
    if ( isset( $query_vars[ AMP_QUERY_VAR ] ) ) {
        $query_vars[ AMP_QUERY_VAR ] = 1;
    }
    return $query_vars;
}
```

mjangda:
Continued in #408

PR #417, opened 2016-07-09
Origin/ampvalidator

iamchrissmith:
Here is the url validation for the amp service to ensure code clears.  We still need to get all builds working in the travis matrix as our server setup is not working for php version 5.2, but the validator tool is working when the travis server sets up.  This appears to be tied to the enabling and use of sudo, because of this we are forced our of Travis's container environment which causes php5.2 to error at the environment setup.

You can see the same error with the _s repo [here](https://travis-ci.org/Automattic/_s/jobs/108258924).

Their solving comment is [here](https://github.com/Automattic/_s/blob/6ae11474be9dc76404c15f5d1b752059d3ad52c4/.travis.yml#L7)

Because we need sudo for WP-CLI and other server commands, I am investigating work arounds.  In the meantime, I have set 5.2 to allowed failures


bin/run-amp-validator.js
```diff
@@ -0,0 +1,228 @@
...
+    });
+
+    it('All URLs correctly validate', function(){
+        ourResults.length.should.equal(testUrls.length);
+        ourResults.should.all.be.equal('PASS');
+    });
+    it('No Errors found', function(){
+        ourResults.length.should.equal(testUrls.length);
+        ourErrors.length.should.equal(0);
+    });
```
mjangda: It feels a bit odd to run assertions for the test up bits; might be cleaner just to throw an error if getting the URLs fails, for example.

Also, could we not run each URL validation as a separate assertion?

bin/install-amp-validator-test-data.sh
```diff
@@ -0,0 +1,102 @@
...
+# Check if our test data exists. If it already does not, we will get it from github.
+if ! [ -f "wptest.xml" ]
+then
+
+	# Get WP Test data if a newer file exists.
+	cd tests/assets && { curl -OL -z wptest.xml https://raw.githubusercontent.com/manovotny/wptest/master/wptest.xml; cd - ; }
+
+fi
+
+if [ "${TRAVIS}" = "true" ]; then
```
mjangda: Why do we have separate flows for Travis and local testing here?

bin/run-amp-validator.js
```diff
@@ -0,0 +1,228 @@
...
+    return resolver.promise;
+};
+
+/**
+ * Timeout promise
+ * adapted from: http://exploringjs.com/es6/ch_promises.html
+ * @param ms integer time in milliseconds before timing out our promise with a reject()
+ * @param promise string name of promise function to timeout
+ */
+var timeout = function (ms, promise) {
```
mjangda: This might be better name `promiseWithTimeout` or similar.

bin/run-amp-validator.js
```diff
@@ -0,0 +1,228 @@
...
+                                    ourResults.push('FAIL');
+                                }
+                                i++;
+                                resolve();
+                            }).catch(function (reason) {
+                                i++;
+                                ourErrors.push(reason);
+                                console.error('Error or timeout', reason);
+                                reject(reason);
+                            });
```
mjangda: This block is very deeply nested and somewhat hard to follow. It would be nice to flatten further.

bin/run-amp-validator.js
```diff
@@ -0,0 +1,228 @@
...
+                });
+        });
+    });
+};
+
+describe('AMP Validation Suite', function() {
+    this.timeout(20000);
+
+    before( function() {
+        var child = exec('wp post list --post_type=post --posts_per_page=-1 --post_status=publish --post_password="" --format=json --fields=url --quiet --skip-plugins=wordpress-importer');
```
mjangda: The URL fetching could be broken out into a separate function?

bin/install-amp-validator-test-data.sh
iamchrissmith: We have to install WP-CLI on Travis every time. I also wanted to give the local user the option of running only one or none of the data imports and of noting being forced to have their URL structure rewritten.

bin/run-amp-validator.js
iamchrissmith: Yes, I changed the URLs assertion out of mocha and into just a process exit.

However, I cannot run the validation for each URL individually.  I tried that, but mocha defines its tests before any code runs, so while we can create tests using .each(), because the URL array hasn't been populated yet, it will define 0 URL tests.  I could get around this by hardcoding the number of test URLs, but then we have to remember to update that number if the number of URLs changes in the future (or add a test to see if the numbers equal as a warning system).

bin/run-amp-validator.js
iamchrissmith: Good idea, change made

bin/run-amp-validator.js
iamchrissmith: This has been better addressed with [this commit](https://github.com/iamchrissmith/amp-wp/commit/a43e140bca67b87b200fdb33fc3faca8d6ad6bef)

Issue #340, opened 2016-03-17
1GB error log full with "php_network_getaddresses: getaddrinfo failed: "


thagxt:
Hi, my error log gets flooded with these errors, daily. If I don't clear the logs it goes up to 1GB. 
I suppose it's some kind of DNS error and server setup... how can I fix it?

```
[15-Mar-2016 23:25:55 UTC] PHP Warning:  fopen(): php_network_getaddresses: getaddrinfo failed: Temporary failure in name resolution in /path/to/public_html/wp-content/plugins/amp/includes/lib/class-fastimage.php on line 31
[15-Mar-2016 23:25:55 UTC] PHP Warning:  fopen(http://40.media.tumblr.com/tumblr_lp1qopjYyX1qcmkeso1_500.jpg): failed to open stream: php_network_getaddresses: getaddrinfo failed: Temporary failure in name resolution in /path/to/public_html/wp-content/plugins/amp/includes/lib/class-fastimage.php on line 31
[15-Mar-2016 23:26:05 UTC] PHP Warning:  fopen(): php_network_getaddresses: getaddrinfo failed: Temporary failure in name resolution in /path/to/public_html/wp-content/plugins/amp/includes/lib/class-fastimage.php on line 31
```

mjangda:
This is fixed as of #480

PR #418, opened 2016-07-11
Check that the image exists before passing to FastImage

mlteal:
FastImage runs fopen() before checking if file_exists() first. If an image no longer exists, attempting to run open on it will trigger a PHP warning.

Fixes the PHP warnings mentioned in issue #340


mjangda:
Does `file_exists` work with URLs?

mlteal:
@mjangda 🤔 honestly I don't think it should... Just haven't had time to actually look through the test failure. This patch is already working fine in a production environment for me, though, which I guess is strange! I'll have to switch to check for 404 headers on the image instead.

mjangda:
Closing as `file_exists` doesn't work with URLs. There are some improvements to image fetching being worked on in #480 as well.

Issue #397, opened 2016-06-06
Co-authors-plus template override


jakewrfoster:
In reference to the PR that fixed #137:
[This action](https://github.com/Automattic/Co-Authors-Plus/blob/master/php/integrations/amp.php#L3) currently overrides the AMP custom template `/amp/meta-author.php`. In order to avoid confusion, either: 
a) [the docs](https://github.com/Automattic/amp-wp#example-override-author-template-from-theme) should be updated to reflect this, or
b) within the `verify_and_include( $file, $template_type )` method, [the AMP templating system](https://github.com/Automattic/amp-wp/blob/master/includes/class-amp-post-template.php#L377) should be called _after_ [the filter is applied](https://github.com/Automattic/amp-wp/blob/master/includes/class-amp-post-template.php#L382) .

I am happy to provide a PR for either of these options if you would let me know which you are more likely to push through.

Thanks!
Jake

EDIT: update links

ebiewald:
Looks like this has been brought up to Co-Authors Plus too https://github.com/Automattic/Co-Authors-Plus/issues/360

mjangda:
I think the filter should continue run after (as a "last-resort" option for other plugins/themes to integrate and modify).

I was thinking either that we:

1) add a new filter before the `locate_template` lookup; or
2) let the existing filter know that we found a theme template so that anyone that hooks in can act accordingly.

romancandlethoughts:
@mjangda, 

Introducing a second filter to override the manual template does nothing to address the confusion that led to the creation of this issue. I would sooner deprecate the manual template construct altogether and have it _just_ be a filter rather than have any option for a filter overriding the manual templating construct. This way, theme devs would at least have absolute control (or knowledge, at least) over the priority. If there is both a filter and a manual template construct, it is very unclear as to what will take precedence (and which plugin it would be coming from). I am guessing that for a lot of theme devs, this could be something that proves very challenging/time consuming to debug. 

All confusion seems to be avoided if the filter is fired before the manual templating constuct. I can't imagine a scenario wherein a dev would introduce those template files and wonder why a third-party plugin _isn't_ affecting or overriding the output. It would also be the first place a dev would think to look if they _wanted_ to see plugin overrides. 

[...]

PR #422, opened 2016-07-21
Allow manual template to override template filter

jakewrfoster:
In reference to #397


westonruter:
There are conflicts in the PR, and besides we'll be discouraging the use of the built-in classic templates in favor sites opting for full AMP theme support going forward, where the plugin will fully re-use a theme's existing templates. That being the case, Co-Authors Plus will be fully supported just the same as on the non-AMP site. So I'm closing this PR.

Issue #394, opened 2016-05-27
Support for canonical AMP (AMP only)


pbakaus:
We, the AMP team, recently started to feel confident enough to encourage sites to build their sites directly for AMP. This means that instead of having a 'full' canonical page and an AMP page associated with it, you'd simply build your whole site with AMP.

I have a blog post coming talking in detail about our reasoning and strategy behind this (mostly centered around preventing the problems that m.dot sites bring), but I'd love to see support for it in amp-wp.

This would require the following:

1) Transform the default post output to AMP (i.e. img to amp-img)
2) First-class support for AMP components in the editor

Would love to hear what you think, and if feasible. Happy to work with you on a PR.

mjangda:
Sounds like a logical future path for the plugin. We've chatted about WordPress themes built entirely with AMP, which is a pretty cool idea. Of course, there are some (fun) challenges in the way :)

Would love to hear more about what you have in mind. Want to ping me to chat further? Hangouts (batmoo@gmail.com) or Skype (mjangda) are probably easiest ways to reach me.

pbakaus:
Sure! Send you an invite to chat on Hangouts :)

pbakaus:
@mjangda I just took a stab at adding this functionality and it all seems to be working. Have a look at the PR here: https://github.com/Automattic/amp-wp/pull/425

amedina:
Closing this one (duplicate); see https://github.com/Automattic/amp-wp/issues/425, and https://github.com/Automattic/amp-wp/pull/668

PR #188, opened 2016-02-16
Add support for pages

technosailor:
This addresses Issue #176

PR #425, opened 2016-07-29
Add ability to go canonical AMP (AMP-only)

pbakaus:
This directly addresses #394 and adds the following:
1. Configuration page under Settings with one setting (_Canonical AMP_), off by default right now
2. A WP option called `amp_canonical` that the plugin, plus any other plugins and themes can use to do certain things differently when signalled the blog is in "full AMP mode".
3. A theme support flag called `amp` that the theme can use to signal it supports AMP. 
4. Renders the post content of all posts (only single posts right now) as AMP and does a few tricks to make WP behave nicely.

Importantly, this PR does not address the actual used theme – it's still the theme's responsibility right now to make sure no external JS or CSS is loaded.
## Where can I see a demo?

Glad you're asking. A live example of a canonical AMP blog + theme using this plugin can be found here: [https://paulbakaus.com](https://paulbakaus.com)
## How do I make my theme validate?

Here's how a theme would conform:
### 1. Add `amp` theme support

The theme must signal that it supports AMP in order for the plugin to enable canonical AMP mode, like so:

```
add_theme_support( 'amp' );
```
### 2. Inline stylesheet

Something like:

```
function add_inline_stylesheet() {
    if( get_option( 'amp_canonical') ) {
        $css = file_get_contents(get_stylesheet_uri());
        ?>
        <style amp-custom><?= $css; ?></style>
        <?php
    }
}

add_action( 'wp_head', 'add_inline_stylesheet' );
```
### 3. Provide valid header bits

Include the flash in the html tag (doesn't hurt even without AMP), make sure that the charset is lowercase and that the correct viewport is set.

```
<html ⚡ <?php language_attributes(); ?>>
    <head>
        <meta charset="<?= strtolower(get_bloginfo( 'charset', 'display' )); ?>">
        <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
```
### 4. Embed AMP analytics

Use AMP analytics instead of other analytics solutions based on the AMP_CANONICAL check:

```
if ( get_option( 'amp_canonical') ) {
?>
<amp-analytics type="googleanalytics" id="analytics-ga">
  {
    "vars": {
      "account": "UA-2427667-1"
    },
    "triggers": {
      "trackPageview": {
        "on": "visible",
        "request": "pageview"
      }
    }
  }
  </script>
</amp-analytics>
<?php
}
```
### 5. Optional: Add AMP-specific styles

Currently, stolen from the styles in the default template. These are not required but helpful:

```
.amp-wp-enforced-sizes {
    /** Our sizes fallback is 100vw, and we have a padding on the container; the max-width here prevents the element from overflowing. **/
    max-width: 100%;
}

.amp-wp-unknown-size img {
    /** Worst case scenario when we can't figure out dimensions for an image. **/
    /** Force the image into a box of fixed dimensions and use object-fit to scale. **/
    object-fit: contain;
}

amp-carousel > amp-img > img {
    object-fit: contain;
}
```

This is my first real attempt at writing Wordpress-compliant code and I haven't touched PHP for ages, so a lot of it is probably not compliant or can be done in better ways. Very much looking for feedback and somebody who can take it and polish it if needed.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -15,6 +15,10 @@ define( 'AMP__FILE__', __FILE__ );
 define( 'AMP__DIR__', dirname( __FILE__ ) );
 
 require_once( AMP__DIR__ . '/includes/amp-helper-functions.php' );
+require_once( AMP__DIR__ . '/options.php' );
+
+// Determine if we want to treat the whole page as AMP
+define( 'AMP_CANONICAL', get_option( 'amp_canonical', 0 ) );
 
 register_activation_hook( __FILE__, 'amp_activate' );
 function amp_activate(){
@@ -39,8 +43,10 @@ function amp_init() {
 
 	load_plugin_textdomain( 'amp', false, plugin_basename( AMP__DIR__ ) . '/languages' );
 
-	add_rewrite_endpoint( AMP_QUERY_VAR, EP_PERMALINK );
-	add_post_type_support( 'post', AMP_QUERY_VAR );
+	if( !AMP_CANONICAL ) {
+		add_rewrite_endpoint( AMP_QUERY_VAR, EP_PERMALINK );
+		add_post_type_support( 'post', AMP_QUERY_VAR );
+	}
 
 	add_filter( 'request', 'amp_force_query_var_value' );
 	add_action( 'wp', 'amp_maybe_add_actions' );
@@ -82,6 +88,8 @@ function amp_maybe_add_actions() {
 
 	if ( $is_amp_endpoint ) {
 		amp_prepare_render();
+	} else if(AMP_CANONICAL && $supports) {
+		amp_render_canonical();
 	} else {
 		amp_add_frontend_actions();
 	}
@@ -113,4 +121,13 @@ function amp_render() {
 	$template = new AMP_Post_Template( $post_id );
 	$template->load();
 	exit;
+}
+
+function amp_render_canonical() {
+	require_once( AMP__DIR__ . '/includes/amp-canonical-actions.php' );
+}
+
+// load high priority filters for canonical AMP
+if( AMP_CANONICAL ) {
+	require_once( AMP__DIR__ . '/includes/amp-canonical-filters.php' );
 }
--- a/amp.php
+++ b/amp.php
@@ -17,9 +17,6 @@ define( 'AMP__DIR__', dirname( __FILE__ ) );
 require_once( AMP__DIR__ . '/includes/amp-helper-functions.php' );
 require_once( AMP__DIR__ . '/options.php' );
 
-// Determine if we want to treat the whole page as AMP
-define( 'AMP_CANONICAL', get_option( 'amp_canonical', 0 ) );
-
 register_activation_hook( __FILE__, 'amp_activate' );
 function amp_activate(){
 	amp_init();
@@ -43,7 +40,8 @@ function amp_init() {
 
 	load_plugin_textdomain( 'amp', false, plugin_basename( AMP__DIR__ ) . '/languages' );
 
-	if( !AMP_CANONICAL ) {
+	// Add rewrite endpoints if we either don't want canonical AMP, or if the theme doesn't support it
+	if( ! get_option( 'amp_canonical') || ! get_theme_support('amp') ) {
 		add_rewrite_endpoint( AMP_QUERY_VAR, EP_PERMALINK );
 		add_post_type_support( 'post', AMP_QUERY_VAR );
 	}
@@ -88,7 +86,7 @@ function amp_maybe_add_actions() {
 
 	if ( $is_amp_endpoint ) {
 		amp_prepare_render();
-	} else if(AMP_CANONICAL && $supports) {
+	} else if( get_option( 'amp_canonical') && $supports && get_theme_support('amp') ) {
 		amp_render_canonical();
 	} else {
 		amp_add_frontend_actions();
@@ -128,6 +126,6 @@ function amp_render_canonical() {
 }
 
 // load high priority filters for canonical AMP
-if( AMP_CANONICAL ) {
+if( get_option( 'amp_canonical') ) {
 	require_once( AMP__DIR__ . '/includes/amp-canonical-filters.php' );
 }
--- a/includes/amp-canonical-actions.php
+++ b/includes/amp-canonical-actions.php
@@ -2,6 +2,7 @@
 
 require_once( AMP__DIR__ . '/includes/utils/class-amp-dom-utils.php' );
 require_once( AMP__DIR__ . '/includes/utils/class-amp-html-utils.php' );
+require_once( AMP__DIR__ . '/includes/utils/class-amp-string-utils.php' );
 
 require_once( AMP__DIR__ . '/includes/class-amp-content.php' );
 
--- a/includes/amp-canonical-actions.php
+++ b/includes/amp-canonical-actions.php
@@ -19,7 +19,7 @@ require_once( AMP__DIR__ . '/includes/embeds/class-amp-instagram-embed.php' );
 require_once( AMP__DIR__ . '/includes/embeds/class-amp-vine-embed.php' );
 require_once( AMP__DIR__ . '/includes/embeds/class-amp-facebook-embed.php' );
 
-function build_post_content() {
+function amp_canonical_retrieve_content() {
 
 	$post = get_post();
 
@@ -56,7 +56,7 @@ function build_post_content() {
 }
 
 // Generate the AMP post content early on (runs the_content filters but skips our filter below)
-$GLOBALS['amp_content'] = build_post_content();
+$GLOBALS['amp_content'] = amp_canonical_retrieve_content();
 
 // "the_content" filter was already invoked now, so attempt remove all filters
 remove_all_filters('the_content');
--- a/includes/amp-canonical-actions.php
+++ b/includes/amp-canonical-actions.php
@@ -57,13 +57,17 @@ function build_post_content() {
 // Generate the AMP post content early on (runs the_content filters but skips our filter below)
 $GLOBALS['amp_content'] = build_post_content();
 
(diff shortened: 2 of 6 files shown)


mjangda:
Thanks for tackling this @pbakaus! Looking pretty good at a quick glance. I'll take it for a more thorough test drive and code review next week.

pbakaus:
Sounds good @mjangda! I just followed up with a few fixes, mostly to make it play nicer in the WP world by using get_option instead of a constant, and adding a theme_support flag so themes can signal they're AMP ready. This gives us an easy way forward to show valid AMP themes.

pbakaus:
friendly ping!

mjangda:
Considering this is still pretty experimental, I wonder if we should remove the options page + toggle for now and let power-users enable via the `add_theme_support` call.

Mostly trying to avoid a barrage of support requests complaining about all the indexing errors they're getting from GWT (we get enough of those as it is :))

pbakaus:
:) We can, but does that work for WP-hosted blogs as well (I have never modified a theme in a hosted WP blog). If so, I'm cool with it.

pbakaus:
Oh also, I'm looking for advice on how to properly add canonical AMP support for non-single pages (e.g. search, homepage, archive etc.). Any thoughts?

pbakaus:
@mjangda let me know which way to proceed.

includes/amp-canonical-filters.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+// before Jetpack ever gets loaded, we need to remove a link rel prefetch for canonical AMP support
```
mjangda: Jetpack adds <a href="https://github.com/Automattic/jetpack/pull/1399/files#diff-9ba9fa5bb52bd6cf36c6c0bdd5ed4830R4939">a bunch of `dns-prefetch` tags</a> (not just for VideoPress) so a better approach might be to unhook it completely:
```
function amp_canonical_disable_jetpack_dns_fetch() {
    if ( class_exists( 'Jetpack' ) ) {
        remove_action( 'wp_head', array( 'Jetpack', 'dns_prefetch' ) ); 
    }
}
add_action( 'wp_head', 'amp_canonical_disable_jetpack_dns_fetch', 0 ); // Hook early so we can unhook Jetpack
```

includes/amp-helper-functions.php
```diff
@@ -18,7 +18,7 @@ function amp_get_permalink( $post_id ) {
 
 function post_supports_amp( $post ) {
 	// Because `add_rewrite_endpoint` doesn't let us target specific post_types :(
-	if ( ! post_type_supports( $post->post_type, AMP_QUERY_VAR ) ) {
+	if (!AMP_CANONICAL && !post_type_supports( $post->post_type, AMP_QUERY_VAR ) ) {
```
mjangda: `AMP_CANONICAL` was removed. Might be nice to have a helper function since we do the same checks a few places.

includes/amp-canonical-actions.php
```diff
@@ -0,0 +1,108 @@
...
+require_once( AMP__DIR__ . '/includes/sanitizers/class-amp-audio-sanitizer.php' );
+
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-twitter-embed.php' );
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-youtube-embed.php' );
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-gallery-embed.php' );
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-instagram-embed.php' );
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-vine-embed.php' );
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-facebook-embed.php' );
+
+function build_post_content() {
```
mjangda: Function should be prefixed with `amp_canonical_` (just to be safe).

jdevalk:
Seems to me this needs #188 to be fully useful. Anything we from Yoast can do to help speed this whole thing up? It's going very slowly whereas AMP development in general seems to be nice and fast, and we'd love to make this all work.

includes/amp-canonical-actions.php
```diff
@@ -0,0 +1,108 @@
...
+			'content_max_width' => $content_max_width,
+		)
+	);
+
+	return $amp_content;
+
+}
+
+// Generate the AMP post content early on (runs the_content filters but skips our filter below)
+$GLOBALS['amp_content'] = build_post_content();
```
moorscode: I would advice strongly against introducing any `$GLOBALS`.

includes/amp-canonical-actions.php
moorscode: And it really doesn't build a post_content, it retrieves a new AMP_Content object. The name really confuses me.

includes/amp-canonical-actions.php
pbakaus: I haven't touched PHP in a while – can you point me to a better way of doing this?

includes/amp-canonical-filters.php
pbakaus: Excellent, I've used your code instead.

pbakaus:
@mjangda please take another look.

I've fixed a bunch of issues, and have reworded the provided option, clearly stating it's experimental, and linking out to the amended readme to explain exactly what it does, and doesn't.

I do think it's important to be able to toggle it on/off, even in hosted environments, and the theme having support for AMP should not be the signal for "I want it all turned into AMP". Themes that support AMP should allow for both render mods – non-AMP and AMP, and it should be up to the user to decide.

What are your remaining blockers to get this merged and deployed as new version?

pbakaus:
@jdevalk while I think #188 would be beneficial, I don't think it should be a blocking requirement. 

This PR is useful on its own, and will turn all posts into valid AMP when done correctly (like on my blog), while rendering everything else normally – the theme still looks the same, the transition for the user is seamless.

options.php
```diff
@@ -0,0 +1,78 @@
...
+add_action( 'admin_init', 'amp_register_settings' );
+add_action( 'admin_menu', 'amp_custom_admin_menu' );
+
+/**
+ * Register the settings
+ */
+function amp_register_settings() {
+     register_setting(
+          'amp_options',  // settings section
+          'amp_canonical' // setting name
```
mjangda: We should pass in a third arg here, which at a minimum, specifies the `type`, `sanitize_callback`, and `default`. (see https://developer.wordpress.org/reference/functions/register_setting/#parameters)

amp.php
```diff
@@ -40,8 +41,11 @@ function amp_init() {
 
 	load_plugin_textdomain( 'amp', false, plugin_basename( AMP__DIR__ ) . '/languages' );
 
-	add_rewrite_endpoint( AMP_QUERY_VAR, EP_PERMALINK );
-	add_post_type_support( 'post', AMP_QUERY_VAR );
+	// Add rewrite endpoints if we either don't want canonical AMP, or if the theme doesn't support it
+	if( ! get_option( 'amp_canonical') || ! get_theme_support('amp') ) {
```
mjangda: We should probably have a helper function to check if canonical mode is enabled (and then use it wherever applicable).

amp.php
```diff
@@ -116,6 +122,15 @@ function amp_render() {
 	exit;
 }
 
+function amp_render_canonical() {
+	require_once( AMP__DIR__ . '/includes/amp-canonical-actions.php' );
+}
+
+// load high priority filters for canonical AMP
+if( get_option( 'amp_canonical') ) {
+	require_once( AMP__DIR__ . '/includes/amp-canonical-filters.php' );
```
mjangda: We can move this to `amp_render_canonical`.

options.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+add_action( 'admin_init', 'amp_register_settings' );
+add_action( 'admin_menu', 'amp_custom_admin_menu' );
```
mjangda: We need an easy way to disable this the options page in certain contexts (e.g. across all of WordPress.com). To do so, we could:

- Rename `amp_register_settings` to `amp_admin_init`
- Move the `admin_menu` hook into `amp_admin_init`
- Add a filter in `amp_admin_init` that allows skipping registration of the admin pages.

includes/amp-canonical-actions.php
```diff
@@ -0,0 +1,108 @@
...
+}
+
+function amp_canonical_add_scripts() {
+
+	$amp_runtime_script = 'https://cdn.ampproject.org/v0.js';
+
+	// Always include AMP form & analytics, as almost every template has search form somewhere and is tracking page views
+	$scripts = array_merge(
+		array('amp-form' => 'https://cdn.ampproject.org/v0/amp-form-0.1.js'),
+		array('amp-analytics' => 'https://cdn.ampproject.org/v0/amp-analytics-0.1.js'),
```
mjangda: Should we leave this to themes/plugins to manage instead of assuming that all pages will have a form and/or analytics?

includes/amp-helper-functions.php
```diff
@@ -18,7 +18,7 @@ function amp_get_permalink( $post_id ) {
 
 function post_supports_amp( $post ) {
 	// Because `add_rewrite_endpoint` doesn't let us target specific post_types :(
-	if ( ! post_type_supports( $post->post_type, AMP_QUERY_VAR ) ) {
+	if (! get_option( 'amp_canonical') && ! post_type_supports( $post->post_type, AMP_QUERY_VAR ) ) {
```
mjangda: Should we always `return true` when in canonical mode?

amp.php
```diff
@@ -116,6 +122,15 @@ function amp_render() {
 	exit;
 }
 
+function amp_render_canonical() {
```
mjangda: We should rename this to `amp_add_canonical_actions` for clarity.

mjangda:
Continued in #668

Issue #328, opened 2016-03-10
Don't load Merriweather by default


corysimmons:
Isn't the whole point of AMP to remove unnecessary downloads? Why force a font download on people? Just provide a Georgia stack instead.

mjangda:
Merriweather looks much nicer than Georgia :) Plus, we're serving from a CDN so the download should be super fast.

But, overall, this is a valid point. We'll re-asses.

corysimmons:
> Merriweather looks much nicer than Georgia

:100:% subjective. Personally I like Georgia better. Merriweather has a dumb question mark.

> we're serving from a CDN so the download should be super fast

Slippery slope reasoning like this was literally why AMP was invented.

Frits1980:
I second the reasons corysimmons gave.
AMP shouldn't be unnecessarily dependent off any CDN.

Besides The extra DNS call's could easily make the AMP page run slow again when on slow network. So it totally defeats the purpose of AMP.

vijayaraghavanramanan:
I suggest the usage of fontfaceobserver. 

https://github.com/bramstein/fontfaceobserver

With that one can display text in Georgia and swap to Merriweather immediately after the font is downloaded.

Frits1980:
I still think layout and design should be kept dynamic and configurable by the developer/administrator instead of being implanted in core code.

corysimmons:
fontfaceobserver looks cool, but why are we doing _any_ fonts by default?

What if I don't want my visitors to have to download a font on their small mobile phone plan?

This is actually fairly relevant to me. I'm about to downgrade my phone plan to a bare minimum of slow data and every kb counts.

I suppose the underlying point is people are going to use the default theme for this (millions probably already are with wp.com and VIP) -- why are we adding a significant amount of download data to people by default? Just because some amp-wp designer thought Merriweather looked prettier than Georgia?

There's a big difference in data size between setting a few minimal styles for a default theme, and loading an entire font.

corysimmons:
Let's ask the core contributors of AMP what they think.

@cramforce @dvoytenko @erwinmombay @Meggin

Do you think amp-wp's default theme should load a font that will be served to millions of websites by default because someone thinks it looks slightly prettier than a system font?

cramforce:
I'd recommend using [amp-font](https://www.ampproject.org/docs/reference/extended/amp-font.html) to mitigate the effect of the font download on page load. This is particular important on iOS which has an infinite timeout by default (other browsers wait 3s)

[...]

Issue #200, opened 2016-02-18
Filterable font


mjangda:
Right now, we force Merriweather and Open Sans on people as it's included in the head. It would be nice if users could swap out the fonts with something of their own choice.

natebot:
Happy to take this one on @mjangda. Do you think it best if the filter goes into `styles.php` with Merriweather as the default?  I can provide PR and then we can discuss the merits of how to handle classes like `.content` via filter vs overwriting in custom css in header?

westonruter:
This is largely obsolete now that 0.7 and 1.0 encourage the use of the active theme's templates when serving AMP.

PR #437, opened 2016-08-18
Switch back to Google Fonts

mjangda:
Instead of loading multiple fonts files, let's just use Google Fonts.
We've also moved the font loading to an action so it can be disabled via
`remove_action`.

We're going to look into adding a toggle to disable this via the Customizer once that lands and possibly a font selector as well.

See #200 
See #328


Frits1980:
As long as it is by choice it's fine by me :)

mjangda:
Obviated by f9cd054

PR #449, opened 2016-08-23
fix replace_node_with_children

Maxime2:
We need to sanitize children of removed node as if there are more than one those starting from second would be missed in sanitisation in parent's foreach loop.


mjangda:
Thanks for the PR. Can you share some example HTML that this fixes?

Maxime2:
Something like that:
`<a><span style="color: red;">Red</span>&amp;<span style="color: blue;">Orange</span></a>`

mjangda:
Continued in #462

Issue #386, opened 2016-05-05
Additional styling and customizations for default template


ericlindley-g:
(pending #199)

Suggested template mocks below, with three variations, based on what the publisher has hypothetically supplied/specified: one with a hero image at the top of the page, one without, and one without a publisher avatar image or logo.

Additional requests not shown in the mocks
- Hide the category if none is specified (so "uncategorized" doesn't show)
- If the publisher can't specify a banner logo at the top of the page, enable specifying a font color and background color for their publication title

![amp-template](https://cloud.githubusercontent.com/assets/16807766/15060505/009342d6-12e0-11e6-83e1-9418f6fd92db.png)

westonruter:
This is obsolete now that the plugin encourages users to re-use their existing active theme in AMP.

PR #453, opened 2016-08-24
[WIP] Refresh AMP Template

allancole:
- Removed previous amp template
- Using JS to generate color settings in customizer live preview and
using PHP to generate color setting on live published site (uses less
css this way)

See #386

### DIFF
--- a/includes/admin/class-amp-customizer.php
+++ b/includes/admin/class-amp-customizer.php
@@ -160,8 +160,8 @@ class AMP_Template_Customizer {
 	 */
 	public function register_panel() {
 		$this->wp_customize->add_panel( self::PANEL_ID, array(
-			'type'            => 'amp',
-			'title'           => __( 'AMP', 'amp' ),
+			'type'  => 'amp',
+			'title' => __( 'AMP', 'amp' ),
 		) );
 	}
 
@@ -197,7 +197,7 @@ class AMP_Template_Customizer {
 		// Navbar background color control.
 		$this->wp_customize->add_control(
 			new WP_Customize_Color_Control( $this->wp_customize, 'amp_navbar_background_color', array(
-				'label'    => __( 'Header Background & Link Color', 'amp' ),
+				'label'    => __( 'Header Background Color & Link Color', 'amp' ),
 				'section'  => 'amp_navbar_section',
 				'priority' => 20
 			) )
--- a/includes/admin/class-amp-customizer.php
+++ b/includes/admin/class-amp-customizer.php
@@ -260,9 +260,6 @@ class AMP_Template_Customizer {
 	}
 }
 
-// Output custom CSS to live site
-add_action( 'wp_head' , array( 'AMP_Template_Customizer' , 'header_output' ) );
-
 function amp_sanitize_color_scheme( $value ) {
     if ( ! in_array( $value, array( 'light', 'dark', 'default' ) ) )
         $value = 'light';
--- a/includes/admin/class-amp-customizer.php
+++ b/includes/admin/class-amp-customizer.php
@@ -121,26 +121,33 @@ class AMP_Template_Customizer {
 	 * @access public
 	 */
 	public function register_settings() {
-		// Navbar text color setting.
+
+		// Header text color setting
 		$this->wp_customize->add_setting( 'amp_navbar_color', array(
 			'default'           => '#ffffff',
 			'sanitize_callback' => 'sanitize_hex_color',
 			'transport'         => 'postMessage'
 		) );
 
-		// Navbar background color setting.
+		// Header background color
 		$this->wp_customize->add_setting( 'amp_navbar_background_color', array(
 			'default'           => '#0a89c0',
 			'sanitize_callback' => 'sanitize_hex_color',
 			'transport'         => 'postMessage'
 		) );
 
-		// Radio Input               =
+		// Header background image upload
+		$this->wp_customize->add_setting( 'amp_navbar_background_image', array(
+			'default'    => '',
+			'transport'  => 'postMessage'
+		) );
+
+		// Background color scheme
 		$this->wp_customize->add_setting( 'amp_background_color', array(
 			'default'           => 'standard',
 			'sanitize_callback' => 'amp_sanitize_color_scheme',
 			'transport'         => 'postMessage'
-		));
+		) );
 
 		do_action( 'amp_customizer_register_settings', $this->wp_customize );
 	}
@@ -166,7 +173,7 @@ class AMP_Template_Customizer {
 	 */
 	public function register_sections() {
 		$this->wp_customize->add_section( 'amp_navbar_section', array(
-			'title' => __( 'Navigation Bar', 'amp' ),
+			'title' => __( 'Color Options', 'amp' ),
 			'panel' => self::PANEL_ID,
 		) );
 	}
@@ -190,9 +197,17 @@ class AMP_Template_Customizer {
 		// Navbar background color control.
 		$this->wp_customize->add_control(
 			new WP_Customize_Color_Control( $this->wp_customize, 'amp_navbar_background_color', array(
-				'label'    => __( 'Header Background Color', 'amp' ),
+				'label'    => __( 'Header Background & Link Color', 'amp' ),
 				'section'  => 'amp_navbar_section',
 				'priority' => 20
+			) )
+		);
+
+		$this->wp_customize->add_control(
+			new WP_Customize_Image_Control( $this->wp_customize, 'amp_navbar_background_image', array(
+				'label'    => __( 'Header Background Image', 'amp' ),
+				'section'  => 'amp_navbar_section',
+				'priority' => 30
 			) )
 		);
 
--- a/includes/admin/class-amp-customizer.php
+++ b/includes/admin/class-amp-customizer.php
@@ -200,7 +200,6 @@ class AMP_Template_Customizer {
 		$this->wp_customize->add_control( 'amp_background_color', array(
 			'label'      => __( 'Background Color Scheme', 'amp' ),
 			'section'    => 'amp_navbar_section',
-		//	'settings'   => 'amp_theme_options[color_scheme]',
 			'type'       => 'radio',
 			'priority'   => 30,
 			'choices'    => array(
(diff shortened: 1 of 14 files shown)


includes/admin/class-amp-customizer.php
```diff
@@ -182,12 +189,46 @@ public function register_controls() {
...
+
+	/**
+	* This will output the custom WordPress settings to the live theme's WP head.
+	*
+	* Used by hook: 'wp_head'
+	*
+	* @see add_action('wp_head',$func)
+	* @since MyTheme 1.0
+	*/
+	public static function header_output() { ?>
```
mjangda: Should we remove this method?

templates/featured-image.php
```diff
@@ -0,0 +1,68 @@
...
+
+	$srcset .= '" ';
+}
+
+// Regex to find the featured image by ID
+$regex1 = sprintf( '#<amp-img.+class=("|"[^"]+ )wp-image-%d("| [^"]+").+>(\s*<.+>)*[^ ]?</amp-img>#im', get_post_thumbnail_id( $post_ID ) );
+$regex2 = sprintf( '#<figure.+id=("|"[^"]+ )attachment_%d("| [^"]+").+>(\s*<.+>)*[^ ]?</figure>#im', get_post_thumbnail_id( $post_ID ) );
+
+// Get the post content
+$content = $this->get( 'post_amp_content' ); // amphtml content; no kses
```
mjangda: We'll want to move most of this processing/logic out of the template and into the `AMP_Post_Template` class instead (to keep the template clean and allow easier filter-based overrides).

templates/featured-image.php
```diff
@@ -0,0 +1,68 @@
...
+ * If a featured image exists and an image with the same
+ * ID exists in the_content(), skip the featured image markup
+ * - Prevents duplicate images
+ */
+if ( false == preg_match( $regex1, $content ) || false == preg_match( $regex2, $content ) ) { ?>
+
+<figure class="amp-wp-article-featured-image wp-caption">
+	<amp-img alt="<?php echo wp_kses_data( $this->get( 'post_title' ) ); ?>" src="<?php the_post_thumbnail_url(); ?>" <?php echo $srcset; ?> height="<?php echo $full_height; ?>" width="<?php echo $full_width; ?>" layout="responsive" itemprop="image"></amp-img>
+	<?php if ( $caption ) { ?>
+		<p class="wp-caption-text"><?php echo $caption; ?></p>
```
mjangda: We may need to pass the `caption` through `AMP_Content`  in case is has HTML content that isn't compatible with AMP and needs to be transformed.

templates/footer-wordads.php
```diff
@@ -0,0 +1,8 @@
+<div class="adspace">
```
mjangda: We'll want to remove this as it's WP.com-specific

templates/footer.php
```diff
@@ -0,0 +1,7 @@
+<footer class="amp-wp-footer" role="contentinfo" itemscope="itemscope" itemtype="http://schema.org/WPFooter">
+	<div>
+		<h2><?php echo esc_html( $this->get( 'blog_name' ) ); ?></h2>
+		<p><?php _e( 'Powered by WordPress.com', 'amp' ); ?></p>
```
mjangda: This should be "WordPress" but we should make it so WP.com can override as necessary.

templates/meta-comments-link.php
```diff
@@ -0,0 +1,3 @@
+	<div class="amp-wp-meta amp-wp-comment-link">
+		<a class="amp-wp-comments-link-button" href="<?php echo get_comment_link() ?>" title="Leave a Comment" itemprop="discussionUrl">Leave a Comment</a>
```
mjangda: `get_comment_link` should be wrapped in `esc_url`

mjangda: `Leave a Comment` needs to be wrapped in `esc_html_e`

templates/meta-sharing.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
mjangda: We'll need to remember to take this out before merging (and add it in via jetpack-helper instead).

templates/single.php
```diff
@@ -1,24 +1,49 @@
 <!doctype html>
-<html amp <?php language_attributes(); ?>>
```
mjangda: `language_attributes` should be added back in (may have been lost in a merge)

templates/single.php
```diff
@@ -1,24 +1,49 @@
...
+<html amp>
 <head>
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=1,user-scalable=no">
 	<?php do_action( 'amp_post_template_head', $this ); ?>
-
+	<script async custom-element="amp-social-share" src="https://cdn.ampproject.org/v0/amp-social-share-0.1.js"></script>
```
mjangda: This should be removed when `meta-sharing` is.

templates/featured-image.php
```diff
@@ -0,0 +1,68 @@
...
+
+/*
+ * If a featured image exists and an image with the same
+ * ID exists in the_content(), skip the featured image markup
+ * - Prevents duplicate images
+ */
+if ( false == preg_match( $regex1, $content ) || false == preg_match( $regex2, $content ) ) { ?>
+
+<figure class="amp-wp-article-featured-image wp-caption">
+	<amp-img alt="<?php echo wp_kses_data( $this->get( 'post_title' ) ); ?>" src="<?php the_post_thumbnail_url(); ?>" <?php echo $srcset; ?> height="<?php echo $full_height; ?>" width="<?php echo $full_width; ?>" layout="responsive" itemprop="image"></amp-img>
```
mjangda: Should we use the alt for the image instead of the title here? The title may not actually accurately describe the image.

mjangda: Regardless of how we approach the alt, we should use `esc_attr` instead of `wp_kses_data` for escaping.

mjangda: `full_height` and `full_width` should be escaped.

templates/style.php
```diff
@@ -1,4 +1,48 @@
...
+
+} else {
+
+	// Default Calypso-based colors
+	$theme_color      = '#fff';
+	$text_color       = '#3d596d';
+	$muted_text_color = '#87A6BC';
+	$border_color     = '#c8d7e1';
+
+} ?>
```
mjangda: Same as with the featured image, we'll want to move this processing to `AMP_Post_Template`

includes/admin/class-amp-customizer.php
allancole: Yup, I’ll get rid of it.

templates/featured-image.php
allancole: Ahhh, good idea! I’ll work that out.

templates/featured-image.php
allancole: Will do.

templates/featured-image.php
allancole: Ohhh, right. Agreed. I’ll work that out.

templates/footer-wordads.php
allancole: Yeah, this is just a place holder, so I’ll remove it.

How will ads be supported going forward?

templates/footer.php
allancole: Sure, I’ll change it to WordPress for now.

I’m not sure how we could connect it to the WP.com overrides, but I’ll ask Ernesto who worked on that new footer override script to see what we can do.

templates/meta-comments-link.php
allancole: Will do

templates/meta-sharing.php
allancole: Gotcha. In the interest of time, can I get you take care of this part? I don’t know my way around JetPack much at all.

templates/single.php
allancole: Will do.

templates/single.php
allancole: Noted.

templates/style.php
allancole: Okay, I’ll try and work this out.

includes/admin/class-amp-customizer.php
allancole: Removed

templates/featured-image.php
allancole: Having a little trouble with this one. I’m sure I could figure this out with some time, but it’s a little outside of my theming expertise. Could you take care of this one as well?

templates/featured-image.php
allancole: Looking like i’m going to need help with this as well.

templates/footer.php
allancole: Temporarily changed to just use “WordPress”

sixhours:
This is looking really good!

Couple things I found, which may or may not be relevant:

* Getting an error in the Customizer when in regular preview:

> Warning: call_user_func_array() expects parameter 1 to be a valid callback, class 'AMP_Template_Customizer' does not have a method 'header_output' in /srv/www/wordpress-trunk/wp-includes/plugin.php on line 524

* "Background Color Scheme" offers three options; I think Light or Dark makes more sense? If we need to indicate a default, maybe Light (Default) and Dark?
* "Header Background & Link Color" reads as "Header Background Color & Header Link Color" to me. Maybe "Header Background Color & Link Color" ? I know it's wordy, but I thought it meant links in the header would be this color on first read.
* Clicking links (ie. the site title or the category links) in the AMP preview brings me to a broken screen: https://cloudup.com/ippA-iW66bP
* Sharing links look awkward when breaking to a new line (not sure you can do much about this: https://cloudup.com/caKu5seq8xo)
* "Back to top" link doesn't work.
* Line height on the site title in the footer looks a bit tall: https://cloudup.com/i_DImm0Ns8V
* The site title font size in the footer looks a bit big overall; I thought it was a link to the next article at first.

davidakennedy:
@allancole asked me to provide some feedback on this, so here goes:

This is looking good. I've tested with custom styles applied, [like this](https://github.com/Automattic/amp-wp#completely-override-css), and without.

* Getting a fatal: `Fatal error: Call to undefined function sharing_display() in /srv/www/wordpress-trunk/wp-content/plugins/amp-wp-add-amp-template/templates/meta-sharing.php on line 5`
* Looks like the HTML markup and class names have changed, which means my custom styles are broke. Not a deal breaker, but a bummer.
* It's really weird you want you go to AMP in the `Appearance > AMP` menu, you then have to go through two Customizer panels to get to options... Seems like that could be easier to get to?
* Live changes in the Customizer did not work for me. Possibly related, getting an error in the `amp.js` file: `error.js:75 No messaging channel: No messaging channel: timeout​​​`.

Let me know if I can answer any questions! Nice work!

mjangda:
Continued in #464

PR #462, opened 2016-09-02
fix replace_node_with_children

mjangda:
Continued from #449 by @Maxime2

> We need to sanitize children of removed node as if there are more than one those starting from second would be missed in sanitisation in parent's foreach loop.

Includes some tests.


mjangda:
Thanks again for the PR @Maxime2!

PR #464, opened 2016-09-07
Refresh AMP Template

mjangda:
Continuing from @allancole's #453 (now in a branch directly in this repo instead of a fork).

Introduces an updated template with support for:

- more customizer options (dark/light colour scheme);
- <strike>header image;</strike> (will be added in the future)
- featured image;
- footer.


mjangda:
Feedback from @davidakennedy:

> This is looking good. I've tested with custom styles applied, [like this](https://github.com/Automattic/amp-wp#completely-override-css), and without.
>
> * Getting a fatal: `Fatal error: Call to undefined function sharing_display() in /srv/www/wordpress-trunk/wp-content/plugins/amp-wp-add-amp-template/templates/meta-sharing.php on line 5`
> * Looks like the HTML markup and class names have changed, which means my custom styles are broke. Not a deal breaker, but a bummer.
> * It's really weird you want you go to AMP in the `Appearance > AMP` menu, you then have to go through two Customizer panels to get to options... Seems like that could be easier to get to?
> * Live changes in the Customizer did not work for me. Possibly related, getting an error in the `amp.js` file: `error.js:75 No messaging channel: No messaging channel: timeout​​​`.
>
> Let me know if I can answer any questions! Nice work!

mjangda:
Feedback from @sixhours:

> This is looking really good!
>
> Couple things I found, which may or may not be relevant:
>
> * Getting an error in the Customizer when in regular preview:
>
> `Warning: call_user_func_array() expects parameter 1 to be a valid callback, class 'AMP_Template_Customizer' does not have a method 'header_output' in /srv/www/wordpress-trunk/wp-includes/plugin.php on line 524`
>
> * "Background Color Scheme" offers three options; I think Light or Dark makes more sense? If we need to indicate a default, maybe Light (Default) and Dark?
> * "Header Background & Link Color" reads as "Header Background Color & Header Link Color" to me. Maybe "Header Background Color & Link Color" ? I know it's wordy, but I thought it meant links in the header would be this color on first read.
> * Clicking links (ie. the site title or the category links) in the AMP preview brings me to a broken screen: https://cloudup.com/ippA-iW66bP
> * Sharing links look awkward when breaking to a new line (not sure you can do much about this: https://cloudup.com/caKu5seq8xo)
> * "Back to top" link doesn't work.
> * Line height on the site title in the footer looks a bit tall: https://cloudup.com/i_DImm0Ns8V
> * The site title font size in the footer looks a bit big overall; I thought it was a link to the next article at first.

assets/js/amp-customizer-preview.js
```diff
@@ -4,15 +4,85 @@
...
+			$( '.amp-wp-article a, .amp-wp-article a:visited, .amp-wp-footer a, .amp-wp-footer a:visited' ).css( 'color', to );
+			$( 'blockquote, .amp-wp-byline amp-img' ).css( 'border-color', to );
+		} );
+	} );
+
+	// Add custom-background-image body class when background image is added.
+	wp.customize( 'amp_navbar_background_image', function( value ) {
+		value.bind( function( newVal, oldVal ) {
+			if ( newVal ) {
+				$( '.amp-wp-header' ).addClass( 'header-background-image' ).css( 'background-image', 'url(' + newVal + ')' );
```
mjangda: This URL/value should probably be escaped.

mjangda:
@allancole I wonder if we should use options instead of theme mods. Since mods are tied to the currently active theme, users would lose their AMP customizations if they ever switch themes.

allancole:
Good catch @mjangda. We’ll definitely need to use site options over theme mods. I’ll try to work this out today.

templates/header.php
```diff
@@ -1,13 +1,11 @@
-<?php $site_icon_url = $this->get( 'site_icon_url' ); ?>
```
mjangda: Sadly, there are more breaking changes than just this one :(

It's why we've including a back-compat mode (#492; automatically enabled for VIPs for now and more details to come) but I guess it makes sense to keep the file name the same here to simplify the upgrade path.

danielbachhuber: > > It's why we've including a back-compat mode (#492; automatically enabled for VIPs for now and more details to come)

Sounds good.

Issue #395, opened 2016-06-03
Inline text styles added by WP stripped out


ebiewald:
I've noticed that text in the WP content that is underlined, colored, or aligned is not appearing as such in AMP. I know this is because WP added inline styles to those tags, which isn't supported by AMP, so they're being stripped out. If the plugin could add a class to those tags in place of the inline styles, so you can still style the text as expected, that would be really awesome. 

Is there any current work around in place? Or is there any discussion of something be added to the plugin to handle all the WP wysiwyg options? If not, maybe add a disclaimer to the README? Thanks in advance for your help, it's much appreciated! 

Ellen

mjangda:
Can you share some example HTML?

ebiewald:
Of course. When you are writing a post in WordPress and you select text and underline it in the WYSIWYG, when you view the post WordPress has added a <span> and an inline style like this: 
`<span style="text-decoration:underline;">underlined text</span>` 
However that same text in the AMP version of the post (after running through the amp-wp sanitization) is just: 
`<span>underlined text</span>`
Same thing happens with colored or aligned text. It would be great if the inline style could be replaced with a class in those cases, something like:
`<span class="underlined">underlined text</span>`
Is that something you'd foresee being possible in the plugin? Is there something already there that I'm missing? It would be awesome if all the default WordPress text styles were maintained in the amp template via the plugin.

mjangda:
Yeah, it's a bit annoying that WordPress does this but I guess `<u>` was deprecated for a little while.

The plugin should probably handle this. The best option seems to be to look for `<span>`s with `style="text-decoration: underline;"` and convert them to a `<u>` instead.

mjangda:
Related: #344 -- we need to handle alignment since that uses `style` as well.

ebiewald:
The same happens for coloring text via the WYSIWYG, it adds an inline `style="color: #{color};"`

mjangda:
Outside of using a full-on CSS parser/extractor (which might be overkill), I guess we can probably get by with some regex?

coreymckrill:
Since the Editor could potentially add any arbitrary CSS rules to a style attribute, it seems like we need to find a generic way to handle it, rather than trying to make tag conversions for every possible case. It also needs to happen server-side since AMP pages [can't have any author-written JS](https://www.ampproject.org/docs/reference/spec.html).

So here's one approach:

* Prior to stripping out the style atts (or maybe during):
  * Add a unique class name to each node that has a style att
[...]

Issue #362, opened 2016-04-04
Image not centered when the 'caption' shortcode is used and center alignment is set


gnotaras:
Image with a caption text are surrounded by the `[caption ...] .... [/caption]` shortcode in the post editor. If those images are set to be aligned, eg center aligned, they still don't appear centered in the AMP page.

It seems that the `aligncenter` class (or other alignment class) is missing from the `<amp-img ... </amp-img>` block.

On the contrary, images without a caption (no `caption` shortcode) appear centered in the AMP page, if that is what has been set in the editor.

jiwicky:
you can hook into 'amp_post_template_css' and set 
```
figure {
  width: 100%;
}
```
That's what worked for me for now, till it's rolled in.

mjangda:
This is fixed.

![image](https://cloud.githubusercontent.com/assets/86105/25546893/f3943afa-2c32-11e7-9612-85479a63d4b2.png)

amedina:
Closing as it is fixed. Thanks @mjangda.

Issue #358, opened 2016-04-02
The style attribute of div elements that might surround images is removed.


gnotaras:
Hi all,

Sometimes plugins may add a `div` element, which has the `style` attribute set, around images _at runtime_. In the AMP version of the page, the `style` attribute of the surrounding `div` is stripped.

For testing purposes, manually add a `div` element around an image in the post editor. The image should be one that is attached to the post but without any caption set and without making use of the `caption` shortcode.

```
<div id="my-surrounding-div" style="width: 543px;">
    <img src="myimage.png" width="543" height="234" />
</div>
```

In the AMP version of the page the outer div becomes:

```
<div id="my-surrounding-div">
  ...
</div>
```

It would be nice if the style attribute was retained, at least its `width` property, which is important in cases in which, for example, a caption is added to the image on the fly (width is important for proper display of the caption).

George

gnotaras:
Just noticed that this issue is related to #344

westonruter:
This is fixed in 0.6.x in my testing. Inline styles get converted into style rules.

Issue #344, opened 2016-03-22
How to fix 'wp-caption' container size?


Berdych:
Original page - http://screenshotuploader.com/s/1603tjpuo
AMP page - http://screenshotuploader.com/s/160331bcq

Standard shortcode from WordPress:
`[caption id="attachment_58" align="alignright" width="300"]<img class="wp-image-58 size-medium" src="http://rehub.chernil.com.ua/wp-content/uploads/2016/02/Z5ZVY4FM28-300x200.jpg" alt="Z5ZVY4FM28" width="300" height="200" /> Several dental practitioners recommend using an antimicrobial mouth wash that will help minimize the build-from tartar...[/caption]`

westonruter:
This is fixed in 0.6.x in my testing. Inline styles get converted into style rules.

PR #470, opened 2016-09-13
Preserve Inline Styles

coreymckrill:
Adds the `AMP_Style_Sanitizer` sanitizer class, which runs before the Blacklist class, and finds all of the element nodes within the content that have a `style` attribute. For each one, it stores the contents of `style` and generates a unique class name, which it adds to the element. Then, during template rendering, it outputs these class names and their styles in the `amp-custom` stylesheet in the document head.

This preserves critical style information that gets stripped from the elements by the Blacklist class, which appears to be the root cause of the problems in #344, #358, #362, and #395.

This also includes unit tests for the new sanitizer class. Testing on the performance impact of this additional sanitizer class was done using [this script](https://github.com/coreymckrill/amp-sanitize-cases).

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -8,6 +8,10 @@ require_once( AMP__DIR__ . '/includes/sanitizers/class-amp-base-sanitizer.php' )
 class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	private $styles = array();
 
+	public function get_styles() {
+		return $this->styles;
+	}
+
 	public function sanitize() {
 		$body = $this->get_body_node();
 		$this->collect_styles_recursive( $body );
@@ -25,12 +29,12 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			if ( $style ) {
 				$style = $this->process_style( $style );
 
-				if ( $style ) {
+				if ( ! empty( $style ) ) {
 					$class_name = $this->generate_class_name( $style );
 					$new_class  = trim( $class . ' ' . $class_name );
 
 					$node->setAttribute( 'class', $new_class );
-					$this->styles[ $class_name ] = $style;
+					$this->styles[ '.' . $class_name ] = $style;
 				}
 
 				$node->removeAttribute( 'style' );
@@ -47,10 +51,10 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 
 	private function process_style( $string ) {
 		// Filter properties
-		$string = safecss_filter_attr( $string );
+		$string = safecss_filter_attr( esc_html( $string ) );
 
 		if ( ! $string ) {
-			return '';
+			return array();
 		}
 
 		// Normalize order
@@ -73,32 +77,11 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			}
 		}
 
-		if ( empty( $arr ) ) {
-			return '';
-		}
-
-		return implode( "; ", $arr ) . ';';
+		return $arr;
 	}
 
-	private function generate_class_name( $string ) {
+	private function generate_class_name( $data ) {
+		$string = maybe_serialize( $data );
 		return 'amp-wp-inline-' . md5( $string );
-	}
-
-	public function get_styles() {
-		if ( ! empty( $this->styles ) ) {
-			ob_start();
-			?>
-
-/* Inline Styles */
-<?php foreach ( $this->styles as $class_name => $style ) : ?>
-.<?php echo $class_name; ?> { <?php echo $style; ?> }
-<?php endforeach; ?>
-
-<?php
-
-			return ob_get_clean();
-		}
-
-		return '';
 	}
 }
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -36,6 +36,8 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 					$node->setAttribute( 'class', $new_class );
 					$this->styles[ $class_name ] = $style;
 				}
+
+				$node->removeAttribute( 'style' );
 			}
 		}
 
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -27,6 +27,8 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			$class = $node->getAttribute( 'class' );
 
 			if ( $style ) {
+				$style = $this->process_style( $style );
+
 				$class_name = $this->generate_class_name( $style );
 				$new_class  = trim( $class . ' ' . $class_name );
 
@@ -43,8 +45,19 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 		}
 	}
 
+	private function process_style( $string ) {
+		// Filter properties
+		$string = safecss_filter_attr( $string );
+
+		// Normalize order
+		$arr = array_map( 'trim', explode( ';', $string ) );
+		sort( $arr );
+
+		return implode( ";\n", $arr ) . ';';
+	}
+
 	private function generate_class_name( $string ) {
-		return 'amp-style-' . md5( $string );
+		return 'amp-inline-style-' . md5( $string );
 	}
 
 	public function append_styles() {
@@ -52,7 +65,10 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 
 /* Inline Styles */
 <?php foreach ( $this->styles as $class_name => $style ) : ?>
-.<?php echo $class_name; ?> { <?php echo $style; ?> }
+.<?php echo $class_name; ?> {
+	<?php echo $style; ?>
+
+}
 <?php endforeach; ?>
 
 		<?php
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -1,0 +1,60 @@
+<?php
+
+require_once( AMP__DIR__ . '/includes/sanitizers/class-amp-base-sanitizer.php' );
+
+/**
+ * Collects inline styles and outputs them in the amp-custom stylesheet.
+ */
+class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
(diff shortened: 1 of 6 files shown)


tests/test-amp-style-sanitizer.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	private function generate_class_name( $string ) {
+		return 'amp-inline-style-' . md5( $string );
+	}
+
+	private function wrap_style( $array ) {
+		ob_start(); ?>
+
+/* Inline Styles */
```
mjangda: We're duplicating the `generate_class_name` and style output behaviours between the sanitizer and the tests, which means that if the implementation in the sanitizer changes but has a bug, we could end up not catching it.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -0,0 +1,102 @@
...
+
+		if ( empty( $arr ) ) {
+			return '';
+		}
+
+		return implode( ";\n\t", $arr ) . ';';
+	}
+
+	private function generate_class_name( $string ) {
+		return 'amp-inline-style-' . md5( $string );
```
mjangda: It would be nice to prefix this as `amp-wp-` for consistency

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -0,0 +1,102 @@
...
+			foreach( $overflow as $index => $rule ) {
+				unset( $arr[ $index ] );
+			}
+		}
+
+		if ( empty( $arr ) ) {
+			return '';
+		}
+
+		return implode( ";\n\t", $arr ) . ';';
```
mjangda: I don't think we should worry about whitespace.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -0,0 +1,102 @@
...
+ */
+class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
+	private $styles = array();
+
+	public function sanitize() {
+		$body = $this->get_body_node();
+		$this->collect_styles_recursive( $body );
+
+		if ( ! empty( $this->styles ) ) {
+			add_action( 'amp_post_template_css', array( $this, 'append_styles' ), 99 );
```
mjangda: Any ideas on how we could skip the need to hook into actions here and pass back the generated styles some other way? (We've tried to keep Sanitizers pretty focused on just DOM manipulation and kept them free from actions/filters for various reasons.)

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -0,0 +1,102 @@
...
+
+			if ( $style ) {
+				$style = $this->process_style( $style );
+
+				if ( $style ) {
+					$class_name = $this->generate_class_name( $style );
+					$new_class  = trim( $class . ' ' . $class_name );
+
+					$node->setAttribute( 'class', $new_class );
+					$this->styles[ $class_name ] = $style;
```
mjangda: Why don't we just remove the style attribute as well?

includes/sanitizers/class-amp-style-sanitizer.php
coreymckrill: Fixed with 43c6972cb17b977083f5d71cee0ce93c575b7816

includes/sanitizers/class-amp-style-sanitizer.php
coreymckrill: The reasoning is that this helps avoid duplication of generated style rules in the stylesheet. Without it, `color: #fff;` and `color :  #fff;` would count as two separate styles.

It is an edge case, so if you feel strongly that it's not worth the extra processing, I'm happy to remove it.

includes/sanitizers/class-amp-style-sanitizer.php
coreymckrill: Since the Blacklist sanitizer is already designed to remove attributes, including `style`, I thought it best followed the single responsibility principle to leave the removal part out of the Style sanitizer.

There could be a scenario where someone changes the order of the sanitizers and the Blacklist runs first, but if that happens, the Style sanitizer won't do anything anyway, since there won't be any remaining style attributes.

includes/sanitizers/class-amp-style-sanitizer.php
coreymckrill: I initially had the same thought. I wanted to pass the data to `AMP_Post_Template` and add the action there, but I couldn't figure out a way to add data to the class without changing the access of some of the methods from private to public. Do you think it would be worth modifying `AMP_Post_Template` to make this possible?

tests/test-amp-style-sanitizer.php
coreymckrill: Good point. Fixed with 28092c4de741f9ee05fd5d8320d9ead4112eedac

includes/sanitizers/class-amp-style-sanitizer.php
coreymckrill: Another approach could be to _not_ remove the style attribute in Blacklist, and just handle everything in the Style class. It seems like this could be confusing to other devs, though, since the Blacklist is designed specifically for removing tags and attributes...

includes/sanitizers/class-amp-style-sanitizer.php
mjangda: 👍

includes/sanitizers/class-amp-style-sanitizer.php
mjangda: Not sure I understand your example. What would the original `style` attribute look like in this edge case?

AMP has limits on how much CSS you can include (although, not 100% sure if whitespace counts) so we should try to keep it as slim as possible. Plus, no need to send extra bytes if we can avoid it.

includes/sanitizers/class-amp-style-sanitizer.php
mjangda: What if we just added `get_styles` to the Base Sanitizer?

includes/sanitizers/class-amp-style-sanitizer.php
mjangda: I think in this particular case, it's probably fine for the Style Sanitizer to remove the attribute once it's done.

tests/test-amp-style-sanitizer.php
mjangda: 👍

includes/sanitizers/class-amp-style-sanitizer.php
coreymckrill: > What would the original style attribute look like in this edge case?

Example markup:

```
<p>I really like the color <span style="color: #00ff00;">green</span>.</p>

<p>Miguel really likes <span style="color:      #00ff00;">green</span> also.</p>
```

Without whitespace normalizing, the resulting `amp-custom` styles would be:

```
/* Inline Styles */
.amp-wp-inline-style-04a69db6d1a130e8e077df8a467e51dc {
	color:     #00ff00;
}
.amp-wp-inline-style-bb01159393134c225a1df0d44226c3d2 {
	color: #00ff00;
}
```

With whitespace normalizing:

```
/* Inline Styles */
.amp-wp-inline-style-bb01159393134c225a1df0d44226c3d2 {
	color: #00ff00;
}
```

includes/sanitizers/class-amp-style-sanitizer.php
coreymckrill: 👍 Fixed with 72883957e6b5f92dee5a8cad5cbb53bdabb7a67a

includes/sanitizers/class-amp-style-sanitizer.php
coreymckrill: @mjangda is this what you're thinking of? bcc1a8537e0aa9d0074b1ceb2aca2fabf87a8e10

includes/sanitizers/class-amp-style-sanitizer.php
mjangda: Oh, I think we're talking about different things :) 

My comment (a bit unclear, I realize) was just referring to the addition of `\n\t`; not the use of `trim` (which is a good idea as you've outlined).

includes/sanitizers/class-amp-style-sanitizer.php
mjangda: Nice, exactly!

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -0,0 +1,107 @@
...
+
+		if ( empty( $arr ) ) {
+			return '';
+		}
+
+		return implode( ";\n\t", $arr ) . ';';
+	}
+
+	private function generate_class_name( $string ) {
+		return 'amp-wp-inline-style-' . md5( $string );
```
mjangda: Minor: let's remove `-style` from the class name.

includes/sanitizers/class-amp-style-sanitizer.php
coreymckrill: Oh! Gotcha. The `\n\t` just makes it easier to read the style rules when they're printed in the doc header, but I'm fine taking them out.

includes/sanitizers/class-amp-style-sanitizer.php
coreymckrill: Fixed with 4d8fc50

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -0,0 +1,104 @@
...
+
+		return implode( "; ", $arr ) . ';';
+	}
+
+	private function generate_class_name( $string ) {
+		return 'amp-wp-inline-' . md5( $string );
+	}
+
+	public function get_styles() {
+		if ( ! empty( $this->styles ) ) {
```
mjangda: Let's actually move have this just return the style array and have our `amp_post_template_add_styles` callback handle the rendering. That way, we maintain consistency with how `get_scripts()` works and avoid the use of `ob_` (and likely makes the tests a bit cleaner too).

coreymckrill: Yeah, this does sound a bit cleaner. Ok, stay tuned.

coreymckrill: 914ecc7

mjangda:
Just two small notes and I think we're good to merge!

coreymckrill:
Cool! Here are the changes: 5d69505

mjangda:
💥 💥 💥 

Thanks @coreymckrill!

Issue #465, opened 2016-09-07
Don't use amp-youtube for YouTube Playlists


mjangda:
We support oEmbeds for YouTube playlists in the following forms:
- `https://www.youtube.com/playlist?list=PL9tY0BWXOZFvR0U9MnfnQSMym0fFgL5cT`
- `https://www.youtube.com/embed/videoseries?list=PL9tY0BWXOZFvR0U9MnfnQSMym0fFgL5cT`

However, AMP's `amp-youtube` element currently only handles individual videos. We should just show these as links or use `amp-iframe` instead.

Originally reported by @k-atul in #463

PR #473, opened 2016-09-14
Youtube playlists: Use the standard oembed handler

coreymckrill:
From #465 : 

> AMP's `amp-youtube` element currently only handles individual videos.

Playlists are therefore output as just links. However, the standard oembed handler for Youtube URLs takes a URL and outputs an iframe. Since AMP has a sanitizer for handling iframes, we can simply output the iframe when the URL is for a playlist instead of an individual video, and then the iframe sanitizer will take care of the rest.

### DIFF
--- a/includes/embeds/class-amp-youtube-embed.php
+++ b/includes/embeds/class-amp-youtube-embed.php
@@ -64,6 +64,12 @@ class AMP_YouTube_Embed_Handler extends AMP_Base_Embed_Handler {
 	}
 
 	public function oembed( $matches, $attr, $url, $rawattr ) {
+		// Use the standard oembed handler if it's a playlist
+		$parsed_url = wp_parse_url( $url );
+		if ( isset( $parsed_url['path'] ) && false !== strpos( $parsed_url['path'], 'playlist' ) ) {
+			return wp_oembed_get( $url );
+		}
+
 		return $this->shortcode( array( $url ) );
 	}
 
--- a/tests/test-amp-youtube-embed.php
+++ b/tests/test-amp-youtube-embed.php
@@ -38,6 +38,11 @@ class AMP_YouTube_Embed_Test extends WP_UnitTestCase {
 				'https://www.youtube.com/playlist?list=PLtn57NsUdLz7Phy4bzX0f3d5_eAQQ7FCA' . PHP_EOL,
 				'<p><iframe width="660" height="371" src="https://www.youtube.com/embed/videoseries?list=PLtn57NsUdLz7Phy4bzX0f3d5_eAQQ7FCA" frameborder="0" allowfullscreen></iframe></p>' . PHP_EOL,
 			),
+
+			'url_playlist_broken' => array(
+				'https://www.youtube.com/playlist?asdf=PLtn57NsUdLz7Phy4bzX0f3d5_eAQQ7FCA' . PHP_EOL,
+				'<p>https://www.youtube.com/playlist?asdf=PLtn57NsUdLz7Phy4bzX0f3d5_eAQQ7FCA</p>' . PHP_EOL,
+			),
 		);
 	}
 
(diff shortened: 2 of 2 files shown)


includes/embeds/class-amp-youtube-embed.php
```diff
@@ -64,6 +64,12 @@ public function shortcode( $attr ) {
 	}
 
 	public function oembed( $matches, $attr, $url, $rawattr ) {
+		// Use the standard oembed handler if it's a playlist
+		$parsed_url = wp_parse_url( $url );
+		if ( isset( $parsed_url['path'] ) && false !== strpos( $parsed_url['path'], 'playlist' ) ) {
+			return wp_oembed_get( $url );
```
mjangda: What happens if the iframe sanitizer runs before the YouTube sanitizer?

coreymckrill: The Youtube class is an embed handler, rather than a sanitizer, and from what I can tell, [the embeds always run before the sanitizers](https://github.com/Automattic/amp-wp/blob/0.3.3/includes/class-amp-content.php#L35-L41). Am I getting that wrong?

mjangda: Oh no, you're right -- I got mixed up :)

Is there a way we can just bail out of our custom oembed handler and fallback directly to core's oembed handler instead?

mjangda:
It would be nice to add some tests for this.

Also, the referenced ticket also has this example URL `https://www.youtube.com/embed/videoseries?list=PL9tY0BWXOZFvR0U9MnfnQSMym0fFgL5cT` that we should catch as well?

includes/embeds/class-amp-youtube-embed.php
coreymckrill: > Is there a way we can just bail out of our custom oembed handler and fallback directly to core's oembed handler instead? (i.e. instead of calling `wp_oembed_get` in our handler)

I think this is as close to that as we can get. All of the methods that match the URL to an oembed provider and fetch the provider data live in the `WP_oEmbed` class, and `wp_oembed_get` is a wrapper that does everything we need, including ensuring that class is loaded and available and fetching the data. I haven't run across a global function that specifically handles the Youtube playlists.

coreymckrill:
> Also, the referenced ticket also has this example URL `https://www.youtube.com/embed/videoseries?list=PL9tY0BWXOZFvR0U9MnfnQSMym0fFgL5cT` that we should catch as well?

I thought about this. Youtube URLs with the `videoseries` path [don't actually match a provider pattern](https://github.com/WordPress/WordPress/blob/4.6.1/wp-includes/class-oembed.php#L58-L63) or get rendered by the standard oembed handler, they just appear as links. So if we wanted to support it, we'd need to add a filter to the `oembed_providers` hook, assuming Youtube also supports it on their end. My reasoning was that if the non-AMP version of the post doesn't render it, the AMP version doesn't need to either.

> It would be nice to add some tests for this.

Good idea! Stay tuned.

coreymckrill:
Ok, I've added a unit test for the embed handler. I looked over the existing tests for the iframe sanitizer and it looks like the markup that gets generated for playlists already has good coverage.

coreymckrill:
Added another unit test for a non-functional playlist URL.

It looks like my tests are failing in Travis CI because the computed width and height for the iframe (based on `$content_width` I think) are different in that environment than they are on my local machine. I'm investigating...

includes/embeds/class-amp-youtube-embed.php
coreymckrill: Update: we might be able to skip all of this by just removing `playlist` from [the regex](https://github.com/Automattic/amp-wp/blob/0.3.3/includes/embeds/class-amp-youtube-embed.php#L8) in the Youtube embed handler. That seems to have the same effect, because then the pattern doesn't match and it just gets caught later by the standard oembed. I'm guessing now that this is what you were suggesting above, @mjangda ;)

Thoughts? We could just close this pull request and I could open another one with a one-line change.

mjangda: That's a very good idea :)

Thoughts on how we could catch this version of the playlist URL as well: `https://www.youtube.com/embed/videoseries?list=PL9tY0BWXOZFvR0U9MnfnQSMym0fFgL5cT`?

coreymckrill:
Closing in favor of https://github.com/Automattic/amp-wp/pull/489

PR #480, opened 2016-09-16
Amp faster image

gititon:


mjangda:
Can we restore tabs (instead of spaces) for indentation in all the files that were using them? (WordPress coding standards uses tabs over spaces.)

tests/test-amp-image-dimension-extractor.php
```diff
@@ -2,8 +2,8 @@
 
 define( 'AMP_IMG_DIMENSION_TEST_VALID_FILE', dirname( __FILE__ ) . '/assets/wordpress-logo.png' );
 define( 'AMP_IMG_DIMENSION_TEST_INVALID_FILE', dirname( __FILE__ ) . '/assets/not-exists.png' );
-define( 'IMG_350', 'https://i0.wp.com/placehold.it/350x150.png');
-define( 'IMG_1024', 'https://i0.wp.com/placehold.it/1024x768.png');
+define( 'IMG_350', 'http://i0.wp.com/placehold.it/350x150.png');
+define( 'IMG_1024', 'http://i0.wp.com/placehold.it/1024x768.png');
```
mjangda: Why does Travis need `http` URLs?

tests/test-amp-image-dimension-extractor.php
```diff
@@ -54,9 +54,9 @@ function test__normalize_url( $source_url, $expected_url ) {
 class AMP_Image_Dimension_Extractor__By_Downloading__Test extends WP_UnitTestCase {
 
 	function test__valid_image_file() {
-		$source = 'https://i0.wp.com/placehold.it/350x150.png';
+		$source = 'https://placehold.it/350x150.png';
```
mjangda: This [comment](https://github.com/Automattic/amp-wp/pull/480#pullrequestreview-1423633) ("The commit message doesn't match the changes here. The use of Photon seems to be dropped here.") was meant to go here (Still getting used to Github's new review interface :))

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -47,73 +71,157 @@ public static function normalize_url( $url ) {
...
+
+    /**
+     * Fetch dimensions of remote images
+     *
+     * @param array $urls_to_fetch Image src urls to fetch
+     * @param array $images Array to populate with results of image/dimension inspection
+     */
+    private static function fetch_images($urls_to_fetch, &$images, $mode) {
+        //User Faster Image when PHP version supports it (it contains a closure that could not be ported to 5.2)
+        if ( $mode === 'synchronous' || strnatcmp( phpversion(), '5.3.0' ) < 0 ) {
```
mjangda: Are we reasonably confident that `curl_multi_` and any of the other fasterimage dependencies are available in a stock install of 5.3+?

includes/lib/class-faster-image-b52f1a8-faster-image.php
```diff
@@ -0,0 +1,189 @@
...
+ * @version 0.01
+ */
+class Faster_Image_B52f1a8_Faster_Image
+{
+    /**
+     * The default timeout
+     *
+     * @var int
+     */
+    protected $timeout = 10;
```
mjangda: `10s` seems like a long default. Should we adjust to something lower?

includes/lib/class-faster-image-b52f1a8-faster-image.php
```diff
@@ -0,0 +1,189 @@
...
+        curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
+        curl_setopt($ch, CURLOPT_BUFFERSIZE, 256);
+        curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, 0);
+        curl_setopt($ch, CURLOPT_SSL_VERIFYHOST, 0);
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+        curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, $this->timeout);
+        curl_setopt($ch, CURLOPT_TIMEOUT, $this->timeout);
+
+        #  Some web servers require the useragent to be not a bot. So we are liars.
+        curl_setopt($ch, CURLOPT_USERAGENT, 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_8_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/27.0.1453.110 Safari/537.36');
```
mjangda: I'm assuming this is from the original fasterimage library? We should be good citizens and not lie and update this :)

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -17,54 +17,90 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
...
-				$new_tag = 'amp-anim';
-			} else {
-				$new_tag = 'amp-img';
-			}
-
-			$new_node = AMP_DOM_Utils::create_node( $this->dom, $new_tag, $new_attributes );
-			$node->parentNode->replaceChild( $new_node, $node );
-		}
-	}
+	public function sanitize()
```
mjangda: Some of the coding standards in this file (and other AMP plugin -- i.e. not `lib`) have been reverted to not follow WordPress Coding Standards (e.g. tabs, brace location, spacing around parenthesis, etc.). These should be fixed.

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -17,54 +17,90 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
...
+            // Determine which images need their dimensions determined/extracted
+            if ( !$node->hasAttribute('width') || ! $node->hasAttribute('height')) {
+                $need_dimensions[ $node->getAttribute('src') ][] = $node;
+            } else {
+                $have_dimensions[ $node->getAttribute('src') ][] = $node;
+            }
+        }
+        $this->determine_dimensions($need_dimensions);
+
+        //Calling function twice to save overhead of merging arrays just to send single array to function
```
mjangda: I'm not sure I understand this. Can you please clarify?

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -2,20 +2,44 @@
 
 class AMP_Image_Dimension_Extractor {
 	static $callbacks_registered = false;
+    const STATUS_FAILED_LAST_ATTEMPT = 'failed';
+    const STATUS_IMAGE_EXTRACTION_FAILED= 'failed';
 
-	static public function extract( $url ) {
+    static public function extract( $urls ) {
```
mjangda: We're changing the signature of a `public` function and later filters, which will break for any users calling this directly or hooking in to those filters.

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -2,20 +2,44 @@
...
+            $url = self::normalize_url($url);
+        }
 
-		return apply_filters( 'amp_extract_image_dimensions', false, $url );
+		$dimensions = apply_filters( 'amp_extract_image_dimensions', $urls );
+
+        if ( has_filter('amp_extract_image_dimensions') ) {
+            return $dimensions;
+        } else {
+            return self::modify_return_value_for_tests_that_disable_extraction_filter($dimensions);
```
mjangda: This isn't ideal. Why should this code need to worry about tests?

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -47,73 +71,157 @@ public static function normalize_url( $url ) {
...
+                    // Ensure we're the only one trying to extract dimensions for this image at the moment so we don't duplicate requests
+                    $transient_lock_name = sprintf( 'amp_lock_%s', $url_hash );
+                    if ( false !== get_transient( $transient_lock_name ) ) {
+                        $all_dimensions[$url] = false;
+                        continue;
+                    } else {
+                        $urls_to_fetch[$url]['url'] = $url;
+                        $urls_to_fetch[$url]['transient_name'] = $transient_name;
+                        $urls_to_fetch[$url]['transient_lock_name'] = $transient_lock_name;
+                        set_transient($transient_lock_name, 1, MINUTE_IN_SECONDS);
```
mjangda: We're getting pretty deeply nested here. Any way we can flatten the conditional blocks a bit?

tests/test-amp-image-dimension-extractor.php
```diff
@@ -2,6 +2,8 @@
 
 define( 'AMP_IMG_DIMENSION_TEST_VALID_FILE', dirname( __FILE__ ) . '/assets/wordpress-logo.png' );
 define( 'AMP_IMG_DIMENSION_TEST_INVALID_FILE', dirname( __FILE__ ) . '/assets/not-exists.png' );
+define( 'IMG_350', 'http://i0.wp.com/placehold.it/350x150.png');
+define( 'IMG_1024', 'http://i0.wp.com/placehold.it/1024x768.png');
```
mjangda: Any way we could reasonably/easily mock the remote fetches so that tests can work offline (and not rely on an external service)?

gititon:
Will do. When PHPstorm said it was using WordPress coding standards, I'd
assumed it was using tabs.

gititon:
It doesn't/shouldn't. I had changed them to http while investigating why
tests were failing and forgot to change them back. I will!

gititon:
I didn't realize at first why certain tests were failing on Travis when
they passed locally and made a few changes to see if I could narrow it
down. Apologies if I used a confusing commit message (can't see what it was
at the moment as I'm on my phone)

gititon:
Thanks, Mo! I've replied to a few of your comments and will action on them
as soon as I can. I have to get going, but I will look at the rest of your
comments/feedback as soon as I'm able. (It's my daughters birthday tomorrow
so I may not be able to get to it then)

Have a great night,
Mike

includes/sanitizers/class-amp-img-sanitizer.php
gititon: The comment precedes the invocation of the same function twice, each time with a different array as its argument. My first thought was to merge the two arrays and call the function only once to make the code more tidy, but I realized it was not worth the overhead of merging the two arrays. I left the comment in case anyone else had the same idea. I'll remove it, though, since it probably only creates confusion.

includes/lib/class-faster-image-b52f1a8-faster-image.php
gititon: Agreed, I've changed it to 4 for starters. I was inclined to lower it right to 2. What do you think?

includes/lib/class-faster-image-b52f1a8-faster-image.php
gititon: Yes, it came that way. I've replaced with the User Agent as reported by PHP.

includes/sanitizers/class-amp-img-sanitizer.php
gititon: That's what I get for trusting PHPStorm when it told me it would use WordPress standard formatting! I've made sure the files conform to WordPress coding standards.

includes/utils/class-amp-image-dimension-extractor.php
gititon: I removed the nested conditionals. continue to the rescue!

includes/utils/class-amp-image-dimension-extractor.php
gititon: The functions normally invoked by the amp_extract_image_dimensions filter modify the format/content of the array returned by AMP_Image_Dimension_Extractor::extract. When the filter is disabled, extract() returns an array that code downstream from it isn't expecting/can't properly parse. modify_return_value_for_tests_that_disable_extraction_filter() normalized the array in such a scenario so that it would work properly downstream. However, revisiting this code now, though the function name is descriptive, it's possible the filter may be removed/disabled for some other reason, so I will give the function a more general name.

tests/test-amp-image-dimension-extractor.php
gititon: I restored the use of Photon. (I'd checked/tried many different things when investigating test failures and forgot to put this one back afterward).

includes/utils/class-amp-image-dimension-extractor.php
gititon: I hadn't remembered doing anything specific to install multi curl after installing curl (I didn't need to install anything else). When Googling how to install multi curl didn't turn up any separate instructions from how to install curl, I performed a simple experiment to confirm that the availability of multi curl was tied to the availability of curl:

if ( function_exists('curl_exec') ) {
	echo "CURL EXISTS\n";
} else {
	echo "CURL DOES NOT EXIST\n";
}

if ( function_exists('curl_multi_exec') ) {
        echo "MULTI CURL EXISTS\n";
} else {
        echo "MULTI CURL DOES NOT EXIST\n";
}

Sure enough, when I disable the curl PHP extension multi curl becomes unavailable as well, and likewise when I re-enable it. Therefore, it doesn't seem like there are dependencies beyond curl.

That said, because curl itself is a dependency, and we don't want installations using php 5.3+ without curl to have errors, I've added a check for the existence of multi_curl_exec() and fall back to the original FastImage library if it isn't present.

includes/utils/class-amp-image-dimension-extractor.php
gititon: Would you prefer that I add and call an additional method instead, e.g. extract_multi(), leaving the original extract() and its signature intact? It wouldn't be of much use because of the way image dimension extraction works with my changes, but it would still be there. I'd assumed the only client of AMP_Image_Dimension_Extractor would be the AMP-WP plugin since it's a utility included specifically for use by the plugin.

My changes only require that the amp_extract_image_dimensions filter have 1 parameter. Should I change it back to 2, regardless, in case anyone is hooking into the filter from elsewhere?

I had removed the extract_from_attachment_metadata filter for the reasons I explained in my blog post. Should I restore it for the sake of anyone who may have been hooking into it?

I'd appreciate your guidance regarding the above since these are areas where more WordPress specific experience would be especially useful. That said, I like to think I learn quickly!

tests/test-amp-image-dimension-extractor.php
gititon: I'll give this some thought and push the changes I've made in response to your comments in the meantime.

tests/test-amp-image-dimension-extractor.php
gititon: I finally debugged why Travis CI was failing on the PHP 5.2 tests with the changes I tried to push this morning. Sure enough, it was because I'd changed the test image urls back to https, and fopen() (used by the original FastImage library) didn't like it.

"AMP_Image_Dimension_Extractor__By_Downloading__Test::test__multiple_valid_image_files
fopen(): Unable to find the wrapper "https" - did you forget to enable it when you configured PHP?"

It looks like this is happening because Travis' PHP 5.2 environment/installation doesn't load the php_openssl extension (I didn't make an explicit deploy just to confirm this, but that's what the sources I'm reading about the https wrapper error are suggesting). 

I've reverted the test urls to http so that the tests pass in Travis' PHP 5.2 environment.

includes/lib/class-faster-image-b52f1a8-faster-image.php
mjangda: 4 seems more reasonable to handle slower server connections / larger files

includes/lib/class-faster-image-b52f1a8-faster-image.php
mjangda: I think we should probably either use the same user agent that WordPress' HTTP library does (see WordPress HTTP API) or maybe something specific to the plugin (like `amp-wp, v0.3`)?

includes/utils/class-amp-image-dimension-extractor.php
mjangda: Right, so it's unlikely that people are calling `extract` directly, but they would be hooking into `amp_extract_image_dimensions` ([we're doing it ourselves on WP.com](https://github.com/Automattic/amp-wp/blob/master/wpcom-helper.php#L106)) which is now changed from two parameters (`bool|array $dimensions, string $url`) to one param (`array $urls`). That's definitely going to break stuff.

```
-		return apply_filters( 'amp_extract_image_dimensions', false, $url );
+		$dimensions = apply_filters( 'amp_extract_image_dimensions', $urls );
```

So yeah, we may need a new method + filter (since the existing method+filter didn't really plan for parallel/batch processing).

gititon: Thanks, Mo. I began working on this this morning but had some questions I wanted to run by you.

At first I thought you meant we needed to preserve the pre-existing functionality somehow and were asking me to change the name of the modified filter and restore the original one. I wondered, though, how the original one, which took an array and a single image URL as arguments, would be invoked, since I changed the plugin to process arrays of image URLs.

It just occurred to me that you may simply be saying we need to change the name of the filter I modified so that any existing downstream hooks that rely on its original parameters/behavior do not get invoked when the filters for it are applied.

Can you please confirm whether you mean we simply need to rename it (and make corresponding changes to the hooks in the plugin that use it) or if the original filter signature/single url functionality needs to be restored as well even though the modified plugin does not invoke/use it?

Thanks,
Mike

mjangda: > we need to change the name of the filter I modified so that any existing downstream hooks that rely on its original parameters/behavior do not get invoked when the filters for it are applied.

Yes, exactly that. Ideally we'd still be able to provide backwards compatibility here and the old filter continues to work as well. Maybe we can can run each image/dimension through the old filter after the fact?

gititon: To be sure, when you say "[m]aybe we can can run each image/dimension through the old filter after the fact?" you're suggesting that after we process the images in batch via my code changes, we iterate over $urls and call apply_filters( 'amp_extract_image_dimensions', false, $url ) on each one? I imagine that would require that publishers who wanted to leverage the original filter would therefore have to suppress/remove the new batch one, like how the WP.com "optimizations" currently suppress the existing callbacks on the original filter. Otherwise, if the batch filter were allowed to run first like it would by default, when the original filter kicked in for each url afterward there would be nothing for any custom callback added to it to do, since the image dimensions would have already been fetched (or marked failed) by the batch filter.

For this reason, it seems any publisher who has added custom callbacks to the original filter (and would like to see them continue to offer some utility) would have to modify code in any event to suppress the new batch filter from running first in order to ensure their custom callbacks weren't rendered inert/useless (by virtue of the batch filter already determining all image dimensions).

Where a code change would be required in any event (by a publisher who added custom callbacks to the original filter and wanted to continue to use them) by either:

A) Suppressing the new batch filter to ensure image dimensions haven't already been fetched so any callbacks added to the original filter actually have something to do; OR

B) Modifying their custom callback(s) to work with the batch filter instead of the original filter

wouldn't it be a win from a maintenance/complexity standpoint to require the latter instead of the former? Assuming a publisher was rational and only added a custom callback to the original filter that extracted dimensions from an image at a URL via some different mechanism (like any callback added to this filter is presumably supposed to), it seems it would be a non breaking change in any event since the new batch filter would simply do the image extraction work unless/until code is changed to suppress it.

Aside from the issue of the batch filter negating the usefulness of calling the original filter for each url unless/until the batch filter is explicitly suppressed/removed, for the majority of cases where publishers haven't added custom callbacks to the original filter that they're interested in maintaining, and simply leave the plugin as is, wouldn't calling the original filter for each url after the batch filter ran result in needlessly iterating over the urls in the great majority of installations/implementations?

To recap, where a code change would seemingly be required by publishers who wanted to retain a custom callback's functionality either way, it seems ideal from a maintenance/complexity standpoint not to carry the original filter forward.

If you agree, I can update wpcom-helper.php wpcom_amp_extract_image_dimensions_add_custom_callbacks() to use the batch filter as well.

If you would prefer to see the original filter invoked for each url subsequent to the invocation of the batch filter despite the above considerations, I am happy to do that as well.

Please let me know your thoughts.

Thanks,
Mike

mjangda: Thanks for your detailed thoughts on this! The main reason I brought it up is that backwards compatibility in the WordPress ecosystem is often preferred over the cost of complexity (there are a lot of reasons for this and many pros/cons on both sides).

In any case, your arguments seem reasonable and it makes sense to remove the existing filter and introduce the new batch one.

> If you agree, I can update wpcom-helper.php wpcom_amp_extract_image_dimensions_add_custom_callbacks() to use the batch filter instead of the original one.

Yes, it'd be great to see how the existing callbacks we have look with the batch approach instead.

includes/utils/class-amp-image-dimension-extractor.php
gititon: I was able to get rid of this.

includes/lib/class-faster-image-b52f1a8-faster-image.php
gititon: I updated it to use to "amp-wp v<plugin version number>" as the user agent.

includes/utils/class-amp-image-dimension-extractor.php
gititon: You're welcome! I know we talked about it on Slack, but just to close the loop here in case anyone else has been following along, I pushed these changes last week.

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -18,51 +18,85 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
...
-
-			// Try to extract dimensions for the image, if not set.
-			if ( ! isset( $new_attributes['width'] ) || ! isset( $new_attributes['height'] ) ) {
-				$dimensions = AMP_Image_Dimension_Extractor::extract( $new_attributes['src'] );
-				if ( is_array( $dimensions ) ) {
-					$new_attributes['width'] = $dimensions[0];
-					$new_attributes['height'] = $dimensions[1];
-				}
+			// Determine which images need their dimensions determined/extracted.
+			if ( ! $node->hasAttribute( 'width' ) || ! $node->hasAttribute( 'height' ) ) {
```
mjangda: An img with attributes `width="" height=""` will end up in the else clause. (Although, I guess this is an existing bug and maybe better solved in a separate issue/PR).

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -2,18 +2,22 @@
...
+	static public function extract( $urls ) {
 		if ( ! self::$callbacks_registered ) {
 			self::register_callbacks();
 		}
 
-		$url = self::normalize_url( $url );
-		if ( false === $url ) {
-			return false;
+		foreach ( $urls as &$url ) {
+			$url = self::normalize_url( $url );
```
mjangda: `normalize_url` can return `false`, which we should probably just ignore (i.e. don't pass through to extract dimensions).

tests/test-amp-image-dimension-extractor.php
```diff
@@ -49,92 +51,160 @@ function test__normalize_url( $source_url, $expected_url ) {
...
+	function test__invalid_image_file() {
+		$sources = array(
+			AMP_IMG_DIMENSION_TEST_INVALID_FILE,
+		);
+		$expected = array(
+			AMP_IMG_DIMENSION_TEST_INVALID_FILE => false,
+		);
 
-		$dimensions = AMP_Image_Dimension_Extractor::extract_from_attachment_metadata( false, $source );
+		$sources = array_fill_keys( $sources, false );
```
mjangda: Rather than calling `array_fill_keys` for the test cases, we should just set up the `sources` variables with the data format that the `extract_by_*` method expects (i.e. remove as much logic from tests as possible).

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -18,51 +18,85 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
...
 			}
+		}
+		$this->determine_dimensions( $need_dimensions );
 
-			// Final fallback when we have no dimensions.
-			if ( ! isset( $new_attributes['width'] ) || ! isset( $new_attributes['height'] ) ) {
-				$new_attributes['width'] = isset( $this->args['content_max_width'] ) ? $this->args['content_max_width'] : self::FALLBACK_WIDTH;
-				$new_attributes['height'] = self::FALLBACK_HEIGHT;
+		$this->adjust_and_replace( $need_dimensions );
+		$this->adjust_and_replace( $have_dimensions );
```
mjangda: So, reading through this a few times, I wonder if we could simplify by doing two separate passes: once for images with dimensions, then fetch any missing dimensions and repeat for remaining images.

Right now, it feels we've optimized for the `needs_dimension` use case and are probably doing multiple unnecessary loops (loop 1: figure out which nodes need dimensions; loop 2+: [nested] process need_dimensions; loop 3+: [nested] process have_dimensions).

We could probably combine 1 and 3 and do both things in the same pass:

```
for ( img_nodes ) {
    if ( has_dimensions ) {
        process_node
    } else {
        array_push( node, remaining_nodes )
    }
}

if ( remaining_nodes ) {
    determine dimensions for remaining_nodes

    for ( remaining_nodes ) {
        process_node
    }
}
```

it's also possible that I may be overlooking something :) Thoughts?

includes/lib/class-faster-image-b52f1a8-faster-image.php
```diff
@@ -0,0 +1,194 @@
...
+     * @var int
+     */
+    protected $timeout = 4;
+
+	/**
+	 * Backup user agent
+	 *
+	 * @var string
+	 */
+	protected $user_agent = 'wp-amp, v' . AMP__VERSION;
```
gititon: I actually thought about doing this when I initially made the change, but I considered the way most other user agents work. Web browsers don't include anything in their user agents that indicate who the particular user is or where the requested resources are going. I was concerned that by doing so for WP-AMP, we could be putting WordPress publishers who hot link to images they don't host (of which I imagine there are many) at a disadvantage, e.g. by facilitating the blocking of their site from accessing the images via their user agent. Still, I suppose that is better than the possibility of the WP-AMP user agent being blocked altogether. I've added the site url and made it filterable.

includes/sanitizers/class-amp-img-sanitizer.php
gititon: I fixed this.

tests/test-amp-image-dimension-extractor.php
gititon: Agreed, done!

includes/utils/class-amp-image-dimension-extractor.php
gititon: Done!

PR #484, opened 2016-09-22
Make amp page work in secondary languages with WPML

dgwatkins:
AMP pages are only working in the default language with the WPML plugin installed.
This problem was reported in WPML forums and we found a workaround that I have included here.

https://wpml.org/forums/topic/wp-amp-pages-only-in-english-but-not-in-other-langauge/

Is this solution acceptable to merge?

If you need a WPML subscription to test this let me know and I will arrange it.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -101,7 +101,7 @@ function amp_add_post_template_actions() {
 }
 
 function amp_prepare_render() {
-	add_action( 'template_redirect', 'amp_render' );
+	add_action( 'template_redirect', 'amp_render', 9 );
 }
 
 function amp_render() {


mjangda:
Thanks for the PR @dgwatkins!

Can you share more details about how the priority change fixes the original issue (i.e. what is WPML doing that is requiring the priority change)?

Also, in the forum thread you noted "we need to add higher priority, for example 9", although, 9 is lower than the default priority of 10. Was that just a typo?

dgwatkins:
I have realized that this has to be fixed on the WPML side of things.
Im sorry for the noise.

dgwatkins:
Just a quick update:
Next version of WPML will include a proper fix for this issue.
It has been tested successfully already and its pending QA.

Issue #354, opened 2016-03-28
Fallback for iframe/audio/video


mjangda:
These AMP elements should have a fallback in case the media item fails to load (or is incompatible with AMP, e.g. non-https media).

westonruter:
I'm not sure what we'd put in as a fallback. A link to the media?

westonruter:
Fallbacks for iframes and media elements have been implemented in #1913 and #1861.

Issue #350, opened 2016-03-24
AMP validation errors: amp-audio requires https?


drewhackney:
I’ve got a couple blog posts that are giving AMP errors as reported by the user who received a notification from Google:

http://studio-lu.net/2012/03/19/welcome-home-cowboy-bob/amp/#development=1

"http://studio-lu.net/2012/03/19/welcome-home-cowboy-bob/amp/:314:38 Invalid URL protocol 'http:' for attribute 'src' in tag 'amp-audio > source'. (see https://www.ampproject.org/docs/reference/extended/amp-audio.html)"

http://studio-lu.net/2012/02/19/in-search-of-my-grandfathers-laugh/amp/#development=1

"http://studio-lu.net/2012/02/19/in-search-of-my-grandfathers-laugh/amp/:307:38 Invalid URL protocol 'http:' for attribute 'src' in tag 'amp-audio > source'. (see https://www.ampproject.org/docs/reference/extended/amp-audio.html)"

Thanks!

mjangda:
#310 improved this a bit and we can improve further in #354.

Issue #288, opened 2016-02-29
Wordpress post embeds fail "wp-embedded-content"


laserjobs:
Posts with Wordpress Embeds fail and give a error of "Unkown syntax errors" in the Search Console. 
`Unkown syntax errors:	src=http://www.godairyfree.org/recipes/smores-ice-cream-sandwiches/embed#?secret=xxxxxxxx`

webpage code:
```
<p><iframe class="wp-embedded-content" sandbox="allow-scripts" security="restricted" style="display:none;" src="http://www.godairyfree.org/recipes/smores-ice-cream-sandwiches/embed#?secret=xxxxxx" data-secret="xxxxxxxxx" width="600" height="338" title="Embedded WordPress Post" frameborder="0" marginwidth="0" marginheight="0" scrolling="no"></iframe></p>
```

AMP code:

```
<amp-iframe class="wp-embedded-content amp-wp-enforced-sizes" sandbox="allow-scripts" src="http://www.godairyfree.org/recipes/smores-ice-cream-sandwiches/embed#?secret=xxxxxxx" width="600" height="338" frameborder="0" sizes="(min-width: 600px) 600px, 100vw"><div placeholder="" class="amp-wp-iframe-placeholder"></div></amp-iframe>
```

laserjobs:
here is a link of the problem:
http://www.godairyfree.org/news/dairy-free-super-bowl-recipes/amp

laserjobs:
How can we get around not using https?
https://www.ampproject.org/docs/guides/amp_replacements.html#display-an-iframe

mjangda:
In terms of rendering it, there's not much we can do since `https` is a strict requirement for AMP

We could, however, include a `fallback` element that has a link to the embed though.

laserjobs:
So it looks like it would validate with SSL but I am sure like myself there is many not using https

https://www.godairyfree.org/recipes/smores-ice-cream-sandwiches/embed#?secret=xxxxxxxxxxxFailed to load resource: net::ERR_SSL_PROTOCOL_ERROR

[...]

PR #493, opened 2016-09-29
Embed fallbacks

coreymckrill:
This is a response to #354. Other related issues are #130, #288, #316, and #350.

The goal of this is to provide a consistent user experience when embedded media can't load because it's not from an https source. A few different options were discussed to make this happen, but this is what we settled on:

> Remove the (embedded) element completely, and replace it with a blockquote or some other block-level element containing the source URL and perhaps a message like “Could not load.”

Here are the types of embeds that should be covered:
- [x] Audio
- [x] Video
- [x] iFrame

### DIFF
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -106,9 +106,22 @@ abstract class AMP_Base_Sanitizer {
 		$protocol = strtok( $src, ':' );
 
 		if ( $protocol === $src && ( $https_required || $force_https ) ) {
-			// src has relative protocol, ie //example.com/asdf, so add https.
-			$src = set_url_scheme( $src, 'https' );
-		} else if ( 'https' !== $protocol ) {
+			if ( 0 === strpos( $src, '//' ) ) {
+				// src has relative protocol, ie //example.com/asdf, so add https
+				$src = set_url_scheme( $src, 'https' );
+			} else if ( 0 === strpos( $src, '/' ) ) {
+				// src is URL relative to the site root
+				$src = home_url( $src );
+			} else {
+				// src is URL relative to current URI
+				global $wp;
+				$src = home_url( trailingslashit( $wp->request ) . $src );
+			}
+
+			$protocol = strtok( $src, ':' );
+		}
+
+		if ( 'https' !== $protocol ) {
 			// Check if https is required
 			if ( $https_required ) {
 				// Remove the src. Let the implementing class decide what do from here.
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -121,4 +121,25 @@ abstract class AMP_Base_Sanitizer {
 
 		return $src;
 	}
+
+	protected function create_fallback_node( $content, $container_el = 'blockquote', $attributes = array() ) {
+		$defaults = array(
+			'class' => 'amp-wp-fallback',
+		);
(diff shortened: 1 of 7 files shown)


coreymckrill:
So far I've added fallbacks for audio and video. iFrame is still in the works.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -102,14 +102,18 @@ public function add_or_append_attribute( &$attributes, $key, $value, $separator
 	 * @return string
 	 */
 	public function maybe_enforce_https_src( $src, $force_https = false ) {
+		$https_required = isset( $this->args['require_https_src'] ) && true === $this->args['require_https_src'];
 		$protocol = strtok( $src, ':' );
-		if ( 'https' !== $protocol ) {
+
+		if ( $protocol === $src && ( $https_required || $force_https ) ) {
+			// src has relative protocol, ie //example.com/asdf, so add https.
+			$src = set_url_scheme( $src, 'https' );
```
mjangda: What would happen with relative URLs, like `/path/to/file.mp3`?

includes/sanitizers/class-amp-video-sanitizer.php
```diff
@@ -32,21 +42,46 @@ public function sanitize() {
...
+							esc_url( array_shift( $orig_src ) )
+						)
+					);
+
+					$fallback_node->appendChild( $fallback_content );
+
+					$node->parentNode->replaceChild( $fallback_node, $node );
+				} else {
+					$node->parentNode->removeChild( $node );
+				}
```
mjangda: This block of code (and the similar ones across the other sanitizers) could probably be abstracted to a new method?

includes/sanitizers/class-amp-video-sanitizer.php
```diff
@@ -32,21 +42,46 @@ public function sanitize() {
...
+						$this->dom,
+						'blockquote',
+						array(
+							'class' => 'amp-wp-fallback amp-wp-video-fallback'
+						)
+					);
+
+					$fallback_content = $this->dom->createDocumentFragment();
+					$fallback_content->appendXML( sprintf(
+							wp_kses( __( 'Could not load <a href="%s">video</a>.', 'amp' ), array( 'a' => array( 'href' => true ) ) ),
```
mjangda: Should we switch the statement from an error ("Could not load") to more of a action ("Watch video"). It's not entirely accurate to say that we couldn't load the video. (Note: this applies to all the sanitizers)

coreymckrill: We could use "Watch video" and "Listen to audio", but I'm not sure what we'd do for the iframe one...

includes/sanitizers/class-amp-video-sanitizer.php
coreymckrill: Good idea. 2365364

includes/sanitizers/class-amp-base-sanitizer.php
coreymckrill: Alright, I took a stab at handling relative paths along with the relative protocol scenario: f7dafe5

includes/sanitizers/class-amp-video-sanitizer.php
coreymckrill: Also, I'm not sure "Could not load" is really misleading, since the AMP code _is_ preventing non-https sources from loading.

westonruter:
Fallbacks for iframes and media elements have been implemented in #1913 and #1861.

PR #531, opened 2016-10-12
sanitise language_attributes()

Maxime2:
Some plugins, e.g. Add Meta Tags, may alter language_attributes() using 'language_attributes' filter and adding XML-related attributes which are disallowed (see https://www.ampproject.org/docs/reference/spec#required-markup )

So we need to sanitise this as well. This patch solves this problem.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -59,6 +59,7 @@ function amp_init() {
 	add_post_type_support( 'post', AMP_QUERY_VAR );
 
 	add_filter( 'request', 'amp_force_query_var_value' );
+	add_filter( 'language_attributes', 'amp_language_attributes', PHP_INT_MAX, 2 );
 	add_action( 'wp', 'amp_maybe_add_actions' );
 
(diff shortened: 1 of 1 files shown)


mjangda:
Thanks for the PR.

The reason we added `language_attributes` was to make sure `lang` and `dir` attributes were added for non-English sites.

I think instead of trying to strip out `prefix` and `xmlns` attributes, it might be better just to add `lang` and `dir` attributes manually (especially since there may be other blacklisted attributes as well). Thoughts?

Maxime2:
Yes we seems may just put lang and dir attributes manually having <html amp> specified.
Though, there could be more valid attributes, especially those accessibility related, https://www.w3.org/TR/html5/semantics.html#the-html-element

JoryHogeveen:
I would say this could have its own function/filter that adds the default `lang` and `dir`. Other plugins can hook into that filter to add their own AMP specific parameters.

Or use the currenct `$doctype` in `language_attributes()` parameter to pass a value like `amphtml` so plugin's can validate before adding anything.

mjangda:
This was fixed in #537

PR #533, opened 2016-10-13
update readme for meta hooks

troyxmccall:
reflects updated independent header_meta and footer_meta filters in https://github.com/Automattic/amp-wp/blob/6ec40876bc7e0f56b665d161d3542811189883ff/templates/single.php

ref: https://github.com/Automattic/amp-wp/commit/6ec40876bc7e0f56b665d161d3542811189883ff


mjangda:
Nice, thanks Troy.

PR #536, opened 2016-10-13
Include the max-width with the iframe sanitizer

srtfisher:
Introduced with 40660765f66b50b72f162a34b6fffa1d1a60dadc, unless we pass in the to all sanitizers the width of the iframe (when used with a percentage) [will break](https://github.com/Automattic/amp-wp/blob/master/includes/sanitizers/class-amp-base-sanitizer.php#L45) when sanitizing the width since `content_max_width` is never passed in.

### DIFF
--- a/includes/class-amp-post-template.php
+++ b/includes/class-amp-post-template.php
@@ -232,7 +232,8 @@ class AMP_Post_Template {
 				 'AMP_Video_Sanitizer' => array(),
 				 'AMP_Audio_Sanitizer' => array(),
 				 'AMP_Iframe_Sanitizer' => array(
-					 'add_placeholder' => true,
+					'add_placeholder' => true,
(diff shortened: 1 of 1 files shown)


srtfisher:
The other solution so that all sanitizers get access to the content width would be to change how the base sanitizer determines the content width.

mjangda:
The content_max_width is already passed to all sanitizers. There was a bug introduced in 0.4.1 that caused it not to be correctly passed in but fixed in e06e9bc913deb06b50e17599145ca33cd6294c3a. Can you see if that solves your issue?

srtfisher:
Ah yes, that would solve it for us. Sorry about not realizing that was fixed at first, going off the current version on VIP.

mjangda:
@srtfisher FYI, the fix has been deployed to WP.com

srtfisher:
Sweet!

PR #540, opened 2016-10-14
Allow the new `html_tag_attributes` to be filtered

JoryHogeveen:
Related: https://github.com/Automattic/amp-wp/pull/537


mjangda:
Thanks for the PR!

You can currently modify the html attributes via the `amp_post_template_data` filter. Do you think there's a significant benefit to adding this new one as well?

JoryHogeveen:
Hi @mjangda 
Sorry, didn't notice that filter.
I guess there is no real need for a dedicated filter then. Up to you!

PR #549, opened 2016-10-26
Fix template/meta-author on older versions of WP

jasongill:
Versions of WordPress prior to 4.2.0 did not include the get_avatar_url function, and will fail on meta-author.php. It appears that there was code in place using "if function_exists" to check for the existence of get_avatar_url, but it was not placed properly in the file. This change corrects the code so that get_avatar_url is only used if the function exists, which was the original intent of the if statement.


mjangda:
Nice catch, although, I would highly recommending updating to a newer version of WordPress soon; v4.2 is very old :)

PR #575, opened 2016-11-27
Update class-amp-img-sanitizer.php

silverma:
If only one image dimension is set, the img sanitizer would ignore it and use the actual image sizes instead. This proposed change checks if only one dimension was set, and if so, calculates the other.

### DIFF
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -42,6 +42,14 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 					$new_attributes['width'] = $dimensions[0];
 					$new_attributes['height'] = $dimensions[1];
 				}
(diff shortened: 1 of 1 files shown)


westonruter:
See also #695.

westonruter:
@silverma Is this still relevant after #979?

silverma:
> @silverma Is this still relevant after #979?

@westonruter for some reason I'm not able to reply to the comment above so replying here. I haven't looked at this in quite a while, so I'm not sure. If I remember correctly the specific problem was when the attribute wasn't included, not that it was there and set to 0. Also, the proposed change suggested to auto calculate the missing attribute. If those cases are covered, I'm guessing it should be good!

westonruter:
I'm pretty sure this has been fixed in 0.7.1 via #1117: 

https://github.com/Automattic/amp-wp/blob/90acca403917c2e7b7f1b5985b03bdee7ffc6697/includes/sanitizers/class-amp-img-sanitizer.php#L187-L210

PR #576, opened 2016-11-27
Update class-amp-img-sanitizer.php

silverma:
If only one dimension is set, don't ignore it; use that dimension to calculate the other.


mjangda:
Duplicate of #575

PR #588, opened 2016-12-16
in readme, replace references to '/classes/' with '/includes/'

combs:
It seems like this plugin uses `/includes/` as a path now and not `/classes/`. In this PR I updated readme.md to reflect this. Apologies if I am misunderstanding something.


mjangda:
Thanks for the PR. Those are example paths that might exist in your own theme/plugins, not the paths for the `amp-wp` plugin.

combs:
ah ha! OK, thank you.

PR #605, opened 2017-01-09
Added AMP Cache utilities to update AMP cache on post updates.

### LINKED ISSUE #598: Send update pings to AMP Caches
On post update and delete. See ~https://developers.google.com/amp/cache/update-ping~ https://developers.google.com/amp/cache/update-cache

delputnam:
@mjangda this is to handle issue #598 

Fixes #598

### DIFF
--- a/includes/admin/class-amp-cache-utilities.php
+++ b/includes/admin/class-amp-cache-utilities.php
@@ -225,10 +225,11 @@ class AMP_Cache_Utilities {
 	 */
 	public static function get_amp_cache_url_for_post( $post ) {
 		$amp_cache_resource_path = self::get_amp_cache_path_for_post( $post );
-		if ( $amp_cache_resource_path ) {
-			return self::$amp_cache_url_base . '/' . ltrim( $amp_cache_resource_path, '/' );
+		if ( ! $amp_cache_resource_path ) {
+			return false;
 		}
-		return false;
+		
+		return self::$amp_cache_url_base . '/' . ltrim( $amp_cache_resource_path, '/' );
 	}
 
 	/**
@@ -242,10 +243,11 @@ class AMP_Cache_Utilities {
 	 */
 	public static function get_amp_cache_update_url_for_post( $post ) {
 		$amp_cache_resource_path = self::get_amp_cache_path_for_post( $post );
-		if ( $amp_cache_resource_path ) {
-			return self::$amp_cache_update_url_base . '/' . ltrim( $amp_cache_resource_path, '/' );
+		if ( ! $amp_cache_resource_path ) {
+			return false;
 		}
-		return false;
+		
+		return self::$amp_cache_update_url_base . '/' . ltrim( $amp_cache_resource_path, '/' );
 	}
 
 	/**
@@ -259,10 +261,11 @@ class AMP_Cache_Utilities {
 	 */
 	public static function get_amp_cache_url_for_resource( $url, $content_type, $scheme = null ) {
 		$amp_cache_resource_path = self::get_amp_cache_path_for_url( $url, $content_type );
-		if ( $amp_cache_resource_path ) {
-			return self::$amp_cache_url_base . '/' . ltrim( $amp_cache_resource_path, '/' );
+		if ( ! $amp_cache_resource_path ) {
+			return false;
 		}
-		return false;
+		
+		return self::$amp_cache_url_base . '/' . ltrim( $amp_cache_resource_path, '/' );
 	}
 
 	/**
@@ -276,10 +279,11 @@ class AMP_Cache_Utilities {
 	 */
 	public static function get_amp_cache_update_url_for_resource( $url, $content_type, $scheme = null ) {
 		$amp_cache_resource_path = self::get_amp_cache_path_for_url( $url, $content_type );
-		if ( $amp_cache_resource_path ) {
-			return self::$amp_cache_update_url_base . '/' . ltrim( $amp_cache_resource_path, '/' );
+		if ( ! $amp_cache_resource_path ) {
+			return false;
 		}
-		return false;
+
+		return self::$amp_cache_update_url_base . '/' . ltrim( $amp_cache_resource_path, '/' );
 	}
 
 	/**
--- a/includes/admin/class-amp-cache-utilities.php
+++ b/includes/admin/class-amp-cache-utilities.php
@@ -51,6 +51,11 @@ abstract class AMP_Cache_Utilities {
 	 * need to update the AMP Cache when a post is updated or about to be deleted.
 	 */
 	public static function amp_add_cache_update_actions() {
+		// Skip AMP Cache updates when importing.
+		if ( defined( 'WP_IMPORTING' ) && ( true === WP_IMPORTING ) ) {
+			return;
+		}
+
 		// Hooking this to the post_updated action so this will fire any time a post
 		// is updated in any way (including status transitions).
 		add_action( 'post_updated',  array( 'AMP_Cache_Utilities', 'post_updated' ), 10, 3 );
--- a/includes/admin/class-amp-cache-utilities.php
+++ b/includes/admin/class-amp-cache-utilities.php
@@ -1,0 +1,292 @@
+<?php
+/**
+ * In addition to handling updates to the AMP cache when a post is created
+ * or updated, this class is meant to be a general toolbox for the AMP cache
+ * and this has a number of utility functions for use with the AMP Cache.
+ * 
+ * The original driver for this class was `amp_add_cache_update_actions()` 
+ * which adds hooks to ping the AMP cache when a post is updated or deleted.
+ *
+ * But, this class also includes a number of functions that could be used for other
+ * features within the plugin. For example, if there was a need to cache images
+ * or other resources (say, fonts), then `get_amp_cache_url_for_resource()`
+ * and `get_amp_cache_update_url_for_resource()` can be used to generate
+ * the correct URLs for retrieving and caching the resource respectively.
+ *
+ * This is the reason that not all of the functions included in the class directly
+ * relate to updating the AMP cache on a post update/delete.
+ *
+ * Note: The reason this is an `abstract` class is because all of the 
+ * properties and methods are static and thus, the class may be used without
+ * being instantiated. `abstract` is just a reminder to call the functions
+ * statically and that you don't need to isntantyiate an instance of the
+ * class to use it. So, while theres's nothing preventing this class from
+ * being extended, that wasn't the primary driver for making it abstract.
+ */
+abstract class AMP_Cache_Utilities {
+
+	/**
+	 * These are the valid types of content that can be cached:
+	 * 
+	 * 'c' - content (a post)
+	 * 'i' - an image
+	 * 'r' - a resource (like a font)
+	 *
+	 * See: https://developers.google.com/amp/cache/overview
+	 */
+	static $amp_valid_content_types = array( 'c', 'i', 'r' );
+
+	/**
+	 * This is the base URL of the AMP cache CDN.
+	 */
+	static $amp_cache_url_base = 'https://cdn.ampproject.org';
+
+	/**
+	 * This is the base URL of the AMP cache ping-update URL
+	 */
+	static $amp_cache_update_url_base = 'https://cdn.ampproject.org/update-ping';
+
+	/**
+	 * Hooks the 'post_updated' and 'before_delete_post' actions to determine if we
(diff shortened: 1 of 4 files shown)


includes/admin/class-amp-cache-utilities.php
```diff
@@ -0,0 +1,251 @@
+<?php
+abstract class AMP_Cache_Utilities {
+
+	static $amp_valid_content_types = array( 'c', 'i', 'r' );
```
mjangda: I'm surprised this passed tests under PHP 5.2. I was under the impression that properties couldn't be defined as arrays in the class definition. Maybe it's different for static properties?

includes/admin/class-amp-cache-utilities.php
```diff
@@ -0,0 +1,251 @@
+<?php
+abstract class AMP_Cache_Utilities {
```
mjangda: Out of curiosity, why is this an abstract class?

includes/admin/class-amp-cache-utilities.php
delputnam: `static` is fine.  You can't have a `const` array.

mjangda: Ah, that makes sense :)

includes/admin/class-amp-cache-utilities.php
delputnam: It doesn't have to be, but this way, someone won't accidentally instantiate multiple copies of it wastefully.

includes/admin/class-amp-cache-utilities.php
```diff
@@ -0,0 +1,251 @@
...
+
+		// determine $content_type, if not specified
+		if ( null == $content_type ) {
+			$post_type = get_post_type( $post );
+			switch ( $post_type ) {
+				case 'attachment':
+					$content_type = 'i';
+					break;
+				case 'post':
+					$content_type = 'c';
```
mjangda: We should handle custom post types here. Could just put the `post` and `default` cases together.

includes/admin/class-amp-cache-utilities.php
```diff
@@ -0,0 +1,251 @@
...
+			return false;
+		}
+
+		// If there is no host part to this URL, return failure.
+		if ( ! isset( $parsed_url['host'] ) ) {
+			return false;
+		}
+
+		// If there is no scheme specified in the parameter list and this is a protocol
+		// relative URL, then we can't figure out whether this should be https or http.
```
mjangda: If its protocol-relative, we can probably assume that both http or https are support and just pick one?

includes/admin/class-amp-cache-utilities.php
```diff
@@ -0,0 +1,251 @@
...
+	 * 
+	 * @param  int/WP_Post	$post 	The post to update in the AMP cache.
+	 * @return bool       			true if the cache was updated, false otherwise.
+	 */
+	public static function do_amp_update_ping( $post ) {
+		$update_ping_url = self::get_amp_cache_update_url_for_post( $post );
+		if ( ! $update_ping_url ) {
+			return false;
+		}
+		$response = wp_remote_get( $update_ping_url );
```
mjangda: We should set a low timeout, maybe even `1s`?

tests/test-amp-cache-utilities.php
```diff
@@ -0,0 +1,509 @@
...
+				false,
+			),
+		);
+	}
+
+	/**
+	 * @dataProvider get_amp_cache_path_for_url_data
+	 * @group amp-cache-path-test	
+	 */
+	public function test_get_amp_cache_path_for_url( $data, $expected_ache_path ) {
```
mjangda: Typo? (missing `c` in `expected_ache_path`)

tests/test-amp-cache-utilities.php
```diff
@@ -0,0 +1,509 @@
...
+				true,
+			),
+			'tag_fail' => array(
+				array(
+					'factory_type' => 'tag',
+					'http_response_code' => 204,
+				),
+				false,
+			),
+			'user_fail' => array(
```
mjangda: We don't really need to worry about users and tags since they're not "post" objects.

includes/admin/class-amp-cache-utilities.php
delputnam: Yes. This is something I meant to ask your opinion about. I was torn between setting a low timeout and making this a non-blocking `GET` given that there is absolutely nothing we can do if this fails anyway.

The only failures I could simulate involved general network failures (can't look up in DNS, etc.). Every `GET` returns 204 even if the URL doesn't exist. Thus, I was wondering if we even need to wait around to see what the response is.

includes/admin/class-amp-cache-utilities.php
delputnam: I was thought about this too and was wondering if I should set up a filter to allow only certain custom post types to be included since some custom post types might not be appropriate for the AMP cache.

includes/admin/class-amp-cache-utilities.php
delputnam: The AMP Cache doesn't treat http and https the same.  For example, if you ping https://cdn.ampproject.org/update-ping/c/s/example.com, but the URL is actually http://example.com, then nothing gets cached.

So, I suppose that another option is to try both .../c/s/example.com and .../c/example.com instead of failing.

tests/test-amp-cache-utilities.php
delputnam: Yep, I just wanted to make sure that they would be rejected correctly by the function and not actually pinged if someone tried to use this code to cache them. Maybe I should have called this test `reject_user_post_type_success` to be more explicit about what I'm trying to accomplish.

includes/admin/functions.php
```diff
@@ -20,6 +21,9 @@ function amp_init_customizer() {
 
 	// Add a link to the Customizer
 	add_action( 'admin_menu', 'amp_add_customizer_link' );
+
+	// Hook the functions to update the AMP cache on post update/deletion.
+	AMP_Cache_Utilities::amp_add_cache_update_actions();
```
delputnam: Good point.  The file is now `required` and the function hooked on `admin_init` at the bottom of amp.php.

includes/admin/class-amp-cache-utilities.php
```diff
@@ -0,0 +1,251 @@
...
+	static $amp_valid_content_types = array( 'c', 'i', 'r' );
+	static $amp_cache_url_base = 'https://cdn.ampproject.org';
+	static $amp_cache_update_url_base = 'https://cdn.ampproject.org/update-ping';
+
+	/**
+	 * Hooks the 'post_updated' and 'before_delete_post' actions to determine if we
+	 * need to update the AMP Cache when a post is updated or about to be deleted.
+	 */
+	public static function amp_add_cache_update_actions() {
+		// Hooking this to the post_updated action so this will fire any time a post
```
delputnam: Yep.  Done.

includes/admin/class-amp-cache-utilities.php
mjangda: Yeah, we don't really need to wait for the response. Probably a good to just do both, i.e. make it non-blocking AND set a low timeout (since not all transports support non-blocking requests).

includes/admin/class-amp-cache-utilities.php
mjangda: We can just use the `post_type_supports` function to check if the post type supports it.

Alternatively, AMP has a function that can check at the individual level: `post_supports_amp`

tests/test-amp-cache-utilities.php
mjangda: Users aren't "post types", so calling the function with a user object would actually be using the function incorrectly.

includes/admin/class-amp-cache-utilities.php
delputnam: Thanks, I'm using `post_type_supports`.

tests/test-amp-cache-utilities.php
delputnam: Ah...got it.  Now I understand your point.  Thanks!

tests/test-amp-cache-utilities.php
delputnam: fixed.

tests/test-amp-cache-utilities.php
delputnam: I removed those test cases.

includes/admin/class-amp-cache-utilities.php
delputnam: For now, I'm assuming 'http' for protocol relative URLs.

includes/admin/class-amp-cache-utilities.php
delputnam: If `post_supports_amp()` returns true, then I'm assuming the content type should be 'c' regardless of the post_type.

includes/admin/class-amp-cache-utilities.php
delputnam: Yep.  `timeout` is now 1 second and `blocking` is set to `false`.

includes/admin/class-amp-cache-utilities.php
```diff
@@ -0,0 +1,251 @@
...
+	 * @param  int/WP_Post 	$post_id     The post that was being updated
+	 * @param  WP_Post 		$post_after  A copy of the post after the update.
+	 * @param  WP_Post		$post_before A copy of the post before the update.
+	 * @return bool         true is the cache was updated, false otherwise.
+	 */
+	public static function post_updated( $post_id, $post_after, $post_before ) {
+		// if post_status is 'publish' or was 'publish' but now is not, update cache
+		if ( ( 'publish' == $post_after->post_status ) ||
+			( ( 'publish' != $post_after->post_status ) && ( 'publish' == $post_before->post_status ) ) ) {
+			return self::do_amp_update_ping( $post_id );
```
delputnam: I'm using `post_supports_amp` to check for this now.

includes/admin/class-amp-cache-utilities.php
```diff
@@ -0,0 +1,251 @@
...
+	 * need to update the AMP Cache when a post is updated or about to be deleted.
+	 */
+	public static function amp_add_cache_update_actions() {
+		// Hooking this to the post_updated action so this will fire any time a post
+		// is updated in any way (including status transitions).
+		add_action( 'post_updated',  array( 'AMP_Cache_Utilities', 'post_updated' ), 10, 3 );
+		
+		// Hooking the call to update the cache *before* the post is updated in case we need
+		// to access any metadata for future functionality.
+		add_action( 'before_delete_post',  array( 'AMP_Cache_Utilities', 'do_amp_update_ping' ) );
```
delputnam: Now using `post_supports_amp` to make sure the post supports AMP. Also returning early if `post_status` is not `publish` when the `before_delete_post` action is called.

amedina:
@mjangda @delputnam What happened with this effort? Is it close to landing? Could we get it to the finish line, or what is missing?

davisshaver:
@amedina @mjangda Bumping this. What could I do to help?

Also this PR encompasses content updates. I believe I ran into an issue this morning with template updates. See https://github.com/ampproject/amphtml/issues/13708.

A `Purge All` is probably not feasible but it is a thought...

westonruter:
@davisshaver I suggest merging the commits from this PR into a new branch off of `develop` and then open a new PR with the updated state to move it forward anew.

NicktheGeek:
The `update-ping` method is deprecated. This is the new method and it requires an RSA key to validate the request.
https://developers.google.com/amp/cache/update-cache

Wondering if this can still be implemented via the plugin though.

westonruter:
@nickFalcone Thanks for the follow-up.

> This is the new method and it requires an RSA key to validate the request. ¶ Wondering if this can still be implemented via the plugin though.

I think so. There is a [`openssl_public_encrypt()`](http://php.net/manual/en/function.openssl-public-encrypt.php) function in PHP.

archon810:
@westonruter Do we have any idea when this could be implemented? We're looking to launch AMP with the official WP plugin, but would definitely need the update functionality. Should we implement it ourselves or wait for the plugin to do so? Private/public keys are already generated.

westonruter:
@archon810 We don't have this on the immediate radar for working on. If you would like to contribute a PR with what you come up with that would be great.

westonruter:
This needs to be revisited. The `update-ping` endpoint was deprecated and the plugin codebase has changed a lot since this PR was opened. So I'm closing this, expecting a new PR will use parts of the old one.

Issue #564, opened 2016-11-04
Trigger a doing it wrong when calling is_amp_endpoint() too early


johnbillion:
Similar to WordPress core's conditional functions, the `is_amp_endpoint()` function will return false data if it's called too early. This is [documented in the function's docblock](https://github.com/Automattic/amp-wp/blob/6d300c63a86c7f4f5437352cdc057944292d240d/includes/amp-helper-functions.php#L39).

Calling it too early should function should trigger a `_doing_it_wrong()` call in the same way that core's conditional functions do.

LeleDev:
Is for this reason that it always return false if called from plugins?

mjangda:
Yep, that's a good idea.

johnbillion:
We can into this issue because we were calling `is_amp_endpoint()` from within a callback on the `amp_init` hook and it was returning `false` all the time.

PR #612, opened 2017-01-17
Trigger a doing it wrong when calling is_amp_endpoint() too early

delputnam:
@mjangda This is to address issue #564.


mjangda:
👍

Issue #124, opened 2016-01-25
Add Support for Jetpack / WP.com related posts


mjangda:

gibrown:
Took a quick look at how to implement this. We currently only render related posts on the client side for Jetpack. I think with the implied caching of AMP that it would be better to render the related posts on the page load. Links of course should get changed to point to the amp version of the pages.

Doing this requires adding a new method of rendering the related posts in https://github.com/Automattic/jetpack/blob/master/modules/related-posts/jetpack-related-posts.php

mjangda:
We could possible use `amp-list` (I even wrote an example for it :) https://github.com/Automattic/amp-wp#custom-embed-handler) although, it expects the endpoint to return data in a specific format (namely, it wants a JSON object with an `items` key).

gibrown:
Interesting, ya that could be made to work. Right now related posts passes the query through to the site where various filters are applied rather than hitting our public api directly.

The benefit of embedding it in the original html is that we can probably use any existing filters or customizations that the user has applied. But maybe those customizations wouldn't make sense anyways with AMP.

gibrown:
Oh, also we should add some sort of click tracking for clicking on the related posts, but I think this is a general problem with Jetpack Related posts.

gibrown:
Per https://www.alexkras.com/google-may-be-stealing-your-mobile-traffic/ this should maybe be implemented in general. When a related posts system exists, then use that. When it doesn't exist, then show the X most recent posts other than the current post. Try to keep the user on the current site.

westonruter:
This should rather be added to the Jetpack plugin itself. See #1021.

westonruter:
For the Jetpack issue specifically for Related Posts compatibility with AMP, see https://github.com/Automattic/jetpack/issues/9556

PR #617, opened 2017-01-24
Add/jetpack related hosts

delputnam:
@mjangda, This PR is to address issue #124.

To render related posts using Jetpack's engine, on AMP pages, this PR uses amp-list to access a site's related posts feed and renders it in a manner similar to the related posts built into Jetpack.

Other (non-Jetpack) related posts modules could also take advantage of this embed handler by hooking the `amp_related_posts_register_embed` and `amp_related_posts_unregister_embed` actions to add their own callbacks and filters.

There are a number of other hooks and filters available:
* `amp_related_posts_scripts` can be used to filter scripts to be added
* `amp_related_posts_options` can add/remove/update the options array
* `amp_related_posts_styles` filter can be used to modify the custom css
* `amp_related_posts_url` to generate the related posts data endpoint
* `amp_related_posts_template` to modify the template used to generate the related posts output

`amp_modify_jetpack_related_post_links()` was added to `jetpack-helper.php`to modify the related posts URLs so that they returned AMP URLs.

`AMP_Blacklist_Sanitizer` was modified to allow mustache template variable replacement in the `href` attribute of `<a>` tags.

For testing in a local development environment, here is a simple plugin I put together that will allow Jetpack's Related Posts module to run without a connection to WordPress.com: https://gist.github.com/delputnam/e6bb2f596f2995ae42f00b20957b5b07

### DIFF
--- a/includes/class-amp-post-template.php
+++ b/includes/class-amp-post-template.php
@@ -19,6 +19,7 @@ require_once( AMP__DIR__ . '/includes/embeds/class-amp-gallery-embed.php' );
 require_once( AMP__DIR__ . '/includes/embeds/class-amp-instagram-embed.php' );
 require_once( AMP__DIR__ . '/includes/embeds/class-amp-vine-embed.php' );
 require_once( AMP__DIR__ . '/includes/embeds/class-amp-facebook-embed.php' );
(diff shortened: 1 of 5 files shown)


westonruter:
This should rather be added to the Jetpack plugin itself. See #1021.

PR #618, opened 2017-01-24
Also need the amp-list script for the related posts example

delputnam:
@mjangda,

While working on the related posts embed module, I noticed that we also need to specify `amp-list-0.1.js` in `get_scripts()`.


mjangda:
Good catch!

PR #624, opened 2017-02-02
Use the correct class name when creating a Stream

mjangda:
Should be the properly prefixed name: `Stream_17b32f3_Stream`


mjangda:
Test failures are nightly only and unrelated.

PR #637, opened 2017-02-16
Add support for amp-dailymotion

amedina:
Add amp-dailymotion support.
cc: @mjangda, @pietergreyling, @pbakaus

### DIFF
--- a/includes/embeds/class-amp-dailymotion-embed.php
+++ b/includes/embeds/class-amp-dailymotion-embed.php
@@ -1,0 +1,101 @@
+<?php
+
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-base-embed-handler.php' );
+
+// Much of this class is borrowed from Jetpack embeds
+class AMP_DailyMotion_Embed_Handler extends AMP_Base_Embed_Handler {
+
+	const URL_PATTERN = '#https?:\/\/(www\.)?dailymotion\.com\/video\/.*#i';
+	const RATIO = 0.5625;
+
+	protected $DEFAULT_WIDTH = 600;
+	protected $DEFAULT_HEIGHT = 338;
+
+	private static $script_slug = 'amp-dailymotion';
+	private static $script_src = 'https://cdn.ampproject.org/v0/amp-dailymotion-0.1.js';
+
+	function __construct( $args = array() ) {
+		parent::__construct( $args );
+
+		if ( isset( $this->args['content_max_width'] ) ) {
+			$max_width = $this->args['content_max_width'];
+			$this->args['width'] = $max_width;
+			$this->args['height'] = round( $max_width * self::RATIO );
+		}
+	}
+
+	function register_embed() {
+		wp_embed_register_handler( 'amp-dailymotion', self::URL_PATTERN, array( $this, 'oembed' ), -1 );
+		add_shortcode( 'dailymotion', array( $this, 'shortcode' ) );
+	}
+
+	public function unregister_embed() {
+		wp_embed_unregister_handler( 'amp-dailymotion', -1 );
+		remove_shortcode( 'dailymotion' );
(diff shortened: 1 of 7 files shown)


includes/embeds/class-amp-dailymotion-embed.php
```diff
@@ -0,0 +1,101 @@
+<?php
+
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-base-embed-handler.php' );
+
+// Much of this class is borrowed from Jetpack embeds
+class AMP_DailyMotion_Embed_Handler extends AMP_Base_Embed_Handler {
+
+	const URL_PATTERN = '#https?:\/\/(www\.)?dailymotion\.com\/video\/.*#i';
+	const RATIO = 0.5625;
```
dandv: Commend on that magic number?

tests/test-amp-dailymotion-embed.php
```diff
@@ -0,0 +1,68 @@
...
+			),
+
+			'url_with_title' => array(
+				'http://www.dailymotion.com/video/x5awwth_snatched-official-trailer-2-hd_shortfilms' . PHP_EOL,
+				'<p><amp-dailymotion data-videoid="x5awwth" layout="responsive" width="600" height="338"></amp-dailymotion></p>' . PHP_EOL
+			),
+
+			'shortcode_unnamed_attr_as_url' => array(
+				'[dailymotion https://www.dailymotion.com/video/x5awwth]' . PHP_EOL,
+                '<amp-dailymotion data-videoid="x5awwth" layout="responsive" width="600" height="338"></amp-dailymotion>' . PHP_EOL
```
mjangda: Minor: fix indentation here.

includes/embeds/class-amp-dailymotion-embed.php
```diff
@@ -0,0 +1,101 @@
...
+		if ( empty( $url ) ) {
+			return '';
+		}
+
+		$video_id = $this->get_video_id_from_url( $url );
+
+		return $this->render( array(
+			'url' => $url,
+			'video_id' => $video_id,
+		) );
```
mjangda: It looks like the shortcode for Dailymotion expects the `id` not the URL, so we may need to modify some of the handling here.

The `oembed` method should call `get_video_id_from_url` to extract the `video_id` and then call `render` directly.

The `shortcode` method can mostly stay as-is, expect we can change references from `$url` => `$video_id` and remove the call to `get_video_id_from_url`. We also need to add a branch to the first `if` block to look for `$attr['id']` and use that if set.

The tests should be updated accordingly as well.

amedina: Makes sense. Updated the pieces appropriately. The shortcode tests pass, and they work on the AMP pages, but it seems there is an issue with the canonical ([dailymotion id=<video_id>] is not being rendered)

includes/embeds/class-amp-dailymotion-embed.php
```diff
@@ -0,0 +1,101 @@
...
+			)
+		);
+	}
+
+	private function get_video_id_from_url( $url ) {
+		$parsed_url = parse_url( $url );
+		parse_str( $parsed_url['path'], $path );
+        $tok = explode("/", $parsed_url['path']);
+		$tok = explode("_", $tok[2]);
+        $video_id = $tok[0];
```
amedina: Fixed.

tests/test-amp-dailymotion-embed.php
amedina: Fixed.

includes/embeds/class-amp-dailymotion-embed.php
amedina: Makes sense. Will update.

PR #640, opened 2017-02-18
Add support for amp-soundcloud

amedina:
Getting this PR started for review and discussion.

### DIFF
--- a/includes/embeds/class-amp-soundcloud-embed.php
+++ b/includes/embeds/class-amp-soundcloud-embed.php
@@ -1,0 +1,95 @@
+<?php
+
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-base-embed-handler.php' );
+
+class AMP_SoundCloud_Embed_Handler extends AMP_Base_Embed_Handler {
+	const URL_PATTERN = '#https?://api\.soundcloud\.com/tracks/.*#i';
+	protected $DEFAULT_HEIGHT = 200;
+
+	private static $script_slug = 'amp-soundcloud';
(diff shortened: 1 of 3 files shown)


includes/embeds/class-amp-soundcloud-embed.php
```diff
@@ -0,0 +1,97 @@
...
+				'height' => $this->args['height'],
+			)
+		);
+	}
+
+	private function get_track_id_from_url( $url ) {
+		$parsed_url = parse_url( $url );
+		parse_str( $parsed_url['path'], $path );
+		$tok = explode("/", $parsed_url['path']);
+		$tok = explode("_", $tok[2]);
```
mjangda: Do we need the `_` explode here?

amedina: My bad; not needed. Fixed.

westonruter:
@amedina as part of

westonruter:
@amedina as part of #806 I found the implementation here to not be complete. In particular, an oEmbed for SoundCloud looks like:

https://soundcloud.com/jack-villano-villano/mozart-requiem-in-d-minor

Not like:

https://api.soundcloud.com/tracks/89299804

Additionally, I found the `soundcloud` shortcode handling wasn't working as expected. For one thing, the Jetpack plugin only recognizes the `url` attribute whereas WordPress.com also apparently supports the `id` attribute.

So I've amended the PR #829 to fix up the SoundCloud handling: https://github.com/Automattic/amp-wp/pull/829/commits/0b729b1a728fbfd0faa9a7842ec757d1b8a5b4bd

PR #642, opened 2017-02-24
Add more support for RTL

SaidElBakkali:


westonruter:
We can re-open this PR but in general going forward we'll be recommending that users opt for AMP theme support (new as of v0.7) so that they can fully re-use their existing theme templates and stylesheets which would support RTL.

PR #654, opened 2017-03-07
PHPCS: code style fixes

paulschreiber:
`find . -name \*.php| xargs -n 1 phpcs -p -s -v --colors --standard=WordPress-VIP`

* change parse_url() to wp_parse_url()
* use strict equality and inequality checks
* use strict comparison with in_array() checks
* use snake_case for variable names (instead of ALL_CAPS)
* use yoda condition checks


mjangda:
Broken tests are unrelated (see #658).

PR #655, opened 2017-03-07
move template actions in to a class to avoid polluting global namespace

paulschreiber:


mjangda:
This will break back-compat (sites may be removing actions or explicitly referencing functions) and PHP compat (`[` arrays are PHP5.4+) :(

paulschreiber:
I pulled out the `array()` change.

Can we shim something out so removing actions still works? (Shimming functions will be easy.) This is < 1.0, so isn't breaking backward compatibility ok, with notice?

mjangda:
> Can we shim something out so removing actions still works? (Shimming functions will be easy.) 

Shimming adds a cost and I'm not sure we gain much overall with this change? Not polluting the global namespace would be nice but the functions are sufficiently named that it should avoid conflict. If I'm missing something, let me know.

> This is < 1.0, so isn't breaking backward compatibility ok, with notice?

Generally speaking, `1.0` isn't really a thing for us :) `0.4` is just as stable as `1.0` would be.

PR #670, opened 2017-03-25
Replace explode with preg_split.

trepmal:
In cases where Jetpack's `safecss_filter_attr` is used in place of core's function (and possibly other scenarios), there may be semi-colons in unexpected places.

For example, `background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAA...);`

In that case, `explode` will break the style, splitting it in the middle of a value. With `preg_split` we can specify semi-colons that are not inside parentheses.

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -56,8 +56,11 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			return array();
 		}
 
-		// Normalize order
-		$styles = array_map( 'trim', preg_split("/;(?![^(]*\))/", $string ) );
+		// safecss returns a string but we want individual rules.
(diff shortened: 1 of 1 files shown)


trepmal:
Related: #604, #610

mjangda:
Thanks for the PR, @trepmal. I'm working on fixing the (unrelated broken) tests but it would be good to add some tests for this particular fix and an inline comment with details on what the regex is trying to catch.

trepmal:
Writing a test against core is proving tricky since `safecss_filter_attr` itself explodes on semi-colon ([#](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/kses.php#L1698)), so styles are broken before it gets to the plugin.

At the moment, it's impossible to bypass `safecss_filter_attr` without hacking core.

However, there are some tickets hoping to fix/modernize/filter `safecss_filter_attr` which could introduce the same issue in a non-hacking-core way.

https://core.trac.wordpress.org/ticket/24157
https://core.trac.wordpress.org/ticket/37134

Issue #658, opened 2017-03-08
Fix tests


mjangda:
The tests that fetch the remote image image are failing because we get a `400` back.

Need reliably accessible versions of the following images (or mocked versions):

```
http://i0.wp.com/placehold.it/350x150.png
http://i0.wp.com/placehold.it/1024x768.png
```

PR #672, opened 2017-03-25
Fix Tests

mjangda:
Update travis config for PHP7 + phpunit 6 compat. And update image URL to one that works.


mjangda:
Fixed #658

Issue #660, opened 2017-03-10
Can't turn off featured images


piersb:
Back in 0.3, featured images weren't turned on by default. Now in 0.4, they are. 

Problem: If you *don't* want a featured image at the top of your amp page, then tough. There's no way to turn 'em off.

Workaround:

Because amp-wp requires there to be a featured image (otherwise it bails), we can intercept by using the hook on get_the_post_thumbnail, and returning false.

```
// only do it on an amp page
	add_action( 'pre_amp_render_post', 'amp_add_custom_actions' );

function amp_add_custom_actions() {
	add_filter( 'post_thumbnail_html', 'amp_remove_featured_image'  );
}

// pretend there's no featured image, whether there is or isn't
static function amp_remove_featured_image() {
	return false;
}
```

piersb:
Aside - I'm not sure how to change the plugin to make this easier. Because we build the data structure privately, how and where (and should) you allow hooks within it?

Would this (untested) code change, for example, be a really good or really dumb idea?

```
private function add_data_by_key( $key, $value ) {
list ($key, $value) = add_filter('manipulate_data_by_key', array( $key, $value) );
	$this->data[ $key ] = $value;
}
```

Pro: It would allow hooking into the ampification
Con: It would allow hooking into the ampification

mjangda:
This should work:

```
add_filter( 'amp_post_template_data', function( $data ) {
    $data['featured_image'] = false;
    return $data;
} );
```

dan-eth:
@Mjangda

Should that code be inserted into the 

amp/templates/featured-image.php OR

amp/includes/amp-post-template-functions.php ?

mjangda:
@dan-eth Please don't modify the plugin directly. Please see https://github.com/Automattic/amp-wp#where-do-i-put-my-code for details on where your custom code should go.

dan-eth:
@Mjangda
Thanks for the advice bro. 

But you Automatic people shouldn't have to be enablers in Google's attempt to impose its half-baked ideas on the entire Web.

You should have told them that people are self - hosting their blogs with WordPress for a reason.

We could have as well be using Google-Plus, if we wanted google dictating everything.

I wonder if the "Powered by WordPress" advertisement at the bottom of AMP pages was the price to get Automatic on board with the annoying project of Google. 

Have a nice weekend anyway.

piersb:
@mjangda Thanks. I can confirm that the code works, and have created a pull request which updates the documentation appropriately.

PR #676, opened 2017-03-27
Update documentation re: featured images

piersb:
Updates readme.md to reflect the fact that the plugin now includes featured images in amp pages by default, and shows a way to turn them off for those that preferred the earlier behaviour, per issue #660


mjangda:
Thanks @piersb!

PR #679, opened 2017-03-29
parse query string

amyevans:
This fixes a bug where the query string was not parsed, so `$w` and `$h` were always `false`.


mjangda:
Good catch! Thank you!

PR #681, opened 2017-04-05
Add JavaScript info to readme

jacklenox:
Some very basic material on the "right way" to use custom JavaScript in an AMP template.

For background, see: #64445-z


mjangda:
As discussed, the AMP spec doesn't currently allow custom Javascript so going to close this out.

PR #690, opened 2017-04-28
Add empty logo as a fallback to structure metadata

mjangda:
To keep GWT happy as publisher logo is required.


mjangda:
Actually, might be better to hook in and inject via a filter as this could break current implementations.

westonruter:
Closing as merge conflicts and if we do this it would probably be better to use the WordPress logo as the default. In fact, this is what we are doing now:

https://github.com/ampproject/amp-wp/blob/85e45e895629bddf674e8ca2181128a1bcc6ad8c/includes/amp-helper-functions.php#L1451-L1454

PR #695, opened 2017-05-15
Fix for AMP height & width invalid attributes

onlineth:
Sometimes, there are dimensions for height and not the width and vice-versa. This leaves one attribute with no value which results as an invalid layout property. With this, now it will use the dimension for the given attribute and for the other dimension that wasn't given, it will use the default dimension.

### DIFF
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -84,6 +84,12 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 		$old_attributes = AMP_DOM_Utils::get_node_attributes_as_assoc_array( $node );
 		$new_attributes = $this->filter_attributes( $old_attributes );
 		$new_attributes = $this->enforce_sizes_attribute( $new_attributes );
+		if ( $new_attributes['height'] === '' ) {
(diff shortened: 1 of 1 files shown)


onlineth:
Agreeing with https://github.com/Automattic/amp-wp/pull/693#issuecomment-300697143, the unit tests do need to be updated.

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -84,6 +84,12 @@ private function adjust_and_replace_node( $node ) {
 		$old_attributes = AMP_DOM_Utils::get_node_attributes_as_assoc_array( $node );
 		$new_attributes = $this->filter_attributes( $old_attributes );
 		$new_attributes = $this->enforce_sizes_attribute( $new_attributes );
+		if ( $new_attributes['height'] === '' ) {
+			$new_attributes['height'] = self::FALLBACK_HEIGHT;
+		}
+		if ( $new_attributes['width'] === '' ) {
```
westonruter: Instead of strict equality with an empty string, perhaps `empty()` just in case the value is `null` or `false`.

westonruter:
See also #575.

westonruter:
I believe this was fixed with #979. Please let me know if otherwise.

PR #702, opened 2017-05-30
Add basic support for amp-vimeo 

amedina:


includes/embeds/class-amp-vimeo-embed.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-base-embed-handler.php' );
+
+// Much of this class is borrowed from Jetpack embeds
+class AMP_Vimeo_Embed_Handler extends AMP_Base_Embed_Handler {
+
+	const URL_PATTERN = '#https?:\/\/(www\.)?vimeo\.com\/*\/*\/.*#i';
```
mjangda: Since we're only handling URLs in the format `https://vimeo.com/172355597` should we limit this regex to something simpler like `'#https?:\/\/(www\.)?vimeo\.com\/\d+#i'`

amedina: Would that pattern capture the different cases? I had the URL_PATTERN as you suggested, then I though that the set of possibilities is: 

`https://vimeo.com/*
https://vimeo.com/*/*/video/*
https://vimeo.com/album/*/video/*
https://vimeo.com/channels/*/*
https://vimeo.com/groups/*/videos/*
https://vimeo.com/ondemand/*/*`

And the video-id is the last component in each case. The specified pattern matches all these cases and the function to extract the video id gets the last element. What do you think?

amedina:
Made the change of URL_PATTERN , as suggested.

PR #703, opened 2017-05-31
Add basic support for embedding Pinterest pins

amedina:


includes/embeds/class-amp-pinterest-embed.php
```diff
@@ -0,0 +1,70 @@
...
+			'url' => false,
+		) );
+
+		if ( empty( $args['url'] ) ) {
+			return '';
+		}
+
+		$this->did_convert_elements = true;
+
+		$t = AMP_HTML_Utils::build_tag(
```
mjangda: Looks like this is unused?

amedina: Oops ; my bad! Removed.

mjangda:
Updated code looks code; let's rebase with #706 once that's merged to make sure tests all pass and then we can merge.

PR #706, opened 2017-06-02
Fix URL handling for schemaless URLS in PHP5.3

### LINKED ISSUE #699: PHP 5.3 Test Failures
https://travis-ci.org/Automattic/amp-wp/jobs/231044444

```
1) AMP_Tag_And_Attribute_Sanitizer_Attr_Spec_Rules_Test::test_validate_attr_spec_rules with data set "test_attr_spec_rule_disallowed_domain_fail" (array(0, 'form', 'action', '//cdn.ampproject.org', true, true, 'check_attr_spec_rule_disallowed_domain'), 'fail')
--- Expected
+++ Actual
@@ @@
-'fail'
+'not_applicable'
/home/travis/build/Automattic/amp-wp/tests/amp-tag-and-attribute-sanitizer-private-methods-tests.php:604
```

mjangda:
`parse_url` doesn't handle schemaless URLs well in 5.3. `wp_parse_url` has a shim that fixes those bugs.

Fixes #699

### DIFF
--- a/includes/class-amp-post-template.php
+++ b/includes/class-amp-post-template.php
@@ -3,6 +3,7 @@
 require_once( AMP__DIR__ . '/includes/utils/class-amp-dom-utils.php' );
 require_once( AMP__DIR__ . '/includes/utils/class-amp-html-utils.php' );
 require_once( AMP__DIR__ . '/includes/utils/class-amp-string-utils.php' );
+require_once( AMP__DIR__ . '/includes/utils/class-amp-wp-utils.php' );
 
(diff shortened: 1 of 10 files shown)


mjangda:
Oh, there's actually a bug in wp_parse_url in WordPress 4.6 that ignores the second param (so you always get an array), which is why the tests are still breaking. Need more work.

mjangda:
Okay, should be fine now; @amedina can you take another look?

This fixes the broken tests and potential compatibility issues with `parse_url` in older PHP versions (< 5.4).

PR #714, opened 2017-06-14
Enable adding AMP Analytics components using the Options API

amedina:
The goal is to provide a simple mechanism for users to add AMP analytics components to  posts via a configuration panel of analytics options. 

This is an initial PR to kick off the review process. Things missing: (1) testing; (2) ironing some rough edges (e.g. details on the AMP top-level menu); (3) Styling of the Analytics options page


amedina:
screen shot 2017-06-13 at 5 56 06 pm

amedina:
screen shot 2017-06-19 at 10 34 19 am

includes/options/views/class-amp-analytics-options-serializer.php
```diff
@@ -0,0 +1,41 @@
...
+		if ( !( empty( $_POST['vendor-type'] ) ||
+		     empty( $_POST['config'] ) ) ) {
+
+			if ( empty( $_POST['id-value'] ) ) {
+				$_POST['id-value'] = md5( $_POST['config'] );
+			}
+
+			$new_analytics_options = array(
+				$_POST['id-value'],
+				$_POST['vendor-type'],
```
mjangda: We need to validate or sanitize the vendor type (maybe check against a list of allowed types)?

includes/options/views/class-amp-analytics-options-serializer.php
```diff
@@ -0,0 +1,41 @@
...
+				stripslashes( $_POST['config'] )
```
mjangda: We need to validate that `config` is valid JSON.

includes/options/views/class-amp-analytics-options-submenu-page.php
```diff
@@ -0,0 +1,54 @@
...
+
+	private function render_option($id = "", $type = "", $config = "") {
+			?>
+			<div class="analytics-data-container">
+				<form method="post" action="<?php echo esc_html( admin_url( 'admin-post.php' ) ); ?>">
+					<h2>Analytics Component: <?php echo ($type ? $type . ':' : '') . substr($id, -6) ?></h2>
+					<div class="options">
+						<p>
+							<label>Type: </label>
+							<input class="option-input" type="text" name=vendor-type value="<?php echo $type; ?>" />
```
mjangda: `esc_attr` for $type

includes/options/views/class-amp-analytics-options-submenu-page.php
```diff
@@ -0,0 +1,54 @@
...
+							<label>Id: </label>
+							<input type="text" name=id value="<?php echo substr($id, -6); ?>" text="alberto" readonly />
```
mjangda: `text="alberto"`? :)

mjangda: `esc_attr` for the `$id`

includes/options/views/class-amp-analytics-options-submenu-page.php
```diff
@@ -0,0 +1,54 @@
...
+                            <input type="hidden" name=id-value value="<?php echo $id; ?>" />
```
mjangda: `esc_attr` for `$id`

includes/options/views/class-amp-analytics-options-submenu-page.php
```diff
@@ -0,0 +1,54 @@
...
+						</p>
+						<p>
+							<label>JSON Configuration:</label>
+							<br />
+							<textarea rows="10" cols="100" name="config"><?php echo stripslashes($config); ?></textarea>
```
mjangda: `esc_textarea` for `$config`.

tests/test-amp-analytics-options.php
```diff
@@ -0,0 +1,222 @@
...
+	}';
+	private $serializer;
+
+	public function setUp() {
+		$this->serializer = new Analytics_Options_Serializer();
+	}
+
+	public function tearDown() {
+		global $wpdb;
+		$wpdb->query( 'ROLLBACK' );
```
mjangda: Is this necessary? I think the class already does this for us?

includes/options/views/class-amp-analytics-options-serializer.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+class Analytics_Options_Serializer {
+
+	public static function save() {
```
mjangda: As discussed, we should look into using the settings API for this instead of manually handling the save (e.g. https://konstantin.blog/2012/the-wordpress-settings-api/). Right now, this handler doesn't have proper capability checks or CSRF protection, which we get "for free" with the settings API.

We can register the setting with `register_setting` and with the `sanitize_callback` where most of the logic from this `save` would move to. Also, thinking to the future, it might be good to nest options under an `amp_settings` option so the structure looks something like:

```
amp_settings => array(
    amp_analytics => array(
        'googleanalytics-basd12' => array( ... ),
        'parsely-sua231' => array( ... ),
    ),
)
```

If we want to keep it as a top-level option for analytics, we need to rename it to `amp_analytics_config` (to avoid conflicts with other plugins/themes).

I'll see if I can pull together an example of moving things over.

includes/admin/amp-wp-admin-styles.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+add_action('admin_head', 'amp_options_styles');
+function amp_options_styles() {
```
amedina: (1) Agree. Moved.
(2) What is the best way to restrict it to the analytics options page? The page does not use a template; should I use the URL parameter 'page='amp-analytics'?

includes/admin/functions.php
```diff
@@ -20,26 +24,12 @@ function amp_init_customizer() {
...
- */
-function amp_add_customizer_link() {
-	// Teensy little hack on menu_slug, but it works. No redirect!
-	$menu_slug = add_query_arg( array(
-		'autofocus[panel]'         => AMP_Template_Customizer::PANEL_ID,
-		'return'                   => rawurlencode( admin_url() ),
-		AMP_CUSTOMIZER_QUERY_VAR   => true,
-	), 'customize.php' );
+	// Add a link to Settings
+	add_action( 'admin_menu', 'amp_add_amp_options_link' );
```
amedina: Agree. Moved.

includes/admin/functions.php
```diff
@@ -69,3 +59,67 @@ function amp_admin_get_preview_permalink() {
...
+function amp_add_amp_options_link() {
+	$amp_options = new AMP_Options_Menu();
+	$amp_options->init();
+}
+
+/**
+ * Grab the analytics options from the DB and return $analytics option
+ * @return array
+ */
+function get_analytics_component_fields($option) {
```
amedina: Done.

includes/admin/functions.php
```diff
@@ -69,3 +59,67 @@ function amp_admin_get_preview_permalink() {
...
+
+	$id= $option[0];
+	$type = $option[1];
+	$config = $option[2];
```
amedina: Nice. Done.

includes/admin/functions.php
```diff
@@ -69,3 +59,67 @@ function amp_admin_get_preview_permalink() {
...
+
+	$fields = array();
+	$component_index = $type . '-' . $id;
+	$fields['id'] =  $component_index;
+	$fields['type'] = $type;
+	$fields['attributes'] = array();
+
+	$analytics_json = json_decode( stripslashes( $config ), true );
```
amedina: Yeah; we don't.

includes/options/class-amp-analytics-options-submenu.php
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	public function init() {
+		$this->add_submenu();
+	}
+
+	private function add_submenu() {
+		add_submenu_page(
+			$this->parent_menu_slug,
+			'AMP Analytics Options',
```
amedina: Done.

includes/options/views/class-amp-analytics-options-serializer.php
amedina: (1) Good idea re:Thinking about the future. The options are saved now according to the hierarchy you suggested. 
(2) Thinking about using the Settings API. Need to change everything? :(

tests/test-amp-analytics-options.php
amedina: True; leftover from trials.

includes/options/views/class-amp-analytics-options-serializer.php
amedina: Done. Added a validation function and the save function proceeds only if content of config is valid JSON.

includes/options/views/class-amp-analytics-options-submenu-page.php
amedina: Done.

includes/options/views/class-amp-analytics-options-serializer.php
amedina: How should we do this? Checking a list of allowed types? Wonder if we should let the user ensure that the type entered is correct, since right now they are doing it manually. WDYT?

includes/admin/functions.php
```diff
@@ -20,26 +24,12 @@ function amp_init_customizer() {
...
 
-	// Add the theme page.
-	$page = add_theme_page(
-		__( 'AMP', 'amp' ),
-		__( 'AMP', 'amp' ),
-		'edit_theme_options',
-		$menu_slug
-	);
+	// Trigger analytics options serializer on analytics option's save
+	add_action( 'admin_post_analytics_options', 'Analytics_Options_Serializer::save' );
```
amedina: Let's discuss.

includes/options/views/class-amp-analytics-options-submenu-page.php
amedina: Done.

includes/options/views/class-amp-analytics-options-submenu-page.php
amedina: Done.

includes/options/views/class-amp-analytics-options-submenu-page.php
amedina: text="alberto"? Of course! :) ==> removed.

tests/test-amp-analytics-options.php
```diff
@@ -75,8 +88,9 @@ function test_no_options() {
 	 * Test that exactly one analytics component is inserted into the DB
 	 */
 	function test_one_option_inserted() {
-		/* Delete analytics options, if any */
-		delete_option( 'analytics' );
+
+		// Delete options, if any
+		delete_option( 'amp-options' );
```
amedina: Related to the ROLLBACK comment before: some tests were failing seemingly due the DB not being rolled back between tests; it should be but... Adding this delete_option for now.

includes/admin/amp-wp-admin-styles.php
mjangda: We can use this hook instead: https://developer.wordpress.org/reference/hooks/admin_print_styles-hook_suffix/ 

```
add_action( 'admin_print_styles-' . $this->menu_slug, array( $this, 'amp_options_styles' ) );
```

includes/options/views/class-amp-analytics-options-serializer.php
mjangda: If validation fails when an invalid type is used, we should definitely check against a list -- maybe the UI should provide a dropdown instead? (Only downside would be keeping this list updated.)

Otherwise we can leave is freeform and just sanitize the value using `sanitize_key`.

includes/admin/amp-wp-admin-styles.php
amedina: Hmm. I replaced the action added to admin_head with this one and the styles do not get applied.

includes/options/views/class-amp-analytics-options-serializer.php
amedina: [Discussed offline] Keeping the use of the Options API.

includes/options/views/class-amp-analytics-options-serializer.php
amedina: [Discussed offline] Going with the sanitize_key option to start with.

readme.md
```diff
@@ -548,12 +548,46 @@ to remove it from the callback chain if you don't want it to, e.g.
...
+
 ## Analytics
 
+There are two options you can follow to include analytics tags in your posts.
 
-To output proper analytics tags, you can use the `amp_post_template_analytics` filter:
+### Plugin Analytics Options
 
-```
+The plugin defines an analytics option to enable the addition of
```
amedina: The links in the image references should be changed to the master branch before merging.

includes/admin/functions.php
```diff
@@ -69,3 +52,78 @@ function amp_admin_get_preview_permalink() {
...
+/**
+ * Registers a top-level menu for AMP configuration options
+ */
+function amp_add_options_menu() {
+	$amp_options = new AMP_Options_Menu();
+	$amp_options->init();
+}
+add_action('admin_menu','amp_add_options_menu');
+// Action to trigger analytics options serializer on analytics option's save
+add_action( 'admin_post_analytics_options', 'Analytics_Options_Serializer::save' );
```
mjangda: We should prefix the `admin_post_` action so it becomes `admin_post_amp_analytics_options`

includes/options/views/class-amp-analytics-options-serializer.php
```diff
@@ -0,0 +1,66 @@
...
+
+require_once( AMP__DIR__ . '/includes/utils/class-amp-html-utils.php' );
+
+class Analytics_Options_Serializer {
+
+	public static function save() {
+
+		if ( is_admin() ) {
+			// Request must come from user with right capabilities
+			if ( ! current_user_can( 'administrator' ) ) {
```
mjangda: We should use `manage_options` (capability) instead of a role check.

includes/options/views/class-amp-analytics-options-serializer.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+require_once( AMP__DIR__ . '/includes/utils/class-amp-html-utils.php' );
+
+class Analytics_Options_Serializer {
+
+	public static function save() {
+
+		if ( is_admin() ) {
```
mjangda: Why is this check needed?

includes/options/views/class-amp-analytics-options-serializer.php
```diff
@@ -0,0 +1,66 @@
...
+		if ( ! ( empty( $_POST['vendor-type'] ) || empty( $_POST['config'] ) ) &&
+				$is_valid_json ) {
+
+			if ( empty( $_POST['id-value'] ) ) {
+				$_POST['id-value'] = md5( $_POST['config'] );
+			}
+
+			// Prepare the data for the new analytics setting
+			$new_analytics_option = array(
+				$_POST['id-value'],
```
mjangda: The id needs to be sanitized.

includes/options/views/class-amp-analytics-options-serializer.php
```diff
@@ -0,0 +1,66 @@
...
+				unset( $amp_analytics[ $inner_option_name ] );
+			} else {
+				$amp_analytics[ $inner_option_name ] = $new_analytics_option;
+			}
+			$amp_options[ $option_name ] = $amp_analytics;
+			update_option( 'amp-options' , $amp_options, false );
+		}
+		// Redirect to keep the user in the analytics options page
+		// Wrap in is_admin() to enable phpunit tests to exercise this code
+		if ( is_admin() ) {
```
mjangda: Instead of adding logic to handle special cases for tests, we should split this into two methods: `save` and `handle_analytics_submit`

```
function save() {
    capability check
    nonce check

    options = get_options

    $options = handle_analytics_submit( $options, $_POST )

    redirect
}

function handle_analytics_submit( $options, $data ) {
    // validation and other logic goes here

    return $options;
}
```

`handle_analytics_submit` can then be more easily tested.

includes/options/views/class-amp-analytics-options-submenu-page.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+require_once( AMP__DIR__ . '/includes/options/views/class-amp-analytics-options-serializer.php' );
+
+class AMP_Analytics_Options_Submenu_Page {
+
+	private function render_option($id = "", $type = "", $config = "") {
+			?>
+			<div class="analytics-data-container">
+				<form method="post" action="<?php echo esc_html( admin_url( 'admin-post.php' ) ); ?>">
```
mjangda: `esc_url` instead of `esc_html`

includes/options/views/class-amp-analytics-options-submenu-page.php
```diff
@@ -0,0 +1,77 @@
...
+					<h2><?php echo __( 'Analytics Component', 'amp' ) ?>: <?php echo ($type ? $type . ':' : '') . substr($id, -6) ?></h2>
```
mjangda: Output of $type and $id should be wrapped in `esc_html`

includes/options/views/class-amp-analytics-options-submenu-page.php
```diff
@@ -0,0 +1,77 @@
...
+					<div class="options">
+						<p>
+							<label><?php echo __( 'Type', 'amp' ) ?>: </label>
+							<input class="option-input" type="text" name=vendor-type value="<?php echo esc_attr( $type ); ?>" />
+							<label><?php echo __( 'Id', 'amp' ) ?>: </label>
+							<input type="text" name=id value="<?php echo substr(esc_attr( $id ), -6); ?>" readonly />
```
mjangda: `esc_attr` should be outside `substr`

includes/options/views/class-amp-analytics-options-submenu-page.php
```diff
@@ -0,0 +1,77 @@
...
+                            <input type="hidden" name=id-value value="<?php echo $id; ?>" />
```
mjangda: `$id` needs `esc_attr`

includes/options/views/class-amp-analytics-options-submenu-page.php
```diff
@@ -0,0 +1,77 @@
...
+		<div class="wrap">
+			<h1><?php echo esc_html( get_admin_page_title() ); ?></h1>
+		<?php
+
+		// If redirected from serializer, check if action succeeded
+		global $_GET;
+		if ( isset( $_GET['valid'] ) ) {
+            if ( ! esc_attr( $_GET['valid'] ) ) {
+                ?>
+                <h3 style='color:red'><?php echo 'Action not taken: invalid input!'; ?></h3>
```
mjangda: We should use WordPress standard admin notices here: https://digwp.com/2016/05/wordpress-admin-notices/

includes/options/views/class-amp-options-menu-page.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+class AMP_Options_Menu_Page {
+	public function render() {
+		?>
+		<div class="ampoptions-admin-page">
+			<h1><?php echo __( 'AMP Plugin Options', 'amp' ) ?></h1>
+				<p>
+					This admin panel menu contains configuration options
```
mjangda: Needs translation.

includes/options/class-amp-analytics-options-submenu.php
```diff
@@ -0,0 +1,50 @@
...
+	public function __construct( $parent_menu_slug ) {
+		$this->parent_menu_slug = $parent_menu_slug;
+		$this->menu_slug = 'amp-analytics-options';
+		$this->menu_page = new AMP_Analytics_Options_Submenu_Page();
+	}
+
+	public function init() {
+		$this->add_submenu();
+		add_action(
+			'admin_head',
```
mjangda: We still need to limit this to the analytics page.

includes/options/views/class-amp-analytics-options-serializer.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+require_once( AMP__DIR__ . '/includes/utils/class-amp-html-utils.php' );
+
+class Analytics_Options_Serializer {
```
mjangda: Let's rename this to `AMP_Options_Manager` and we can use it for other options we decide to add in the future.

includes/admin/functions.php
```diff
@@ -69,3 +52,78 @@ function amp_admin_get_preview_permalink() {
...
+}
+
+/**
+ * Registers a top-level menu for AMP configuration options
+ */
+function amp_add_options_menu() {
+	$amp_options = new AMP_Options_Menu();
+	$amp_options->init();
+}
+add_action('admin_menu','amp_add_options_menu');
```
mjangda: For any new code, let's be mindful of WordPress coding standards around spacing and indents: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/

e.g. this would become 

```
add_action( 'admin_menu', 'amp_add_options_menu' );
```

Would be good to especially clean this up in any new files.

amedina: Totally agree; my bad. I went over all the files in this PR and checked for coding standard violations; found several (mostly missing spaces).

includes/admin/functions.php
amedina: Done.

includes/options/views/class-amp-analytics-options-serializer.php
amedina: Good idea. Done.

includes/options/views/class-amp-analytics-options-serializer.php
amedina: This check prevent the checks to be run when the code is invoked from the test suite. Is there other way to prevent the tests from breaking (without the check was getting a permission error when running test_amp-analytics_options.php)

includes/options/views/class-amp-analytics-options-serializer.php
amedina: Done.

includes/options/views/class-amp-analytics-options-serializer.php
amedina: Done.

includes/options/views/class-amp-analytics-options-submenu-page.php
amedina: Done.

includes/options/views/class-amp-analytics-options-submenu-page.php
amedina: Done.

includes/options/views/class-amp-analytics-options-submenu-page.php
amedina: Done

includes/options/views/class-amp-analytics-options-submenu-page.php
amedina: Done.

includes/options/views/class-amp-analytics-options-submenu-page.php
amedina: Very nice; was not aware of these. Done.

includes/options/views/class-amp-analytics-options-serializer.php
amedina: That is the way to do it!. Done.

includes/options/class-amp-analytics-options-submenu.php
amedina: Sorry I did not address this in the previous review; what is the standard way to achieve this? Could you point me to an example?

mjangda: This should work:

```
add_action( 'admin_print_styles-' . $this->menu_slug, array( $this, 'amp_options_styles' ) );
```

Alternatively:

```
add_action( 'admin_print_styles', array( $this, 'add_options_styles' ) );

function add_options_styles() {
    $screen = get_current_screen();
    if ( $this->menu_slug !== $screen->id ) {
        return;
    }

    // output styles
}
```

mjangda:
Looks like there are some test fatals related to the updated code:

PHP 5.2
https://travis-ci.org/Automattic/amp-wp/jobs/248203194

```
Fatal error: Call to undefined function json_last_error() in /home/travis/build/Automattic/amp-wp/includes/utils/class-amp-html-utils.php on line 24
```


PHP 5.3+
https://travis-ci.org/Automattic/amp-wp/jobs/248203203

```
1) AMP_Render_Post_Test::test__invalid_post
pre_amp_render_post action fire when it should not have.
Failed asserting that 3 matches expected 0.
/home/travis/build/Automattic/amp-wp/tests/test-amp-render-post.php:9
2) AMP_Render_Post_Test::test__valid_post
Failed asserting that 4 matches expected 1.
```

includes/options/views/class-amp-options-menu-page.php
amedina: Done.

PR #723, opened 2017-07-04
Abstract notion of AMP Templates

amedina:
Second step in road to Canonical AMP mode: define an AMP_template class and have AMP_Post_Template inherit from it; then we will add template classes for different post/content types (e.g. pages, home, etc.).


includes/sanitizers/require-sanitizers.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+require_once( AMP__DIR__ . '/includes/sanitizers/class-amp-style-sanitizer.php' );
+require_once ( AMP__DIR__ . '/includes/sanitizers/class-amp-blacklist-sanitizer.php' );
```
mjangda: Minor: the spacing for many of the require statements, seems off. Should be:

`require( '' );`

includes/templates/class-amp-post-template.php
```diff
@@ -1,96 +1,52 @@
...
-			 * @since 0.4
-			 *.
-			 * @param	array	$analytics	An associative array of the analytics entries we want to output. Each array entry must have a unique key, and the value should be an array with the following keys: `type`, `attributes`, `script_data`. See readme for more details.
-			 * @param	object	$post	The current post.
-			 */
-			'amp_analytics' => apply_filters( 'amp_post_template_analytics', array(), $this->post ),
-			);
+		parent::__construct();
+		
+		$this->post_id = $post_id;
```
mjangda: We should keep $this->ID for back-compat (or maybe add a magic `__get` method to handle this?)

includes/templates/class-amp-post-template.php
```diff
@@ -1,96 +1,52 @@
...
+				 * Add amp-analytics tags.
+				 *
+				 * This filter allows you to easily insert any amp-analytics tags without needing much heavy lifting.
+				 *
+				 * @since 0.4
+				 *.
+				 * @param	array	$analytics	An associative array of the analytics entries we want to output. Each array entry must have a unique key, and the value should be an array with the following keys: `type`, `attributes`, `script_data`. See readme for more details.
+				 * @param	object	$post	The current post.
+				 */
+				'amp_analytics' => apply_filters( 'amp_post_template_analytics', array(), $this->post ),
```
mjangda: This can be probably be moved to `AMP_Template`?

includes/templates/class-amp-post-template.php
```diff
@@ -260,11 +170,11 @@ private function build_post_content() {
 
 		$this->add_data_by_key( 'post_amp_content', $amp_content->get_amp_content() );
 		$this->merge_data_for_key( 'amp_component_scripts', $amp_content->get_amp_scripts() );
-		$this->merge_data_for_key( 'post_amp_styles', $amp_content->get_amp_styles() );
+		$this->merge_data_for_key( 'amp_styles', $amp_content->get_amp_styles() );
```
mjangda: We probably shouldn't change this for back-compat reasons. Or we should find a workaround that gracefully handles the change.

includes/templates/class-amp-template.php
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+	
+	private function verify_and_include( $file, $template_type ) {
+		$located_file = $this->locate_template( $file );
+		if ( $located_file ) {
+			$file = $located_file;
+		}
+		
+		$file = apply_filters( 'amp_post_template_file', $file, $template_type, $this->post );
```
mjangda: `$this->post` will be null for all non-post templates and we'll need to think about how to handle that.

mjangda:
```
if ( 'AMP_Post_Template' === get_class( $this ) ) {
    $file = apply_filters( 'amp_post_template_file', $file, $template_type, $this->post );
}

$file = apply_filters( 'amp_template_file', $file, $template_type );
```

includes/sanitizers/require-sanitizers.php
amedina: Fixed.

includes/templates/class-amp-post-template.php
amedina: Added magic method as discussed.

includes/templates/class-amp-post-template.php
amedina: Agree. Done.

includes/templates/class-amp-post-template.php
amedina: Approach her can be: (1) Keep `post_amp_styles` as part of the AMP_Post_Template class, and (2) keep the more general `amp_styles` as part of the new template classes in the hierarchy.

includes/embeds/require-embeds.php
```diff
@@ -0,0 +1,12 @@
...
+
+require_once ( AMP__DIR__ . '/includes/embeds/class-amp-twitter-embed.php' );
+require_once ( AMP__DIR__ . '/includes/embeds/class-amp-youtube-embed.php' );
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-dailymotion-embed.php' );
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-vimeo-embed.php' );
+require_once( AMP__DIR__ . '/includes/embeds/class-amp-soundcloud-embed.php' );
+require_once ( AMP__DIR__ . '/includes/embeds/class-amp-gallery-embed.php' );
+require_once ( AMP__DIR__ . '/includes/embeds/class-amp-instagram-embed.php' );
+require_once ( AMP__DIR__ . '/includes/embeds/class-amp-vine-embed.php' );
+require_once ( AMP__DIR__ . '/includes/embeds/class-amp-facebook-embed.php' );
```
mjangda: Minor: spacing.

includes/templates/class-amp-post-template.php
```diff
@@ -1,96 +1,38 @@
 <?php
 
-require_once( AMP__DIR__ . '/includes/utils/class-amp-dom-utils.php' );
-require_once( AMP__DIR__ . '/includes/utils/class-amp-html-utils.php' );
-require_once( AMP__DIR__ . '/includes/utils/class-amp-string-utils.php' );
+require_once ( AMP__DIR__ . '/includes/utils/class-amp-dom-utils.php' );
+require_once ( AMP__DIR__ . '/includes/utils/class-amp-html-utils.php' );
+require_once ( AMP__DIR__ . '/includes/utils/class-amp-string-utils.php' );
```
mjangda: Minor: spacing.

includes/templates/class-amp-post-template.php
```diff
@@ -100,65 +42,29 @@ public function __construct( $post_id ) {
...
-		$this->data[ $key ] = $value;
-	}
-
-	private function merge_data_for_key( $key, $value ) {
-		if ( is_array( $this->data[ $key ] ) ) {
-			$this->data[ $key ] = array_merge( $this->data[ $key ], $value );
-		} else {
-			$this->add_data_by_key( $key, $value );
-		}
+	public function load_post_template() {
```
mjangda: I would suggest keeping this method as `load` with a default param:

```
public function load( $template = 'single' ) {
    parent::load( $template );
}
```

includes/templates/class-amp-post-template.php
```diff
@@ -260,11 +166,14 @@ private function build_post_content() {
 
 		$this->add_data_by_key( 'post_amp_content', $amp_content->get_amp_content() );
 		$this->merge_data_for_key( 'amp_component_scripts', $amp_content->get_amp_scripts() );
+		// Keeping both style arrays for backward compatibility
+		// post_amp_styles will be deprecated in the future
+		$this->merge_data_for_key( 'amp_styles', $amp_content->get_amp_styles() );
```
mjangda: I think to handle backwards compatibility here, we need to add a special handler in the `get` method that returns `amp_styles` when someone requests `post_amp_styles`. Maybe something like:

```
public function get( $property, $default = null ) {
    if ( 'post_amp_styles' === $property ) {
        _doing_it_wrong( __METHOD__, 'The `post_amp_styles` key has been deprecated; use `amp_styles` instead.', '0.6' );
        return $this->get( 'amp_styles', $default );
    } elseif ( 'amp_styles' === $property ) {
        // merge the values of  `$this->data[ 'post_amp_styles' ]` and `$this->data[ 'amp_styles' ]` and return
    }

    // ...
}
```

includes/templates/class-amp-post-template.php
amedina: Fixed.

amedina:
@mjangda when you have a chance take a look at the latest review; after that if we are good to go I will merge. Thanks!

includes/templates/class-amp-template.php
```diff
@@ -0,0 +1,168 @@
...
+			 * Add amp-analytics tags.
+			 *
+			 * This filter allows you to easily insert any amp-analytics tags without needing much heavy lifting.
+			 *
+			 * @since 0.4
+			 *.
+			 * @param	array	$analytics	An associative array of the analytics entries we want to output. Each array entry must have a unique key, and the value should be an array with the following keys: `type`, `attributes`, `script_data`. See readme for more details.
+			 * @param	object	$post	The current post.
+			 */
+			'amp_analytics' => apply_filters( 'amp_post_template_analytics', array(), $this->post ),
```
mjangda: Do we want to rename this filter as well (with proper back-compat handling)? One issue is that because of load order, `$this->post` will be null (`AMP_Template::__construct` is called before `$this->post` is assigned in `AMP_Post_Template`.

Maybe we change this filter to `amp_template_analytics` and remove the post object and move this into the `AMP_Post_Template` constructor where we initialize the other post data?

westonruter:
The abstraction makes sense here, but I don't see how the abstraction gets us anywhere beyond where we are today, at least with this PR alone. The missing piece seems to be a way to programmatically switch between `AMP_Template` subclasses depending on the post type. For example, when registering a custom post type there could be an `amp_template_class` property in the call to `register_post_type()`. This would be analogous to what is done with the REST API and `WP_REST_Posts_Controller` subclasses and the `rest_controller_class` property. See https://github.com/WordPress/wordpress-develop/blob/9f77ec1/src/wp-includes/rest-api.php#L179-L186

This property could then be used in `amp_render_post()`, where the registered post type of the could be looked up and then `amp_template_class` instantiated if it is set, or otherwise fall-back to 
`AMP_Post_Template`.

Alternatively, or additionally, there could be an `amp_template_class` filter that would allow a theme or plugin to supply their own subclass to `AMP_Template` dynamically.

westonruter:
We are going the route of pure WP templates when theme support is present (now in `develop`), so we don't need to build upon `AMP_Post_Trmplate` to achieve Canonical AMP.

PR #735, opened 2017-07-27
For normalizing urls with parameters

victorfreitas:


mjangda:
Thanks @victorfreitas! Can you share some details on what this is fixing? Some example URLs/permalinks structures would be helpful. It would also be nice to add some tests for this new function.

victorfreitas:
Hi @mjangda ,

When using parameter in permalink, the amp url breaks.

**Example:**

`<?php
$permalink = 'http://domain.com/category/post-name/?trace=1519021197&source=single';
$amp_url   = trailingslashit( $permalink ) . user_trailingslashit( AMP_QUERY_VAR, 'single_amp' );
?>`

> Result $amp_url is: http://domain.com/category/post-name/?trace=1519021197&source=single/amp/
> Correct is: http://domain.com/category/post-name/amp/?trace=1519021197&source=single

Thank you!

includes/amp-helper-functions.php
```diff
@@ -11,12 +11,28 @@ function amp_get_permalink( $post_id ) {
...
 
+/**
+ * For normalizing urls with parameters
+ */
+function amp_normalize_url_param( $url ) {
+	if ( false === strpos( $url, '?' ) ) {
+		return trailingslashit( $url ) . user_trailingslashit( AMP_QUERY_VAR, 'single_amp' );
+	}
+
+	$parts   = explode( '?', $url, 2 );
```
westonruter: Instead of exploding on `?`, this should use `wp_parse_url()` to parse out the URL parts and then either amend the query string with the `AMP_QUERY_VAR` param or amend the `AMP_QUERY_VAR` slug to the path.

westonruter: For example, it is possible that `?` appears in a fragment identifier: `http://example.com/#/?foo/bar`

westonruter:
This has been fixed in in 0.7 by opting to always use the `amp` query param if there are other query params in the permalink:

https://github.com/Automattic/amp-wp/blob/40d1945778c9917ed258e74595d504ad31719d39/includes/amp-helper-functions.php#L70-L71

PR #742, opened 2017-08-04
Force is_amp_endpoint to true when rendering.

mjangda:
If amp_render_post is called directly outside of the standard endpoint, is_amp_endpoint will return false, which is not ideal for any code that expects to run in an AMP context (e.g. inside shortcodes).

Let's force the value to be true while we render AMP content.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -78,9 +78,7 @@ function is_amp_endpoint() {
 		_doing_it_wrong( __FUNCTION__, sprintf( esc_html__( "is_amp_endpoint() was called before the 'parse_query' hook was called. This function will always return 'false' before the 'parse_query' hook is called.", 'amp' ) ), '0.4.2' );
 	}
 
-	$is_amp_endpoint = false !== get_query_var( AMP_QUERY_VAR, false );
-
-	return apply_filters( 'is_amp_endpoint', $is_amp_endpoint );
+	return false !== get_query_var( AMP_QUERY_VAR, false );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


includes/amp-helper-functions.php
```diff
@@ -44,7 +44,9 @@ function is_amp_endpoint() {
 		_doing_it_wrong( __FUNCTION__, sprintf( esc_html__( "is_amp_endpoint() was called before the 'parse_query' hook was called. This function will always return 'false' before the 'parse_query' hook is called.", 'amp' ) ), '0.4.2' );
 	}
 
-	return false !== get_query_var( AMP_QUERY_VAR, false );
+	$is_amp_endpoint = false !== get_query_var( AMP_QUERY_VAR, false );
```
amedina: Why not just getting the result of `get_query_var( AMP_QUERY_VAR, false );`?

tests/test-amp-render-post.php
```diff
@@ -21,4 +21,33 @@ public function test__valid_post() {
...
+		$user_id = $this->factory->user->create();
+		$post_id = $this->factory->post->create( array(
+			'post_author' => $user_id,
+		) );
+
+		$pre_is_amp_endpoint = is_amp_endpoint();
+
+		add_action( 'pre_amp_render_post', array( $this, '_check_is_amp_endpoint' ) );
+
+		// Need to use ob here since the method echos
```
amedina: Could we clarify this comment a little further?

includes/amp-helper-functions.php
westonruter: Since it can be either an empty string or ''1''. If a URL is requested as `/?amp` or `/?amp=1` then this should cause AMP to be enabled in both cases.

westonruter:
I think it's better to just make sure the AMP query var is set during rendering, instead of introducing a new filter. Some plugins may be looking for the presence of the query var rather than using the `is_amp_endpoint()` function.

PR #754, opened 2017-08-16
Strip out tags with bad/missing mandatory attrs

### LINKED ISSUE #747: Javascript Error in Google Console When Wordads Is Enabled
Two users reported this error in the Google Search Console when Wordads was enabled with AMP: 

`Accelerated Mobile Pages > User-authored JavaScript found on page (Critical 
issue). `

One user tracked it down to this:
```
<script><![CDATA[ 
(function(g){g.__ATA.initAd({sectionId:1304173541, width:300, 
height:250});})(window); 
]]></script> 
```
@dbspringer had a look and thinks it's something in the code here. Can someone take a look?

Internal reference: 627192-zen, 565431-hc

### LINKED ISSUE #752: User-authored JavaScript found on page (Critical issue)
I’ve recently(starting 08/07/2017) started getting this error in my webmaster console and I currently have 71 errors and still counting.

The google AMP test page shows that is it **NOT a valid AMP page** and don’t know what’s causing this as I didn’t do any recent changes for any of my plugins besides regular updates.

Can someone tell what’s going on and what should I do in order to solve this issue?

I see that my amp pages are getting deindexed from google results so this is critical.

Someof the urls that need attention are:

https://www.hlbenefits.com/cla-safflower-oil-reviews/amp/
https://www.hlbenefits.com/how-know-probiotic-supplements-working/amp/
https://www.hlbenefits.com/best-vegan-and-vegetarian-protein-sources/amp/

mjangda:
If I add a script tag to the body and it's not a JSON metadata type, we should strip out the tag. Without this fix, we strip out the bad attribute but leave the tag in which breaks validation.

Fixes #747, #752

### DIFF
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -83,7 +83,6 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 
 		// Remove nodes with non-whitelisted tags.
 		if ( ! $this->is_amp_allowed_tag( $node ) ) {
-
 			// If it's not an allowed tag, replace the node with it's children
 			$this->replace_node_with_children( $node );
 			// Return early since this node no longer exists.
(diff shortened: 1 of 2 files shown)


mjangda:
Test cases:

```
Empty script tags (should be removed): <script></script>

No type attr (should be removed): <script>adsbygoogle = window.adsbygoogle || []).push({});</script>

With type attr (should be removed): <script type="text/javascript">console.log( 'console!' );</script>

With type json (should stay): <script type="application/json">{}</script>
```

The first three are fixed with this (and stripped out). The fourth fails because DOMDocument wraps the script content in `CDATA`, which is not as big an issue and needs to be fixed separately.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -109,52 +108,62 @@ private function process_node( $node ) {
...
-					}
+			// If there is exactly one attr_spec with a max score, use that one.
+			if ( 1 == count( $spec_ids_sorted ) ) {
+				$attr_spec_list = $rule_spec_list_to_validate[ $spec_ids_sorted[0] ][AMP_Rule_Spec::ATTR_SPEC_LIST];
+			// Otherwise...
+			} else {
+				// This shoud not happen very often, but...
+				// If we're here, then we have no idea which spec should
+				// be used. Try to merge the top scoring ones and cross
+				// your fingers.
```
amedina: Let's rephrase this comment if possible.

PR #767, opened 2017-08-23
Allow the AMP submit-error and submit-success tags

stephywells:
Currently, the submit-error and submit-success are being removed from the div inside the form. This change allows them to stick around.


stephywells:
Nevermind. I'll go ahead and close this. It works as long as the submit-success div is not inside the fieldset.

PR #780, opened 2017-09-22
Make redirect for non-supported AMP pages a 301

jdevalk:
The 302 redirect for AMP to non-AMP cannot be cached and can also not transfer value properly for Google. Changing to a 301 fixes this.


mjangda:
Thanks @jdevalk!

PR #789, opened 2017-10-20
Require `$sanitizer_classes` to be an array

### LINKED ISSUE #788: AMP_Content_Sanitizer::sanitize assumes `$sanitizer_classes` will be an array
AMP_Content_Sanitizer::sanitize assumes `$sanitizer_classes` will be an array and does no validation.

```
Warning: Invalid argument supplied for foreach() in /var/www/wp-content/plugins/amp/includes/class-amp-content.php on line 100
```

joshbetz:
Require `$sanitizer_classes` to be an array because we use it in `foreach()`.

Fixes #788

### DIFF
--- a/includes/class-amp-content.php
+++ b/includes/class-amp-content.php
@@ -92,7 +92,7 @@ class AMP_Content {
 }
 
 class AMP_Content_Sanitizer {
-	public static function sanitize( $content, $sanitizer_classes, $global_args = array() ) {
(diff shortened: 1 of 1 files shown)


joshbetz:
Maybe we'd want to default that value to an empty array?

mjangda:
> Maybe we'd want to default that value to an empty array?

The method won't do much if you don't pass in a valid array or sanitizer class names so the type hint makes sense instead of a default empty array.

PR #790, opened 2017-10-23
Updating iframe placeholder. Using amp-image instead of a div.

bribeiro:


mjangda:
Thanks for the PR @bribeiro! Can you provide a link to the AMP docs where they state that an amp-img is preferred over a div for validation?

bribeiro:
Here it is, https://www.ampproject.org/docs/reference/components/amp-iframe 
This is also related to this issue:  https://github.com/Automattic/amp-wp/issues/783 
I'm sorry I didn't link it in my PR.

Cheers

amedina:
Thanks @bribeiro for the PR. Actually it does not make a difference to have a div or an amp-img element; any element would work well. The docs suggest an amp-img as a possibility, but do not recommend any specific element.

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -124,9 +124,11 @@ private function filter_attributes( $attributes ) {
 	}
 
 	private function build_placeholder( $parent_attributes ) {
-		$placeholder_node = AMP_DOM_Utils::create_node( $this->dom, 'div', array(
+		$placeholder_node = AMP_DOM_Utils::create_node( $this->dom, 'amp-img', array(
 			'placeholder' => '',
+			'layout' => 'fill',
 			'class' => 'amp-wp-iframe-placeholder',
+			'src' => get_stylesheet_directory_uri() . '/images/amp-iframe-placeholder.png'
```
westonruter: This isn't accounting for whether or not the image exists.

westonruter:
It might be nice to use `amp-img` pointing to a website thumbnail image service as the placeholder image. But it's probably better to just introduce a filter that would allow a theme to dynamically supply the placeholder content returned by `build_placeholder`, whether that be an `am-img` or a div, and then it can supply something dynamic and not assume the presence of a static file in a given theme.

bribeiro:
Thinking again, you're right @amedina, the `amp-img` would be great for videos, but not for every type of content.

bribeiro:
@westonruter makes sense.

westonruter:
Closing in favor of https://github.com/ampproject/amp-wp/pull/4292.

You can now specify a placeholder by putting an `img` as a child element of the `iframe`.

PR #793, opened 2017-11-03
Support extracting dimensions for single URLs

mjangda:
Back-compat for sites using the class directly and relying on it for extracting images for just a single URL.

Fixes issues when doing something like: `AMP_Image_Dimension_Extractor::extract( $image_url )` where `$image_url` is a string and throws `Invalid argument supplied for foreach()` warnings.


mdbitz:
Hi @westonruter Weston sorry this has sat, It's been on the back burner for quite a while. Let me or Mo know if any concerns.

Thanks.

westonruter:
See also #1314 which undoes a bit of this PR: https://github.com/Automattic/amp-wp/pull/1314#pullrequestreview-143667183

PR #795, opened 2017-11-05
Add code quality checking via phpcs, eslint, jscs, and jshint

westonruter:
* Limit checking of code to the lines that are modified in scope of the pull request; this will allow for adherence to linting rules to be applied incrementally in the context of PRs without having to fix everything up front in order to get a passing build.
* Fix i18n errors related to UnorderedPlaceholdersText and MissingArgDomain.
* Add [wp-dev-lib](https://github.com/xwp/wp-dev-lib/), which is added as submodule. 
* Re-use phpunit bootstrap from wp-dev-lib.
* Update Travis to use container-based environment and use caching to improve performance.


amedina:
Ready to merge.

PR #800, opened 2017-11-06
Support extra attributes for amp-iframe

pdufour:


westonruter:
@pdufour Would you please elaborate on the changes here? According to the spec, I don't that `amp-iframe` supports `heights` or `media` attributes: https://www.ampproject.org/docs/reference/components/amp-iframe#attributes

pdufour:
They are common attributes for all amp elements that support `layout=responsive`

https://www.ampproject.org/docs/reference/common_attributes#heights

http://jsbin.com/nacecaboqi/edit?html,output (Scroll down for iframe)

westonruter:
Can you provide some steps to reproduce the problem? I pasted the AMP HTML from http://jsbin.com/nacecaboqi/edit?html,output into post content and `heights` attribute was not stripped out.

westonruter:
Also, can you re-check in the `develop` branch now that #823 has been merged?

pdufour:
You're right, I'm on an old branch. It is working with the `heights` attribute. I assume the `iframe-sanitizer` is deprecated now?

westonruter:
No, `iframe-sanitizer` isn't deprecated, but the `heights` attribute is not HTML5 so it isn't something that would be sanitized from an `iframe` tag to convert to `amp-iframe`.

PR #811, opened 2017-11-23
Add admin screen for managing which post types have AMP support

### LINKED ISSUE #798: Allow Custom Post Types to be enabled via admin/customizer
**[Edited] Refer to the latest [acceptance criteria](https://github.com/Automattic/amp-wp/issues/798#issuecomment-346595030).**
___
As [noted](https://github.com/Automattic/amp-wp#custom-post-type-support) in the readme:

> By default, the plugin only creates AMP content for posts. You can add support for other post_types using the post_type parameter used when registering the custom post type (assume our post_type is xyz-review):

This requires someone to write code for adding support. However, this ability to opt-in to post type support could instead be presented in the admin (or in the Customizer) so that the user doesn't have to write any code at all.

The list of post types could be presented with checkboxes for deciding whether or not AMP support should be enabled. The checkbox for generic posts would be shown with a checked checkbox that is also disabled since support is built-in and cannot be disabled. After that a checkbox for pages could be presented, once we have pages listed and enabled by default (once we have #176) but then perhaps allow the user to disable support by unchecking it if they don't want it. Then each of the other public post types could be listed, with the checkboxes by default unchecked. Any post types that have hard-coded post type support in the theme or plugin would be disabled and checked.

ThierryA:
This PR add a "AMP -> Settings" submenu above the "Analytics" submenu (some refactoring was needed to achieve that). The new Settings page has the Post Types support settings allowing users to enable/disable AMP for custom post types. The "post" post type is always on as per the ACs.

[Refer to the ACs](https://github.com/Automattic/amp-wp/issues/798#issuecomment-346595030) for more info...

* [x] Update readme to indicate code no longer required.
* [x] Update unit tests.

Close #798


includes/admin/functions.php
```diff
@@ -80,18 +80,28 @@ function amp_add_customizer_link() {
...
 function amp_add_options_menu() {
 	if ( ! is_admin() ) {
 		return;
 	}
 
-	$show_options_menu = apply_filters( 'amp_options_menu_is_enabled', true );
-	if ( true !== $show_options_menu ) {
+	/**
+	 * Filter whether to enable the AMP settings.
+	 *
```
westonruter: Could you find when this filter was introduced and add a `@since` tag for the version?

includes/amp-helper-functions.php
```diff
@@ -18,8 +18,11 @@ function amp_get_permalink( $post_id ) {
 }
 
 function post_supports_amp( $post ) {
-	// Because `add_rewrite_endpoint` doesn't let us target specific post_types :(
-	if ( ! post_type_supports( $post->post_type, AMP_QUERY_VAR ) ) {
+	// Listen to post types settings.
+	$is_enabled_via_setting = (bool) AMP_Settings_Post_Types::get_instance()->get_settings_value( $post->post_type );
```
westonruter: Question: Instead of sourcing the enabled state via the settings directly, why not just add a separate action hook  that then just calls `add_post_type_support()` for each post type enabled be admin settings. This would allow plugins to query for AMP post type support via just `post_type_supports()` calls, and it could allow a plugin to turn off post type support for AMP via just `remove_post_type_support()`. Then this `post_supports_amp()` function could be left unchanged, as the plugin would essentially be providing a UI for doing the same `add_post_type_support()` calls that existing themes/plugins are doing today.

includes/settings/class-amp-settings-post-types.php
```diff
@@ -0,0 +1,133 @@
...
+	public function get_supported_post_types() {
+		$core = get_post_types( array(
+			'name' => 'post',
+		), 'objects' );
+		$cpt  = get_post_types( array(
+			'public'   => true,
+			'_builtin' => false,
+		), 'objects' );
+
+		return $core + $cpt;
```
westonruter: I think `array_merge()` should be used here. Using `+` with arrays can have unexpected results if the items are numerically-indexed: https://stackoverflow.com/a/7059731/93579

includes/settings/class-amp-settings-post-types.php
```diff
@@ -0,0 +1,133 @@
+<?php
+/**
+ * Post types settings.
+ *
+ * @package AMP
+ */
+
+/**
+ * Settings post types class.
```
westonruter: Let's add `@since` tags to the classes and methods being introduced.

includes/settings/class-amp-settings-post-types.php
```diff
@@ -0,0 +1,133 @@
...
+			'description' => __( 'Enable/disable AMP post type(s) support', 'amp' ),
+		);
+	}
+
+	/**
+	 * Initialize.
+	 */
+	public function init() {
+		add_action( 'admin_init', array( $this, 'register_settings' ) );
+		add_action( 'update_option_' . AMP_Settings::SETTINGS_KEY, 'flush_rewrite_rules' );
```
westonruter: Good you thought of this.

includes/settings/class-amp-settings-post-types.php
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	/**
+	 * Register the current page settings.
+	 */
+	public function register_settings() {
+		register_setting( AMP_Settings::SETTINGS_KEY, AMP_Settings::SETTINGS_KEY );
```
westonruter: This should add a `sanitize_callback` here for good measure, even though it is only ever read out with a cast to boolean.

includes/settings/class-amp-settings-post-types.php
```diff
@@ -0,0 +1,133 @@
...
+
+	/**
+	 * Get the instance of AMP_Settings_Post_Types.
+	 *
+	 * @return object $instance AMP_Settings_Post_Types instance.
+	 */
+	public static function get_instance() {
+		static $instance;
+
+		if ( ! $instance instanceof AMP_Settings_Post_Types ) {
```
westonruter: Operator precedence here makes me nervous. Can this be changed to:

```
! ( $instance instanceof AMP_Settings_Post_Types )
```

templates/admin/settings/fields/checkbox-post-types.php
```diff
@@ -0,0 +1,21 @@
...
+
+// Check referrer.
+if ( ! $this instanceof AMP_Settings_Post_Types ) {
+	return;
+}
+?>
+<fieldset>
+	<?php foreach ( $this->get_supported_post_types() as $post_type ) : ?>
+		<label>
+			<input type="checkbox" value="1" name="<?php echo esc_attr( $this->get_setting_name( $post_type->name ) ); ?>"<?php checked( true, (bool) $this->get_settings_value( $post_type->name ) || post_type_supports( $post_type->name, AMP_QUERY_VAR ) ); ?><?php disabled( true, post_type_supports( $post_type->name, AMP_QUERY_VAR ) ); ?>> <?php echo esc_html( $post_type->label ); ?>
```
westonruter: My comment about `is_enabled_via_setting` above will impact this. In particular, the `disabeld()` call would need to be changed to:

```php
disabled( true, post_type_supports( $post_type->name, AMP_QUERY_VAR ) && ! $this->get_settings_value( $post_type->name ) )
```

includes/amp-helper-functions.php
ThierryA: Good question, that is how I had it initially but here is the issue I faced. If `post_type_supports()` is set by a theme/plugin then the admin setting must be checked and disabled according to the AC. The `post_type_supports()` support is usually declared in a `init` action which runs before the admin checks so if we add `post_type_supports` based on the db value saved then the checkbox becomes disabled (unless we declare it in an action running later which doesn't feel right). So I added a flag to differentiate the `post_type_supports()` that was added by the setting and the one that would be declared by a theme/plugin to prevent this behaviour. The issue is that if one is saved in the db and then a plugin/theme add support for it, then it won't be following the intended behaviour.
I will take another stab at this right now and post and update.

ThierryA: Alright, [this commit](https://github.com/Automattic/amp-wp/pull/811/commits/09d3685989b114173389895614f9927a51947cb2) declare the `post_type_supports()` based on the admin settings. To do so, it runs it in a `after_setup_theme` callback so that it can be overwritten by theme/plugins which would declare `add_post_type_support()` or `remove_post_type_support()` invoked through an `init` action. 

Regarding the admin option disabled state, since we can't flag a post type which is not enabled by setting and removed by plugin/theme we can't disable the checkbox but the `AMP_Settings_Post_Types::errors()` takes care of this scenario if a user try to enable a post type support but a theme/plugin declares `remove_post_type_support()`.

One scenario that could be seen as an issue is if a plugin doesn’t have `add_post_type_support()` in its current version and the AMP settings are saved. If in the newer version the plugin add `add_post_type_support()` then the checkbox in the admin wouldn’t be disabled since it is flagged that the post type support is managed by the setting is saved in the db. For that reason, I would be in favor of removing the "inconsistent" disabled state and add an error handling in case a plugin/theme calls `add_post_type_support()` and a user tries to disable it. That would be a complete bullet proof solution. Thoughts?

amp.php
```diff
@@ -73,6 +72,19 @@ function amp_init() {
...
 
+/**
+ * Define AMP query var.
+ *
+ * This function must be invoked through the 'after_setup_theme' action to allow
+ * the AMP setting to declare the post types support earlier than plugins/theme.
+ *
+ * @since 0.6
+ */
+function define_query_var() {
```
westonruter: We should add hook docs for this filter here since we're touching it.

It was introduced in 0.3.2 via https://github.com/Automattic/amp-wp/commit/c486a1c62f10385bb32d66641a935b621fc3d821

includes/amp-post-types-support.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+function amp_custom_post_types_support() {
+	// Listen to post types settings.
+	foreach ( AMP_Settings_Post_Types::get_instance()->get_settings() as $post_type => $enabled ) {
+		if ( true === $enabled ) {
+			add_post_type_support( $post_type, AMP_QUERY_VAR );
+		}
+	}
+}
+add_action( 'after_setup_theme', 'amp_custom_post_types_support' );
```
westonruter: I wonder if this shouldn't have a lower priority, like `5`. 

The _s starter theme and Twenty Seventeen among others do their `add_theme_support` calls in `after_setup_theme` at priority 10. 
https://github.com/Automattic/_s/blob/efd37d13a6622bcd20e8c7b35309102c06981c71/functions.php#L10-L84
https://github.com/WordPress/wordpress-develop/blob/a825a181e11b1a89e76d79d22bd2c220f45b3741/src/wp-content/themes/twentyseventeen/functions.php#L20-L217

So if they also do calls to `add_post_type_support()` I'd expect them to do it here as well, for example: https://github.com/ryelle/Anadama-React/blob/21c8f550bc3ea6bbb7407891bf06845316b0ebf6/functions.php#L78-L82

includes/amp-helper-functions.php
westonruter: > If in the newer version the plugin add add_post_type_support() then the checkbox in the admin wouldn’t be disabled since it is flagged that the post type support is managed by the setting is saved in the db

What if the plugin instead had a constant that defined the post types that the AMP plugin supports by default? This could then be used in the `amp_core_post_types_support()`  function, and then when deciding whether or not to disable the checkbox it can look at the constant rather than (or in addition to?) doing the `post_type_supports()` call?

westonruter:
@ThierryA My reply to your comment got collapsed since it was part of a patch that had a change: https://github.com/Automattic/amp-wp/pull/811#discussion_r153298086

includes/amp-helper-functions.php
ThierryA: > One scenario that could be seen as an issue is if a plugin doesn’t have `add_post_type_support()` in its current version and the AMP settings are saved. If in the newer version the plugin add `add_post_type_support()` then the checkbox in the admin wouldn’t be disabled since it is flagged that the post type support is managed by the setting is saved in the db
Plugin wouldn't be able to change the constant unless we add a filter, hence my suggestion to add a filter via PM earlier this week. That said, this would take care of plugins which would want to declare `remove_post_type_support()` but have the checkbox disabled which is a pity if we are going to root of adding an extra flag.

If we are going to add an extra step, then I would vote for a filter like `amp_post_types_support_checkbox_disabled`. That way we can let the error handling taking care of overwrite restrictions on save if the plugin didn't add `amp_post_types_support_checkbox_disabled` and allow plugins to force disable the checkbox to enhance user experience. Alternatively, we could just not have the disabled state and let error handle all scenarios on save which will be functionality bullet proof and consistent. Thoughts?

westonruter: I'm having a hard time running through the scenarios in my head. Can we create a little sample plugin that registers some post types, enables support for some of them, and disables support for others?

ThierryA: @westonruter following on our discussion at WCUS, I pushed a commit which removed the disabled state and handle all scenarios using an errors handling on save. This covers all type of scenarios in a consistent and bullet proof way.

While it was nice to have the disable state on load from a user experience perspective, I believe the fact that it was so inconsistent (no way around it) didn't justify to have it and actually turned into a bad user experience as no one could really figure out what was going on. Please take it for a spin and let me know what you think.

ThierryA:
@westonruter this PR is good to go unless we decide to make other changes regarding the admin checkbox disable status (see [my reply here](https://github.com/Automattic/amp-wp/pull/811#discussion_r154537549)). Thanks,

westonruter:
I'm going to be pushing an update here to merge the new methods into the existing menu and options manager classes.

westonruter:
Unit tests need to be updated but I think 1befc79 will get us better set for future migration to Customizer while starting to unify option handling.

includes/options/class-amp-options-menu.php
```diff
@@ -1,54 +1,201 @@
...
+				<?php if ( $is_builtin ) : ?>
+					<input type="hidden" name="<?php echo esc_attr( $id ); ?>" value="1">
+				<?php endif; ?>
+				<input
+					type="checkbox"
+					value="1"
+					id="<?php echo esc_attr( $id ); ?>"
+					name="<?php echo esc_attr( $id ); ?>"
+					<?php checked( true, post_type_supports( $post_type->name, AMP_QUERY_VAR ) ); ?>
+					<?php disabled( $is_builtin ); ?>
```
westonruter: @ThierryA This is what I was talking about in regards to hard-coding a set of post types that are always active, and thus are disabled in the UI. Maybe we could instead just hide them altogether, but it turns out to be useful to have them present. The hidden input above ensures that there are some checked checkboxes which means we will get a non-empty array passed in to validate.

westonruter:
@ThierryA ok, this is ready for your final review and merge!

tests/test-class-amp-options-manager.php
```diff
@@ -0,0 +1,151 @@
...
+			$id => array(
+				'id'     => $id,
+				'type'   => 'foo',
+				'config' => '{"very_good":true}',
+				'delete' => true,
+			),
+		) );
+		$entries = AMP_Options_Manager::get_option( 'analytics' );
+		$this->assertCount( 1, $entries );
+		$this->assertArrayNotHasKey( $id, $entries );
```
ThierryA: Would be good to cleanup the option here not to leave traces behind which might affect other tests.

tests/test-class-amp-post-type-support.php
```diff
@@ -0,0 +1,82 @@
...
+		$this->assertEquals( array( 'post' ), AMP_Post_Type_Support::get_builtin_supported_post_types() );
+	}
+
+	/**
+	 * Test get_eligible_post_types.
+	 *
+	 * @covers AMP_Post_Type_Support::get_eligible_post_types()
+	 */
+	public function test_get_eligible_post_types() {
+		register_post_type( 'book', array(
```
ThierryA: These post types are not removed on tear down which could affect other tests.

tests/test-class-amp-post-type-support.php
```diff
@@ -0,0 +1,82 @@
...
+		);
+	}
+
+	/**
+	 * Test add_post_type_support.
+	 *
+	 * @covers AMP_Post_Type_Support::add_post_type_support()
+	 */
+	public function test_add_post_type_support() {
+		register_post_type( 'book', array(
```
ThierryA: These post types are not removed on tear down which could affect other tests.

includes/class-amp-post-type-support.php
```diff
@@ -0,0 +1,69 @@
...
+	 *
+	 * This function should only be invoked through the 'after_setup_theme' action to
+	 * allow plugins/theme to overwrite the post types support.
+	 *
+	 * @since 0.6
+	 */
+	public static function add_post_type_support() {
+		$post_types = array_merge(
+			self::get_builtin_supported_post_types(),
+			array_keys( array_filter( AMP_Options_Manager::get_option( 'supported_post_types', array() ) ) )
```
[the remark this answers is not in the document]
westonruter: True. Eventually I think we should just store an array of post types rather than an array mapping post type names to `true`. In fact, there's no reason why we can't do that now.

tests/test-class-amp-options-manager.php
westonruter: Unnecessary because tests in core automatically clean up after themselves: https://github.com/WordPress/wordpress-develop/blob/1c2b0762c7f1a91394c633ae812505919bd297ed/tests/phpunit/includes/testcase.php#L102-L182

tests/test-class-amp-post-type-support.php
westonruter: Unnecessary because tests in core automatically clean up after themselves: https://github.com/WordPress/wordpress-develop/blob/1c2b0762c7f1a91394c633ae812505919bd297ed/tests/phpunit/includes/testcase.php#L102-L182

tests/test-class-amp-post-type-support.php
westonruter: Unnecessary because tests in core automatically clean up after themselves: https://github.com/WordPress/wordpress-develop/blob/1c2b0762c7f1a91394c633ae812505919bd297ed/tests/phpunit/includes/testcase.php#L102-L182

includes/options/views/class-amp-options-manager.php
```diff
@@ -27,58 +152,46 @@ public static function update_option( $option, $value ) {
...
-			unset( $amp_analytics[ $entry_id ] );
-		} else {
-			$amp_analytics[ $entry_id ] = array(
-				'type' => $entry_vendor_type,
-				'config' => $entry_config,
-			);
-		}
+		// Redirect to keep the user in the analytics options page.
+		// Wrap in is_admin() to enable phpunit tests to exercise this code.
+		wp_safe_redirect( admin_url( 'admin.php?page=amp-analytics-options&settings-updated=1' ) );
```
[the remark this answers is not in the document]
westonruter: Yes, but this is how it is currently in the released version of the plugin: https://github.com/Automattic/amp-wp/blob/6df01e9c90f2de3c4cb16952401a285992d11021/includes/options/views/class-amp-options-manager.php#L42

I didn't want to spend more time on improving analytics settings in the admin since we want to move them to the Customizer.

includes/options/views/class-amp-options-manager.php
```diff
@@ -19,6 +63,87 @@ public static function get_option( $option, $default = false ) {
...
+		);
+
+		$options = array_merge(
+			$defaults,
+			self::get_options()
+		);
+
+		// Validate post type support.
+		if ( isset( $new_options['supported_post_types'] ) ) {
+			$options['supported_post_types'] = array();
```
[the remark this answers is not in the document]
westonruter: On 74 it is to define it if it is not defined, as a default. Here we explicitly reset the previously saved value before gathering the new post types.

includes/options/views/class-amp-options-manager.php
```diff
@@ -19,6 +63,87 @@ public static function get_option( $option, $default = false ) {
...
 
+	/**
+	 * Validate options.
+	 *
+	 * @param array $new_options Plugin options.
+	 * @return array Options.
+	 */
+	public static function validate_options( $new_options ) {
+		$defaults = array(
+			'supported_post_types' => array(),
```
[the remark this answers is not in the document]
westonruter: Yes, that would be an improvement. Not a blocker though.

includes/options/class-amp-options-menu.php
```diff
@@ -1,54 +1,184 @@
...
+				$id         = AMP_Options_Manager::OPTION_NAME . "[supported_post_types][{$post_type->name}][]";
+				$is_builtin = in_array( $post_type->name, $builtin_support, true );
+				?>
+				<?php if ( $is_builtin ) : ?>
+					<input type="hidden" name="<?php echo esc_attr( $id ); ?>" value="1">
+				<?php endif; ?>
+				<input
+					type="checkbox"
+					value="1"
+					id="<?php echo esc_attr( $id ); ?>"
```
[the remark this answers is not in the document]
westonruter: I don't see the need to change. We should embrace the future 😄

includes/options/class-amp-options-menu.php
```diff
@@ -1,54 +1,184 @@
...
+			}
 		}
 	}
+
+	/**
+	 * Post types support section renderer.
+	 *
+	 * @since 0.6
+	 */
+	public function render_post_types_support() {
```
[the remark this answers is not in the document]
westonruter: Yeah, it's because we eventually want to move to Customizer and so adding another class temporarily for rendering HTML would seem undesirable if we're just going to delete it in the next release.

includes/options/class-amp-options-menu.php
```diff
@@ -1,54 +1,184 @@
...
-	function remove_toplevel_menu_item() {
-		global $submenu;
-		if ( isset( $submenu['amp-plugin-options'][0] ) ) {
-			unset( $submenu['amp-plugin-options'][0] );
+	/**
+	 * Check for errors with updating the supported post types.
+	 *
+	 * @since 0.6
+	 */
+	protected function check_supported_post_type_update_errors() {
```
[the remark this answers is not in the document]
westonruter: Yes, it's still needed because the error checking can only be done for these errors after WordPress finishes initializing, since we have to wait for themes and plugins to have called `add_post_type_support()` or `remove_post_type_support()` in order to detect whether there has been an error scenario.

includes/options/class-amp-options-menu.php
```diff
@@ -1,54 +1,184 @@
...
 
+		add_submenu_page(
+			AMP_Options_Manager::OPTION_NAME,
+			__( 'AMP Settings', 'amp' ),
+			__( 'General', 'amp' ),
+			'manage_options',
+			AMP_Options_Manager::OPTION_NAME
+		);
+
+		add_settings_section(
```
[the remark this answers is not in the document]
westonruter: That's a good question. I debated that. I think I settled on this because the sections and fields only make sense in the context of the pages that show them, whereas the setting itself would be relevant both on the admin pages _and_ the Customizer.

includes/class-amp-post-type-support.php
```diff
@@ -0,0 +1,69 @@
...
+ * Class AMP_Post_Type_Support.
+ */
+class AMP_Post_Type_Support {
+
+	/**
+	 * Add hooks.
+	 */
+	public static function init() {
+		add_action( 'after_setup_theme', array( __CLASS__, 'add_post_type_support' ), 5 );
+		if ( did_action( 'after_setup_theme' ) ) {
```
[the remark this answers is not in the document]
westonruter: When a plugin invokes `init`. I don't have a scenario where this would happen, so we can remove it.

includes/options/class-amp-options-menu.php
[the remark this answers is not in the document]
ThierryA: That makes sense, thanks for sharing your thoughts in that.

includes/options/class-amp-options-menu.php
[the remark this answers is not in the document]
ThierryA: So my bad, what I meant was why having this method here (structurally) as opposed to moving it to the `AMP_Options_Manager` class where the other error handlings are.

includes/options/class-amp-options-menu.php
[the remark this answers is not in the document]
ThierryA: Let's embrace the future then 😗

tests/test-class-amp-post-type-support.php
ThierryA: I might be missing something but I don't see where the registered post types are removed, for instance if remove the post type registration in `test_add_post_type_support()` the tests still pass.

westonruter: I ended up explicitly unregistering the post types in 8c140631320654af9d2ba2316731feb61c69374d because it seems core isn't fully cleaning up after itself as I thought.

westonruter:
@ThierryA there is nothing left outstanding here I believe.

ThierryA:
Thanks @westonruter, looking great. This is good to go!

Issue #709, opened 2017-06-08
Block AMP on specific pages


doodirock:
**[Edited] Refer to the latest [acceptance criteria](https://github.com/Automattic/amp-wp/issues/709#issuecomment-346621232).**
___
We have a lot of cases where AMP wont work because of some interactive content we have on a particular page.  Is there any way to disable amp in the rel link base on a post ID?

nicopujol:
We implemented something similar but the other way around: we insert the interactive content via the Ad Inserter plugin on regular pages. The plugin allows you to blacklist (not load) the content for certain pages, in this case we black listed */amp/ for that content not to load on AMP pages.

doodirock:
Great idea, however instead I do not want any AMP page at all to load.  Not just the interactive content.

westonruter:
@doodirock There is a filter you can use to disable AMP support for a given post.

Let's say you add a custom field in postmeta called `amp_skipped` which if it has an non-empty value it should cause the post to get skipped:

![image](https://user-images.githubusercontent.com/134745/32515106-dd643a46-c3b3-11e7-991c-3e66d468ec6b.png)

You can then add support for that custom postmeta by adding the following to your theme or plugin:

```php
add_filter( 'amp_skip_post', function( $skipped, $post_id ) {
	if ( get_post_meta( $post_id, 'amp_skipped', true ) ) {
		$skipped = true;
	}
	return $skipped;
}, 10, 2 );
```

So then the question remains for this issue as to whether there should be an actual UI in the AMP plugin to offer “Disable AMP paired mode”. There could be a checkbox in the publish metabox, for example, that offered this.

westonruter:
@amedina Having a toggle like this may be critical for adding support for pages (#176), since pages often have ad hoc structure. We could potentially use the opt-out checkbox instead of forcibly skipping AMP for pages that have custom templates selected.

westonruter:
There could be an interesting crossover with the AMP Validator here and we could essentially auto-populate the checkbox. Consider this:

1. I add an embed to the content that is not supported by AMP.
2. I update the post/page.
3. After updating the source of the AMP page is scraped and sent off to the validator
4. If there are validation errors, then the AMP checkbox would be unchecked and disabled, and a warning notice could appear notifying them that AMP is not available because of certain embeds.
5. After removing the offending embeds and saving, then the validation would happen again and if it passes, the warning would be removed and the checkbox would again be enabled, giving the user the option to forcibly override it or not.

[...]

PR #812, opened 2017-11-24
#799 Add AMP preview button

### LINKED ISSUE #799: Add post preview support for AMP
**[Edited] Refer to the latest [acceptance criteria](https://github.com/Automattic/amp-wp/issues/799#issuecomment-346622668).**
___
There is currently no way to preview how a post will look in AMP mode prior to publishing. We could add an AMP button next to existing Preview button which opens post with `?amp=1` in URL. Given post url of `http://example.com/?p=2043&preview=true` you just add `&amp=1` to it, then you get the AMP version of the post preview in your desktop browser.

Since “mobile” specifically in AMP itself is obsolete (as it is just as relevant for tablet and desktop), it is important to see how AMP looks across all devices sizes (Canonical AMP). If a user wants to see what mobile looks like they can just resize their browser window (or optionally include device preview buttons in admin bar like in Customizer).

ThierryA:
This PR adds the AMP preview button on posts which have AMP support.

amp preview button

[Refer to the ACs](https://github.com/Automattic/amp-wp/issues/799#issuecomment-346622668) for more info...

Close #799


assets/css/amp-post-meta-box.css
```diff
@@ -0,0 +1,22 @@
+/**
```
ThierryA: It seems overkill to enqueue a file for this small amount of CSS but it is done that way in preparation of #709.

assets/js/amp-post-meta-box.js
```diff
@@ -0,0 +1,57 @@
...
+		data: {},
+
+		/**
+		 * Boot plugin.
+		 *
+		 * @since 0.6
+		 * @param {Object} data Object data.
+		 * @return {void}
+		 */
+		boot: function( data ) {
```
ThierryA: It seems overkill to have a `boot()` method just to call `addPreviewButton()` method but it is structured that way in preparation of #709.

assets/js/amp-post-meta-box.js
```diff
@@ -0,0 +1,57 @@
...
+
+			$previewBtn
+				.clone()
+				.insertAfter( $previewBtn )
+				.addClass( 'amp-preview' )
+				.prop( {
+					'href': this.data.previewLink,
+					'id': 'amp-' + $previewBtn.prop( 'id' ),
+					'target': 'amp-' + $previewBtn.prop( 'target' )
+				} )
```
westonruter: This isn't working yet with autosave revisions. If you try clicking the AMP preview before an autosave happens, then the opened window will not show the changes you've made.

This logic needs to be incorporated: https://github.com/WordPress/wordpress-develop/blob/a825a181e11b1a89e76d79d22bd2c220f45b3741/src/wp-admin/js/post.js#L388-L418

assets/css/amp-post-meta-box.css
```diff
@@ -0,0 +1,22 @@
...
+/* Core preview button */
+#preview-action.has-next-sibling .preview:first-of-type {
+    border-top-right-radius: 0;
+	border-bottom-right-radius: 0;
+	float: none;
+}
+
+/* AMP preview button */
+#preview-action.has-next-sibling .preview.amp-preview {
+    border-top-left-radius: 0;
```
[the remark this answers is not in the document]
ThierryA: Oops, thanks for picking that one up.

I just switch to Atom last week and I guess this is the first flaw I see. For those who use Atom here is what I see by default.
atom issue
*Tips:* in settings->editor enable "Show invisible" and then you get 
screen shot 2017-11-28 at 10 35 34 am
which is already an improvement. But better, install "[Tabs to Space](https://atom.io/packages/tabs-to-spaces)" package and set it on save which will automatically convert such inconvenience.

includes/admin/class-amp-post-meta-box.php
```diff
@@ -0,0 +1,67 @@
...
+	public function init() {
+		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_admin_assets' ) );
+	}
+
+	/**
+	 * Enqueue admin assets.
+	 *
+	 * @since 0.6
+	 */
+	public function enqueue_admin_assets() {
```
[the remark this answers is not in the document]
ThierryA: I am not sure if I am missing something but from my perspective it is not loading on all pages since it checks if the `$post->type` is set and the `post_supports_amp()`. It does load on the `edit.php`, `post.php` and `post-new.php` if the post type as AMP support. That said it isn't really necessary to load it on `edit.php` so I will add a check to strictly load it on `post.php` and `post-new.php`.

assets/js/amp-post-meta-box.js
ThierryA: Good catch! To prevent rewriting core functionalities and apply DRY principals, I added some logic to trigger the Core Preview but but append the amp query var at the end if the AMP preview button is clicked. That way the events propagate correctly and it also keeps the preview in the same target which is probably better from a user experience perspective.

includes/admin/class-amp-post-meta-box.php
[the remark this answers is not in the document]
westonruter: @ThierryA oh yes, I see you're right. and that it wouldn't get enqueued. But it would get enqueued if an admin screen happened to define a global `$post` and didn't clean it up. So yeah, I think it is best to explicitly check if the `current_screen` is the post editor. Maybe `'post' === get_current_screen()->base` would be a good check?

assets/js/amp-post-meta-box.js
ThierryA: @westonruter [this commit](https://github.com/Automattic/amp-wp/pull/812/commits/779f76317301c3ae1843936d697b58b92cc2bd16) address this issue, it should be working as expected now.

ThierryA:
@westonruter kindly review the changes, this should be ready to merge.

westonruter:
Closing in favor of #813, as I got the two branches intertwined.

PR #813, opened 2017-11-27
Add post preview for AMP and allow AMP to be disabled on a per-post basis

### LINKED ISSUE #709: Block AMP on specific pages

### LINKED ISSUE #799: Add post preview support for AMP

ThierryA:
This PR adds the AMP preview button on posts which have AMP support.

amp preview button

[Refer to the ACs](https://github.com/Automattic/amp-wp/issues/799#issuecomment-346622668) for more info...

Merges PR #812. ~This PR is branched off `feature/799-post-preview` (#812) which is currently pending for review. Once #812 is reviewed and merged, the `develop` must be merged into this branch and this PR may be reviewed.~

Fixes #709.
Fixes #799.


includes/admin/class-amp-post-meta-box.php
```diff
@@ -0,0 +1,159 @@
...
+	 * @since 0.6
+	 * @param object $post \WP_POST object.
+	 */
+	public function render_status( $post ) {
+		$verify = (
+			isset( $post->ID )
+			&&
+			isset( $post->post_type )
+			&&
+			post_type_supports( $post->post_type, AMP_QUERY_VAR )
```
ThierryA: This differ from the acceptance criteria, I would suggest to add the ability to disable AMP on a per post basis for all post types which have post type support set, not only pages (especially now that WordPress Core has [Post Type Templates)](https://make.wordpress.org/core/2016/11/03/post-type-templates-in-4-7/). Thoughts @amedina @westonruter?

ThierryA: This differ from from the acceptance criteria (point 9) in a sense that the AMP status option would not be displayed at all if the post type does not have AMP support set to avoid visual noise. The thinking behind this is that the AMP status (enabled/disabled) should not be editable for post types (page included) if it doesn't have AMP support set.

If we really want to show the status as disabled, one approach would be to change to edit section with a text indicating that AMP post type support is not enabled for the current post with a link to the AMP settings page where user may enable the post type support.

Thoughts @amedina @westonruter?

westonruter:
I've temporarily changed the base branch from `develop` to `feature/799-post-preview` for the sake of review.

includes/admin/class-amp-post-meta-box.php
```diff
@@ -64,4 +91,69 @@ public function enqueue_admin_assets() {
 		) );
 	}
 
+	/**
+	 * Render AMP status.
+	 *
+	 * @since 0.6
+	 * @param object $post \WP_POST object.
```
westonruter: This line should read:

```
* @param WP_Post $post Current post.
```

includes/admin/class-amp-post-meta-box.php
```diff
@@ -64,4 +91,69 @@ public function enqueue_admin_assets() {
...
+			! wp_is_post_revision( $post_id )
+			&&
+			! wp_is_post_autosave( $post_id )
+		);
+
+		if ( true === $verify ) {
+			update_post_meta(
+				$post_id,
+				self::POST_META_KEY,
+				sanitize_key( wp_unslash( $_POST[ self::POST_META_KEY ] ) )
```
westonruter: Note that this line will continue to be valid in WPCS once https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1222 is merged because `sanitize_key()` is marked in that PR as an `autoSlashingFunctions` (because it doesn't return any chars that will include slashes).

includes/admin/class-amp-post-meta-box.php
```diff
@@ -16,17 +16,44 @@ class AMP_Post_Meta_Box {
...
+	 * @since 0.6
+	 * @const string
+	 */
+	const NONCE_NAME = 'amp-status';
+
+	/**
+	 * The nonce action.
+	 *
+	 * @since 0.6
+	 * @const string
```
westonruter: I think `@var` is actually what should be used here, counter-intuitively. See https://stackoverflow.com/a/18446766/93579

assets/css/amp-post-meta-box.css
```diff
@@ -20,3 +20,27 @@
...
 }
+
+/* AMP status */
+.misc-amp-status > i {
+	float: left;
+	background: transparent url( '../images/amp-icon.svg' ) no-repeat left;
+	background-size: 17px;
+	width: 17px;
+    height: 17px;
+    margin: 0 8px 0 1px;
```
westonruter: Looks like some inconsistent indentation here.

templates/admin/amp-status.php
```diff
@@ -0,0 +1,34 @@
...
+ * @package AMP
+ */
+
+// Check referrer.
+if ( ! ( $this instanceof AMP_Post_Meta_Box ) ) {
+	return;
+}
+?>
+<div class="misc-pub-section misc-amp-status">
+	<i></i>
```
westonruter: Why an `i` here and not a `span`?

includes/amp-helper-functions.php
```diff
@@ -23,6 +23,11 @@ function post_supports_amp( $post ) {
 		return false;
 	}
 
+	// Listen to post meta.
+	if ( ! isset( $post->ID ) || 'disabled' === get_post_meta( $post->ID, AMP_Post_Meta_Box::POST_META_KEY, true ) ) {
+		return false;
```
westonruter: Question: instead of hard-coding a `return false` here, what if we instead here did:

```php
$skip = true;
```

If the default value for `$skip` were `false`, then this `$skip` could then be passed in below as value for the `amp_skip_post` filter. This would allow plugins to still be able to override the value.

_But_ maybe that would be a premature optimization. It would also be bad if a user indicates that they want to disable AMP for a given post, but then it still ends up getting AMP enabled on the frontend due to a filter that they don't know about.

includes/admin/class-amp-post-meta-box.php
westonruter: @ThierryA I think you're right. The AMP status toggle shouldn't be shown at all if the post type doesn't support AMP to begin with.

includes/admin/class-amp-post-meta-box.php
ThierryA: Good to know 😄  Would you like to have it removed and add a whitelisting comment until it is merged?

includes/amp-helper-functions.php
ThierryA: > But maybe that would be a premature optimization. It would also be bad if a user indicates that they want to disable AMP for a given post, but then it still ends up getting AMP enabled on the frontend due to a filter that they don't know about.

That would justified if we want to allow plugins to overwrite that admin value. That said doing that wouldn't make the admin option reflect the correct value so it would be better for plugins to use the `get_post_metadata` filter which would cover both, admin and frontend.

templates/admin/amp-status.php
ThierryA: I guess there have been a debate for a while as "what should be used for icons". Font awesome, Facebook etc. exclusively use the `i` tag for icons for example. I am personally in favor of using the `i` tag because it adds a semantic meaning (even though it doesn't mean icon per se) while `span` doesn't.
That said, the various places in the WP admin which are not using the dashicon font use the `span` tag so I think it would be better to keep it as consistent as possible. [This commit](https://github.com/Automattic/amp-wp/pull/813/commits/e4f6c8733ebadf0766242a7c6e39e04393ca15e2) changes the `i` tag to a `span`.

ThierryA:
@westonruter unless we decide to [change this](https://github.com/Automattic/amp-wp/pull/813#discussion_r154534193), this PR is ready to go.

westonruter:
Merge conflict needs to be resolved.

ThierryA:
Thanks @westonruter, all your changes are approved. I pushed two other commits:
- 802bcfb which improves the preview button styling, specifically making sure that the preview has the same background color (previously white) as the preview button and keep the exact same disabled styling (the border was previously lighter as the entire button opacity was reduced rather than just the background image).
- 1b6cc54 which is a slight code improvement to apply DRY principals and avoid using and extra variable unnecessary.

Ready for review final review and merge.

PR #819, opened 2017-12-06
#796 Move AMP Customizer into main Customizer

### LINKED ISSUE #796: Merge AMP Customizer with main Customizer
**[Edited] Refer to the latest [acceptance criteria] https://github.com/Automattic/amp-wp/issues/796#issuecomment-347296271**
___

Currently if you want to Customizer your AMP template you have to go to Appearance > AMP in the admin, and this will open the Customizer in a blank slate with only the AMP panel registered. The latest post is loaded in to the preview, and navigation of the site is disabled. There are a couple problems with this:

1. It is difficult to discover the AMP Customizer if you don't know specifically to look for that Appearance > AMP option.
2. If you add AMP support for other post types, then you can't preview how they look in the Customizer since they cannot be navigated to.

So I propose that we merge the AMP panel into the regular Customizer. We can allow the user to navigate around the entirety of their site in the Customizer preview, but then depending on whether or not the AMP panel is expanded then automatically toggle the preview URL to include the AMP version if it is available. If AMP is not available for a given URL then the Customizer can show a notification in the AMP panel or any of its sub-sections that AMP is not available, and that they should navigate to a URL that is AMP compatible. The notification could have a link to take them to the most recent post in the preview.

There can still be an Appearance > AMP link in the admin menu, but it can just deep-link to the AMP panel and supply the `url` of the most recent post as the preview to start with.

I consider this to be a dependency for adding page support (#176), because if you can't navigate to pages (or other custom post types) in the Customizer, then you can't customize their appearance.

kaitnyl:
This PR removed the previous AMP Customizer that was fixed to one static post. It merges the AMP Customizer options into the main Customizer and provides a toggle to view all posts and pages if an AMP version is available. 

[Acceptance criteria is here](https://github.com/Automattic/amp-wp/issues/796#issuecomment-347296271).

Close #796 

- [x] Make sure admin menu links to proper location.
- [x] Add selective refresh previewing of site title changes.
- [ ] ~Add site icon previewing.~


assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,194 @@
...
+		// Adding checkbox toggle before device selection.
+		$( '.devices-wrapper' ).before( '<label class="amp-toggle">' +
+			'<input type="checkbox">' +
+			'<span class="slider"></span>' +
+		'</label>' );
+
+		// Notification tooltip for AMP toggle.
+		$( '.amp-toggle input' ).before( '<span class="tooltip">' +
+			'This page is not AMP compatible.<br>' +
+			'<a data-post="' + window.ampPost + '">Navigate to an AMP compatible page</a>' +
```
westonruter: This needs be translated.

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,194 @@
...
+		} )( api.previewer.previewUrl.validate );
+
+		// Adding checkbox toggle before device selection.
+		$( '.devices-wrapper' ).before( '<label class="amp-toggle">' +
+			'<input type="checkbox">' +
+			'<span class="slider"></span>' +
+		'</label>' );
+
+		// Notification tooltip for AMP toggle.
+		$( '.amp-toggle input' ).before( '<span class="tooltip">' +
```
westonruter: We should to use `wp.template` here instead of string concatenation. If the template is output via PHP we can also then make use of i18n in PHP to get the translated string.

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,194 @@
...
+
+		/**
+		 * Central function to control UI toggle elements.
+		 *
+		 * @param {bool} checked If toggle input should be checked.
+		 * @param {bool} disabled If toggle input should be disabled.
+		 * @param {bool} notification If tooltip notification should appear.
+		 * @return {void}
+		 */
+		function toggleControl( checked, disabled, notification ) {
```
westonruter: Can this function take an object as its sole param instead of three positional params? I think that will make it easier to read.

includes/admin/class-amp-customizer.php
```diff
@@ -127,41 +71,43 @@ public function register_settings() {
...
-			array( 'jquery', 'customize-preview', 'wp-util' ),
+			amp_get_asset_url( 'js/amp-customize-controls.js' ),
+			array( 'jquery', 'customize-controls' ),
 			$version = false,
-			$footer = true
+			$footer  = true
 		);
 
-		do_action( 'amp_customizer_enqueue_preview_scripts', $this->wp_customize );
+		wp_localize_script( 'amp-customizer', 'ampPost', amp_admin_get_preview_permalink() );
```
westonruter: I'm seeing this output a variable containing `http://src.wordpress-develop.dev/?preview=true&p=53/amp/`. It seems `amp_admin_get_preview_permalink()` has a bug with how it adds the endpoint.

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,194 @@
...
+			if ( url.length ) {
+				api.previewer.previewUrl.set( setCurrentAmpUrl( url ) );
+				toggleControl( false, false, false );
+			}
+		} );
+
+		// Main control for toggling AMP preview.
+		$( '#customize-footer-actions' ).on( 'change', '.amp-toggle input', function( event ) {
+			ampToggle = $( this ).is( ':checked' );
+			toggleAmpView();
```
westonruter: I don't think the toggle should be toggle-able if the currently-previewed URL is not AMP-compatible. It should be `disabled` in that case, where clicking it just then shows the tooltip.

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,194 @@
...
+			var url = $( this ).data( 'post' );
+			if ( url.length ) {
+				api.previewer.previewUrl.set( setCurrentAmpUrl( url ) );
+				toggleControl( false, false, false );
+			}
+		} );
+
+		// Main control for toggling AMP preview.
+		$( '#customize-footer-actions' ).on( 'change', '.amp-toggle input', function( event ) {
+			ampToggle = $( this ).is( ':checked' );
```
westonruter: I think the state here should be moved to `wp.customize.state`. For example, at the top of this file do:

```js
wp.customize.state.add( 'ampEnabled', new wp.customize.Value() );
wp.customize.state.add( 'ampAvailable', new wp.customize.Value() );
```

The `ampEnabled` state can manage whether the toggle is on or off, whereas the `ampAvailable` state can manage whether or not AMP is available in the preview. Then it can be populated via messages from the preview via:

```js
wp.customize.bind( 'ready', function( available ) {
    wp.customize.previewer.bind( 'ampAvailable', function( available ) {
        wp.customize.state( 'ampAvailable' ).set( available );
    } );
} );
```

Then the `disabled` state of the toggle can be managed by the state like so:

```js
wp.customize.state( 'ampAvailable' ).bind( function( available ) {
     $( '.amp-toggle input' ).prop( 'disabled', available );
} );
```

As noted above about sending messages from the preview about whether AMP is supported, there can then be a message receiver that then updates this state. In the preview it could do:

```js
var ampAvailable = <?php echo wp_json_encode( is_singular() && post_supports_amp( get_queried_object() ) ); ?>;
//...
wp.customize.bind( 'preview-ready', function() {
    wp.customize.preview.send( 'ampAvailable', ampAvailable );
} );
```

Clicking on the checkbox can just change the `ampEnabled`state, something like this:

```js
$( '.amp-toggle input' ).on( 'click', function( event ) {
    event.preventDefault();
    wp.customize.state( 'ampEnabled' ).set( ! wp.customize.state( 'ampEnabled' ).get() );
} );
```

And likewise the state of the button then can then in turn be managed by this state as well, as well as then setting the `previewUrl`:

```js
wp.customize.state( 'ampEnabled' ).bind( function( enabled ) {
    $( '.amp-toggle input' ).prop( 'checked', enabled );
    toggleAmpView();
} );
```

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,194 @@
...
+				var val = prevValidate.call( this, value );
+				if ( val ) {
+					val = setCurrentAmpUrl( val );
+				}
+				return val;
+			};
+		} )( api.previewer.previewUrl.validate );
+
+		// Adding checkbox toggle before device selection.
+		$( '.devices-wrapper' ).before( '<label class="amp-toggle">' +
```
westonruter: This can make use of a `wp.template`.

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	/**
+	 * Hook up all AMP preview interactions once panel is ready.
+	 *
+	 * @param {wp.customize.Panel} panel The AMP panel.
+	 * @return {void}
+	 */
+	function panelReady( panel ) {
+		var ampToggle = false;
```
westonruter: Remove this in favor of referring to `wp.customize.state( 'ampEnabled' )`

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,194 @@
...
+
+		/**
+		 * Make current URL AMPified if toggle is on.
+		 *
+		 * @param {string} url URL.
+		 * @return {string} AMPified URL.
+		 */
+		function setCurrentAmpUrl( url ) {
+			if ( ! ampToggle && isAmpUrl( url ) ) {
```
westonruter: The `ampToggle` here could be defined locally as:

```js
var ampToggle = wp.customize.state( 'ampEnabled' ).get() && wp.customize.state( 'ampAvailable' ).get();
```

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,194 @@
...
+			}
+			return url;
+		}
+
+		/**
+		 * Swap to AMP version of URL in preview.
+		 *
+		 * @return {void}
+		 */
+		function toggleAmpView() {
```
westonruter: This might be better named `updatePreviewUrl`

includes/admin/class-amp-customizer.php
westonruter: Fixed in 0885b48

kaitnyl:
@westonruter, with regards to [your comment here](https://github.com/Automattic/amp-wp/pull/819#discussion_r155152263), I think we should have a better way of identifying if the page can be AMPified. The current approach waits until the preview attempts to load the AMP'd page and then decides based on content that's returned to the preview (if the page has AMP metadata). 

As the user navigates, we should detect if the current page can be AMPified and enable/disable the toggle before they attempt to use it. This would allow us to disable the AMP toggle before attemping to load the AMP URL in the preview and prevent that initial attempt to display the AMP version. 

I'm thinking once the preview is ready/changes URLs, make an Ajax call that would enable or disable the bottom AMP toggle. The downside would be more backend communication as the user navigates, but upside is it would disable the instance where a user attempts to toggle on a page that isn't AMPified resulting in the preview trying to show it and then quickly changing back (2 refreshes). 

Any thoughts/opinions on this?

assets/js/amp-customize-controls.js
kaitnyl: This function can likely be removed altogether due to swapping to `wp.customize.state` 😄

westonruter:
@kaitnyl I think the better way is outlined in https://github.com/Automattic/amp-wp/pull/819#discussion_r155155742. I don't see a need to do an additional Ajax request when the preview URL changes because the preview window is going to be issuing a request anyway, and when the preview loads, it can just send a message to the controls app to indicate whether or not AMP is available. If it is not available, then the toggle can be disabled. With the toggle disabled, then there would be no attempt to load the AMPed version and so no refreshes.

assets/js/amp-customize-controls.js
```diff
@@ -1,19 +1,27 @@
...
 	 */
 	function isAmpUrl( url ) {
-		var urlParser = document.createElement( 'a' );
+		var urlParser = document.createElement( 'a' ),
+			regexParam = new RegExp( "(^|\\?|&)" + ampVars.query + "=1(?=&|$)" ),
+			regexEndpoint = new RegExp( "\\/" + ampVars.query + "\\/?$" );
+
 		urlParser.href = url;
-		if ( /(^|\?|&)amp=1(?=&|$)/.test( urlParser.search ) ) {
+		if ( regexParam.test( urlParser.search ) ) {
```
westonruter: Instead of `regexParam` you can make use of `wp.customize.utils.parseQueryString()` and do something like:

```js
if ( ! _.isUndefined( wp.customize.utils.parseQueryString( urlParser.search.substr( 1 ) )[ ampVars.query ] ) ) {
```

assets/js/amp-customize-controls.js
```diff
@@ -23,10 +31,13 @@
...
-		var urlParser = document.createElement( 'a' );
+		var urlParser = document.createElement( 'a' ),
+			regexParam = new RegExp( "(^|\\?|&)" + ampVars.query + "=1" ),
+			regexEndpoint = new RegExp( "\\/" + ampVars.query + "\\/?$" );
+
 		urlParser.href = url;
-		urlParser.pathname = urlParser.pathname.replace( /\/amp\/?$/, '' );
-		urlParser.search = urlParser.search.replace( /(^|\?|&)amp=1/, '' );
+		urlParser.pathname = urlParser.pathname.replace( regexEndpoint, '' );
+		urlParser.search = urlParser.search.replace( regexParam, '' );
```
westonruter: Per above, instead of `regexParam` I suggest doing something like:

```
if ( urlParser.search.length > 1 ) {
    var params = wp.customize.utils.parseQueryString( urlParser.search.substr( 1 );
    delete params[ ampVars.query ];
    urlParser.search = $.param( params );
}
```

This is similar to what is done in core: https://github.com/WordPress/wordpress-develop/blob/72b74509f3bfb71c884bf8cd9f580c026079498a/src/wp-admin/js/customize-controls.js#L6547-L6559

includes/admin/class-amp-customizer.php
```diff
@@ -86,7 +87,14 @@ public function add_customizer_scripts() {
 			$footer  = true
 		);
 
-		wp_localize_script( 'amp-customizer', 'ampPost', amp_admin_get_preview_permalink() );
+		wp_localize_script( 'amp-customizer', 'ampVars', array(
```
westonruter: Instead of adding a global variable, I suggest that the `amp-customizer` script export a boot method which we can then pass an inline script into. For example: 

https://github.com/Automattic/amp-wp/blob/eeb502b28807c1be80e59937fab6112e406b204b/assets/js/amp-post-meta-box.js#L51-L59

https://github.com/Automattic/amp-wp/blob/eeb502b28807c1be80e59937fab6112e406b204b/includes/admin/class-amp-post-meta-box.php#L95-L111

This will avoid an additional global var, it will make the JS more encapsulated and easier to test, and `wp_localize_script()` has the nasty side effect of converting an scalar values to strings (which wouldn't apply here though).

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,164 @@
...
+		}
+
+		// AMP panel triggers the input toggle for AMP preview.
+		panel.expanded.bind( function() {
+			api.state( 'ampEnabled' ).set( panel.expanded.get() );
+		} );
+
+		// Open AMP panel if mobile device selected.
+		api.previewedDevice.bind( function( device ) {
+			panel.expanded.set( 'mobile' === device );
```
westonruter: Is this right? The AC says:

> When a user selects to view the mobile version of a page that is AMP supported, the new toggle will automatically default to AMP. The user will be able to change it to Original if they wish to do so.

I think this means that `ampEnabled` should be set to `true` when switching to the mobile `previewedDevice`. I would think it could be annoying for the expanded panel to change when the user just changes the previewed device, as it would remove the controls context for where they were previously working.

kaitnyl:
@weston for your [comment here](https://github.com/Automattic/amp-wp/pull/819#pullrequestreview-81422199) regarding the `amp_customizer_is_enabled` filter, I've made changes for this within https://github.com/Automattic/amp-wp/pull/819/commits/0d1790bb0cb0752a08395fd7b1e687dce60d95b5.

I've moved the filter to control only the UI portion of the `AMP_Customizer_Design_Settings` class. We still need some things in that class to load for the AMP preview in the Customizer to function correctly, and moving the [filter here](https://github.com/Automattic/amp-wp/pull/819/commits/0d1790bb0cb0752a08395fd7b1e687dce60d95b5#diff-c001fcbf113f636c24f6a66210306211) will ensure the AMP panel is still added (although empty and hidden).

westonruter:
@kaitnyl I think that the toggle needs some sort of hover and focus styling? Refer to Gutenberg.

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,201 @@
...
+			}
+		} );
+
+		// Main controls for toggling AMP preview.
+		$( '#customize-footer-actions' ).on( 'click', '.amp-toggle', function() {
+			var $input = $( 'input', $( this ) ),
+				$tooltip = $( '.tooltip', $( this ) ),
+				tooltipTimer = 5000;
+
+			if ( $input.prop( 'disabled' ) ) {
```
ThierryA: It would be good to prevent the tooltip from disappearing in the users mouse is over the tooltip.

assets/js/amp-customize-controls.js
```diff
@@ -0,0 +1,201 @@
...
+			api.state( 'ampAvailable' ).set( data.available );
+		} );
+		function setInitialAmpEnabledState( data ) {
+			api.state( 'ampEnabled' ).set( data.enabled );
+			api.previewer.unbind( 'amp-status', setInitialAmpEnabledState );
+		}
+		api.previewer.bind( 'amp-status', setInitialAmpEnabledState );
+
+		// Persist the presence or lack of the amp=1 param when navigating in the preview,
+		// even if current page is not yet supported.
```
ThierryA: Small detail, this doesn't follow WP JS code standards (see [here](https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#comments)).

includes/settings/class-amp-customizer-design-settings.php
```diff
@@ -62,37 +127,97 @@ public static function register_customizer_ui( $wp_customize ) {
...
+			'container_inclusive' => true,
+		) );
+	}
+
+	/**
+	 * Render header bar template.
+	 */
+	public static function render_header_bar() {
+		if ( is_singular() ) {
+			amp_load_classes();
```
ThierryA: Make a note to remove all `amp_load_classes()` in #828.

westonruter:
@ThierryA 

> I picked up that AC point 3 (below) does not work for me while testing this locally:
>> In the event a user navigates to a page that is not AMP enabled a notification will display within the AMP panel to inform the user that they should navigate to a URL that is AMP compatible.

This is currently indicated by the toggle being grayed out, and then you can click on the toggle to show the notification. Maybe the notification should be displayed in addition to the toggle being grayed?

includes/settings/class-amp-customizer-design-settings.php
westonruter: Done

assets/js/amp-customize-controls.js
westonruter: Or when keyboard focus is on the link in the tooltip.

ThierryA:
Great improvements @westonruter ❤️

PR #825, opened 2017-12-07
Add page support

ThierryA:
This PR adds support for pages, [refer to the ACs](https://github.com/Automattic/amp-wp/issues/176#issuecomment-346622268) for more info...

Fixes #176.
Closes #188.
Closes #816.

_Inspired by #188 & #816_
_Kudos to @technosailor, @mjangda and @mpszone for their contribution._

PR #826, opened 2017-12-07
Move AMP initialization logic into after_setup_theme, conditioned on amp_is_enabled filter

westonruter:
Move AMP initialization logic to `after_setup_theme` to only run if `amp_is_enabled` filter applies as `true`.

Amends #813.
See #709.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -99,7 +99,6 @@ function amp_init() {
 
 	load_plugin_textdomain( 'amp', false, plugin_basename( AMP__DIR__ ) . '/languages' );
 
-	amp_after_setup_theme();
 	add_rewrite_endpoint( AMP_QUERY_VAR, EP_PERMALINK );
 
 	add_filter( 'request', 'amp_force_query_var_value' );
@@ -199,6 +198,13 @@ function amp_render_post( $post ) {
 
 	amp_load_classes();
 
+	/**
+	 * Fires before rendering a post in AMP.
+	 *
+	 * @since 0.2
+	 *
+	 * @param int $post_id Post ID.
+	 */
 	do_action( 'pre_amp_render_post', $post_id );
 
 	amp_add_post_template_actions();
--- a/amp.php
+++ b/amp.php
@@ -30,6 +30,7 @@ require_once AMP__DIR__ . '/includes/actions/class-amp-paired-post-actions.php';
 
 register_activation_hook( __FILE__, 'amp_activate' );
 function amp_activate() {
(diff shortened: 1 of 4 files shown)


westonruter:
@ThierryA In 856bb27 I fixed a defect across the AMP plugin whereby if you added a `add_filter( 'amp_is_enabled', '__return_false' )` to your theme, only _parts_ of the plugin would be disabled.

westonruter:
I cherry-picked 36881a4 into #825, so this can be rebased once that PR is merged.

amp.php
```diff
@@ -50,19 +51,55 @@ function amp_deactivate() {
...
+	 * Triggers on init when AMP plugin is active.
+	 *
+	 * @since 0.3
+	 */
 	do_action( 'amp_init' );
 
 	load_plugin_textdomain( 'amp', false, plugin_basename( AMP__DIR__ ) . '/languages' );
 
-	amp_define_query_var();
+	amp_after_setup_theme();
```
ThierryA: Any specific reason to call `amp_after_setup_theme()` here since the `amp_init()` is invoked in a `init` action which is declared in the `amp_after_setup_theme()`?

amp.php
```diff
@@ -159,10 +173,12 @@ function amp_prepare_render() {
  * @since 0.1
  */
 function amp_render() {
-
 	// Note that queried object is used instead of the ID so that the_preview for the queried post can apply.
-	amp_render_post( get_queried_object() );
-	exit;
+	$post = get_queried_object();
+	if ( $post instanceof WP_Post ) {
```
ThierryA: Should it be exciting even if the page can't be rendered?

westonruter:
@ThierryA yes, I'll rebase 36881a4 out of the PR.

amp.php
westonruter: Yes, it is very exciting. 😉 

In practice this function only ever gets called if the queried object is already guaranteed to be a post, so this is more for the sake of static analysis. But yeah, it could be better to not exit here.

amp.php
westonruter: Good catch. Yeah, this should have already been called so it should be removed entirely.

amp.php
ThierryA: Haha, very exciting indeed 😂

westonruter: I think we can just leave this as-is right now. We'll need to revisit how this works when we start rendering non-posts in AMP anyway.

Issue #827, opened 2017-12-08
Add support for using regular WordPress theme templates in AMP, in paired or canonical modes


mikeschinkel:
We have a project where we need to build the entire site in AMP, and we are finding several of the architecture decisions for this plugin to make that more difficult than it needs to be. 

It seems everything assumes routing through `template_redirect` and then to be served inside `AMP_Post_Template->verify_and_include()` and making the `$this` variable visible within the themes _(and hence other AMP plugins are using it)_.

When I discovered this is surprised me because I would have assumed it would follow WordPress architecture, hook `'template_include'` and provide template tags and other template-related functions _(or maybe an `WP_AMP` singleton to encapsulate all the AMP-related functionality.)_

Right now it seems that anything other that supporting other AMP-equivalent templates other than  `single.php` would be a huge amount of work and I'm not even sure where someone would even start to add that support with this code base.

Do you think the ship has sailed on the current architecture or that we could revisit this and move in the direction of supporting a more WordPress-compatible approach?  I need to deliver something for a client and I am trying to decide if I should contribute to this project, fork it and modify it, or start from scratch.

westonruter:
One of my goals for AMP is to make the template system more WordPressy. Actually, the ultimate goal is to make WordPress themes themselves to be AMP-compatible, that is, Canonical AMP (#668). This will require that we make sure that WordPress's own templating system can be made to output only that which is AMP compatible, or to output-buffer the output and sanitize the entire rendered HTML. More discovery needs to be done on the best way to implement the AMP canonical themes.

Nevertheless, in the mean time for the exiting paired mode and the current AMP theming system, I do agree we should switch to `template_include` filter instead if possible. We are adding support for pages in #825 which adds a new `page.php`. It seems we could add support for `template_include` here by applying the filter in `\AMP_Post_Template::load()` and reworking how `load_parts` gets called.

If you can identify a backwards-compatible PR then I'd be happy to review.

mikeschinkel:
@westonruter Interesting.  Thanks for the super-fast reply.  Let me see what I can come up with.

mikeschinkel:
@westonruter How stable is the current `develop` branch?  Should I work off that or `master` or `0.5.1`?

westonruter:
@mikeschinkel use `develop` as the base branch. It is stable and now the default branch.

mikeschinkel:
[...]

PR #828, opened 2017-12-08
Added autoloader to reduce complexity; fix phpcs issues

mikeschinkel:
@westonruter As I started working on a refactoring as per [here](https://github.com/Automattic/amp-wp/issues/827) I found myself struggling with the complexity of all the classes and making sure they were all loaded in the correct order.  So I went out on a limb and implemented a simple, fast classmap autoloader and removed all the now-unnecessary `require_once()` calls throughout the plugin. I am really hoping you will see this as a plus and accept it, and then I can move forward with the next step of refactor as in the issue above.

While I was at it PhpStorm pointed out several errors and oversights and I created an individual commit for each of those, but the first commit has everything needed for and related to the autoloader.


.gitmodules
```diff
@@ -1,4 +0,0 @@
-[submodule "dev-lib"]
-	path = dev-lib
-	url = https://github.com/xwp/wp-dev-lib.git
-	branch = master
```
westonruter: This file needs to be restored.

autoloader.php
```diff
@@ -0,0 +1,153 @@
...
+		$filepath = $this->_classmap[ $class_name ];
+		require( '!' !== $filepath[ 0 ] ? __DIR__ . "/includes/{$filepath}.php" : substr( $filepath, 1 ) );
+	}
+
+	/**
+	 * Trigger creation of AMP_Autoloader singleton instance and return it.
+	 *
+	 * @return AMP_Autoloader
+	 */
+	public static function instance() {
```
westonruter: There aren't any singletons in this plugin, so I think static methods should be used instead for consistency.

autoloader.php
```diff
@@ -0,0 +1,153 @@
...
+
+	/**
+	 * Registers to autoloader to PHP.
+	 *
+	 * Called at the end of this file; calling a second time has no effect.
+	 */
+	public static function register() {
+		static $registered = false;
+		if ( ! $registered ) {
+			spl_autoload_register( [ static::instance(), '_autoload' ] );
```
westonruter: Syntax error in PHP 5.2

autoloader.php
```diff
@@ -0,0 +1,153 @@
...
+	}
+}
+
+/**
+ * Call method to register this autoloader with PHP.
+ */
+AMP_Autoloader::register();
+
+
+
```
westonruter: Extra linebreaks at end of file.

includes/admin/functions.php
```diff
@@ -106,6 +106,11 @@ function amp_add_options_menu() {
 }
 add_action( 'wp_loaded', 'amp_add_options_menu' );
 
+/**
```
westonruter: Missing function description, param description, and return description. We've also been adding `@since` tags.

includes/amp-helper-functions.php
```diff
@@ -1,5 +1,10 @@
 <?php
 
+/**
```
westonruter: Missing function description, param description, and return description. We've also been adding `@since` tags.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -73,6 +71,9 @@ public function sanitize() {
 		}
 	}
 
+	/**
```
westonruter: Missing function description, param description. We've also been adding `@since` tags.

.gitmodules
mikeschinkel: I deleted it because it was blocking my ability to commit and I thought my pull request would not have included the deletion. I am not exactly sure what I was doing wrong or how to fix it _(I gave up on submodules long ago as I always had issues with them.)_ 

Is there any chance that you can accept the pull request after I fix everything else, and then restore manually?

autoloader.php
mikeschinkel: Funny, I debated using static but assumed you'd prefer instances. :-)

autoloader.php
mikeschinkel: 5.2?  :-o
Fixed.

autoloader.php
mikeschinkel: Fixed.

autoloader.php
westonruter: Yeah. Currently the plugin supports PHP 5.2: https://github.com/Automattic/amp-wp/blob/eeb502b28807c1be80e59937fab6112e406b204b/.travis.yml#L45-L47

.gitmodules
westonruter: The lack of the submodule is causing Travis CI to fail: https://travis-ci.org/Automattic/amp-wp/jobs/313361694#L186

So it needs to be restored before we merge the PR so that the automated tests and checks can be performed.

What is the error you're having?

autoloader.php
westonruter: Normally I'd prefer non-singleton instances, but given most classes are just being used as namespaces with static methods, using static here again seems to be most in keeping with everything else.

.gitmodules
mikeschinkel: Let me see if I can figure it out.  

I was using SourceTree to allow me to make partial commits of files and it said I needed the submodule but was locked in the dialog and would not let me exit.  When I deleted the file SourceTree recognized the change and allowed my commit to continue.

I will try reverting this commit and see if I can get things to work from the command line.  If I have an issue I will let you know otherwise I will resolve it and we can just move forward.

includes/amp-helper-functions.php
mikeschinkel: Your requests to document this caused me to go down the rabbit hole for the past _(far too many)_ hours.  Every time I would document something it would cause PhpStorm to flag something else as being _"off"_, so I ended up documented all of the sanitizers.  

Note I found a few places that would throw errors, e.g. `DOMNode $nodes` using methods that are only on `DOMElement` so I added a variable of logic to ensure errors are not thrown during the normal course of usage when `WP_DEBUG` is set to `true`. One such example is I added the following in several places:

        return;
    }

I hope that was okay...

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
mikeschinkel: Documented.

autoloader.php
mikeschinkel: Updated to use static methods.

autoloader.php
mikeschinkel: Noted.

includes/admin/functions.php
mikeschinkel: I do not understand that function's parameter well enough to fully document so I backed out the partial PHPDoc I had added.

includes/amp-helper-functions.php
westonruter: The node type checks are probably a good idea, as I can see that clearly some of the situations would result in text nodes being used as elements. It looks like some of the instances you added would never result in a non-element being iterated (e.g. `$this->dom->getElementsByTagName( self::$tag );`) but it doesn't hurt.

westonruter:
We kinda opened a can of worms here because now there are many PHPCS violations being reported, because of the various files being touched, and wp-dev-lib will only report PHPCS errors surrounding actual patches. The PHPCS issues need to be fixed at some point anyway, but I don't know if now is the time.

I am noticing, however, one PHPUnit failure that must be fixed:

```
There was 1 failure:

1) AMP_Iframe_Converter_Test::test__args__placeholder
--- Expected
+++ Actual
@@ @@
-'<amp-iframe src="https://example.com/video/132886713" width="500" height="281" sandbox="allow-scripts allow-same-origin" sizes="(min-width: 500px) 500px, 100vw" class="amp-wp-enforced-sizes"><div placeholder="" class="amp-wp-iframe-placeholder"></div></amp-iframe>'
+'<amp-iframe src="https://example.com/video/132886713" width="500" height="281" sandbox="allow-scripts allow-same-origin" sizes="(min-width: 500px) 500px, 100vw" class="amp-wp-enforced-sizes"><div placeholder="" class="amp-wp-enforced-sizes" src="https://example.com/video/132886713" width="500" height="281" sandbox="allow-scripts allow-same-origin" sizes="(min-width: 500px) 500px, 100vw"></div></amp-iframe>'

.../amp/tests/test-amp-iframe-sanitizer.php:167
```

mikeschinkel:
@westonruter I have been trying to recreate a local PHPUnit setup to run the tests locally and while I've gotten past numerous issues I've found a roadblock on this error message:

> _"[_Unable to locate wordpress-tests-lib_](https://github.com/xwp/wp-dev-lib/blob/master/phpunit-plugin-bootstrap.php#L31)"_

It appears to be calling a `includes/functions.php` if it can find the appropriate directory but I don't know what `functions.php` is suppose to contain.

Google is failing me to find the right solution to this. I'm sure I could power through it given enough time but I am hoping you can make me more productive and help me know how to solve this one.

Help?

westonruter:
@mikeschinkel Yeah, there's a couple options here.

For one, you can add the following to your `.bash_profile`:

```sh
export WP_TESTS_DIR=~/path/to/wordpress-develop/tests/phpunit/
```

That assumes you have a fully up-and-running [wordpress-develop](https://github.com/WordPress/wordpress-develop/) core unit tests on the machine. 

Otherwise, you can just run the unit tests inside of VVV. You can see that it has this out of the box: https://github.com/Varying-Vagrant-Vagrants/VVV/blob/5b4ee4b/config/bash_profile#L29-L33

mikeschinkel:
@westonruter Thanks.  That got me past that roadblock.  However the next roadblock is trying to use either PHPUnit `4.8` or `5.7`.  With PHPUnit `4.8.9` and PHP `5.3` I get failures at:

> `Parse error: syntax error, unexpected '[' in /.../plugins/amp-wp/vendor/doctrine/instantiator/src/Doctrine/Instantiator/Instantiator.php on line 45` 

Clearly I have a later version of `doctrine/instantiator` but since my `composer.json` does not reference it directly I have no clue how to correct this.

 With PHPUnit `5.7.9` and PHP `5.6` I get failures at:

> `PHP Warning:  Unsupported declare 'strict_types' in /.../plugins/amp-wp/vendor/sebastian/global-state/src/Snapshot.php on line 11` 

When I switch to PHPUnit `6.5.4` and PHP `7.0.17` I can get tests to run, but there are more than one tests failed.  Should I work through all the failures?

mikeschinkel:
@westonruter Good news!  I was about to get all `501` tests to pass, at least with PHPUnit `6.5.4` and PHP `7.0`.  I'll have a commit very soon.

mikeschinkel:
@westonruter In order to get testing working I needed to add a `composer.json`.  Any reason for me not to include that in the PR?

mikeschinkel:
@westonruter   I see that PHP 7.1 is failing so I am working to debug and correct it. Just FYI.

westonruter:
> In order to get testing working I needed to add a composer.json. Any reason for me not to include that in the PR?

By did you need to add Composer to get it to work? In any case, it makes sense to include. I think it could look like https://github.com/WordPress/better-code-editing/blob/master/composer.json

mikeschinkel:
@westonruter Yes, because I was trying to use PhpStorm to run the tests _(as it provides such a nice interface)_ and for some reason PhpStorm would not recognize the `.phar`; it would only recognize when I pointed to the Composer-generated `autoload.php`.

I added one without `"dist"` &mdash; for obvious reasons &mdash; and I omitted a version number because I have found adding one makes it difficult for the person wanting to use a specific version when the Git tag does not exactly match the version in the `composer.json` file; kind of a _"[_man with two watches_](https://en.wikipedia.org/wiki/Segal%27s_law)"_ problem.

mikeschinkel:
@westonruter Finally!  What a rabbit hole that was.  :-)

Let me know if you have an issues otherwise I will start working on how to implement a full theme in AMP and run the ideas by you either in issue #827 or via a new pull request.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1,22 +1,56 @@
...
 
+	/**
+	 * Get list of CSS styles in HTML content of DOMDocument ($this->dom)
+	 *
+	 * @since 0.4
+	 *
+	 * @return string[]
+	 */
 	public function get_styles() {
+		if ( ! $this->did_convert_elements ) {
```
westonruter: Why this if `$this->styles` is already an `array()`?

mikeschinkel: Consistency.  But to your point I guess it is not needed.

mikeschinkel:
Woohoo!  :-D

PR #831, opened 2017-12-12
Show panel and section notifications when AMP is unavailble

westonruter:
In WordPress 4.9 and above, panels and sections support notifications (see https://core.trac.wordpress.org/ticket/38794). We can use these notifications to more prominently communicate that a user needs to navigate to an AMP-compatible URL to preview changes, similar to what is done in core with the header video control in the header media section.

This PR adds notifications to the AMP panel and any sections contained inside of it: 

image

image

Amends #796


csossi:
warnings appearing fine

Issue #849, opened 2018-01-09
Allow themes to provide alternate templates in “paired mode”


ThierryA:
### Acceptance Criteria

AC1: Using “paired mode,” themes can specify an alternative path to use for AMP templates (see @westonruter’s [comment proposed](https://github.com/Automattic/amp-wp/issues/827#issuecomment-351827719)).
AC2: When themes do this, the templates in that directory will replace the entire “paired mode” template system in the plugin. The AMP toggle developed in 0.6 will continue to be available in this mode.

#### Tasks

- [x] Add theme support property to indicate path to paired-mode templates, for the theme root, where the template hierarchy will be evaluated. This may be called `theme_root` or `template_directory` or `template_path` or something else that makes the most sense.
- [x] And add callback function which is used to determine whether a given request should be routed to the paired mode. Have default logic that tries to identify the template for a given request and if it is is present, then make AMP available. Otherwise, don’t make it available. For example, if requesting a single page, but there is no page.php or index.php in the AMP template root, then AMP would not be available by default.
- [x] Update `amp_prepare_render()` to _prevent_ calling `add_action( 'template_redirect', 'amp_render' );` when `current_theme_supports( 'amp' )`, as we need to _not_ use `amp_render()` since it uses the old paired-mode theme system; instead, this function should do whatever is required (e.g. add `template_include` filter), per the next task (implement template hierarchy). 
- [x] Implement the same template hierarchy resolution logic, ideally re-use what is in core. Figure out how to make use of the custom theme root for AMP templates. 
- [x] Continue outputting the AMP rel link in head for non-AMP templates and other things that happen in current paired mode.  And also include relevant links in head for AMP templates. Toggle what is output in wp_head based on whether it is an AMP response or not.

PR #852, opened 2018-01-10
Issue 848: Allow theme support of canonical AMP

### LINKED ISSUE #848: Allow themes to register support for AMP canonical mode
### Acceptance Criteria

AC1: Allow themes to register support for AMP canonical with `add_theme_support( 'amp' )` (As @westonruter’s [comment proposed](https://github.com/Automattic/amp-wp/issues/827#issuecomment-351827719)).
AC2: If support is added, this would remove paired mode. The site would only serve AMP, and there wouldn’t be `/amp` or `?amp` URLs.

#### Tasks

- [x] Check if AMP theme support exists, and if so, do not activate  the current AMP templating system and the current paired mode.
- [x] When AMP theme support is present, prevent AMP rel link in head and other frontend actions that assume paired mode.

kienstra:
**Request For Review**

Hi @ThierryA or @westonruter,
Could you please review this pull request for Issue #848?

If the theme runs `add_theme_support( 'amp' )` without any other arguments, the plugin will be in 'canonical mode.' The theme will still use its own template(s).

But if any other argumnents are passed to `add_theme_support( 'amp' )`, it will retain 'paired mode.' That case is in #849.

When in 'canonical mode':
* There's no Customizer 'AMP' panel
* There's no AMP 'rel' link
* The plugin doesn't use its template system
* The `post.php` page's 'Publish' meta box doesn't have the AMP preview link, nor the "AMP: Enabled" section:
removed-in-canonical

Also, should we keep the 'AMP Settings' page?
post-type-support

When you add or remove support for a custom post type, `amp_is_canonical()` [short-circuits this](https://github.com/Automattic/amp-wp/compare/add/848-theme-support-canonical?expand=1#diff-4805c837d3390517773926fb26726bcdR140) if it is `true`, and prevents the plugin from using its own templates and running frontend actions. 

Fixes #848.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -5,7 +5,7 @@
  * Plugin URI: https://github.com/automattic/amp-wp
  * Author: Automattic
  * Author URI: https://automattic.com
- * Version: 0.6-alpha
+ * Version: 0.7-alpha
  * Text Domain: amp
  * Domain Path: /languages/
  * License: GPLv2 or later
@@ -15,7 +15,7 @@
 
 define( 'AMP__FILE__', __FILE__ );
 define( 'AMP__DIR__', dirname( __FILE__ ) );
-define( 'AMP__VERSION', '0.6-alpha' );
+define( 'AMP__VERSION', '0.7-alpha' );
 
 require_once AMP__DIR__ . '/includes/class-amp-autoloader.php';
 AMP_Autoloader::register();
--- a/amp.php
+++ b/amp.php
@@ -127,7 +127,7 @@ function amp_force_query_var_value( $query_vars ) {
  * @return void
  */
 function amp_maybe_add_actions() {
-	if ( ! is_singular() || is_feed() ) {
+	if ( amp_is_canonical() || ! is_singular() || is_feed() ) {
 		return;
 	}
 
@@ -137,7 +137,7 @@ function amp_maybe_add_actions() {
 	global $wp_query;
 	$post = $wp_query->post;
 
-	$supports = post_supports_amp( $post ) && ! amp_is_canonical();
+	$supports = post_supports_amp( $post );
 
 	if ( ! $supports ) {
 		if ( $is_amp_endpoint ) {
--- a/amp.php
+++ b/amp.php
@@ -117,6 +117,15 @@ function amp_force_query_var_value( $query_vars ) {
 	return $query_vars;
 }
 
+/**
+ * Conditionally add AMP actions or render the 'paired mode' template(s).
+ *
+ * If the request is for an AMP page and this is in 'canonical mode,'
+ * redirect to the non-AMP page.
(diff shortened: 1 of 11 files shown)


westonruter:
> Also, should we keep the 'AMP Settings' page?

Great point. No, this should not be kept when in canonical mode.

amp.php
```diff
@@ -145,6 +154,31 @@ function amp_maybe_add_actions() {
...
+ * @return boolean Whether this is in AMP 'canonical mode'.
+ */
+function amp_is_canonical() {
+	$support = get_theme_support( 'amp' );
+	if ( true === $support ) {
+		return true;
+	}
+	if ( is_array( $support ) ) {
+		$args = array_shift( $support );
+		if ( empty( $args['template_path'] ) ) {
```
westonruter: The use of `template_path` will probably need to be iterated upon. For example, `theme_root` per `\WP_Theme::get_theme_root()` or `template_dir` per `\WP_Theme::get_template_directory()`. This will need some more discovery in #849 as to what makes the most sense. To me it seems the `theme_root` and `template_directory` are the same. /cc @kaitnyl

amp.php
```diff
@@ -128,7 +137,7 @@ function amp_maybe_add_actions() {
 	global $wp_query;
 	$post = $wp_query->post;
 
-	$supports = post_supports_amp( $post );
+	$supports = post_supports_amp( $post ) && ! amp_is_canonical();
```
westonruter: Might as well move the `! amp_is_canonical()` check up to the top of the function:

```php
if ( amp_is_canonical() || ! is_singular() || is_feed() ) {
```

kienstra:
**Thanks For Reviewing**
*Will Disable 'AMP Settings' Section*

Hi @westonruter,
Thanks a lot for reviewing this. I'll remove the "AMP Settings" post type section.

westonruter:
> Thanks a lot for reviewing this. I'll remove the "AMP Settings" post type section.

@kienstra This could be a bit tricky because without the post types settings section, the AMP > General admin page would have nothing on it. Maybe for now the best thing would be to just prevent rendering that settings section and instead show a message that canonical AMP is enabled, and so all post types will be shown regardless. Or rather, modify the section to display every post type with the checkbox `checked` and `disabled` along with the notice and without the submit button.

kienstra:
**Applied Suggestion For 'AMP Settings'**

Hi @westonruter,
Thanks for your suggestion for the 'AMP Settings' page, which the [commit](https://github.com/Automattic/amp-wp/pull/852/commits/4a1f621051fba96a4cd2ab54e07e655264917cec) above applies.

Here's the 'AMP Settings' page when the theme registers support for AMP 'canonical mode':
theme-registers-support-canonical-amp

PR #853, opened 2018-01-10
Remove pound symbol from header id

eshannon3:
Removed the # symbol from <header> id. It was causing 404 errors and
prevented the "Back to top" link in the footer from working.


westonruter:
@ThierryA we might as well merge this into `0.6` as well.

PR #856, opened 2018-01-11
#849 Add paired mode custom templates

kaitnyl:
This PR will allow users to specify a custom AMP template folder from within their own theme for #849.

===

**Completed**

- Using "paired mode," themes can specify an alternative path to use for AMP templates (example below).
- Templates in the custom directory will replace the entire "paired mode" template system in our plugin. The template hierarchy loaded is hooked into WP's `template_include` in order to follow typical template schemas.
- Callback added to further control custom template usage when referencing custom directory (example below). 
- All "default" template output still available when not using a custom template folder. 
- Toggle within Customizer still available when using custom templates.

Testing example inside a theme:
```
	'template_dir'       => 'amp-templates',
	'available_callback' => function() {
		return is_single(); // Change this up to enable/disable custom templates per $post
	},
) );
```
_Updated as of b10c0fe._

===

**In question**

- Update amp_prepare_render() to prevent calling add_action( 'template_redirect', 'amp_render' )

This part of the acceptance criteria may not be needed (but could be useful for clarity)? Before we get to `amp_prepare_render()`, we're initially inside `amp_maybe_add_actions()` and that will prevent calling `amp_prepare_render()` depending on [this line](https://github.com/Automattic/amp-wp/pull/856/files#diff-4805c837d3390517773926fb26726bcdR130).


includes/templates/class-amp-post-template.php
```diff
@@ -80,14 +80,50 @@ class AMP_Post_Template {
...
+	 */
+	public $custom_template_dir;
+
+	/**
+	 * If template system is currently active.
+	 *
+	 * @since 0.7
+	 * @var bool
+	 */
+	public $active;
```
kaitnyl: I feel the term "active" might be too ambiguous here. This is what the callback property within the theme support addition controls, and is used when we do have a custom template included but want to control if it should be AMP or default WP templates (since the custom will try to follow WP template hierarchy and we may not always want it to).

kaitnyl:
@ThierryA and @westonruter, would either of you mind reviewing this PR for #849? 😃

includes/templates/class-amp-post-template.php
```diff
@@ -80,14 +80,50 @@ class AMP_Post_Template {
...
+		if ( is_array( $support ) ) {
+			$args = array_shift( $support );
+
+			if ( ! empty( $args['template_path'] ) ) {
+				$this->custom_template_dir = $args['template_path'];
+			}
 
-		$this->template_dir = apply_filters( 'amp_post_template_dir', AMP__DIR__ . '/templates' );
+			if ( isset( $args['active_callback'] ) ) {
+				$this->active = $args['active_callback']();
```
westonruter: I think this may need to be `call_user_func( $args['active_callback'] )` for PHP 5.2.

westonruter:
@kaitnyl It looks like I pushed up some code after you pushed up yours. Take a look at this line of my #857 PR here:

https://github.com/Automattic/amp-wp/blob/20838af95a8445f01a3a578ada9d53543aed3761/amp.php#L129-L146

I think that instead of adding code to `AMP_Post_Template` we can instead make a clean break and instead add your code to a new `AMP_Paired_Mode_Actions`. This would make sense I think because in our new paired mode system, it's not limited to _posts_ at all, which the existing templating system is limited to. This would then answer your question about `amp_prepare_render()`, because with the PR #857 this never gets called at all when a theme supports `amp`.

includes/templates/class-amp-post-template.php
westonruter: Nevermind! I was way wrong. This works even back in PHP 4: https://3v4l.org/TcVa0

westonruter:
@kaitnyl The size of this PR increased due to the merge of both #857 and #858.

includes/class-amp-theme-support.php
```diff
@@ -14,6 +14,84 @@ class AMP_Theme_Support {
...
+	public static function is_paired_available() {
+		$support = get_theme_support( 'amp' );
+		if ( empty( $support ) || amp_is_canonical() ) {
+			return false;
+		}
+
+		$args = array_shift( $support );
+
+		// @todo We might want to rename active_callback to available_callback..
+		if ( isset( $args['active_callback'] ) && is_callable( $args['active_callback'] ) ) {
```
westonruter: Should we rename `active_callback` to `available_callback`? It would seem a little strange for an _active_ callback to return `true` if on the frontend in paired mode, since `is_amp_endpoint()` would be false.

includes/class-amp-theme-support.php
```diff
@@ -54,6 +132,44 @@ public static function register_hooks() {
...
+	 * @param array $templates Template hierarchy.
+	 * @returns array Templates.
+	 */
+	public static function filter_paired_template_hierarchy( $templates ) {
+		$support = get_theme_support( 'amp' );
+		$args    = array_shift( $support );
+		if ( isset( $args['template_path'] ) ) {
+			$amp_templates = array();
+			foreach ( $templates as $template ) {
+				$amp_templates[] = $args['template_path'] . '/' . $template;
```
westonruter: Since `template_path` is a _relative_ path from the theme root to locate the AMP templates directory, is this still the best name?

includes/class-amp-theme-support.php
kaitnyl: I agree the name isn't the most accurate. I've updated it to `available_callback` in b10c0fe, but `display_callback` might even be a more clear. Thoughts?

amp.php
```diff
@@ -131,9 +131,9 @@ function amp_maybe_add_actions() {
 	// Add hooks for when a themes that support AMP.
 	if ( current_theme_supports( 'amp' ) ) {
 		if ( amp_is_canonical() || is_amp_endpoint() ) {
-			AMP_Theme_Support::register_hooks();
+			AMP_Theme_Support::init();
 		} else {
-			AMP_Frontend_Actions::register_hooks();
+			amp_add_frontend_actions();
```
ThierryA: Why calling this here again since it is already a `wp_head` callback and it has a conditional statement only to load in paired mode?

includes/class-amp-theme-support.php
kaitnyl: I've updated these to use `template_dir` which is what `AMP_Post_Template` is calling it, too.  Setting this will look like the following, which should be easier since it'll be done inside of a theme:
```
	'template_dir' => 'amp-templates',
```

includes/class-amp-theme-support.php
```diff
@@ -14,6 +14,83 @@ class AMP_Theme_Support {
...
+	public static function is_paired_available() {
+		$support = get_theme_support( 'amp' );
+		if ( empty( $support ) || amp_is_canonical() ) {
+			return false;
+		}
+
+		$args = array_shift( $support );
+
+		if ( isset( $args['available_callback'] ) && is_callable( $args['available_callback'] ) ) {
+			return $args['available_callback']();
```
ThierryA: Using `call_user_func()` would probably be more appropriate here.

includes/amp-frontend-actions.php
```diff
@@ -0,0 +1,33 @@
...
+	/**
+	 * Filters whether to show the amphtml link on the frontend.
+	 *
+	 * @since 0.2
+	 */
+	if ( false === apply_filters( 'amp_frontend_show_canonical', true ) ) {
+		return;
+	}
+
+	$amp_url = amp_get_permalink( get_queried_object_id() );
```
ThierryA: Not really a need to store this in a variable, the following would be readable too:

``` PHP
printf(
	'<link rel="amphtml" href="%s">',
	esc_url( amp_get_permalink( get_queried_object_id() ) )
);
```

includes/class-amp-theme-support.php
```diff
@@ -14,6 +14,83 @@ class AMP_Theme_Support {
...
+		'attachment',
+	);
+
+	/**
+	 * Initialize.
+	 */
+	public static function init() {
+		require_once AMP__DIR__ . '/includes/amp-post-template-actions.php';
+		if ( amp_is_canonical() ) {
+			$is_amp_endpoint = ( false !== get_query_var( AMP_QUERY_VAR, false ) ); // Because is_amp_endpoint() now returns true if amp_is_canonical().
```
ThierryA: There isn't really a need to store this in a variable, the condition could be added directly in the if statement below.

includes/class-amp-theme-support.php
westonruter: I suppose it would be more readable.

includes/class-amp-theme-support.php
```diff
@@ -14,6 +14,83 @@ class AMP_Theme_Support {
...
+		}
+		return true;
+	}
+
+	/**
+	 * Register hooks for paired mode.
+	 */
+	public static function register_paired_hooks() {
+		foreach ( self::$template_types as $template_type ) {
+			add_filter( "{$template_type}_template_hierarchy", array( __CLASS__, 'filter_paired_template_hierarchy' ) );
```
westonruter: It would be preferred if there was a `template_hierarchy` filter in core so that the list of types wouldn't have to be hard-coded. But alas, there is no such filter: https://github.com/WordPress/wordpress-develop/blob/fb5b6170564b403944a8fbf8a4e6eb877b26a092/src/wp-includes/template.php#L30-L44

includes/class-amp-theme-support.php
westonruter: I think I like `available` more. It sounds more like something that would return a boolean. There is a `display_callback` in core, but it is used in widgets to actually render something out. Otherwise, in core, there is `active_callback` and `video-active-callback`.

Core usage: https://gist.github.com/westonruter/49552860d1e57a20547bcebd090ef269

I'm torn between _available_ and _active_.

includes/class-amp-theme-support.php
westonruter: Right, there is the `amp_post_template_dir` filter:

```php
$this->template_dir = apply_filters( 'amp_post_template_dir', AMP__DIR__ . '/templates' );
```

One difference is that the `template_dir` is absolute here, whereas what we're doing in the theme support is relative. I doubt there will be confusion anyway.

amp.php
westonruter: @ThierryA I don't understand. This function is only being called here in the `amp_maybe_add_actions()` function. It gets called here at the top in the case where the theme supports `amp`. Otherwise, it gets called below when the theme does not.

Issue #864, opened 2018-01-16
List AMP support for all native WordPress widgets


kienstra:
**Acceptance criteria:**
AC1: List the AMP support for all native WordPress widgets, as produced by the script in #839. 

This from AC5 in #839:
`AC5: One story will be created as an outcome of the Discovery to address enhanced/added support for sidebar widgets.`

When AMP doesn't support a widget, either describe it in detail here, or open a sub-issue.

There doesn't seem to be [an issue](https://github.com/Automattic/amp-wp/issues?utf8=%E2%9C%93&q=is%3Aissue+widget+) for this yet.

kienstra:
**Widget AMP Support**
(Comment WIP)

Widget | Status | Note
--- | --- | ---
[Audio](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/widgets/class-wp-widget-media-audio.php) | 🔸 | Mainly functional, but has an AMP error
[Archives](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/widgets/class-wp-widget-archives.php) *without* 'dropdown' |  ✅ 
[Archives](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/widgets/class-wp-widget-archives.php) *with* 'dropdown' | 🔸 |  AMP error from the script on the `onchange` attribute
[Calendar](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/widgets/class-wp-widget-calendar.php) |  ✅ 


**AMP Errors**
[Audio Widget](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/widgets/class-wp-widget-media-audio.php)
* The tag 'audio' may only appear as a descendant of tag 'noscript'. Did you mean 'amp-audio'? (2 occurrences) 
Though the plugin's [sanitize function](https://github.com/Automattic/amp-wp/blob/b94b10bade418d7710138a8d7bb88ccc39ce5140/includes/templates/class-amp-content.php#L184) doesn't yet apply to widgets.

[Archives Widget](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/widgets/class-wp-widget-archives.php) 
* The attribute 'onchange' may not appear in tag 'select':
```<select id="archives-dropdown-106" name="archive-dropdown" onchange='document.location.href=this.options[this.selectedIndex].value;'>```

westonruter:
@kienstra these are all supported now via #870, right?

[...]

PR #870, opened 2018-01-17
Issue #864: Native WordPress widget support, including subclasses

kienstra:
**WIP Pull Request**

This is the work in progress to support all native WordPress widgets (#864). The build will fail now, as the filtering in widget() isn't present.

There isn't yet a way to instantiate the class `AMP_Widgets()`. And feel free to comment on the location of these files. I thought it'd be good to place them in a new `includes/widgets/` directory.

This unregisters the 'Archives' and 'Categories' widgets. And registers new widgets that subclass them.

@todo: 
* Filter the output of `widget()` in these subclasses.
* Add support for the remaining widgets, mainly with subclasses.
* Handle the widgets that depend on scripts, and dequeue the remaining scripts.


includes/widgets/class-amp-widgets.php
```diff
@@ -0,0 +1,63 @@
...
+	 *
+	 * Uses the action 'wp_footer' in order to prevent
+	 * 'wp_print_footer_scripts' from outputting the scripts.
+	 *
+	 * @return void.
+	 */
+	public function dequeue_scripts() {
+		wp_dequeue_script( 'wp-mediaelement' );
+		wp_dequeue_script( 'mediaelement-vimeo' );
+		wp_dequeue_style( 'wp-mediaelement' );
```
westonruter: I don't think this is necessary. The plugin is already unhooking `wp_print_footer_scripts`. It may be useful to keep track of when certain scripts are enqueued, as that can be a signal that certain functionality needs to be incorporated using AMP components.

includes/widgets/class-amp-widgets.php
```diff
@@ -0,0 +1,63 @@
...
+
+	/**
+	 * Add the actions.
+	 *
+	 * @return void.
+	 */
+	public function init() {
+		add_action( 'widgets_init', array( $this, 'register_widgets' ) );
+		add_action( 'show_recent_comments_widget_style', '__return_false' );
+		add_action( 'wp_footer', array( $this, 'dequeue_scripts' ) );
```
westonruter: Per below, this shouldn't be necessary.

includes/widgets/class-amp-widgets.php
```diff
@@ -0,0 +1,63 @@
...
+class AMP_Widgets {
+
+	/**
+	 * Add the actions.
+	 *
+	 * @return void.
+	 */
+	public function init() {
+		add_action( 'widgets_init', array( $this, 'register_widgets' ) );
+		add_action( 'show_recent_comments_widget_style', '__return_false' );
```
westonruter: Actually `remove_filter` here. Nevertheless, should this be added inside of a subclass of `WP_Widget_Recent_Comments`?

kienstra: Hi @westonruter,
Thanks for reviewing this. That's a good point to move this into a subclass of `WP_Widget_Recent_Comments`. How does [this subclass look](https://github.com/Automattic/amp-wp/pull/870/commits/be58d0243f1ef4e3533e853ff6bb436231104d4f#diff-c5a7c81e8eaa3508f2f9881ca24ccd43R20)? It might be better if I avoided adding a filter on the `__construct()` method, though.


>Actually remove_filter here

I initially used `add_action` instead of `add_filter`:
```
add_action( 'show_recent_comments_widget_style', '__return_false' );
```
But wouldn't this be enough to [prevent outputting the styling](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/widgets/class-wp-widget-recent-comments.php#L53):
```
add_filter( 'show_recent_comments_widget_style', '__return_false' );
```

includes/widgets/class-amp-widgets.php
kienstra: Thanks, @westonruter. This [commit](https://github.com/Automattic/amp-wp/pull/870/commits/be58d0243f1ef4e3533e853ff6bb436231104d4f#diff-4a5a57759183e45d3afe37b6f63e406cL57) deletes `dequeue_scripts()`.

includes/widgets/class-amp-widgets.php
kienstra: Thanks, this [commit](https://github.com/Automattic/amp-wp/pull/870/commits/be58d0243f1ef4e3533e853ff6bb436231104d4f#diff-4a5a57759183e45d3afe37b6f63e406cL21) deletes this `add_action()` call.

kienstra:
#875 might handle sanitizing the entire page, making these commits unnecessary.

includes/widgets/class-amp-widgets.php
westonruter: The subclass looks good.

includes/widgets/class-amp-widget-categories.php
```diff
@@ -0,0 +1,31 @@
...
+	 * @todo filter $output, to strip the <script> tag.
+	 * @see https://github.com/Automattic/amp-wp/issues/864
+	 * @param array $args Widget display data.
+	 * @param array $instance Data for widget.
+	 * @return void.
+	 */
+	public function widget( $args, $instance ) {
+		ob_start();
+		parent::widget( $args, $instance );
+		$output = ob_get_clean();
```
westonruter: @kienstra Here we need to handle the dropdown behavior. I think this should strip out the `script` tag here via regex:

```php
$output = preg_replace( '#<script.+?</script>#', '', $output );
```

And then what it needs to do is inject two additional things:

1. Add am `id` attribute to the `<form>` with a random value, or an auto-incremented one: `widget-categories-dropdown-%d`, e.g. `widget-categories-dropdown-123`
2. Inject an `on` attribute onto the `<select>` element like `on="change:widget-categories-dropdown-123.submit"`.

Reference: https://www.ampproject.org/docs/reference/components/amp-form#input-events

👉 N.B. WordPress 4.9 first introduced the `<form>` element here: https://github.com/WordPress/wordpress-develop/commit/b7c70cad148a5a4c57b8403ed4ba5008f821a67a

So if no `<form>` is present, then it should wrap the `<select>` with one.

On second thought, instead of output-buffering the output and modifying it, just copy the existing logic from `\WP_Widget_Categories::widget()` in WP 4.9 and modify it here to be valid AMP.

kienstra: **Applied Suggestion**
*Thanks For The Explanation*

Hi @westonruter,
Thanks for your great explanation of how to make the dropdown work with AMP. How does [this commit](https://github.com/Automattic/amp-wp/pull/870/commits/fb9b451c2286f52b3a35766c52525fc5ff167868) look? There are still Travis errors that I need to fix.

It mainly copies  `\WP_Widget_Categories::widget()` from 4.9, and makes the changes you suggested. 

It's worked locally, with valid AMP.

ThierryA:
@kienstra this PR may be helpful to look at as it uses the WP core widgets in the static templates with the necessary HTML modifications be AMP valid.

includes/widgets/class-amp-widgets.php
```diff
@@ -0,0 +1,53 @@
...
+	 * @return array $widgets An associative array, with the previous WP widget mapped to the new AMP widget.
+	 */
+	public function get_widgets() {
+		return array(
+			'WP_Widget_Archives'        => 'AMP_Widget_Archives',
+			'WP_Widget_Categories'      => 'AMP_Widget_Categories',
+			'WP_Widget_Media_Audio'     => 'AMP_Widget_Media_Audio',
+			'WP_Widget_Media_Gallery'   => 'AMP_Widget_Media_Gallery',
+			'WP_Widget_Media_Image'     => 'AMP_Widget_Media_Image',
+			'WP_Widget_Media_Video'     => 'AMP_Widget_Media_Video',
```
westonruter: I can see Travis is failing due to these media widgets not existing in 4.7. So I suggest before returning this array to loop over the keys and remove any for which `! class_exists( $key )`.

includes/widgets/class-amp-widget-categories.php
```diff
@@ -0,0 +1,102 @@
...
+	 * This enables adding an 'on' attribute, with the id of the form.
+	 * So changing the dropdown value will redirect to the category page, with valid AMP.
+	 *
+	 * @param array $args Widget display data.
+	 * @param array $instance Data for widget.
+	 * @return void.
+	 */
+	public function widget( $args, $instance ) {
+		static $first_dropdown = true;
+		// @codingStandardsIgnoreLine
```
westonruter: What is this for? Is it for the text domain? If so, I suggest editing the `phpcs.xml` to include `default`, such as:

```xml
<property name="text_domain" value="amp,default" />
```

And then add `'default'` as the text domain for `__( 'Categories' )` and others.

includes/widgets/class-amp-widget-categories.php
```diff
@@ -0,0 +1,102 @@
...
+			<?php
+			$cat_args['title_li'] = '';
+			/**
+			 * Filters the arguments for the Categories widget.
+			 *
+			 * @since 2.8.0
+			 * @since 4.9.0 Added the `$instance` parameter.
+			 *
+			 * @param array $cat_args An array of Categories widget options.
+			 * @param array $instance Array of settings for the current widget.
```
westonruter: Instead of copying the phpdoc, you can `/** This filter is documented in ... */`

amp.php
```diff
@@ -71,6 +71,7 @@ function amp_after_setup_theme() {
 	}
 
 	add_action( 'init', 'amp_init' );
+	add_action( 'init', 'amp_add_widget_support' );
```
westonruter: Should this only be done if `amp_is_canonical()`?

includes/widgets/class-amp-widget-categories.php
```diff
@@ -0,0 +1,102 @@
...
+class AMP_Widget_Categories extends WP_Widget_Categories {
+
+	/**
+	 * Adds an 'on' attribute to the category dropdown's <select>.
+	 *
+	 * @param string $dropdown The markup of the category dropdown.
+	 * @return string $dropdown The filtered markup of the dropdown.
+	 */
+	public function modify_select( $dropdown ) {
+		$new_select = sprintf( '<select on="change:widget-categories-dropdown-%d.submit"', esc_attr( $this->number ) );
```
westonruter: Brilliant use of `$this->number` here.

amp.php
kienstra: Hi @westonruter, good catch. It should at least check that it [is_amp_endpoint()](https://github.com/Automattic/amp-wp/blob/develop/includes/amp-helper-functions.php#L120). We wouldn't want these subclass widgets on a standard WP page.

includes/widgets/class-amp-widget-categories.php
kienstra: Thanks, Weston!

includes/widgets/class-amp-widgets.php
kienstra: Hi @weston,
That's a good idea. Here's the [class_exists()](https://github.com/Automattic/amp-wp/pull/870/files#diff-4a5a57759183e45d3afe37b6f63e406cR55) check, and the function is slightly refactored.

includes/widgets/class-amp-widget-categories.php
kienstra: Thanks, @westonruter. That `@codingStandardsIgnoreLine` was for the text domain. Thanks for suggesting a solution. It's [applied here](https://github.com/Automattic/amp-wp/pull/870/files#diff-6b216d983cf0a62e9c648934c5687456R10).

includes/widgets/class-amp-widget-categories.php
kienstra: Thanks, [this line](https://github.com/Automattic/amp-wp/pull/870/files#diff-93b74d684677d7f182014458dee2af14R42) applies your suggestion.

amp.php
westonruter: The theme should be styling whatever markup the the widgets in core (or the plugin) output. The AMP widget functionality you're adding here should only apply if `is_amp_endpoint()` because this will true both when viewing AMP in paired mode and when canonical.

kienstra: Thanks, @westonruter. You probably saw it, but [this commit](https://github.com/Automattic/amp-wp/pull/870/commits/74d727b1f96035d01a1db594502891cc4bc6ed7a) applies your suggestion. 

I had tried to wrap the action hook in the conditional:

```add_action( 'init', 'amp_add_widget_support' )```
But this produced a notice:
```
is_amp_endpoint was called <strong>incorrectly</strong>. is_amp_endpoint() was called before the &#39;parse_query&#39; hook was called.
```

Comment (westonruter) [reply amp.php]: I've pushed a suggested change in dc3f5b4 to fix.

File: includes/widgets/class-amp-widget-categories.php (hunk 7 above)
Comment (westonruter) [reply includes/widgets/class-amp-widget-categories.php]: You can do the same with the `widget_categories_args` filter here.

File: includes/class-amp-theme-support.php [hunk 10]
```diff
@@ -154,6 +154,8 @@ public static function register_paired_hooks() {
 	 */
 	public static function register_hooks() {
 
+		add_action( 'widgets_init', array( __CLASS__, 'register_widgets' ) );
```
kienstra: Hi @westonruter, 
It looks like '`widgets_init'` has already fired by the time `register_hooks()` executes. 

`AMP_Theme_Support::init()` [is called on the 'wp' action](https://github.com/Automattic/amp-wp/blob/develop/amp.php#L137), which is after `'widgets_init'`.

My suggestion of checking `is_amp_endpoint()` probably led to this. That function [needs to be called](https://github.com/Automattic/amp-wp/blob/develop/includes/amp-helper-functions.php#L126) before `'parse_query'`, which is already after `'widgets_init'`.

westonruter:
Any of the widgets that are subclassed to merely do output buffering would be able to be removed once #888 is merged.

includes/class-amp-theme-support.php
westonruter: @kienstra great point. It looks like maybe the theme-support logic needs to be removed from `amp_maybe_add_actions()` to instead be called earlier? But yeah, tough spot with the dependency on the `parse_query` action.

kienstra: **Moved register_widgets()**

Hi @westonruter,
What do you think of [this commit](https://github.com/Automattic/amp-wp/pull/870/commits/6053f9a0d4a4fff51a423845ecd28e86903d1988), which ensures that `register_widgets()` fires in the right place?

kienstra:
*Regression: amp-form extension*

It looks like the 'Archives' and 'Categories' widgets need the `amp-form` extension:

```
The tag 'FORM [method=GET]' requires including the 'amp-form' extension JavaScript
```

This [edit|https://github.com/Automattic/amp-wp/commit/da6cee6c2784f58d145a37bb06ad18197efee420#diff-2585472f207548f626a43a7ff3cab922L448] might be related to it. But I'll look at this more tomorrow (Tuesday).

kienstra:
**Will Look At 'Categories' and 'Archives' Regression**

Hi @westonruter,
The full-page sanitization in #888 simplifies this a lot. The [commit](https://github.com/Automattic/amp-wp/pull/870/commits/c28578cd93d32ce78541772d448a35705d40e660) above removes the widget subclasses that only sanitized the output.

There looks to be an issue with the 'Archives' and 'Categories' dropdowns being stripped in sanitization. I'll look at that.

includes/widgets/class-amp-widget-archives.php
```diff
@@ -0,0 +1,122 @@
...
+
+	/**
+	 * Add a form extension script for AMP to output.
+	 *
+	 * @param array $scripts The AMP scripts.
+	 * @return array $scripts The filtered AMP scripts.
+	 */
+	public function form_script( $scripts ) {
+		if ( ! isset( $scripts['amp-form'] ) ) {
+			$scripts['amp-form'] = 'https://cdn.ampproject.org/v0/amp-form-latest.js';
```
westonruter: This is unnecessary once you merged from `develop` in 62a924ff761b0a24985b4974d97aa77ffda466b7

includes/widgets/class-amp-widget-archives.php
```diff
@@ -0,0 +1,103 @@
...
+	 *
+	 * @since 0.7.0
+	 *
+	 * @param array $args Widget display data.
+	 * @param array $instance Data for widget.
+	 * @return void.
+	 */
+	public function widget( $args, $instance ) {
+		if ( ! is_amp_endpoint() ) {
+			parent::widget( $args, $instance );
```
westonruter: Excellent move to defer the `is_amp_endpoint()` check to when the `widget` is actually called.

Issue #862, opened 2018-01-15
Support submitting comments to WordPress


ThierryA:
### Acceptance Criteria

AC1: The commenting experience that is native to WordPress needs to be supported in the AMP theme.
AC2: A user will need the ability to access the comment form, enter comment, and successfully submit a comment, as they do today.

#### Tasks
- [x] Add comment list template
- [x] Add comment form template + submission

alcurrie:
**These are the same notes as in #862**
Functionality looks good and works well, but I noticed a few items that were not as expected:

1) **Issue:** Comment moderation does not display comments for previously approved comment authors. 
**Expected**
In the Discussion Settings, when the Before a comment appears:  Comment must be manually approved (is 'unchecked'), and  Comment Author must have a previously approved comment is checked  (See screen cast of settings  https://cl.ly/442k1O2z1J0b)
I expect that when I enter the same Name and email as a user whose comments have already been approved, that my comment will display right away. 
**Actual**
Even if the comment author has been previously approved, comment does not display. 

2) The confirmation of the comment post, is too close to the Post comment button:
![screen shot 2018-02-05 at 7 27 41 pm](https://user-images.githubusercontent.com/18579988/35835801-cf36935e-0aaa-11e8-90fd-ff75d201dd61.png)

3) From a usability perspective the comments seem quite far away from the actual article with several other articles between the bottom of the article the user is commenting on, with 3 other articles between the article and the comments on the article
https://cl.ly/1x0g0H0K421k

csossi:
see Anne Louise's comments above

Issue #797, opened 2017-11-06
Add support for comments


westonruter:
In order to make AMP Canonical possible (#668) we'll need native support for WordPress commenting. We could create an `amp-wordpress-comments` component, and extend `comment_form()` and `wp_list_comments()` to accept `format=amp` which would then render this. 

See also:
https://www.ampproject.org/docs/tutorials/login_requiring 
https://ampbyexample.com/samples_templates/comment_section/

amedina:
@westonruter creating an amp-wordpress-comments component may not be necessary (and it would be hard to land it into the library); but the referenced methods can be extended to generate a  AMP-compatible format.

westonruter:
I saw the [`amp-facebook-comments`](https://github.com/ampproject/amphtml/tree/master/extensions/amp-facebook-comments) component so I thought that might be a path, but yeah, it probably wouldn't be necessary since we could just use `amp-form` and other existing components.

amedina:
Thinking more in general, I wonder if it would make sense to propose the creation of a set of AMP components that are specific to WordPress; given the prevalence of WP in the web ecosystem, the requirement of "general utility" would be satisfied. Will run the idea around to gather perspective.

DavidCramer:
further information on this is on https://github.com/Automattic/amp-wp/issues/862

westonruter:
The bare minimum it seems to add AMP support for the comment list is just to make sure that Gravatars use `amp-img` as opposed to `img`. In short, the following could be adapted and integrated into the theme support class:

```php
add_filter( 'get_avatar', function( $avatar ) {
	$avatar = str_replace( '<img', '<amp-img', $avatar );
	return $avatar;
} );
```

westonruter:
It would be great to eventually support [`amp-list`](https://www.ampproject.org/docs/reference/components/amp-list) for comments, but that will require a lot more work, especially to support nested comment replies.

alcurrie:
Functionality looks good and works well, but I noticed a couple of items that were not as expected:

1) **Issue:** Comment moderation does not display comments for previously approved comment authors. 
**Expected**
In the Discussion Settings, when the Before a comment appears:  Comment must be manually approved (is 'unchecked'), and  Comment Author must have a previously approved comment is checked:   https://cl.ly/442k1O2z1J0b
I expect that when I enter the same Name and email as a user whose comments have already been approved, that my comment will display right away. 
**Actual**
Even if the comment author has been previously approved, comment does not display. 

[...]

PR #871, opened 2018-01-17
Add support for comments

DavidCramer:
See #797, #862.


includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -0,0 +1,74 @@
...
+			return;
+		}
+
+		for ( $i = $num_nodes - 1; $i >= 0; $i-- ) {
+			$node = $nodes->item( $i );
+			if ( ! $node instanceof DOMElement ) {
+				continue;
+			}
+
+			if ( ! $node->hasAttribute( 'action' ) || '' === $node->getAttribute( 'action' ) ) {
```
westonruter: Per the spec: https://www.ampproject.org/docs/reference/components/amp-form#target

> This attribute is required for `method=GET`. For `method=POST`, the `action` attribute is invalid, use `action-xhr` instead.

Is the second part of this conditional intended to check if `'get' === $node->getAttribute( 'method' )`? Since the default method is `get` then I think there should be something above that does:

```php
$method = 'get';
if ( $node->hasAttribute( 'method' ) ) {
    $method = strtolower( $node->getAttribute( 'method' ) );
}
```

And then use the `$method` variable going forward.

includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -0,0 +1,74 @@
...
+				$node->parentNode->removeChild( $node );
```
westonruter: I don't think this is right. If `method` is `get` and there is no `action`, then what this really should do I think is just set the `action` to be `esc_url_raw( 'https://' . $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'] )`, because the default `action` for a `form` is the URL that the form appears on.

includes/utils/class-amp-dom-utils.php
```diff
@@ -35,7 +33,7 @@ public static function get_dom_from_content( $content ) {
 		 * Add utf-8 charset so loadHTML does not have problems parsing it.
 		 * See: http://php.net/manual/en/domdocument.loadhtml.php#78243
 		 */
-		$result = $dom->loadHTML( '<html><head><meta http-equiv="content-type" content="text/html; charset=utf-8"></head><body>' . $content . '</body></html>' );
+		$result = $dom->loadHTML( $document );
```
westonruter: Changes here are conflicting in `develop`.

includes/class-amp-theme-support.php
```diff
@@ -347,7 +347,14 @@ public static function finish_output_buffering( $output ) {
 			1
 		);
 
+		$dom = AMP_DOM_Utils::get_dom( $output );
+		// Sanitize forms in the document.
+		$sanitizer = new AMP_Form_Sanitizer( $dom );
+		$sanitizer->sanitize();
+
 		// @todo Add more validation checking and potentially the whitelist sanitizer.
+		$output = $dom->saveHTML();
```
westonruter: @DavidCramer I chatted about this with @ThierryA, but I think we should avoid having to load the entire response into the DOM for sanitization. I think that we should limit sanitization to just the post content, 3rd-party widgets, and other “leaf nodes” which we know are liable to be invalid AMP.

Is this right? Or am I off track? My worry is that loading the HTML into a DOMDocument for every request will be too heavy. But at the same time, if we're already spinning up a DOMDocument for every instance of `the_content()` maybe it is actually better to just do one for the entire HTML. This would certainly simplify somethings for sanitizing widgets (cc @kienstra). It could also mean that we wouldn't have to have this ad hoc logic to search for elements requiring components: https://github.com/Automattic/amp-wp/blob/e60cb152a50e2ffbf5504d41aee859ad1d0e0baa/includes/class-amp-theme-support.php#L448-L455

Maybe we should decide to not prematurely optimize and instead keep things simple at first and just go ahead and sanitize the entire response. Thoughts?

westonruter: See findings at https://github.com/Automattic/amp-wp/issues/875#issuecomment-358491587

It seems we should definitely not sanitize the entire output buffer.

DavidCramer: @westonruter whats the consensus now on this? It looks like your findings indicated it may be a good idea to sanitize the whole buffer output. I personally think this is the most efficient. I spoke with @ThierryA about this yesterday and he was going to do some tests, like you did.

westonruter: I'm leaning toward sanitizing the entire output buffer now.

westonruter: @DavidCramer do you want to build off of https://github.com/Automattic/amp-wp/issues/875#issuecomment-358560145 in a new PR to serve as a basis for what you're introducing here with form sanitization?

westonruter:
@DavidCramer FYI: There won't need to be a `get_scripts()` method defined in the form sanitizer once #882 is merged.

ThierryA:
@DavidCramer #882 was just merged to `develop` so you can apply @westonruter [suggestion here](https://github.com/Automattic/amp-wp/pull/871#issuecomment-358901499).

includes/class-amp-comment-walker.php
```diff
@@ -0,0 +1,107 @@
...
+		$template  = '<comments-template>';
+		$template .= parent::paged_walk( $elements, $max_depth, $page_num, $per_page, $args[0] );
+		$template .= '</comments-template>';
+
+		$url = get_rest_url( get_current_blog_id(), 'amp/v1/comments/' . get_the_ID() );
+		if ( strpos( $url, 'http:' ) === 0 ) {
+			$url = substr( $url, 5 );
+		}
+		// @todo Identify arguments and make filterable/settable.
+		$template .= '<amp-list src="' . esc_attr( $url ) . '" height="400" single-item="true" layout="fixed-height">';
```
westonruter: What about calling the actual comments walker here too and provide the content as [placeholder/fallback content](https://www.ampproject.org/docs/reference/components/amp-list#placeholder-and-fallback)? Maybe that is overkill.

DavidCramer: That's not a bad idea. However, it would increase page weight. But setting Adobe kind of fallback is great idea.

includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -0,0 +1,78 @@
...
+			}
+
+			$method = 'get';
+			if ( $node->hasAttribute( 'method' ) ) {
+				$method = strtolower( $node->getAttribute( 'method' ) );
+			}
+
+			// Get the action URL.
+			if ( ! $node->hasAttribute( 'action' ) ) {
+				$action_url = esc_url_raw( '//' . $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'] ); // WPCS: ignore. input var okay, sanitization ok.
```
DavidCramer: @westonruter I don't normally like using `$_SERVER` if I can avoid it, but this is from your suggestion here: https://github.com/Automattic/amp-wp/pull/871#discussion_r162148603
I feel this should be a little cleaner, but can't think how. any suggestions?

includes/class-amp-theme-support.php
```diff
@@ -540,6 +540,23 @@ public static function filter_the_content( $content ) {
...
+		$component_types = apply_filters( 'amp_component_types', array(
+			'amp-mustache' => 'template',
+		) );
+
+		if ( ! empty( $component_types[ $component ] ) ) {
+			return $component_types[ $component ];
+		}
+		return 'element';
+	}
+
```
DavidCramer: @westonruter component scripts are have `custom-element` in the script tags, however `amp-mustache` used for `amp-list` does not, instead it has `custom-template`. see https://www.ampproject.org/docs/reference/components/amp-mustache
It looks like this is the only component that's different, however I separated it to its own method in case the spec changes, or additionals are added in future.

DavidCramer:
@westonruter Regarding moving the sanitisation from `the_content` to the output buffer: I needed this in order to have the comments to work, which I have done in my dev, but have not added to this PR. Is that still under discussion or could I do a PR on it based on what I've already got?

westonruter:
@DavidCramer we'll go ahead and sanitize the entire `body`, or actually, the entire `html` document.  I was also going to work on a PR, so if you can get one up first that is good as well.

includes/class-amp-theme-support.php
westonruter: I don't think get_component_type would be needed if we sanitize the entire body, right?

includes/sanitizers/class-amp-form-sanitizer.php
westonruter: I think this is right what you have here, except that there should a beautiful `wp_unslash()` call added to it, since `$_SERVER` vars are auto-slashed for back-compat reasons:

```
esc_url_raw( '//' . wp_unslash( $_SERVER['HTTP_HOST'] ) . wp_unslash( $_SERVER['REQUEST_URI'] ) );
```

The one for `HTTP_HOST` is somewhat overkill.

includes/amp-rest-functions.php
```diff
@@ -0,0 +1,69 @@
...
+		'parent'  => $parent,
+		'order'   => 'ASC',
+	) );
+
+	$return = array();
+	foreach ( $comments as $comment ) {
+		$GLOBALS['comment']    = $comment; // WPCS: override ok.
+		$comment->comment_date = get_comment_date();
+		$comment->comment_time = get_comment_time();
+		$comment->comment      = amp_get_comments_recursive( $id, $comment->comment_ID );
```
westonruter: I believe this will have performance problems, as it could result in a lot of DB queries. If you look at `wp_list_comments()` I believe you can see it gets a flat list of all comments for the post, and then it passes it into `Walker::paged_walk()` which then goes about identifying which comments are children of other comments.

westonruter: That is, `AMP_Comment_Walker:: paged_walk()` here.

westonruter:
@DavidCramer I cherry-picked one of your commits and added another to implement output-buffered `body` sanitization in #888.

includes/amp-rest-functions.php
DavidCramer: @westonruter Agreed. Can revise it to be more efficient.

DavidCramer: @westonruter how about this? b7c5963

includes/class-amp-theme-support.php
DavidCramer: @westonruter I think we still do, since this is done after the sanitisation and the scripts are placed in. All the scripts are `custom-element` but the template is `custom-template`.

includes/sanitizers/class-amp-form-sanitizer.php
DavidCramer: agreed on the overkill. but will unslash the request uri.

DavidCramer:
@westonruter You can start reviewing this, but there is an issue with the url template strings I need to fix. I'm needing to head off now, but will finish when I'm back. If not by the time you get on, I'll sort it in the morning.

westonruter:
> there is an issue with the url template strings I need to fix

Shoot. This is a side effect of switching from `saveXML` to `saveHTML` in #891.

Similar to what was done in #895 with `AMP_DOM_Utils::convert_amp_bind_attributes()`, we need to do a temporary replacement of `amp-mustache` tokens during parsing and serialization, then swap them back after. I'm working on the change.

westonruter:
We talked it over and we decided to try the route of `amp-live-list` instead and see how it goes. 🎉

PR #872, opened 2018-01-17
Improved page default status logic and add overwrite ability; ensure front page and page for posts can be served as AMP when enabled

### LINKED ISSUE #837: Allow AMP to be disabled by default for individual Posts
#813 allowed AMP to be disabled Per-Post which is excellent, but it seems that there's no reliable way to invert the assumption that all Posts for each enabled Post Type should load AMP by default. 

The way the Post Meta currently works only stores data if AMP is disabled, so attempting to Filter that output using the `get_{$meta_type}_metadata` Filter would not be reliable as you cannot differentiate between a Post that has not been edited since the AMP plugin was activated and a Post that has AMP enabled (As not checking "Disabled" just deletes the Post Meta, which for all intents and purposes looks the same as a Post that has not been edited since enabling the plugin).

The `amp_skip_post` Filter exists, but without knowing whether the Post has AMP intentionally enabled for the Post or not it is not much help here.

### LINKED ISSUE #867: Version 0.6 get_support_errors() filter?
I have previously provided support in a theme for having a chosen page as AMP front page by manually changing the global WP_Query, setting proper template and calling `amp_prepare_render()` when applicable.

0.6 introduced a problem when trying to render an AMP front page that is also the main front page due to this code [here](https://github.com/Automattic/amp-wp/blob/0.6/includes/class-amp-post-type-support.php#L85):

```php
// Homepage and page for posts are not supported yet.
if ( 'page' === get_post_type( $post ) && 'page' === get_option( 'show_on_front' ) ) {
	if ( (int) get_option( 'page_for_posts' ) === (int) $post->ID ) {
		$errors[] = 'page-for-posts';
	} elseif ( (int) get_option( 'page_on_front' ) === (int) $post->ID ) {
		$errors[] = 'page-on-front';
	}
}
```

It forces the AMP to be disabled in the [metabox](https://i.imgur.com/J43zf73.png) and redirects to canonical when visiting it.

I think if the `AMP_Post_Type_Support::get_support_errors()` [return array](https://github.com/Automattic/amp-wp/blob/0.6/includes/class-amp-post-type-support.php#L110) had a filter I could just unset the `'page-for-posts'` and `'page-on-front'` error values and would be able to keep the support for front pages.

ThierryA:
This PR implements the workflow [discussed here](https://github.com/Automattic/amp-wp/issues/837#issuecomment-358144627). For pages with custom templates assigned or for pages assigned to the front page or posts page, the default status will be disabled. For all others, the default status will be enabled.

The default status has a filter to allow devs to change the default status to disabled as such `add_filter( 'amp_status_default_enabled', '__return_false' );`

Closes #837
Closes #867

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -158,20 +158,27 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
 	}
 }
 
+/**
+ * Add AMP actions when the request can be served as AMP.
+ *
+ * Actions will only be added if the request is for a singular post (including front page and page for posts), excluding feeds.
+ *
+ * @since 0.2
+ */
 function amp_maybe_add_actions() {
-	if ( ! is_singular() || is_feed() ) {
+	global $wp_query;
+	if ( ! ( is_singular() || $wp_query->is_posts_page ) || is_feed() ) {
 		return;
 	}
-
 	$is_amp_endpoint = is_amp_endpoint();
 
-	// Cannot use `get_queried_object` before canonical redirect; see https://core.trac.wordpress.org/ticket/35344
-	global $wp_query;
-	$post = $wp_query->post;
-
-	$supports = post_supports_amp( $post );
-
-	if ( ! $supports ) {
+	/**
+	 * Queried post object.
+	 *
+	 * @var WP_Post $post
+	 */
+	$post = get_queried_object();
+	if ( ! post_supports_amp( $post ) ) {
 		if ( $is_amp_endpoint ) {
 			wp_safe_redirect( get_permalink( $post->ID ), 301 );
 			exit;
--- a/amp.php
+++ b/amp.php
@@ -75,6 +75,7 @@ function amp_after_setup_theme() {
 	add_filter( 'amp_post_template_analytics', 'amp_add_custom_analytics' );
 	add_action( 'wp_loaded', 'amp_post_meta_box' );
 	add_action( 'wp_loaded', 'amp_add_options_menu' );
+	add_action( 'parse_query', 'amp_correct_query_when_is_front_page' );
 	AMP_Post_Type_Support::add_post_type_support();
 }
 add_action( 'after_setup_theme', 'amp_after_setup_theme', 5 );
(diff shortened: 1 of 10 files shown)


westonruter:
@ThierryA do you think `amp_status_default_enabled` is better than `amp_post_status_default_enabled` since the status in question is tied to a given `$post`?

What would make most sense in light of theme support being added in 0.7? In 0.7 there is the `available_callback` which I see we're missing an important piece before calling. Namely, inside of `\AMP_Theme_Support::is_paired_available()` we need to add:

```diff
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -77,6 +77,10 @@ class AMP_Theme_Support {
 			return false;
 		}
 
+		if ( is_singular() && ! post_supports_amp( get_queried_object() ) ) {
+			return false;
+		}
+
 		$args = array_shift( $support );
 
 		if ( isset( $args['available_callback'] ) && is_callable( $args['available_callback'] ) ) {

```

Naturally this shouldn't be part of this PR for 0.6. 

I suppose what this comes down to is whether there would be a default status for non-post things, like terms. In that case, there could conceivably be a `amp_term_status_default_enabled` and a `term_supports_amp()` (however unlikely). Not a big deal either way for the filter naming, but maybe better to be explicit that it is specifically regarding posts to be more future proof.

westonruter:
As an extra safeguard, @ThierryA and I have discussed having the Page support not enabled by default. To enable Page support, the user would just enable Page checkbox in the AMP settings. By leaving it off by default, it will be safer to upgrade to 0.6.

westonruter:
The three new commits here ensure that when you enable AMP for the homepage and posts (blog) page, that the AMP versions of them can indeed be served and that the templates can be filtered as required.

For example, if you want to serve out a custom page template for the homepage and posts page, you could put an `amp/home.php` and a `amp/blog.php` in your theme, and use them via the following in your `functions.php`:

```php
function xyz_filter_home_and_blog_amp_post_template( $template, $template_type, $post ) {
	if ( 'page' === $template_type && 'page' === get_option( 'show_on_front' ) ) {
		if ( (int) get_option( 'page_on_front' ) === $post->ID ) {
			$template = dirname( __FILE__ ) . '/amp/home.php';
		} elseif ( (int) get_option( 'page_for_posts' ) === $post->ID ) {
			$template = dirname( __FILE__ ) . '/amp/blog.php';
		}
	}
	return $template;
}
add_filter( 'amp_post_template_file', 'xyz_filter_home_and_blog_amp_post_template', 10, 3 );
```

The posts page template would naturally need to have The Loop to actually show the posts in addition to the post content, and this could be accomplished by coping the `single.php` from the template and adding this after the `.amp-wp-article`:

```php
	<ul>
		<?php while ( have_posts() ) : the_post(); ?>
			<li><a href="<?php the_permalink(); ?>"><?php the_title(); ?></a></li>
		<?php endwhile; ?>
	</ul>
	<?php the_posts_pagination(); ?>
```

/cc @mikeschinkel

ThierryA:
> @ThierryA do you think amp_status_default_enabled is better than amp_post_status_default_enabled since the status in question is tied to a given $post?

@westonruter agreed, I [pushed a commit](https://github.com/Automattic/amp-wp/pull/872/commits/0a7f28c6ec106c3f634525353c0d7a091c99793f) which renames the filter to `amp_post_status_default_enabled`.

ThierryA:
Thanks for addressing the post meta sanitization issue @westonruter, approved!

mikeschinkel:
@westonruter Are you assuming this means AMP support for a full site?

westonruter:
@mikeschinkel no, but this change here means that AMP is supported for all pages, including the homepage and blog page. AMP support for the entire site is being worked on in the context of `amp` theme support. Just wanted you to be aware of this update for 0.6.

mikeschinkel:
@westonruter Thanks.  FYI, I ran into issues with the architecture and decided to work on a clean room implementation for my needs but with the potential of offering it back to this project if you are interested in it.  

It is going slower than I'd like, but for supporting AMP I'm actually using the data in [the protoascii files](https://github.com/ampproject/amphtml/blob/master/validator/validator-main.protoascii) to drive conversion.  The plan is to write a parser that converts protoascii into an object that I can `var_dump()` and then include that file in the plugin when needed, and my code then uses the rules in that file to decide how to generate code.

I think that's the best approach from a maintaining compatibility with future changes approach. Thoughts?

westonruter:
@mikeschinkel The plugin also is using data from the protoascii files. See https://github.com/Automattic/amp-wp/blob/develop/bin/amphtml-update.py

It generates [`class-amp-allowed-tags-generated.php`](https://github.com/Automattic/amp-wp/blob/develop/includes/sanitizers/class-amp-allowed-tags-generated.php) which is then used for whitelist sanitization via [`AMP_Tag_And_Attribute_Sanitizer`](https://github.com/Automattic/amp-wp/blob/develop/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php). 

The sanitizer can then be called via:

```php
$sanitized_content = AMP_Content_Sanitizer::sanitize( $content, amp_get_content_sanitizers(), $args )
```

Where `amp_get_content_sanitizers()` is a new helper function in 0.7-alpha.

mikeschinkel:
@westonruter Interesting. Will check it out as I work on my user case the next few days.

PR #878, opened 2018-01-18
Remove double border from grouped preview buttons

ktmn:
https://i.imgur.com/aui7Ypd.png

!important is needed for when the AMP preview button is .disabled

### DIFF
--- a/assets/css/amp-post-meta-box.css
+++ b/assets/css/amp-post-meta-box.css
@@ -8,7 +8,12 @@
 .wp-core-ui #preview-action.has-amp-preview #post-preview {
 	border-top-right-radius: 0;
 	border-bottom-right-radius: 0;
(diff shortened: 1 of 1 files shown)


westonruter:
@ktmn how about the hover states? Is there a problem with one of the buttons not getting a darker border on all sides?

westonruter:
Sorry, I'm re-looking at the PR and I see it does handle the hover state.

ThierryA:
@ktmn thanks for your suggestion and contribution. The preview button group styling follows the same button group styling used in WP Customizer (see below):

screen shot 2018-01-18 at 9 39 38 pm

ktmn:
I see. Looked like an oversight to me but if you want to keep it consistent with the customizer it's fine.

PR #882, opened 2018-01-19
Let whitelist sanitizer dictate the required AMP scripts via spec 

### LINKED ISSUE #875: Improve ability to use sanitizers to identify required AMP components from output buffer
Theme support currently has the following logic that us run on the buffered output:

https://github.com/Automattic/amp-wp/blob/e60cb152a50e2ffbf5504d41aee859ad1d0e0baa/includes/class-amp-theme-support.php#L448-L455

This is not ideal. Normally when the sanitizer run, they identify the AMP component scripts that are required once they detect that a given AMP component is present in the page. When the sanitizer is not involved in sanitizing however, then the plugin currently requires manual inclusion of the required AMP components (cf. #874):

https://github.com/Automattic/amp-wp/blob/e60cb152a50e2ffbf5504d41aee859ad1d0e0baa/includes/amp-post-template-actions.php#L120-L131

This isn't good. Merely the presence of an `<amp-analytics>` element in the response should _automatically_ result in the `amp-analytics` AMP component script being included. The same should go for when a `form` is included in a template, which currently results in issues like #802.

So one way or another, we need to identify which elements are used in the page and then add the required AMP components. With #857 we now have the output buffering so we _could_ just take the entire output and feed it into the sanitizers once, instead of doing so once for each instance of `the_content`. It would certainly be simpler to do this, but my main concern is what would be more performant.

If we didn't sanitize the entire buffered output, then we'd need another mechanism to look through the raw unparsed HTML to identify whether a given element is present which needs an AMP component. Searching unparsed HTML is going to be less reliable than looking at the DOM.

I've written more about this in https://github.com/Automattic/amp-wp/pull/871/files#r162198204

westonruter:
* Include `requires_extension` and `also_requires_tag_warning` from spec in `AMP_Allowed_Tags_Generated` so that the required AMP components can be determined for a given tag spec and returned via `AMP_Tag_And_Attribute_Sanitizer::get_scripts()`.
* Use `latest` versions of AMP component scripts instead of `0.1`. (Discussed with @amedina.)
* Eliminate the `get_scripts()` method implementations from all of the sanitizers since all of the scripts now identified by whitelist sanitizer.
* Remove temporary duplicated/incompleted/brittle AMP script component lookups in `AMP_Theme_Support::get_amp_component_scripts()`.

With the changes here, you can just drop in AMP components to your post content (or to the theme templates w/ theme support) and AMP component scripts will automatically be added for them. For example, try pasting the following into a new post:

```php
<amp-gist
    data-gistid="b9bb35bc68df68259af94430f012425f"
    layout="fixed-height"
    height="225">
</amp-gist>

  <amp-ad type="a9"
    width="300"
    height="250"
    data-aax_size="300x250"
    data-aax_pubname="test123"
    data-aax_src="302">
  </amp-ad>
```

Automatically you'll then see these scripts added to the `head`:

* `<script custom-element="amp-video" src="https://cdn.ampproject.org/v0/amp-video-latest.js" async></script>`
* `<script custom-element="amp-gist" src="https://cdn.ampproject.org/v0/amp-gist-latest.js" async></script>`

See #875.


ThierryA:
@westonruter how about running the embeds through the `AMP_Base_Sanitizer` class that we can suppress the embeds `get_scripts()`?

DavidCramer:
Sweet!

westonruter:
> how about running the embeds through the AMP_Base_Sanitizer class that we can suppress the embeds get_scripts()?

Excellent point. The embeds don't need to implement `get_scripts()` either.

PR #887, opened 2018-01-20
Introduce AMP_WP_Styles for doing enqueued scripts inline

### LINKED ISSUE #886: Honor filtering of Stylesheet URI to allow child themes.
Currently, there is no way to use a child theme with AMP due to the hard coded `get_template_directory() . '/style.css'`.

https://github.com/Automattic/amp-wp/blob/89e4227edf41a8fe0408e8e855a6dcbbd4ccc13c/includes/class-amp-theme-support.php#L391

As I can see it, there are two options:

1. Use `get_stylesheet_uri()`: this would allow child themes to work and allow users to customize the location of the stylesheet (we usually keep it stored in an `assets/css` directory).
2. Use `get_stylesheet_directory() . '/style.css'`: this is still a bit restrictive, but would allow the `style.css` file to either be from the child or current theme. Since child themes require a `style.css` file, this should always be present when using a child theme.

westonruter:
This makes enqueueing CSS much more seamless. The only requirement is that you enqueue CSS that exists on the filesystem in a theme/plugin, wp-admin, or wp-includes. It also handles added inline styles, locale stylesheets, RTL styles, and Custom CSS.

The `AMP_WP_Styles` can then be used to house the logic for “tree shaking” the CSS rules as well. For tree-shaking, we'd need to add a basic CSS tokenizer-parser, with a method that takes the `$dom` object from #875 for the buffer that is output, and then this can be used to grab all of the `class` attributes; or it could just use a regexp to grab all of the `class="(.+?)"` from the output buffer.

Fixes #886.

### DIFF
--- a/includes/class-amp-wp-styles.php
+++ b/includes/class-amp-wp-styles.php
@@ -58,7 +58,7 @@ class AMP_WP_Styles extends WP_Styles {
 
 		if ( ! preg_match( '/\.(css|less|scss|sass)$/i', $css_path ) ) {
 			/* translators: %1$s is stylesheet handle, %2$s is stylesheet URL */
-			return new WP_Error( 'amp_css_path_not_found', sprintf( __( 'Skipped stylesheet %1$s which does not have recognized CSS file extension (%2$s).', 'amp' ), $handle, $src ) );
+			return new WP_Error( 'amp_css_bad_file_extension', sprintf( __( 'Skipped stylesheet %1$s which does not have recognized CSS file extension (%2$s).', 'amp' ), $handle, $src ) );
 		}
 
 		if ( ! $css_path || false !== strpos( '../', $css_path ) || 0 !== validate_file( $css_path ) || ! file_exists( $css_path ) ) {
@@ -135,7 +135,7 @@ class AMP_WP_Styles extends WP_Styles {
 		foreach ( array_filter( array( $css_file_path, $css_rtl_file_path ) ) as $css_path ) {
 			$css = file_get_contents( $css_path ); // phpcs:ignore -- It's a local filesystem path not a remote request.
 			if ( 'all' !== $media ) {
-				$css .= sprintf( '@media %s { %s }', $media, $css );
+				$css = sprintf( '@media %s { %s }', $media, $css );
 			}
 			$this->print_code .= $css;
 		}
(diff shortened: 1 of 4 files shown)


includes/class-amp-wp-styles.php
```diff
@@ -0,0 +1,216 @@
...
+		$src = preg_replace( ':[\?#].+:', '', $src );
+
+		$src = esc_url_raw( $src );
+
+		// @todo Explicitly not using includes_url() or content_url() since filters may point outside filesystem?
+		$includes_url = includes_url( '/' );
+		$content_url  = content_url( '/' );
+		$admin_url    = get_admin_url( null, '/' );
+		$css_path     = null;
+		if ( 0 === strpos( $src, $content_url ) ) {
```
ThierryA: What if the assets are in root folder? For instance could use the filesystem to create dir and cache all CSS which is then enqueued. [Here here is a function](https://github.com/Getbeans/Beans/blob/development/lib/api/utilities/functions.php#L174) which converts url to path, has be perfected over the years to fix all bugs reported on all sorts of various WP sites structure.

tests/test-class-amp-wp-styles.php
```diff
@@ -0,0 +1,199 @@
...
+		parent::tearDown();
+		$wp_styles = null;
+	}
+
+	/**
+	 * Get wp_styles().
+	 *
+	 * @return WP_Styles|AMP_WP_Styles Styles.
+	 */
+	protected function amp_wp_styles() {
```
ThierryA: This method can be removed and `AMP_Theme_Support::override_wp_styles()` can be used instead if/when the comment left on the `override_wp_styles()` are applied.

includes/class-amp-wp-styles.php
westonruter: No assets from core, themes or plugins would be in the root folder. So I don't think we should accommodate them. Since we're actually getting the file source from the filesystem (rather than relying on whether the requesting user has HTTP access) and outputting it, I think we should be very restrictive about what we allow. For example, if there was a `/private/secrets.js` or something, we would not want this file to be enqueued. True we're talking about CSS here, but still.

includes/class-amp-wp-styles.php
```diff
@@ -0,0 +1,216 @@
...
+	 *
+	 * @since 0.7
+	 * @see WP_Styles::_css_href()
+	 *
+	 * @param string $src The source URL of the enqueued style.
+	 * @param string $handle The style's registered handle.
+	 * @return string|WP_Error Style's absolute validated filesystem path, or WP_Error when error.
+	 */
+	public function get_validated_css_file_path( $src, $handle ) {
+		if ( ! is_bool( $src ) && ! preg_match( '|^(https?:)?//|', $src ) && ! ( $this->content_url && 0 === strpos( $src, $this->content_url ) ) ) {
```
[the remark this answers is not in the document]
westonruter: It's like this because it is forked/copied from core: https://github.com/WordPress/wordpress-develop/blob/4fb6c02b9961d41dcaaf0ea04e24223993fa8667/src/wp-includes/class.wp-styles.php#L327

includes/class-amp-wp-styles.php
ThierryA: Sounds good, I attempted to test on Windows as I recalled issues when doing URL to path replacements but gave up due to issues encountered when trying to run a stable local install on my virtual machine. I will leave that task to QA (Claudio).

PR #888, opened 2018-01-23
Sanitize entire HTML output when theme support is present

### LINKED ISSUE #875: Improve ability to use sanitizers to identify required AMP components from output buffer

westonruter:
See #875.

This largely eliminates the need for `amp_component_scripts` entirely because the AMP component scripts are discovered via the whitelist sanitizer in #882. So for example, the Form sanitizer in #871 can be used without needing to define its `get_scripts()` method.

For another PR:

* Fix character encoding issue. See https://github.com/Automattic/amp-wp/issues/855#issuecomment-359703587
* Add elements from `head` to sanitization.

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -495,13 +495,16 @@ class AMP_Theme_Support {
 		 * from outside the body from being part of the whitelist sanitizer when it runs when theme support is not present,
 		 * as otherwise elements from the HEAD could get added to the BODY.
 		 */
-		$sanitized_inner_body = AMP_DOM_Utils::get_content_from_dom( $dom );
-		$output = preg_replace( '#(<body.*?>)(.+)(</body>)#si', '$1' . $sanitized_inner_body . '$3', $output );
+		$output = preg_replace(
+			'#(<body.*?>)(.+)(</body>)#si',
+			'$1' . AMP_DOM_Utils::get_content_from_dom( $dom ) . '$3',
+			$output
+		);
 
 		// Inject required scripts.
 		$output = preg_replace(
 			'#' . preg_quote( self::COMPONENT_SCRIPTS_PLACEHOLDER, '#' ) . '#',
-			self::get_amp_component_scripts( $output ),
+			self::get_amp_component_scripts(),
 			$output,
 			1
 		);
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -482,6 +482,11 @@ class AMP_Theme_Support {
 			'content_max_width' => ! empty( $content_width ) ? $content_width : AMP_Post_Template::CONTENT_MAX_WIDTH, // Back-compat.
 		);
 
+		$assets = AMP_Content_Sanitizer::sanitize_document( $dom, self::$sanitizer_classes, $args );
+
+		self::$amp_scripts = array_merge( self::$amp_scripts, $assets['scripts'] );
+		self::$amp_styles  = array_merge( self::$amp_styles, $assets['styles'] );
+
 		/*
 		 * @todo The sanitize method needs to be updated to sanitize the entire HTML element and not just the BODY.
 		 * This will require updating mandatory_parent_blacklist in amphtml-update.py to include elements that appear in the HEAD.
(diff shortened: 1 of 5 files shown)


includes/class-amp-theme-support.php
```diff
@@ -492,10 +470,32 @@ public static function start_output_buffering() {
...
+		 * since the the old paired mode is sending content from the *body* we'll need to be able to filter out the elements
+		 * from outside the body from being part of the whitelist sanitizer when it runs when theme support is not present,
+		 * as otherwise elements from the HEAD could get added to the BODY.
+		 */
+		list( $sanitized_inner_body, $scripts, $styles ) = AMP_Content_Sanitizer::sanitize( $dom, self::$sanitizer_classes, $args );
+
+		self::$amp_scripts = array_merge( self::$amp_scripts, $scripts );
+		self::$amp_styles  = array_merge( self::$amp_styles, $styles );
+
+		$output = preg_replace( '#(<body.*?>)(.+)(</body>)#si', '$1' . $sanitized_inner_body . '$3', $output );
```
DavidCramer: @westonruter here we're replacing the body with the sanitised body. This is because it's returning `AMP_DOM_Utils::get_content_from_dom( $dom )` from the sanitize method. Wouldn't this be easier if the method returned `$dom->saveHTML()` if the content passed is a complete document? This would stop the need to replace the body after it was sanitized.

westonruter: Yes, potentially. The `sanitize` method is better called `sanitize_content`. Better handling of sanitizing entire documents should be done.

westonruter: @DavidCramer how about this: https://github.com/Automattic/amp-wp/pull/888/commits/876c22fab1d70f9990ae42a52e1cc1bfd96c91d9

DavidCramer: @westonruter Thats nice, makes way for a full doc later on. I see that you made a comment about the `head` needs updating to `mandatory_parent_blacklist` so that makes sense now.

includes/class-amp-theme-support.php
```diff
@@ -492,10 +470,33 @@ public static function start_output_buffering() {
...
+		 * This will ensure that the scripts and styles that plugins output via wp_head() will be sanitized as well. However,
+		 * since the the old paired mode is sending content from the *body* we'll need to be able to filter out the elements
+		 * from outside the body from being part of the whitelist sanitizer when it runs when theme support is not present,
+		 * as otherwise elements from the HEAD could get added to the BODY.
+		 */
+		$sanitized_inner_body = AMP_DOM_Utils::get_content_from_dom( $dom );
+		$output = preg_replace( '#(<body.*?>)(.+)(</body>)#si', '$1' . $sanitized_inner_body . '$3', $output );
 
 		// Inject required scripts.
 		$output = preg_replace(
```
DavidCramer: @westonruter `get_amp_component_scripts` doesn't require the argument anymore.

westonruter:
Hummm, something isn't working quite right.

If I add this to a template:

```
<script type="text/javascript">bad</script>
```

It is not getting stripped out as expected.

DavidCramer:
hmm. testing as well.

westonruter:
😅 whew, it's just because #884 wasn't merged yet from the `0.6` branch. Fixed with merging in ef69f2d.

tests/test-class-amp-theme-support.php
```diff
@@ -67,4 +67,53 @@ public function test_is_paired_available() {
...
+					data-aax_pubname="test123"
+					data-aax_src="302"></amp-ad>
+				<?php wp_footer(); ?>
+			</body>
+		</html>
+		<?php
+		$original_html  = trim( ob_get_clean() );
+		$sanitized_html = AMP_Theme_Support::finish_output_buffering( $original_html );
+
+		$this->assertContains( '<meta charset="utf-8">', $sanitized_html );
```
ThierryA: @westonruter @DavidCramer these tests could have a bit more coverage here. I believe it is ok to do that at a later stage since the other components sanitization (form etc.) are still WIP.

Issue #841, opened 2018-01-04
Support all default embeds where relevant


MackenzieHartung:
### Acceptance Criteria

AC1: Add AMP support for “Media Video Playlist” content
AC2: Add AMP support for “Media Audio Playlist” content
AC3: Add AMP support for “Issu Embed” content
AC4: Enhance the support for “Post Embed” content. Currently, Post Embeds doubles content in places. The issue stems from the p.wp-embedded-content not getting replaced with the iframe.wp-embedded-content.
AC5: Enhance the support for “CollegeHumor Video Embed”. Currently the video player appears, but does not play. Ensure the video content plays.
AC6: Enhance the support for “Meetup Embed”. Currently, the markup appears, but there is no styling.  Ensure the styling works properly in AMP.
AC7: Enhance the support for “Reddit Embed”. Currently, some markups appear, but without styling. Ensure that the styling works properly in AMP.
AC8: Enhance the support for the “Screencast Embed”. Currently, even with the domain added to Chrome’s Flash “Allow” list, when playing it displays “Unable to display content. Adobe Flash is required.”. Ensure that the error does not display, and that the video plays successfully.
AC9: Enhance the support for “Tumblr Embed”. Currently, only the link displays. Ensure that the content displays properly within AMP.
AC10: Enhance the support for the “WordPress Plugin Directory Embed”. Currently it displays mostly as expected, but with slightly different styling and an extra link. Ensure that the styling appears properly, and an extra link is not displayed.

See #806  as reference

kienstra:
**Current Embed Support**

The matrix at the top of [this pull request](https://github.com/Automattic/amp-wp/pull/889) seems to track the current state of embed support. But I haven't verified it.

kienstra:
**Test Page**

Hi @csossi,
Could you please use this [test page](https://dev-ampconfdemo.pantheonsite.io/?page_id=313&preview=true), which @westonruter created?

csossi:
Thanks, gents - on it

csossi:
@westonruter, @kienstra  - QA Results:
- CloudUp Image Embed (blank/white space)
- CollegeHumor Embed - (blank/black space)
- Flickr Video Embed - image appears, but no embedded video
- Photobucket Embed - displays URL only
- Screencast embed - "Unable to display content. Adobe Flash os required"

Please let me know if any of these are "ok as is"

westonruter:
> CollegeHumor Embed - (blank/black space)

[...]

PR #889, opened 2018-01-23
#841 Improve support for default embeds

kaitnyl:
This addresses issue #841.
See full list of previous embed compatibility [here](https://github.com/Automattic/amp-wp/issues/806#issuecomment-350790526). 

I'm just going to highlight the updated/troubled ones below:

AMP Support | Status* | Note
------------ | ------------- | -----
Media Video playlist | ❌ | The amp-video element doesn't currently support playlists. I think we can add to this, but since it's already implemented at their level, we likely need some discussion around how to approach this.
Media Audio playlist | ❌ | The amp-audio element doesn't currently support playlists (see above).
Post Embed | 🔸 | Blockquote containing link is required because links in iframe cannot be clicked. See follow-up in #914.
CollegeHumor Video Embed | 🔸 | The embed attempts to load http content through an https amp-iframe src (amp-iframe requires https). This causes the browser to reject it.
Issuu Embed | ✅ | There's no AMP element ready for this type of element, but I've added a solution that wraps it within an amp-iframe so it will display.
Meetup Embed | ✅ | This embed uses an inline style tag. I've captured it from the default embed and output it in the head, while the embed will output similar to desktop.
Reddit Embed | ✅ | Now appears the same as desktop version via the amp-reddit element.
Screencast Embed | 🔸 | I'm seeing the same error appear on the desktop version. This looks like it might be the downside of Screencast using .swfs.
Tumblr Embed | 🔸 | There's no AMP element ready for this type of element, but I've added a solution that wraps it within an amp-iframe so it will display. Remaining issue is that the iframe does not auto-resize as needed.
WordPress Plugin Directory Embed | ✅ | Updated to no longer display the extra blockquote above it.

_*Status as of this diff._

Should we try and create our own version of amp-video and amp-audio element playlists? If we do now, they also likely wouldn't be valid AMP elements anymore.


kaitnyl:
Hi @ThierryA and @westonruter!

I'd love to get your input on some of the embed notes listed above. Ones like CollegeHumor and Screencast seem like they may be a bit incompatible here, but we do have some room for improvement with the addition of amp-video and amp-audio playlists.

includes/embeds/class-amp-reddit-embed-handler.php
```diff
@@ -0,0 +1,84 @@
...
+	 * Unregister embed.
+	 */
+	public function unregister_embed() {
+		wp_embed_unregister_handler( 'amp-reddit' );
+	}
+
+	/**
+	 * Include the required AMP Reddit scripts.
+	 */
+	public function get_scripts() {
```
westonruter: It should be unnecessary to implement `get_scripts()` because now the whitelist sanitizer will automatically discover the `amp-reddit` script as being required through sanitization when it runs across the `amp-reddit` element.

includes/embeds/class-amp-reddit-embed-handler.php
```diff
@@ -0,0 +1,84 @@
+<?php
+/**
+ * Class AMP_Reddit_Embed_Handler
+ *
+ * @package AMP
+ */
+
+/**
+ * Class AMP_Reddit_Embed_Handler
```
westonruter: Please add `@since` tags to the phpdoc.

includes/embeds/class-amp-reddit-embed-handler.php
```diff
@@ -0,0 +1,84 @@
...
+	 * @const string
+	 */
+	const URL_PATTERN = '#https?://(www\.)?reddit\.com/r/[^/]+/comments/.*#i';
+
+	/**
+	 * Script slug.
+	 *
+	 * @var string
+	 */
+	private static $script_src = 'https://cdn.ampproject.org/v0/amp-reddit-0.1.js';
```
westonruter: Per below, the `$script_src` needn't be defined at the embed level. See #885.

includes/embeds/class-amp-reddit-embed-handler.php
```diff
@@ -0,0 +1,84 @@
...
+			return '';
+		}
+
+		return AMP_HTML_Utils::build_tag(
+			'amp-reddit',
+			array(
+				'layout'         => 'responsive',
+				'data-embedtype' => 'post',
+				'width'          => '596',
+				'height'         => '141',
```
westonruter: Why are the `width` and `height` being hard-coded like this?

includes/embeds/class-amp-tumblr-embed-handler.php
```diff
@@ -0,0 +1,71 @@
+<?php
+/**
+ * Class AMP_Tumblr_Embed_Handler
+ *
+ * @package AMP
+ */
+
+/**
+ * Class AMP_Tumblr_Embed_Handler
```
westonruter: Please add `@since 0.7` tags to the phpdoc.

westonruter:
@kaitnyl Sorry for the delay:

> MeetUp: This one is a bit tricky. We want the embed that AMP is currently using, but there's inline styling that we don't have added. We can add it inline but risk not being valid with the CSS sanitizer. I'll continue investigating how to approach this one.

Leaving the CSS on the embed output should successfully get sanitized by the CSS sanitizer. It will move any inline `style` attributes to the HEAD.

> The embed attempts to load http content through an https amp-iframe src (amp-iframe requires https). This causes the browser to reject it.

Yeah, this embed doesn't work on desktop either. So it is broken on their side and we shouldn't try to fix it on our side.

> I'm seeing the same error appear on the desktop version. This looks like it might be the downside of Screencast using .swfs.

If it is trying to embed a SWF then I think it should be intentionally not supported.

> Should we try and create our own version of amp-video and amp-audio element playlists? If we do now, they also likely wouldn't be valid AMP elements anymore.

Good question. One possibility would be to just create endpoints from WP that render out the playlist in standalone form and then render it in an `amp-iframe`. That would work but it wouldn't be ideal. In particular it would not be ideal because each embed would get its own separate copy of MediaElement.js and any other scripts that the playlists depend on

westonruter:
> The amp-video element doesn't currently support playlists. I think we can add to this, but since it's already implemented at their level, we likely need some discussion around how to approach this. ¶ Should we try and create our own version of amp-video and amp-audio element playlists? If we do now, they also likely wouldn't be valid AMP elements anymore.

Good question. One possibility would be to just create endpoints from WP that render out the playlist in standalone form and then render it in an `amp-iframe`. This would be like the current Post Embeds. That would work but it wouldn't be ideal. In particular it would not be ideal because each embed would get its own separate copy of MediaElement.js and any other scripts that the playlists depend on.

What would be great to see is a custom AMP implementation of the audio and video playlist which made use of `amp-bind` and `amp-state` to manage the playback of `amp-video` and `amp-audio` components. A tutorial for how to make use of these can be seen at <https://www.ampproject.org/docs/tutorials/interactivity>. This should be done in a separate PR as it would be more work, but it will be extremely interesting and exciting!

westonruter:
Oh, and I should also mention `amp-selector` as the other key component that could be used to create playlists.

includes/embeds/class-amp-reddit-embed-handler.php
```diff
@@ -0,0 +1,85 @@
...
+	 * @var int
+	 */
+	protected $DEFAULT_HEIGHT = 141; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.MemberNotSnakeCase
+
+	/**
+	 * Default width.
+	 *
+	 * @var int
+	 */
+	protected $DEFAULT_WIDTH = 596; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.MemberNotSnakeCase
```
westonruter: Maybe the oEmbed handler can make a request off to a URL like https://www.reddit.com/comments/7t6msa.json?limit=1&app=embed to obtain the `width`/`height` specific for the given Reddit embed.

kaitnyl: I've gone ahead and made it follow 100%/100% for the responsive layout type. This solves longer embeds using media that were getting cut off by the fixed dimensions, but now suffers from the same issue as reported [here](https://github.com/ampproject/amphtml/issues/11869).

The dimensions given in https://www.reddit.com/comments/7t6msa.json?limit=1&app=embed are just for the media attachments and wouldn't encompass the entire embed content.

includes/embeds/class-amp-meetup-embed-handler.php
```diff
@@ -0,0 +1,72 @@
...
+			return '';
+		}
+
+		$content = wp_oembed_get( $args['url'] );
+		preg_match( '#<style.+</style>#', $content, $result );
+
+		// Outlying style tag from oembed.
+		if ( isset( $result[0] ) ) {
+			$css = str_replace( '<style type="text/css">', '<style amp-meetup>', $result[0] );
+			echo $css; // WPCS: XSS ok.
```
westonruter: In the builtin paired mode theme, the CSS is getting printed before the `<html>` starts. It should get removed entirely I believe.

westonruter: One additional reason why the CSS should be discarded is that it includes `!important` properties, which are illegal in AMP. If it didn't, then maybe it could make sense to move the CSS to the `amp-custom` style in the `head`. But alas, I think we just have to strip it entirely.

includes/embeds/class-amp-meetup-embed-handler.php
```diff
@@ -0,0 +1,71 @@
...
+	public function render( $args ) {
+		$args = wp_parse_args( $args, array(
+			'url' => false,
+		) );
+
+		if ( empty( $args['url'] ) ) {
+			return '';
+		}
+
+		$content = wp_oembed_get( $args['url'] );
```
westonruter: I'm seeing that `wp_oembed_get()` here is resulting in an HTTP request with every call. I was expecting it to be cached. Apparently to cache we need to use `$wp_embed->shortcode( array(), $args['url'] )`. This seems to be a problem in core as well with the Video widget: https://github.com/WordPress/wordpress-develop/blob/afb4f45e62520397b147e560e051d3940797121e/src/wp-includes/widgets/class-wp-widget-media-video.php#L144

I think core is erroneously doing an HTTP request for every load of a page with a video widget that is a non-YouTube and non-Vimeo oEmbed.

includes/embeds/class-amp-tumblr-embed-handler.php
```diff
@@ -0,0 +1,74 @@
...
+		$args = wp_parse_args( $args, array(
+			'url' => false,
+		) );
+
+		if ( empty( $args['url'] ) ) {
+			return '';
+		}
+
+		// Must enforce https for amp-iframe, but editors can supply either on desktop.
+		$args['url'] = str_replace( 'http://', 'https://', $args['url'] );
```
westonruter: Unfortunately it seems this is not working fully. When attempting to embed https://ifpaintingscouldtext.tumblr.com/post/92003045635/grant-wood-american-gothic-1930 then Tumblr will redirect from HTTPS to HTTP for some reason. When serving a page over HTTPS, as they should be, then the iframe is blocked from rendering.

kienstra:
**Playlist Embeds**

PR #954 (WIP) is handling playlist embeds.

PR #891, opened 2018-01-23
Update serialization to use HTML instead of XML

westonruter:
* Updates minimum required PHP version to 5.3. The AMP plugin currently supports PHP 5.2. Because of this it has meant that `DOMDocument::saveHTML($node)` has not available since it was introduced in 5.3. Using `saveXML` instead causes other problems, including injection of CDATA sections, unbooleanized attributes, and self-closing tags. Instead of trying to work around all of these issues to continue supporting PHP 5.2, I think it would be more sensible to update the minimum requirement to PHP 5.3.
* Fully deprecates blacklist sanitizer.


westonruter:
Rebased to resolve merge conflict. Former `HEAD` was 25b1fd9.

PR #900, opened 2018-01-24
#874 Add analytics to theme support

### LINKED ISSUE #874: Add analytics to theme support
Note comment:

https://github.com/Automattic/amp-wp/blob/e60cb152a50e2ffbf5504d41aee859ad1d0e0baa/includes/class-amp-theme-support.php#L169

With the initial introduction of theme support in in #857, the injection of analytics functionality was not ported over from the AMP post templates. This needs to be done.

kaitnyl:
Fixes #874.

The `amp-analytics` component and script are now output with theme support in the same fashion as the default AMP templates.

- I've added it to `wp_footer` since the default AMP template version is added inside `amp_post_template_footer`.
- The filter is using [amp_add_custom_analytics](https://github.com/Automattic/amp-wp/blob/develop/includes/admin/functions.php#L121) to get the configs from the backend.


kaitnyl:
Hi @ThierryA and @westonruter, would one of you mind reviewing this PR? It should be a quick win to get analytics integrated to new themes.

includes/class-amp-theme-support.php
```diff
@@ -331,6 +333,37 @@ public static function add_meta_viewport() {
...
+	 */
+	public static function add_analytics_data() {
+		$analytics_entries = apply_filters( 'amp_post_template_analytics', array() );
+
+		if ( empty( $analytics_entries ) ) {
+			return;
+		}
+
+		// Can enter multiple configs within backend.
+		foreach ( $analytics_entries as $id => $analytics_entry ) {
```
westonruter: Since the logic here is largely now a duplicate of `\amp_post_template_add_analytics_data()`, should a new function be added to `amp-helper-functions.php` called something like: `\amp_print_analytics()` which takes the `$analytics_entries` as its one arg? Then `\amp_post_template_add_analytics_data()` can be updated to call this common function.

includes/class-amp-theme-support.php
```diff
@@ -331,6 +333,37 @@ public static function add_meta_viewport() {
 		echo '<meta name="viewport" content="width=device-width,minimum-scale=1">';
 	}
 
+	/**
+	 * Print analytics data.
+	 */
+	public static function add_analytics_data() {
+		$analytics_entries = apply_filters( 'amp_post_template_analytics', array() );
```
westonruter: The other instances of the `amp_post_template_analytics` filter being applied are passing a `$post` as its second argument. This would be needed when a plugin needs to vary the analytics by the current post. Maybe this should, to maximize back-compat with existing plugins that filter `amp_post_template_analytics` do something like this:

```php
$analytics_entries = amp_add_custom_analytics( array() );
if ( is_singular() ) {
    $analytics_entries = apply_filters( 'amp_post_template_analytics', $analytics_entries, get_queried_object() );
}
```

For that matter, the current `amp_add_custom_analytics()` is somewhat bizzare in `amp.php`. Maybe it would be better to add another helper function to `amp-helper-functions.php` like `amp_get_analytics()` which then contains the body of what `amp_add_custom_analytics()` has today. Then at the end of that function, it can apply the `amp_post_template_analytics` filter per above. The `amp_get_analytics()` function could then be used in `\AMP_Post_Template::__construct()` instead of manually applying the `amp_post_template_analytics` filter there.

This could then be removed from `amp.php`:

```php
add_filter( 'amp_post_template_analytics', 'amp_add_custom_analytics' );
```

includes/class-amp-theme-support.php
```diff
@@ -331,6 +333,37 @@ public static function add_meta_viewport() {
 		echo '<meta name="viewport" content="width=device-width,minimum-scale=1">';
 	}
 
+	/**
+	 * Print analytics data.
+	 */
+	public static function add_analytics_data() {
```
westonruter: Add:

```php
/** This filter is documented in includes/templates/class-amp-post-template.php */
```

kaitnyl:
@westonruter I've move the analytics functions around and cleaned it up to reduce duplicate code. 

There are a few tests for analytics already. It looks like it goes over the saving/retrieving within the backend options, but it also confirms the output of script tag and one/multiple `amp-analytics` elements. Should new tests be added for [amp_get_analytics](https://github.com/Automattic/amp-wp/pull/900/commits/5dc9455c4fe301b1c61560fc8133b3b9108d1c3c#diff-5e173f2331b78ccd3da05258383dcc55R158) and [amp_print_analytics](https://github.com/Automattic/amp-wp/pull/900/commits/5dc9455c4fe301b1c61560fc8133b3b9108d1c3c#diff-5e173f2331b78ccd3da05258383dcc55R195)?

includes/amp-helper-functions.php
```diff
@@ -149,6 +149,76 @@ function amp_print_boilerplate_code() {
...
+		 * Add amp-analytics tags.
+		 *
+		 * This filter allows you to easily insert any amp-analytics tags without needing much heavy lifting.
+		 *
+		 * @since 0.4
+		 *
+		 * @param array   $analytics An associative array of the analytics entries we want to output. Each array entry must have a unique key, and the value should be an array with the following keys: `type`, `attributes`, `script_data`. See readme for more details.
+		 * @param WP_Post $post      The current post.
+		 */
+		$analytics_entries = apply_filters( 'amp_post_template_analytics', $analytics_entries, get_queried_object() );
```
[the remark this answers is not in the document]
kaitnyl: This makes sense. The filter being named specifically for the post template class could have it stay in that mode, while a new generic filter can be used going forward. I'll jump on these updates.

kaitnyl:
@westonruter I've restored the old `amp_add_custom_analytics()` per above. Within legacy, we end up using the old filter in case any existing plugins/themes are already making use of it.

For paired mode, we use the newer `add_action( 'wp_footer', 'amp_print_analytics' );` which has a new filter that's used in paired mode.

kienstra:
**Steps To Test**

1. Navigate to the [test site's AMP > Analytics page](https://test-ampconfdemo.pantheonsite.io/wp-admin/admin.php?page=amp-analytics-options)
2. In `Type`, enter `googleanalytics`
3. In `JSON Configuration`, enter [this configuration](https://github.com/Automattic/amp-wp/blob/develop/tests/test-amp-analytics-options.php#L15-L34):
```
{
	"requests": {
		"event": "https://example.com/..."
	},
	"triggers": {
		"trackPageview": {
			"on": "visible",
			"request": "event",
			"visibilitySpec": {
				"selector": "#cat-image-id",
				"visiblePercentageMin": 20,
				"totalTimeMin": 500,
				"continuousTimeMin": 200
			},
			"vars": {
				"eventId": "catview"
			}
		}
	}
}
```
4. View the HTML source of any page on the [test site](https://test-ampconfdemo.pantheonsite.io) (in Chrome, Developer > View Source)
5. Expected: there is an `<amp-analytics` tag with the data that you entered above. Example:
```
<amp-analytics id="ffc8f76dfe25" type="googleanalytics"><script type="application/json">{"requests":{"event":"https:\/\/example.com\/..."},"triggers":{"trackPageview":{"on":"visible","request":"event","visibilitySpec":{"selector":"#cat-image-id","visiblePercentageMin":20,"totalTimeMin":500,"continuousTimeMin":200},"vars":{"eventId":"catview"}}}}</script></amp-analytics>
```
5. Delete the analytics data you entered on the [AMP Analytics page](https://test-ampconfdemo.pantheonsite.io/wp-admin/admin.php?page=amp-analytics-options)

csossi:
verified in QA

Issue #843, opened 2018-01-04
Incorporate AMP validator into post authorship flow, to alert if content within the post editor is not compatible with AMP


MackenzieHartung:
### Acceptance Criteria

AC1: On saving a post on /wp-admin/post.php, if there’s post content that’s not compatible with AMP, display an error at the top of the page.
AC2: The error will not display specifically what is incompatible.
AC3: The error will display the following text: “Notice: your post fails AMP validation”.

kienstra:
**Request To Work On This**

Hi @ThierryA and @westonruter,
Could I please work on this? Though this is in [Sprint 3](https://github.com/Automattic/amp-wp/projects/3), It looks like we're doing well on [Sprint 2](https://github.com/Automattic/amp-wp/projects/2). But maybe you have other priorities in mind.

If so, I'll write a simple solution design here so we're on the same page.

Thanks!

westonruter:
@kienstra Please do. I see this highly related to #842, but probably a different mechanism. For content, I suppose it would involve hooking into the whitelist sanitizer to find elements that are dropped due to being illegal? If this could identify the specific paragraph in the content that has invalid content in it, that would be ideal. Naturally this would straightforward to accomplish with Gutenberg, and perhaps Gutenberg should be the focus for this here then.

For #842 my thought was do do something at a higher level, when hooks are run, to run the validator on the entire page output rather than on just the content. 

I don't see these two as being mutually exclusive. One is focused on the content authorship—telling users what they can't author—and the other is focused on the site administration—telling users what plugins are doing illegally.

kienstra:
**In Development**

Hi @westonruter,
Thanks for your details here. If it's alright, I'll post an update soon.

westonruter:
Let's chat about the approach to take here.

kienstra:
**Possible Approach**

Hi @westonruter,
Thanks for waiting here. One approach that I've prototyped is front-end validation in Gutenberg ([screencast](https://cloudup.com/cPYRW-m2Ysf)).

Unfortunately, this only applies on editing a block. Not on saving it, as you mentioned. 

Here's a very rough prototype that you could past in the console on a Gutenberg page.

```
var isValidAMP,
	previousWp = wp,
	el = wp.element.createElement,
	Notice = wp.components.Notice,
	blockType = 'core/html',
	htmlBlock = wp.blocks.unregisterBlockType( blockType ),
	OriginalBlockEdit = htmlBlock.edit;

/**
[...]

Issue #845, opened 2018-01-04
Conduct a Discovery to determine support for Gutenberg blocks.


MackenzieHartung:
### Acceptance Criteria

AC1: Create a wp-cli script to output all native Gutenberg blocks on a test page to identify which will need to be supported (see this [script](https://github.com/Automattic/amp-wp/blob/develop/bin/create-embed-test-post.php) for reference).

kienstra:
**Example wp-cli Scripts**

 Hi @DavidCramer,
It looks like you wrote the `wp-cli` script [create-comments-on-test-post.php](https://github.com/Automattic/amp-wp/blob/develop/bin/create-comments-on-test-post.php), so you probably have a good handle on this. This issue's script for Gutenberg blocks should probably be similar to [create-embed-test-post.php](https://github.com/Automattic/amp-wp/blob/develop/bin/create-embed-test-post.php).

kienstra:
**Permutations Of Blocks**

We should probably output all of the blocks in [blocks/library/](https://github.com/WordPress/gutenberg/tree/8b4a61df4b2b1b57c99cb26a8b7f6978ae2f071f/blocks/library). Here are some blocks that I think could use permutations:

* [Categories](https://github.com/WordPress/gutenberg/tree/8b4a61df4b2b1b57c99cb26a8b7f6978ae2f071f/blocks/library/categories) 
1. With dropdown
2. Without dropdown

* [Columns](https://github.com/WordPress/gutenberg/tree/8b4a61df4b2b1b57c99cb26a8b7f6978ae2f071f/blocks/library/columns)
2 and 4 columns

* [Cover Image](https://github.com/WordPress/gutenberg/tree/8b4a61df4b2b1b57c99cb26a8b7f6978ae2f071f/blocks/library/cover-image) 
1. Fixed Background on
2. Fixed Background off

* [Embed](https://github.com/WordPress/gutenberg/tree/8b4a61df4b2b1b57c99cb26a8b7f6978ae2f071f/blocks/library/embed)
2 or 3 embed URLs should be enough. #841 handles embed support, though not in Gutenberg.

* [Gallery](https://github.com/WordPress/gutenberg/tree/8b4a61df4b2b1b57c99cb26a8b7f6978ae2f071f/blocks/library/gallery)
2 and 3 columns

[...]

PR #902, opened 2018-01-25
[WIP] Issue #843: Validate Gutenberg blocks for AMP compliance

kienstra:
**WIP Pull Request**

Hi @westonruter and @ThierryA,
Here's the beginning of the script to validate Gutenberg blocks. The script is similar to the one [pasted in Issue 843](https://github.com/Automattic/amp-wp/issues/843#issuecomment-360251366). But it applies AMP validation to every block.

As @westonruter suggested, it gets the content by calling the block's `save()` function. As this is what goes to the server.

We might consider not displaying notices for all blocks. Some are valid AMP once the back-end sanitizer alters them. Like the 'Audio' block.

This still works mainly like in the [earlier screencast](https://cloudup.com/cPYRW-m2Ysf), just for all of the blocks. Though I need to test this on more of them.

I still need to handle [Travis issues](https://travis-ci.org/Automattic/amp-wp/jobs/333145167).

For full-page validation on saving, especially in the "classic" editor, we might do something like [this snippet](https://github.com/ampproject/amphtml/issues/9371#issuecomment-306603777).

See #843.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -89,6 +89,7 @@ function amp_after_setup_theme() {
 
 	add_action( 'init', 'amp_init' );
 	add_action( 'widgets_init', 'AMP_Theme_Support::register_widgets' );
+	add_action( 'enqueue_block_editor_assets', 'AMP_Theme_Support::enqueue_gutenberg' );
 	add_action( 'admin_init', 'AMP_Options_Manager::register_settings' );
 	add_filter( 'amp_post_template_analytics', 'amp_add_custom_analytics' );
 	add_action( 'wp_loaded', 'amp_post_meta_box' );
--- a/amp.php
+++ b/amp.php
@@ -89,7 +89,7 @@ function amp_after_setup_theme() {
 
 	add_action( 'init', 'amp_init' );
 	add_action( 'widgets_init', 'AMP_Theme_Support::register_widgets' );
-	add_action( 'enqueue_block_editor_assets', 'AMP_Theme_Support::enqueue_gutenberg' );
+	add_action( 'admin_enqueue_scripts', 'AMP_Theme_Support::enqueue_editor' );
 	add_action( 'admin_init', 'AMP_Options_Manager::register_settings' );
 	add_filter( 'amp_post_template_analytics', 'amp_add_custom_analytics' );
 	add_action( 'wp_loaded', 'amp_post_meta_box' );
--- a/assets/js/amp-editor-validation.js
+++ b/assets/js/amp-editor-validation.js
@@ -8,7 +8,7 @@
  */
 
 /* exported ampGutenberg */
-var ampGutenberg = ( function( $ ) {
+var ampEditorValidation = ( function( $ ) {
 	'use strict';
 
 	var component = {
@@ -16,6 +16,11 @@ var ampGutenberg = ( function( $ ) {
 		 * Holds data.
 		 */
 		data: {},
+
+		/**
+		 * The response from the amp.validator for valid AMP.
+		 */
+		validStatus: 'PASS',
 
 		/**
 		 * Boot module.
@@ -39,17 +44,18 @@ var ampGutenberg = ( function( $ ) {
 		 * @returns void.
 		 */
 		getScript: function() {
-			var previousWp = wp;
 			$.getScript( 'https://cdn.ampproject.org/v0/validator.js', function() {
-				wp = previousWp;
-				component.processBlocks();
+				component.validatePage();
+				if ( 'undefined' !== typeof wp.blocks ) {
+					component.processBlocks();
+				}
 			} );
 		},
 
 		/**
 		 * Gets all of the registered blocks, and overwrites their edit() functions.
 		 *
-		 * THe new edit() functions will check if the content is AMP-compliant.
+		 * The new edit() functions will check if the content is AMP-compliant.
 		 * If not, the block will display a notice.
 		 *
 		 * @returns {void}
@@ -90,7 +96,7 @@ var ampGutenberg = ( function( $ ) {
 				}
 
 				// If validation fails, prepend a Notice to the block.
-				if ( ! component.isValidAMP( content ) ) {
+				if ( ! component.isValidAMP( content, true ) ) {
 					result.unshift( el(
 						Notice,
 						{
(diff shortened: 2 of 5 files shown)


westonruter:
As we talked about on Slack…

The AMP validator running client-side on the Gutenberg blocks is running on the pre-sanitized output. So if you add an image then it's going to marked as invalid because it is not an `amp-img`. Now, one possibility would be to try modifying all of the blocks to use AMP components instead of HTML, but that seems not scalable or maintainable, at all. So… perhaps what is needed is an endpoint where we can POST some HTML to then have it run through the sanitization logic and then get back the result. This would mean that the block validation wouldn't be _live_, but it would mean that it wouldn't be reporting errors for things that are going to be sanitized on the frontend.

This will depend on #912.

pbakaus:
I actually have different opinion about this. Within the AMP team, we are fairly confident in our belief that the sanitizers/converters of this world will never be a perfect solution. The exception would be if they happen at a very granular level, but even then I'm curious how you'd 1:1 translate something like the layout attribute in `<amp-img>`.

So while we see the need for the sanitizers to exist, we'd like to strive for a world where they are not needed, and a CMS like WP can generate native AMP. As example, I'd much rather patch the_thumbnail to produce valid AMP, than run the final result through a sanitizer.

For Gutenberg, it is a little more tricky due to third party Gutenblocks. But I'd love to investigate if we can have at least the native, bundled blocks generate valid AMP from the get go. Please feel free to tell me if I'm delusional :)

westonruter:
@pbakaus Thanks for the input. I agree that the ideal is that the CMS itself would be outputting AMP components from the start. And it's true that sanitizers won't be a perfect solution, and users would be able to have more tailored controls provided for them if AMP components were the target. But this really would require that the CMS only ever be supporting AMP period.

For the featured image (`the_post_thumbnail()`) this is something that could be made to output an `amp-img` tag if `current_theme_supports( 'amp' )`. That would be straightforward, as this is dynamically-generated. It is more complicated with content that is stored in the DB. For content created with the classic editor and Gutenberg alike, the images get written into `post_content` as `` tags. Now, we could wrap the `Image` block's `save` function to replace [using](https://github.com/WordPress/gutenberg/blob/f2bbf5f48aefda63e00b40ea8362a10934dacca9/blocks/library/image/index.js#L167) `img` with `amp-img`:

```diff
- const image = <img src={ url } alt={ alt } { ...extraImageProps } />;
+ const image = <amp-img src={ url } alt={ alt } { ...extraImageProps } />;
```

However, this would then mean the AMP tag would be persistently stored in the DB. If the user deactivates the AMP plugin then all of their images will be broken. The problem is worse if the site operates in a paired mode, with AMP and non-AMP versions of content: the post rendered in non-AMP would be broken. The problem is somewhat mitigated if the site is AMP Canonical, which is what we're working on now, but this requires opt-in theme support. If the user switches to a theme that doesn't support `amp` then it would be the same as if the AMP plugin were disabled, and the images would be broken.

The only way I can think of to address this issue long term is for Gutenberg blocks to rely more heavily on server-side rendering. There are two kinds of blocks in Gutenbeg: static and dynamic. The vast majority of the blocks are static, in that the block's attributes are largely serialized to HTML   which then allows them to be portable and resilient to not being lost when a plugin that introduced the block is deactivated: it's essentially built-in to have a fallback rendering. Dynamic blocks, on the other hand, have no serialized HTML that is saved: everything is generated by PHP on the frontend. The Latest Posts block is the best example of this. A concern with such blocks is that the editing interface still renders using JS via React whereas there is a _separate_ PHP-based rendering function on the frontend. This results in a duplication of rendering systems. Perhaps this is acceptable for the core blocks where the user experience warrants having duplicated rendering logic, but for most developers I can't imaging they'd want to implement a shortcode-like PHP renderer in addition to a React-based renderer. They'd want to just do the PHP one alone.

We could implement a PHP renderer for the `Image` block. This could be the granular-level conversion you are describing. Additional attributes for managing `layout` could be added as block attributes and then used in the dynamic PHP rendering of the block, which would parse out the `` and convert it into `<amp-img>` with the additional layout block attributes added to it at runtime. I think this would work, but it would involve having to start maintaining PHP-based renderers for all of the relevant core blocks that need AMP-specific versions. This would involve a maintenance overhead, but if limited to the core blocks then it should be ok.

For 3rd-party blocks that don't have dynamic renderers, however, we'll need the sanitizer. So I think that we should take these two separate routes in parallel, but always falling back to sanitization to enforce compliance.

westonruter:
Here is a proof of concept for what such a dynamic granular handling of the Image block could look like from the PHP side of things. There'd probably need to be some layout layout options added to the block's `edit` function as well.

```php
add_action( 'init', function() {
	if ( ! function_exists( 'register_block_type' ) ) {
		return;
	}

	register_block_type( 'core/image', array(
		'render_callback' => function( $attributes, $raw_content ) {

			// Return raw content if not serving AMP.
			if ( ! function_exists( 'is_amp_endpoint' ) || ! is_amp_endpoint() ) {
				return $raw_content;
			}

			// Defer to sanitizer if image is external.
			if ( empty( $attributes['id'] ) || 0 !== strpos( get_post_mime_type( $attributes['id'] ), 'image/' ) ) {
				return $raw_content;
			}

			$html_attr = '';

			list( $src, $width, $height ) = wp_get_attachment_image_src ( $attributes['id'], 'large' );
			if ( false === strpos( 'width="', $raw_content ) ) {
				$html_attr .= sprintf( ' width="%d"', $width );
			}
			if ( false === strpos( 'height="', $raw_content ) ) {
				$html_attr .= sprintf( ' height="%d"', $height );
			}

			return preg_replace(
				'#<img\s([^>]*?)/?>#',
				'<amp-img $1 ' . $html_attr . '></amp-img>',
				$raw_content
			);
		},
	) );
} );
```

westonruter:
I should also mention that Gutenberg may eventually include built-in PHP implementation of content selector used in JS to parse out the `img` attributes. This would avoid having to use regular expressions as seen here.

westonruter:
As an alternative to implementing custom block renderers like above in https://github.com/Automattic/amp-wp/pull/902#issuecomment-361044978 perhaps a better approach would be add AMP processing instructions to the HTML which the pre-processor (sanitizer) can use to convert the HTML into AMP as required. For example, if the `core/image` block in Gutenberg added a `layout` control with options for `fixed`, `responsive`, `fill`, etc. Then the `layout` property could get serialized into the HTML as a data attribute, e.g. `<img data-amp-layout="responsive"...`. This would allow for the markup to be portable between AMP and non-AMP 

We can shift from thinking of the whitelist sanitizer just about sanitization but rather now more about pre-processing/transformation. With additional attributes placed on elements this can then inform the pre-processor to output exactly what should be output, rather than having to make a best guess for the lowest-common denominator.

pbakaus:
Thanks for the thoughtful response @westonruter, and apologies for the delay.

I was initially under the impression that most of Gutenberg would help separate concerns in WordPress, so to eventually not have any HTML in the database anymore, but sad to hear that's not the case. The fallback reasoning is a fair one though.

Is there a way to store to alternate copies of a block in the database, where if you've enabled AMP, it saves the generated `<amp-img>` but also stores the `` as fallback, for if you disable AMP, for instance?

If not, then I agree that the best path might be to map existing and new classes that change the layout of an image in WordPress to their AMP counterparts when parsed in the sanitizer. That also means that in the image block, we need to introduce the layouting capabilities of AMP for the non-AMP world, so that the user experience is consistent. It would be weird if we did the data attribute thing to say, for instance, that an image when rendered in AMP should fill to the width and height of its parent, but the rule would be ignored for non-AMP.

westonruter:
@pbakaus:

> Is there a way to store to alternate copies of a block in the database, where if you've enabled AMP, it saves the generated <amp-img> but also stores the  as fallback, for if you disable AMP, for instance?

There is not really an existing mechanism to store different variations of a block. It could perhaps be shimmed in to store different versions in meta, but then there's perhaps the worse problem of duplicate content.

> If not, then I agree that the best path might be to map existing and new classes that change the layout of an image in WordPress to their AMP counterparts when parsed in the sanitizer. That also means that in the image block, we need to introduce the layouting capabilities of AMP for the non-AMP world, so that the user experience is consistent. It would be weird if we did the data attribute thing to say, for instance, that an image when rendered in AMP should fill to the width and height of its parent, but the rule would be ignored for non-AMP.

Actually, it's pretty standard for some things in WordPress to only work in the context of a given theme being active. For example, Gutenberg is adding the ability for images to be full-bleed as one of the layout options in the block UI. In order for this option to be shown, however, a theme has to explicitly `add_theme_support( 'align-wide' )`. Similarly, Gutenberg allows for for blocks to provide different color choices for a block, but a theme can restrict which colors are available via `add_theme_support( 'editor-color-palette', …)`. 

If a user switches to a theme that doesn't have support for these then the options are not available. I think this actually would work perfectly for `layout` in AMP. We can add the `layout` option when the theme (or a plugin) has done `add_theme_support('amp')`. The `layout` would be ignored if switching themes, in the same way that WordPress displays fallback content for blocks when the block definition is no longer registered.

pbakaus:
@westonruter ok got it, thanks for the clarification. I mostly talked about the case where, say, you're publishing both non-AMP and AMP pages (e.g. through paired mode) on your Wordpress instance with one theme. In this case, I think it would be majorly confusing if there would be a lot of options in Gutenberg's image block, some of which would only apply to the AMP-rendered version. As long as the experience is consistent *within* a theme, I'm a happy camper :)

westonruter:
@pbakaus Yeah, that's a good point. This really presumes a native AMP site, without a separate non-AMP version, at least for the singular template. Otherwise we'd either have to mark the AMP-only settings, or re-implement the features available in AMP for non-AMP responses. That being said, what if we just embraced “dirty AMP” for serving out AMP components in non-AMP documents?

pbakaus:
@westonruter not a bad thought in general, but risky, as the AMP v0 library assumes it's the only DOM manipulator on the page, and thus has a pretty simplistic world view where it doesn't observe mutations, scroll etc. So dirty AMP only works if there's no JS on the page that manipulates the DOM.

ThierryA:
AMP offers a lot of great components which we can leverage not only in the context of image layout more also to build custom blocks such as carousel for example. That would obviously not work for paired mode unless we build a non AMP fallback or embrasse the risk of dirty mode (which isn't bulletproof at the moment).
I would personally not be in favor of adding fallbacks as it is unecessarly re-inventing the wheel and will be a nightmare to maintain. So I would vote to either make AMP component work in "dirty AMP" in a bullet proof way **if possible** (to investigate further) or make native AMP a requirement in order to benefit from all the "advanced" features.

westonruter:
The discovery that has come out of this PR has been really insightful and important going forward. However, I think that the specific approach taken in this PR for validation (loading the `validator.js` from the AMP project for client-side validation) should give way to the validation of server-side rendering which is now enabled with #971 powered by the plugin itself. In short, with that PR after each save we run the the post through the preprocessor to validate the content for errors. It currently has support for detecting validation errors introduced by shortcodes, and the same approach can then also be applied to blocks. Then it is a matter of taking the validation results and displaying them meaningfully with their corresponding blocks.

As I see them, the next steps for Gutenberg support are:

1. Complete discovery of Gutenberg blocks and their base compatibility (whether anything gets removed through validation) with the AMP plugin out of the box. See #845.
2. Integrate block validation results with the UI to reveal when a validation error does occur. Normally such errors will not happen for core blocks, but for the Custom HTML block in particular it is a distinct possibility. It is also key for plugins which introduce blocks so that their AMP validity can also be reported. 
3. Extend core blocks which have validation errors so that the appropriate AMP markup is output to faithfully render the block when the page is served as AMP. This would be done via some combination of the `render_callback` that integrates with a new block sanitizer/preprocessor.
4. Further extend core blocks to leverage AMP functionality, such as `layout`.
5. Introduce blocks that feature AMP-specific components (e.g. `amp-ad` and even `amp-story` in #968) which are served in both AMP and non-AMP documents as Dirty AMP (if possible).

That's what I see as the next steps on the roadmap for Gutenberg/AMP integration.

PR #904, opened 2018-01-25
Avoid potential issues with get_the_date filters

kartben:
Using get_post_time enables to make sure that the publish date cannot be altered by any potential filter associated with get_the_date function.


westonruter:
What if they _want_ the published date to be changed? Also, the `get_post_time()` function likewise has a `get_post_time` filter, just as `get_the_date()` has `get_the_date`. So they'd still be able to filter it. If filtering is truly not to be allowed, them something like `mysql2date( 'U', $post->post_date )` would be needed.

PR #906, opened 2018-01-26
Prevent saveHTML from munging amp-mustache curly braces via URL-encoding

westonruter:
Cherry-picked from #871. URL-encoding of braces in `href` attributes is a side effect of switching from `saveXML` to `saveHTML` in #891.

### DIFF
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -60,6 +60,22 @@ class AMP_DOM_Utils {
 		$document = self::convert_amp_bind_attributes( $document );
 
 		/*
+		 * Prevent amp-mustache syntax from getting URL-encoded in attributes when saveHTML is done.
+		 * While this is applying to the entire document, it only really matters inside of <template>
+		 * elements, since URL-encoding of curly braces in href attributes would not normally matter.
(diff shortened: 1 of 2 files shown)


westonruter:
(Yes, this is yet something else that could be part of `AMP_DOMDocument`, but let's not go there quite yet.)

includes/utils/class-amp-dom-utils.php
```diff
@@ -285,6 +333,14 @@ public static function get_content_from_dom_node( $dom, $node ) {
 
 		$html = self::restore_amp_bind_attributes( $html );
 
+		// Restore amp-mustache placeholders which were replaced to prevent URL-encoded corruption by saveHTML.
```
ThierryA: Perhaps something like the prototype replacement below would have been sufficient:

``` PHP
$html = preg_replace_callback(
	"#href=[\"|'](%7B.*%7D)?[\"|']+#",
	function( $matches ) {
		return urldecode( $matches[0] );
	},
	$html
);
```

westonruter: I didn't want to do replacements on encodings that were original. And I wanted to limit the replacements to braces.

PR #909, opened 2018-01-26
Add amp-live-list comments

### LINKED ISSUE #797: Add support for comments

### LINKED ISSUE #862: Support submitting comments to WordPress

DavidCramer:
- [x] Decide on whether to enforce `div` container instead of `ol`. Ideal is `ol`.
- [x] Figure out how to reset `form` after submission.
- [ ] Triggering an `update` does not force a request to update. In short term, just show message saying it will be available soon.
- [x] Page without comments does not output an `amp-live-list` to start with.
- [x] Admin notice on the discussion settings screen to indicate that comments in ascending order is not supported yet.

Fixes #862.
Fixes #797.


westonruter:
As discussed, to fully support comments in the AMP plugin for WordPress, the `amp-live-list` component really needs to support the sort order, but it is not yet: https://github.com/ampproject/amphtml/issues/5396

The ordering of WP comments is configurable in the admin:

image

For the immediate term we'll be forcing the option to be “newer” and display a notice:

image

So when `amp` theme support is present, we'll need to add a filter for `option_comment_order` to have it return `'desc'`. And then some JS will need to be injected onto the Discussion admin screen to disable the `select#comment_order` element and then to inject the following HTML at the end of its containing fieldset:

```php
<div class="notice notice-info inline"><p><?php echo wp_kses_post( __( 'Note: AMP does not yet <a href="https://github.com/ampproject/amphtml/issues/5396" target="_blank">support ascending</a> comments with newer entries appearing at the bottom.', 'amp' ) ); ?></p></div>
```

amp.php
```diff
@@ -361,3 +363,52 @@ function amp_redirect_old_slug_to_new_url( $link ) {
...
+	return 'amp_send_error_json';
+}
+
+/**
+ * Send error creating message.
+ *
+ * @param string $message The error message to send.
+ */
+function amp_send_error_json( $message ) {
+	header( 'HTTP/1.1 400 BAD REQUEST' );
```
westonruter: Use `status_header( 400 )` instead.

includes/class-amp-comment-walker.php
```diff
@@ -0,0 +1,69 @@
...
+	 * @see Walker::end_el()
+	 * @see wp_list_comments()
+	 *
+	 * @param string     $output  Used to append additional content. Passed by reference.
+	 * @param WP_Comment $comment The current comment object. Default current comment.
+	 * @param int        $depth   Optional. Depth of the current comment. Default 0.
+	 * @param array      $args    Optional. An array of arguments. Default empty array.
+	 */
+	public function end_el( &$output, $comment, $depth = 0, $args = array() ) {
+		$output .= '<amp-comment data-sort-time="' . strtotime( $comment->comment_date ) . '"  />';
```
westonruter: I don't think an `amp-comment` element should be introduced here. Rather, the `data-sort-time` attribute should just be added to the existing element. So instead of overriding the `end_el` method, you should instead extend the protected `comment` method and do something like:

```php
protected function comment( $comment, $depth, $args ) {
    ob_start();
    parent::comment( $comment, $depth, $args );
    $html = ltrim( ob_get_clean() );
    $html = preg_replace( '/^(<\w+)(?=\s)/', '$1 data-sort-time="' . strtotime( $comment->comment_date ) . '" );
    echo $html;
}
```

By doing this, there may not be a need for `AMP_Comments_Sanitizer`, except for adding the `on` attribute to the `form#commentform` element.

includes/sanitizers/class-amp-comments-sanitizer.php
```diff
@@ -0,0 +1,71 @@
...
+					if ( $child instanceof DOMElement ) {
+						$time = $child->getElementsByTagName( 'amp-comment' )->item( 0 )->getAttributeNode( 'data-sort-time' );
+						$child->setAttributeNode( $time );
+					}
+					$wrapper->appendChild( $child );
+				}
+			}
+			$node->replaceChild( $wrapper, $node->firstChild );
+			$node->parentNode->parentNode->replaceChild( $node, $node->parentNode );
+		}
```
westonruter: Per above, I don't think this `for` loop will be necessary.

includes/class-amp-theme-support.php
```diff
@@ -275,6 +279,44 @@ public static function register_content_embed_handlers() {
...
+		// @todo Make threaded work by setting 'data-update-time' of the parent first level parent with the last timestamp.
+		$args['max_depth'] = 1;
+
+		return $args;
+	}
+
+	/**
+	 * Adds the form submit success and fail templates.
+	 *
+	 * @param string $post_id The post ID if action is 'comment_form'.
```
westonruter: Instead of overloading the `add_amp_comment_form_templates` method, let's have two separate methods for before and after.

includes/class-amp-theme-support.php
```diff
@@ -275,6 +279,44 @@ public static function register_content_embed_handlers() {
...
+			$output .= '<div id="amp-comment-form-fields">';
+		} else {
+			$output .= '</div>';
+			$output .= '<div submit-success><template type="amp-mustache">';
+			$output .= esc_html__( 'Your comment has been posted.', 'amp' );
+			$output .= sprintf( '<a class="amp-view-new-comment" href="{{comment_link}}">%s</a>', esc_html__( 'View Comment', 'amp' ) );
+			$output .= '</div></template>';
+			$output .= '<div submit-error><template type="amp-mustache">';
+			$output .= '<p class="amp-comment-submit-error">{{{error}}}</p>';
+			$output .= '</div></template>';
```
westonruter: Instead of creating a string of HTML and then echoing it, how about just `?><div>...</div><?php`?

includes/class-amp-comment-walker.php
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		$args = array_slice( func_get_args(), 4 );
+
+		$output  = '<amp-live-list layout="container" data-poll-interval="15000" data-max-items-per-page="20" id="amp-live-comments-list">';
+		$output .= '<div items>';
+		$output .= parent::paged_walk( $elements, $max_depth, $page_num, $per_page, $args[0] );
+		$output .= '</div>';
+		$output .= '<button update on="tap:amp-live-comments-list.update" class="ampstart-btn ml1 caps">' . esc_html__( 'You have updates', 'amp' ) . '</button>';
+		$output .= '<div pagination></div>';
```
westonruter: Missing pagination links.

includes/class-amp-comment-walker.php
```diff
@@ -0,0 +1,69 @@
...
+	 * @return string XHTML of the specified page of elements.
+	 */
+	public function paged_walk( $elements, $max_depth, $page_num, $per_page ) {
+		if ( empty( $elements ) || $max_depth < - 1 ) {
+			return '';
+		}
+
+		$args = array_slice( func_get_args(), 4 );
+
+		$output  = '<amp-live-list layout="container" data-poll-interval="15000" data-max-items-per-page="20" id="amp-live-comments-list">';
```
westonruter: The `data-max-items-per-page` attribute should get pulled from the `comments_per_page`?

westonruter: The `id` should probably be unique. It's possible that more than one comment loop could be displayed on a page. So there should perhaps be a new `static` member variable that increments as the method is called.

includes/class-amp-comment-walker.php
```diff
@@ -0,0 +1,69 @@
...
+	 * @see Walker::paged_walk()
+	 * @param array $elements List of comment Elements.
+	 * @param int   $max_depth The maximum hierarchical depth.
+	 * @param int   $page_num The specific page number, beginning with 1.
+	 * @param int   $per_page Per page counter.
+	 *
+	 * @return string XHTML of the specified page of elements.
+	 */
+	public function paged_walk( $elements, $max_depth, $page_num, $per_page ) {
+		if ( empty( $elements ) || $max_depth < - 1 ) {
```
westonruter: Trivial: Remove space between `-` and `1`

includes/class-amp-theme-support.php
```diff
@@ -275,6 +279,44 @@ public static function register_content_embed_handlers() {
...
+	 * Add the comments template placeholder marker
+	 *
+	 * @param array $args the args for the comments list..
+	 * @return array Args to return.
+	 */
+	public static function amp_set_comments_walker( $args ) {
+		$amp_walker     = new AMP_Comment_Walker();
+		$args['walker'] = $amp_walker;
+		// @todo Make threaded work by setting 'data-update-time' of the parent first level parent with the last timestamp.
+		$args['max_depth'] = 1;
```
westonruter: Yeah, we need to be able to remove this constraint.

amp.php
```diff
@@ -361,3 +363,52 @@ function amp_redirect_old_slug_to_new_url( $link ) {
 
 	return $link;
 }
+
+/**
+ * Hook into a comment submission if an AMP xhr post request.
+ */
+function amp_hook_comments_maybe() {
```
westonruter: Rename this to `amp_handle_comment_post`.

amp.php
```diff
@@ -361,3 +363,52 @@ function amp_redirect_old_slug_to_new_url( $link ) {
...
+
+	$method        = filter_input( INPUT_SERVER, 'REQUEST_METHOD', FILTER_SANITIZE_STRING );
+	$is_amp_submit = filter_input( INPUT_GET, '__amp_source_origin', FILTER_VALIDATE_URL );
```
westonruter: The variable name here implies it is a boolean. Let's change it to `$amp_source_origin`

amp.php
```diff
@@ -125,6 +125,8 @@ function amp_init() {
 	if ( class_exists( 'Jetpack' ) && ! ( defined( 'IS_WPCOM' ) && IS_WPCOM ) ) {
 		require_once( AMP__DIR__ . '/jetpack-helper.php' );
 	}
+
+	amp_hook_comments_maybe();
```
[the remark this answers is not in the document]
DavidCramer: Nice one! will do.

amp.php
```diff
@@ -361,3 +363,52 @@ function amp_redirect_old_slug_to_new_url( $link ) {
...
+	$method        = filter_input( INPUT_SERVER, 'REQUEST_METHOD', FILTER_SANITIZE_STRING );
+	$is_amp_submit = filter_input( INPUT_GET, '__amp_source_origin', FILTER_VALIDATE_URL );
+
+	if ( 'post' !== strtolower( $method ) || is_null( $is_amp_submit ) ) {
+		return;
+	}
+
+	// Add amp comment hooks.
+	add_filter( 'comment_post_redirect', 'amp_comment_post_handler', PHP_INT_MAX, 2 );
+	add_filter( 'wp_die_handler', 'amp_set_comment_submit_die_handler' );
```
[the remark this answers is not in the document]
DavidCramer: awesome! I forgot that the min version was increased. great news!

amp.php
```diff
@@ -361,3 +363,52 @@ function amp_redirect_old_slug_to_new_url( $link ) {
...
+}
+
+/**
+ * Handle the comment return.
+ *
+ * @param string     $location The location to redirect to.
+ * @param WP_Comment $comment The comment that was posted.
+ */
+function amp_comment_post_handler( $location, $comment ) {
+	$comment->comment_link = get_comment_link( $comment->comment_ID );
```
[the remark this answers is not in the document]
DavidCramer: Strange that it's not complaining for me. will check to see if the inspections are set up for that.

includes/class-amp-comment-walker.php
DavidCramer: In that instance, we could do the same string replace for the comment form attribute.
I was hesitant to extend the comment function since there was a lot of code in it and it worries me overriding such large methods, but your solution there is elegant, so will do that. Thanks!

includes/class-amp-comment-walker.php
DavidCramer: 😐  ... 🤦‍♂️

DavidCramer: Good point! will address it.

includes/class-amp-comment-walker.php
DavidCramer: Yup, still learning on how that works. Will add it in when I have more information on it.

includes/class-amp-theme-support.php
DavidCramer: Ye, this is temporary, I figured out how to do it. will experiment and make it work!

includes/class-amp-theme-support.php
DavidCramer: The idea was to have it pass through a filter before outputting it. Would that be a good idea?

DavidCramer:
@westonruter Thanks for the thorough review. Good suggestions made. Will work on these and update promptly. Thanks!

includes/class-amp-theme-support.php
westonruter: I don't think we need to add a filter here? In any case, output buffering could be used here to capture the string. On the other hand, the HTML could also get loaded from an external template file which could potentially be overridden by a theme. I'd have to refresh my memory on how comment forms are normally extended.

includes/class-amp-comment-walker.php
DavidCramer: Was thinking the same thing. Will experiment and on it.

includes/class-amp-comment-walker.php
DavidCramer: @westonruter since pagination isn't exactly handled by amp-live-list, should we just omit it, and use pagination as provided by the theme?

westonruter: I think you may want to just put a call to `get_the_comments_navigation()` inside `<div pagination></div>`. It's important that the `update`, `items`, and `pagination` elements all be siblings:

> The amp-live-list component has 3 sections. We'll refer to these sections as "reference points" and they are denoted by an attribute. These reference points _must be a direct child of the amp-live-list component_.

DavidCramer: cool, will do! thanks.

amp.php
```diff
@@ -361,3 +363,52 @@ function amp_redirect_old_slug_to_new_url( $link ) {
...
+}
+
+/**
+ * Send error creating message.
+ *
+ * @param string $message The error message to send.
+ */
+function amp_send_error_json( $message ) {
+	header( 'HTTP/1.1 400 BAD REQUEST' );
+	wp_send_json( array( 'error' => strip_tags( $message, 'strong' ) ) );
```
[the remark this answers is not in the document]
DavidCramer: @westonruter I used striptags because `amp-mustache` takes the text from element so it doesn't go deeper. meaning that anything within child elements are totally removed. So `<p><strong>ERROR</strong>:Comment is empty.</p>` becomes `:Comment is empty`

assets/js/amp-admin-comments.js
```diff
@@ -0,0 +1,16 @@
...
+	$( document ).on( 'ready', function() {
+		var orderSelect = $( '#comment_order' ),
+			notice = $( '#amp-comment-notice' );
+		if ( orderSelect.length && notice.length ) {
+			orderSelect.prop( 'disabled', true );
+			notice.appendTo( orderSelect.parent() );
+		}
+	} );
+
+})( jQuery );
```
DavidCramer: @westonruter I think this could easily have been a simple inline script alongside the notice, but I made it a asset file for now, just in case.

DavidCramer:
@westonruter You can review this again now I think I have addressed all issues as well as made changes necessary for https://github.com/xwp/ampconf/pull/52

amp.php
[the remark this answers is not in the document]
westonruter: Great to know. Thanks for sharing.

assets/js/amp-admin-comments.js
westonruter: Thanks. I've decided to go ahead and inline it. See 310a74ade4777089816ef0136a93bc74da3b1b54

DavidCramer: Perfect. thought that would be the case. thanks for sorting it for me.

includes/class-amp-theme-support.php
```diff
@@ -532,6 +533,93 @@ public static function add_canonical_link() {
...
+	 *
+	 * @since 0.7
+	 * @see get_cancel_comment_reply_link()
+	 *
+	 * @param string $formatted_link The HTML-formatted cancel comment reply link.
+	 * @param string $link           Cancel comment reply link URL.
+	 * @param string $text           Cancel comment reply link text.
+	 * @return string Cancel reply link.
+	 */
+	public function filter_cancel_comment_reply_link( $formatted_link, $link, $text ) {
```
DavidCramer: @westonruter should be a static function.

PR #910, opened 2018-01-26
AMP Rewrite endpoint enables infinite loops on relative page links.

mdbitz:
This update ensures that 404 Page Not Found is triggered instead of returning the amp article for urls that contain content after the amp endpoint. /amp/XXXX/XXXX

For example if you have a post with the permalink https://my.site/12/2017/post-link and the content contains a link for /contact.

https://my.site/12/2017/post-link/amp will have a relative link of https://my.site/12/2017/post-link/amp/contact

This will return the amp page with a relative link of https://my.site/12/2017/post-link/amp/contact/contact


mdbitz:
HI Weston:

I'm re-opening this as it can lead to massive resource usage from googlebot infinitely indexing the site. WordPress allows users to enter any link they want into content and it typically isn't validated to be a correct link. As AMP rewrite rules work is that it captures any url of the format AMP/XXXXXX.

When any page contains a link that is relative `href="blue"` `href="www.google.com"` or etc when the page is visited it re-renders the current page just with links now a level deeper.

For example create a page that has a single relative link `<a href="red">RED</a>`. If you access the amp version of the page you will see that the page you are directed to is: /amp/red and it renders the current page but now if you click the link it goes to amp/red/red and so forth. When google bot indexes the site it tries to follow all these pages and gets in an indexing loop.

The issue is even more pronounced when there are more relative links on a page.

amp.php
```diff
@@ -133,8 +133,28 @@ function amp_init() {
...
+
+			// If there is something after /amp/, pretend we didn't hit that rewrite rule.
+			if ( $wp_rewrite->use_trailing_slashes && '' !== $query_vars[ AMP_QUERY_VAR ] ) {
+
+				// Check for extra-long slugs, which will truncate and remove the /amp/* part.
+				if ( strlen( $query_vars['name'] ) > 190 ) {
+					// Give WP_Query an improbable string to get a 404.
+					$query_vars['name'] = substr( $query_vars['name'], 0, - 20 ) . substr( md5( $query_vars['name'] ), 0, 20 );
+				} else {
+					$query_vars['name'] = $query_vars['name'] . "/amp/" . $query_vars[ AMP_QUERY_VAR ];
```
westonruter: Note that `"/amp/"` here doesn't account for `AMP_QUERY_VAR` being set to something other than `amp`.

amp.php
```diff
@@ -133,8 +133,28 @@ function amp_init() {
 // Make sure the `amp` query var has an explicit value.
 // Avoids issues when filtering the deprecated `query_string` hook.
 function amp_force_query_var_value( $query_vars ) {
-	if ( isset( $query_vars[ amp_get_slug() ] ) && '' === $query_vars[ amp_get_slug() ] ) {
-		$query_vars[ amp_get_slug() ] = 1;
+
+	if ( isset( $query_vars[ AMP_QUERY_VAR ] ) ) {
```
westonruter: Note that `amp_get_slug()` should be used now instead of `AMP_QUERY_VAR`.

westonruter:
This also needs unit tests.

postphotos:
This is an interesting bug. It would be great if it could get fixed!

westonruter:
I propose something like this instead: https://gist.github.com/westonruter/20d8ef0cea36bcd27bdcd8d011570a58

westonruter:
Here's what I think a better approach is: do a 301 redirect to the AMP permalink for the current post. This will have the effect of stripping off the extra `amp` endpoint matter.

```diff
diff --git a/amp.php b/amp.php
index 88d3ab98..9839728c 100644
--- a/amp.php
+++ b/amp.php
@@ -317,6 +317,14 @@ function amp_maybe_add_actions() {
 	}
 
 	if ( $is_amp_endpoint ) {
+
+		// Prevent infinite URL space under /amp/ endpoint.
+		$endpoint = get_query_var( amp_get_slug(), false );
+		if ( ! isset( $_GET[ amp_get_slug() ] ) && false !== $endpoint && '1' !== (string) $endpoint && '' !== $endpoint ) {
+			wp_safe_redirect( amp_get_permalink( $post->ID ), 301 );
+			exit;
+		}
+
 		amp_prepare_render();
 	} else {
 		amp_add_frontend_actions();
```

westonruter:
@mdbitz @scottsweb Heres a PR which iterates on my above suggestion: #1846.

Please check whether that indeed fixes the issue as you observed.

westonruter:
Closing in favor of #1846.

Issue #842, opened 2018-01-04
Display error when a plugin is activated which tries to do something incompatible with AMP


MackenzieHartung:
### Acceptance Criteria

AC1: On activating a plugin on wp-admin/plugins.php, if the plugin does something that’s not compatible with AMP, display an error at the top of the page
AC2: The error message will display the following text:. “<plugin name> is not compatible with AMP, and could result in content displaying improperly on your site.”

#### Tasks
- [x] Add a "reporting" flag whenever something is removed by the output buffer sanitizer and use that to build the "Bad Plugin" error.
- [x] Report other things than just node removal, including enqueued scripts and styles.
- [x] Report errors related to unknown sources.
- [x] Add method to disable AMP sanitization to be able to debug what is going on?
- [x] Improve initialization process for how validation is requested.

kienstra:
**Question About Approach**

Hi @westonruter,
Using the [sanitization reporting PR](https://github.com/Automattic/amp-wp/pull/912), what do you think about this approach:

1. [Before activating a plugin](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-admin/includes/plugin.php#L603) on `/wp-admin/plugins.php`,
2. Make a GET [request for the header](https://developer.wordpress.org/reference/functions/wp_remote_retrieve_header/) of a single post page, like `https://local.envi/?p=1633`
3. Store the `AMP-Validation-Error` from the header. Example:
```
{
"has_error": true,
"removed_nodes": {"script": 1},
"removed_attributes": null
}
```
3. After the [plugin activation](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-admin/includes/plugin.php#L645), make another request like in step 2, and compare the `AMP-Validation-Error`values.
4. If there are `'removed_nodes'` in step 3 that weren't present in step 2, display an error message.
5. Possibly display in the notice the node(s) that the plugin added that the sanitizer removed.

Thanks, Weston 😄

westonruter:
@kienstra I think that this will not give us complete results. A plugin may not do anything upon activation. It may require a user to actually do something, like add a widget or shortcode. 

I think the best we can probably do here is:

1. Watch for a plugin enqueueing a script or stylesheet. If a script or stylesheet gets removed by sanitizer, look for the `wp-content/(themes|plugins)/(.+)` directory the file is located in. And this would indicate which plugin or theme is doing something bad.
[...]

PR #912, opened 2018-01-27
Add sanitization reporting

### LINKED ISSUE #843: Incorporate AMP validator into post authorship flow, to alert if content within the post editor is not compatible with AMP

westonruter:
- [x] Add callback mechanism for being able to be informed of when an element or attribute is removed. (mainly applied in [e9f39](https://github.com/Automattic/amp-wp/pull/912/commits/e9f394a8acf473785c72eca3d914844073d6b382), pending review)
- [x] Add REST API endpoint allowing for HTML to be `POST`'ed and get back results for what changes were made. An element removal could be considered a sanitization error. The request would require an `edit_post` cap and require a nonce (see this [screencast](https://cloudup.com/cshOVEISKLg))
- [x] For normal frontend GET requests, consider being able to return validation errors in response header when requested via query param. The errors can be gathered during sanitization and then added before the output buffer is sent. ([applied](https://cloudup.com/c29EF6yH898), pending review)

Relates to content validation (#843) and also determining when a theme or plugin does something invalid (#842).

Fixes #843.


kienstra:
**In Development**

Hi @westonruter,
Thanks for starting this pull request, and for the outline of how to proceed. I'm working on this now.

kienstra:
**Screencast Of REST API Responses**

Here's a [screencast](https://cloudup.com/cshOVEISKLg) of the current validation via the REST API. The schema of the response will probably change as I look at outputting it in a frontend GET request.

includes/class-amp-theme-support.php
```diff
@@ -506,13 +506,25 @@ public static function get_amp_component_scripts() {
...
-	 * Finish output buffering.
+	 * Get the result of the output buffering, and add a header.
 	 *
 	 * @todo Do this in shutdown instead of output buffering callback?
+	 * @param string $output Buffered output.
+	 * @return string Finalized output.
+	 */
+	public static function finish_buffer_add_header( $output ) {
+		$markup = self::finish_output_buffering( $output );
+		AMP_Mutation_Utils::add_header();
```
westonruter: We'll want to limit this to only be added when a user specifically requests this additional information. Like there should be a nonce that must be present to authorize the reporting.

includes/class-amp-theme-support.php
```diff
@@ -523,6 +535,7 @@ public static function finish_output_buffering( $output ) {
 		$dom  = AMP_DOM_Utils::get_dom( $output );
 		$args = array(
 			'content_max_width' => ! empty( $content_width ) ? $content_width : AMP_Post_Template::CONTENT_MAX_WIDTH, // Back-compat.
+			'mutation_callback' => 'AMP_Mutation_Utils::track_removed',
```
westonruter: Per above, this should be conditional based on whether an authorized nonce is present.

includes/class-amp-theme-support.php
```diff
@@ -506,13 +506,25 @@ public static function get_amp_component_scripts() {
 	 * Start output buffering.
 	 */
 	public static function start_output_buffering() {
-		ob_start( array( __CLASS__, 'finish_output_buffering' ) );
+		ob_start( array( __CLASS__, 'finish_buffer_add_header' ) );
```
westonruter: We can leave this as `finish_output_buffering` I think.

kienstra:
**Example Of GET Response Headers**
_Question About Schema_

Hi @westonruter,
Here's a [screencast](https://cloudup.com/c29EF6yH898) of the frontend response headers showing AMP errors. What do you think about the schema of the response? 

amp-validation-issue

includes/class-amp-theme-support.php
kienstra: Thanks, @westonruter. That's a good idea to add the header only for users with a nonce.

westonruter:
@kienstra thanks for that great video. Really cool to see.

Aside: you can still use Postman with authenticated requests if you just install the [basic auth](https://github.com/WP-API/Basic-Auth) plugin.

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -192,4 +192,28 @@ private function build_placeholder( $parent_attributes ) {
...
+	/**
+	 * Removes a child of a node.
+	 *
+	 * Also, calls the mutation callback for it.
+	 * This tracks all the nodes that were removed.
+	 *
+	 * @since 0.7
+	 *
+	 * @param object $child The node to remove.
+	 * @param object $parent The parent node for which to remove the child (optional).
```
westonruter: Instead of `object` these should probably be `DOMNode`.

Also, is the `$parent` even needed since it will always be available via `$child->nodeParent`?

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -192,4 +192,28 @@ private function build_placeholder( $parent_attributes ) {
...
+	 * @return void.
+	 */
+	public function remove_child( $child, $parent = null ) {
+		if ( ( null === $parent ) && method_exists( $child->parentNode, 'removeChild' ) ) { // phpcs:ignore WordPress.NamingConventions.ValidVariableName.NotSnakeCaseMemberVar.
+			$child->parentNode->removeChild( $child ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.NotSnakeCaseMemberVar.
+		} elseif ( method_exists( $parent, 'removeChild' ) ) {
+			$parent->removeChild( $child );
+		}
+
+		if ( isset( $this->args['mutation_callback'] ) ) {
```
westonruter: It's possible that `remove_child` could be called without any node being removed, per the above `if`/`elseif` conditions. Should this check to see if something was actually removed? I'm curious as to the scenarios when the `else` condition would be met.

includes/utils/class-amp-mutation-utils.php
```diff
@@ -0,0 +1,212 @@
...
+	 * @return array $response The AMP validity of the markup.
+	 */
+	public static function get_response( $markup = null ) {
+		if ( isset( $markup ) ) {
+			self::process_markup( $markup );
+		}
+		$response = array(
+			'has_error'          => self::was_node_removed(),
+			'removed_nodes'      => self::$removed_nodes,
+			'removed_attributes' => self::$removed_attributes,
```
westonruter: It would be useful if the processed markup were also returned here. It could then be used for previewing, for example.

includes/sanitizers/class-amp-iframe-sanitizer.php
kienstra: Thanks, @westonruter. This [commit](https://github.com/Automattic/amp-wp/pull/912/commits/90090b0727cd652c197be8391bb21f8d72614793) changes the `@param` type to `DOMNode`, and removes the second parameter. Thanks for pointing out that the second isn't needed.

includes/sanitizers/class-amp-iframe-sanitizer.php
kienstra: Thanks, that's a good point. This [commit](https://github.com/Automattic/amp-wp/pull/912/commits/90090b0727cd652c197be8391bb21f8d72614793) from above has the `if` conditional nested, to ensure there was a removal before calling the `'mutation_callback'`.

includes/class-amp-theme-support.php
kienstra: Thanks, this [commit](https://github.com/Automattic/amp-wp/pull/912/commits/83dac8735704fe3d0e0c4e1dc8e2e97d33785ffa) changes the name back to `finish_output_buffering()`.

includes/utils/class-amp-mutation-utils.php
kienstra: Yes, that would help. Do you think we should return the processed markup only if we're validating a limited amount of markup, like a single Gutenberg block? We could detect this by whether `get_response()` is called with a `$markup` argument. 

```
if ( isset( $markup ) ) {
        $response['processed_markup] = $markup,
);
```

If `get_response()` is called without a `$markup` argument, it's validating the entire document.

kienstra: Thanks, @westonruter. 

[These commits](https://github.com/Automattic/amp-wp/compare/70521ef12c20139796108d09a588cf83de238486...0b6a4eef902ab76238ce7f758cbff9fb0b5b0474) output the `'processed_markup'`, when `$markup` is passed to the function.
request-postman-validate

includes/utils/class-amp-mutation-utils.php
```diff
@@ -0,0 +1,215 @@
...
+	 * It will return the validation errors the sanitizers found in rendering the page.
+	 *
+	 * @param string $markup To validate for AMP compatibility (optional).
+	 * @return array $response The AMP validity of the markup.
+	 */
+	public static function get_response( $markup = null ) {
+		$response = array();
+		if ( isset( $markup ) ) {
+			self::process_markup( $markup );
+			$response['processed_markup'] = esc_html( $markup );
```
kienstra: `esc_html()` probably isn't the best solution, but something is needed. Without it, almost the whole tag is stripped. Here's the result of this from Postman:
```
{"has_error":false,"removed_nodes":null,"removed_attributes":{"onchange":1},"processed_markup":"<amp-youtube onchange='someFunction()' data-videoid='XOY3ZUO6P0k' layout='responsive' width='694' height='390'><\/amp-youtube>"}
```

It looks like it escapes `/` characters:
```
<\/amp-youtube>
```

kienstra:
**Proposed Validation Message**

Here's the current message on updating a post in `/wp-admin/post.php`:

notice-amp-validation

includes/class-amp-theme-support.php
kienstra: Applied with a nonce, details to come shortly.

kienstra: `AMP_Mutation_Utils::add_header()` [now has a check for authorization](https://github.com/Automattic/amp-wp/pull/912/files#diff-60304ef5c3b3f17aeb304813b8716ebcR287) via `self::is_authorized()`. That verifies the nonce and capability.

includes/class-amp-theme-support.php
kienstra: Good point. This is [now only added](https://github.com/Automattic/amp-wp/pull/912/files#diff-2585472f207548f626a43a7ff3cab922R797) if `AMP_Mutation_Utils::is_authorized()`. That checks for the nonce and capability.

kienstra:
**Nonces Applied For /wp-admin/post.php**

There's now [nonce verification](https://github.com/Automattic/amp-wp/pull/912/files#diff-60304ef5c3b3f17aeb304813b8716ebcR271) on updating a post on `post.php`. It uses the `'update-post_' . $post_id` that's already in `$_REQUEST`. 

But that's a narrow use case. I might have to create a custom nonce, especially with validation on plugin activation (#842), or for Gutenberg.

ThierryA:
@kienstra could you merge the latest `develop` and fix merge conflict on this PR, thanks 😄

kienstra:
**Will Merge Develop**

Hi @ThierryA,
Sure, I'll merge in `develop`. The [last time](https://github.com/Automattic/amp-wp/pull/912/commits/46e708482b03f58bc5e5dab3e318b8985d8ebacb) I did this, it didn't resolve the conflict. But I'll work on this.

kienstra:
**Fixing My Mistakes In Resolving Merge Conflicts**

Hi @ThierryA,
I also see an issue with this PR. I think they're related to how I resolved the merge conflicts. I'm working on them now.

includes/utils/class-amp-mutation-utils.php
```diff
@@ -0,0 +1,325 @@
...
+	 * Output AMP validation data in the response header of a frontend GET request.
+	 *
+	 * This must be called before the document output begins.
+	 * Because the document is buffered,
+	 * The sanitizers run after the 'send_headers' action.
+	 * So it's not possible to call this function on that hook.
+	 *
+	 * @return void.
+	 */
+	public static function add_header() {
```
kienstra: This isn't actually used yet. I think it was a foundation for validating plugins on activation (#842)

phpcs.xml
```diff
@@ -11,6 +11,9 @@
 		</properties>
 	</rule>
 
+	<rule ref="WordPress.Files.FileName.InvalidClassFileName">
+		<exclude-pattern>tests/*</exclude-pattern>
+	</rule>
```
kienstra: Almost all of the test files begin with `test-`, while this rule states that they should begin with `class-`.

includes/utils/class-amp-mutation-utils.php
```diff
@@ -0,0 +1,325 @@
...
+		}
+		AMP_Content_Sanitizer::sanitize( $markup, amp_get_content_sanitizers(), $args );
+	}
+
+	/**
+	 * Registers the REST API endpoint for validation.
+	 *
+	 * @return void.
+	 */
+	public static function amp_rest_validation() {
```
kienstra: This isn't currently used, though it would be helpful for Gutenberg block validation.

kienstra:
**Ready For Review**

Hi @ThierryA,
This pull request is ready for review, with the error addressed.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -704,6 +704,9 @@ private function sanitize_disallowed_attributes_in_node( $node, $attr_spec_list
 
 		foreach ( $attrs_to_remove as $attr ) {
 			$node->removeAttributeNode( $attr );
+			if ( isset( $this->args['mutation_callback'], $attr->name ) ) {
+				call_user_func( $this->args['mutation_callback'], $node, 'removed_attr', $attr->name );
```
ThierryA: I believe `AMP_Mutation_Util::ATTRIBUTE_REMOVED` could be used here.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -304,4 +304,46 @@ public function maybe_enforce_https_src( $src, $force_https = false ) {
...
+	 *
+	 * @param DOMNode $node The node for which to remove the attribute.
+	 * @param string  $attribute The attribute to remove from the node.
+	 * @return void.
+	 */
+	public function remove_attribute( $node, $attribute ) {
+		if ( method_exists( $node, 'removeAttribute' ) ) {
+			$node->removeAttribute( $attribute );
+			if ( isset( $this->args['mutation_callback'] ) ) {
+				call_user_func( $this->args['mutation_callback'], $node, 'removed_attr', $attribute );
```
ThierryA: I believe `AMP_Mutation_Util::ATTRIBUTE_REMOVED` could be used here.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -304,4 +304,46 @@ public function maybe_enforce_https_src( $src, $force_https = false ) {
...
+	 * @since 0.7
+	 *
+	 * @param DOMNode $child The node to remove.
+	 * @return void.
+	 */
+	public function remove_child( $child ) {
+		if ( method_exists( $child->parentNode, 'removeChild' ) ) { // phpcs:ignore WordPress.NamingConventions.ValidVariableName.NotSnakeCaseMemberVar.
+			$child->parentNode->removeChild( $child ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.NotSnakeCaseMemberVar.
+			if ( isset( $this->args['mutation_callback'] ) ) {
+				call_user_func( $this->args['mutation_callback'], $child, 'removed' );
```
ThierryA: I believe `AMP_Mutation_Util::NODE_REMOVED` could be used here.

includes/utils/class-amp-mutation-utils.php
```diff
@@ -0,0 +1,325 @@
...
+ * Class AMP_Mutation_Utils
+ *
+ * @since 0.7
+ */
+class AMP_Mutation_Utils {
+
+	/**
+	 * The argument if an attribute was removed.
+	 *
+	 * @const array.
```
ThierryA: For constants the `@var` tag must be used for the PHPDoc to be valid. This comment applies in multiple constants in this PR.

includes/utils/class-amp-mutation-utils.php
```diff
@@ -0,0 +1,325 @@
...
+		} elseif ( ( self::NODE_REMOVED === $removal_type ) && isset( $node->nodeName ) ) {
+			self::$removed_nodes = self::increment_count( self::$removed_nodes, $node->nodeName ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.NotSnakeCaseMemberVar
+		}
+	}
+
+	/**
+	 * Tracks when a sanitizer removes an attribute or node.
+	 *
+	 * @param array  $histogram The count of attributes or nodes removed.
+	 * @param string $key The attribute or node name removed.
```
ThierryA: It is a wonder Travis didn't pickup alignment comments miss alignment, it might only sniff the variables alignment actually.

includes/utils/class-amp-mutation-utils.php
```diff
@@ -0,0 +1,325 @@
...
+	/**
+	 * Displays an error message on /wp-admin/post.php if the saved content is not valid AMP.
+	 *
+	 * Use $_GET, as get_query_var won't return the value.
+	 * This displays at the top of the 'Classic' editor.
+	 *
+	 * @return void.
+	 */
+	public static function display_error() {
+		$error = isset( $_GET[ self::ERROR_QUERY_KEY ] ) ? sanitize_text_field( wp_unslash( $_GET[ self::ERROR_QUERY_KEY ] ) ) : ''; // WPCS: CSRF ok.
```
ThierryA: For compliance sake, it would be good to check nonce here.

includes/utils/class-amp-mutation-utils.php
```diff
@@ -0,0 +1,325 @@
...
+ *
+ * @package AMP
+ */
+
+/**
+ * Class AMP_Mutation_Utils
+ *
+ * @since 0.7
+ */
+class AMP_Mutation_Utils {
```
ThierryA: What does Mutation stands for here, it could because I am native French but `AMP_Validation_Utils` would make more sense to me.

ThierryA: From an architectural perspective, it is interesting that all methods are `public static`. Are they really all meant to but used in that purposes?

kienstra: You're tight that `AMP_Validation_Utils` would be better. Mutation refers to the `mutation_callback`, which tracks when the sanitizer removes an attribute or node.

includes/utils/class-amp-mutation-utils.php
kienstra: Thanks, this [commit](https://github.com/Automattic/amp-wp/commit/f996673feb2e80585722fc7a9b787e968739f4a9) changes the tags to `@var`.

includes/utils/class-amp-mutation-utils.php
kienstra: Should this also align the comments, like:
```
@param array  $histogram The count of attributes or nodes removed.
@param string $key       The attribute or node name removed.
```

includes/utils/class-amp-mutation-utils.php
kienstra: This [commit](https://github.com/Automattic/amp-wp/commit/f8aeca80dd2fe7589d6097b4c2a802ea6d4fb679) changes the name of the classes.

includes/sanitizers/class-amp-base-sanitizer.php
kienstra: Thanks, this [commit](https://github.com/Automattic/amp-wp/commit/69317b877679ceb4bfb4a319244de0995f1b1222) uses that constant instead of the string.

includes/sanitizers/class-amp-base-sanitizer.php
kienstra: Thanks, this [commit](https://github.com/Automattic/amp-wp/commit/69317b877679ceb4bfb4a319244de0995f1b1222) from above substitutes the constant.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
kienstra: Thanks, this [commit](https://github.com/Automattic/amp-wp/commit/69317b877679ceb4bfb4a319244de0995f1b1222) from above substitutes the constant.

includes/utils/class-amp-mutation-utils.php
kienstra: It would be much better to not have all of these methods `static`. This might mean bootstrapping the class somewhere else. 

It's currently calling `AMP_Validation_Utils::init();` [here in amp.php](https://github.com/Automattic/amp-wp/commit/f8aeca80dd2fe7589d6097b4c2a802ea6d4fb679#diff-4805c837d3390517773926fb26726bcdR98). It might help if we could instantiate the class somewhere, like:
```
$validation_utils = new AMP_Validation_Utils();
validation_utils->init();
```

kienstra: [amp_post_meta_box()](https://github.com/Automattic/amp-wp/blob/develop/includes/admin/functions.php#L152) might be a good model for this.

ThierryA: Sure, this is not really a pressing need or a blocker though, it doesn't have to be addressed in this PR.

includes/utils/class-amp-mutation-utils.php
ThierryA: Yes

includes/utils/class-amp-mutation-utils.php
kienstra: Good point, @ThierryA. This [commit](https://github.com/Automattic/amp-wp/commit/583a6bcf549283c167673624ed5be141f3be347a) adds nonce verification, via `check_admin_referer()`.

includes/utils/class-amp-mutation-utils.php
kienstra: This [commit](https://github.com/Automattic/amp-wp/commit/cd910ff648356a9d454ca4c3148d6f9935544260) aligns the spacing of the comments. The pre-commit hook blocked aligning this in [two](https://github.com/Automattic/amp-wp/pull/912/files#diff-7d234989187d87a372e2aa6aa1299a71R245) [places](https://github.com/Automattic/amp-wp/pull/912/files#diff-7d234989187d87a372e2aa6aa1299a71R313), possibly because there was only one `@param` and a `@return`.

kienstra:
**Applied Code Review Suggestions**

Hi @ThierryA,
Thanks for waiting for this. All of your suggestions above are applied, exception for the (good) question about whether all of the methods should be static.

includes/class-amp-theme-support.php
```diff
@@ -752,7 +752,9 @@ public static function start_output_buffering() {
 	 * @see AMP_Theme_Support::start_output_buffering()
 	 */
 	public static function finish_output_buffering() {
-		echo self::prepare_response( ob_get_clean() ); // WPCS: xss ok.
+		$output = self::prepare_response( ob_get_clean() );
+		AMP_Validation_Utils::add_header();
```
westonruter: I'm going to remove thus until we have it being utilized in the plugin.

kienstra: Sure, that's fine.

PR #914, opened 2018-01-29
[WIP] Add post embed support

### LINKED ISSUE #809: Improve handling of WordPress post embeds
The handling of post embeds isn't great in AMP right now because the `p.wp-embedded-content` isn't getting replaced with the `iframe.wp-embedded-content` like it does in the non-AMP version. The result is some doubling of content.

Compare in a post I just published:

[Canonical (non-AMP) version](https://weston.ruter.net/2017/11/16/mel-choyce-recap-on-wordpress-4-9/):

> image

[AMP version](https://weston.ruter.net/2017/11/16/mel-choyce-recap-on-wordpress-4-9/amp/):

> image

westonruter:
This is split out from #889 for #841. We need a few things to fully support post embeds, including:

* Move the contents of the `blockquote.wp-embedded-content` into the `amp-iframe` as the `placeholder`. Note that this must not include the `blockquote` itself, because `wpautop` will break it.
* Allow clicking inside the post iframe to cause `_top` navigation.
* Allow iframe self-resize in AMP: https://www.ampproject.org/docs/reference/components/amp-iframe#iframe-resizing
* Consider inlining the embeds and not using an iframe at all for embeds from elsewhere on the site.

See previously for some key considerations: https://github.com/Automattic/amp-wp/issues/288

A core patch may be required here as well.

Closes #809.


swissspidy:
@westonruter Shall we close this in favor of #3465?

westonruter:
Yes. Thanks!

PR #918, opened 2018-01-30
Fix parsing of HTML5 empty media elements including source and track

### LINKED ISSUE #62: amp-video/amp-audio supports limited children
amp-video only supports `<source>` `<div fallback>` and `<div placeholder>` as children.

amp-audio only supports  `<source>` and `<div fallback>`.

We should reformat the child nodes to match this.

westonruter:
Fixes #62.

### DIFF
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -75,6 +75,13 @@ class AMP_DOM_Utils {
 			$document
 		);
 
+		// Force all self-closing tags to have closing tags since DOMDocument isn't fully aware.
+		$document = preg_replace(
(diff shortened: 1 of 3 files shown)


includes/utils/class-amp-dom-utils.php
```diff
@@ -75,6 +75,13 @@ public static function get_dom( $document ) {
 			$document
 		);
 
+		// Force all self-closing tags to have closing tags since DOMDocument isn't fully aware.
+		$document = preg_replace(
+			'#<(' . join( self::$self_closing_tags, '|' ) . ')[^>]*>(?!</\1>)#',
```
ThierryA: No changes require: while both work, usually the glue string is the first parameter when using `join()` which is an alias of `implode()`.

PR #920, opened 2018-01-30
Fix AMP preview icon in Firefox

zigasancin:
The AMP preview icon isn't displaying correctly in Firefox - see screenshot below (probably because of a different interpretation of the text-indent css rule). My proposed fix works in latest releases of Firefox, Chrome, Safari, Edge and IE.

fix-amp-wp-preview-icon-in-firefox


westonruter:
Sure enough! Not sure how that was missed. With patch, it looks like the following in Chrome, Firefox, and Safari:

image

This could be cherry-picked into a 0.6.1 release if we do one.

PR #923, opened 2018-02-01
Handle general form posts

### LINKED ISSUE #911: Add automatic support for handling form POST post submissions that redirect
All AMP `post` forms use XHR to submit the request. When we sanitize a `form` (#907) to upgrade a `form[method=post][action=/foo]` to `form[method=post][action-xhr=/foo]`, we also need to make sure that the result of the form being submitted is handled as well. In particular, when we rewrite an `action` to an `action-xhr` we should add a query var to the URL like `_wp_amp_action_xhr_converted=1`(e.g. the `action` here could get sanitized to be `/foo?_wp_amp_action_xhr_converted=1`) so that when WP calls `wp_redirect()` we can hook into it and automatically sent the `AMP-Redirect-To` and `Access-Control-Expose-Headers` headers instead: https://www.ampproject.org/docs/reference/components/amp-form#redirecting-after-a-submission

Otherwise, if the `action-xhr` attribute was supplied to begin with, then no such redirect interception would be done, as then it would be assumed that the theme/plugin is already properly handling the XHR request client-side.

DavidCramer:
Fixes #911

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -424,6 +424,14 @@ function amp_prepare_xhr_post() {
 	} else {
 		// Add amp comment hooks.
 		add_filter( 'wp_redirect', 'amp_handle_general_post', PHP_INT_MAX, 2 );
+		add_action( 'template_redirect', function() {
+			/*
+			 * Buffering starts here, so unlikely the form has a redirect,
+			 * so force a redirect to the same page.
+			 */
+			$location = esc_url_raw( wp_unslash( $_SERVER['REQUEST_URI'] ) ); // WPCS: CSRF ok, input var ok.
+			amp_handle_general_post( $location );
+		}, 0 );
 	}
 	// Add die handler for AMP error display.
 	add_filter( 'wp_die_handler', function() {
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -412,7 +412,7 @@ function amp_handle_xhr_request() {
 	}
 
 	if ( isset( $pagenow ) && 'wp-comments-post.php' === $pagenow ) {
-		// This only runs on wp-comments-post.php.
+		// We don't need any data, so just send a success.
 		add_filter( 'comment_post_redirect', function() {
 			// We don't need any data, so just send a success.
 			wp_send_json_success();
@@ -422,7 +422,6 @@ function amp_handle_xhr_request() {
 		add_filter( 'wp_redirect', 'amp_intercept_post_request_redirect', PHP_INT_MAX, 2 );
 		amp_handle_xhr_headers_output();
 	}
-
 }
 
 /**
@@ -456,18 +455,37 @@ function amp_handle_xhr_headers_output() {
 }
 
 /**
- * Handle a general, non comment AMP XHR post.
+ * Intercept the response to a non-comment POST request.
  *
  * @since 0.7.0
  * @param string $location The location to redirect to.
  */
 function amp_intercept_post_request_redirect( $location ) {
-	$host = wp_parse_url( $location, PHP_URL_HOST );
-	if ( is_null( $host ) ) {
-		$location = home_url( $location );
+
+	// Make sure relative redirects get made absolute.
+	$parsed_location = array_merge(
+		array(
+			'scheme' => 'https',
+			'host'   => wp_parse_url( home_url(), PHP_URL_HOST ),
+			'path'   => strtok( wp_unslash( $_SERVER['REQUEST_URI'] ), '?' ),
+		),
+		wp_parse_url( $location )
+	);
+
+	$absolute_location = $parsed_location['scheme'] . '://' . $parsed_location['host'];
+	if ( isset( $parsed_location['port'] ) ) {
+		$absolute_location .= ':' . $parsed_location['port'];
 	}
-	header( 'AMP-Redirect-To: ' . $location );
-	header( 'Access-Control-Expose-Headers: AMP-Redirect-To;' );
+	$absolute_location .= $parsed_location['path'];
+	if ( isset( $parsed_location['query'] ) ) {
+		$absolute_location .= '?' . $parsed_location['query'];
+	}
+	if ( isset( $parsed_location['fragment'] ) ) {
+		$absolute_location .= '#' . $parsed_location['fragment'];
+	}
+
+	header( 'AMP-Redirect-To: ' . $absolute_location );
+	header( 'Access-Control-Expose-Headers: AMP-Redirect-To' );
 	// Send json success as no data is required.
 	wp_send_json_success();
 }
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -406,7 +406,7 @@ function amp_print_schemaorg_metadata() {
  *
  * @since 0.7.0
  */
-function amp_prepare_xhr_post() {
+function amp_handle_xhr_request() {
 	global $pagenow;
 	if ( ! isset( $_GET['__amp_source_origin'] ) || ! isset( $pagenow ) ) { // WPCS: CSRF ok. Beware of AMP_Theme_Support::purge_amp_query_vars().
 		return;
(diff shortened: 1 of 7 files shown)


DavidCramer:
@westonruter I think this is done and ready for you to review.

includes/amp-helper-functions.php
```diff
@@ -406,17 +406,33 @@ function amp_print_schemaorg_metadata() {
...
+	} elseif ( ! isset( $_GET['_wp_amp_action_xhr_converted'] ) ) { // WPCS: CSRF ok.
+		// submission was from a set action-xhr, implying it's being handled already.
+		return;
+	} else {
+		// Add amp redirect hooks.
+		add_filter( 'wp_redirect', 'amp_handle_general_post', PHP_INT_MAX, 2 );
+		add_action( 'template_redirect', function() {
+			/*
+			 * Buffering starts here, so unlikely the form has a redirect,
+			 * so force a redirect to the same page.
```
westonruter: This is clever, but I'm worried it could have unintended results. In particular, when a form submission does not do a redirect after a `POST` then the data from the `POST` submission is often displayed in some way on the response. If we do the redirect, then it could cause the user to be redirected back to a blank page. Consider something like this:

```php
<form method="post">
     <?php if ( ! empty( $_POST ) ) : ?>
          Submission successful!
          <pre><?php echo esc_html( wp_json_encode( wp_unslash( $_POST ) ) ); ?></pre>
     <?php else : ?>
         <input name="data"><button>Submit</button>
     <?php endif; ?>
</form>
```

Something like this will break if we do the redirect here.

includes/amp-helper-functions.php
```diff
@@ -406,17 +406,33 @@ function amp_print_schemaorg_metadata() {
  *
  * @since 0.7.0
  */
-function amp_prepare_comment_post() {
-	if ( ! isset( $_GET['__amp_source_origin'] ) ) { // WPCS: CSRF ok. Beware of AMP_Theme_Support::purge_amp_query_vars().
+function amp_prepare_xhr_post() {
```
westonruter: Let's rename this to `amp_handle_xhr_request`.

includes/amp-helper-functions.php
```diff
@@ -406,17 +406,33 @@ function amp_print_schemaorg_metadata() {
...
-
+	if ( 'wp-comments-post.php' === $pagenow ) {
+		// This only runs on wp-comments-post.php.
+		add_filter( 'comment_post_redirect', function() {
+			// We don't need any data, so just send a success.
+			wp_send_json_success();
+		}, PHP_INT_MAX, 2 );
+	} elseif ( ! isset( $_GET['_wp_amp_action_xhr_converted'] ) ) { // WPCS: CSRF ok.
+		// submission was from a set action-xhr, implying it's being handled already.
+		return;
```
westonruter: I suggest inverting the isset check and moving the redirect filter from else to here.

includes/amp-helper-functions.php
```diff
@@ -438,3 +454,18 @@ function amp_prepare_comment_post() {
...
 	header( 'AMP-Access-Control-Allow-Source-Origin: ' . $origin, true );
 }
+
+/**
+ * Handle a general, non comment AMP XHR post.
+ *
+ * @since 0.7.0
+ * @param string $location The location to redirect to.
+ */
+function amp_handle_general_post( $location ) {
```
westonruter: Let's call this `amp_intercept_post_request_redirect`.

includes/amp-helper-functions.php
```diff
@@ -438,3 +454,18 @@ function amp_prepare_comment_post() {
...
+
+	$url = site_url( $location );
```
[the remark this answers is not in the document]
DavidCramer: @westonruter Sorry I always get those two mixed up. was supposed to be `home_url()` since the `$location` is a relative link. But this may or maynot be, so should check for host first.
Good catch :)

includes/amp-helper-functions.php
DavidCramer: @westonruter The `wp_die_handler` and `AMP-Access-Control-Allow-Source-Origin` are fall throughs for both the comments and general submissions. moving the filter in here would make these happen for handled and unhandled. But I see what you're saying. and know how to simplify it.

DavidCramer: @westonruter see 83a0799 does that work?

includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -108,4 +112,43 @@ public function sanitize() {
...
+		$node->appendChild( $this->create_error_template() );
+	}
+
+	/**
+	 * Creates a error handler element node.
+	 *
+	 * @link https://www.ampproject.org/docs/reference/components/amp-form
+	 * @since 0.7
+	 *
+	 * return DOMElement
```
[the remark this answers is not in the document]
DavidCramer: @westonruter I really need to focus on phpdoc tags. thanks.

includes/amp-helper-functions.php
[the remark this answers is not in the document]
DavidCramer: @westonruter this is somewhat a hack and will probably have side effects see a9d9bfd

DavidCramer:
@westonruter can review this again when you're ready.

includes/amp-helper-functions.php
```diff
@@ -406,17 +406,54 @@ function amp_print_schemaorg_metadata() {
...
+			wp_send_json_success();
+		}, PHP_INT_MAX, 2 );
+		amp_handle_xhr_headers_output();
+	} elseif ( isset( $_GET['_wp_amp_action_xhr_converted'] ) ) { // WPCS: CSRF ok.
+		// Add amp redirect hooks.
+		add_filter( 'wp_redirect', 'amp_intercept_post_request_redirect', PHP_INT_MAX, 2 );
+		add_action( 'template_redirect', function() {
+			// grab post data.
+			$transient_name = uniqid();
+			set_transient( $transient_name, wp_unslash( $_POST ), 60 ); // WPCS: CSRF ok, input var ok.
```
westonruter: This is clever, but I do fear it could lead to unexpected results like you said. So I think for now we should omit it and instead revisit it later if we find that it is a common problem.

westonruter:
@DavidCramer this is looking good. One thing left is to add unit tests. Please add them specifically for:

* `amp_intercept_post_request_redirect()`
* `amp_handle_xhr_headers_output()`
* `amp_handle_xhr_request()`

Really cool to see that with your changes here, the Jetpack contact form _just works_.

includes/amp-helper-functions.php
DavidCramer: @westonruter Ye I thought so to. I was going to just remove it, but wanted to see if I could at least make your example work. It's better not being there for now.

DavidCramer:
@westonruter I used the jetpack contact form as well as a basic custom made for for testing, so I'm glad they work without effort. Will get onto those unit tests.

includes/amp-helper-functions.php
```diff
@@ -429,12 +442,50 @@ function amp_prepare_comment_post() {
 			if ( is_wp_error( $error ) ) {
 				$error = $error->get_error_message();
 			}
-			$error = strip_tags( $error, 'strong' );
-			wp_send_json( compact( 'error' ) );
+			$amp_mustache_allowed_html_tags = array( 'strong', 'b', 'em', 'i', 'u', 's', 'small', 'mark', 'del', 'ins', 'sup', 'sub' );
```
DavidCramer: Oh nice one!

DavidCramer:
@westonruter finally managed to push these unit tests. I'm not that good with unit tests yet, but fortunately @ThierryA lent a hand. You can review this now.

includes/amp-helper-functions.php
DavidCramer: @westonruter this is a little hard to test as it's on a die handler and causes much issues. I left it out of the burrent test. Perhaps you could give a suggestion?

westonruter:
Cool, I haven't used `xdebug_get_headers()` previously. That's a handy one to know. But is it what requires the use of `@runInSeparateProcess`? That really slows the tests down so we should try to find an alternative.

DavidCramer:
@westonruter unfortunately  `@runInSeparateProcess` is required. The setup process echos status which then gives you `headers already sent` causing the test to fail. Generally we would group them and have them as exclusions so you can run them individually. Which is what WordPress does for things like ajax tests.

Issue #861, opened 2018-01-15
Create a WP-CLI Script to Output Comments to a test page


ThierryA:
**Acceptance criteria:**
AC1: Create a `wp-cli` script in the directory [bin/](https://github.com/Automattic/amp-wp/tree/develop/bin).
AC2. This will create a post with a list of comments, in order to test their AMP compatibility (see this [script](https://github.com/Automattic/amp-wp/blob/develop/bin/create-embed-test-post.php) for reference).

#### Tasks

- [x] Create script, like was done for embeds, that generates a post with a list of comments.

PR #924, opened 2018-02-01
Add WP-CLI Script to create Comments

DavidCramer:
#861


ThierryA:
@DavidCramer I pushed two commit to reduce complexity a bit. I like your approach to get content using the Lorem API but it is maybe a bit over thinking, it shouldn’t be a blocker to run the script and a simple sample string does the job.
Please take a look at my changes to double check them and this will be good to go.

DavidCramer:
@ThierryA I agree. I looked at the changes and its nice and simple. good work on that content generator. nice and simple. I may steal that..

PR #926, opened 2018-02-03
Fix dispatch_key handling for NAME_VALUE_DISPATCH; add cdata validation

westonruter:
This fixes the longstanding annoyance for when the spec was read, the one `value` had to be removed for `data-multi-size` or else unit tests would fail. It turns out to be due to the `dispatch_key` not being supported. So this is now fixed and the list of tags and attributes can be re-generated at will.

In addition to `dispatch_key` being included in the generated PHP file, so too `cdata` is now included, with validation for blacklisted regex and amp-keyframes styles. This is needed for validating elements in the `head`. See #926.

### DIFF
--- a/includes/sanitizers/class-amp-rule-spec.php
+++ b/includes/sanitizers/class-amp-rule-spec.php
@@ -17,6 +17,7 @@ abstract class AMP_Rule_Spec {
 	 */
 	const ATTR_SPEC_LIST = 'attr_spec_list';
 	const TAG_SPEC       = 'tag_spec';
+	const CDATA          = 'cdata';
 
 	/**
 	 * AMP attr_spec value check results
--- a/includes/sanitizers/class-amp-rule-spec.php
+++ b/includes/sanitizers/class-amp-rule-spec.php
@@ -47,6 +47,23 @@ abstract class AMP_Rule_Spec {
 	const VALUE_REGEX             = 'value_regex';
 	const VALUE_REGEX_CASEI       = 'value_regex_casei';
 
+	/*
+	 * DispatchKeyTypes:
+	 * https://github.com/ampproject/amphtml/blob/eda1daa8c40f830207edc8d8088332b32a15c1a4/validator/validator.proto#L111-L120
+	 */
+
+	// Indicates that the attribute does not form a dispatch key.
+	const NONE_DISPATCH = 0;
+
+	// Indicates that the name of the attribute alone forms a dispatch key.
+	const NAME_DISPATCH = 1;
+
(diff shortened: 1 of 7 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -105,6 +112,33 @@ public function collect_style_elements() {
...
+	 * Validate amp-keyframe style.
+	 *
+	 * @since 0.7
+	 * @link https://github.com/ampproject/amphtml/blob/b685a0780a7f59313666225478b2b79b463bcd0b/validator/validator-main.protoascii#L1002-L1043
+	 *
+	 * @param DOMElement $style Style element.
+	 * @return true|WP_Error Validity.
+	 */
+	private function validate_amp_keyframe( $style ) {
+		if ( strlen( $style->textContent ) > 500000 ) {
```
ThierryA: Maybe better to store 500000 in constant.

includes/sanitizers/class-amp-rule-spec.php
```diff
@@ -47,6 +48,23 @@
 	const VALUE_REGEX             = 'value_regex';
 	const VALUE_REGEX_CASEI       = 'value_regex_casei';
 
+	/*
+	 * DispatchKeyTypes:
+	 * https://github.com/ampproject/amphtml/blob/eda1daa8c40f830207edc8d8088332b32a15c1a4/validator/validator.proto#L111-L120
+	 */
+
+	// Indicates that the attribute does not form a dispatch key.
```
ThierryA: Would be good to add valid PHPDoc with @since tags for these constances.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -88,6 +88,13 @@ public function collect_style_elements() {
 				continue;
 			}
 
+			if ( 'body' === $style_element->parentNode->nodeName && $style_element->hasAttribute( 'amp-keyframes' ) ) {
+				$validity = $this->validate_amp_keyframe( $style_element );
```
ThierryA: I guess this doesn't really have to be stored in a var and could part of the conditional statement above.

westonruter:
Thanks, I'll address these in the next PR.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: The reason behind is is that in #912 we need to add reporting and `$validity` is a `WP_Error` instance, so we'd want to report it when this is merged into that PR.

PR #931, opened 2018-02-05
Prepare AMP response at shutdown instead of output buffer callback

westonruter:
This will make it easier to debug things since printing can be done in shutdown function but cannot in output buffer callback.

Furthermore, this prevents a fatal error if `print_r()` is attempted in the output buffer callback:

> PHP Fatal error:  print_r(): Cannot use output buffering in output buffering display handlers in .../wp-content/plugins/amp/includes/class-amp-theme-support.php on line 863

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -809,8 +809,7 @@ class AMP_Theme_Support {
 	 * @see AMP_Theme_Support::start_output_buffering()
 	 */
 	public static function finish_output_buffering() {
-		$output = ob_get_clean();
-		echo self::prepare_response( $output ); // WPCS: xss ok.
+		echo self::prepare_response( ob_get_clean() ); // WPCS: xss ok.
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/class-amp-theme-support.php
```diff
@@ -791,36 +791,54 @@ protected static function ensure_required_markup( DOMDocument $dom ) {
...
 
 	/**
 	 * Finish output buffering.
 	 *
-	 * @todo Do this in shutdown instead of output buffering callback? This will make it easier to debug things since printing can be done in shutdown function but cannot in output buffer callback.
+	 * @since 0.7
+	 * @see AMP_Theme_Support::start_output_buffering()
+	 */
+	public static function finish_output_buffering() {
+		$output = ob_get_clean();
```
ThierryA: I would advise to avoid storing the output in a variable unless necessary, especially in the context of the output buffer. For instance the memory goes from 3018376 to 3006088 on my local install by moving `ob_get_clean()` to `self::prepare_response( ob_get_clean() );`.

Issue #930, opened 2018-02-05
Extend style sanitizer to incude CSS parser


westonruter:
In order to be able to reliably excise `!important` from styles (#927), we need a proper CSS parser to be able to parse the rules so that higher-specificity selectors can be used instead of `!important`. 

Additionally, a CSS parser used in conjunction with sanitization where the document has been output buffered and parsed into a DOM tree will allow us to strip out all CSS rules from the document that are not relevant to the current page, such as by checking to see if a rule references a class name that does not exist on the current page.

This will also address minification in #688. Class names could be minified in the CSS rules as well as in the document's elements that have the `class` attributes.

Current work can be seen in the [`add/css-parser`](https://github.com/Automattic/amp-wp/tree/add/css-parser) branch. The idea is to use the same phpegjs library that is being used by Gutenberg to parse blocks. The PEG grammar needs to be written for CSS. The grammar currently in the branch needs to be simplified to reduce the granularity. We don't need each token separately parsed for a given selector, but rather just the entire selector string as a whole.

As part of this we should probably augment `\AMP_Theme_Support::ensure_required_markup()` to ensure the `amp-boilerplate` is present. Likewise, the way that `AMP_WP_Styles` waits to output styles should be removed in favor of just outputting them right away to then allow the sanitizer to collect them. The CSS sanitizer can then be responsible actually for adding the `style[amp-custom]` to the `head`.

This will supersede and incorporate #610 to switch from blacklist to whitelist. It will also need to improve the approach to scrubbing `!important` in #927. See https://github.com/Automattic/amp-wp/issues/927#issuecomment-363199810.

westonruter:
In Twenty Fifteen, on a single page that is fully populated with all embeds, widgets, and has a comment in it, the `style.css` (97KB) coverage as reported by Chrome is:

![image](https://user-images.githubusercontent.com/134745/37860965-566b7f86-2eed-11e8-9182-04dc04dadd7f.png)

So if we were able to delete all of the unused bytes, the the CSS would be below 50KB.

Additionally, if the CSS were also minified with all whitespace and comments stripped, then this would bring the initial size from 97KB down to 77KB.

Here's a list of the class names used by frequency: https://gist.github.com/westonruter/29c8e0284f39ace5c95cbe7d81546b6d

PR #935, opened 2018-02-06
Update style sanitizer approach

### LINKED ISSUE #927: Remove all instances of !important in CSS.
When parsing the CSS, it would be good to remove all instances of `!important` as these invalidate the AMP markup.

westonruter:
* Eliminate custom `WP_Styles` in favor of consolidating logic in `AMP_Style_Sanitizer`. This ensures that plugins that print `link` tags directly will automatically get them included. The same goes for `style` elements that a plugin prints.
* External stylesheets and `style` elements now get concatenated in document order to ensure proper cascade.
* Include CSS comment before each concatenated stylesheet indicating its source. This helps with debugging.
* Run `!important` and overflow removal logic on inlined external stylesheets as well as style elements and style attributes.
* Make sure that `!important` does not appear _anywhere_ in CSS, including comments, as otherwise invalid and entire CSS gets stripped by whitelist sanitizer.
* Stop concatenating CSS stylesheets to output if surpasses max size; show HTML comment containing the stylesheet URL that was excluded due to size. This helps with debugging.
* Enforce that style sanitizer and whitelist sanitizer always run last.

Fixes #927. However, the longer-term fixes will need to be done as part of #930. This PR lays the groundwork for #930.


amedina:
@weston are we avoiding this: "Run !important removal logic" as discussed with @pbakaus?

westonruter:
@amedina Not yet. That will be addressed as part of #930. We can't do it yet because we don't have a CSS parser. 

(I'll be squashing the commits once I identify why Travis is failing.)

westonruter:
Wow, that was one of the most obscure things I've ever had to debug.

Issue #933, opened 2018-02-06
Add default stylesheet for AMP content when in theme support mode


westonruter:
As discovered in https://github.com/Automattic/amp-wp/pull/921#issuecomment-363275594 there are some CSS rules that the sanitizer expects to be present which are not currently being output when there is theme support. In particular, these styles seem to be necessary:

https://github.com/Automattic/amp-wp/blob/7b88ae0acd450e1658f1248d11319a01e7c5e920/templates/style.php#L39-L49

An audit needs to be done to see which styles are needing to be present in the default styles, without requiring each theme to implement these on their own. Given that the plugin will automatically concatenate CSS into the `style[amp-custom]` element, a new CSS file can be added and enqueued normally when an AMP page is rendered in theme support.

Note that the CSS enqueued in theme support should not include any of the styles that are used in the legacy paired mode templates.

Issue #919, opened 2018-01-30
Sanitizer: Request to convert <img> to <amp-img> correctly


kienstra:
As @westonruter [mentioned](https://github.com/xwp/ampconf/issues/48#issuecomment-361698890) in the 'ampconf' theme, the sanitizer doesn't look to be converting `<img>` tags to `<amp-img>` correctly:

* `width` and `height` attributes are unexpectedly different
* Changes in the `srcset`

Here's @westonruter's [temporary workaround](https://github.com/xwp/ampconf/commit/d26f628d0a25207ec8809ac1d0315814480b7cb7) for reference.

kienstra:
**Will Work On After #864**

Hi @westonruter,
If it's alright, I'll work on this after the video issue in #864.

kienstra:
**Request For Another Look**

Hi Koop or @westonruter,
Could you please let me know if you still see this issue? I can't reproduce the issue [mentioned](https://github.com/xwp/ampconf/issues/48#issuecomment-361698890) in #48, looking at a [single.php page](https://github.com/xwp/ampconf/blob/develop/single.php). Maybe something fixed it in the meantime.

Without the [AMP plugin](github.com/Automattic/amp-wp) active, 
```
the_post_thumbnail( 'ampconf-1040x400' );
```
produced:
```
<img width="1040" height="400" src="http://local.envi/wp-content/uploads/2018/01/American_bison_k5680-1-1040x400.jpg" class="attachment-ampconf-1040x400 size-ampconf-1040x400 wp-post-image" alt="" srcset="http://local.envi/wp-content/uploads/2018/01/American_bison_k5680-1-1040x400.jpg 1040w, http://local.envi/wp-content/uploads/2018/01/American_bison_k5680-1-768x295.jpg 768w" sizes="(max-width: 1040px) 100vw, 1040px">
```
With the plugin active, that function seems to produce the same `width`, `height`, and `srcset`:
```
[...]

Issue #876, opened 2018-01-18
Deprecate AMP_WP_Utils


westonruter:
All of the methods in the AMP_WP_Utils class are for WP<4.7 to support `wp_parse_url()`. The entire class should be deprecated and any references to `AMP_WP_Utils::parse_url()` should be replaced with straight calls to `wp_parse_url()`.

PR #937, opened 2018-02-06
Issue #864: Remove 'sizes' attribute for <amp-iframe> and <amp-video>.

kienstra:
**Request For Code Review**

Hi @westonruter,
Could you please review this pull request, which removes the `sizes` attribute for `<amp-iframe>` and `<amp-video>`?

Before, there was a workaround to ensure these didn't overflow. I had added `style="max-width:100%"`
This removes that workaround, and instead sets the layout to 'responsive.'

We don't need to remove `sizes` from `<amp-img>`, right? The discussion in [PR #921](https://github.com/Automattic/amp-wp/pull/921) only briefly mentioned `<amp-img>`, though the referenced [amphtml issue # 11575](https://github.com/ampproject/amphtml/issues/11575) discusses `<amp-img>`.

**Update**
Sorry, Weston, I might have missed the real need of removing `sizes`. It sounds like the main need was removing them from images. If there's still a need to remove them from images, I'd be happy to do that.


includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -80,8 +80,10 @@ public function sanitize() {
 
 			$this->did_convert_elements = true;
 
-			$new_attributes = $this->enforce_fixed_height( $new_attributes );
-			$new_attributes = $this->enforce_sizes_attribute( $new_attributes );
+			$new_attributes = $this->set_layout( $new_attributes );
+			if ( isset( $new_attributes['width'] ) && isset( $new_attributes['height'] ) ) {
+				$this->add_or_append_attribute( $new_attributes, 'class', 'amp-wp-enforced-sizes' );
+			}
```
kienstra: These 3 lines are copied from `enforce_sizes_attribute()`.

westonruter:
I'm not clear on whether removing `sizes` is what we're supposed to do, and whether it is just for `amp-iframe` or `amp-img`. I'm waiting on confirmation for https://github.com/Automattic/amp-wp/pull/921#issuecomment-363275594 I think. I'm honestly a bit lost. In any case, I've deployed your changes up to a new multidev environment at https://sizesfix-ampconfdemo.pantheonsite.io/ 

Would you get some QA support for check if the changes help or hinder?

westonruter:
Would this potentially solve #919 as well?

pbakaus:
Commented on the other one – for the reasons outlined, I'd definitely recommend removing.

westonruter:
@pbakaus Thank you for clarifying. 

@kienstra Please remove `sizes` for images as well.

kienstra:
**Thank, Sizes Removed From Images**

Hi @pbakaus and @westonruter,
Thanks for your guidance on the `sizes` attribute. The [commit above](https://github.com/Automattic/amp-wp/pull/937/commits/6848541e9aa680f8756719d338d9358d1c4e516a) removes them from `<amp-img>` elements.

kienstra:
**amp-video, amp-iframe**

Hi @westonruter,
The `<amp-video>` looks to display well with no `sizes`, and with `layout="responsive"`. This ensures that it doesn't overflow its container, [like a sidebar](https://sizesfix-ampconfdemo.pantheonsite.io/category/gaming/).

It also [doesn't look to expand](https://sizesfix-ampconfdemo.pantheonsite.io/2018/02/14/test-embed-post/) larger than its `width` or `height`, so users shouldn't have an unexpected change with this version:

amp-video-no-sizes

But the `<amp-iframe>` does expand to fill its container when it has `layout="responsive"`. This might cause surprises with the new version. 

So I removed my addition of `layout="responsive"` to the `<amp-iframe>`, and [instead re-added](https://github.com/Automattic/amp-wp/pull/937/commits/11405b3c40346605fe6ba08b72f67f46d02be261#diff-6b823d0b2316959c565a48afd24d021aR84) `style="max-width:100%"`. 

This is just a workaround, and it might help to eventually have `.amp-wp-enforced-sizes` in the `<style amp-custom>`.

westonruter:
> This is just a workaround, and it might help to eventually have .amp-wp-enforced-sizes in the <style amp-custom>.

See #933.

westonruter:
I've deployed the changes to https://sizesfix-ampconfdemo.pantheonsite.io/

Question: Should this eliminate the need for https://github.com/xwp/ampconf/blob/992fa40c75b9be898326b186a4c3efd0dc788325/inc/template-functions.php#L72-L112 ?

kienstra:
**New Commit**

Hi @westonruter,
Thanks for deploying the changes. This [new commit](https://github.com/Automattic/amp-wp/pull/937/commits/6bd4326a88d60539b8568e2551f774e278f0720c) removes the `sizes` attribute, where this [previous commit](https://github.com/Automattic/amp-wp/pull/937/commits/6848541e9aa680f8756719d338d9358d1c4e516a) only prevented it from being added if it wasn't present.

The [filters you mentioned](https://github.com/xwp/ampconf/blob/992fa40c75b9be898326b186a4c3efd0dc788325/inc/template-functions.php#L72-L112) override the removal of the `sizes`. When they're present, this plugin's removal of the `sizes` doesn't apply. 

Still, I need to look at the underlying cause causes of those filters, and whether it's appropriate to remove them.

kienstra:
**PR To Remove Workaround**

Hi @westonruter,
The [theme PR #81](https://github.com/xwp/ampconf/pull/81) removes the [workaround filters](https://github.com/xwp/ampconf/blob/992fa40c75b9be898326b186a4c3efd0dc788325/inc/template-functions.php#L72-L112) you mentioned. In testing it, I noticed an issue in theme styling. I described it on the ticket.

westonruter:
Deployed latest to https://sizesfix-ampconfdemo.pantheonsite.io/

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -119,7 +119,6 @@ private function filter_attributes( $attributes ) {
 				case 'alt':
 				case 'class':
 				case 'srcset':
-				case 'sizes':
```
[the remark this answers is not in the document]
kienstra: That sounds good, @delawski. If it's alright, I'll work on this in several hours.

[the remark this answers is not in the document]
westonruter: @delawski @kienstra Adding a `layout` to an HTML `img` element makes me nervous. I think it would be better actually to introduce an `data-amp-layout` attribute which the sanitizer could convert into a `layout` attribute when converting an `img` into an `amp-img`. This would allow for regular plain HTML in post content to be decorated with the `data-amp-layout` attribute and for the content to be still served as valid HTML. This will also allow for such `img[data-amp-layout]` to be styled with CSS in non-AMP responses. This would play nicely into adding AMP support for Gutenberg blocks as described in https://github.com/Automattic/amp-wp/pull/902#issuecomment-361310795

We'll need to allow `data-amp-layout` attribute to Kses for it to be saved with `unfiltered_html`.

includes/utils/class-amp-wp-utils.php
```diff
@@ -58,4 +58,28 @@ protected static function _wp_translate_php_url_constant_to_key( $constant ) {
...
+	 * @since 0.7
+	 *
+	 * @param array  $context      Allowed tags and their allowed attributes.
+	 * @param string $context_type Type of context.
+	 * @return array $context      Filtered allowed tags and attributes.
+	 */
+	public static function add_layout( $context, $context_type ) {
+		if ( 'post' !== $context_type ) {
+			return $context;
+		}
```
kienstra: Is it alright to only filter the allowed attributes [in a 'post' context](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/kses.php#L850)?

includes/utils/class-amp-wp-utils.php
```diff
@@ -58,4 +58,28 @@ protected static function _wp_translate_php_url_constant_to_key( $constant ) {
...
+		$img            = isset( $context['img'] ) ? $context['img'] : array();
+		$context['img'] = array_merge(
+			$img,
+			array(
+				'data-amp-layout' => true,
+			)
+		);
+		return $context;
+	}
+
```
westonruter: This shouldn't be added to `AMP_WP_Utils` since it is marked for deletion in #876.

includes/sanitizers/class-amp-img-sanitizer.php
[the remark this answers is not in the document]
kienstra: Thanks, @westonruter. The 2 commits below convert `'data-amp-layout'` to  `'layout'` in the image sanitizer, and allow `'data-amp-layout'` in `wp_kses()`, when in a `'post'` context.

includes/utils/class-amp-wp-utils.php
westonruter: I'd say it should be allowed in any context where an `img` is allowed with a `width` and `height`.

kienstra:
**Will Resolve Merge Conflicts**

Next, I'll merge in `develop` and resolve the conflicts.

includes/utils/class-amp-wp-utils.php
```diff
@@ -58,4 +58,28 @@ protected static function _wp_translate_php_url_constant_to_key( $constant ) {
...
+		if ( 'post' !== $context_type ) {
+			return $context;
+		}
+		$img            = isset( $context['img'] ) ? $context['img'] : array();
+		$context['img'] = array_merge(
+			$img,
+			array(
+				'data-amp-layout' => true,
+			)
+		);
```
westonruter: So this can be changed to:

```php
if ( ! empty( $context['img']['width'] ) && ! empty( $context['img']['height'] ) ) {
    $context['img']['data-amp-layout'] = true;
}
```

includes/utils/class-amp-wp-utils.php
kienstra: Thanks, how about [AMP_HTML_Utils](https://github.com/Automattic/amp-wp/blob/develop/includes/utils/class-amp-html-utils.php)?

westonruter: Let's just put it in `AMP_Theme_Support` for now. It's the only class that is using it, and we'll probably break up that class into separate classes later anyway.

includes/utils/class-amp-wp-utils.php
kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/pull/937/commits/a233b59e1d4088eae1f614fc4ed856c700959558) uses your code in that filter.

includes/utils/class-amp-wp-utils.php
kienstra: Thanks, I'll move it to `AMP_Theme_Support`.

includes/utils/class-amp-wp-utils.php
kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/pull/937/commits/a233b59e1d4088eae1f614fc4ed856c700959558) allows `data-amp-layout` whenever an `` permits `width` and `height`.

includes/utils/class-amp-wp-utils.php
kienstra: This [commit](https://github.com/Automattic/amp-wp/pull/937/commits/011e47838e5154bb77af4f36010ff56d8bedd9eb) moves `add_layout()` to `AMP_Theme_Support`.

includes/class-amp-theme-support.php
```diff
@@ -833,4 +834,20 @@ public static function prepare_response( $response, $args = array() ) {
...
+
+	/**
+	 * Adds 'data-amp-layout' to the allowed <img> attributes for wp_kses().
+	 *
+	 * @since 0.7
+	 *
+	 * @param array $context Allowed tags and their allowed attributes.
+	 * @return array $context Filtered allowed tags and attributes.
+	 */
+	public static function add_layout( $context ) {
```
westonruter: I think this should be named a bit more explicitly, like: `whitelist_layout_in_wp_kses_allowed_html()`.

kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/pull/937/commits/e344418132bc56fbf4d0c6083f399061f0ab4976) renames the function to `whitelist_layout_in_wp_kses_allowed_html()`.

westonruter:
@delawski Would you give the new changes a spin?

delawski:
@westonruter Sure, I've [replaced `layout` with `data-amp-layout` ](https://github.com/xwp/ampconf/pull/81/commits/ca9e3fe9fba86a64acbca8e65500a550887ef8c0)and it works like a charm!

westonruter:
@delawski I deployed the latest changes to https://sizesfix-ampconfdemo.pantheonsite.io/

I'm seeing some strangeness compared to https://dev-ampconfdemo.pantheonsite.io/

The header logo is vertically stretched, and the image crops don't seem to be right.

delawski:
@westonruter I will investigate. There's no such issue on my local, that's why I've missed it. I'll check what might be wrong.

delawski:
@westonruter @kienstra So it turns out that for a larger logo file the aspect ratio of the image was not preserved. I've updated the CSS so that this should not be the case for the logo.

However, the bigger issue I see is that all the images that are used in the post content don't have any `layout` value set. I think that setting `layout="responsive"` would be a good default value to use. Otherwise the `layout="fixed"` value is set by AMP which is not the way the images are sized on the non-AMP pages.

When it comes to the featured article image, the aspect ratio / crop you can see on https://dev-ampconfdemo.pantheonsite.io/ is incorrect according to the mock-ups in Figma. In the previous implementation it was one of the limitations we didn't really have time to overcome. All images were using the same aspect ratio. Now, as we use `object-fit` and `layout="fill"` for the posts' featured images, we can use the correct crops.

kienstra:
**Possible Approach To Adding layout="responsive"**

We could do something [similar to AMP_Video_Sanitizer](https://github.com/Automattic/amp-wp/pull/937/files#diff-7f309eb76f4897b98188810f9479b4b8R54) for [AMP_Img_Sanitizer::adjust_and_replace_node()](https://github.com/Automattic/amp-wp/blob/develop/includes/sanitizers/class-amp-img-sanitizer.php#L204):

```diff
private function adjust_and_replace_node( $node ) {
        $old_attributes = AMP_DOM_Utils::get_node_attributes_as_assoc_array( $node );
	$new_attributes = $this->filter_attributes( $old_attributes );
	$new_attributes = $this->enforce_sizes_attribute( $new_attributes );
+	if ( empty( $new_attributes['layout'] ) && ! empty( $new_attributes['height'] ) && ! empty( $new_attributes['width'] ) ) {
+		$new_attributes['layout'] = 'responsive';
+	}
```

westonruter:
I'm not totally aware of all the `layout` intricacies here so it's hard for me to evaluate the differences. I think that @mehigh is going to chime in here to confirm with any additional insights. 

I will note that in the Twenty Ten theme, _without_ this branch checked out, the header image is *incorrectly* sized:

image

```html
<amp-img src="https://src.wordpress-develop.test/wp-content/themes/twentyten/images/headers/path.jpg" width="940" height="198" alt="" sizes="(min-width: 640px) 640px, 100vw" class="amp-wp-enforced-sizes"></amp-img>
```

But _with_ this branch checked out, it is sized *correctly*:

image


```html
<amp-img src="https://src.wordpress-develop.test/wp-content/themes/twentyten/images/headers/path.jpg" width="940" height="198" alt="" class="amp-wp-enforced-sizes"></amp-img>
```

With the AMP plugin deactivated, it looks like:

image

```html
<img src="https://src.wordpress-develop.test/wp-content/themes/twentyten/images/headers/path.jpg" width="940" height="198" alt="" />
```

So aside from the margin issue, this branch renders things the same as it should without AMP (which is great 👍).

If the change as proposed by @kienstra is applied then it renders _very wrong_:

image

```html
<amp-img src="https://src.wordpress-develop.test/wp-content/themes/twentyten/images/headers/path.jpg" width="940" height="198" alt="" layout="responsive" class="amp-wp-enforced-sizes"></amp-img>
```

kienstra:
**Question About Layout**

Hi @mehigh,
Could you please help with the [discussion here](https://github.com/Automattic/amp-wp/pull/937/#issuecomment-370415351) about the `layout`?

This PR [strips the sizes attribute](https://github.com/Automattic/amp-wp/pull/937/files#diff-20fe23f4cf9210c1bde24280f116244aL122) from `<amp-img>` tags.

Before, when the `sizes` attribute was present, I think [AMP inferred a responsive layout](https://www.ampproject.org/docs/design/responsive/control_layout).

With this PR, I think fewer images are getting a layout of `responsive`. 

@delawski suggested [above](https://github.com/Automattic/amp-wp/pull/937/#issuecomment-370415351) that we might want to add `layout="responsive"`

Any help you could lend to this would be great. Thanks a lot!

mehigh:
Hi,

It took me a while to get my head around this, but I got to a solution.

Here's the PR issued against this branch (as I don't have push access to this repo):
https://github.com/Automattic/amp-wp/pull/1045

This is the specific commit fixing it:
https://github.com/Automattic/amp-wp/pull/1045/commits/103dd811444b18122bd4254a2bc6b81d51a963b7

And I've also merged in develop and handled the merge conflicts.

Here's a video showing the fix:
http://cloud.urldocs.com/3L340Z0I0l0f

Thanks for being patient.

kienstra:
**Thanks**
*Question About Doing Same Thing For Other Elements*

Hi @mehigh,
Thanks a lot for your [commit](https://github.com/Automattic/amp-wp/commit/103dd811444b18122bd4254a2bc6b81d51a963b7) to fix the image size issue. If it's alright, I cherry-picked it into this pull request. I had to change the target branch of this PR to `0.7` for the upcoming release.

You already did us a big favor by working on this so quickly, so no pressure if you can't get to this.

But does the [commit I made above](https://github.com/Automattic/amp-wp/pull/937/commits/02db2293d2d13fb3c8c64a6fe56461581c92bd1a) look reasonable? It does something similar to what you did, but for `<amp-video>` and `<amp-iframe>`.

This could prevent the `<amp-iframe>` from expanding to fill its container. Before this PR, most embed iframes had a height and width, so I think AMP [inferred a layout](https://www.ampproject.org/docs/design/responsive/control_layout#what-if-the-layout-attribute-isn%E2%80%99t-specified?) of `fixed`. And they wouldn't expand to fill their container. 

By using `layout: intrinsic`, it looks like this avoids the iframe filling the container:

layout-intrinsic

To test this, I made this change to `AMP_Base_Sanitizer::set_layout()` (only for testing):
```diff
	public function set_layout( $attributes ) {
+		$attributes['width']  = 300;
+		$attributes['height'] = 300;
		if ( empty( $attributes['height'] ) ) {
```

And pasted this in the post content:
```
<h1>Amazon Com Smile Embed</h1>
https://smile.amazon.com/dp/B00DPM7TIG
<h1>Amazon Co Smile Embed</h1>
https://smile.amazon.co.uk/dp/B00DPM7TIG
```

mehigh:
@kienstra 
amp-video with intrinsic works really well and it's even recommended in the docs, so no comments there.

As for iframes, there isn't a one-size-fits-all solution. Some iFrames can be responsive (hence intrinsic works for these types of iframes allowing them to grow and occupy the available space), and some can be fixed, it all depends on what rules are inside of that iFrame, data we can't know from outside.
And knowing that, intrinsic makes more sense than responsive for amp-iframe as well, but there can be scenarios where a fixed layout would make more sense. 
That's why we need the iframes to receive a fixed-height layout when no width is present. Think the FB like button (it used to be an iframe eons ago) -> that was not responsive, as the button didn't grow to 400px width if you would have placed it in such a large container. 
As long as it's catered for iframes with height-only dimension provided, I'm good with the default being intrinsic.

kienstra:
**Thanks!**

Hi @mehigh,
Thanks a lot for your [comments above](https://github.com/Automattic/amp-wp/pull/937#issuecomment-377679021) about `<amp-iframe>` and `<amp-video>`.

>That's why we need the iframes to receive a fixed-height layout when no width is present. 

Thanks, that helps a lot clarify this. [This line on this PR](https://github.com/Automattic/amp-wp/pull/937/commits/02db2293d2d13fb3c8c64a6fe56461581c92bd1a#diff-b71112cf4b967ce0ca2fb315ef3245c3L225) ensures that `amp-iframes` with no `width` have `layout="fixed-height"`.

kienstra:
**Request For Final Review**

Hi @westonruter,
It's a holiday weekend, so there's no hurry to review this. But when you have a chance, you please give this a final review?

@mehigh's comments and [commit](https://github.com/Automattic/amp-wp/commit/103dd811444b18122bd4254a2bc6b81d51a963b7) here helped a lot to clarify this PR. All of his points are applied. 

Also, this [existing assertion](https://github.com/Automattic/amp-wp/blob/develop/tests/test-amp-iframe-sanitizer.php#L31) ensures that `amp-iframes` with no `width` have `layout="fixed-height"`.

The [snipped that I posted here](https://github.com/Automattic/amp-wp/pull/937#issuecomment-370893403) that turned out to not work was luckily never part of the diff of this PR.

This looks good so far in my testing. Of course, we'll also validate all of this in the 3 testing environments for `0.7`.

kienstra:
**Request For Final Review**

Hi @westonruter,
It's a holiday weekend, so there's no hurry to review this. But when you have a chance, you please give this a final review?

@mehigh's comments and [commit](https://github.com/Automattic/amp-wp/commit/103dd811444b18122bd4254a2bc6b81d51a963b7) here helped a lot to clarify this PR.

Also, this [existing assertion](https://github.com/Automattic/amp-wp/blob/develop/tests/test-amp-iframe-sanitizer.php#L31) ensures that `amp-iframes` with no `width` have `layout="fixed-height"`.

The [snipped that I posted](https://github.com/Automattic/amp-wp/pull/937#issuecomment-370893403) which turned out to not work was luckily never part of the diff.

It turns out that my idea of using layout=“intrinsic” for `<amp-video>` also doesn’t work, as the <amp-video> [specification](https://www.ampproject.org/docs/reference/components/amp-video) doesn’t allow this. I [reverted it](https://github.com/Automattic/amp-wp/pull/937/commits/f5a3c7380abdc427abd2eae8e93be633ff2d7414). 

Strangely, this [document](https://www.ampproject.org/docs/design/responsive/control_layout) encourages using `intrinsic` for `<amp-video>`:

>This layout works very well for most AMP elements, including amp-img, amp-video, etc. 


Using layout=“responsive” for `<amp-video>` looks to preserve the size, which is good. For example, this video doesn’t fill its container, and retains its size:
```
[video src=https://videos.files.wordpress.com/DK5mLrbr/video-ca6dc0ab4a_hd.mp4 width=400 height=300]
```
Of course, we'll also validate all of this in the 3 testing environments for `0.7`.

mehigh:
The mixed messages around intrinsic being viable or not for amp-video is odd.

PR #942, opened 2018-02-08
Fix comments not showing due to caching

### LINKED ISSUE #939: Comments posted not appearing.
When posting a comment as a logged in user, the comment is not picked up and no load button is shown. Reloading the page does show the comment though.

Public posted comments are also shown as expected.

DavidCramer:
in #939 new the comments after the second poll due to the timestamp not changing.
the polled url contains the timestamp of the latest item or latest updated item.
the URL is like this `amp_latest_update_time=1518071595` on the first poll, a site with a cache sill see the timestamp as a unique url and deliver a fresh page. If there are no new comments, or updated comments, the next poll will be exactly the same.

In some caching plugins, you can set it to clear on update of content ( not sure if this includes a comment ) On Server based caching, this might be even more of a problem.

This change sets the `data-update-time` on the last comment to the current timestamp. This will force the next poll to use this timestamp in the polled URL.

Fixes #939.


westonruter:
@DavidCramer I'm not sure I understand the solution. If the problem is the full-cache (which shouldn't apply for a logged-in user) then if I leave a comment and then wait for 30 seconds after the second poll has been made, then if I post another comment will it not still use the same `updated` time as the last time and result in the cached response being returned?

Your right that the full-page cache should be getting flushed when new comments are posted, and that the full-page cache should exclude `amp_latest_update_time` from consideration when caching the URL.

westonruter:
Closing for now per https://github.com/Automattic/amp-wp/issues/939#issuecomment-378020664

PR #951, opened 2018-02-10
Update sanitization reporting

westonruter:
Follow-up on #912 for #843.

- [x] Eliminate sanitization reporting during `save_post` action in favor of doing it inline when loading the edit post screen.
- [x] Make sure that AMP embeds are run prior to sanitization.

This PR could also include parts from #842 if we want.


westonruter:
The fix to embeds here should prevent this from happening on #954:

image

PR #952, opened 2018-02-10
Add customizer theme support

### LINKED ISSUE #901: Add Customizer integrations when AMP theme support is present
Live preview is not working in the Customizer preview when a page is rendered as AMP.

westonruter:
- [x] Make sure sanitizer is run in rendered partials.
- [x] ~Add AMP component scripts when selective refresh results in a new component added.~ Irrelevant since AMP apparently does this automatically.
- [x] Make sure scripts are allowed to be output when `is_customize_preview()`. Scripts need to be enqueued regardless of the user being in the `iframe` since frontend previewing depends on JS injecting the `customize_changeset_uuid` into requests.
- [x] Suspend CSS sanitizer to facilitate JS-based live previews.
- [x] Suspend limit on CSS size so that `customize-preview.css` can be included.

Fixes #901.


westonruter:
@DavidCramer FYI: Note the re-organization here for form handling. This will probably impact https://github.com/Automattic/amp-wp/issues/936

DavidCramer:
Thanks @westonruter .

includes/class-amp-theme-support.php
```diff
@@ -178,6 +207,7 @@ public static function register_hooks() {
 		 */
 		add_action( 'template_redirect', array( __CLASS__, 'start_output_buffering' ), 0 );
 
+		// Commenting hooks.
```
ThierryA: During the upcoming cleanup, I would suggest to move comments, widgets etc. in their on classes.

includes/class-amp-theme-support.php
```diff
@@ -247,13 +277,100 @@ public static function purge_amp_query_vars() {
...
+		if ( empty( self::$purged_amp_query_vars['__amp_source_origin'] ) ) {
 			return;
 		}
 
+		if ( isset( $pagenow ) && 'wp-comments-post.php' === $pagenow ) {
+			// We don't need any data, so just send a success.
+			add_filter( 'comment_post_redirect', function() {
+				// We don't need any data, so just send a success.
+				wp_send_json_success();
+			}, PHP_INT_MAX, 2 );
```
ThierryA: Why passing 2 args here, I believe this is perhaps a left over?

includes/class-amp-theme-support.php
```diff
@@ -247,13 +277,100 @@ public static function purge_amp_query_vars() {
...
+			self::handle_xhr_headers_output();
+		} elseif ( ! empty( self::$purged_amp_query_vars['_wp_amp_action_xhr_converted'] ) ) {
+			add_filter( 'wp_redirect', array( __CLASS__, 'intercept_post_request_redirect' ), PHP_INT_MAX, 2 );
```
ThierryA: Why passing 2 args here, I believe this is perhaps a left over?

includes/class-amp-theme-support.php
```diff
@@ -721,6 +838,27 @@ protected static function ensure_required_markup( DOMDocument $dom ) {
...
+	 * Prevent enqueueing customize-preview styles if not in customizer preview iframe.
+	 * These are only needed for when there is live editing of content, such as selective refresh.
+	 *
+	 * @since 0.7
+	 */
+	public static function dequeue_customize_preview_scripts() {
+
+		// Dequeue styles unnecessary unless in customizer preview iframe when editing (such as for edit shortcuts).
+		if ( ! self::is_customize_preview_iframe() ) {
+			wp_dequeue_style( 'customize-preview' );
```
ThierryA: What about just doing `wp_deregister_style( 'customize-preview' );` so that we don't have to loop through dependencies?

westonruter: That could work, but it would also result in something like Query Monitor issuing a warning about a dependency missing.

PR #961, opened 2018-02-14
Remove AMP admin meta box from non-public post types

### LINKED ISSUE #957: AMP meta box is active on post types that are not public
The AMP plugin adds the meta box in the back-end for post types that are not public. A good example I stumbled upon now while in the BuddyPress email area.

BuddyPress uses post types to manager its emails. The post types are not public. Still, in the admin area, one can see the notice that AMP is not active because of post type support.

post-type-support

Clicking on the link sends the user to the AMP settings page where the email post type, correctly, by the way, is not available for activation.

The verification made on `AMP_Post_Meta_Box::render_status()` that uses `post_supports_amp()` is not enough. There is a filter available that plugins can use, but I believe this should be fixed in the AMP plugin.

Theoretically, non-public post types will not be used in the front-end and so, will have no AMP support. So the meta box in the back-end is unnecessary.

What do you guys think? :)

renatonascalves:
Fixes #957.

### DIFF
--- a/includes/admin/class-amp-post-meta-box.php
+++ b/includes/admin/class-amp-post-meta-box.php
@@ -120,9 +120,9 @@ class AMP_Post_Meta_Box {
 		$validate = (
 			isset( $screen->base )
 			&&
-			is_post_type_viewable( $post->post_type )
-			&&
 			'post' === $screen->base
+			&&
+			is_post_type_viewable( $post->post_type )
 		);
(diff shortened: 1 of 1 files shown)


renatonascalves:
@westonruter Could you review it when you have a chance? Thank you! :)

westonruter:
By any chance is your `DB_CHARSET` not set to `utf8`?

PR #971, opened 2018-02-20
Add validation error reporting for plugins and themes

### LINKED ISSUE #842: Display error when a plugin is activated which tries to do something incompatible with AMP

### LINKED ISSUE #843: Incorporate AMP validator into post authorship flow, to alert if content within the post editor is not compatible with AMP

kienstra:
This pull request for #842 shows which plugin echoed invalid markup. 

validate-query-var

If the user has the `'edit_post'` capability, and the query var of `amp_validate` is present, it adds a response header. 

This has all of the data the the response on `post.php` has, in addition to a `'plugins_with_invalid_output'` value. This array has all of the plugins that output invalid markup.

But maybe this isn't the best way to display an error.

The main item remaining for me is the [failed Travis build](https://travis-ci.org/Automattic/amp-wp/builds/343686096?utm_source=github_status&utm_medium=notification). The tests seem to stop, without error, on PHP < 5.5. This might be related to the [closure](https://github.com/Automattic/amp-wp/pull/971/files#diff-7d234989187d87a372e2aa6aa1299a71R359).

Fixes #842.
Fixes #843.

### DIFF
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -77,9 +77,11 @@ abstract class AMP_Base_Sanitizer {
 	protected $root_element;
 
 	/**
-	 * The plugin or theme that is outputting markup, if any.
+	 * Stack of the plugin or theme that is outputting markup, if any.
 	 *
-	 * @var null|string
+	 * Each item in the stack is an array containing the 'type' (theme, plugin, mu-plugin) and the 'name' of the extension.
+	 *
+	 * @var array[][]
 	 */
 	public $current_sources = array();
 
@@ -336,10 +338,10 @@ abstract class AMP_Base_Sanitizer {
 		if ( isset( $this->args[ AMP_Validation_Utils::CALLBACK_KEY ] ) ) {
 			call_user_func( $this->args[ AMP_Validation_Utils::CALLBACK_KEY ],
 				array(
-					'node'   => $child,
-					'parent' => $parent,
-				),
-				$this->current_sources
+					'node'    => $child,
+					'parent'  => $parent,
+					'sources' => $this->current_sources,
+				)
 			);
 		}
 	}
@@ -365,10 +367,10 @@ abstract class AMP_Base_Sanitizer {
 				$element->removeAttributeNode( $attribute );
 				call_user_func( $this->args[ AMP_Validation_Utils::CALLBACK_KEY ],
 					array(
-						'node'   => $attribute,
-						'parent' => $element,
-					),
-					$this->current_sources
+						'node'    => $attribute,
+						'parent'  => $element,
+						'sources' => $this->current_sources,
+					)
 				);
 			}
 		} elseif ( is_string( $attribute ) ) {
@@ -379,22 +381,21 @@ abstract class AMP_Base_Sanitizer {
 	}
 
 	/**
-	 * Sets the current plugin that is outputting markup, if there is one.
+	 * Updates the captured current plugin that is outputting markup.
 	 *
 	 * @since 0.7
 	 *
-	 * @param DOMNode $node The node to check for the presence of a plugin in a comment.
+	 * @param DOMComment $node The node to check for the presence of a plugin in a comment.
 	 * @return void
 	 */
 	public function capture_current_source( $node ) {
-		preg_match( ':(before|after)\:(.*):s', $node->nodeValue, $matches );
-		if ( ! isset( $matches[1], $matches[2] ) ) {
+		if ( ! preg_match( '#(?P<closing>/)?(?P<type>theme|plugin|mu-plugin):(?P<name>.*)#s', $node->nodeValue, $matches ) ) {
 			return;
-		} elseif ( 'after' === $matches[1] ) {
-			$this->current_sources[] = $matches[2];
-		} elseif ( 'before' === $matches[1] ) {
+		}
+		if ( ! empty( $matches['closing'] ) ) {
 			array_pop( $this->current_sources );
+		} else {
+			$this->current_sources[] = wp_array_slice_assoc( $matches, array( 'type', 'name' ) );
 		}
 	}
-
 }
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -321,21 +321,19 @@ abstract class AMP_Base_Sanitizer {
 	 *
 	 * @since 0.7
 	 *
-	 * @param DOMNode|DOMElement $child The node to remove.
+	 * @param DOMNode|DOMElement $node The node to remove.
+	 * @param array              $args Additional args to pass to validation error callback.
+	 *
 	 * @return void
 	 */
-	public function remove_invalid_child( $child ) {
-		$parent = $child->parentNode;
-		if ( isset( $this->args['remove_invalid_callback'] ) ) {
-			call_user_func( $this->args['remove_invalid_callback'],
-				array(
-					'node'   => $child,
-					'parent' => $parent,
-				)
+	public function remove_invalid_child( $node, $args = array() ) {
+		if ( isset( $this->args['validation_error_callback'] ) ) {
+			call_user_func( $this->args['validation_error_callback'],
+				array_merge( compact( 'node' ), $args )
 			);
 		}
 		if ( empty( $this->args['disable_invalid_removal'] ) ) {
-			$child->parentNode->removeChild( $child );
+			$node->parentNode->removeChild( $node );
 		}
 	}
 
@@ -349,18 +347,21 @@ abstract class AMP_Base_Sanitizer {
 	 *
 	 * @param DOMElement     $element   The node for which to remove the attribute.
 	 * @param DOMAttr|string $attribute The attribute to remove from the element.
+	 * @param array          $args      Additional args to pass to validation error callback.
 	 * @return void
 	 */
-	public function remove_invalid_attribute( $element, $attribute ) {
-		if ( isset( $this->args['remove_invalid_callback'] ) ) {
+	public function remove_invalid_attribute( $element, $attribute, $args = array() ) {
+		if ( isset( $this->args['validation_error_callback'] ) ) {
 			if ( is_string( $attribute ) ) {
 				$attribute = $element->getAttributeNode( $attribute );
 			}
 			if ( $attribute ) {
-				call_user_func( $this->args['remove_invalid_callback'],
-					array(
-						'node'   => $attribute,
-						'parent' => $element,
+				call_user_func( $this->args['validation_error_callback'],
+					array_merge(
+						array(
+							'node' => $attribute,
+						),
+						$args
 					)
 				);
 				if ( empty( $this->args['disable_invalid_removal'] ) ) {
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -393,9 +393,9 @@ abstract class AMP_Base_Sanitizer {
 			return;
 		}
 		if ( ! empty( $matches['closing'] ) ) {
-			array_pop( $this->current_sources );
-		} else {
 			$this->current_sources[] = wp_array_slice_assoc( $matches, array( 'type', 'name' ) );
+		} else {
+			array_pop( $this->current_sources );
 		}
 	}
 }
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -54,6 +54,7 @@ abstract class AMP_Base_Sanitizer {
 	 *      @type array $amp_bind_placeholder_prefix
 	 *      @type bool $allow_dirty_styles
 	 *      @type bool $allow_dirty_scripts
+	 *      @type bool $disable_invalid_removal
 	 *      @type callable $remove_invalid_callback
 	 * }
 	 */
@@ -75,15 +76,6 @@ abstract class AMP_Base_Sanitizer {
 	 * @var DOMElement
 	 */
 	protected $root_element;
-
-	/**
-	 * Stack of the plugin or theme that is outputting markup, if any.
-	 *
-	 * Each item in the stack is an array containing the 'type' (theme, plugin, mu-plugin) and the 'name' of the extension.
-	 *
-	 * @var array[][]
-	 */
-	public $current_sources = array();
 
 	/**
 	 * AMP_Base_Sanitizer constructor.
@@ -334,15 +326,16 @@ abstract class AMP_Base_Sanitizer {
 	 */
 	public function remove_invalid_child( $child ) {
 		$parent = $child->parentNode;
-		$child->parentNode->removeChild( $child );
-		if ( isset( $this->args[ AMP_Validation_Utils::CALLBACK_KEY ] ) ) {
-			call_user_func( $this->args[ AMP_Validation_Utils::CALLBACK_KEY ],
+		if ( isset( $this->args['remove_invalid_callback'] ) ) {
+			call_user_func( $this->args['remove_invalid_callback'],
 				array(
-					'node'    => $child,
-					'parent'  => $parent,
-					'sources' => $this->current_sources,
+					'node'   => $child,
+					'parent' => $parent,
 				)
 			);
+		}
+		if ( empty( $this->args['disable_invalid_removal'] ) ) {
+			$child->parentNode->removeChild( $child );
 		}
 	}
 
@@ -359,43 +352,27 @@ abstract class AMP_Base_Sanitizer {
 	 * @return void
 	 */
 	public function remove_invalid_attribute( $element, $attribute ) {
-		if ( isset( $this->args[ AMP_Validation_Utils::CALLBACK_KEY ] ) ) {
+		if ( isset( $this->args['remove_invalid_callback'] ) ) {
 			if ( is_string( $attribute ) ) {
 				$attribute = $element->getAttributeNode( $attribute );
 			}
 			if ( $attribute ) {
-				$element->removeAttributeNode( $attribute );
-				call_user_func( $this->args[ AMP_Validation_Utils::CALLBACK_KEY ],
+				call_user_func( $this->args['remove_invalid_callback'],
 					array(
-						'node'    => $attribute,
-						'parent'  => $element,
-						'sources' => $this->current_sources,
+						'node'   => $attribute,
+						'parent' => $element,
 					)
 				);
+				if ( empty( $this->args['disable_invalid_removal'] ) ) {
+					$element->removeAttributeNode( $attribute );
+				}
 			}
-		} elseif ( is_string( $attribute ) ) {
-			$element->removeAttribute( $attribute );
-		} else {
-			$element->removeAttributeNode( $attribute );
-		}
-	}
-
-	/**
-	 * Updates the captured current plugin that is outputting markup.
-	 *
-	 * @since 0.7
-	 *
-	 * @param DOMComment $node The node to check for the presence of a plugin in a comment.
-	 * @return void
-	 */
-	public function capture_current_source( $node ) {
-		if ( ! preg_match( '#(?P<closing>/)?(?P<type>theme|plugin|mu-plugin):(?P<name>.*)#s', $node->nodeValue, $matches ) ) {
-			return;
-		}
-		if ( ! empty( $matches['closing'] ) ) {
-			$this->current_sources[] = wp_array_slice_assoc( $matches, array( 'type', 'name' ) );
-		} else {
-			array_pop( $this->current_sources );
+		} elseif ( empty( $this->args['disable_invalid_removal'] ) ) {
+			if ( is_string( $attribute ) ) {
+				$element->removeAttribute( $attribute );
+			} else {
+				$element->removeAttributeNode( $attribute );
+			}
 		}
 	}
 }
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -283,10 +283,7 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 	private function process_node( $node ) {
 
 		// Don't process text or comment nodes.
-		if ( XML_COMMENT_NODE === $node->nodeType ) {
-			$this->capture_current_source( $node );
-			return;
-		} elseif ( XML_TEXT_NODE === $node->nodeType || XML_CDATA_SECTION_NODE === $node->nodeType ) {
+		if ( XML_TEXT_NODE === $node->nodeType || XML_COMMENT_NODE === $node->nodeType || XML_CDATA_SECTION_NODE === $node->nodeType ) {
 			return;
 		}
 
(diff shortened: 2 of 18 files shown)


westonruter:
Figuring out _where_ to display validation errors is tough. Normally this is something that we'd want to show in the admin bar, but this isn't available to us… yet. I think it should be a goal for us to get the admin bar to be fully supported, and then we can show the errors in it. But otherwise, what we may want to do is introduce a new admin screen for “AMP status”. This could eventually interface with the Google Search Console, maybe, to get an external view of AMP validity. This can be augmented with an _internal_ view of validity, being populated with the results of what you are building here. 

Maybe what we should do is introduce a new `amp_validation_error` post type to store logs of such errors. Such a post could include the error, the URL where the error occurred, and plugin/theme responsible, and so on. Then upon plugin activation, a request could be made back to the homepage with the query var auth key to request a validation check, and instead of the errors being returned as an HTTP response header they could be added as post type entries. (Warnings could be displayed at this point as well.) This process could then also be done at WP Cron to regularly check the homepage and maybe the most recently-published post. 

By having the errors logged then a user could be informed of them asynchronously, and with the original URL being present, the validation errors could be re-checked after the fact. There could be a button like “Re-check”.

Duplicates should be avoided. So I suggest storing the validation error and the responsible plugin/theme in the post content and then taking the data and creating an md5 hash which then can be used as the `post_name`. If such a post already exists with that `post_name`, then no new post would be inserted, but rather perhaps a postmeta for that post could be modified to amend the URL for where the validation error is encountered.

Eventually we'll want to include a WP-CLI command to crawl the entire site and find all validation errors with these being all logged.

includes/utils/class-amp-validation-utils.php
```diff
@@ -483,4 +487,62 @@ public static function add_validation_callback( $sanitizers ) {
...
+		if ( ! isset( $post_id ) ) {
+			$post_id = wp_insert_post( array(
+				'post_type'    => self::POST_TYPE_SLUG,
+				'post_name'    => $post_name,
+				'post_content' => $encoded_errors,
+			) );
+		}
+
+		// Store the URL where the error(s) appeared.
+		$meta = get_post_meta( $post_id, self::URLS_VALIDATION_ERROR, true );
```
westonruter: Instead of `true` to return a single meta and then append to it with `update_post_meta()` this could instead skip straight to doing `add_post_meta( $post_id, self::URLS_VALIDATION_ERROR, $url, true )`.

includes/utils/class-amp-validation-utils.php
```diff
@@ -483,4 +487,62 @@ public static function add_validation_callback( $sanitizers ) {
 		return $sanitizers;
 	}
 
+	/**
+	 * Registers the post type to store the validation errors.
+	 *
+	 * @return void.
+	 */
+	public static function post_type() {
```
westonruter: Let's keep method names verby. So this should be `register_post_type`

includes/utils/class-amp-validation-utils.php
```diff
@@ -272,4 +461,88 @@ public static function display_error( $response ) {
 		echo '</div>';
 	}
 
+	/**
+	 * Whether to validate the front end response.
+	 *
+	 * @return boolean
+	 */
+	public static function do_validate_front_end() {
```
westonruter: The “do” here sounds like it is doing doing something, but actually it's about whether it _should_ do something. So I suggest renaming this to `should_validate_frontend`, maybe.

includes/utils/class-amp-validation-utils.php
```diff
@@ -272,4 +461,88 @@ public static function display_error( $response ) {
...
+	 * @return void.
+	 */
+	public static function post_type() {
+		register_post_type(
+			self::POST_TYPE_SLUG,
+			array(
+				'labels'   => array(
+					'name' => _x( 'AMP Validation Errors', 'post type general name', 'amp' ),
+				),
+				'supports' => false,
```
westonruter: This should also have `'public' => false`, no?

includes/utils/class-amp-validation-utils.php
```diff
@@ -41,16 +83,39 @@ class AMP_Validation_Utils {
...
+	 * Gets the source of the asset.
+	 *
+	 * This attempts to find if it's from a theme or a plugin.
+	 * In that case, it also returns the name of the theme or plugin.
+	 *
+	 * @param string $asset The asset path to search for the source.
+	 * @return array $source The asset type and source.
+	 */
+	public static function get_source( $asset ) {
+		preg_match( ':wp-content/(themes|plugins|)/(.+?)/:s', $asset, $matches );
```
westonruter: Since `WP_CONTENT_DIR` can be customized, I think this shouldn't be explicitly looking for `wp-content`.

westonruter: Instead of `(.+?)` it may probably be better as `([^/]+)`

westonruter: I notice `(themes|plugins|)` has a bar with nothing after it. Shouldn't it be `(themes|plugins)`?

westonruter: Finally, both the themes dir and plugins dir can be customized via `WP_PLUGIN_DIR`. So this should be looking at it as well.

westonruter: Likewise, `WPMU_PLUGIN_DIR` for `mu-plugins`.

westonruter: Also, `mu-plugins` needs to be accounted for.

includes/utils/class-amp-validation-utils.php
```diff
@@ -231,14 +301,133 @@ public static function validate_content( $post ) {
...
+				$result = call_user_func_array( $function, array() );
+			} elseif ( $accepted_args >= func_num_args() ) {
+				$result = call_user_func_array( $function, $args );
+			} else {
+				$result = call_user_func_array( $function, array_slice( $args, 0, intval( $accepted_args ) ) );
+			}
+			$output = ob_get_clean();
+
+			if ( ! empty( $output ) ) {
+				printf( '<!--before:%s-->', esc_attr( $callback['plugin'] ) );
```
westonruter: I think the source type should also be included in these comments. So it should be like `<!--before:plugin:jetpack-->`, for example. This would help in the case where a theme and plugin have the same name.

includes/utils/class-amp-validation-utils.php
```diff
@@ -231,14 +301,133 @@ public static function validate_content( $post ) {
...
+			} elseif ( is_object( $callback ) && ( 'Closure' === get_class( $callback ) ) ) {
+				$reflection = new ReflectionFunction( $callback );
+			}
+
+			$file = isset( $reflection ) ? $reflection->getFileName() : null;
+			if ( ! isset( $file ) ) {
+				return null;
+			}
+			$source_data = self::get_source( $file );
+			if ( 'plugins' === $source_data['type'] ) {
```
westonruter: What about `themes` and `mu-plugins`?

includes/utils/class-amp-validation-utils.php
```diff
@@ -231,14 +301,133 @@ public static function validate_content( $post ) {
...
+	public static function callback_wrappers() {
+		global $wp_filter;
+		if ( ! self::do_validate_front_end() ) {
+			return;
+		}
+		$pending_wrap_callbacks = array();
+		foreach ( $wp_filter as $filter_tag => $wp_hook ) {
+			foreach ( $wp_hook->callbacks as $priority => $callbacks ) {
+				foreach ( $callbacks as $callback ) {
+					$plugin = self::get_plugin( $callback['function'] );
```
westonruter: Also needs to account for `themes` and `mu-plugins`.

includes/utils/class-amp-validation-utils.php
```diff
@@ -231,14 +301,133 @@ public static function validate_content( $post ) {
...
+		}
+	}
+
+	/**
+	 * Gets the plugin of the callback, if one exists.
+	 *
+	 * @param string|array $callback The callback for which to get the plugin.
+	 * @return string|null $plugin   The plugin to which the callback belongs, or null.
+	 */
+	public static function get_plugin( $callback ) {
```
westonruter: This function seems like it should be merged with `get_source`, and it can then return an array including the source type (e.g. `plugin`) and the source name (e.g. `jetpack`).

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -225,6 +225,7 @@ private function process_node( $node ) {
 
 		// Don't process text or comment nodes.
 		if ( XML_TEXT_NODE === $node->nodeType || XML_COMMENT_NODE === $node->nodeType || XML_CDATA_SECTION_NODE === $node->nodeType ) {
+			$this->set_plugin_output( $node );
```
westonruter: Let's refactor this to:

```php
if ( XML_COMMENT_NODE === $node->nodeType ) {
    $this->set_plugin_output( $node );
    return;
} elseif ( XML_TEXT_NODE === $node->nodeType || XML_CDATA_SECTION_NODE === $node->nodeType ) {
    return;
}
```

And then remove the check inside the `set_plugin_output` method. And that method can be renamed to something like `capture_current_source` as this should be for plugins and themes.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -351,4 +358,25 @@ public function remove_invalid_attribute( $element, $attribute ) {
...
+	 */
+	public function set_plugin_output( $node ) {
+		if ( XML_COMMENT_NODE === $node->nodeType ) {
+			preg_match( ':(before|after)\:(.*):s', $node->nodeValue, $matches );
+			if ( ! isset( $matches[1], $matches[2] ) ) {
+				return;
+			} elseif ( 'after' === $matches[1] ) {
+				$this->current_plugin_output = $matches[2];
+			} elseif ( 'before' === $matches[1] ) {
+				$this->current_plugin_output = null;
```
westonruter: I think that `current_plugin_output` should actually be an array, a stack of sources. So this should do:

```
$this->current_source[] = $matches[2];
```

And then `above` it can just `array_pop`. This is because a theme could invoke a plugin and there can be nesting.

includes/utils/class-amp-validation-utils.php
kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/commit/4955686933b6aaef1807cc1e1901b8a52c438415) renames it to `register_post_type()`.

includes/utils/class-amp-validation-utils.php
kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/commit/4955686933b6aaef1807cc1e1901b8a52c438415) adds  `'public' => false`.

includes/utils/class-amp-validation-utils.php
kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/commit/fda14ea7f84fed6990d60571a629e013bb0053e4) renames the function.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
kienstra: Thanks, [these commits](https://github.com/Automattic/amp-wp/compare/fda14ea7f84fed6990d60571a629e013bb0053e4...3d1e6e3dd2a70ce4b7cb319f2bef0cfc56fc9856) use your code above to only send comment nodes to `set_plugin_output()`. They also rename that function to `capture_current_source()`, and remove a conditional inside it.

includes/utils/class-amp-validation-utils.php
kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/commit/66ab20d3cae169cec044ea26331bec2362eb0322) moves `get_plugin()` into `get_source()`. As you mentioned, it returns an array with the `'type'` and `'source'`.

includes/sanitizers/class-amp-base-sanitizer.php
kienstra: Thanks, [commit](https://github.com/Automattic/amp-wp/pull/971/commits/32a58fcd21710124cf563e457880f0e80a7fb4d5) renames `$current_plugin_output` to `$current_sources`. And it makes it into an array (stack).

includes/utils/class-amp-validation-utils.php
kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/commit/87df66ac81f186c05e2e07ad849f96b86836b142) begins with `add_post_meta()`, and only runs `update_post_meta()` if there was already a value for that key. Is that what you had in mind?

includes/utils/class-amp-validation-utils.php
kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/pull/971/commits/66ab20d3cae169cec044ea26331bec2362eb0322#diff-7d234989187d87a372e2aa6aa1299a71R370) changes the regex.

kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/pull/971/commits/66ab20d3cae169cec044ea26331bec2362eb0322#diff-7d234989187d87a372e2aa6aa1299a71R370) changes the regex. I still need to accomodate `mu-plugins`.

includes/utils/class-amp-validation-utils.php
kienstra: Sure, [this commit](https://github.com/Automattic/amp-wp/commit/76ec4dfc770305a0624e16ffe9ec0ee84729d732) adds the type of source to the comments, like you suggested.

includes/utils/class-amp-validation-utils.php
kienstra: Good points. [This commit](https://github.com/Automattic/amp-wp/commit/ea16d64c6284a9b68522e9d7f1aa2b031f94bbd2) looks for themes, plugins, and mu-plugins. There's probably room to improve this, though.

includes/utils/class-amp-validation-utils.php
kienstra: This now accommodates themes and plugins, via the new function [get_source()](https://github.com/Automattic/amp-wp/pull/971/files#diff-7d234989187d87a372e2aa6aa1299a71R349).

includes/utils/class-amp-validation-utils.php
kienstra: This [now accommodates](https://github.com/Automattic/amp-wp/pull/971/files#diff-7d234989187d87a372e2aa6aa1299a71R349) themes and mu-plugins.

kienstra:
**Applied Review Suggestions**

Hi @westonruter,
Thanks for your suggestions here. I applied all of them, though the work to apply your [solution design](https://github.com/Automattic/amp-wp/pull/971#issuecomment-367180550) is still in progress.

includes/utils/class-amp-validation-utils.php
kienstra: [This commit](https://github.com/Automattic/amp-wp/commit/ea16d64c6284a9b68522e9d7f1aa2b031f94bbd2) accommodates `mu-plugins`.

kienstra:
**Working On Displaying Error on plugins.php**

Hi @westonruter,
I'm working on displaying errors on `/wp-admin/plugins.php` when a plugin has invalid content.

westonruter:
@kienstra Should it be on `plugins.php` or should it be on another AMP status page? Let's talk about where it would be best to show that. I think it's right that there should be some indicator on the plugins page for which plugins are reported to have invalid content, but it should probably link over to whatever AMP status page has a list of all the invalidities. Maybe we can re-use the post list table to list all of the AMP invalidation errors.

kienstra:
**plugins.php Page**

Hi @westonruter,
Yes, it'd be good if the `/wp-admin/plugins.php` page only had a simple indication that there's an error, and a link to the `'AMP Status'` page:

possible-warning-activating-plugin

You're probably right that we could reuse the [existing logic](https://github.com/Automattic/amp-wp/blob/develop/includes/utils/class-amp-validation-utils.php#L245) that outputs the post errors, with some refactoring.

kienstra:
**Possible UI For Validation Errors**

Hi @westonruter,
What do you think about this possible UI?

ui-amp-validation-errors

It could reuse [your logic](https://github.com/Automattic/amp-wp/blob/develop/includes/utils/class-amp-validation-utils.php#L260) for outputting the validation errors in the editor.

kienstra:
**Possible Change**

In the mockup above, maybe 'URLs With Validation Errors,' should be 'URLs With Incompatible Plugins Or Themes.' These errors will only appear if there's a validation issue from a plugin or theme. Not from post content.

includes/utils/class-amp-validation-utils.php
```diff
@@ -273,4 +501,185 @@ public static function display_error( $response ) {
...
+	 * After the preprocessors run, this gets the validation response if the query var is present.
+	 * It then stores the response in a custom post type.
+	 * If there's already an error post for the URL, but there's no error anymore, it deletes it.
+	 *
+	 * @return int|null $post_id The post ID of the custom post type used, or null.
+	 */
+	public static function store_validation_errors() {
+		if ( ! self::should_validate_front_end() ) {
+			return null;
+		}
```
kienstra: There should be a better way to limit who can make a request that stores validation errors. Right now, [should_validate_front_end()](https://github.com/Automattic/amp-wp/pull/971/files#diff-7d234989187d87a372e2aa6aa1299a71R510) checks for capability and the presence of the `validation` query var. 

But this needs to be able to make requests from the back-end. Like in [validate_home()](https://github.com/Automattic/amp-wp/pull/971/files#diff-7d234989187d87a372e2aa6aa1299a71R646), where it requests the front page on activating a plugin.

In that case, there won't be a logged-in user, as it's currently using [wp_remote_get()](https://github.com/Automattic/amp-wp/pull/971/files#diff-7d234989187d87a372e2aa6aa1299a71R647). Maybe there's a better way to make a request for a front-end page.

kienstra: There should be a better way to limit who can make a request that stores validation errors. Right now, [should_validate_front_end()](https://github.com/Automattic/amp-wp/pull/971/files#diff-7d234989187d87a372e2aa6aa1299a71R510) checks for capability and the presence of the `validation` query var. 

But this needs to be able to make requests from the back-end. Like in [validate_home()](https://github.com/Automattic/amp-wp/pull/971/files#diff-7d234989187d87a372e2aa6aa1299a71R646), where it requests the front page on activating a plugin.

In that case, there won't be a logged-in user, as it's currently using [wp_remote_get()](https://github.com/Automattic/amp-wp/pull/971/files#diff-7d234989187d87a372e2aa6aa1299a71R647). Maybe there's a better way to make a request for a front-end page.

westonruter: One alternative is what was implemented in 4.9 for the scraping of plugin activation errors which is very similar to the validation checks we're doing in AMP. See:

* https://github.com/WordPress/wordpress-develop/blob/6e5a560be0c01e236b8d43da3ac61c50f34297d0/src/wp-admin/includes/file.php#L496-L505
* https://github.com/WordPress/wordpress-develop/blob/6e5a560be0c01e236b8d43da3ac61c50f34297d0/src/wp-includes/load.php#L1211-L1254

I don't know if we should go the route of storing the nonce in a transient.

On the other hand, you can see in the same example that `wp_remote_get()` can be passed the user's current cookies to make the request as an authenticated user:

https://github.com/WordPress/wordpress-develop/blob/6e5a560be0c01e236b8d43da3ac61c50f34297d0/src/wp-admin/includes/file.php#L525-L540

So that would allow for the capability check to continue to be used.

westonruter:
> What do you think about this possible UI?

@kienstra What about re-using a post list table that lists out every validation error? There could be columns for listing out the invalid elements and attributes, as well as a column for the source and which extension is causing the problem. There can also be a row action to re-check, or else the errors can be bulk-deleted or bulk re-checked using the post list table UI.

kienstra:
**Possible Validation Page**

Hi @westonruter,
Good idea to display this in a post table. How does this look?

possible-validation-page

>as well as a column for the source and which extension is causing the problem. 

Would that be the AMP extension?

This reuses [your logic](https://github.com/Automattic/amp-wp/blob/develop/includes/utils/class-amp-validation-utils.php#L262-L271) from the editor for outputting the removed elements and attributes.

I have this partially applied in code, though there are several points remaining, like the re-check and the bulk actions.

includes/utils/class-amp-validation-utils.php
westonruter: I updated to pass cookies in fe0e6cc.

westonruter:
> Good idea to use a post table. How does this look?

Yes, that looks much better.

westonruter:
I'm working on adding shortcode support now.

kienstra:
Thanks, @westonruter. I'm working on implementing the 're-check' on `/wp-admin/plugins.php`.

kienstra:
**Cron Job Commits**

Hi @westonruter,
My [cron job commits](https://github.com/Automattic/amp-wp/pull/971/commits/7c5402ef580bb4ddb6e03a3e6a243bb073c236a1) above might need improvement. They use one of the [strategies you mentioned](https://github.com/Automattic/amp-wp/pull/971/#discussion_r171433334), storing a nonce in a transient. Though I'm not sure if it does that in the best way.

kienstra:
**post.php Meta Boxes**

Hi @westonruter,
If it's alright, I'm working on the `/wp-admin/post.php` meta boxes now.

westonruter:
@kienstra ¡Adelante!

kienstra:
**Meta Boxes On post.php**

Hi @westonruter,
Here's what the `/wp-admin/post.php` page looks like so far:

post-php-validation-page

I still need to add the list of URLs that have the error to the full-width meta box, and 'recheck' links for them.

kienstra:
**URLs on /wp-admin/post.php** 

Hi @westonruter,
Here's the display of URLs that the [commit above](https://github.com/Automattic/amp-wp/pull/971/commits/81dcb8e691d4563fac5355607afa6285e010c432) adds:

error-ui

westonruter:
With dd9d629 the AMP Validation Errors are shown in the admin menu:

image

With e6d12c8 the AMP Validation Errors are shown in the At a Glance widget:

image

westonruter:
- [ ] Fix redirect from amp endpoint to non-AMP when AMP is native.

PR #978, opened 2018-02-26
Refines handling of 0 and unset height/width attrs

davisshaver:
See https://github.com/Automattic/amp-wp/issues/974


davisshaver:
CI shows some warnings due to `=>` spacing in the update test file, however based on the existing spacing I think we want to ignore these. See how `=>` do not align in current https://github.com/Automattic/amp-wp/blob/develop/tests/test-amp-img-sanitizer.php#L15

Issue #974, opened 2018-02-25
<img> tag with decimal in width or height will fail validation as <amp-img>


davisshaver:
Need to investigate whether this is responsive images or AMP WP plugin. Can be assigned to me. Will write additional test case.

westonruter:
@davisshaver could you also please take a look at the case when `0` is supplied for the `width` or `height`? See https://wordpress.org/support/topic/amp-img-not-setting-height-and-width/

PR #979, opened 2018-02-26
Refines handling of 0 and empty height/width attributes

davisshaver:
Alternate branch of https://github.com/Automattic/amp-wp/pull/978#issuecomment-368572107 w/ spacing update.

Fixes https://github.com/Automattic/amp-wp/issues/974

### DIFF
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -174,11 +174,22 @@ abstract class AMP_Base_Sanitizer {
 	 * @return float|int|string Returns a numeric dimension value, or an empty string.
 	 */
 	public function sanitize_dimension( $value, $dimension ) {
-		if ( empty( $value ) ) {
+
+		/**
+		 * Allows 0 as valid dimension.
+		 */
+		if ( null === $value ) {
 			return '';
 		}
 
 		if ( false !== filter_var( $value, FILTER_VALIDATE_INT ) ) {
+			return absint( $value );
+		}
+
+		/**
+		 * Allows floats but parses them to integers.
+		 */
+		if ( false !== filter_var( $value, FILTER_VALIDATE_FLOAT ) ) {
 			return absint( $value );
 		}
 
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -175,22 +175,14 @@ abstract class AMP_Base_Sanitizer {
 	 */
 	public function sanitize_dimension( $value, $dimension ) {
 
-		/**
-		 * Allows 0 as valid dimension.
-		 */
+		// Allows 0 to be used as valid dimension.
 		if ( null === $value ) {
 			return '';
 		}
 
-		if ( false !== filter_var( $value, FILTER_VALIDATE_INT ) ) {
-			return absint( $value );
-		}
-
-		/**
-		 * Allows floats but parses them to integers.
-		 */
-		if ( false !== filter_var( $value, FILTER_VALIDATE_FLOAT ) ) {
-			return absint( $value );
+		// Accepts both integers and floats & prevents negative values.
+		if ( is_numeric( $value ) ) {
+			return max( 0, floatval( $value ) );
 		}
 
 		if ( AMP_String_Utils::endswith( $value, 'px' ) ) {
@@ -263,7 +255,12 @@ abstract class AMP_Base_Sanitizer {
 			$max_width = $this->args['content_max_width'];
 		}
 
-		$attributes['sizes'] = sprintf( '(min-width: %1$dpx) %1$dpx, 100vw', absint( $max_width ) );
+		// Allows floats and integers but prevents negative values.
+		// Uses string format to prevent additional modification.
+		$attributes['sizes'] = sprintf(
+			'(min-width: %1$spx) %1$spx, 100vw',
+			max( 0, floatval( $max_width ) )
+		);
 
 		$this->add_or_append_attribute( $attributes, 'class', 'amp-wp-enforced-sizes' );
 
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -165,8 +165,7 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 					$class = $node->hasAttribute( 'class' ) ? $node->getAttribute( 'class' ) . ' amp-wp-unknown-size' : 'amp-wp-unknown-size';
 					$node->setAttribute( 'class', $class );
 				} elseif (
-					! is_numeric( $node->getAttribute( 'height' )
-				) ) {
+					! is_numeric( $node->getAttribute( 'height' ) ) ) {
 					$height = self::FALLBACK_HEIGHT;
 					$node->setAttribute( 'height', $height );
 					$class = $node->hasAttribute( 'class' ) ? $node->getAttribute( 'class' ) . ' amp-wp-unknown-size amp-wp-unknown-height' : 'amp-wp-unknown-size amp-wp-unknown-height';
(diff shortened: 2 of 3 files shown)


westonruter:
@davisshaver in #974 you call out numbers with decimals specifically but I don't see any tests that have float dimensions.

davisshaver:
@westonruter See https://github.com/Automattic/amp-wp/pull/979/files#diff-40978c54c34050065a79a2c817f6b0e0R45

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -174,14 +174,25 @@ protected function get_body_node() {
 	 * @return float|int|string Returns a numeric dimension value, or an empty string.
 	 */
 	public function sanitize_dimension( $value, $dimension ) {
-		if ( empty( $value ) ) {
+
+		/**
```
westonruter: Minor thing, but this should be just:

```php
// Allows 0 as valid dimension.
```

Same for the comment below.

The `/**` comments are for phpdoc specifically, and there is no need for multi-line here.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -174,14 +174,25 @@ protected function get_body_node() {
...
 
 		if ( false !== filter_var( $value, FILTER_VALIDATE_INT ) ) {
 			return absint( $value );
 		}
 
+		/**
+		 * Allows floats but parses them to integers.
+		 */
+		if ( false !== filter_var( $value, FILTER_VALIDATE_FLOAT ) ) {
+			return absint( $value );
```
westonruter: Should this not be `floor( $value )`?

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -154,18 +154,29 @@ private function determine_dimensions( $need_dimensions ) {
...
 					$node->setAttribute( 'class', $class );
+				} elseif (
+					! is_numeric( $node->getAttribute( 'height' ) ) ) {
+					$height = self::FALLBACK_HEIGHT;
+					$node->setAttribute( 'height', $height );
+					$class = $node->hasAttribute( 'class' ) ? $node->getAttribute( 'class' ) . ' amp-wp-unknown-size amp-wp-unknown-height' : 'amp-wp-unknown-size amp-wp-unknown-height';
+					$node->setAttribute( 'class', $class );
+				} elseif (
+					! is_numeric( $node->getAttribute( 'width' ) )
+				) {
```
westonruter: Indentation of this `elseif` line is inconsistent with the previous multi-line `elseif`.

davisshaver:
@westonruter Updated, sry about that.

tests/test-amp-img-sanitizer.php
```diff
@@ -12,42 +12,62 @@ public function setUp() {
...
+
+			'image_with_empty_height'                  => array(
+				'<p><img src="http://placehold.it/300x300" width="300" height="" /></p>',
+				'<p><amp-img src="http://placehold.it/300x300" width="300" height="400" class="amp-wp-unknown-size amp-wp-unknown-height amp-wp-enforced-sizes" sizes="(min-width: 300px) 300px, 100vw"></amp-img></p>',
+			),
+
+			'image_with_zero_width'                    => array(
+				'<p><img src="http://placehold.it/300x300" width="0" height="300" /></p>',
+				'<p><amp-img src="http://placehold.it/300x300" width="0" height="300" sizes="(min-width: 0px) 0px, 100vw" class="amp-wp-enforced-sizes"></amp-img></p>',
+			),
```
westonruter: Please also include a test with a zero width _and_ height so we can ensure that the [support topic](https://wordpress.org/support/topic/amp-img-not-setting-height-and-width/) is covered:

```html
<img src="https://example.com/img.jpg" width="0" height="0" border="0">
```

davisshaver:
Thanks for the review, will clean this up tomorrow AM!

includes/sanitizers/class-amp-base-sanitizer.php
davisshaver: @westonruter Maybe? I am looking at [the docs for floor()](http://php.net/manual/en/function.floor.php) and [round()](http://php.net/manual/en/function.round.php). Both return floats. What about `return absint( round( $value, 0 ) )`?

westonruter: `absint( -1.5 )` is `1` but `floor( -1.5 )` is `-1`. In other words, we don't want to lose the sign, do we?

Somewhat strange in the docs there for `floor()`:

> Returns the next lowest integer value (as float)

So I suppose then this is what should be done:

```php
return (int) floor( $value );
```

The `@return` phpdoc should be updated to remove `float`?

westonruter: Correction: `floor( -1.5 )` is `2`, as is `floor( -1.1 )`. Maybe `round()` is better then:

```php
return (int) round( $value, 0 );
```

westonruter: @davisshaver Hold on. I don't see how decimals are invalid AMP:

image

davisshaver: Agree about updating the `@return` statement.

I'd lean towards `absint()` to match the behavior on values that are already `int`, otherwise we may pass a negative height/width.

westonruter: True, negative numbers are illegal. But decimal-point values _are_ allowed. So then I think it should be the following with the two conditionals being combined:

```php
if ( is_numeric( $value ) ) {
    return max( 0, floatval( $value ) );
}
```

davisshaver: Thanks for the discussion. I will update the PR tomorrow for another consideration

davisshaver:
@westonruter all set for another 👀 ! Much appreciated

tests/test-amp-img-sanitizer.php
westonruter: Could you add a test for this?

davisshaver: Done ✔️

sach-patel:
Found this while searching for a solution where images have fractional value for the height property. It is weird but it happened during migration from Drupal to WP. Is there any way to tackle such images? Thanks.

westonruter:
@sach-patel I wasn't able to see floats causing any validation errors.

westonruter:
@davisshaver I have a related bug if you're interested: https://github.com/Automattic/amp-wp/issues/1030

sach-patel:
@westonruter
Below is an example;
Post: https://thebrag.com/tldr-week-cinema-bad-cop-bad-cop/
AMP version: https://thebrag.com/tldr-week-cinema-bad-cop-bad-cop/amp/

img src="https://i.imgur.com/d1NBb20.gif" alt="" width="425" height="238.85" tag has been converted to amp-anim src="https://i.imgur.com/d1NBb20.gif" alt="" width="425" height="" sizes="(min-width: 425px) 425px, 100vw" class="amp-wp-enforced-sizes"

Google SC is reporting this as "Invalid value for attribute in layout 'height' in tag 'amp-anim'". Is there any quick solution or patch?

Thanks.

westonruter:
@sach-patel the fix is coming in the 0.7 version of the plugin. I suggest testing 0.7-beta2 which should be out this week.

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -154,18 +154,30 @@ private function determine_dimensions( $need_dimensions ) {
 				if ( ! $node instanceof DOMElement ) {
 					continue;
 				}
-
-				// Provide default dimensions for images whose dimensions we couldn't fetch.
-				if ( false !== $dimensions ) {
-					$node->setAttribute( 'width', $dimensions['width'] );
-					$node->setAttribute( 'height', $dimensions['height'] );
```
westonruter: These should not have been removed. See #1117.

PR #983, opened 2018-03-01
added host validation for urls

### LINKED ISSUE #977: Malformed url in links creates an invalid AMP page
If an editor ads a link like `<a href="http://foo%20bar">foo bar</a>` by mistake, the generated amp page will be invalid.

I think it would be great to add an AMP_Link_Sanitizer class that checks the urls of the links and remove the <a> tag if the url is malformed.

rubengonzalezmrf:
Fixes #977.


tests/test-tag-and-attribute-sanitizer.php
```diff
@@ -680,6 +680,11 @@ public function get_body_data() {
 				'<a class="foo" href="">value</a>',
 			),
 
+			'a_with_wrong_host'                                         => array(
+				'<a class="foo" href="http://foo bar">value</a>',
+				'<a class="foo" href="">value</a>',
+			),
+
```
westonruter: Please add test cases for examples found in https://github.com/Automattic/amp-wp/issues/977#issuecomment-369230272

includes/sanitizers/class-amp-allowed-tags-generated.php
```diff
@@ -28,6 +28,7 @@ class AMP_Allowed_Tags_Generated {
 						'value_url' => array(
 							'allow_empty' => true,
 							'allow_relative' => true,
+							'valid_host' => true,
```
westonruter: This isn't going to work since this file is generated from the spec via [`amphtml-update.py`](https://github.com/Automattic/amp-wp/blob/develop/bin/amphtml-update.py), so your change here will be lost.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -699,6 +699,19 @@ private function validate_attr_spec_list_for_node( $node, $attr_spec_list ) {
 				}
 			}
 
+			/*
+			 * If given attribute's value is a URL with a host, the host must
+			 * be valid
+			 */
+			if ( isset( $attr_spec_rule[ AMP_Rule_Spec::VALUE_URL ][ AMP_Rule_Spec::VALID_HOST ] ) ) {
```
westonruter: Instead of introducing `AMP_Rule_Spec::VALID_HOST` here, which isn't actually part of the spec, I suggest you validate the host whenever `isset( $attr_spec_rule[ AMP_Rule_Spec::VALUE_URL ] )`. Should it not be validated for every attribute that is a URL value?

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1122,6 +1138,40 @@ private function check_attr_spec_rule_value_regex_casei( $node, $attr_name, $att
 		return AMP_Rule_Spec::NOT_APPLICABLE;
 	}
 
+	/**
+	 * Check if attribute has a valid host value
+	 *
```
westonruter: Please add a `@since 0.7` here

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1122,6 +1138,40 @@ private function check_attr_spec_rule_value_regex_casei( $node, $attr_name, $att
...
+	 */
+	private function check_attr_spec_rule_valid_host( $node, $attr_name, $attr_spec_rule ) {
+		if ( isset( $attr_spec_rule[ AMP_Rule_Spec::VALID_HOST ] ) ) {
+			if ( $node->hasAttribute( $attr_name ) ) {
+				$attr_value   = $node->getAttribute( $attr_name );
+				$attr_value   = preg_replace( '/\s*,\s*/', ',', $attr_value );
+				$urls_to_test = explode( ',', $attr_value );
+
+				foreach ( $urls_to_test as $url ) {
+					$url_host = AMP_WP_Utils::parse_url( urldecode( $url ), PHP_URL_HOST );
```
westonruter: As noted in #876, `AMP_WP_Utils` is obsolete and you can use `wp_parse_url()` directly.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1122,6 +1138,40 @@ private function check_attr_spec_rule_value_regex_casei( $node, $attr_name, $att
...
+					if ( $url_host && preg_match( '/[!"#$%&\'()*+,\/:;<=>?@[\]^`{|}~\s]/i', $url_host ) ) {
```
westonruter: I should think that a whitelist should be used rather than a blacklist here.

rubengonzalezmrf: I tend to think like you, but the amp validator is doing it this way, looking for blacklisted chars, so i thought it would be better to follow the same approach.

westonruter: So this is supposed to match everything matched by `hostCharIsValid()` in https://github.com/ampproject/amphtml/blob/af1e3a550feeafd732226202b8d1f26dcefefa18/validator/engine/parse-url.js#L69-L103 ?

westonruter:
I'm getting the same failures locally.

westonruter:
If I add `post_content` that includes:

```html
<a class="foo" href="http://foo bar">value</a>

<a class="foo" href="mail to:foo@bar.com">value</a>
```

I get out:

```html
<p><a class="foo" href="">value</a></p>

<p><a class="foo" href="mail%20to:foo@bar.com">value</a></p>
```

So it seems the `mail to` check is failing.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1122,6 +1138,49 @@ private function check_attr_spec_rule_value_regex_casei( $node, $attr_name, $att
...
+
+				foreach ( $urls_to_test as $url ) {
+					// Check if the host contains invalid chars.
+					$url_host = wp_parse_url( urldecode( $url ), PHP_URL_HOST );
+					if ( $url_host && preg_match( '/[!"#$%&\'()*+,\/:;<=>?@[\]^`{|}~\s]/i', $url_host ) ) {
+						return AMP_Rule_Spec::FAIL;
+					}
+
+					// Check if the protocol contains invalid chars.
+					$url_scheme = AMP_WP_Utils::parse_url( $url, PHP_URL_SCHEME );
```
westonruter: Use `wp_parse_url()` instead of `AMP_WP_Utils::parse_url()` here.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1122,6 +1138,49 @@ private function check_attr_spec_rule_value_regex_casei( $node, $attr_name, $att
...
+	 *      - AMP_Rule_Spec::FAIL - $attr_name has a value that does *not* match rule.
+	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
+	 *                                        is no rule for this attribute.
+	 */
+	private function check_attr_spec_rule_valid_url( $node, $attr_name, $attr_spec_rule ) {
+		if ( isset( $attr_spec_rule[ AMP_Rule_Spec::VALUE_URL ] ) ) {
+			if ( $node->hasAttribute( $attr_name ) ) {
+				$attr_value   = $node->getAttribute( $attr_name );
+				$attr_value   = preg_replace( '/\s*,\s*/', ',', $attr_value );
+				$urls_to_test = explode( ',', $attr_value );
```
westonruter: I'm curious about the comma-separated list of URLs. Naturally this shouldn't be allowed in an `href` or `src`. What elements do this?

westonruter: You can also combine these two lines into:

```php
$urls_to_test = preg_split( '/\s*,\s*/', $attr_value );
```

rubengonzalezmrf: an image with a srcset attribute would have a list of comma separated urls, so i thought it could be a good idea to check all of them

rubengonzalezmrf:
I have post content now with the same 2 examples you added, and i get out the expected results. But unit tests fails anyway

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1122,6 +1138,48 @@ private function check_attr_spec_rule_value_regex_casei( $node, $attr_name, $att
...
+	 *      - AMP_Rule_Spec::PASS - $attr_name has a value that matches the rule.
+	 *      - AMP_Rule_Spec::FAIL - $attr_name has a value that does *not* match rule.
+	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
+	 *                                        is no rule for this attribute.
+	 */
+	private function check_attr_spec_rule_valid_url( $node, $attr_name, $attr_spec_rule ) {
+		if ( isset( $attr_spec_rule[ AMP_Rule_Spec::VALUE_URL ] ) ) {
+			if ( $node->hasAttribute( $attr_name ) ) {
+				$attr_value   = $node->getAttribute( $attr_name );
+				$urls_to_test = preg_split( '/\s*,\s*/', $attr_value );
```
westonruter: Since only `srcset` has comma-separated list of URLs, maybe this should check of `'srcset' === $attr_name` for whether to split. Maybe that's overkill. My only concern would be if a non-`srcset` URL happening to have a comma in it and that this would result in the attribute being invalid. 

The alternative could be to do this: 

```php
$urls_to_test = preg_split( '#\s*,\s*(?=https?://)#', $attr_value );
```

That will ensure that it will only split where a comma is followed by another URL.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1122,6 +1138,46 @@ private function check_attr_spec_rule_value_regex_casei( $node, $attr_name, $att
...
+				$urls_to_test = preg_split( '/\s*,\s*/', $node->getAttribute( $attr_name ) );
+				foreach ( $urls_to_test as $url ) {
+					// Check if the host contains invalid chars.
+					$url_host = wp_parse_url( urldecode( $url ), PHP_URL_HOST );
+					if ( $url_host && preg_match( '/[!"#$%&\'()*+,\/:;<=>?@[\]^`{|}~\s]/i', $url_host ) ) {
+						return AMP_Rule_Spec::FAIL;
+					}
+
+					// Check if the protocol contains invalid chars.
+					$dots_pos = strpos( $url, ':' );
```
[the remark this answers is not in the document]
rubengonzalezmrf: The point is that `mail%20to:foo@bar.com` won't be correctly parsed by wp_parse_url, the scheme is null

PR #984, opened 2018-03-01
To avoid deprecated argument notice, do not pass $post to amp_get_con…

sarutole:
…tent_embed_handlers() and amp_get_content_sanitizers().

### DIFF
--- a/includes/templates/class-amp-post-template.php
+++ b/includes/templates/class-amp-post-template.php
@@ -310,8 +310,8 @@ class AMP_Post_Template {
 	private function build_post_content() {
 		$amp_content = new AMP_Content(
 			$this->post->post_content,
-			amp_get_content_embed_handlers( $this->post ),
-			amp_get_content_sanitizers( $this->post ),
+			amp_get_content_embed_handlers(),
+			amp_get_content_sanitizers(),
 			array(
 				'content_max_width' => $this->get( 'content_max_width' ),
 			)


westonruter:
How is it that you are getting the deprecation notice? The `build_post_content` method only applies when `amp` theme support is not present, and the deprecation notice is only shown when `amp` support _is_ present.

sarutole:
Thank you, your comment helped me realize I was going against the flow in some ways.

I had it set up so that theme did not call `wp_head` and `wp_footer` so that plugins would not pollute head and footer with their styles and scripts.

Now I get the contents of the output buffer right _before_ `AMP_Theme_Support::prepare_response()` and replace `<link>` elements with what is needed for my theme. Now AMP-WP can be used without modification. 

Thank for prompt review.

westonruter:
@sarutole what `link` elements are you replacing? Manipulating the output buffer shouldn't really be needed by other integrations.

sarutole:
I removed the `link` elements that were referring to stylesheets enqueued by various plugins, for example a stylesheet for contact forms. I am aware it is possible to dequeue them, but it is not practical because any newly added plugin can enqueue some more. Or am I missing something?

westonruter:
@sarutole actually, dequeueing is exactly what you should do. All you need to do is take a whitelist approach, for example:

```php
add_action( 'wp_print_styles', function() {
	$allowed_stylesheets = array(
		'amp-default',
		get_stylesheet() . '-style',
		get_template() . '-style',
	);
	foreach ( array_diff( wp_styles()->queue, $allowed_stylesheets ) as $handle ) {
		wp_dequeue_style( $handle );
	}
} );
```

Issue #945, opened 2018-02-08
Deprecate AMP_QUERY_VAR constant and amp_query_var filter


westonruter:
Currently one is able to override the `amp` query var and endpoint to be something totally different:

https://github.com/Automattic/amp-wp/blob/c701b0b8614e4b454bf620950018d0f143796d3c/amp.php#L80-L88

This ability complicates some things because we have to wait until the theme has fully loaded before we can start looking for whether it is an AMP request or not.

I want to find out if this ability is actually being used. And if so, why. If it is not, then we should remove it.

PR #986, opened 2018-03-02
Set AMP_QUERY_VAR fallback when AMP is inactive

mjangda:
If AMP is deactivated via the `amp_is_enabled` filter, the `AMP_QUERY_VAR` ends up undefined. This causes warnings when a helper function like `is_amp_endpoint()` is called. Setting a fallback value prevents this.

This was happening with a site running both `amp-wp` (but deactivated via the filter) and `lazy-load` (which calls `is_amp_endpoint()`):

```
Warning: Use of undefined constant AMP_QUERY_VAR - assumed 'AMP_QUERY_VAR' (this will throw an Error in a future version of PHP) in wp-content/plugins/amp/includes/amp-helper-functions.php
```

To reproduce:

```
add_filter( 'amp_is_enabled', '__return_false' );

add_action( 'wp_head', function() {
    is_amp_endpoint();
} );
```

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -55,7 +55,7 @@ function amp_deactivate() {
 	// We need to manually remove the amp endpoint
 	global $wp_rewrite;
 	foreach ( $wp_rewrite->endpoints as $index => $endpoint ) {
-		if ( AMP_QUERY_VAR === $endpoint[1] ) {
+		if ( amp_get_slug() === $endpoint[1] ) {
 			unset( $wp_rewrite->endpoints[ $index ] );
 			break;
 		}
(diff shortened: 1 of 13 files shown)


mjangda:
We may also want to consider hardening up the helper functions as well to return appropriate values when AMP is disabled via the filter.

westonruter:
See also #945. What have been the use cases for `AMP_QUERY_VAR` in the wild?

mjangda:
One example is http://indianexpress.com/article/india/breakthrough-journalist-gauri-lankesh-murder-first-arrest-5084078/lite/ (they use `lite` instead of `amp`).

Deprecation is fine too :)

westonruter:
> We may also want to consider hardening up the helper functions as well to return appropriate values when AMP is disabled via the filter.

I think this might be the better way to go instead of defining `AMP_QUERY_VAR` as `false`, as if someone _does_ later do define `AMP_QUERY_VAR` then they'll get a notice about it already being defined. Or rather, if changing the slug is to be deprecated, then maybe we should just stop using `AMP_QUERY_VAR` altogether and instead introduce an `amp_get_slug()` function instead.

westonruter:
Rebased to fix merge conflict. Former `head` was 82dced4a4.

mjangda:
> we should just stop using AMP_QUERY_VAR altogether and instead introduce an amp_get_slug() function instead.

Sounds like a good idea.

westonruter:
@ThierryA I just wanted you to be aware of it.

PR #992, opened 2018-03-08
Prevent schema.org duplicates

### LINKED ISSUE #962: Only add Schema.org metadata if not already added by another plugin
Currently when theme support is enabled, `amp_print_schemaorg_metadata()` is called at the `wp_head` action. However, if another plugin is adding this Schema.org metadata then we should defer to the meta that it adds to prevent duplicates. So I think we should augment `AMP_Theme_Support::ensure_required_markup()` to check if the `script` is present after output-buffering, and if so, then add `amp_get_schemaorg_metadata()` to the `head` element.

Something similar should be done eventually for the app manifest.

For both Schema.org metadata and app manifest, core should eventually be doing these itself, and that will help facilitate themes and plugins coordinating on the data that is output.

oscarssanchez:
Request For Review

Hi @westonruter,

Could you use this PR, which addresses #962?

Originally, I used json_decode( $script->nodeValue ) to convert the schema.org json data to an array. However, that brought some issues with get_mustache_tag_placeholders(), as it adds some extra markup to the schema.org script tag. 

I opted to use preg_match() as I saw the schema.org URL sometimes can either start with http or https. 

Fixes #962.

### DIFF
--- a/includes/amp-post-template-actions.php
+++ b/includes/amp-post-template-actions.php
@@ -14,6 +14,7 @@ function amp_post_template_init_hooks() {
 	add_action( 'amp_post_template_head', 'amp_post_template_add_scripts' );
 	add_action( 'amp_post_template_head', 'amp_post_template_add_fonts' );
 	add_action( 'amp_post_template_head', 'amp_post_template_add_boilerplate_css' );
+	add_action( 'amp_post_template_head', 'amp_print_schemaorg_metadata' );
 	add_action( 'amp_post_template_head', 'amp_add_generator_metadata' );
 	add_action( 'amp_post_template_css', 'amp_post_template_add_styles', 99 );
 	add_action( 'amp_post_template_data', 'amp_post_template_add_analytics_script' );
--- a/includes/amp-post-template-actions.php
+++ b/includes/amp-post-template-actions.php
@@ -14,7 +14,6 @@ function amp_post_template_init_hooks() {
 	add_action( 'amp_post_template_head', 'amp_post_template_add_scripts' );
 	add_action( 'amp_post_template_head', 'amp_post_template_add_fonts' );
 	add_action( 'amp_post_template_head', 'amp_post_template_add_boilerplate_css' );
-	add_action( 'amp_post_template_head', 'amp_print_schemaorg_metadata' );
 	add_action( 'amp_post_template_head', 'amp_add_generator_metadata' );
 	add_action( 'amp_post_template_css', 'amp_post_template_add_styles', 99 );
 	add_action( 'amp_post_template_data', 'amp_post_template_add_analytics_script' );
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -794,7 +794,7 @@ class AMP_Theme_Support {
 	 *
 	 * @param DOMDocument $dom Doc.
 	 */
-	protected static function ensure_required_markup( DOMDocument $dom ) {
+	public static function ensure_required_markup( DOMDocument $dom ) {
 		$head = $dom->getElementsByTagName( 'head' )->item( 0 );
 		if ( ! $head ) {
 			$head = $dom->createElement( 'head' );
@@ -828,14 +828,21 @@ class AMP_Theme_Support {
 			) );
 			$head->insertBefore( $meta_viewport, $meta_charset->nextSibling );
 		}
-		if ( ! self::schema_org_present( $dom ) ) {
+		// Prevent schema.org duplicates.
+		$schema_org_script = null;
+		foreach ( $head->getElementsByTagName( 'script' ) as $script ) {
+			if ( 'application/ld+json' === $script->getAttribute( 'type' ) && preg_match( '/{"@context":"https?:[\\\]\/[\\\]\/schema\.org/', $script->nodeValue ) ) {
+				$schema_org_script = $script->nodeValue;
+				break;
+			}
+		}
+		if ( ! $schema_org_script ) {
 			$script = $dom->createElement( 'script', wp_json_encode( amp_get_schemaorg_metadata() ) );
 			AMP_DOM_Utils::add_attributes_to_node( $script, array(
 				'type' => 'application/ld+json',
 			) );
 			$head->appendChild( $script );
 		}
-
 		// Ensure rel=canonical link.
 		$rel_canonical = null;
 		foreach ( $head->getElementsByTagName( 'link' ) as $link ) {
@@ -1012,21 +1019,5 @@ class AMP_Theme_Support {
 		);
 
 		return $response;
-	}
-
-	/**
-	 * Checks if there is already a schema.org script in the head.
-	 *
-	 * @param DOMDocument $dom Representation of the document.
-	 * @return bool
-	 */
-	public static function schema_org_present( DOMDocument $dom ) {
-		$head = $dom->getElementsByTagName( 'head' )->item( 0 );
-		foreach ( $head->getElementsByTagName( 'script' ) as $script ) {
-			if ( 1 === preg_match( '/{"@context":"https?:[\\\]\/[\\\]\/schema\.org/', $script->nodeValue ) ) {
-				return true;
-			}
-		}
-		return false;
 	}
 }
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -829,18 +829,16 @@ class AMP_Theme_Support {
 			$head->insertBefore( $meta_viewport, $meta_charset->nextSibling );
 		}
 		// Prevent schema.org duplicates.
-		$schema_org_script = null;
+		$has_schema_org_metadata = false;
 		foreach ( $head->getElementsByTagName( 'script' ) as $script ) {
-			if ( 'application/ld+json' === $script->getAttribute( 'type' ) && preg_match( '/{"@context":"https?:[\\\]\/[\\\]\/schema\.org/', $script->nodeValue ) ) {
-				$schema_org_script = $script->nodeValue;
+			if ( 'application/ld+json' === $script->getAttribute( 'type' ) && false !== strpos( $script->nodeValue, 'schema.org' ) ) {
+				$has_schema_org_metadata = true;
 				break;
 			}
 		}
-		if ( ! $schema_org_script ) {
+		if ( ! $has_schema_org_metadata ) {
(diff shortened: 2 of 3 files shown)


includes/amp-post-template-actions.php
```diff
@@ -14,7 +14,6 @@ function amp_post_template_init_hooks() {
 	add_action( 'amp_post_template_head', 'amp_post_template_add_scripts' );
 	add_action( 'amp_post_template_head', 'amp_post_template_add_fonts' );
 	add_action( 'amp_post_template_head', 'amp_post_template_add_boilerplate_css' );
-	add_action( 'amp_post_template_head', 'amp_print_schemaorg_metadata' );
```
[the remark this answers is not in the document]
oscarssanchez: Hi @westonruter 

Thanks for your feedback. I corrected that mistake and restored the line.

tests/test-class-amp-theme-support.php
```diff
@@ -330,4 +330,49 @@ public function test_handle_xhr_request() {
...
+	/**
+	 * Test schema_org_present().
+	 *
+	 * @dataProvider get_script_data
+	 * @covers AMP_Theme_Support::schema_org_present()
+	 * @param string  $script The value of the script.
+	 * @param boolean $expected The expected result.
+	 */
+	public function test_schema_org_present( $script, $expected ) {
+		$page = '<html><head><script>%s</script></head><body>Test</body></html>';
```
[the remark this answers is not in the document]
oscarssanchez: Very good observation, i added the type attribute on the new tests for ensure_required_markup()

includes/class-amp-theme-support.php
```diff
@@ -1007,4 +1013,20 @@ public static function prepare_response( $response, $args = array() ) {
...
+	/**
+	 * Checks if there is already a schema.org script in the head.
+	 *
+	 * @param DOMDocument $dom Representation of the document.
+	 * @return bool
+	 */
+	public static function schema_org_present( DOMDocument $dom ) {
+		$head = $dom->getElementsByTagName( 'head' )->item( 0 );
+		foreach ( $head->getElementsByTagName( 'script' ) as $script ) {
+			if ( 1 === preg_match( '/{"@context":"https?:[\\\]\/[\\\]\/schema\.org/', $script->nodeValue ) ) {
```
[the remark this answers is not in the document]
oscarssanchez: Removed the integer comparison and added your suggestion, indeed it makes it more precise.

includes/class-amp-theme-support.php
```diff
@@ -1007,4 +1013,20 @@ public static function prepare_response( $response, $args = array() ) {
...
 		return $response;
 	}
+
+	/**
+	 * Checks if there is already a schema.org script in the head.
+	 *
+	 * @param DOMDocument $dom Representation of the document.
+	 * @return bool
+	 */
+	public static function schema_org_present( DOMDocument $dom ) {
```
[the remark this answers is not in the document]
oscarssanchez: I added your suggestions. 

I also did the test for ensure_required_markup, however, I only made tests for the schema.org script. Could you use tests for the entire function?

includes/class-amp-theme-support.php
```diff
@@ -829,7 +828,21 @@ protected static function ensure_required_markup( DOMDocument $dom ) {
 			) );
 			$head->insertBefore( $meta_viewport, $meta_charset->nextSibling );
 		}
-
+		// Prevent schema.org duplicates.
+		$schema_org_script = null;
+		foreach ( $head->getElementsByTagName( 'script' ) as $script ) {
+			if ( 'application/ld+json' === $script->getAttribute( 'type' ) && preg_match( '/{"@context":"https?:[\\\]\/[\\\]\/schema\.org/', $script->nodeValue ) ) {
```
westonruter: The regular expression here could be brittle. There's no guarantee that the `@context` will be the first key, and there's no guarantee there wouldn't be whitespace between the strings. I think doing a simple `false !== strpos( $script->nodeValue, 'schema.org' )` could be sufficient here.

includes/class-amp-theme-support.php
```diff
@@ -829,7 +828,21 @@ protected static function ensure_required_markup( DOMDocument $dom ) {
...
+				$schema_org_script = $script->nodeValue;
```
westonruter: Since the value isn't being used, it may be better to just store a flag like `$has_schema_org_metadata = true`.

includes/class-amp-theme-support.php
```diff
@@ -829,7 +828,21 @@ protected static function ensure_required_markup( DOMDocument $dom ) {
...
+				break;
+			}
+		}
+		if ( ! $schema_org_script ) {
+			$script = $dom->createElement( 'script', wp_json_encode( amp_get_schemaorg_metadata() ) );
+			AMP_DOM_Utils::add_attributes_to_node( $script, array(
```
westonruter: Since you're using `$dom->createElement()` here then I think it makes more sense to just do `$script->setAttribute( 'type', 'application/ld+json' )`

tests/test-class-amp-theme-support.php
```diff
@@ -330,4 +330,25 @@ public function test_handle_xhr_request() {
...
+		$dom  = new DOMDocument();
+		$dom->loadHTML( $page );
+		AMP_Theme_Support::ensure_required_markup( $dom );
+		$this->assertContains( $schema_test_value, $dom->saveHTML() );
+
+		$page = '<html><head<script type="application/ld+json">{"@context":"http:\/\/schema.org","publisher":{"@type":"Organization","name":"Test Blog"}}</script>></head><body>Test</body></html>';
+		$dom  = new DOMDocument();
+		$dom->loadHTML( $page );
+		AMP_Theme_Support::ensure_required_markup( $dom );
+		$this->assertEquals( substr_count( $dom->saveHTML(), $schema_test_value ), 1 );
```
westonruter: Also should add a test for when JSON is output but the slashes are not escaped.

westonruter: And add a test for when `@context` is not the first key.

includes/class-amp-theme-support.php
westonruter: We may not even need to look at the value. Just checking if the type is `application/ld+json` may be just fine.

westonruter:
One more higher-level question: when I first raised this issue I was thinking that there could only be one such metadata script in the document. However, apparently it's fine to have multiple: https://stackoverflow.com/questions/30723531/best-json-ld-practices-using-multiple-script-elements

The main concern here for the AMP plugin is to allow for other plugins to take control of the generation of the Schema.org metadata instead of the defaults that the AMP plugin outputs, but then to have the AMP plugin output a good fallback when others don't do so.

I'm feeling like if there is _any_ JSON-LD `script` in the `head` (which mentions `schema.org`) then we can take this as the signal to just skip outputting the default metadata from the plugin.

includes/class-amp-theme-support.php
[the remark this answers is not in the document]
westonruter: Yes, this method needs more tests, so if that's something you want to add that would be great.

includes/class-amp-theme-support.php
westonruter: Why can't this do `json_decode( $script->nodeValue, false )` and check if `wp_parse_url( $script['@context'], PHP_URL_HOST ) === 'schema.org`?

oscarssanchez: Thanks, it is true, i made the corresponding correction.

includes/class-amp-theme-support.php
oscarssanchez: Also true, corrected it.

tests/test-class-amp-theme-support.php
oscarssanchez: I made tests for when @anothercontext is the first key and has schema.org present, I hope i understood correctly.

includes/class-amp-theme-support.php
oscarssanchez: Hi @westonruter, 

Trying to use `wp_parse_url` with `json_decode( $script->nodeValue, false )` breaks the site, but it works fine when nodeValue is converted to an array. The logic however stops working.

I've found that if you comment out these lines in `class-amp-dom-utils.php`, the logic works again:

		$placeholders = self::get_mustache_tag_placeholders();
		$document     = str_replace(
			array_keys( $placeholders ),
			array_values( $placeholders ),
			$document
		);

I opted to keep your previous suggestions since i don't know how important are those lines.

includes/class-amp-theme-support.php
oscarssanchez: Hi @westonruter 

I opted to search for both the attribute and the 'schema.org' substring as you mentioned it should check for any JSON-LD script in the head which mentions 'schema.org' in any form.

includes/class-amp-theme-support.php
westonruter: OK, I think this makes sense. If I have some JSON that looks like:

```php
{"foo": {"bar":}}
```

The last `}}` there is going to get replaced with a placeholder (since this token is returned by `get_mustache_tag_placeholders`) before parsing the DOM, to then get put back in `AMP_DOM_Utils::get_content_from_dom_node()`. That means we cannot parse JSON from the parsed DOM currently, which is not ideal. To fix this we'd need to limit the replacements to only happen inside of `href` and `src` attributes only, using something like this:

```diff
diff --git a/includes/utils/class-amp-dom-utils.php b/includes/utils/class-amp-dom-utils.php
index 5288c04..f20fba1 100644
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -77,9 +77,15 @@ class AMP_DOM_Utils {
 		 * elements since it is faster and wouldn't change the outcome.
 		 */
 		$placeholders = self::get_mustache_tag_placeholders();
-		$document     = str_replace(
-			array_keys( $placeholders ),
-			array_values( $placeholders ),
+		$document     = preg_replace_callback(
+			'#\s(src|href)=("[^"]+"|\'[^\']+\')#s',
+			function( $matches ) use ( $placeholders ) {
+				return str_replace(
+					array_keys( $placeholders ),
+					array_values( $placeholders ),
+					$matches[0]
+				);
+			},
 			$document
 		);
```

That would probably be something good to add here to ensure we can read JSON out of the DOM.

westonruter: I've merged the fix for this in #1019 via https://github.com/Automattic/amp-wp/pull/1019/commits/4a10ee479650016e1baa80c1f5c6dbe566247bf5

PR #994, opened 2018-03-08
Fix Backtrack Limit Exhausted

### LINKED ISSUE #993: 500 Error when PREG Backtrack Limit Exhausted
PHP versions 5.2.x have a default value of `100000` on `pcre.backtrack_limit`.  We stumbled on the issue in `0.7-beta` with the work supporting amp-bind.

We are using `amp-iframe` as a container for a table with a lot of rows and some javascript to help filter that table.  We're injecting the iframe source through the `srcdoc` attribute.  The regex matching tag attributes eventually will hit backtrack limit, fail with a `PREG_BACKTRACK_LIMIT_ERROR` silently, and `AMP_Dom_Utils::convert_amp_bind_attributes` will return `null`.

Further along the stack, this `null` value will be fed to `DOMDocument` which will cause a fatal error.
I'm opening up a PR shortly with some failing tests.  I've made an attempt at modifying the regex, but I can't get all the tests to pass with my changes so I'm opening up this issue to the broader community.

christianc1:
## Work In Progress for #993 

Adding some tests to make the issue apparent.

Fixes #993.

### DIFF
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -233,7 +233,7 @@ class AMP_DOM_Utils {
 
 		$html = preg_replace_callback(
 			// Match all start tags that probably contain a binding attribute.
-			'#<(?P<name>[a-zA-Z0-9_\-]+)(?P<attrs>\s+[^>]+\]=[^>]+)\s*>#',
+			'#<(?P<name>[a-zA-Z0-9_\-]+)(?P<attrs>\s[^>]+\]=[^>]+)>#',
 			$replace_callback,
 			$html
 		);
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -64,8 +64,8 @@ class AMP_DOM_Utils {
 
 		$dom = new DOMDocument();
 
-			// @todo In the future consider an AMP_DOMDocument subclass that does this automatically. See <https://github.com/Automattic/amp-wp/pull/895/files#r163825513>.
-			$document = self::convert_amp_bind_attributes( $document );
+		// @todo In the future consider an AMP_DOMDocument subclass that does this automatically. See <https://github.com/Automattic/amp-wp/pull/895/files#r163825513>.
+		$document = self::convert_amp_bind_attributes( $document );
 
 		/*
 		 * Prevent amp-mustache syntax from getting URL-encoded in attributes when saveHTML is done.
@@ -238,7 +238,15 @@ class AMP_DOM_Utils {
 			$html
 		);
 
-		// If the conversion failed, pass the original HTML as DOMDocument may still be able to use it.
+		/**
+		 * If the regex engine incurred an error during processing, for example exceeding the backtrack
+		 * limit, $converted will be null. In this case we return the originally passed document to allow
+		 * DOMDocument to attempt to load it.  If the AMP HTML doesn't make use of amp-bind or similar
+		 * attributes, then everything should still work.
+		 *
+		 * See https://github.com/Automattic/amp-wp/issues/993 for additional context on this issue.
+		 * See http://php.net/manual/en/pcre.constants.php for additional info on PCRE errors.
+		 */
 		return ( ! is_null( $converted ) ) ? $converted : $html;
 	}
 
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -64,8 +64,10 @@ class AMP_DOM_Utils {
 
 		$dom = new DOMDocument();
 
-		// @todo In the future consider an AMP_DOMDocument subclass that does this automatically. See <https://github.com/Automattic/amp-wp/pull/895/files#r163825513>.
-		$document = self::convert_amp_bind_attributes( $document );
+		if ( apply_filters( 'amp_convert_bind_attributes', true, $document ) ) {
+			// @todo In the future consider an AMP_DOMDocument subclass that does this automatically. See <https://github.com/Automattic/amp-wp/pull/895/files#r163825513>.
+			$document = self::convert_amp_bind_attributes( $document );
+		}
 
 		/*
 		 * Prevent amp-mustache syntax from getting URL-encoded in attributes when saveHTML is done.
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -64,6 +64,17 @@ class AMP_DOM_Utils {
 
 		$dom = new DOMDocument();
 
+		/**
+		 * Convert Bind Attributes.
+		 *
+		 * Use this filter to skip converting amp-bind style attributes. On large documents that
+		 * are known not to contain amp-bind style attributes, it's more performant to skip the
+		 * conversion.  Use of this filter is only recommended if documents are very large and
+		 * at risk of exceeding regex engine limits.
+		 *
+		 * @param bool   $convert  True if bind attributes should be converted, false otherwise. Default true.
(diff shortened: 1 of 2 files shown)


tests/test-class-amp-dom-utils.php
```diff
@@ -145,4 +145,58 @@ public function test_html5_empty_elements() {
...
+	 *
+	 * @return array An array of arrays holding an integer representation of iterations.
+	 */
+	public function get_table_row_iterations() {
+		return array(
+			array( 1 ),
+			array( 10 ),
+			array( 100 ),
+			array( 1000 ),
+			array( 10000 ),
```
westonruter: Per https://github.com/Automattic/amp-wp/issues/993#issuecomment-372139130 I think this 10,000 value should just be removed. I don't think it is reasonable to expect a `srcdoc` to be this large, let alone an AMP document as a whole.

tests/test-class-amp-dom-utils.php
```diff
@@ -145,4 +145,58 @@ public function test_html5_empty_elements() {
...
+		}
+
+		$html .= '</table></body></html>';
+
+		$to_convert = '<amp-iframe
+		sandbox="allow-scripts"
+		srcdoc="' . htmlentities( $html ) . '">
+		</amp-iframe>';
+
+		$converted = AMP_DOM_Utils::convert_amp_bind_attributes( $to_convert );
```
westonruter: The `$converted` variable isn't being used, so should it even be set?

tests/test-class-amp-dom-utils.php
```diff
@@ -145,4 +145,58 @@ public function test_html5_empty_elements() {
...
+				</tr>
+			';
+		}
+
+		$html .= '</table></body></html>';
+
+		$to_convert = '<amp-iframe
+		sandbox="allow-scripts"
+		srcdoc="' . htmlentities( $html ) . '">
+		</amp-iframe>';
```
westonruter: I think it is better to use `sprintf()` here:

```php
$to_convert = sprintf( 
    '<amp-iframe sandbox="allow-scripts" sandbox="allow-scripts" srcdoc="%s"></amp-iframe>',
    htmlentities( $html )
);
```

tests/test-class-amp-dom-utils.php
christianc1: If the regex engine fails *for any reason*, `$converted` will be `null` here.  I was originally testing for `null` in this test case, `$this->assertNotNull( $converted );`. I then later opted to fail with a specific `PREG_BACKTRACK_LIMIT_ERROR` to give more insight into why the test actually failed.  I'll remove this.

tests/test-class-amp-dom-utils.php
christianc1: 👍

includes/utils/class-amp-dom-utils.php
```diff
@@ -64,8 +64,21 @@ public static function get_dom( $document ) {
...
+		 * are known not to contain amp-bind style attributes, it's more performant to skip the
+		 * conversion.  Use of this filter is only recommended if documents are very large and
+		 * at risk of exceeding regex engine limits.
+		 *
+		 * @param bool   $convert  True if bind attributes should be converted, false otherwise. Default true.
+		 * @param string $document Valid HTML document possibly containing bind attributes to be converted.
+		 */
+		if ( apply_filters( 'amp_convert_bind_attributes', true, $document ) ) {
+			// @todo In the future consider an AMP_DOMDocument subclass that does this automatically. See <https://github.com/Automattic/amp-wp/pull/895/files#r163825513>.
+			$document = self::convert_amp_bind_attributes( $document );
```
westonruter: Note that while the conversion is conditional here, the restoration is not conditional. Below you can see `restore_amp_bind_attributes` gets called even though no conversion may have been done. This isn't necessarily a problem, but it is a wasted search/replace operation.

includes/utils/class-amp-dom-utils.php
```diff
@@ -231,14 +244,15 @@ public static function convert_amp_bind_attributes( $html ) {
...
+		$converted = preg_replace_callback(
 			// Match all start tags that probably contain a binding attribute.
-			'#<(?P<name>[a-zA-Z0-9_\-]+)(?P<attrs>\s+[^>]+\]=[^>]+)\s*>#',
+			'#<(?P<name>[a-zA-Z0-9_\-]+)(?P<attrs>\s[^>]+\]=[^>]+)>#',
 			$replace_callback,
 			$html
 		);
 
-		return $html;
+		// If the conversion failed, pass the original HTML as DOMDocument may still be able to use it.
```
westonruter: Maybe good to explain why it failed. Linking back the GitHub issue and mentioning backtracking would be good.

includes/utils/class-amp-dom-utils.php
```diff
@@ -64,8 +64,21 @@ public static function get_dom( $document ) {
...
-		// @todo In the future consider an AMP_DOMDocument subclass that does this automatically. See <https://github.com/Automattic/amp-wp/pull/895/files#r163825513>.
-		$document = self::convert_amp_bind_attributes( $document );
+		/**
+		 * Convert Bind Attributes.
+		 *
+		 * Use this filter to skip converting amp-bind style attributes. On large documents that
+		 * are known not to contain amp-bind style attributes, it's more performant to skip the
+		 * conversion.  Use of this filter is only recommended if documents are very large and
+		 * at risk of exceeding regex engine limits.
+		 *
```
westonruter: Add `@since 0.7`

includes/utils/class-amp-dom-utils.php
```diff
@@ -64,8 +64,21 @@ public static function get_dom( $document ) {
...
+		 * @param bool   $convert  True if bind attributes should be converted, false otherwise. Default true.
+		 * @param string $document Valid HTML document possibly containing bind attributes to be converted.
+		 */
+		if ( apply_filters( 'amp_convert_bind_attributes', true, $document ) ) {
```
westonruter: So you see this being necessary? Is it required in your use case? How big are the HTML documents you're working with?

tests/test-class-amp-dom-utils.php
```diff
@@ -145,4 +145,57 @@ public function test_html5_empty_elements() {
...
+		$to_convert = sprintf(
+			'<amp-iframe sandbox="allow-scripts" srcdoc="%s"> </amp-iframe>',
+			htmlentities( $html )
+		);
+
+		AMP_DOM_Utils::convert_amp_bind_attributes( $to_convert );
+
+		if ( preg_last_error() === PREG_BACKTRACK_LIMIT_ERROR ) {
+			$this->fail( 'Failed when backtrack limit was exhausted.' );
+		}
```
westonruter: When there is no error this test will run without any assertions. That will mean this test will appear as a warning when PHPUnit is called with ` --strict`.

I think it would be better to do:

```php
$this->assertSame( PREG_NO_ERROR, preg_last_error(), 'Probably failed when backtrack limit was exhausted.' );
```

includes/utils/class-amp-dom-utils.php
christianc1: The HTML document which led to this error was the only one I've identified, so my sample size is small lol.

In my particular case, without this filter, the document will be passed through the conversion method and `DOMDocument` will be able to load it.  This filter is not strictly necessary, I just thought it was a convenient way for a developer to bypass an operation that they know will not serve a purpose.  That said, I don't want to pollute the codebase with obscure filters that might be misused.  I'll leave it up to you to decide whether it should be included.  I'm happy to remove it.

christianc1: I re-thought this, this filter only confuses the issue.  I'll remove, I don't see a strong enough use case to include it.

PR #995, opened 2018-03-08
Consistently use wp_parse_url(); deprecate AMP_WP_Utils

### LINKED ISSUE #876: Deprecate AMP_WP_Utils

paulschreiber:
* Consistently use wp_parse_url() instead of a combination of `parse_url()` and `AMP_WP_Utils::parse_url()`.
* Remove AMP_WP_Utils class and tests 
* Bump required version of WordPress to 4.8

Fixes #876.


paulschreiber:
I removed the version bump, restored the class and marked the class and its function as deprecated.

PR #996, opened 2018-03-08
phpcs: set color=true in output

paulschreiber:
Updates `composer.json` to colorize output of `vendor/bin/phpcs`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,5 +9,10 @@
         "wp-coding-standards/wpcs": "^0.14.0",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.4.4",
         "wimg/php-compatibility": "^8.1"
+    },
+    "scripts": {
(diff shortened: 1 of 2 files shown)


westonruter:
If we're going to set the color, it would be better to do this in the ruleset itself: https://github.com/Automattic/amp-wp/blob/develop/phpcs.xml

See also https://make.xwp.co/2017/06/04/reducing-command-line-args-required-for-running-phpcs/

And an argument for why color should not be default: https://github.com/WordPress/gutenberg/pull/1685#issuecomment-312780011

paulschreiber:
I moved the argument to phpcs.xml.

westonruter:
Still, I'm not sure that colors should be the default. See: https://github.com/WordPress/gutenberg/pull/1685#issuecomment-312780011

PR #1000, opened 2018-03-08
improve sanitization in jetpack_amp_build_stats_pixel_url()

paulschreiber:
* avoid double-calling `rawurlencode()` on `$data` — `array_map()` already does this
* sanitize HTTP_HOST

### DIFF
--- a/jetpack-helper.php
+++ b/jetpack-helper.php
@@ -63,7 +63,7 @@ function jetpack_amp_build_stats_pixel_url() {
 		$data = compact( 'v', 'j', 'blog', 'post', 'tz', 'srv' );
 	}
 
-	$data['host'] = isset( $_SERVER['HTTP_HOST'] ) ? rawurlencode( $_SERVER['HTTP_HOST'] ) : ''; // input var ok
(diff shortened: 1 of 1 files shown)


jetpack-helper.php
```diff
@@ -63,7 +63,7 @@ function jetpack_amp_build_stats_pixel_url() {
 		$data = compact( 'v', 'j', 'blog', 'post', 'tz', 'srv' );
 	}
 
-	$data['host'] = isset( $_SERVER['HTTP_HOST'] ) ? rawurlencode( $_SERVER['HTTP_HOST'] ) : ''; // input var ok
+	$data['host'] = isset( $_SERVER['HTTP_HOST'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_HOST'] ) ) : ''; // input var ok
```
westonruter: `sanitize_text_field` is not equivalent to `rawurlencode`

westonruter:
See https://github.com/Automattic/amp-wp/pull/1002#issuecomment-371617248

jetpack-helper.php
westonruter: Nevermind. I supose this change is ok. You can cherry-pick this into #1002 as well.

PR #1012, opened 2018-03-10
Wrap registered widgets in callbacks to track widget validation errors

### LINKED ISSUE #1005: Add reporting for validation errors coming from widgets
Just as we are now detect the specific shortcode that is the source for an AMP validation error in #971 (and as we'll do for blocks in Gutenberg), we also need to do the same for widgets. When a widget renders something that not valid AMP, we need to indicate:

1. The widget that did something invalid.
2. The theme/plugin source for where the widget came from.

westonruter:
* Include core as a source in addition to plugins and themes.
* Refactor source comments to only consist of JSON-encoded source data.
* Include the function name in the source.

Example validation data for when a Custom HTML widget is added which includes a `script` tag:

image

Fixes #1005.

### DIFF
--- a/includes/utils/class-amp-validation-utils.php
+++ b/includes/utils/class-amp-validation-utils.php
@@ -284,7 +284,8 @@ class AMP_Validation_Utils {
 	 * Add hooks for doing validation during preprocessing/sanitizing.
 	 */
 	public static function add_validation_hooks() {
-		self::callback_wrappers();
+		self::wrap_widget_callbacks();
+		add_action( 'all', array( __CLASS__, 'wrap_hook_callbacks' ) );
 		add_filter( 'do_shortcode_tag', array( __CLASS__, 'decorate_shortcode_source' ), -1, 2 );
 		add_filter( 'amp_content_sanitizers', array( __CLASS__, 'add_validation_callback' ) );
 	}
@@ -602,6 +603,7 @@ class AMP_Validation_Utils {
 	 * @return string HTML Comment.
 	 */
 	public static function get_source_comment( array $source, $is_start = true ) {
+		unset( $source['reflection'] );
 		return sprintf(
 			'<!--%samp-source-stack %s-->',
 			$is_start ? '' : '/',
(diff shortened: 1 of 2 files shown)


westonruter:
- [ ] Add some tests specifically for registered widgets.
- [ ] Try using `all` action instead of wrapping all callbacks up-front.

westonruter:
Ready for review!

kienstra:
**Will Review**

Hi @westonruter,
Sorry for the delay here. I'll review this late today (Sunday) if that's alright.

tests/test-class-amp-validation-utils.php
```diff
@@ -68,16 +68,33 @@ class Test_AMP_Validation_Utils extends \WP_UnitTestCase {
...
 		$this->node   = $dom_document->createElement( self::TAG_NAME );
 		AMP_Validation_Utils::reset_validation_results();
+		$this->original_wp_registered_widgets = $GLOBALS['wp_registered_widgets'];
+	}
+
+	/**
+	 * After a test method runs, reset any state in WordPress the test method might have changed.
+	 */
+	public function tearDown() {
+		$GLOBALS['wp_registered_widgets'] = $this->original_wp_registered_widgets; // WPCS: override ok.
```
kienstra: Good idea to reset the original global `$wp_registered_widgets`.

westonruter: It should be something that gets reset for us automatically, but it wasn't so this is to ensure we reset the global state.

includes/utils/class-amp-validation-utils.php
```diff
@@ -597,47 +598,34 @@ public static function print_edit_form_validation_status( $post ) {
...
-	 * @return string HTML Comment.
-	 */
-	public static function get_source_comment_end( $type, $name ) {
-		return sprintf( '<!--/amp-source-stack:%s:%s-->', $type, $name );
+	public static function get_source_comment( array $source, $is_start = true ) {
+		unset( $source['reflection'] );
+		return sprintf(
+			'<!--%samp-source-stack %s-->',
+			$is_start ? '' : '/',
+			str_replace( '--', '', wp_json_encode( $source ) )
```
kienstra: It's nice how this adds the json-encoded source data to the comment.

PR #1013, opened 2018-03-12
Facilitate using AMP components outside of AMP documents (AMP in PWA, “dirty AMP”)

### LINKED ISSUE #982: Ensure support for Dirty AMP
AMP can be used as a component library even on non-AMP pages. This is called “Dirty AMP”. When a theme is doing this they will have to enqueue the AMP component scripts on their own whereas when AMP is active it will automatically add the component scripts to the `head` based on which components are actually used in the page. So if AMP components are being used like this on non-AMP pages then we need to make sure that the AMP component scripts are not duplicated.

westonruter:
* Add AMP scripts the WordPress way: register all AMP component scripts as defined in spec via WP_Scripts via new function `amp_register_default_scripts()`, allowing simple enqueueing via `wp_enqueue_script( 'amp-list' )`, for example.
* Also register the AMP Shadow API as a default script.
* Use `wp_print_scripts()` with filtering `script_loader_tag` to output scripts with necessary attributes. See https://core.trac.wordpress.org/ticket/12009
* Update whitelist sanitizer to allow, validate, and sanitize AMP scripts.
* Eliminate needless amp_component_scripts filter.

With this in place, you can leverage AMP components entirely outside of pages served as AMP. For example, to add the quadradic formula to the footer of a site, served as AMP or not, all that is needed is:

```php
<?php
add_action( 'wp_enqueue_scripts', function() {
	wp_enqueue_script( 'amp-mathml' );
} );

	?>
	<amp-mathml
		layout="container"
		data-formula="\[x = {-b \pm \sqrt{b^2-4ac} \over 2a}.\]"
	></amp-mathml>
	<?php
} );
```

This not only adds the `amp-mathml` component script to the `head`, but it also automatically outputs the AMP runtime script, since it is marked as a dependency of all registered AMP component scripts.

Fixes #982.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -210,28 +210,15 @@ function amp_register_default_scripts( $wp_scripts ) {
 	$extensions = array();
 	foreach ( AMP_Allowed_Tags_Generated::get_allowed_tags() as $allowed_tag ) {
 		foreach ( $allowed_tag as $rule_spec ) {
-			if ( ! isset( $rule_spec[ AMP_Rule_Spec::TAG_SPEC ] ) ) {
-				continue;
-			}
-			$tag_spec = $rule_spec[ AMP_Rule_Spec::TAG_SPEC ];
-			if ( ! empty( $tag_spec['also_requires_tag_warning'] ) ) {
-				$extensions[] = strtok( $tag_spec['also_requires_tag_warning'][0], ' ' );
-			}
-			if ( ! empty( $tag_spec['requires_extension'] ) ) {
-				$extensions = array_merge( $extensions, $tag_spec['requires_extension'] );
+			if ( ! empty( $rule_spec[ AMP_Rule_Spec::TAG_SPEC ]['requires_extension'] ) ) {
+				$extensions = array_merge(
+					$extensions,
+					$rule_spec[ AMP_Rule_Spec::TAG_SPEC ]['requires_extension']
+				);
 			}
 		}
 	}
-
-	/**
-	 * List of components that are custom elements.
-	 *
-	 * Per the spec, "Most extensions are custom-elements." In fact, there is only one custom template.
-	 *
-	 * @link https://github.com/ampproject/amphtml/blob/cd685d4e62153557519553ffa2183aedf8c93d62/validator/validator.proto#L326-L328
-	 * @link https://github.com/ampproject/amphtml/blob/cd685d4e62153557519553ffa2183aedf8c93d62/extensions/amp-mustache/validator-amp-mustache.protoascii#L27
-	 */
-	$custom_templates = array( 'amp-mustache' );
+	$extensions = array_unique( $extensions );
 
 	foreach ( $extensions as $extension ) {
 		$src = sprintf(
@@ -246,15 +233,6 @@ function amp_register_default_scripts( $wp_scripts ) {
 			array( 'amp-runtime' ),
 			null
 		);
-		$attributes = array(
-			'async' => true,
-		);
-		if ( in_array( $extension, $custom_templates, true ) ) {
-			$attributes['custom-template'] = $extension;
-		} else {
-			$attributes['custom-element'] = $extension;
-		}
-		$wp_scripts->add_data( $extension, 'amp_script_attributes', $attributes );
 	}
 }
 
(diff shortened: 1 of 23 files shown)


westonruter:
It would be nice if this were further enhanced to automatically move the AMP scripts from the footer or anywhere else in the `body` to the `head` automatically. This would allow us to simplify the above to as follows:

```php
	wp_print_scripts( 'amp-mathml' );
	?>
	<amp-mathml layout="container" data-formula="\[x = {-b \pm \sqrt{b^2-4ac} \over 2a}.\]"></amp-mathml>
	<?php
} );
```

westonruter:
> It would be nice if this were further enhanced to automatically move the AMP scripts from the footer or anywhere else in the body to the head automatically.

This is done on 0154620.

includes/amp-helper-functions.php
```diff
@@ -167,13 +167,133 @@ function amp_get_asset_url( $file ) {
...
+	 * List of components that are custom elements.
+	 *
+	 * Per the spec, "Most extensions are custom-elements." In fact, there is only one custom template.
+	 *
+	 * @link https://github.com/ampproject/amphtml/blob/cd685d4e62153557519553ffa2183aedf8c93d62/validator/validator.proto#L326-L328
+	 * @link https://github.com/ampproject/amphtml/blob/cd685d4e62153557519553ffa2183aedf8c93d62/extensions/amp-mustache/validator-amp-mustache.protoascii#L27
+	 */
+	$custom_templates = array( 'amp-mustache' );
+
+	foreach ( $extensions as $extension ) {
```
westonruter: Probably should do an array_unique call here.

includes/amp-helper-functions.php
```diff
@@ -167,13 +167,133 @@ function amp_get_asset_url( $file ) {
...
+
+	// Get all AMP components as defined in the spec.
+	$extensions = array();
+	foreach ( AMP_Allowed_Tags_Generated::get_allowed_tags() as $allowed_tag ) {
+		foreach ( $allowed_tag as $rule_spec ) {
+			if ( ! isset( $rule_spec[ AMP_Rule_Spec::TAG_SPEC ] ) ) {
+				continue;
+			}
+			$tag_spec = $rule_spec[ AMP_Rule_Spec::TAG_SPEC ];
+			if ( ! empty( $tag_spec['also_requires_tag_warning'] ) ) {
```
westonruter: I don't think this statement is relevant here

includes/amp-helper-functions.php
```diff
@@ -167,13 +167,133 @@ function amp_get_asset_url( $file ) {
...
+		);
+		$attributes = array(
+			'async' => true,
+		);
+		if ( in_array( $extension, $custom_templates, true ) ) {
+			$attributes['custom-template'] = $extension;
+		} else {
+			$attributes['custom-element'] = $extension;
+		}
+		$wp_scripts->add_data( $extension, 'amp_script_attributes', $attributes );
```
westonruter: This really isn't necessary. In the script_loader_tag filter we can just look to see if the script comes from the AMP CDN and if so then automatically add async and the required custom attribute.

westonruter:
Ready for review.

kienstra:
**Reviewing Now**

Thanks, @westonruter. I'm reviewing this now.

includes/amp-helper-functions.php
```diff
@@ -167,13 +167,135 @@ function amp_get_asset_url( $file ) {
...
+	// Add each attribute (if it hasn't already been added).
+	foreach ( $attributes as $key => $value ) {
+		if ( ! preg_match( ":\s$key(=|>|\s):", $tag ) ) {
+			if ( true === $value ) {
+				$attribute_string = sprintf( ' %s', esc_attr( $key ) );
+			} else {
+				$attribute_string = sprintf( ' %s="%s"', esc_attr( $key ), esc_attr( $value ) );
+			}
+			$tag = preg_replace(
+				':(?=></script>):',
```
kienstra: Should the slash be escaped?

includes/class-amp-theme-support.php
```diff
@@ -119,6 +119,12 @@ public static function finish_init() {
 			self::register_paired_hooks();
 		}
 
+		// Enqueue AMP runtime.
+		wp_enqueue_script( 'amp-runtime' );
+
+		// Enqueue default styles expected by sanitizer.
+		wp_enqueue_style( 'amp-default', amp_get_asset_url( 'css/amp-default.css' ) );
```
kienstra: It looks like this is just moved from [enqueue_amp_default_styles()](https://github.com/Automattic/amp-wp/pull/1013/files#diff-2585472f207548f626a43a7ff3cab922L754), which only passed 2 arguments to `wp_enqueue_style()`. But could this use a `$version` as the 4th argument: `AMP__VERSION`?

includes/class-amp-theme-support.php
```diff
@@ -1032,13 +975,33 @@ public static function prepare_response( $response, $args = array() ) {
...
+			}
+		}
+
+		// Print all scripts, some of which may have already been printed and inject into head.
+		ob_start();
+		wp_print_scripts( array_keys( $amp_scripts ) );
+		$script_tags = ob_get_clean();
+		if ( ! empty( $script_tags ) ) {
+			$response = preg_replace(
+				'#(?=</head>)#',
```
kienstra: Should the slash be escaped?

includes/class-amp-theme-support.php
westonruter: Oh good point! Yes,  this is needed.

includes/amp-helper-functions.php
westonruter: No. This is doing a lookahead search to find `></script>` in the string to inject content before it.

PR #1015, opened 2018-03-12
Prevent showing details link on edit post screen if validation status post type is not registered

westonruter:
If the theme does not have `amp` support then the `amp_validation_status` is not registered, and the admin screen is not accessible. So there should be no link to access the edit post screen attempted to be presented, since `get_edit_post_link` will return empty since the user cannot edit the post type that doesn't exist.


kienstra:
**Reviewing Now**

I'm reviewing this PR now.

PR #1016, opened 2018-03-12
Keep track of the callbacks run (and their sources) when a filter is applied

westonruter:
* Wrap the_content and the_excerpt in AMP stack source comments to reveal which filters are responsible for the resulting content, since not any one filter callback can be identified as the culprit for invalid content.
* Remove spurious apostrophe from callback class name.
* Prevent PHP notice when doing summarize_validation_errors if a source has no type or name.

If I enter an excerpt that contains:

```html
This is <b onclick=EVIL>awesome</b> food
```

Now this results in the validation error containing the filter that is applied (`the_excerpt`), the current post global, and the filter callbacks added and their sources:

> image

This will greatly help with debugging AMP validation errors related to _filters_, where previously there was only good validation debugging for _actions_.

### DIFF
--- a/includes/utils/class-amp-validation-utils.php
+++ b/includes/utils/class-amp-validation-utils.php
@@ -185,6 +185,15 @@ class AMP_Validation_Utils {
 	public static $posts_pending_frontend_validation = array();
 
 	/**
+	 * Current sources gathered for a given hook currently being run.
+	 *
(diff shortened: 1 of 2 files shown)


westonruter:
Here's what it looks like when there is invalid content output by a shortcode: both the shortcode being applied _and_ the filter (`the_content`) is captured in the source stack trace:

image

PR #1018, opened 2018-03-13
Issue #845: Script to create Gutenberg test page

kienstra:
**Pull Request**

This PR for Issue #845 uses the [test fixtures](https://github.com/WordPress/gutenberg/tree/master/blocks/test/fixtures) in Gutenberg, but excludes the embeds. Those fixtures usually don't have actual URLs.

From the plugin root, run this command from inside an environment like [VVV](https://github.com/Varying-Vagrant-Vagrants/VVV/):

`bash bin/create-gutenberg-test-post.sh`

This clones Gutenberg if it's not present, or pulls the master branch. Then, it runs the `wp-cli` script to create the test page, based on the [test fixtures](https://github.com/WordPress/gutenberg/tree/master/blocks/test/fixtures).

- [x] Add some permutations (see #845)
- [x] Add documentation in [contributing.md](https://github.com/Automattic/amp-wp/blob/develop/contributing.md)
- [x] Create a test [reusable block](https://github.com/WordPress/gutenberg/tree/master/blocks/library/block)
- [x] Test this more

This uses logic from [create-embed-test-post.php](https://github.com/Automattic/amp-wp/blob/develop/bin/create-embed-test-post.php) and [amphtml-update.sh](https://github.com/Automattic/amp-wp/blob/develop/bin/amphtml-update.sh)

### DIFF
--- a/bin/create-gutenberg-test-post.php
+++ b/bin/create-gutenberg-test-post.php
@@ -59,7 +59,7 @@ function amp_get_block_permutations() {
 		),
 		array(
 			'title'   => 'Columns, With 2 Columns',
-			'content' => '<!-- wp:columns {"columns":2} --><div class="wp-block-columns has-2-columns"><!-- wp:paragraph {"layout":"column-1"} -->	<p class="layout-column-1">Column One, Paragraph One</p><!-- /wp:paragraph --><!-- wp:paragraph {"layout":"column-1"} --><p class="layout-column-1">Column One, Paragraph Two</p><!-- /wp:paragraph --><!-- wp:paragraph {"layout":"column-2"} --><p class="layout-column-2">Column Two, Paragraph One</p><!-- /wp:paragraph --></div><!-- /wp:columns -->',
+			'content' => '<!-- wp:columns {"columns":2} --><div class="wp-block-columns has-2-columns"><!-- wp:paragraph {"layout":"column-1"} --><p class="layout-column-1">Column One, Paragraph One</p><!-- /wp:paragraph --><!-- wp:paragraph {"layout":"column-1"} --><p class="layout-column-1">Column One, Paragraph Two</p><!-- /wp:paragraph --><!-- wp:paragraph {"layout":"column-2"} --><p class="layout-column-2">Column Two, Paragraph One</p><!-- /wp:paragraph --></div><!-- /wp:columns -->',
 		),
 		array(
 			'title'   => 'Cover Image With Fixed Background',
--- a/bin/create-gutenberg-test-post.php
+++ b/bin/create-gutenberg-test-post.php
@@ -104,14 +104,15 @@ function amp_get_block_permutations() {
  */
 function amp_create_gutenberg_test_post( $content ) {
 	$slug            = 'amp-test-gutenberg-blocks';
-	$page            = get_page_by_path( $slug );
+	$title           = 'AMP Test Gutenberg Blocks';
+	$page            = get_page_by_path( "/{$slug}/" );
 	$failure_message = 'The test page could not be added, please try again.';
 	if ( $page ) {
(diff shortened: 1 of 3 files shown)


kienstra:
**Request For Review**

Could you please review this pull request for #845?

bin/create-gutenberg-test-post.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
kienstra: Is this script needed? It clones Gutenberg if it's not present, then runs [create-gutenberg-test-post.php](https://github.com/Automattic/amp-wp/pull/1018/files#diff-2520e4e1c81073b0b11d09153c68d994R1).

bin/create-gutenberg-test-post.php
```diff
@@ -0,0 +1,168 @@
...
+			'title'   => '(Reusable) Block With Video',
+			'content' => sprintf( '<!-- wp:block {"ref":%d} /-->', amp_create_reusable_block() ),
+		),
+		array(
+			'title'   => 'Categories With Dropdown',
+			'content' => '<!-- wp:core/categories {"showPostCounts":false,"displayAsDropdown":true,"showHierarchy":false} /-->',
+		),
+		array(
+			'title'   => 'Columns, With 2 Columns',
+			'content' => '<!-- wp:columns {"columns":2} --><div class="wp-block-columns has-2-columns"><!-- wp:paragraph {"layout":"column-1"} -->	<p class="layout-column-1">Column One, Paragraph One</p><!-- /wp:paragraph --><!-- wp:paragraph {"layout":"column-1"} --><p class="layout-column-1">Column One, Paragraph Two</p><!-- /wp:paragraph --><!-- wp:paragraph {"layout":"column-2"} --><p class="layout-column-2">Column Two, Paragraph One</p><!-- /wp:paragraph --></div><!-- /wp:columns -->',
```
kienstra: This might need to be `wp:core/columns` instead of `wp:columns`, and likewise for the rest of the line.

kienstra: Applied with [3e41cca](https://github.com/Automattic/amp-wp/pull/1018/commits/3e41cca7395724274a3efb7833ce14a1d7ee70d0).

bin/create-gutenberg-test-post.sh
westonruter: It doesn't hurt, but it probably isn't necessary. The `amp_create_gutenberg_test_post()` function could just throw an exception if Gutenberg isn't active (e.g. if `! function_exists( 'do_blocks' )`)

bin/create-gutenberg-test-post.php
westonruter: Gutenberg creates blocks with `wp:columns`: https://github.com/Automattic/amp-wp/pull/1019/files#diff-ae6e35f6dd94607ca0bb8954533a1cb5R174

But they are synonymous. The `core/` is implict when no namespace is present: https://github.com/WordPress/gutenberg/blob/11187b4093df792f4a5caefc765acca98d5fbb04/lib/blocks.php#L163-L164

bin/create-gutenberg-test-post.php
```diff
@@ -0,0 +1,168 @@
...
+ * @package AMP
+ */
+
+/**
+ * Gets many of the Gutenberg fixture blocks in blocks/tests/fixtures/.
+ *
+ * @throws Exception If this is script is not run inside the plugin directory.
+ * @return string $content Post content with all Gutenberg blocks.
+ */
+function amp_get_blocks() {
```
westonruter: Since we're using PHP 5.3, let's just go ahead and use namespaces for this file. This will help prevent confusion with autocompletion. So then this could be renamed to just `get_test_data_blocks()`.

PR #1019, opened 2018-03-13
Add AMP validation checking for Gutenberg blocks

### LINKED ISSUE #1009: [Gutenberg] Integrate validator with Gutenberg blocks
**As a WordPress site running Gutenberg, my core blocks' AMP-HTML markup should be validated and results should appear in the Compatibility Tool.**

- [x] AC1: Integrate block validation results with the UI to reveal when a validation error does occur. Normally such errors will not happen for core blocks, but for the Custom HTML block in particular it is a distinct possibility. It is also key for plugins which introduce blocks so that their AMP validity can also be reported.

westonruter:
- [x] Add source comments around each Gutenberg block to track validation errors.
- [x] Pass validation errors back in REST API response when Gutenberg saves a post. (Discovery needed.)
- [x] Show validation status in editor after saving post. 

An invalid `script` inside of an HTML block which is inside of a Column block has sources that look like this:

image

Fixes #1009.


westonruter:
@kienstra feel free to pick this up from me.

kienstra:
Thanks, @westonruter! It'll be great to work on this PR. If it's alright, tomorrow is probably the earliest I could get to it. I'm working on #1018 now.

westonruter:
There is a bug with how Mustache replacements are being done which is corrupting the JSON in the block attributes. I'm working on a fix for that. Not a blocker for work being done on this.

tests/test-class-amp-validation-utils.php
```diff
@@ -131,6 +131,89 @@ public function test_add_validation_hooks() {
...
+			'latest_posts' => array(
+				'<!-- wp:latest-posts /-->',
+				'<!--amp-source-stack {"block_name":"core\/latest-posts","type":"plugin","name":"gutenberg","function":"render_block_core_latest_posts"}--><ul class="wp-block-latest-posts aligncenter"></ul><!--/amp-source-stack {"block_name":"core\/latest-posts","type":"plugin","name":"gutenberg","function":"render_block_core_latest_posts"}-->',
+				array(
+					'element' => 'ul',
+					'blocks'  => array( 'core/latest-posts' ),
+				),
+			),
+			'columns'      => array(
+				"<!-- wp:columns -->\n<div class=\"wp-block-columns has-2-columns\">\n    <!-- wp:quote {\"layout\":\"column-1\",\"foo\":{\"bar\":1}} -->\n    <blockquote class=\"wp-block-quote layout-column-1\">\n        <p>A quotation!</p><cite>Famous</cite></blockquote>\n    <!-- /wp:quote -->\n\n    <!-- wp:html {\"layout\":\"column-2\"} -->\n    <div class=\"layout-column-2\">\n        <script>\n            document.write('Not allowed!');\n        </script>\n    </div>\n    <!-- /wp:html -->\n</div>\n<!-- /wp:columns -->",
```
westonruter: Gutenberg creates blocks with `wp:columns`: https://github.com/Automattic/amp-wp/pull/1019/files#diff-ae6e35f6dd94607ca0bb8954533a1cb5R174

But they are synonymous. The `core/` is implict when no namespace is present: https://github.com/WordPress/gutenberg/blob/11187b4093df792f4a5caefc765acca98d5fbb04/lib/blocks.php#L163-L164

westonruter:
Rebased to include test page generated from #1018. Former `head` was 4a10ee4.

includes/utils/class-amp-validation-utils.php
```diff
@@ -303,6 +303,97 @@ public static function add_validation_hooks() {
...
+			// Prepend rendered block with start source comment.
+			if ( $source ) {
+				$rendered_block = self::get_source_comment( $source, true ) . $rendered_block;
+			}
+
+			// Render nested blocks.
+			if ( ! empty( $block['innerBlocks'] ) ) {
+				$rendered_inner_blocks = self::render_blocks( $block['innerBlocks'] );
+
+				// Inject rendered inner blocks before closing tag of outer block (or at end of string if no closing). @todo Confirm approach here.
```
westonruter: I've raised this in WordPress `#core-editor`:

image


https://wordpress.slack.com/archives/C02QB2JS7/p1521043684000458

westonruter:
Let's not merge this as part of 0.7 since nested blocks are still in flux.

westonruter:
@kienstra Here's an idea: what if we piggy-backed on the autosave functionality to incorporate the validation results? This could avoid the need to trigger an update when changing each block. We would be piggy-backing on the autosave interval. 

Upon saving draft we'll want to do the loopback request to get the full validation results anyway and display them in a notification. (Also, a block may enqueue scripts which would need to be caught but `process_markup` does not currently do that since it does not do `wp_head` or `wp_footer`. ) The results after this save we'll want to add to the blocks. So in addition to there being a validation message at the top, there should also be the specific validation errors added to their respective individual blocks inside so that users know which ones specifically are problematic.

In short, I think perhaps this REST API endpoint should not be used, but rather a new REST _field_ should be added to the existing post/page endpoints. When a `POST`/`PUT` request is made to create/update a post, this can trigger the same loopback request we're doing during `save_post` now, and then the array of validation errors could be returned as the contents of the field for the client to then display in the notification error and with each block. This could be done both during autosave and regular save requests.

kienstra:
**Validation On Saving Post**

Hi @westonruter,
Thanks, that's a good idea to make a request to get the validation results on updating the post. The block `edit()` functions can fire a few times a second, so it was going to be hard to hook into those or [componentDidMount()](https://github.com/Automattic/amp-wp/pull/1019/files#diff-b5da96daddf181169cf709cf45f4e478R122).

I'll look at how we might display the notices when the page autosaves. So far, it looks [setting the state](https://github.com/Automattic/amp-wp/pull/1019/files#diff-b5da96daddf181169cf709cf45f4e478R161) of the `AMPNotice` component from outside the block might be hard. But there's probably another approach.

westonruter:
> So far, it looks setting the state of the AMPNotice component from outside the block might be hard. But there's probably another approach.

I think that it shouldn't actually be part of the state then in this case. Instead the AMP validation results should be stored in the store that then sends the data down as props. So then you just update the store with the validation results, and then the edit function for a block can determine (somehow) which validation errors are relevant to itself and then display them.

I'm guessing this is what you'll want to use: https://github.com/WordPress/gutenberg/tree/master/data

kienstra:
**Thanks**

Hi @westonruter,
Thanks, that looks like a better approach.

includes/utils/class-amp-validation-utils.php
```diff
@@ -1791,4 +1878,60 @@ public static function get_recheck_link( $post, $redirect_url, $recheck_url = nu
...
+
+	/**
+	 * Adds a field to the REST API responses to display the validation status.
+	 *
+	 * @param array  $post_data  Data for the post.
+	 * @param string $field_name The name of the field to add.
+	 * @return array|null $validation_data Validation data if it's available, or null.
+	 */
+	public static function rest_field_amp_validation( $post_data, $field_name ) {
+		$validation_post = self::get_validation_status_post( $post_data['link'] );
```
westonruter: Excellent re-use of this function. 

One key thing is that instead of `$post_data['link']` it should do `amp_get_permalink( $post_data['id'] )` so that it can work in paired mode.

Idea: If this post is empty or the post type is not viewable, should we take the post content and try to validate it? In other words, this:

https://github.com/Automattic/amp-wp/blob/0bd1d58451d7a78b125e8eac2f1b7d02ebfb124a/includes/utils/class-amp-validation-utils.php#L549-L560

includes/utils/class-amp-validation-utils.php
```diff
@@ -1869,4 +1878,60 @@ public static function get_recheck_link( $post, $redirect_url, $recheck_url = nu
...
+	}
+
+	/**
+	 * Adds fields to the REST API responses, in order to display validation errors.
+	 *
+	 * @return void
+	 */
+	public static function add_rest_api_fields() {
+		register_rest_field(
+			array( 'post', 'page' ),
```
westonruter: When `! amp_is_canonical()` then this should be the REST object types for all post types that have `amp` post type support. Otherwise, it should be added for all post types which have the `editor` post type support.

assets/js/amp-block-validation.js
```diff
@@ -0,0 +1,185 @@
...
+		 *
+		 * @see https://riad.blog/2017/10/16/one-thousand-and-one-way-to-extend-gutenberg-today/
+		 * @param {string} blockType the type of the block, like 'core/paragraph'.
+		 * @returns {void}
+		 */
+		overwriteEdit: function( blockType ) {
+			let block = wp.blocks.unregisterBlockType( blockType );
+			let OriginalBlockEdit = block.edit;
+
+			block.edit = class AMPNotice extends wp.element.Component {
```
westonruter: ES6 classes aren't going to work in IE11.

assets/js/amp-block-validation.js
```diff
@@ -0,0 +1,185 @@
...
+			};
+
+			wp.blocks.registerBlockType( blockType, block );
+		}
+
+	};
+
+	return module;
+
+} )( window.jQuery );
```
westonruter: We should eliminate jQuery as being a dependency for this script. Extending Gutenberg should be jQuery-free.

includes/utils/class-amp-validation-utils.php
kienstra: Good idea, @westonruter. [This commit](https://github.com/Automattic/amp-wp/pull/1019/commits/ed4b03ba8a01dd276cc79378456d100a84088552) abstracts that snippet above into [get_existing_validation_errors()](https://github.com/Automattic/amp-wp/pull/1019/commits/ed4b03ba8a01dd276cc79378456d100a84088552#diff-7d234989187d87a372e2aa6aa1299a71R636), and reuses it here. If there are no existing errors, it [validates the front-end](https://github.com/Automattic/amp-wp/pull/1019/commits/ed4b03ba8a01dd276cc79378456d100a84088552#diff-7d234989187d87a372e2aa6aa1299a71R1968). 

The only issue is that this delays the response if it validates several posts.

assets/js/amp-block-validation.js
kienstra: Thanks, @westonruter. [This commit](https://github.com/Automattic/amp-wp/pull/1019/commits/ea1c37af19209b6851b261cb115105a5b5367bb4) removes the ES6 class.

assets/js/amp-block-validation.js
kienstra: Thanks, @westonruter. [This commit](https://github.com/Automattic/amp-wp/pull/1019/commits/ea1c37af19209b6851b261cb115105a5b5367bb4) removes the `jQuery` dependency.

assets/js/amp-block-validation.js
```diff
@@ -0,0 +1,140 @@
...
+			}
+
+			currentPost[ module.data.errorKey ].forEach( function( validationError ) {
+				if ( validationError.hasOwnProperty( 'sources' ) ) {
+					validationError.sources.forEach( function( source ) {
+						if ( source.hasOwnProperty( 'block_name' ) ) {
+							if ( blocksWithErrors.hasOwnProperty( source.block_name ) ) {
+								blocksWithErrors[ source.block_name ].push( validationError ); // jscs:ignore requireCamelCaseOrUpperCaseIdentifiers
+							} else {
+								blocksWithErrors[ source.block_name ] = [ validationError ]; // jscs:ignore requireCamelCaseOrUpperCaseIdentifiers
```
westonruter: I think it would be very useful to capture the `block_attrs` as well here. The `block_name` and `block_attrs` together are what could help to narrow down the validation errors for a given block in `getBlockValidationErrors`.

assets/js/amp-block-validation.js
```diff
@@ -0,0 +1,140 @@
...
+		 * Gets the validation errors for a specific block if they exist.
+		 *
+		 * In module.blocksWithErrors, the errors are stored by block name.
+		 * This finds the validation errors for a specific block, based on its content.
+		 *
+		 * @todo: search more specifically, as this currently only checks for matching of the node_name and node_attributes.
+		 * @param {Object} props Properties for the block.
+		 * @return {Array} The validation error(s) for the block, or an empty array.
+		 */
+		getBlockValidationErrors: function( props ) {
```
westonruter: Per above, I think there should be two args passed in here: `name` and `attributes`.

westonruter: Correction: I see there is already `props.attributes`. So this could then be used to look up `rawErrors`.

assets/js/amp-block-validation.js
kienstra: Thanks, @westonruter. That's a good idea. [These commits](https://github.com/Automattic/amp-wp/compare/743cca49a132a147085125a274e23a0bb62c057c...b9aa16c39b1452b0eee17c76305bba48f5918887) add `block_attrs` to `blocksWithErrors`.

kienstra:
**Current UI**

The errors now appear inline, above individual blocks:

notice-error-codes

Here's a [screencast](https://drive.google.com/file/d/1nQEKdNgh0utHyzWBlaAOMdVcswM4V26Z/view).

This now displays these errors:
* `invalid_element`
* `invalid_attribute`

It doesn't display these errors:
* `css_overflow_property_removed`
* `css_important_removed`
* `enqueued_script`
This does detect inline scripts, like in the 'Categories' block with a dropdown. It [looks like](https://wp.zacgordon.com/2017/12/26/how-to-add-javascript-and-css-to-gutenberg-blocks-the-right-way-in-plugins-and-themes/) the best practice for blocks to enqueue scripts is with the action `enqueue_block_assets`, and it would probably be hard to find the exact block. But if the block enqueued a script in the `render_callback`, this wouldn't detect it.

kienstra:
**Next Steps**

This is mainly ready for code review, but there are also some console errors to address:

* `react-dom.d3299c8d.js:894 Warning: Can only update a mounted or mounting component. `
* `Uncaught TypeError: Cannot read property 'dom' of undefined`

westonruter:
This is looking good! Would it be possible to avoid linking off to the validation error admin screen and instead allow the details to be rendered in the notification upon expansion? This would allow, for example, the validation errors to be displayed for posts that don't have a public URL which is required for a validation error to be logged.

kienstra:
Hi @westonruter, thats a good idea to display the validation errors on expanding them. I'll look at how to do that.

kienstra:
**Possible Notice UI**

Hi @westonruter,
What do you think of this possible UI for notices?

error-notice-collapsed

On clicking to expand the panel:
panel-expanded

assets/js/amp-block-validation.js
```diff
@@ -0,0 +1,166 @@
...
+		},
+
+		/**
+		 * Gets the block types with errors.
+		 *
+		 * Iterates through the 'amp_validation_errors' from the REST API response.
+		 * This returns an object, with block types as the keys, and error arrays as the values.
+		 * The block's overridden edit() method can then get the errors for its block type.
+		 * This finds the validation errors for a specific block, based on its attributes or content.
+		 * @todo: keep refining how this finds if the errors match.
```
westonruter: I think I know what we need to do here. We need to grab the validation errors, group them by `postId` and `blockIndex`, then align with `getBlockOrder` upon update of the `restValidationErrorsField`. I'll push up a bit more on this tomorrow.

includes/utils/class-amp-validation-utils.php
```diff
@@ -1794,6 +1917,85 @@ public static function get_recheck_link( $post, $redirect_url, $recheck_url = nu
...
+	 * @param string          $field_name The name of the field to add.
+	 * @param WP_REST_Request $request    The name of the field to add.
+	 * @return array|null $validation_data Validation data if it's available, or null.
+	 */
+	public static function rest_field_amp_validation( $post_data, $field_name, $request ) {
+		$post_id = $post_data['id'];
+		$post    = get_post( $post_id );
+		if ( in_array( $request->get_method(), array( 'PUT', 'POST' ), true ) ) {
+			if ( ! isset( self::$posts_pending_frontend_validation[ $post_id ] ) ) {
+				self::$posts_pending_frontend_validation[ $post_id ] = true;
```
westonruter: I just realized we may need to do something special during autosave here.

assets/js/amp-block-validation.js
```diff
@@ -0,0 +1,166 @@
...
+		/**
+		 * Update blocks' validation errors in the store.
+		 *
+		 * @param {Object[]} validationErrors - Validation errors.
+		 * @return {void}
+		 */
+		updateBlocksValidationErrors: function updateBlocksValidationErrors( validationErrors ) {
+			var blocksValidationErrors, blockOrder, currentPost, selectors;
+			selectors = wp.data.select( 'core/editor' );
+			blockOrder = selectors.getBlockOrder();
```
westonruter: There is a race condition here where:

1. I add an AMP invalidity to the first block.
2. I hit Save.
3. The server is slow to respond, and while waiting to respond I move a block.
4. When the server responds, the block indices in the validation errors are no longer aligned with the order of blocks in the editor and so validation errors are mismatched.

To fix this problem, we need to capture the `blockOrder` at the moment that saving is initiated instead of using the state at the time that saving is finished.

assets/js/amp-block-validation.js
```diff
@@ -0,0 +1,177 @@
...
+					getBlockValidationErrors: function( state, uid ) {
+						return state.blockValidationErrorsByUid[ uid ] || [];
+					}
+				}
+			} );
+
+			wp.data.subscribe( function() {
+				var currentPost, thisValidationErrors;
+
+				// @todo Gutenberg currently is not persisting isDirty state if changes are made during save request. Block order mismatch.
```
westonruter: See https://github.com/WordPress/gutenberg/issues/6112

assets/js/amp-block-validation.js
```diff
@@ -0,0 +1,177 @@
...
+
+				// Find the inner-most nested block source only; ignore any nested blocks.
+				for ( i = validationError.sources.length - 1; 0 <= i; i-- ) {
+					source = validationError.sources[ i ];
+
+					if ( ! source.block_name || postId !== source.post_id ) {
+						continue;
+					}
+
+					// @todo Cross-check source.block_name with wp.data.select( 'core/editor' ).getBlock( matchedBlockUid ).name?
```
westonruter: Indeed, this is going to be required because my approach here to use the `blockOrder` along is somewhat naïve since a plugin could _add_ or _remove_ additional blocks during `the_content` filter. Nevertheless, this should hopefully be an edge case.

westonruter:
This is currently blocked by https://github.com/WordPress/gutenberg/issues/6134

westonruter:
Autosaves just got improved as well, so this needs to be accounted for: https://github.com/WordPress/gutenberg/pull/4156

westonruter:
@kienstra this is now unblocked and ready for review/merge. The latest state of `master` in Gutenberg works with this PR (so first do `git pull; npm install; npm run build` for that plugin).

There will be more work to do on it of course, but the initial Gutenberg block validation seems important to get merged to start testing and _using_. 

The easiest way to test this is to add a Custom HTML block and add a `script` tag to it. Upon saving you should see the notification at the top and with the block. Upon removal of the `script` and re-saving you should see the notice with the block disappear. There may be non-block validation errors (e.g. regarding enqueued CSS) which persist at the top and this is expected. You can ignore this warning in the console because it is regarding a problem with the Custom HTML block:

> Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.

kienstra:
**Approved**

Hi @westonruter,
Great job with the Gutenberg validation here. It displays errors where expected:

amp-validation-error-gutenberg

It's nice how the notice at the top of the editor says "But none are directly due to content here."

And as expected, there was no block validation error on the page created by the `WP-CLI` script [create-gutenberg-test-post.php](https://github.com/Automattic/amp-wp/blob/develop/bin/create-gutenberg-test-post.php).

Also, maybe there could be a `margin-left: 40px` to "indent" the errors slightly:

block-margin-left

includes/utils/class-amp-validation-utils.php
```diff
@@ -1794,6 +1924,108 @@ public static function get_recheck_link( $post, $redirect_url, $recheck_url = nu
...
+	 * First, get existing errors for the post.
+	 * If there are none, validate the post and return any errors.
+	 *
+	 * @param array           $post_data  Data for the post.
+	 * @param string          $field_name The name of the field to add.
+	 * @param WP_REST_Request $request    The name of the field to add.
+	 * @return array|null $validation_data Validation data if it's available, or null.
+	 */
+	public static function get_amp_validity_rest_field( $post_data, $field_name, $request ) {
+		unset( $field_name );
```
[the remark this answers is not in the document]
westonruter: It's to prevent PhpStorm from flagging an unused function argument.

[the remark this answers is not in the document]
kienstra: Thanks, that makes sense.

kienstra:
**Request For Testing**

Hi @csossi,
Could you please test this?

1. Go to https://native-ampconfdemo.pantheonsite.io/
2. Create a new post
3. Add a Custom HTML block 
4. Put this in the block:
```html
<script>doThis();</script>
<button onclick="doSomething();">
	Click Me
</button>
```
5. Publish the post
6. Expected: there's a notice above the block, and at the top of the editor:
post-validation
7. Expected: clicking "Review issues" in the top notice leads to a page "Invalid AMP Page (URL)"

csossi:
verified in QA
image

kienstra:
**Request For Regression Testing**

 Hi @csossi,
Thanks for testing this. Could you please retest this to ensure there wasn't a regression, using the [test steps above](https://github.com/Automattic/amp-wp/pull/1019#issuecomment-395772343)?

Also, could you please test for edge cases by creating several blocks, moving and duplicating blocks, and other combinations?

Here's the staging site:
https://paired-ampconfdemo.pantheonsite.io/

csossi:
verified in QA

PR #1020, opened 2018-03-15
#936 - comment confirmation message

### LINKED ISSUE #936: Change comment confirmation message based on approval settings
As mentioned in https://github.com/Automattic/amp-wp/issues/797#issuecomment-363272601 and https://github.com/Automattic/amp-wp/issues/862#issuecomment-363275416
The message indicated "..subject to moderation".
This message should be adjusted based on the moderation settings.
An indication of refresh times should also be mentioned.

DavidCramer:
This adds a setting in the discussion page to allow for the changing of the success message when posting a comment.

Fixes #936

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -567,7 +567,7 @@ class AMP_Theme_Support {
 		?>
 		<p>
 			<label for="amp_comment_notice">
-				<?php echo esc_html__( 'This message is to give a user posting a comment that it was successful, but there will be a delay.', 'amp' ); ?>
+				<?php echo esc_html__( 'This message is to indicate to a user that the comment was posted, but there will be a delay in displaying it.', 'amp' ); ?>
 			</label>
 		</p>
 		<p>
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -550,8 +550,12 @@ class AMP_Theme_Support {
 	 * @since 0.7
 	 */
 	public static function get_comment_moderation_notice() {
-		$default = __( 'Your comment has been posted, but may be subject to moderation.', 'amp' );
-		return get_option( 'amp_comment_notice', $default );
+		$default = __( 'Your comment has been posted and will show shortly.', 'amp' );
+		$message = trim( get_option( 'amp_comment_notice' ) );
+		if ( empty( $message ) ) {
+			$message = $default;
+		}
+		return $message;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


DavidCramer:
initially we were going to use an alternate success message if the user has had an accepted comment before, however this is more complex to do since you need to check etc.
Also, when you post a comment that is to be moderated first, the comment still shows but has a note stating the comment is awaiting moderation.
So the function of this is more to indicate to the user that the comment will show shortly.

screen shot 2018-03-15 at 12 06 39

DavidCramer:
@westonruter this is ready for review now. I'm unsure of the approach. I think it's ok, but a second opinion will be good.

westonruter:
@DavidCramer I think there may be a better way. We actually can determine if the comment was approved via in the `\AMP_Theme_Support::handle_xhr_request()` call. We can pass back the approval status by changing it to look like this:

```php
add_filter( 'comment_post_redirect', function( $location, $comment ) {
	wp_send_json( array(
		'approved' => '1' === (string) $comment->comment_approved,
	) );
}, PHP_INT_MAX, 2 );
```

Then inside of `AMP_Theme_Support::add_amp_comment_form_templates` it can do a conditional message based on this data returned in `AMP_Theme_Support::add_amp_comment_form_templates()`, for example:

```php
<div submit-success>
		{{#approved}}
			<?php esc_html_e( 'Your comment has been posted.', 'amp' ); ?>
		{{/approved}}
		{{^approved}}
			<?php esc_html_e( 'Your comment is awaiting moderation.', 'default' ); ?>
		{{/approved}}
	</template>
</div>
```

Note how it's re-using the default moderation string in core. I couldn't find a default message regarding a comment being accepted, however.

westonruter:
@DavidCramer on second thought, it may be better to return the message in the same way is being done for the die handler:

```php
add_filter( 'comment_post_redirect', function( $location, $comment ) {
	if ( '1' === (string) $comment->comment_approved ) {
		$message = __( 'Your comment has been approved.', 'amp' );
	} else {
		$message = __( 'Your comment is awaiting moderation.', 'default' );
	}
	$message = apply_filters( 'amp_comment_submitted_message', $message, $comment );
	wp_send_json( array(
		'message' => $message,
		'approved' => $comment->comment_approved,
	) );
}, PHP_INT_MAX, 2 );
```

Then inside of `AMP_Theme_Support::add_amp_comment_form_templates()` it can just do:

```php
<div submit-success>
		{{{message}}}
	</template>
</div>
```

This will allow for the message to be filtered and for plugins to add support for custom comment statuses.

PR #1022, opened 2018-03-15
#1010: [Gutenberg] Fix issues with rendering native blocks

### LINKED ISSUE #1010: [Gutenberg] Fixing any issues with rendering native blocks in AMP
**As a WordPress site running Gutenberg, my core/native blocks should be renderable as AMP-HTML compatible markup.**

AC1: Any issues previously identified with core blocks from #845 should be fixed here.

miina:
Fixes the issue for `Categories` with dropdown found within #845.

Wraps `select` with `form` tags and adds `on:change` to `select`.

Fixes #1010 .


includes/amp-helper-functions.php
```diff
@@ -424,22 +424,23 @@ function amp_get_content_embed_handlers( $post = null ) {
...
-			'AMP_Meetup_Embed_Handler'      => array(),
-			'AMP_Vine_Embed_Handler'        => array(),
-			'AMP_Facebook_Embed_Handler'    => array(),
-			'AMP_Pinterest_Embed_Handler'   => array(),
-			'AMP_Playlist_Embed_Handler'    => array(),
-			'AMP_Reddit_Embed_Handler'      => array(),
-			'AMP_Tumblr_Embed_Handler'      => array(),
-			'AMP_Gallery_Embed_Handler'     => array(),
-			'WPCOM_AMP_Polldaddy_Embed'     => array(),
+			'AMP_Gutenberg_Categories_Handler' => array(),
```
westonruter: Since “Gutenberg“ project name will not be carried over when merged into core, we should just call it `AMP_Categories_Block_Handler`.

includes/embeds/class-amp-gutenberg-categories-handler.php
```diff
@@ -0,0 +1,98 @@
...
+ * @since 1.0
+ */
+class AMP_Gutenberg_Categories_Handler extends AMP_Base_Embed_Handler {
+
+	/**
+	 * Register embed.
+	 */
+	public function register_embed() {
+		if ( function_exists( 'gutenberg_init' ) ) {
+			add_action( 'the_post', array( $this, 'override_category_block_render_callback' ) );
```
westonruter: I don't think this needs to wait for `the_post` action. I think it can just do `$this-> override_category_block_render_callback()`

includes/embeds/class-amp-gutenberg-categories-handler.php
```diff
@@ -0,0 +1,98 @@
...
+	/**
+	 * Unregister embed.
+	 */
+	public function unregister_embed() {}
+
+	/**
+	 * Override the output of Gutenberg core/category block.
+	 */
+	public function override_category_block_render_callback() {
+		if ( is_amp_endpoint() ) {
```
westonruter: Note that `register_embed` will have only been called if it is an AMP endpoint to begin with. So this `override_category_block_render_callback` method can actually be eliminated in favor of moving the logic to `register_embed` directly, minus the `is_amp_endpoint` check.

includes/embeds/class-amp-gutenberg-categories-handler.php
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @package AMP
+ */
+
+/**
+ * Class AMP_Gutenberg_Categories_Handler
+ *
+ * @since 1.0
+ */
+class AMP_Gutenberg_Categories_Handler extends AMP_Base_Embed_Handler {
```
westonruter: As noted above, the “Gutenberg” name is going to be dropped with the core merge so we can change it to just `AMP_Categories_Block_Handler`. But given that there is only one block that needs to be overridden, maybe it would be better to just call it `AMP_Core_Block_Handler`. This then can include the logic for the `categories` block and we can add support for any other core blocks that need special handling instead of creating a separate class for each, which I think is somewhat overkill.

includes/embeds/class-amp-gutenberg-categories-handler.php
```diff
@@ -0,0 +1,98 @@
...
+	public function register_embed() {
+		if ( function_exists( 'gutenberg_init' ) ) {
+			add_action( 'the_post', array( $this, 'override_category_block_render_callback' ) );
+		}
+	}
+
+	/**
+	 * Unregister embed.
+	 */
+	public function unregister_embed() {}
```
westonruter: This should undo what is done in `register_embed`. The original `render_callback` for the `core/categories` block should be saved and then restored here.

Issue #1014, opened 2018-03-12
Include AMP content in REST API responses for embedding (AMP in PWA)


westonruter:
In #1013 a script for the [Shadow AMP API](https://www.ampproject.org/docs/integration/pwa-amp/amp-in-pwa) is registered. To facilitate AMP in PWA there needs to be an easy way to obtain AMP content to embed into a page. One easy way would just be to add an `amp` property to go alongside the [existing `raw` and `rendered` properties of the `content` field](https://github.com/WordPress/wordpress-develop/blob/5f56921131043db31353f795c2123fbd4fa13671/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L1503-L1508).

The logic used to obtain the `content.amp` value would be similar to:

https://github.com/ampproject/amp-wp/blob/e2c5b193a6350282cf9387c7890fc99c97ab0156/includes/utils/class-amp-validation-utils.php#L357-L367

The REST API endpoint is extensible for plugins to add a new property here. For example, the [`rest_prepare_{$this->post_type}`](https://github.com/WordPress/wordpress-develop/blob/5f56921131043db31353f795c2123fbd4fa13671/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L1606) filter could be used. Nevertheless, the schema should also be updated to advertise this new property.

westonruter:
See also https://github.com/WordPress/gutenberg/issues/7668

ataylorme:
A second endpoint for the tree shaken CSS would be great too

luisherranz:
An array of scripts would be useful as well:

```js
{
  ...
  content: {
    rendered: "....<iframe src="https://youtube....></iframe>",
    amp: "...<amp-youtube ...>..."
  },
  amp: {
    styles: "....",
    scripts: [
      {
        "custom-element": "amp-youtube",
        src: "https://cdn.ampproject.org/v0/amp-youtube-0.1.js"
      }
    ]
  }
}
```

luisherranz:
I'd love to share another point here, just in case it matters during the implementation:
This `content.amp` field is not only useful for the Shadow AMP API, but also for **Headless WP frameworks used to render full AMP pages**, like [Next.js 8+](https://nextjs.org/docs#amp-support) or Frontity's ([upcoming support](https://community.frontity.org/t/amp-support/388)).

luisherranz:
I've started a PR to work on this.

swissspidy:
[...]

PR #1023, opened 2018-03-15
Move enqueuing of assets to 'wp_enqueue_scripts'

kienstra:
**Request For Code Review**

Hi @westonruter,
@kopepasah found an issue where the `amp-runtime` script appeared in the `wp-admin/edit.php` page.

As you mentioned, this should be enqueued on the action `wp_enqeue_scripts`.

So this does that, and wraps the enqueuing in `if ( is_amp_endpoint() )` This also appeared where they were enqueued before.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -171,6 +171,10 @@ function post_supports_amp( $post ) {
  * @return bool Whether it is the AMP endpoint.
  */
 function is_amp_endpoint() {
+	if ( is_admin() || is_feed() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) ) {
+		return false;
+	}
+
 	if ( amp_is_canonical() ) {
 		return true;
 	}
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -100,6 +100,8 @@ class AMP_Theme_Support {
 		 * action to template_redirect--the wp action--is used instead.
 		 */
 		add_action( 'wp', array( __CLASS__, 'finish_init' ), PHP_INT_MAX );
+
+		add_action( 'wp_enqueue_scripts', array( __CLASS__, 'enqueue_assets' ) );
 	}
 
 	/**
@@ -118,12 +120,6 @@ class AMP_Theme_Support {
 		} else {
 			self::register_paired_hooks();
 		}
-
-		// Enqueue AMP runtime.
-		wp_enqueue_script( 'amp-runtime' );
-
-		// Enqueue default styles expected by sanitizer.
-		wp_enqueue_style( 'amp-default', amp_get_asset_url( 'css/amp-default.css' ), array(), AMP__VERSION );
 
 		self::add_hooks();
 		self::$sanitizer_classes = amp_get_content_sanitizers();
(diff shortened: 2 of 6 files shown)


includes/class-amp-theme-support.php
```diff
@@ -1016,4 +1012,20 @@ public static function prepare_response( $response, $args = array() ) {
 
 		return $response;
 	}
+
+	/**
+	 * Enqueue AMP assets if this is an AMP endpoint.
+	 *
+	 * @since 0.7
```
kienstra: Maybe this should be `1.0` instead. I'm happy to change it.

westonruter: 0.7 is correct. I changed the branch to merge this PR.

kienstra:
**Working On `is_amp_endpoint()`**

Hi @westonruter,
I'm working on `is_amp_endpoint():

```diff
function is_amp_endpoint() {
+	if ( is_admin() || is_feed() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) ) {
+		return false;
+	}

	if ( amp_is_canonical() ) {
		return true;
	}
```

I need to address some failed unit tests from that change. If it's alright, I'll come back to this soon.

kienstra:
With the [commit above](https://github.com/Automattic/amp-wp/pull/1023/commits/919b84bab634b83de22e86cc71cf639eea932de1), this PR is ready review.

tests/test-class-amp-meta-box.php
```diff
@@ -67,6 +67,7 @@ public function test_enqueue_admin_assets() {
 		// Test inline script boot.
 		$this->assertTrue( false !== stripos( wp_json_encode( $script_data ), 'ampPostMetaBox.boot(' ) );
 		unset( $GLOBALS['post'] );
+		unset( $GLOBALS['current_screen'] );
```
kienstra: This is needed to counteract `set_current_screen( 'post.php' )` above. This caused `is_admin()` to be true on later tests.

includes/amp-helper-functions.php
```diff
@@ -171,6 +171,10 @@ function post_supports_amp( $post ) {
  * @return bool Whether it is the AMP endpoint.
  */
 function is_amp_endpoint() {
+	if ( is_admin() || is_feed() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) ) {
```
westonruter: We may need to revisit this with #1014 since we may want to render AMP content for a single property in the REST API response. We don't need to worry about it yet.

tests/test-class-amp-meta-box.php
westonruter: FYI: Probably better to such things in `tearDown` to ensure they are run regardless of whether the tests fail in this function. If a test fails, then the vars wouldn't get cleared.

includes/class-amp-theme-support.php
```diff
@@ -100,6 +100,8 @@ public static function init() {
 		 * action to template_redirect--the wp action--is used instead.
 		 */
 		add_action( 'wp', array( __CLASS__, 'finish_init' ), PHP_INT_MAX );
+
+		add_action( 'wp_enqueue_scripts', array( __CLASS__, 'enqueue_assets' ) );
```
westonruter: Putting this here caused a regression where the `amp-runtime` is enqueued even on non-AMP pages. PR to fix in https://github.com/Automattic/amp-wp/pull/1033

PR #1026, opened 2018-03-19
[Gutenberg] 1008: Extend core blocks to add AMP-specific functionality.

### LINKED ISSUE #1008: [Gutenberg] Extend existing core blocks for AMP specific functionality
**As a WordPress site running Gutenberg, my core/native blocks should have extended AMP-HTML markup features.**

- [ ] AC1: Extend core blocks which have validation errors so that the appropriate AMP markup is output to faithfully render the block when the page is served as AMP. This would be done via some combination of the render_callback that integrates with a new block sanitizer/preprocessor.

- [ ] AC2: Further extend core blocks to leverage AMP functionality, such as layout.

miina:
Fixes #1008.

~~Note that this PR is not a complete solution. It doesn't fully implement AMP Layout, also it doesn't include overriding `render_callback`s of dynamic blocks. This is partly implementing AMP Layout of all the core blocks (except for the front-end view of dynamic blocks which could be likely done in a similar way to #1010). The logic is not tested on all the blocks and was mainly using `paragraph` and `code` blocks as examples during development, and some others briefly, it's possible that some of the blocks will need separate logic.~~

**What does the PR do?**
- Adds new `ampLayout` attribute to all the core blocks which are converted to AMP elements leveraging the filter `blocks.registerBlockType`;
- Extends `edit` of each core block with filter `blocks.BlockEdit`. Adds `Amp Layout` select control to Inspector Controls (settings under Block tab when clicking on a block) which currently allows choosing between options that the specific element supports;
- ~~Extends `save` of each (non-dynamic) core block with filter `blocks.getSaveElement`. Wraps the saved block in `<amp-layout layout=[ampLayout value] width=1 height=1></ampLayout>` tags.~~
- Edit: adds `noloading` support for relevant blocks;
- Edit: makes `amp-carousel` optional for gallery shortcode;

**Todo**
- [x] Add AMP Layout attribute to embeds;
- [x] Add AMP Layout attribute to audio, video, image blocks, gallery;
- [x] Make amp-carousel optional for gallery;
- [x] Add `noloading` support for image, video, and other embeds;
- [x] Add tests;
- [x] Test everything;
- [x] Reflect amp layout changes in editor?


westonruter:
Oh, I wasn't aware of the [`amp-layout`](https://www.ampproject.org/docs/reference/components/amp-layout) component. In my mind I was thinking specifically of the [`layout`](https://www.ampproject.org/docs/guides/responsive/control_layout#the-layout-attribute) attribute. I think perhaps the attribute route may be better to focus on instead. In fact, some work has been done to support the `layout` attribute in https://github.com/Automattic/amp-wp/pull/937#discussion_r171668582 whereby the sanitizer is modified to convert `data-amp-layout` attributes over to `layout` attributes when rendering as AMP. I think this could potentially make it much easier to implement this since it would mean adding a new data attribute to _existing_ elements rather than adding new wrapper elements. Also it would mean the `data` attribute could be rendered in non-AMP responses and still be valid HTML5.

westonruter:
Another AMP attribute that could be good to add support for is `noloading` for image, video, and other embed blocks: https://www.ampproject.org/docs/reference/common_attributes#noloading

westonruter:
For `layout`, one thing that comes to mind is how the editor can visually indicate the difference between the various `layout` attributes. To do so it would seem that we should enqueue the `amp-runtime` script and try to let it apply the changes to the styling based on changes to the `layout` attribute. (Actually, for the edit function be easier to use the `amp-layout` element because then changes to its `layout` attribute should get picked up by the AMP runtime?) Adding AMP components to the Gutenberg editor is new territory as far as I know, so some discovery is needed here.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -0,0 +1,53 @@
...
+
+	/**
+	 * Enqueue filters for extending core blocks attributes.
+	 * Has to be loaded before registering the blocks in registerCoreBlocks.
+	 */
+	public function add_editor_filters() {
+		wp_enqueue_script(
+			'amp-editor-blocks',
+			amp_get_asset_url( 'js/amp-editor-blocks.js' ),
+			array( 'underscore', 'wp-hooks' ),
```
westonruter: I think this should include `amp-runtime`

assets/js/amp-editor-blocks.js
```diff
@@ -0,0 +1,112 @@
...
+							options: component.data.ampLayoutOptions,
+							onChange: function( ampLayout ) {
+								props.setAttributes( { ampLayout: ampLayout } );
+							}
+						} )
+					)
+				);
+			return [
+				inspectorControls,
+				el( BlockEdit, _.assign( { key: 'original' }, props ) ),
```
westonruter: The following is what I have in mind, though it is pretty broken in my testing:

```diff
diff --git a/assets/js/amp-editor-blocks.js b/assets/js/amp-editor-blocks.js
index c4d826a..c1ad182 100644
--- a/assets/js/amp-editor-blocks.js
+++ b/assets/js/amp-editor-blocks.js
@@ -82,9 +82,14 @@ var ampEditorBlocks = ( function() {
 						} )
 					)
 				);
+
+			const original = el( BlockEdit, _.assign( { key: 'original' }, props ) );
+
 			return [
 				inspectorControls,
-				el( BlockEdit, _.assign( { key: 'original' }, props ) ),
+				wp.element.createElement( 'amp-layout',
+					{ layout: attributes.ampLayout, width: 1, height: 1, children: original }
+				),
 			];
 		};
 	};
```

westonruter:
I've also raised this in the amp-wp Slack channel: 

> Something else: in regards to leveraging AMP in the authorship of Gutenberg blocks, it is clear that we need to be able to load the AMP runtime into the Gutenberg editor itself. The editor naturally is not an AMP document so this is the “dirty AMP” scenario. We need to be able to render AMP components by React. For example, there is an open PR now to explore adding an AMP `layout` control blocks. In order to be able to preview how the layout will impact the block, the layout logic applied by the AMP runtime should be getting applied in the context of the Gutenberg edit function. I recall one thing needed by the runtime to be used in “dirty mode” is to disable resource management from running. There are other things that may need to be confirmed, such as dynamically re-applying class names in response to `layout` attribute changes at runtime. We'd appreciate any input you have

https://amphtml.slack.com/archives/C4J8QHGBZ/p1521501878000295

miina:
Thanks for looking into the PR and for the comments, @westonruter, very helpful thoughts!

On `layout` vs `amp-layout` tag -- looks like in case of tags that are AMP elements or get converted to AMP elements (e.g. `img` => `amp-img`) we could just add the attribute `layout`. However, this wouldn't work in case of non-amp elements, e.g. `p` or `pre` etc. So basically `layout` attributes works on AMP components and `amp-layout` supports any HTML as children.

Are you thinking that maybe we should add `layout` support only to AMP elements and not to other HTML? One option would be to use both -- `amp-layout` in case of non-AMP elements and `layout` attribute in case of AMP elements. It's true that using `amp-layout` as a wrapper for AMP elements that already support the `layout` attribute seems like overkill.

Thoughts?

miina:
In case of AMP elemenets and `layout` attribute -- as discussed in https://github.com/Automattic/amp-wp/pull/937#discussion_r171668582 it seems that would generally be better to use `data-amp-layout` instead of just `layout`, and then it should get converted correctly in the sanitizer. Do I understand correctly that this is something that doesn't exist right now and needs adding to the sanitizers within this PR?

westonruter:
@miina:
> On `layout` vs `amp-layout` tag -- looks like in case of tags that are AMP elements or get converted to AMP elements (e.g. `img` => `amp-img`) we could just add the attribute `layout`. However, this wouldn't work in case of non-amp elements, e.g. `p` or `pre` etc. So basically `layout` attribute works on AMP components and `amp-layout` supports any HTML as children.
>
> Are you thinking that maybe we should add `layout` support only to AMP elements and not to other HTML? One option would be to use both -- `amp-layout` in case of non-AMP elements and `layout` attribute in case of AMP elements. It's true that using `amp-layout` as a wrapper for AMP elements that already support the `layout` attribute seems like overkill.

Great points. One idea is that there could be an `data-amp-layout` attribute added to the block's wrapper element regardless if it will be converted to an AMP component or a standard HTML5 element, and then during post-processing (sanitizer) it could look to see if any such attributes are present and then create wrapper `amp-layout` elements with the `layout` attribute for HTML5 elements, but convert `data-amp-layout` attributes to just `layout` on any AMP components.

In other words, given a `` this could result in `<amp-img layout="responsive">` whereas `<div data-amp-layout="responsive">` could result in `<amp-layout layout="responsive"><div>…`.

As I noted in https://github.com/Automattic/amp-wp/pull/937#issuecomment-371665126 I'm not totally familiar with the intricacies of the AMP layout system, so take my input with a grain of salt.

> it seems that would generally be better to use `data-amp-layout` instead of just layout, and then it should get converted correctly in the sanitizer. Do I understand correctly that this is something that doesn't exist right now and needs adding to the sanitizers within this PR?

This is partially implemented in #937 but it is not yet merged. You could copy the relevant code and put it into this PR. In particular, that PR is specifically looking at adding `layout` to images. You'd be looking to add it to all elements. (Note the use of `wp_kses_allowed_html` filter to ensure the data attributes are not removed.)

pbakaus:
Note on `<amp-layout>`: This component should really be treated as an additional component, not a way to wrap everything in, as the inner portion won't magically adapt to the situation. So if you create a `fill` amp-layout, the inner p won't magically fill as well.

TBH, I'd not worry about the `amp-layout`-as-wrapper scenario too much right now, and only make it available as optional Gutenblock, and use the layout attribute for all other core AMP components.

miina:
@westonruter Picked up the PR today to continue work here, was looking at `amp-image`, `amp-audio`, and `amp-video`. Some notes on that -- it looks like although many layout options are supported for these elements, it might not make sense to display these as options for the layout for some blocks, for example in case of `amp-video` it doesn't look like the height and width are set by just uploading the video, meaning that some layout options for `amp-video` wouldn't work, and would use the fallback of `fixed-height` layout with default `height`.  As an example `responsive` and `intrinsic` layout don't seem to make sense in case of `amp-video`, there are probably some others for other elements. Alternative would be adding additional fields to be able to set the height and width in the editor, too. Thoughts?

Also, started looking into adding Layout attribute to embeds and was wondering what would be a good approach -- the embed methods don't have access to the attributes saved within Gutenberg nor their parent HTML element (which has the `data-amp-layout` attribute), thus it's not so straightforward to add the layout attribute. One option could be for example filtering the content and looking for blocks that include the embed, and then adding additional param to the end of the embed URL with layout, which would be remove later. E.g. if the original URL is `http://youtube.com/video123` then somehow adding `?amp-data-layout='something'` to the ending, and then in the `AMP_YouTube_Embed_Handler::shortcode` remove that and add the layout attribute to the `amp-youtube` tag. Or perhaps it would be possible to override the Gutenberg embed block's `save` method to add the layout directly there. Or perhaps there's some other better way. Thoughts?

westonruter:
> it doesn't look like the height and width are set by just uploading the video

I'm looking at a video that I uploaded in my install and I'm seeing there is a `_wp_attachment_metadata` postmeta that contains:

```json
{"filesize":24004120,
"mime_type":"video\/mp4",
"length":9,
"length_formatted":"0:09",
"width":1920,
"height":1080,
"fileformat":"mp4",
"dataformat":"quicktime",
"audio":{"dataformat":"mp4","codec":"ISO\/IEC 14496-3 AAC","sample_rate":48000,"channels":1,"bits_per_sample":16,"lossless":false,"channelmode":"mono"},
"created_timestamp":1512335233}
```

So at least for this video there is indeed `width`/`height`. For some reason these dimensions fetched and output by the default video shortcode, but it seems it is available for us to use via `wp_get_attachment_metadata()`.

> Also, started looking into adding Layout attribute to embeds and was wondering what would be a good approach

So the problem is that a Twitter block, could look like this:

```html
<!-- wp:core-embed/twitter {"url":"https://twitter.com/westonruter/status/988577526741061632","type":"rich","providerNameSlug":"twitter"} -->
<figure class="wp-block-embed-twitter wp-block-embed is-type-rich is-provider-twitter">
    https://twitter.com/westonruter/status/988577526741061632
</figure>
<!-- /wp:core-embed/twitter -->
```

And get output as this:

```html
<figure class="wp-block-embed-twitter wp-block-embed is-type-rich is-provider-twitter">
    <amp-twitter data-tweetid="988577526741061632" layout="responsive" width="600" height="480"></amp-twitter>
</figure>
```

And if `data-amp-layout` were added to `figure.wp-block-embed` that the markup generated by the embed handler wouldn't be able to see it, right?

Here's an idea: what if you go ahead and add a `data-amp-layout` attribute to the `figure` element, and then in the sanitizer that is already looking for `data-amp-layout` attributes, look for it being on any elements that contain a single element that itself has a `layout` attribute, and in that case cause the outer layout to override the inner layout? In other words, if wanting to have a `fixed` layout for a Tweet with markup stored like:

```html
<figure data-amp-layout="fixed" class="wp-block-embed-twitter wp-block-embed is-type-rich is-provider-twitter">
    <amp-twitter data-tweetid="988577526741061632" layout="responsive" width="600" height="480"></amp-twitter>
</figure>
```

This could get post-processed by a sanitizer into:

```html
<figure class="wp-block-embed-twitter wp-block-embed is-type-rich is-provider-twitter">
    <amp-twitter data-tweetid="988577526741061632" layout="fixed" width="600" height="480"></amp-twitter>
</figure>
```

The question then in my mind is whether that “inheritance override” behavior should be limited to `figure.wp-block-embed` elements, or if it should apply to any `[data-amp-layout] > [layout]` elements. Maybe the override behavior should be made explicit by calling it a `data-amp-layout-override` attribute instead of just `data-amp-layout`, or maybe that is unnecessary.

includes/sanitizers/class-amp-block-sanitizer.php
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * @package AMP
+ */
+
+/**
+ * Class AMP_Block_Sanitizer
+ *
+ * Modifies elements created as blocks to match the blocks' AMP-specific configuration.
+ */
+class AMP_Block_Sanitizer extends AMP_Base_Sanitizer {
```
miina: @westonruter That's still WIP but if you happen to have time, could you check if this approach for embeds seems to make sense?

includes/embeds/class-amp-gallery-embed.php
```diff
@@ -128,7 +128,10 @@ public function shortcode( $attr ) {
 	 * @param array  $attributes Shortcode attributes.
 	 * @return string $html Markup for the gallery.
 	 */
-	public function override_gallery( $html, $attributes ) {
+	public function maybe_override_gallery( $html, $attributes ) {
+		if ( isset( $attributes['amp-carousel'] ) && false === filter_var( $attributes['amp-carousel'], FILTER_VALIDATE_BOOLEAN ) ) {
+			return $html;
```
miina: @westonruter Added editor support for making displaying gallery shortcode as `amp-carousel` optional by adding a shortcode attribute `amp-carousel`. The attribute is hidden from the user in editor view, and the Inspector Control is displayed only if `[gallery` is found in the shortcode text, seems to be working as expected.
screen shot 2018-04-26 at 4 30 05 pm

Do you think this approach makes sense? Thoughts?

miina:
Added support for `noloading` within 7bf9df3.

includes/sanitizers/class-amp-block-sanitizer.php
```diff
@@ -0,0 +1,131 @@
...
+	protected function set_attributes( $node, $parent_node, $layout, $attributes ) {
+
+		// The width has to be unset / auto in case of fixed-height.
+		if ( 'fixed-height' === $layout ) {
+			if ( ! isset( $attributes['height'] ) ) {
+				$node->setAttribute( 'height', self::FALLBACK_HEIGHT );
+			}
+			$node->setAttribute( 'width', 'auto' );
+
+			// @todo Perhaps the height is not in px, is it possible?
```
westonruter: According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-height

> The intrinsic height of the image in pixels. In HTML 4, the height could be defined pixels or as a percentage. In HTML5, however, the value must be in pixels.

So I'd say we could assume pixel value. But just in case, you could could do something like this:

```php
$height = $node->getAttribute( 'height' );
if ( is_numeric( $height ) ) {
    $height .= 'px';
}
$parent_node->setAttribute( 'style', "height: $height; width: auto;" );
```

includes/sanitizers/class-amp-block-sanitizer.php
```diff
@@ -0,0 +1,131 @@
...
+	public static $amp_embeds = array(
+		'amp-youtube',
+		'amp-facebook',
+		'amp-twitter',
+		'amp-vimeo',
+		'amp-instagram',
+		'amp-dailymotion',
+		'amp-hulu',
+		'amp-reddit',
+		'amp-soundcloud',
```
westonruter: Instead of enumerating out all of the embeds, what about this: look for `figure` elements that contain the `wp-block-embed` class, and then if this `figure` element contains a `DOMElement` whose name begins with `amp-` then assume it is an AMP embed component. This will help prevent us having to maintain this list here in response to Gutenberg changes and spec changes.

assets/js/amp-editor-blocks.js
```diff
@@ -0,0 +1,374 @@
...
+				name = props.name,
+				ampLayout,
+				inspectorControls,
+				width;
+
+			ampLayout = attributes.ampLayout;
+
+			if ( 'core/shortcode' === name ) {
+				// Lets remove amp-carousel from from edit view.
+				if ( component.hasGalleryShortcodeCarouselAttribute( attributes.text ) ) {
```
westonruter: I'm getting a JS error here when first adding a shortcode because `text` prop is undefined. So I think the value passed into this function should be `attributes.text || ''`.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -0,0 +1,68 @@
...
+ * Class AMP_Editor_Blocks
+ */
+class AMP_Editor_Blocks {
+
+	/**
+	 * Init.
+	 */
+	public static function init() {
+		if ( function_exists( 'gutenberg_init' ) ) {
+			$self = new self();
```
westonruter: I don't think this is right. Below you do `$editor_blocks = new AMP_Editor_Blocks()` so should not it just be using `$this` here instead of creating a second instance of itself?

includes/embeds/class-amp-gallery-embed.php
westonruter: Yes, I think that makes sense and it seems to work great.

At first I was thinking that carousel option should also be made available to the Gallery block in addition to the `gallery` shortcode, but maybe that's something that will be coming in Gutenberg as a separate block type in its own right.

includes/sanitizers/class-amp-block-sanitizer.php
```diff
@@ -0,0 +1,136 @@
...
+				if ( ! in_array( $child_node->tagName, self::$amp_embeds, true ) ) {
+					continue;
+				}
+				$amp_el_found = true;
+
+				if ( isset( $attributes['data-amp-layout'] ) ) {
+					$child_node->setAttribute( 'layout', $attributes['data-amp-layout'] );
+				}
+				if ( isset( $attributes['data-amp-noloading'] ) && true === filter_var( $attributes['data-amp-noloading'], FILTER_VALIDATE_BOOLEAN ) ) {
+					$child_node->setAttribute( 'noloading', '' );
```
westonruter: Instead of setting the attributes here, why not set them in the `set_layout_attributes` method?

includes/sanitizers/class-amp-block-sanitizer.php
westonruter: I think you're on the right track!

kienstra:
**Layout And Noloading Attributes*

Hi @miina,
Following up on your request for more functional testing here, the `layout` and `noloading` settings have worked for all of the blocks that I tested:
* Image (the noloading attribute also appears when added in the UI)
* YouTube Embed
* Audio

layout-select-element

miina:
@kienstra Thank you for testing. Guess the PR should be ready for review.
(cc @westonruter)

westonruter:
@miina I tried adding `intrinsic` layout to an image. Then I tried switching to `fixed-height` and upon doing so I got an error in my console:

> Warning: Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.

Additionally, the window froze, apparently due to an infinite loop until finally an error was thrown:

> Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

It seems to be _specifically_ a problem when setting the layout to `fixed-height`. I don't get this error when switching to another layout.

westonruter:
@miina also, as we chatted about today, if I add a `responsive` layout for example and save the post. And then I switch the AMP plugin branch back to `develop` and reload, I then get:

image

So I think we need to save out the `layout` in a `class` instead of `data-amp-layout`, but this layout-specific class shouldn't appear in the “Additional CSS Class” field when the AMP plugin is active. I think this is similar to what you did for the `gallery` shortcode.

miina:
@westonruter:
Fix for `fixed-height` issue added (separate commit b107296). Also reworked the code to use CSS classes instead of `data-amp-*`. Note that had to use REST API filters for saving the class to post_content without showing it in editor, let me know if this approach seems reasonable.

One additional note, this appears actually on `develop` branch, too. Getting this error when using an image block in editor:
screen shot 2018-05-04 at 5 31 15 pm
The `responseText` in the error is this:
screen shot 2018-05-04 at 5 34 09 pm
Looks like in case of `attachment` the `content` is not set and thus the error.

Can you also see this error or is it just me?

includes/embeds/class-amp-gallery-embed.php
miina: Do you think that just using `amp-carousel` attribute within post content when using Classic Editor for gallery shortcode could be a fix for #1065, too?

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -342,4 +345,108 @@ public function remove_invalid_attribute( $element, $attribute, $args = array()
...
+	}
+
+	/**
+	 * Set AMP attributes.
+	 *
+	 * @param array $attributes Array of attributes.
+	 * @param array $amp_data Array of AMP attributes.
+	 * @return array
+	 */
+	public function set_data_amp_attributes( $attributes, $amp_data ) {
```
westonruter: I think instead of `set_` this should be prefixed with `filter_`, since it's not actually setting any attributes on a node. That would align with `filter_attributes` method in the image sanitizer.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -342,4 +345,108 @@ public function remove_invalid_attribute( $element, $attribute, $args = array()
...
+
+	/**
+	 * Set attributes to node's parent element according to layout.
+	 *
+	 * @param DOMNode $node Node.
+	 * @param array   $new_attributes Attributes array.
+	 * @param string  $layout Layout.
+	 * @return array New attributes.
+	 */
+	public function set_attachment_layout_attributes( $node, $new_attributes, $layout ) {
```
westonruter: Same as above:

> I think instead of `set_` this should be prefixed with `filter_`, since it's not actually setting any attributes on a node. That would align with `filter_attributes` method in the image sanitizer.

But that gets tricky here because here there are side-effects on the parent node. So it's not so clear what is the best to do.

assets/js/amp-editor-blocks.js
```diff
@@ -0,0 +1,416 @@
...
+						}, props ) )
+					];
+				}
+			} else if ( -1 !== component.data.mediaBlocks.indexOf( name ) || 0 === name.indexOf( 'core-embed/' ) ) {
+				inspectorControls = component.setUpInspectorControls( props );
+			}
+
+			if ( attributes.ampLayout ) {
+				if ( 'core/image' === name ) {
+					component.setImageBlockLayoutAttributes( props, attributes.ampLayout, inspectorControls );
```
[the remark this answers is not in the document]
westonruter: I think this should be done in some `onChange` callback somehow.

assets/js/amp-editor-blocks.js
```diff
@@ -0,0 +1,416 @@
...
+			],
+			embedBlocks = [
+				'core-embed/youtube',
+				'core-embed/facebook',
+				'core-embed/instagram'
+			];
+
+		_.each( component.data.ampLayoutOptions, function( option ) {
+			// Exclude options from layout that are not supported.
+			if ( 'core/image' === blockName || 'core/video' === blockName || 'core-embed/twitter' === blockName ) {
```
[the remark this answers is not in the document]
miina: Changed within dbc1c7c, used `notAvailable` param for `ampLayoutOptions`, let me know if `isAvailable` param seems to make more sense.

assets/js/amp-editor-blocks.js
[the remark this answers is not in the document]
miina: Changed within dbc1c7c.

includes/embeds/class-amp-gallery-embed.php
miina: Yes, it would, just thinking if this should be the official fix that closes #1065, too, or if an alternative approach would be needed for that.

includes/sanitizers/class-amp-base-sanitizer.php
miina: Changed this to `filter_` within dbc1c7c since even with setting the parent node attributes it seems to be more consistent. One option could be to separate setting the parent node attributes from the filtering method into a separate method, would that be a more clear / better alternative? Thoughts?

includes/admin/class-amp-editor-blocks.php
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	/**
+	 * Whitelist used data-amp-* attributes.
+	 *
+	 * @param array $context Array of contexts.
+	 * @return mixed Modified array.
+	 */
+	public function whitelist_layout_in_wp_kses_allowed_html( $context ) {
+		foreach ( $context as $tag ) {
```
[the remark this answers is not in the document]
miina: That was not working, indeed. Changed within dbc1c7c.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -342,4 +345,108 @@ public function remove_invalid_attribute( $element, $attribute, $args = array()
...
+
+		// If either height or width is missing, try to get these from original file.
+		if ( empty( $new_attributes['width'] ) || empty( $new_attributes['height'] ) ) {
```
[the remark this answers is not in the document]
miina: Moved within dbc1c7c, missed the fetching image dimensions logic within `AMP_Img_Sanitizer` when creating this, thanks for noting that!

miina:
@westonruter On editor freeze with Gallery block -- just to confirm, do I understand correctly that the editor freeze is with inserting the default Gallery block and not the gallery shortcode into Shortcode block? Asking since couldn't reproduce the issue at this moment (using Gutenberg 2.9.1 on WP 4.9.6 on VVV), in theory this PR shouldn't influence the Gallery block. Does the freeze occur when clicking the block or already when inserting the block without any additional action?

assets/js/amp-editor-blocks.js
```diff
@@ -0,0 +1,448 @@
...
+							props.setAttributes( { ampLayout: value } );
+							if ( 'core/image' === props.name ) {
+								component.setImageBlockLayoutAttributes( props, value );
+							}
+						}
+					} ),
+					el( ToggleControl, {
+						label: __( 'AMP loading indicator disabled' ),
+						checked: ampNoLoading,
+						onChange: function() {
```
westonruter: For some reason, the first time I check the checkbox I get a warning:

> Warning: A component is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components

Maybe it's a bug in Gutenberg. It doesn't seem to cause a problem.

westonruter:
I just pulled down the latest from Gutenberg and I'm not able to reproduce the editor freezing when adding a Gallery block.

PR #1029, opened 2018-03-20
Do redirect if amp comments_live_list support is not declared; vary comment success message by approval status

### LINKED ISSUE #936: Change comment confirmation message based on approval settings

### LINKED ISSUE #1028: Let amp-live-list for comments be opt-in
We should add a theme support flag for `comments_live_list` to control whether or not comment submission is intercepted. Without flag, default full-page-reload behavior should be retained to allow themes without `amp-live-list` in comments to work normally.

The theme can do:

```php
	'comments_live_list' => true,
) );
```

And then the plugin can read it like this:

```php
<?php
add_filter( 'comment_post_redirect', function( $url, $comment ) {
	$theme_support = get_theme_support( 'amp' );

	// Return with simply success when amp-live-list will show newly-posted comment.
	if ( ! empty( $theme_support['comments_live_list'] ) ) {
		wp_send_json_success();
	}

	// Add the comment ID to the URL to force AMP to refresh the page.
	$url = add_query_arg( 'comment', $comment->comment_ID, $url );

	return $url;
}, PHP_INT_MAX, 2 );
```

This will allow `amp` support to be added to an existing theme without requiring its comments template to be retrofitted with `amp-live-list`.

Alternatively, if we think that `amp-live-list` should _really_ be the default we want to always be used, then there could instead be an opt-out flag.

DavidCramer:
fixes #1028 
fixes #936

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -369,8 +369,10 @@ class AMP_Theme_Support {
 		}
 
 		// Send AMP response header.
-		$origin = esc_url_raw( self::$purged_amp_query_vars['__amp_source_origin'] );
-		self::send_header( 'AMP-Access-Control-Allow-Source-Origin', $origin, array( 'replace' => true ) ); // @todo The $origin needs to be validated.
+		$origin = wp_validate_redirect( wp_sanitize_redirect( esc_url_raw( self::$purged_amp_query_vars['__amp_source_origin'] ) ) );
+		if ( $origin ) {
+			self::send_header( 'AMP-Access-Control-Allow-Source-Origin', $origin, array( 'replace' => true ) ); // @todo The $origin needs to be validated.
+		}
 
 		// Intercept POST requests which redirect.
 		add_filter( 'wp_redirect', array( __CLASS__, 'intercept_post_request_redirect' ), PHP_INT_MAX );
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -335,9 +335,17 @@ class AMP_Theme_Support {
 				if ( empty( $theme_support['comments_live_list'] ) ) {
 					header( 'AMP-Redirect-To: ' . $url, true );
 				}
+				// Create a success message to display to the user.
+				if ( '1' === (string) $comment->comment_approved ) {
+					$message = __( 'Your comment has been posted and has been approved.', 'amp' );
+				} else {
+					$message = __( 'Your comment has been posted, and is awaiting moderation.', 'default' );
+				}
+
+				$message = apply_filters( 'amp_comment_submitted_message', $message, $comment );
 
 				// We don't need any data, so just send a success.
-				wp_send_json_success();
+				wp_send_json_success( $message );
 
 			}, PHP_INT_MAX, 2 );
 			self::handle_xhr_headers_output();
@@ -528,7 +536,7 @@ class AMP_Theme_Support {
 		?>
 		<div submit-success>
 			<template type="amp-mustache">
-				<?php esc_html_e( 'Your comment has been posted, but may be subject to moderation.', 'amp' ); ?>
+				{{{data}}}
 			</template>
 		</div>
 		<div submit-error>
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -501,7 +501,11 @@ class AMP_Theme_Support {
 			wp_parse_url( $location )
 		);
 
-		$absolute_location = $parsed_location['scheme'] . '://' . $parsed_location['host'];
+		$absolute_location = '';
+		if ( 'https' === $parsed_location['scheme'] ) {
+			$absolute_location .= $parsed_location['scheme'] . ':';
+		}
+		$absolute_location .= '//' . $parsed_location['host'];
 		if ( isset( $parsed_location['port'] ) ) {
 			$absolute_location .= ':' . $parsed_location['port'];
 		}
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -359,7 +359,7 @@ class AMP_Theme_Support {
 	}
 
 	/**
-	 * Hook into a form submissions, such as the comment form or some other form submission.
+	 * Hook into a POST form submissions, such as the comment form or some other form submission.
 	 *
 	 * @since 0.7.0
 	 */
@@ -371,7 +371,7 @@ class AMP_Theme_Support {
 		// Send AMP response header.
 		$origin = wp_validate_redirect( wp_sanitize_redirect( esc_url_raw( self::$purged_amp_query_vars['__amp_source_origin'] ) ) );
 		if ( $origin ) {
-			self::send_header( 'AMP-Access-Control-Allow-Source-Origin', $origin, array( 'replace' => true ) ); // @todo The $origin needs to be validated.
+			self::send_header( 'AMP-Access-Control-Allow-Source-Origin', $origin, array( 'replace' => true ) );
 		}
 
 		// Intercept POST requests which redirect.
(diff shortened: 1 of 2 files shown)


DavidCramer:
@westonruter you can take a look at this now. I've included #936 as well, since they are closely related.

DavidCramer:
@westonruter - I was basing it on your second thought - https://github.com/Automattic/amp-wp/pull/1020#issuecomment-373889605

DavidCramer:
@westonruter - I see I missed the filter, sorry. ignore me, I'll revised it.

includes/class-amp-theme-support.php
```diff
@@ -324,10 +324,30 @@ public static function handle_xhr_request() {
...
-			add_filter( 'comment_post_redirect', function() {
+			add_filter( 'comment_post_redirect', function( $url, $comment ) {
+				// Get theme support.
+				$theme_support = get_theme_support( 'amp' );
+
+				// Add the comment ID to the URL to force AMP to refresh the page.
+				$url = add_query_arg( 'comment', $comment->comment_ID, $url );
+
+				// Send redirect header if amp-live-list has been opted-out.
+				if ( empty( $theme_support['comments_live_list'] ) ) {
```
westonruter: This doesn't work because `$theme_support` is an array of the args so it needs to be looking at ` $theme_support[0]['comments_live_list']`

includes/class-amp-theme-support.php
```diff
@@ -324,10 +324,30 @@ public static function handle_xhr_request() {
...
+					header( 'AMP-Redirect-To: ' . $url, true );
```
westonruter: This shouldn't be necessary because `wp_redirect()` is currently being called. It only needs to return the `$url`

includes/class-amp-theme-support.php
```diff
@@ -324,10 +324,30 @@ public static function handle_xhr_request() {
...
+				}
+				// Create a success message to display to the user.
+				if ( '1' === (string) $comment->comment_approved ) {
+					$message = __( 'Your comment has been posted and has been approved.', 'amp' );
+				} else {
+					$message = __( 'Your comment has been posted, and is awaiting moderation.', 'default' );
+				}
+
```
westonruter: Missing phpdoc.

includes/class-amp-theme-support.php
```diff
@@ -324,10 +324,30 @@ public static function handle_xhr_request() {
...
+				// Send redirect header if amp-live-list has been opted-out.
+				if ( empty( $theme_support['comments_live_list'] ) ) {
+					header( 'AMP-Redirect-To: ' . $url, true );
+				}
+				// Create a success message to display to the user.
+				if ( '1' === (string) $comment->comment_approved ) {
+					$message = __( 'Your comment has been posted and has been approved.', 'amp' );
+				} else {
+					$message = __( 'Your comment has been posted, and is awaiting moderation.', 'default' );
```
westonruter: This string has the `default` text domain because it is re-using a string in core:

https://github.com/WordPress/wordpress-develop/blob/8f95800d52c1736d651ae6e259f90ad4a0db2c3f/src/wp-includes/class-walker-comment.php#L285

We need it to keep it the same for it to re-use core translations.

includes/class-amp-theme-support.php
```diff
@@ -516,7 +536,7 @@ public static function add_amp_comment_form_templates() {
 		?>
 		<div submit-success>
 			<template type="amp-mustache">
-				<?php esc_html_e( 'Your comment has been posted, but may be subject to moderation.', 'amp' ); ?>
+				{{{data}}}
```
westonruter: This should have a paragraph wrapping it, like for the error message.

westonruter: The error handler sends back an object with an `error` key. I think this should be similarly looking for an object returned with just a `message` key.

includes/class-amp-theme-support.php
```diff
@@ -324,10 +324,30 @@ public static function handle_xhr_request() {
...
+				} else {
+					$message = __( 'Your comment has been posted, and is awaiting moderation.', 'default' );
+				}
+
+				$message = apply_filters( 'amp_comment_submitted_message', $message, $comment );
+
 				// We don't need any data, so just send a success.
-				wp_send_json_success();
-			}, PHP_INT_MAX );
+				wp_send_json_success( $message );
```
westonruter: We should probably apply kses to this to only allow the allowed amp-mustache markup.

includes/class-amp-theme-support.php
```diff
@@ -324,10 +324,30 @@ public static function handle_xhr_request() {
...
+				// Create a success message to display to the user.
+				if ( '1' === (string) $comment->comment_approved ) {
+					$message = __( 'Your comment has been posted and has been approved.', 'amp' );
+				} else {
+					$message = __( 'Your comment has been posted, and is awaiting moderation.', 'default' );
+				}
+
+				$message = apply_filters( 'amp_comment_submitted_message', $message, $comment );
+
 				// We don't need any data, so just send a success.
```
westonruter: This comment is no longer true.

westonruter:
@DavidCramer Please test these changes with the corresponding changes in the theme PR https://github.com/xwp/ampconf/pull/85

Test submitting comments with the `comments_live_list` flag enabled and disabled. Also test Jetpack comment form to ensure it continues to work as expected. It is working for me.

DavidCramer:
@westonruter Thanks for those changes. I've gone over them to get better understanding of it all. Will review and test xwp/ampconf#85

includes/class-amp-theme-support.php
```diff
@@ -310,65 +319,174 @@ public static function purge_amp_query_vars() {
...
+	 *                                code. Default empty.
+	 * @param string|array|int $args {
+	 *     Optional. Arguments to control behavior. If `$args` is an integer, then it is treated
+	 *     as the response code. Default empty array.
+	 *
+	 *     @type int $response The HTTP response code. Default 200 for Ajax requests, 500 otherwise.
+	 * }
+	 */
+	public static function handle_wp_die( $error, $title = '', $args = array() ) {
+		$status_code = 500;
```
ThierryA: That could move in a else statement at the end of the statements below.

includes/class-amp-theme-support.php
```diff
@@ -310,65 +319,174 @@ public static function purge_amp_query_vars() {
...
 		}
 
-		if ( isset( $pagenow ) && 'wp-comments-post.php' === $pagenow ) {
-			// We don't need any data, so just send a success.
-			add_filter( 'comment_post_redirect', function() {
-				// We don't need any data, so just send a success.
-				wp_send_json_success();
-			}, PHP_INT_MAX );
-			self::handle_xhr_headers_output();
-		} elseif ( ! empty( self::$purged_amp_query_vars['_wp_amp_action_xhr_converted'] ) ) {
```
westonruter: I realized that we're now not properly looking at `_wp_amp_action_xhr_converted`.

includes/class-amp-theme-support.php
```diff
@@ -310,65 +319,174 @@ public static function purge_amp_query_vars() {
...
+	 * @return string URL.
+	 */
+	public static function filter_comment_post_redirect( $url, $comment ) {
+		$theme_support = get_theme_support( 'amp' );
+
+		// Cause a page refresh if amp-live-list is not implemented for comments via add_theme_support( 'amp', array( 'comments_live_list' => true ) ).
+		if ( empty( $theme_support[0]['comments_live_list'] ) ) {
+
+			// Add the comment ID to the URL to force AMP to refresh the page.
+			$url = add_query_arg( 'comment', $comment->comment_ID, $url );
```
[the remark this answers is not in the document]
westonruter: Good point, but the idea is that eventually this should not be necessary at all. I've opened an issue in amphtml to address this and I'll add a comment here to reference it: https://github.com/ampproject/amphtml/issues/14170

includes/class-amp-theme-support.php
westonruter: On further thought, I don't think the logic this function cares if `_wp_amp_action_xhr_converted`. If someone is implementing their own `action-xhr` endpoint, then they're not going to be doing `wp_redirect()` anyway, but rather should be responding with their own appropriate `wp_send_json()` call.

PR #1040, opened 2018-03-23
Check if  argument is empty and initialize it properly

amedina:
Check for the edge case when the $analytics argument passed to amp_print_analytics() is empty and initialize it properly. This happened because the function is hooked into `wp_footer` (````add_action( 'wp_footer', 'amp_print_analytics' );```` and when `do_action( 'wp_footer' )` is called, it passes an empty string as the first arg to amp_print_analytics.

### DIFF
--- a/tests/test-amp-analytics-options.php
+++ b/tests/test-amp-analytics-options.php
@@ -257,7 +257,34 @@ class AMP_Analytics_Options_Test extends WP_UnitTestCase {
 		amp_print_analytics( $analytics );
 		$output = ob_get_clean();
 
-		$this->assertEquals( 0, strpos( $output, '<amp-analytics' ) );
+		$this->assertStringStartsWith( '<amp-analytics', $output );
+		$this->assertContains( 'type="googleanalytics"><script type="application/json">{"requests":{"event":', $output );
+	}
+
+	/**
+	 * Test amp_print_analytics() when empty, called via wp_footer.
+	 *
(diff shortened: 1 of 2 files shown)


tests/test-amp-analytics-options.php
```diff
@@ -257,7 +257,34 @@ public function test_amp_print_analytics() {
...
+	/**
+	 * Test amp_print_analytics() when empty, called via wp_footer.
+	 *
+	 * Note that wp_footer action passes empty string to any handlers.
+	 * This test asserts that an issue discovered in PHP 7.1 is fixed.
+	 *
+	 * @see AMP_Theme_Support::add_hooks() Where add_action( 'wp_footer', 'amp_print_analytics' ) is done.
+	 * @covers \amp_print_analytics()
+	 */
+	public function test_amp_print_analytics_when_empty() {
```
westonruter: Without the change applied to `amp_print_analytics` this test fails with the following in PHP 7.1:

```
There was 1 error:
1) AMP_Analytics_Options_Test::test_amp_print_analytics_when_empty
Illegal string offset 'adc84c40ea28'
/tmp/wordpress/src/wp-content/plugins/amp/includes/amp-helper-functions.php:361
/tmp/wordpress/src/wp-content/plugins/amp/includes/amp-helper-functions.php:379
/tmp/wordpress/src/wp-content/plugins/amp/tests/test-amp-analytics-options.php:280
```

PR #1041, opened 2018-03-23
Fix the order in which ensure_required_markup() is called

amedina:
The order in which `ensure_required_markup` was being called was causing the viewport meta tag to be stripped out by the sanitizer.

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -887,11 +887,6 @@ class AMP_Theme_Support {
 	public static function ensure_required_markup( DOMDocument $dom ) {
 		$xpath = new DOMXPath( $dom );
 
-		// First ensure the mandatory amp attribute is present on the html element, as otherwise it will be stripped entirely.
-		if ( ! $dom->documentElement->hasAttribute( 'amp' ) && ! $dom->documentElement->hasAttribute( '⚡️' ) ) {
-			$dom->documentElement->setAttribute( 'amp', '' );
-		}
-
 		$head = $dom->getElementsByTagName( 'head' )->item( 0 );
 		if ( ! $head ) {
 			$head = $dom->createElement( 'head' );
@@ -952,15 +947,6 @@ class AMP_Theme_Support {
 				'href' => self::get_current_canonical_url(),
 			) );
 			$head->appendChild( $rel_canonical );
-		}
-
-		// Make sure scripts from the body get moved to the head.
-		$scripts = array();
-		foreach ( $xpath->query( '//body//script[ @custom-element or @custom-template ]' ) as $script ) {
-			$scripts[] = $script;
-		}
-		foreach ( $scripts as $script ) {
-			$head->appendChild( $script );
 		}
 	}
 
@@ -1098,9 +1084,25 @@ class AMP_Theme_Support {
 		}
 		$dom = AMP_DOM_Utils::get_dom( $response );
 
-		self::ensure_required_markup( $dom );
+		$xpath = new DOMXPath( $dom );
+
+		// Make sure scripts from the body get moved to the head.
+		$scripts = array();
+		foreach ( $xpath->query( '//body//script[ @custom-element or @custom-template ]' ) as $script ) {
+			$scripts[] = $script;
+		}
+		foreach ( $scripts as $script ) {
+			$head->appendChild( $script );
+		}
+
+		// First ensure the mandatory amp attribute is present on the html element, as otherwise it will be stripped entirely.
+		if ( ! $dom->documentElement->hasAttribute( 'amp' ) && ! $dom->documentElement->hasAttribute( '⚡️' ) ) {
+			$dom->documentElement->setAttribute( 'amp', '' );
+		}
 
 		$assets = AMP_Content_Sanitizer::sanitize_document( $dom, self::$sanitizer_classes, $args );
+
+		self::ensure_required_markup( $dom );
 
 		// @todo If 'utf-8' is not the blog charset, then we'll need to do some character encoding conversation or "entityification".
 		if ( 'utf-8' !== strtolower( get_bloginfo( 'charset' ) ) ) {
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -885,8 +885,6 @@ class AMP_Theme_Support {
 	 * @param DOMDocument $dom Doc.
 	 */
 	public static function ensure_required_markup( DOMDocument $dom ) {
-		$xpath = new DOMXPath( $dom );
-
 		$head = $dom->getElementsByTagName( 'head' )->item( 0 );
 		if ( ! $head ) {
 			$head = $dom->createElement( 'head' );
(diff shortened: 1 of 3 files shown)


westonruter:
This is a regression introduced by me in https://github.com/Automattic/amp-wp/commit/0154620e624db02944920681999e34b7149d7abe

includes/class-amp-theme-support.php
```diff
@@ -1098,10 +1084,26 @@ public static function prepare_response( $response, $args = array() ) {
...
-		self::ensure_required_markup( $dom );
+		$xpath = new DOMXPath( $dom );
+
+		// Make sure scripts from the body get moved to the head.
+		$scripts = array();
+		foreach ( $xpath->query( '//body//script[ @custom-element or @custom-template ]' ) as $script ) {
+			$scripts[] = $script;
+		}
+		foreach ( $scripts as $script ) {
+			$head->appendChild( $script );
```
westonruter: The `$head` is undefined. We should grab it with `DOMXPath` and only do the script gathering if it is defined.

westonruter:
It would be good to add a new test case for the bug this is fixed so that it doesn't re-surface.  For example another test like `\Test_AMP_Theme_Support::test_prepare_response()` could be made which specifically tests the validation of a non-AMP theme like Twenty Fifteen. Take some of the actual output of Twenty Fifteen (in particular the AMP-invalid meta viewport) and running it through the sanitizer to ensure the expected viewport comes out the other end. Similarly, inside the body there could be a call to `<?php wp_print_scripts( 'amp-mathml' ); ?>` and then test to make sure that the `script` tag gets moved to the head properly.

tests/test-class-amp-theme-support.php
```diff
@@ -212,6 +212,54 @@ public function test_prepare_response() {
...
+			</head>
+			<body>
+				<?php wp_print_scripts( 'amp-mathml' ); ?>
+			</body>
+		</html>
+		<?php
+		$original_html  = trim( ob_get_clean() );
+		$removed_nodes  = array();
+		$sanitized_html = AMP_Theme_Support::prepare_response( $original_html, array(
+			'validation_error_callback' => function( $removed ) use ( &$removed_nodes ) {
```
westonruter: I don't think we need `validation_error_callback` here since `$removed_nodes` is never checked. Alternative, you could `$this->assertEmpty( $removed_nodes )` to verify that nothing got stripped out.

tests/test-class-amp-theme-support.php
```diff
@@ -212,6 +212,54 @@ public function test_prepare_response() {
...
+				$removed_nodes[ $removed['node']->nodeName ] = $removed['node'];
+			},
+		) );
+
+		$this->assertContains( '<meta charset="' . get_bloginfo( 'charset' ) . '">', $sanitized_html );
```
westonruter: Since AMP must be UTF-8, should this use not be a literal UTF-8?

tests/test-class-amp-theme-support.php
```diff
@@ -212,6 +212,54 @@ public function test_prepare_response() {
 		$this->assertInstanceOf( 'DOMAttr', $removed_nodes['onclick'] );
 	}
 
+	/**
+	 * Test validate_non_amp_theme.
+	 *
+	 * @global WP_Widget_Factory $wp_widget_factory
+	 * @global WP_Scripts $wp_scripts
+	 * @covers AMP_Theme_Support::validate_non_amp_theme()
```
westonruter: There is no such method as `AMP_Theme_Support::validate_non_amp_theme()`. It should be `AMP_Theme_Support:: prepare_response()`.

tests/test-class-amp-theme-support.php
```diff
@@ -212,6 +212,54 @@ public function test_prepare_response() {
...
+	 */
+	public function test_validate_non_amp_theme() {
+		global $wp_widget_factory, $wp_scripts;
+		$wp_scripts = null;
+
+		add_theme_support( 'amp' );
+		AMP_Theme_Support::init();
+		AMP_Theme_Support::finish_init();
+		$wp_widget_factory = new WP_Widget_Factory();
+		wp_widgets_init();
```
westonruter: Since no widgets are being tested in this response, I think the references to `$wp_widget_factory` and `wp_widgets_init()` in this PR can be removed.

tests/test-class-amp-theme-support.php
amedina: Agree; changed to `$this->assertContains( '<meta charset="utf-8">', $sanitized_html );`

tests/test-class-amp-theme-support.php
```diff
@@ -212,6 +212,49 @@ public function test_prepare_response() {
...
 
+	/**
+	 * Test validate_non_amp_theme.
+	 *
+	 * @global WP_Widget_Factory $wp_widget_factory
+	 * @global WP_Scripts $wp_scripts
+	 * @covers AMP_Theme_Support::prepare_response()
+	 */
+	public function test_validate_non_amp_theme() {
+		global $wp_widget_factory, $wp_scripts;
```
ThierryA: `$wp_widget_factory` doesn't seem to be used in this method and the `$wp_scripts` scripts are already reset in the `tearDown()` method.

tests/test-class-amp-theme-support.php
```diff
@@ -212,6 +212,49 @@ public function test_prepare_response() {
...
+				<meta name="viewport" content="width=device-width">
+				<?php wp_head(); ?>
+			</head>
+			<body>
+				<?php wp_print_scripts( 'amp-mathml' ); ?>
+			</body>
+		</html>
+		<?php
+		$original_html  = trim( ob_get_clean() );
+		$removed_nodes  = array();
```
ThierryA: `$removed_nodes` doesn't seem to be used in this method.

includes/class-amp-theme-support.php
```diff
@@ -1098,10 +1082,30 @@ public static function prepare_response( $response, $args = array() ) {
...
 		$dom = AMP_DOM_Utils::get_dom( $response );
 
-		self::ensure_required_markup( $dom );
+		$xpath = new DOMXPath( $dom );
+
+		$head = $dom->getElementsByTagName( 'head' )->item( 0 );
+
+		if ( isset( $head ) ) {
+			// Make sure scripts from the body get moved to the head.
+			$scripts = array();
```
ThierryA: Any reason to store the scripts in a var and then run two loops rather than doing the following:
``` PHP
foreach ( $xpath->query( '//body//script[ @custom-element or @custom-template ]' ) as $script ) {
	$head->appendChild( $script );
}
```

tests/test-class-amp-theme-support.php
amedina: Done.

includes/class-amp-theme-support.php
amedina: Makes sense :)

tests/test-class-amp-theme-support.php
```diff
@@ -220,7 +220,6 @@ public function test_prepare_response() {
 	 * @covers AMP_Theme_Support::prepare_response()
 	 */
 	public function test_validate_non_amp_theme() {
-		global $wp_widget_factory, $wp_scripts;
 		$wp_scripts = null;
```
ThierryA: This line still needs to be removed and this PR will be good to go. Thanks @amedina, great stuff!

PR #1042, opened 2018-03-24
Deploy to test environment when PR is merged

westonruter:
For the given Pantheon environments, deploy when a non-PR branch build is done made for a branch that is able to be used as a multidev environment (subdomain) name, except the `develop` branch is mapped to the `master` branch (and the `dev` environment). When a commit is pushed to a release branch like `0.7` then the branch on Pantheon becomes `v0-7`.

Plugin builds get versions that correspond to the commit hash and current datetime so that there is no question as to whether the desired version is present, for example via the `generator` meta tag:

```html
<meta name="generator" content="AMP Plugin v1.0-alpha-22c1619-20180413T033020">
```


westonruter:
@kienstra upon merging you _should_ shortly thereafter see v0.7 deployed to http://v0-7-ampconfdemo.pantheonsite.io/

PR #1043, opened 2018-03-25
Fix AMP Status toggle boolean - Issue #1011

### LINKED ISSUE #1011: AMP status toggle appears in Redirect Rule post in Safe Redirect Manager
This may be a bug with the Safe Redirect Manager, but the AMP status toggle should not appear when editing a redirect rule:

image

rickalee:
Update boolean for adding AMP Status toggle to post_type_supports() instead of is_post_type_viewable() to fix issue #1011 since Safe Redirect Manager sets public_queryable to true in post type registration. 

Fixes #1011

### DIFF
--- a/includes/admin/class-amp-post-meta-box.php
+++ b/includes/admin/class-amp-post-meta-box.php
@@ -167,7 +167,7 @@ class AMP_Post_Meta_Box {
 		$verify = (
 			isset( $post->ID )
 			&&
-			is_post_type_viewable( $post->post_type )
+			post_type_supports( $post->post_type, amp_get_slug() )
 			&&
 			current_user_can( 'edit_post', $post->ID )
 			&&


includes/admin/class-amp-post-meta-box.php
```diff
@@ -167,7 +167,7 @@ public function render_status( $post ) {
 		$verify = (
 			isset( $post->ID )
 			&&
-			is_post_type_viewable( $post->post_type )
+			post_type_supports( $post->post_type, amp_get_slug() )
```
westonruter: I'm not sure if this is the right approach. It would mean that a user would never be presented with the prompt to enable the post type support as shown here:

https://github.com/Automattic/amp-wp/blob/4f2f8dd09a974a6ddf4c28b21e41aa341c79ca0c/templates/admin/amp-status.php#L49-L52

PR #1048, opened 2018-03-30
Add runtime CSS minification, !important replacement, and tree shaking

### LINKED ISSUE #688: CSS in Reader mode is not minified
The AMP CSS inserting by the plugin (`.alignright` through `.back-to-top`) is not minified.
screen shot 2017-04-24 at 4 48 15 pm

### LINKED ISSUE #930: Extend style sanitizer to incude CSS parser

### LINKED ISSUE #990: Incorporate Server Timing API
To help facilitate debugging issues related to the performance of the preprocessor, we should leverage the new [Server Timing API](https://w3c.github.io/server-timing/) to report how long each sanitizer is taking to run, how long WordPress took to generate the page prior to the sanitizers running, etc.

Relates to #959.

westonruter:
- [x] Add support for [Server Timing](https://w3c.github.io/server-timing/) to reveal the time required for the plugin to render a response. Move header-sending logic to `AMP_Response_Headers` class. See #990.
- [x] Minify CSS to remove whitespace and comments. See #688.
- [x] Sanitize CSS by means of a CSS parser instead of relying on regular expressions.
- [x] Remove illegal CSS properties (`-moz-binding` and `behavior`).
- [x] Reduce length of CSS class names generated for `style` attributes.
- [x] If `amp-keyframes` appears anywhere but end of `body`, just move it to the end of the `body`.
- [x] Implement the [replace-important](https://www.npmjs.com/package/replace-important) approach to eliminating `!important` qualifiers.
- [x] Delete illegal CSS @-rules.
- [x] Make sure css_spec gets parsed and included in AMP_Allowed_Tags_Generated.
- [x] Limit `@keyframes` to opacity, transform and -vendorPrefix-transform. 
- [x] Make sure legacy post template still renders properly with stylesheet changes.
- [x] Delete CSS rules that don't apply to the current page based on the class names used, if the total CSS size would otherwise be greater than 50KB. (CSS tree shaking.)
- [x] Delete selectors that reference class names that don't exist.
- [x] Handle case of false negatives when a selector contains `:not(.classname)`. Also strip strings.
- [x] Figure out how to gracefully deal with Genericons and Dashicons.
- [x] Use expiring transient instead of indefinite cache when external cache is not available.
- [x] Improve handling of specificity, including giving higher selector specificity to rules from inline styles. Reduce length of fake ID.
- [x] Address relative URLs, such as for background images.

Fixes #990.
Fixes #930.
Fixes #688.
Closes #610.

Ideas for later:
* Look at other tree-shaking methods beyond class names.
* Add tree-shaking based on IDs.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -2,10 +2,21 @@
 
 Thanks for taking the time to contribute!
 
-To clone this repository
-``` bash
-$ git clone --recursive git@github.com:Automattic/amp-wp.git
+To start, clone this repository into your WordPress install being used for development:
+
+```bash
+cd wp-content/plugins && git clone --recursive git@github.com:Automattic/amp-wp.git amp
 ```
+
+If you happened to have cloned without `--recursive` previously, please do `git submodule update --init` to ensure the [dev-lib](https://github.com/xwp/wp-dev-lib/) submodule is available for development.
+
+Lastly, to get the plugin running in your WordPress install, run `composer install` and then activate the plugin via the WordPress dashboard or `wp plugin activate amp`.
+
+To install the `pre-commit` hook, do `bash dev-lib/install-pre-commit-hook.sh`.
+
+Note that pull requests will be checked against [WordPress-Coding-Standards](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards) with PHPCS, and for JavaScript linting is done with ESLint and (for now) JSCS and JSHint.
+
+To run the Grunt commands, please first `npm install -g grunt-cli` and then `npm install`.
 
 ## Updating Allowed Tags And Attributes
 
@@ -72,6 +83,7 @@ When you push a commit to your PR, Travis CI will run the PHPUnit tests and snif
 
 Contributors who want to make a new release, follow these steps:
 
+0. Do `grunt build` and install the `amp.zip` onto a normal WordPress install running a stable release build; do smoke test to ensure it works.
 1. Bump plugin versions in `package.json` (×1), `package-lock.json` (×1, just do `npm install` first), `composer.json` (×1), and in `amp.php` (×2: the metadata block in the header and also the `AMP__VERSION` constant).
 2. Add changelog entry to readme.
 3. Merge release branch into `master`.
--- a/includes/class-amp-response-headers.php
+++ b/includes/class-amp-response-headers.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * Class AMP_Response_Headers
+ *
+ * @since 1.0
+ * @package AMP
+ */
+
+/**
+ * Class AMP_Response_Headers
+ */
+class AMP_Response_Headers {
+
+	/**
+	 * Headers sent (or attempted to be sent).
+	 *
+	 * @since 1.0
+	 * @see AMP_Theme_Support::send_header()
+	 * @var array[]
+	 */
+	public static $headers_sent = array();
+
+	/**
+	 * Send an HTTP response header.
+	 *
+	 * This largely exists to facilitate unit testing but it also provides a better interface for sending headers.
+	 *
+	 * @since 0.7.0
+	 *
+	 * @param string $name  Header name.
+	 * @param string $value Header value.
+	 * @param array  $args {
+	 *     Args to header().
+	 *
+	 *     @type bool $replace     Whether to replace a header previously sent. Default true.
+	 *     @type int  $status_code Status code to send with the sent header.
+	 * }
+	 * @return bool Whether the header was sent.
+	 */
+	public static function send_header( $name, $value, $args = array() ) {
+		$args = array_merge(
+			array(
+				'replace'     => true,
+				'status_code' => null,
+			),
+			$args
+		);
+
+		self::$headers_sent[] = array_merge( compact( 'name', 'value' ), $args );
+		if ( headers_sent() ) {
+			return false;
+		}
+
+		header(
+			sprintf( '%s: %s', $name, $value ),
+			$args['replace'],
+			$args['status_code']
+		);
+		return true;
+	}
+
+	/**
+	 * Send Server-Timing header.
+	 *
+	 * @since 1.0
+	 * @todo What is the ordering in Chrome dev tools? What are the colors about?
+	 * @todo Is there a better name standardization?
+	 * @todo Is there a way to indicate nested server timings, so an outer method's own time can be seen separately from the inner method's time?
+	 *
+	 * @param string $name        Name.
+	 * @param float  $duration    Duration. If negative, will be added to microtime( true ). Optional.
+	 * @param string $description Description. Optional.
+	 * @return bool Return value of send_header call.
+	 */
+	public static function send_server_timing( $name, $duration = null, $description = null ) {
+		$value = $name;
+		if ( isset( $description ) ) {
+			$value .= sprintf( ';desc=%s', wp_json_encode( $description ) );
+		}
+		if ( isset( $duration ) ) {
+			if ( $duration < 0 ) {
+				$duration = microtime( true ) + $duration;
+			}
+			$value .= sprintf( ';dur=%f', $duration * 1000 );
+		}
+		return self::send_header( 'Server-Timing', $value, array( 'replace' => false ) );
+	}
+}
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -44,29 +44,21 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	private $stylesheets = array();
 
 	/**
-	 * Spec for style[amp-custom] cdata.
-	 *
-	 * @since 1.0
-	 * @var array
-	 */
-	private $style_custom_cdata_spec;
-
-	/**
-	 * Spec for style[amp-keyframes] cdata.
-	 *
-	 * @since 1.0
-	 * @var array
-	 */
-	private $style_keyframes_cdata_spec;
-
-	/**
-	 * Current CSS size.
+	 * Current amp-custom CSS size.
 	 *
 	 * Sum of CSS located in $styles and $stylesheets.
 	 *
 	 * @var int
 	 */
 	private $current_custom_size = 0;
+
+	/**
+	 * Spec for style[amp-custom] cdata.
+	 *
+	 * @since 1.0
+	 * @var array
+	 */
+	private $style_custom_cdata_spec;
 
 	/**
 	 * The style[amp-custom] element.
@@ -76,12 +68,28 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	private $amp_custom_style_element;
 
 	/**
-	 * The found style[amp-keyframe] elements.
+	 * The found style[amp-keyframe] stylesheets.
 	 *
 	 * @since 1.0
-	 * @var DOMElement[]
+	 * @var string[]
 	 */
-	private $amp_keyframes_style_elements = array();
+	private $keyframes_stylesheets = array();
+
+	/**
+	 * Current amp-keyframes CSS size.
+	 *
+	 * @since 1.0
+	 * @var int
+	 */
+	private $current_keyframes_size = 0;
+
+	/**
+	 * Spec for style[amp-keyframes] cdata.
+	 *
+	 * @since 1.0
+	 * @var array
+	 */
+	private $style_keyframes_cdata_spec;
 
 	/**
 	 * Regex for allowed font stylesheet URL.
@@ -109,6 +117,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	/**
 	 * XPath.
 	 *
+	 * @since 1.0
 	 * @var DOMXPath
 	 */
 	private $xpath;
@@ -116,6 +125,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	/**
 	 * Amount of time that was spent parsing CSS.
 	 *
+	 * @since 1.0
 	 * @var float
 	 */
 	private $parse_css_duration = 0.0;
@@ -325,24 +335,22 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 * @param DOMElement $element Style element.
 	 */
 	private function process_style_element( DOMElement $element ) {
-		if ( $element->hasAttribute( 'amp-keyframes' ) ) {
-			$validity = $this->validate_amp_keyframe( $element );
-			if ( is_wp_error( $validity ) ) {
-				$this->remove_invalid_child( $element, array(
-					'message' => $validity->get_error_message(),
-				) );
-			}
-			return;
-		}
 
-		$stylesheet = trim( $element->textContent );
+		// @todo Any @keyframes rules could be removed from amp-custom and instead added to amp-keyframes.
+		$is_keyframes = $element->hasAttribute( 'amp-keyframes' );
+		$stylesheet   = trim( $element->textContent );
+		$cdata_spec   = $is_keyframes ? $this->style_keyframes_cdata_spec : $this->style_custom_cdata_spec;
 		if ( $stylesheet ) {
-			$stylesheet = $this->process_stylesheet( $stylesheet, $element );
+			$stylesheet = $this->process_stylesheet( $stylesheet, $element, array(
+				'allowed_at_rules'   => $cdata_spec['css_spec']['allowed_at_rules'],
+				'property_whitelist' => $cdata_spec['css_spec']['allowed_declarations'],
+				'validate_keyframes' => $cdata_spec['css_spec']['validate_keyframes'],
+			) );
 		}
 
 		// Remove if surpasses max size.
 		$length = strlen( $stylesheet );
-		if ( $this->current_custom_size + $length > $this->style_custom_cdata_spec['max_bytes'] ) {
+		if ( ( $is_keyframes ? $this->current_keyframes_size : $this->current_custom_size ) + $length > $cdata_spec['max_bytes'] ) {
 			$this->remove_invalid_child( $element, array(
 				'message' => __( 'Too much CSS enqueued.', 'amp' ),
 			) );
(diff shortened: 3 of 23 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -349,230 +429,544 @@ private function process_link_element( DOMElement $element ) {
...
+
+		$stylesheet = '';
+		foreach ( $parsed['stylesheet'] as $stylesheet_part ) {
+			if ( is_array( $stylesheet_part ) ) {
+				list( $selectors_parsed, $declaration_block ) = $stylesheet_part;
+
+				$selectors = array();
+				if ( $should_tree_shake ) {
+					foreach ( $selectors_parsed as $selector => $class_names ) {
+						if ( count( $class_names ) === count( array_intersect( $class_names, $this->used_class_names ) ) ) {
```
westonruter: Better to use `array_diff` here instead maybe.

westonruter:
See in action for Twenty Fifteen theme at https://treeshaking-ampconfdemo.pantheonsite.io/

amedina:
The capabilities being added in this PR are great @westonruter 

Adding here[ this document](https://goo.gl/WGiSYM) where @pbakaus which outlines a two-tiered  optimization approach; let's review it and see if that approach can be used to expand your solution here.

includes/class-amp-response-headers.php
```diff
@@ -0,0 +1,88 @@
...
+		return true;
+	}
+
+	/**
+	 * Send Server-Timing header.
+	 *
+	 * @since 1.0
+	 * @todo What is the ordering in Chrome dev tools? What are the colors about?
+	 * @todo Is there a better name standardization?
+	 * @todo Is there a way to indicate nested server timings, so an outer method's own time can be seen separately from the inner method's time?
```
igrigorik: - Order in CDT is unspecified, ST does not provide startTime or epoch
- No guidance on naming either
- No support for nesting (intentionally :))

includes/class-amp-response-headers.php
```diff
@@ -0,0 +1,88 @@
...
+	 *
+	 * @param string $name        Name.
+	 * @param float  $duration    Duration. If negative, will be added to microtime( true ). Optional.
+	 * @param string $description Description. Optional.
+	 * @return bool Return value of send_header call.
+	 */
+	public static function send_server_timing( $name, $duration = null, $description = null ) {
+		$value = $name;
+		if ( isset( $description ) ) {
+			$value .= sprintf( ';desc=%s', wp_json_encode( $description ) );
```
igrigorik: Is description carrying a JSON blob? Technically, that's allowed since we just accept a string, but not something we want to encourage.. Passing around large JSON blobs via headers is not a good idea.

pbakaus:
My $0.02: We should be extra careful to not strip out stuff that could break things on the page, without a proper way to flag and notify the webmaster. So essentially, we have a bunch of categories of optimizations:

1. Minification (generally safe)
2. Substitution (Also generally safe, e.g. the `!important` trick)
3. Filtering (Not safe)

If filtering is required to make something AMP valid, it's probably better to either:

1. Not filter, and throw warning that AMP is invalid because of these things
2. Filter, but throw strong visual warnings on the theme page / wp-admin

I'm in the former camp, but either way probably works.

includes/class-amp-response-headers.php
westonruter: The description is just a string so `wp_json_encode` is just being used as a shortcut to add quote marks around a string literal. That's probably overkill and it shows just strip out any double quote marks provided in the `$description`.

westonruter:
@pbakaus Yes, the filtering (tree shaking) is required in order to get the WordPress theme+plugin stylesheets to be valid AMP if they weren't written specifically for AMP to begin with. So the changes here allow the Twenty Fifteen theme to be served as AMP without doing any changes. Nevertheless, the point you make in the doc about `amp-live-list` is a key one: content could be added to the page later that depends on a rule that is filtered. So I think the solution here is to only do the filtering if the size is greater than 50KB and then to show the warning as you suggest.

pbakaus:
@westonruter SGTM! For sub 50kb, we can still do any optimizations that are safe, of course.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -349,230 +431,741 @@ private function process_link_element( DOMElement $element ) {
...
+				}
+				$extension   = preg_replace( ':.+/(.+-)?:', '', $mime_type );
+				$guessed_url = preg_replace(
+					':(?<=\.)\w+(\?.*)?(#.*)?$:', // Match the file extension in the URL.
+					$extension,
+					$source_file_url->getURL()->getString(),
+					1,
+					$count
+				);
+				if ( $count ) {
```
westonruter: We should really confirm whether or not the guessed file does exist. While it does in the core themes for Genericons and in core itself for Dashicons, another theme may decide to not bundle the WOFF file at all.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -349,230 +431,741 @@ private function process_link_element( DOMElement $element ) {
...
+				$properties = $rules->getRules();
+				foreach ( $properties as $property ) {
+					$vendorless_property_name = preg_replace( '/^-\w+-/', '', $property->getRule() );
+					if ( ! in_array( $vendorless_property_name, $options['property_whitelist'], true ) ) {
+						$validation_errors[] = array(
+							'code'           => 'illegal_css_property',
+							'property_name'  => $property->getRule(),
+							'property_value' => $property->getValue(),
+						);
+						$rules->removeRule( $property->getRule() );
```
westonruter: This removal should be removed since it will break many keyframe animations out there which don't use hardware accelerated properties.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -349,230 +431,741 @@ private function process_link_element( DOMElement $element ) {
...
+			foreach ( $parsed['validation_errors'] as $validation_error ) {
+				call_user_func( $this->args['validation_error_callback'], array_merge( $validation_error, compact( 'node' ) ) );
+			}
+		}
+
+		$stylesheet = '';
+		foreach ( $parsed['stylesheet'] as $stylesheet_part ) {
+			if ( is_array( $stylesheet_part ) ) {
+				list( $selectors_parsed, $declaration_block ) = $stylesheet_part;
+				if ( $should_tree_shake ) {
```
westonruter: This logic needs to be deferred to when all stylesheets are finished processing to then decide whether or not tree-shaking should be performed if it surpasses the limit.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -349,230 +431,741 @@ private function process_link_element( DOMElement $element ) {
...
+		} else {
+			foreach ( $options['property_blacklist'] as $illegal_property_name ) {
+				$properties = $ruleset->getRules( $illegal_property_name );
+				foreach ( $properties as $property ) {
+					$validation_errors[] = array(
+						'code'           => 'illegal_css_property',
+						'property_name'  => $property->getRule(),
+						'property_value' => $property->getValue(),
+					);
+					$ruleset->removeRule( $property->getRule() );
```
westonruter: Consider skipping the removal (or let it be configurable) since it could have unintended results, but since `-moz-binding` and `behavior` are the only blacklisted properties (for now), it may not be the end of the world if they are removed regardless.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -349,230 +431,741 @@ private function process_link_element( DOMElement $element ) {
...
+			$properties = $ruleset->getRules();
+			foreach ( $properties as $property ) {
+				$vendorless_property_name = preg_replace( '/^-\w+-/', '', $property->getRule() );
+				if ( ! in_array( $vendorless_property_name, $options['property_whitelist'], true ) ) {
+					$validation_errors[] = array(
+						'code'           => 'illegal_css_property',
+						'property_name'  => $property->getRule(),
+						'property_value' => $property->getValue(),
+					);
+					$ruleset->removeRule( $property->getRule() );
```
westonruter: Consider just reporting validation error instead of removing, or having an option to configure the behavior.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -349,230 +431,741 @@ private function process_link_element( DOMElement $element ) {
...
+			$options
+		);
 
-		if ( $this->keyframes_max_size && strlen( $style->textContent ) > $this->keyframes_max_size ) {
-			return new WP_Error( 'max_bytes', __( 'amp-keyframes is too large', 'amp' ) );
-		}
+		$stylesheet        = array();
+		$validation_errors = array();
+		try {
+			$parser_settings = Sabberworm\CSS\Settings::create()->withMultibyteSupport( false );
```
westonruter: I think `->withMultibyteSupport( false )` can be removed. I only added it to see if it would improve performance, but it only helped marginally, and since the results are cached it doesn't matter so much.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -349,230 +431,741 @@ private function process_link_element( DOMElement $element ) {
...
+		return $validation_errors;
+	}
+
+	/**
+	 * Replace !important qualifiers with more specific rules.
+	 *
+	 * @since 1.0
+	 * @see https://www.npmjs.com/package/replace-important
+	 * @see https://www.ampproject.org/docs/fundamentals/spec#important
+	 * @todo Further tailor for specificity. See https://github.com/ampproject/ampstart/blob/4c21d69afdd07b4c60cd190937bda09901955829/tools/replace-important/lib/index.js#L87-L126.
```
westonruter: This has been implemented.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -349,230 +431,741 @@ private function process_link_element( DOMElement $element ) {
...
-		 * @todo This removal needs to be reported.
-		 */
-		if ( false !== strpos( $value, 'important' ) ) {
-			$value = preg_replace( '/\s*\!\s*important$/', '', $value );
+		if ( $this->current_custom_size + $length > $this->style_custom_cdata_spec['max_bytes'] ) {
+			$this->remove_invalid_attribute( $element, $style_attribute, array(
+				/* translators: %d is the number of bytes over the limit */
+				'message' => sprintf( __( 'Too much CSS enqueued (by %d bytes).', 'amp' ), ( $this->current_custom_size + $length ) - $this->style_custom_cdata_spec['max_bytes'] ),
+			) );
+			return;
```
westonruter: This needs to be eliminated. At the very least it should be done after all stylesheets have been gathered and filtered (tree shook). But even then it can result in broken styles if the styles are removed.

includes/class-amp-response-headers.php
igrigorik: Gotcha, well.. If someone does pass a JSON payload, technically they're not wrong. I just wanted to nudge us away from doing so if that's what we're considering. :)

contributing.md
```diff
@@ -2,11 +2,22 @@
...
-To clone this repository
-``` bash
-$ git clone --recursive git@github.com:Automattic/amp-wp.git
+To start, clone this repository into your WordPress install being used for development:
+
+```bash
+cd wp-content/plugins && git clone --recursive git@github.com:Automattic/amp-wp.git amp
 ```
 
+If you happened to have cloned without `--recursive` previously, please do `git submodule update --init` to ensure the [dev-lib](https://github.com/xwp/wp-dev-lib/) submodule is available for development.
```
kienstra: It's a good idea to add this.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -349,230 +431,741 @@ private function process_link_element( DOMElement $element ) {
...
+				}
+				if ( ! empty( $selectors ) ) {
+					$stylesheet .= implode( ',', $selectors ) . $declaration_block;
+				}
+			} else {
+				$stylesheet .= $stylesheet_part;
+			}
 		}
+
 		return $stylesheet;
```
kienstra: Would it help to call the `validation_error_callback` here if `$stylesheet` is empty and it's from a `<link>`? This would account for when a stylesheet has no rules that pertain to DOM classes. Like if a plugin enqueues a stylesheet for its shortcode on every request.

```
if ( empty( $stylesheet ) && ( 'link' === $node->nodeName ) && ! empty( $this->args['validation_error_callback'] ) ) {
        // Call validation callback
```
        
We'd probably have to deprecate [amp-default.css](https://github.com/Automattic/amp-wp/blob/develop/assets/css/amp-default.css) to do this.

westonruter:
@amedina yes, I referred to `replace-important` for the implementation here. I think it is improved on here in a couple ways:

* It uses `_` instead of `FK_ID` as the dummy ID for the `:not()` selector. This is important because the `:not()` has to repeated to achieve the required specificity.
* It also accounts for adding specificity to rules like `html { color:red !important; }` in that it amends `:not(#_)` to the `html` root selector, as opposed to adding an additional ancestor selector (which shouldn't be valid). So here this becomes `html:not(#_) { color:red; }` instead of `:root:not(#_) html { color:red; }`.

westonruter:
> the point you make in the doc about amp-live-list is a key one: content could be added to the page later that depends on a rule that is filtered

Just occurred to me that we could help guard against this from happening by adding logic to skip removing selectors that reference `amp-live-list` or `amp-list` elements. Also, we could skip selectors that reference `div[submit-success]` and `div[submit-error]`. 

Nevertheless, there would be no way to statically account for class names assigned via `amp-bind`, so that's a danger, unless we adopt some some naming convention for `amp-bind`-supplied class names used in selectors.

@pbakaus Any other key dynamic content cases to have in mind?

pbakaus:
@westonruter sounds sensible! Thats the only ones I have in mind right now.

camelburrito:
@westonruter usually the ones associated with `amp-bind` can be searchable in the dom (string search), even that would be a little risky because there could be string concat etc happening. 
Naming conventions seems to be a sane idea. I guess these could be optional flag that we can enable (like list of prefixes that should not be deleted?)


And on the replace-important improvements feel free to edit the code here 
https://github.com/ampproject/ampstart/tree/master/tools/replace-important
I can help you push it live. If that would help you keep it more modular.

westonruter:
> I guess these could be optional flag that we can enable (like list of prefixes that should not be deleted?)

@camelburrito I added initial support for this via the `dynamic_element_selectors` argument in 
bede9de. By default it only includes the known dynamic content ancestors, but a given theme could amend it with additional dynamic selectors to exclude from removal.

@kienstra This is ready for review to merge so we can get this testing. I want to follow up later with some more improvements including the added ability to knowingly bypass removal of elements, attributes, and styles that are invalid but which a given site cannot afford to be removed. I'm thinking we'd want CSS over the 50KB limit to default to not be removed, even if that means it is invalid AMP. In terms of implementation here, I think this can be implemented by allowing the `validation_error_callback` to return `false` as a way to prevent removal from happening. That would allow a theme/plugin to decide on a case-by-case basis which things should get removed. This is in regards to https://github.com/Automattic/amp-wp/pull/1048#pullrequestreview-108681682 and https://github.com/Automattic/amp-wp/pull/1048#discussion_r178599928

tests/test-amp-style-sanitizer.php
```diff
@@ -154,31 +200,75 @@ public function test_body_style_attribute_sanitizer( $source, $expected_content,
...
+			'style_on_root_element' => array(
+				'<html amp style="color:red;"><head><meta charset="utf-8"><style amp-custom>html { background-color: blue !important; }</style></head><body>Hi</body></html>',
+				array(
+					'html:not(#_):not(#_){background-color:blue;}',
+					':root:not(#_):not(#_):not(#_):not(#_):not(#_) .amp-wp-10b06ba{color:red;}',
+				),
+				array(),
+			),
+			'styles_with_dynamic_elements' => array(
+				implode( '', array(
```
kienstra: Great idea to use `implode()` here for the long document.

tests/test-amp-style-sanitizer.php
```diff
@@ -130,11 +166,19 @@ public function get_body_style_attribute_data() {
 	 * @param string $source               Source.
 	 * @param string $expected_content     Expected content.
 	 * @param string $expected_stylesheets Expected stylesheets.
+	 * @param array  $expected_errors      Expected error codes.
 	 */
-	public function test_body_style_attribute_sanitizer( $source, $expected_content, $expected_stylesheets ) {
+	public function test_body_style_attribute_sanitizer( $source, $expected_content, $expected_stylesheets, $expected_errors = array() ) {
```
kienstra: Maybe the 3rd `$expected_errors` parameter isn't necessary, given that the `dataProvider` [get_link_and_style_test_data](https://github.com/Automattic/amp-wp/pull/1048/files#diff-78b16003565f81870be4068700bdbfc5R200) always has a 3rd array item of `array()`. But this isn't a blocker to approving.

westonruter: Yeah, you're right. But I suppose it aligns with the other tests and opens the door to testing for validation errors in the future.

PR #1050, opened 2018-03-30
Output admin notice if persistent object caching is not present

### LINKED ISSUE #960: Add admin notice if persistent object caching is not present
The plugin really performs best (as does WordPress generally) when a persistent object cache is present. We detect whether it is present and inform if it is missing.

oscarssanchez:
Hi @westonruter,

Could you use this PR which addresses #960 ?

I was not sure where it was best to write the method in, so I opted to write it in **class-amp-validation-utils** but i'm open to your suggestions.

I'm also not sure if the output message would be good as it is now. I opted to include a link to [https://codex.wordpress.org/Class_Reference/WP_Object_Cache#Persistent_Caching](url) since users might not know what persistent object cache is and how to enable it.

git

Fixes #960.

### DIFF
--- a/includes/utils/class-amp-validation-utils.php
+++ b/includes/utils/class-amp-validation-utils.php
@@ -214,7 +214,7 @@ class AMP_Validation_Utils {
 		add_filter( 'bulk_actions-edit-' . self::POST_TYPE_SLUG, array( __CLASS__, 'add_bulk_action' ), 10, 2 );
 		add_filter( 'handle_bulk_actions-edit-' . self::POST_TYPE_SLUG, array( __CLASS__, 'handle_bulk_action' ), 10, 3 );
 		add_action( 'admin_notices', array( __CLASS__, 'remaining_error_notice' ) );
-		add_action( 'admin_notices', array( __CLASS__, 'persistent_caching_object_notice' ) );
+		add_action( 'admin_notices', array( __CLASS__, 'persistent_object_caching_notice' ) );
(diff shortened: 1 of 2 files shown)


oscarssanchez:
Thanks for your feedback @kienstra . 

I made the corresponding changes.

includes/utils/class-amp-validation-utils.php
```diff
@@ -1791,4 +1792,22 @@ public static function get_recheck_link( $post, $redirect_url, $recheck_url = nu
...
+	 * @return void
+	 */
+	public static function persistent_object_caching_notice() {
+		if ( ! wp_using_ext_object_cache() && 'toplevel_page_amp-options' === get_current_screen()->id ) {
+			printf(
+				'<div class="notice notice-warning is-dismissible"><p>%s <a href="%s">%s</a></p><button type="button" class="notice-dismiss"><span class="screen-reader-text">%s</span></button></div>',
+				esc_html__( 'The AMP plugin performs at its best when persistent object cache is enabled.', 'amp' ),
+				esc_url( 'https://codex.wordpress.org/Class_Reference/WP_Object_Cache#Persistent_Caching' ),
+				esc_html__( 'More details', 'amp' ),
+				esc_html__( 'Dismiss this notice.', 'amp' )
```
westonruter: I think we should eliminate the dismissal because as soon as the page reload the notice comes right back.

westonruter: @oscarssanchez once this change is made the PR will be good to merge.

oscarssanchez:
HI @westonruter,

Thanks for your comments, i just did the changes.

Issue #1030, opened 2018-03-20
Prevent AMP validation error due to img src containing spaces


westonruter:
If I add an `img` with spaces as the `src` like so:

```html
<img src="   " width="100" height="100">
```

Then this gets output as:

```html
<amp-img src="" width="100" height="100" sizes="(min-width: 100px) 100px, 100vw" class="amp-wp-enforced-sizes"></amp-img>
```

AMP complains about this:

> Missing URL for attribute 'src' in tag 'amp-img'.

If an `img` is added with an empty `src` then the `amp-img` gets removed entirely. That should also be the behavior when `src` contains just whitespace.

Support topic: https://wordpress.org/support/topic/google-console-detected-amp-error/

PR #1054, opened 2018-04-03
Sanitize <img> tags when src attribute contains only spaces. 

amedina:
This PR fixes issue #1030.


westonruter:
Re-basing for the 0.7 branch since this is a nice little bugfix for the next release.

PR #1058, opened 2018-04-04
Update versions tested in Travis build matrix

westonruter:
* Add PHP 7.2 to Travis test matrix; WordPress currently recommends PHP 7.2.
* Include testing PHP 5.6 (the most popular PHP version), as well as 5.5.
* Include testing WP 4.8 and trunk.
* Skip running PHPCS on every build; only needs to be run once.
* Clean up some PHPCS issues along the way to cause PHP to be checked in the build.

### DIFF
--- a/wpcom-helper.php
+++ b/wpcom-helper.php
@@ -45,14 +45,15 @@ add_filter( 'amp_post_template_metadata', 'jetpack_amp_post_template_metadata', 
 /**
  * Add publisher and image metadata.
  *
+ * @since 0.3
+ *
  * @param array   $metadata Metadata array.
  * @param WP_Post $post     Post.
- *
  * @return array Modified metadata array.
  */
 function jetpack_amp_post_template_metadata( $metadata, $post ) {
 	if ( isset( $metadata['publisher'] ) && ! isset( $metadata['publisher']['logo'] ) ) {
-		$metadata = wpcom_amp_add_blavatar_to_metadata( $metadata, $post );
+		$metadata = wpcom_amp_add_blavatar_to_metadata( $metadata );
 	}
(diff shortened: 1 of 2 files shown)


wpcom-helper.php
```diff
@@ -24,14 +29,27 @@ function jetpack_amp_disable_the_content_filters( $post_id ) {
...
+ * Add publisher and image metadata.
+ *
+ * @param array   $metadata Metadata array.
+ * @param WP_Post $post     Post.
+ *
+ * @return array Modified metadata array.
+ */
 function jetpack_amp_post_template_metadata( $metadata, $post ) {
 	if ( isset( $metadata['publisher'] ) && ! isset( $metadata['publisher']['logo'] ) ) {
 		$metadata = wpcom_amp_add_blavatar_to_metadata( $metadata, $post );
```
kienstra: It'd probably be good to remove the second argument `$post` here, as the function function [now only accepts](https://github.com/Automattic/amp-wp/pull/1058/files#diff-e3795a514c859a54511c3580d489cab3R73) one argument.

wpcom-helper.php
```diff
@@ -1,17 +1,22 @@
 <?php
+/**
+ * WPCOM-specific things.
+ *
+ * @todo Move this into Jetpack. See https://github.com/Automattic/amp-wp/issues/1021
+ * @package AMP
+ */
```
kienstra: Nice additions of DocBlocks and improvements in coding style here.

wpcom-helper.php
westonruter: Good catch. Fixed in f683d9f.

Issue #1003, opened 2018-03-08
Add ability to acknowledge and suppress/ignore specific validation errors


westonruter:
With #971 there is now reporting on the validation errors that the sanitizer scrubs from the output to ensure valid AMP. For example (and these screens need better UX/design):

* ![image](https://user-images.githubusercontent.com/134745/37178250-d3353c82-22d6-11e8-9504-678a10e71977.png)
* ![image](https://user-images.githubusercontent.com/134745/37178264-d9062644-22d6-11e8-8650-2da2b7b865fe.png)
* ![image](https://user-images.githubusercontent.com/134745/37178294-e108f83a-22d6-11e8-83f1-67279c8344ac.png)

I think it is clear that there should be a way to dismiss certain validation errors from being reported. When something is dismissed, it would then not show up again. 

For example, if a theme uses an illegal `overflow` CSS property, then this is going to get reported on every URL. But if it doesn't cause a big problem for the AMP rendering the page, then this is something the user probably would like to ignore. If you have a plugin that does one thing you know is bad, and the removal doesn't cause any problems in AMP, then the user should be able to just _acknowledge_ it was removed and cease getting notified about it.

postphotos:
Hi @westonruter, I plan to write a formal user story and ACs, but dropping this in for now:
```
Validation errors exist now for a given URL and issues will be grouped together in the same post.
- if you had 100 posts and each had a different script tag, you want to understand what's unique.
- have a count of unique instances of the post so you can know what is unique so you can priortize what should be fixed.
- Ability to suppress errors for not valid AMP so you don't get reported ("deal with it later") - santizer can take care of it anyway.
```
Let me know if there's any details here I'm missing.

Related #1006, which should inform/be informed by progress in both places.

westonruter:
Yeah, I think this needs some UX insights to figure out what would be the most beneficial for users. It's closely related to my https://github.com/Automattic/amp-wp/issues/1060#issuecomment-378991968 and the story of activating AMP on a site. It could go so far as a “wizard”-like setup flow. But beyond the initial setup, there is then the ongoing monitoring of validity, as new content is added and plugins are added/removed.

westonruter:
I think a nice flow for end-users to enable AMP theme support without doing coding:

1. I activate the AMP plugin. 
2. The plugin sees that I don't have AMP theme support.
3. The plugin asks if you would like to try AMP theme support on your theme.
[...]

Issue #958, opened 2018-02-14
Enable generation of Optimized (aka Transformed) AMP pages


westonruter:
We can speed up the serving of AMP pages by pre-rendering them and serving them in the same way they are served from an AMP cache. In short, the processes performed by the [AMP Optimizer](https://github.com/ampproject/amp-toolbox/tree/master/packages/optimizer) should be ported into the plugin. Something important to note here that when pre-rendering is performed, there will somewhat counter-intuitively need to be a paired mode even when a site is AMP Canonical. The reason is:

> Warning: optimized AMPs will no longer be valid AMPHTML. Don't use this to optimize AMPs meant for being used by platforms such as Google Search.

So the pre-rendered AMP version of a page would would be canonical URL, and there would be an `amphtml` link to the same exact content without the pre-rendering logic applied. This is similar to the existing paired mode works, _except_ here the canonical version is generated _from_ the AMP version via pre-rendering. So it's important to note that this pre-rendering should not be offered on sites that are not using Canonical AMP, as it would not make sense to have a pre-rendered AMP version alongside the valid AMP and the non-AMP content.

As for where the pre-rendering should be performed, it should be done right after the call to `AMP_Content_Sanitizer::sanitize_document()` in `AMP_Theme_Support::prepare_response()`:

https://github.com/Automattic/amp-wp/blob/1ab37c654d8c36cb1e17f1c997a6a2f00002273c/includes/class-amp-theme-support.php#L813-L815

For example:

```php
$assets = AMP_Content_Sanitizer::sanitize_document( $dom, self::$sanitizer_classes, $args );
self::ensure_required_markup( $dom );
if ( ! is_amp_endpoint() ) {
    do_the_function_that_runs_amp_prerendering( $dom ); // 👈
}
```

Since `is_amp_endpoint()` currently always returns `true` if `amp_is_canonical()` then this will need to be amended to support the case for pre-rendering. In other words, even on a canonical-site there will be the need for separate AMP endpoint URLs. So the `if ( amp_is_canonical() ) { return true; }` conditional may need to be removed here. Related to #934.

👉 For more on server-side rendering, see [Faster AMP on the origin: AMP + SSR = ⚡](https://blog.amp.dev/2019/08/07/faster-amp-on-the-origin-amp-ssr/).

[...]

PR #1063, opened 2018-04-06
[WIP] Add dynamic handling of validation errors

### LINKED ISSUE #956: [SUGGESTION] Regarding 0.7 Progress Feature Enabler
👍 After trying, beta testing, and watching the AMPConf talk about `0.7` version of AMP-WP I'm super excited about the AMP canonical pages. Love the work you guys have done to improve this plugin. While I understand the need for AMP valid pages — I'd like to suggest that allow theme authors to move towards AMP Canonical in a progressive one feature at a time manner. 

✅ E.g. I have a lot of JS in my theme at https://WPCouple.com and I'd love to take advantage of CSS inlining feature and make my theme compatible to go all in on AMP Canonical mode but I'd love to have time to do that, so instead of going all in — I'd want to enable one feature at a time (since I don't want all of my JS stripped out right away, but want to make use of as many AMP components that I can start using now). Let's call it a progressive way of getting towards an all AMP WP theme instead of completely changing the theme to do that.

🤞This way my theme might not be AMP valid but I'd still benefit from the fast page speed. Also, that's what most of the other old themes that are non-amp would want to do.

Would love to hear thoughts of @westonruter @ThierryA and @amedina on this 👍 
Peace ✌️

westonruter:
Follow up on https://github.com/Automattic/amp-wp/pull/1048#issuecomment-379167176

> I want to follow up later with some more improvements including the added ability to knowingly bypass removal of elements, attributes, and styles that are invalid but which a given site cannot afford to be removed. I'm thinking we'd want CSS over the 50KB limit to default to not be removed, even if that means it is invalid AMP. In terms of implementation here, I think this can be implemented by allowing the `validation_error_callback` to return `false` as a way to prevent removal from happening. That would allow a theme/plugin to decide on a case-by-case basis which things should get removed. This is in regards to https://github.com/Automattic/amp-wp/pull/1048#pullrequestreview-108681682 and https://github.com/Automattic/amp-wp/pull/1048#discussion_r178599928

There are a few validation errors that should specifically be not sanitized by default since they could seriously impact a site:

* Keyframe properties that aren't hardware accelerated.
* CSS exceeding 50KB.
* External stylesheets that can't be inlined.
* …

In addition to allowing the user to manually acknowledge whether a given validation error should be considered critical or be ignored (#1003), there also needs to be a way to dynamically make this decision based on a user-supplied validation callback, a capability which would also be used in #1087 to check whether a validation error is a deal breaker for serving AMP for the current page.

# Todo

- [ ] When a deal-breaker validation error occurs, prevent AMP from being served from a given URL.
- [ ] Make sure that error codes are distinct enough to be differentiated.
- [ ] Make sure that validation errors still display when editing a post.
- [x] Make sure that sources are located when not serving validation errors from cache (pass `$node` to methods).
- [ ] 🚫 ~Decide on which validation errors should not result in sanitization by default (if any).~
- [ ] 🚫 ~Consider having a AMP admin setting for whether to allow select validation errors when they occur.~
- [ ] 🚫 ~Look at factoring in sanitization-skipping to paired mode, to serve `canonical` dirty AMP documents but valid non-canonical (`amphtml`) ones. This is similar to incorporating prerendering into the postprocessing flow (#958).~

Closes #956

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -139,7 +139,9 @@ function amp_init() {
 
 	add_rewrite_endpoint( amp_get_slug(), EP_PERMALINK );
 
-	AMP_Validation_Utils::init();
+	AMP_Validation_Utils::init( array(
+		'debug' => isset( $_REQUEST[ AMP_Validation_Utils::DEBUG_QUERY_VAR ] ), // phpcs:ignore WordPress.CSRF.NonceVerification.NoNonceVerification
+	) );
 	AMP_Theme_Support::init();
 	AMP_Post_Type_Support::add_post_type_support();
 	add_filter( 'request', 'amp_force_query_var_value' );
--- a/includes/class-amp-response-headers.php
+++ b/includes/class-amp-response-headers.php
@@ -63,9 +63,6 @@ class AMP_Response_Headers {
 	 * Send Server-Timing header.
 	 *
 	 * @since 1.0
-	 * @todo What is the ordering in Chrome dev tools? What are the colors about?
-	 * @todo Is there a better name standardization?
-	 * @todo Is there a way to indicate nested server timings, so an outer method's own time can be seen separately from the inner method's time?
 	 *
 	 * @param string $name        Name.
 	 * @param float  $duration    Duration. If negative, will be added to microtime( true ). Optional.
@@ -75,7 +72,7 @@ class AMP_Response_Headers {
 	public static function send_server_timing( $name, $duration = null, $description = null ) {
 		$value = $name;
 		if ( isset( $description ) ) {
-			$value .= sprintf( ';desc=%s', wp_json_encode( $description ) );
+			$value .= sprintf( ';desc="%s"', str_replace( array( '\\', '"' ), '', substr( $description, 0, 100 ) ) );
 		}
 		if ( isset( $duration ) ) {
 			if ( $duration < 0 ) {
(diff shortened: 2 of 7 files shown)


westonruter:
@pbakaus Are there other such validation errors that come to mind which would leave a site worse off than sanitizing them from the response?

westonruter:
As of the current state of this PR the following is how a theme could on its own opt-in to allow illegal CSS properties (such as in keyframe animations):

```php
add_filter( 'amp_content_sanitizers', function( $sanitizer_classes ) {
	foreach ( array_keys( $sanitizer_classes ) as $sanitizer_class ) {
		$original_callback = null;
		if ( isset( $sanitizer_classes[ $sanitizer_class ]['validation_error_callback'] ) ) {
			$original_callback = $sanitizer_classes[ $sanitizer_class ]['validation_error_callback'];
		}

		$sanitizer_classes[ $sanitizer_class ]['validation_error_callback'] = function( $validation_error ) use ( $original_callback ) {
			$retval = call_user_func( $original_callback, $validation_error );
			if ( isset( $validation_error['code'] ) && 'illegal_css_property' === $validation_error['code'] ) {
				return false;
			}
			return $retval;
		};
	}
	return $sanitizer_classes;
}, 20 );
```

Naturally this could be made more user friendly, such as with a filter.

westonruter:
Ah, rather than letting the validation callback suppress sanitization of the error, the callback could instead be used to determine whether or not the validation error is “critical”, and when the first such error is encountered, switch to serving non-AMP.

westonruter:
Commits cherry-picked into #1093.

Issue #1062, opened 2018-04-06
Width attribute on table col elements causes validation errors


westonruter:
When `post_content` contains HTML such as:

```html
<table class="auto-style1 sws_custom_table" style="width: 100%; display: table; background-color: #ffffff; border: 2px solid #eaeaea;" border="0" width="630" cellspacing="0"><colgroup> <col style="mso-width-source: userset; mso-width-alt: 8988; width: 190pt;" width="253" /> <col style="mso-width-source: userset; mso-width-alt: 6826; width: 144pt;" width="192" /> <col style="mso-width-source: userset; mso-width-alt: 8135; width: 172pt;" width="229" /></colgroup>
...</table>
```

This is getting converted to AMP as:

```html
<table class="auto-style1 sws_custom_table amp-wp-inline-976a106f7543a9a2e8e3c8a97560be03" width="630"><colgroup><col width="253" class="amp-wp-inline-16106abec734b745f40158cc486aa0fe"/><col width="192" class="amp-wp-inline-91e2d2d529220418b62e99acb85f0928"/><col width="229" class="amp-wp-inline-43ca072c586bb2eaad4c6de773864f95"/></colgroup>
...</table>
```

This results in a validation error:

> DISALLOWED_HTML: The attribute 'width' may not appear in tag 'col'.

_First of all, maybe it is a bug in AMP that this attribute is not allowed._ We should double check that first.

Otherwise, if it is really not allowed, then whitelist sanitizer may be incorrectly merging the width layout attribute among the attributes allowed on an element: 

https://github.com/Automattic/amp-wp/blob/4e1cd227b10f2fcd6add44a24b078158aad41e53/includes/sanitizers/class-amp-allowed-tags-generated.php#L10573

See here:

https://github.com/Automattic/amp-wp/blob/4e1cd227b10f2fcd6add44a24b078158aad41e53/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L1468

`layout_allowed_attributes` is not globally-allowed it seems. Maybe that condition needs to be removed.

As reported on https://wordpress.org/support/topic/not-a-valid-amp-page-7/

westonruter:
I can see that the problem here goes beyond just `col` element. The sanitizer is also not stripping `width` attribute from elements that don't even allow in regular HTML, for example:

```html
<span width="123">not right</span>
```

The sanitizer erroneously lets the `width` persist.

westonruter:
[...]

Issue #494, opened 2016-09-30
Aligned images with inline width larger than max get displayed off-center


mjangda:
Regression from #470 

Swapping the inline style from `width` to `max-width` fixes the issue.

Reproducible with theme unit test data in the "Markup: Image Alignment" post.

PR #1064, opened 2018-04-07
[WIP] Convert width attribute in col tags to equivalent CSS rule

amedina:
This addresses issue #1062.


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -183,6 +183,15 @@ public function sanitize() {
 			$elements[] = $element;
 		}
 
+		// If 'width' attribute is present for 'col' tag, convert to proper CSS rule.
+		foreach ( $this->dom->getElementsByTagName( 'col' ) as $col ) {
+			$width_attr = $col->getAttribute( 'width' );
+			if ( ! empty( $width_attr ) ) {
+				$col->setAttribute( 'style', $width_attr . 'px' );
```
westonruter: I think this should also account for percentage widths. Also, if it has the `*` syntax then I think it should be skipped so that the whitelist sanitizer will then report the attribute as a violation.

westonruter: Also, I see there is a CSS syntax error here where it needs to be `width: {$width_attr}px`. Because there is no `width` property name this is causing a CSS syntax error and resulting in the lack of a class name being added in the unit test below.

westonruter: Also, this should account for whether an `style` attribute already exists, and if so, append to the existing one.

westonruter: Lastly, I wonder how this is going to play nicely with the auto-conversion of `width` into `max-width` (introduced in v0.4). This is seen in `0.7` branch:

https://github.com/Automattic/amp-wp/blob/5ba9f5bbf21ea0dfdf44d08899940c3e7d228afa/includes/sanitizers/class-amp-style-sanitizer.php#L54

And here in `develop` branch:

https://github.com/Automattic/amp-wp/blob/40d1945778c9917ed258e74595d504ad31719d39/includes/sanitizers/class-amp-style-sanitizer.php#L792-L801

This was introduced in https://github.com/Automattic/amp-wp/issues/494 via https://github.com/Automattic/amp-wp/pull/495. Nevertheless, in #610 it was suggested to be removed:

> Notably, the current AMP spec does not […] specify that width should be replaced by max-width, thus these rules are no longer enforced.

So we should investigate why that conversion was introduced in the first place, and potentially come up with a better fix.

tests/test-amp-style-sanitizer.php
```diff
@@ -120,6 +120,12 @@ public function get_body_style_attribute_data() {
 					'div > span { font-weight:bold; font-style: italic; }',
 				),
 			),
+
+			'col_with_width_attribute' => array(
+				'<table><colgroup><col width="253"/></colgroup></table>',
+				'<table><colgroup><col></colgroup></table>',
```
westonruter: There is an unexpected lack of an `class` attribute here.

tests/test-amp-style-sanitizer.php
```diff
@@ -120,6 +120,12 @@ public function get_body_style_attribute_data() {
...
+				array(),
```
westonruter: The `width` should have been extracted to a stylesheet here.

tests/test-amp-style-sanitizer.php
```diff
@@ -120,6 +120,12 @@ public function get_body_style_attribute_data() {
...
+			),
```
westonruter: Please add a test for when a `col` has an existing `style` attribute to make sure its properties don't get clobbered.

westonruter: Also add a test for percentage-based `width` and for width with `*` syntax.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -183,6 +183,17 @@ public function sanitize() {
 			$elements[] = $element;
 		}
 
+		// If 'width' attribute is present for 'col' tag, convert to proper CSS rule.
+		foreach ( $this->dom->getElementsByTagName( 'col' ) as $col ) {
+			$width_attr = $col->getAttribute( 'width' );
+			if ( ! empty( $width_attr ) && ! strpos( $width_attr, '*' ) ) {
+				strpos( $width_attr, '%' )
+					? $col->setAttribute( 'style', 'width: ' . $width_attr )
+					: $col->setAttribute( 'style', 'width: ' . $width_attr . 'px' );
```
westonruter: Using a ternary expression as a control statement is usually discouraged. I recommend doing something like this:

```php
$width_style = "width: $width_attr";
if ( is_numeric( $width_attr ) ) {
    $width_attr .= 'px';
}
$col->setAttribute( 'style', $width_style ); // @todo Still need to be wary of there being an existing style attr.
```

amedina: Nice. Will do.

includes/class-amp-theme-support.php
```diff
@@ -220,6 +220,14 @@ public static function add_hooks() {
...
 		remove_action( 'wp_head', 'wp_oembed_add_host_js' );
 
+		// Prevent MediaElement.js scripts/styles from being enqueued.
+		add_filter( 'wp_video_shortcode_library', function() {
+			return 'amp';
+		} );
+		add_filter( 'wp_audio_shortcode_library', function() {
+			return 'amp';
+		} );
+
```
westonruter: This change is here because I haven't yet merged #106 into `develop` from `0.7`.

westonruter:
@mehigh Could we have your insights on `width` vs `max-width` and whether automatically converting them (#495) it was the best solution to the problem reported in #494 (“Aligned images with inline width larger than max get displayed off-center”)? In #610 it was suggested to be removed:

> Notably, the current AMP spec does not […] specify that width should be replaced by max-width, thus these rules are no longer enforced.

It feels like it fixes a symptom but isn't the right solution to the underlying problem. This PR includes the reversion of the `width`/`max-width` change, but it should include a fix to the underlying problem then.

tests/test-amp-style-sanitizer.php
```diff
@@ -156,6 +148,28 @@ public function get_body_style_attribute_data() {
...
+				array(
+					':root:not(#_):not(#_):not(#_):not(#_):not(#_) .amp-wp-cd7753e{width:50%;}',
+				),
+			),
+
+			'col_with_star_width_attribute' => array(
+				'<table><colgroup><col width="0*"/></colgroup></table>',
+				'<table><colgroup><col width="0*"></colgroup></table>',
+				array(),
+			),
```
westonruter: Let's add one more test with a `col` that has `Let's add a `width=50 style="background-color: red; width: 60px;"`. This will ensure that (1) the inline style overrides any `width` attribute, and (2) the new `width` style is correctly merged with the existing styles. In particular, this shoes that `width:50px;` should be _prepended_ and not _appended_ because we need the `width:60px` to “win” over `width:50px`.

tests/test-amp-style-sanitizer.php
```diff
@@ -156,6 +148,36 @@ public function get_body_style_attribute_data() {
...
+
+			'col_with_width_attribute_and_existing_style' => array(
+				'<table><colgroup><col width="50" style="background-color: red; width: 60px"/></colgroup></table>',
+				'<table><colgroup><col class="amp-wp-c8aa9e9"></colgroup></table>',
+				array(
+					':root:not(#_):not(#_):not(#_):not(#_):not(#_) .amp-wp-c8aa9e9{width:50px;width:60px;background-color:red;}',
```
westonruter: Interesting the behavior of the CSS parser here. It groups properties with the same name when serializing, but it preserves the order. Given a source stylesheet:

```css
#bard {
	width: 1%;
	color: red;
	width: 123px;
	color: green;
	width: 456%;
	color: blue;
	width: 789em;
}
```

This gets serialized out as:

```css
.bard{width:1%;width:123px;width:456%;width:789em;color:red;color:green;color:blue;}
```

So everything looks in in order here.

mehigh:
@westonruter sorry I didn't get to this in a decent amount of time.
Instead of replacing the width by max-width, a solution used almost everywhere on the internets is to introduce a max-width: 100%; on the images. That way they would 'stay put' regardless on whether the width attribute is larger than the container or not. 
But there are edge cases, and nuances, as with any solutions.

PR #1066, opened 2018-04-10
Detect when ob_flush() is called prematurely in the course of output-buffering document for AMP theme support

westonruter:
At the moment this is a workaround to detect when the output buffer passed into `AMP_Theme_Support::finish_output_buffering()` is truncated due to a WordPress install doing `ob_flush()` prematurely. For example, on Pressable it appears that some code such as the following is running:

```php
add_action( 'get_footer', function() {
	ob_flush();
} );
```

This breaks the output buffer started in `AMP_Theme_Support::start_output_buffering()` at the `wp` action. The output buffer needs to remain intact until the `shutdown` action when `AMP_Theme_Support::finish_output_buffering()` then passes the buffer into `AMP_Theme_Support::prepare_response()`. Note that this happens at `shutdown` with a priority of 0 because `wp_ob_end_flush_all()` runs at `shutdown` priority 1.

We should try to find a better solution to ensure the integrity of the output buffer.

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1034,6 +1034,15 @@ class AMP_Theme_Support {
 			return $response;
 		}
 
+		// Account for case where ob_flush() was called prematurely.
+		if ( false === strpos( $response, '<html' ) ) {
+			$error = sprintf(
+				'<div style="color:red; background: white; padding: 0.5em; position: fixed; z-index: 100000; bottom: 0; border: dashed 1px red;">%s</div>',
+				wp_kses_post( __( '<strong>AMP Plugin Error</strong>: It appears that your WordPress install prematurely flushed the output buffer. You will need to disable AMP theme support until that is fixed.', 'amp' ) )
+			);
+			return $error . $response;
+		}
+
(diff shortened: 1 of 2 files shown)


seanosh:
Thanks for the report @westonruter. The problem isn't specific to Pressable per se, but caused by a dependency of a plugin that is 'must use' in the environment you're seeing this behavior in. Now that we've tracked it down, we'll get it fixed.

westonruter:
@seanosh Excellent! What is the code that is causing the issue? It is something that may need to be accounted for in other hosting environments?

seanosh:
@westonruter I don't know that it applies to other environments unless they too are adding "Powered by WordPress.com" to their footers :)

westonruter:
@seanosh good to know 😄Please let me know when the fix is deployed and we'll test. 

@kienstra in the mean time, the changes in this PR would be should helpful for catching this problem for any other WordPress installs that may do something similar. Please review.

kienstra:
**Thanks, Reviewing Now**

Hi @westonruter,
Thanks for this PR. I'm reviewing it now.

westonruter:
> Now that we've tracked it down, we'll get it fixed.

@seanosh Do you have an ETA on when the will be deployed?

PR #1070, opened 2018-04-11
Fix domain check for WP.com hosted files

mjangda:
We want the querystring extraction to apply to either `*.wp.com` or `*.files.wordpress.com` domains. Previously, this would only apply for `*.wp.com` due to the incorrect conditional.

This change is in the `wpcom-helper.php` file and only applies to WP.com-hosted sites.

props @vortfu


westonruter:
I've rebased for 0.7.

PR #1071, opened 2018-04-12
Fix libxml issues with parsing character encoding

### LINKED ISSUE #1067: In paired mode, sanitization replaces certain punctuation with "â"
I'm in the early stages of updating my theme to add AMP support in "paired mode" per 0.7. Something is replacing certain punctuation marks (curly quotes, curly apostrophes, and em-dashes at least) with "â".

I suspect this may relate to the quote issues mentioned in #855, however my blog is already fully UTF-8 so I would think no encoding conversions would be necessary.

It looks like this is happening only to special characters that are part of the raw post content or the template itself. In cases where the raw post content has plain (straight) quotes and apostrophes, they are ultimately rendered correctly (with their "curly" versions, as presumably replaced by `wptexturize`).

If I remove AMP support from my theme and just use the legacy templates, these characters render fine. These characters also display fine on my non-AMP endpoints. So I think it must be caused by the full-HTML sanitization being done per #888.

Details:
* My database and all of its tables are in `utf8`, and my `config.php` also has `DB_CHARSET` set to `utf8`.
* In all cases, the rendered page has `<meta charset="UTF-8">` and the `Content-Type` header also has `charset=UTF-8`.
* The PHP version is 5.6.32.
* The version of this plugin I'm using is commit `5ba9f5b` from the 0.7 branch.

douglyuckling:
Fixes #1067.

### DIFF
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -138,9 +138,9 @@ class AMP_DOM_Utils {
 		/*
 		 * Remove pre-HTML5-style encoding declaration if added above.
 		 */
-		$metaHttpEquivElement = $dom->getElementById("meta-http-equiv");
-		if ( $metaHttpEquivElement ) {
-			$metaHttpEquivElement->parentNode->removeChild($metaHttpEquivElement);
+		$meta_http_equiv_element = $dom->getElementById( 'meta-http-equiv' );
+		if ( $meta_http_equiv_element ) {
+			$meta_http_equiv_element->parentNode->removeChild( $meta_http_equiv_element );
 		}
 
 		return $dom;
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -109,6 +109,18 @@ class AMP_DOM_Utils {
 		}
 
 		/*
+		 * Add a pre-HTML5-style declaration of the encoding since libxml<2.8 doesn't recognize
+		 * HTML5's meta charset. See <https://bugzilla.gnome.org/show_bug.cgi?id=655218>.
+		 */
+		if ( version_compare( LIBXML_DOTTED_VERSION, '2.8', '<' ) ) {
+			$document = preg_replace(
+				'#<meta[^>]+charset="([^"]+)"#i',
+				'<meta http-equiv="Content-Type" content="text/html; charset=$1" id="meta-http-equiv">$0',
+				$document
+			);
+		}
+
+		/*
 		 * Wrap in dummy tags, since XML needs one parent node.
 		 * It also makes it easier to loop through nodes.
 		 * We can later use this to extract our nodes.
@@ -121,6 +133,14 @@ class AMP_DOM_Utils {
 
 		if ( ! $result ) {
 			return false;
+		}
+
+		/*
+		 * Remove pre-HTML5-style encoding declaration if added above.
+		 */
+		$metaHttpEquivElement = $dom->getElementById("meta-http-equiv");
+		if ( $metaHttpEquivElement ) {
+			$metaHttpEquivElement->parentNode->removeChild($metaHttpEquivElement);
 		}
 
 		return $dom;
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -90,13 +90,15 @@ class AMP_DOM_Utils {
 			$document
 		);
 
-		/*
-		 * Replace noscript elements with placeholders since libxml<2.8 can parse them incorrectly.
-		 * When appearing in the head element, a noscript can cause the head to close prematurely
-		 * and the noscript gets moved to the body and anything after it which was in the head.
-		 * See <https://stackoverflow.com/questions/39013102/why-does-noscript-move-into-body-tag-instead-of-head-tag>.
-		 */
+		// Deal with bugs in older versions of libxml.
+		$added_back_compat_meta_content_type = false;
 		if ( version_compare( LIBXML_DOTTED_VERSION, '2.8', '<' ) ) {
+			/*
+			 * Replace noscript elements with placeholders since libxml<2.8 can parse them incorrectly.
+			 * When appearing in the head element, a noscript can cause the head to close prematurely
+			 * and the noscript gets moved to the body and anything after it which was in the head.
+			 * See <https://stackoverflow.com/questions/39013102/why-does-noscript-move-into-body-tag-instead-of-head-tag>.
+			 */
 			$document = preg_replace_callback(
 				'#<noscript[^>]*>.*?</noscript>#si',
 				function( $matches ) {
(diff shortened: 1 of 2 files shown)


douglyuckling:
By the way, the [libxml bug](https://bugzilla.gnome.org/show_bug.cgi?id=655218) doesn't clearly state what version it was fixed in, but the commit ID is in the comments there so I used `git tag --contains` with that commit ID to determine that the fix was first included in libxml 2.8.0.

includes/utils/class-amp-dom-utils.php
```diff
@@ -123,6 +135,14 @@ function( $matches ) {
 			return false;
 		}
 
+		/*
+		 * Remove pre-HTML5-style encoding declaration if added above.
+		 */
+		$meta_http_equiv_element = $dom->getElementById( 'meta-http-equiv' );
```
westonruter: This could be made a tiny more performant by first checking `version_compare( LIBXML_DOTTED_VERSION, '2.8', '<' )`. The result of the first call could be put inside a variable to re-use here.

includes/utils/class-amp-dom-utils.php
```diff
@@ -108,6 +108,18 @@ function( $matches ) {
 			);
 		}
 
+		/*
+		 * Add a pre-HTML5-style declaration of the encoding since libxml<2.8 doesn't recognize
+		 * HTML5's meta charset. See <https://bugzilla.gnome.org/show_bug.cgi?id=655218>.
+		 */
+		if ( version_compare( LIBXML_DOTTED_VERSION, '2.8', '<' ) ) {
+			$document = preg_replace(
+				'#<meta[^>]+charset="([^"]+)"#i',
```
westonruter: It looks like this may be missing the trailing `>` so the replacement would seem to result in doubled `>>` being added?

includes/utils/class-amp-dom-utils.php
```diff
@@ -108,6 +108,18 @@ function( $matches ) {
...
+				'<meta http-equiv="Content-Type" content="text/html; charset=$1" id="meta-http-equiv">$0',
```
westonruter: Why the `$0`?

includes/utils/class-amp-dom-utils.php
westonruter: In fact, you can see there is another such check for the libxml version on line 99. So all of these checks could re-use this same var, e.g.:

```php
$is_old_libxml = version_compare( LIBXML_DOTTED_VERSION, '2.8', '<' );
```

douglyuckling: Okay, wrapping this in a version check makes sense to me, as does introducing a variable. I'd be inclined to name the variable something more specific, like `$is_pre_2_8_libxml`, but if you have a specific preference for the name I'll go with that.

includes/utils/class-amp-dom-utils.php
douglyuckling: Because I'm not replacing the original `meta charset` tag, just inserting the new `meta` tag before it. (That's also why I don't include the `>` in he regex, per your previous comment.)

I see that at a glance it looks like the aim is to replace the whole original meta tag, so I should probably add a comment to clarify what's going on.

includes/utils/class-amp-dom-utils.php
westonruter: Either is fine with me. Since we're only testing for one version of libxml currently, the a simple “old“ name should suffice.

includes/utils/class-amp-dom-utils.php
westonruter: Thanks. I understand now. In that case you might want to use a lookahead match instead, like so:

```php
$document = preg_replace(
    '#(?=<meta[^>]+charset="([^"]+)")#i',
    '<meta http-equiv="Content-Type" content="text/html; charset=$1" id="meta-http-equiv">',
    $document
);
```

Not really a big deal.

includes/utils/class-amp-dom-utils.php
```diff
@@ -108,6 +108,18 @@ function( $matches ) {
...
+				$document
+			);
```
westonruter: Might want to add a limit of `1`.

includes/utils/class-amp-dom-utils.php
westonruter: What about the case where there is `<meta charset=utf-8>` with no quote marks? Sure this is probably unlikely.

douglyuckling:
Regarding adding a unit test, that will require actually setting up a proper PHP development environment. :-P (So far I've gotten away with just FTPing my hacked PHP files up to a dev clone of the blog on our hosting provider.)

I'll see about setting up my dev environment tonight. Is there a guide for dev environment setup you can recommend? The instructions in the [theme handbook](https://developer.wordpress.org/themes/getting-started/setting-up-a-development-environment/) seem like a good start, but they don't cover the `composer` and `wp` commands I see mentioned in `contributing.md`.

Sorry, in my day job I do JVM-based web development, but until now I haven't really touched PHP in well over a decade. Speaking of my day job, I should probably get back to that...

westonruter:
No worries. I can add a unit test instead if you like.

includes/utils/class-amp-dom-utils.php
douglyuckling: Oh, yeah, I should have thought of that. My only defense is that it was after midnight. :-P

includes/utils/class-amp-dom-utils.php
douglyuckling: Mm, yeah. I'll include a test case for that and for single quotes as well. I should also account for possible whitespace around the `=`.

douglyuckling:
If you're willing, that would be much appreciated.

It occurs to me there should also maybe be a check after the call to `loadHTML` to make sure that `$dom->encoding` isn't empty. If it is, there should maybe be a warning logged somewhere that the encoding couldn't be detected and therefore special characters will likely likely be corrupted in the output. That could be a huge time saver for anyone trying to track down encoding issues in the future (e.g. a different libxml bug, or something the above regex doesn't account for).

westonruter:
For unit testing, the easiest way to do that is to use the [VVV](https://github.com/Varying-Vagrant-Vagrants/VVV) environment. Clone the plugin into the `wordpress-develop` `src` install. Then when SSH'ed into the vagrant box, you can run `phpunit`. A bit of that is documented in:

https://github.com/Automattic/amp-wp/blob/develop/contributing.md#amp-contributing-guide
https://github.com/Automattic/amp-wp/blob/develop/contributing.md#phpunit-testing

It should be fleshed out a bit more.

westonruter:
I'll make the changes.

westonruter:
> If it is, there should maybe be a warning logged somewhere that the encoding couldn't be detected and therefore special characters will likely likely be corrupted in the output. That could be a huge time saver for anyone trying to track down encoding issues in the future (e.g. a different libxml bug, or something the above regex doesn't account for).

It's a good concern but I think that it shouldn't ever happen in our case here because we explicitly are inserting a `meta charset` if one is not originally present:

https://github.com/Automattic/amp-wp/blob/40d1945778c9917ed258e74595d504ad31719d39/includes/class-amp-theme-support.php#L1009-L1021

westonruter:
@douglyuckling Would you please test https://github.com/Automattic/amp-wp/pull/1071/commits/aad19e1b097d269fca76dbe91ca4db4dcfb5c6d9 to make sure it works on your environment?

douglyuckling:
Yep, works like a charm! Thanks!

PR #1075, opened 2018-04-15
Ensure layout attributes are only allowed on supporting elements

### LINKED ISSUE #1062: Width attribute on table col elements causes validation errors

westonruter:
The whitelist sanitizer is currently incorrectly allowing layout attributes (`layout`, `width`, `height`, etc) on all elements. This PR ensures that the list of layout attributes are only merged into the attribute list for tag specs that actually declare they have `amp_layout`. Additionally, the value of the `layout` attribute itself is validated against the tag spec's `supported_layouts`.

This is a companion PR with #1064 to fix #1062.

### DIFF
--- a/bin/amphtml-update.py
+++ b/bin/amphtml-update.py
@@ -469,6 +469,15 @@ def GetTagRules(tag_spec):
 	if tag_spec.HasField('unique_warning'):
 		tag_rules['unique_warning'] = tag_spec.unique_warning
 
+	if tag_spec.HasField('amp_layout'):
+		amp_layout = {}
+		for field in tag_spec.amp_layout.ListFields():
+			if 'supported_layouts' == field[0].name:
+				amp_layout['supported_layouts'] = [ val for val in field[1] ]
+			else:
+				amp_layout[ field[0].name ] = field[1]
+		tag_rules['amp_layout'] = amp_layout
+
 	logging.info('... done')
 	return tag_rules
 
--- a/includes/sanitizers/class-amp-allowed-tags-generated.php
+++ b/includes/sanitizers/class-amp-allowed-tags-generated.php
@@ -105,6 +105,14 @@ class AMP_Allowed_Tags_Generated {
 					),
 				),
 				'tag_spec' => array(
+					'amp_layout' => array(
+						'supported_layouts' => array(
+							6,
+							2,
+							7,
+							4,
+						),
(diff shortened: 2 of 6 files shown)


kienstra:
**Reviewing Now**

Hi @westonruter,
Thanks for this PR, fixing the sanitizer's handling of `layout` attributes. I'm reviewing this now.

PR #1077, opened 2018-04-17
Fix handling of font stylesheets with non-HTTPS scheme or scheme-less URLs

### LINKED ISSUE #1072: Sanitizer removes fonts that omit the scheme from the URL
When my theme enqueues a font like this:
```php
wp_enqueue_style( 'hemingway_googleFonts', '//fonts.googleapis.com/css?family=Raleway' );
```
the sanitizer strips it out. Adding the `https:` scheme to the URL fixes the problem, so I think the sanitizer maybe needs to be a little more tolerant with the URL.

(If this behavior is actually intentional, then I think the error message could use improvement. The error message I got was misleading: "Skipped stylesheet which does not have recognized CSS file extension (//fonts.googleapis.com/css).")

westonruter:
* When someone enqueues a font CDN stylesheet without the required HTTPS scheme, automatically supply it before checking against the regex and ensure the link has the HTTPS URL set during processing.
* Improve URL checking by first ensuring the host name matches.
* Ignore differences for URL scheme (even scheme-less URLs) when comparing URLs since the scheme is irrelevant.

Fixes #1072

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -311,8 +311,28 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			$url = $this->base_url . $url;
 		}
 
-		// Strip query and fragment from URL.
-		$url = preg_replace( ':[\?#].*$:', '', $url );
+		$remove_url_scheme = function( $schemed_url ) {
+			return preg_replace( '#^\w+:(?=//)#', '', $schemed_url );
+		};
+
+		// Strip URL scheme, query, and fragment.
+		$url = $remove_url_scheme( preg_replace( ':[\?#].*$:', '', $url ) );
+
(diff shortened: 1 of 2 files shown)


kienstra:
Hi @westonruter, thanks for this PR. I'm reviewing it now.

kienstra:
**Approved**

Hi @westonruter,
This pull request looks good, and it worked locally.

When enqueuing the Google Font file that @douglyuckling listed in #1072, the `<link>` appeared as expected, using both "Paired mode" (with a child theme of [twentyten](https://wordpress.org/themes/twentyten/)) and "Native AMP" (with the [ampconf theme](github.com/xwp/ampconf/)):
```php
wp_enqueue_style( 'hemingway_googleFonts', '//fonts.googleapis.com/css?family=Raleway' );
```

Also, as expected, the `<link>` appeared with `https`, even if I changed the enqueuing URL to `http://fonts...`.

douglyuckling:
Yep, works for me, too. Thanks!

PR #1092, opened 2018-04-20
Ensure all sanitizers run when building featured image

### LINKED ISSUE #1090: The tag 'amp-anim' requires including the 'amp-anim' extension JavaScript.
Google Search Console is returning this issue for makup on our site. 

> The tag 'amp-anim' requires including the 'amp-anim' extension JavaScript.

Worth noting that we're using code on branch v0.7 as per your recommendation in ticket #1076 

screen shot 2018-04-20 at 1 46 55 pm

westonruter:
This is a regression introduced by #882 in c62dad662229f71e3ec7c6407c3fd834eb37e159.

This ensures that the `amp-anim` script, for example, will be assured to be included. It also ensures that if a plugin adds other markup to a featured image via the `post_thumbnail_html` filter that it will be properly sanitized.

Fixes #1090.

### DIFF
--- a/includes/templates/class-amp-post-template.php
+++ b/includes/templates/class-amp-post-template.php
@@ -349,7 +349,7 @@ class AMP_Post_Template {
 
 		list( $sanitized_html, $featured_scripts, $featured_styles ) = AMP_Content_Sanitizer::sanitize(
 			$featured_html,
-			array( 'AMP_Img_Sanitizer' => array() ),
+			amp_get_content_sanitizers( $this->post ),
 			array(
 				'content_max_width' => $this->get( 'content_max_width' ),
 			)


includes/templates/class-amp-post-template.php
```diff
@@ -349,7 +349,7 @@ private function build_post_featured_image() {
 
 		list( $sanitized_html, $featured_scripts, $featured_styles ) = AMP_Content_Sanitizer::sanitize(
 			$featured_html,
-			array( 'AMP_Img_Sanitizer' => array() ),
+			amp_get_content_sanitizers( $post_id ),
```
kienstra: Should this pass a `WP_Post` instead of `$post_id`? It looks like `amp_get_content_sanitizers()` [expects a WP_Post](https://github.com/Automattic/amp-wp/blob/develop/includes/amp-helper-functions.php#L465) as the argument.

kienstra: Should this pass a `WP_Post` instead of `$post_id`? It looks like `amp_get_content_sanitizers()` [expects a WP_Post](https://github.com/Automattic/amp-wp/blob/develop/includes/amp-helper-functions.php#L465) as the argument.

westonruter: @kienstra Good catch. Fixed in 5045563

kienstra:
**Approved**

Hi @westonruter,
Thanks, this looks good.

PR #1093, opened 2018-04-21
Refactor storage of validation errors to use taxonomy terms

### LINKED ISSUE #1003: Add ability to acknowledge and suppress/ignore specific validation errors

### LINKED ISSUE #1031: Capture script tag contents in AMP validator errors
The AMP validator utilities should capture script tag contents when it is less than 1KB (or some threshold) so that we can gather data about the scripts which are removed, so we can begin to identify the most common scripts that should get their own dedicated post-processor to convert to AMP (e.g. Google Analytics, AdSense, etc).

### LINKED ISSUE #1087: Dynamically allow AMP responses based on validation status
The sanitizers are designed to force a page served as AMP to be valid, without any illegal markup or missing markup which would cause a document to be invalid and not able to be served from an AMP cache. While 1.0-alpha now has the ability to inform users of the validation errors that are occurring on a site, it doesn't currently use this information to help make decisions for the user. The sanitizer may very well remove something that completely breaks the page for the user. So what is needed is a way to detect when such a critical (deal breaker) validation error occurs, and when it does, prevent AMP from being served from a given URL. This is essentially building upon the enable/disable toggle that exists currently in the paired mode: there could in essence be an in-between behavior like “enable when possible” (which could make the whole AMP enable/disable toggle obsolete). This ties in with some ideas from #934 (Add theme support scenarios: allow paired mode to serve some content exclusively in AMP): even though a site may be Native AMP normally, when such a deal breaker validation error occurs, AMP should be disabled on a given response.

What determines what a critical validation error is? I think in general any validation error should by default be considered critical. Removing something from the page should never be done without the user approving it. So this is where #1003 comes in with the ability to acknowledge validation errors that occur on a site. If a user has acknowledged a given validation error by saying it should be ignored, then a response that includes such a validation error should continue to be served as AMP. Otherwise, by default if there is a validation error detected in a response and this validation error has not knowingly been ignored by the user, then AMP should be prevented in the response. (Likewise, if a user acknowledges a validation error as being critical, then AMP would be prevented. Acknowledging a validation error marks it as “read”.)

The default behavior for an unacknowledged validation error should be configurable so that instead of defaulting to unacknowledged errors being treated as critical, they can instead be treated as ignored. This could be accomplished by a filter.

The handling of a critical validation error would look something like this:

1. Theme enables `amp` support for a given request.
2. User makes first request to a given URL.
3. Sanitizer detects validation errors.
4. Validation errors are looked up to see if they have been acknowledged and ignored. A validation callback could also be provided to let a site dynamically judge whether something is critical, via #1063.
5. If all validation errors have been ignored, then the AMP response finalizes.
6. Otherwise, if there are critical validation errors then the given URL needs to be flagged as excluded from AMP and instead of the output-buffered response finalizing, the response needs to redirect to the non-AMP version of the page. (Special consideration is needed here for Native AMP sites, such as in #934.)

This whole process could be done en masse upon first activating the plugin through a sort of onboarding/wizard/setup flow when activating the plugin for the first time or switching to a new theme, per https://github.com/Automattic/amp-wp/issues/1003#issuecomment-379833095:

> 1. I activate the AMP plugin. 
> 2. The plugin sees that I don't have AMP theme support.
> 3. The plugin asks if you would like to try AMP theme support on your theme.
> 4. It runs validation report and lets you browse the site with AMP active just for yourself.
> 5. Then at the end you decide whether you want to enable AMP for the theme, or re-use the legacy post templates, or whether you want to use paired mode to selectively opt-in to certain templates being available in AMP.
> 6. AMP theme support is enabled by the plugin as desired.

westonruter:
This will include the plumbing to be able to achieve dynamically allowing AMP responses based on validation status (#1087). It should include the ability to acknowledge a certain validation error as being ignored (#1003, #1063).

- [x] Add “term statuses” (via `term_group`) for acknowledged and ignored, with ability for user to change them.
- [x] Add ability to acknowledge/ignore errors from the Invalid URL post edit screen.
- [x] Move `sources` to be stored in the `amp_invalid_url` posts themselves. This addresses issue with validation errors having dynamic sources, such as `post_id`.
- [x] Capture inline `script` contents in validation errors. Fixes #1031.
- [x] Merge `enqueued_script` validation error into `removed_element` validation error, combining `sources` into latter.
- [x] Detect when critical validation errors are present after post-processor runs, and refresh with non-AMP version. 
- [x] Refactor prototype code into better-organized class(es).
- [x] Accepted validation errors for CSS need to be applied when pulling parsed CSS from cache.
- [x] When serving dirty AMP, ensure `document.write` is disabled for sake of `amp-live-list`.
- [ ] Rewrite unit tests.

Fixes #1003.
Fixes #1087.


westonruter:
Current view of the new validation errors, with bulk actions for acknowledge/ignore and inline actions to do the same:

image

The URLs column indicates the number of URLs that have the error, which links to a new “Invalid Pages” screen (formerly Validation Status) with the URLs listed:

image

westonruter:
There is also a new admin menu item for validation errors which is separate from the Invalid Pages screen:

image

ThierryA:
This is awesome @westonruter, such a great step forward! First round feedback:
1. It would be good to have a way to go back to the "blocking status". So for example a user might mark an issue as "Acknowledged" or "Ignored" but later realize there is still an issue in which case they might revert the status. Right now that status is called "New" but the naming could change.
2. What is the difference between "Acknowledged" and "Ignored" from a functional perspective?
3. 4b34bdc breaks bulk actions, not only preventing the users from bulk deleting but also from doing any other bulk actions when validation errors still have associated invalid URLs.

westonruter:
@ThierryA 
> 1. It would be good to have a way to go back to the "blocking status". So for example a user might mark an issue as "Acknowledged" or "Ignored" but later realize there is still an issue in which case they might revert the status. Right now that status is called "New" but the naming could change.
> 3. What is the difference between "Acknowledged" and "Ignored" from a functional perspective?

Actually, “acknowledge” has the same effect as “new”: both are blockers. The difference is that “new” is like unread. The idea is that maybe a user would like to default the behavior of new validation errors to be ignored by default instead of acknowledged by default. I tried to initially explain what acknowledging means in `aria-label` attributes:

* **Acknowledge:** Acknowledging an error marks it as read. AMP validation errors prevent a URL from being served as AMP.
* **Ignore:** Ignoring an error prevents it from blocking a URL from being served as AMP.

I'm not totally sold on the naming of “acknowledge” and “ignore” but I couldn't think of anything better.

> It would be good to have a way to preview a page with validation error in AMP mode which would help to debug the issues reported.

Agreed. I've got this jotted down as a todo item.

westonruter:
> 4b34bdc breaks bulk actions, not only preventing the users from bulk deleting but also from doing any other bulk actions when validation errors still have associated invalid URLs.

You're right. Unfortunately [`WP_Terms_List_Table::column_cb()` uses `delete_term`](https://github.com/WordPress/wordpress-develop/blob/f9e4467b3db7d8bc83a82fded8326158309ec107/src/wp-admin/includes/class-wp-terms-list-table.php#L344) for whether to show the checkbox. Need to figure out an alternative.

westonruter:
Validation errors are now sorted by creation date, with default being in descending order, with the creation date column being sortable in ascending order by clicking the header:

image

Validation errors are also sortable by frequency (the number of URLs that have the error occurring):

image

westonruter:
Now see the error status at a glance in the list of invalid AMP URLs:

image

westonruter:
In the same way that there are inline row actions to acknowledge or ignore a given validation error on the edit terms screen, so too there are now the same action links provided when listing the validation errors on a given invalid AMP URL post:

image

westonruter:
The edit post screen for an Invalid AMP Page now displays its actual URL and as a link so you can visit the URL to review the post on the frontend. (The view URL will need to be getting an `amp_forced` query param when the auto-disable functionality is triggered due to a URL having new/acknowledged errors which block it from being served as AMP.) This screen also now features the counts for the various error types present on the URL:

image

The URL shown in the post list table is also now shortened to remove the scheme and host name:

image

DavidCramer:
@westonruter I took some time today going over the whole validation system, and it's crazy cool!

Some feedback.
I was initially confused as @ThierryA pointed out, as to the functional differences between `Acknowledged` and `Ignore`. My first thought was `ignore` tells the AMP plugin to ignore it and leave the invalid markup on the page. Where is `Acknowledged` meant that you are acknowledging it being removed. 
Now reading your explanation to @ThierryA's question on functionality, I understand the difference. I still think the confirmation message is a little confusing though.

A suggestion on terminology. It seams that Ignore and Acknowledge is similar. Perhaps `Dismiss` would be better with an option like `Dismiss` for` 1 hour`, `1 day`, `1 week`, `indefinitely`. Kinda like, saying "ye thanks, I'm still working on it, remind me later."

The confirmation messages also made me feel a little confused `Ignored 1 error. It will no longer block related URLs from being served as AMP.` to me this told me that the pages will now be AMP valid. but was confused when I checked the page and it was valid since the invalid markup was being stripped.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -290,18 +289,22 @@ public function maybe_enforce_https_src( $src, $force_https = false ) {
...
 			);
+			if ( is_bool( $result ) ) {
+				$should_remove = $result;
+			}
 		}
-		if ( empty( $this->args['disable_invalid_removal'] ) ) {
+		if ( $should_remove ) {
 			$node->parentNode->removeChild( $node );
 		}
+		return $should_remove;
```
westonruter: In canonical AMP, whenever `$should_remove` is `false`, then this needs to trigger dirty AMP mode, where `amp` attribute is removed from the `html` element.

includes/class-amp-theme-support.php
```diff
@@ -1079,16 +1082,23 @@ public static function prepare_response( $response, $args = array() ) {
...
 		self::ensure_required_markup( $dom );
 
+		if ( AMP_Validation_Utils::has_blocking_validation_errors() ) {
+			if ( amp_is_canonical() ) {
+				$dom->documentElement->removeAttribute( 'amp' );
+			} else {
+				self::redirect_canonical_amp( false );
+				return esc_html__( 'Redirecting to non-AMP version.', 'amp' );
+			}
+		}
```
westonruter: This conditional is the critical piece to the whole thing.

includes/utils/class-amp-validation-utils.php
```diff
@@ -733,6 +896,81 @@ public static function locate_sources( DOMNode $node ) {
...
+						$script_dependency
+						&&
+						$script_dependency->src
+						&&
+						// Script attribute is haystack because includes protocol and may include query args (like ver).
+						false !== strpos( $src, preg_replace( '#^https?:(?=//)#', '', $script_dependency->src ) )
+					);
+					if ( $is_matching_script ) {
+						$sources = array_merge(
+							self::$enqueued_script_sources[ $enqueued_script_handle ],
```
westonruter: The `$enqueued_script_handle` also needs to be passed to the validation error to include as the `handle`.

includes/utils/class-amp-validation-utils.php
```diff
@@ -733,6 +896,81 @@ public static function locate_sources( DOMNode $node ) {
...
 		}
+
+		$is_enqueued_link = (
+			$node instanceof DOMElement
+			&&
+			'link' === $node->nodeName
+			&&
+			preg_match( '/(?P<handle>.+)-css$/', (string) $node->getAttribute( 'id' ), $matches )
+			&&
+			isset( self::$enqueued_style_sources[ $matches['handle'] ] )
```
westonruter: Need to pass the handle to the style validation error.

westonruter: Done in 3c0d724

includes/utils/class-amp-validation-utils.php
westonruter: Done in 3c0d724

westonruter:
Adding invalid markup to Gutenberg results in a notice:

> image

Clicking Review Issues takes you to the “Invalid AMP Page” screen to review the issues with the URL:

> image

On a non-canonical site, accessing the AMP URL will result in a redirect to the non-AMP version of the page. On a canonical site, the validation errors which have been marked as accepted will be sanitized (stripped out), and all validation errors which are either new or rejected will remain as-is in the document. The `amp` attribute on the root `html` element is removed in this latter case.

When all of the validation issues have been accepted, then the URL is eligible to be served as AMP:

> image

And in the editor the validation notice is not shown since all of the errors have been accepted:

> image

The unique occurrences of validation errors can be listed irrespective of the URLs they appear on:

image

There is URL count for the number of times the error occurs on the site, and clicking on the count shows the list of pages that have that error.

includes/class-amp-response-headers.php
```diff
@@ -75,7 +72,7 @@ public static function send_header( $name, $value, $args = array() ) {
 	public static function send_server_timing( $name, $duration = null, $description = null ) {
 		$value = $name;
 		if ( isset( $description ) ) {
-			$value .= sprintf( ';desc=%s', wp_json_encode( $description ) );
+			$value .= sprintf( ';desc="%s"', str_replace( array( '\\', '"' ), '', substr( $description, 0, 100 ) ) );
```
ThierryA: Replacing `_` with space would have a more elegant result for class names.

includes/class-amp-theme-support.php
```diff
@@ -100,8 +100,13 @@ public static function init() {
 			return;
 		}
 
+		// @todo Rename to query var to indicate whether sources should be obtained. It's sources that are needing to be to be conditional.
+		AMP_Validation_Manager::init( array(
+			'locate_sources' => AMP_Validation_Manager::should_validate_response(),
+			'debug'          => isset( $_REQUEST[ AMP_Validation_Manager::DEBUG_QUERY_VAR ] ), // phpcs:ignore WordPress.CSRF.NonceVerification.NoNonceVerification
```
ThierryA: No cap check at this specific processing time on purpose?

includes/options/class-amp-options-manager.php
```diff
@@ -259,4 +260,20 @@ public static function update_analytics_options( $data ) {
...
 		return self::update_option( 'analytics', wp_unslash( $data ) );
 	}
+
+	/**
+	 * Outputs an admin notice if persistent object cache is not present.
+	 *
+	 * @return void
+	 */
+	public static function persistent_object_caching_notice() {
+		if ( ! wp_using_ext_object_cache() && 'toplevel_page_' . self::OPTION_NAME === get_current_screen()->id ) {
```
ThierryA: `'toplevel_page_' . self::OPTION_NAME` can be replaced with `get_plugin_page_hookname( self::OPTION_NAME, null )` to avoid hard coded prefix.

includes/class-amp-response-headers.php
westonruter: This logic here is specifically for ensuring the well-formedness of the `Server-Timing` header. If we want to use spaces instead of underscores then we should do that at:

https://github.com/Automattic/amp-wp/blob/9dde3226a5998cab2ba12983d63f0fa8111fb338/includes/templates/class-amp-content-sanitizer.php#L96

includes/class-amp-theme-support.php
westonruter: The `debug` flag is primarily used as a flag for whether to `remove_source_comments`. This only runs if `AMP_Validation_Manager::should_validate_response()` returns `true`, and that method does the cap check. It's used in one more place as a way to prevent validation errors from being sanitized, and here no cap check is involved, which may or may not be a good idea.

But I think this debug flag should go away entirely in favor of giving the ability to toggle whether individual validation errors should be sanitized or not in a preview interface. There should then be a separate param specifically for skipping `remove_source_comments`.

includes/options/class-amp-options-manager.php
westonruter: I tried swapping that out and it worked but it broke the unit test. Feel free to amend the PR with the change.

westonruter:
I've added the rudimentary ability to preview changes to the statuses of validation errors:

image

Clicking “Preview” takes you to a URL like:

```
https://example.com/2018/05/25/foo/?amp_validate=47cd785376&amp_validation_error_status%5Bddecaba254d1ffcd034c9ab7add5196f%5D=1
```

You can see how the URL behaves with any number of validation errors accepted (sanitized).

Again, the UI for the invalid URLs screen and validation errors screen needs to be completely redone. The UI code in place now should be considered a prototype. Ideally there would be REST API endpoints added for managing the validation error status and invalid URL posts, and the current screens for managing them should be refactored to use a JS-based interface, such as with React.

@ThierryA Given these known issues with the UI and the outstanding need to update unit tests (which will likely change greatly with changes to the UI) I suggest getting this merged sooner rather than later, and then we can follow up with additional PRs to iterate on what is included here.

westonruter:
I've updated the design of the Invalid AMP Page screen to work a lot more like a regular edit post screen, with buttons inside the normal side metabox:

image

includes/class-amp-autoloader.php
```diff
@@ -89,7 +89,9 @@ class AMP_Autoloader {
 		'AMP_DOM_Utils'                               => 'includes/utils/class-amp-dom-utils',
 		'AMP_HTML_Utils'                              => 'includes/utils/class-amp-html-utils',
 		'AMP_Image_Dimension_Extractor'               => 'includes/utils/class-amp-image-dimension-extractor',
-		'AMP_Validation_Utils'                        => 'includes/utils/class-amp-validation-utils',
+		'AMP_Validation_Manager'                      => 'includes/validation/class-amp-validation-manager',
```
kienstra: It's nice how this is now organized in the `validation/` directory.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -549,22 +607,47 @@ private function process_link_element( DOMElement $element ) {
...
+				array( 'property_whitelist', 'property_blacklist', 'stylesheet_url', 'allowed_at_rules' )
+			),
+			wp_array_slice_assoc(
+				$this->args,
+				array( 'should_locate_sources', 'parsed_cache_variant' )
+			)
 		);
 
-		$cache_key = md5( $stylesheet . serialize( $cache_impacting_options ) ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
+		$cache_key = md5( $stylesheet . wp_json_encode( $cache_impacting_options ) );
```
kienstra: It's nice how using `wp_json_encode()` removes the need to ignore a `phpcs` violation.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -0,0 +1,1219 @@
...
+	 * @param int|WP_Post $post Post of amp_invalid_url type.
+	 */
+	public static function display_invalid_url_validation_error_counts_summary( $post ) {
+		$result = array();
+		$counts = self::get_invalid_url_validation_error_counts( $post );
+		if ( $counts[ AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_NEW_STATUS ] ) {
+			$result[] = esc_html( sprintf(
+				/* translators: %s is count */
+				__( '&#x2753; New: %s', 'amp' ),
+				number_format_i18n( $counts[ AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_NEW_STATUS ] )
```
kienstra: Nice use of [number_format_i18()](https://codex.wordpress.org/Function_Reference/number_format_i18n), I didn't know about that.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -0,0 +1,1219 @@
...
+			$stored_validation_errors[] = compact( 'term_slug', 'sources' );
+		}
+
+		$post_content = wp_json_encode( $stored_validation_errors );
+		$placeholder  = 'amp_invalid_url_content_placeholder' . wp_rand();
+
+		// Guard against Kses from corrupting content by adding post_content after content_save_pre filter applies.
+		$insert_post_content = function( $post_data ) use ( $placeholder, $post_content ) {
+			$should_supply_post_content = (
+				isset( $post_data['post_content'] )
```
kienstra: Maybe this line could also check `$post_data['post_type']`, instead of the `isset( $post_data['post_type'] )` check [4 lines below](https://github.com/Automattic/amp-wp/pull/1093/files#diff-57518db0811ee918fee17d27459ed4ecR368):
```php
isset( $post_data['post_content'], $post_data['post_type'] )
```

tests/test-class-amp-validation-utils.php
```diff
@@ -1,12 +1,12 @@
 <?php
```
kienstra: It might be good to move these tests to separate files, like `test-class-amp-validation-manager.php`.

westonruter: Yes, absolutely agree. We'll follow up with new PR to add/update the tests and also to polish the UI. 👍

westonruter:
It seems there are networking connectivity issues being experienced by GitHub or Travis that are causing jobs to fail. I'm going to go ahead and merge this even though the build isn't finishing.

PR #1097, opened 2018-04-26
Ensure only remaining AMP scripts are printed when preparing response

westonruter:
This fixes a bug exposed when testing the plugin with W3 Total Cache. That plugin prints out a script at the `wp_print_scripts` action which is triggered when calling `wp_print_scripts()`. We can avoid this duplication of script printing by directly calling `wp_scripts()->do_items()` instead of `wp_print_scripts()` which is essentially a wrapper for the former. The being printed here are the AMP component scripts that were discovered during the sanitization process. As such these scripts are run _after_ sanitization has been completed, thus it is important that only the remaining AMP scripts be output here. The changes here ensure this.

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1128,9 +1128,17 @@ class AMP_Theme_Support {
 			}
 		}
 
-		// Print all scripts, some of which may have already been printed and inject into head.
+		/*
+		 * Print additional AMP component scripts which have been discovered by the sanitizers, and inject into the head.
+		 * Before printing the AMP scripts, make sure that no plugins will be manipulating the output to be invalid AMP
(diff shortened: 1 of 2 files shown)


westonruter:
I have one more thing I want to do here to harden this.

kienstra:
**Looks Good, Works As Expected**

Hi @westonruter,
Sorry for the delay here. To second @amedina's point, this looks good.

I also saw the issue of the issue of [W3 Total Cache](https://wordpress.org/plugins/w3-total-cache/) outputting that `<script>`.

But with this PR applied, that script doesn't appear. And as expected, the component scripts still appear, like https://cdn.ampproject.org/v0/amp-bind-latest.js

disallowed-script-cache

PR #1101, opened 2018-04-29
Fix fatal errors related to output buffering and excessive response header sizes

westonruter:
* Introduce `AMP_Validation_Utils::can_output_buffer()` which looks at the call stack to determine whether an output buffer display handler callback is currently being called; if so, output buffering is prevented since it would result in a fatal error. Note that the call stack is only examined during validation (which require authorized user), not during normal responses for visitors. There is no other way I could find to determine if currently in a display handler.
* Eliminate the sending of validation errors via HTTP response header in favor of only sending via HTML comment. When many validation errors are present the excessive size of a response header can cause a fatal error.
* Switch to using output buffer callback for overall response buffer, as well as using a named method for wrapping hooks with source comments; this ensures the named function can be inspected in the call stack as opposed to the closure. Supersedes #1085.
* Start output buffering as early as possible in `template_redirect` hook.
* Add `AMP_Theme_Support::is_output_buffering()` method which is used to prevent source locators from buffering to add source comments if the overall response has not started buffering.
* Eliminate use of output buffer which was required to completely re-use `wp_scripts()->do_items()`. Supersedes #1097. 

The changes in this PR have been tested to work in the following hosting environments:

1. WordPress.com Business Plan (Pressable)
2. WP Engine
3. Pantheon
4. Dreamhost shared hosting
5. Local by Flywheel
6. VVV

### DIFF
--- a/includes/utils/class-amp-validation-utils.php
+++ b/includes/utils/class-amp-validation-utils.php
@@ -943,7 +943,7 @@ class AMP_Validation_Utils {
 			return false;
 		}
 
-		// Abort when in shutdown since printing we know to be in buffering display handler, and no output allowed anyway.
+		// Abort when in shutdown since output has finished, when we're likely in the overall output buffering display handler.
 		if ( did_action( 'shutdown' ) ) {
 			return false;
 		}
--- a/tests/test-class-amp-theme-support.php
+++ b/tests/test-class-amp-theme-support.php
@@ -879,6 +879,8 @@ class Test_AMP_Theme_Support extends WP_UnitTestCase {
 	 * Test start_output_buffering.
 	 *
 	 * @covers AMP_Theme_Support::start_output_buffering()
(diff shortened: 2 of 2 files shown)


includes/utils/class-amp-validation-utils.php
```diff
@@ -925,6 +929,39 @@ public static function get_source( $callback ) {
...
+	 * @return bool Whether output buffering is allowed.
+	 */
+	public static function can_output_buffer() {
+
+		// Output buffering for validation can only be done while overall output buffering is being done for the response.
+		if ( ! AMP_Theme_Support::is_output_buffering() ) {
+			return false;
+		}
+
+		// Abort when in shutdown since printing we know to be in buffering display handler, and no output allowed anyway.
```
amedina: Nit pick: the phrasing in this comment is hard to parse.

Issue #1104, opened 2018-04-30
Images unexpectedly expand to fill containers


kienstra:
As @westonruter pointed out, images can sometimes appear horizontally stretched.

This issue sometimes appears when images have less width than their container. 

<img width="1291" alt="layout-intrinsic-example" src="https://user-images.githubusercontent.com/4063887/39452231-1bca4a0c-4c97-11e8-8fb1-fa62e829da3c.png">

**Appears on**
* Paired Mode child themes of Twenty Ten, Twenty Fifteen, Twenty Sixteen, and Twenty Seventeen
* Native AMP [child theme of Twenty Fifteen](https://amp-make-xwp.pantheonsite.io/2016/07/24/dependently-contextual-customizer-controls/)
* Legacy Templating ([screenshot](https://cloudup.com/c00IeqnePhu))

**Steps To Reproduce**

1. Checkout the `0.7` branch of this plugin
2. Activate a Paired Mode child theme of Twenty Fifteen (update: or simply use any core theme, not a child theme)
3. Create a new post
4. Add an image that's less wide than `600px`, like [this image](https://cloudup.com/cQZwaa4TeRd) (update: and center-align the image)
5. Save the post, and visit the front-end on a desktop display (appending `/amp` or `?amp=1` to the URL)
Expected: it appears at its natural dimensions (300x232 for the example above)
Actual: it appears stretched (see image above)

This may be related to #937.

kienstra:
**Initial Findings**

So far, this looks to be related to the `aligncenter` class. 

That has the style rule `display: block;`
```css
.aligncenter {
    clear: both;
    display: block;
    margin-left: auto;
    margin-right: auto;
}
```
And that seems to override the rule that AMP has for `layout="intrinsic"`
```css
.i-amphtml-layout-intrinsic {
    display: inline-block;
    ...
}
```

<img width="1286" alt="align-center" src="https://user-images.githubusercontent.com/4063887/39454324-387f55bc-4ca0-11e8-9b61-f60ea98609e4.png">

postphotos:
Hi @kienstra, Thanks for hopping in and thanks @westonruter for reporting! I added a user story/AC here.

westonruter:
The `margin` on the `.amp-wp-enforced-sizes` also seems to be a problem, since it is blowing away the `auto` margins:

![image](https://user-images.githubusercontent.com/134745/39455637-6ed5788a-4c96-11e8-982d-181687d9663b.png)

[...]

PR #1109, opened 2018-05-02
Address an issue of stretched <amp-img> when using '.aligncenter'.

kienstra:
**Request For Code Review**

Hi @mehigh and @westonruter,
Could you please review this pull request, which addresses the problem in #1104 of image stretching? Thanks to @delawski for the solution.

This doesn't address [this comment](https://github.com/Automattic/amp-wp/issues/1104#issuecomment-385586532) in that issue, where "legacy templating" centered images, unless they had an `alignleft` or `alignright` class.

**Background**
There was an issue where the `aligncenter` class overrode the `"display: inline-block"` rule of AMP's `layout=intrinsic attribute`. So this strips that class, and instead wraps the image in a `<figure>` to center it.

**Testing**
The fastest way to test this is to simply use legacy templating, with any theme. Add an image that's less wide than about `600px`, and center-align it.

### DIFF
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -248,8 +248,8 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 	 * @since 0.7
 	 * @see https://github.com/Automattic/amp-wp/issues/1104
 	 *
-	 * @param DOMNode|DOMElement $node The <amp-img> node.
-	 * @return DOMNode|DOMElement $node The <amp-img> node, possibly wrapped in a <figure>.
+	 * @param DOMElement $node The <amp-img> node.
+	 * @return DOMElement $node The <amp-img> node, possibly wrapped in a <figure>.
 	 */
 	public function handle_centering( $node ) {
 		$align_class = 'aligncenter';
@@ -257,7 +257,7 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 		$width       = $node->getAttribute( 'width' );
 
 		// If this doesn't have a width attribute, centering it in the <figure> wrapper won't work.
-		if ( ( false === strpos( $classes, $align_class ) ) || empty( $width ) ) {
+		if ( empty( $width ) || ! in_array( $align_class, preg_split( '/\s+/', trim( $classes ) ), true ) ) {
 			return $node;
 		}
 
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -219,6 +219,7 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 			$new_tag = 'amp-img';
 		}
 		$new_node = AMP_DOM_Utils::create_node( $this->dom, $new_tag, $new_attributes );
+		$new_node = $this->handle_centering( $new_node );
 		$node->parentNode->replaceChild( $new_node, $node );
 	}
 
@@ -235,5 +236,44 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 		$ext  = self::$anim_extension;
 		$path = AMP_WP_Utils::parse_url( $url, PHP_URL_PATH );
 		return substr( $path, -strlen( $ext ) ) === $ext;
+	}
+
+	/**
+	 * Handles an issue with the aligncenter class.
+	 *
+	 * If the <amp-img> has the class aligncenter, this strips the class and wraps it in a <figure> to center the image.
+	 * There was an issue where the aligncenter class overrode the "display: inline-block" rule of AMP's layout="intrinsic" attribute.
+	 * So this strips that class, and instead wraps the image in a <figure> to center it.
+	 *
+	 * @since 0.7
+	 * @see https://github.com/Automattic/amp-wp/issues/1104
+	 *
+	 * @param DOMNode|DOMElement $node The <amp-img> node.
+	 * @return DOMNode|DOMElement $node The <amp-img> node, possibly wrapped in a <figure>.
+	 */
+	public function handle_centering( $node ) {
+		$align_class = 'aligncenter';
+		$classes     = $node->getAttribute( 'class' );
+		$width       = $node->getAttribute( 'width' );
+
+		// If this doesn't have a width attribute, centering it in the <figure> wrapper won't work.
+		if ( ( false === strpos( $classes, $align_class ) ) || empty( $width ) ) {
+			return $node;
+		}
+
+		// Strip the class, and wrap the <amp-img> in a <figure>.
+		$classes = trim( str_replace( $align_class, '', $classes ) );
+		$node->setAttribute( 'class', $classes );
+		$figure = AMP_DOM_Utils::create_node(
+			$this->dom,
+			'figure',
+			array(
+				'class' => $align_class,
+				'style' => "max-width: {$width}px;",
+			)
+		);
+		$figure->appendChild( $node );
+
+		return $figure;
 	}
 }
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -236,4 +237,43 @@ private function is_gif_url( $url ) {
...
+	 *
+	 * @param DOMNode|DOMElement $node The <amp-img> node.
+	 * @return DOMNode|DOMElement $node The <amp-img> node, possibly wrapped in a <figure>.
+	 */
+	public function handle_centering( $node ) {
+		$align_class = 'aligncenter';
+		$classes     = $node->getAttribute( 'class' );
+		$width       = $node->getAttribute( 'width' );
+
+		// If this doesn't have a width attribute, centering it in the <figure> wrapper won't work.
```
kienstra: Still, we'd expect there to always be a `$width`, as `self::FALLBACK_WIDTH` is applied.

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -236,4 +237,43 @@ private function is_gif_url( $url ) {
...
+		if ( ( false === strpos( $classes, $align_class ) ) || empty( $width ) ) {
```
kienstra: Still, we'd expect there to always be a `$width`, as `self::FALLBACK_WIDTH` is applied.

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -236,4 +237,43 @@ private function is_gif_url( $url ) {
...
+			return $node;
+		}
+
+		// Strip the class, and wrap the <amp-img> in a <figure>.
+		$classes = trim( str_replace( $align_class, '', $classes ) );
```
kienstra: We could probably also use this logic to center images in legacy templating if they don't have an `alignleft` or `alignright` class. In `0.6`, they were centered in those cases ([more details](https://github.com/Automattic/amp-wp/issues/1104#issuecomment-385586532)). We could use this to detect "legacy templating":

```php
! current_theme_supports( 'amp' )
```

includes/sanitizers/class-amp-img-sanitizer.php
westonruter: It would be a bit more robust to check for the class via:

```php
in_array( $align_class, preg_split( '/\s+/', trim( $classes ) ), true )
```

includes/sanitizers/class-amp-img-sanitizer.php
westonruter: If we did that, then I'd prefer to pass that flag as an argument to the class as a whole instead of calling this global function. For example, there could be an arg like `auto_align_center` passed in:

https://github.com/Automattic/amp-wp/blob/cd03ee45171e41b3c7ae111afcda58584f96a523/includes/amp-helper-functions.php#L482

If `! current_theme_supports( 'amp' )`.

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -236,4 +237,43 @@ private function is_gif_url( $url ) {
...
+	 * There was an issue where the aligncenter class overrode the "display: inline-block" rule of AMP's layout="intrinsic" attribute.
+	 * So this strips that class, and instead wraps the image in a <figure> to center it.
+	 *
+	 * @since 0.7
+	 * @see https://github.com/Automattic/amp-wp/issues/1104
+	 *
+	 * @param DOMNode|DOMElement $node The <amp-img> node.
+	 * @return DOMNode|DOMElement $node The <amp-img> node, possibly wrapped in a <figure>.
+	 */
+	public function handle_centering( $node ) {
```
westonruter: I like this approach!

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -236,4 +237,43 @@ private function is_gif_url( $url ) {
...
+	 *
+	 * If the <amp-img> has the class aligncenter, this strips the class and wraps it in a <figure> to center the image.
+	 * There was an issue where the aligncenter class overrode the "display: inline-block" rule of AMP's layout="intrinsic" attribute.
+	 * So this strips that class, and instead wraps the image in a <figure> to center it.
+	 *
+	 * @since 0.7
+	 * @see https://github.com/Automattic/amp-wp/issues/1104
+	 *
+	 * @param DOMNode|DOMElement $node The <amp-img> node.
+	 * @return DOMNode|DOMElement $node The <amp-img> node, possibly wrapped in a <figure>.
```
westonruter: There is no need for `DOMNode` in the phpdoc tags here. It will always be a `DOMElement` (and `DOMElement` is a subclass of `DOMNode`).

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -236,4 +237,43 @@ private function is_gif_url( $url ) {
...
+
+	/**
+	 * Handles an issue with the aligncenter class.
+	 *
+	 * If the <amp-img> has the class aligncenter, this strips the class and wraps it in a <figure> to center the image.
+	 * There was an issue where the aligncenter class overrode the "display: inline-block" rule of AMP's layout="intrinsic" attribute.
+	 * So this strips that class, and instead wraps the image in a <figure> to center it.
+	 *
+	 * @since 0.7
+	 * @see https://github.com/Automattic/amp-wp/issues/1104
```
westonruter: NormallyI use `@link` for URLs and `@see` for symbols, but apparently they are synonymous.

tests/test-amp-img-sanitizer.php
```diff
@@ -128,6 +128,16 @@ public function get_data() {
...
+
+			'image_center_aligned'                     => array(
+				'<img class="aligncenter" src="http://placehold.it/350x150" width="350" height="150" />',
+				'<figure class="aligncenter" style="max-width: 350px;"><amp-img class="amp-wp-enforced-sizes" src="http://placehold.it/350x150" width="350" height="150" layout="intrinsic"></amp-img></figure>',
+			),
+
+			'image_left_aligned'                       => array(
+				'<img class="alignleft" src="http://placehold.it/350x150" width="350" height="150" />',
+				'<amp-img class="alignleft amp-wp-enforced-sizes" src="http://placehold.it/350x150" width="350" height="150" layout="intrinsic"></amp-img>',
+			),
```
westonruter: Should there be a test for when an image has a caption? Or is that unnecessary because in that case the `figure` will have the class name instead of the `img`?

includes/sanitizers/class-amp-img-sanitizer.php
kienstra: Thanks, @westonruter. 

@mehigh and I discussed how the legacy templating in `0.6` centers images that don't have `alignleft` and `alignright`, and maybe we don't need to continue that behavior in `0.7`. 

At least images with `alignnone` shouldn't be centered. They're centered in `0.6 ` ([example](https://ryankienstra.wordpress.com/2018/05/01/test-page/amp/)), but they're not centered in `0.7`.

includes/sanitizers/class-amp-img-sanitizer.php
kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/pull/1109/commits/25c2efcb7e0cfbfb77d3f1f6815a34a560760afe) makes that change.

includes/sanitizers/class-amp-img-sanitizer.php
kienstra: Thanks!

includes/sanitizers/class-amp-img-sanitizer.php
kienstra: Thanks, [this commit](https://github.com/Automattic/amp-wp/pull/1109/commits/25c2efcb7e0cfbfb77d3f1f6815a34a560760afe) removes `DOMNode` from the tags.

tests/test-amp-img-sanitizer.php
kienstra: Yes, good idea. [This commit](https://github.com/Automattic/amp-wp/pull/1109/commits/79958f52660ff435832530a0ce1f77046a1d3a00) adds a test for an image with a caption. It ensures that `<figure>` elements don't have the class `aligncenter` stripped, only `<amp-img>` elements.

kienstra:
**Thanks, Applied Code Review Suggestions**

Hi @westonruter,
Thanks for reviewing this. Your suggestions here are applied.

PR #1111, opened 2018-05-03
Prevent CSS tree shaking from removing selectors with classes mentioned in [class] amp-bind attributes

westonruter:
This is a follow-up on #1048, and in particular this https://github.com/Automattic/amp-wp/pull/1048#issuecomment-379117270 from @camelburrito:

> usually the [class names] associated with `amp-bind` can be searchable in the dom (string search), even that would be a little risky because there could be string concat etc happening

I suspect that dynamic class names (constructed via concatenation) would be less likely than static ones, since dynamic class names would depend on having a correspondingly dynamic stylesheet. This doesn't seem likely. 

With the changes in this PR, with the following markup:

```html
<amp-state id="mySidebar">
		{"expanded": false}
	</script>
</amp-state>
<aside class="sidebar" [class]="mySidebar.expanded ? 'expanded' : ''">...</aside>
```
And given the following CSS:

```css
.sidebar { display:none }
.sidebar.expanded { display:block }
```
The second style rule will no longer be removed when tree shaking is being performed.

This should greatly reduce the amount of times that someone has to manually inform style sanitizer of the `dynamic_element_selectors`, for example:

```php
	if ( ! isset( $sanitizers['AMP_Style_Sanitizer']['dynamic_element_selectors'] ) ) {
		$sanitizers['AMP_Style_Sanitizer']['dynamic_element_selectors'] = array();
	}
		$sanitizers['AMP_Style_Sanitizer']['dynamic_element_selectors'],
		array(
			'amp-list',
			'amp-live-list',
			'[submit-error]',
			'[submit-success]',
			'.expanded', // <= Added.
			'.sidebar-open', // <= Added.
		)
	);
	return $sanitizers;
} );
```

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -225,7 +225,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 
 			// Find all [class] attributes and capture the contents of any single- or double-quoted strings.
 			foreach ( $this->xpath->query( '//*/@' . AMP_DOM_Utils::get_amp_bind_placeholder_prefix() . 'class' ) as $bound_class_attribute ) {
-				if ( preg_match_all( '/([\'"])([^\1]+?)\1/', $bound_class_attribute->nodeValue, $matches ) ) {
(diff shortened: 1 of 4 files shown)


camelburrito:
Thank you!! This looks great!

PR #1114, opened 2018-05-03
Fix inconsistency between singular and plural

garrett-eclipse:
This is a very minor localization fix to avoid errors in GlotPress/GlotDict while translating. Removed the extra trailing space from the plural string to avoid the error message seen here;
https://i.imgur.com/6EAQII5.png


westonruter:
Thank you.

PR #1116, opened 2018-05-04
Add workaround to preserve CSS calc() functions

westonruter:
It turns out that CSS `calc()` functions aren't parsed properly by the CSS parser. There is an old issue open to support them: https://github.com/sabberworm/PHP-CSS-Parser/issues/79. In the mean time, this makes sure that these functions are preserved by replacing them with placeholders prior to parsing, and then returning the original functions after parsing.

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -713,6 +713,8 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 					$open_parens++;
 				} elseif ( ')' === $css[ $final_offset ] ) {
 					$open_parens--;
+				} elseif ( ';' === $css[ $final_offset ] || '}' === $css[ $final_offset ] ) {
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -665,6 +690,54 @@ function ( $value ) {
...
+			$css_length   = strlen( $css );
+			$open_parens  = 1;
+			$start_offset = $match_offset + strlen( $match_string );
+			$final_offset = $start_offset;
+			for ( ; $final_offset < $css_length; $final_offset++ ) {
+				if ( '(' === $css[ $final_offset ] ) {
+					$open_parens++;
+				} elseif ( ')' === $css[ $final_offset ] ) {
+					$open_parens--;
+				}
```
westonruter: Probably should add a case here to catch `;` or `}` and `break` because clearly that indicates unbalanced parentheses and we should skip to looking for the next match from that point forward.

kienstra: It looks like the `elseif` block below now handles that case.

PR #1121, opened 2018-05-05
[Gutenberg] Add AMP Carousel for Gallery and AMP Lightbox features for Gallery and Image

miina:
Based on #1026.

Does the following:
- Adds AMP Carousel toggle to Gallery block which allows displaying the gallery as carousel instead of default WP gallery.
- Adds "Add lightbox effect" toggle to Gallery and Image core blocks.
screen shot 2018-05-08 at 4 03 09 pm


includes/sanitizers/class-amp-gallery-block-sanitizer.php
```diff
@@ -0,0 +1,91 @@
...
+			$images     = $node->getElementsByTagName( 'a' );
+			$num_images = $images->length;
+
+			if ( 0 === $num_images ) {
+				continue;
+			}
+
+			$attributes   = array(
+				'width'  => self::FALLBACK_WIDTH,
+				'height' => self::FALLBACK_HEIGHT,
```
miina: Not sure what could be a better alternative for setting width and height. Thoughts?

westonruter: Could the `AMP_Gallery_Block_Sanitizer` get added to the sanitizers list so that it is forced to be run after the `AMP_Img_Sanitizer`? Then you would be able to use the `width` and `height` of the images inside of the gallery to obtain the `width` and `height` for the overall gallery?

miina: Added `AMP_Gallery_Block_Sanitize::get_carousel_height()` which gets the heights of the images and sets the gallery height according to the smallest image. However, saw that basically all the images had the default height and width set (`400px`) within `AMP_Image_Sanitizer::determine_dimensions()`, so also added `AMP_Base_Sanitizer::set_attachment_width_and_height_by_src()` which gets the height and width from the image itself and is used in `AMP_Image_Sanitizer::determine_dimensions()` before the logic that adds the fallback width and height. Does this make sense, or does it interfere with the logic that sets the fallback width and height?

miina:
Note: Also added `amp-image-lightbox`  implementation for Image and Gallery blocks within this PR since it was quite related.

includes/sanitizers/class-amp-gallery-block-sanitizer.php
miina: Thinking that this might also change the looks for default gallery in an unexpected way.

assets/js/amp-editor-blocks.js
```diff
@@ -0,0 +1,596 @@
...
+		}
+		if ( component.hasAmpCarouselSet( attributes || '' ) ) {
+			ampClassName += ' amp-carousel';
+		}
+		if ( component.hasAmpLightboxSet( attributes || '' ) ) {
+			ampClassName += ' amp-lightbox';
+		}
+
+		if ( '' !== ampClassName && attributes.className !== ampClassName ) {
+			props.className = ampClassName.trim();
```
kienstra: Hi @miina,
In my browser, this didn't assign the new `props.className` value, causing the lightbox to not appear on the front-end.

But with this change (borrowing from Weston's solution [here](https://github.com/Automattic/amp-wp/blob/7439d2ab5eebd0ee071749a849fda946a2eff4be/assets/js/amp-block-validation.js#L357)), it did:

```js
props = _.extend( {}, props, { className: ampClassName.trim() } );
```

miina: Hey @kienstra, thanks for finding this, makes sense that the `props.className` can't be directly set this way.

assets/js/amp-editor-blocks.js
```diff
@@ -211,6 +242,15 @@ var ampEditorBlocks = ( function() {
 			ampLayout = attributes.ampLayout;
 
 			if ( 'core/shortcode' === name ) {
+				// Lets remove amp-carousel from edit view.
+				if ( component.hasGalleryShortcodeCarouselAttribute( attributes.text || '' ) ) {
+					props.setAttributes( { text: component.removeAmpCarouselFromShortcodeAtts( attributes.text ) } );
```
westonruter: This will need to be moved to an `onChange` handler.

assets/js/amp-editor-blocks.js
```diff
@@ -211,6 +242,15 @@ var ampEditorBlocks = ( function() {
...
+				}
+				// Lets remove amp-lightbox from edit view.
+				if ( component.hasGalleryShortcodeLightboxAttribute( attributes.text || '' ) ) {
+					props.setAttributes( { text: component.removeAmpLightboxFromShortcodeAtts( attributes.text ) } );
```
westonruter: Per above, this will need to be moved to an `onChange` handler.

assets/css/amp-default.css
```diff
@@ -3,3 +3,11 @@
 	/** Force the image into a box of fixed dimensions and use object-fit to scale. **/
 	object-fit: contain;
 }
+
+#amp-image-lightbox-1 img {
```
westonruter: I don't think we should be targeting `img` elements in selectors. I should think that only targeting `amp-img` should be done since `img` should really be the domain of the `amp-img` to control?

assets/js/amp-editor-blocks.js
```diff
@@ -267,48 +315,171 @@ var ampEditorBlocks = ( function() {
...
 		}
 
+		return el( SelectControl, {
+			label: label,
+			value: ampLayout,
+			options: component.getLayoutOptions( name ),
+			onChange: function( value ) {
+				props.setAttributes( { ampLayout: value } );
+				if ( 'core/image' === props.name ) {
+					component.setImageBlockLayoutAttributes( props, value );
```
westonruter: Per above, this will need to be moved to an `onChange` handler.

includes/embeds/class-amp-gallery-embed.php
```diff
@@ -113,9 +119,24 @@ public function shortcode( $attr ) {
...
+		$args = array(
 			'images' => $urls,
-		) );
+		);
+		if ( ! empty( $atts['lightbox'] ) ) {
+			$args['lightbox'] = true;
+			$lightbox_tag     = AMP_HTML_Utils::build_tag(
+				'amp-image-lightbox',
+				array(
+					'id'                           => AMP_Base_Sanitizer::AMP_IMAGE_LIGHTBOX_ID,
```
westonruter: Why the static `'amp-image-lightbox-1'`? Should this not have some auto-increment number instead that is appended?

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -384,6 +396,12 @@ public function filter_data_amp_attributes( $attributes, $amp_data ) {
 		if ( isset( $amp_data['noloading'] ) ) {
 			$attributes['data-amp-noloading'] = '';
 		}
+		if ( isset( $amp_data['lightbox'] ) ) {
+			$attributes['data-amp-lightbox'] = '';
+			$attributes['on']                = 'tap:' . self::AMP_IMAGE_LIGHTBOX_ID;
```
westonruter: Per above, I think this needs to auto-increment a number to append to `amp-image-lightbox-`.

assets/css/amp-default.css
miina: Looks like `amp-image-lightbox` component is empty by default and clicking on image which is targeting this lightbox will insert the `img` element into the `amp-image-lightbox`, the `amp-image-lightbox` component doesn't include an additional `amp-img` component within, targeting `amp-img` doesn't seem to be possible in this case.
For example after clicking on an an image the lightbox HTML looks like this:
screen shot 2018-05-21 at 10 34 06 pm
Thoughts?

includes/embeds/class-amp-gallery-embed.php
miina: Used `'amp-image-lightbox-1'` originally since thought that maybe there will be more than one, however, looks like using one should be enough, maybe we could just leave it as `'amp-image-lightbox'`. Thoughts?

assets/js/amp-editor-blocks.js
miina: Hmm, removing the attributes from the saved shortcode text on loading the block isn't related to any other attribute changing, the objective here was to not display the saved attributes within the shortcode already when the block is initially loaded, e.g. display just `[gallery]` instead of `[gallery amp-lightbox=true]`. Moving this part to `onChange` handler would cause the AMP attributes show up when first loading the block and remove these once `onChange` triggers.

Should we just display the attributes (in case they're not default) within the shortcode instead of trying to hide them? Thoughts?

assets/js/amp-editor-blocks.js
miina: This one is already in `onChange` handler.

westonruter:
For some reason I get this warning when I click the Carousel or Lightbox toggles the first time:

> Warning: A component is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components

westonruter:
At some point it would probably be a good idea to hide the gallery column count when slideshow is enabled:

image

assets/css/amp-default.css
```diff
@@ -12,4 +12,12 @@ amp-fit-text h4,
...
 	font-size: inherit;
-}
\ No newline at end of file
+}
+
+#amp-image-lightbox div div > :first-child {
+	object-fit: contain;
+}
+
+.entry__content .wp-block-gallery[class*="columns-"] figure {
```
westonruter: `entry__content` is not a standard class name.

assets/css/amp-default.css
```diff
@@ -12,4 +12,12 @@ amp-fit-text h4,
 amp-fit-text h5,
 amp-fit-text h6 {
 	font-size: inherit;
-}
\ No newline at end of file
+}
+
+#amp-image-lightbox div div > :first-child {
+	object-fit: contain;
```
westonruter: This doesn't seem to be having any effect. I don't think it is needed?

assets/css/amp-default.css
```diff
@@ -12,4 +12,12 @@ amp-fit-text h4,
...
+}
+
+.entry__content .wp-block-gallery[class*="columns-"] figure {
+	width: 100%;
```
westonruter: I think this may just be needed in the [ampnews](https://github.com/xwp/ampnews/) theme. It doesn't seem necessary in Twenty Fifteen. It's probably just that the ampnews theme lacks all of the required styles.

includes/sanitizers/class-amp-gallery-block-sanitizer.php
```diff
@@ -78,35 +78,30 @@ public function sanitize() {
...
 				'type'   => 'slides',
 				'layout' => 'fixed-height',
-			);
-			$amp_carousel        = AMP_DOM_Utils::create_node( $this->dom, 'amp-carousel', $carousel_attributes );
-
-			for ( $j = $num_images - 1; $j >= 0; $j-- ) {
-				$amp_carousel->appendChild( $images->item( $j ) );
+			) );
+			foreach ( $images as $image ) {
+				$amp_carousel->appendChild( $image );
```
miina: @westonruter `for ( $j = $num_images - 1; $j >= 0; $j-- ) {` was used here since when looping directly through the nodes (`foreach ( $images as $image ) {`) it's not working as expected, for example when adding a gallery with 5 images locally and using AMP Carousel it will just display the first three in the gallery. That's happening when testing locally now as well, is it working as expected for you?

westonruter: PR to fix opened in #1187.

kienstra:
**Request For Testing**

Hi @csossi,
Could you please test this?

1. Go to https://native-ampconfdemo.pantheonsite.io/
2. Create a new post, using Gutenberg (the default)
3. Add a "Gallery Block," and add images to the gallery
4. Publish the post
5. Toggle the "Display as AMP carousel" and "Add lightbox effect" options that Miina showed in her [comment](https://github.com/Automattic/amp-wp/pull/1121#issue-186168582).
6. Go to the front-end of the post
7. Expected: The gallery displays as a carousel if selected
8. Expected: On selecting "Add lightbox effect," clicking an image expands it to fill the display

csossi:
@kienstra - Works fine on front-end, but noted that a message appeared in the admin on publishing post. Not sure if relevant here, but wanted you to see it:
image
The link here takes me to:
image

kienstra:
**Thanks, Should Be Fine**

Hi @csossi,
Thanks for letting me know about that. The validation error should be fine, as it's not directly related to the post content.

PR #1125, opened 2018-05-07
Force videopress to use html5 player for AMP

yurynix:
Videopress embed uses an `iframe` and a `script` tag,
`script` tag is not allowed in AMP, so it's stripped and a warning is generated.

screen shot 2018-05-07 at 10 19 17

This PR tries to force html5 player, so the player would be actually a `<video />` tag
that would be converted to `<amp-video />`.

### DIFF
--- a/includes/utils/class-amp-validation-utils.php
+++ b/includes/utils/class-amp-validation-utils.php
@@ -554,7 +554,6 @@ class AMP_Validation_Utils {
 
 		// If no results from URL are available, validate post content outside frontend context.
 		if ( empty( $validation_errors ) && post_type_supports( $post->post_type, 'editor' ) ) {
-			do_action( 'pre_amp_render_post', $post->ID );
 			self::process_markup( $post->post_content );
 			$validation_errors = array_merge(
 				$validation_errors,
--- a/includes/utils/class-amp-validation-utils.php
+++ b/includes/utils/class-amp-validation-utils.php
@@ -554,6 +554,7 @@ class AMP_Validation_Utils {
 
 		// If no results from URL are available, validate post content outside frontend context.
 		if ( empty( $validation_errors ) && post_type_supports( $post->post_type, 'editor' ) ) {
+			do_action( 'pre_amp_render_post', $post->ID );
 			self::process_markup( $post->post_content );
 			$validation_errors = array_merge(
 				$validation_errors,
--- a/jetpack-helper.php
+++ b/jetpack-helper.php
@@ -14,6 +14,7 @@ function amp_jetpack_mods() {
 	}
 	amp_jetpack_disable_sharing();
 	amp_jetpack_disable_related_posts();
+	add_filter( 'videopress_shortcode_options', 'amp_videopress_enable_freedom_mode' );
 }
 
 function amp_jetpack_disable_sharing() {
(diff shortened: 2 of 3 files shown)


yurynix:
Despite the code in this PR, I still see the validation warning, I'm not sure why...

yurynix:
Ok, got it, it's because of [this](https://github.com/Automattic/amp-wp/blob/develop/includes/utils/class-amp-validation-utils.php#L593) validation, so shortcodes are processed, but there is no way to hook to it for allowing modification for AMP.

Maybe `do_action( 'pre_amp_render_post', $post_id );` ?

jetpack-helper.php
```diff
@@ -69,3 +70,19 @@ function jetpack_amp_build_stats_pixel_url() {
...
+ * @param array $options videopress shortcode options.
+ * @return array videopress shortcode options with `freedom` set to true
+ */
+function amp_videopress_enable_freedom_mode( $options ) {
+	$options['freedom'] = true;
+	return $options;
+}
+
+if ( isset( $_GET[ AMP_Validation_Utils::VALIDATE_QUERY_VAR ] ) ) { // WPCS: CSRF ok.
+	add_filter( 'videopress_shortcode_options', 'amp_videopress_enable_freedom_mode' );
```
westonruter: This shouldn't be necessary now that you have https://github.com/Automattic/amp-wp/pull/1125/commits/b6a4671d12b56a849ccd2d8477b28b68f4f20ef7 right?

includes/utils/class-amp-validation-utils.php
```diff
@@ -590,6 +590,7 @@ public static function print_edit_form_validation_status( $post ) {
 
 		// If no results from URL are available, validate post content outside frontend context.
 		if ( empty( $validation_errors ) && post_type_supports( $post->post_type, 'editor' ) ) {
+			do_action( 'pre_amp_render_post', $post->ID );
```
westonruter: This should have the phpdoc added to it on the line before:

```php
/** This action is documented in amp.php */
```

yurynix:
@westonruter Thank you for the review, I'll address your comments tomorrow unless someone beats me to it.

I'm not really clear what you meant by your comment here though:
https://github.com/Automattic/amp-wp/pull/1125#pullrequestreview-118081543

Why hooking to `template_redirect` action is better than `pre_amp_render_post`?
Sorry if that's obvious, It's my first time around that repo.

westonruter:
@yurynix In v0.7 we've added `amp` theme support where you can have the entire theme be served as AMP, not just the post templates. For example, you could have your blog index template be served as AMP now (see example site: https://ampdemo.xwp.io/). In that case, there is no `$post` to be passed as the arg for the `pre_amp_render_post` action. But you can see from the source that the `pre_amp_render_post` action actually runs at `template_redirect` (priority 10) anyway:

https://github.com/Automattic/amp-wp/blob/7439d2ab5eebd0ee071749a849fda946a2eff4be/amp.php#L326-L383

So by doing `amp_jetpack_mods()` at `template_redirect` priority 9, then this ensures that the mods are present whether WordPress is serving the old AMP post templates or whether `amp` theme support is present.

westonruter:
@yurynix also, you may want to rebase your commits onto the `0.7` branch so that the changes could be included in 0.7.1.

includes/utils/class-amp-validation-utils.php
westonruter: Actually this changes can be reverted because this entire `if` statement is going to be removed in https://github.com/Automattic/amp-wp/pull/1132

westonruter:
@yurynix I've rebased onto `0.7` so please reset your branch to the remote. I've also made the changes I described. Please test to confirm it works as expected.

includes/utils/class-amp-validation-utils.php
yurynix: Just saw that now, we can drop my last 2 commits, sorry for the noise.

yurynix:
I've tested it and it works great 👍 
Thanks.

PR #1127, opened 2018-05-07
[WIP] Add AMP_Script_Sanitizer to automatically detect and replace known scripts 

### LINKED ISSUE #1032: Add script sanitizer to convert the most common scripts to AMP
WordPress users very commonly add script tags to Custom HTML widgets and directly to theme templates to show ads, add analytics, etc. Assuming that the vast majority of these scripts are very templated (e.g. including via Google Tag Manager), then it should be possible to reliably identify the each script and extract the relevant information to convert to AMP. 

See also #1031 which could serve as a way to build up a database of scripts for prioritization.

amedina:
WordPress users very commonly add script tags to Custom HTML widgets and directly to theme templates to show ads, add analytics, etc. Assuming that the vast majority of these scripts are very templated (e.g. including via Google Tag Manager), then it should be possible to reliably identify each script and extract the relevant information to convert to AMP.  This pull request addresses this. 

Fixes #1032


postphotos:
Hi @amedina, this is a great idea and should help many sites as they enable AMP! Thanks for sharing this PR.

swissspidy:
Just stumbled upon this and realized that the Site Kit plugin has similar script detection for Google Analytics and Google Tag Manager. So there is some overlap at the moment.

westonruter:
@swissspidy I don't think this is going to go anywhere, since it is not scalable. Better to let Site Kit and other plugins focus on their respective domains.

PR #1131, opened 2018-05-07
Switch to installing wp-dev-lib via composer instead of via submodule

westonruter:
Todo:

- [x] Switch to installing `dev-lib` via NPM (or Composer) instead of using submodule. See https://github.com/xwp/wp-dev-lib/pull/278
- [x] Make sure Travis CI does `composer install` first so that `dev-lib` is available.
- [x] Update instructions to remove references to submodules.
- [ ] ~Install WP-CLI's i18n package via Composer.~ (see https://github.com/ampproject/amp-wp/pull/1789)

Other:

- [ ] ~Automatically `composer install` when doing `npm install`.~
- [ ] ~Warn when `composer` is not installed.~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ jobs:
       php: "7.3"
       env: WP_VERSION=5.0
     - php: "5.4"
-      env: WP_VERSION=4.9    DEV_LIB_SKIP=composer,phpcs
+      env: WP_VERSION=4.9    DEV_LIB_SKIP=phpcs
     - php: "5.5"
       env: WP_VERSION=latest DEV_LIB_SKIP=phpcs
     - php: "5.6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ cache:
     - $HOME/deployment-targets
 
 install:
-  - if [[ $DEV_LIB_SKIP =~ composer ]]; then composer install --no-dev; fi
   - nvm install 8.11.4 && nvm use 8.11.4
   - composer install
   - export DEV_LIB_PATH=vendor/xwp/wp-dev-lib/scripts
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,15 +21,16 @@ cache:
 install:
   - if [[ $DEV_LIB_SKIP =~ composer ]]; then composer install --no-dev; fi
   - nvm install 8.11.4 && nvm use 8.11.4
-  - export DEV_LIB_PATH=dev-lib
-  - source $DEV_LIB_PATH/travis.install.sh
+  - composer install
+  - export DEV_LIB_PATH=vendor/xwp/wp-dev-lib/scripts
+  - source "$DEV_LIB_PATH/travis.install.sh"
   - npx grunt shell:webpack_production
 
 script:
-  - source $DEV_LIB_PATH/travis.script.sh
+  - source "$DEV_LIB_PATH/travis.script.sh"
 
 after_script:
-  - source $DEV_LIB_PATH/travis.after_script.sh
+  - source "$DEV_LIB_PATH/travis.after_script.sh"
 
 jobs:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
 # Tell Travis CI we're using PHP
 language: php
 
-# Newer versions like trusty don't have PHP 5.2 or 5.3
-# https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming
-dist: precise
+# Using trusty instead of precise because we don't need PHP 5.2 or 5.3 anymore.
+dist: trusty
+
+addons:
+  apt:
+    packages:
+      # Needed for `xmllint`.
+      - libxml2-utils
 
 notifications:
   email:
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     "wp-coding-standards/wpcs": "1.2.1",
     "dealerdirect/phpcodesniffer-composer-installer": "0.4.4",
     "phpcompatibility/php-compatibility": "9.1.1",
-    "xwp/wp-dev-lib": "dev-feature/composer-first"
+    "xwp/wp-dev-lib": "1.0.0"
   },
   "config": {
     "platform": {
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
   "require-dev": {
     "wp-coding-standards/wpcs": "1.2.1",
     "dealerdirect/phpcodesniffer-composer-installer": "0.4.4",
-    "phpcompatibility/php-compatibility": "9.1.1"
+    "phpcompatibility/php-compatibility": "9.1.1",
+    "xwp/wp-dev-lib": "dev-feature/composer-first"
   },
   "config": {
     "platform": {
--- a/contributing.md
+++ b/contributing.md
@@ -45,7 +45,20 @@ an explicit [Code of Conduct](https://github.com/ampproject/amp-wp/blob/develop/
 To start, clone this repository into your WordPress install being used for development:
 
 ```bash
-cd wp-content/plugins && git clone --recursive git@github.com:ampproject/amp-wp.git amp
+cd wp-content/plugins && git clone git@github.com:ampproject/amp-wp.git amp
+```
+
+Then install the packages:
+
+```bash
+composer install
+npm install
+```
+
+And lastly, do a build of the JavaScript:
+
+```bash
+npm run build
 ```
 
 Lastly, to get the plugin running in your WordPress install, run `composer install` and then activate the plugin via the WordPress dashboard or `wp plugin activate amp`.
@@ -63,7 +76,6 @@ To edit JavaScript code which is built/complied, run `npm run dev` to watch the 
 To create a build of the plugin for installing in WordPress as a ZIP package, do:
 
 ```bash
-git submodule update --init # (if you haven't done so yet)
 composer install # (if you haven't done so yet)
 npm install # (if you haven't done so yet)
 npm run build
@@ -136,7 +148,7 @@ When you push a commit to your PR, Travis CI will run the PHPUnit tests and snif
 
 Contributors who want to make a new release, follow these steps:
 
-0. Do `git submodule update --init --recursive && npm install && composer selfupdate && composer install`.
+0. Do `npm install && composer selfupdate && composer install`.
 1. Bump plugin versions in `amp.php` (×2: the metadata block in the header and also the `AMP__VERSION` constant). Verify via `npx grunt shell:verify_matching_versions`.
 2. Add changelog entry to readme.
 3. Do `npm run build` and install the `amp.zip` onto a normal WordPress install running a stable release build; do smoke test to ensure it works.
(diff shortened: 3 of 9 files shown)


westonruter:
@felixarntz What do you think about this including a change:

```diff
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
  },
  "main": "blocks/index.js",
  "scripts": {
+  "install": "composer install",
   "build": "grunt build && grunt create-build-zip",
   "deploy": "grunt deploy",
   "dev": "cross-env BABEL_ENV=default webpack --watch"
```

Would this be a good idea or would it be trying to do too much? It's not really so hard to do `npm install && composer install` vs just `npm install`. Also, arguments may need to be passed like `composer install --prefer-source` which wouldn't be possible if `composer install` were invoked automatically upon `npm install`. So I think I just answered my own question 😄 

This depends on https://github.com/xwp/wp-dev-lib/pull/278 which upon merging will make a 0.2 version available for us to reference.

westonruter:
Something else we should explore down the road is automatic installation of the `pre-commit` hook as described here: https://github.com/xwp/wp-dev-lib/#configure-the-git-pre-commit-hook

A problem is that Lando isn't playing nice with `pre-commit` so perhaps we should hold off on that for now: https://github.com/lando/lando/issues/904

felixarntz:
@westonruter 
Agreed with your own response, I think it's usually most flexible to not intertwine `composer.json` commands or shortcuts with `package.json` commands or shortcuts. In the case pointed out, I'd definitely say it's not worth it.

Regarding the pre-commit hook, I think we should move forward with this now, as it shouldn't rely on Lando anyway. IMO such a hook should not be tied to a certain environment, it should only rely on the tools that are installed from the repo itself. In that regard, we should also consider installing PHPCS and PHPUnit ourselves. Note that even with Lando, a pre commit hook relying on the tools directly could still be run either from inside the container or outside of it.

I didn't yet review, but definitely approve of this change. Is it ready for review yet? Cause there are still a few boxes to be checked in the description.

composer.json
```diff
@@ -19,7 +19,8 @@
   "require-dev": {
     "wp-coding-standards/wpcs": "1.2.1",
     "dealerdirect/phpcodesniffer-composer-installer": "0.4.4",
-    "phpcompatibility/php-compatibility": "9.1.1"
+    "phpcompatibility/php-compatibility": "9.1.1",
+    "xwp/wp-dev-lib": "dev-feature/composer-first"
```
kasparsd: @westonruter This can be set to `^1.0` now.

contributing.md
```diff
@@ -45,14 +45,25 @@ an explicit [Code of Conduct](https://github.com/ampproject/amp-wp/blob/develop/
...
+And lastly, do a build of the JavaScript:
+
+```bash
+npm run build
+```
 
 Lastly, to get the plugin running in your WordPress install, run `composer install` and then activate the plugin via the WordPress dashboard or `wp plugin activate amp`.
 
-To install the `pre-commit` hook, do `bash dev-lib/install-pre-commit-hook.sh`.
+To install the `pre-commit` hook, do `bash vendor/xwp/wp-dev-lib/scripts/install-pre-commit-hook.sh`.
```
kasparsd: @westonruter Maybe use `brainmaestro/composer-git-hooks` with Composer or husky with npm [per dev-lib docs](https://github.com/xwp/wp-dev-lib/tree/a458f7915c672e0d6431499040d6d8cd135a532a#configure-the-git-pre-commit-hook)?

felixarntz: > Maybe use `brainmaestro/composer-git-hooks` with Composer or `husky` with npm per dev-lib docs?

+1

westonruter:
> In that regard, we should also consider installing PHPCS and PHPUnit ourselves. Note that even with Lando, a pre commit hook relying on the tools directly could still be run either from inside the container or outside of it.

We are actually installing `phpcs` and `phpunit` ourselves via Composer, but the problem is invoking them inside of Lando from the host machine. For `phpcs` this isn't a problem really since it's likely that PHP is installed on the host, but for `phpunit` there has to be the database connection and WordPress environment which is inside of Lando. And running `lando phpunit` is not working inside of the `pre-commit` hook. 

> I didn't yet review, but definitely approve of this change. Is it ready for review yet? Cause there are still a few boxes to be checked in the description.

Yes, it is ready. I've updated the todos based on your confirmation of `composer install`.

contributing.md
westonruter: Agreed. However, given that the `pre-commit` hook doesn't work yet with Lando, I was thinking to defer this to another PR.

composer.json
westonruter: Thanks. I've pinned it to `1.0.0` in 744077a9 since Renovate will open PRs to bump when new versions are available.

westonruter:
> We are actually installing `phpcs` and `phpunit` ourselves via Composer

My mistake. Actually we aren't currently installing `phpunit` and that was due to compatibility with PHP 5.3 I believe, which is no longer a problem.

westonruter:
Installing `phpunit` via Composer I believe is still a problem because one version is needed by PHP 5.x but another by PHP 7, which is why Gutenberg also does not install PHPUnit via `composer.json` but does so imperatively in:

https://github.com/WordPress/gutenberg/blob/480e47742b419e8cddbcc49556b1783c14da7051/bin/install-php-phpunit.sh#L21-L26

kasparsd:
> one version is needed by PHP 5.x but another by PHP 7, which is why Gutenberg also does not install PHPUnit via composer.json but does so imperatively

@westonruter Could we run PHP unit tests only in PHP 7. Are there tests exclusive to older versions of PHP?

westonruter:
> Could we run PHP unit tests only in PHP 7. Are there tests exclusive to older versions of PHP?

@kasparsd Perhaps, but not running tests in PHP<7 would be dangerous because some issues related to specific PHP versions only are exposed when running PHPUnit tests on those PHP versions.

kasparsd:
@westonruter Would https://github.com/PHPCompatibility/PHPCompatibility take care of all the PHP version-specific issues?

Core is running tests with a different version of phpunit based on the version of PHP:

https://github.com/WordPress/wordpress-develop/blob/a0978e8222e09acaf41788cbd4eadae3b60879bc/.travis.yml#L66-L87

I'm having trouble thinking of a unit test that would fail with a different version of phpunit and not be related to PHP syntax/feature differences.

westonruter:
> Would https://github.com/PHPCompatibility/PHPCompatibility take care of all the PHP version-specific issues?

It would catch a lot but it wouldn't catch everything.

> I'm having trouble thinking of a unit test that would fail with a different version of phpunit and not be related to PHP syntax/feature differences.

The different version of phpunit isn't the problem. The problem is the different versions of PHP. There is code that targets specific PHP versions (which come with different library versions):

https://github.com/ampproject/amp-wp/blob/1bac4d525c37c8f0d6ae4d4987741d445251297a/includes/utils/class-amp-dom-utils.php#L77-L79

https://github.com/ampproject/amp-wp/blob/1bac4d525c37c8f0d6ae4d4987741d445251297a/includes/class-amp-http.php#L205-L211

This latter case is irrelevant now given the upgrade to PHP 5.4, but still. Note the `phpcs:ignore` for `PHPCompatibility` because we're explicitly needing to use code that is for an older version. So for other cases like this, the PHPCompatibility tests wouldn't help make sure that the code actually works in the older PHP versions.

felixarntz:
> And running `lando phpunit` is not working inside of the pre-commit hook.

I would strongly advise to not make this project depend on Lando in any way, for the same reason I disagree on #1825. IMO a pre-commit hook should apply to the system that it is run from: It should just work after running `npm install` and `composer install`. In other words, the pre-commit hook should call `vendor/bin/phpunit` (as installed from Composer, more on that below) instead of `lando phpunit`.

> Installing `phpunit` via Composer I believe is still a problem because one version is needed by PHP 5.x but another by PHP 7

We should still install these dependencies via Composer. For CI, this can easily be worked around by putting specific clauses in place for the older versions. In terms of workload and custom tweaks, this is similar to the additional work that needs to be made to install PHPUnit manually in every CI process, but it has the benefit that the default setup on your dev environment is at least proper and not relying on something from the outside that you might not have installed. I've used the approach of installing PHPUnit from Composer in multiple plugins, even those going as far back as supporting PHP 5.2, and there's no issue when working around this in `.travis.yml` to support the very old versions.

> Yeah, `wp-dev-lib` installs a version of phpunit depending on the version of PHP running on Travis. However, it would be better to fix that for consistent tests both locally and on Travis.

I wasn't aware of that. That dependency installing PHPUnit is perfectly fine with me, we should ensure to use it though (again, except for the Travis-CI builds which don't even support Composer). The 4+ is the correct version constraint there.

To summarize these ideas, we should use PHPUnit via Composer everywhere, except in some CI environments, where something like this in `.travis.yml` works great:
```yml
  - |
    case "$TRAVIS_PHP_VERSION" in
      5.6|5.5|5.4|5.3)
        composer global require "phpunit/phpunit:^4"
        composer install
        PHPUNIT_BIN="phpunit"
        ;;
      5.2)
        PHPUNIT_BIN="phpunit"
        ;;
      *)
        composer install
        PHPUNIT_BIN="vendor/bin/phpunit"
        ;;
    esac
```
Then, the file later calls `$PHPUNIT_BIN` (with args as necessary). The snippet here even works for 5.2 and 5.3, which we obviously wouldn't need (particularly great not to support 5.2 since that doesn't even support Composer).

Btw I completely agree with PHPCompatibility not being sufficient to discover issues with old versions. PHPUnit tests need to run for every PHP version we support.

> Using phpunit 4+ will also require removing the Composer platform lock to PHP 5.4 or else it will refuse to install anything newer than phpunit 4:

That's a good point, I didn't think about it when I introduced that lock. We should probably fix that. It would be perfect to have the possibility to only set the fixed platform to 5.4 when doing the build. I've looked into this for about an hour now, but there seems to be no proper support for this. I'll look into it in the near future again, for now let's keep `config.platform` to 5.4, even if that means we always run an old PHPUnit version, even locally.

.travis.yml
```diff
@@ -21,23 +26,24 @@ cache:
 install:
   - if [[ $DEV_LIB_SKIP =~ composer ]]; then composer install --no-dev; fi
```
kasparsd: @westonruter We now require `composer install` on all builds.

kasparsd:
> We should still install these dependencies via Composer.

Could we do something like:

```json
"require-dev": {
	"phpunit/phpunit": "^4.0 || ^5.0 || ^6.0 || ^7.0"
},
```

per https://getcomposer.org/doc/articles/versions.md#version-range

It should then pickup the version supported by the version of PHP available on the system.

.travis.yml
westonruter: Good catch. I should have removed this.

westonruter:
> Could we do something like:
> 
> ```json
> "require-dev": {
> 	"phpunit/phpunit": "^4.0 || ^5.0 || ^6.0 || ^7.0"
> },
> ```
> 
> per https://getcomposer.org/doc/articles/versions.md#version-range

@kasparsd That looks promising! The only thing is that the `config.platform` would need to be removed, yes?

kasparsd:
> The only thing is that the config.platform would need to be removed, yes?

@westonruter Yes, correct.

kasparsd:
@westonruter Maybe leave the phpunit update and the `pre-commit` autoinstall to a different PR?

felixarntz:
> That looks promising! The only thing is that the config.platform would need to be removed, yes?

Regarding the PHPUnit version, it could also use `>=4.0`, which would be a bit simpler to read. However, we shouldn't remove `config.platform` until we've found a better replacement for that IMO. Alternatively, _if_ we remove it, we need to ensure to always have all entries in `require` use fixed versions (no ranges of any kind) that we know to work with PHP 5.4. The `require-dev` versions could use ranges since those aren't included in a plugin build anyway.

felixarntz:
> Maybe leave the phpunit update and the `pre-commit` autoinstall to a different PR?

Probably even two PRs :)

PR #1132, opened 2018-05-07
Limit validation to AMP theme support

### LINKED ISSUE #1130: amp plugin causing inability to view posts
When I click on edit link of post, I am taken to blank screen. We have deactivated the plugin..any resolution. when deactivated, we can see blog post contents...when activated, see this image.
screen shot 2018-05-07 at 12 23 19 pm

westonruter:
* Fixes problems related to `the_content` filters being applied in the admin. See #1130.
* Eliminates validation errors from being displayed when `the_content` filters apply when creating new posts.
* Reduces influx of support topics related to warning messages; the debug link is not helpful unless theme support is present.

To test:

```bash
git clone --recursive https://github.com/Automattic/amp-wp.git amp
cd amp
npm install
composer install
npm run build
```

This will create an `amp.zip` file that you can install on your site. Note you'll have to first deactivate and uninstall the v0.7.0 AMP plugin before you can upload and activate this ZIP on the plugins admin screen. Ideally you could test this on a staging site.

Support tickets:
* https://wordpress.org/support/topic/warning-there-is-content-which-fails-amp-validation/
* https://wordpress.org/support/topic/warning-there-is-content-which-fails-amp-validation-it-will-be-stripped-when-s/
* Fixes #1130

### DIFF
--- a/includes/utils/class-amp-validation-utils.php
+++ b/includes/utils/class-amp-validation-utils.php
@@ -202,13 +202,10 @@ class AMP_Validation_Utils {
 	 * @return void
 	 */
 	public static function init() {
-		if ( current_theme_supports( 'amp' ) ) {
-			add_action( 'init', array( __CLASS__, 'register_post_type' ) );
-			add_filter( 'dashboard_glance_items', array( __CLASS__, 'filter_dashboard_glance_items' ) );
-			add_action( 'rightnow_end', array( __CLASS__, 'print_dashboard_glance_styles' ) );
-			add_action( 'save_post', array( __CLASS__, 'handle_save_post_prompting_validation' ), 10, 2 );
-		}
-
+		add_action( 'init', array( __CLASS__, 'register_post_type' ) );
+		add_filter( 'dashboard_glance_items', array( __CLASS__, 'filter_dashboard_glance_items' ) );
+		add_action( 'rightnow_end', array( __CLASS__, 'print_dashboard_glance_styles' ) );
+		add_action( 'save_post', array( __CLASS__, 'handle_save_post_prompting_validation' ), 10, 2 );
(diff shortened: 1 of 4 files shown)


kienstra:
**Thanks, Will Review Soon**

Hi @westonruter,
Good idea here. If it's alright, I'll start reviewing in about an hour.

includes/utils/class-amp-validation-utils.php
```diff
@@ -535,64 +532,44 @@ public static function print_edit_form_validation_status( $post ) {
 			return;
 		}
 
-		$url                    = null;
-		$validation_status_post = null;
-		$validation_errors      = array();
+		// Skip if the post type is not viewable on the frontend, since we need a permalink to validate.
+		if ( ! is_post_type_viewable( $post->post_type ) ) {
+			return;
```
kienstra: Good idea to exit if the post type isn't viewable.

includes/utils/class-amp-validation-utils.php
```diff
@@ -535,64 +532,44 @@ public static function print_edit_form_validation_status( $post ) {
...
-			$validation_errors = array_merge(
-				$validation_errors,
-				self::$validation_errors
-			);
-			self::reset_validation_results();
-
-			// Make sure original post is restored after applying shortcodes which could change it.
-			$GLOBALS['post'] = $post; // WPCS: override ok.
-			setup_postdata( $post );
+		if ( $validation_status_post ) {
```
kienstra: Could this be an else block of the earlier `if block?
```php
if ( ! $validation_status_post ) {
	return;
} else { 
	$validation_errors = json_decode( $validation_status_post->post_content, true );
}
```

westonruter: Actually, there's no need for the `if` at all here. Good point.

PR #1141, opened 2018-05-10
Soft-deprecate AMP_Base_Sanitizer::get_body_node()

westonruter:
There's no need to hard-deprecate this method with a warning. And it may be useful still to have this method to get the `body` when the `html` element is not desired.

See https://wordpress.org/support/topic/amp_base_sanitizerget_body/

### DIFF
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -159,11 +159,10 @@ abstract class AMP_Base_Sanitizer {
 	 * Get HTML body as DOMElement from DOMDocument received by the constructor.
 	 *
 	 * @deprecated Just reference $root_element instead.
-	 * @return DOMElement The body or html element.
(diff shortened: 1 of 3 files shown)


kienstra:
**Include in 0.7.1?**

Hi @westonruter,
Do you think it'd be possible to include this in the `0.7.1` release? There have been a few [support](https://wordpress.org/support/topic/php-notice-amp_base_sanitizerget_body_node/
) topics for this, and in one I accidentally said that this fix would be in the next release.

westonruter:
@kienstra good catch. I merged this into `develop` accidentally. I've cherry-picked cbc0ae5 into the 0.7 branch: c65cbf85d0110620c0fcb29c3db12348f2aa1df3

kienstra:
Thanks, @westonruter!

kienstra:
This topic might also be related to this issue:
https://wordpress.org/support/topic/amp-validation-errors-5/

PR #1143, opened 2018-05-11
Proposing amp_post_template_body_open hook (AMP Reader mode version of wp_body_open)

koconder:
Proposing a hook for post body, this will be part of an upcoming fix changes for third party analytics tracking tags and some A/B testing tools which require actions after opening body.

### DIFF
--- a/templates/html-start.php
+++ b/templates/html-start.php
@@ -24,4 +24,4 @@
 </head>
 
 <body class="<?php echo esc_attr( $this->get( 'body_class' ) ); ?>">
-<?php do_action( 'wp_body_open', $this ); ?>
+<?php do_action( 'amp_wp_body_open', $this ); ?>
--- a/templates/html-start.php
+++ b/templates/html-start.php
@@ -24,4 +24,4 @@
 </head>
 
 <body class="<?php echo esc_attr( $this->get( 'body_class' ) ); ?>">
-<?php do_action( 'amp_wp_body_open', $this ); ?>
+<?php do_action( 'amp_post_template_body_open', $this ); ?>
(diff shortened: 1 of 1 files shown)


westonruter:
Which such AMP components require placement as the first child of the `body`?

westonruter:
It's not a requirement for [`amp-experiment`](https://www.ampproject.org/docs/reference/components/amp-experiment) or [`amp-analytics`](https://www.ampproject.org/docs/reference/components/amp-analytics).

koconder:
@westonruter some analytics tools such as Google Tag Manager AMP requires injection of the amp-analytics post-body. I will source some more reliable information, as documentation is scarce.

hkchakladar:
@westonruter So what's the hook available to insert auto-ads code?

westonruter:
@hkchakladar For the legacy post templates embedded with the plugin, you can just use the `amp_post_template_footer ` action.

koconder:
Reopening https://make.wordpress.org/themes/2019/03/29/addition-of-new-wp_body_open-hook/

koconder:
Updated to use https://developer.wordpress.org/reference/hooks/wp_body_open/ instead as per the new wp_body_open() command

koconder:
@westonruter i have pushed it back to where it originally was as amp_wp_body_open() - Let me know what we need in order to get this in, as wp_body_open() was much needed for some time and we need an AMP equivalent.

My original commit was using amp_post_template_body()

koconder:
@westonruter any luck reviewing my comments above?

westonruter:
I've updated the `amp-auto-ads` example plugin to make use of this new hook: https://gist.github.com/westonruter/a41573b932e24810b09949136b9a8445

koconder:
Thanks @westonruter

PR #1147, opened 2018-05-14
Adding id as a valid attribute for amp-iframe

juanchaur1:
The id attribute of any iframe is being removed in the amp-iframe-sanitizer. I believed it's save to pass it as it's done with the classes, size, etc.

### DIFF
--- a/includes/sanitizers/class-amp-allowed-tags-generated.php
+++ b/includes/sanitizers/class-amp-allowed-tags-generated.php
@@ -1657,9 +1657,7 @@ class AMP_Allowed_Tags_Generated {
 						'value' => '',
 					),
 					'sandbox' => array(),
-					'scrolling' => array(
-						'value_regex' => 'auto|yes|no',
-					),
+
 					'src' => array(
 						'blacklisted_value_regex' => '__amp_source_origin',
 						'value_url' => array(
--- a/includes/sanitizers/class-amp-allowed-tags-generated.php
+++ b/includes/sanitizers/class-amp-allowed-tags-generated.php
@@ -1657,7 +1657,9 @@ class AMP_Allowed_Tags_Generated {
 						'value' => '',
 					),
 					'sandbox' => array(),
-
+					'scrolling' => array(
+						'value_regex' => 'auto|yes|no',
+					),
 					'src' => array(
 						'blacklisted_value_regex' => '__amp_source_origin',
 						'value_url' => array(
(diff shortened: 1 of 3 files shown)


includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -135,6 +136,7 @@ private function filter_attributes( $attributes ) {
 				case 'sandbox':
 				case 'class':
 				case 'sizes':
+				case 'id':
```
westonruter: Instead of just allowing `id` we should probably obtain all attributes that are allowed on `amp-iframe`, including `data-*` ones. For non-`data-*` attributes, the list of attributes can be obtained via `AMP_Allowed_Tags_Generated::get_allowed_tag( 'amp-iframe' )`. The list of globally-allowed attributes can be obtained via `AMP_Allowed_Tags_Generated::get_allowed_attributes()`...

On second thought, we it would be better to just copy _all_ of the attributes over from the `iframe` because then the whitelist sanitizer can then report when invalid attributes are present, as opposed to silently stripping them out.

juanchaur1:
Following your advice I am just normalizing specific attributes in amp-iframe and letting AMP_Tag_And_Attribute_Sanitizer do the rest.

includes/sanitizers/class-amp-allowed-tags-generated.php
```diff
@@ -1657,9 +1657,7 @@ class AMP_Allowed_Tags_Generated {
 						'value' => '',
 					),
 					'sandbox' => array(),
-					'scrolling' => array(
-						'value_regex' => 'auto|yes|no',
-					),
+
```
westonruter: Why is this change here? Any changes to this generated file should be done via [`amphtml-update.sh`](https://github.com/Automattic/amp-wp/blob/develop/bin/amphtml-update.sh)

juanchaur1:
@westonruter I believe it's ready to be merge now 👍

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -161,7 +153,11 @@ private function filter_attributes( $attributes ) {
 					}
 					break;
 
+				case 'scrolling':
+					break;
```
westonruter: @juanchaur1 I don't think this is right because `scrolling` is allowed on `amp-iframe`:

https://github.com/ampproject/amphtml/blob/0a7e981cf65ef61394fe36f532c2b26a1b653606/extensions/amp-iframe/validator-amp-iframe.protoascii#L56-L58

In any case, it would be better to let the whitelist sanitizer remove it so that a validation error would be reported.

PR #1149, opened 2018-05-15
Deprecate Jetpack helper and some parts of WPCOM helper

### LINKED ISSUE #1021: Move WPCOM and Jetpack-specific codebase into the Jetpack plugin
As we work to add AMP support to the WordPress ecosystem it is not feasible for the AMP plugin itself to support all of the plugins. Instead, the plugins themselves should be supporting AMP. So the current Jetpack-specific code should be moved into the AMP plugin. The new Lazy Images feature of Jetpack actually was built with [AMP-support baked in](https://github.com/Automattic/jetpack/blob/7c3d69bbf27e761e515d758fcdd6fbb6199e5836/modules/lazy-images/lazy-images.php#L104-L107). 

Code in scope for moving to Jetpack:

* Jetpack-specific embeds.
* `/jetpack-helper.php`
* `/wpcom`
* `/wpcom-helper.php`

gravityrail:
Fixes #1021

Removes the Jetpack helper after Jetpack version 6.2. Also moves some wpcom-helper functions over to Jetpack, where appropriate.

Once https://github.com/Automattic/jetpack/pull/9458 is merged, this can be merged too, as AMP support will be explicitly included in Jetpack.


westonruter:
This is part of https://github.com/Automattic/amp-wp/issues/1021

westonruter:
@gravityrail Would you merge this PR if the state of Jetpack is ready for it?

westonruter:
@gravityrail I just realized that we may have a timing issue with this PR and Jetpack. If AMP 1.0 is released after Jetpack 6.2, is that going to cause a problem? If so, we may want to merge these commits into the 0.7 branch to ensure that it is released before Jetpack.

PR #1150, opened 2018-05-16
Add support for extracting (pixel) dimensions from SVG images

westonruter:
When there are `img` tags that reference SVG images and the `img` does not include the `width` or `height` in HTML, we need to obtain the dimensions of the underlying SVG image in its own `width` and `height` attributes in a similar way to how we extract dimensions from raster image formats like PNG and JPEG.

- [x] Extract width/height from `viewBox` if `width` or `height` attributes aren't pixel values.

### DIFF
--- a/includes/lib/fastimage/class-fastimage.php
+++ b/includes/lib/fastimage/class-fastimage.php
@@ -70,7 +70,12 @@ class FastImage
                 case chr(0x89).'P':
                     return $this->type = 'png';
                 default:
-                    return false;
+                    $this->strpos = 0;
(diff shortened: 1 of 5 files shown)


includes/lib/fastimage/class-fastimage.php
```diff
@@ -70,7 +70,14 @@ public function getType()
                 case chr(0x89).'P':
                     return $this->type = 'png';
                 default:
-                    return false;
+                    $this->strpos = 0;
```
amedina: Just a clarifying question: why can't the SVG part be on its own 'case' branch rather than on the 'default' one?

PR #1151, opened 2018-05-16
[Gutenberg] Add amp-fit-text support to text blocks

miina:
Adds `amp-fit-text` toggle and controls to the following blocks:
- paragraph;
- heading;
- subheading;
- code;
- quote;

Settings when amp-fit-text is selected:
screen shot 2018-05-16 at 9 16 08 pm

Currently has the default layout of `fixed-height`, perhaps this could be changed with a next iteration once AMP Layout has been implemented.


westonruter:
Conflicts need resolving.

miina:
Develop merged & conflicts resolved.

assets/js/amp-editor-blocks.js
```diff
@@ -314,6 +345,77 @@ var ampEditorBlocks = ( function() {
...
+			minFont = props.attributes.minFont,
+			maxFont = props.attributes.maxFont,
+			height = props.attributes.height,
+			isSelected = props.isSelected,
+			el = wp.element.createElement,
+			InspectorControls = wp.blocks.InspectorControls,
+			TextControl = wp.components.TextControl,
+			ToggleControl = wp.components.ToggleControl,
+			PanelBody = wp.components.PanelBody,
+			label = 'Use AMP Fit Text';
```
westonruter: Needs translation

assets/js/amp-editor-blocks.js
```diff
@@ -314,6 +345,77 @@ var ampEditorBlocks = ( function() {
...
+				el( InspectorControls, { key: 'inspector' },
+					el( PanelBody, { title: component.data.ampSettingsLabel },
+						el( ToggleControl, {
+							label: label,
+							checked: ampFitText,
+							onChange: function() {
+								props.setAttributes( { ampFitText: ! ampFitText } );
+							}
+						} ),
+						el( TextControl, {
```
westonruter: Can the `font-size-picker` control be used instead?

assets/js/amp-editor-blocks.js
```diff
@@ -314,6 +345,77 @@ var ampEditorBlocks = ( function() {
...
+							value: minFont,
+							onChange: function( nextMinFont ) {
+								props.setAttributes( { minFont: nextMinFont } );
+							}
+						} ),
+						el( TextControl, {
+							label: __( 'Max font (px)' ),
+							value: maxFont,
+							onChange: function( nextMaxFont ) {
+								props.setAttributes( { maxFont: nextMaxFont } );
```
westonruter: Make sure the `height` is set to be the `Math.max()` of the `nextMaxFont` and its current value.

assets/js/amp-editor-blocks.js
```diff
@@ -314,6 +345,77 @@ var ampEditorBlocks = ( function() {
...
+							value: height,
+							onChange: function( nextHeight ) {
+								props.setAttributes( { height: nextHeight } );
+							}
+						} ),
+						el( TextControl, {
+							label: __( 'Min font (px)' ),
+							value: minFont,
+							onChange: function( nextMinFont ) {
+								props.setAttributes( { minFont: nextMinFont } );
```
westonruter: Abort if `nextMinFont` is larger than `maxFont`

assets/js/amp-editor-blocks.js
```diff
@@ -314,6 +345,77 @@ var ampEditorBlocks = ( function() {
...
+
+		if ( ampFitText ) {
+			return isSelected && (
+				el( InspectorControls, { key: 'inspector' },
+					el( PanelBody, { title: component.data.ampSettingsLabel },
+						el( ToggleControl, {
+							label: label,
+							checked: ampFitText,
+							onChange: function() {
+								props.setAttributes( { ampFitText: ! ampFitText } );
```
westonruter: Should `ampFitText` be replaced with `event.checked`?

westonruter: Can this also result in the `font-size-picker` and `range-control` under Text Settings to be hidden when `ampFitText` is true?

miina: Most of the core blocks seem to use negating the current value for toggle controls (as far as I've seen), thought of staying consistent with the core blocks when using that. Thoughts?

miina:
@westonruter Changes made. It's still possible to set larger minFont than maxFont, for example when first setting `minFont` to 30 and then setting `maxFont` to something lower. However, adding the check for aborting `maxFont` change when it's lower than `minFont` would make editing the values very unfriendly for the user. E.g. if `maxFont` and `minFont` are both 15 and the user would like to set `maxFont` to 25, then removing `5` from `15` wouldn't be possible since then the temporary `1` would be lower than the `minFont` value.

Would be great if the InspectorControls would have validation option (for example with option to display error message) and disable saving in case of incorrect values instead of automatically changing other values. Wondering if something like this is planned, tried to find an issue for it but without success.

westonruter:
@pbakaus with 937035f I've made it so that if you are using an AMP block in content that the required AMP component scripts will get added to the page even in non-AMP documents. This ensures that the components actually render instead of being empty, which seems better. The downside here is that the AMP runtime is trying to take control of the entire page, not just the components you're using. So for example attempting to submit the comment form results in an error:

image

I think that the AMP runtime should perhaps skip doing anything other than render specific components when the document lacks the boilerplate or the `amp` attribute on the `html` element. So that the AMP components can be better used as a component library. Would this be “dirty AMP”?

pbakaus:
@westonruter good call. Yes, I think it's fair to file this on the amphtml repo as issue, should be a fairly easy fix.

westonruter:
> I think it's fair to file this on the amphtml repo as issue

https://github.com/ampproject/amphtml/issues/15583

PR #1165, opened 2018-05-22
[Gutenberg] Add amp-mathml block

miina:
Adds Gutenberg block for `amp-mathml` width input for formula and toggle for displaying inline.

screen shot 2018-05-22 at 8 38 24 pm


miina:
@westonruter Currently the block is not displaying the converted formula within editor, do you think that's OK for the first iteration, or should we try to implement displaying the converted formula in editor, too?

westonruter:
> Currently the block is not displaying the converted formula within editor, do you think that's OK for the first iteration, or should we try to implement displaying the converted formula in editor, too?

Yes, this is OK for the first iteration. Once AMP can be embedded inside of React components, then we can hook up `amp-mathml` proper in the preview.

blocks/amp-mathml/index.js
```diff
@@ -0,0 +1,79 @@
...
+								label={ __( 'Display inline' ) }
+								checked={ displayInline }
+								onChange={ () => ( setAttributes( { displayInline: ! displayInline } ) ) }
+							/>
+						</PanelBody>
+					</InspectorControls>
+				),
+				<PlainText
+					key='formula'
+					tagName='div'
```
westonruter: I'm getting a warning here:

> Warning: React does not recognize the `tagName` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `tagname` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

PR #1173, opened 2018-05-24
1171: Ensure translatable strings in blocks can actually be translated

### LINKED ISSUE #1171: Ensure translatable strings in blocks can actually be translated
Currently the blocks have `wp.i18n.__( … )` calls but the strings are not getting extracted for translation.

See also https://make.wordpress.org/core/2018/05/01/javascript-internationalization-the-missing-pieces/

miina:
Add creating `amp.pot` file from blocks' JS files and add to inline script for making sure that the strings can be translated.

Fixes #1171.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -31,7 +31,15 @@ npm install # (if you haven't done so yet)
 npm run build
 ```
 
-This will create an `amp.zip` in the plugin directory which you can install. The contents of this ZIP are also located in the `build` directory which you can `rsync` somewhere as well. 
+This will create an `amp.zip` in the plugin directory which you can install. The contents of this ZIP are also located in the `build` directory which you can `rsync` somewhere as well.
+
+TO create a build of the plugin as it will be deployed to WordPress.org, run:
+
+```bash
+npm run build-release
+```
+
+Note that this will currently take much longer than a regular build because it generates the files required for translation. You also must have WP-CLI installed with the `i18n` package.
 
 ## Updating Allowed Tags And Attributes
 
@@ -98,7 +106,7 @@ When you push a commit to your PR, Travis CI will run the PHPUnit tests and snif
 
 Contributors who want to make a new release, follow these steps:
 
-1. Do `npm run build` and install the `amp.zip` onto a normal WordPress install running a stable release build; do smoke test to ensure it works.
+1. Do `npm run build-release` and install the `amp.zip` onto a normal WordPress install running a stable release build; do smoke test to ensure it works.
 2. Bump plugin versions in `package.json` (×1), `package-lock.json` (×1, just do `npm install` first), `composer.json` (×1), and in `amp.php` (×2: the metadata block in the header and also the `AMP__VERSION` constant).
 3. Add changelog entry to readme.
 4. Draft blog post about the new release.
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -42,7 +42,16 @@ module.exports = function( grunt ) {
 			verify_matching_versions: {
 				command: 'php bin/verify-version-consistency.php'
 			},
-			create_release_zip: {
+			webpack_production: {
+				command: 'cross-env BABEL_ENV=production webpack'
+			},
+			pot_to_php: {
+				command: 'npm run pot-to-php'
+			},
+			makepot: {
+				command: 'wp i18n make-pot .'
+			},
+			create_build_zip: {
 				command: 'if [ ! -e build ]; then echo "Run grunt build first."; exit 1; fi; if [ -e amp.zip ]; then rm amp.zip; fi; cd build; zip -r ../amp.zip .; cd ..; echo; echo "ZIP of build: $(pwd)/amp.zip"'
 			}
 		},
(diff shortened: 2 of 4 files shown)


includes/admin/class-amp-editor-blocks.php
```diff
@@ -105,6 +105,10 @@ public function enqueue_block_editor_assets() {
 			true
 		);
 
-		wp_add_inline_script( 'amp-editor-blocks', sprintf( 'ampEditorBlocks.boot();' ) );
+		wp_add_inline_script(
+			'amp-editor-blocks',
+			'wp.i18n.setLocaleData( ' . wp_json_encode( gutenberg_get_jed_locale_data( 'amp' ) ) . ', "amp" );' . sprintf( 'ampEditorBlocks.boot();' ),
```
miina: At this moment it's just including the `amp.pot` file created by `babel`, not including the PHP file translations, wondering if we should create a separate locale for the blocks and load only that specific file within the blocks instead?

languages/amp.pot
```diff
@@ -0,0 +1,4 @@
+msgid ""
+msgstr ""
+"Content-Type: text/plain; charset=utf-8\n"
+"X-Generator: babel-plugin-makepot\n"
```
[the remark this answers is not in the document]
westonruter: I think we should follow Gutenberg's lead here. The POT file is not committed to version control: https://github.com/WordPress/gutenberg/tree/master/languages

The POT file generation should be done as part of the build.

includes/admin/class-amp-editor-blocks.php
westonruter: That's one of the questions raised in https://make.wordpress.org/core/2018/05/01/javascript-internationalization-the-missing-pieces/ under “Loading only specific set of translations”.

> Use a different text domain per JavaScript module. Export a JSON file per text domain. This is appealing, but has to be ruled out quickly: the text domain is not known to GlotPress and is not stored in the database or anything.

If Babel is just creating an `amp.pot` for the JS strings, then we can just use that as-is and ignore the `amp.pot` for all other purposes. In other words, maybe there could be an alternative to `gutenberg_get_jed_locale_data()` which allows you to pass a path to a POT file instead (or maybe there is an underlying function that could be used instead). And then this POT file would _not_ be used for translating the plugin as a whole. Instead, Babel should generate a dummy PHP file that contains all of the strings in it for translation by GlotPress. You can see that Gutenberg does this by generating a `languages/gutenberg-translations.php` file in addition to the `languages/gutenberg.pot`.

Not sure if this is possible, but it's what I'd be looking into.

miina: If I understand correctly then the following is true on using different text domain per JS module, however, it should be possible for separating the editor blocks translation file for all the blocks from the rest in case of this plugin since `babel` is creating the `.pot` file based on the `block` directory files only.
>Use a different text domain per JavaScript module. Export a JSON file per text domain. This is appealing, but has to be ruled out quickly: the text domain is not known to GlotPress and is not stored in the database or anything.

Added a commit where `amp-js` locale is registered for editor blocks JS translation and is creating a separate `amp-js.pot` file for translation of the blocks, leaving `amp.pot` available for PHP and any other translations outside of `blocks` directory: 8ac7759. At least it seems to be working locally when translating some block-related strings within `amp-js-et.po` and strings from PHP files within `amp-et.po`.

Do you think this could be suitable setup for the plugin?

westonruter: I don't know how well this will work with GlotPress on translate.wordpress.org. My understanding is that the POT files are not read and that only the strings used in PHP files are parsed and made available for translation. How would GlotPress be told of the POT files in the first place? I could very well be wrong. @swissspidy Do you have insights to share on the approach?

swissspidy: Regarding telling about POT files: it‘s one of the possible features for JS 18N, see our latest chat summary.

swissspidy: For reference: https://github.com/swissspidy/gutenberg-i18n-block

miina: Good to know that the JS i18n defining is happening right now, taking a look at the chat summary. Thanks for the clarification about the limitations of the current system.

contributing.md
```diff
@@ -31,7 +31,15 @@ npm install # (if you haven't done so yet)
...
-This will create an `amp.zip` in the plugin directory which you can install. The contents of this ZIP are also located in the `build` directory which you can `rsync` somewhere as well. 
+This will create an `amp.zip` in the plugin directory which you can install. The contents of this ZIP are also located in the `build` directory which you can `rsync` somewhere as well.
+
+TO create a build of the plugin as it will be deployed to WordPress.org, run:
+
+```bash
+npm run build-release
+```
+
+Note that this will currently take much longer than a regular build because it generates the files required for translation. You also must have WP-CLI installed with the `i18n` package.
```
swissspidy: @westonruter You might wanna add a link to https://github.com/wp-cli/i18n-command

swissspidy: Regarding the performance: a quick guess is that the command currently scans every folder looking for PHP files to parse. We might wanna exclude this directory (and others) by default. See https://github.com/wp-cli/i18n-command/blob/90f3f8e80a8e7db71e3b1e2d2b26140d5dd9e9fe/src/WordPressCodeExtractor.php#L115-L135 and https://github.com/wp-cli/i18n-command/issues/27.

If you try running `wp i18n make-pot` without a node_modules folder being there, it should definitely be very fast. If not, it's something else.

PR #1175, opened 2018-05-25
1094: Transform CSS selectors according to sanitizer HTML element to AMP component conversions

### LINKED ISSUE #1094: Transform CSS selectors according to sanitizer HTML element to AMP component conversions
In Twenty Ten without AMP, the header looks like this:

image

In AMP, however, it looks like this:

image

Notice the white gap appearing in AMP where a border appears on the non-AMP version. This is due in part to Twenty Ten having a stylesheet with:

```css
#branding img {
    border-top: 4px solid #000;
    border-bottom: 1px solid #000;
    display: block;
    float: left;
}
```

When the image sanitizer converts `img` to `amp-img`, this style selector then stops working as expected. Really, the style sanitizer should be converting `img` elements in stylesheets to `amp-img`. All such conversions which are done by the sanitizers should be communicated down to the style sanitizer so that it can make the necessary replacements when it processes a stylesheet.

In other words, I don't think this is something that a theme necessarily should be expected to fix by switching to use a CSS class name or manipulate with a custom sanitizer (e.g. https://github.com/Automattic/amp-wp/pull/1074).

miina:
Modify `AMP_Style_Sanitizer` so that it would convert non-amp CSS selectors to amp component selectors in case it seems to be necessary. Fixes issue with potential broken design, for example with Twenty Ten header.

Fixes #1094.


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1446,51 +1528,35 @@ function( $selector ) {
...
+							$amp_selectors_parsed = $this->get_ampified_selectors_parsed( $parsed_selector, $selector );
+
+							// @todo Maybe we should also make sure here that the original selector is not needed?
+							if ( is_array( $amp_selectors_parsed ) && $this->should_include_selector( $dynamic_selector_pattern, $amp_selectors_parsed['parsed_selector'], $amp_selectors_parsed['selector'] ) ) {
+
+								// @todo Check if there could be selector's array?
+								$selectors[] = $amp_selectors_parsed['selector'];
+							} else {
+								$selectors = array_keys( $selectors_parsed );
+							}
```
miina: @westonruter Started adding the logic within this method since it seemed to be related with tree shaking in terms of adding/not adding the selector if it's not found after replacing. That's still WIP (e.g. currently works only in case `! $should_tree_shake` and is missing most of the mappings) but let me know if there are any general with this approach.

miina: Hmm, maybe one potential issue could be that the size of the style could exceed the limit if doing the replacement here.

miina: Moving the logic to `process_css_declaration_block` to avoid issues with style size changing after the size check.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1435,6 +1441,35 @@ private function finalize_styles() {
 		}
 	}
 
+	/**
+	 * Ampify CSS selectors.
+	 *
+	 * @param Selector $selector CSS Selector.
+	 */
+	private function ampify_selector( $selector ) {
```
miina: @westonruter Reworked the logic to replace the components before tree shaking instead (compared to the previous commit). Wondering if it's OK to just replace the selectors or is it possible that this way we'll be replacing a selector that's actually in use.

Also, in case of img it can be replaced with either `amp-img` or `amp-playbuzz`, thinking if we should add an additional selector for potential `amp-anim`, too.

assets/css/amp-default.css
```diff
@@ -1,4 +1,4 @@
-.amp-wp-unknown-size img {
+.amp-wp-unknown-size .i-amphtml-replaced-content {
```
westonruter: It is illegal in AMP to use `i-amphtml-replaced-content` class names. See:

> Internal AMP class names prefixed with `-amp-` and `i-amp-` are disallowed in AMP HTML.

https://www.ampproject.org/docs/fundamentals/spec#classes

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: It would be ideal if the selector mappings could be added to each sanitizer. For example, on `AMP_Img_Sanitizer` there could be a `get_selector_mapping()` method which returns:

```php
array(
    'img' => array( 'amp-img', 'amp-anim' )
)
```

Whereas on the `AMP_Audio_Sanitizer` class it would return:

```php
array(
    'video' => array( 'amp-video ),
)
```

In other words, there could be multiple AMP components that could be converted so the selector should do replacements for each one, potentially duplicating a selector however many times is needed.

As for how to get the `AMP_Style_Sanitizer` to be able to invoke such a `get_selector_mapping` method on the other sanitizers, it could be done similar to as proposed for passing embed handlers to an embed sanitizer: https://github.com/Automattic/amp-wp/pull/1128/files#diff-2585472f207548f626a43a7ff3cab922R150

I have a concern with that approach of passing class instances as sanitizer args, however. It could cause problems with the response caching introduced in #1156, since the cache key is computed by passing the sanitizer args:

https://github.com/Automattic/amp-wp/blob/ab7204af35a8eb68bd8840a2e3531df8ae73c21f/includes/class-amp-theme-support.php#L1073-L1078

/cc @juanchaur1

DavidCramer:
@miina it's also important to note that `<amp-img>`  wraps a normal `` tag, however this tag is inserted after the page renders which means that the element didn't exist when the tree shaking takes place. So the style `.amp-wp-unknown-size img` would have been removed since it was not pointing to an existing element.

A suggestion would be to maybe have a rule that any class starting with `amp-wp-*` be exempted.

assets/css/amp-default.css
miina: Yes, think I even saw a todo note recently somewhere as well to remove these from CSS within style sanitizer and realized that it's illegal. Will look into which selector to use.

miina:
@DavidCramer Do you mean "exempted" as excluded from sanitizing and never removed from CSS?

DavidCramer:
@miina yes, but only if there is an element with the class.
example:
```css
.amp-wp-unknown-size img {
....
}
```
if it finds an element with `class="amp-wp-unknown-size"` it will be left in as it can be assumed that the `img` will exist but doesn't currently.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1435,6 +1441,35 @@ private function finalize_styles() {
...
+		// @todo What about amp-anim, amp-playbuzz?
```
westonruter: I've got an idea I'm working on for this…

westonruter:
@miina there, I feel pretty happy with 7a218aa as something to iterate on. I left some todos for some additional changes, other than the ones you already have outstanding.

includes/templates/class-amp-content-sanitizer.php
```diff
@@ -96,6 +116,11 @@ public static function sanitize_document( &$dom, $sanitizer_classes, $args ) {
 			AMP_Response_Headers::send_server_timing( 'amp_sanitize', -$sanitize_class_start, $sanitizer_class );
 		}
 
+		// Let the sanitizers.
+		foreach ( $sanitizers as $sanitizer ) {
+			$sanitizer->after_sanitize( $sanitizers );
+		}
```
westonruter: Since we're not using this, I think we should just kill `after_sanitize`.

includes/templates/class-amp-content-sanitizer.php
```diff
@@ -84,6 +92,18 @@ public static function sanitize_document( &$dom, $sanitizer_classes, $args ) {
 				continue;
 			}
 
+			$sanitizers[ $sanitizer_class ] = $sanitizer;
+		}
+
+		// Let the sanitizers know about each other prior to sanitizing.
+		foreach ( $sanitizers as $sanitizer ) {
+			$sanitizer->before_sanitize( $sanitizers );
```
westonruter: I do think we should rename this to `init`

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1754,6 +1789,40 @@ private function finalize_styles() {
...
+	 * Convert CSS selectors.
+	 *
+	 * @param DeclarationBlock $ruleset Ruleset.
+	 */
+	private function ampify_ruleset_selectors( $ruleset ) {
+		$selectors    = array();
+		$replacements = 0;
+		foreach ( $ruleset->getSelectors() as $old_selector ) {
+			$edited_selectors = array( $old_selector->getSelector() );
+			foreach ( $this->selector_mappings as $html_selector => $amp_selectors ) { // Note: The $selector_mappings array contains ~6 items.
```
westonruter: These notes are here because 4 level of loop nesting is normally something that is a big red flag for performance.

assets/css/amp-default.css
westonruter: Since tree shaking isn't done for attribute values, I suggest the way to fix this is to use an attribute selector like `[src]`.

PR #1181, opened 2018-05-29
Fetch (local) stylesheets with @import instead of removing

### LINKED ISSUE #1091: Fetch stylesheets referenced by @import CSS rules instead of removing them
When an `@import` rule is encountered in a stylesheet it is currently getting, it is currently getting deleted because it is not valid AMP:

https://github.com/Automattic/amp-wp/blob/31016b1cc92f8a2e3b75f57dac034b76efafda2e/includes/sanitizers/class-amp-style-sanitizer.php#L692-L697

However, there is no reason why we have to do this (note the `@todo`'s related to this in the style sanitizer). We can instead rather fetch the contents of the `@import`'ed stylesheet as we're doing for external stylesheets loaded form `link` tags:

https://github.com/Automattic/amp-wp/blob/31016b1cc92f8a2e3b75f57dac034b76efafda2e/includes/sanitizers/class-amp-style-sanitizer.php#L440-L456

It is _import_-ant to support `@import` because it used to be the recommended way for child themes to include the `style.css` from the parent theme.

Note that we can also support loading `@import`'ed stylesheets from external domains (and not the filesystem) which will also be done for `link` tags in #1083.

miina:
Adds support for `@import` stylesheets. Local stylesheets are inlined, and external stylesheets are reported as sanitizeable validation errors.

Fixes #1091.

### DIFF
--- a/tests/test-amp-style-sanitizer.php
+++ b/tests/test-amp-style-sanitizer.php
@@ -801,4 +801,66 @@ class AMP_Style_Sanitizer_Test extends WP_UnitTestCase {
 			$this->assertEmpty( $link );
 		}
 	}
+
+	/**
+	 * Get data for CSS @import test.
+	 *
+	 * @return array
+	 */
+	public function get_css_import_data() {
+		$css_url             = admin_url( 'css/login.css' );
+		$html                = sprintf( '<html><head><link rel="stylesheet" href="%s"></head><style>@import url("https://something.example.com/style.css"); body { color:red; }</style></html>', $css_url ); // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedStylesheet
+		$stylesheet_patterns = array(
+			'/' . preg_quote( 'input[type="checkbox"]:disabled' ) . '/',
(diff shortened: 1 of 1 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -837,10 +879,7 @@ private function process_css_list( CSSList $css_list, $options ) {
 					$css_list->remove( $css_item );
 				}
 			} elseif ( $css_item instanceof Import ) {
-				$validation_errors[] = array(
-					'code'    => 'illegal_css_import_rule',
-					'message' => __( 'CSS @import is currently disallowed.', 'amp' ),
-				);
+				$this->parse_import_stylesheet( $css_item );
```
miina: @westonruter Still WIP but started adding the support for @imported stylesheets, any thought on the approach?

westonruter: Yes, your approach is great. I went ahead and iterated on it because of the changes introduced by the refactored error validation handling.

Issue #1148, opened 2018-05-14
Discontinue use of amp endpoint in favor of query var when amp theme support is present


westonruter:
It is currently very difficult to add dynamic switching between AMP and non-AMP responses in plugins because `is_amp_endpoint()` has to be called after the `parse_query` action. This means littering `is_amp_endpoint()` calls into the functions that run during the generation of the page, including enqueueing scripts, generating shortcodes, and rendering widgets. The use of the endpoint is also troublesome with flushing the rewrite rules (e.g. #801). Also, the `/amp/` endpoint is only eligible for being used on non-hierarchical endpoints. As noted in a [support topic](https://wordpress.org/support/topic/0-6-0-not-working-for-pages/#post-9903766):

> The use of the `?amp` query var is required for pages and other such hierarchical post types in order to prevent ambiguity from happening. Consider you have a website about guitars and you publish pages about your favorite accessories, including which amp you like the most. So you publish a page with the URL path `/guitars/favorites/amp/`. If the `/amp/` endpoint were used for pages, then WordPress would not know whether you are trying to get the AMP version of the Favorites page or the non-AMP version of the page about your favorite amp.
> 
> Here is the logic behind when a query var is chosen instead of an endpoint slug: 
> 
> https://github.com/Automattic/amp-wp/blob/4744f2def2769f497fdb13fd4f1e08620c1a35d2/includes/amp-helper-functions.php#L37-L38
> 
> So you can see that `?amp` is used when pretty permalinks aren’t enabled (of course), when a hierarchical post type is used, or when there are other query vars present.
> 
> We confirmed with the Google search team that using `?amp` is perfectly valid from their side of things.

This is bound up with #945, deprecating the `AMP_QUERY_VAR` constant and `amp_query_var` filter.

When all is said and done, `isset( $_GET['amp'] )` is all that would be needed to check if the AMP version is being requested. Note that if `amp` is not available for a given URL, then a redirect to the non-AMP version of the page will still be done.

westonruter:
Aside, this is how to prevent using the `amp` endpoint today:
```php
// Force using query var instead of endpoint.
add_filter( 'amp_pre_get_permalink', function( $pre, $post_id ) {
	unset( $pre );
	return add_query_arg( amp_get_slug(), '', get_permalink( $post_id ) );
}, 10, 2 );
```

mor10:
tracking for future update of WP Rig.

westonruter:
[...]

Issue #1006, opened 2018-03-09
[Compatibility Tool] Improve experience for user based on AMP compatibility


postphotos:
**As a WordPress site, my Compatibility Tool should give visual context to the validation data. A user should be able to understand this data and understand how to action what's being displayed.**

- [x] AC1: Assess and outline all output from the Compatability Tool.
- [x] AC2: ~Create a new experience flow for this that more elegantly allows a user to understand errors and issues~: Closed in favor subticket #1359 
- [ ] AC3: Write steps for the user (i.e. documentation for the product site, wiki, etc.) to be able to resolve any issues, based on final implementation of design.

While there's a decent reporting mechanism currently built for this plugin, it'd be great if, beyond output, if something was more contextual to non-technical users.
https://files.slack.com/files-pri/T02UB976M-F9N2ADXQF/image.png

postphotos:
The main issue description has been updated with a formal Story and Acceptance Criteria.

kienstra:
**Question About Design**

Hi @westonruter,
It sounds like we're going to ask @jwold to work on the design of the 'Compatibility Tool.' Should I update him on the current state of the 'Compatibility Tool,' maybe with a screencast or a detailed comment here?

westonruter:
@kienstra yes, that sounds good. I think he's seen the screenshots already to a degree. But having a screencast would be helpful.

kienstra:
**Request For Design Work, Update On Current State**

Hi @jwold,
Are you available to work on the design of the 'Compatibility Tool' (AC2 on this issue)? This [screencast](https://drive.google.com/file/d/14IPryAcCLQZHlpXpN9hOvJ3dw9BNwlGG/view) shows the current state, using the [dev site](https://dev-ampconfdemo.pantheonsite.io). 

Also, [this 'Steps To Test' comment](https://github.com/Automattic/amp-wp/issues/842#issuecomment-371644473) describes more 'Recheck' user flows.

postphotos:
Hi @kienstra - thank you for this. I just watched your screencast.

A request and a few questions regarding this component (cc: @jwold):
1) Can you (as a `gist` perhaps, or in a throwaway repo?) share on that `AMP-invalid` plugin?

2) Thinking about desired functionality for MVP here, the Validation Status page is indeed confusing.

[...]

PR #1183, opened 2018-05-30
Test AMP compatibility of entire site

### LINKED ISSUE #1007: [Compatibility Tool] Better tooling for evaluating plugins
**As a WordPress site, my Compatibility Tool should have better tooling for evaluating plugins' AMP compatibility programmatically.**

- [ ] AC1: Currently, AMP validation only occurs on plugin activation and as a call inside WP CRON: This story addresses user experience. We would like to additionally allow a site admin to manually trigger a check of the plugins' AMP compatibility ~inside the user interface from #1006~ *via WP-CLI*.
- [ ] AC2: A validation test should be able to evaluate an entire site even if AMP theme support is not declared yet. (i.e. Be able to force the equivalent of AMP theme support when crawling.)
- [ ] AC3: The compatibility tool should be able to target the various templates that now are presented to the user (as as described by @westonruter [below](https://github.com/Automattic/amp-wp/issues/1007#issuecomment-403096232), related #1235.)

The goal is to spawn validation process and get results programmatically for an entire site. WP-Cron and WP-CLI are in view here, as is integrating with logic in XML Sitemap generators to obtain list of URLs potentially.

kienstra:
**Pull Request For #1007**

This crawls the site, and test for AMP validity by template/content type.

- [x] Validate the homepage, as the existing query won't include it if it's not a static page.
- [x] Should this also validate the URLs of [categories](https://codex.wordpress.org/Function_Reference/_get_category_link) and [tags](https://codex.wordpress.org/Function_Reference/get_tag_link)?
- [x] Consider whether reusing [AMP_Validation_Manager::validate_queued_posts_on_frontend()](https://github.com/Automattic/amp-wp/blob/6ec7e1b44b01aae2ae1d49dd062f39100f985f53/includes/validation/class-amp-validation-manager.php#L247) is the best way way to validate these URLs. It will probably need to be changed to batch the requests to the front-end, and I'm not sure if that will interfere with the existing functionality. 
- [x] Consider whether this really needs its own class. At the moment, it's probably not big enough to need it.

Fixes #1007.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -550,7 +550,7 @@ function amp_print_analytics( $analytics ) {
 	// Can enter multiple configs within backend.
 	foreach ( $analytics_entries as $id => $analytics_entry ) {
 		if ( ! isset( $analytics_entry['type'], $analytics_entry['attributes'], $analytics_entry['config_data'] ) ) {
-			/* translators: %1$s is analytics entry ID, %2$s is actual entry keys. */
+			/* translators: 1: the analytics entry ID, 2: comma-separated list of the actual entry keys. */
 			_doing_it_wrong( __FUNCTION__, sprintf( esc_html__( 'Analytics entry for %1$s is missing one of the following keys: `type`, `attributes`, or `config_data` (array keys: %2$s)', 'amp' ), esc_html( $id ), esc_html( implode( ', ', array_keys( $analytics_entry ) ) ) ), '0.3.2' );
 			continue;
 		}
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -287,13 +287,13 @@ function is_amp_endpoint() {
 		return false;
 	}
 
-	/**
+	/*
 	 * If this is a URL for validation, and validation is forced for all URLs, return true.
 	 * Normally, this would be false if the user has deselected a template,
 	 * like by unchecking 'Categories' in 'AMP Settings' > 'Supported Templates'.
 	 * But there's a flag for the WP-CLI command that sets this query var to validate all URLs.
 	 */
-	if ( isset( $_GET[ AMP_Validation_Manager::VALIDATE_QUERY_VAR ], $_GET[ AMP_CLI::FORCE_VALIDATION_QUERY_VAR ] ) ) { // WPCS: CSRF OK.
+	if ( AMP_Validation_Manager::is_theme_support_forced() ) {
 		return true;
 	}
 
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -293,7 +293,7 @@ function is_amp_endpoint() {
 	 * like by unchecking 'Categories' in 'AMP Settings' > 'Supported Templates'.
 	 * But there's a flag for the WP-CLI command that sets this query var to validate all URLs.
 	 */
-	if ( isset( $_GET[ AMP_Validation_Manager::VALIDATE_QUERY_VAR ], $_GET[ AMP_Site_Validation::FORCE_VALIDATION_QUERY_VAR ] ) ) { // WPCS: CSRF OK.
+	if ( isset( $_GET[ AMP_Validation_Manager::VALIDATE_QUERY_VAR ], $_GET[ AMP_CLI::FORCE_VALIDATION_QUERY_VAR ] ) ) { // WPCS: CSRF OK.
 		return true;
 	}
 
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -287,6 +287,16 @@ function is_amp_endpoint() {
 		return false;
 	}
 
+	/**
+	 * If this is a URL for validation, and validation is forced for all URLs, return true.
+	 * Normally, this would be false if the user has deselected a template,
+	 * like by unchecking 'Categories' in 'AMP Settings' > 'Supported Templates'.
+	 * But there's a flag for the WP-CLI command that sets this query var to validate all URLs.
+	 */
+	if ( isset( $_GET[ AMP_Validation_Manager::VALIDATE_QUERY_VAR ], $_GET[ AMP_Site_Validation::FORCE_VALIDATION_QUERY_VAR ] ) ) { // WPCS: CSRF OK.
+		return true;
+	}
+
 	$availability = AMP_Theme_Support::get_template_availability();
 	return amp_is_canonical() ? $availability['supported'] : ( $has_amp_query_var && $availability['supported'] );
 }
--- a/includes/class-amp-cli.php
+++ b/includes/class-amp-cli.php
@@ -119,7 +119,7 @@ class AMP_CLI {
 	 *
 	 * Templates are each a separate type, like those for is_category() and is_tag().
 	 * Also, each post type is a separate type.
-	 * This value is overridden if the WP-CLI command has an --n argument, like --n=10.
+	 * This value is overridden if the WP-CLI command has an --limit argument, like --limit=10.
 	 *
 	 * @var int
 	 */
@@ -142,11 +142,11 @@ class AMP_CLI {
 	 *
 	 * ## OPTIONS
 	 *
-	 * [--limit=<templates>]
+	 * [--limit=<count>]
 	 * : The maximum number of URLs to validate for each template and content type.
-	 * ----
+	 * ---
 	 * default: 100
-	 * ----
+	 * ---
 	 *
 	 * [--include=<templates>]
 	 * : Only validates a URL if one of the conditionals is true.
@@ -183,8 +183,8 @@ class AMP_CLI {
 
 		if ( ! current_theme_supports( 'amp' ) ) {
 			if ( self::$force_crawl_urls ) {
-				/**
-				 * There is no theme support added programatically or via options.
+				/*
+				 * There is no theme support added programmatically or via options.
 				 * So the validation taxonomy and post type would not normally be registered.
 				 * Register them here in order to use them to crawl the site.
 				 */
--- a/includes/class-amp-cli.php
+++ b/includes/class-amp-cli.php
@@ -169,18 +169,14 @@ class AMP_CLI {
 			self::$force_crawl_urls = true;
 		}
 
-		$theme_support_amp = current_theme_supports( 'amp' );
-
-		if ( ! $theme_support_amp ) {
+		if ( ! current_theme_supports( 'amp' ) ) {
 			if ( self::$force_crawl_urls ) {
 				/*
 				 * There is no theme support added programmatically or via options.
-				 * So the validation taxonomy and post type would not normally be registered.
-				 * Register them here in order to use them to crawl the site.
+				 * So make sure that theme support is present so that AMP_Validation_Manager::validate_url()
+				 * will use a canonical URL as the basis for obtaining validation results.
 				 */
 				add_theme_support( 'amp' );
-				AMP_Validation_Error_Taxonomy::register();
-				AMP_Invalid_URL_Post_Type::register();
 			} else {
 				WP_CLI::error(
 					sprintf(
@@ -252,23 +248,19 @@ class AMP_CLI {
 		);
 
 		// Output a table of validity by template/content type.
-		if ( $theme_support_amp ) {
-			$url_more_details = add_query_arg(
-				'post_type',
-				AMP_Invalid_URL_Post_Type::POST_TYPE_SLUG,
-				admin_url( 'edit.php' )
-			);
-			/* translators: %s is the URL to the admin */
-			WP_CLI::line( sprintf( __( 'For more details, please see: %s', 'amp' ), $url_more_details ) );
-		} else {
-			WP_CLI::warning( __( 'You cannot currently access the detailed validation results because you have not switched to native or paired AMP template modes.', 'amp' ) );
-		}
-
 		WP_CLI\Utils\format_items(
 			'table',
 			$table_validation_by_type,
 			array( $key_template_type, $key_url_count, $key_validity_rate )
 		);
+
+		$url_more_details = add_query_arg(
+			'post_type',
+			AMP_Invalid_URL_Post_Type::POST_TYPE_SLUG,
+			admin_url( 'edit.php' )
+		);
+		/* translators: %s is the URL to the admin */
+		WP_CLI::line( sprintf( __( 'For more details, please see: %s', 'amp' ), $url_more_details ) );
 	}
 
 	/**
--- a/includes/class-amp-cli.php
+++ b/includes/class-amp-cli.php
@@ -57,7 +57,7 @@ class AMP_CLI {
 	 *
 	 * @var string
 	 */
-	const MAXIMUM_URLS_ARGUMENT = 'limit';
+	const LIMIT_URLS_ARGUMENT = 'limit';
 
 	/**
 	 * The WP CLI progress bar.
@@ -123,7 +123,7 @@ class AMP_CLI {
 	 *
 	 * @var int
 	 */
-	public static $maximum_urls_to_validate_for_each_type = 100;
+	public static $limit_type_validate_count;
 
 	/**
 	 * The validation counts by type, like template or post type.
@@ -144,6 +144,9 @@ class AMP_CLI {
 	 *
 	 * [--limit=<templates>]
 	 * : The maximum number of URLs to validate for each template and content type.
+	 * ----
+	 * default: 100
+	 * ----
 	 *
 	 * [--include=<templates>]
 	 * : Only validates a URL if one of the conditionals is true.
@@ -162,6 +165,9 @@ class AMP_CLI {
 	 */
 	public function validate_site( $args, $assoc_args ) {
 		unset( $args );
+		self::$include_conditionals      = array();
+		self::$force_crawl_urls          = false;
+		self::$limit_type_validate_count = (int) $assoc_args[ self::LIMIT_URLS_ARGUMENT ];
 
 		/*
 		 * Handle the argument and flag passed to the command: --include and --force.
@@ -173,10 +179,6 @@ class AMP_CLI {
 			self::$force_crawl_urls     = true;
 		} elseif ( isset( $assoc_args[ self::FLAG_NAME_FORCE_VALIDATION ] ) ) {
 			self::$force_crawl_urls = true;
-		}
-
-		if ( ! empty( $assoc_args[ self::MAXIMUM_URLS_ARGUMENT ] ) ) {
-			self::$maximum_urls_to_validate_for_each_type = intval( $assoc_args[ self::MAXIMUM_URLS_ARGUMENT ] );
 		}
 
 		$number_urls_to_crawl = self::count_urls_to_validate();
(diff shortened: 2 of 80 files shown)


includes/validation/class-amp-site-validation.php
```diff
@@ -0,0 +1,50 @@
...
+	 * @todo: Consider wrapping this function with another, as different use cases will probably require a different return value or display.
+	 * For example, the <button> in /wp-admin that makes an AJAX request for this will need a different response than a WP-CLI command.
+	 * @return array $validation_result The post ID as the index, and the result of validation as the value.
+	 */
+	public static function validate_site_urls() {
+		$site_post_ids = self::get_post_ids();
+		foreach ( $site_post_ids as $id ) {
+			AMP_Validation_Manager::$posts_pending_frontend_validation[ $id ] = true;
+		}
+		return AMP_Validation_Manager::validate_queued_posts_on_frontend();
```
westonruter: As you suggested, this `validate_queued_posts_on_frontend` method is probably not relevant. The main thing will be to:
1. Obtain the list of URLs on the site, including homepage, posts, pages, categories, archives, etc. A good sitemap XML generator plugin should have the logic you need for this.
2. If in paired mode, append `?amp` to each of the URLs discovered. This will be used instead of `amp_get_permalink()` because the endpoint will be eliminated when AMP theme support is present. See #1148. Some consideration in paired mode may be needed for whether AMP is available or not for a given post.
3. For each AMP URL, call `\AMP_Validation_Manager::validate_url()` and pass the results into `\AMP_Invalid_URL_Post_Type::store_validation_errors()`.
4. Chunk the results to prevent timing out. This is particularly key in WP-Cron and Ajax. In WP-CLI this isn't relevant, and a progress meter should be shown.

kienstra: Thanks for your fast reply, @westonruter! These details really help.

includes/validation/class-amp-site-validation.php
```diff
@@ -0,0 +1,76 @@
...
+	 * Gets the front-end links for public taxonomy terms, like categories and tags.
+	 *
+	 * For example, https://example.org/?cat=2
+	 * This includes categories and tags, and any more that are registered.
+	 * But it excludes post_format terms.
+	 *
+	 * @param int $number_links The maximum amount of links to get (optional).
+	 * @return string[] $links The term links in an array.
+	 */
+	public static function get_taxonomy_links( $number_links = 200 ) {
```
westonruter: There should be pagination support here. The `WP_Term_Query` supports an `$offset` arg. I suggest that this method only return links for one specific taxonomy. The site crawler will probably need to make multiple calls to such a method to iterate over all of the taxonomy terms in batches. One way to do it would be to get the terms sorted by ID in ascending order. That ensures that when you paginate through the terms you won't miss terms that get added while you're processing through them.

includes/validation/class-amp-site-validation.php
```diff
@@ -0,0 +1,76 @@
...
+		$public_taxonomies = get_taxonomies( array(
+			'public' => true,
+		) );
+
+		// It doesn't seem necessary to get links for post format terms, like asides, galleries, or quotes.
```
westonruter: Actually, I think it is necessary. There are archive links for post formats: https://codex.wordpress.org/Function_Reference/get_post_format_link

westonruter:
I think the “Check Site” button would make sense on the “Invalid AMP Pages” screen. Clicking it should show a spinner and some progress indicator probably. 

And as noted on https://github.com/Automattic/amp-wp/issues/1184#issuecomment-393256957:

> I've been considering that we should add a new AMP admin screen to manage settings regarding theme support, including:
> 
> * Forcibly enable `amp` theme support for the current theme, even if it lacks `add_theme_support('amp')`. This will be particularly useful as the plugin will support all the core themes in AMP by default, and it would allow 
> * Decide whether to have paired mode.
> * Preemptively opt-in to tree shaking, since it is going to be needed in the vast majority of cases anyway. Options here could be “Ask me” (block AMP without accepting), “As needed” (sometimes), and “Always”.

Having a “Check Site Compatibility” button would make a lot of sense on such a Theme Support admin screen. As part of this, we should allow for the `amp_validate` query parameter to force the plugin to behave as if `amp` theme support is present. In that way, a site could be checked for AMP compatibility before even enabling `amp` theme support via an opt-in toggle on that admin screen.

includes/validation/class-amp-site-validation.php
kienstra: Thanks, commit [db54a](https://github.com/Automattic/amp-wp/pull/1183/commits/db54a537a5267eb49c6a80cbbdaeb1b3475ab091) removes the exemption of `post_format` terms.

includes/validation/class-amp-site-validation.php
kienstra: Thanks, good ideas. [These commits](https://github.com/Automattic/amp-wp/compare/db54a537a5267eb49c6a80cbbdaeb1b3475ab091...26553102bc9a2f02e58b80bb9ce54159b67e6ae6) apply your suggestions, as I understand them. This now has `$taxonomy` and `$offset` parameters.

kienstra:
**[WIP] WP-CLI Output**
When you run `wp eval-file bin/validate-site.php`

wp-cli-output

And when it finishes:
p-cli-after

westonruter:
That UI for WP-CLI looks perfect 👌

bin/validate-site.php
```diff
@@ -0,0 +1,60 @@
...
+				$url_more_details
+			)
+		);
+	} catch ( \Exception $e ) {
+		\WP_CLI::error( $e->getMessage() );
+	}
+} else {
+	echo "Please run this script with WP-CLI via: wp eval-file bin/validate-site.php\n";
+	exit( 1 );
+}
```
westonruter: Instead of adding this as a `bin` script it should get registered as an actual command with WP-CLI. For example, `wp amp validate-site`. The command needs to be included with the plugin when it is released, and the `bin` directory is just for development.

includes/validation/class-amp-site-validation.php
```diff
@@ -0,0 +1,154 @@
...
+			$offset     = 0;
+
+			while ( ! empty( $permalinks ) ) {
+				self::validate_urls( $permalinks );
+				$offset    += self::BATCH_SIZE;
+				$permalinks = self::get_post_permalinks( $post_type, self::BATCH_SIZE, $offset );
+			}
+
+			if ( $wp_cli_progress ) {
+				$wp_cli_progress->tick();
```
westonruter: This could get de-coupled from WP-CLI by passing a `$progress_callback` function instead which is called, and then you could call the method like so:

```php
\AMP_Site_Validation::validate_entire_site_urls( function() use ( $wp_cli_progress ) {
    $wp_cli_progress->tick()
} ) 
```

bin/validate-site.php
```diff
@@ -0,0 +1,60 @@
...
+ * Crawl the entire site to validate it, and get the results.
+ *
+ * @return array[] {
+ *     @type int  $number_crawled The number of URLs visited, regardless of the validation results.
+ *     @type int  $number_invalid The number of URLs that have AMP validation error(s).
+ * }
+ */
+function crawl_site() {
+	\WP_CLI::log( 'Crawling the entire site to test for AMP validity. This might take a while...' );
+	$count_post_types_and_taxonomies = count( get_post_types( array( 'public' => true ), 'names' ) ) + count( get_taxonomies( array( 'public' => true ) ) );
```
westonruter: This should count the number of posts and terms, not just the number of types.

includes/validation/class-amp-site-validation.php
```diff
@@ -0,0 +1,154 @@
...
+		self::validate_urls( home_url( '/' ) );
+
+		// Validate all public, published posts.
+		$public_post_types = get_post_types( array( 'public' => true ), 'names' );
+		foreach ( $public_post_types as $post_type ) {
+			$permalinks = self::get_post_permalinks( $post_type, self::BATCH_SIZE );
+			$offset     = 0;
+
+			while ( ! empty( $permalinks ) ) {
+				self::validate_urls( $permalinks );
```
westonruter: Per above, the tick should be done with each URL that is processed, not after each type is processed.

kienstra:
Hi @westonruter,
Thanks for reviewing this. You make good points, like registering an actual WP-CLI command. 

If it's alright, I'm going to leave this work for the next phase.

kienstra:
**Demo Screencast**

Here's a [screencast](https://drive.google.com/file/d/1G_OATyt-_SWXmXs6LKOurEnKp05WQzjq/view?usp=sharing) of the current work in progress of the WP-CLI command.

kienstra:
**WP-CLI**

Here's the current output from the new WP-CLI command of `wp amp validate-site`

amp-wp-cli-result

Of course, this is open for discussion. I though it would be good to also show the amount of unaccepted errors.

includes/validation/class-amp-site-validation.php
```diff
@@ -0,0 +1,279 @@
...
+ */
+class AMP_Site_Validation {
+
+	/**
+	 * The size of the batch of URLs to validate.
+	 * Grouped in a batch to avoid making loopback requests for all of a site's URLs at the same time.
+	 *
+	 * @var int
+	 */
+	const BATCH_SIZE = 100;
```
kienstra: This `BATCH_SIZE` is probably only appropriate for this WP-CLI command. A possible `/wp-admin` button that crawls the site would probably need a lower value.

includes/validation/class-amp-site-validation.php
```diff
@@ -0,0 +1,279 @@
...
+			WP_CLI::add_command( 'amp', array( __CLASS__, 'crawl_site' ) );
+		}
+	}
+
+	/**
+	 * Crawls the entire site to validate it, and gets the results.
+	 *
+	 * @param array $args The arguments for the command.
+	 */
+	public function crawl_site( $args ) {
```
kienstra: This and `init()` are the only methods in the class that aren't tested, as they depend on the class `WP_CLI`.

includes/validation/class-amp-site-validation.php
```diff
@@ -0,0 +1,279 @@
...
+			WP_CLI::add_command( 'amp', array( __CLASS__, 'crawl_site' ) );
+		}
+	}
+
+	/**
+	 * Crawls the entire site to validate it, and gets the results.
+	 *
+	 * @param array $args The arguments for the command.
+	 */
+	public static function crawl_site( $args ) {
```
kienstra: This and `init()` are the only methods in the class that aren't tested, as they depend on the class `WP_CLI`.

includes/validation/class-amp-site-validation.php
```diff
@@ -0,0 +1,279 @@
...
+
+		/**
+		 * Because of the posts_per_page => -1 value, this is only suited for use in WP-CLI.
+		 * This is necessary to pass as an argument to WP_CLI\Utils\make_progress_bar(),
+		 * in order to show the progress of every post.
+		 */
+		$post_query   = new WP_Query( array(
+			'post_type'      => $public_post_types,
+			'fields'         => 'ids',
+			'posts_per_page' => -1,
```
kienstra: `'posts_per_page' => -1,` is a little troubling, as it could crash this when a site has tens or hundreds of thousands of posts.

westonruter: Indeed. It should do batches of posts at 100 per query, and keep querying while incrementing the page until there are none left. I suggest sorting posts by ID in ascending order, as that will ensure newly published posts don't result in accidental skipping of posts.

kienstra: Thanks, good idea.

kienstra:
@ThierryA made a good point that this WP-CLI script shouldn't crawl endpoints that the user has chosen to not display as AMP.

For example, if the content type or template is unchecked in "Supported Templates":

should-not-crawl

westonruter:
> this WP-CLI script shouldn't crawl endpoints that the user has chosen to not display as AMP.

That is a good point. That should be the default behavior to not check templates that opt-out of AMP. However, there should be a flag to override this in case the user wants in the future to re-enable AMP for those templates and they want to check compatibility. Additionally, there could be a way to invoke the command with a whitelist of specific templates to check, like `--include=is_singular,is_category`.

kienstra:
**Thanks!**

Hi @westonruter,
Thanks a lot for your work to make the Travis build pass.

kienstra:
There's now a display of validation results by type:
validation-results-by-type

kienstra:
**Table Output**

Thanks to @westonruter's suggestion, this command now outputs a table:
table

includes/amp-helper-functions.php
```diff
@@ -287,6 +287,16 @@ function is_amp_endpoint() {
...
 	}
 
+	/**
+	 * If this is a URL for validation, and validation is forced for all URLs, return true.
+	 * Normally, this would be false if the user has deselected a template,
+	 * like by unchecking 'Categories' in 'AMP Settings' > 'Supported Templates'.
+	 * But there's a flag for the WP-CLI command that sets this query var to validate all URLs.
+	 */
+	if ( isset( $_GET[ AMP_Validation_Manager::VALIDATE_QUERY_VAR ], $_GET[ AMP_Site_Validation::FORCE_VALIDATION_QUERY_VAR ] ) ) { // WPCS: CSRF OK.
+		return true;
```
kienstra: This might not be the best approach.

kienstra:
**Example Command**

```bash
wp amp validate-site --n=5 --force-validation
```

`--n=5`
makes this validate only 5 URLs for each template or post type. Otherwise, this will take much longer to run.

`--force-validation`
 makes this ignore the options in in 'AMP Settings' > 'Supported Templates,' and validate templates even if they may be unchecked

`--include=is_tag,is_author` 
Validates only tag and author templates

kienstra:
**Request For Review**

Hi @westonruter,
Thanks for all of your help on this PR.

Could you please review this again? Like we talked about, it now gets the AMP validity of the site [by template and content type](https://github.com/Automattic/amp-wp/pull/1183#issuecomment-414783863).

When testing this, passing `--n=5` will probably help. Otherwise, this [will validate 100 URLs](https://github.com/Automattic/amp-wp/blob/907facd78848a8bff643fb20479a205b2091daca/includes/validation/class-amp-site-validation.php#L141) for each template/content type. We might also change that default value.

```bash
$ wp amp validate-site --n=5 --force-validation
```

Also, this doesn't register the command at all in 'Classic' mode, as the [class is bootstrapped](https://github.com/Automattic/amp-wp/blob/907facd78848a8bff643fb20479a205b2091daca/includes/validation/class-amp-validation-manager.php#L160) in `AMP_Validation_Manager`. Should that change?

Thanks, Weston!

westonruter:
> Also, this doesn't register the command at all in 'Classic' mode, as the class is bootstrapped in AMP_Validation_Manager. Should that change?

That's a good point. Yes, we should register the WP-CLI command unconditionally. In the future there will be other commands we'll introduce that are independent of validation, like switching between template the modes. So there could be a general `AMP_CLI`class that is used instead of `AMP_Site_Validation`, and most of the validation-specific helper methods of `AMP_Site_Validation` could go over to `AMP_Validation_Manager`.

kienstra:
Hi @westonruter,
That's a good point, and thanks for the steps to validate in 'Classic' mode. I'll work on that today if that's alright.

kienstra:
Hi @westonruter, the [2 commits above](https://github.com/Automattic/amp-wp/compare/2637fb803d611743d907ddf56e4addf297808f82...bfadbe203ae3607fa6eb24342265497470a4ec42) apply your suggestions for forcing theme support when in Classic mode. I hope they capture what you had in mind.

Here's the error that appears when in 'Classic' mode and `--force` isn't present:

wp-cli-error

includes/class-amp-cli.php
```diff
@@ -0,0 +1,613 @@
...
+		 * The WP-CLI command should indicate which templates are crawled, not the /wp-admin options.
+		 */
+		if ( ! empty( $assoc_args[ self::INCLUDE_ARGUMENT ] ) ) {
+			self::$include_conditionals = explode( ',', $assoc_args[ self::INCLUDE_ARGUMENT ] );
+			self::$force_crawl_urls     = true;
+		} elseif ( isset( $assoc_args[ self::FLAG_NAME_FORCE_VALIDATION ] ) ) {
+			self::$force_crawl_urls = true;
+		}
+
+		if ( ! current_theme_supports( 'amp' ) ) {
```
kienstra: This conditional will capture the Template Mode being Classic, though maybe you had in mind a more narrow check, like:
```php
'disabled' === AMP_Options_Manager::get_option( 'theme_support' )
```

westonruter:
The last piece here seemed to me to be providing a way for the user to access the validation results for their `wp amp validate-site` call even while AMP theme support is turned off. This is implemented in edcb23ae, where the post type and taxonomy are unconditionally registered but they are only shown in the admin menu if theme support is present. Otherwise, links to the screens are shown if there are invalid URL posts that exist:

image

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -109,6 +109,22 @@ public static function register() {
...
+	 * Determine whether the admin menu item should be included.
+	 *
+	 * @return bool Whether to show in menu.
+	 */
+	public static function should_show_in_menu() {
+		global $pagenow;
+		if ( current_theme_supports( 'amp' ) ) {
+			return true;
+		}
+		if ( 'edit.php' === $pagenow && ( isset( $_GET['post_type'] ) && self::POST_TYPE_SLUG === $_GET['post_type'] ) ) { // WPCS: CSRF OK.
```
hellofromtonya: Why not return the result of the conditional instead of explicitedly retuning `true` or `false`?

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -162,6 +162,22 @@ public static function register() {
...
+	 * Determine whether the admin menu item should be included.
+	 *
+	 * @return bool Whether to show in menu.
+	 */
+	public static function should_show_in_menu() {
+		global $pagenow;
+		if ( current_theme_supports( 'amp' ) ) {
+			return true;
+		}
+		if ( 'edit-tags.php' === $pagenow && ( isset( $_GET['taxonomy'] ) && self::TAXONOMY_SLUG === $_GET['taxonomy'] ) ) { // WPCS: CSRF OK.
```
hellofromtonya: Same here....return the result of the conditional expression instead of explicit `true`or `false`.  It’s less code and more clear.

includes/validation/class-amp-invalid-url-post-type.php
westonruter: Sure. Done in a83a569f.

includes/class-amp-cli.php
```diff
@@ -0,0 +1,608 @@
...
+
+		if ( empty( $table_validation_by_type ) ) {
+			WP_CLI::error( __( 'No validation results were obtained from the URLs.', 'amp' ) );
+			return;
+		}
+
+		WP_CLI::success(
+			sprintf(
+				/* translators: $1%d is the number of URls crawled, $2%d is the number of validation issues, $3%d is the number of unaccepted issues, $4%s is the list of validation by type, $5%s is the link for more details */
+				__( '%3$d crawled URLs have unaccepted issue(s) out of %2$d total which AMP validation issue(s); %1$d URLs were crawled.', 'amp' ),
```
kienstra: Maybe
>total which AMP validation issue(s)

should be:

>total with AMP validation issue(s)

kienstra:
**Commands Work As Expected**

Hi @westonruter,
The WP-CLI commands here work as expected. Including:

`wp amp validate-site`

`wp amp validate-site --include=is_author`

`wp amp validate-site --force` (when in Classic mode)

`wp amp validate-site --limit=2`

limit-2

Also, great idea to add links for the invalid URLs and errors in Classic mode:

classic-validation

postphotos:
Hi @westonruter - really like the progress [here](https://github.com/Automattic/amp-wp/pull/1183#issuecomment-416783852) and thanks for tagging this in #1359. 

Also wanted to call out #1273 here - the idea I spelled out there could try to surface errors next to where toggles exist (so that a user is more likely to select modes when they might be afraid to do this). I'm not entirely sure this is necessary or how straightforward implementing it might be, but it's captured in the backlog regardless.

Both that feature (and #1273) are trying to take an approach to accomplish the inverse goals of #1366: Instead of surfacing settings info in the error listings, surface the error listings in the settings. 

The key decisions I still think are up for discussion (past `v1.0-stable`) will reside around how much information a user should see and in what context that's necessary, but based on the work presented in #1006 I'm happy with the current efforts. Thanks for your help here!

cc: @jwold

PR #1188, opened 2018-05-31
Add/1096 convert css original child

DavidCramer:
@miina this solves the issue of the child element from being removed.

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -190,6 +190,13 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	private $selector_mappings = array();
 
 	/**
+	 * Original element replaced and expanded to the amp equivalent child.
+	 *
+	 * @var array
+	 */
+	private $replaced_internal_selector = array();
+
+	/**
 	 * Get error codes that can be raised during parsing of CSS.
 	 *
 	 * This is used to determine which validation errors should be taken into account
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1808,10 +1813,13 @@ private function ampify_ruleset_selectors( $ruleset ) {
 					if ( ! $count ) {
 						continue;
 					}
+					$selectors[]   = $edited_selector . '>' . $html_selector;
 					$replacements += $count;
 					while ( ! empty( $amp_selectors ) ) { // Note: This array contains only a couple items.
 						$amp_selector       = array_shift( $amp_selectors );
-						$edited_selectors[] = preg_replace( $html_pattern, $amp_selector, $original_selector, -1, $count );
+						$edited_selector    = preg_replace( $html_pattern, $amp_selector, $original_selector, -1, $count );
+						$edited_selectors[] = $edited_selector;
```
miina: Started thinking that maybe we should only replace it with `amp-img>img` and not `img` at all (same with others)? Adding two selectors would add the same style to both `amp-img` and to `img`, could that cause issues? Also, wondering if in any case we would always need the style for the child element (`img`) and not the `amp-img` at all. Would we need it for the `amp-img` in some cases? Thoughts?

In this case we could actually just replace the child selector rules, this would also solve the issue for `AMP_Playbuzz` where `amp-playbuzz>.pb_feed` would never exist.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -311,6 +311,11 @@ private function get_used_tag_names() {
 				$used_tag_names[ $el->tagName ] = true;
 			}
 			$this->used_tag_names = array_keys( $used_tag_names );
+			foreach ( $this->selector_mappings as $html_tag => $amp_tags ) {
+				if ( 0 < count( array_intersect( $this->used_tag_names, $amp_tags ) ) ) {
+					$this->used_tag_names[] = $html_tag;
```
westonruter: Note that `selector_mappings` does not consist of tag names but rather selector terms, for example `.pb_feed`.

westonruter:
Thanks for identifying this issue. I've pushed what I think is a more robust solution in https://github.com/Automattic/amp-wp/pull/1175/commits/91e3eef4cd601421a429589e65299ae4f9200f69

PR #1190, opened 2018-06-01
Improve handling of validation errors in response cache and admin display

westonruter:
* Ensure response cache is varied by validation error sanitization status. If the response had been cached when all validation errors had marked as rejected (not-sanitized), then when one or more of those validation errors have their status changed to accepted (sanitized) then needs to make sure that the response cache will be updated. Otherwise, a validation error's status could change but no change would be reflected on the frontend.
* Improve searchability of validation errors in invalid AMP URL posts by storing full text of validation error in the `post_content`, instead of just storing the `sources`. Note that this means the validation error data is duplicated in the `amp_validation_error` term `description` as well as in the `post_content` of the `amp_invalid_url` post type. This is not concerning because the size is small, the data will not be updated once written, and denormalization improves performance for searching.
* Improve the “incompatible sources” summary so it shows more than just “core: wp-includes”.
* Account for `amp_validation_error_sanitized filter` when displaying validation errors in admin. (And let second filter argument be flatter as just a `$error` array, instead of `$context` containing `$error`.) Prevent offering to change validation error status when filter is forcing a certain status. For example, if I want to make sure that `invalid_attribute` validation errors are always sanitized, I could add this:

```php
add_filter( 'amp_validation_error_sanitized', function( $sanitized, $error ) {
	if ( $error['code'] === 'invalid_attribute' ) {
		$sanitized = true;
	}
	return $sanitized;
}, 10, 2 );
```

In the UI now, when looking at an invalid AMP URL post, this is indicated:

image

Likewise, in the list of validation errors, the checkboxes for doing bulk actions on such sanitization-forced errors are removed, as are the row actions:

image

### DIFF
--- a/includes/validation/class-amp-invalid-url-post-type.php
+++ b/includes/validation/class-amp-invalid-url-post-type.php
@@ -198,20 +198,12 @@ class AMP_Invalid_URL_Post_Type {
 				continue;
 			}
 			$term = get_term_by( 'slug', $stored_validation_error['term_slug'], AMP_Validation_Error_Taxonomy::TAXONOMY_SLUG );
-			if ( ! $term ) {
-				continue;
-			}
-			if ( $args['ignore_accepted'] && AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === $term->term_group ) {
+			if ( $term && $args['ignore_accepted'] && AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === $term->term_group ) {
 				continue;
 			}
 			$errors[] = array(
 				'term' => $term,
-				'data' => array_merge(
-					json_decode( $term->description, true ),
-					array(
-						'sources' => $stored_validation_error['sources'],
-					)
-				),
+				'data' => $stored_validation_error['data'],
 			);
 		}
 		return $errors;
@@ -318,15 +310,6 @@ class AMP_Invalid_URL_Post_Type {
 
 		$terms = array();
 		foreach ( $validation_errors as $data ) {
-			/*
-			 * Exclude sources from data since not available unless sources are being obtained,
-			 * and thus not able to be matched when hashed.
-			 */
-			$sources = null;
-			if ( isset( $data['sources'] ) ) {
-				$sources = $data['sources'];
-			}
-
 			$term_data = AMP_Validation_Error_Taxonomy::prepare_validation_error_taxonomy_term( $data );
 			$term_slug = $term_data['slug'];
 			if ( ! isset( $terms[ $term_slug ] ) ) {
@@ -352,7 +335,7 @@ class AMP_Invalid_URL_Post_Type {
 				$terms[ $term_slug ] = $term;
 			}
 
-			$stored_validation_errors[] = compact( 'term_slug', 'sources' );
+			$stored_validation_errors[] = compact( 'term_slug', 'data' );
 		}
 
 		$post_content = wp_json_encode( $stored_validation_errors );
--- a/includes/validation/class-amp-invalid-url-post-type.php
+++ b/includes/validation/class-amp-invalid-url-post-type.php
@@ -175,14 +175,16 @@ class AMP_Invalid_URL_Post_Type {
 	 * @param array       $args {
 	 *     Args.
 	 *
-	 *     @type bool $ignore_accepted Exclude validation errors that are accepted. Default false.
+	 *     @type bool $ignore_accepted           Exclude validation errors that are accepted. Default false.
+	 *     @type bool $check_forced_sanitization Check whether sanitization status is forced via amp_validation_error_sanitized filter. Adds a check_forced_sanitization value to each returned array.
 	 * }
 	 * @return array List of errors.
 	 */
 	public static function get_invalid_url_validation_errors( $post, $args = array() ) {
 		$args   = array_merge(
 			array(
-				'ignore_accepted' => false,
+				'ignore_accepted'           => false,
+				'check_forced_sanitization' => true,
 			),
 			$args
 		);
@@ -198,13 +200,30 @@ class AMP_Invalid_URL_Post_Type {
 				continue;
 			}
 			$term = get_term_by( 'slug', $stored_validation_error['term_slug'], AMP_Validation_Error_Taxonomy::TAXONOMY_SLUG );
-			if ( $term && $args['ignore_accepted'] && AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === $term->term_group ) {
+
+			$forced_sanitization = null;
+			if ( $args['check_forced_sanitization'] ) {
+				$error = $stored_validation_error['data'];
+				unset( $error['sources'] );
+
+				/** This filter is documented in amp/includes/validation/class-amp-validation-manager.php */
+				$forced_sanitization = apply_filters( 'amp_validation_error_sanitized', $forced_sanitization, $error );
+			}
+
+			if ( $args['ignore_accepted'] && ( true === $forced_sanitization || ( $term && AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === $term->term_group ) ) ) {
 				continue;
 			}
-			$errors[] = array(
+			if ( $term && null !== $forced_sanitization ) {
+				$term->term_group = $forced_sanitization ? AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS : AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_REJECTED_STATUS;
+			}
+			$error = array(
 				'term' => $term,
 				'data' => $stored_validation_error['data'],
 			);
+			if ( $args['check_forced_sanitization'] ) {
+				$error = array_merge( $error, compact( 'forced_sanitization' ) );
+			}
+			$errors[] = $error;
 		}
 		return $errors;
 	}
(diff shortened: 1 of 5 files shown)


westonruter:
Unit tests are still needed for the validation classes, but let's not let that hold this up.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -966,33 +968,49 @@ public static function print_validation_errors_meta_box( $post ) {
 		} );
 		</script>
 
+		<?php
+		$forced_sanitized_count = 0;
```
ThierryA: Since the notice is displayed if `$forced_sanitized_count` is bigger than 0, we could display the notice as soon as a `$validation_error['forced_sanitization']` is found and `break` to avoid unnecessarily continuing the loop.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -966,33 +968,49 @@ public static function print_validation_errors_meta_box( $post ) {
...
+		<?php endif; ?>
+
 		<div class="amp-validation-errors">
 			<ul>
 				<?php foreach ( $validation_errors as $error ) : ?>
 					<?php
 					$collapsed_details = array();
 					$term              = $error['term'];
 					$select_name       = sprintf( '%s[%s]', AMP_Validation_Manager::VALIDATION_ERROR_TERM_STATUS_QUERY_VAR, $term->slug );
+					$term_status       = $term->term_group;
```
ThierryA: Not sure I see the reason to store this in a variable.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -155,11 +155,43 @@ public static function register() {
 			),
 		) );
 
+		add_filter( 'user_has_cap', array( __CLASS__, 'filter_user_has_cap_for_hiding_term_list_table_checkbox' ), 10, 3 );
```
ThierryA: Is this filter reset somewhere?

includes/validation/class-amp-validation-manager.php
```diff
@@ -368,6 +368,44 @@ public static function has_cap() {
...
+	 * @param array $error Validation error.
+	 * @return bool Whether error should be sanitized.
+	 */
+	public static function is_validation_error_sanitized( $error ) {
+		$term_data = AMP_Validation_Error_Taxonomy::prepare_validation_error_taxonomy_term( $error );
+
+		$term = get_term_by( 'slug', $term_data['slug'], AMP_Validation_Error_Taxonomy::TAXONOMY_SLUG );
+		if ( isset( self::$validation_error_status_overrides[ $term_data['slug'] ] ) ) {
+			$sanitized = AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === self::$validation_error_status_overrides[ $term_data['slug'] ];
+		} elseif ( ! empty( $term ) && AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === $term->term_group ) {
```
ThierryA: This could be simplified to:
``` PHP
// ...
else {
	$sanitized = ! empty( $term ) && AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === $term->term_group;
}
```

includes/validation/class-amp-validation-manager.php
```diff
@@ -368,6 +368,44 @@ public static function has_cap() {
...
+		 * validation request and so they are not suitable for plugins to vary
+		 * sanitization by. Note that returning false this indicates that the
+		 * validation error should not be considered a blocker to render AMP.
+		 *
+		 * @since 1.0
+		 *
+		 * @param bool|null $sanitized Whether sanitized, or null if no default predetermined.
+		 * @param array     $error     Validation error being sanitized.
+		 */
+		$sanitized = apply_filters( 'amp_validation_error_sanitized', $sanitized, $error );
```
ThierryA: `return apply_filters( 'amp_validation_error_sanitized', $sanitized, $error );`

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -966,33 +968,49 @@ public static function print_validation_errors_meta_box( $post ) {
 		} );
 		</script>
 
+		<?php
+		$forced_sanitized_count = 0;
+		foreach ( $validation_errors as $validation_error ) {
```
ThierryA: Another approach:
``` PHP
$forced_sanitized_count = array_reduce(
	$validation_errors,
	function( $sum, $validation_error ) {
		return $sum + (int) isset( $validation_error['forced_sanitization'] );
	}
);
```

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -198,21 +200,30 @@ public static function get_invalid_url_validation_errors( $post, $args = array()
...
+				/** This filter is documented in amp/includes/validation/class-amp-validation-manager.php */
+				$forced_sanitization = apply_filters( 'amp_validation_error_sanitized', $forced_sanitization, $error );
 			}
-			if ( $args['ignore_accepted'] && AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === $term->term_group ) {
+
+			if ( $args['ignore_accepted'] && ( true === $forced_sanitization || ( $term && AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === $term->term_group ) ) ) {
 				continue;
 			}
-			$errors[] = array(
+			if ( $term && null !== $forced_sanitization ) {
```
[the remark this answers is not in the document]
westonruter: Yes. `true` means that the validation error is forced to be accepted (sanitized), whereas`false` means that the validation error is forced to be rejected (not sanitized). A `null` value means that there is no forcing because no filter is applying to force the sanitized state.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -198,21 +200,30 @@ public static function get_invalid_url_validation_errors( $post, $args = array()
...
+			if ( $args['check_forced_sanitization'] ) {
+				$error = $stored_validation_error['data'];
+				unset( $error['sources'] );
+
+				/** This filter is documented in amp/includes/validation/class-amp-validation-manager.php */
+				$forced_sanitization = apply_filters( 'amp_validation_error_sanitized', $forced_sanitization, $error );
 			}
-			if ( $args['ignore_accepted'] && AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === $term->term_group ) {
+
+			if ( $args['ignore_accepted'] && ( true === $forced_sanitization || ( $term && AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPTED_STATUS === $term->term_group ) ) ) {
```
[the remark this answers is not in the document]
westonruter: Actually, the `$term->term_group` property is always set because it is defined on the `WP_Term` class:

https://github.com/WordPress/wordpress-develop/blob/a26c24226c6b131a0ed22c722a836c100d3ba254/src/wp-includes/class-wp-term.php#L43-L49

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -198,21 +200,30 @@ public static function get_invalid_url_validation_errors( $post, $args = array()
 				continue;
 			}
 			$term = get_term_by( 'slug', $stored_validation_error['term_slug'], AMP_Validation_Error_Taxonomy::TAXONOMY_SLUG );
-			if ( ! $term ) {
-				continue;
+
+			$forced_sanitization = null;
+			if ( $args['check_forced_sanitization'] ) {
+				$error = $stored_validation_error['data'];
```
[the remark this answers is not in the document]
westonruter: You're right but you only had this problem because the data format of the post type is changed as of 3b484e8. Now that the validation errors are stored in the post type itself, we don't need to parse it out of the term as well. So it is expected that there would be a PHP warning as you observed. But it will go away once re-checking. Since we're in alpha I didn't think it was important to ensure back-compat for a previous alpha state.

includes/validation/class-amp-invalid-url-post-type.php
westonruter: Originally I showed a different message when _all_ of the sanitizations were forced vs _some_. So yes, this could `break` now.

includes/validation/class-amp-invalid-url-post-type.php
westonruter: You're right. It's no longer needed since the `term_group` is overridden in `get_invalid_url_validation_errors`.

includes/validation/class-amp-validation-error-taxonomy.php
westonruter: No. It should not be reset. This is a permanent filter for the user caps.

westonruter:
@ThierryA there, I'm feeling way better about this after https://github.com/Automattic/amp-wp/pull/1190/files/3e2ba8a..5458e9a

westonruter:
In 6dbdb0d a key improvement is made when accepting/rejecting validation errors from the invalid URL screen: now the URL will be automatically re-checked after the validation errors have been updated. This is very important in the case of errors like `removed_unused_css_rules` in particular, because if it is sanitized then all of the subsequent `excessive_css` validation errors automatically go away:

image

Now when changing the status of just `removed_unused_css_rules` and clicking Update, the result is then:

image

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -966,6 +1009,22 @@ public static function print_validation_errors_meta_box( $post ) {
 		} );
 		</script>
 
+		<?php
+		$has_forced_sanitized = false;
+		foreach ( $validation_errors as $validation_error ) {
+			if ( $validation_error['forced'] ) {
```
kienstra: You're probably aware of this, but I saw notices on an "Invalid AMP Page (URL)" pages:

>( ! ) Notice: Undefined index: forced in /srv/www/loc/public_html/wp-content/plugins/amp/includes/validation/class-amp-invalid-url-post-type.php on line 1015

amp-invalid-url

This occurred even after clicking "Re-check."

But these could be from errors saved from before this PR. I tried bulk moving "Invalid Pages" to the trash, but this didn't help.

Still, you [mentioned](https://github.com/Automattic/amp-wp/pull/1190#discussion_r192407969) that we don't need backwards compatibility with another version of `1.0`. And the way validation errors are stored is very different from in `0.7`.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -792,6 +806,10 @@ public static function handle_validation_error_status_update() {
 		if ( empty( $_POST[ AMP_Validation_Manager::VALIDATION_ERROR_TERM_STATUS_QUERY_VAR ] ) || ! is_array( $_POST[ AMP_Validation_Manager::VALIDATION_ERROR_TERM_STATUS_QUERY_VAR ] ) ) {
 			return;
 		}
+		$post = get_post();
+		if ( ! $post || self::POST_TYPE_SLUG !== get_post_type( $post ) ) {
```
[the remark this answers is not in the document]
westonruter: Yes, it wouldn't make any difference though. 👍

PR #1194, opened 2018-06-02
Let query var always be used and always be 'amp' when theme support is added

### LINKED ISSUE #945: Deprecate AMP_QUERY_VAR constant and amp_query_var filter

### LINKED ISSUE #1129: Deprecate the pre_amp_render_post action (when AMP theme support is present)
In v0.7 we've added `amp` theme support where you can have the entire theme be served as AMP, not just the post templates. For example, you could have your blog index template be served as AMP now (see example site: https://ampdemo.xwp.io/). In that case, there is no `$post_id` to be passed as the arg for the `pre_amp_render_post` action. But you can see from the source that the `pre_amp_render_post` action actually runs at `template_redirect` (priority 10) anyway:

https://github.com/Automattic/amp-wp/blob/7439d2ab5eebd0ee071749a849fda946a2eff4be/amp.php#L326-L383

As such, we should deprecate the `pre_amp_render_post` action entirely in favor of themes and plugins doing:

```php
add_action( 'template_redirect', function() {
    if ( is_amp_endpoint() ) {
        // ...
    }
}, 9 );
```

The `pre_amp_render_post` action should only emit a deprecation warning when `amp` theme support is present.

We could also consider adding a new action that gets triggered here instead, like `pre_amp_render_template` which does not pass any `$post_id` arg.

See #1125.

### LINKED ISSUE #1148: Discontinue use of amp endpoint in favor of query var when amp theme support is present

westonruter:
* Update `amp_get_slug()` and `amp_get_permalink()` to always use `amp` query var when theme support added (and not endpoint). When `amp` theme support is present, the `amphtml` URLs will always end in `?amp` as opposed to `/amp/`. The `amp_pre_get_permalink` and `amp_get_permalink` filters thus no longer apply, though the `/amp/` endpoint will continue to be added for back-compat. If a theme adds `amp` theme support and yet a theme also defines `AMP_QUERY_VAR` to something other than `amp` then a notice is raised.
* Mark `pre_amp_render_post` action as deprecated in favor of the `template_redirect` action with a check to `is_amp_endpoint()`. The legacy `amp_render()` function is updated to run at `template_redirect` priority 11 instead of 10 to make it easier on themers to use the default priority of 10.
* Mark all legacy functions (ones not run when `amp` theme support present) as deprecated (without raising notice).
 

Fixes #1148.
Fixes #1129.
Fixes #945.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -331,10 +331,11 @@ function amp_add_post_template_actions() {
  * Add action to do post template rendering at template_redirect action.
  *
  * @since 0.2
+ * @since 1.0 The amp_render() function is called at template_redirect action priority 11 instead of priority 10.
  * @deprecated This function is not used when 'amp' theme support is added.
  */
 function amp_prepare_render() {
-	add_action( 'template_redirect', 'amp_render' );
+	add_action( 'template_redirect', 'amp_render', 11 );
 }
 
 /**
@@ -389,6 +390,8 @@ function amp_render_post( $post ) {
 
 	/**
 	 * Fires before rendering a post in AMP.
+	 *
+	 * This action is not triggered when 'amp' theme support is present. Instead, you should use 'template_redirect' action and check if `is_amp_endpoint()`.
 	 *
 	 * @since 0.2
 	 *
--- a/amp.php
+++ b/amp.php
@@ -113,6 +113,13 @@ add_filter( 'script_loader_tag', 'amp_filter_script_loader_tag', PHP_INT_MAX, 2 
 function amp_after_setup_theme() {
 	amp_get_slug(); // Ensure AMP_QUERY_VAR is set.
 
+	/**
+	 * Filters whether AMP is enabled on the current site.
+	 *
+	 * Useful if the plugin is network activated and you want to turn it off on select sites.
+	 *
+	 * @since 0.2
+	 */
 	if ( false === apply_filters( 'amp_is_enabled', true ) ) {
 		return;
 	}
@@ -183,6 +190,7 @@ function amp_force_query_var_value( $query_vars ) {
  * If the request is for an AMP page and this is in 'canonical mode,' redirect to the non-AMP page.
  * It won't need this plugin's template system, nor the frontend actions like the 'rel' link.
  *
+ * @deprecated This function is not used when 'amp' theme support is added.
  * @global WP_Query $wp_query
  * @since 0.2
  * @return void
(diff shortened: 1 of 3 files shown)


mor10:
What does this mean for themes?

PR #1199, opened 2018-06-06
Add theme support settings to admin screen; prevent serving dirty AMP

### LINKED ISSUE #1192: Avoid serving dirty AMP until it is supported by the runtime
Since dirty AMP is not supported yet (see https://github.com/ampproject/amphtml/issues/15583), on native AMP sites (where there is no non-AMP version) we should go ahead and auto-accept (sanitize) validation errors to remove the invalid AMP markup from the page. This could be simply implemented via:

```php
    add_filter( 'amp_validation_error_sanitized', '__return_true' );
}
```

_However_ what also needs to be done is that the validation errors need to _always_ be shown in the editor instead of suppressing them once accepted.

This would basically mean that this logic would not normally run:

https://github.com/Automattic/amp-wp/blob/90acca403917c2e7b7f1b5985b03bdee7ffc6697/includes/class-amp-theme-support.php#L1141-L1153

The invalid AMP URL page should show a special warning _when in native AMP_ instead of what is currently displayed:

> This site is configured to automatically decide whether or not some validation errors may be accepted. For these errors, you cannot change their status below.

### LINKED ISSUE #1196: Add theme support mode selection in AMP general settings page
In 0.6 we added the ability for users to select which post types they wanted to make available in AMP instead of requiring them to include `add_post_type_support()` calls in their theme. In 0.7 the `amp` theme support flag was added, but there was no UI to enable it. We should allow for a user to enable `amp` theme support in the admin just as they can do for post types. 

Since the core themes for Twenty Seventeen, Twenty Sixteen, and Twenty Fifteen all are compatible with AMP via the core theme sanitizer in the plugin (via #1074), users can be encouraged to switch from the legacy templates to their theme's templates just via a toggle.

This is a natural place to incorporate the utility for doing a scan of a site to check its compatibility (see #1007).

This screen could be extended to also provide the user with control over sanitization behavior. (This is also related to the need to disable serving dirty AMP for the time being, per #1192). For example, there should be a toggle to automatically accept tree shaking. Additionally, there could be a flag for whether to automatically accept the sanitization of all validation errors.

With #934 there would also need to be a way to indicate a subset-native mode, where AMP is served on canonical URLs for some subset of the site (e.g. singular template) but on others there is no AMP version available, not even on a separate AMP-specific URL (in paired mode).

westonruter:
* Add admin settings for picking between disabled, paired, and native theme support. See #1196.
* Prevent making AMP Gutenberg blocks available when AMP is not native.
* Prevent serving dirty AMP by forcibly sanitizing all validation errors when `amp_is_canonical()`. See #1192.
* Add checkbox for automatically allowing tree shaking. This is not shown when native/canonical because dirty AMP is not allowed. 
* Add checkbox for automatically sanitizing all validation errors (including tree shaking).
* Make explicitly clear that unaccepted validation errors will block rendering on AMP
* When validation errors are automatically sanitized, ensure the terms' term_group is updated when re-checking.
* Update `AMP_Options_Manager::get_options()` and to return default values.
* Prevent showing `amphtml` link in paired mode when it is known that the URL has blocking validation errors.
* Add link to General settings admin screen in plugin links.
* Improve UI for distinguishing between validation errors which are rejected but which are also forced to be sanitized (such as in native AMP mode). These are shown as flagged.
* Fix: Remove menu-toggle in Twenty Fifteen if there is nothing to open.
* Add an AMP status item to the admin bar to give easy access to the Invalid URL post for any given URL, with an indicator for whether AMP is available, or else to initiate a validation check if no such post exists.

Fixes #1196.
Fixes #1192.

image

image

image

image


includes/options/class-amp-options-menu.php
```diff
@@ -71,6 +97,142 @@ public function add_menu_items() {
...
+						<?php esc_html_e( 'Your install is configured via a theme or plugin to automatically sanitize any AMP validation error that is encountered.', 'amp' ); ?>
+					</p>
+				</div>
+				<input type="hidden" name="<?php echo esc_attr( AMP_Options_Manager::OPTION_NAME . '[force_sanitization]' ); ?>" value="<?php echo AMP_Options_Manager::get_option( 'force_sanitization' ) ? 'on' : ''; ?>">
+			<?php else : ?>
+				<div class="amp-force-sanitize">
+					<p>
+						<label for="force_sanitization">
+							<input id="force_sanitization" type="checkbox" name="<?php echo esc_attr( AMP_Options_Manager::OPTION_NAME . '[force_sanitization]' ); ?>" <?php checked( AMP_Options_Manager::get_option( 'force_sanitization' ) ); ?>>
+							<?php esc_html_e( 'Automatically accept sanitization for any AMP validation error that is encountered.', 'amp' ); ?>
```
kienstra: It's great how this conditionally displays on selecting "Paired":
conditional-display

includes/options/class-amp-options-menu.php
```diff
@@ -71,6 +97,142 @@ public function add_menu_items() {
...
+			<?php elseif ( $should_have_theme_support ) : ?>
+				<div class="notice notice-success notice-alt inline">
+					<p><?php esc_html_e( 'Your active theme is known to work well in paired or native mode.', 'amp' ); ?></p>
+				</div>
+			<?php endif; ?>
+			<dl>
+				<dt>
+					<input type="radio" id="theme_support_disabled" name="<?php echo esc_attr( AMP_Options_Manager::OPTION_NAME . '[theme_support]' ); ?>" value="disabled" <?php checked( $theme_support, 'disabled' ); ?> <?php disabled( ! $theme_support_mutable ); ?>>
+					<label for="theme_support_disabled">
+						<strong><?php esc_html_e( 'Disabled', 'amp' ); ?></strong>
```
kienstra: If somebody looked at this really quickly, it might look like AMP is disabled, not just theme support:
classic-templates

Maybe "Disabled" could be "Classic Templates (Disabled)."

includes/admin/class-amp-editor-blocks.php
```diff
@@ -43,8 +43,16 @@ public function init() {
 		if ( function_exists( 'gutenberg_init' ) ) {
 			add_action( 'enqueue_block_editor_assets', array( $this, 'enqueue_block_editor_assets' ) );
 			add_filter( 'wp_kses_allowed_html', array( $this, 'whitelist_block_atts_in_wp_kses_allowed_html' ), 10, 2 );
-			add_filter( 'the_content', array( $this, 'tally_content_requiring_amp_scripts' ) );
-			add_action( 'wp_print_footer_scripts', array( $this, 'print_dirty_amp_scripts' ) );
```
kienstra: Do you think the methods `tally_content_requiring_amp_scripts()` and `print_dirty_amp_scripts()` could be removed, given that they're not used anymore? Maybe the [comment](https://github.com/Automattic/amp-wp/pull/1199/files#diff-af86f8db0fb2fe24a8aea92de5ee5128R48) below could [link to them](https://github.com/Automattic/amp-wp/blob/5f635d27d64f50eeba4c0710c84d929342f967b5/includes/admin/class-amp-editor-blocks.php#L156-L176).

includes/options/class-amp-options-menu.php
westonruter: That's a great point. Agreed.

And “Paired” doesn't really make sense to users at first look either, nor does “Native”.

Maybe:

* Classic Templates (Legacy)
* Active Theme's Templates with Separate URLs (Paired)
* Active Theme's Templates without Separate URLs (Canonical/Native)

I don't feel great about these these latter two.

includes/admin/class-amp-editor-blocks.php
westonruter: You're probably right. It's easy enough to look back in history to find the removal of these methods to resurrect them. 👍

includes/options/class-amp-options-menu.php
westonruter: Renamed Disabled to Classic in 5bae33e. Needs more iteration.

westonruter:
In a5756d9 added Settings link to plugins list table:

image

includes/options/class-amp-options-menu.php
ThierryA: To look at it from a different angle, perhaps we should consider using an other term than "Theme support". I understand the reasoning behind it as it is defined using `add_theme_support` but what this feature really does is controlling the AMP Mode. So perhaps the Setting itself should be called "AMP Mode" and then have:
- Legacy
- Paired
- Native

westonruter: Yes, that makes perfect sense. 👍 

Maybe “Template Mode” would be better since it's clearly related to AMP already.

westonruter:
BTW, the PR regarding URL handling (https://github.com/Automattic/amp-wp/pull/1203) is a dependency for the PR regarding the new validation error screen because it ensures that the `amp_invalid_url` post for a given URL will always be stored with the canonical URL which will then allow for us to reliably add an “AMP Status” item to the admin bar. This is important because if there are unaccepted validation errors on a paired URL then they will get redirected to the non-AMP version and this admin bar item can indicate why. Likewise, we need to omit the `amphtml` link from the canonical page if there are unaccepted validation errors on the page.

postphotos:
This is great stuff, @westonruter! 🎂 🎉 I really like the thought you've put into this.

westonruter:
Now there is an admin bar menu item to link to the AMP version of the page (when available):

image

As well as the AMP validation status (when it is known):

image

If you click on the AMP link and there are validation errors, you'll get redirected back to the canonical URL with the admin bar updated to indicate this:

image

Clicking on the validate link takes you to the Invalid AMP URL edit post screen which lists out all of the errors so you can take action. This validate admin menu item is present even if there are no known validation errors:

image

kienstra:
**Moving To "Ready For Merging"**

At some point, it'd be appropriate to test this. But as work on the Template Modes is ongoing, I'd propose that we don't need to test this now.

But if you have other ideas, feel free to move this back to "Ready For QA."

PR #1202, opened 2018-06-06
Adding sanitizer to support amp-o2-player

juanchaur1:
Adding support for [amp-o2-player](https://www.ampproject.org/docs/reference/components/amp-o2-player)

e.g: If we have the following markup

```html
<div id="579a115aa01b4d2db062aa85" class="vdb_player vdb_579a115aa01b4d2db062aa85545db032e4b0af1a81424b48">
        <script type="text/javascript" src="http://delivery.vidible.tv/jsonp/pid=579a115aa01b4d2db062aa85/545db032e4b0af1a81424b48.js"></script>
    </div>
```
We should be able to render:
```html
<amp-o2-player data-pid="579a115aa01b4d2db062aa85"
  data-bcid="545db032e4b0af1a81424b48"
  data-macros="m.playback=click"
  layout="responsive"
  width="480"
  height="270">
</amp-o2-player>
```


includes/sanitizers/class-amp-o2-player-sanitizer.php
```diff
@@ -0,0 +1,137 @@
...
+
+	/**
+	 * AMP_O2_Player_Sanitizer constructor.
+	 *
+	 * @param DOMDocument $dom  Represents the HTML document to sanitize.
+	 * @param array       $args Args.
+	 */
+	public function __construct( DOMDocument $dom, array $args = array() ) {
+		parent::__construct( $dom, $args );
+		$this->xpath = new DOMXPath( $dom );
```
westonruter: The `xpath` setting here could just as well be moved to the `sanitize` method since that is where it is used. This would avoid needing to override the constructor at all, and if the sanitizers are instantiated but never call `sanitize` then we could avoid needlessly instantiating this object (however rare that may be).

includes/sanitizers/class-amp-o2-player-sanitizer.php
```diff
@@ -0,0 +1,137 @@
...
+
+	/**
+	 * Height to set for 02 Player elements.
+	 *
+	 * @var string
+	 */
+	private static $height = '270';
+
+	/**
+	 * Width to set for 02 Player elements.
```
westonruter: Should be `O2` not `02`.

includes/sanitizers/class-amp-o2-player-sanitizer.php
```diff
@@ -0,0 +1,137 @@
...
+
+	/**
+	 * Amp O2 Player class.
+	 *
+	 * @var string CSS class to identify O2 Player <div> to replace with AMP version.
+	 */
+	public static $xpath_selector = '//div[ contains( @class, \'vdb_player\' ) ]/script';
+
+	/**
+	 * Height to set for 02 Player elements.
```
westonruter: Should be `O2` not `02`.

includes/sanitizers/class-amp-o2-player-sanitizer.php
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Class AMP_O2_Player_Sanitizer
+ *
+ * Converts <div class="vdb_player><script></script></div> embed to <amp-o2-player>
+ *
+ * @see https://www.ampproject.org/docs/reference/components/amp-o2-player
+ */
+class AMP_O2_Player_Sanitizer extends AMP_Base_Sanitizer {
+	const URL_PATTERN = '#.*delivery.vidible.tv\/jsonp\/pid=(.*)\/vid=(.*)\/(.*).js.*#i';
```
westonruter: Would be good to add phpdoc for this.

westonruter: Might want to use named capture groups to make it easier to reference the `$matches`.

includes/sanitizers/class-amp-o2-player-sanitizer.php
```diff
@@ -0,0 +1,137 @@
...
+ * Class AMP_O2_Player_Sanitizer
+ *
+ * @package AMP
+ */
+
+/**
+ * Class AMP_O2_Player_Sanitizer
+ *
+ * Converts <div class="vdb_player><script></script></div> embed to <amp-o2-player>
+ *
```
westonruter: Could you add `@since 1.0` tag to the class (at least)?

juanchaur1: done

PR #1203, opened 2018-06-07
Improve handling of URLs and location redirects in AMP

### LINKED ISSUE #1166: Increase timeout for validate URL requests and handle timeout errors
Currently there is a 5 second timeout when doing a `validate_url` request:

https://github.com/Automattic/amp-wp/blob/a68c93fd62df93c6c79cf99d4917bb912e11b684/includes/utils/class-amp-validation-utils.php#L1426-L1432

Given the added overhead of locating sources and bypassing the post-processor cache (#959), it can happen that a request to validate a URL can timeout, especially when the stylesheets have not yet been processed and cached.

So not only do we need to increase this timeout, we also need to add error reporting for when `validate_url` returns with an error, including in bulk/inline re-check, during post save in classic editor, and doing a REST API `POST`/`PUT`. At the moment, such validation requests fail silently, meaning a user can be shown that there is no change to the improvement of validation errors.

westonruter:
* Automatically redirect from `/amp/` endpoint to `?amp` when `amp` theme support is present. This is important for plugins to be able to reliably call `is_amp_endpoint()`. See #1148.
* Add `amp_get_current_url()` helper function.
* Let `amp_add_amphtml_link()` replace the poorly-named `amp_frontend_add_canonical()` function and improve how it is referenced.
* Fix `old_slug_redirect_url` handling for AMP theme support. This ensures that AMP URLs redirect properly when changing a post's slug.
* Fix handling of re-validating URLs when redirects happen or post slugs change. Fix validation when switching between native and paired modes. Only store canonical URLs for `amp_invalid_url` posts.
* Restore passing of current user cookies when validating a URL to access private/drafted posts.
* Increase timeout for validate_url requests; show errors when fetching fails
* Fix issue with `get_current_screen()` not returning an object.

Fixes #1166.


gravityrail:
> Automatically redirect from /amp/ endpoint to ?amp when amp theme support is present. This is important for plugins to be able to reliably call is_amp_endpoint().

I'm not able to reproduce this behaviour. I tried with/without `add_theme_support( 'amp' )`, with/without trailing `/amp/`, and I wasn't able to trigger a redirect to `?amp`. I see that you have injected your link redirection code into the action `old_slug_redirect_url` but I don't see anywhere in your branch that that filter is called - maybe there's a change missing from this PR?

westonruter:
@gravityrail humm, the `old_slug_redirect_url` is for a different case for when the slug changes. The redirection from `/amp/` to `?amp` is performed here in the `ensure_proper_amp_location` method:

https://github.com/Automattic/amp-wp/blob/236be4babd249e27b98a60526355d036ef4872d4/includes/class-amp-theme-support.php#L166-L206

I just double-checked and it's working for me:

```
curl -i https://src.wordpress-develop.test/2018/06/06/initial333/amp/
HTTP/1.1 302 Found
Location: https://src.wordpress-develop.test/2018/06/06/initial333/?amp
```

😕

gravityrail:
Ah! Now I'm able to reproduce the desired behaviour.

The trick is to enable paired mode, which allows specifying a theme directory to AMP-ify:

```
add_theme_support( 'amp', array( 'template_dir' => './' ) );
```

Then you ALSO need to make sure you accept any validation errors that have been generated (otherwise it always redirects from AMP to non-AMP versions of page)

artkanna:
Hi! Will  `Fix old_slug_redirect_url handling for AMP theme support` retroactively fix amp redirects for pages that were previously redirected - before the release? Thank you!

westonruter:
@artkanna Can you give an example of pages that were previously redirected?

Issue #968, opened 2018-02-19
Implement AMP Stories


davisshaver:
**ISSUE 1: As a WordPress author, I need to be able to manage content for AMP stories using Gutenberg.**
Key PR: #1215 

- [x] AC1: I should be able to create between 1 and _n_ pages for AMP stories.
   - [x] At this time, _n_ is assumed to be up to 20 pages (though this is not a hard limit).
- [x] AC2: I should be able to rearrange pages inside AMP stories.
- [x] AC3: I also, within each page, need to be able to create _x_ number of layers (CTA, grid layer, etc). 
   - [x] _x_ is assumed to be up to 5 layers at this point, but is not a hard limit.
   - [x] Only one CTA layer is allowed, and the interface should anticipate this.
- [ ] AC3: I should be able to rearrange layers inside AMP pages.
- [ ] AC4: Within each layer, I also need to be able to set layouts (thirds, fill, horizontal, vertical, etc.)
- [ ] AC5: Also within each layer, I need to be able to add my own blocks which are allowed by AMP (e.g. paragraph, image, video, etc).
- [ ] AC6: I should be able to preview the AMP story within the editor. I should also be able to edit each page next to each other. The current mode should be obvious to the author (edit or preview).
- [ ] AC7: I need a basic way to set default animations on pages, layers, elements for an AMP story.
- [ ] AC8: Each setting panel for each level (page, layer, element) should be clearly labeled as a user navigates through and selects them. There should be an intuitive way to be able to navigate through them. There should be an intuitive way to move items through layers if needed.
- [ ] AC9: I need a clear way to understand what I’ve selected is being edited. For example, a border or shading around an HTML element clearly shows a web browser console selecting that DOM element. 

# AMP Dependencies

- [ ] Expose lifecycle of AMP components to enable constructing them dynamically https://github.com/ampproject/amphtml/issues/17548

# Gutenberg Dependencies

- [x] Nested blocks (see https://github.com/WordPress/gutenberg/pull/3745 and [InnerBlocks](https://github.com/WordPress/gutenberg/blob/master/blocks/inner-blocks/README.md))
- [x] [Locked post type block templates](https://github.com/WordPress/gutenberg/blob/master/docs/templates.md#custom-post-types)
- [x] Restricting blocks to given post type: https://github.com/WordPress/gutenberg/pull/6208
- [x] Child blocks: https://github.com/WordPress/gutenberg/issues/5540
[...]

PR #1215, opened 2018-06-16
[WIP] Introduce amp-story support via Gutenberg blocks

westonruter:
**As a developer for the AMP for WordPress plugin, I need introductory support for managing AMP stories via Gutenberg blocks.**

- [x] AC1: Introduce a general structure and schema to what a block is, in relationship to new "presentation types" inside Gutenberg.
- [x] AC2: Propose a UI that could work using this general structure.
- [x] AC3: Store the "AMP story" as a custom post type.
- [ ] AC4: Inform #968 of any progress, and break out any new tasks as needed.

Subtasks:
- [x] Create system for editing AMP Story grid layers while they appear on top of each other (see https://github.com/Automattic/amp-wp/pull/1347);
- [x] Add default background settings (color) for AMP Story Page block (see https://github.com/Automattic/amp-wp/pull/1380);
- [x] Animations for frontend (see https://github.com/Automattic/amp-wp/pull/1385);
- [x] UI fixes (see https://github.com/Automattic/amp-wp/pull/1388);
- [ ] the “Classic Editor” option should not be available for AMP Stories;
- [x] Fix selection of layers (`selectBlock`);
- [ ] Finalize styling;
---

❗️ _Note that no design/UX have been incorporated into this PR yet. As of now it is purely an exploration into the technical details of using Gutenberg blocks for implementing an AMP Story editor._

This PR introduces an `amp_story` custom post type, and when editing this post there are `<amp-story-page>` and `<amp-story-grid-layer>` blocks that can be added, with generic blocks allowed in the latter as depicted in the [AMP Story docs](https://www.ampproject.org/docs/reference/components/amp-story):

image

Inspired by the color scheme above, the blocks are color-coded (for testing, except got the blue and green colors reversed) as followed:

screen shot 2018-06-16 at 7 26 25 pm

screen shot 2018-06-16 at 7 26 55 pm

A custom template is made available for the `amp_story` post type which renders out the standalone `<amp-story>`:

screen shot 2018-06-16 at 7 27 28 pm

See #968.


westonruter:
Rebased to fix merge conflicts.

Some problems with the current implementation:

* After inserting an `amp-story-page` it is difficult to add the first `amp-story-grid-layer` block inside of it.
* It is difficult to select the `amp-story-page` block to insert a new `amp-story-grid-layer` block inside of it. There should be a block template for `amp-story-page`.
* I can erroneously insert an `amp-story-page` block _inside_ of an `amp-story-grid-layer` block. 
* I can erroneously insert an `amp-story-grid-layer` block inside of another `amp-story-grid-layer` block.
* I can erroneously insert an `amp-story-grid-layer` as a root block, not under an `amp-story-page`.

westonruter:
I thought I found a solution to the problem to being able to erroneously be able to insert an `amp-story-page` at any level other than the root by specifying the `allowedBlocks` as well as the `parent` block: 9cd428e. However, this resulted in an infinite recursion error:

> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

```
"Invariant Violation: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    at invariant (https://src.wordpress-develop.test/wp-content/plugins/gutenberg/vendor/react-dom.24169eaf.js:59:15)
    at scheduleWork$1 (https://src.wordpress-develop.test/wp-content/plugins/gutenberg/vendor/react-dom.24169eaf.js:16799:11)
    at Object.enqueueForceUpdate (https://src.wordpress-develop.test/wp-content/plugins/gutenberg/vendor/react-dom.24169eaf.js:11914:5)
    at t.Component.forceUpdate (https://src.wordpress-develop.test/wp-content/plugins/gutenberg/vendor/react.832d746b.js:542:16)
    at t.value (https://src.wordpress-develop.test/wp-content/plugins/gutenberg/build/components/index.js?ver=1531156454:37:139734)
    at updateClassInstance (https://src.wordpress-develop.test/wp-content/plugins/gutenberg/vendor/react-dom.24169eaf.js:12340:18)
    at updateClassComponent (https://src.wordpress-develop.test/wp-content/plugins/gutenberg/vendor/react-dom.24169eaf.js:13731:20)
    at beginWork (https://src.wordpress-develop.test/wp-content/plugins/gutenberg/vendor/react-dom.24169eaf.js:14402:14)
    at performUnitOfWork (https://src.wordpress-develop.test/wp-content/plugins/gutenberg/vendor/react-dom.24169eaf.js:16441:12)
    at workLoop (https://src.wordpress-develop.test/wp-content/plugins/gutenberg/vendor/react-dom.24169eaf.js:16480:24)"
```
image

So I've reverted that.

westonruter:
> It is difficult to select the `amp-story-page` block to insert a new `amp-story-grid-layer` block inside of it. There should be a block template for `amp-story-page`.

I believe this is mostly fixed by cd9cdbe.

jwold:
@westonruter I've been thinking through the UI implications a bit today. I'm not quite sure about the layers aspect, so I'd like to chat about that further. You can see some of my sketches below:

ideas

jwold:
Quick question @westonruter, could Layer and Page settings be included in the same area? Trying to understand why they would need to be selected differently or have settings in different areas. 

It seems they always have a direct relationship to each other. Are there situations where a page could have 2 layers? Or a layer is across 2 pages?

bigmazi:
Hi,
I say shortly and I will have hope you are understand. They need to work in model:
Matrix in Matrix to be safe :)

If you will have any question or needs please let me know I will help with pleasure.

Of course my job I will do also but just let me know what to do to do not double our yobs.

Thanks and to hear you

Also I have names If You know what I mean

westonruter:
@jwold A page can have one or more layers. It has to have at least one layer, but it can have any number. A layer cannot span pages. So there is a direct hierarchical relationship.

bigmazi:
Hi,
I say shortly and I will have hope you are understand. They need to work in
model:
Matrix in Matrix to be safe :)

If you will have any question or needs please let me know I will help with
pleasure.

Of course my job I will do also but just let me know what to do to do not
double our yobs.

Thanks and to hear you

westonruter:
@bigmazi I have no idea what you are talking about.

jwold:
@westonruter and @postphotos I'd love some feedback on this idea (and anyone else who can weigh in)

First, here's the screenshot:

new mockup 3

Second, here's my video walkthrough: https://v.usetapes.com/LERPRZwnLk

westonruter:
@jwold It may be that the need for the tabs here is due to Gutenberg not making it super easy to select parent blocks when nesting blocks. Let's say that we do need the tabs for now to work with that. I think, however, we only need one tab/shortcut: select the page block. And this may be something to propose to fix in Gutenberg itself. It may be that the nesting hierarchy should be exposed in the sidebar, header where each nested level gets its own Block tab (horizontal space availability notwithstanding):

image

But, given that there can and would be multiple layers, how would you select a specific layer to manipulate? I don't see a way in your mock to manipulate layers yet. If you select the page and open the block settings sidebar there at this point could be a _list of the layers_ in the page. Inside of the sidebar the user could _reorder_ the layers, remove layers, or add new layers (including a final CTA layer). I think this would be easier to work with than try to replicate the block reordering controls for manipulating layer ordering.

This https://github.com/WordPress/gutenberg/issues/7931 issue in [[Component] Nested / Inner Blocks](https://github.com/WordPress/gutenberg/issues?q=is%3Aissue+is%3Aopen+nested+block+parent+label%3A%22%5BComponent%5D+Nested+%2F+Inner+Blocks%22) may be relevant here as well, since we will probably want to disable the reordering controls for the layer blocks. We'd also potentially need to change the orientation of the reordering controls _inside_ a layer for the blocks when the horizontal template is selected.

jwold:
@westonruter and @postphotos I'd love some feedback on where I've landed so far with the AMP Stories. Here's a brief video: https://v.usetapes.com/GXK3Q9MvsR

And here's the full wireframe: 

amp stories

jwold:
@westonruter great points that you brought up. I'm still thinking through them. 

In the meantime you can see an updated wireframe that takes into account some of your thoughts:

new mockup 3

cc @amedina

jwold:
> Tapping the layer would select the layer on the page, to show that layer in isolation and allow its elements to be manipulated. So then to get back to the view of all layers combined, how would the user do this? The button in the layer looks like a radio.

I was thinking that the radio like button would serve like in Illustrator or Photoshop where it's a target/highlight button to show all elements in that layer. The reason is because I want to be able to edit and change the settings in layers WITHOUT having all the view change everytime I click a different layer, I only want it to change when I change a setting. This is important because I'll likely (as a user) want to click through layers and look at all the options, and having the view changing when I did this would be weird. 

In addition, I was thinking "show all layers" would work well to allow users to see all the layers beyond just the one they've highlighted. 

> What happens when the user clicks on the story with the intention of editing what they click? Should that result in the top-most layer automatically being selected and the others hidden? But then what if they wanted to edit something that appears in the background? That could be jarring as then the thing they want to edit would then just disappear.

Could this be solved by having the other layers go to 50% opacity? They'd still be there, just not fully visible. 

> Also, perhaps upon selecting a page that the sidebar would need to be shown (even if it was closed) so that they could see what the layers are.

Agreed

jwold:
@westonruter What would happen if I were to move a layer up and down inside of the AMP Story Page on the left, should it also move the layers around on the sidebar?

westonruter:
> Page 1 of 1

I think in the editor, the pages should be all visible and seen by scrolling down in the window, similar to how slides are presented in presentation apps when in edit mode.

> Select to show layer

Actually I think it would be “select to _edit_ layer”. Also, it would have the effect of _isolating_ the layer.

> In addition, I was thinking "show all layers" would work well to allow users to see all the layers beyond just the one they've highlighted.

So that would have the effect of clearing the selected radio button? What if there was an actual button that said “Edit” and then this switched do “Done”?

> Could this be solved by having the other layers go to 50% opacity? They'd still be there, just not fully visible.

Potentially, yes. We'd also need the non-edited layers to be essentially inert, so that in addition to `opacity:0.5` they'd need to be read-only and `pointer-events:none` so that you click through any layers on top of the layer you're editing.

> What would happen if I were to move a layer up and down inside of the AMP Story Page on the left, should it also move the layers around on the sidebar?

I don't think there would be up/down arrows for layers in the editor. The only way to re-order them would be via the sidebar. The arrows with the blocks in the editor would only be for the blocks inside of a layer.

westonruter:
Related Gutenberg issue: https://github.com/WordPress/gutenberg/issues/7694 (Blocks: Consider a "passthrough" supports for nested block wrappers). This would seem to be seems to be key for the layer blocks.

jwold:
@westonruter just thinking through another example of an AMP Story. A few questions came up, welcome your thoughts! https://d.pr/v/uZrEQN

cc @amedina

westonruter:
Another related issue: https://github.com/WordPress/gutenberg/issues/6459 (Show block breadcrumbs for selected block when inline toolbars are enabled)

westonruter:
@jwold I think the first slide should have the “Your 5-minute fix” and other content as the bottom third of the same layer that has the WaPo logo in the first layer. I don't think these would be pulled from the metadata for the page automatically as a special case for AMP Story, as part of the template. If such automatic-population were desired, then they should create a Gutenberg block for surfacing that information from the post metadata. This is something I know is needed by Gutenberg more broadly for customization, to be able to construct a template entirely with blocks (e.g. title, byline, date, etc).

Something that @mtias pointed out as well that we should think about is how the background image is achieved. In particular, Gutenberg has the Cover Image block currently which allows you to put text over an image:

> image

Nevertheless, the Cover Image block is not flexible enough for replicating that `amp-story-page` in [that AMP Story](https://www.washingtonpost.com/graphics/2018/national/amp-stories/the-5-biggest-political-headlines-facebook-and-paul-ryan/). In particular, there aren't controls for managing the font size or positioning of the text. The Cover Image block could be decomposed into Image block and a Heading block if Gutenberg allowed them to be layered on top of each other.

The vertical centering of the text on the grid layer is achieved with a custom stylesheet that is doing the following (so is not strictly part of the format):

> image

The caption at the bottom of the page is indeed a separate layer, and it is positioned on the bottom again with custom CSS not unique to the AMP story format:

> image

The text is centered vertically on the last slide with the same custom CSS associated with the `text-slide` class as per above. The button at the bottom is actually placed by the `amp-story-cta-layer`.

blocks/amp-story/amp-story-cta-layer.js
```diff
@@ -0,0 +1,58 @@
...
+		 * <amp-story-cta-layer>:
+		 *   mandatory_ancestor: "AMP-STORY-PAGE"
+		 *   descendant_tag_list: "amp-story-cta-layer-allowed-descendants"
+		 *
+		 * https://github.com/ampproject/amphtml/blob/87fe1d02f902be97b596b36ec3421592c83d241e/extensions/amp-story/validator-amp-story.protoascii#L172-L188
+		 */
+
+		edit() {
+			return (
+				<InnerBlocks key='contents' allowedBlocks={ ALLOWED_BLOCKS } />
```
miina: @westonruter Using `allowedBlocks` here should enforce not being able to use `amp-story-page` within the inner blocks -> it should be allowed only on the root level this way.

blocks/amp-story/amp-story-grid-layer.js
```diff
@@ -0,0 +1,124 @@
...
+	return props;
+}
+
+wp.hooks.addFilter(
+	'blocks.registerBlockType',
+	'amp/set-block-parents',
+	setBlockParent
+);
+
+// @todo Perhaps this should wait behind https://github.com/WordPress/gutenberg/issues/8589?
```
miina: Commented this out since with using `allowedBlocks` param for the inner blocks it should already restrict the usage of the blocks. Since only `amp-story-page` should be allowed on the root level, perhaps it wouldn't make sense to unregister other blocks here? Maybe we should wait for https://github.com/WordPress/gutenberg/issues/8589 to be able to use `amp-story-page` as the default instead.

blocks/amp-story/amp-story-page.js
```diff
@@ -0,0 +1,93 @@
...
+						required={ true }
+						label={ __( 'ID', 'amp' ) }
+						value={ props.attributes.id }
+						onChange={ value => ( setAttributes( { id: value } ) ) }
+					/>
+				</InspectorControls>,
+				! props.attributes.id && (
+					<Notice status="error" isDismissible={ false }>{ __( 'You must supply an ID for the page.', 'amp' ) }</Notice>
+				),
+				<InnerBlocks key='contents' template={ TEMPLATE } allowedBlocks={ ALLOWED_BLOCKS } />
```
miina: This should ensure not being able to add `amp-story-page` as a child for `amp-story-page`.

blocks/amp-story/amp-story-grid-layer.js
```diff
@@ -0,0 +1,124 @@
...
+const {
+	InspectorControls,
+	InnerBlocks
+} = wp.editor;
+const {
+	SelectControl
+} = wp.components;
+
+const ALLOWED_BLOCKS = [
+	'core/code',
```
miina: Removed `button` from the list since according to the [specs](https://www.ampproject.org/docs/reference/components/amp-story#children) it seems that the grid layer does not allow buttons/links, only the CTA layer does.

westonruter:
@miina deploy errors should be fixed after merging this PR into this branch: https://github.com/Automattic/amp-wp/pull/1345

blocks/amp-story/amp-story-page.js
miina: @westonruter Just saw that you tried with `allowedBlocks` within https://github.com/Automattic/amp-wp/commit/9cd428e38535c0b48b01b25dc6259a92c1eac2ff and it created an error. Are you getting the error now as well? I'm not seeing the error at this moment, wondering if I should actually.

westonruter: Yes, I had tried it but I was getting an error. However, if you're not getting an error now, then perhaps the cause was fixed in Gutenberg.

westonruter:
Current state:

image

image

postphotos:
Hi @miina, this is really coming along since our first designs! My first round of tirekicking revealed a few surfaced items for discussion:

1) The custom post type for `AMP Stories` probably should not surface if Gutenberg is not installed. (I forgot to install it on my local and discovered this.) 😄 
screen shot 2018-09-12 at 1 24 51 am

2) As opposed to prompting the user on every layer creation, we should make CTA layer an option (less friction to add layers) inside the layers panel. Same goes, probably, for `<amp-story-bookend>` when we get there.
screen shot 2018-09-12 at 1 49 21 am

3) Styles from Gutenberg _feel_ like they should match preview/post. It's an almost perfect match, but the rendered content looks different than editing experience, especially here. Not sure if this is related to work still in progress, but calling it out anyway... unless I'm missing something here.
screen shot 2018-09-12 at 1 37 02 am

4) Double hinting of two blocks on top of each other is confusing if I create multiple layers. (i.e. hide placeholder on non-active layers.) 
screen shot 2018-09-12 at 1 33 33 am

5) In the current `amp-stories` branch, the editing bar gets in the way sometimes. 
screen shot 2018-09-12 at 1 34 12 am

6) Active vs non-active layers needs clearer differentiation - it's still kind of hard to tell what layer I'm on. cc @jwold 

7) I can resize an image larger than the viewer, which breaks the whole thing.
screen shot 2018-09-12 at 1 39 18 am

8) Could not figure out how to set a background to a layer (or full canvas image) to replicate a given AMP layout. (First thing I was trying to do.)

Awesome work so far! cc: @westonruter

miina:
Thanks for the thoughts and bringing out the issues you found, @postphotos!

For now I'm thinking of mainly looking into the bugs, and the improvements of UX we could leave for later after discussing, does this sound reasonable?

1. Correct, will fix!
>1. The custom post type for AMP Stories probably should not surface if Gutenberg is not installed. (I forgot to install it on my local and discovered this.) 😄

2. Do you mean that instead of having the CTA block in the block inserter interface instead we could have it in the right sidebar as a toggle option? Since the parent of the CTA block is `Page` then it would have to be in the `Page` block's settings. Is that what you are thinking of? And then a new Grid Layer would be automatically inserted just when clicking the `+` icon? If yes, then I'm not convinced about this since it feels like the default way how to add blocks in Gutenberg is via block inserter. Also, in this case the user would need to go to `Page` settings first to add the CTA Layer and the user would have to _know_ that it's there instead of using the block inserter as the default behavior of Gutenberg is. However, if the general feeling is that it's a better option then we can go for it. On `amp-story-bookend` -- note that this block would be on the same level as `amp-story-page` so it would be inserted from the same block inserter as a new page. Also, perhaps I misunderstood your thought!
>2. As opposed to prompting the user on every layer creation, we should make CTA layer an option (less friction to add layers) inside the layers panel. Same goes, probably, for <amp-story-bookend> when we get there.

3. The screenshots you shared with the console error shouldn't be happening and is not related to being WIP. Is there a way you could recreate the situation so that I could access this as well? The way how AMP Story is displayed on desktop if this:
screen shot 2018-09-12 at 10 22 36 am
So it's weird that it's displaying full desktop width in your screenshot. Note that when the dev tools are opened the browser displays the story only when the device toolbar is on (mobile view). 
>3. Styles from Gutenberg feel like they should match preview/post. It's an almost perfect match, but the rendered content looks different than editing experience, especially here. Not sure if this is related to work still in progress, but calling it out anyway... unless I'm missing something here.


On 4., 5., 7. -- good points, will look into it!

8. Do you mean that you would like to have a layer that would fill the whole page with an image? For that you would have to:
- Set the layer template to `Fill`;
- Add an image block within that layer;
- Set an image that would cover the whole page -- note that it should be with correct dimensions, otherwise the image will cover the background only partly since the AMP Story is not going to stretch the image, for example this has incorrect dimensions:
screen shot 2018-09-12 at 10 37 26 am
>8. Could not figure out how to set a background to a layer (or full canvas image) to replicate a given AMP layout. (First thing I was trying to do.)

postphotos:
**No further actions probably needed here until after the MVP of Stories is met, but I'd like to know about background and opacities**

Thanks @miina for updating the branch with the PR, looks good! I've responded to all of your comments and shaken loose a few more things. 😄 

> For now I'm thinking of mainly looking into the bugs, and the improvements of UX we could leave for later after discussing, does this sound reasonable?

1) Yes indeed! cc: @jwold 


> > The custom post type for AMP Stories probably should not surface if Gutenberg is not installed. (I forgot to install it on my local and discovered this.) 😄
> Correct, will fix! 

2) Just calling out the message from Slack: We'd also want to prevent Classic Editor editing from the "Edit AMP Stories" CPT screen. I'll add this to #968 as a new requirement.

> > As opposed to prompting the user on every layer creation, we should make CTA layer an option (less friction to add layers) inside the layers panel. Same goes, probably, for  when we get there.
> 
> > The screenshots you shared with the console error shouldn't be happening and is not related to being WIP. Is there a way you could recreate the situation so that I could access this as well? The way how AMP Story is displayed on desktop if this:
> >   screen shot 2018-09-12 at 10 22 36 am
> >  So it's weird that it's displaying full desktop width in your screenshot. Note that when the dev tools are opened the browser displays the story only when the device toolbar is on (mobile view).
> >
> Do you mean that instead of having the CTA block in the block inserter interface instead we could have it in the right sidebar as a toggle option? Since the parent of the CTA block is `Page` then it would have to be in the `Page` block's settings. Is that what you are thinking of? And then a new Grid Layer would be automatically inserted just when clicking the `+` icon? If yes, then I'm not convinced about this since it feels like the default way how to add blocks in Gutenberg is via block inserter. Also, in this case the user would need to go to `Page` settings first to add the CTA Layer and the user would have to _know_ that it's there instead of using the block inserter as the default behavior of Gutenberg is. However, if the general feeling is that it's a better option then we can go for it. On `amp-story-bookend` -- note that this block would be on the same level as `amp-story-page` so it would be inserted from the same block inserter as a new page. Also, perhaps I misunderstood your thought!


3) Currently, we have two separate blocks for managing the two different kinds of layers. I suggest we merge them.

In short, I'd say a "layers type" toggle or dropdown would be more appropriate.

I think a user should be able to create layers even more easily than now - unless have a reason for more layer types than two, we should _strongly_ find a way to merge this. CTA layers aren't likely to occur on many pages for a given story (there might be a few) so trying to de-escalate that feature means the regular layers become easier to use.

Suppose a story has 10 pages, each with 5 layers. That'd be 50 regular layers for a given AMP Story, with maybe be two or three CTA layers. I'd say make it easier for 47 of those clicks so that user doesn't mind doing an extra action three times (as opposed to 47 extra actions.)

So unless we have the time, I'd consider this an enhancement for now, so the work you have completed suffices for MVP. 👍**No actions needed for this until a future date.** cc: @jwold 

And FWIW `amp-story-bookend` should not be part of this ticket or the initial effort of #968. 😄 I see that solution looking similar to this one, where ideally bookend is an option only available on the last page in a story. 

> > Could not figure out how to set a background to a layer (or full canvas image) to replicate a given AMP layout. (First thing I was trying to do.)
> Do you mean that you would like to have a layer that would fill the whole page with an image? For that you would have to:
> 
> * Set the layer template to `Fill`;
> * Add an image block within that layer;
> * Set an image that would cover the whole page -- note that it should be with correct dimensions, otherwise the image will cover the background only partly since the AMP Story is not going to stretch the image, for example this has incorrect dimensions:
>   screen shot 2018-09-12 at 10 37 26 am

That worked for me! Shaking loose bugs and questions:

4) If I pull the bottom handle first in fill mode, it goes wonky like this:
screen shot 2018-09-12 at 5 09 03 pm

Not sure if this is an issue with us or Gutenberg or is simply WIP related. 

5) This is more of an enhancement, but this plus button isn't in an ideal place yet. I struggled yesterday to place an image on as I assumed it'd be a "background image." The circle around the plus is also not aligned correctly. I'm trying to understand if there's currently a `background` property or if there could be. I think this may be a question for the AMP team directly (@amedina?) but It felt intuitive to add a background color to a layer or object. In reviewing this doc on [AMP Stories](https://www.ampproject.org/docs/reference/components/amp-story#amp-story-grid-layer), I don't see an opinion on whether we could do this. There could be a conscious choice _why_ but I don't see it yet.

In reality some blocks in G (i.e. `Paragraph` and `Headline`) support this, but I wonder if there's a way to replicate this kind of contentless overlay, either as a block or later property. 

A few of the demo AMP stories I've seen look to have used semi-transparent opaque images, but a few bits of CSS is definitely easier to generate and manage, and definitely more performant. I'm spitballing here, but hopefully the right solution comes loose. 🎨 
screen shot 2018-09-12 at 5 27 38 pm

@westonruter @jwold Any thoughts on backgrounds here?

6) In that same vein, it could be useful to consider suggesting adding either `opacity` or `rgba (#,#,#,#)` values to our text blocks to allow for a half-step of what I'm describing here,
 and perhaps it is a contribution we can share with Gutenberg for all text blocks. 

FWIW [there is a pattern](https://github.com/WordPress/gutenberg/issues/9329) for this already that uses an opacity slider. 👍  


7) Should image blocks in the whole AMP Story view support `Write Caption...`? It feels like it shouldn't.
screen shot 2018-09-12 at 5 11 22 pm

westonruter:
> Any thoughts on backgrounds here?

So setting a background image in addition to the background color? The background color is just being set with CSS and so the same could be done for a background image. That would avoid having to create a layer just for a single fill image block, so maybe that would be good. But then there would also be two ways to do the same thing. If the background image were set in the block inspector controls, then there would also potentially need to be controls for managing the position of the background image similar to in the Customizer. That might be overkill however.

jwold:
@miina if we change the template type for the layer will the icon also change to the right type?

screen shot on 2018-09-14 at 08_57_21

miina:
@jwold Yes, it does change the icon as well.

postphotos:
So glad we're making progress here! 🎉 

A few new thoughts @miina, and I'd consider these enhancements to be part of MVP, unless there's any objection:

1) **Block settings confusion.** When adding a new layer, a few things happen that confuse me:
 a) The "block" settings switches from whatever layer/block I was on (layer _n_ or block on said layer) to the new layer's new block (layer _n + 1_)
 b) This happens to always be a `paragraph` block that is automatically created, and I don't see layer options. 
 c) I struggle to create a non-`paragraph` block. I have to remove the paragraph, add a new block (using the plus block). I try to convert the block to another kind of block using the text dialogue, but that doesn't work if it's a video/photo. 

I would rather:
 a) When adding a new layer, I see its settings first
 b) No block is created, through some Gutenberg pattern I'm suggested options to create a block inside that layer (photo, video, paragraph, headline, quote, etc.) If I don't have text (on the average most AMP Stories content isn't likely to be `paragraph` text) there needs to be an easier option to set it here.

2) **Layer management.** As a user, I need some basic way to reorder layers. Once opacity got into the mix, I found it hard to compose. Without changing the current layers dialogue, it'd be nice to be able to take layer _n_ and move it up or down in the view, to either layer _n - 1_ or layer _n + 1_. 

As a future goal past MVP we could consider reusing the arrow toggles from Gutenberg, but that'd need design feedback. cc: @jwold 

3) **Opacity.** This looks great! It introduces a concept of compositing that I love. Can we change the label to "Layer Opacity" to be slightly more specific?

4) **Default new layers.** When adding a new layer, it's always a thirds layout. Can we, instead, change the default to be whatever is the top layer's layout? 

When creating a background composition I'd rather have multiple fills, for example, and then break into sub layouts than always start at thirds.

postphotos:
In regard to 1b mentioned above, I think we need to design a solution (or force open the "add new block") that allows a user to select non-paragraph blocks more easily.

This might be the toggle that is force opened as the layer creation event happens: 
screen shot 2018-09-14 at 5 11 26 pm

So they'd see the above plus the layers panel:
screen shot 2018-09-14 at 5 13 05 pm

miina:
Hey @postphotos, thank you for the comments.

1: This is actually currently a known issue and should be related to https://github.com/WordPress/gutenberg/issues/9212, we should try to fix it within the next phase and possibly within Gutenberg, unfortunately this probably won't be part of MVP.

2: The default arrows shouldn't be modified and should appear in the left top of a given Layer -- is this missing for you? Or perhaps you were thinking of an alternative?
screen shot 2018-09-17 at 9 17 41 am


3. Sure, makes sense, will change the label.

4. At this moment the default template should be `Vertical`, chose this since it matches the best the general Gutenberg editor flow compared to other templates. In your testing is the default `thirds` instead?
To understand your thought correctly: are you thinking that maybe by default a new layer should have the same template as the previous one (the one which was on top before the new layer was added on top of that)? For example if the template of a layer is `fill` and then a new layer is added on top of it, the new layer would also get the template `fill` instead of the current default? And then if the added layer's template would be changed to `thirds` and a new layer added, the new layer would have `thirds` as the template, too? Let me know if I understood correctly what you meant.
>Default new layers. When adding a new layer, it's always a thirds layout. Can we, instead, change the default to be whatever is the top layer's layout?


Thank you, Leo!

postphotos:
Thanks for fixing the dependency issues and merging `develop` back into this branch @miina! 

@amedina, Here's an up-to-date version of the AMP Stories plugin with all changes to date and the AMP Stories editor Proof-of-Concept enabled:
[amp.zip](https://github.com/Automattic/amp-wp/files/2474177/amp.zip)

amedina:
Is the build process fixed? i.e. can the zip file be generated locally?

postphotos:
@amedina yep! All you have to do is:
1) follow the [install instructions](https://github.com/Automattic/amp-wp/blob/develop/contributing.md) to setting up the repo locally.
2) Create a new branch if you haven't already for AMP Stories: `git checkout -b amp-stories`
3) Fetch the latest changes from Github: `git pull origin amp-stories`
4) Rerun the build process: `npm run build`

It should just work! 🎉 

It'll also generate a packaged `amp.zip` file in the root of the plugin directory, e.g. `/wp-content/plugins/amp/amp.zip`

westonruter:
Just noticed that the poster image warning is being added to non-story posts incorrectly. 

image

- [ ] Limit warning to `amp_story` post type, and also improve the message to be just, “You must provide a featured image.”

westonruter:
Closing in favor of #1884 which picks up where this PR left off.

Issue #1213, opened 2018-06-14
Automatically promote contents of noscript elements as acceptable replacement for adjoining script tag


westonruter:
If a theme or plugin outputs something like:

```html
Will this script run? <script>document.write('YES!');</script><noscript>No!</noscript>
```

It would make sense to me if the AMP plugin would automatically take the child elements of the `noscript` element to take the replace of the `script` element that is sanitized from the document. This would ensure that any fallback behavior would run as intended for browsers that have JS disabled.

For example, Google Tag Manager advises using the [following HTML](https://developers.google.com/tag-manager/quickstart):

```html
<!-- Google Tag Manager -->
<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
})(window,document,'script','dataLayer','GTM-XXXX');</script>
<!-- End Google Tag Manager -->

<!-- Google Tag Manager (noscript) -->
<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-XXXX"
height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
<!-- End Google Tag Manager (noscript) -->
```

When this gets served as AMP, then this could get all replaced with the contents of `noscript`, namely:

```html
<iframe src="https://www.googletagmanager.com/ns.html?id=GTM-XXXX"
height="0" width="0" style="display:none;visibility:hidden"></iframe>
```

This would make sense as part of the `AMP_Script_Sanitizer` as proposed in #1032. When the script sanitizer does this replacement, it could result in bypassing the pass-through of the `script` element to the whitelist sanitizer and thus avoid raising a blocking validation error. And again, the fallback behavior would be preserved.

postphotos:
Thanks @westonruter! Agreed here, thanks for opening this ticket.

Calling out what should be obvious (but hasn't been stated): `<noscript>` is indeed, permitted in AMP, and has been available for [quite a while](https://github.com/ampproject/amphtml/issues/1279).

westonruter:
[...]

PR #1226, opened 2018-06-25
Add AMP_Script_Sanitizer to replace noscript elements with their contents

westonruter:
As noted in [Implementing Interactivity](https://github.com/Automattic/amp-wp/wiki/Implementing-Interactivity):

> the baseline experience after enabling AMP on an existing theme is that it largely behaves as if JavaScript is disabled in the user's browser. Many themes and plugins already have fallbacks in case JavaScript is not available so that will serve them well in AMP. Therefore, most of the work entailed is to progressively enhance a theme for AMP to restore features that were lost during this graceful no-JS degradation to AMP

Nevertheless, this is not totally true at present: the plugin's postprocessor does not currently do any transformation of `noscript` elements meaning that the no-JS fallbacks will be not be successfully used in AMP pages. 

So this PR ensures that that a non-AMP document like:

```html
<script>document.write('Scripts!')</script><noscript><em>No scripts!</em></noscript>
```

Will get served in AMP as:

```html
<!--noscript--><em>No scripts!</em><!--/noscript-->
```

Thus it encourages the use of no-JS fallbacks for the baseline AMP experience as well as for visitors who have JS turned off in their browsers.

This will be further built upon in #1213 and #1032 with more intelligence about whether to promote the `noscript` element's contents or rather to convert a companion `script` to an AMP equivalent.

The base of this PR comes from https://github.com/Automattic/amp-wp/issues/1127

### DIFF
--- a/includes/sanitizers/class-amp-script-sanitizer.php
+++ b/includes/sanitizers/class-amp-script-sanitizer.php
@@ -28,7 +28,7 @@ class AMP_Script_Sanitizer extends AMP_Base_Sanitizer {
 		$noscripts = $this->dom->getElementsByTagName( 'noscript' );
 
 		for ( $i = $noscripts->length - 1; $i >= 0; $i-- ) {
-			$noscript = $noscripts[ $i ];
+			$noscript = $noscripts->item( $i );
 
 			// Skip AMP boilerplate.
 			if ( $noscript->firstChild instanceof DOMElement && $noscript->firstChild->hasAttribute( 'amp-boilerplate' ) ) {
--- a/includes/sanitizers/class-amp-script-sanitizer.php
+++ b/includes/sanitizers/class-amp-script-sanitizer.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * Class AMP_Script_Sanitizer
+ *
+ * @since 1.0
+ * @package AMP
+ */
+
+/**
+ * Class AMP_Script_Sanitizer
+ *
+ * @since 1.0
+ */
+class AMP_Script_Sanitizer extends AMP_Base_Sanitizer {
+
+	/**
+	 * Sanitize noscript elements.
+	 *
+	 * Eventually this should also handle script elements, if there is a known AMP equivalent.
+	 * If nothing is done with script elements, the whitelist sanitizer will deal with them ultimately.
+	 *
+	 * @todo Eventually this try to automatically convert script tags to AMP when they are recognized. See <https://github.com/Automattic/amp-wp/issues/1032>.
+	 * @todo When a script has an adjacent noscript, consider removing the script here to prevent validation error later. See <https://github.com/Automattic/amp-wp/issues/1213>.
+	 *
+	 * @since 1.0
+	 */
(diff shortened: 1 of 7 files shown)


includes/sanitizers/class-amp-script-sanitizer.php
```diff
@@ -0,0 +1,49 @@
...
+			if ( $noscript->firstChild instanceof DOMElement && $noscript->firstChild->hasAttribute( 'amp-boilerplate' ) ) {
+				continue;
+			}
+
+			$fragment = $this->dom->createDocumentFragment();
+			$fragment->appendChild( $this->dom->createComment( 'noscript' ) );
+			while ( $noscript->firstChild ) {
+				$fragment->appendChild( $noscript->firstChild );
+			}
+			$fragment->appendChild( $this->dom->createComment( '/noscript' ) );
```
kienstra: It's nice how this outputs the comment `<!--noscript-->`. This should help with debugging if needed, like on [WordPress.org support topics](https://wordpress.org/support/plugin/amp).

tests/test-amp-script-sanitizer.php
```diff
@@ -0,0 +1,98 @@
...
+		$whitelist_sanitizer = new AMP_Tag_And_Attribute_Sanitizer( $dom );
+		$whitelist_sanitizer->sanitize();
+		$content = AMP_DOM_Utils::get_content_from_dom( $dom );
+		$this->assertEquals( $expected, $content );
+	}
+
+	/**
+	 * Test style[amp-boilerplate] preservation.
+	 */
+	public function test_boilerplate_preservation() {
```
kienstra: Nice thorough testing of the boilerplate not being removed.

westonruter:
@kienstra thank you. I'm very confused about why the tests are failing. They all pass for me locally.

kienstra:
@westonruter, that's strange, the tests pass locally for me also.

westonruter:
Force-pushed a tweak. Maybe there was a caching problem. The test is failing as if these lines are not present:

```diff
--- a/tests/test-amp-script-sanitizer.php
+++ b/tests/test-amp-script-sanitizer.php
@@ -38,8 +38,6 @@ class AMP_Script_Sanitizer_Test extends WP_UnitTestCase {
 	 */
 	public function test_noscript_promotion( $source, $expected = null ) {
 		$dom       = AMP_DOM_Utils::get_dom_from_content( $source );
-		$sanitizer = new AMP_Script_Sanitizer( $dom );
-		$sanitizer->sanitize();
 		$whitelist_sanitizer = new AMP_Tag_And_Attribute_Sanitizer( $dom );
 		$whitelist_sanitizer->sanitize();
 		$content = AMP_DOM_Utils::get_content_from_dom( $dom );
```

westonruter:
I'm pretty sure the problem is due to an old version of libxml and this bit of logic:

https://github.com/Automattic/amp-wp/blob/70711f0cd647f233b02ffa1bbbd58496eeeea510/includes/utils/class-amp-dom-utils.php#L77-L110

PR #1235, opened 2018-06-29
Add support for allowing a site subset to be native AMP

### LINKED ISSUE #934: Add theme support scenarios: allow paired mode to serve some content exclusively in AMP
The plugin in `amp` theme support currently supports two modes:

1. Canonical: Everything is canonical AMP.
2. Paired: Everything is available in non-AMP and _some_ content is also available in AMP, with _different_ templates used in AMP mode.

There is a third case that should be supported and that is:

3. Mixed-Some-Optionally-AMP: All content is available in non-AMP, and some content is also available in AMP, but they use the _same theme templates_. 

Now that we have full-document sanitization in #929, the a non-AMP template can actually be served as totally valid AMP, with `amp` attribute added to the `html` element, meta viewport being added, and CSS boilerplate code. So we can actually use the exact same templates in AMP as non-AMP. We don't readily make this easy in paired mode, as to do paired mode the `amp` theme support must define a `template_dir` for where to locate the AMP templates, like:

```php
    'template_dir' => 'amp-templates/'
) );
```

When a paired-mode site only allows a subset of content to be shown in AMP (e.g. just singular posts), they can explicitly indicate this with the `available_callback` property:

```php
    'template_dir' => 'amp-templates/'
    'available_callback' => 'is_single',
) );
```

However, in order to support re-using templates between AMP and non-AMP we should also be able to do:

```php
    'available_callback' => 'is_single',
) );
```

Without this, you have to do a needless hack of setting the `template_dir` to `.` or add a `single.php` in the template directory that just does `require __DIR__ . '../single.php';`

Lastly, there is a fourth mode that should be doable:

4. Mixed-Some-Only-AMP: Some content is not available in AMP, and everything else is _only_ available in AMP.

In the fourth mode there should not be two versions of any one URL. There would be no `/amp/` endpoints, and no `amphtml` links on AMP content. Content in AMP and non-AMP would be disjoint sets. To do this, there would need to probably need to be something different from `available_callback` in the theme support, or rather instead of this callback returning a boolean flag it should actually be tri-state:

* AMP Omitted
* AMP Optional
* AMP Only

For example, if you wanted to serve AMP-only posts, AMP optionally for pages, and omit AMP everywhere else, you could do:

```php
    'available_callback' => function() {
        if ( is_single() ) {
            return AMP_ONLY; // No amphtml version offered. AMP 100%.
        } elseif (  is_page()) {
            return AMP_OPTIONAL; // Non-AMP by default. Link to amphtml URL offered to opt-in.
        } else {
            return AMP_OMITTED; // Non-AMP always. No link to amphtml URL offered at all.
        }
    }
) );
```

westonruter:
It's been a journey trying to figure out the best way to empower both end users and theme developers over which parts of their site are served as AMP, and whether AMP is served as canonical or is available in paired mode.

ℹ️  The commits in this PR will be squashed/rebased when merged, so please excuse the messy commit history in this branch.

When in native/paired template modes, the Post Types section of the admin screen is extended to also display a selection of the templates for which AMP can be served on the site:

image

The supportable templates are organized hierarchically according to the WordPress template hierarchy, so you enable the Category template for AMP but the general Archives template is disabled, then the Category template status will override the Archives template status.

The available templates can be amended via the `amp_supportable_templates` filter. This filter can also be used by a theme/plugin to force AMP to be enabled or disabled for a given template, such as here to forbid AMP from the Category templates (since maybe they have some required custom JS):

```php
add_filter( 'amp_supportable_templates', function( $templates ) {
	$templates['is_category']['supported'] = false;
	return $templates;
} );
```

Or to force AMP to be enabled for the author template:

```php
add_filter( 'amp_supportable_templates', function( $templates ) {
	$templates['is_author']['supported'] = false;
	return $templates;
} );
``` 

You can use this filter to add new templates as well. For example if you wanted to give the user control over the year archive template (take note of the `parent`):

```php
add_filter( 'amp_supportable_templates', function( $templates ) {
	$templates['is_year'] = array(
		'label'  => __( 'Year', 'example' ),
		'parent' => 'is_date',
	);
	return $templates;
} );
```

That then is presented as:

image

Lastly, there is a checkbox to automatically serve all templates and post types as AMP:

image

The availability of this option may will be determined whether or not a theme/plugin is forcibly setting the supported status for a given template.

You may also force all templates to be served as AMP at the theme support level:

```php
    'templates_supported' => 'all'
) );
```

Or force certain templates to be supported or unsupported:

```php
    'templates_supported' => array(
        'is_category' => false,
        'is_author' => true,
    )
) );
```

# Changelog Summary

* The template hierarchy is now exposed in the admin screen allowing the user to specify which kinds of templates should be served as AMP.
* The `amp_supportable_templates` filter allows you to customize the template hierarchy, add recognition for new template types, and force certain templates to be supported or forbidden in AMP.
* There is a new `templates_supported` arg on the `amp` theme support which can take a value of 'all' to force AMP to be available on all templates, or else an array mapping template IDs (e.g. `is_search`) to whether they are supported.
* The “Post” post type can now be turned off instead of always being forcibly enabled. 
* When `amp` theme support is present, the static homepage and page for posts are both enabled by default, as are pages with custom templates assigned.
* The template mode is now explicitly indicated via a `mode` theme support arg, which can either be `native` or `paired`.
* The `template_dir` can now be used in native mode as well.
* You can now define `add_theme_support( 'amp', array( 'mode' => 'native', 'optional' => true ) )` where the `optional` flag allows you to bypass actually activating the support in code, but forcing the mode to be `native` if they do add support. This allows the user to define `comments_live_list` in the theme without forcing AMP support to be present.
* Loading AMP component scripts has been restored on a native-mode site where AMP content appears on a non-AMP page.
* The `post_supports_amp()` logic has been consolidated into `AMP_Post_Type_Support::get_support_errors()`.
* ❗️ Breaking change: The `available_callback` has been eliminated.
* The restriction that `is_amp_endpoint()` cannot be called prior to `parse_query` action has been restored.

# Todo

- [x] Determine how a theme can force AMP to be used on all templates. If `mode` is `native` and AMP components are being used in all templates, then the theme needs to force all responses to be AMP. This could be done by a theme setting `supported` to true for all the `amp_supportable_templates`, but it seems there should be an easier way.
- [x] Add an “other” option to determine if AMP gets served by default for any templates not explicitly called out. This would be used in particular for `template_redirect` templates that are not loaded normally.
- [x] Address todos.
- [x] Should there be a way for a theme to explicitly mark templates as requiring for AMP other than by using the filter? An `all_templates_supported` flag on the theme support as a shortcut for using the filter? There could be a `templates_supported` arg when defining theme support which can take either `'all'` or an array of the template conditions which then become immutable.
- [x] Restore facilitation of serving AMP components in non-AMP documents if mode is native.

Fixes #934.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -296,7 +296,6 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  *         'template_dir' => 'my-amp-templates',
  *     ) );
  *
- * @todo Why do we even need this function?
  * @see \AMP_Theme_Support::is_paired_available()
  *
  * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is native and there is not separate AMP URL current URL.
@@ -319,18 +318,9 @@ function amp_is_canonical() {
 		if ( ! empty( $args['template_dir'] ) ) {
 			return false;
 		}
-//
-//		// Otherwise, the available_callback dictates whether AMP is available.
-//		if ( isset( $args['available_callback'] ) && is_callable( $args['available_callback'] ) ) {
-//			return call_user_func( $args['available_callback'] );
-//		}
 	}
 	return true;
 }
-//
-//function amp_is_available() {
-//
-//}
 
 /**
  * Load classes.
--- a/amp.php
+++ b/amp.php
@@ -295,15 +295,6 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  *          ),
  *      ) );
  *
- * Or if you want to prevent AMP from being used on the front page and on any Other template that is unrecognized:
- *
- *      add_theme_support( 'amp', array(
- *          'templates_supported' => array(
- *              'is_front_page' => false,
- *              'unrecognized' => false,
- *          ),
- *      ) );
- *
  * Or if you want to force AMP to be used on all templates:
  *
  *      add_theme_support( 'amp', array(
--- a/amp.php
+++ b/amp.php
@@ -301,6 +301,7 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  *          'templates_supported' => 'all',
  *      ) );
  *
+ * @see AMP_Theme_Support::read_theme_support()
  * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is native and there is not separate AMP URL current URL.
  */
 function amp_is_canonical() {
@@ -311,9 +312,22 @@ function amp_is_canonical() {
 	$mode    = 'native';
 	$support = get_theme_support( 'amp' );
 	if ( is_array( $support ) ) {
-		$args = array_shift( $support );
+		$args    = array_shift( $support );
+		$support = AMP_Options_Manager::get_option( 'theme_support' );
+
+		// If support is optional, look at DB option if mode is not explicitly set in theme support.
+		if ( ! empty( $args['optional'] ) ) {
+			if ( 'disabled' === $support ) {
+				return false;
+			} elseif ( ! isset( $args['mode'] ) ) {
+				return 'native' === $support;
+			}
+		}
+
 		if ( isset( $args['mode'] ) ) {
 			$mode = $args['mode'];
+		} elseif ( 'disabled' !== $support ) {
+			$mode = $support; // Supplied via admin screen.
 		} elseif ( ! empty( $args['template_dir'] ) ) {
 			$mode = 'paired'; // If there is a template_dir, then paired mode is implied.
 		}
--- a/amp.php
+++ b/amp.php
@@ -301,6 +301,10 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is native and there is not separate AMP URL current URL.
  */
 function amp_is_canonical() {
+	if ( ! current_theme_supports( 'amp' ) ) {
+		return false;
+	}
+
 	$support = get_theme_support( 'amp' );
 	if ( true === $support ) {
 		return true;
@@ -310,8 +314,8 @@ function amp_is_canonical() {
 		$args = array_shift( $support );
 
 		// If paired arg is supplied, then it is never canonical.
-		if ( ! empty( $args['paired'] ) ) {
-			return false;
+		if ( isset( $args['paired'] ) ) {
+			return ! $args['paired'];
 		}
 
 		// If there is a template_dir, then paired mode is implied.
--- a/amp.php
+++ b/amp.php
@@ -272,15 +272,36 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
 }
 
 /**
- * Whether this is in 'canonical mode.'
+ * Whether this is in 'canonical mode'.
  *
- * Themes can register support for this with `add_theme_support( 'amp' )`.
- * Then, this will change the plugin from 'paired mode,' and it won't use its own templates.
- * Nor output frontend markup like the 'rel' link. If the theme registers support for AMP with:
- * `add_theme_support( 'amp', array( 'template_dir' => 'my-amp-templates' ) )`
- * it will retain 'paired mode.
+ * Themes can register support for this with `add_theme_support( 'amp' )`, or via the
+ * following so that only so that single blog posts will be native/canonical, pages are paired,
+ * and everything else has AMP unavailable:
  *
- * @return boolean Whether this is in AMP 'canonical mode'.
+ *      add_theme_support( 'amp', array(
+ *          'template_dir'       => 'amp-templates/', // Optional. In case you need to override the template as a whole.
+ *          'available_callback' => function() {
+ *              // @todo Warning: If a plugin or theme calls is_amp_endpoint() before parse_query() then the conditionals will not work!
+ *              if ( is_single() ) {
+ *                  return 'native';
+ *              } elseif ( is_page() ) {
+ *                  return 'paired'; // Or 'true'.
+ *              } else {
+ *                  return false;
+ *              }
+ *          },
+ *      ) );
+ *
+ * Then, this will change the plugin so that it won't run in 'paired mode' with separate URLs.
+ * Neither will it output the rel=amphtml link on the frontend.
+ *
+ * Paired mode will be retained if the theme registers support for AMP with just a template_dir and no available_callback:
+ *
+ *     add_theme_support( 'amp', array(
+ *         'template_dir' => 'my-amp-templates',
+ *     ) );
+ *
+ * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is native and there is not separate AMP URL current URL.
  */
 function amp_is_canonical() {
 	$support = get_theme_support( 'amp' );
@@ -289,8 +310,21 @@ function amp_is_canonical() {
 	}
 	if ( is_array( $support ) ) {
 		$args = array_shift( $support );
-		if ( empty( $args['template_dir'] ) ) {
+
+		$is_native = (
+			isset( $args['available_callback'] )
+			&&
+			is_callable( $args['available_callback'] )
+			&&
+			'native' === call_user_func( $args['available_callback'] )
+		);
+		if ( $is_native ) {
 			return true;
+		}
+
+		// If there is no available_callback and yet there is a template_dir, then paired mode is implied.
+		if ( empty( $args['available_callback'] ) && ! empty( $args['template_dir'] ) ) {
+			return false;
 		}
 	}
 	return false;
--- a/amp.php
+++ b/amp.php
@@ -279,27 +279,25 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  * and everything else has AMP unavailable:
  *
  *      add_theme_support( 'amp', array(
- *          'template_dir'       => 'amp-templates/', // Optional. In case you need to override the template as a whole.
- *          'available_callback' => function() {
- *              // @todo Warning: If a plugin or theme calls is_amp_endpoint() before parse_query() then the conditionals will not work!
- *              if ( is_single() ) {
- *                  return 'native';
- *              } elseif ( is_page() ) {
- *                  return 'paired'; // Or 'true'.
- *              } else {
- *                  return false;
- *              }
- *          },
+ *          'available_callback' => 'is_single',
  *      ) );
  *
  * Then, this will change the plugin so that it won't run in 'paired mode' with separate URLs.
- * Neither will it output the rel=amphtml link on the frontend.
+ * Neither will it output the rel=amphtml link on the frontend. To serve a site in paired mode, add the paired arg:
  *
- * Paired mode will be retained if the theme registers support for AMP with just a template_dir and no available_callback:
+ *      add_theme_support( 'amp', array(
+ *          'paired' => true,
+ *          'available_callback' => 'is_single',
+ *      ) );
+ *
+ * Paired mode will also be retained if the theme registers support for AMP with just a template_dir and no available_callback:
  *
  *     add_theme_support( 'amp', array(
  *         'template_dir' => 'my-amp-templates',
  *     ) );
+ *
+ * @todo Why do we even need this function?
+ * @see \AMP_Theme_Support::is_paired_available()
  *
  * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is native and there is not separate AMP URL current URL.
  */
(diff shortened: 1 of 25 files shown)


includes/amp-helper-functions.php
```diff
@@ -239,80 +239,62 @@ function amp_add_amphtml_link() {
...
 	}
 
-	return false;
+	if ( current_theme_supports( 'amp' ) ) {
+		$template_available = true === AMP_Theme_Support::get_template_availability();
+		return amp_is_canonical() ? $template_available : $has_amp_query_var;
+	} else {
+
+		// Check if the queried object supports AMP.
+		if ( is_singular() || ( $wp_query instanceof WP_Query && $wp_query->is_posts_page ) ) {
```
kienstra: Good idea to check that `$wp_query->is_posts_page`

includes/class-amp-theme-support.php
```diff
@@ -133,26 +141,56 @@ public static function init() {
 		add_action( 'wp', array( __CLASS__, 'finish_init' ), PHP_INT_MAX );
 	}
 
+	/**
+	 * Determine whether theme support as added via option.
```
kienstra: A very minor point, but this might be:
>theme support **was** added

includes/class-amp-theme-support.php
```diff
@@ -303,13 +337,246 @@ public static function is_customize_preview_iframe() {
...
+		 *
+		 * A theme or plugin can force a given template to be supported or not by preemptively
+		 * setting the 'supported' flag for a given template. Otherwise, if the flag is undefined
+		 * then the user will be able to toggle it themselves in the admin.
+		 *
+		 * @since 1.0
+		 *
+		 * @param array $templates Supportable templates.
+		 */
+		$templates = apply_filters( 'amp_supportable_templates', $templates );
```
kienstra: This filter is already well-documented.

 But maybe it could also mention that each `$template` in `$templates` [can have](https://github.com/Automattic/amp-wp/pull/1235/files#diff-2585472f207548f626a43a7ff3cab922R386) an `$id` or `'callback'` that is a method of `WP_Query` (if I understand this right). Or it can simply be a `callable`.

includes/options/class-amp-options-menu.php
```diff
@@ -272,45 +275,148 @@ public function render_validation_handling() {
...
+			<ul>
+			<?php foreach ( array_map( 'get_post_type_object', AMP_Post_Type_Support::get_eligible_post_types() ) as $post_type ) : ?>
+				<li>
+					<?php $element_id = AMP_Options_Manager::OPTION_NAME . "-supported_post_types-{$post_type->name}"; ?>
+					<input
+						type="checkbox"
+						id="<?php echo esc_attr( $element_id ); ?>"
+						name="<?php echo esc_attr( $element_name ); ?>"
+						value="<?php echo esc_attr( $post_type->name ); ?>"
+						<?php checked( true, post_type_supports( $post_type->name, amp_get_slug() ) ); ?>
```
kienstra: The first argument of `checked()` probably isn't needed.

includes/options/class-amp-options-menu.php
```diff
@@ -272,45 +275,148 @@ public function render_validation_handling() {
...
+			<p class="description">
+				<?php esc_html_e( 'This will allow all of the URLs on your site to be served as AMP by default.', 'amp' ); ?>
+			</p>
+		</fieldset>
+
+		<fieldset id="singular_templates_fieldset">
+			<?php $element_name = AMP_Options_Manager::OPTION_NAME . '[supported_post_types][]'; ?>
+			<h4 class="title"><?php esc_html_e( 'Content Types', 'amp' ); ?></h4>
+			<p>
+				<?php esc_html_e( 'The following content types will be available as AMP by default, but you can override this on an item-by-item basis:', 'amp' ); ?>
```
kienstra: Maybe this would be enough:
>The following content types will be available as AMP:

If they're checked by default, that might communicate "will be available as AMP **by default**"

content-types

includes/class-amp-theme-support.php
```diff
@@ -303,13 +337,246 @@ public static function is_customize_preview_iframe() {
...
+	/**
+	 * Get the templates which can be supported.
+	 *
+	 * @return array Supportable templates.
+	 */
+	public static function get_supportable_templates() {
+		$templates = array(
+			'is_singular' => array(
+				'label'       => __( 'Singular', 'amp' ),
+				'description' => __( 'Required for the above content types.', 'amp' ),
```
kienstra: clicking-content-type

includes/class-amp-theme-support.php
```diff
@@ -118,7 +126,7 @@ public static function init() {
 			$args = array_shift( $support );
 			if ( ! is_array( $args ) ) {
 				trigger_error( esc_html__( 'Expected AMP theme support arg to be array.', 'amp' ) ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
```
kienstra: This appears on the AMP Settings page `/wp-admin/admin.php?page=amp-options`
>( ! ) Warning: array_merge(): Argument #2 is not an array in /srv/www/loc/public_html/wp-content/plugins/amp/includes/class-amp-theme-support.php on line 189
>( ! ) Notice: Expected AMP theme support arg to be array. in /srv/www/loc/public_html/wp-content/plugins/amp/includes/class-amp-theme-support.php on line 128

This occurs with Twenty Seventeen. That would make sense, as it doesn't do `add theme_support( 'amp' )`, so `$support` is `array()`

And `$args` is null, because `$args = array_shift( $support );`

includes/options/class-amp-options-menu.php
westonruter: Actually they would be available as AMP by default, no?

rheinardkorf:
@westonruter works really well with custom templates in plugins as well.  Thanks for doing this.

westonruter:
Ready for review and testing. With this merged, I believe we're ready for beta.

amp.php
```diff
@@ -272,28 +272,53 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
...
 function amp_is_canonical() {
-	$support = get_theme_support( 'amp' );
-	if ( true === $support ) {
-		return true;
+	if ( ! current_theme_supports( 'amp' ) ) {
+		return false;
 	}
+
+	$mode    = 'native';
+	$support = get_theme_support( 'amp' );
```
amedina: May there be a more differentiating naming for `current_theme_supports` vs. `get_theme_support`? Probably I am missing some of the context but the naming appear a little confusing.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -43,6 +43,22 @@ public function init() {
 		if ( function_exists( 'gutenberg_init' ) ) {
 			add_action( 'enqueue_block_editor_assets', array( $this, 'enqueue_block_editor_assets' ) );
 			add_filter( 'wp_kses_allowed_html', array( $this, 'whitelist_block_atts_in_wp_kses_allowed_html' ), 10, 2 );
+
+			/*
+			 * Dirty AMP is required when a site is in native mode but not all templates are being served
```
amedina: This comment is hard to follow; perhaps not mentioning "Dirty AMP" but explaining directly the semantics of the filter and action functions may be clearer. Just a thought.

amp.php
westonruter: The `current_theme_supports()` and `get_theme_support()` are WordPress core functions. They are somewhat confusing. The former returns whether support has been added (boolean), _and_ applies a filter to let plugins control whether support is present. The `get_theme_support()` on the other hand does not apply a filter, but rather it returns the args that were passed in for the original `add_theme_support()` call.

includes/admin/class-amp-post-meta-box.php
```diff
@@ -143,12 +143,20 @@ public function enqueue_admin_assets() {
...
+			$support_errors = AMP_Post_Type_Support::get_support_errors( $post );
+		}
+
 		wp_add_inline_script( self::ASSETS_HANDLE, sprintf( 'ampPostMetaBox.boot( %s );',
 			wp_json_encode( array(
 				'previewLink'     => esc_url_raw( add_query_arg( amp_get_slug(), '', get_preview_post_link( $post ) ) ),
 				'canonical'       => amp_is_canonical(),
-				'enabled'         => post_supports_amp( $post ),
-				'canSupport'      => count( AMP_Post_Type_Support::get_support_errors( $post ) ) === 0,
+				'enabled'         => empty( $support_errors ),
```
amedina: Very nice semantics.

includes/admin/class-amp-post-meta-box.php
```diff
@@ -170,23 +178,32 @@ public function render_status( $post ) {
...
-			! amp_is_canonical()
 		);
 
 		if ( true !== $verify ) {
 			return;
 		}
 
-		$errors = AMP_Post_Type_Support::get_support_errors( $post );
-		$status = post_supports_amp( $post ) ? self::ENABLED_STATUS : self::DISABLED_STATUS;
+		if ( current_theme_supports( 'amp' ) ) {
```
amedina: Would it be possible having a comment for this `if-then-else`?

includes/options/class-amp-options-menu.php
kienstra: Yes, they would be available by default. But I was thinking that it might not be necessary to mention that. 

But on second thought, that description is probably helpful. Most people probably don't want to disable AMP on posts, for example.

includes/admin/class-amp-post-meta-box.php
westonruter: To explain why the condition is here?

includes/class-amp-theme-support.php
kienstra: Thanks, this looks good.

includes/options/class-amp-options-menu.php
kienstra: Thanks, this looks good.

westonruter:
I realized that we don't need to remove `available_callback`. We can still support for legacy installs (0.7) it but mark it as deprecated. In short, we just need to add this to the top of `\AMP_Theme_Support::get_template_availability()`:

```php
		// Support available_callback from 0.7, though it is deprecated.
		if ( isset( $theme_support_args['available_callback'] ) && is_callable( $theme_support_args['available_callback'] ) ) {
			_doing_it_wrong( 'add_theme_support', esc_html__( 'The available_callback is deprecated when adding amp theme support in favor of declaratively setting the supported_templates.', 'amp' ), '1.0' );
			return call_user_func( $theme_support_args['available_callback'] );
		}
```

I'll do that.

includes/options/class-amp-options-menu.php
```diff
@@ -272,45 +281,168 @@ public function render_validation_handling() {
...
+		<fieldset id="supported_post_types_fieldset">
+			<?php $element_name = AMP_Options_Manager::OPTION_NAME . '[supported_post_types][]'; ?>
+			<h4 class="title"><?php esc_html_e( 'Content Types', 'amp' ); ?></h4>
+			<p>
+				<?php esc_html_e( 'The following content types will be available as AMP:', 'amp' ); ?>
+			</p>
+			<ul>
+			<?php foreach ( array_map( 'get_post_type_object', AMP_Post_Type_Support::get_eligible_post_types() ) as $post_type ) : ?>
+				<li>
+					<?php $element_id = AMP_Options_Manager::OPTION_NAME . "-supported_post_types-{$post_type->name}"; ?>
```
kienstra: **Media Pages In "Classic" Mode**

This isn't the most common use-case. But it looks like on the "Media" pages in "Classic" mode, the `href` in the `<link rel="amphtml">` leads to a 404 page. It ends with `/amp`, which results in a 404 page. But the `?amp` URL works.

Steps to reproduce:
1. Activate Twenty Sixteen
2. On the "AMP Settings" page, select "Template Mode" > "Classic," and check "Supported Templates" > "Media"
classic-media-templates
3. In the Media Library, click an image, and click "View Attachment Page":
media-page

4. In the source of that page, the `<link rel="amphtml">` has an `href` with `/amp`, not `?amp`:

```html
<link rel="amphtml" href="https://loc.test/lorem-2000bridge/amp/">
```
source-media-page

5. Expected: that URL is a valid AMP page
6. Actual: it's a 404 page:
ends-amp

7. Going to the same URL, but ending in `?amp` results in a valid AMP page:
ends-amp-query-var

kienstra: **Media Pages In "Classic" Mode**

This isn't the most common use-case. But it looks like on the "Media" pages in "Classic" mode, the `href` in the `<link rel="amphtml">` leads to a 404 page. It ends with `/amp`, which results in a 404 page. But the `?amp` URL works.

Steps to reproduce:
1. Activate Twenty Sixteen
2. On the "AMP Settings" page, select "Template Mode" > "Classic," and check "Supported Templates" > "Media"
classic-media-templates
3. In the Media Library, click an image, and click "View Attachment Page":
media-page

4. In the source of that page, the `<link rel="amphtml">` has an `href` with `/amp`, not `?amp`:

```html
<link rel="amphtml" href="https://loc.test/lorem-2000bridge/amp/">
```
source-media-page

5. Expected: that URL is a valid AMP page
6. Actual: it's a 404 page:
ends-amp

7. Going to the same URL, but ending in `?amp` results in a valid AMP page:
ends-amp-query-var

This seems to exist even after flushing permalinks.

westonruter:
Using `available_callback` now works again but it is deprecated, and the admin shows:

image

includes/options/class-amp-options-menu.php
westonruter: Good catch. I believe this is fixed after 7091b35.

kienstra: Thanks, your commit [7091b35](https://github.com/Automattic/amp-wp/commit/7091b35ec7f316680cf5d1c6cf77a66dd6aa47b8) fixed this, and the `<link rel="amphtml">` now leads to a valid AMP page.

includes/class-amp-theme-support.php
jwold: @kienstra I like the idea of making it simpler for people, so that if they select a content type the singular template would be auto selected.

Is there ever a situation where I would select a singular template and NOT select a content type?

westonruter: @jwold Yes, it would probably be common to have the Singular template checked but not have all Content Types checked. What would probably not be as common is to have Singular unchecked but the Content Types checked. There is a scenario where it could be used though: maybe the post archive page for a given CPT can be served as AMP but the singular CPT template cannot. That would surely be rare, but even so the “Required“ comment above isn't really true.

Issue #1245, opened 2018-07-03
"Template Mode" is set to "Native," and <input> elements are disabled


kienstra:
This might be the intended behavior, but when using a Core theme like Twenty Seventeen, the "Template Mode" is always "Native," and the `<input>` elements are disabled.

**Steps to reproduce:**
1. Activate Twenty Seventeen
2. Go to the "AMP Settings Page"
3. Expected (I think): you can select the "Template Mode" of "Classic," "Paired," or "Native"
4. Actual: "Native" is selected, and the `<input>` elements are disabled:

<img width="1257" alt="native-selected" src="https://user-images.githubusercontent.com/4063887/42242255-dee26782-7ed2-11e8-8658-617e9136bf3d.png">

This existed even after doing `wp option delete amp-options`, to ensure that a previous option I selected wasn't interfering.

Maybe the [disabled() calls](/Users/ryankienstra/Projects/vvv2/www/loc/public_html/wp-content/plugins/amp/includes/options/class-amp-options-menu.php) should be `disabled( ! $mode_mutable )` instead of `disabled( ! $theme_support_mutable )`.

westonruter:
Fixed by #1246.

PR #1246, opened 2018-07-03
Use $mode_mutable to only disable Template Mode <input> elements in some cases.

kienstra:
**Request For Review**

Hi @westonruter,
Could you please review this PR for #1245? It might not be the correct solution. 

But it worked locally to avoid disabling the `<input>` elements for the Template Modes.

If I understand this right, it looks like the `<input>` elements for "Classic," "Paired," and "Native" should only be disabled (or hidden) if the theme passes a `'mode'` argument, like:
```php
	'mode' => 'native',
) );
```
Then, it would look like:
theme-support-mode

Still, I might not have understood the intent.

### DIFF
--- a/includes/options/class-amp-options-menu.php
+++ b/includes/options/class-amp-options-menu.php
@@ -154,7 +154,7 @@ class AMP_Options_Menu {
 			<?php endif; ?>
 			<dl>
 				<dt>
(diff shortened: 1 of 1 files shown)


kienstra:
Also, even with this PR, it's not possible to select "Classic," and see it after clicking "Save Changes." 

**Steps To Reproduce**

1. Activate Twenty Seventeen
2. Go to the "AMP Settings" page
3. Select "Template Mode" > "Classic"
4. Click "Save Changes"
5. Expected: "Classic" is selected
6. Actual: "Native" is selected

native-amp-settings

It looks like it saves the "Classic" option correctly (as `'disabled'`).

change-theme-support

kienstra:
The [issue above](https://github.com/Automattic/amp-wp/pull/1246#issuecomment-402288725) is addressed in #1248.

PR #1247, opened 2018-07-04
Keep the Template Mode of Classic if selected, and the theme doesn't override it

kienstra:
**Request For Review**

Hi @westonruter,
Could you please review this, which addresses the issue in [#1246 (comment)](https://github.com/Automattic/amp-wp/pull/1246#issuecomment-402288725)? There are steps to reproduce there.

This might not be the right solution. But this allows displaying the Template Mode of "Classic." 

It also seems to still have the right behavior if a theme adds a `'mode'` argument, like:

```php
	'comments_live_list' => true,
	'mode'               => 'native',
) );
```

In that example, it forces Native mode, even if Classic (disabled) is actually saved as the option. This is the same behavior as before this commit.

theme-native-forced

Still, I might have overlooked other use-cases, or the intent of that line of code.

### DIFF
--- a/includes/options/class-amp-options-menu.php
+++ b/includes/options/class-amp-options-menu.php
@@ -136,7 +136,7 @@ class AMP_Options_Menu {
 
 		$mode_mutable = ! isset( $support_args['mode'] );
 
-		if ( ! $theme_support_mutable || ( ! $mode_mutable && 'disabled' !== $theme_support ) ) {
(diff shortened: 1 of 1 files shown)


westonruter:
I think this will be made obsolete by #1248.

kienstra:
Thanks, closing in favor of #1248.

PR #1248, opened 2018-07-04
Simplify theme support args

### LINKED ISSUE #1238: Decide whether declared theme support should include 'optional'
In #1235 support was added for theme support to be declared but optionally. For example:

```php
    'optional' => true,
    'comments_live_list' => true,
) )
```

The idea here is that this would allow a theme to pre-configure a theme to have support but for the plugin to not actually apply the `amp` theme support unless the user opts-in via the admin screen. This would allow a user to activate such a theme but retain the Classic template mode until they decide they want it. This doesn't seem like a very compelling use case and it could be eliminated.

westonruter:
* Eliminate `optional` flag when adding `amp` theme support. If theme support is present, then the admin screen will show it is enabled without a way to turn it off. If no `amp` theme support flag is present, then the radio buttons are presented to toggle the mode.
* Eliminate the ability to `add_theme_support('amp')` and then let the `native` and `paired` modes be toggled between. If `add_theme_support('amp')` is present, then AMP is made available in native mode.
* Instead of passing a `mode` theme support arg, switch to a simpler `paired` toggle. For example, `add_theme_support('amp', array('paired' => true))`. You can still use AMP-specific templates in native mode via `add_theme_support('amp', array('paired' => false, 'template_dir' => 'amp'))`, though this should be uncommon.
* Fix issue with paired templates not being accessible in Customizer preview.
* Prevent response caching in Customizer preview.
* Redirect to AMP-less variable if AMP is not available for this URL and yet the query var is present.

# Paired mode set by theme support

image

# Native mode set by theme support

image

# Theme support not present

image

Fixes #1238

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -278,12 +278,24 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  *
  *      add_theme_support( 'amp' );
  *
- * This will serve templates in native AMP by default but the user would be able to change the template mode
- * from native to paired in the admin. To force only native to be available, such as when you are using AMP components
- * in your theme templates, do:
+ * This will serve templates in native AMP, allowing you to use AMP components in your theme templates.
+ * If you want to make available in paired mode, where templates are served in AMP or non-AMP documents, do:
  *
  *      add_theme_support( 'amp', array(
- *          'mode' => 'native',
+ *          'paired' => true,
+ *      ) );
+ *
+ * Paired mode is also implied if you define a template_dir:
+ *
+ *      add_theme_support( 'amp', array(
+ *          'template_dir' => 'amp',
+ *      ) );
+ *
+ * If you want to have AMP-specific templates in addition to serving native AMP, do:
+ *
+ *      add_theme_support( 'amp', array(
+ *          'paired'       => false,
+ *          'template_dir' => 'amp',
  *      ) );
  *
  * If you want to force AMP to always be served on a given template, you can use the templates_supported arg,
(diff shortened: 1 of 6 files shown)


amp.php
```diff
@@ -309,30 +321,13 @@ function amp_is_canonical() {
...
-		if ( isset( $args['mode'] ) ) {
-			$mode = $args['mode'];
-		} elseif ( 'disabled' !== $support ) {
-			$mode = $support; // Supplied via admin screen.
-		} elseif ( ! empty( $args['template_dir'] ) ) {
-			$mode = 'paired'; // If there is a template_dir, then paired mode is implied.
-		}
+	$args = AMP_Theme_Support::get_theme_support_args();
+	if ( isset( $args['paired'] ) ) {
+		return empty( $args['paired'] );
```
kienstra: Good idea to use the `paired` flag for theme support.

includes/class-amp-theme-support.php
```diff
@@ -144,107 +131,77 @@ public static function init() {
...
-				if ( isset( self::$initial_theme_support_args['available_callback'] ) ) {
-					_doing_it_wrong( 'add_theme_support', esc_html__( 'The available_callback is deprecated when adding amp theme support in favor of declaratively setting the supported_templates.', 'amp' ), '1.0' );
-				}
+			$args = self::get_theme_support_args();
+
+			// Validate theme support usage.
+			$keys = array( 'template_dir', 'comments_live_list', 'paired', 'templates_supported', 'available_callback' );
+
+			if ( count( array_diff( array_keys( $args ), $keys ) ) !== 0 ) {
+				_doing_it_wrong( 'add_theme_support', esc_html( sprintf(  // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
```
kienstra: Good idea output this notice.

Issue #1230, opened 2018-06-27
Add AMP enabled/disabled toggle to Gutenberg editor


westonruter:
The classic post editor has a toggle to enable/disable AMP for a given post:

![image](https://user-images.githubusercontent.com/134745/41952109-1569d2d6-7983-11e8-95b3-e3e097108b66.png)

This needs to be added to Gutenberg as well.

westonruter:
@jwold This needs some Gutenberg design insights.

westonruter:
The AMP-specific blocks should not be made available on the Inserter when AMP is disabled for a given post, even in native mode.

jwold:
@westonruter what would happen in this scenario:

1. Enable AMP in a Gutenberg post
2. Add an AMP-specific block
3. Disable Gutenberg
4. What happens to that block that's already on the page? Should we display a notice around it that it's incompatible, should we gray it out?

jwold:
![screen shot on 2018-07-10 at 08_04_29](https://user-images.githubusercontent.com/3411173/42519016-f1b2f68e-8417-11e8-8384-812dc4549a29.png)

For the interface in Gutenberg I'm thinking something as simple as this ^

Any concerns?

westonruter:
@jwold that mock looks good to me. I'm assuming Gutenberg has is extensible to allow new toggles to be added in that location.

> What happens to that block that's already on the page? Should we display a notice around it that it's incompatible, should we gray it out?

Good question. In the case where you've added an AMP-specific block to content but then you disable AMP, the next time you access the post then the block would be unrecognized and it would show the same notice as if a block from a plugin were in content but then the plugin was deactivated:

![image](https://user-images.githubusercontent.com/134745/42525790-e4b118d8-8428-11e8-8471-aaac2ea89d80.png)

This seems acceptable to me.

jwold:
> In the case where you've added an AMP-specific block to content but then you disable AMP, the next time you access the post then the block would be unrecognized and it would show the same notice as if a block from a plugin were in content but then the plugin was deactivated:

I'm good with this approach, my only question is if we should change the wording at all to be specific, such as, 

"This block requires enabling AMP in the document settings"
[ Re-enable AMP button ]   [ Edit as HTML button ]

westonruter:
That screenshot is just the default view as presented in Gutenberg for a block that is no longer recognized. I'm not sure how easy it is to customize it for specific blocks. That would be nice for the AMP case here.

[...]

PR #1252, opened 2018-07-05
Add filter to render admin metabox.

schrapel:
We want to sometimes hide the metabox entirely, so would need this filter added.

### DIFF
--- a/includes/admin/class-amp-post-meta-box.php
+++ b/includes/admin/class-amp-post-meta-box.php
@@ -180,7 +180,7 @@ class AMP_Post_Meta_Box {
 			current_user_can( 'edit_post', $post->ID )
 		);
 
-		if ( true !== $verify ) {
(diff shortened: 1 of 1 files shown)


westonruter:
Is the reason for this PR to fix a similar issue as is reported in #1011?

It would be preferable if one could just do:

```php
remove_action( 'post_submitbox_misc_actions', array( $post_meta_box, 'render_status' ) ); // ⛔️
```

 

But the problem is that the instantiated `AMP_Post_Meta_Box` class is not stored in a public variable in the `amp_post_meta_box()` function.

For now I would recommend you just hide it with CSS as needed:

```css
.misc-pub-section.misc-amp-status { display:none; }
```

We're going to need to address this for Gutenberg as well in #1230, and I don't know if a filter is going to be the right approach. I feel that adding a new filter just for this purpose of hiding the metabox is perhaps missing a deeper fix that is needed, so I'm going to close this for now.

PR #1261, opened 2018-07-08
Install service worker to cache AMP CDN assets, cache Google Fonts, and cache images (following AMP by Example)

### LINKED ISSUE #1210: Use service worker to cache AMP component scripts and other assets
Once we have a working prototype of registering service worker logic to cache assets in the [pwa-wp](https://github.com/xwp/pwa-wp) feature plugin, we should immediately leverage this to cache AMP component scripts. This will allow AMP to components to used offline, such as to do development on a plane.

westonruter:
Changes:

* In AMP responses, prevents outputting the `script` for installing the service worker as output by PWA-WP and instead outputs an `amp-install-service-worker` tag instead.
* Add an endpoint specifically for handling the iframe-based service worker installation from AMP Cache.
* Implement remaining aspects of @sebastianbenz's [minimal service worker for websites built with AMP](https://github.com/ampproject/amp-by-example/blob/master/boilerplate-generator/templates/files/serviceworkerJs.js) which aren't already implemented in the PWA plugin, namely:
  * `cdn_script_caching`: The runtime precaching of AMP runtime and key component scripts; using stale-while-revalidate for those and other CDN assets that are loaded at runtime.
  * `google_fonts_caching`: Cache-first strategy for images. Disabled by default.
  * `image_caching`: Caching of Google Fonts, unless the PWA plugin's integration isn't also running. Disabled by default.

To enable all integrations, add to your theme:

```php
add_theme_support( 'amp', [
	'service_worker' => [
		'cdn_script_caching'   => true,
		'google_fonts_caching' => true,
		'image_caching'        => true,
	],
] );
```

To disable all integrations, you can also do:

```php
add_theme_support( 'amp', [
	'service_worker' => false,
] );
```

With this PR and the PWA-WP feature plugin, running Twenty Sixteen in my development environment yields these Lighthouse results:

image

# Todo

- [x] Add tests
- [x] Implement caching with Workbox 
- [x] “SWR still needs some reasonable expiry (say 8 or 30 days) and https://fonts.gstatic.com  should probably use cacheFirst instead of SWR.” (@cramforce [tweet](https://twitter.com/cramforce/status/1016700943130214400))
- [ ] ~“consider using networkFirst for documents. This might be controversial, but it would be amazing. Trying that for AMP only might be a good first step to rolling it out to the rest of WP.” (@cramforce [tweet](https://twitter.com/cramforce/status/1016700943130214400))~ (We can address this with integration of amp-sw.)

Fixes #1210 

Depends on:

* https://github.com/xwp/pwa-wp/pull/14 ✅ 
* https://github.com/xwp/pwa-wp/pull/27 ✅
* https://github.com/xwp/pwa-wp/pull/53 ✅
* https://github.com/xwp/pwa-wp/pull/57 ✅

Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/2950523/amp.zip) (v1.1-alpha-20190311T031056Z-a1cb0cdc)


includes/class-amp-service-workers.php
```diff
@@ -0,0 +1,146 @@
...
+	 * @see wp_print_service_workers()
+	 * @link https://www.ampproject.org/docs/reference/components/amp-install-serviceworker#data-iframe-src-(optional)
+	 */
+	public static function handle_service_worker_iframe_install() {
+		if ( empty( $GLOBALS['wp']->query_vars[ self::INSTALL_SERVICE_WORKER_IFRAME_QUERY_VAR ] ) ) {
+			return;
+		}
+
+		$scope = $GLOBALS['wp']->query_vars[ self::INSTALL_SERVICE_WORKER_IFRAME_QUERY_VAR ];
+		if ( 'front' !== $scope && 'admin' !== $scope ) {
```
westonruter: This needs to be updated to use the `WP_Service_Worker::SCOPE_*` constants.

includes/class-amp-service-workers.php
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	/**
+	 * Handle request to install service worker via iframe.
+	 *
+	 * @see wp_print_service_workers()
+	 * @link https://www.ampproject.org/docs/reference/components/amp-install-serviceworker#data-iframe-src-(optional)
+	 */
+	public static function handle_service_worker_iframe_install() {
+		if ( empty( $GLOBALS['wp']->query_vars[ self::INSTALL_SERVICE_WORKER_IFRAME_QUERY_VAR ] ) ) {
```
westonruter: This can use `isset()` as shown in the feature plugin.

includes/class-amp-service-workers.php
```diff
@@ -0,0 +1,146 @@
...
+		}
+
+		if ( empty( $front_handles ) ) {
+			return; // No service worker scripts are installed.
+		}
+
+		$src        = wp_get_service_worker_url( 'front' );
+		$iframe_src = add_query_arg(
+			self::INSTALL_SERVICE_WORKER_IFRAME_QUERY_VAR,
+			'front',
```
westonruter: This needs to be updated to use the `WP_Service_Worker::SCOPE_FRONT` constant.

includes/class-amp-service-workers.php
```diff
@@ -0,0 +1,146 @@
...
+	 */
+	public static function install_service_worker() {
+		if ( ! function_exists( 'wp_service_workers' ) || ! function_exists( 'wp_get_service_worker_url' ) ) {
+			return;
+		}
+
+		// Get the frontend-scoped service worker scripts.
+		$front_handles = array();
+		foreach ( wp_service_workers()->registered as $handle => $item ) {
+			if ( 'all' === $item->args['scope'] || 'front' === $item->args['scope'] ) {
```
westonruter: This needs to be updated to use the `WP_Service_Worker::SCOPE_FRONT` constant, in a bitwise `&` condition.

westonruter:
I'm removing this from 1.0. It depends on the PWA plugin and/or core support, so this is something that should land later when WP support for service workers are more mature.

westonruter:
Important [note](https://docs.google.com/document/d/169XUxtSSEJb16NfkrCr9y5lqhUR7vxXEAsNxBzg07fM/edit#heading=h.1qclfiu0nxi5) from @sebastianbenz in regards to whether a service worker should be used if a site is not a single-page app:

> If none of these patterns are used and it’s not possible to cache the whole site (which only is reasonable for very small sites), a serviceworker might have a [negative performance impact](https://developers.google.com/web/updates/2017/02/navigation-preload). The best thing in this case is to **not** use a serviceworker.

In this case, if the service worker is merely being used to cache AMP CDN assets then the service worker should only be used during development (when `WP_DEBUG` is enabled).

postphotos:
This is fantastic, @westonruter! Those Lighthouse scores are quite impressive. Looking forward to seeing this merge in as PWA advances forward in development.

felixarntz:
This needs a few minor adjustments after https://github.com/xwp/pwa-wp/pull/57:
* Callbacks for `wp_front_service_worker` now receive a `WP_Service_Worker_Registry`. The methods for cached and precached routes are now part of that class, so no adjustments are needed in that regard.
* The constants for cache strategies and cache names are now part of `WP_Service_Worker_Registry`, so these references need to be adjusted.

felixarntz:
We should probably wait with merge until the follow-up discussion in https://github.com/xwp/pwa-wp/pull/57 has led to a decision.

kienstra:
**Question About Moving To v1.1 Board**

Hi @westonruter,
It's nice how your PR #1386 applies some of this for the `1.0` release.

Would you mind if I moved this to the [v1.1](https://github.com/Automattic/amp-wp/projects/7) board?

Thanks!

westonruter:
@kienstra yes, you're right. It's not a 1.0 board item anymore. I removed it.

westonruter:
I'm picking this up again so we can get it merged as part of 1.1.

westonruter:
I pushed the code up to http://pwa-westonruter.pantheonsite.io/ as a demo.

westonruter:
Ideally we would incorporate functionality from amp-sw: https://github.com/ampproject/amp-sw

westonruter:
For facilitating the incorporation of `amp-sw.js` into the AMP plugin, see https://github.com/ampproject/amp-sw/issues/12

includes/class-amp-service-worker.php
```diff
@@ -0,0 +1,316 @@
...
+	/**
+	 * Init.
+	 */
+	public static function init() {
+		if ( ! class_exists( 'WP_Service_Workers' ) ) {
+			return;
+		}
+
+		$theme_support = AMP_Theme_Support::get_theme_support_args();
+		if ( empty( $theme_support['service_worker'] ) ) {
```
westonruter: I wonder if we should opt-in or opt-out? If they have the PWA plugin, perhaps this should be enough of a signal that they want this functionality? In that case, this could be:

```suggestion
		if ( isset( $theme_support['service_worker'] ) && false === $theme_support['service_worker'] ) {
```

westonruter:
The amp-sw integration shouldn't block this from being merged, nevertheless.

includes/class-amp-service-worker.php
```diff
@@ -0,0 +1,316 @@
...
+	/**
+	 * Add runtime caching of Google Fonts with stale-while-revalidate strategy for stylesheets and cache-first strategy for webfont files.
+	 *
+	 * @link https://developers.google.com/web/tools/workbox/guides/common-recipes#google_fonts
+	 * @link https://github.com/ampproject/amp-by-example/blob/4593af61609898043302a101826ddafe7206bfd9/boilerplate-generator/templates/files/serviceworkerJs.js#L76-L103
+	 * @link https://github.com/xwp/pwa-wp/blob/master/integrations/class-wp-service-worker-fonts-integration.php
+	 *
+	 * @param WP_Service_Worker_Scripts $service_workers Service workers.
+	 */
+	public static function add_google_fonts_caching( $service_workers ) {
```
felixarntz: I think this method should be removed as the PWA plugin is already taking care of this functionality. Since we integrate with it anyway, I think we should rely on its features as much as possible.

Together with the other AMP integrations relying on the PWA constant anyway, this would be taken care of already.

includes/class-amp-service-worker.php
```diff
@@ -0,0 +1,316 @@
...
+		);
+
+		$theme_support = AMP_Theme_Support::get_theme_support_args();
+		if ( ! empty( $theme_support['comments_live_list'] ) ) {
+			$precached_handles[] = 'amp-live-list';
+		}
+
+		if ( amp_get_analytics() ) {
+			$precached_handles[] = 'amp-analytics';
+		}
```
felixarntz: Couldn't we cover other AMP extensions here as well? What would be the downsides of including them in this list unconditionally?

Alternatively, we could at least make the list filterable.

includes/class-amp-service-worker.php
```diff
@@ -0,0 +1,316 @@
...
+	/**
+	 * Add runtime caching for scripts loaded from the AMP CDN with a stale-while-revalidate strategy.
+	 *
+	 * @link https://github.com/ampproject/amp-by-example/blob/4593af61609898043302a101826ddafe7206bfd9/boilerplate-generator/templates/files/serviceworkerJs.js
+	 *
+	 * @param WP_Service_Worker_Scripts $service_workers Service worker registry.
+	 */
+	public static function add_cdn_script_caching( $service_workers ) {
+		if ( ! ( $service_workers instanceof WP_Service_Worker_Scripts ) ) {
+			_doing_it_wrong( __METHOD__, esc_html__( 'Expected argument to be WP_Service_Worker_Cache_Registry.', 'amp' ), '1.1' );
```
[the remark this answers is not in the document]
westonruter: I believe the reason for this change is because the argument was different in previous iteration of the plugin, so this is to prevent a fatal error from happening.

includes/class-amp-service-worker.php
westonruter: The downside would be needlessly precaching AMP scripts and wasting the user's storage.

The intention here is that the service worker can automatically precache scripts that we know will be used purely based on configuration. If a theme wants to precache other scripts then they could do so via:

```php
add_action(
	'wp_front_service_worker',
	function ( WP_Service_Worker_Scripts $sw_scripts ) {
		$sw_scripts->precaching_routes()->register( wp_scripts()->registered['amp-carousel']->src );
	}
);
```

Alternatively, we could include a `precached_scripts` in the theme support args which takes a list of script handles, which would be more user friendly.

includes/class-amp-service-worker.php
```diff
@@ -0,0 +1,316 @@
...
+	/**
+	 * Add runtime image caching from the origin with a cache-first strategy.
+	 *
+	 * @link https://github.com/ampproject/amp-by-example/blob/4593af61609898043302a101826ddafe7206bfd9/boilerplate-generator/templates/files/serviceworkerJs.js#L60-L74
+	 *
+	 * @param WP_Service_Worker_Scripts $service_workers Service workers.
+	 */
+	public static function add_image_caching( $service_workers ) {
+		if ( ! ( $service_workers instanceof WP_Service_Worker_Scripts ) ) {
+			_doing_it_wrong( __METHOD__, esc_html__( 'Expected argument to be WP_Service_Worker_Scripts.', 'amp' ), '1.1' );
```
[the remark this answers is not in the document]
westonruter: See above. How about replacing with a type hint in the next release?

Doing it this way also gives us more flexibility to modify the PWA plugin's API without causing a fatal error in the AMP plugin.

includes/class-amp-service-worker.php
```diff
@@ -0,0 +1,316 @@
...
+			return;
+		}
+
+		$service_workers->caching_routes()->register(
+			'^' . preg_quote( set_url_scheme( content_url( '/' ), 'https' ), '/' ) . '[^\?]+?\.(?:png|gif|jpg|jpeg|svg|webp)(\?.*)?$',
+			array(
+				'strategy'  => WP_Service_Worker_Caching_Routes::STRATEGY_CACHE_FIRST,
+				'cacheName' => 'images', // @todo This needs to get the proper prefix in JS.
```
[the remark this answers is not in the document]
westonruter: This is somewhat of an open question, which is beyond the scope of this PR really. I can move it to a PWA issue.  See https://github.com/xwp/pwa-wp/issues/146

westonruter:
CLA bot is complaining about `felixarntz@users.noreply.github.com`. So we'll just ignore that until merging, at which time we'll re-flip the label.

includes/class-amp-service-worker.php
felixarntz: Makes sense.

I think applying a filter here would be a good start. I'm not sure we should add more complexity to the theme support arguments. For example, these precached scripts should be provided somewhere under the service_worker key of AMP theme support, but this would cause for an even deeper nested array.

Having a filter allows advanced users to adjust this, and then in the future we can figure out how to make this more straightforward, or whether we can possibly auto-detect it somehow.

includes/class-amp-service-worker.php
felixarntz: Resolved per our face to face discussion.

westonruter: See 05172f8d

includes/class-amp-service-worker.php
```diff
@@ -0,0 +1,316 @@
...
+		add_action( 'parse_request', array( __CLASS__, 'handle_service_worker_iframe_install' ) );
+		add_action( 'wp', array( __CLASS__, 'add_install_hooks' ) );
+
+		if ( true === $theme_support['service_worker'] || ! empty( $theme_support['service_worker']['cdn_script_caching'] ) ) {
+			add_action( 'wp_front_service_worker', array( __CLASS__, 'add_cdn_script_caching' ) );
+		}
+		if ( true === $theme_support['service_worker'] || ! empty( $theme_support['service_worker']['image_caching'] ) ) {
+			add_action( 'wp_front_service_worker', array( __CLASS__, 'add_image_caching' ) );
+		}
+		if ( true === $theme_support['service_worker'] || ! empty( $theme_support['service_worker']['google_fonts_caching'] ) ) {
```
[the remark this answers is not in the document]
westonruter: See 05172f8d

includes/class-amp-service-worker.php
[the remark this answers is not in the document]
felixarntz: Ah, good catch. I think in this case we should omit the `_doing_it_wrong()` though and just return. This method is not manually called by anybody, so neither the user nor the developer is really "at fault" for this happening. It will be the case if someone uses this version of the AMP plugin with version 0.1 of the PWA plugin.

includes/class-amp-service-worker.php
westonruter: How about deferring this decision for later? Themes can use the `wp_front_service_worker` hook to accomplish this for now. We can then determine later what level of abstraction is right based on usage.

includes/class-amp-service-worker.php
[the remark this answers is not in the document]
felixarntz: Same as above. I'd argue we should remove the `_doing_it_wrong()` because nobody is really doing anything wrong, they're just on an old version.

Because of that, I'd say we wait a little further with a type hint, just to decrease the chances of somebody using PWA plugin 0.1.

includes/class-amp-service-worker.php
```diff
@@ -0,0 +1,316 @@
...
+	/**
+	 * Add runtime caching of Google Fonts with stale-while-revalidate strategy for stylesheets and cache-first strategy for webfont files.
+	 *
+	 * @link https://developers.google.com/web/tools/workbox/guides/common-recipes#google_fonts
+	 * @link https://github.com/ampproject/amp-by-example/blob/4593af61609898043302a101826ddafe7206bfd9/boilerplate-generator/templates/files/serviceworkerJs.js#L76-L103
+	 * @link https://github.com/xwp/pwa-wp/blob/master/integrations/class-wp-service-worker-fonts-integration.php
+	 *
+	 * @param WP_Service_Worker_Scripts $service_workers Service workers.
+	 */
+	public static function add_google_fonts_caching( $service_workers ) {
```
felixarntz: This needs to have the check for the correct class as well, like the above (without `_doing_it_wrong()`).

includes/class-amp-service-worker.php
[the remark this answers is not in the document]
felixarntz: As discussed in our face to face conversation, let's rely on theme support and provide the default values in an array. AMP CDN script caching should be opt-out, the other two should be opt-in. This matches the suggestions from AMP by Example.

includes/class-amp-service-worker.php
westonruter: Not like this?

```diff
diff --git a/includes/class-amp-service-worker.php b/includes/class-amp-service-worker.php
index 37dbc0e1..4de47a37 100644
--- a/includes/class-amp-service-worker.php
+++ b/includes/class-amp-service-worker.php
@@ -157,6 +157,10 @@ class AMP_Service_Worker {
 	 * @param WP_Service_Worker_Scripts $service_workers Service workers.
 	 */
 	public static function add_google_fonts_caching( $service_workers ) {
+		if ( ! ( $service_workers instanceof WP_Service_Worker_Scripts ) ) {
+			_doing_it_wrong( __METHOD__, esc_html__( 'Expected argument to be WP_Service_Worker_Scripts.', 'amp' ), '1.1' );
+			return;
+		}
 
 		// The PWA plugin also automatically adds runtime caching for Google Fonts when WP_SERVICE_WORKER_INTEGRATIONS_ENABLED is set.
 		if ( class_exists( 'WP_Service_Worker_Fonts_Integration' ) ) {
```

felixarntz: See the other comments above.

includes/class-amp-service-worker.php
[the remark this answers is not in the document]
westonruter: @felixarntz then what about just changing the message to say: "Please upgrade to PWA plugin v0.2".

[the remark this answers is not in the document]
westonruter: Like so:

```diff
diff --git a/includes/class-amp-service-worker.php b/includes/class-amp-service-worker.php
index 37dbc0e1..73a2280e 100644
--- a/includes/class-amp-service-worker.php
+++ b/includes/class-amp-service-worker.php
@@ -83,7 +83,7 @@ class AMP_Service_Worker {
 	 */
 	public static function add_cdn_script_caching( $service_workers ) {
 		if ( ! ( $service_workers instanceof WP_Service_Worker_Scripts ) ) {
-			_doing_it_wrong( __METHOD__, esc_html__( 'Expected argument to be WP_Service_Worker_Cache_Registry.', 'amp' ), '1.1' );
+			_doing_it_wrong( __METHOD__, esc_html__( 'Please update to PWA v0.2. Expected argument to be WP_Service_Worker_Cache_Registry.', 'amp' ), '1.1' );
 			return;
 		}
 
@@ -125,7 +125,7 @@ class AMP_Service_Worker {
 	 */
 	public static function add_image_caching( $service_workers ) {
 		if ( ! ( $service_workers instanceof WP_Service_Worker_Scripts ) ) {
-			_doing_it_wrong( __METHOD__, esc_html__( 'Expected argument to be WP_Service_Worker_Scripts.', 'amp' ), '1.1' );
+			_doing_it_wrong( __METHOD__, esc_html__( 'Please update to PWA v0.2. Expected argument to be WP_Service_Worker_Scripts.', 'amp' ), '1.1' );
 			return;
 		}
 
@@ -157,6 +157,10 @@ class AMP_Service_Worker {
 	 * @param WP_Service_Worker_Scripts $service_workers Service workers.
 	 */
 	public static function add_google_fonts_caching( $service_workers ) {
+		if ( ! ( $service_workers instanceof WP_Service_Worker_Scripts ) ) {
+			_doing_it_wrong( __METHOD__, esc_html__( 'Please update to PWA v0.2. Expected argument to be WP_Service_Worker_Scripts.', 'amp' ), '1.1' );
+			return;
+		}
 
 		// The PWA plugin also automatically adds runtime caching for Google Fonts when WP_SERVICE_WORKER_INTEGRATIONS_ENABLED is set.
 		if ( class_exists( 'WP_Service_Worker_Fonts_Integration' ) ) {
```

It think it's important to keep _some_ kind of notice, as otherwise it just silently fails.

PR #1271, opened 2018-07-18
Add an admin pointer for version 1.0

### LINKED ISSUE #1254: Show admin pointer for AMP admin screen when updating to 1.0
Since 1.0 introduces an admin UI for managing the theme support mode, including allowing the user to opt-in to paired and native modes without having to make any code changes, we should add an admin pointer to invite users to try out the new modes:

image

This could be paired with the ability to do a scan of their site to check for compatibility (see #1007).

kienstra:
**Request For Review**

Hi @hellofromtonya,
Thanks for your PR #1274. I'm going to review it next.

It would be great if you could review this. Then, maybe @westonruter could give it another look and merge when it's ready.

* The admin pointer should appear on checking out this branch and loading `/wp-admin`.
* Mainly follows Weston's mockup in Issue 1254.
* Enqueues the script only if this pointer hasn't been dismissed.
* Based on [this tutorial](https://code.tutsplus.com/articles/integrating-with-wordpress-ui-admin-pointers--wp-26853)
* Copies some of the structure of [amp-block-validation.js](https://github.com/Automattic/amp-wp/blob/024e450798c35a89c3390c410ab42522a924f712/assets/js/amp-block-validation.js)

Closes #1254

amp-admin-pointer-image

### DIFF
--- a/includes/admin/class-amp-admin-pointer.php
+++ b/includes/admin/class-amp-admin-pointer.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * Admin pointer class.
+ *
+ * @package AMP
+ * @since 1.0
+ */
+
+/**
+ * AMP_Admin_Pointer class.
+ *
+ * Outputs an admin pointer to show the new features of v1.0.
+ * Based on https://code.tutsplus.com/articles/integrating-with-wordpress-ui-admin-pointers--wp-26853
+ *
+ * @since 1.0
+ */
+class AMP_Admin_Pointer {
+
+	/**
+	 * The ID of the template mode admin pointer.
+	 *
+	 * @var string
+	 */
+	const TEMPLATE_POINTER_ID = 'amp_template_mode_pointer_10';
+
+	/**
+	 * The slug of the script.
+	 *
+	 * @var string
+	 */
+	const SCRIPT_SLUG = 'amp-admin-pointer';
+
+	/**
+	 * Initializes the class.
+	 *
+	 * @since 1.0
+	 */
+	public function init() {
+		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_pointer' ) );
+	}
+
+	/**
+	 * Enqueues the pointer assets.
+	 *
+	 * If the pointer has not been dismissed, enqueues the style and script.
+	 * And outputs the pointer data for the script.
+	 *
+	 * @since 1.0
+	 */
+	public function enqueue_pointer() {
+		$dismissed = explode( ',', strval( get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) ) );
+
+		// Exit if the pointer has been dismissed.
+		if ( in_array( self::TEMPLATE_POINTER_ID, $dismissed, true ) ) {
+			return;
+		}
+
+		wp_enqueue_style( 'wp-pointer' );
+
+		wp_enqueue_script(
+			self::SCRIPT_SLUG,
+			amp_get_asset_url( 'js/' . self::SCRIPT_SLUG . '.js' ),
+			array( 'jquery', 'wp-pointer' ),
+			AMP__VERSION,
+			true
+		);
+
+		wp_add_inline_script(
+			self::SCRIPT_SLUG,
+			sprintf( 'ampAdminPointer.load( %s );', wp_json_encode( $this->get_pointer_data() ) )
+		);
+	}
+
+	/**
+	 * Gets the pointer data to pass to the script.
+	 *
+	 * @since 1.0
+	 * @return array
+	 */
+	public function get_pointer_data() {
+		return array(
+			'pointer' => array(
+				'pointer_id' => self::TEMPLATE_POINTER_ID,
+				'target'     => '#toplevel_page_amp-options',
+				'options'    => array(
+					'content'  => sprintf(
+						'<h3>%s</h3><p><strong>%s</strong></p><p>%s</p>',
+						__( 'AMP', 'amp' ),
+						__( 'New AMP Template Modes', 'amp' ),
+						__( 'You can now reuse your theme\'s styles in AMP responses, using "Paired" or "Native" mode.', 'amp' )
+					),
+					'position' => array(
+						'edge'  => 'left',
+						'align' => 'middle',
+					),
+				),
+			),
+		);
+	}
+}
--- a/includes/admin/class-amp-admin-pointer.php
+++ b/includes/admin/class-amp-admin-pointer.php
@@ -48,10 +48,7 @@ class AMP_Admin_Pointer {
 	 * @since 1.0
 	 */
 	public function enqueue_pointer() {
-		$dismissed = explode( ',', strval( get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) ) );
-
-		// Exit if the pointer has been dismissed.
-		if ( in_array( self::TEMPLATE_POINTER_ID, $dismissed, true ) ) {
+		if ( $this->is_pointer_dismissed() ) {
 			return;
 		}
 
@@ -69,6 +66,22 @@ class AMP_Admin_Pointer {
 			self::SCRIPT_SLUG,
 			sprintf( 'ampAdminPointer.load( %s );', wp_json_encode( $this->get_pointer_data() ) )
 		);
+	}
+
+	/**
+	 * Whether the AMP admin pointer has been dismissed.
+	 *
+	 * @since 1.0
+	 * @return boolean
+	 */
+	protected function is_pointer_dismissed() {
+		$dismissed = get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true );
+		if ( empty( $dismissed ) ) {
+			return false;
+		}
+		$dismissed = explode( ',', strval( $dismissed ) );
+
+		return in_array( self::TEMPLATE_POINTER_ID, $dismissed, true );
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


includes/admin/class-amp-admin-pointer.php
```diff
@@ -0,0 +1,100 @@
...
+ * @since 1.0
+ */
+class AMP_Admin_Pointer {
+
+	/**
+	 * The ID of the template mode admin pointer.
+	 *
+	 * @var string
+	 */
+	const TEMPLATE_POINTER_ID = 'amp_template_mode_pointer_10';
```
kienstra: When you're testing this, you might have to change this locally. Once you dismiss the pointer, it won't display again. Unless you manually update the user meta value.

includes/admin/functions.php
```diff
@@ -166,3 +166,13 @@ function amp_editor_core_blocks() {
...
+
+/**
+ * Bootstrap the AMP admin pointer class.
+ *
+ * @since 1.0
+ */
+function amp_admin_pointer() {
+	$admin_pointer = new AMP_Admin_Pointer();
+	$admin_pointer->init();
+}
```
kienstra: This bootstrapping is repetitive. It'd be nice if `amp_post_meta_box()`, `amp_editor_core_blocks()`, and `amp_admin_pointer()` could be combined into one function, like `amp_admin_bootstrap()`. 

But in the unlikely case that a user [unhooked](https://github.com/Automattic/amp-wp/pull/1271/files#diff-4805c837d3390517773926fb26726bcdL155) one of these from `wp_loaded`, this would break that.

tests/test-class-amp-admin-pointer.php
```diff
@@ -0,0 +1,93 @@
...
+	 * @covers AMP_Admin_Pointer::get_pointer_data()
+	 */
+	public function test_get_pointer_data() {
+		$pointer_data = $this->instance->get_pointer_data();
+		$pointer      = $pointer_data['pointer'];
+		$this->assertContains( '<h3>AMP</h3><p><strong>New AMP Template Modes</strong></p>', $pointer['options']['content'] );
+		$this->assertEquals(
+			array(
+				'align' => 'middle',
+				'edge'  => 'left',
```
kienstra: This might need improvement. I think `'edge' => 'left'` is probably right, but the alignment might still be improved.

includes/admin/class-amp-admin-pointer.php
```diff
@@ -0,0 +1,100 @@
...
+				'options'    => array(
+					'content'  => sprintf(
+						'<h3>%s</h3><p><strong>%s</strong></p><p>%s</p>',
+						__( 'AMP', 'amp' ),
+						__( 'New AMP Template Modes', 'amp' ),
+						__( 'You can now reuse your theme\'s styles in AMP responses, using "Paired" or "Native" mode.', 'amp' )
+					),
+					'position' => array(
+						'edge'  => 'left',
+						'align' => 'middle',
```
kienstra: This might need improvement. I think `'edge' => 'left'` is probably correct, but the alignment might still be improved.

includes/admin/class-amp-admin-pointer.php
```diff
@@ -0,0 +1,100 @@
...
+	/**
+	 * Enqueues the pointer assets.
+	 *
+	 * If the pointer has not been dismissed, enqueues the style and script.
+	 * And outputs the pointer data for the script.
+	 *
+	 * @since 1.0
+	 */
+	public function enqueue_pointer() {
+		$dismissed = explode( ',', strval( get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) ) );
```
hellofromtonya: To optimize this code, I've move the `get_user_meta()` out of it.  Then check if there are any dismissed pointers to consider. If yes, explode then. 

To keep the method focused on the enqueueing tasks, I'd abstract it into a new method:

```
	protected function has_pointer_been_dismissed() {
		$dismissed = get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true );
		if ( empty( $dismissed ) ) {
			return false;
		}
		$dismissed = explode( ',', strval( $dismissed ) );

		return in_array( self::TEMPLATE_POINTER_ID, $dismissed, true );
	}
```

includes/admin/class-amp-admin-pointer.php
```diff
@@ -0,0 +1,100 @@
...
+
+		// Exit if the pointer has been dismissed.
+		if ( in_array( self::TEMPLATE_POINTER_ID, $dismissed, true ) ) {
```
hellofromtonya: Here, this can now be simplified to this:

```
		if ( $this->has_pointer_been_dismissed() ) {
			return;
		}
```

These changes make it to read and quickly know what's going on for the guard clause.

includes/admin/functions.php
hellofromtonya: Then the questions are:  

- Do we need to provide the ability to unhook any of those components? 
- Do we need to provide the ability to unhook any of these components?

```
	add_action( 'wp_loaded', 'amp_editor_core_blocks' );
	add_action( 'wp_loaded', 'amp_post_meta_box' );
	add_action( 'wp_loaded', 'amp_editor_core_blocks' );
	add_action( 'wp_loaded', 'amp_add_options_menu' );
	add_action( 'wp_loaded', 'amp_admin_pointer' );
```

Changing that to `add_action( 'wp_loaded', 'amp_admin_bootstrap' );` and then letting `amp_admin_bootstrap()` manage the order and initializations results in less callbacks having to be processed, i.e. only 1 instead of 5.  It'll be slightly faster.

kienstra: Good questions. I think it'd be best to avoid renaming the functions above, as a user could have unhooked them. 

But it's not a strongly-held belief, and there probably aren't many users who would have unhooked them.

Another option is renaming `amp_admin_pointer()` to `amp_admin_bootstrap()`, without making any other change.

Then, any future admin classes could add their bootstrapping there.

hellofromtonya: >Another option is renaming amp_admin_pointer() to amp_admin_bootstrap(), without making any other change.

This option is future-thinking as you are planning for future bootstrapping functionality.  That's good.

Now, one could argue that confusion may be introduced when a dev is deciding where to bootstrap.  Adding an inline comment above the action hook may elevate that confusion.

hellofromtonya: I'd vote to review how we want to bootstrap and consider what callbacks we can consolidate.  That said, I think it's outside the scope of this particular PR.  We may want to open a ticket for it and begin the discussion process beyond ours here.

kienstra:
Hi @hellofromtonya,
Thanks for your great suggestions here. Using `show` was a [great idea](https://github.com/Automattic/amp-wp/pull/1271#discussion_r203884628).

There are still a few open discussions here. I'm happy to add `has_pointer_been_dismissed()` if you'd like.

includes/admin/class-amp-admin-pointer.php
hellofromtonya: Correct. If the user has never dismissed a pointer, then the user meta does not exist or is empty.  In that case, it optimizes the code in that it does not to run `strval()`,  `explode()`, or `in_array()`.

hellofromtonya: I also find it expresses the intent more clearly to help us more quickly read and understand what's going on.

hellofromtonya: @kienstra You could name that guard clause method something less passive such as 

`is_pointer_dismissed()`

includes/admin/functions.php
kienstra: Sounds good, Tonya.

hellofromtonya:
>I'm happy to add has_pointer_been_dismissed() if you'd like.

Your code works well.  Great job!

But I'd vote for abstracting the dismissed guard clause for the following reasons:

1. It clear communicates intent of what that code is doing.
2. It's more performant _when_ the user has not dismissed pointers.
3. It's more readable and quickly understandable.

Plus, I'd change the method name to `is_pointer_dismissed()` vs. the passive name I originally gave.

I'll leave it up to you whether you wish to change it or not.  I'm approving the PR as your code does work.

includes/admin/class-amp-admin-pointer.php
kienstra: Thanks, commit [e60f0](https://github.com/Automattic/amp-wp/pull/1271/commits/e60f0f67fced870d11e9e87690198cc32ad39db8) adds `is_pointer_dismissed()`.

kienstra:
**Thanks, Applied Suggestion**

Hi @hellofromtonya,
Thanks for reviewing this, and adding so much thought into your comments.

Commit [e60f0](https://github.com/Automattic/amp-wp/pull/1271/commits/e60f0f67fced870d11e9e87690198cc32ad39db8) adds `is_pointer_dismissed()`.

includes/admin/class-amp-admin-pointer.php
kienstra: Thanks, this is applied in commit [e60f0](https://github.com/Automattic/amp-wp/pull/1271/commits/e60f0f67fced870d11e9e87690198cc32ad39db8#diff-7ec3f55e1310503aa82ce61984d3f482R51).

includes/admin/functions.php
westonruter: Yes, I think we should try to get away from bootstrapping like this because the class instantiation variable that gets created is not accessible anywhere. We should instead of an AMP plugin manager that is responsible for instantiating classes and then keeping around the references to the class instances for the plugin (and others) to access later.

PR #1275, opened 2018-07-21
Add Gutenberg 'Enable AMP' toggle

### LINKED ISSUE #1230: Add AMP enabled/disabled toggle to Gutenberg editor

kienstra:
This adds the 'Enable AMP' toggle to the Gutenberg editor.

- [x] This should have the initial enabled value, probably from `AMP_Post_Meta_Box`.
- [x]  It could also have the notifications that the classic editor has, if needed.

Closes #1230

enable-amp


westonruter:
> It could also have the notifications that the classic editor has, if needed.

Meaning the notices shown when expanding the toggle to edit? Yes, these should be shown as well. In Gutenberg perhaps these notices would be shown unconditionally since there wouldn't be any toggle to expand. So this will require exporting `$errors` from PHP to JS as found here:

https://github.com/Automattic/amp-wp/blob/e3d036c2db60b215636a2b665414b0e599e9ad97/includes/admin/class-amp-post-meta-box.php#L187-L203

kienstra:
**Notices**

Hi @westonruter,
>Meaning the notices shown when expanding the toggle to edit? 
Yes, that's what I had in mind. Good point that they would have to be shown without the toggle to expand.

assets/js/amp-block-editor-toggle.js
```diff
@@ -0,0 +1,121 @@
+/**
```
westonruter: Ideally this file would be converted into a module similar to all of the scripts for blocks in https://github.com/Automattic/amp-wp/tree/develop/blocks

This needs to be done for https://github.com/Automattic/amp-wp/blob/develop/assets/js/amp-block-validation.js and https://github.com/Automattic/amp-wp/blob/develop/assets/js/amp-editor-blocks.js as well. This would allow for being able to use Babel to handle i18n and it would allow for the scripts to all be concatenated into a single build script.

kienstra:
**Hiding Toggle, Displaying Notice**

The next step is to apply something like the Classic editor UI. Where if AMP is disabled [via errors](https://github.com/Automattic/amp-wp/pull/1275/commits/25156f4027b6d39a62578a76360a15ef0c09e15e#diff-87f0ef41a80eed0bfff797f932886230R197), you can't select "Enabled."

errors-amp-disabled

webpack.config.js
```diff
@@ -4,7 +4,8 @@ const path = require( 'path' );
 
 module.exports = {
 	entry: {
-		'./assets/js/amp-blocks-compiled': './blocks/index.js'
+		'./assets/js/amp-blocks-compiled': './blocks/index.js',
+		'./assets/js/amp-block-editor-toggle-compiled': './assets/src/amp-block-editor-toggle.js'
```
westonruter: I think these two two should be combined into a single entry point. In other words, `./blocks/index.js` could just import the module for the block editor toggle, and then there wouldn't be a need for `amp-block-editor-toggle-compiled.js`

kienstra: Good idea, I'll apply that.

kienstra:
**Current Notice Display**

Here's an example of how the Notice looks so far when AMP can't be enabled:

amp-notice

westonruter:
That looks good 👍

webpack.config.js
kienstra: Could this instead combine `amp-block-editor-toggle-compiled.js` with any future modules, like [amp-block-validation.js](https://github.com/Automattic/amp-wp/blob/e3d036c2db60b215636a2b665414b0e599e9ad97/assets/js/amp-block-validation.js) (when that's converted to a module)?

It looks like `assets/js/amp-blocks-compiled.js` is [only enqueued if amp_is_canonical()](https://github.com/Automattic/amp-wp/blob/e3d036c2db60b215636a2b665414b0e599e9ad97/includes/admin/class-amp-editor-blocks.php#L127).

westonruter: @kienstra Yes, `amp-block-validation.js` should also be converted to a module and included in this bundle. 100%

Good point on `amp-blocks-compiled.js`. Maybe that would make sense to be a separate bundle then. Either that, or there could be an arg passed for whether the blocks should be registered. But if not, then why include the file in the first place?

westonruter: So `amp-block-validation.js` and `amp-block-editor-toggle-compiled.js` should definitely be combined.

kienstra: Thanks, combining `amp-block-validation.js` and `amp-block-editor-toggle-compiled.js` sounds good. 

If it's alright, I'll open a new issue to convert `amp-block-validation.js` to a module and compile it in the same file.

kienstra:
The Notice in the block editor now has the same links as the Classic Editor:

links-like-classic-editor


And here's the display in the unlikely case that there are 2 messages in a Notice:
two-messages-notice

webpack.config.js
westonruter: Sounds good. 👍

westonruter: I know that @hellofromtonya is currently working on `amp-block-validation.js` so it would be good to not step on her toes 😄

westonruter: @kienstra One more thing for that issue you'll create: please add to it the need to generate source maps.

hellofromtonya: @westonruter, @kienstra and I synced up in slack.  He'll open the issue.  Once we resolve the issues in `amp-block-validation.js`, then we can green light the task.

assets/js/amp-block-editor-toggle.js
kienstra: Sure, Issue #1298 now tracks bundling this `amp-block-editor-toggle.js` module with `amp-block-validation.js`. We could also add converting `amp-editor-blocks.js` to a module.

webpack.config.js
kienstra: Issue #1298 now tracks bundling [amp-block-editor-toggle.js](https://github.com/Automattic/amp-wp/blob/0735ed02ef01f85782d38e2a7de56e42b085c8e9/assets/src/amp-block-editor-toggle.js#L1) with [amp-block-validation.js](https://github.com/Automattic/amp-wp/blob/21ebbc69289ec46a6752df39d2170a625c70faf7/assets/js/amp-block-validation.js). It also mentions that it's probably best to wait for #1293.

templates/admin/amp-status.php
```diff
@@ -39,36 +39,7 @@
 			</fieldset>
 		<?php else : ?>
 			<div class="inline notice notice-warning notice-alt">
-				<p>
-					<?php
-					$error_messages = array();
```
kienstra: This section is moved to [get_raw_error_messages()](https://github.com/Automattic/amp-wp/pull/1275/files#diff-87f0ef41a80eed0bfff797f932886230R301).

kienstra:
**Request For Code Review**

Hi @westonruter,
Could you please review this PR? 

Notice with a link:
notice-with-link

Notice without a link:
notice-without-a-link

You know this, but for this PR's record, please run this to compile the script:
`npm run dev`

includes/admin/class-amp-post-meta-box.php
```diff
@@ -202,13 +279,99 @@ public function render_status( $post ) {
...
+	 * AMP can be disabled for several reasons,
+	 * like if the user disables it in Supported Templates > Content Types.
+	 * Then, it would not make sense to allow the user to enable AMP in the post editor,
+	 * as it's already disabled.
+	 * So this gets the message(s) that should display.
+	 *
+	 * @param array $raw_error_messages The raw error messages, possibly with URLs.
+	 * @return array $error_messages The error message(s), as an array of strings.
+	 */
+	public function get_classic_editor_error_messages( $raw_error_messages ) {
```
westonruter: There is duplication between PHP here and the JS in `PluginPostStatusInfo`. I appreciate the concern to not `dangerouslySetInnerHTML` but since the error messages are coming from a trusted source, I think it would be better to do so in this case. Even in Gutenberg, for example, there is a [`RawHTML`](https://github.com/WordPress/gutenberg/blob/master/packages/element/src/raw-html.js) component for this purpose. This component is publicly available already in Gutenberg via `wp.element.RawHTML`.

kienstra: Thanks, I'm working on this now.

kienstra: [These commits](https://github.com/Automattic/amp-wp/compare/9f0e248b5863563171294a8ce52cfb3c67f0c27e...230d87a42a2b0f73074736b0db11142a68ad6399) simplify the error messages to use `RawHTML` like you suggested.

kienstra:
Thanks, @westonruter!

PR #1278, opened 2018-07-23
Add filter for changing the amphtml url

rebeccahum:
It would be nice to have a filter where we can change the `amphtml` URL rendered for formatting purposes.

Ref: #80636-z

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -230,7 +230,7 @@ function amp_add_amphtml_link() {
 	}
 
 	if ( $amp_url ) {
-		printf( '<link rel="amphtml" href="%s">', esc_url( $amp_url ) );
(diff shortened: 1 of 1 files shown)


rebeccahum:
@westonruter Hi!  So, we have a client that'd like to strip the trailing slash off the of the URL.  For `amp_get_permalink`, the filter only runs if the `is_singular()` condition is met (where it isn't in their use case): https://github.com/Automattic/amp-wp/blob/364e03b32d8cbf5052dbfa63cb33cd6082082ef5/includes/amp-helper-functions.php#L199

westonruter:
@rebeccahum so you're referring to the URLs with the `amp` query var, such as `https://example.com/about/?amp`? You want to to be linked to as `https://example.com/about?amp`?

How does this matter in practice?

PR #1281, opened 2018-07-24
[WIP] Fix wide image to contain in content

### LINKED ISSUE #1086: Revisit aligned images with inline width larger than max getting displayed off-center
**As a user, I expect my images inside a post to be aligned correctly, even if they are larger than the `max-width` of the given viewport.**
- [x] AC1: Identify when and why the viewport breaks.
- [x] AC2: A user who has a wider-than-viewport image expects their image to display in margin; using the plugin's sanitizer, we expect this to be fixed automatically. 

---

The conversion of `width` to `max-width` was originally added in #495. This was removed recently in #1064 as it was not determined to be the best solution to the problem reported in #494 (“Aligned images with inline width larger than max get displayed off-center”)? In #610 it was suggested to be removed:

> Notably, the current AMP spec does not […] specify that width should be replaced by max-width, thus these rules are no longer enforced.

It feels like it fixes a symptom but isn't the right solution to the underlying problem. 

In https://github.com/Automattic/amp-wp/pull/1064#issuecomment-382826877 @mehigh comments:

> Instead of replacing the width by max-width, a solution used almost everywhere on the internets is to introduce a max-width: 100%; on the images. That way they would 'stay put' regardless on whether the width attribute is larger than the container or not. But there are edge cases, and nuances, as with any solutions.

So we need to (1) replicate the original issue and (2) determine what the best fix is.

hellofromtonya:
This PR solves the problem of wide images overflowing the content width.  It adds an inline style of `width: auto` to the `<figure>` element when it has the class attribute `'wp-block-image'` but not `is-resized`. 

Why?  

Gutenberg adds the class attributes and sets the CSS rule to:

```css
.wp-block-image {
    width: -webkit-fit-content;
    width: -moz-fit-content;
    width: fit-content;
}
```

The width value of `fit-content` is causing wider images to overflow.  See the details in issue #1086 more information as well as screenshots.  You can view the example posts here: [Gutenberg post](https://paired-ampconfdemo.pantheonsite.io/2018/07/24/wide-image-alignments-with-gutenberg-image-blocks/?amp) and [Classic Editor post](https://paired-ampconfdemo.pantheonsite.io/2018/07/24/wide-image-alignments-with-classic-editor/?amp).

By adding the inline CSS rule, wider images rescale and fit within the content.

It's worth noting that the issue does not occur with the Classic Editor.

Closes #1086.

### DIFF
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -229,7 +229,7 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 	/**
 	 * Make final modifications to DOMNode
 	 *
-	 * @param DOMNode $node The DOMNode to adjust and replace.
+	 * @param DOMElement $node The DOMNode to adjust and replace.
 	 */
 	private function adjust_and_replace_node( $node ) {
 
@@ -269,7 +269,7 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 	 */
 	private function maybe_add_lightbox_attributes( $attributes, $node ) {
 		$parent_node = $node->parentNode;
-		if ( 'figure' !== $parent_node->tagName ) {
+		if ( ! ( $parent_node instanceof DOMElement ) || 'figure' !== $parent_node->tagName ) {
 			return $attributes;
 		}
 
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -339,4 +340,36 @@ public function handle_centering( $node ) {
...
 		return $figure;
 	}
+
+	/**
+	 * Add an inline style to set the `<figure>` element's width to `auto` instead of `fit-content`.
+	 *
+	 * @since 1.0
+	 * @see https://github.com/Automattic/amp-wp/issues/1086
+	 *
+	 * @param DOMNode $node The DOMNode to adjust and replace.
```
westonruter: I think we should be able to safely replace `DOMNode` with `DOMElement` here. We already know it is an element because we just called it via `AMP_DOM_Utils::create_node()`.

hellofromtonya:
We will go ahead and merge this PR as it fixes the overflowing image issue.  But we'll keep issue #1086 open to track for regression once the AMPHTML team resolves the open issue.

PR #1287, opened 2018-07-25
Report removing unrecognized elements.

### LINKED ISSUE #1100: Unrecognized elements are silently removed without validation error reporting
When coming across an element that is not recognized at all, the following code in `process_node` will be invoked:

https://github.com/Automattic/amp-wp/blob/1e2cd22f1bfe8ab9d833c901e02111120f52fe72/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L284-L292

In the call to `replace_node_with_children` it will call `remove_node` instead of `remove_invalid_child`, and only the latter will report the removal. This needs to be improved to report invalidity when it is being removed for that reason.

hellofromtonya:
Fixes the problem of silently removing unrecognized elements.  Uses the same code pattern and methodology as `remove_invalid_child`:

- Prevent double-reporting nodes that are rejected.
- Check through should_sanitize_validation_error().
- Replace if true.
- Else, store in the `should_not_replace_nodes` property.

Closes #1100.

[See this comment](https://github.com/Automattic/amp-wp/issues/1100#issuecomment-407879611) where the issue was recreated.

### DIFF
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -86,6 +86,13 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 	protected $script_components = array();
 
 	/**
+	 * Keep track of nodes that should not be replaced to prevent duplicated validation errors since sanitization is rejected.
+	 *
+	 * @var array
+	 */
+	protected $should_not_replace_nodes = array();
+
+	/**
 	 * AMP_Tag_And_Attribute_Sanitizer constructor.
 	 *
 	 * @since 0.5
@@ -412,7 +419,7 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 					$attr_spec_list = $first_spec[ AMP_Rule_Spec::ATTR_SPEC_LIST ];
 				}
 			}
-		} // End if().
+		}
 
 		if ( ! empty( $attr_spec_list ) && $this->is_missing_mandatory_attribute( $attr_spec_list, $node ) ) {
 			$this->remove_node( $node );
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1656,34 +1663,46 @@ private function get_ancestor_with_matching_spec_name( $node, $ancestor_spec_nam
...
 	 * @since 0.3.3
+	 * @since 1.0 Fix silently removing unrecognized elements.
+	 * @see https://github.com/Automattic/amp-wp/issues/1100
 	 *
 	 * @param DOMNode $node Node.
 	 */
 	private function replace_node_with_children( $node ) {
 
+		// If node has no children or no parent, just remove the node.
 		if ( ! $node->hasChildNodes() || ! $node->parentNode ) {
```
westonruter: Should not there be a `$this->should_sanitize_validation_error( array(), compact( 'node' ) )` call in here as well? Otherwise the removal won't be reported as a validation error.

hellofromtonya: @westonruter the `if` code block flows through to `remove_node()` where it’s handled.  I did confirm it there.  When I write the error unit tests, I’ll validate that too to ensure we have reporting for both paths.

hellofromtonya:
@westonruter I've added tests to ensure invalid elements with and without child nodes do get the right validation errors. Check out [commit 65c9cde](https://github.com/Automattic/amp-wp/pull/1287/commits/65c9cde0ed176d9f2a99255bcecf75c9b91a023a).

westonruter:
Good work! Great to have this as part of 1.0. This will help a lot when someone tries to use an AMP component that is misspelled or tries to use a new AMP component that is not yet recognized due to the spec not being updated.

PR #1291, opened 2018-07-30
Change the order of template modes in admin interface

### LINKED ISSUE #1264: Change the order of template modes in admin interface
**As a user, I should see "Native" as the first toggleable option for AMPhtml so that I am most likely to choose this option.**

- [ ] AC1: Instead of the ordering list of "Classic, Paired, Native" I should see "Native, Paired, Classic" as the three options.

---

Currently the operational mode of the plugin can be configured in the admin via toggle buttons. The current ordering of the buttons reflects the chronological evolution of the modes: `Classic`, `Paired`, `Native`. Proposal: Change the order to have `Native` first to implicitly encourage that as the best option.

screen shot 2018-07-10 at 2 24 23 pm

AdelDima:
capture

Fixes #1264.

### DIFF
--- a/includes/options/class-amp-options-menu.php
+++ b/includes/options/class-amp-options-menu.php
@@ -177,7 +177,7 @@ class AMP_Options_Menu {
 					</dt>
 					<dd>
 						<?php esc_html_e( 'Display AMP responses in classic (legacy) post templates in a basic design that does not match your theme\'s templates.', 'amp' ); ?>
-					</dd>					
+					</dd>
 				</dl>
 			</fieldset>
 		<?php endif; ?>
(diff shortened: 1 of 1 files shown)


includes/options/class-amp-options-menu.php
```diff
@@ -150,15 +150,15 @@ public function render_theme_support() {
 						<p><?php esc_html_e( 'Your active theme is known to work well in paired or native mode.', 'amp' ); ?></p>
 					</div>
 				<?php endif; ?>
-				<dl>
+				<dl>
```
westonruter: Looks like there is trailing whitespace introduced on some lines which should get trimmed.

AdelDima: Sorry, this was my first contribution, I fixed it now.

hellofromtonya: Welcome and thank you for your contribution, @AdelDima.

PR #1293, opened 2018-07-30
Fix displaying AMP validation error at block level in the Gutenberg editor

### LINKED ISSUE #1292: AMP validation error is not displaying at the block level in the Gutenberg editor.
**As a user, I expect AMP validation notices to appear on each block that has an AMP issue.**

- [x] AC1: Identify when and why the block level notices do not appear.
- [x] AC2: Ensure block level validation notices appear on with the block after saving and when the editor is in a clean state.

## What we want

An AMP validation error should display at the block level within the Gutenberg editor to provide the user with more information at a granular level.  

This is the UI notice we want:

amp-issue-block-level-notification

## Noted problems 

There are multiple contributing factors for this issue.  Here is a list:

- [x] Gutenberg update changed filter hook name from `blocks.BlockEdit` to `editor.BlockEdit`.
- [x] Gutenberg update changed the block `uid` property from `id` to `clientId`.
- [x] Block uids go out-of-sync.  Witnessed on page/editor load and intermittently.
- [x]  Duplicated blocks cause all block level notices to disappear.
- [x] The warning notice remains after removing all affected blocks.
- [x] Both the top level and block level notices remain when turning off AMP on a post, i.e. switching the AMP toggle to off.

hellofromtonya:
This PR fixes the problems noted in #1292:

- [x] Changes the filter hook name to `editor.BlockEdit`.  This filter was changed in [Gutenberg 3.3](https://github.com/WordPress/gutenberg/blob/41fd116ca1f73a9b5c962c66ea0c4aae338ea35e/docs/reference/deprecated.md#330) as noted here:

>blocks.BlockEdit filter removed. Please use editor.BlockEdit instead.

- [x] Changes the block `uid` to `ownProps.clientId`.  The `id` property was changed in [Gutenberg 3.5.0](https://github.com/WordPress/gutenberg/blob/41fd116ca1f73a9b5c962c66ea0c4aae338ea35e/docs/reference/deprecated.md#350) to `.clientId` as noted here:

>All references to a block's `uid` have been replaced with equivalent props and selectors for `clientId`.

These two changes result in displaying the AMP validation error at the affected block upon saving or updating:

amp-issue-block-level-notification

- [x] Handling when blocks get out of sync.

- [x] Handling notices when duplicating blocks.

- [x] Reset warning notice after removing all affected blocks.

- [x] Display block-level notices in native mode or when `force_sanitization` is enabled.

- [x] Display block-level notices on first loading the post into the editor.
When first opening a post in the editor, the validation notices are displayed.  They update when saving.

- [x] Remove and disable notice handling when the AMP toggle is off.
When the post's AMP toggle is off, the top and block level notices are removed.  It then bails out.  When the toggle is turned to on, the notices reappear after saving.

Closes #1292.

### DIFF
--- a/assets/js/amp-block-validation.js
+++ b/assets/js/amp-block-validation.js
@@ -382,7 +382,7 @@ var ampBlockValidation = ( function() { // eslint-disable-line no-unused-vars
 
 			return wp.data.withSelect( function( select, ownProps ) {
 				return _.extend( {}, ownProps, {
-					ampBlockValidationErrors: select( module.storeName ).getBlockValidationErrors( ownProps.id )
+					ampBlockValidationErrors: select( module.storeName ).getBlockValidationErrors( ownProps.clientId )
 				} );
 			} )( AmpNoticeBlockEdit );
 		}
--- a/assets/js/amp-block-validation.js
+++ b/assets/js/amp-block-validation.js
@@ -72,29 +72,29 @@ var ampBlockValidation = ( function() { // eslint-disable-line no-unused-vars
 			return wp.data.registerStore( module.storeName, {
 				reducer: function( _state, action ) {
 					var state = _state || {
-						blockValidationErrorsByUid: {}
+						blockValidationErrorsByClientId: {}
 					};
 
 					switch ( action.type ) {
 						case 'UPDATE_BLOCKS_VALIDATION_ERRORS':
 							return _.extend( {}, state, {
-								blockValidationErrorsByUid: action.blockValidationErrorsByUid
+								blockValidationErrorsByClientId: action.blockValidationErrorsByClientId
 							} );
 						default:
 							return state;
 					}
 				},
 				actions: {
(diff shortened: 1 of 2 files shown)


hellofromtonya:
@westonruter The recent commits fixed 2 issues:

1. Waiting for the editor to load up before handling the validation error state change.
2. Handling when the blocks get out of sync.

For the last one, the blocks get out of sync.  It happens every time when the editor is loading on page load.  And I've seen it in other intermittent editing/saving workflow.  With these changes, it now checks if the block order changed.  If yes, it resets the last validation errors that were previously stored.

Can you test it on your machine?

BTW We still have an issue when duplicating a block.  I'll continue working on that problem.

kienstra:
**Notices Above Blocks**

It looks like the Notices above blocks now appear as expected, at least on the blocks I tested:

notices-appear-above-blocks

notice-above

Also, a Notice appears above a duplicate block on updating the post:
duplicate-and-update

westonruter:
Great work!

PR #1299, opened 2018-08-01
Improve validation and presentation of analytics form

westonruter:
When reviewing #1296 I was inspired with some quick wins to greatly facilitate the input of analytics data, which I expect will also lessen a lot of support topics on the issue.

* Use HTML5 form validation to prevent invalid JSON from being submitted and then lost when the page reloads.
* Parse the textarea value as JSON at input event to ensure JSON object is supplied.
* Add `required` attribute to the type and config fields to prevent empty values from being submitted.
* Add red border around invalid textures.
* Pretty-print saved JSON when rendering a config.
* Let initial value of analytics textarea be an empty object `{}`.
* Do client-side JSON validation.
* Add monospace font to texture.
* Add `{...}` as placeholder for textarea.
* Add intro text to explain required format for analytics JSON config. Let text be expanded when there are no entries saved.
* Add `googleanalytics` placeholder for type.

# Before:

image

image

# After:

image

In the future we could go a step further and add the CodeMirror editor: https://make.wordpress.org/core/2017/10/22/code-editing-improvements-in-wordpress-4-9/

### DIFF
--- a/includes/options/views/class-amp-analytics-options-submenu-page.php
+++ b/includes/options/views/class-amp-analytics-options-submenu-page.php
@@ -64,7 +64,7 @@ class AMP_Analytics_Options_Submenu_Page {
 								name="<?php echo esc_attr( $id_base . '[config]' ); ?>"
 								class="amp-analytics-input"
 								placeholder="{...}"
-								title="<?php esc_attr_e( 'A JSON object begins with a "{" and ends with a "}". Do not include any HTML tags like "<amp-analytics>" or "<script>".', 'amp' ); ?>"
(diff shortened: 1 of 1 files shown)


tryazi:
Does this analysis section work after the new version. I'm taking many url hits but I can't see the tracking

`{
    "vars": {
        "account": "UA-XXXX-1"
    },
    "triggers": {
        "trackPageview": {
            "on": "visible",
            "request": "pageview"
        }
    }
}`

westonruter:
@tryazi Please share the URL to your AMP page.

tryazi:
example
https://www.google.com/amp/s/www.odevbul.org/konu/yasadiginiz-yeri-yoneten-kisiler-olmasaydi-hangi-sorunlar-ortaya-cikardi%3fbbp_reply_to=12730&_wpnonce=306789645b&amp

tryazi:
Thank you. Wait a little longer. Hope it gets better with time

PR #1306, opened 2018-08-01
Fix the allowed attributes for `amp-position-observer`

felixarntz:
I tried to use `amp-position-observer` with the plugin earlier and noticed that it strips out the `target-id` attribute as invalid. That is because the whitelist for that tag only allows a `target` attribute which appears to be wrong to me. [The AMP documentation states that `target-id` is the correct name of the attribute.](https://www.ampproject.org/docs/reference/components/amp-position-observer#target-id-(optional))

### DIFF
--- a/includes/sanitizers/class-amp-allowed-tags-generated.php
+++ b/includes/sanitizers/class-amp-allowed-tags-generated.php
@@ -2519,7 +2519,7 @@ class AMP_Allowed_Tags_Generated {
 					'noloading' => array(
 						'value' => '',
 					),
-					'target' => array(),
+					'target-id' => array(),
 					'viewport-margins' => array(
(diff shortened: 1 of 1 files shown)


westonruter:
Thanks for pointing this out. In this case it means it is time to do another update of the generated PHP from the spec. Previous PRs where this was done:

* https://github.com/Automattic/amp-wp/pull/943
* https://github.com/Automattic/amp-wp/pull/1024
* https://github.com/Automattic/amp-wp/pull/1057
* https://github.com/Automattic/amp-wp/pull/1172

Here are some instructions: https://github.com/Automattic/amp-wp/blob/develop/contributing.md#updating-allowed-tags-and-attributes

westonruter:
I'll try to re-generating from the spec by tomorrow unless you beat me to it.

westonruter:
It turns out that `target-id` is actually incorrect in the docs:

image

It _is_ supposed to be just `target`: https://github.com/ampproject/amphtml/blob/1143ef048f4845ff0ae21a78c07053646508fc36/extensions/amp-position-observer/validator-amp-position-observer.protoascii#L45

image

PR #1309, opened 2018-08-02
Preserve whitespace when serializing the DOM as HTML

### LINKED ISSUE #1304: Blank lines are added when serializing HTML between elements that should not have whitespace
Class AMP_Content_Sanitizer in class-amp-content.php

$content containing `><` from the editor is fine up until where it gets sanitized. At this point a blank line appears to get inserted into $sanitzed_content.   If you trace it through I think there is a problem in class AMP_Content_Sanitizer with this file being called class-amp-dom-utils.php - it is doing something strange. In other words

`<li><strong>` is in $content (the editor) but in $sanitzed_content you get `<li>BLANK LINE<strong>`

Another example 

```
<h2>Blal Bla <span class="green">more text</span></h2> gets rendered as 

<h2>Blal Bla <span class="green">more text</span>
</h2>
```

westonruter:
Add workaround to prevent PHP from adding whitespace to format the serialization of HTML. This prevents bugs related to spaces being added between elements which can break layouts. For example, without this fix, the following HTML:
```html
<style>
.cols-test > li {
    width: 50%;
    display: inline-block;
    outline: solid 1px red;
}
</style>
<ul class="cols-test"><li>column 1</li><li>column 2</li></ul>
```
Which should get rendered as:
image

Will get serialized instead as:

```html
<ul class="cols-test">
<li>column 1</li>
<li>column 2</li>
</ul>
```

And displayed as (with text selected to show whitespace insertion):

image

For some more background on the changes here, see https://stackoverflow.com/q/51660286/93579

Fixes #1304.

### DIFF
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -410,14 +410,27 @@ class AMP_DOM_Utils {
 			}
 		}
 
-		/*
-		 * Temporarily add fragment boundary comments in order to locate the desired node to extract from
-		 * the given HTML document. This is required because libxml seems to only preserve whitespace when
-		 * serializing when calling DOMDocument::saveHTML() on the entire document. If you pass the element
-		 * to DOMDocument::saveHTML() then formatting whitespace gets added unexpectedly. This is seen to
-		 * be fixed in PHP 7.3, but for older versions of PHP the following workaround is needed.
-		 */
-		if ( version_compare( PHP_VERSION, '7.3', '<' ) ) {
+		if ( version_compare( PHP_VERSION, '7.3', '>=' ) ) {
+			$html = $dom->saveHTML( $node );
+		} else {
+			/*
+			 * Temporarily add fragment boundary comments in order to locate the desired node to extract from
+			 * the given HTML document. This is required because libxml seems to only preserve whitespace when
+			 * serializing when calling DOMDocument::saveHTML() on the entire document. If you pass the element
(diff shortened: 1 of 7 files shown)


westonruter:
Note that to test, do not use the `<ul>` example in post content. There is a `the_content` filter which adds line breaks for that. Instead try the markup originally reported in #1304.

westonruter:
Here is a build to test: [amp.zip](https://github.com/Automattic/amp-wp/files/2255287/amp.zip) (v1.0-beta1-95dc4069-20180802T220041Z)

Frenchomatic:
@westonruter Just to confirm in case you missed it in my post - this fixes it all for me both in your plugin and the AMP for WP one too which uses the utils folder in particular - thank you.

Zabi09:
@westonruter 
We really appreciate your hard work in fixing the spaces issue from the source code and we have tested the commits on PHP Version 7.2.5 and it was working perfectly,

Will you please tell us how to test the same with the 7.3 version of PHP and for your information we have downloaded the beta version of 7.3 version but unable to check it

Please let us know So that we can check it and update the code on our AMPforWP plugin

westonruter:
@Zabi09 From what I can tell the issue is no longer present in PHP 7.3: https://3v4l.org/TS2tX

This being the case we can bypass the whole logic in:

https://github.com/Automattic/amp-wp/blob/95dc40692df5f3216d5b0ffa35136ab6f0b1383d/includes/utils/class-amp-dom-utils.php#L413-L414

I'll be updating the Travis build matrix to include PHP 7.3 once it is available: https://github.com/travis-ci/travis-ci/issues/9717

Issue #1237, opened 2018-07-02
Images on attachment page appear stretched


westonruter:
In Twenty Seventeen at least, an image on the attachment template looks stretched:

![image](https://user-images.githubusercontent.com/134745/42182514-456e1fc2-7df3-11e8-9537-598739aef71c.png)

```html
<amp-img width="300" height="300" src="https://src.wordpress-develop.test/wp-content/uploads/2018/05/cropped-wordpress-amp-300x300.png" class="attachment-medium size-medium amp-wp-enforced-sizes" alt="" srcset="https://src.wordpress-develop.test/wp-content/uploads/2018/05/cropped-wordpress-amp-300x300.png 300w, https://src.wordpress-develop.test/wp-content/uploads/2018/05/cropped-wordpress-amp-150x150.png 150w, https://src.wordpress-develop.test/wp-content/uploads/2018/05/cropped-wordpress-amp-100x100.png 100w, https://src.wordpress-develop.test/wp-content/uploads/2018/05/cropped-wordpress-amp-270x270.png 270w, https://src.wordpress-develop.test/wp-content/uploads/2018/05/cropped-wordpress-amp-192x192.png 192w, https://src.wordpress-develop.test/wp-content/uploads/2018/05/cropped-wordpress-amp-180x180.png 180w, https://src.wordpress-develop.test/wp-content/uploads/2018/05/cropped-wordpress-amp-32x32.png 32w, https://src.wordpress-develop.test/wp-content/uploads/2018/05/cropped-wordpress-amp.png 512w" sizes="100vw" layout="intrinsic"></amp-img>
```

Non-AMP version:

![image](https://user-images.githubusercontent.com/134745/42182488-32a9e52e-7df3-11e8-95af-477d3ff4c35e.png)

```html
[...]

PR #1321, opened 2018-08-06
Filter the attachment image attributes for the Twenty Seventeen theme.

hellofromtonya:
Per issue #1237, the Twenty Seventeen theme is adding `$attr['sizes']  = '100vw';` which stretches images on the attachment page when in AMP mode.

AMP uses a combination of the `layout` and `sizes` attributes to determine how to set the inline style width.

This PR filters `'wp_get_attachment_image_attributes'` to change the layout to responsive and sizes to the calculated image sizes per `wp_get_attachment_image_sizes()`.

~- Changing the layout to responsive ensures the image renders within the parent container in all viewports.~
- Changing the `sizes` value provides the information AMP needs to set the inline width.

### Before

Notice the differences in non-AMP vs. AMP rendering _before_ this PR is applied.  The images are stretched due to the `sizes` and `layout` attributes.

Here is a 150x150 thumbnail image before this PR is applied:

before-stretched-150x150

Here is a 1200w image before this PR is applied:

before-stretched-1200w

### After

Now let's compare both of the above images after applying this PR.  Notice the images render the same in non-AMP and AMP modes without stretching.

after-stretched-150x150

after-stretched-1200w

### Is this a fix or bandaid?

There is an [open issue with the AMPHTML project](https://github.com/ampproject/amphtml/issues/17053) about images being stretched.  IMO this PR fixes the layout and sizes for the Twenty Seventeen theme as it is adjusting the sizes to `100vw` when on the attachment page.  Therefore, I think this PR could be rolled in without worrying about the changes from the open issue.

### DIFF
--- a/includes/sanitizers/class-amp-core-theme-sanitizer.php
+++ b/includes/sanitizers/class-amp-core-theme-sanitizer.php
@@ -78,6 +78,7 @@ class AMP_Core_Theme_Sanitizer extends AMP_Base_Sanitizer {
 				'//header[@id = "masthead"]//a[ contains( @class, "menu-scroll-down" ) ]',
 			),
 			'set_twentyseventeen_quotes_icon'     => array(),
+			'add_twentyseventeen_attachment_image_attributes' => array(),
 		),
 
 		// Twenty Sixteen.
@@ -342,6 +343,30 @@ class AMP_Core_Theme_Sanitizer extends AMP_Base_Sanitizer {
 
 			return $content;
 		} );
+	}
+
+	/**
+	 * Add filter to adjust the attachment image attributes to ensure attachment pages have a consistent <amp-img> rendering.
+	 *
+	 * This is only used in Twenty Seventeen.
+	 *
+	 * @since 1.0
+	 * @link https://github.com/WordPress/wordpress-develop/blob/ddc8f803c6e99118998191fd2ea24124feb53659/src/wp-content/themes/twentyseventeen/functions.php#L545:L554
+	 */
+	public static function add_twentyseventeen_attachment_image_attributes() {
+		add_filter( 'wp_get_attachment_image_attributes', function ( $attr, $attachment, $size ) {
+			if ( ! is_attachment() ) {
+				return $attr;
+			}
+
+			$sizes = wp_get_attachment_image_sizes( $attachment->ID, $size );
+			if ( false !== $sizes ) {
+				$attr['layout'] = 'responsive';
+				$attr['sizes']  = $sizes;
+			}
+
+			return $attr;
+		}, 11, 3 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


westonruter:
@hellofromtonya does this negatively effect the behavior in the featured image?

It seems the [purpose of `100vw`](https://github.com/WordPress/wordpress-develop/blob/c52e7c79b2fd6381ba89fb8c40624e6621284743/src/wp-content/themes/twentyseventeen/functions.php#L534-L554) is to:

> Add custom image sizes attribute to enhance responsive image functionality for post thumbnails.

Does that negatively impact such an image?

hellofromtonya:
@westonruter 

>does this negatively effect the behavior in the featured image?

It is only on the [attachment page](https://github.com/Automattic/amp-wp/pull/1321/files#diff-5d05e505f93786061c2078486784c46dR358) and does not impact any other page.  Therefore, it will not impact the featured image.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -344,6 +345,30 @@ public static function set_twentyseventeen_quotes_icon() {
...
+	 */
+	public static function add_twentyseventeen_attachment_image_attributes() {
+		add_filter( 'wp_get_attachment_image_attributes', function ( $attr, $attachment, $size ) {
+			if ( ! is_attachment() ) {
+				return $attr;
+			}
+
+			$sizes = wp_get_attachment_image_sizes( $attachment->ID, $size );
+			if ( false !== $sizes ) {
+				$attr['layout'] = 'responsive';
```
kienstra: This is a a minor point, but would it be possible to not set the `layout` here, and let the image sanitizer [default to intrinsic](https://github.com/Automattic/amp-wp/blob/c236cc472ddcdfa22b373d4d32f24665034422f8/includes/sanitizers/class-amp-img-sanitizer.php#L247) (if there's a `height` and `width`)?

So far, it looks like the image displays the same with `intrinsic` and `responsive`. But you've tested this more 😄 

Of course, you have thought about this:
>Changing the layout to responsive ensures the image renders within the parent container in all viewports.

hellofromtonya: @kienstra with the `intrinsic` layout, we then get into the issue over wider images overflowing the parent container.  But let me do a little more investigation to gather some test data for us.

kienstra: Thanks, @hellofromtonya!

hellofromtonya: @kienstra Upon review, we can allow the layout to default and be handled in the image sanitizer.

Why?  The wide image overflow problem is tied to the Gutenberg image block.  That image block adjusts the width on the added parent `<figure>`.  When the layout is set to `intrinsic`, it sets the width to `fit-content`.  Combining that width with the inline width that AMP sets causes the overflowing problem.

Here we do not have that issue.  Therefore, we do not need to force the layout attribute. 

I've removed it in [commit 74dc97a](https://github.com/Automattic/amp-wp/pull/1321/commits/74dc97a0554eaed405c5fc26e7302248c628ff25).

kienstra: Thanks, @hellofromtonya! Sorry to nitpick 😄

kienstra:

If it's alright, I'm moving this to 'Ready For Merging,' as I don't think it needs functional beyond the testing already done for this PR. Feel free to move it back.

PR #1325, opened 2018-08-07
Improve post processor response cache

### LINKED ISSUE #1239: Twenty Seventeen's use of uniqid() breaks post-processor response cache
As a site admin running AMP on my WordPress site, I should be able to know when elements cannot be cached and I should be able to cache unique elements (via usage of `uniqid()` for example) when a WP theme or plugin generates this.

- [ ] AC1: Using `TwentySeventeen` and its uniqid method break caching as described below. The plugin should automatically catch when a theme (such as this one) breaks caching, reporting it if it cannot be fixed automatically.
- [ ] AC2: Propose patch to Twenty Seventeen to use an alternative to use a counter or something more stable than `uniqid()`.
- [ ] AC3: Introduce new filter to apply to buffered content which can be used (sparingly) to fixup content that would otherwise break response caching.

---

Twenty Seventeen uses `uniqid()` to generate IDs for HTML elements:

https://github.com/WordPress/wordpress-develop/blob/8e96abbcc2214c3566b814619c4b038d59ad77d9/src/wp-content/themes/twentyseventeen/searchform.php#L13

https://github.com/WordPress/wordpress-develop/blob/8e96abbcc2214c3566b814619c4b038d59ad77d9/src/wp-content/themes/twentyseventeen/inc/icon-functions.php#L77-L94

Unfortunately this causes the response to differ for each request and causes a cache MISS for the plugin's AMP response cache. The result is that the post-processor will always be invoked on a site with Twenty Seventeen, and the object cache will get polluted with values that never get used.

I'm not sure if there is something we can do to fix this. It would have been preferable if Twenty Seventeen had instead introduced a unique-ID generator like:

```php
function twentyseventeen_uniqid() {
    static $id = 0;
    return ++$id;
}
```

As this would mean the IDs would be unique within a response, but they would be repeated across multiple responses.

hellofromtonya:
Randomly generated content can occur on a web page.  When that happens, the response will differ for each request causing cache misses for the plugin's AMP response cache.  As @westonruter notes in Issue #1239:

>the object cache will get polluted with values that never get used.

This PR detects continual cache misses by caching the page's URL and then comparing the number of cached URLs to a threshold.  When that threshold is exceeded, it disables the response cache and then notifies the admin.

TODO
- [x] Detect continual cache misses
- [x] Disable caching
- [x] Notify
- [x] Provide a UI to reset the option's value and re-enable response caching.

Closes #1239.

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -2013,9 +2013,7 @@ class AMP_Theme_Support {
 	 */
 	private static function check_for_cache_misses() {
 		// If the cache miss threshold is exceeded, return true.
-		$cache_miss_url     = get_option( self::CACHE_MISS_URL_OPTION, false );
-		$exceeded_threshold = ! empty( $cache_miss_url );
-		if ( $exceeded_threshold ) {
+		if ( self::exceeded_cache_miss_threshold() ) {
 			return array( true, null );
 		}
 
@@ -2048,6 +2046,18 @@ class AMP_Theme_Support {
 		if ( get_option( self::CACHE_MISS_URL_OPTION ) ) {
 			delete_option( self::CACHE_MISS_URL_OPTION );
 		}
+	}
+
+	/**
+	 * Checks if cache miss threshold has been exceeded.
+	 *
+	 * @since 1.0
+	 *
+	 * @return bool
+	 */
+	public static function exceeded_cache_miss_threshold() {
+		$url = get_option( AMP_Theme_Support::CACHE_MISS_URL_OPTION, false );
+		return ! empty( $url );
 	}
 
 	/**
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -156,6 +156,8 @@ class AMP_Theme_Support {
 		 * action to template_redirect--the wp action--is used instead.
 		 */
 		add_action( 'wp', array( __CLASS__, 'finish_init' ), PHP_INT_MAX );
+
+		add_action( 'amp_reenable_response_cache', array( __CLASS__, 'reset_cache_miss_url_option' ) );
 	}
 
 	/**
@@ -2037,6 +2039,17 @@ class AMP_Theme_Support {
 		update_option( self::CACHE_MISS_URL_OPTION, amp_get_current_url() );
 		AMP_Options_Manager::update_option( 'enable_response_caching', false );
 		return array( true, null );
+	}
+
+	/**
+	 * Reset the cache miss URL option.
+	 *
+	 * @since 1.0
+	 */
+	public static function reset_cache_miss_url_option() {
+		if ( get_option( self::CACHE_MISS_URL_OPTION ) ) {
+			delete_option( self::CACHE_MISS_URL_OPTION );
+		}
 	}
 
 	/**
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -113,13 +113,6 @@ class AMP_Theme_Support {
 	protected static $support_added_via_option = false;
 
 	/**
-	 * Post-processor cache key.
-	 *
-	 * @var string
-	 */
-	public static $post_processor_cache_key;
-
-	/**
 	 * Initialize.
 	 *
 	 * @since 0.7
@@ -1782,11 +1775,10 @@ class AMP_Theme_Support {
 		$ampless_url = amp_remove_endpoint( $current_url );
 
 		// When response caching is enabled, determine if it should be turned off for cache misses.
-		$caches_for_url                 = null;
-		self::$post_processor_cache_key = null;
+		$caches_for_url = null;
 		if ( true === $args['enable_response_caching'] ) {
-			self::$post_processor_cache_key = md5( $current_url );
-			$caches_for_url                 = wp_cache_get( self::$post_processor_cache_key, self::POST_PROCESSOR_CACHE_EFFECTIVENESS );
+			$post_processor_cache_key = md5( $current_url );
+			$caches_for_url           = wp_cache_get( $post_processor_cache_key, self::POST_PROCESSOR_CACHE_EFFECTIVENESS );
 
 			$args['enable_response_caching'] = (
 				empty( $caches_for_url )
@@ -1839,14 +1831,14 @@ class AMP_Theme_Support {
 				return $response_cache['body'];
 			}
 
-			$cache_response = function( $body, $validation_results ) use ( $response_cache_key, $caches_for_url ) {
+			$cache_response = function( $body, $validation_results ) use ( $response_cache_key, $post_processor_cache_key, $caches_for_url ) {
 				if ( empty( $caches_for_url ) ) {
 					$caches_for_url = array( $response_cache_key );
 				} else {
 					$caches_for_url[] = $response_cache_key;
 				}
 				wp_cache_set(
-					AMP_Theme_Support::$post_processor_cache_key,
+					$post_processor_cache_key,
 					$caches_for_url,
 					AMP_Theme_Support::POST_PROCESSOR_CACHE_EFFECTIVENESS,
 					MONTH_IN_SECONDS
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1780,6 +1780,15 @@ class AMP_Theme_Support {
 		if ( true === $args['enable_response_caching'] ) {
 			$post_processor_cache_key = md5( $current_url );
 			$caches_for_url           = wp_cache_get( $post_processor_cache_key, self::POST_PROCESSOR_CACHE_EFFECTIVENESS );
+
+			// Turn off response caching when the number of cached URLs exceeds the threshold.
+			$args['enable_response_caching'] = (
+				empty( $caches_for_url )
+				||
+				! is_array( $caches_for_url )
+				||
+				count( $caches_for_url ) < self::CACHE_MISS_THRESHOLD
+			);
 		}
 
 		// Return cache if enabled and found.
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -113,13 +113,6 @@ class AMP_Theme_Support {
 	protected static $support_added_via_option = false;
 
 	/**
-	 * Response cache key.
-	 *
-	 * @var string
-	 */
-	public static $response_cache_key;
-
-	/**
 	 * Post-processor cache key.
 	 *
 	 * @var string
@@ -1805,11 +1798,10 @@ class AMP_Theme_Support {
 		}
 
 		// Return cache if enabled and found.
-		$cache_response           = null;
-		self::$response_cache_key = '';
+		$cache_response = null;
 		if ( true === $args['enable_response_caching'] ) {
 			// Set response cache hash, the data values dictates whether a new hash key should be generated or not.
-			self::$response_cache_key = md5( wp_json_encode( array(
+			$response_cache_key = md5( wp_json_encode( array(
 				$args,
 				$response,
 				self::$sanitizer_classes,
@@ -1817,7 +1809,7 @@ class AMP_Theme_Support {
 				AMP__VERSION,
 			) ) );
 
-			$response_cache = wp_cache_get( self::$response_cache_key, self::RESPONSE_CACHE_GROUP );
+			$response_cache = wp_cache_get( $response_cache_key, self::RESPONSE_CACHE_GROUP );
 
 			// Make sure that all of the validation errors should be sanitized in the same way; if not, then the cached body should be discarded.
 			$blocking_error_count = 0;
@@ -1847,11 +1839,11 @@ class AMP_Theme_Support {
 				return $response_cache['body'];
 			}
 
-			$cache_response = function( $body, $validation_results ) use ( $caches_for_url ) {
+			$cache_response = function( $body, $validation_results ) use ( $response_cache_key, $caches_for_url ) {
 				if ( empty( $caches_for_url ) ) {
-					$caches_for_url = array( AMP_Theme_Support::$response_cache_key );
+					$caches_for_url = array( $response_cache_key );
 				} else {
-					$caches_for_url[] = AMP_Theme_Support::$response_cache_key;
+					$caches_for_url[] = $response_cache_key;
 				}
 				wp_cache_set(
 					AMP_Theme_Support::$post_processor_cache_key,
@@ -1861,7 +1853,7 @@ class AMP_Theme_Support {
 				);
 
 				return wp_cache_set(
-					AMP_Theme_Support::$response_cache_key,
+					$response_cache_key,
 					compact( 'body', 'validation_results' ),
 					AMP_Theme_Support::RESPONSE_CACHE_GROUP,
 					MONTH_IN_SECONDS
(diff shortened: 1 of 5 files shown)


hellofromtonya:
@westonruter Commits through to [1d1f399](https://github.com/Automattic/amp-wp/pull/1325/commits/1d1f39991924c1a31784ee04febbb8e56ffd4088) are a prototype per [our discussions here](https://github.com/Automattic/amp-wp/issues/1239#issuecomment-410879591).  It's implementing:

1. [Cache the URLs](https://github.com/Automattic/amp-wp/pull/1325/files#diff-2585472f207548f626a43a7ff3cab922R1796).
2. Then [determine if we should turn off the response caching](https://github.com/Automattic/amp-wp/pull/1325/files#diff-2585472f207548f626a43a7ff3cab922R1798).
3. If caching is enabled, then inside of the caching closure, [we'll cache the new URL](https://github.com/Automattic/amp-wp/pull/1325/files#diff-2585472f207548f626a43a7ff3cab922R1856).

Take a look and see what you think.  BTW Notices are not included yet.

includes/class-amp-theme-support.php
```diff
@@ -1760,19 +1788,36 @@ public static function prepare_response( $response, $args = array() ) {
 		$current_url = amp_get_current_url();
 		$ampless_url = amp_remove_endpoint( $current_url );
 
+		// When response caching is enabled, determine if it should be turned off for cache misses.
+		$caches_for_url                 = null;
+		self::$post_processor_cache_key = null;
+		if ( true === $args['enable_response_caching'] ) {
+			self::$post_processor_cache_key = md5( $current_url );
+			$caches_for_url                 = wp_cache_get( self::$post_processor_cache_key, self::POST_PROCESSOR_CACHE_EFFECTIVENESS );
```
westonruter: Sone of the subsequent checks could perhaps be simplified by checking if this is an array here, and if not, just set it to be an empty array.

includes/class-amp-theme-support.php
```diff
@@ -98,6 +112,20 @@ class AMP_Theme_Support {
 	 */
 	protected static $support_added_via_option = false;
 
+	/**
+	 * Response cache key.
+	 *
+	 * @var string
+	 */
+	public static $response_cache_key;
```
westonruter: Is this necessary? Aren't references limited to a single method? Since this is a static class it would be better of we limited the state that is persisted in it's class variables IMO.

includes/class-amp-theme-support.php
```diff
@@ -1802,9 +1847,21 @@ public static function prepare_response( $response, $args = array() ) {
 				return $response_cache['body'];
 			}
 
-			$cache_response = function( $body, $validation_results ) use ( $response_cache_key ) {
+			$cache_response = function( $body, $validation_results ) use ( $caches_for_url ) {
+				if ( empty( $caches_for_url ) ) {
+					$caches_for_url = array( AMP_Theme_Support::$response_cache_key );
+				} else {
+					$caches_for_url[] = AMP_Theme_Support::$response_cache_key;
```
westonruter: Per above, this could be the sole line instead of the if/else.

includes/class-amp-theme-support.php
hellofromtonya: @westonruter I added this public property for the following purposes:

1. to give us the ability to deeply test by knowing what the cache key is and then checking for it in the cache. 
2. to simplify access to the logic within the caching closure, i.e. meaning that we don't have to pass it via `use`.

It serves no other external purpose as part of an API.

Let's think about it.

From a testing point of view, we can _infer_ that we disabled the cache by the state of the post-processor cache.  While that's not a direct test on the response cache, both are within the closure.  Therefore, it's a minimal risk to test by inference.

Hmm, I like that approach better actually.  We get the benefit only exposing what is needed externally while avoided the extra work of resetting persistent state.

I'll make that change.

westonruter:
For the notice, are you thinking there would be a new key added to the `amp-options` option which would flag whether or not the response cache is enabled?

hellofromtonya:
>For the notice, are you thinking there would be a new key added to the amp-options option which would flag whether or not the response cache is enabled?

Well that's a good question @westonruter.  I think to find the solution, we need to explore what we want to report back to the developer/admin.   

We are disabling the cache on a page-by-page basis.  Do we want to give a list of URLs (i.e. that have cache disabled)? Or do we want a general notice?

Following our other solution paths, we may want to provide a list and notice.

westonruter:
@hellofromtonya humm, how about showing the notice on the general settings page, similar to how we are showing a notice for not having the object cache active. 

I didn't think about letting the cache be disabled on a page-by-page basis. I realize that is how it is implemented now. I think that once we get to the point where the uncached count is greater than `CACHE_MISS_THRESHOLD` we can just set an option to stop the caching. In the future we could go to a page-by-page basis, but I think that could be overkill for now.

westonruter:
Also, currently the `$post_processor_cache_key` is specific to the current URL, so there would be no way to query for all URLs that are surpassing the `CACHE_MISS_THRESHOLD`, since the object cache cannot be queried. Maybe a compromise is that we could store the URL that causes the `CACHE_MISS_THRESHOLD` to be surpassed, and then use that URL in the notice.

Something else: if there are logged-in users and the REST API is being used on the frontend, it is likely that the API nonce is going to be output. This nonce will be unique for each user. This being the case, it is very likely that the `CACHE_MISS_THRESHOLD` will be very quickly passed, since each user landing on the homepage would then cause the increment to go up. Maybe this is expected and is proper behavior, but it's also something to be aware of as it could make the cache very prone to being turned off.

westonruter:
Maybe such caching should just be disabled entirely if the user is logged in.

Also, I just realized something else, when we do:

```php
$caches_for_url[] = $response_cache_key;
```

We could easily cause for the homepage, for example, to trigger the cache to get disabled once you've published 3 posts, each causing the homepage to change. So what I think we're missing here is a _timestamp_ for when the cache MISS happened. And then what we'd want to do is only consider such entries that have happened within a recent time period (e.g. 10 minutes) for whether the `CACHE_MISS_THRESHOLD` has been surpassed.

hellofromtonya:
>Also, currently the $post_processor_cache_key is specific to the current URL, so there would be no way to query for all URLs that are surpassing the CACHE_MISS_THRESHOLD, since the object cache cannot be queried.

Correct.  

>Maybe a compromise is that we could store the URL that causes the `CACHE_MISS_THRESHOLD` to be surpassed, and then use that URL in the notice.

I agree.  That's the next step as we think about how to notify.

>We could easily cause for the homepage, for example, to trigger the cache to get disabled once you've published 3 posts, each causing the homepage to change. So what I think we're missing here is a timestamp for when the cache MISS happened. And then what we'd want to do is only consider such entries that have happened within a recent time period (e.g. 10 minutes) for whether the `CACHE_MISS_THRESHOLD` has been surpassed.

How about instead if we set the [cache expiration](https://github.com/Automattic/amp-wp/pull/1325/files#diff-2585472f207548f626a43a7ff3cab922R1842) to say 10 minutes?

westonruter:
> How about instead if we set the cache expiration to say 10 minutes?

Yeah, that sounds good.

hellofromtonya:
@westonruter I'm thinking we want to provide a list of URLs with links to the impacted page.  I'm picturing an area on our General Settings Page that adds a section for caching _when_ we have something in that option.  

We'll need a way to clear out or acknowledge those URLs.

includes/options/class-amp-options-manager.php
```diff
@@ -319,4 +320,27 @@ public static function persistent_object_caching_notice() {
...
+		}
+
+		$cache_miss_url = get_option( AMP_Theme_Support::CACHE_MISS_URL_OPTION, false );
+		if ( empty( $cache_miss_url ) ) {
+			return;
+		}
+
+		printf(
+			'<div class="notice notice-warning"><p>%s <a href="%s">%s</a></p></div>',
+			esc_html__( 'Response caching was disabled due to exceeding the cache miss threshold.', 'amp' ),
```
westonruter: We'll probably need to figure out a way to explain what this means to non-developers. Maybe there should be a link to the Wiki with more information about what causes this? 

Also, should the notice be dismissible?

Should there be a checkbox to re-enable? If so, then the notice could be displayed inline with the checkbox.

hellofromtonya: >We'll probably need to figure out a way to explain what this means to non-developers. Maybe there should be a link to the Wiki with more information about what causes this?

Agreed.  We need more work on the messaging.  I've noted it as a `@todo`.

>should the notice be dismissible?

I think so.  I think it's okay that the developer or admin dismisses this warning. 

Thinking through it, I'd like to use this top-level warning as an _alert_: "Hey, we automatically turned off caching. Go to the AMP > General page for more information."  Use it to alert them that we did something automatically and it needs their attention.

>Should there be a checkbox to re-enable? If so, then the notice could be displayed inline with the checkbox.

Agreed. This is the next step, i.e. adding a re-enable checkbox that appears when caching is disabled.

includes/options/class-amp-options-manager.php
```diff
@@ -319,4 +327,26 @@ public static function persistent_object_caching_notice() {
...
+		}
+
+		if ( self::get_option( 'enable_response_caching' ) ) {
+			return;
+		}
+
+		printf(
+			'<div class="notice notice-warning is-dismissible"><p>%s <a href="%s">%s</a></p></div>',
+			esc_html__( "The AMP plugin's response cache disabled due to detecting randomly generated content.", 'amp' ),
+			esc_url( 'https://github.com/Automattic/amp-wp/wiki/Response-cache#automatically-disabling-of-the-response-cache' ),
```
hellofromtonya: @westonruter I opened a [new Wiki page](https://github.com/Automattic/amp-wp/wiki/Response-cache) and started the process of explaining the cache.  Once we get the UI done, then screenshots can be added to the page to provide users/developers more clarity.

includes/options/class-amp-options-manager.php
```diff
@@ -48,6 +49,7 @@ public static function register_settings() {
 
 		add_action( 'update_option_' . self::OPTION_NAME, array( __CLASS__, 'maybe_flush_rewrite_rules' ), 10, 2 );
 		add_action( 'admin_notices', array( __CLASS__, 'persistent_object_caching_notice' ) );
+		add_action( 'admin_notices', array( __CLASS__, 'render_cache_miss_notice' ) );
```
hellofromtonya: @westonruter As we are automatically turning off the response cache, I'm thinking we may want to notify not only on this settings page, but on other pages too.  Why? We can't be certain the admin/developer will come to this page.  At the very least, I'm thinking this notice should be on all AMP settings' pages.  What do you think?

westonruter: What if we limit it to the general settings page but add an indicator to the admin menu item so that the user knows there has something changed on this page?

includes/options/class-amp-options-manager.php
```diff
@@ -199,6 +200,12 @@ public static function validate_options( $new_options ) {
 		// Store the current version with the options so we know the format.
 		$options['version'] = AMP__VERSION;
 
+		// Handle the caching option.
+		$options['enable_response_caching'] = ! empty( $new_options['enable_response_caching'] );
+		if ( $options['enable_response_caching'] ) {
+			do_action( 'amp_reenable_response_cache' );
```
westonruter: Wouldn't it not be better to call the method here rather than indirectly trigger the action that the method is is connected to? Is there any utility for the `amp_reenable_response_cache` action generally?

includes/options/class-amp-options-menu.php
```diff
@@ -109,6 +109,17 @@ public function add_menu_items() {
 			)
 		);
 
+		add_settings_field(
```
westonruter: This setting should only be added if `wp_using_ext_object_cache()` returns true, as we should only do response caching if we know that the cached value will persist across requests.

tests/test-class-amp-options-manager.php
```diff
@@ -100,12 +101,12 @@ public function test_get_and_set_options() {
 				'disable_admin_bar'       => false,
 				'all_templates_supported' => true,
 				'supported_templates'     => array( 'is_singular' ),
+				'enable_response_caching' => true,
```
westonruter: The default value here can be the return value of `wp_using_ext_object_cache()`.

includes/options/class-amp-options-manager.php
```diff
@@ -31,6 +31,7 @@ class AMP_Options_Manager {
 		'disable_admin_bar'       => false,
 		'all_templates_supported' => true,
 		'supported_templates'     => array( 'is_singular' ),
+		'enable_response_caching' => true,
```
westonruter: The default value here can be the return value of `wp_using_ext_object_cache()`.

includes/options/class-amp-options-manager.php
```diff
@@ -198,6 +200,12 @@ public static function validate_options( $new_options ) {
 		// Store the current version with the options so we know the format.
 		$options['version'] = AMP__VERSION;
 
+		// Handle the caching option.
+		$options['enable_response_caching'] = ! empty( $new_options['enable_response_caching'] );
```
westonruter: The condition here should probably be `wp_using_ext_object_cache() && ! empty( $new_options['enable_response_caching'] )`

includes/options/class-amp-options-manager.php
hellofromtonya: Upon review, you're right.  There is not need for a separation action here.  Switching to call direct.

includes/options/class-amp-options-menu.php
hellofromtonya: True.  Good point.

includes/options/class-amp-options-manager.php
hellofromtonya: One way to accomplish this is to set the default state within `get_options()`:

```
self::$defaults['enable_response_caching'] = wp_using_ext_object_cache();
return array_merge( self::$defaults, $options );
```

includes/options/class-amp-options-manager.php
hellofromtonya: @westonruter I like that approach in general.  I think it spans more than just this PR and edge case.  Thinking about, it seems we could have a general indicator to alert of something that needs attention.  That indicator could link to the plugin's settings page where more information lives.

If we go for that approach, I think that's a separate feature, issue, and PR.

hellofromtonya:
@westonruter When we changed the conditions by using `wp_using_ext_object_cache()`, the notices were displaying incorrectly compared to the state.  In the commit [
e575cde](https://github.com/Automattic/amp-wp/pull/1325/commits/e575cdef2c9755521584e61599966f4b1e0d3ba8), I fixed it so that the notice appears when:

1. `wp_using_ext_object_cache()` is enabled.
2. The `enable_response_caching` option is disabled.
3. The cache miss threshold was exceeded.

includes/options/class-amp-options-manager.php
```diff
@@ -355,8 +355,8 @@ public static function render_cache_miss_notice() {
 
 		printf(
 			'<div class="notice notice-warning is-dismissible"><p>%s <a href="%s">%s</a></p></div>',
-			esc_html__( "The AMP plugin's response cache disabled due to detecting randomly generated content.", 'amp' ),
-			esc_url( 'https://github.com/Automattic/amp-wp/wiki/Response-cache#automatically-disabling-of-the-response-cache' ),
+			esc_html__( "The AMP plugin's post-processor cache disabled due to the detection of highly-variable content.", 'amp' ),
```
hellofromtonya: Much better @westonruter. Thank you 👍

westonruter:
@hellofromtonya Great work. This was tough!

hellofromtonya:
Collaborative effort between both of us.  Thanks for your help and for first identifying this as an issue.

PR #1338, opened 2018-08-14
Use all eligible post types when `all_templates_supported` is selected.

### LINKED ISSUE #1302: Changing settings can result in erroneous notices regarding post types
I've found sometimes when switching between template modes or supported templates that I get blasted with error notices regarding post type support, even when I made no changes to the post types:

image

hellofromtonya:
Per [issue 1302](https://github.com/Automattic/amp-wp/issues/1302#issuecomment-412910057), post type notices are being blasted when:

1. There are no options saved in the database AND
2. You switch from Classic to either Paired or Native.

The root cause for this problem is that the method `AMP_Options_Manager::check_supported_post_type_update_errors()` is only considering the selected post types when the `all_templates_supported` option is selected.

This PR fixes the problem by using all of the eligible post types (and not just the ones that are individually selected) when `all_templates_supported` option is selected.

Fixes #1302.

### DIFF
--- a/includes/options/class-amp-options-manager.php
+++ b/includes/options/class-amp-options-manager.php
@@ -208,8 +208,12 @@ class AMP_Options_Manager {
 	 * @see add_settings_error()
 	 */
 	public static function check_supported_post_type_update_errors() {
-		$supported_types = self::get_option( 'supported_post_types', array() );
-		foreach ( AMP_Post_Type_Support::get_eligible_post_types() as $name ) {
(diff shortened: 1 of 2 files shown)


kienstra:

If it's alright, I'm moving this to 'Ready For Merging.' It [looks like](https://github.com/Automattic/amp-wp/issues/1302#issuecomment-412910057) reproducing this on the [staging site](https://paired-ampconfdemo.pantheonsite.io) will be complex, though we could create a new site for this if you think it's important.

PR #1341, opened 2018-08-16
Fix Jed localization error in Classic mode

westonruter:
Fixes issue when open the Gutenberg editor in Classic mode:

image


westonruter:
Thanks. The issue is that `amp-editor-blocks-build` is not enqueued in classic mode, so that is why it switches to `wp-i18n`.

PR #1342, opened 2018-08-16
Try: Use amp-mathml and amp-fit-text components in Gutenberg block edit function's React components

westonruter:
In order to be able to create content using AMP components, it is important that the AMP components be able to be rendered in the editing context. For the next generation editor of WordPress, Gutenberg, this means being able to use AMP components inside of React components, and that changes to the AMP components' attributes will cause the shadow DOM to dynamically update as expected.

This does work as expected `amp-mathml` component:

maxresdefault

This may be because it doesn't have much shadow DOM to speak of, other than a single `iframe` child. 

However, using the `amp-fit-text` component in React does not work as expected:

maxresdefault (1)

Note that even though the `amp-fit-text` component's `height`, `min-font-size`, and `max-font-size` attributes are being modified, the changes to them are not getting applied to the shadow DOM.

### DIFF
--- a/assets/css/amp-editor-blocks.css
+++ b/assets/css/amp-editor-blocks.css
@@ -2,3 +2,6 @@
 .is-amp-fit-text + .blocks-font-size > .components-font-size-picker__custom-input {
     display: none;
 }
+.edit-post-visual-editor amp-fit-text p {
+    font-size: inherit;
+}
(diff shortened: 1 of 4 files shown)


swissspidy:
Quick note: we built our own amp-fit-text implementation for the AMP Stories editor. Maybe we could re-use that here as well?

swissspidy:
Created a new #3434 issue for amp-fit-text support in the editor.

PR #1347, opened 2018-08-20
[WIP] [amp-stories sub-task] Grid layers layout

miina:
This is a sub-PR of #1215 and should be merged to that PR.

The goal of this PR is to creating user interface for being able to edit AMP Story Grid Layers on top of each other within Gutenberg editor, using CSS Grid to lay out the blocks inside, and build the control for selecting a given layer.

### DIFF
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -1,14 +1,35 @@
-
-.editor-block-list__layout {
-	background-color: #ffdddd;
-}
-.editor-block-list__layout .editor-block-list__layout {
-	background-color: white;
-}
 div[data-type="amp/amp-story-page"] {
-	background-color: #ddffdd;
+	border: 1px solid #ffdddd;
 }
-div[data-type="amp/amp-story-grid-layer"] {
-	background-color: #ddddff;
+
+div[data-type="amp/amp-story-page"],
+div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__layout {
+	min-height: 450px;
+}
+div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__layout .editor-block-list__layout {
+	background-color: initial;
+	min-height: initial;
+}
+
+.editor-block-list__layout div[data-type="amp/amp-story-grid-layer"] {
+	position: absolute;
+	height: 100%;
+	width: 100%;
+}
+
+.editor-block-list__layout div[data-type="amp/amp-story-cta-layer"] {
+	position: absolute;
+	height: 20%;
+	width: 100%;
+	bottom: 0;
+}
+
+.editor-block-list__layout div[data-type="amp/amp-story-grid-layer"],
+.editor-block-list__layout div[data-type="amp/amp-story-cta-layer"] {
+	border: 1px solid #ddffdd;
+}
+
+.editor-block-list__layout div[data-type="amp/amp-story-grid-layer"].is-selected ~ div[data-type="amp/amp-story-grid-layer"] {
+	display: none;
 }
 


blocks/amp-story/block-selector.js
```diff
@@ -0,0 +1,73 @@
...
+				className += ' is-selected';
+			}
+			let blockType = getBlockType( block.name );
+			links.push(
+				<li className={ className } key={ 'selector-' + index }>
+					<a onClick={ ( e ) => {
+						e.stopPropagation();
+						if ( getSelectedBlock.clientId !== block.clientId ) {
+							// @todo This selects the first inner child instead for some reason.
+							selectBlock( block.clientId );
```
miina: @westonruter For some reason `selectBlock` here is triggering `selectBlock` for its first inner block, too, and ends up selecting the first inner block of the layer (e.g. the first paragraph within the layer block) instead of the layer itself. Any thoughts on why this might happen? (If not, I'll just continue looking into it, perhaps it's an issue to report.)

blocks/amp-story/block-selector.js
```diff
@@ -0,0 +1,73 @@
...
+
+		_.forEachRight( rootBlock.innerBlocks, function( block, index ) {
+			let className = 'component-editor__selector';
+			if ( isBlockSelected( block.clientId ) || hasSelectedInnerBlock( block.clientId ) ) {
+				className += ' is-selected';
+			}
+			let blockType = getBlockType( block.name );
+			links.push(
+				<li className={ className } key={ 'selector-' + index }>
+					<a onClick={ ( e ) => {
```
westonruter: Let's have this be a `button` so that it can be focused and accessible.

blocks/amp-story/block-selector.js
```diff
@@ -0,0 +1,73 @@
...
+		let className = 'component-editor__selector';
+		if ( isBlockSelected( this.props.rootClientId ) ) {
+			className += ' is-selected';
+		}
+
+		links.push(
+			<li className={ className } key='page-selector'>
+				<a onClick={ () => {
+					selectBlock( this.props.rootClientId );
+				}}>Page</a>
```
westonruter: “Page” here needs to be translatable.

blocks/amp-story/block-selector.js
```diff
@@ -0,0 +1,73 @@
...
+			let blockType = getBlockType( block.name );
+			links.push(
+				<li className={ className } key={ 'selector-' + index }>
+					<a onClick={ ( e ) => {
+						e.stopPropagation();
+						if ( getSelectedBlock.clientId !== block.clientId ) {
+							// @todo This selects the first inner child instead for some reason.
+							selectBlock( block.clientId );
+						}
+					}}>{ blockType.title }</a>
```
westonruter: This can just be “Layer” perhaps. In the end, I don't think that we'll be displaying the label and that we'd just have an icon instead, with perhaps a tooltip.

blocks/amp-story/block-selector.js
```diff
@@ -0,0 +1,73 @@
...
+		} );
+
+		let className = 'component-editor__selector';
+		if ( isBlockSelected( this.props.rootClientId ) ) {
+			className += ' is-selected';
+		}
+
+		links.push(
+			<li className={ className } key='page-selector'>
+				<a onClick={ () => {
```
westonruter: Per above, this should be `button`.

blocks/amp-story/block-selector.js
westonruter: I don't know why. Is this why I am unable to insert a grid layer upon selecting the page?

blocks/amp-story/block-selector.js
```diff
@@ -0,0 +1,73 @@
...
+					selectBlock( this.props.rootClientId );
+				}}>Page</a>
+			</li>
+		);
+
+		return (
+			<ul className="editor-selectors">
```
westonruter: This should probably be an `ol` since the layers are ordered.

assets/css/amp-editor-story-blocks.css
```diff
@@ -1,14 +1,56 @@
+div[data-type="amp/amp-story-page"] {
+	border: 1px solid #ffdddd;
+}
 
-.editor-block-list__layout {
-	background-color: #ffdddd;
+div[data-type="amp/amp-story-page"],
+div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__layout {
+	min-height: 450px;
```
westonruter: On desktop an AMP story has the following style rule:

```css
[desktop]>amp-story-page {
    margin: auto!important;
    max-height: 75vh!important;
    max-width: 45vh!important;
    min-width: 320px!important;
    min-height: 533px!important;
}
```

Would that ensure that the page is displayed with the right aspect ratio and dimensions?

blocks/amp-story/block-selector.js
```diff
@@ -0,0 +1,92 @@
...
+				}}>
+					{ __( 'Page', 'amp' ) }
+				</Button>
+			</li>
+		);
+
+		// @todo Creating a custom inserter since the default inserter doesn't allow taking the root client ID dynamically. Change if that becomes available.
+		return (
+			<ul className="editor-selectors">
+				<LayerInserter { ...inserterProps }/>
```
miina: Added a custom block inserter which just displays the relevant items in the list since it looks like the default `Inserter` is automatically taking the `clientId` of the selected block and then inserts a block after that. Since the list of layouts is fixed with the page block then this would mean that the inserter would always take the `amp-story-page` as the selected block and thus add the new block to the root level instead of inside the AMP Story page block (since the inserter inserts AFTER the selected block by default).

We could potentially create another issue in Gutenberg repo for allowing the `Inserter` to also accept the selected block ID as a param, however, it's likely not the most pressing issue right now and might take some time to get implemented.

(As of now) the custom inserter basically just displays the allowed blocks and that's it:
screen shot 2018-08-22 at 9 37 58 pm

@westonruter Do you have any thoughts on if this makes sense for now?

westonruter: That sounds good!

blocks/amp-story/block-selector.js
```diff
@@ -0,0 +1,97 @@
...
+						e.stopPropagation();
+
+						// @todo This selects the first inner child instead for some reason. Note that this also creates a new paragraph as the first child is it doesn't exist.
+						selectBlock( block.clientId );
+
+						// @todo This is a temporary workaround for selecting the correct block. Remove when possible.
+						let timeout = 50;
+						setTimeout( function() {
+							selectBlock( block.clientId );
+						}, timeout );
```
miina: Looks like using `selectBlock` a second time with timeout results in selecting the correct block eventually, however, there is another issue as well. Noticed that not the first inner child is selected but the first inner text block is selected. For example if the Grid Layer only has one Image block as the inner child and then `selectBlock` is used then a paragraph block is automatically created within this Grid Layer.

We should definitely look into the issues with `selectBlock` / alternatives at some moment ourselves.

miina:
@westonruter The PR of adding the control for selecting layers and using CSS grid for the inner blocks is ready for review now. There are some things that still need improving but maybe is better to do later / separately:
- `selectBlock` not working as expected but added a temporary workaround;
- Styling fixes -- maybe that's better to do when other pieces are also in place, such as the icons, amp-story-page background color selection etc.;

westonruter:
@miina would you update to fix the merge conflicts introduced by merging #1380?

miina:
@westonruter Done!

PR #1357, opened 2018-08-25
Fetch CSS over HTTP when URL lacks extension; convert font CDN stylesheets @imports to convert to links instead of fetching

### LINKED ISSUE #1317: Remove restriction that external stylesheets have .css extension
Initially the inclusion of external styles were limited to those that were on the filesystem. But since #1083 we also fetch external styles over HTTP and then cache them. So now if a URL lacks the `.css` extension (or it is not found on the filesystem even) we can just fallback to fetching the stylesheet over the network.

This issue was noticed on WordPress.com VIP where [automatic CSS concatenation](https://vip.wordpress.com/documentation/vip-go/vip-go-faq/#how-does-file-concatenation-and-minification-work-on-vip-go) results in a stylesheet URL that does not have a static filename extension.

westonruter:
If there is a stylesheet `link` with a URL which does not end in `css`, fetch it over HTTP instead of throwing a validation error.

Additionally, if there is a stylesheet that contains an `@import` to a stylesheet URL which is whitelisted CDN for fonts such as:

```css
@import "https://fonts.googleapis.com/css?family=Merriweather:300";
```

Then instead of fetching the CSS to inline, opt instead to convert into a normal `link` tag:

```html
<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Merriweather:300">
```

Fixes #1317

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -474,7 +474,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 * @see WP_Styles::_css_href()
 	 *
 	 * @param string   $url The file URL.
-	 * @param string[] $allowed_extensions Allowed file extensions.
+	 * @param string[] $allowed_extensions Allowed file extensions for local files.
 	 * @return string|WP_Error Style's absolute validated filesystem path, or WP_Error when error.
 	 */
 	public function get_validated_url_file_path( $url, $allowed_extensions = array() ) {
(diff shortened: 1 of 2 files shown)


westonruter:
See also https://github.com/Automattic/nginx-http-concat/pull/41 which would bypasse VIP's http-concat for CSS when `is_amp_endpoint()`. Having this in place will improve the AMP integration yet further, as it would eliminate redundant and unnecessary work.

kienstra:

If it's alright, I'm moving this to 'Ready For Merging,' as I don't think it needs functional testing. Feel free to move it back.

PR #1375, opened 2018-08-29
Display when validation results are stale due to active theme/plugin changes

westonruter:
Given validation errors are caused by an active theme or plugin, once the offending theme or plugin is deactivated any validation results obtained while active are no longer fresh. If another theme or plugin has been activated, then different validation errors could be present. In both cases, the user needs to be informed that the validation results need to be refreshed by re-checking the URL.

This pull request adds the underlying building blocks needed for displaying staleness, however the UI should not be considered finalized:

image

image

### DIFF
--- a/includes/validation/class-amp-invalid-url-post-type.php
+++ b/includes/validation/class-amp-invalid-url-post-type.php
@@ -743,6 +743,9 @@ class AMP_Invalid_URL_Post_Type {
 			esc_url( self::get_recheck_url( $post ) ),
 			esc_html__( 'Re-check', 'amp' )
 		);
+		if ( self::get_post_staleness( $post ) ) {
+			$actions[ self::VALIDATE_ACTION ] = sprintf( '<em>%s</em>', $actions[ self::VALIDATE_ACTION ] );
+		}
 
 		return $actions;
 	}
@@ -1157,7 +1160,7 @@ class AMP_Invalid_URL_Post_Type {
 								esc_html_e( 'Different theme and plugins were active when these results were obtained. Please re-check.', 'amp' );
 							} elseif ( ! empty( $staleness['theme'] ) ) {
 								esc_html_e( 'A different theme was active when these results were obtained. Please re-check.', 'amp' );
-							} elseif ( ! empty( $staleness['plugin'] ) ) {
+							} elseif ( ! empty( $staleness['plugins'] ) ) {
 								esc_html_e( 'Different plugins were active when these results were obtained. Please re-check.', 'amp' );
 							}
 							echo '</p></div>';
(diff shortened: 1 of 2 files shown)


includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -1096,6 +1145,24 @@ public static function print_status_meta_box( $post ) {
...
+						$staleness = self::get_post_staleness( $post );
+						if ( ! empty( $staleness ) ) {
+							echo '<div class="notice notice-info notice-alt inline"><p>';
+							echo '<b>';
+							esc_html_e( 'Stale results', 'amp' );
+							echo '</b>';
+							echo '<br>';
+							if ( ! empty( $staleness['theme'] ) && ! empty( $staleness['plugins'] ) ) {
+								esc_html_e( 'Different theme and plugins were active when these results were obtained. Please re-check.', 'amp' );
+							} elseif ( ! empty( $staleness['theme'] ) ) {
```
hellofromtonya: @westonruter The `'theme'` key may not exist in the returned `$staleness` array.  We either need to initialize `$staleness` with this key back here on [line 479)[https://github.com/Automattic/amp-wp/pull/1375/files#diff-57518db0811ee918fee17d27459ed4ecR479] or check `isset()`.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -1096,6 +1145,24 @@ public static function print_status_meta_box( $post ) {
...
+								esc_html_e( 'A different theme was active when these results were obtained. Please re-check.', 'amp' );
+							} elseif ( ! empty( $staleness['plugin'] ) ) {
```
hellofromtonya: @westonruter Shouldn't this key be `'plugins'`?  And this conditional has the same concern of `isset()` as above.

includes/validation/class-amp-invalid-url-post-type.php
westonruter: Actually, `empty()` and `isset()` are both PHP language constructs which will not complain about undefined array indexes. So using `isset( $staleness['theme'] ) && (bool) $staleness['theme']` is functionally equivalent to `! empty( $staleness['theme'] )`.

The absence of the `theme` key is important for the return value to satisfy an empty condition, as `array( 'theme' => array() )` is not `empty` for the overall check above in:

https://github.com/Automattic/amp-wp/blob/54b1272d05060d70d3ab2ace696fd9f7b3e540ed/includes/validation/class-amp-invalid-url-post-type.php#L1149-L1150

hellofromtonya: @westonruter True. Valid point. Thank you.

includes/validation/class-amp-invalid-url-post-type.php
westonruter: Good catch. Fixed in 6b5cb8be.

kienstra:
**Request To Test**

Hi @csossi,
Could you please test this?

1. On the [staging site](https://paired-ampconfdemo.pantheonsite.io/wp-admin/plugins.php?plugin_status=all&paged=1&s), activate a plugin like 'AMP Invalid'
2. Expected: The [Invalid URLs page](https://paired-ampconfdemo.pantheonsite.io/wp-admin/edit.php?post_type=amp_invalid_url) has 'Stale results' text:

invalid-urls

3. Expected: On clicking a single URL as shown above, there should be a notice in the 'Status' box like:
stale-results-single

csossi:
Verified in QA

PR #1376, opened 2018-08-29
AMP Mode Notice for Index listings

### LINKED ISSUE #1366: AMP Mode Notice for Index listings
As a user, I should be presented with a notice that reveals what mode of AMP I’m using so I understand what each action does based on the current mode. 

Without this notice, a given user may think actions take effect on their site, but when auto-sanitize mode is turned on (either by force for "Native" mode or optionally for "Paired" mode), 

- [ ] AC1: Surface a given notice based on the mode (from the settings page) in each of the views where actions can occur.
- [ ] AC2: Propose copy (and perhaps a suggestion to turn on Auto Suppression) for Native, Paired (non-auto) and Paired (auto) modes. Also, in Paired Mode (non-auto) offer a dismissable notice to help the user turn this on to enable auto sanitization sitewide.

Notes:
- If a user is in Paired Mode with auto-sanitize or in Native AMP mode, each error is but a notice and does not affect the site.
- If a user is in Paired Mode without auto-sanitize, each action might affect which pages are served as AMP. A given user, by not taking action on errors, would prevent AMP from being served on those pages.
- If a user is in Classic Mode, they should know they won't get the index listings (nor the single listings) as the sanitizer does not run in this mode.

miina:
Fixes #1366.

Adds admin notice to `Invalid Pages` and `Validation Errors` index listings according to the AMP Mode and configuration.

The notices:
- Native mode:
`The site is using native AMP mode, the validation errors found are already automatically handled.`
- Paired mode with auto-sanitization:
`The site is using paired AMP mode with auto-sanitization turned on, the validation errors found are already automatically handled.`
- Paired mode without auto-sanitization:
`The site is using paired AMP mode without auto-sanitization, the validation errors found require action and influence which pages are shown in AMP. For automatically handling the errors turn on auto-sanitization from Validation Handling settings. (<--- link)`
- Classic mode:
`The site is using classic AMP mode, your theme templates are not used and the errors below are irrelevant.`
<--- Although the index pages are not visible when in classic mode, it is possible to see this notice when turning on Classic mode in one tab and refreshing an index page in the other tab, thus the notice.


includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -894,6 +894,34 @@ public static function print_admin_notice() {
 				esc_html__( 'Dismiss this notice.', 'amp' )
 			);
 		}
+
+		if ( 'post' !== get_current_screen()->base ) {
+			// Display admin notice according to the AMP mode.
+			$theme_support     = AMP_Options_Manager::get_option( 'theme_support' );
+			$auto_sanitization = AMP_Options_Manager::get_option( 'force_sanitization' );
+
+			if ( 'native' === $theme_support ) {
```
kienstra: It might be more accurate to use `amp_is_canonical()` to detect Native AMP. That will also detect when a theme registers support for AMP via `add_theme_support( 'amp' )`.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -894,6 +894,34 @@ public static function print_admin_notice() {
...
+				$message = __( 'The site is using native AMP mode, the validation errors found are already automatically handled.', 'amp' );
+			} elseif ( 'paired' === $theme_support && $auto_sanitization ) {
```
kienstra: Likewise, it might be more accurate to use `AMP_Theme_Support::is_paired_available()` to detect Paired Mode. That will detect when a theme registers Paired Mode, as well as when a user selects it via 'AMP Settings' > 'Template Mode.'

westonruter: Actually `AMP_Theme_Support::is_paired_available()` won't work because it is specifically intended to be called on the frontend in the context of a given `WP_Query`. I think that checking for `AMP_Options_Manager::get_option( 'theme_support' )` is correct.

westonruter: On second thought, per the comment above and whether theme support was added via `add_theme_support()`... 

```php
if ( amp_is_canonical() ) {
	$template_mode = 'native';
} elseif ( current_theme_supports( 'amp' ) ) {
	$template_mode = 'paired';
} else {
	$template_mode = 'classic';
}
```

PR #1380, opened 2018-08-29
[amp-stories sub-task] Add background color to amp-story-page block

miina:
This is a sub-PR of #1215 and should be merged to that PR.

The goal of this PR is to add background color setting for the `amp-story-page` block. The default color is white.

screen shot 2018-08-29 at 6 05 48 pm

### DIFF
--- a/blocks/amp-story/amp-story-page.js
+++ b/blocks/amp-story/amp-story-page.js
@@ -113,7 +113,7 @@ export default registerBlockType(
 			return [
 				<InspectorControls key='controls'>
 					<PanelColorSettings
-						title={ __( 'Page Background Color Settings', 'amp' ) }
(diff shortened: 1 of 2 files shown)


blocks/amp-story/amp-story-page.js
```diff
@@ -62,6 +64,10 @@ export default registerBlockType(
 				source: 'attribute',
 				selector: 'amp-story-page',
 				attribute: 'id'
+			},
+			backgroundColor: {
+				type: 'string',
+				default: '#ffffff'
```
westonruter: It would be nice if Gutenberg had a `selector` for `style`.

westonruter: Now proposed in https://github.com/WordPress/gutenberg/issues/9439

Issue #1298, opened 2018-07-31
Convert block validation script to a module, and bundle it with block toggle module


kienstra:
AC1. Convert [amp-block-validation.js](https://github.com/Automattic/amp-wp/blob/21ebbc69289ec46a6752df39d2170a625c70faf7/assets/js/amp-block-validation.js) to an [ES6 module](http://exploringjs.com/es6/ch_modules.html)
AC2. Create a new file that will `import` both this module and [amp-block-editor-toggle.js](https://github.com/Automattic/amp-wp/pull/1275/files#diff-f10def20383407dba28e5820c57ba9b1R1) (though that might have since been renamed). This can be similar to [blocks/index.js](https://github.com/Automattic/amp-wp/blob/21ebbc69289ec46a6752df39d2170a625c70faf7/blocks/index.js).
AC3. Add that new file with the `import` statements to [webpack.config.js](https://github.com/Automattic/amp-wp/blob/21ebbc69289ec46a6752df39d2170a625c70faf7/webpack.config.js#L7), so it will be compiled with `npm run dev`
AC4. Ensure that the compiled files have sourcemaps.

This should probably wait for the work on `amp-block-validation.js` in #1293

kienstra:
Hi @postphotos,
If it's alright, I opened this issue with basic ACs. Feel free to edit this if you'd like.

This arose from the work on [https://github.com/Automattic/amp-wp/pull/1275#discussion_r206317004](https://github.com/Automattic/amp-wp/pull/1275#discussion_r206317004).

hellofromtonya:
@kienstra PR #1293 is now merged.  You have a green light to convert the `amp-block-validation.js` script into an E6 module.

kienstra:
Thanks, @hellofromtonya! I'm not sure if I'll work on this issue soon, though.

kienstra:
**Adding Another Script To The Requirements**

Per [this conversation](https://github.com/Automattic/amp-wp/pull/1418#discussion_r218880084), please also convert [amp-invalid-url-post-edit-screen.js](https://github.com/Automattic/amp-wp/blob/bdf411e7e326dfb3069eb8578aee11ab561bed29/assets/js/amp-invalid-url-post-edit-screen.js) to an ES6 module.

westonruter:
I'm going to close this because this will come automatically once we refactor the admin screens to use JS-driven interfaces as opposed to the PHP-driven ones we've largely tried to re-use from what WordPress provides.

PR #1385, opened 2018-08-30
[amp-stories sub-task] Add front-end animation.

miina:
This is a sub-PR of #1215 and should be merged to that PR. (The name of the branch is misleading, please ignore that.)

This PR adds three controls to the core blocks allowed within the `amp-story-page` and also to AMP Story Layer blocks (CTA and Grid).
AMP Story Grid Layer block:
screen shot 2018-08-30 at 9 42 51 pm
Core blocks:
screen shot 2018-08-30 at 9 43 34 pm

Note that currently the maximum delay and maximum duration are set to 5000ms.


miina:
@westonruter This should be ready for review.

assets/js/amp-story-editor-blocks.js
```diff
@@ -131,14 +219,26 @@ var ampStoryEditorBlocks = ( function() { // eslint-disable-line no-unused-vars
...
-		// Add the "thirds" template position option.
+		// Add the "thirds" template position option and animation settings.
 		if ( -1 !== component.data.allowedBlocks.indexOf( name ) ) {
 			if ( ! settings.attributes ) {
 				settings.attributes = {};
 			}
 			settings.attributes.ampStoryPosition = {
 				type: 'string'
 			};
+			// @todo We could map all the blocks to their tag and use attribute as source instead.
```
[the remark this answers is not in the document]
westonruter: Or maybe it's rather because we don't know the element name to select?

blocks/amp-story/amp-story-cta-layer.js
```diff
@@ -99,9 +121,16 @@ export default registerBlockType(
 						<Notice status="error" isDismissible={ false }>{ __( 'Multiple CTA Layers are not allowed. Please remove all but one.', 'amp' ) }</Notice>
 					);
 				}
-				return (
+				return [
+					<InspectorControls key='controls'>
+						<PanelBody key='animation' title={ __( 'CTA Layer Animation' ) }>
```
westonruter: Missing `amp` text domain.

blocks/amp-story/amp-story-grid-layer.js
```diff
@@ -100,6 +117,11 @@ export default registerBlockType(
 						] }
 						onChange={ value => ( setAttributes( { template: value } ) ) }
 					/>
+					<PanelBody key='animation' title={ __( 'Grid Layer Animation' ) }>
```
westonruter: Missing `amp` text domain.

blocks/utils.js
```diff
@@ -82,3 +83,115 @@ export function getLayoutControls( props, ampLayoutOptions ) {
...
+				{
+					value: 'zoom-out',
+					label: __( 'Zoom Out', 'amp' )
+				}
+			] }
+			onChange={ value => ( setAttributes( { animationType: value } ) ) }
+		/>,
+		<RangeControl
+			key='duration'
+			label={ __( 'Duration (ms)' ) }
```
westonruter: Missing `amp` text domain.

blocks/utils.js
```diff
@@ -82,3 +83,115 @@ export function getLayoutControls( props, ampLayoutOptions ) {
...
+			value={ attributes.animationDuration }
+			onChange={ ( value ) => setAttributes( { animationDuration: value } ) }
+			min='0'
+			max='5000'
+		/>,
+		<RangeControl
+			key='delay'
+			label={ __( 'Delay (ms)' ) }
```
westonruter: Missing `amp` text domain.

assets/js/amp-story-editor-blocks.js
```diff
@@ -39,6 +40,84 @@ var ampStoryEditorBlocks = ( function() { // eslint-disable-line no-unused-vars
 					value: 'lower-third',
 					label: __( 'Lower Third', 'amp' )
 				}
+			],
+			ampAnimationTypeOptions: [
```
westonruter: This list is duplicated with the array in `getAmpStoryAnimationControls`, is it not? I think there should be a common `getAnimationTypeOptions` function in the `utils` module to eliminate this duplication. Even if the list is not exactly the same, the function could still return the full list and then you could `filter` out the ones that aren't allowed.

includes/sanitizers/class-amp-allowed-tags-generated.php
```diff
@@ -13323,6 +13323,9 @@ class AMP_Allowed_Tags_Generated {
 		'amp-fx' => array(
 			'value_regex_casei' => '(fade-in|fade-in-scroll|fly-in-bottom|fly-in-left|fly-in-right|fly-in-top|parallax)(\\s|fade-in|fade-in-scroll|fly-in-bottom|fly-in-left|fly-in-right|fly-in-top|parallax)*',
 		),
+		'animate-in' => array(),
+		'animate-in-delay' => array(),
+		'animate-in-duration' => array(),
```
westonruter: This manual editing of the generated tags shouldn't be required. These should be automatically populated from `bin/amphtml-update.sh`. If not, then we have a bug to fix in `bin/amphtml-update.py`.

westonruter: I'm woking on this in https://github.com/Automattic/amp-wp/pull/1386. I can see there is indeed a problem with how the script handles `$REFERENCE_POINTS` in the spec.

westonruter: In the mean time, we can ignore this and merge this as-is.

assets/js/amp-story-editor-blocks.js
miina: Yes, it is duplicated, however, the `utils.js` gets compiled, not sure what would be the best way to use the module like this within `amp-story-editor-blocks.js`, thoughts?

Created a separate helpers file for AMP Stories specifically within 4923569 since the `utils.js` always gets enqueued, however, `getAmpStoryAnimationControls` is needed for AMP Stories post type only, thought that then maybe we could add an inline script for defining the `getAnimationTypeOptions` globally to avoid duplication but not sure if that would be a good way to go.

assets/js/amp-story-editor-blocks.js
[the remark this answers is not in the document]
miina: Changed these settings to `source: attribute` by using `parseInt()` within 4923569. Exception is for `core/list` since it could be either `ul` or `ol`.

assets/js/amp-story-editor-blocks.js
westonruter: Oh, right. So this is actually dependent on #1298 which will convert `amp-story-editor-blocks.js` into a module. So once that happens then this problem would go away, right?

PR #1386, opened 2018-08-30
Add support for reference points and update spec to file revision 720 (v1534879991178)

westonruter:
* Add support for handling reference points in validator spec, including support for attributes used by `amp-next-page`, `amp-story`, `amp-lightbox-gallery`, and `amp-live-list`. This removes hard-coded list of whitelisted attribute pattern which were erroneously being allowed on every element, not just the descendants of elements with reference points.
* Add special case for handling reference points for `amp-selector` since its attributes can be added to any descendant not just direct children.
* Add components for `amp-image-slider` and `amp-viqeo-player`.
* Add `when-ended=continue` to `amp-date-countdown`.
* Add `[open]` to `amp-lightbox`.
* Add `disable-double-tap` to `amp-pan-zoom`.

### DIFF
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -240,7 +240,7 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 	}
 
 	/**
-	 * Sanitize the <video> elements from the HTML contained in this instance's DOMDocument.
+	 * Sanitize the elements from the HTML contained in this instance's DOMDocument.
 	 *
 	 * @since 0.5
 	 */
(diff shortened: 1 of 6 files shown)


includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -276,6 +276,79 @@ public function sanitize() {
...
+						 * sanitizer does not yet handle the mandatory_oneof constraint.
+						 */
+						unset( $reference_point_spec_attr['mandatory'] );
+
+						$rule_spec[ AMP_Rule_Spec::ATTR_SPEC_LIST ][ $attr_name ] = $reference_point_spec_attr;
+					}
+				}
+			}
+		}
+		return $rule_spec;
```
westonruter: Most of this code is refactored into a separate method from below. See https://github.com/Automattic/amp-wp/pull/1386/commits/002bda35034eca01febaf681906ed9624059bf31 for the actual changes introduced.

PR #1391, opened 2018-09-03
[amp-stories sub-task] Add icons to layer selector.

miina:
Adds icons to layer selector component:
screen shot 2018-09-03 at 2 23 04 pm

### DIFF
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -150,24 +150,66 @@ div[data-amp-position="lower-third"] {
 	display: none;
 }
 
-.editor-selectors {
+.edit-post-visual-editor ul.editor-selectors {
 	position: absolute;
-	right: -150px;
+	right: -155px;
 	bottom: 0;
-	width: 100px;
+	width: 120px;
 	z-index: 80;
+	list-style-type: none;
+	padding: 0;
+	margin: 0;
 }
 
 .editor-selectors .component-editor__selector {
-	margin-bottom: 2px;
+	background: no-repeat left top;
+	padding: 2px 0 0 30px;
+	height: 40px;
+	margin-left: 5px;
+}
+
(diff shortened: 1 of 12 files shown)


assets/css/amp-editor-story-blocks.css
```diff
@@ -150,24 +150,66 @@ div[data-amp-position="lower-third"] {
...
+	background: no-repeat left top;
+	padding: 2px 0 0 30px;
+	height: 40px;
+	margin-left: 5px;
 }
 
-.editor-selectors .component-editor__selector button {
-	cursor: pointer;
+.editor-selectors .component-editor__selector.template-fill {
+	background-image: url( '../images/grid-fill-inactive.svg' );
```
westonruter: In the Gutenberg Inserter, the SVG images are actually inline in the document as opposed to being included from a separate file as a background-image in CSS. I don't know if necessarily one way is better than the other in this case, but I wanted to mention that.

blocks/amp-story/block-selector.js
```diff
@@ -34,7 +34,8 @@ class BlockSelector extends Component {
 		let hasCtaLayer = false;
 
 		window.lodash.forEachRight( rootBlock.innerBlocks, function( block, index ) {
-			let className = 'component-editor__selector';
+			let template = ! block.attributes.template ? 'vertical' : block.attributes.template;
```
westonruter: This could be simplified to:

```js
const template = block.attributes.template || 'vertical';
```

Not an important change, however.

blocks/amp-story/block-selector.js
```diff
@@ -34,7 +34,8 @@ class BlockSelector extends Component {
...
+			let className = 'component-editor__selector template-' + template;
```
westonruter: Or even the above could just be put right here:

```js
let className = 'component-editor__selector template-' + ( block.attributes.template || 'vertical' );
```

PR #1393, opened 2018-09-03
Add missing amp-block-validation dependency.

felixarntz:
Currently, in Gutenberg I receive an error in the console because the `amp-block-validation.js` file tries to access the `wpAmpEditor` global. That global is set via `wp_localize_script()` attached to the `amp-block-editor-toggle-compiled.js` file. As the latter is not a dependency of `amp-block-validation`, it may load later, causing the error to be thrown.

By adding `amp-block-editor-toggle-compiled.js`, this issue can be fixed. As a side observation, we might also want to replace `wp_localize_script()` with a call to `wp_add_inline_script()`, but that isn't part of the bugfix, so I didn't include it here.


westonruter:
Looks like unit test needs to be updated for this change. Also, feel free to include the switch to `wp_add_inline_script()`.

PR #1396, opened 2018-09-04
Add embed support for Twitter timelines via new amp-twitter attributes

felixarntz:
`amp-twitter` now supports arbitrary data attributes for Twitter timelines, based on a required `data-timeline-source-type` attribute. All other arguments to pass along to the Twitter script must be provided as individual data attributes prefixed with "timeline-". As an example, if you set `data-timeline-source-type` to "profile", you also need to pass a `screenName` argument, which would need to happen via `data-timeline-screen-name`.

This PR adds support for Twitter timelines, by automatically parsing embed URLs into appropriate `amp-twitter` elements. The code supports `profile`, `likes` and `list` source types.

For reference:
* https://www.ampproject.org/docs/reference/components/amp-twitter
* https://developer.twitter.com/en/docs/twitter-for-websites/javascript-api/guides/scripting-factory-functions

### DIFF
--- a/includes/embeds/class-amp-twitter-embed.php
+++ b/includes/embeds/class-amp-twitter-embed.php
@@ -11,7 +11,14 @@
  *  Much of this class is borrowed from Jetpack embeds
  */
 class AMP_Twitter_Embed_Handler extends AMP_Base_Embed_Handler {
-	const URL_PATTERN = '#http(s|):\/\/twitter\.com(\/\#\!\/|\/)([a-zA-Z0-9_]{1,20})\/status(es)*\/(\d+)#i';
+
+	/**
+	 * URL pattern for a Tweet URL.
+	 *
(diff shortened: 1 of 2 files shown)


felixarntz:
When testing the functionality, parsing into the correct tag works, however on the frontend the corresponding iframes only render as blank. I'd need to investigate why this happens, but maybe someone with more AMP experience has a quick idea about it.

westonruter:
@felixarntz Is your testing site in HTTPS? AMP requires HTTPS for `amp-iframe` to render:

> An amp-iframe must only request resources via HTTPS, from a data-URI, or via the srcdoc attribute.

https://www.ampproject.org/docs/reference/components/amp-iframe

felixarntz:
@westonruter Ah yes, that was it. I should really setup SSL for my VVV sites.

Issue #1361, opened 2018-08-27
Implement improved error listing view


postphotos:
As an developer using the AMP for WordPress plugin when viewing AMP errors by type, I seek improved usability though color highlighting, improved wording and new visual styles.

Efforts to implement the dropdown facet filters should be discussed here: #1363

Based on "errors by type" screen from ~[here](https://sketch.cloud/s/PoWy8/all/page-1/amp-validation-errors) [here](https://sketch.cloud/s/Vow17/all/page-1/amp-validation-errors)._
(From Ryan) [This](https://sketch.cloud/s/VoDKr/all/page-1/amp-validation-errors) is the updated design, right?~ FINAL DESIGN: [ [Page 1](https://sketch.cloud/s/xQRzx/all/page-1/amp-validation-errors-copy) ] [ [Page 2](https://sketch.cloud/s/xQRzx/all/page-1/amp-validation-errors) ]

As an developer using the AMP for WordPress plugin when viewing AMP errors by URL, I seek improved usability though color highlighting, improved wording and new visual styles.

_Based on "Errors by URL" screens from [here](https://sketch.cloud/s/PoWy8/all/page-1/amp-validation-urls)._

- [ ] AC1: Implement new design elements.
  - [ ] New header labels for each column: Error Inventory, Status, Details, Error Type, Last Seen, Details, Last Seen and Found URLs
  - [ ] Implement a type column using the new language here, referencing the mocks and work IP elsewhere (see #1360)
   - [ ]  Implement new icons and language for each rendered error (which respects Stale states): 
     - [ ] ~`Suppressed` (formerly~ "Accepted"
     - [ ] ~`Identified` (formerly~ "New"
     - [ ] ~`To Fix Later` (formerly~ "Rejected" 
     - [ ] No counts should appear next to errors, as the plugin already does.
   - [ ] Implement new language for each rendered error action: 
     - [ ] ~`Suppressed` (formerly~ "Accepted"
     - [ ] ~`Identified` (formerly~ "New"
     - [ ] ~`To Fix Later` (formerly~ "Rejected" 
   - [ ] Implement a tooltip helper for the "Status" column that allows us to describe what error states mean in case a user is stuck.
   - [ ] Change "Created Date" to more accurate language: "Last Seen"
- [ ] I should also still be able to take bulk actions on these errors.
- [ ] Search, error count and pagination (regular feature of these index listings of errors) should remain.
[...]

PR #1397, opened 2018-09-04
Improve URL Listing view

### LINKED ISSUE #1362: Implement improved URL listing view
As an developer using the AMP for WordPress plugin when viewing AMP errors by URL, I seek improved usability though color highlighting, improved wording and new visual styles.

_Based on "Errors by URL" screens from ~[here](https://sketch.cloud/s/PoWy8/all/page-1/amp-validation-urls) [here](https://sketch.cloud/s/Vow17/all/page-1/amp-validation-urls).
(From Ryan) [This](https://sketch.cloud/s/VoDKr/all/page-1/amp-validation-urls) is the updated design, right?~_ Final design: [ [Page 1 ](https://sketch.cloud/s/xQRzx/all/page-1/amp-validation-urls) ] [ [Page 2](https://sketch.cloud/s/xQRzx/all/page-1/amp-validation-urls-stale) ]

- [ ] AC1: Implement new design elements.
  - [ ] New header labels for each column: URL, Status, Invalid, Sources, Last Checked
  - [ ] For error "Sources," incorporate the updated design to better organize sources that can be expanded or collapsed. ~from #1006 (color syntax to separate source of plugin, theme, color) that should consider a11y~
  - [ ] Surface the validation of whether a given URL is delivering AMP in the index view by conditionally showing the "AMP Logo" (as it's done elsewhere) when errors are suppressed (as shown in the mocks.)
   - [ ] Implement new icons and language for each rendered error (which respects `Stale` states):
     - [ ] ~`Suppressed` (formerly~ "Accepted"
     - [ ] ~`Identified` (formerly~ "New"
     - [ ] ~`To Fix Later` (formerly~ "Rejected" 
     - [ ] Counts should appear next to each error, as the plugin already does.
     - [ ] "New" errors should reflect auto-santization. If auto is on (in Native or in Paired), the icon should be a flag. If auto is off (in Paired) the icon should be a checkmark and red. See [here]([url](https://docs.google.com/a/xwp.co/document/d/1gpv9PpdRT_fAvp76JOberevmkZRmtpHHyBFT40vgw7I/edit?disco=uiAAAAB0tmy7Y)).
   - [ ] Implement new language for each rendered error action: 
     - [ ] ~`Suppress` (formerly~ "Accept"
     - [ ] ~`To Fix Later` (formerly~ "Reject" 
   - [ ] Implement a tooltip helper for the "Status" column that allows us to describe what error states mean in case a user is stuck.
   - [ ] Combine the `Found Errors` and `Found Attributes` into a single column as shown in the mocks. Also include CSS errors so that all `Invalid` errors are displayed in this summary view.
   - [ ] Change "Created Date" to more accurate language: "Last Checked"
- [ ] Search, error count and pagination (regular feature of these index listings of errors) should remain.

- [ ] AC2: I should be able to view the other index listing (`Errors by Type`) screen by clicking on the button next to the page title in this listing. The wording of the page title and sidebar for this feature should match the mocks.

~Efforts to implement the backlog item of dropdown facet filters should be discussed here: #1363

_Proposed/Backlog:_
- [x] AC3: I should be able to filter them by error status and type. (see #1373)~

miina:
Fixes #1362.

Updates the Error by URLs view according to [this sketch](https://sketch.cloud/s/Vow17/all/page-1/amp-validation-errors-for-amp-test-embeds-copy).


includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -439,10 +446,12 @@ public static function summarize_validation_errors( $validation_errors ) {
 			}
 
 			if ( ! empty( $validation_error['sources'] ) ) {
-				$source = array_shift( $validation_error['sources'] );
+				$sources = array_shift( $validation_error['sources'] );
 
-				if ( isset( $source['type'], $source['name'] ) ) {
-					$invalid_sources[ $source['type'] ][] = $source['name'];
+				foreach ( $sources['sources'] as $source ) {
```
miina: ^ Changed this since the sources were not properly displaying in the table -- the `type` and `name` were never set since the actual structure of `$validation_error` seems to be like this:
screen shot 2018-09-04 at 5 21 41 pm

Meaning that the `type` and `name` would be set in `$validation_error['sources']['sources'][0]`, not in `$validation_error['sources']`.

However, this change broke the tests since the input data of test is assuming that a validation error looks like this instead:
screen shot 2018-09-04 at 5 43 21 pm

Is the structure of validation error incorrect for some reason and it should be as assumed in the tests? Maybe the fix should be removing the second nested `sources` instead, or is it like this for some reason? Thoughts?

cc @kienstra @westonruter

miina:
@kienstra Created this WIP PR as a start for resolving #1362, however, I'm going to offline from Wed-Fri (5-7th of September) so if you'd like to pick this up meanwhile or implement together with #1361 (there are a few matching tasks) then feel free to do so.

includes/validation/class-amp-validation-error-taxonomy.php
kienstra: Hi @miina,
That's interesting. I couldn't reproduce that issue.

It looks like there aren't double-nested `'sources'` arrays, like in your first screenshot when I run the `develop` branch:

name-type-local

Maybe the validation error data model changed.

It might help to delete all validation error posts and terms, and regenerate them (thanks to Weston's code, though I might not have copied it right):

```bash
wp post delete $(wp post list --post_type='amp_invalid_url' --format=ids)
wp term list amp_validation_error --field=term_id | xargs wp term delete amp_validation_error
```
You might already have some logic to output validation errors, but here's a [simple plugin](https://gist.github.com/kienstra/c204c13a0974d949ed741f37f8f9e14f) I've used (also partially copied from Weston)

And the WP-CLI script can then regenerate errors:
```bash 
cd /path/in/vvv/wp-content/plugins/amp
wp amp validate-site limit=5
```

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -162,6 +164,19 @@ public static function add_admin_hooks() {
 		} );
 	}
 
+	/**
+	 * Enqueue style.
+	 */
+	public static function enqueue_admin_assets() {
+		// Styles.
+		wp_enqueue_style(
```
westonruter: This should only be enqueued for the post list screen for this specific post type

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -162,6 +164,19 @@ public static function add_admin_hooks() {
...
+			'amp-admin-tables',
+			amp_get_asset_url( 'css/admin-tables.css' ),
```
westonruter: Was this CSS file not committed?

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -162,6 +164,19 @@ public static function add_admin_hooks() {
 		} );
 	}
 
+	/**
+	 * Enqueue style.
+	 */
+	public static function enqueue_admin_assets() {
```
westonruter: See also https://github.com/Automattic/amp-wp/pull/1414 where an `enqueue_edit_post_screen_scripts` method is introduced.

kienstra:
**Request For Review**

Hi @westonruter,
Could you please review this, now that the Travis build passes?

Thanks!

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -583,92 +626,78 @@ public static function output_custom_column( $column_name, $post_id ) {
...
+							$sources_container_classes .= ' collapsed';
+						}
+						$output[] = '<div class="' . $sources_container_classes . '">';
+						$output[] = implode( '<br/>', array_unique( $sources['core'] ) );
+						$output[] = '</div>';
+						$output[] = '</div>';
+					}
+					if ( isset( $sources['theme'] ) ) {
+						$output[] = '<div class="source">';
+						$output[] = sprintf( '<span class="dashicons dashicons-admin-appearance"></span><strong>%s</strong>', esc_html( $sources['theme']['name'] ) );
```
westonruter: I'm getting a PHP notice here:

image

The name isn't stored in the validation error (no should it be). The `$sources['theme']` variable contains an array like `[ 'twentyseventeen' ]`. So this should be looping over each `$theme` and get `wp_get_theme( $theme )->get('Name')`. If the theme does not exist, then the slug should be shown instead (and obviously the result is stale in this case).

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -583,92 +626,78 @@ public static function output_custom_column( $column_name, $post_id ) {
...
+						$count                     = count( array_unique( $sources['plugin'] ) );
+						$sources_container_classes = 'sources-container sources-plugins';
+						if ( 1 === $count ) {
+							$output[] = sprintf( '<span class="dashicons dashicons-admin-plugins"></span></span><strong>%s</strong><br/>', esc_html__( 'Plugin', 'amp' ) );
+						} else {
+							$output[]                   = sprintf( '<span class="dashicons dashicons-admin-plugins"></span><strong>%s (%d)</strong><span class="dashicons toggle-sources dashicons-arrow-down"></span>', esc_html__( 'Plugins', 'amp' ), $count );
+							$sources_container_classes .= ' collapsed';
 						}
+						$output[] = '<div class="' . $sources_container_classes . '">';
+						$output[] = implode( '<br/>', array_unique( $sources['plugin'] ) );
```
westonruter: This should obtain the actual names of the plugins using `get_plugin_data()` and getting the `Name` returned. If the plugin doesn't exist then show the slug (and this result is obviously stale).

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -169,6 +170,37 @@ public static function add_admin_hooks() {
 		} );
 	}
 
+	/**
+	 * Enqueue style.
+	 */
+	public static function enqueue_admin_assets() {
```
westonruter: Can this name be made more specific? There is `enqueue_edit_post_screen_scripts`. So should it be `enqueue_post_list_screen_scripts`?

assets/js/amp-admin-tables.js
```diff
@@ -0,0 +1,46 @@
...
+		allSourcesHideShow: function() {
+			$( '.double-arrow' ).on( 'click', function() {
+				$( '.double-arrow' ).find( '.dashicons' ).toggleClass( 'dashicons-arrow-down dashicons-arrow-up' );
+				$( 'span.dashicons.toggle-sources' ).toggleClass( 'dashicons-arrow-down dashicons-arrow-up' );
+				$( '.sources-container' ).toggleClass( 'collapsed' );
+			} );
+		},
+
+		addViewErrorsByTypeLinkButton: function() {
+			$( '.wp-heading-inline' ).after( '<a href="' + ampAdminTables.errorsByTypeLink + '" class="page-title-action">View errors by <strong>Type</strong></a>' );
```
westonruter: This shouldn't be using string concatenation for constructing HTML. It would be preferable to do `var link = $( '<a>...</a>' );` And then do `link.prop('href', ampAdminTables.errorsByTypeLink).`

westonruter: The text here is not translatable. I suggest doing this: at the admin footer action whenever you are printing scripts, output right to the page the entire link tag with `esc_url()` and `esc_html_e()` and such. Give that link an ID and include a `hidden` attribute. Then in this `addViewErrorsByTypeLinkButton` function, just move the element in the DOM and remove the `hidden` attribute.

assets/js/amp-admin-tables.js
```diff
@@ -0,0 +1,46 @@
...
+		},
+
+		boldURLInPageTitle: function() {
```
westonruter: We can remove this if we go with “Invalid URL” per Slack convo.

assets/js/amp-admin-tables.js
```diff
@@ -0,0 +1,46 @@
+/* global jQuery, ampAdminTables */
+( function( $ ) {
+	'use strict';
+
+	var adminTables = {
```
westonruter: Some of the code here might be duplicated with https://github.com/Automattic/amp-wp/blob/develop/assets/src/amp-validation-error-detail-toggle.js

Can we unify them?

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -85,8 +85,8 @@ public static function register() {
 			self::POST_TYPE_SLUG,
 			array(
 				'labels'       => array(
-					'name'               => _x( 'Invalid AMP Pages (URLs)', 'post type general name', 'amp' ),
-					'menu_name'          => __( 'Invalid Pages', 'amp' ),
+					'name'               => _x( 'Errors by URL', 'post type general name', 'amp' ),
+					'menu_name'          => __( 'Errors by URL', 'amp' ),
```
westonruter: Let's change these to “Invalid URLs” throughout. As part of this, we can remove code related to bolding the “URLs” string.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -286,25 +318,34 @@ public static function display_invalid_url_validation_error_counts_summary( $pos
...
+				$icon = 'flag';
+			} else {
+				$icon = 'yes';
+			}
+			$result[] = sprintf(
 				/* translators: %s is count */
-				__( '&#x2753; New: %s', 'amp' ),
+				__( '<span class="dashicons dashicons-%1$s identified"></span><span class="error-status identified">%2$s: %3$s</span>', 'amp' ),
+				$icon,
+				esc_html__( 'Identified', 'amp' ),
```
westonruter: This can move back to “New”

tests/validation/test-class-amp-invalid-url-post-type.php
```diff
@@ -219,7 +219,7 @@ public function test_get_invalid_url_validation_errors() {
 		ob_start();
 		AMP_Invalid_URL_Post_Type::display_invalid_url_validation_error_counts_summary( $invalid_url_post_id );
 		$summary = ob_get_clean();
-		$this->assertContains( 'New: 1', $summary );
+		$this->assertContains( 'Identified: 1', $summary );
```
westonruter: Move back to “New”

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -286,25 +318,34 @@ public static function display_invalid_url_validation_error_counts_summary( $pos
 
 		$result = array();
 		if ( $counts['new'] ) {
-			$result[] = esc_html( sprintf(
+			if ( current_theme_supports( 'amp' ) && AMP_Validation_Manager::is_sanitization_forcibly_accepted() ) {
```
westonruter: We can probably remove `current_theme_supports( 'amp' ) &&`

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -542,18 +583,20 @@ public static function add_post_columns( $columns ) {
 		$columns = array_merge(
 			$columns,
 			array(
-				'error_status' => esc_html__( 'Error Status', 'amp' ),
-				AMP_Validation_Error_Taxonomy::REMOVED_ELEMENTS => esc_html__( 'Removed Elements', 'amp' ),
-				AMP_Validation_Error_Taxonomy::REMOVED_ATTRIBUTES => esc_html__( 'Removed Attributes', 'amp' ),
-				AMP_Validation_Error_Taxonomy::SOURCES_INVALID_OUTPUT => esc_html__( 'Incompatible Sources', 'amp' ),
+				'error_status' => sprintf( '%s<span class="dashicons dashicons-editor-help"></span>', esc_html__( 'Status', 'amp' ) ),  // @todo Create actual tooltip.
```
westonruter: The `'error_status'` here can be replaced with the `AMP_Validation_Error_Taxonomy::ERROR_STATUS` constant I believe.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -286,25 +318,34 @@ public static function display_invalid_url_validation_error_counts_summary( $pos
...
 				number_format_i18n( $counts['accepted'] )
-			) );
+			);
 		}
 		if ( $counts['rejected'] ) {
-			$result[] = esc_html( sprintf(
+			$result[] = sprintf(
 				/* translators: %s is count */
-				__( '&#x274C; Rejected: %s', 'amp' ),
+				__( '<span class="dashicons dashicons-warning rejected"></span><span class="error-status rejected">%1$s: %2$s</span>', 'amp' ),
```
westonruter: I don't think the `span` shouldn't be included in the translation string.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -286,25 +318,34 @@ public static function display_invalid_url_validation_error_counts_summary( $pos
...
-			) );
+			);
 		}
 		if ( $counts['accepted'] ) {
-			$result[] = esc_html( sprintf(
-				/* translators: %s is count */
-				__( '&#x2705; Accepted: %s', 'amp' ),
+			$result[] = sprintf(
+				/* translators: 1. Title, 2. %s is count */
+				__( '<span class="amp-logo-icon"></span><span class="error-status accepted">%1$s: %2$s</span>', 'amp' ),
```
westonruter: I don't think the `span` shouldn't be included in the translation string.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -286,25 +318,34 @@ public static function display_invalid_url_validation_error_counts_summary( $pos
...
-			$result[] = esc_html( sprintf(
+			if ( current_theme_supports( 'amp' ) && AMP_Validation_Manager::is_sanitization_forcibly_accepted() ) {
+				$icon = 'flag';
+			} else {
+				$icon = 'yes';
+			}
+			$result[] = sprintf(
 				/* translators: %s is count */
-				__( '&#x2753; New: %s', 'amp' ),
+				__( '<span class="dashicons dashicons-%1$s identified"></span><span class="error-status identified">%2$s: %3$s</span>', 'amp' ),
```
westonruter: I don't think the `span` shouldn't be included in the translation string.

assets/js/amp-admin-tables.js
jacobschweitzer: @westonruter  I started working on this but realized a dependency of the script you mentioned is `wp-dom-ready` which is a part of Gutenberg. So this script doesn't load for me since on my test site Gutenberg is not activated. Are we assuming that everyone will have Gutenberg as well or why is this a dependency? Or did I miss something in the build process?

kienstra: Hi @westonruter and @jacobschweitzer,
Do you think we should add `dom-ready` to [package.json](https://github.com/Automattic/amp-wp/blob/e63a879c6047ae5a10fdd6d0ea028b83eb37a42f/package.json#L12), like how `i18n` is included there:

```json
  "@wordpress/dom-ready": "^2.0.0",
  "@wordpress/i18n": "^1.1.0",

```

Still, `npm run build` seemed to build [amp-validation-error-detail-toggle.js](https://github.com/Automattic/amp-wp/blob/21413ba755917c6d6888c98fcc2ac5608e118266/assets/src/amp-validation-error-detail-toggle.js) fine in my local, and that script ran without error on the 'AMP Validation Error Index' page.

jacobschweitzer: @kienstra I've added that line to `package.json` and ran `npm run build` / `npm run dev` but the script still won't load on the Error Index page. Am I missing something here? I did have to install webpack and wp-cli but I have those now so I'm not sure if there is another dependency I'm missing since there aren't any errors in the build process now.

jacobschweitzer:
@westonruter , @kienstra found that if you remove the `domReady` from the webpack config, remove it as a dependency of the script loading, and do `npm run dev` again we can get the [script](https://github.com/Automattic/amp-wp/blob/develop/assets/src/amp-validation-error-detail-toggle.js) to run successfully. It is included here: 56915dab8efae6b90db9894feb46756071ed1da2  but it is a workaround, so if you can come up with a fix or have another idea that'd be awesome. 

Otherwise these changes are ready for review. Thanks @westonruter !

assets/src/amp-validation-error-detail-toggle.js
```diff
@@ -54,7 +61,20 @@ function addToggleListener() {
 	} );
 }
 
+function addViewErrorsByTypeLinkButton() {
+	if ( 'undefined' === typeof errorIndexAnchor || 'undefined' === typeof errorIndexLink ) {
+		return;
+	}
+	const heading = document.querySelector( '.wp-heading-inline' );
+	const link = document.createElement( 'a' );
+	link.innerHTML = errorIndexAnchor;
```
westonruter: `innerText` should be used instead of `innerHTML` because the latter is susceptible to an injection attack, if `errorIndexAnchor` is coming from an untrusted source.

assets/src/amp-validation-error-detail-toggle.js
```diff
@@ -54,7 +61,20 @@ function addToggleListener() {
...
+	link.setAttribute( 'href', errorIndexLink );
+	link.setAttribute( 'class', 'page-title-action' );
+	heading.after( link );
+}
```
westonruter: Either this function or `AMP_Validation_Error_Taxonomy::render_link_to_errors_by_url()` should be refactored to be taking the same approach.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -585,92 +637,88 @@ public static function output_custom_column( $column_name, $post_id ) {
...
-							$sources[] = sprintf( '%s: <code>%s</code>', esc_html( $type ), esc_html( $name ) );
+					$sources = $error_summary[ AMP_Validation_Error_Taxonomy::SOURCES_INVALID_OUTPUT ];
+					$output  = array();
+
+					if ( isset( $sources['plugin'] ) ) {
+						$output[]     = '<details class="source">';
+						$plugin_names = array();
+						$plugin_slugs = array_unique( $sources['plugin'] );
+						foreach ( $plugin_slugs as $plugin_slug ) {
+							$plugin_data = get_plugin_data( WP_PLUGIN_DIR . '/' . $plugin_slug . '/' . $plugin_slug . '.php' );
```
westonruter: This won't work for plugins that don't have directories or which don't use the slug as the plugin's main PHP file.

image

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -585,92 +637,88 @@ public static function output_custom_column( $column_name, $post_id ) {
...
+						$output[] = '</div>';
+						$output[] = '</details>';
 					}
-					echo implode( ', ', $sources ); // WPCS: XSS ok.
+					if ( isset( $sources['theme'] ) ) {
+						$output[] = '<div class="source">';
+						$output[] = '<span class="dashicons dashicons-admin-appearance"></span>';
+						$themes   = array_unique( $sources['theme'] );
+						foreach ( $themes as $theme_slug ) {
+							$output[] = sprintf( '<strong>%s</strong><br/>', esc_html( $theme_slug ) );
```
westonruter: Still needing to get `wp_get_theme( $theme_slug )->get( 'Name' )` here.

Issue #1384, opened 2018-08-30
Default to standard mode when compatible theme/plugins active


westonruter:
~For brand new plugin activations, we should consider to default to the new Transitional mode.~ (As of v2.0, Transitional mode is the default when an AMP-compatible theme is active.) This would be instead of defaulting to the Reader mode, which is what the default would continue to be for existing users of the plugin when they upgrade. For new users, the Transitional mode can be much superior to the Reader mode because not only are same theme templates being used, but also any validation errors are actually reported to the user. ~In the Reader mode, validation errors are silently sanitized which is dangerous since it can mean ads and other critical site functionality is removed.~ (Not as of #4480.)

We should also consider defaulting to the new Transitional mode if the user has an active theme which is known to be compatible with AMP (e.g. the core themes). We can also consider the same for compatible. plugins. We could also consider doing this by default if no `excessive_css` validation errors are raised, as this should mean that the paired mode templates will at least be stylistically the same as the canonical templates.

This would hopefully increase the adoption of the new AMP theme support.

See also #1288.

westonruter:
We could also show a deprecated notice to the users of the Classic mode to get them to switch to the new Paired mode. However, we should consider only doing this if the active theme does _not_ contain an `amp` directory, since this directory entails that they have a customized theme for the classic mode and thus probably they don't have a blue bar. 

Then in the future we can even remove Classic mode from being presented to users entirely. Similar to Links/Blogroll in WordPress, which are only available for grandfathered sites that had it enabled in the past, we can similarly hide the Classic option for new plugin activations in the future.

postphotos:
As we've talked about a bunch, the "blue bar" classic experience isn't that great, especially with the leaps and bounds of functionality we've added in v0.7 and v1.0. 

Further, there's [this description](https://searchengineland.com/google-will-require-amp-canonical-pages-match-february-1-2018-286921) of AMP + Google Search from Search Engine Land:
> Though AMP isn't a ranking signal, non-compliant pages won't appear in Top Stories or other AMP-specific slots in search results.

Paired mode more closely meets this "canonical AMP" requirement, and Native AMP does it to 100%.

alcurrie:
Based on the discussion in #1288 re.   I've added a new issue What to do with sites who run Classic Mode who can run Paired or Native modes in v1.1. #1538 - there may be some overlap with this discussion though.

westonruter:
[...]

Issue #1288, opened 2018-07-26
Onboarding Wizards


postphotos:
As a new user activating the AMP WordPress plugin for the first time, I need a user experience that helps me easily understand AMP, this plugin's features and what I can (and can't) do with it out of the box. With a streamlined experience, I should be able to activate (and understand) AMP with the least amount of steps necessary and the greatest amount of clarity possible.

- [x] AC1: ~Design a user flow that informs windows, screens, processes and messages they might receive.~ Duplicate #1359.
- [x] AC2: Evaluate this in comparison to the current user flow to see how to best roll out these features.
- [x] AC3: Suggest any new tweaks to existing settings pages, interfaces, etc. 
- [x] ~AC4: Suggest what to do with sites who run Classic Mode who can run Paired or Native modes.~  (moved to issue #1536 ) 
- [x] ~AC5: obtain the validation results for the most recently-published post and make them available to the user even before activating paired or classic modes.~ (suggested [here](https://github.com/Automattic/amp-wp/issues/1288#issuecomment-418892411).)(moved to issue #1537 ) 

_Placeholder ticket to be able to discuss new user experience questions for @postphotos, @jwold and anyone else who'd like to chime in._

Related #1006, #1273, #1254, #1235, #1359.

jwold:
![img_f14381971e46-1](https://user-images.githubusercontent.com/3411173/43288410-942f7c5e-90dc-11e8-9ae7-257e42e76b68.jpeg)

Thoughts from call with Leo and I

postphotos:
Hi @jwold - story and AC have been drafted here.

jwold:
<img width="1366" alt="validation user journey" src="https://user-images.githubusercontent.com/3411173/44753577-6f7bb600-aad3-11e8-8bb6-cadf3ff41109.png">

Had a great discussion with @postphotos today to think through the implications of presenting options to the user when they need to decide whether to go with classic, paired, or native mode.

westonruter:
Another AC is that we should make sure we obtain the validation results for the most recently-published post and make them available to the user even before activating paired or classic modes.  Actually, the validation status for each template type should be obtained and then this can be shown alongside each template in the hierarchy, as well as in some summary view with native/paired options.

alcurrie:
[...]

PR #1402, opened 2018-09-05
Default to automatic sanitization and tree shaking

westonruter:
It seems to me that having the vast majority of sites will need tree shaking enabled by default, since themes and plugins will be surely enqueueing more than 50KB of CSS. Additionally, almost every site will have plugins that are enqueueing JavaScript onto the page.

It seems to me that we should default to the plugin getting _some_ positive results upon switching to the paired mode rather than automatically redirecting to the non-AMP version with errors. The user will find out about the errors as they go when they save posts. At the very least, this should reduce the pain for new users trying out the paired mode, which should provide a much better user experience compared with the classic paired mode. Since the classic paired mode automatically removes invalid markup, having the new paired mode also default to this mode seems reasonable, especially since there will be warnings that this is occurring. They can then turn off auto-sanitization via the checkbox if it's not the behavior they want. But it seems like it will be the most frequent default that users will want.

Additionally, with improvements to the #1288, the validation status can be determined ahead of time before even enabling the native/paired modes.

Related to #1384 which is go go a step further and default to the new paired mode for new activations.


hellofromtonya:
>At the very least, this should reduce the pain for new users trying out the paired mode, which should provide a much better user experience compared with the classic paired mode.

The _provide a much better user experience_ is key here.  You're right that the vast majority of users will need tree shaking by default.

PR #1417, opened 2018-09-10
[WIP] [amp-stories sub-task] Fix layer selection.

miina:
Adds being able to select layer when clicking on icon as well. Sets timeout to `1` instead of `50` for the `selectBlock` workaround.

WIP since needs confirming if we should handle the workaround for `selectBlock` in a different way.


westonruter:
@miina Feel free to merge once you're ready. I tested and the layer selection buttons work better.

PR #1424, opened 2018-09-12
Ensure only the post being edited in the classic editor is queued for validation

### LINKED ISSUE #1404: Validator performance with bulk publishing posts
When creating lots of posts through a importing script, the server CPU usage gets really high because it runs the validation for all posts at once.

I've seen this happen with WordPress Importer and WooCommerce import. Althought the WooCommerce support is not available yet, this issue has to do with any kind of bulk publishing posts.

I didn't check the source yet but maybe we could use a flag, or even better, a queue for validator hooked function, to prevent it to be executed for lots of posts at once. And it could be a simple variable set in the AMP configuration a sort of "limit of simultaneous validator instances running".

westonruter:
When a post is saved via the classic editor it needs to be re-validated on the frontend to show the validation warning:

image

However, at present the `save_post` action is being used in a way that other post edits are also causing re-validation to happen for them as well. This is particularly problematic when doing a mass import of posts.

So what is needed is to restrict the `save_post` action to only follow-through when we are _explicitly_ saving changes for a _specific_ post in the classic editor.

Note that this is all much cleaner for Gutenberg because we can rely on the REST API field to manage the re-validation.

Fixes #1404.

### DIFF
--- a/includes/validation/class-amp-validation-manager.php
+++ b/includes/validation/class-amp-validation-manager.php
@@ -486,14 +486,32 @@ class AMP_Validation_Manager {
 	/**
 	 * Handle save_post action to queue re-validation of the post on the frontend.
 	 *
+	 * This is intended to only apply to post edits made in the classic editor.
+	 *
+	 * @see AMP_Validation_Manager::get_amp_validity_rest_field() The method responsible for validation post changes via Gutenberg.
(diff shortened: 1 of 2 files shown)


ricardobrg:
@westonruter it worked perfectly with WooCommerce CSV import and WordPress xml import.

PR #1427, opened 2018-09-13
Work around syntax error caused by the pot-to-php script

### LINKED ISSUE #1416: Syntax error in amp-translations.php can block deployments
The `pot-to-php` script is used to generate an `amp-translations.php` file for WordPress.org to use for translations:

https://github.com/Automattic/amp-wp/blob/b3e1b4d07dda70c5070d30f7426f849eb3a39823/package.json#L34

However, the `amp-translations.php` file has a PHP syntax error with a spurious comma on line 6:

```php
<?php
/* THIS IS A GENERATED FILE. DO NOT EDIT DIRECTLY. */
$generated_i18n_strings = array(
	/* translators: Copyright (C) 2018 WordPress.com VIP, XWP, Google, and contributors
This file is distributed under the same license as the AMP plugin. */
,

	/* Plugin Name of the plugin */
	__( 'AMP', 'amp' ),
```

This syntax error doesn't cause a problem for the purposes of translating on WordPress.org since the file isn't actually executed, but rather the gettext function calls are just extracted.

However, the presence of this syntax error causes problems when attempting to deploy a build of the plugin to a hosting environment where PHP syntax checking is part of a `pre-receive` hook, so a git push can then fail. So far this is known to be an issue on WP Engine and it would also be a problem on WordPress.com VIP.

swissspidy:
This is an attempt to fix #1416 for the time being.

I still have to look into how the pot-to-php script handles comments exactly and how it can be solved. We likely don't need that script anymore in the long run though.

Also, I need to find some time to look into the WP-CLI command. It should be possible to skip the file comment entirely by using `--file-comment=""` instead of using a somewhat hacky `--file-comment="*/\'\'/*"` workaround. This would also mostly eliminate the need to fix the pot-to-php script.

Fixes #1416.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -49,7 +49,7 @@ module.exports = function( grunt ) {
 				command: 'npm run pot-to-php'
 			},
 			makepot: {
-				command: 'wp i18n make-pot . languages/amp-js.pot --include="*.js"'
+				command: 'wp i18n make-pot . languages/amp-js.pot --include="*.js" --file-comment="*/\'\'/*"'
 			},
 			create_build_zip: {
 				command: 'if [ ! -e build ]; then echo "Run grunt build first."; exit 1; fi; if [ -e amp.zip ]; then rm amp.zip; fi; cd build; zip -r ../amp.zip .; cd ..; echo; echo "ZIP of build: $(pwd)/amp.zip"'
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -46,7 +46,7 @@ module.exports = function( grunt ) {
 				command: 'cross-env BABEL_ENV=production webpack'
 			},
 			pot_to_php: {
-				command: 'npm run pot-to-php'
+				command: 'npm run pot-to-php && php -l languages/amp-translations.php'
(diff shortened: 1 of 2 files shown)


westonruter:
@swissspidy the fix doesn't seem to be working:

```php
<?php
/* THIS IS A GENERATED FILE. DO NOT EDIT DIRECTLY. */
$generated_i18n_strings = array(
	/* translators: *//* */
,

	/* Plugin Name of the plugin */
	__( 'AMP', 'amp' ),
```

Checking syntax:

```
php -l build/languages/amp-translations.php
PHP Fatal error:  Cannot use empty array elements in arrays in build/languages/amp-translations.php on line 5
```

westonruter:
Adding more slashes fixes the problem: e301ee6.

swissspidy:
Thanks for the review!

The slashes might be an environment thing. I thought it worked fine when I tested it the other day...

In the meantime I submitted a PR to the WP-CLI command to make `--file-comment=""` work: https://github.com/wp-cli/i18n-command/pull/82

PR #1429, opened 2018-09-14
Update validation model with auto-accepted (instead of forcibly sanitized) and statuses for new-accepted/new-rejected

westonruter:
We currently have 3 term statuses:

* new = 0
* accepted = 1
* rejected = 2

So what we talked about was combining new with both accepted and rejected to then have 4:

* new rejected = 0 = `0b00`
* new accepted = 1 = `0b01`
* ack accepted = 3 = `0b11`
* ack rejected = 2 = `0b10`

In this arrangement, we have a nice feature: a bitmask of `0b10` will match all acknowledged (not-new) validation errors, whereas a bitmask of `0b01` will match all accepted validation errors (regardless of whether new or acknowledged).

* New-accepted and new-rejected validation errors are styled similar to unmoderated comments.
* The force-sanitization option is replaced with an auto-accept sanitization option. Validation errors can still be rejected after the fact.
* Validation errors will be marked as new-accepted if the auto-accept sanitization checkbox is checked, _or_ the user has native mode enabled.
* The preview button is restored on the Invalid URL screen in native mode.
* When a validation error is explicitly marked as rejected in native mode, then the `amp` attribute is removed from the `html` element.
* Replace trashing Invalid URLs with permanent deletion.
* Add `wp amp reset-site-validation` command to purge site of validation data.
* Stop hiding empty validation error terms and add an “Clear Empty” button to remove those with no URLs associated with them.

# Todo

- [x] Figure out why validation errors get marked as new even though previously they were accepted, when re-validating a trashed invalid URL.
- [x] Add a button that allows the user to delete all terms with 0 count. This explicit request to delete 0-count terms could then be preceded by a call to `wp_update_term_count_now()` for each term to make sure it is actually 0 before we go ahead and delete them.
- [x] Make sure updating invalid URL post won't modify new validation errors if no selection is made.
- [ ] ~Update filters so that “Accepted” matches both new and acknowledged, and the same for “Rejected” (which should be linked to from auto-accept checkbox label). New can match both new-accepted and new-rejected still.~ (To explore at another time, see `update/error-status-dropdown-options` branch)
- [ ] ~Delete taxonomy terms entirely when an invalid URL post is deleted. When the count goes to zero.~


westonruter:
@amedina This is ready for initial testing according to the revised debugging workflow. PTAL and check if it aligns with your expectations.

westonruter:
@amedina There is now a “Clear Empty” button that appears when there are validation errors with no associated URLs (i.e. they have all been forgotten):

image

assets/css/amp-validation-error-taxonomy.css
```diff
@@ -159,6 +159,27 @@ details[open] .details-attributes__summary::after, details.single-error-detail[o
...
+	color: #006505;
+}
+
+.row-actions .amp_validation_error_reject > a {
+	color: #a00;
+}
+
 .notice.accept-reject-error {
 	display: flex;
 }
```
kienstra: In a different CSS file, what do you think about changing this width to `20%`:

https://github.com/Automattic/amp-wp/blob/e67234d86375cbc3454416e84365219eab1d12df/assets/css/amp-validation-single-error-url.css#L118-L120

The 'status' icon sometimes appears above the `<select>` when the `<select>` is wider. Though this happens for all of them at slightly more narrow screen widths.

singel-url

includes/class-amp-cli.php
```diff
@@ -263,6 +263,69 @@ public function validate_site( $args, $assoc_args ) {
...
+	 * ## EXAMPLES
+	 *
+	 *     wp amp reset-site-validation --yes
+	 *
+	 * @subcommand reset-site-validation
+	 * @param array $args       Positional args. Unused.
+	 * @param array $assoc_args Associative args.
+	 * @throws Exception If an error happens.
+	 */
+	public function reset_site_validation( $args, $assoc_args ) {
```
kienstra: Great idea, this is something I've had to do manually with longer WP-CLI commands to delete the error posts and terms.

kienstra: This worked well:

reset-site-validation

no-invalid-urls

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -1196,38 +1398,62 @@ public static function filter_terms_clauses_for_description_search( $clauses, $t
...
+						'amp'
+					),
+					$count
+				);
+			} elseif ( self::VALIDATION_ERROR_REJECT_ACTION === $actioned ) {
+				$message = sprintf(
+					/* translators: %s is number of errors rejected */
+					_n(
+						'Rejected %s error. It will continue to block related URLs from being served as AMP.',
+						'Rejected %s errors. They will continue to block related URLs from being served as AMP.',
```
kienstra: This looks good:

notice-reject-bulk-action

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -1196,38 +1398,62 @@ public static function filter_terms_clauses_for_description_search( $clauses, $t
...
+
+			if ( $message ) {
+				printf( '<div class="notice notice-success is-dismissible"><p>%s</p></div>', esc_html( $message ) );
+			}
 		}
 
-		if ( $message ) {
-			printf( '<div class="notice notice-success is-dismissible"><p>%s</p></div>', esc_html( $message ) );
+		// Show success message for clearing empty terms.
+		if ( isset( $_GET[ self::VALIDATION_ERRORS_CLEARED_QUERY_VAR ] ) ) { // WPCS: csrf ok.
```
kienstra: This looks good. After 'forgetting' some errors and clicking 'Clear empty,' this appeared:

after-clear-empty

tests/validation/test-class-amp-validation-error-taxonomy.php
```diff
@@ -513,6 +705,38 @@ public function test_add_error_type_clauses_filter() {
 		$this->assertEquals( $initial_clauses, apply_filters( $tested_filter, $initial_clauses, $taxonomies ) );
 	}
 
+	/**
+<<<<<<< HEAD
```
kienstra: It looks like this is still here from resolving the (many) merge conflicts.

kienstra:
**Errors Appear As Expected**

Paired Mode, 'Automatically accept sanitization' unchecked:

paired-new-rejected

Paired Mode,  'Automatically accept sanitization' checked:

paired-auto

Native Mode:
native-mode

assets/css/amp-validation-error-taxonomy.css
kienstra: This styling issue isn't caused by this PR, it's just more apparent when the `<select>` has longer text, like `'New Rejected'`.

kienstra: Applied in https://github.com/Automattic/amp-wp/pull/1429/commits/276c5a59495af6c3c2c1cfc3d4c484aa51083e88

tests/validation/test-class-amp-validation-error-taxonomy.php
kienstra: Fixed with https://github.com/Automattic/amp-wp/pull/1429/commits/cd2010670c8e0ab8e4baa303148612f2c29debde

PR #1426, opened 2018-09-12
Expose and store queried object for validated URL; show edit link

westonruter:
Applies the following part of the [design](https://sketch.cloud/s/VoDKr/all/page-1/amp-validation-errors-for-amp-test-embeds-notification-message-on-the-top-right) from #1365:

image

PR #1433, opened 2018-09-14
Prevent store_validation_errors from clobbering global post when empty invalid_url_post supplied

westonruter:
Another fix for regression introduced #1426. See also #1428.

When creating a post in Gutenberg and publishing, and then making a change to then update, at that point the invalid URL post replaces the post you're editing.

The issue is that I switched from a boolean check here https://github.com/Automattic/amp-wp/pull/1426/files#diff-57518db0811ee918fee17d27459ed4ecL358:

image

So switching back to an `empty()` check fixes the problem.


includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -360,9 +360,11 @@ public static function get_url_from_post( $post ) {
 	public static function store_validation_errors( $validation_errors, $url, $args = array() ) {
```
amedina: Not for now but may consider splitting this function? Kind of long.

PR #1437, opened 2018-09-17
Update admin screen access

westonruter:
While the AMP validation screens are generally available to users who can `edit_posts`, in reality this should be tightened up to just administrators, with authors then only able to access the `amp_invalid_url` screens with just the posts that they themselves were responsible for creating in the course of creating regular posts.

* Only show reciprocal links on Invalid URLs and Error Index screens when user has authorization, respectively.
* Do not show the recheck row action for invalid URLs if they do not have permission.
* Hide AMP admin bar link if user doesn't have authorization.
* Harmonize approach for adding wp-heading-inline links. See https://github.com/Automattic/amp-wp/pull/1397#discussion_r217902283.
* Allow non-admin users to access AMP settings screen, but let all fields be disabled so that they can only see how it is configured.
* Do not give non-admins access to the Error Index screen.

### DIFF
--- a/includes/options/class-amp-options-menu.php
+++ b/includes/options/class-amp-options-menu.php
@@ -55,7 +55,7 @@ class AMP_Options_Menu {
 		add_menu_page(
 			__( 'AMP Options', 'amp' ),
 			__( 'AMP', 'amp' ),
-			'manage_options',
+			'edit_posts',
 			AMP_Options_Manager::OPTION_NAME,
 			array( $this, 'render_screen' ),
 			self::ICON_BASE64_SVG
@@ -65,7 +65,7 @@ class AMP_Options_Menu {
 			AMP_Options_Manager::OPTION_NAME,
 			__( 'AMP Settings', 'amp' ),
 			__( 'General', 'amp' ),
-			'manage_options',
+			'edit_posts',
 			AMP_Options_Manager::OPTION_NAME
 		);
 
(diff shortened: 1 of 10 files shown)


westonruter:
@postphotos this fixes the issue you noted when a non-admin tries to access AMP screens.

includes/options/class-amp-options-menu.php
```diff
@@ -553,14 +553,21 @@ public function render_screen() {
 			AMP_Options_Manager::check_supported_post_type_update_errors();
 		}
 		?>
+		<?php if ( ! current_user_can( 'manage_options' ) ) : ?>
+			<div class="notice notice-info">
+				<p><?php esc_html_e( 'You do not have permission to modify these settings. They are shown here for your reference. Please contact your administrator to make changes.', 'amp' ); ?></p>
```
kienstra: Good idea to display this notice.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -1019,12 +1052,19 @@ public static function handle_validate_request() {
 				if ( ! $post || self::POST_TYPE_SLUG !== $post->post_type ) {
 					throw new Exception( 'invalid_post' );
 				}
+				if ( ! current_user_can( 'edit_post', $post->ID ) ) {
+					throw new Exception( 'unauthorized' );
```
kienstra: Good idea to check the capability here.

kienstra:
**Approved**

Hi @weston,
This PR looks good, and it's nice how it changes the capabilities for the validator.

This looked good, when testing it for an administrator, editor, and author.

authors-view-page

PR #1438, opened 2018-09-17
Add caption support to all amp-instagram embeds

### LINKED ISSUE #822: Add caption support to all amp-instagram embeds
We've started to receive "Social Media Issues" emails from the Google Search Console because our `amp-instagram` embeds are missing captions which we display on the canonical by default.

They state: "Instagram posts embedded in the article are missing captions, which leads to context being lost."

Caption support for `amp-instagram` embeds was only added at the end of March this year.
https://github.com/ampproject/amphtml/pull/8243

I propose making the necessary changes to `amp-wp/includes/embeds/class-amp-instagram-embed.php` to allow captions to be shown as default for all `amp-instagram` embeds to avoid future "Social Media Issues" for publishers.

If people are happy for this go ahead, I'll knock up the needed PR as soon as I can.

chandrapatel:
Fixes #822

### DIFF
--- a/includes/embeds/class-amp-instagram-embed.php
+++ b/includes/embeds/class-amp-instagram-embed.php
@@ -144,13 +144,18 @@ class AMP_Instagram_Embed_Handler extends AMP_Base_Embed_Handler {
 	private function create_amp_instagram_and_replace_node( $dom, $node ) {
 		$instagram_id = $this->get_instagram_id_from_url( $node->getAttribute( 'data-instgrm-permalink' ) );
 
-		$new_node = AMP_DOM_Utils::create_node( $dom, $this->amp_tag, array(
+		$node_args = array(
 			'data-shortcode' => $instagram_id,
-			'data-captioned' => '',
 			'layout'         => 'responsive',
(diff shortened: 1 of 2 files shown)


westonruter:
Should the attribute be included in the `create_amp_instagram_and_replace_node` method as well?

includes/embeds/class-amp-instagram-embed.php
```diff
@@ -145,6 +146,7 @@ private function create_amp_instagram_and_replace_node( $dom, $node ) {
 
 		$new_node = AMP_DOM_Utils::create_node( $dom, $this->amp_tag, array(
 			'data-shortcode' => $instagram_id,
+			'data-captioned' => '',
```
chandrapatel: @westonruter Do we need to add this attribute only if `<blockquote>` element has `data-instgrm-captioned` attribute?

chandrapatel: Thanks, I've made changes and PR updated.

westonruter:
Thank you very much!

PR #1440, opened 2018-09-17
Checks for INTL_IDNA_VARIANT_UTS46

### LINKED ISSUE #1439: Warning: Use of undefined constant INTL_IDNA_VARIANT_UTS46
```
( ! ) Warning: Use of undefined constant INTL_IDNA_VARIANT_UTS46 - assumed
'INTL_IDNA_VARIANT_UTS46' (this will throw an Error in a future version of PHP) in
/home/public_html/wp-content/plugins/amp/includes/class-amp-http.php on line 205
```

From looking online, this may be due to a really old version of ICU running on the machine, but something that we may want to account for?

kraftbj:
Fixes #1439 

Since ICU, which provides the `INTL_IDNA_VARIANT_UTS46` may not be kept up to date the same as PHP, a separate check here will prevent the undeclared constant warning.

The downside is `idn_to_utf8` function defaults to using `INTL_IDNA_VARIANT_2003` which would then throw a deprecated warning in PHP 7.2+. If PHP 7.4, it is planned that the default will be `INTL_IDNA_VARIANT_UTS46` which would likely return the original issue.

@westonruter What would you prefer? Checking for UTS46 and taking the deprecated in PHP 7.2 for sites running ancient ICU or check for the UTS46 variant along with the `idn_to_utf8` exists check? Or a third option?

### DIFF
--- a/includes/class-amp-http.php
+++ b/includes/class-amp-http.php
@@ -201,7 +201,7 @@ class AMP_HTTP {
 		 */
 		foreach ( $domains as $domain ) {
 			if ( function_exists( 'idn_to_utf8' ) ) {
-				if ( version_compare( PHP_VERSION, '5.4', '>=' ) ) {
+				if ( version_compare( PHP_VERSION, '5.4', '>=' ) && defined( 'INTL_IDNA_VARIANT_UTS46' ) ) {
 					$domain = idn_to_utf8( $domain, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46 ); // phpcs:ignore PHPCompatibility.PHP.NewFunctionParameters.idn_to_utf8_variantFound, PHPCompatibility.PHP.NewConstants.intl_idna_variant_uts46Found
 				} else {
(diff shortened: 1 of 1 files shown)


kraftbj:
After a bit more research and touching base with sysadmin, the setup of the testing box is a default CentOS6/cPanel-powered box with PHP updated. So, I think we'll need to account for both lacking `INTL_IDNA_VARIANT_UTS46` and a deprecated `INTL_IDNA_VARIANT_2003`.

kraftbj:
> So if ICU is at an old version as before PHP 5.4 but the current version of PHP is 7.2+?

Yup, ICU is independent of PHP, apparently. 5.4+ integrated the "wrapper", but isn't an indicator of what version of ICU is available to a given system. So yes, PHP 7.2+ with older ICU <4.6, we're stuck between a rock and a hard place.

For now, I think what I have is decent. I would expect most people with 7.2+ with outdated ICU would be fewer than people on 5.4+ with outdated ICU. Long-term, I need to chew on it more.

PR #1442, opened 2018-09-18
Display a welcome notice on the main 'AMP Settings' page

### LINKED ISSUE #1381: Implement settings splash notice
**As a new user of the AMP WP plugin, I need a splash screen that helps with understanding of what the plugin does, what settings do and how to solve AMP errors.**

- [x] AC1: WORDING - Identify or write copy that helps explain what settings do, what next steps are past activation (related #1366).  
- [ ] AC2: DESIGN - Finalize design (related #1288) of what this should look like in a high fidelity fashion with all art assets prepared for the plugin. 
   - [ ] Link to the product site, where a user would be able to understand what the plugin does, how AMP works and why they should use this tool.
- [ ] AC3: INTERFACE - On the settings page, implement the ability to have a dismissible splash "panel" that sits above the toggles.
- [ ] AC4: Implement new settings toggle design (related #1288) from AC2.  
- [ ] AC5: Combine AC1, AC2 and AC3, AC4 work in visual styles and functionality.

cc: @jwold @jillchu

kienstra:
* Adds a placeholder notice to the 'AMP Settings' page, until the copy is available
* Once the user dismisses this, it will never appear again

amp-settings-notice

Closes #1381


includes/options/class-amp-options-manager.php
```diff
@@ -327,6 +328,45 @@ public static function update_analytics_options( $data ) {
...
+	 * @todo: Add the full copy for this notice when it is decided.
+	 */
+	public static function render_welcome_notice() {
+		if ( 'toplevel_page_' . self::OPTION_NAME !== get_current_screen()->id ) {
+			return;
+		}
+
+		$notice_id = 'welcome-notice-1';
+		$dismissed = get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true );
+		if ( in_array( $notice_id, explode( ',', strval( $dismissed ) ), true ) ) {
```
kienstra: **Question About Admin Pointer User Meta**

Hi @westonruter,
What do you think about how this uses the user meta for admin pointers to keep track of whether this notice was dismissed? 

This is very similar to how [this admin pointer PR](https://github.com/Automattic/amp-wp/pull/1271/files#diff-7ec3f55e1310503aa82ce61984d3f482R78) does this.

Of course, this isn't actually an admin pointer, so it's not ideal. 

But this is a little simpler, as it can reuse the `dismiss-wp-pointer' action handler when the user dismisses it.

Thanks, Weston!

includes/options/class-amp-options-manager.php
```diff
@@ -327,6 +328,45 @@ public static function update_analytics_options( $data ) {
...
+	 * So once the user dismisses this notice, it will never appear again.
+	 *
+	 * @todo: Add the full copy for this notice when it is decided.
+	 */
+	public static function render_welcome_notice() {
+		if ( 'toplevel_page_' . self::OPTION_NAME !== get_current_screen()->id ) {
+			return;
+		}
+
+		$notice_id = 'welcome-notice-1';
```
westonruter: This should have some `amp` prefix.

includes/options/class-amp-options-manager.php
```diff
@@ -327,6 +328,45 @@ public static function update_analytics_options( $data ) {
...
+		}
+
+		?>
+		<div class="notice notice-info is-dismissible" id="<?php echo esc_attr( $notice_id ); ?>"><p><?php esc_html_e( 'Welcome to AMP for WordPress', 'amp' ); ?></p></div>
+
+		<script>
+		( function( $ ) {
+			$( function() {
+				// On dismissing the notice, make a POST request to store this notice with the dismissed WP pointers so it doesn't display again.
+				$( document ).on( 'click', '#<?php echo $notice_id; // WPCS: XSS OK. ?> .notice-dismiss', function() {
```
westonruter: Instead of: `'#<?php echo $notice_id; // WPCS: XSS OK. ?> .notice-dismiss'` 
Do this: `<?php echo wp_json_encode( "#$notice_id .notice-dismiss" ); ?>`

includes/options/class-amp-options-manager.php
```diff
@@ -327,6 +328,45 @@ public static function update_analytics_options( $data ) {
...
+					console.log( 'clicked' );
+					$.post( ajaxurl, {
+						pointer: '<?php echo $notice_id; // WPCS: XSS OK. ?>',
```
westonruter: As above, better to do `pointer: <?php echo wp_json_encode( $notice_id ); ?>`,

includes/options/class-amp-options-manager.php
westonruter: My main concern is what happens over time. If each new version of the plugin will include new information that is shown in this message, then the `dismissed_wp_pointers` usermeta will grow endlessly.

You could instead consider what was added for the Try Gutenberg screen:

https://github.com/WordPress/wordpress-develop/blob/5d477aa70613850a2e78cdaecb7e2de365de7134/src/wp-admin/js/dashboard.js#L34-L75

https://github.com/WordPress/wordpress-develop/blob/5d477aa70613850a2e78cdaecb7e2de365de7134/src/wp-admin/includes/ajax-actions.php#L1487-L1498

But maybe this is all overkill. For a one-off, I don't see a reason why to not use the admin-pointers. They're being used for non-pointer things already, including the theme and plugin editor warnings:

https://github.com/WordPress/wordpress-develop/blob/5d477aa70613850a2e78cdaecb7e2de365de7134/src/wp-admin/plugin-editor.php#L282-L294

https://github.com/WordPress/wordpress-develop/blob/5d477aa70613850a2e78cdaecb7e2de365de7134/src/wp-admin/theme-editor.php#L301-L313

https://github.com/WordPress/wordpress-develop/blob/5d477aa70613850a2e78cdaecb7e2de365de7134/src/wp-admin/js/theme-plugin-editor.js#L134-L150

So I'd say go ahead with pointers.

includes/options/class-amp-options-manager.php
kienstra: Thanks, https://github.com/Automattic/amp-wp/pull/1442/commits/8f02acf695c24ad6b5a2045a988e439564119381#diff-6dc720c5e5352d31e656ef4973984528R357 applies your suggestion.

includes/options/class-amp-options-manager.php
kienstra: Thanks, https://github.com/Automattic/amp-wp/pull/1442/commits/8f02acf695c24ad6b5a2045a988e439564119381#diff-6dc720c5e5352d31e656ef4973984528R360 applies your snippet.

includes/options/class-amp-options-manager.php
```diff
@@ -327,6 +328,45 @@ public static function update_analytics_options( $data ) {
...
+		}
+
+		?>
+		<div class="notice notice-info is-dismissible" id="<?php echo esc_attr( $notice_id ); ?>"><p><?php esc_html_e( 'Welcome to AMP for WordPress', 'amp' ); ?></p></div>
+
+		<script>
+		( function( $ ) {
+			$( function() {
+				// On dismissing the notice, make a POST request to store this notice with the dismissed WP pointers so it doesn't display again.
+				$( document ).on( 'click', <?php echo wp_json_encode( "#$notice_id .notice-dismiss" ); ?>, function() {
```
westonruter: Another point, this code is using event delegation when it doesn't need to. It should instead just be:

```
$( <?php echo wp_json_encode( "#$notice_id .notice-dismiss" ); ?> ).on( 'click', function() {...});
```

kienstra: Hi Weston,
Good point. 

But it looks like the `button.notice-dismiss` might be created with JS when the `is-dismissible` class is present, and might not yet exist at the time the `<script>` above runs. 

```html
<button type="button" class="notice-dismiss">
```

So the `click` handler with that selector doesn't run.

Maybe if this script ran later, the `button.notice-dismiss` would be present.

westonruter: ok, that makes sense

kienstra:
**Notice Copy Applied**
https://github.com/Automattic/amp-wp/pull/1442/commits/6e9a09e753091d74a5dbc5d904dbf7d1fc49d8fb applies the notice copy from https://github.com/Automattic/amp-wp/issues/1381#issuecomment-422358593:

notice-amp

includes/options/class-amp-options-manager.php
```diff
@@ -327,6 +328,63 @@ public static function update_analytics_options( $data ) {
...
+		if ( in_array( $notice_id, explode( ',', strval( $dismissed ) ), true ) ) {
+			return;
+		}
+
+		?>
+		<div class="notice notice-info is-dismissible" id="<?php echo esc_attr( $notice_id ); ?>">
+			<span class="amp-logo-icon"></span>
+			<h1><?php esc_html_e( 'Welcome to the AMP for WordPress plugin v1.0', 'amp' ); ?></h1>
+			<p><?php esc_html_e( 'Thank you for installing! Bring the speed and features of the open source AMP project to your site, the WordPress way, complete with the tools to support content authoring and website development.', 'amp' ); ?></p>
+			<h2><?php esc_html_e( 'What&#39;s New', 'amp' ); ?></h2>
```
kienstra: The `&#39;` character here might not be the best way to render the `'`.

kienstra:

Hi @westonruter,
Could you please have another look at this PR, especially the [latest commit](https://github.com/Automattic/amp-wp/pull/1442/commits/6e9a09e753091d74a5dbc5d904dbf7d1fc49d8fb) that [applies the notice copy](https://github.com/Automattic/amp-wp/pull/1442#issuecomment-422608434)? 

Thanks, Weston!

kienstra:
*'Thank you for installing!'*

Ideally, this would apply the 'Thank you for installing!' only on installing `v1.0`, not upgrading from `0.7.2` to this.

includes/options/class-amp-options-manager.php
westonruter: That's right. It should be `&#8217;` instead.

includes/options/class-amp-options-manager.php
westonruter: At least we can do this:
```js
$( <?php echo wp_json_encode( "#$notice_id" ); ?> ).on( 'click', '.notice-dismiss', function() {
```
That will scope the delegated event to the containing element.

includes/options/class-amp-options-manager.php
```diff
@@ -327,6 +328,63 @@ public static function update_analytics_options( $data ) {
...
+						action: 'dismiss-wp-pointer'
+					} );
+				} );
+			} );
+		}( jQuery ) );
+		</script>
+
+		<style>
+			.amp-logo-icon {
+				background-image: url( "<?php echo esc_url( amp_get_asset_url( 'images/amp-logo-icon.svg' ) ); ?>" );
```
westonruter: HTML entities don't render properly inside of CSS. So best to just remove `esc_url()` I believe.

includes/options/class-amp-options-manager.php
```diff
@@ -327,6 +328,63 @@ public static function update_analytics_options( $data ) {
...
+			<h2><?php esc_html_e( 'What&#39;s New', 'amp' ); ?></h2>
+			<p><?php esc_html_e( 'From granular controls that help you create AMP content, to Core Gutenberg support, to a sanitizer that only shows visitors error-free pages, to a full error workflow for developers, this v1.0 release makes it easier than ever to bring a rich, performant experience to your WordPress site via AMP HTML.', 'amp' ); ?></p>
+		</div>
+
+		<script>
+		( function( $ ) {
+			$( function() {
+				// On dismissing the notice, make a POST request to store this notice with the dismissed WP pointers so it doesn't display again.
+				$( document ).on( 'click', <?php echo wp_json_encode( "#$notice_id .notice-dismiss" ); ?>, function() {
+					console.log( 'clicked' );
```
westonruter: Debug code to be removed.

includes/options/class-amp-options-manager.php
```diff
@@ -327,6 +328,63 @@ public static function update_analytics_options( $data ) {
...
+			<span class="amp-logo-icon"></span>
+			<h1><?php esc_html_e( 'Welcome to the AMP for WordPress plugin v1.0', 'amp' ); ?></h1>
+			<p><?php esc_html_e( 'Thank you for installing! Bring the speed and features of the open source AMP project to your site, the WordPress way, complete with the tools to support content authoring and website development.', 'amp' ); ?></p>
+			<h2><?php esc_html_e( 'What&#39;s New', 'amp' ); ?></h2>
+			<p><?php esc_html_e( 'From granular controls that help you create AMP content, to Core Gutenberg support, to a sanitizer that only shows visitors error-free pages, to a full error workflow for developers, this v1.0 release makes it easier than ever to bring a rich, performant experience to your WordPress site via AMP HTML.', 'amp' ); ?></p>
+		</div>
+
+		<script>
+		( function( $ ) {
+			$( function() {
```
westonruter: This function should be combined with its parent function:

```js
jQuery( function( $ ) {
/* ... */
} );
```

includes/options/class-amp-options-manager.php
```diff
@@ -327,6 +328,63 @@ public static function update_analytics_options( $data ) {
...
+
+		$notice_id = 'amp-welcome-notice-1';
+		$dismissed = get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true );
+		if ( in_array( $notice_id, explode( ',', strval( $dismissed ) ), true ) ) {
+			return;
+		}
+
+		?>
+		<div class="notice notice-info is-dismissible" id="<?php echo esc_attr( $notice_id ); ?>">
+			<span class="amp-logo-icon"></span>
```
westonruter: I think this should just be a regular `img`.

jwold:
@kienstra 

I'd like to suggest the following changes to the welcome banner (note that I changed a few lines of text as well). What do you think?

illustration

And the icon:

[icon.svg.zip](https://github.com/Automattic/amp-wp/files/2399501/icon.svg.zip)

cc @postphotos

jwold:
Going to actually post this to 1381, since that's the real issue :)

PR #1443, opened 2018-09-18
Replace the 'Settings saved' notice, only on changing the Template Mode

### LINKED ISSUE #1399: Settings callout for "view your site as amp" after initial save
**As a user on first activation of the plugin, I should have a clear call-to-action after deciding my AMP settings for my site to view my site as AMP or to debug errors.**

 In current form, a user completes the form on the settings page and is presented with a notice that says `Settings Saved.` - this intends to replace this with new copy and directions. While it's important to notify a user that the tool works as expected, a user also needs a clear throughline from activation of the plugin to engagement with the AMP version of their to understand what the point of this tool is and how it works.

As part of the work in #1006, #1288, #1359 it was clear that a new information design for the settings page was needed. By providing a strong CTA after initial save, a user should engage with the AMP site more readily.

- [ ] AC1: Write copy that encourages a user to engage with the site given the condition. Suggestions : `View the AMP version of your site!` or `View the Classic AMP version of your site. (We strongly recommend upgrading to Paired Mode.)` or `View the Native AMP version of your site.`
- [ ] AC2: Based on the outcomes of #1381, implement the ability to surface a call-to-action that would appear after a user "saves" the settings.
- [ ] AC3: Integrate copy from AC1 into the new CTA feature.

kienstra:
* The copy can still change, but this now has [the copy here](https://github.com/Automattic/amp-wp/issues/1399#issuecomment-420247537) for all 3 template modes
* Once this displays, it never displays again

Todos include:
- [x] At the moment, this display on the first change of _any_ setting on this page, even 'Automatically accept sanitization...' 
It might be best to only display this message when one of the template modes changes.
- [x] On first activating the plugin, there probably won't be any validation errors (see https://github.com/Automattic/amp-wp/issues/1399#issuecomment-422278943). So the 'or Review errors' part of the notice might not apply. This could query for errors, and conditionally show that text.

paired-mode

Closes #1399


westonruter:
> It might be best to only display this message when one of the template modes changes.

👍 

> * Once this displays, it never displays again

I don't think this should be the case necessarily. If it only shows when changing the template mode, then in reality it won't show very much.

> 2. On first activating the plugin, there probably won't be any validation errors (see https://github.com/Automattic/amp-wp/issues/1399#issuecomment-422278943). So the 'or Review errors' part of the notice might not apply. This could query for errors, and conditionally show that text.

This surfaces the discussion we had awhile ago about showing the validation status alongside each of the supportable templates.

When visiting the AMP settings screen, I think we need to make sure that there are some invalid URLs captured for us to show. This could be done server-side when switching the mode, or when visiting this settings screen there could be Ajax requests triggered to validate the latest URL of each supported template type.

That may be overkill to start with, so perhaps the best thing would be upon switching the mode to obtain the first post type that is supported by AMP, and obtain the first published post of that type that has AMP enabled, and then re-validate that URL. When the page then reloads there will be validation results available to show.

includes/options/class-amp-options-menu.php
```diff
@@ -573,4 +574,47 @@ public function render_screen() {
...
+						break;
+					case 'classic':
+						$message = esc_html__( 'Classic Mode activated! View your site as AMP now. We recommend upgrading to Native or Paired mode.', 'amp' );
+						break;
+				}
+
+				if ( isset( $message ) ) {
+					$wp_settings_errors[ $key ]['message'] = $message; // WPCS: Global override OK.
+					// Ensure that this notice does not display again.
+					update_user_meta( get_current_user_id(), $meta_key, true );
```
westonruter: We can remove the user meta flag. The notice should be shown each time the user changes the mode.

includes/options/class-amp-options-menu.php
```diff
@@ -573,4 +574,47 @@ public function render_screen() {
...
+		$meta_key = 'amp_view_your_site_notice';
+		$user_id  = get_current_user_id();
+
+		// Exit if this setting isn't present, or if this notice has already displayed.
+		if ( ! get_settings_errors( 'general' ) || get_user_meta( $user_id, $meta_key, true ) ) {
+			return;
+		}
+
+		foreach ( $wp_settings_errors as $key => $setting ) {
+			if ( 'general' === $setting['setting'] && 'settings_updated' === $setting['code'] ) {
```
westonruter: If you call `add_settings_error()` inside the option sanitizer callback, where we can detect the mode change, then this will get stored in a transient to then automatically be displayed with the `admin_notices` when the page reloads.

What this won't do is hide the default settings updated notice.

Note that `add_settings_error()` has a `$type` parameter which can be `"updated"`, so not an error at all.

includes/options/class-amp-options-menu.php
kienstra: Good point. https://github.com/Automattic/amp-wp/pull/1443/commits/3a752500c5573e53217d4abfb33ea7172b88b82c removes the user meta as a condition of displaying this message.

includes/options/class-amp-options-menu.php
kienstra: Great idea. https://github.com/Automattic/amp-wp/pull/1443/commits/cc2e62fbfcbc21d980a509dbe4276bc9e95f7c58 uses `add_settings_error()` in the `sanitize_callback`. 

Though I'm not sure why, it seems like the default message doesn't display when this is added. 

on-paired-mode

kienstra:
**Validating On Switching Template Modes**

Hi @westonruter,
>...perhaps the best thing would be upon switching the mode to obtain the first post type that is supported by AMP, and obtain the first published post of that type that has AMP enabled, and then re-validate that URL. When the page then reloads there will be validation results available to show.

Good idea, it would make sense to validate a URL on switching template modes.

1. Do you think this could reuse [AMP_Validation_Manager:: validate_after_plugin_activation()](https://github.com/Automattic/amp-wp/blob/9444a7578474b3ff29810e44d6eada22851bbed7/includes/validation/class-amp-validation-manager.php#L1656) to do this? 
2. If by chance there's no error after validating that URL, should this still have the 'or Review Errors' text, with a link to an empty [Invalid URLs page](https://sketch.cloud/s/xQRzx/all/page-1/amp-validation-urls)? Or should this possibly query for `amp_invalid_url`, and only show the message with 'or Review Errors' if an error exists?

westonruter:
@kienstra using `validate_after_plugin_activation` isn't quite right because it sets that transient. Otherwise, it's right. Maybe most of the guts of that method should be moved to another method, like `validate_latest_post()` which would store the results and and return the ID for the `amp_invalid_url` post. 

In reality, the `$validation_errors` should not be stored in the transient. Only the ID should be stored. The plugin can then read the ID and then fetch the validation errors from `AMP_Invalid_URL_Post_Type::get_invalid_url_validation_errors()` instead of from the transient.

And yes, if there are no unaccepted validation errors on the URL then that “Review Errors” link should be removed.

westonruter:
Here's where I'm going with the messages. They vary based on the mode switched to, and there are links to view either the validated URL on the site, and/or to view the validation errors for that URL.

## Success message when activating native mode:

image

## Message when enabling paired mode but there are validation errors blocking it:

image

## When activating classic mode: 

image

PR #1448, opened 2018-09-20
AMP validation - add pointer tooltips for status columns

### LINKED ISSUE #1400: Tooltips for Compatibility Tool
**As a developer user debugging an error in the Compatibility Tool, I should have tooltips that hint at what given columns do to explain what they mean and how I should use that information.**

As a user gets stuck or confused, they should have both user-friendly hints and more verbose documentation at what the plugin does, why it's organized this way and how the tool intends to allow them to activate AMP quickly.

- [x] AC1: Based on #1361, #1362, #1363 and #1364 decide which columns need tooltips. Currently the Error by Type, Error by URL and Single screens all have tooltips on the header of the `Status` and `Detail` columns.
- [ ] AC2: For each tooltip, write copy and link to relevant URLs on the Plugin product site. (similar to #1401.)
- [ ] AC3: Implement each tooltip (per view, per column) based on what is decided in AC1/AC2.
- [ ] AC4: Re-implement to meet the spec of #1479.

johnwatkins0:
This partially resolves #1400 by adding a tooltip to the headings of "Status" columns on AMP validation screens. It's not fully resolved because (1) we'll need to put in the final text, and (2) the styling and positioning of the tooltips could potentially use a little fine-tuning. Further, the tooltips could easily be created with the `Tooltip` component from the `@wordpress/components` package where Gutenberg-derived assets are available --  but I didn't go this route now because that tooltip looks much different from the one that had been decided on for this task.

Edit: Added the tooltip to the "Details" column as well. There's a little bit of conflicting information on this in the tickets/comps. It can easily be removed.

### DIFF
--- a/assets/css/amp-validation-tooltips.css
+++ b/assets/css/amp-validation-tooltips.css
@@ -1,0 +1,11 @@
+.tooltip {
+    margin: 0 6px;
+}
+
+.tooltip * {
+    display: none;
+}
+
+.wp-pointer--tooltip {
+    transform: translateX(-52px);
+}
--- a/assets/css/amp-validation-tooltips.css
+++ b/assets/css/amp-validation-tooltips.css
@@ -1,11 +1,7 @@
-.tooltip {
+
+/* @todo This should be moved to admin-tables.css which is then enqueued on both screens. */
+.tooltip-button {
     margin: 0 6px;
-}
-
-.tooltip * {
-    display: none;
-}
-
-.wp-pointer--tooltip {
-    transform: translateX(-52px);
+    cursor: pointer;
+    color: #767676;
 }
--- a/assets/src/amp-validation-tooltips.js
+++ b/assets/src/amp-validation-tooltips.js
@@ -5,9 +5,9 @@ import domReady from '@wordpress/dom-ready';
 
 // WIP Pointer function
 function sourcesPointer() {
-	jQuery( '.tooltip' ).on( 'hover', function() {
+	jQuery( document ).on( 'click', '.tooltip-button', function() {
 		jQuery( this ).pointer( {
-			content: this.innerHTML,
+			content: jQuery( this ).next( '.tooltip' ).html(),
 			position: {
 				edge: 'top',
 				align: 'left'
--- a/assets/src/amp-validation-tooltips.js
+++ b/assets/src/amp-validation-tooltips.js
@@ -1,0 +1,22 @@
+/**
+ * WordPress dependencies
+ */
+import domReady from '@wordpress/dom-ready';
+
+// WIP Pointer function
+function sourcesPointer() {
+	jQuery( '.tooltip' ).on( 'hover', function() {
+		jQuery( this ).pointer( {
+			content: this.innerHTML,
+			position: {
+				edge: 'top',
+				align: 'left'
+			},
+			pointerClass: 'wp-pointer wp-pointer--tooltip'
+		} ).pointer( 'open' );
+	} );
+}
+
+domReady( () => {
+	sourcesPointer();
+} );
--- a/includes/validation/class-amp-invalid-url-post-type.php
+++ b/includes/validation/class-amp-invalid-url-post-type.php
@@ -191,22 +191,20 @@ class AMP_Invalid_URL_Post_Type {
 		wp_enqueue_style(
 			'amp-validation-error-taxonomy',
 			amp_get_asset_url( 'css/amp-validation-error-taxonomy.css' ),
-			array( 'common' ),
+			array( 'common', 'amp-validation-tooltips' ),
 			AMP__VERSION
 		);
 		wp_enqueue_script(
 			'amp-validation-error-detail-toggle',
 			amp_get_asset_url( 'js/amp-validation-error-detail-toggle-compiled.js' ),
-			array(),
+			array( 'wp-dom-ready', 'amp-validation-tooltips' ),
 			AMP__VERSION,
 			true
 		);
 		wp_localize_script(
 			'amp-validation-error-detail-toggle',
 			'ampValidationI18n',
-			array(
-				'btnAriaLabel' => esc_attr__( 'Toggle all sources', 'amp' ),
-			)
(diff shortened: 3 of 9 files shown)


kienstra:
Thanks, @johnwatkins0! If it's alright, I'll defer to Weston for the code review.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -191,22 +191,20 @@ public static function enqueue_post_list_screen_scripts() {
...
 			amp_get_asset_url( 'css/amp-validation-error-taxonomy.css' ),
-			array( 'common' ),
+			array( 'common', 'amp-validation-tooltips' ),
 			AMP__VERSION
 		);
 		wp_enqueue_script(
 			'amp-validation-error-detail-toggle',
 			amp_get_asset_url( 'js/amp-validation-error-detail-toggle-compiled.js' ),
-			array(),
+			array( 'wp-dom-ready', 'amp-validation-tooltips' ),
```
kienstra: It looks like `wp-dom-ready` might only be available in Gutenberg (or at least it's merge).

westonruter: It should not be made a script dependency. It is a module dependency.

assets/src/amp-validation-tooltips.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * WordPress dependencies
+ */
+import domReady from '@wordpress/dom-ready';
+
+// WIP Pointer function
+function sourcesPointer() {
+	jQuery( '.tooltip' ).on( 'hover', function() {
```
westonruter: I don't think that `hover` works. The tooltip really needs to not get in the way. It's too easy to trigger it. The problem could be solved if we switch to showing the tooltip on _click_ rather than on hover. This would also make it keyboard accessible.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -702,7 +700,12 @@ public static function add_post_columns( $columns ) {
 		$columns = array_merge(
 			$columns,
 			array(
-				AMP_Validation_Error_Taxonomy::ERROR_STATUS => sprintf( '%s<span class="dashicons dashicons-editor-help"></span>', esc_html__( 'Status', 'amp' ) ),  // @todo Create actual tooltip.
+				AMP_Validation_Error_Taxonomy::ERROR_STATUS => sprintf(
+					'%s<div class="tooltip dashicons dashicons-editor-help"><h3>%s</h3><p>%s</p></div>',
+					esc_html__( 'Status', 'amp' ),
+					__( 'Statuses tooltip title', 'amp' ),
```
westonruter: Placeholder title? 

Should be using `esc_html__()`.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -702,7 +700,12 @@ public static function add_post_columns( $columns ) {
...
+					__( 'An accepted validation error is one that will not block a URL from being served as AMP; the validation error will be sanitized, normally resulting in the offending markup being stripped from the response to ensure AMP validity.', 'amp' )
```
westonruter: Should be using `esc_html__()`.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -635,8 +635,18 @@ public static function add_admin_hooks() {
 			return array(
 				'cb'               => $old_columns['cb'],
 				'error'            => __( 'Error', 'amp' ),
-				'status'           => __( 'Status', 'amp' ),
-				'details'          => __( 'Details', 'amp' ),
+				'status'           => sprintf(
+					'%s<div class="tooltip dashicons dashicons-editor-help"><h3>%s</h3><p>%s</p></div>',
+					__( 'Status', 'amp' ),
+					__( 'Statuses tooltip title', 'amp' ),
+					__( 'An accepted validation error is one that will not block a URL from being served as AMP; the validation error will be sanitized, normally resulting in the offending markup being stripped from the response to ensure AMP validity.', 'amp' )
```
westonruter: Should be using `esc_html__()`.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -635,8 +635,18 @@ public static function add_admin_hooks() {
...
+				),
+				'details'          => sprintf(
+					'%s<div class="tooltip dashicons dashicons-editor-help"><h3>%s</h3><p>%s</p></div>',
+					__( 'Details', 'amp' ),
+					__( 'Details tooltip title', 'amp' ),
+					__( 'An accepted validation error is one that will not block a URL from being served as AMP; the validation error will be sanitized, normally resulting in the offending markup being stripped from the response to ensure AMP validity.', 'amp' )
```
westonruter: Should be using `esc_html__()`.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -702,7 +700,12 @@ public static function add_post_columns( $columns ) {
 		$columns = array_merge(
 			$columns,
 			array(
-				AMP_Validation_Error_Taxonomy::ERROR_STATUS => sprintf( '%s<span class="dashicons dashicons-editor-help"></span>', esc_html__( 'Status', 'amp' ) ),  // @todo Create actual tooltip.
+				AMP_Validation_Error_Taxonomy::ERROR_STATUS => sprintf(
+					'%s<div class="tooltip dashicons dashicons-editor-help"><h3>%s</h3><p>%s</p></div>',
```
westonruter: The `div` needs `tabindex=0` so that it can be focused with the keyboard.

assets/css/amp-validation-tooltips.css
```diff
@@ -0,0 +1,11 @@
+.tooltip {
+    margin: 0 6px;
+}
+
+.tooltip * {
+    display: none;
+}
+
+.wp-pointer--tooltip {
+    transform: translateX(-52px);
```
westonruter: This `transform` seems to be causing jumpiness.

PR #1449, opened 2018-09-21
[WIP] On the single error term page, apply a similar UI to the single URL details

kienstra:
@johnwatkins0 had created a nice display of the details, which expands on clicking the 'Error' row on the single URL page:

single-error-page-details-ui

Something similar to this should be applied to the single taxonomy term UI:

current-ui

### DIFF
--- a/includes/validation/class-amp-invalid-url-post-type.php
+++ b/includes/validation/class-amp-invalid-url-post-type.php
@@ -1158,31 +1158,20 @@ class AMP_Invalid_URL_Post_Type {
 			}
 
 			// @todo Update this to use the method which will be developed in PR #1429 AMP_Validation_Error_Taxonomy::get_term_error() .
-			$description  = json_decode( $error->description, true );
-			$sanitization = \AMP_Validation_Error_Taxonomy::get_validation_error_sanitization( $description );
-			$status_text  = \AMP_Validation_Error_Taxonomy::get_status_text_with_icon( $sanitization );
-			$error_code   = isset( $description['code'] ) ? $description['code'] : 'error';
-			$error_title  = \AMP_Validation_Error_Taxonomy::get_error_title_from_code( $error_code );
+			$description      = json_decode( $error->description, true );
+			$sanitization     = \AMP_Validation_Error_Taxonomy::get_validation_error_sanitization( $description );
+			$status_text      = \AMP_Validation_Error_Taxonomy::get_status_text_with_icon( $sanitization );
+			$error_code       = isset( $description['code'] ) ? $description['code'] : 'error';
+			$error_title      = \AMP_Validation_Error_Taxonomy::get_error_title_from_code( $error_code );
+			$validation_error = json_decode( $error->description, true );
+			?>
+			<div class="notice">
+				<ul>
+					<?php echo AMP_Validation_Error_Taxonomy::render_single_url_error_details( $validation_error, $error ); // WPCS : XSS OK. ?>
+				</ul>
+			</div>
 
-			$output = '';
-			foreach ( $description as $desc_name => $desc_info ) {
-				if ( ! is_array( $desc_info ) ) {
-					$output .= sprintf( '<li><span class="details-attributes__title">%s:</span><br/><span class="details-attributes__value">%s</span></li>', $desc_name, $desc_info );
-				} else {
-					$output .= sprintf( '<ul class="secondary-details-array"><span class="details-attributes__title">%s:</span><br/>', $desc_name );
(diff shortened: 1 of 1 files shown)


kienstra:
**Request To Work On This**

Hi @johnwatkins0,
Could you please work on this?

westonruter:
@kienstra I think this certainly looks better than the red/blue colors:

image

More alignment between the two styles is needed of course. Perhaps it should consistently be fixed font for keys and values?

kienstra:
Hi @westonruter,
>Perhaps it should consistently be fixed font for keys and values?
Good idea.

Is this what you had in mind?

fixed-font

westonruter:
@kienstra  Each value should be have a fixed-width font, and the values should be indented as well. No need for the braces or the quote marks. For the nested children, in that case the key can just be bold with the value not-bold. The top-level keys could have the gray background like you show above. Disclaimer: I'm not a designer.

johnwatkins0:
@westonruter @kienstra Here's how it looks with the commit just pushed: 

invalid_urls_ _local_wordpress_test_ _wordpress

johnwatkins0:
Based on @westonruter comment a little while ago, I'll change "text : javascript" to "**text**: javascript".

kienstra:
Thanks, @johnwatkins0. This looks really good.

johnwatkins0:
@westonruter This is ready for your review.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -1903,8 +1903,15 @@ public static function render_single_url_error_details( $validation_error, $term
 						<div class="detailed">
 							<?php if ( is_string( $value ) ) : ?>
 								<?php echo esc_html( $value ); ?>
-							<?php else : ?>
-								<pre><?php echo esc_html( wp_json_encode( $value, 128 /* JSON_PRETTY_PRINT */ | 64 /* JSON_UNESCAPED_SLASHES */ ) ); ?></pre>
+							<?php elseif ( is_array( $value ) ) : ?>
```
kienstra: This looked to have changed the display of the single error URL details:

single-error-url

kienstra: Addressed with https://github.com/Automattic/amp-wp/pull/1449/commits/5ebc340e40ea9c0ccd744e66a80490fc4f299875:

single-error-url

johnwatkins0: Thanks for catching that, @kienstra

kienstra: All good, thanks a lot for applying this UI.

amedina:
In the single error term page, the top box with the error details should be open by default
screen shot 2018-09-21 at 5 55 23 pm

amedina:
And perhaps the inner components should be open as well; the reason being that the user will go there specifically to look at the details of the errors and the URLs that have it

westonruter:
For the details on the removed, why are the angle brackets removed? It looks truncated. Also, shouldn't it be `monospace`?

image

And the node attributes is mostly just a duplicate of this. There should be a vertical list of attributes instead.

And all except sources should probably be expanded by default.

westonruter:
Need to figure out why the sources are being duplicated for each row:

image

westonruter:
- [ ] The order of the errors no longer corresponds to the document order. I'll look into that.

johnwatkins0:
Just pushed changes for the following: 

- [ ]  In the single error term page, the top box with the error details should be open by default
- [ ] And perhaps the inner components should be open as well; the reason being that the user will go there specifically to look at the details of the errors and the URLs that have it
- [ ]  Also, shouldn't it be monospace?
- [ ] And all except sources should probably be expanded by default.

@westonruter, on "For the details on the removed, why are the angle brackets removed?", that field is the attribute key and values concatenated to look like the removed attributes. I suppose the element could be prepended to make it look like an actual element, but I'm not comfortable enough with all the different conditions to make that happen.

westonruter:
I'm changing the plugin/other sources to eliminate the `details` when there is only one item.

So going from:

image

To:

image

westonruter:
- [ ] The column checkboxes are erroneously including the tooltips:

image

westonruter:
> on "For the details on the removed, why are the angle brackets removed?", that field is the attribute key and values concatenated to look like the removed attributes. I suppose the element could be prepended (with angle brackets) to make it look like an actual element, but I'm not comfortable enough with all the different conditions to make that happen.

Here's what I'll push that uses a `mark` element. For an invalid element:

image

For an invalid attribute:

image


image

PR #1452, opened 2018-09-22
Fix styling issue on Error Index page, where status icon is cut off

kienstra:
**Request For Code Review**

Hi @westonruter,
Could you please review this?

As @postphotos mentioned, on the **AMP Validation Error Index** page, the right part of this icon is cut off, like when the status is 'New Accepted'. 

So this make the column wider, to prevent this. 

This new styling only applies on this **AMP Validation Error Index** page. It's overridden by another rule on the single error URL page.

Before:
icon-cut-off

After (with this PR):
after-not-cut-off


westonruter:
It was still getting cut off on narrow screens, so the `min-width` is forcing it to not get cropped for me.

PR #1453, opened 2018-09-22
Fix output buffer fatal error due to not properly handling class method output callbacks

westonruter:
When activating the `all-in-one-seo-pack` plugin, an error shows up in the error log when validating a URL:

> PHP Fatal error:  ob_start(): Cannot use output buffering in output buffering display handlers

The underlying code in the plugin that causes the problem is the adding of the `All_in_One_SEO_Pack ::output_callback_for_title()` method as an output buffer callback in:

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/982cbb6a6ba183cf538d86b858273505a417720f/aioseop_class.php#L3496

The AMP plugin is supposed to prevent this error from happening via the `AMP_Validation_Manager::can_output_buffer()` method. It turns out there was a flaw in the logic being used to convert a function on the call stack into the format expected by `ob_list_handlers()`. In particular, it was not handling the case where an output callback was a class method. The issue is fixed by properly detecting when a call stack entry is for a class method.

This pull request also improves error message shown when a `validate_url` call fails. Instead of showing an error code like `response_comment_absent` it will now show a full message explaining what is going on. This change also adds WSOD (whitescreen of death detection) as a more specific error case for when the JSON comment is absent:

image

### DIFF
--- a/includes/options/class-amp-options-manager.php
+++ b/includes/options/class-amp-options-manager.php
@@ -463,11 +463,18 @@ class AMP_Options_Manager {
 			if ( is_wp_error( $validation ) ) {
 				$review_messages[] = esc_html( sprintf(
 					/* translators: %1$s is the error message, %2$s is the error code */
-					__( 'However, there was an error when checking the AMP validity for your site: %1$s (code: %2$s) ', 'amp' ),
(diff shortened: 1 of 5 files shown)


postphotos:
This one was fun to find. 🐞 Thanks for digging in, @westonruter!

PR #1458, opened 2018-09-24
Re-use styling for unmoderated comments to apply to new accepted/rejected validation errors

westonruter:
WordPress has a convention for marking items that need to be actioned on, namely for comments that have not been approved yet. This styling would seem to make sense for new validation errors as well:

image

This would help draw attention to the validation errors that are new which the user needs to action on. Now that the number of rows that have this styling would correspond to the error count appearing with Error Index admin menu item:

image

Aside: The styling above was actually implemented in https://github.com/Automattic/amp-wp/pull/1429/commits/58d03c987526fc8f1082b54052e412097e6dabce but it got lost in dead code that no longer was loaded (`amp-validation-error-detail-toggle.js`).

# TODO

- [x] Also apply the same styling to new validation errors appearing on the invalid URL screen, making sure that the style is removed as soon as a user changes a `select` to “Accepted” or “Rejected”.
- [x] Style invalid URLs in the post list table with unapproved style of contain new errors.


johnwatkins0:
Hi, @westonruter. @kienstra asked me to handle

> Also apply the same styling to new validation errors appearing on the invalid URL screen, making sure that the style is removed as soon as a user changes a select to “Accepted” or “Rejected”.

I pushed an update that applies the unmoderated comment styles to new errors in the single URL table (both the main row and the expanded details row when it's open). I'm removing the [WIP] from this PR now.

westonruter:
@johnwatkins0 I added one more todo. In the list of Invalid URLs, we should also show the same styling for Invalid URLs that _contain_ any new validation errors.

assets/js/amp-invalid-url-post-edit-screen.js
```diff
@@ -287,14 +287,32 @@ var ampInvalidUrlPostEditScreen = ( function() { // eslint-disable-line no-unuse
 	 * And sets this as the src of the status icon <img>.
 	 */
 	component.handleStatusChange = function handleStatusChange() {
-		const onChange = function( event ) {
+		const setRowStatusClass = function( { row, select } ) {
+			const statusText = select.options[ select.selectedIndex ].innerText.trim();
+
+			if ( statusText ) {
+				row.classList.toggle( 'new', 'New Rejected' === statusText || 'New Accepted' === statusText );
```
westonruter: @johnwatkins0 This will break when a different locale is active. Instead of looking at the `option` text, you should use the option _value_ which is an integer which will not change.

jacobschweitzer:
@westonruter I've updated this based on the new to-do and your latest comment. Could you take another look at this please? Thanks!

PR #1463, opened 2018-09-24
Fix title display of Invalid URL page

amedina:
Title of the Invalid URL admin screen with characters such as quotes did not displayed correctly.


westonruter:
This is required because the HTML entities are being encoded into JSON and the JS is updating the heading.

PR #1467, opened 2018-09-25
Fix incorrect attribution of theme as source for content validation errors

westonruter:
Given `post_content` such as:

```html
See post on Make XWP:

https://make.xwp.co/2018/09/24/amp-plugin-release-v1-0-beta4/

This is something: <button onclick="ok">food</button>
```

The Invalid URL screen currently displays sources as the following (the New Rejected errors are from the content here):

image

Note how they are incorrectly being attributed to “Twenty Sixteen (?)”. This was due to a change I made in https://github.com/Automattic/amp-wp/pull/1449/commits/3be57670dcf0953573707cb13b924a3696720895 from #1449. But it is clearly not right. The four validation errors should all be contributed to _the content_, not guessing the theme. 

This PR fixes that problem by listing the outermost hook running in the source stack, if available, and if no other source was obtained. So the result is this:

image

Note how each source now indicates `the_content` as the source.

PR #1468, opened 2018-09-25
Add source information for Embeds; discontinue extracting iframe from paragraphs

westonruter:
This implements https://github.com/Automattic/amp-wp/pull/1467#issuecomment-424493866.

Instead of showing invalid markup from oEmbeds as having the `the_content` source as per #1467, actually show that it is an “Embed”:

image

Also reveal the embed source information in the source stack, including the embed URL and its attributes:

```json
    {
        "embed": "https://make.xwp.co/2018/09/24/amp-plugin-release-v1-0-beta4/",
        "attr": {
            "width": 840,
            "height": 1000
        }
    }
```

In order to accomplish the above, we needed to undo an old PR https://github.com/Automattic/amp-wp/pull/112 which extracted an `amp-iframe` from a contained `p`:

https://github.com/Automattic/amp-wp/blob/c5ebe1f38ff58d58ee1cbdb735cd97bb14a316fe/includes/sanitizers/class-amp-iframe-sanitizer.php#L110-L112

The reason for why this was needed is apparently because of the iframe placeholder. A `div` placeholder would get extracted from the `p` by the HTML parser. So an easy fix is to switch this to being a `span` instead.

### DIFF
--- a/includes/sanitizers/class-amp-iframe-sanitizer.php
+++ b/includes/sanitizers/class-amp-iframe-sanitizer.php
@@ -103,18 +103,7 @@ class AMP_Iframe_Sanitizer extends AMP_Base_Sanitizer {
 				$new_node->appendChild( $placeholder_node );
 			}
 
-			$parent_node = $node->parentNode;
-			if ( 'p' !== strtolower( $parent_node->tagName ) ) {
-				$parent_node->replaceChild( $new_node, $node );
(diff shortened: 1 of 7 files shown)


westonruter:
Since the change to the `iframe` is potentially larger impact, this should wait for v1.1.

PR #1473, opened 2018-09-27
Add AMP icon if AMP is enabled on single error page

jacobschweitzer:


westonruter:
@jacobschweitzer Could you please add before/after screenshots in PRs like this?

jacobschweitzer:
@westonruter Of course. Here we are:

Before:
screen shot 2018-09-27 at 2 18 06 pm

After:
screen shot 2018-09-27 at 2 18 17 pm

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -2130,4 +2131,46 @@ public static function filter_bulk_post_updated_messages( $messages, $bulk_count
...
+	/**
+	 * Is AMP Enabled on Post
+	 *
+	 * @param WP_Post $post              Post object to check.
+	 * @param array   $validation_errors Validation errors.
+	 * @param array   $counts            Error counts.
+	 *
+	 * @return bool|void
+	 */
+	public static function is_amp_enabled_on_post( $post, $validation_errors = array(), $counts = array() ) {
```
westonruter: I think it would be better to eliminate the `$validation_errors` and `$counts` params. This simplifies the API. Any previously-obtained terms will already be stored in the object cache, so there is not really much concern about duplicated work.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -2130,4 +2131,46 @@ public static function filter_bulk_post_updated_messages( $messages, $bulk_count
...
+					$counts['new_accepted']++;
+					break;
+				case AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACK_ACCEPTED_STATUS:
+					$counts['ack_accepted']++;
+					break;
+				case AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACK_REJECTED_STATUS:
+					$counts['ack_rejected']++;
+					break;
+			}
+		}
```
westonruter: This `foreach` loop is duplicated with `display_invalid_url_validation_error_counts_summary`. I suggest adding a `protected` method for `count_invalid_url_validation_errors()` which you pass the `$validation_errors` array both here and in `display_invalid_url_validation_error_counts_summary`.

assets/js/amp-invalid-url-post-edit-screen.js
```diff
@@ -391,5 +393,17 @@ const ampInvalidUrlPostEditScreen = ( function() { // eslint-disable-line no-unu
 		}
 	};
 
+	/**
+	 * Adds the AMP icon to the page heading if AMP is enabled on this URL.
+	 */
+	component.showAMPIconIfEnabled = function() {
+		const heading = document.getElementsByClassName( 'wp-heading-inline' );
+		if ( heading[ 0 ] && true === component.data.l10n.amp_enabled ) {
+			let ampIcon = document.createElement( 'span' );
```
westonruter: This should use `const` not `let`.

assets/js/amp-invalid-url-post-edit-screen.js
```diff
@@ -391,5 +393,17 @@ const ampInvalidUrlPostEditScreen = ( function() { // eslint-disable-line no-unu
 		}
 	};
 
+	/**
+	 * Adds the AMP icon to the page heading if AMP is enabled on this URL.
+	 */
+	component.showAMPIconIfEnabled = function() {
+		const heading = document.getElementsByClassName( 'wp-heading-inline' );
```
westonruter: This can use `querySelector` instead of `getElementsByClassName`. So:

```js
const heading = document.querySelector( '.wp-heading-inline' );
```

Then below just check to see if `heading` is truthy as opposed to `heading[0]`

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -1539,15 +1538,17 @@ public static function enqueue_edit_post_screen_scripts() {
 	 */
 	public static function add_edit_post_inline_script() {
 		$current_screen = get_current_screen();
-		if ( 'post' !== $current_screen->base || self::POST_TYPE_SLUG !== $current_screen->post_type ) {
+		if ( 'post' !== $current_screen->base || self::POST_TYPE_SLUG !== $current_screen->post_type || ! isset( $_GET['post'] ) ) { // WPCS: CSRF OK.
 			return;
 		}
 
+		$post = get_post( intval( $_GET['post'] ) );  // WPCS: CSRF OK.
```
westonruter: I don't think this is necessary because the `$post` global will already be populated on the edit post screen. So instead of `$post` this you can just use `get_post()` below.

includes/validation/class-amp-invalid-url-post-type.php
```diff
@@ -2007,7 +2008,7 @@ public static function get_single_url_page_heading() {
 		}
 
 		/* translators: %s is the name of the page with the the validation error(s) */
-		return esc_html( sprintf( __( 'Errors for: %s', 'amp' ), $name ) );
+		return esc_html( sprintf( __( 'Errors for: %1$s', 'amp' ), $name ) );
```
westonruter: This can remain just `%s` because there is only one placeholder.

tests/validation/test-class-amp-invalid-url-post-type.php
```diff
@@ -1061,7 +1061,10 @@ public function test_render_link_to_error_index_screen() {
 	public function test_add_edit_post_inline_script() {
 		global $pagenow;
 
-		$pagenow = 'post.php';
+		$pagenow              = 'post.php';
+		$amp_invalid_url_post = $this->factory()->post->create_and_get( array( 'post_type' => AMP_Invalid_URL_Post_Type::POST_TYPE_SLUG ) );
+		$post                 = $this->factory()->post->create_and_get();
```
westonruter: Make this  `global` here.

includes/validation/class-amp-invalid-url-post-type.php
jacobschweitzer: Oh good catch, I tried adding the icon in here before but realized it was added as text via JS so it didn't work that way.

PR #1475, opened 2018-09-27
Update spec from revision 720 to 734

kienstra:
* Updates the AMP spec, simply from running `bash bin/amphtml-update.sh`
* No other edits
* Basic testing of this looks good so far, but we'll need full end-to-end testing, just like with any release


kienstra:
**Request For Review**

Hi @westonruter,
Could you please review this?

PR #1478, opened 2018-09-28
Improve definition and filtering of sanitizer default args

westonruter:
At the moment the default args for the sanitizers are not mode available for plugins to filter. This means, for example, if a plugin wants to augment default args they have to currently copy all of them:

```php
		// In case another filter already defined dynamic_element_selectors.
		! empty( $sanitizers['AMP_Style_Sanitizer']['dynamic_element_selectors'] ) ? $sanitizers['AMP_Style_Sanitizer']['dynamic_element_selectors'] : array(),
		// Duplicated from protected AMP_Style_Sanitizer::$DEFAULT_ARGS.
		array(
			'amp-list',
			'amp-live-list',
			'[submit-error]',
			'[submit-success]',
		),
		// Duplicated from protected AMP_Style_Sanitizer::$DEFAULT_ARGS.
		array(
			'#jp-relatedposts',
			'.jp-relatedposts',
			'.jp-relatedposts-items',
			'.jp-relatedposts-post',
		)
	);
	return $sanitizers;
} );
```

This is not ideal. This PR allows you to instead do:

```php
		$sanitizers['AMP_Style_Sanitizer']['dynamic_element_selectors'],
		array(
			'#jp-relatedposts',
			'.jp-relatedposts',
			'.jp-relatedposts-items',
			'.jp-relatedposts-post',
		)
	);
	return $sanitizers;
} );
```

The `AMP_Base_Sanitizer` class gets a new `get_default_args()` static method which is called up front when obtaining the list of sanitizers and their args. This then exposes the default args to the `amp_content_sanitizers` filter to customize.

This PR also deprecates the `DEFAULT_ARGS` class variable and consolidates the location of default args to be all defined in this `get_default_args()` methods.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -841,50 +841,37 @@ function amp_get_content_sanitizers( $post = null ) {
 		$post = null;
 	}
 
-	$parsed_home_url = wp_parse_url( get_home_url() );
-	$current_origin  = $parsed_home_url['scheme'] . '://' . $parsed_home_url['host'];
-	if ( isset( $parsed_home_url['port'] ) ) {
-		$current_origin .= ':' . $parsed_home_url['port'];
-	}
-
-	$sanitizers = [
-		'AMP_Core_Theme_Sanitizer'        => [
-			'template'   => get_template(),
-			'stylesheet' => get_stylesheet(),
-		],
-		'AMP_Img_Sanitizer'               => [
-			'align_wide_support' => current_theme_supports( 'align-wide' ),
-		],
-		'AMP_Form_Sanitizer'              => [],
-		'AMP_Comments_Sanitizer'          => [
-			'comments_live_list' => ! empty( $theme_support_args['comments_live_list'] ),
-		],
-		'AMP_Video_Sanitizer'             => [],
-		'AMP_O2_Player_Sanitizer'         => [],
-		'AMP_Audio_Sanitizer'             => [],
-		'AMP_Playbuzz_Sanitizer'          => [],
-		'AMP_Embed_Sanitizer'             => [],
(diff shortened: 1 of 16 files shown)


includes/amp-helper-functions.php
```diff
@@ -632,6 +632,32 @@ function amp_get_content_sanitizers( $post = null ) {
...
+		'AMP_Embed_Sanitizer',
+		'AMP_Iframe_Sanitizer',
+		'AMP_Gallery_Block_Sanitizer', // Note: Gallery block sanitizer must come after image sanitizers since itś logic is using the already sanitized images.
+		'AMP_Block_Sanitizer', // Note: Block sanitizer must come after embed / media sanitizers since it's logic is using the already sanitized content.
+		'AMP_Script_Sanitizer', // Note: This whitelist sanitizer must come at the end to clean up any remaining issues the other sanitizers didn't catch.
+		'AMP_Style_Sanitizer', // Always penultimate.
+		'AMP_Tag_And_Attribute_Sanitizer', // Always ultimate.
+	);
+
+	// @todo This doesn't help for classes that get added via amp_content_sanitizers. Does that matter?
```
westonruter: In practice this probably doesn't matter. If someone is adding a new sanitizer, they'd be doing so via a filter anyway. They would just be required to supply the defaults when they do so. For example:

```php
    require_once __DIR__ . '/class-sanitizer.php'
    return array_merge(
        $sanitizers,
        array(
            '\Foo\Sanitizer' => Sanitizer::get_default_args(),
        )
    );
} );
```

westonruter:
This should be considered as part of the whole refactor.

westonruter:
Closing this as stale, and the approach we take should probably be part of the modularization effort. See #4570.

PR #1481, opened 2018-10-01
[WIP] Add ETags to post-processor response with support for 304 status

westonruter:
We are already computing a cache key for the post-processor results to be stored in the object cache. We can use this same cache key also for the `ETag` of the response. Then a browser can serve the `ETag` value in its `If-None-Match` request header, and WordPress can short-circuit sending any response in favor of sendind `304 Not Modified`.

- [ ] Add some tests.

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1558,18 +1558,28 @@ class AMP_Theme_Support {
 			$args['enable_response_caching']                   = ! $disable_response_caching;
 		}
 
+		/*
+		 * Set response cache hash, the data values dictates whether a new hash key should be generated or not.
+		 * This is also used as the ETag.
+		 */
+		$response_cache_key = md5( wp_json_encode( array(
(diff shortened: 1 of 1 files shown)


includes/class-amp-theme-support.php
```diff
@@ -1558,18 +1558,28 @@ public static function prepare_response( $response, $args = array() ) {
...
+			self::$embed_handlers,
+			AMP__VERSION,
+		) ) );
+
+		// Handle responses that are cached by the browser.
+		if ( isset( $_SERVER['HTTP_IF_NONE_MATCH'] ) && $_SERVER['HTTP_IF_NONE_MATCH'] === $response_cache_key ) {
+			status_header( 304 );
+			return '';
+		}
+		AMP_HTTP::send_header( 'ETag', $response_cache_key );
```
westonruter: This header needs to be sent even when it is a 304 response.

Issue #1479, opened 2018-10-01
Replace WP admin pointers with Gutenberg popovers


westonruter:
**As a user, I should be able to view tooltips, regardless of my accessibility needs.**

- [ ] AC1: Investigate the level of effort to port the Gutenberg popover to the `amp` plugin.
- [ ] AC2: Replace the existing tooltip with the Gutenberg popover that behaves like a modal and plays more nicely with Assistive Technologies (ATs).

---

Per @GaryJones in https://github.com/Automattic/amp-wp/issues/1400#issuecomment-425732009:

> The work seems to be progressing on using the **legacy** pointers. Did the fact they are **not accessible**, and **discouraged**, get ignored? That they may not support keyboard-only users, or tell screen-reader users what the behaviour is?

And @afercia in https://github.com/Automattic/amp-wp/issues/1400#issuecomment-424232988:

> the legacy core "pointers" aren't accessible. They've been discussed a few times in the last years and their usage generally discouraged. Reading in the comments above:
>
>> It opens when the user hovers over the "?" icon 
>
>how about keyboard users? how about communicating to users the "?" icon opens a sort of modal? I'd rather look at the `popover` implementation in Gutenberg, where a toggle button with a proper aria-label and visible button tooltip opens a floating panel that behaves like sort of a modal.

postphotos:
Thanks @westonruter for breaking this out - I've added some ACs here (cc @alcurrie). 👍

westonruter:
This should be tackled with #2316.

jwold:
Weston > These popups are not the correct way to implement a tooltip from an accessibility perspective. Note that there are likely more considerations for accessibility. The primary focus for now though should be moving over to use the Gutenberg component that's responsible for tooltips (found in the edit post screen). That is part of the refactoring of these screens to be using React components. This will happen as part of that refactoring, and we may want to close this issue to say that it will be done as part of #2316.

Related to #1400.

PR #1483, opened 2018-10-03
Update the copy for the 'Details' tooltip

kienstra:
* Changes the 'Details' copy
* The previous text was a placeholder
* This will eventually be converted to a Gutenberg popover (#1479), but this change will apply to the `RC1` release

New 'Details' copy:
new-details-tooltip-copy

Existing 'Status' copy:
status-tooltip-unchanged

### DIFF
--- a/includes/validation/class-amp-invalid-url-post-type.php
+++ b/includes/validation/class-amp-invalid-url-post-type.php
@@ -830,7 +830,7 @@ class AMP_Invalid_URL_Post_Type {
 				esc_html__( 'Details', 'amp' ),
 				esc_attr( sprintf( '<h3>%s</h3><p>%s</p>',
 					esc_html__( 'Details', 'amp' ),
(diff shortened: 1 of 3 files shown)


kienstra:
**Question About Copy**

Hi @westonruter and @postphotos,
What do you think about the tooltip copy, as shown above?

*Details*
>The parent element of where the error occurred.

*Status (Not changed in this PR)*
>An accepted validation error is one that will not block a URL from being served as AMP; the validation error will be sanitized, normally resulting in the offending markup being stripped from the response to ensure AMP validity.

This probably isn't ideal, but maybe it's enough for `1.0`.

kienstra:
(Not ready to merge, needs the copy applied to the **AMP Validation Error Index** page)

kienstra:
Ready to merge when there's agreement on the copy.

kienstra:
Thanks, @westonruter! Ah, I should have opened this to `1.0`.

postphotos:
LGTM, thanks for the work here @kienstra! 🎉

PR #1486, opened 2018-10-03
Ensure font stylesheets are requested in CORS mode in both AMP and non-AMP documents

westonruter:
The AMP plugin is currently using the post-processor to add `crossorigin=anonymous` to font stylesheet links. This is not being done, however, in the non-AMP responses. This is a problem because when a service worker is caching the font stylesheet responses with CORS but then tries to serve it in response to a stylesheet being requested without CORS, then the fetch fails entirely with:

> The FetchEvent resulted in a network error response: an "opaque" response was used for a request whose type is not no-cors

So we need to make sure that such font URLs _always_ get served with CORS (`crossorigin="anonymous"`) when in AMP and not-AMP alike to prevent this problem from happening.

Another benefit of CORS is that it ensures that a service worker caching the external stylesheet will not inflate the storage quota. See: 
* https://developers.google.com/web/tools/workbox/guides/storage-quota#beware_of_opaque_responses
* https://developers.google.com/web/tools/workbox/guides/handle-third-party-requests#cross-origin_requests_and_opaque_responses

This PR is cherry-picking https://github.com/Automattic/amp-wp/pull/1261/commits/02fd0a7b04e29651e0f63d4a5b09ae22210a54e2 from #1261. We'll not ship the PWA plugin integration for 1.0, but this fix needs to be deployed first.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -492,6 +492,48 @@ function amp_filter_script_loader_tag( $tag, $handle ) {
 }
 
 /**
+ * Opt-in to CORS Mode for the font stylesheets.
+ *
+ * This ensures that a service worker caching the external stylesheet will not inflate the storage quota.
+ * This must be done in AMP and non-AMP alike because in paired mode the service worker could cache the
+ * font stylesheets in a non-AMP document without CORS (crossorigin="anonymous") in which case the
+ * service worker could then fail to serve the cached font resources in an AMP document with the warning:
+ *
+ * > The FetchEvent resulted in a network error response: an "opaque" response was used for a request whose type is not no-cors
+ *
+ * @since 1.0
+ * @link https://developers.google.com/web/tools/workbox/guides/storage-quota#beware_of_opaque_responses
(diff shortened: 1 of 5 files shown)


includes/amp-helper-functions.php
```diff
@@ -491,6 +491,48 @@ function amp_filter_script_loader_tag( $tag, $handle ) {
 	return $tag;
 }
 
+/**
+ * Opt-in to CORS Mode for the font stylesheets.
+ *
```
amedina: [Just a suggestion] We may simplify this comment by directly specifying the action: 

"Explicitly opt-in to CORS mode by adding the crossorigin attribute to font stylesheet links. This explicitly triggers a CORS request, and gets back a non-opaque response, ensuring that a service worker caching the external stylesheet will not inflate the storage quota."

kienstra:
**This Looks Good**

Hi @westonruter,
To echo @amedina, this PR looks good. As expected, a font stylesheet in Twenty Seventeen has `crossorigin="anonymous"` in both the AMP and non-AMP URLs:

crossorigin-anonymous-present

Issue #1498, opened 2018-10-09
Create new analytics migration document


davisshaver:
In the existing docs, users are recommend to filter `config_data`, but it looks like actually `config` is checked the custom key values.

Doc: https://github.com/Automattic/amp-wp/wiki/Analytics
Code: https://github.com/Automattic/amp-wp/blob/1.0/includes/amp-helper-functions.php#L568

Another issue is that the filter docs will result in the custom key-values being PHP array, but the filter still runs `json_decode`. A potential patch might be something like, `'config_data' => is_array( $entry['config'] ) ? $entry['config'] : json_decode( $entry['config'] )`.

davisshaver:
https://github.com/Automattic/amp-wp/pull/1499
https://github.com/Automattic/amp-wp/pull/1500

westonruter:
Can you please elaborate on the problem that you're facing? In particular, please share code that you are using which is not working as expected.

davisshaver:
I _think_ I see what’s going on now, could you review https://github.com/Automattic/amp-wp/pull/1499#issuecomment-428385899 and if you agree I can flesh out the docs part?

kienstra:
**Question About Filter**

Hi @davisshaver,
Thanks for bringing this up. You made a [good point](https://github.com/Automattic/amp-wp/pull/1500#discussion_r223864705):

>...that they could use a new section on filtering for native themes?

Like you mentioned, the [Analytics wiki page](https://github.com/Automattic/amp-wp/wiki/Analytics#manually) only documents [an analytics filter](https://github.com/Automattic/amp-wp/blob/469e99213d58d2af83915ca5924657648365bde7/includes/admin/functions.php#L161) for Classic mode (legacy templating).

Would it satisfy this issue and its PRs for me to update the [Analytics wiki page](https://github.com/Automattic/amp-wp/wiki/Analytics#manually) with a section on [how to filter](https://github.com/Automattic/amp-wp/blob/469e99213d58d2af83915ca5924657648365bde7/includes/amp-helper-functions.php#L558) analytics when there's theme support (Native and Paired mode)?

<img width="1058" alt="analytics-wiki-page-manually" src="https://user-images.githubusercontent.com/4063887/48172906-2dd36c80-e2c7-11e8-96dd-eb72f4a4f3ac.png">

Thanks, Davis! It was good to meet you in Chicago at the WordCamp.

davisshaver:
[...]

PR #1499, opened 2018-10-09
Don't try JSON decoding if the filtered analytics config data is already an array

davisshaver:
The [filter docs](https://github.com/Automattic/amp-wp/wiki/Analytics) will result in the custom key-values being PHP array, but the filter still runs `json_decode`. A potential patch might be something like, `'config_data' => is_array( $entry['config'] ) ? $entry['config'] : json_decode( $entry['config'] )`.

See #1498.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -565,7 +565,7 @@ function amp_get_analytics( $analytics = array() ) {
 		$analytics[ $entry_id ] = array(
 			'type'        => $entry['type'],
 			'attributes'  => array(),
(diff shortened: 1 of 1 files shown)


westonruter:
Please merge this into #1500 to have a single PR with your proposed changes for analytics.

westonruter:
There's a merge conflict for this PR. Also, it's not clear whether this is still an issue. If so, please open a new issue and open a new PR to address. Note also that we want the current analytics options to be deprecated in favor of dedicated analytics provider plugins, e.g. Site Kit. See #2317.

PR #1500, opened 2018-10-09
Check filtered analytics entry for `config_data` not `config` to match docs

davisshaver:
In the existing docs, users are recommend to filter `config_data`, but it looks like actually `config` is checked the custom key values.

Doc: https://github.com/Automattic/amp-wp/wiki/Analytics
Code: https://github.com/Automattic/amp-wp/blob/1.0/includes/amp-helper-functions.php#L568

See #1498..

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -408,7 +408,7 @@ function amp_get_analytics( $analytics = array() ) {
 		$analytics[ $entry_id ] = array(
 			'type'        => $entry['type'],
 			'attributes'  => array(),
-			'config_data' => json_decode( $entry['config'] ),
+			'config_data' => json_decode( $entry['config_data'] ),
 		);
 	}
 


includes/amp-helper-functions.php
```diff
@@ -408,7 +408,7 @@ function amp_get_analytics( $analytics = array() ) {
 		$analytics[ $entry_id ] = array(
 			'type'        => $entry['type'],
 			'attributes'  => array(),
-			'config_data' => json_decode( $entry['config'] ),
```
westonruter: Won't this cause all existing data to become invalid?

Also, is `config` ever populated? See:

https://github.com/Automattic/amp-wp/blob/a3983a0758dc5357592f6461c56e5345b78cacb3/includes/options/class-amp-options-manager.php#L206

davisshaver: One quick note, in both this interaction and the one earlier with Medina, I am coming away feeling quite discouraged by the knee-jerk response that I've either missed something (the Parsely AMP docs which I have known about for _years_) or misunderstood something (like the above with `config`). That's not the case. As with Gutenberg a disproportionate number of the commits come from people paid to be coding it.  I think a more formal contribution process would help, as currently messaging seems terse and close/decline decisions seem hasty.

westonruter: I'm sorry you feel discouraged. To be honest I have have trouble fully understanding your PRs, so I'm providing initial feedback on what I _do_ understand. In this PR, for example, there is no description provided.

Furthermore, you can see that your pull request failed the builds due to unit test errors:

```
1) AMP_Analytics_Options_Test::test_analytics_js_added
Undefined index: config_data
/tmp/wordpress/src/wp-content/plugins/amp/includes/amp-helper-functions.php:411
/tmp/wordpress/src/wp-content/plugins/amp/includes/admin/functions.php:127
/tmp/wordpress/src/wp-content/plugins/amp/includes/templates/class-amp-post-template.php:140
/tmp/wordpress/src/wp-content/plugins/amp/amp.php:327
/tmp/wordpress/src/wp-content/plugins/amp/tests/test-amp-analytics-options.php:64
/tmp/wordpress/src/wp-content/plugins/amp/tests/test-amp-analytics-options.php:135
```

I believe the failures are related to my code review comment here.

See unit test failures: https://travis-ci.org/Automattic/amp-wp/jobs/439292180#L284

westonruter:
You don't need to close the PR.

westonruter:
Code review comments requesting changes are given on most PRs, including PRs that I open. Please don't feel that a request for changes means that your PR should be unvalued.

includes/amp-helper-functions.php
davisshaver: It's okay, also I just experienced a weird Github bug where my response disappeared when you replied on the thread. I know you are having trouble understanding it, I was too. But I figured out what happened. It looks like this PR introduced a regression: https://github.com/davisshaver/amp-wp/commit/5dc9455c4fe301b1c61560fc8133b3b9108d1c3c

Essentially the filtered analytics data was previously expected to be in `config_data` and then run through `wp_json_encode()`, so that explains the two PR's.

So the questions are: 1) Should we introduce a patch so that it also checks `config_data`, as there are a number of filters in the wild that would throw a PHP notice currently, and 2) should we update the docs to reflect that `config` should be used instead of `config_data`?

Also just to clarify closing the PR wasn't a passive aggressive thing I had simply realized what was going on and I think we can close this PR in favor of https://github.com/Automattic/amp-wp/pull/1499

westonruter: My question in https://github.com/Automattic/amp-wp/pull/1500#discussion_r223827414 is that when the user adds analytics entries to the DB via the admin screen, the entries are stored in `config` keys:

https://github.com/Automattic/amp-wp/blob/a3983a0758dc5357592f6461c56e5345b78cacb3/includes/options/class-amp-options-manager.php#L206

So if you change this from `config` to `config_data` any existing analytics entries saved in the DB will fail to be served.

davisshaver: Example use in the wild, I think this would throw the PHP notice on a clean AMP WP install: https://github.com/WPPlugins/mediavine-control-panel/blob/b370dd4ad3b7aa52b299957dc1e4f4dafaa0b31a/extensions/amp/amp.php#L113-L129

I don't love it but something like this could work as a patch, probably into #1499 

```
$config = false;
if ( ! empty( $entry['config'] ) ) {
    $config = is_array( $entry['config'] ) ? $entry['config'] : json_decode( $entry['config'] );
} elseif ( ! empty( $entry['config_data'] ) ) {
    $config = is_array( $entry['config_data'] ) ? $entry['config_data'] : json_decode( $entry['config_data'] );
}
if ( ! empty( $config ) ) {
    $analytics[ $entry_id ] = array(
        'type'        => $entry['type'],
        'attributes'  => array(),
        'config_data' => $config,
    );
}
```

davisshaver: Apologies I think that this was also the wrong explanation... The docs were of `amp_post_template_analytics` and I am filtering `amp_analytics_entries`. In my testing I don't believe `amp_post_template_analytics` gets called on native pages. So rather than the docs being wrong, maybe it's better phrased that they could use a new section on filtering for native themes?

westonruter:
As with #1499, there's a merge conflict for this PR. Also, it's not clear whether this is still an issue. If so, please open a new issue and open a new PR to address. Note also that we want the current analytics options to be deprecated in favor of dedicated analytics provider plugins, e.g. Site Kit. See #2317.

PR #1503, opened 2018-10-11
[WIP] Add service worker streaming for AMP documents

westonruter:
This pull request explores what it would look like to improve performance of navigating a site through the use of streaming. See @jeffposnick's article [Beyond SPA](https://developers.google.com/web/updates/2018/05/beyond-spa) for some key inspiration here. The goal is to speed up the perceived performance for navigating a site by showing a meaningful interface and loading indicator immediately. In short, this should reduce the first contentful paint time when navigating an AMP site on origin.

The proposed way that this would work is as follows:

0. The user opts in to AMP streaming via theme support flag and adds a `amp_stream_boundary()` call to their theme's `header.php`. This function outputs a loading message to be displayed while the body is loading.
1. A new query var called `amp_stream_fragment` is added which can take one of two values: `header` and `body` (for example, https://example.com/?amp&amp_stream_fragment=header and https://example.com/?amp&amp_stream_fragment=body). This argument can be added to any URL, and when `header` is supplied the AMP plugin returns the document up until the `amp_stream_boundary()` call and when `body` is supplied the document after the `amp_stream_boundary()` call is returned.
2. The `header` fragment is precached by the service worker. Likely the `title` served with this fragment should also say “Loading...”. This header fragment is also served with an inline `script` containing am `ampStreamCombine()` function which is called later in the body fragment.
3. When the service worker handles a navigation request, it will immediately serve the header fragment which will be displayed while the browser is fetching the body fragment. While this is happening the loading indicator is shown.
4. The AMP plugin serves the body fragment with data extracted from what would have normally been included in an entire response. A JSON representation of the nodes in the `head` and the attributes shown on the `body` are included in this response. This data is then passed into the `ampStreamCombine` function in the head fragment which reconciles the differences between the precached header fragment. In Jeff's example linked above, he used this approach for reconciling the `document.title`. 
5. The `ampStreamCombine` function then deletes any traces of itself so that when the document finishes loading, the AMP runtime is none-the-wiser to the presence of custom scripts during the page load.
6. If the body fragment request fails, then a precached offline page body fragment would be served instead. Eventually, the body fragment responses should use a stale-while-revalidate strategy once there is a mechanism to communicate an update back to the user (see https://github.com/ampproject/amphtml/issues/18615#issuecomment-428006471)

This PR will depend on the [PWA plugin](https://github.com/xwp/pwa-wp/) to operate once it is ready for review.


westonruter:
Yes, we should add tests once 0.2 of the PWA plugin is on WordPress.org. It's hard to test now because the v0.2 of the PWA plugin isn't published for us to download during testing.

PR #1511, opened 2018-10-14
Improve validation of children of invalid elements

### LINKED ISSUE #1493: Unexpected validation Handling of `<details>` tag
The `<details>` HTML tag is not allowed by AMP's specification yet. It will likely be added soon but its current status allows us to see an seemingly unexpected behavior in validation handling. This tag is used as follows:

```html
<details>
  <summary>Copyright 1999-2018.</summary>
   <p>- by Awesome Author. All Rights Reserved.</p>
   <p>All content and graphics on this web site are the property of the author.</p>
</details>
```
Which results in the following view in the non-AMP version:

screen shot 2018-10-07 at 1 15 09 pm

On the AMP side the following takes place:

* The plugin correctly flags the HTML error and points out to the `<details>` tag as the offending markup.
screen shot 2018-10-07 at 1 15 36 pm
screen shot 2018-10-07 at 1 15 45 pm

* Upon accepting the error, the plugin is effectively instructed to remove the details tag
* Upon updating the post, the plugin removes the `details` tag and flags a new error associated with the `<summary>` tag

screen shot 2018-10-07 at 1 16 24 pm
screen shot 2018-10-07 at 1 16 34 pm

**Expected behavior**:
* Since the `summary` tag is a child of the  Details tag, the expected behavior is that removing the details tag should also remove the summary tag and therefore no error should be flagged for the latter once the former has been accepted

kienstra:
## Before
As @amedina reported in #1493, when there is an invalid element like `<details>` and the validation error isn't accepted, its child elements are not validated. 

Copying Alberto's test markup from #1493:

```php
<details>
        <summary>Copyright 1999-2018.</summary>
        <p>- by Awesome Author. All Rights Reserved.</p>
        <p>All content and graphics on this web site are the property of the author.</p>
</details>
```

There was only an error for the `<details>`, even though the nested `<summary>` is also invalid:
details-validation-error

---
On accepting the `<details>` validation error, the `<summary>` is also validated. But the `parent_name` of the `<summary>` should probably be `details`, as it's nested in `<details>`:

details-nested

## After
invalid-elements

## Approach
I'm not sure this is the right solution, especially in ensuring the `parent_name` is correct.

All of the existing unit tests passed, except for one where the `parent_name` is the original parent. Which I think is the expected result.

But adding the `->originalParent` to the `DOMElement` (or [DOMText](https://secure.php.net/manual/en/class.domtext.php)) isn't ideal, as it's [not one of its properties](https://secure.php.net/manual/en/class.domelement.php#domelement.props).

This will of course require full front-end testing in all 3 template modes, like we've done for the other releases.

Closes #1493

### DIFF
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -1807,19 +1807,12 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 			return;
 		}
 
-		/*
-		 * If node has children, replace it with them and push children onto stack
-		 *
-		 * Create a DOM fragment to hold the children
-		 */
-		$fragment = $this->dom->createDocumentFragment();
-
 		// Add all children to fragment/stack.
-		$child = $node->firstChild;
-		while ( $child ) {
-			$fragment->appendChild( $child );
-			$this->stack[] = $child;
-			$child         = $node->firstChild;
+		$children = array();
+		foreach ( $node->childNodes as $child ) {
+			$child->originalParent = $node->nodeName; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.NotSnakeCaseMemberVar
+			$children[]            = $child;
+			$this->stack[]         = $child;
 		}
 
 		// Prevent double-reporting nodes that are rejected for sanitization.
@@ -1830,6 +1823,23 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 		// Replace node with fragment.
 		$should_replace = $this->should_sanitize_validation_error( array(), compact( 'node' ) );
 		if ( $should_replace ) {
+			/*
+			 * If node has children, replace it with them and push children onto stack.
+			 *
+			 * Create a DOM fragment to hold the children.
+			 */
+			$fragment = $this->dom->createDocumentFragment();
+
+			/*
+			 * Add the child nodes to the $fragment just-in-time, as this changes their ->parentNode.
+			 * It can prevent child elements from being validated if they're added to a $fragment in the else block below.
+			 */
+			array_map(
+				function( $child ) use ( $fragment ) {
+					$fragment->appendChild( $child );
+				},
+				$children
+			);
 			$node->parentNode->replaceChild( $fragment, $node );
 		} else {
 			$this->should_not_replace_nodes[ $node->nodeName ][] = $node;
--- a/tests/amp-tag-and-attribute-sanitizer-private-methods-tests.php
+++ b/tests/amp-tag-and-attribute-sanitizer-private-methods-tests.php
@@ -713,6 +713,13 @@ class AMP_Tag_And_Attribute_Sanitizer_Attr_Spec_Rules_Test extends WP_UnitTestCa
 				),
 				'',
 			),
+			'nested_invalid_elements' => array(
+				array(
+					'source' => '<div><details><summary><p>Example Summary</p></summary><p>Example expanded text</p></details></div>',
+					'tag_name' => 'details',
+				),
+				'<div><summary><p>Example Summary</p></summary><p>Example expanded text</p></div>',
+			),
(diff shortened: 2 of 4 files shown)


includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1807,19 +1807,12 @@ private function replace_node_with_children( $node ) {
...
-		 *
-		 * Create a DOM fragment to hold the children
-		 */
-		$fragment = $this->dom->createDocumentFragment();
-
 		// Add all children to fragment/stack.
-		$child = $node->firstChild;
-		while ( $child ) {
-			$fragment->appendChild( $child );
-			$this->stack[] = $child;
```
kienstra: Here, all children were appended to `$fragment`. Even though `$fragment` is only used if `$should_replace`. 

So I think this caused the nodes in `$this->stack` to have no values, and thus not get validated:

this-stack

As a workaround, this stores the nodes in `$children`, and only appends them to `$fragment` if `$should_replace`.

tests/test-tag-and-attribute-sanitizer.php
```diff
@@ -1203,7 +1215,7 @@ public function test_replace_node_with_children_validation_errors() {
 		);
 		$expected_errors[] = array(
 			'node_name'       => 'foo',
-			'parent_name'     => 'body',
+			'parent_name'     => 'divs',
```
kienstra: It seems like the `parent_name` should be `divs`, as that's the parent of `<foo>`. But feel free to offer a different opinion.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1807,19 +1807,12 @@ private function replace_node_with_children( $node ) {
...
-			$child         = $node->firstChild;
+		$children = array();
+		foreach ( $node->childNodes as $child ) {
+			$child->originalParent = $node->nodeName; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.NotSnakeCaseMemberVar
```
kienstra: Without keeping track of the original parent somehow, children of the stripped element will have a different parent reported.

For example, when this is validated:
```html
<div><invalid-element><inner-invalid></inner-invalid></invalid-element></div>
```

This wil strip `<invalid-element>`:
```html
<div><inner-invalid></inner-invalid></div>
```

So `<inner-invalid>` will have a reported parent of `<div>`, even though its original parent was `<invalid-element>`.

tests/amp-tag-and-attribute-sanitizer-private-methods-tests.php
```diff
@@ -713,6 +713,13 @@ public function get_replace_node_with_children_data() {
 				),
 				'',
 			),
+			'nested_invalid_elements' => array(
+				array(
+					'source' => '<div><details><summary><p>Example Summary</p></summary><p>Example expanded text</p></details></div>',
```
westonruter: Now that AMP allows the `details` element (via https://github.com/ampproject/amphtml/pull/18440), the tests here aren't relevant now.

tests/test-tag-and-attribute-sanitizer.php
westonruter: You're right. The same change was required in https://github.com/ampproject/amp-wp/pull/3243/commits/f084b43156c89856ef8dd0f2dcb8e34d20fe71f0 for #3243.

tests/test-tag-and-attribute-sanitizer.php
```diff
@@ -1246,6 +1258,21 @@ public function test_replace_node_with_children_validation_errors() {
 			'node_attributes' => array(),
 		);
 
+		// Both <details> and <summary> are invalid, so there should be errors for both.
+		$content[]         = '<div><details><summary><p>Example Summary</p></summary><p>Example expanded text</p></details><div>';
```
westonruter: As noted above, these elements are no longer invalid. See https://github.com/ampproject/amphtml/pull/18440.

westonruter:
Closing in favor of #3243

PR #1519, opened 2018-10-22
Add support for app shell navigation via AMP Shadow DOM

westonruter:
Note: This is somewhat of a prototype for what could be supported in AMP itself as seamless page transitions: https://github.com/ampproject/amphtml/issues/12981 / https://github.com/ampproject/amphtml/issues/14738

----

For a theme to support app shell via the AMP plugin, the bare minimum that needs to be done is:

0. Activate on [PWA plugin](https://wordpress.org/plugins/pwa).
1. Ensure your theme (and plugins) work entirely in the AMP plugin's [Transitional mode](https://amp-wp.org/documentation/how-the-plugin-works/amp-plugin-serving-strategies/). You must use Transitional mode as opposed to Standard mode for now. Make sure that you have “Serve all templates as AMP regardless of what is being queried” enabled.
2. Identify the element that contains the markup that changes as you navigate from page to page and wrap it with two function calls to indicate where the app shell container is:
```php
<?php amp_start_app_shell_content(); ?>
    <?php get_template_part( 'content' ); ?>
<?php amp_end_app_shell_content(); ?>
```
3. Opt-in to AMP theme support for app shell by adding the following to your `functions.php`:

```php
    'paired' => true,
    'app_shell' => true,
) );
```
4. You should define an `offline.php` template in your theme.
5. [Deprecated] ~You should mark for precaching any scripts and stylesheets needed by the app shell. This can be done via:~
```php
wp_style_add_data( 'foo-special-colors', 'precache', true );
wp_script_add_data( 'bar-special-behavior', 'precache', true );
```

For theme which adds support for AMP app shell see https://github.com/xwp/twentyseventeen-westonson. Note that this theme copies some code from the Twenty Seventeen parent theme in order to add make the necessary modifications. For example, the `js/global.js` file modified as per [this diff](https://gist.github.com/westonruter/b9d7952c0879ea3cda9e0081e387846d). The theme has app shell behind a theme mod flag, so make sure you do `wp theme mod set service_worker_navigation amp_app_shell` via WP-CLI. See also how it [adds skeleton content](https://github.com/xwp/twentyseventeen-westonson/blob/03934b7328f9b18dd979ae97458ad38feb6f0e1a/functions.php#L241-L283) to the app shell. See also plugin to [enable stale-while-revalidate caching strategy for navigation requests](https://gist.github.com/westonruter/f848013108672568be6dcde153f9ca37).

This was presented at CDS 2018; see [related portion of talk](https://youtu.be/s1WrBaAyzAI?t=940).

For a demo of the app shell theme in action, see https://dev-showcase-pwa.pantheonsite.io/

See also this short screen cast: https://youtu.be/oAiIbhHdoXM

For some more background on this, see https://github.com/xwp/pwa-wp/issues/12#issuecomment-401843173

# Todo

- [x] Identify the element that contains the content
- [x] Add a query var which is used to obtain the document without the content being populated; add another query var which is used to omit the content for the shell response. Make sure this is done before sanitizers run.
- [x] Preserve admin bar in shadow DOM.
- [x] Add dedicated element which contains the content regardless of the id.
- [x] Ensure all `style` and `link` elements get preserved during content isolation.
- [x] Add Shadow DOM JS in `outer` shell request.
- [x] Let shadow root element be direct child of content element. This avoids styling problems.
- [x] Synchronize `title`.
- [x] Synchronize `body` classes.
- [x] Synchronize nav menu item classes.
- [x] Add JS which is responsible for history management, navigation, fetching, and invoking Shadow DOM API. Intercept the clicks on links in the shell and in the shadow to then load the new shadow doc.
- [x] Bootstrap spa onto frontend, automatically swap out content with shadow DOM AMP doc when navigating to another page? This handles case of smooth transition when navigating after landing on a site for the first time before the service worker is installed. Afterward, the app she would be served instead.
- [x] Fix lack of trailing slash causes redirect which drops the required app shell query param.
- [x] Trigger an event on the document when a page is transitioned.
- [x] Transfer admin bar from shadow to shell document.
- [x] Use AMP Shadow streaming API.
- [x] Update the class name for the menu item the instant that it is clicked.
- [x] ~Flag that the [deprecation of the non-streaming API](https://github.com/ampproject/amphtml/blob/master/spec/amp-shadow-doc.md#fetching-and-attaching-shadow-docs) would break this.~
- [x] Use async external script. Include in precache.
- [x] ~When declaring theme support, include the script handles to print in addition to shadow API script.~
- [x] Make sure that the AMP runtime does not conflict with our custom script which intercept clicks.
- [x] When content is extracted in shell, make sure that shell serves skeleton initially.
- [x] Make sure app shell has revision. Vary cache by navigation menus, theme mods, etc.
- [x] Figure out injecting into service worker for navigation. Disable preload and then override entirely the URL that is served. 
- [x] Make sure app shell is precached, as well as offline and error inners
- [x] Implement two methods of integrating app shell: bootstrapping app shell onto SW-unmanaged page, and another where app shell loads the content from the start.
- [ ] Synchronize meta (`link`, `meta`, JSON-LD, etc)
- [ ] Omit admin bar from shell? Or rather, prevent app shell entirely if user is logged-in?
- [ ] Allow shadow doc request to respond with 400 if amp is disabled and so then cause navigation at top? No, just do not follow redirects from `?amp` to not-`?amp`. If such a redirect happens, navigate the entire page.
- [ ] Stale while revalidate will be able to update. This can be done even on streaming by leaving script in the page which listens for broadcast update.
- [ ] Make sure that the app shell serves the appropriate initial body class based on whether or not the homepage is initially requested.
- [ ] Preserve amp-analytics in shadow DOM; analytics needs to be excluded from app shell.
- [ ] When requesting app shell contents force accepting sanitization
- [ ] How can we handle a case where a user navigate to a URL that has amp disabled.  Add a fury fragment to indicate that the service worker should bypass the use of the app shell
- [ ] Scroll to content of it is out of view on mobile.
- [ ] Prevent navigation when `#` anchor link.
- [ ] Add support for navigating to anchors on other URLs.
- [ ] Download web components API JS to serve locally.


westonruter:
Note that travis is failing due to Gutenberg being merged into core.

assets/js/amp-service-worker-offline-commenting.js
```diff
@@ -0,0 +1,45 @@
...
+					method: event.request.method,
+					headers: event.request.headers,
+					mode: event.request.mode,
+					credentials: event.request.credentials,
+					referrer: event.request.referrer,
+					redirect: 'follow',
+					body: body
+				} );
+
+				// Add request to queue. @todo Replace when upgrading to Workbox v4!
```
westonruter: @miina Is this still valid?

miina: @westonruter I believe so, that was related to https://github.com/GoogleChrome/workbox/pull/1710 specifically. Looks like it got merged meanwhile.

westonruter: It's required sure enough. The offline commenting is currently failing:

image

westonruter: Fixed in https://github.com/Automattic/amp-wp/pull/1519/commits/1e77c6e56f3ad0b8a39a988f4aab409bb646e326 and https://github.com/xwp/pwa-wp/pull/109

westonruter: Correction: https://github.com/Automattic/amp-wp/pull/1519/commits/ee465e718f52f6ea90d102a9b033240aa24505a6

westonruter:
We'll need to merge the updates to the underlying service worker code in #1261, as it will be merged before this is.

mehigh:
@miina 
I've handled the 8 conflicts and issued a PR against this branch here which handles them:
https://github.com/ampproject/amp-wp/pull/3931 
This is an effort towards getting it closer to being merged.

assets/js/amp-wp-app-shell.js
```diff
@@ -0,0 +1,409 @@
...
+				const newContainer = document.createElement( oldContainer.nodeName );
+				newContainer.setAttribute( 'id', oldContainer.getAttribute( 'id' ) );
+				oldContainer.parentNode.replaceChild( newContainer, oldContainer );
+
+				/*
+				 * For more on this, see:
+				 * https://www.ampproject.org/latest/blog/streaming-in-the-shadow-reader/
+				 * https://github.com/ampproject/amphtml/blob/master/spec/amp-shadow-doc.md#fetching-and-attaching-shadow-docs
+				 */
+				currentShadowDoc = AMP.attachShadowDocAsStream( newContainer, url, {} );
```
dero: @westonruter The `url` parameter here comes through as an [URL object](https://developer.mozilla.org/en-US/docs/Web/API/URL) when the navigation happens through a click or form submit, but the method [expects](https://github.com/ampproject/amphtml/blob/f41de5e5205ae37a823e774af817e2748ad5ed09/src/multidoc-manager.js#L273) and [newly enforces](https://github.com/ampproject/amphtml/commit/f41de5e5205ae37a823e774af817e2748ad5ed09) the parameter to be a `string`.

This newly leads to an error being raised now, which breaks regular navigation and potentially form submissions.

westonruter: So this should be `url.toString()`?

westonruter: Feel free to open a PR to this branch!

dero: Opened https://github.com/ampproject/amp-wp/pull/4261

includes/amp-helper-functions.php
```diff
@@ -1228,6 +1247,63 @@ function amp_wp_kses_mustache( $markup ) {
...
+	if ( ! isset( $support_args['app_shell'] ) ) {
+		return;
+	}
+
+	// Clean output buffer if requesting outer shell, since all content will be omitted from the response.
+	if ( 'outer' === AMP_Theme_Support::get_requested_app_shell_component() ) {
+		ob_end_clean();
+	}
+
+	echo '</div><!-- #amp-app-shell-content -->';
```
delawski:
```suggestion
	printf( '</div><!-- #%s -->', esc_attr( AMP_Theme_Support::APP_SHELL_CONTENT_ELEMENT_ID ) );
```

includes/class-amp-theme-support.php
```diff
@@ -190,13 +205,16 @@ public static function init() {
 			}
 
 			add_action( 'widgets_init', [ __CLASS__, 'register_widgets' ] );
+			add_action( 'parse_query', [ __CLASS__, 'init_app_shell' ], 9 );
 
 			/*
 			 * Note that wp action is use instead of template_redirect because some themes/plugins output
 			 * the response at this action and then short-circuit with exit. So this is why the the preceding
```
delawski:
```suggestion
			 * Note that wp action is used instead of template_redirect because some themes/plugins output
			 * the response at this action and then short-circuit with exit. So this is why the preceding
```

includes/class-amp-theme-support.php
```diff
@@ -456,6 +474,70 @@ static function() {
...
+		// When inner app shell is requested, it is always an AMP request. Do not allow AMP when getting outer app shell for now (but this should be allowed in the future).
+		if ( 'outer' === $requested_app_shell_component ) {
+			add_action(
+				'template_redirect',
+				function() {
+					if ( ! is_amp_endpoint() ) {
+						return;
+					}
+					wp_die(
+						esc_html__( 'Outer app shell can only be requested of the non-AMP version (thus requires paired mode).', 'amp' ),
```
delawski:
```suggestion
						esc_html__( 'Outer app shell can only be requested of the non-AMP version (thus requires Transitional mode).', 'amp' ),
```

PR #1520, opened 2018-10-22
Fix integration with WordPress 5.0

miina:
This PR replaces relying on `gutenberg_init` method with relying on `register_block_type` method.

It's likely that `gutenberg_init` will be removed once the plugin gets merged into core so perhaps it makes sense to not depend on it?


westonruter:
This is good. However, it should be rebased onto the `1.0` branch since it needs to be part of 1.0.

westonruter:
See corresponding dev-lib PR: https://github.com/xwp/wp-dev-lib/pull/284

PR #1527, opened 2018-10-23
Fixes Gallery Block Sanitizer carousel logic

christianc1:

PR #1529, opened 2018-10-24
Address issue where <ul> is converted to an <amp-carousel>

### LINKED ISSUE #1528: AMP Gallery Block Sanitizer transforming all <ul>'s to carousels.
In cc53173, logic was added (from the commit message for back-compat), setting the `AMP_Gallery_Block_Sanitizer::$args['carousel_required']` to true.  This affects all installations that have not explicitly declared some sort of `add_theme_support` for amp.

An issue arises as the logic in `AMP_Gallery_Block_Sanitizer::sanitize` allows through all `<ul>` nodes.  Further in the method, there is a check to see if there are child `<a>` nodes and if there are, pushes them onto an array of images which are then appended to the carousel.

So there are two problems.  

1. As far as I can tell, all `<ul>`'s will be will be transformed to galleries if one of their items contain a link, unless `add_theme_support` is declared.
2. Given the above, any content that is not an image or a link or a descendent of one, is stripped and not rendered to the app template.

Admittedly I'm a bit fuzzy as to the full context here, but I opened https://github.com/Automattic/amp-wp/pull/1527.  If someone can fill in some blanks, I can help contribute a better fix if this PR is missing something.

kienstra:
Thanks to @christianc1 for the detailed report of this in #1528.

# Steps To Reproduce
1. In AMP Settings > Template Mode, select 'Classic'
2. Create a post using Gutenberg
3. Add a Custom HTML block with:
```html
<ul><li><a href="#">click me</a></li></ul>
```
4. Save and go to the AMP URL for the post
5. Expected: this markup shouldn't be changed on the AMP URL
6. Actual: this is wrapped in an `<amp-carousel>`, though it's not a gallery block:

classic-mode-carousel

# Background
[AMP_Gallery_Block_Sanitizer](https://github.com/Automattic/amp-wp/blob/60bf71fa1aa061e23d539a24b00f3e847a4154fb/includes/sanitizers/class-amp-gallery-block-sanitizer.php) iterates through the `<ul>` tags [that have child nodes](https://github.com/Automattic/amp-wp/blob/60bf71fa1aa061e23d539a24b00f3e847a4154fb/includes/sanitizers/class-amp-gallery-block-sanitizer.php#L76).

As @christianc1 mentioned, if this is in Classic mode, the sanitizer [will have](https://github.com/Automattic/amp-wp/commit/cc53173f92f166bc557d37954dece76b6ca651d9#diff-5e173f2331b78ccd3da05258383dcc55R493) the argument of `array( 'carousel_required' => true )`. 

This [causes](https://github.com/Automattic/amp-wp/blob/60bf71fa1aa061e23d539a24b00f3e847a4154fb/includes/sanitizers/class-amp-gallery-block-sanitizer.php#L82) `$is_amp_carousel` to be true, even for something like `<ul><li></li></ul>`.

# Approach
Assuming that the `AMP_Gallery_Block_Sanitizer` is only intended for the Gallery block, restricting it to that block would probably make sense.

It looks like the Gallery block should always have the class of `wp-block-gallery`. Its stylesheet [only uses that class](https://github.com/WordPress/gutenberg/blob/d4e097c14e3841994fe43385f52c39d6e3492557/packages/block-library/src/gallery/style.scss#L1), and the image selector [begins with](https://github.com/WordPress/gutenberg/blob/d4e097c14e3841994fe43385f52c39d6e3492557/packages/block-library/src/gallery/index.js#L24) `ul.wp-block-gallery`.

So this restricts the [Gallery block sanitizer](https://github.com/Automattic/amp-wp/blob/60bf71fa1aa061e23d539a24b00f3e847a4154fb/includes/sanitizers/class-amp-gallery-block-sanitizer.php) to only `ul.wp-block-gallery`, and it simply exits if the class isn't present:

ul-click-me

Closes #1528.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ install:
   - nvm install 8.11.4 && nvm use 8.11.4
   - export DEV_LIB_PATH=dev-lib
   - source $DEV_LIB_PATH/travis.install.sh
-  - mkdir -p $WP_CORE_DIR/src/wp-includes/css/dist/block-library/ && touch $WP_CORE_DIR/src/wp-includes/css/dist/block-library/style.css
 
 script:
   - source $DEV_LIB_PATH/travis.script.sh


kienstra:
**Pending Unit Test Passing**

If it's alright, I'll look at the [failed unit test](https://travis-ci.org/Automattic/amp-wp/jobs/445497877) tomorrow. It passes locally.

westonruter:
It looks like the failed test may be due to something new introduced in WordPress 5.0. 

How does this PR relate to #1527?

kienstra:
Hi @westonruter,
This takes a different approach than #1527, though we mights still consider #1527.

I think there's something needed to ensure the `AMP_Gallery_Block_Sanitizer` only operates on Gallery blocks. Like how this PR [checks](https://github.com/Automattic/amp-wp/pull/1529/files#diff-3f3a86f42d679f81a48116543bcb3b38R86) that the `<ul>` has the class of `wp-block-gallery`.

PR #1533, opened 2018-10-25
🐛 FIX: Show error when DOM extension is not installed #1249

### LINKED ISSUE #1249: Short circuit plugin init and show error when DOM extension is not installed
The plugin currently short-circuits with an admin error notice when PHP>=5.3 is not present. We should do the same when the DOM extension is not installed. This will avoid this error from being raised at runtime: https://wordpress.org/support/topic/error-5xx-on-amp-pages/

AdelDima:
Print admin error notice "The AMP plugin requires DOM extension in PHP. Please contact your host to install DOM extension." when DOMDocument class does not exist.

Fixes #1249.


amp.php
```diff
@@ -30,6 +30,23 @@ function _amp_print_php_version_admin_notice() {
 	return;
 }
 
+/**
+ * Print admin notice regarding DOM extension is not installed
+ *
+ * @since 0.7.2
```
westonruter: This should be changed to `1.1` if merging into `develop`. However, if you rebase your commits onto the `1.0` branch and change the PR to point to the `1.0` branch, then the version can be changed to `1.0` since this would be good to include sooner.

amp.php
```diff
@@ -30,6 +30,23 @@ function _amp_print_php_version_admin_notice() {
 	return;
 }
 
+/**
+ * Print admin notice regarding DOM extension is not installed
```
westonruter: Minor: Add period to end of summary.

westonruter:
The build test failures aren't your fault.

PR #1534, opened 2018-10-25
Update regex for tag selectors

### LINKED ISSUE #1513: Legit CSS selector renamed
In my theme I have some code like this (simplified):

HTML:

```html
<body class="home">
  <div class="content-area">
    <article id="post-1234" class="post-1234 video type-video status-publish has-post-thumbnail hentry">
      <div class="entry-video">
        <iframe width="400" height="225" src="https://www.youtube.com/...."></iframe>
      </div>
    </article>
  </div>
</body>
```

CSS:

```css
.home .content-area .video {
  border-top: 5px solid red;
  padding: 21px .5em .5em;
  background: white;
}
```

With AMP active, it looks like this:

```html
<body class="home">
  <div class="content-area">
    <article id="post-1234" class="post-1234 video type-video status-publish has-post-thumbnail hentry">
      <div class="entry-video">
        <amp-youtube data-videoid="..." layout="responsive" width="600" height="338"></amp-youtube>
      </div>
    </article>
  </div>
</body>
```

CSS:

```css
.home .content-area .amp-video,
.home .content-area .amp-youtube {
  border-top: 5px solid red;
  padding: 21px .5em .5em;
  background: white;
}
```

As a result, my video container is unexpectedly unstyled when using AMP.

I've now fixed this by using `.type-video` instead of `.video` in my selector as a workaround.

However, I'm curious why this happens in the first place. Is `video` some kind of reserved keyword that's documented somewhere?

swissspidy:
This prevents cases where legit CSS selectors were wrongly renamed.

Props @westonruter for the fix 🙂 

Fixes #1513.

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -1982,13 +1982,17 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 				continue;
 			}
 
+			// An element (type) either starts a selector or is preceded by combinator or opening paren.
+			$before_type_selector_pattern = '(?<=^|\(|\s|>|\+|~)';
+			$after_type_selector_pattern  = '(?=$|[^a-zA-Z0-9_-])';
+
 			$edited_selectors = array( $selector );
 			foreach ( $this->selector_mappings as $html_selector => $amp_selectors ) { // Note: The $selector_mappings array contains ~6 items.
-				$html_pattern = '/(?<=^|[^a-z0-9_-])' . preg_quote( $html_selector, '/' ) . '(?=$|[^a-z0-9_-])/i';
+				$html_pattern = '/' . $before_type_selector_pattern . preg_quote( $html_selector, '/' ) . $after_type_selector_pattern . '/i';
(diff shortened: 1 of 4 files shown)


westonruter:
We'll need to add some tests.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1969,7 +1969,7 @@ private function ampify_ruleset_selectors( $ruleset ) {
 				foreach ( $edited_selectors as &$edited_selector ) { // Note: The $edited_selectors array contains only item in the normal case.
 					$original_selector = $edited_selector;
 					$amp_selector      = array_shift( $amp_selectors );
-					$amp_tag_pattern   = '/(?<=^|[^a-z0-9_-])' . preg_quote( $amp_selector, '/' ) . '(?=$|[^a-z0-9_-])/i';
+					$amp_tag_pattern   = '/(?<=^|\s|>)' . preg_quote( $amp_selector, '/' ) . '(?=$|[^a-z0-9_-])/i';
```
westonruter: Also need to account for other possible characters which can precede an element name. I'm thinking in particular:

* Adjacent sibling combinator (`+`)
* General sibling combinator (`~`)

Once PHP-CSS-Parser is updated to actually parse the components of a selector, we can be smarter about the replacements instead of using regular expressions.

PR #1568, opened 2018-10-30
Serve legacy post template CSS as minified

paulschreiber:


paulschreiber:
Should I open an issue for:
>we should output-buffer the CSS and then send it through the AMP_Style_Sanitizer do the minification. Or rather, the PHP-CSS-Parser can be used to do the minification without invoking the full sanitizer.

westonruter:
If you want. Or you can use this PR for that purpose.

westonruter:
Note that output-buffering and minifying the entire legacy post template CSS would then also include any CSS added via the `amp_post_template_css` action, not just the CSS coming from the AMP plugin (which you minified here).

westonruter:
Closing for now since a different approach will be needed.

Issue #1583, opened 2018-11-01
Add CSS code coverage analysis as part of validation process


westonruter:
The CSS tree shaking does not eliminate all rules which do not apply to the current document. While the effectiveness can be improved (e.g. #1492) but it should never achieve 100% unused rule removal. CSS tree-shaking needs to be done at runtime, and the more sophisticated the tree-shaking the slower the time will be for the post-processor to generate the response. A fast tree shaking strategy is to remove selectors that mention an ID that does not exist in the page. Another relatively fast tree shaking strategy is to remove any selectors that mention a class name that is not used in any elements on the page. Expensive tree shaking strategies would include matching combinations of selector parts, especially rules that relate multiple elements together (ancestors, children, siblings, etc).

Something that comes to mind is that we could extend the CSS tree shaking to attempt to shake 100% of the unused CSS rules, but we could limit which tree shaking strategies are employed by default. We could then allow the employed strategies to be configurable so that if you are willing to take the additional performance penalty, then you can do the deeper tree shaking.

In any case, having comprehensive tree shaking strategies would be useful for adding CSS code coverage to the validation results. We could show a warning on a Validated URL screen to show how many of the CSS rules are actually used. Additionally, the source stylesheet should be identified for each rule.

A dependency for achieving this is for the PHP-CSS-Parser to add parsing of CSS selectors: https://github.com/sabberworm/PHP-CSS-Parser/pull/138#issuecomment-418193262 and https://github.com/sabberworm/PHP-CSS-Parser/pull/146.

With that in place, there would be a need to convert CSS selectors into the corresponding XPath queries so that they can be evaluated on a given document.

westonruter:
I don't think this is going to be feasible. Better alternatives to this are using code coverage in  DevTools (#1584), and also to implement CSS sourcemaps (#2917).

PR #1584, opened 2018-11-01
Amend AMP style elements with sourceURL comment for DevTools

westonruter:
When stylesheets include a `sourceURL` comment at the end then they will be listed in DevTools as sources which can by analyzed for CSS code coverage:

image

This change is important because it allows developers to identify which CSS rules specifically need to be removed. As discussed in #1583, the CSS tree shaker in the plugin is not able to be very sophisticated in how deeply it shakes the tree because it must be done at runtime.

So this relates to #1583, but it re-uses DevTools to do the code coverage instead of having to re-implement the functionality on the PHP side. This is not a replacement, however, because the code coverage does not indicate the specific stylesheet sources for where the unused CSS rules are coming from.

Thanks to @paulirish for the tip.

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -1760,6 +1760,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 
 		$stylesheet_sets = array(
 			'custom'    => array(
+				'source_map_comment'  => "\n\n/*# sourceURL=amp-custom.css */",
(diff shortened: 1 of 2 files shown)


kienstra:
**Approved**

Hi @westonruter,
Wow, this'll be really helpful for tree-shaking.

Just like your [screenshot above](https://github.com/Automattic/amp-wp/pull/1584#issue-227790400), I also saw that the coverage appears for `<style amp-custom>`:

amp-custom-coverage

PR #1586, opened 2018-11-02
Add default values for AMP Timeago block

### LINKED ISSUE #1580: Implement unique constraint in whitelist sanitizer
Went to https://native-ampconfdemo.pantheonsite.io/

Got the attached amp validation errors:
image

kienstra:
Thanks to @csossi for reporting this in #1580. 

Creating an AMP Timeago block with the default values causes a validation error in the Chrome extension because it doesn't have the required `height` value:

height-value-required

So this sets a default value of `20` for the height. This is the value in [this example](https://www.ampproject.org/ko/docs/reference/components/amp-timeago#behavior).

It also sets the default layout to `fixed-height`. This is the only layout available that doesn't require a `width` (there's no default width in the editor):

new-default-values

Fixes #1580

### DIFF
--- a/blocks/amp-timeago/index.js
+++ b/blocks/amp-timeago/index.js
@@ -58,6 +58,7 @@ export default registerBlockType(
 			},
 			ampLayout: {
 				type: 'string',
+				default: 'fixed-height',
 				source: 'attribute',
 				selector: 'amp-timeago',
 				attribute: 'layout'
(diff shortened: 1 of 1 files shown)


kienstra:
**Request For Code Review**

Hi @westonruter,
Could you please review this? 

Thanks, and have a great weekend.

PR #1592, opened 2018-11-07
Do not show fallback source as active theme if no validation errors

westonruter:
There is a bug in `AMP_Validated_URL_Post_Type::render_sources_column()` where it will render the active theme with a `(?)` after it (via https://github.com/Automattic/amp-wp/commit/3be57670dcf0953573707cb13b924a3696720895) when it can't find the source for an error, since in that case it is normally in the theme template itself. However, this was incorrectly showing the active theme when there were no validation errors at all to display in the summary at all. So this fix short-circuits the source summary display in this case.

# Before

screen shot 2018-11-07 at 9 28 47 am

# After

screen shot 2018-11-07 at 9 29 45 am

### DIFF
--- a/includes/validation/class-amp-validated-url-post-type.php
+++ b/includes/validation/class-amp-validated-url-post-type.php
@@ -934,6 +934,12 @@ class AMP_Validated_URL_Post_Type {
 			return;
 		}
 
+		// Show nothing if there are no valudation errors.
+		if ( 0 === count( array_filter( $error_summary ) ) ) {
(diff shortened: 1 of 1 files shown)


kienstra:
**Approved**

Hi @westonruter,
Good idea to output the `--` and exit from the method if there's no error.

Just like you [showed above](https://github.com/Automattic/amp-wp/pull/1592#issue-229111046), I also saw that this works as expected.

# Before
validation-before

# After
validation-after

PR #1605, opened 2018-11-10
Prevent a notice from appearing in the Compatibility Tool meta box

kienstra:
It looks like this notice doesn't appear with WordPress `5.0-beta3` and Gutenberg `4.2.0`.

But it appeared on the single URL validation page, with earlier versions of both. I can't reproduce this (even with `WP_DEBUG` set to `true`):

meta-box

Still, it might be good to ensure that in future versions of Core and Gutenberg, that notice doesn't appear.

This might not be the right flag. But this sets `__back_compat_meta_box` to `true`, to indicate that:

>...this meta box should only be loaded in the classic editor interface, and the block editor should not display it. ([source](https://make.wordpress.org/core/2018/11/07/meta-box-compatibility-flags/))

That document also notes:
>If you register your meta box with the __back_compat_meta_box set to true, then this warning will never be displayed with your meta box, now, or in future WordPress releases.

If this notice appeared in WP `5.0-beta3` and Gutenberg `4.2.0`, maybe another flag would be better.

This is based on this Make post that @westonruter mentioned:
See https://make.wordpress.org/core/2018/11/07/meta-box-compatibility-flags/

### DIFF
--- a/includes/validation/class-amp-validated-url-post-type.php
+++ b/includes/validation/class-amp-validated-url-post-type.php
@@ -1599,7 +1599,15 @@ class AMP_Validated_URL_Post_Type {
 	public static function add_meta_boxes() {
 		remove_meta_box( 'submitdiv', self::POST_TYPE_SLUG, 'side' );
 		remove_meta_box( 'slugdiv', self::POST_TYPE_SLUG, 'normal' );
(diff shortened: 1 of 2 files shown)


kienstra:
**Request For Review**

Hi @westonruter,

This uses the `__back_compat_meta_box` flag, which is a little different from what we talked about.

I can't reproduce this notice appearing, but it might be good to pass some flag to prevent this notice in future versions of Core or Gutenberg.

PR #1616, opened 2018-11-15
[WIP] [amp-stories] 1541 - Blacklist root elements which are not Page 

### LINKED ISSUE #1541: It should not be possible for an AMP story author to create blocks created outside of page
**As a AMP story author, I don’t want be able to create blocks outside of the current page**

 - [ ] _AC 1:_  Ensure that it is not  possible to create blocks created outside of page: https://v.usetapes.com/IYEIsJFz2E. 
     - [ ] _AC 2:_  It should be clear that top inserter is used for creating new blocks only in the current page/layer

mehigh:
Fixes #1541

### DIFF
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -231,3 +231,34 @@ div[data-amp-type="grid"] .block-library-image__resizer {
 .post-type-amp_story .components-range-control__number {
 	width: 60px;
 }
+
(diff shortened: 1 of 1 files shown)


miina:
@mehigh Looks like there is a way now how to remove the elements except for Page within Gutenberg, so we're not going to need to hide it with CSS afterall 🎉 

Closing this PR in favor of #1664.

PR #1621, opened 2018-11-16
Short-circuit polldaddy shortcode when no poll or survey supplied

### LINKED ISSUE #1214: Polldaddy embed not working
Polldaddy embed not works with amp plugin.

### LINKED ISSUE #1620: Polldaddy shortcode callback triggered when shortcode not present
When using the plugin in conjunction with the Polldaddy plugin: https://wordpress.org/plugins/polldaddy/ the polldaddy shortcode callback function in `wpcom/class-amp-polldaddy-embed.php` is entered even when there is no polldaddy shortcode on the page. This results in a link at the bottom of the post content with text equal to the post title and url set to `https://polldaddy.com`.

westonruter:
Prevent showing a link when a rating embed is being output.

Aside: This logic should be moved to the ~Polldaddy~ Crowdsignal plugin and not be housed in the AMP plugin.

# Before

image

# After

image

Fixes #1620.

### DIFF
--- a/wpcom/class-amp-polldaddy-embed.php
+++ b/wpcom/class-amp-polldaddy-embed.php
@@ -36,17 +36,21 @@ class WPCOM_AMP_Polldaddy_Embed extends AMP_Base_Embed_Handler {
 	public function shortcode( $attr ) {
 		global $wp_embed;
 
-		$output = '';
-		$url    = 'https://polldaddy.com/';
+		$url = null;
 		if ( ! empty( $attr['poll'] ) ) {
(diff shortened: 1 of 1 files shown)


kienstra:
**Approved**

Hi @westonruter,
This looks good.

This works as expected, assuming I understand it correctly.

This fixes the behavior when the shortcode doesn't have a `poll` or `survey` attribute, like if it's simply `[polldaddy]`

# Before 
polldaddy-before

# After
polldaddy-after

PR #1622, opened 2018-11-16
Gutenberg: Remove 'type' from attributes where 'source' is set.

miina:
Removes 'type' from attributes where 'source' is set, apparently that was deprecated in 4.1.0, removed in 4.2.0 and is now causes issues with blocks and extended features.

Note that `type` isn't mandatory for other cases either (if the source is not set at all).

### DIFF
--- a/assets/js/amp-editor-blocks.js
+++ b/assets/js/amp-editor-blocks.js
@@ -230,25 +230,21 @@ var ampEditorBlocks = ( function() { // eslint-disable-line no-unused-vars
 				settings.attributes = {};
 			}
 			settings.attributes.ampFitText = {
-				type: 'boolean',
 				default: false
 			};
 			settings.attributes.minFont = {
-				type: 'number',
 				default: component.data.fontSizes.small,
 				source: 'attribute',
 				selector: 'amp-fit-text',
 				attribute: 'min-font-size'
 			};
 			settings.attributes.maxFont = {
-				type: 'number',
 				default: component.data.fontSizes.larger,
 				source: 'attribute',
 				selector: 'amp-fit-text',
 				attribute: 'max-font-size'
 			};
 			settings.attributes.height = {
-				type: 'number',
 				default: 50,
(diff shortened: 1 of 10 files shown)


miina:
@kienstra and @westonruter:
This PR should fix most of the custom blocks and some custom AMP features not working in Gutenberg.

For some reason the branch `1.0` is showing some additional file changes as well, would it be better if I'd create the PR into `develop`?

There is another branch doing the same thing but based on `develop` if that's better:
https://github.com/Automattic/amp-wp/compare/fix/block-attributes?expand=1

westonruter:
> For some reason the branch `1.0` is showing some additional file changes as well, would it be better if I'd create the PR into `develop`?

Fixed by cherry-picking your commit onto `1.0` and then force-pushing to this branch.

kienstra:
**Approved**

Hi @miina,
Thanks for fixing this. All of the blocks that this affects looks good, except for some issues not related to this PR (mentioned below).

# amp-fit-text
amp-fit-text-test

# amp-brid-player
amp-brid-player-test

# amp-ima-video
amp-ima-video-test

# amp-jwplayer
amp-jwplayer-test

# Issues Unrelated To PR
The `<amp-springboard-player>` doesn't render, but it doesn't render on [ampbyexample.com](https://ampbyexample.com/components/amp-springboard-player/), either. This could be a matter of finding the right attribute values to test this.

amp-springboard-player

There's a validation error for the `<amp-o2-player>`, though I don't think it's related to this PR. I'll open a separate issue for this:

validation-error-amp-o2-player

kienstra:
**amp-o2-player**

This wasn't related to this PR. But the [problem I mentioned above](https://github.com/Automattic/amp-wp/pull/1622#issuecomment-439726698) with the `<amp-o2-player>` doesn't seem to exist when using the attribute values from [ampbyexample.com](https://ampbyexample.com/components/amp-o2-player/#basic-usage):

amp-o2-player

PR #1625, opened 2018-11-19
[amp-stories sub-task] Adjust default paragraph settings.

### LINKED ISSUE #1565: Add larger default styles for fonts in AMP stories
*As an AMP Author I would like add larger default styles for fonts in AMP stories, so that paragraph text for stories large by default instead of medium*
 - [ ] _AC 1:_ Add larger default styles for fonts (or leverage custom font sizes in Gutenberg) making the default font size for a paragraph large instead of medium.
 - [ ] _AC 2:_  **Optional:**  Make AMP Fit Text the default?  (_Try it and evaluate_)
         -   Rename "AMP Fit Text" to "Automatically fit text to container"
         -  Replicate AMP Fit Text functionality inside of the Gutenberg editor so that it can be previewed.

miina:
Fixes #1565.

- Sets the default font size to `large`.
- Fixes AMP Fit Text for paragraph;
- Removes `type` from settings since that's not supported anymore in case `source` is set;

Note: At this moment it's not possible to use AMP components within editor, it's not possible to replicate the AMP Fit Text behavior within editor with reasonable amount of effort.


miina:
@mehigh This issue also has an optional task to replicate AMP Fit Text behavior in the editor as well. Since at this moment AMP elements are not handled within the editor then replicating AMP Fit Text doesn't seem to be doable with reasonable effort. That's something we have discussed previously as well (not being able to replicate the behavior of AMP elements within editor).

However, since this was added to the AC as an optional then could you verify just in case if replicating AMP Fit Text behavior would be possible with reasonable amount of effort?

Thank you!

mehigh:
This can't be done with CSS only... only with JS.

assets/js/amp-editor-blocks.js
```diff
@@ -503,13 +499,12 @@ var ampEditorBlocks = ( function() { // eslint-disable-line no-unused-vars
...
 					label: __( 'Height' ),
 					value: height,
-					type: 'number',
 					min: 1,
 					onChange: function( nextHeight ) {
 						props.setAttributes( { height: nextHeight } );
 					}
 				} ),
-				maxFont > height && el(
+				parseInt( maxFont ) > parseInt( height ) && el(
```
jacobschweitzer: I think for easier reading and to be DRY, we could `parseInt()` these variables just inside the preceding `if ( ) {` statement. Also, it's good practice to explicitly state the 2nd parameter for `parseInt()` which in this case would be 10 like `parseInt( maxFont, 10 )` .

assets/js/amp-editor-blocks.js
```diff
@@ -526,13 +521,13 @@ var ampEditorBlocks = ( function() { // eslint-disable-line no-unused-vars
 							if ( ! nextMinFont ) {
 								nextMinFont = component.data.fontSizes.small; // @todo Supplying fallbackFontSize should be done automatically by the component?
 							}
-							if ( nextMinFont <= maxFont ) {
+							if ( parseInt( nextMinFont ) <= parseInt( maxFont ) ) {
```
jacobschweitzer: See previous comment regarding `parseInt`

assets/js/amp-editor-blocks.js
```diff
@@ -526,13 +521,13 @@ var ampEditorBlocks = ( function() { // eslint-disable-line no-unused-vars
...
 								props.setAttributes( { minFont: nextMinFont } );
 							}
 						}
 					} )
 				),
-				minFont > maxFont && el(
+				parseInt( minFont ) > parseInt( maxFont ) && el(
```
jacobschweitzer: See previous comment regarding `parseInt`

assets/js/amp-editor-blocks.js
```diff
@@ -231,25 +231,21 @@ var ampEditorBlocks = ( function() { // eslint-disable-line no-unused-vars
 				settings.attributes = {};
 			}
 			settings.attributes.ampFitText = {
-				type: 'boolean',
```
jacobschweitzer: Why remove all the types (just wondering) ?

assets/js/amp-editor-blocks.js
miina: Thanks, makes sense!

assets/js/amp-editor-blocks.js
miina: Apparently using `type` changed with Gutenberg 4.2.0 -- it's still allowed but not mandatory if the `source` is not specified, however, it's not supported if `source` is specified and the attribute won't work correctly (it will save but won't fetch the value as expected).

miina:
@jacobschweitzer: Just made the updates to parsing integers, could you recheck the addressed parts? Thanks!

assets/js/amp-editor-blocks.js
jacobschweitzer: Thanks, you may have just saved me some time/frustration in the future! 👍

mehigh:
Thanks for the confirmation @jacobschweitzer . Let's get this baby merged in.

Issue #1524, opened 2018-10-23
Fix the outline box and block selector to make it clearer what block has been selected. 


alcurrie:
**As a AMP Stories author, I want the interface for the outline box and the block selector improved so that I can easily understand what block has been selected**
 - [ ]  _AC 1:_ Fix/Improve  the interface for the outline box and the block selector to make it clearer what block has been selected
    - [ ]  _AC 2:_ Add Blue outline with highlight around selected block
          **Selected Block** (ie text block) 
![selectedblock](https://user-images.githubusercontent.com/18579988/48522822-1278dc00-e848-11e8-9bd9-5ceb8c19de61.png)
 (_NB: forward/back arrows show for selected block. See_ #1564) 
    - [ ]  _AC 3:_Add Blue outline what will show around Layer when the layer is selected *The blue line around the paragraph should not be displaying when the page is selected - mock up is incorrect.
     **Selected Layer**  (ie thirds layer) 
![selectedlayer](https://user-images.githubusercontent.com/18579988/48522850-1c9ada80-e848-11e8-915b-14d4a2e24f9b.png)
 (_NB: forward/back arrows show for selected Layer. See_ #1564) 
    - [ ]  _AC 4:_Add Blue outline what will show around Story page when the story page is selected  (_NB: forward/back arrows show  for selected story page._ See #1564) 
    - [ ]  _AC 5:_  Add Blur to all elements below the layer that contains the selected element.  For additional clarity for the user, as well as visually highlighting the selected element on a layer,  all elements on the layer below will be blurred, and any elements in the layers above the selected element will be hidden. See discussion in #1564 
    - [ ]  _AC 6:_  Add White shadow below placeholder text so that placeholder text is more visible  on darker backgrounds
    - [ ] _AC 7:_ Ensure that there is a visible hover for block elements (similar to what is default for Gutenberg) and that goes around the whole element. 
 - [ ]  _AC 7:_ The interface change should be aligned with and work well with the other related interface changes.  
    - [x]  Layer inserter will be moved to above the block navigation (Resolved in #1563)  
    - [x]  Forward/back arrows show for selected Block, Layer or AMP story (Resolved in #1564) 
    - [ ]  Highlight (bold) element selected in Block navigation. (Resolved in #1563)  
    - [ ]  Blur all blocks/layers below not selected and hide layers above (Resolved in #1524)

jwold:
[...]

Issue #1522, opened 2018-10-23
Update the border in the AMP Story interface to be more apparent for pages


alcurrie:
As an editor creating an AMP story, I'd like there to be a more apparent border for pages in the AMP Stories editor
- [ ] AC: Make border more apparent for pages - https://d.pr/i/j9e2Nq
     Design for updated border around the active AMP Story page is shown in this mockup: https://cl.ly/fc36d7797fbb

alcurrie:
@jwold  though this task design change, this doesn't seem like this needs design UX feedback as the screenshot in the AC is fairly straightforward, though it may make sense to include it in the overall look at design/UX changes needed.

jwold:
This has been discussed in https://github.com/Automattic/amp-wp/issues/1564, proposing what the border should look like. Feeling pretty good with where we've landed.

alcurrie:
@mehigh  - for additional clarification I've updated the AC to show the reference to border, but as @jwold notes this is also shown in the latest designs discussed/posted in #1564

mehigh:
The colors have been updated and now the page looks like this: http://cloud.urldocs.com/26f832e554cd
Closing as it was merged this along the fixes for 1524
Moving it on to QA.

alcurrie:
Confirmed: https://cl.ly/8da2559e0382

PR #1629, opened 2018-11-20
[amp-stories sub-task] Improve outline of blocks.

mehigh:
Fixes the remaining AC2, AC3, AC4, AC6 and AC7 of #1524

The blue outline is maintained around the currently selected layer and the text as well - even when typing: http://cloud.urldocs.com/7ec491663479


mehigh:
I've also fixed an issue where the warning was not clickable (convert a block -> i was moving from various branches which had or didn't have a functionality) .. 
and I discovered it to be caused by something in the core of G:
.editor-block-list__layout .editor-block-list__block.has-warning .editor-block-list__block-edit > :not(.editor-warning)
in this context, the editor-warning was not a direct child of editor-block-list__block-edit, hence it was not clickable at all. I've patched it, so in an amp stories context this doesn't occur. I'm not sure if we're adding the extra wrapper, or a G update, but it's working OK now.

mehigh:
Fixes #1522 - border & background updated for the page
Fixes #1524 - handling the last feedback of not applying the blue border to the parents of the selected item

Issue #1532, opened 2018-10-24
Improve the flow for adding new page blocks below existing stories


alcurrie:
**As a AMP story author, I want a better user flow for adding new page blocks below existing stories, so I'm able to easily understand how to add new pages to my AMP story**
 - [ ] _AC 1:_ Identify interface changes from what is currently available for blocks in AMP stores, that would improve the flow to add new story pages below existing story pages (i.e. show 'ghosted' page with icon)
          - [ ] Display ghosted Story page by default on a new AMP story or below the last AMP story page as shown in this mockup: https://cl.ly/4840a3512886
          - [ ] On tap/click anywhere on the 'ghosted story' page a new story page is created
          - [ ] When a new story page is created,  the block navigation with the (+) Add Layer icon above the Block Navigation (heading)  and the Story page top level on the block navigation selected
![initialstateblocknavafterpagecreation](https://user-images.githubusercontent.com/18579988/48520951-929b4380-e840-11e8-9c24-0586c22d011c.png)
 - [ ] _AC 2:_ Add page number to the AMP stories page interface to represent what story page the user currently editing.   (_Based on the new, easier story page/layer  manipulation #1564 consider not numbering story pages in order to not cause confusion when pages are 'reordered'._) 
- [ ] _AC 3:_ Update the page icon in the block navigation active story to be aligned with/match the updated ‘add a new page’  icon  https://cl.ly/24a5c2e8ab6e

westonruter:
Note: The big plus icon now seems to be part of Gutenberg itself, although it is not shown in a ghosted page:

![image](https://user-images.githubusercontent.com/134745/47463402-f8912f80-d79a-11e8-8218-f74840266f8a.png)

alcurrie:
@westonruter Thanks for the call out on the plus already being available. I've updated AC1 to take into account that this task maybe partially done, though may need some additional improvements to meet the overall 'goal' in this ticket. 

@jwold -  Based on Weston's screen shot, this looks like we'll need to take this into account when making this interface change to make adding pages easier. 
 From the screen shot Weston shared,  as a user, I'd assume the plus at the top of (inside) the card would add a new layer, and the plus below the card would add a new page. Assuming the plus below the card adds a new page (i.e. already does part of what we're asking for) that would affect the specifics of the approach. We wouldn't need to add a + icon, but I still think there could be improvements in styling and interface to make the process clearer.   I've update AC1 and added a Design_Feedback label as it would be good if you could weigh in on this.

[...]

PR #1632, opened 2018-11-21
[amp-stories sub-task] Improve adding a page block flow

bluestormcreative:
Fixes parts of #1532. 
- Adds styles and new svg icon to turn the default block-appender inserter into a "ghosted" page 
- Hover/focus state on the ghosted page should be confirmed - default is still in place and adds an inset box-shadow and updates the background color to white
- When clicking on the ghosted page inserter, the default action to choose a block is still in place


mehigh:
I'm going to pick this one up and bring it home.

PR #1634, opened 2018-11-21
Fix locale data for WordPress 5.0

### LINKED ISSUE #1618: Rename or replace gutenberg_get_jed_locale_data() usage
`gutenberg_get_jed_locale_data()` is now `wp_get_jed_locale_data()` in WordPress core.

The plugin currently fatals because of that.

The function should be renamed in the plugin and perhaps even considered to replaced in WP 5.0+ as per https://make.wordpress.org/core/2018/11/09/new-javascript-i18n-support-in-wordpress/.

See also #1604.

kienstra:
As @westonruter pointed out, `wp_get_jed_locale_data()` [was just removed from Core](https://core.trac.wordpress.org/changeset/43910).

Because that function was never used in a released version of this plugin, this removes it entirely (see the [0.7 branch](https://github.com/Automattic/amp-wp/tree/0.7)). But maybe that's not a great approach.

`wp_set_script_translations` [was added](https://github.com/WordPress/wordpress-develop/commit/4f05bb81f609559f865a85687166b9d7881639d2) to WP 5.0 on October 25th.

This was released in 5.0 Beta4 on November 13th:
https://wordpress.org/news/2018/11/wordpress-5-0-beta-4/

Fixes #1618

### DIFF
--- a/includes/admin/class-amp-editor-blocks.php
+++ b/includes/admin/class-amp-editor-blocks.php
@@ -138,6 +138,10 @@ class AMP_Editor_Blocks {
 				array( 'wp-blocks', 'lodash', 'wp-i18n', 'wp-element', 'wp-components' ),
 				AMP__VERSION
 			);
+
+			if ( function_exists( 'wp_set_script_translations' ) ) {
+				wp_set_script_translations( 'amp-editor-blocks-build', 'amp' );
+			}
 		}
 
 		wp_enqueue_script(
@@ -155,12 +159,9 @@ class AMP_Editor_Blocks {
 			) ) )
 		);
 
-		$locale_data = function_exists( 'wp_get_jed_locale_data' ) ? wp_get_jed_locale_data( 'amp' ) : gutenberg_get_jed_locale_data( 'amp' );
-		wp_add_inline_script(
-			'wp-i18n',
-			'wp.i18n.setLocaleData( ' . wp_json_encode( $locale_data ) . ', "amp" );',
-			'after'
-		);
+		if ( function_exists( 'wp_set_script_translations' ) ) {
+			wp_set_script_translations( 'amp-editor-blocks', 'amp' );
+		}
 	}
 
 	/**
--- a/includes/admin/class-amp-editor-blocks.php
+++ b/includes/admin/class-amp-editor-blocks.php
@@ -135,7 +135,7 @@ class AMP_Editor_Blocks {
 			wp_enqueue_script(
 				'amp-editor-blocks-build',
 				amp_get_asset_url( 'js/amp-blocks-compiled.js' ),
-				array( 'wp-blocks', 'lodash', 'wp-i18n', 'wp-element', 'wp-components' ),
+				array( 'wp-editor', 'wp-blocks', 'lodash', 'wp-i18n', 'wp-element', 'wp-components' ),
 				AMP__VERSION
 			);
 
--- a/includes/admin/class-amp-editor-blocks.php
+++ b/includes/admin/class-amp-editor-blocks.php
@@ -161,6 +161,13 @@ class AMP_Editor_Blocks {
 
 		if ( function_exists( 'wp_set_script_translations' ) ) {
 			wp_set_script_translations( 'amp-editor-blocks', 'amp' );
+		} elseif ( function_exists( 'wp_get_jed_locale_data' ) || function_exists( 'gutenberg_get_jed_locale_data' ) ) {
+			$locale_data = function_exists( 'wp_get_jed_locale_data' ) ? wp_get_jed_locale_data( 'amp' ) : gutenberg_get_jed_locale_data( 'amp' );
+			wp_add_inline_script(
+				'wp-i18n',
+				'wp.i18n.setLocaleData( ' . wp_json_encode( $locale_data ) . ', "amp" );',
+				'after'
+			);
 		}
 	}
 
--- a/includes/admin/class-amp-post-meta-box.php
+++ b/includes/admin/class-amp-post-meta-box.php
@@ -204,6 +204,10 @@ class AMP_Post_Meta_Box {
 
 		if ( function_exists( 'wp_set_script_translations' ) ) {
 			wp_set_script_translations( self::BLOCK_ASSET_HANDLE, 'amp' );
+		} elseif ( function_exists( 'wp_get_jed_locale_data' ) ) {
+			$script_data['i18n'] = wp_get_jed_locale_data( 'amp' );
+		} elseif ( function_exists( 'gutenberg_get_jed_locale_data' ) ) {
+			$script_data['i18n'] = gutenberg_get_jed_locale_data( 'amp' );
 		}
 
 		wp_add_inline_script(
--- a/includes/admin/class-amp-post-meta-box.php
+++ b/includes/admin/class-amp-post-meta-box.php
@@ -202,7 +202,9 @@ class AMP_Post_Meta_Box {
 			'errorMessages' => $error_messages,
 		);
 
-		$script_data['i18n'] = function_exists( 'wp_get_jed_locale_data' ) ? wp_get_jed_locale_data( 'amp' ) : gutenberg_get_jed_locale_data( 'amp' );
+		if ( function_exists( 'wp_set_script_translations' ) ) {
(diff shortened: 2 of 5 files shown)


kienstra:
Thanks to @miina for pointing out that `wp-editor` '[is needed as a dependency](https://github.com/WordPress/gutenberg/issues/12173#issuecomment-440715061).

tests/test-class-amp-meta-box.php
```diff
@@ -77,7 +77,7 @@ public function test_enqueue_admin_assets() {
 	 * @see AMP_Post_Meta_Box::enqueue_block_assets()
 	 */
 	public function test_enqueue_block_assets() {
-		if ( ! function_exists( 'wp_get_jed_locale_data' ) && ! function_exists( 'gutenberg_get_jed_locale_data' ) ) {
+		if ( ! function_exists( 'the_gutenberg_project' ) ) {
```
miina: Maybe it would be good to use a function here that's very likely going to stay even when Gutenberg gets merged? For example `register_block_type`. Thoughts?

kienstra: Good point. I'm changing that now.

includes/admin/class-amp-post-meta-box.php
```diff
@@ -202,7 +202,9 @@ public function enqueue_block_assets() {
 			'errorMessages' => $error_messages,
 		);
 
-		$script_data['i18n'] = function_exists( 'wp_get_jed_locale_data' ) ? wp_get_jed_locale_data( 'amp' ) : gutenberg_get_jed_locale_data( 'amp' );
+		if ( function_exists( 'wp_set_script_translations' ) ) {
```
miina: Should this still support older versions as well where the `wp_set_script_translations` was not introduce yet?

tests/test-class-amp-meta-box.php
kienstra: Thanks, your suggestion is applied with https://github.com/Automattic/amp-wp/pull/1634/commits/3915ba24fd92b702b5d78895faf61c5f1e46f964

includes/admin/class-amp-post-meta-box.php
miina: I guess the assumption is that Gutenberg will be included in 5.0 and not before, so it would be okay like this as well.

kienstra: Hi @miina,
That's a really good question . This change would leave out people that use some version of Gutenberg, but WordPress earlier than [5-0-beta-4](https://wordpress.org/news/2018/11/wordpress-5-0-beta-4/) (when `wp_set_script_translations()` was released). 

So they wouldn't get translations for the meta box.

Should we expect users that are running Gutenberg to have at least WordPress [5-0-beta-4](https://wordpress.org/news/2018/11/wordpress-5-0-beta-4/)?

I think it'll add a lot of complexity to have conditionals to support this.

```
if ( function_exists( 'wp_set_script_translations' ) ) {
	wp_set_script_translations( self::BLOCK_ASSET_HANDLE, 'amp' );
} elseif ( function_exists( 'wp_get_jed_locale_data' ) ) {
	$script_data['i18n'] = wp_get_jed_locale_data( 'amp' );
} elseif ( function_exists( 'wp_get_jed_locale_data' ) ) {
	$script_data['i18n'] = gutenberg_get_jed_locale_data( 'amp' );
}
```

 If we're going to remove references to `wp_set_script_translations()` and `wp_get_jed_locale_data()`, we should do it before the `1.0` stable release.

kienstra: Still, you raise a good point that this should support older versions. I'm going to add conditionals that support WP before `5.0-beta4`.

kienstra: https://github.com/Automattic/amp-wp/pull/1634/commits/0cac87cc5ed3c310b973dc8e5acdbb415460cadd restores the previous use of `wp_get_jed_locale_data()` or `gutenberg_get_jed_locale_data()` `wp_set_script_translations()` doesn't exist. I still need to address the failed [Travis build](https://travis-ci.org/Automattic/amp-wp/jobs/458111904).

includes/admin/class-amp-editor-blocks.php
```diff
@@ -135,9 +135,13 @@ public function enqueue_block_editor_assets() {
 			wp_enqueue_script(
 				'amp-editor-blocks-build',
 				amp_get_asset_url( 'js/amp-blocks-compiled.js' ),
-				array( 'wp-blocks', 'lodash', 'wp-i18n', 'wp-element', 'wp-components' ),
+				array( 'wp-editor', 'wp-blocks', 'lodash', 'wp-i18n', 'wp-element', 'wp-components' ),
 				AMP__VERSION
 			);
+
+			if ( function_exists( 'wp_set_script_translations' ) ) {
+				wp_set_script_translations( 'amp-editor-blocks-build', 'amp' );
```
miina: Should this also support the previous versions, too?

kienstra: It looks like this could use the `wp.i18n.setLocaleData()` that's [lower in the function](https://github.com/Automattic/amp-wp/pull/1634/files/ad105928431545d1eb9a5fe00e0abf57a3daa70c#diff-af86f8db0fb2fe24a8aea92de5ee5128R166), if I understand this correctly.

miina: You're right!

includes/admin/class-amp-post-meta-box.php
westonruter: We don't need to sorry about back compat for WordPress 5.0 beta releases. Only be concerned with 4.9.x+Gutenberg and 5.0 stable.

kienstra:
Hi @miina, thanks a lot for reviewing this. How do the latest commits look? Luckily, the Travis builds just passed.

kienstra:
Thanks a lot for reviewing this, @miina!

kienstra:
**Request To Merge**

Hi @amedina,
Could you please merge this PR?

(Or @westonruter, assuming you're satisfied with [your point above](https://github.com/Automattic/amp-wp/pull/1634#discussion_r235541059). No pressure, it's one of your very few days off 😄)

westonruter:
@ktmn As I understand, the POT file isn't included in the plugin build because it will be provided by WordPress.org's language packs. So the way to translate the string would be to do so via translate.wordpress.org once the release has been made. A POT file is generated during the build but it is converted into PHP for this purpose.

@swissspidy please confirm.

ktmn:
@westonruter Ok so with WP 5.0 and AMP 1.0.0 I swiched to Finnish language and indeed an `amp-fi.po` showed up in Loco Translate. But it doesn't seem to include some (any?) of the strings that are in scripts (ex. `Add lightbox effect`), even after I "Sync" which finds ~400 additonal strings that are not translated to Finnish I guess. I translated the string "AMP Settings" which appears in the Gutenberg `core/gallery` settings, but the translation is not passed to the script.

Where is the `wp_set_script_translations('amp-editor-blocks', 'amp')` supposed to take the translated strings from and why doesn't it? It just outputs empty `"locale_data": { "messages": { "": {} } }`, while `wp_get_jed_locale_data()` used to actually populate the messages.

swissspidy:
Those strings will be taken from new JSON translation files being shipped for plugins since 5.0.

swissspidy:
Looks like WordPress.org doesn't yet provide the JSON translation files, that's why you get `"locale_data": { "messages": { "": {} } }` at the moment. 🤔 Will have to check it.

swissspidy:
You mean how should it work for any other plugin not on WordPress.org?

First of all, you don‘t need pot-to-php. That‘s obsolete now. You create one POT file with all translatable strings and from there translate them using a tool like Poedit.

Then, you can create JSON files from the resulting PO file using the po2json npm script or ideally WP-CLI (https://github.com/wp-cli/i18n-command), which is better suited for WP‘s needs.

The main reason for this is performance: instead of loading ALL translations it only loads the ones needed for that particular JavaScript file.

Also, there definitely is some documentation already:

https://make.wordpress.org/core/2018/11/09/new-javascript-i18n-support-in-wordpress/
https://wordpress.org/gutenberg/handbook/designers-developers/developers/internationalization/

I am also publishing a very thorough blog post about this today or tomorrow. Happy to share it here.

PR #1630, opened 2018-11-20
[WIP] [amp-stories sub-task] Make it easier to select parents/children within Page.

miina:
Fixes #1563.

- Removes the custom Layer Cards;
- Adds block navigation to the left side of a Page;
- Hides the default block navigation icon;
- Remove custom LayerInserter component and replace with default `Inserter` -- it now accepts `rootClientId` 🎉 ;

Todo:
- Styling;

PR #1636, opened 2018-11-21
[amp-stories sub-task] CTA UI improvements

jacobschweitzer:


blocks/amp-story/amp-story-cta-layer.js
```diff
@@ -105,9 +110,25 @@ export default registerBlockType(
...
+
 				if ( this.props.attributes.hasMultipleCtaBlocks ) {
 					removeBlock( this.props.clientId );
-					dispatch( 'core/editor' ).createWarningNotice( __( 'Multiple CTA Layers are not allowed, the block was removed.', 'amp' ) );
+					noticeMessage = wp.i18n.__( 'Multiple CTA Layers are not allowed, the block was removed.', 'amp' );
+				}
+
+				if ( 0 === blockIndex ) {
+					removeBlock( this.props.clientId );
+					noticeMessage = wp.i18n.__( 'CTA layer is not allowed on the first page, the block was removed.', 'amp' );
```
miina: Actually CTA Layer is allowed on each page -- it has to be the last Layer of a Page, so it can't be the first Layer in case there is another Layer already existing.

blocks/amp-story/block-selector.js
```diff
@@ -48,6 +48,14 @@ class BlockSelector extends Component {
 			let className = 'component-editor__selector template-' + template;
 			if ( isBlockSelected( block.clientId ) || hasSelectedInnerBlock( block.clientId ) ) {
 				className += ' is-selected';
+
+				// Do not blur the CTA layer so the user can edit the elements within it.
+				if ( 'amp/amp-story-cta-layer' === block.name ) {
+					const ampStoryDiv = document.getElementById( 'block-' + block.clientId );
+					if ( ampStoryDiv ) {
+						ampStoryDiv.classList.add( 'is-selected' );
```
miina: @jacobschweitzer What seems to be the issue here with blurring? Perhaps that's something that @mehigh can take a look at within CSS instead? CTA Layer shouldn't probably be blurred at any point since it's always the top layer -- so it wouldn't be possible to select a layer that's on top of CTA Layer.

Mike, do you think you could look into CTA Layer to make sure that it's not being blurred when edited? There seems to be another new CSS-related issue with CTA Layer -- The CTA Layer should always take 20% of the Page height, however, it looks like the CTA Layer has the same height as the other layers:

screen shot 2018-11-22 at 9 25 59 am

blocks/amp-story/amp-story-cta-layer.js
```diff
@@ -9,7 +9,8 @@ const {
 } = wp.blocks;
 const {
 	InspectorControls,
-	InnerBlocks
+	InnerBlocks,
+	RichText
```
miina: AC3 of the issue was to also pre-populate CTA block with a `core/button` block -- is this something that doesn't seem to be possible at this moment? I thought I saw a template variable somewhere yesterday but not seeing it today anymore.

blocks/amp-story/amp-story-cta-layer.js
miina: For reference: https://www.ampproject.org/docs/reference/components/amp-story#amp-story-cta-layer

blocks/amp-story/block-selector.js
miina: Also note that the Layer Cards ( block-selector.js ) will be removed within #1630: that's something that will not exist soon and will be replaced by the Block Navigator on the left.

jacobschweitzer: @miina Here is what I was seeing with the blurring issue:
<a href="https://cl.ly/e1384d9be219" target="_blank"></a>

blocks/amp-story/amp-story-cta-layer.js
jacobschweitzer: Hmm yeah that is interesting, I wonder why I got that console error then. I'll test again and see what happens.

jacobschweitzer: I tried again and got the same error, it is the last layer on the first page:
image

miina: Okay, looks like they have updated AMP but didn't update the docs accordingly.

Thanks for finding and adding this!

Probably it would be good to hide the CTA Layer Icon via CSS from the Inserter of the first Page then, too.

blocks/amp-story/amp-story-cta-layer.js
```diff
@@ -105,9 +109,25 @@ export default registerBlockType(
...
+					id: 'amp-errors-notice-removed-cta'
+				};
+
 				if ( this.props.attributes.hasMultipleCtaBlocks ) {
 					removeBlock( this.props.clientId );
-					dispatch( 'core/editor' ).createWarningNotice( __( 'Multiple CTA Layers are not allowed, the block was removed.', 'amp' ) );
+					noticeMessage = wp.i18n.__( 'Multiple CTA Layers are not allowed, the block was removed.', 'amp' );
+				}
+
+				if ( 0 === blockIndex ) {
```
miina: Maybe this would make sense as `else if ( 0 === blockIndex ) {` ? Since if the block already gets removed within the previous `if` then trying to remove it the second time might cause an error.

blocks/amp-story/block-selector.js
jacobschweitzer: The underlying issue seems to be that the child block gets the `is-selected` class so it is not blurry, but its parent (`div.amp-story-cta-layer`) does not have `is-selected` so it is blurred which makes all its child elements blurred.

miina: @jacobschweitzer In the `helpers.js` file there is the method `maybeIsSelectedParentClass` which is used for adding `is-selected-parent` class to page and also layers -- except for CTA Layer -- perhaps you could implement this for the CTA Layer as well? This might fix the issue and if it does not right away, it would give a CSS selector for fixing the issue via CSS, as it's done for the other layers at this moment. Thoughts?

jacobschweitzer: Ok thanks @miina that looks like a good approach, I'll start working on that.

miina: For reference: https://github.com/Automattic/amp-wp/blob/amp-stories/assets/js/amp-story-editor-blocks.js#L228-L230

PR #1652, opened 2018-11-26
Move AMP Settings in editor to after default settings

### LINKED ISSUE #1645: Should AMP settings panels be open by default
I find it distracting that in many Gutenberg blocks, AMP is the first panel. AMP is important and all but it's not the first thing I wish to interact with. It's above block's native Gutenberg controls and it's open by default. Even if it's kept above the native controls, I would suggest having it collapse by default. If someone wishes to interact with AMP settings, they can expand it.

In Gutenberg, `PanelBody` has the prop `initialOpen` which can control this behavior.

miina:
Fixes #1645.

### DIFF
--- a/assets/js/amp-editor-blocks.js
+++ b/assets/js/amp-editor-blocks.js
@@ -321,10 +321,10 @@ var ampEditorBlocks = ( function() { // eslint-disable-line no-unused-vars
 			}
 
 			return [
-				inspectorControls,
 				el( BlockEdit, _.extend( {
 					key: 'original'
-				}, props ) )
+				}, props ) ),
+				inspectorControls
 			];
 		};
 	};


miina:
@westonruter This moves the AMP controls after the default controls in the sidebar.

However, one note about Gallery and Image blocks -- when these blocks are not populated with image(s) they don't show the default controls. The default controls will be added once an image has been added and it looks like the controls are then added to the end of the existing controls.

This means that if an image has been added and the editor would select the block then the default controls show first. However, if the image has not been added yet and gets added then the newly added Image controls seem to be added to the end thus having the AMP Controls first. It will display correct order after selecting another block for a moment or saving / refreshing the page.

Would this be sufficient to fix #1645? Thoughts?

westonruter:
I think so, yes.

PR #1661, opened 2018-11-27
Address an issue with an invalid embed 

kienstra:
# Steps To Reproduce 
1. Activate a theme that calls `the_content` inside a function, like the Genesis theme Essence Pro
2. Create a new post using the block editor
3. Add a 'Ted' block with this URL:
https://www.ted.com/talks/derek_sivers_how_to_start_a_movement
4. Save the post
5. Visit the validation error page for that post
6. Expected: 'Embed' appears as the sources, as the Ted block is the source of the validation errors:
expected-amp-embed
7. Actual: The theme is reported as the source
actual-parent-theme

Thanks to @hellofromtonya for reporting this.

The Genesis theme has a function `genesis_do_post_content()` that calls `the_content()`:
do_post_content

...so it was reported as the source.

Still, I'm not sure this PR's approach is right.

### DIFF
--- a/includes/validation/class-amp-validated-url-post-type.php
+++ b/includes/validation/class-amp-validated-url-post-type.php
@@ -982,7 +982,7 @@ class AMP_Validated_URL_Post_Type {
 				$output[] = '</details>';
 			}
 		}
-		if ( isset( $sources['theme'] ) ) {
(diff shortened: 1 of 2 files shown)


westonruter:
Good catch.

PR #1664, opened 2018-11-27
Look in entire document for Schema.org metadata not just head

### LINKED ISSUE #1659: RC3 - Structured Data Error
On v1RC3 I get valid AMP but errors on structured data (SD). On the non-AMP test there is 1 entry for for article but on the AMP SD there are two article entries, one with an error and one without. Here is a URL you can test SD on on for non-AMP and AMP.

https://ccountync.com/morehead-city-used-game-consoles-for-sale-nc-xbox-nintendo-playstation-wii/

westonruter:
The JSON-LD `script` elements can appear anywhere in a document, not just the `head`. This fixes the detection to account for that so that when another plugin outputs Schema.org metadata, the AMP plugin won't output a duplicate.

Fixes #1659

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1175,6 +1175,8 @@ class AMP_Theme_Support {
 		 * @var DOMElement $link
 		 */
 
+		$xpath = new DOMXPath( $dom );
+
 		// Make sure the HEAD element is in the doc.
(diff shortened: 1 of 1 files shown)


kienstra:
**Approved**

Hi @westonruter,
This looks good. I also saw that the plugin doesn't output Schema.org metadata if it's already in the document, like in the footer:

schema-appears-footer

And if the metadata isn't already present, the plugin still outputs it like before.

PR #1665, opened 2018-11-28
Update generated tags file to 767

miina:
PR for testing -- looks like updating the file makes the test fails.

### DIFF
--- a/includes/sanitizers/class-amp-allowed-tags-generated.php
+++ b/includes/sanitizers/class-amp-allowed-tags-generated.php
@@ -13,7 +13,7 @@
  */
 class AMP_Allowed_Tags_Generated {
 
-	private static $spec_file_revision = 757;
+	private static $spec_file_revision = 767;
 	private static $minimum_validator_revision_required = 348;
 
 	private static $allowed_tags = array(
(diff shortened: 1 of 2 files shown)


westonruter:
Updating the file from what state? The latest amphtml tag? I've identified an issue with the script where it is checking out the most recent tag and not necessarily the most recent stable release.

miina:
@westonruter I just checked out `1.0`, run the `amphtml-update.sh` and committed the changes.

Was there anything else that should have been done?

westonruter:
The last time I did it, I had to comment out these two lines:

https://github.com/Automattic/amp-wp/blob/b748977321cc3769d0476ceede56c39aa2fed267/bin/amphtml-update.sh#L35-L36

The reason being that the latest tag is not actually the latest stable release, as can be seen at https://github.com/ampproject/amphtml/releases

image

So I manually switched to the latest release tag when I did it. In practice this is probably overkill.

But as can be seen in https://github.com/ampproject/amphtml/pull/19214, the `amp-script` component was commented-out a couple weeks ago. That is why `amp-script` is now failing in the unit tests. So we just need to remove the `amp-script` test.

kienstra:
**Request For Review**

Hi @westonruter,
Could you please review this? It applies your suggestion of [removing the test](https://github.com/Automattic/amp-wp/pull/1665/commits/8dbbe502dd970ebf8820e53f247c89d94c68cb43) for `<amp-script>`, and it [adds a test](https://github.com/Automattic/amp-wp/pull/1665/commits/29f29d7d6161f0af20a5eb2f2bfc15f7c5f0821a) for the newly-allowed element `<amp-delight-player>`. 

Thanks, Weston 😄

PR #1672, opened 2018-11-29
Add CSS style to support RTL language in frontend

### LINKED ISSUE #235: RTL support in template
We should make sure things like the site title, logo, meta are all aligned properly.

mostafaabd:
Add new CSS template loaded for RTL language

Fix #235


westonruter:
I believe it would be better to add the `is_rtl()` checks directly to https://github.com/Automattic/amp-wp/blob/develop/templates/style.php

PR #1680, opened 2018-11-30
Fix tests which broke with Gutenberg 4.6

westonruter:
In Gutenberg 4.6, the `gutenberg_wpautop` function has gone away. So the tests need to be updated to account for that.


kienstra:
Thanks, @westonruter! I'm reviewing this now.

PR #1681, opened 2018-11-30
[Don't merge until after 1.0-RC4] Merge 1.0 into develop, without overwriting version

kienstra:


amp.php
```diff
@@ -5,7 +5,7 @@
  * Plugin URI: https://github.com/automattic/amp-wp
  * Author: WordPress.com VIP, XWP, Google, and contributors
  * Author URI: https://github.com/Automattic/amp-wp/graphs/contributors
- * Version: 1.0-RC3
+ * Version: 1.1-alpha
```
kienstra: The `develop` branch has a version of `1.1-alpha` here, so I'm not sure why this shows a diff.

https://github.com/Automattic/amp-wp/blob/develop/amp.php#L8

PR #1708, opened 2018-12-04
Let readme.md title differ from title in readme.txt

westonruter:
With this change, the GitHub project page will look like this:

> image

Whereas the title on WordPress.org will look like (disregard the "The" in the banner):

> mock-up-launch

Per https://github.com/Automattic/amp-wp/issues/1631#issuecomment-442129515


westonruter:
I also updated the description of the plugin as it appears in the admin:

image

PR #1713, opened 2018-12-06
1643: Add font family option and CSS

jacobschweitzer:


includes/class-amp-story-post-type.php
```diff
@@ -219,4 +214,21 @@ public static function filter_template_include( $template ) {
 		}
 		return $template;
 	}
+
+	/**
+	 * Add CSS to the frontend for AMP Stories.
+	 *
+	 * @see /assets/css/amp-story-frontend.css
+	 */
+	public function add_frontend_css() {
```
miina: This should probably be `static` if it's statically called.

includes/class-amp-story-post-type.php
```diff
@@ -219,4 +214,21 @@ public static function filter_template_include( $template ) {
...
+		$post = get_post();
+		if ( 'amp_story' !== $post->post_type ) {
+			return;
+		}
+
+		$css_src      = AMP__DIR__ . '/assets/css/amp-story-frontend.css';
+		$css_contents = file_get_contents( $css_src );
+
+		wp_add_inline_style( 'wp-block-library', $css_contents );
```
miina: Could we also add the CSS for the editor? So that the editor would see the change there, too?

includes/class-amp-story-post-type.php
jacobschweitzer: 👍

includes/class-amp-story-post-type.php
jacobschweitzer: Done, had to make some changes in the JS too for this to work.

postphotos:
@jacobschweitzer This is looking very promising. Thanks for the work here! 👍

miina:
@jacobschweitzer I'll review this again tomorrow. I can see that Travis is failing though, is this something you could fix meanwhile?

miina:
@jacobschweitzer FYI: Continuing your work here, will fix the coding standard issues myself.

assets/js/amp-story-editor-blocks.js
```diff
@@ -342,6 +342,14 @@ var ampStoryEditorBlocks = ( function() { // eslint-disable-line no-unused-vars
...
 				el = wp.element.createElement;
 
+			if ( attributes.ampFontFamily ) {
+				newProps.wrapperProps[ 'data-font-family' ] = attributes.ampFontFamily;
+				return el(
+					BlockListBlock,
+					newProps
+				);
+			}
+
```
jacobschweitzer: @miina I'm not sure if we need to do some re-factoring here or move this to its own function. As you can see I had it further down in the function before, but that didn't work properly as sometimes it didn't get that far to add the new property.

miina: @jacobschweitzer Just made some changes in the code, thanks for noting that.

PR #1709, opened 2018-12-05
Bump version to 1.0.0, update readme

kienstra:
Following steps 2 and 3 in [Creating a Release](https://github.com/Automattic/amp-wp/blob/32703d696fdcba9b1511fbfb7c3a2366d6ebd999/contributing.md#creating-a-release):
>2. Bump plugin versions in amp.php (×2: the metadata block in the header and also the AMP__VERSION constant)
>3. Add changelog entry to readme.

Todo: After the release, ~I'll do~ someone should do step 10:
>10. Create a new branch off of the release branch, merge develop into it and resolve conflicts (e.g. with version and changelog), and then open pull request to merge changes into develop.

Fixes #1633.

PR #1715, opened 2018-12-06
Fix composer vendor name and use stable version of CSS parser

felixarntz:
While the repo has been moved to the AMP Project organization, `composer.json` still has the old references.

Also, we shouldn't be using a possibly unstable version of a dependency for a stable release, the CSS parser however is currently used as that. In this PR I change it to use the latest stable version of it >= 8 < 9 (8.2 is the latest release of that package).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": "^5.3.6 || ^7.0",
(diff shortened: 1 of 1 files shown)


westonruter:
Most of this is fixed in #1709, so we'll merge these changes there.

PR #1721, opened 2018-12-08
Ensure Schema.org meta data has correct date

kienstra:
Thanks to [Fernando Tellado](https://wordpress.org/support/users/fernandot/) for raising this in:
https://wordpress.org/support/topic/warning-date-expects-parameter-2-to-be-integer-after-1-0-0/

The [documentation of date()](https://secure.php.net/manual/en/function.date.php) requires its 2nd argument to be an `int`.

And the [documentation of get_the_date()](https://developer.wordpress.org/reference/functions/get_the_date/) says it returns a string. 

But locally, using PHP 7.2, it looked like `get_the_date()` returns an int.

I couldn't reproduce this, but it'll probably be good to ensure this is an int. Edge cases might need more thought, like if `get_the_date()` somehow returns `''` or `false`, and `intval()` returns `0` for it.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -870,7 +870,7 @@ function amp_get_schemaorg_metadata() {
 				'@type'            => is_page() ? 'WebPage' : 'BlogPosting',
 				'mainEntityOfPage' => get_permalink(),
 				'headline'         => get_the_title(),
-				'datePublished'    => date( 'c', intval( get_the_date( 'U', $post->ID ) ) ),
+				'datePublished'    => mysql2date( 'c', $post->post_date_gmt, false ),
 				'dateModified'     => mysql2date( 'c', $post->post_modified_gmt, false ),
 			)
 		);
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -870,8 +870,8 @@ function amp_get_schemaorg_metadata() {
 				'@type'            => is_page() ? 'WebPage' : 'BlogPosting',
 				'mainEntityOfPage' => get_permalink(),
 				'headline'         => get_the_title(),
-				'datePublished'    => date( 'c', get_the_date( 'U', $post->ID ) ),
-				'dateModified'     => date( 'c', get_the_date( 'U', $post->ID ) ),
+				'datePublished'    => date( 'c', intval( get_the_date( 'U', $post->ID ) ) ),
+				'dateModified'     => date( 'c', intval( get_the_date( 'U', $post->ID ) ) ),
 			)
 		);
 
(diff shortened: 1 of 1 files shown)


usernamefghh:
pull

westonruter:
Here is the support topic that identifies the underlying problem: https://wordpress.org/support/topic/line-error-shown-with-wp_debug/

includes/amp-helper-functions.php
```diff
@@ -870,8 +870,8 @@ function amp_get_schemaorg_metadata() {
 				'@type'            => is_page() ? 'WebPage' : 'BlogPosting',
 				'mainEntityOfPage' => get_permalink(),
 				'headline'         => get_the_title(),
-				'datePublished'    => date( 'c', get_the_date( 'U', $post->ID ) ),
-				'dateModified'     => date( 'c', get_the_date( 'U', $post->ID ) ),
+				'datePublished'    => date( 'c', intval( get_the_date( 'U', $post->ID ) ) ),
```
```suggestion
				'datePublished'    => mysql2date( 'c', $post->post_date_gmt, false ),
```

includes/amp-helper-functions.php
```diff
@@ -870,8 +870,8 @@ function amp_get_schemaorg_metadata() {
...
+				'dateModified'     => date( 'c', intval( get_the_date( 'U', $post->ID ) ) ),
```
```suggestion
				'dateModified'     => mysql2date( 'c', $post->post_modified_gmt, false ),
```

includes/amp-helper-functions.php
```diff
@@ -870,8 +870,8 @@ function amp_get_schemaorg_metadata() {
 				'@type'            => is_page() ? 'WebPage' : 'BlogPosting',
 				'mainEntityOfPage' => get_permalink(),
 				'headline'         => get_the_title(),
-				'datePublished'    => date( 'c', get_the_date( 'U', $post->ID ) ),
-				'dateModified'     => date( 'c', get_the_date( 'U', $post->ID ) ),
```
westonruter: This is incorrectly using the published date as the modified date.

includes/amp-helper-functions.php
westonruter: This change also needs to be applied.

kienstra:
**Looks Good**

Hi @westonruter,
Your work here is a nice improvement. Like you pointed out, there are now different dates for `datePublished` and `dateModified`.

date-published-modified

And some pages that I've tested in the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/u/0/) look good:
structured-data-testing-tool

This should also prevent the issue in the [support topic](https://wordpress.org/support/topic/warning-date-expects-parameter-2-to-be-integer-after-1-0-0/), as this doesn't use `date()` or `get_the_date()`.

This works in my testing in Classic, Native, and Paired mode, though we wouldn't expect a difference there anyway.

kienstra:
What do you think about including this in `1.0.1`?

westonruter:
Something that just came to mind is that it could be a site actually wants to filter the value of the date to be something other than it is stored in the database. Maybe this is an extreme edge case, but it is something that we should keep on the lookout for. A site could always filter the metadata itself in this case.

PR #1723, opened 2018-12-10
Second argument of the_title filter is an Int not WP_Post.

lesterchan:
The `the_title` applies this filter `filter_the_title_in_post_list_table` with two arguments. The second argument is wrong. It is an Int, not a WP_Post object. See https://codex.wordpress.org/Plugin_API/Filter_Reference/the_title


westonruter:
See also #1725.

PR #1726, opened 2018-12-11
Bump 1.0.1

westonruter:


westonruter:
Here's a build of the plugin with #1725 and #1723 merged: [amp.zip](https://github.com/ampproject/amp-wp/files/2669777/amp.zip)

PR #1727, opened 2018-12-11
Add check for required iconv extension

westonruter:
The PHP-CSS-Parser requires `iconv`.

See https://wordpress.org/support/topic/paired-mode-causes-fatal-error/


kienstra:
**Approved**

Hi @westonruter,
Sorry for the delay. This notice looks good when I change the conditional (my environment has the `iconv` extension):

amp-iconv-extension-notice

PR #1728, opened 2018-12-12
Prevent image with caption from overflowing its container in Classic mode

kienstra:
Thanks to @lumay for raising this in:
https://wordpress.org/support/topic/amp-img-with-caption-center-problem/

# Steps To Reproduce
1. Select 'Classic' mode
2. Activate the Classic Editor plugin, assuming you're running WordPress `5.0`
3. Create a post
4. Add an image with a width of at least `1000px`, and select Full Size
5. Add a caption to the image
6. Update the post and go to its AMP URL
7. Expected: the image stays within its container
8. Actual: it overflows its container:

image-caption-overflows-container

# Background
This only looks to exist in the Classic editor, or when using the Classic block, or when adding a `[caption]` to a Shortcode block.

An image with a caption in the Classic editor creates a `figure.wp-caption`, which often has an inline width style value, like `<figure style="width: 1199px" ...>`

This can cause it to overflow the container.

There used to be logic in `AMP_Style_Sanitizer::filter_style()` [that converted 'width' to 'max-width'](https://github.com/ampproject/amp-wp/blob/da3a779c3ffad5c61c05bba4b83b680de274c5bd/includes/sanitizers/class-amp-style-sanitizer.php#L549).

Now that it's removed, this PR applies a simple style rule as a workaround.

This is in a stylesheet because it only applies to Classic mode.

**This issue doesn't seem to exist in Paired or Native mode**, where themes can add their own `max-width` styling.

# Before
captions-before

# After
captions-after

### DIFF
--- a/templates/style.php
+++ b/templates/style.php
@@ -241,6 +241,10 @@ blockquote p:last-child {
 	margin-left: 1em;
 }
 
+.amp-wp-article-content figure.wp-caption {
+	max-width: 100%;
+}
+
 .amp-wp-article-content amp-img {
 	margin: 0 auto;
 }
--- a/templates/style.php
+++ b/templates/style.php
@@ -241,7 +241,7 @@ blockquote p:last-child {
 	margin-left: 1em;
 }
 
-.amp-wp-article-content figure.wp-caption {
+.amp-wp-article-content .wp-caption {
 	max-width: 100%;
 }
 


kienstra:
**Request For Review**

Hi @westonruter and @amedina,
Could you please review this?

Instead of adding a style rule, a change like this would be possible [here in AMP_Img_Sanitizer:: add_auto_width_to_figure()](https://github.com/ampproject/amp-wp/blob/62c2fcd34b5cb493e788327d9b8fa2b56ce0f17b/includes/sanitizers/class-amp-img-sanitizer.php#L359). But what I came up with was more complex:

```php
/*
 * In Classic mode, prevents an image with a caption from overflowing its container.
 * In v0.7.2, AMP_Style_Sanitizer::filter_style() converted inline styling for width to max-width.
 * For example, it converted <figure style="width: 1000px" ...> to <figure style="max-width: 1000px" ...>.
 * This wouldn't be appropriate in Paired or Native mode, as themes can set this max-width in their stylesheets.
 */
if (
	! current_theme_supports( 'amp' )
	&&
	false === strpos( $class, 'wp-block-image' ) // This fix doesn't apply to the Image block.
	&&
	false !== strpos( $class, 'wp-caption' )
	&&
	$figure->hasAttribute( 'style' )
	&&
	preg_match( '/(^|[;\s])width:/', $figure->getAttribute( 'style' ) )
) {
	$figure->setAttribute( 'style', 'max-width: 100%;' . $figure->getAttribute( 'style' ) );
	return;
}
```

Thanks!

kienstra:
Thanks to @lumay's [feedback](https://wordpress.org/support/topic/amp-img-with-caption-center-problem/#post-10980164), the original commit didn't completely fix this.

https://github.com/ampproject/amp-wp/pull/1728/commits/b9e0463cbe625387b0dde828ff57830b2e755846 is needed to apply to cases where there's no theme support for HTML5.

In that case, the `caption` shortcode [uses a <div>](https://github.com/WordPress/wordpress-develop/blob/7ec7fadfe751795871caf4ca6ca34e92f1450925/src/wp-includes/media.php#L1627-L1628), not a `<figure>`.

So this removes `figure` from the selector.

# Before
images-that-have-captions-before

# After
images-that-have-captions-after

Issue #696, opened 2017-05-18
amp-iframe has negative margins


paulschreiber:
amp-iframe has negative left and right margins:

```
amp-iframe,
amp-youtube,
amp-instagram,
amp-vine {
	background: <?php echo sanitize_hex_color( $border_color ); ?>;
	margin: 0 -16px 1.5em;
}
```

Why is this?

westonruter:
See explanation in https://github.com/ampproject/amp-wp/pull/1740#issuecomment-447393386

PR #1740, opened 2018-12-13
fix css margins for embeds

mostafaabd:
- i test it with YouTube, Instagram embed and its display full width screen, so i remove margin that break layout.

Fix issue #696

### DIFF
--- a/templates/style.php
+++ b/templates/style.php
@@ -287,7 +287,6 @@ amp-youtube,
 amp-instagram,
 amp-vine {
 	background: <?php echo sanitize_hex_color( $border_color ); ?>;
-	margin: 0 -16px 1.5em;
 }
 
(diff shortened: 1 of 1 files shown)


westonruter:
Can you please add before/after screenshots?

Aside: This would only apply to Classic mode.

mostafaabd:
i test it and its only apply to classic mode.
my opinion:  Its look good with margin before i submit PR, full width is perfect for Youtube video or Instagram embed  for small screen resolution.

**before** After text copying and pasting the Youtube video is full width screen

amp_before

**after:**

amp-after

mostafaabd:
**Before:**

amp-2-before


**After:**

amp-2-after

mostafaabd:
yes, its look nice with negative margin

PR #1745, opened 2018-12-14
Introduce new sanitizers to streamline theme support for nav menu mobile toggle and submenu dropdowns

felixarntz:
One of the most common things theme developers still need to get their hands dirty with today when wanting to support AMP is dealing with the common hamburger menu for the navigation.

While there has been good documentation around on how to implement this, it's still quite clunky to do so - although what needs to happen is actually quite straightforward. The AMP plugin has so far automated the implementation for core themes already, only relying on some arguments like element IDs and CSS classes to use. With this PR I propose to streamline this so that _any_ WordPress theme can opt in and thus is no longer required to do the heavy lifting manually.

This PR ensures that a theme can add a `nav_menu_toggle` key when adding theme support, with the value for that being an array with further arguments. Example:

```php
add_theme_support(
    'amp',
    array(
        'nav_menu_toggle' => array(
            'nav_container_id'           => '...',
            'nav_menu_button_id'         => '...',
            'nav_container_toggle_class' => '...',
        ),
    )
);
```

The above is the minimum required. Alternatively to both IDs this PR also supports providing an Xpath query selector (in case the respective element does not use an ID), and you can also pass an optional `nav_menu_button_toggle_class` (in case the class on the button itself should change too).

A new `AMP_Nav_Menu_Toggle_Sanitizer` class is introduced to take care of this.

The logic for taking care of injecting the necessary `amp-state` and adjusting the markup I've mostly copied from `AMP_Core_Theme_Sanitizer::add_nav_menu_toggle()`. That method is now static (so that it is used as a buffering hook early) and has been simplified to dynamically add the required theme support - the heavy lifting is now part of `AMP_Nav_Menu_Toggle_Sanitizer`.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -708,6 +708,7 @@ function amp_get_content_sanitizers( $post = null ) {
 			),
 			'AMP_Block_Sanitizer'             => array(), // Note: Block sanitizer must come after embed / media sanitizers since it's logic is using the already sanitized content.
 			'AMP_Nav_Menu_Toggle_Sanitizer'   => ! empty( $theme_support_args['nav_menu_toggle'] ) ? $theme_support_args['nav_menu_toggle'] : array(),
+			'AMP_Nav_Menu_Dropdown_Sanitizer' => ! empty( $theme_support_args['nav_menu_dropdown'] ) ? $theme_support_args['nav_menu_dropdown'] : array(),
 			'AMP_Script_Sanitizer'            => array(),
 			'AMP_Style_Sanitizer'             => array(),
 			'AMP_Tag_And_Attribute_Sanitizer' => array(), // Note: This whitelist sanitizer must come at the end to clean up any remaining issues the other sanitizers didn't catch.
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -670,7 +670,9 @@ function amp_get_content_embed_handlers( $post = null ) {
  * @return array Embed handlers.
  */
 function amp_get_content_sanitizers( $post = null ) {
-	if ( current_theme_supports( AMP_Theme_Support::SLUG ) && $post ) {
+	$theme_support_args = AMP_Theme_Support::get_theme_support_args();
+
+	if ( is_array( $theme_support_args ) && $post ) {
 		_deprecated_argument( __FUNCTION__, '0.7', esc_html__( 'The $post argument is deprecated when theme supports AMP.', 'amp' ) );
 		$post = null;
 	}
@@ -702,9 +704,10 @@ function amp_get_content_sanitizers( $post = null ) {
 				'add_placeholder' => true,
 			),
 			'AMP_Gallery_Block_Sanitizer'     => array( // Note: Gallery block sanitizer must come after image sanitizers since itś logic is using the already sanitized images.
-				'carousel_required' => ! current_theme_supports( AMP_Theme_Support::SLUG ), // For back-compat.
+				'carousel_required' => ! is_array( $theme_support_args ), // For back-compat.
 			),
 			'AMP_Block_Sanitizer'             => array(), // Note: Block sanitizer must come after embed / media sanitizers since it's logic is using the already sanitized content.
+			'AMP_Nav_Menu_Toggle_Sanitizer'   => ! empty( $theme_support_args['nav_menu_toggle'] ) ? $theme_support_args['nav_menu_toggle'] : array(),
 			'AMP_Script_Sanitizer'            => array(),
 			'AMP_Style_Sanitizer'             => array(),
 			'AMP_Tag_And_Attribute_Sanitizer' => array(), // Note: This whitelist sanitizer must come at the end to clean up any remaining issues the other sanitizers didn't catch.
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -664,6 +664,7 @@ function amp_get_content_embed_handlers( $post = null ) {
  * Get content sanitizers.
  *
  * @since 0.7
+ * @since 1.1 Added AMP_Nav_Menu_Toggle_Sanitizer and AMP_Nav_Menu_Dropdown_Sanitizer.
  *
  * @param WP_Post $post Post that the content belongs to. Deprecated when theme supports AMP, as sanitizers apply
  *                      to non-post data (e.g. Text widget).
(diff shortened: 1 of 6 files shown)


includes/sanitizers/class-amp-nav-menu-toggle-sanitizer.php
```diff
@@ -0,0 +1,157 @@
...
+ */
+
+/**
+ * Class AMP_Nav_Menu_Toggle_Sanitizer
+ *
+ * Handles state for navigation menu toggles, based on theme support.
+ *
+ * @since 1.1.0
+ */
+class AMP_Nav_Menu_Toggle_Sanitizer extends AMP_Base_Sanitizer {
```
westonruter: What about making this instead a `AMP_Nav_Menu_Sanitizer`? Ideally this would handle not just the hamburger menu but also the injection of the submenu expand/collapse buttons. In other words, the logic in the `add_nav_sub_menu_buttons` method of the core theme sanitizer should be included here as well, in addition to the logic from the `add_nav_menu_toggle` class.

Eventually I would hope that AMP could actually provide a built-in nav menu component which could handle all of this as part of the framework instead of requiring each site to copy the same logic over and over again. When/if that happens, then much of this class could be removed in favor of just injecting the AMP component. The learnings from this sanitizer could also directly inform the creation of such a nav menu AMP component as well.

westonruter: Also, perhaps the `amp-sidebar` and `amp-accordion` are already the blessed AMP components for doing what WordPress does in a different way and there isn't anything else that AMP needs to add.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -976,61 +976,36 @@ public static function add_nav_menu_styles( $args = array() ) {
...
-			sprintf(
-				"%s + ( $state_id ? %s : '' )",
-				wp_json_encode( $nav_el->getAttribute( 'class' ) ),
-				wp_json_encode( ' ' . $args['nav_container_toggle_class'] )
-			)
+		$support['nav_menu_toggle'] = array(
+			'nav_container_id'             => $args['nav_container_id'],
+			'nav_menu_button_xpath'        => $args['menu_button_xpath'],
+			'nav_container_toggle_class'   => $args['nav_container_toggle_class'],
+			'nav_menu_button_toggle_class' => $args['menu_button_toggle_class'],
```
westonruter: Why not just use the same keys? This could just be:

```php
$support['nav_menu_toggle'] = $args;
```

felixarntz: Sounds good, I'll change it.

includes/sanitizers/class-amp-nav-menu-toggle-sanitizer.php
felixarntz: > What about making this instead a `AMP_Nav_Menu_Sanitizer`?

Sounds good to me. Is it okay if I rename it to that, and we add the extra submenu functionality in a follow-up PR? It wouldn't require any breaking change anyway, would only be an addition.

> Eventually I would hope that AMP could actually provide a built-in nav menu component 

Agreed, I'd be happy to get involved.

> Also, perhaps the `amp-sidebar` and `amp-accordion` are already the blessed AMP components

Possibly, but I wouldn't want to enable one of these by default as replacement for the navigation menu as it would modify behavior a bit. A future enhancement for using `amp-sidebar` instead of the current nav menu logic via opt-in (by theme support) sounds certainly interesting.

westonruter: > Is it okay if I rename it to that, and we add the extra submenu functionality in a follow-up PR? It wouldn't require any breaking change anyway, would only be an addition.

If you must, but this PR seems like the logical place to extract all the menu-specific code from the core theme sanitizer. Why not do it together?

felixarntz: > If you must, but this PR seems like the logical place to extract all the menu-specific code from the core theme sanitizer.

Alright, I'll do it here. Revisiting this though, I think we should use a separate class for the submenu toggle sanitizer, mainly because of how theme support should be added to these two things. They are not directly related, so I think it should be possible to add theme support individually. As in:

```php
add_theme_support(
    'amp',
    array(
        'nav_menu_toggle'           => array( ... ),
        'nav_menu_dropdown_toggles' => array( ... ),
    )
);
```

instead of just having one generic `nav_menu` key and cramping everything in there. The separation makes it more obvious what you're adding support for. Then each of the two arrays is passed to the respective sanitizer class.

includes/amp-helper-functions.php
```diff
@@ -702,9 +704,11 @@ function amp_get_content_sanitizers( $post = null ) {
 				'add_placeholder' => true,
 			),
 			'AMP_Gallery_Block_Sanitizer'     => array( // Note: Gallery block sanitizer must come after image sanitizers since itś logic is using the already sanitized images.
-				'carousel_required' => ! current_theme_supports( AMP_Theme_Support::SLUG ), // For back-compat.
+				'carousel_required' => ! is_array( $theme_support_args ), // For back-compat.
 			),
 			'AMP_Block_Sanitizer'             => array(), // Note: Block sanitizer must come after embed / media sanitizers since it's logic is using the already sanitized content.
+			'AMP_Nav_Menu_Toggle_Sanitizer'   => ! empty( $theme_support_args['nav_menu_toggle'] ) ? $theme_support_args['nav_menu_toggle'] : array(),
+			'AMP_Nav_Menu_Dropdown_Sanitizer' => ! empty( $theme_support_args['nav_menu_dropdown'] ) ? $theme_support_args['nav_menu_dropdown'] : array(),
```
westonruter: If theme support is not present, then these sanitizers actually should be omitted entirely. Right?

includes/sanitizers/class-amp-nav-menu-toggle-sanitizer.php
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	/**
+	 * If supported per the constructor arguments, inject `amp-state` and bind dynamic classes accordingly.
+	 *
+	 * @since 1.1.0
+	 */
+	public function sanitize() {
+		$theme_support_args = AMP_Theme_Support::get_theme_support_args();
```
westonruter: This needs to be looking at `$this->args` not theme support. Obtaining the global state here will be problematic for response caching.

includes/sanitizers/class-amp-nav-menu-dropdown-sanitizer.php
```diff
@@ -0,0 +1,164 @@
...
+
+	/**
+	 * Add filter to manipulate output during output buffering to add AMP-compatible dropdown toggles.
+	 *
+	 * @since 1.0
+	 *
+	 * @param array $args Args.
+	 */
+	public static function add_buffering_hooks( $args = array() ) {
+		$theme_support_args = AMP_Theme_Support::get_theme_support_args();
```
westonruter: Note that the concern below about calling `AMP_Theme_Support::get_theme_support_args()` inside of the `sanitize` method does not apply here for `add_buffering_hooks`. 

Nevertheless, should it not only be looking at `$args`? The `$args` is coming from the `amp_get_content_sanitizers()` function from an applied `amp_content_sanitizers` filter. So either the theme support args could be passed into that array at that point (as you're already doing above). So is this not redundant here anyway?

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -976,61 +976,31 @@ public static function add_nav_menu_styles( $args = array() ) {
...
-
-		$button_on = sprintf( "tap:AMP.setState({ $state_id: ! $state_id })" );
-		$button_el->setAttribute( 'on', $button_on );
-		$button_el->setAttribute( 'aria-expanded', 'false' );
-		$button_el->setAttribute( AMP_DOM_Utils::get_amp_bind_placeholder_prefix() . 'aria-expanded', "$state_id ? 'true' : 'false'" );
-		$button_el->setAttribute(
-			AMP_DOM_Utils::get_amp_bind_placeholder_prefix() . 'class',
-			sprintf( "%s + ( $state_id ? %s : '' )", wp_json_encode( $button_el->getAttribute( 'class' ) ), wp_json_encode( ' ' . $args['menu_button_toggle_class'] ) )
-		);
+		add_theme_support( AMP_Theme_Support::SLUG, $support );
```
westonruter: Reflecting on the comments I left below, I can see here that there is a chicken-and-egg problem in what I suggested as a fix. This logic is going to run after the `amp_get_content_sanitizers()` function will have already been called and the `amp_content_sanitizers` filter will have already been applied, so it is already too late for the core theme sanitizer to put args in to the sanitizers array.

Another alternative would be to include the theme support args in the `$response_cache_key`, but this also feels not ideal.

😔

includes/amp-helper-functions.php
felixarntz: Good catch!

includes/sanitizers/class-amp-core-theme-sanitizer.php
felixarntz: This is indeed a problem, the current architecture doesn't allow for fixing this properly.

I'm working on a solution that keeps the weirdness fairly small, implementing an additional separate mechanism for "extending" theme support.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -851,24 +890,24 @@ public static function add_nav_menu_styles( $args = array() ) {
 				}
 
 				/* Use sibling selector and re-use class on button instead of toggling toggle-on class on ul.sub-menu */
-				.main-navigation ul .<?php echo esc_html( $args['sub_menu_button_toggle_class'] ); ?> + .sub-menu {
```
westonruter: This may need to be reverted.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -205,66 +199,118 @@ public static function get_acceptable_errors( $template ) {
...
+				return array(
+					'nav_menu_toggle'   => array(
+						'nav_container_id'           => 'secondary',
+						'nav_container_toggle_class' => 'toggled-on',
+						'menu_button_xpath'          => '//header[ @id = "masthead" ]//button[ contains( @class, "secondary-toggle" ) ]',
+						'menu_button_toggle_class'   => 'toggled-on',
+					),
+					'nav_menu_dropdown' => array(
+						'sub_menu_button_class'        => 'dropdown-toggle',
 						'sub_menu_button_toggle_class' => 'toggle-on',
```
westonruter: This can be moved to `$theme_features`?

postphotos:
Awesome, thanks for your work here @felixarntz!

felixarntz:
@westonruter This should be good to merge now.

felixarntz:
@westonruter 
> Could you please update the description to reflect the latest state of the PR that I just merged?

Done!

PR #1746, opened 2018-12-14
Check if WP_CLI is defined and true

srtfisher:
Ensure that `WP_CLI` is true before loading


srtfisher:
I signed it!

westonruter:
> **connivance**
> n. willingness to secretly allow or be involved in wrongdoing, especially an immoral or illegal act.

Is that what you mean? 😄

PR #1748, opened 2018-12-18
Remove Merriweather font in favor of Georgia

### LINKED ISSUE #328: Don't load Merriweather by default

westonruter:
Fixes #328.


mqudsi:
I forgot to say thanks for this!

PR #1754, opened 2018-12-19
Fix Fatal error due to missing arguments

pradeep910:
I faced Fatal Error while using latest 1.0.1 version on a news site which is hosted on WordPress.com VIP platform. 

Looks like in case of concatenated CSS files, somehow its not passing 3rd argument `$href` to the function `amp_filter_font_style_loader_tag_with_crossorigin_anonymous` here -
https://github.com/ampproject/amp-wp/blob/1.0.1/includes/amp-helper-functions.php#L515

After some debugging I found that $tag contains -  
`<link rel='stylesheet' id='all-css-0' href='https://amp-wp.test/_static/??/wp-includes/css/dashicons.css,/wp-includes/css/admin-bar.css?m=1534398924j' type='text/css' media='all' />` 
and $handle contains `h4-global`.

This was the error - 
> `Fatal error: Uncaught ArgumentCountError: Too few arguments to function amp_filter_font_style_loader_tag_with_crossorigin_anonymous(), 2 passed in /wp/wp-includes/class-wp-hook.php on line 286 and exactly 3 expected in /wp-content/plugins/amp/includes/amp-helper-functions.php:515 Stack trace: #0 /wp/wp-includes/class-wp-hook.php(286): amp_filter_font_style_loader_tag_with_crossorigin_anonymous('<link rel='styl...', 'h4-global') #1 /wp/wp-includes/plugin.php(208): WP_Hook->apply_filters('<link rel='styl...', Array) #2 /wp-content/mu-plugins/http-concat/cssconcat.php(130): apply_filters('style_loader_ta...', '<link rel='styl...', 'h4-global') #3 /wp/wp-includes/functions.wp-styles.php(64): WPcom_CSS_Concat->do_items(Array) #4 /wp/wp-includes/class-wp-hook.php(286): wp_print_styles(false) #5 /wp/wp-includes/class-wp-hook.php(310): WP_Hook->apply_filter in /wp-content/plugins/amp/includes/amp-helper-functions.php on line 515`

To fix this error, I have just set the default value to 3rd argument `$href = ''`.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -513,6 +513,12 @@ function amp_filter_script_loader_tag( $tag, $handle ) {
  * @return string Link tag HTML.
  */
 function amp_filter_font_style_loader_tag_with_crossorigin_anonymous( $tag, $handle, $href = '' ) {
+
+	// In the case of concatenated CSS stylesheet links, $href is not being passed to function. Ref - <https://github.com/ampproject/amp-wp/pull/1754#issue-239925151>.
+	if ( empty( $href ) && preg_match( '#<link.+?href=[\'"](.+?)[\'"]#', $tag, $matches ) ) {
+		$href = html_entity_decode( $matches[1], ENT_QUOTES );
+	}
+
 	static $allowed_font_src_regex = null;
 	unset( $handle );
 	if ( ! $allowed_font_src_regex ) {
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -512,7 +512,7 @@ function amp_filter_script_loader_tag( $tag, $handle ) {
  * @param string $href   Link URL.
  * @return string Link tag HTML.
  */
-function amp_filter_font_style_loader_tag_with_crossorigin_anonymous( $tag, $handle, $href ) {
+function amp_filter_font_style_loader_tag_with_crossorigin_anonymous( $tag, $handle, $href = '' ) {
 	static $allowed_font_src_regex = null;
 	unset( $handle );
 	if ( ! $allowed_font_src_regex ) {


westonruter:
@pradeep910 thanks a lot for raising this. It looks like the underlying problem is that `WPcom_CSS_Concat` is not passing all of the required arguments to the `style_loader_tag` filter. You can see that the `$href` argument was introduced in WordPress 4.3:

https://github.com/WordPress/wordpress-develop/blob/d8968636b73092886a54bbe467acf1414afd85b0/src/wp-includes/class.wp-styles.php#L201-L213

This `$href` arg was added to Automattic/nginx-http-compat two years ago:

https://github.com/Automattic/nginx-http-concat/commit/dc302fea4db9a35c1cd25b6662c53b6b469f2ba4

Nevertheless, the `style_loader_tag` filter no longer is applying in Automattic/nginx-http-concat, being replaced with `ngx_http_concat_style_loader_tag`: 

https://github.com/Automattic/nginx-http-concat/commit/01c8291e9be45471f63355d3d5538ef590ec8fc9

Are you experiencing this fatal error in production? Or on a local dev environment? In either case, it seems the best fix is for the underlying Automattic/nginx-http-concat.

includes/amp-helper-functions.php
```diff
@@ -512,7 +512,7 @@ function amp_filter_script_loader_tag( $tag, $handle ) {
  * @param string $href   Link URL.
  * @return string Link tag HTML.
  */
-function amp_filter_font_style_loader_tag_with_crossorigin_anonymous( $tag, $handle, $href ) {
+function amp_filter_font_style_loader_tag_with_crossorigin_anonymous( $tag, $handle, $href = '' ) {
```
westonruter: Also, if we went with this, more would be needed as an empty `$href` would mean that the URL would need to be extracted from the `$tag` via regular expression.

pradeep910:
@westonruter You are right. `WPcom_CSS_Concat` is not passing all arguments. 

We haven't tried it in production yet, but it gives the same error on our staging so I think it will also give the error on production.

I think it's better to have this fix, to avoid accidental Fatal Error in case site is on old WPcom platform. 

Do you think WPcom using this yet? I'm not sure.
> Automattic/nginx-http-concat@01c8291

includes/amp-helper-functions.php
pradeep910: Yes. Concatenated CSS is being skipped currently because of not having `$href`.

But I believe we will have to add this regex to get `href` from `$tag`, if we have to go with this solution.

westonruter:
@philipjohn Can you shed light on this?

includes/amp-helper-functions.php
westonruter: That's right. The `$href` would need to be extracted from the `$tag` if it is empty. For example:

```php
if ( empty( $href ) && preg_match( '#<link.+?href="(.+?)"#', $tag, $matches ) ) {
    $href = html_entity_decode( $tag[1], ENT_QUOTES );
}
```

Ultimately this would be a fix for themes and plugins which are incorrectly applying the `style_loader_tag` filter.

pradeep910: @westonruter 
Made small changes to allow single or double quotes in regex.
```php 
if ( empty( $href ) && preg_match( '#<link.+?href=[\'"](.+?)[\'"]#', $tag, $matches ) ) {
    $href = html_entity_decode( $matches[1], ENT_QUOTES );
}
```

Since we get tag like this:
```
<link rel='stylesheet' id='all-css-0' href='https://amp-wp.test/_static/??/wp-includes/css/dashicons.css,/wp-includes/css/admin-bar.css?m=1534398924j' type='text/css' media='all' /> 
```

westonruter:
@pradeep910 Is this still an issue? If so, I suggest looping in someone from VIP via a Zendesk ticket.

pradeep910:
@westonruter Just got back from long vacation. I will ask someone from VIP about this. 
Closing this.

PR #1781, opened 2018-12-27
Prevent re-bundling WordPress packages

westonruter:
* Eliminate re-bundling of `@wordpress/dom-ready` and `@wordpress/i18n` to instead rely on existing bundles if available. Ensure both are marked as `devDependencies`.
* Include new standalone bundles of `@wordpress/dom-ready` and `@wordpress/i18n` if not available (e.g. in WordPress 4.9 without Gutenberg).
* Eliminate use of JSHint to avoid having to maintain duplicate duplicate linter as ESLint.

~The `@wordpress/dom-ready` package is bundled with both WordPress 5.0 and Gutenberg as `wp-dom-ready`, so we should use it.~

~Also, `@wordpress/i18n` is not even used in the project anymore, so we don't need it at all.~

~**Problem:** If on 4.9 and Gutenberg is not installed, then we need to polyfill `wp-dom-ready` with `jQuery.ready()`. Nevertheless, we are currently depending on `wp-i18n` even though it may not be available either. So we might need to make the plugin require WordPress 5.0 or WordPress 4.9 with Gutenberg as the minimum version. Or perhaps in that case we should still bundle those dependencies as separate scripts that we register? If we want to support 4.9 without Gutenberg, then I suppose this is what we'd have to do. @felixarntz thoughts?~

Closes #1763.


felixarntz:
@westonruter Couldn't we bundle the `wp-i18n` and `wp-dom-ready` scripts in compiled versions and register them if they aren't yet available (by core)? I think that's the least error-prone solution, doesn't require too many changes. Whenever we decide to drop 4.x support, we can remove them without further change.

assets/src/amp-validation-tooltips.js
```diff
@@ -17,6 +12,5 @@ function sourcesPointer() {
 	} );
 }
 
-domReady( () => {
-	sourcesPointer();
-} );
+// Run at DOM ready.
+jQuery( sourcesPointer );
```
felixarntz: I might miss something, but shouldn't this use `wp.domReady` somehow?

includes/amp-helper-functions.php
```diff
@@ -334,6 +334,16 @@ function amp_add_generator_metadata() {
  */
 function amp_register_default_scripts( $wp_scripts ) {
 
+	// Polyfill dependencies that are registered in Gutenberg and WordPress 5.0.
+	if ( ! has_action( 'wp_default_scripts', 'gutenberg_register_packages_scripts' ) ) {
```
felixarntz: I think it's okay having this, but strictly it isn't needed right? It only covers the plugin anyway I think?

felixarntz: How that? Shouldn't we just _always_ want to check whether these two scripts are registered, and if not do so ourselves?

westonruter: That being said, `gutenberg_register_packages_scripts()` explicitly accounts for this case in `gutenberg_register_packages_scripts()` by [calling `gutenberg_override_script()`](https://github.com/WordPress/gutenberg/blob/80be381921ce818c0381e877f510bbc0be0fcf24/lib/client-assets.php#L144-L159) which does `wp_deregister_script()` first. So it looks like the condition is indeed unnecessary.

westonruter:
Ok, issues have been addressed.

PR #1789, opened 2018-12-31
Improve performance of make-pot call

westonruter:
Without this `--exclude` argument, running `time lando grunt shell:makepot` takes ~73 seconds to complete. What's more is that the generated `amp-js.pot` file includes references to files in the `build/` directory even though this directory is not included in the `--include` argument. When the `--exclude` is included, the time to run `time lando grunt shell:makepot` cuts down to ~41 seconds. Even though this is entirely too long to run (which is likely due to a Lando problem), it shows that adding the `--exclude` is cutting down on processing files unnecessarily.

This PR also uses Composer to explicitly mark the `wp-cli/i18n-command` dependency with a pinned version.

Lastly, it adds a missing `translators` comment.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -38,7 +38,8 @@ module.exports = function( grunt ) {
 				command: 'npm run pot-to-php && php -l languages/amp-translations.php'
 			},
 			makepot: {
-				command: 'wp i18n make-pot . languages/amp-js.pot --include="*.js" --file-comment="*/null/*"' // The --file-comment is a temporary workaround for <https://github.com/ampproject/amp-wp/issues/1416>.
+				// TODO: The exclude argument to make-pot does not seem to actually exclude. Without including the --exclude argument, the generated amp-js.pot still includes references to the build/ directory.
+				command: 'wp i18n make-pot . languages/amp-js.pot --include="assets/**/*.js,blocks/**/*.js" --exclude="*.*,back-compat/,bin/,build/,dev-lib/,includes/,languages/,node_modules/,templates/,tests/,third_party/,vendor/,wpcom/" --file-comment="*/null/*"' // The --file-comment is a temporary workaround for <https://github.com/ampproject/amp-wp/issues/1416>.
 			},
 			create_build_zip: {
 				command: 'if [ ! -e build ]; then echo "Run grunt build first."; exit 1; fi; if [ -e amp.zip ]; then rm amp.zip; fi; cd build; zip -r ../amp.zip .; cd ..; echo; echo "ZIP of build: $(pwd)/amp.zip"'
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -39,7 +39,7 @@ module.exports = function( grunt ) {
 			},
 			makepot: {
 				// TODO: The exclude argument to make-pot does not seem to actually exclude. Without including the --exclude argument, the generated amp-js.pot still includes references to the build/ directory.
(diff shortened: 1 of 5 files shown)


Gruntfile.js
```diff
@@ -38,7 +38,8 @@ module.exports = function( grunt ) {
 				command: 'npm run pot-to-php && php -l languages/amp-translations.php'
 			},
 			makepot: {
-				command: 'wp i18n make-pot . languages/amp-js.pot --include="*.js" --file-comment="*/null/*"' // The --file-comment is a temporary workaround for <https://github.com/ampproject/amp-wp/issues/1416>.
+				// TODO: The exclude argument to make-pot does not seem to actually exclude. Without including the --exclude argument, the generated amp-js.pot still includes references to the build/ directory.
```
westonruter: @swissspidy Can you advise regarding this?

westonruter:
@felixarntz Does running `lando grunt shell:makepot` on your machine similarly take a long time to complete?

swissspidy:
Quick question: Does the plugin still support WordPress 4.9 with the Gutenberg plugin or is Gutenberg support aimed at WordPress 5.0 only?

westonruter:
Support is intended for 4.9 without Gutenberg as well. For example https://github.com/ampproject/amp-wp/pull/1781

swissspidy:
WordPress.org should handle string extraction of JavaScript files by now, as you can see here: https://translate.wordpress.org/projects/wp-plugins/amp/dev/en-gb/default?filters%5Boriginal_id%5D=7004670

That also means WordPress.org ships language packs with JSON translation files in them since WordPress 5.0. So there's no need to manually build POT files anymore.

Note: I published an in-depth guideline for WordPress JS I18N on my website: https://pascalbirchler.com/internationalization-in-wordpress-5-0/

`wp_set_script_translations` handles loading these JSON files for you.

`wp_get_jed_locale_data` only existed in the 5.0 Betas but has been removed after that. Any checks for that function should be removed as it does not exist.

`gutenberg_get_jed_locale_data` is for older versions of Gutenberg running on WordPress 4.9. It loads the translations from the PO files. IIRC WordPress.org also adds the JS translations to PO files, so that should work for these people.

westonruter:
@swissspidy ok, thank you. So we can eliminate `makepot` entirely?

So given this example: 

https://github.com/ampproject/amp-wp/blob/8f7a60ac3484ca6aa7b1ba603af33046a482ca3e/includes/admin/class-amp-editor-blocks.php#L162-L171

Are you saying that we don't need the `elseif` at all anymore? But I see that `wp_set_script_translations()` is not part of Gutenberg, so would 4.9 installs using Gutenberg just not be able to have translated strings?

Also, will `wp_set_script_translations()` take the _entire_ plugin language pack for the locale and export it to JSON? It doesn't seem there is a way to subset it for what is actually needed in JS? So this means that PHP-only strings would be exported as well, right?

swissspidy:
Yep, there's no need for that anymore. WordPress.org handles the make-pot part nowadays.

> Are you saying that we don't need the elseif at all anymore? But I see that `wp_set_script_translations()` is not part of Gutenberg, so would 4.9 installs using Gutenberg just not be able to have translated strings?

4.9 + Gutenberg should still have `gutenberg_get_jed_locale_data()`.

> Also, will `wp_set_script_translations()` take the entire plugin language pack for the locale and export it to JSON?

No :-) That's the whole point of this new function. It creates multiple little JSON files, see https://make.wordpress.org/core/2018/11/09/new-javascript-i18n-support-in-wordpress/ for more info.

However, `gutenberg_get_jed_locale_data()` does that.

You could shim `wp_set_script_translations()` in the plugin though in that case.

westonruter:
@swissspidy thank you very much. I've amended the PR then to eliminate POT generation entirely. The result is that there is that there is no longer a `languages/amp-translations.php` in the build which will be committed to WordPress.org. Can you confirm if the PR looks good now?

PR #1790, opened 2019-01-01
Make get_validated_url_file_path() better compatible with register_theme_directory()

westonruter:
In the wordpressdev Lando setup, [`register_theme_directory()` is used](https://github.com/felixarntz/wordpressdev/blob/01721d8aeb010b7608615857a0a854c82f3ade6c/public/content/mu-plugins/allow-using-default-themes.php#L13) to ensure that core-bundled themes can continue to be used in addition to the themes defined in the `WP_CONTENT_DIR`. This is causing a problem for `\AMP_Style_Sanitizer::get_validated_url_file_path()`:

image

In order to fix this specific problem, we need to explicitly allow `ABSPATH . 'wp-content'` as a filesystem directory that files can be referenced from.

I don't believe this is the ultimate solution for any issue that may arise from having custom directories for themes and plugins, but for the time being at least it fixes this issue in the Lando environment.

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -543,6 +543,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 		$includes_url = $remove_url_scheme( includes_url( '/' ) );
 		$content_url  = $remove_url_scheme( content_url( '/' ) );
 		$admin_url    = $remove_url_scheme( get_admin_url( null, '/' ) );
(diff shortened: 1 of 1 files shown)


westonruter:
@felixarntz I thought about that but we'd also need to figure out the URL for the theme directory. The case here was easy because it is just the site URL plus `wp-content`. So I think we should do that eventually, but this fixes the immediate problem.

felixarntz:
@westonruter Sounds good. Would be great to track this in a follow-up issue so we don't forget it, even (or especially) if we don't wanna work on it right away.

westonruter:
Thanks. Opened #1793 to track this.

PR #1794, opened 2019-01-04
Add _doing_it_wrong() when calling is_amp_endpoint() before queried object is available

westonruter:
Ever since 0.4.2, the AMP plugin has issued a `_doing_it_wrong()` when `is_amp_endpoint()` is called before the `parse_query` action. This was needed to ensure that the `amp` query var was parsed and available for inspection. 

It turns out that this did not go far enough. As of 1.0 the `is_amp_endpoint()` needs to look at the queried object as well to determine whether AMP has been enabled/disabled for a given post on singular queries. The queried object is not available at `parse_query` since the posts have not been queried yet at this point. Rather, it is only at the `wp` action that [we can be assured](https://github.com/WordPress/wordpress-develop/blob/67e12d960b166b12f7eee511bc8abf798533a872/src/wp-includes/class-wp.php#L737-L750) that `$wp_query->post` has been populated and thus that the queried object is present for inspection.

Also make sure that only a `WP_Post` is passed into `AMP_Post_Type_Support::get_support_errors()`.

Without the changes in this PR, if you have a plugin that does:

```php
add_action( 'parse_query', function() {
	if ( function_exists( 'is_amp_endpoint' ) && is_amp_endpoint() ) {
			echo "HELLO PAIRED AMP!";
		} );
		add_action( 'amp_post_template_footer', function() {
			echo "HELLO CLASSIC AMP!";
		} );
	} else {
			echo "HELLO NOT AMP!";
		} );
	}
} );
```

You will get PHP notices when viewing a singular post in paired mode:

> Notice: Trying to get property 'post_type' of non-object in /app/public/content/plugins/amp/includes/class-amp-post-type-support.php on line 80
> Notice: Trying to get property 'ID' of non-object in /app/public/content/plugins/amp/includes/class-amp-post-type-support.php on line 101

Changing `parse_query` to `wp` fixes the issue in plugin code.

Issue discovered originally in a Jetpack PR: https://github.com/Automattic/jetpack/pull/10945#issuecomment-451299275

### DIFF
--- a/includes/widgets/class-amp-widget-text.php
+++ b/includes/widgets/class-amp-widget-text.php
@@ -22,6 +22,7 @@ if ( class_exists( 'WP_Widget_Text' ) ) {
 		 * @return string $html The markup, unaltered.
 		 */
 		public function inject_video_max_width_style( $matches ) {
+			wp();
 			if ( is_amp_endpoint() ) {
 				return $matches[0];
 			}
--- a/includes/widgets/class-amp-widget-text.php
+++ b/includes/widgets/class-amp-widget-text.php
@@ -22,7 +22,6 @@ if ( class_exists( 'WP_Widget_Text' ) ) {
 		 * @return string $html The markup, unaltered.
 		 */
 		public function inject_video_max_width_style( $matches ) {
-			wp();
 			if ( is_amp_endpoint() ) {
 				return $matches[0];
 			}
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -283,8 +283,15 @@ function is_amp_endpoint() {
 		return true;
 	}
 
-	$availability = AMP_Theme_Support::get_template_availability();
-	return amp_is_canonical() ? $availability['supported'] : ( $has_amp_query_var && $availability['supported'] );
+	if ( ! did_action( 'wp' ) ) {
+		_doing_it_wrong( __FUNCTION__, sprintf( esc_html__( "is_amp_endpoint() was called before the 'wp' action which means it will not have access to the queried object to determine if it is an AMP response, thus neither the amp_skip_post filter nor the AMP enabled publish metabox toggle will not be considered.", 'amp' ) ), '1.0.2' );
+		$supported = true;
+	} else {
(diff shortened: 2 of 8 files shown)


includes/widgets/class-amp-widget-text.php
```diff
@@ -22,6 +22,7 @@ class AMP_Widget_Text extends WP_Widget_Text {
 		 * @return string $html The markup, unaltered.
 		 */
 		public function inject_video_max_width_style( $matches ) {
+			wp();
```
felixarntz: This worries me a bit. Why do we need to run this manually in regular code?

westonruter: Because `is_amp_endpoint()` now warns of `_doing_it_wrong` if the `wp` action has not been done.

westonruter: Why is it worrisome?

felixarntz: This function bootstraps WordPress, I think it shouldn't ever be called by anything but WordPress core itself. What is the problem here that might cause `wp()` to not have run before?

westonruter: It doesn't bootstrap WordPress. It just sets up the main WordPress query. 

Unit tests reset `$wp` after each test run:

https://github.com/WordPress/wordpress-develop/blob/67e12d960b166b12f7eee511bc8abf798533a872/tests/phpunit/includes/testcase.php#L162

And they essentially call `wp()` when calling the unit test `go_to()` method:

https://github.com/WordPress/wordpress-develop/blob/67e12d960b166b12f7eee511bc8abf798533a872/tests/phpunit/includes/testcase.php#L631-L637

felixarntz: Right, but this instance here is not in a test, that's what gets me curious. Again, why is this necessary? What might cause `wp()` to not have run before a call to this?

westonruter: Oh, woops. You're right. I added it to the wrong place.

westonruter: Fixed: 744202d

PR #1805, opened 2019-01-10
Use constants instead of strings in AMP_Post_Meta_Box::sanitize_status()

paulschreiber:


paulschreiber:
@googlebot  I signed it!

PR #1807, opened 2019-01-11
[WIP] Fix reading of small SVG images to determine dimensions

### LINKED ISSUE #204: Special handling for smileys
If we have an image with the `wp-smiley` smiley class, it should probably display inline and have a forced size. Currently, it gets inferred as `layout=responsive`, which isn't great.

westonruter:
This issue came up via this support topic: https://wordpress.org/support/topic/how-to-show-emoji-show-share-button-and-hide-powered-by-on-amp-page/

With this raw post content:

```html
<p>Image: <img class="emoji loaded" draggable="false" src="https://s.w.org/images/core/emoji/11/svg/1f642.svg" alt="🙂"></p>

<p>Character: 🙂</p>
```

In HTML non-AMP this gets rendered as:

image

Note that the emoji character is being converted into an image by WordPress's emoji handling. In AMP this is disabled, so the character passes through as a character but the image is incorrectly rendered in AMP:

image

The underlying HTML is:

```html
<p>Image: <amp-img class="emoji loaded amp-wp-unknown-size amp-wp-unknown-width amp-wp-unknown-height amp-wp-enforced-sizes" draggable="false" src="https://s.w.org/images/core/emoji/11/svg/1f642.svg" alt="🙂" width="525" height="400" layout="intrinsic"></amp-img></p>

<p>Character: 🙂</p>
```

The problem turns out to be an issue with `FasterImage` and how it was modified in https://github.com/ampproject/amp-wp/pull/1150 to allow extracting pixel dimensions from SVG images. The problem is that it is attempting to read 1024 bytes of data from the image, but the [above SVG](https://s.w.org/images/core/emoji/11/svg/1f642.svg) is only 525 bytes long. This results in a `WillWashburn\Stream\StreamBufferTooSmallException` exception being thrown, and the image size failing to be obtained, and thus the fallback image size is being used (which is obviously too big).

So this PR fixes that problem by allowing small SVG images to be read without throwing an error.

# Work Outstanding

## Before

image

## After

image

While this is way better, it is still not perfect. The SVG image has a viewbox of 36x36 so that is what is assigned as the `width` & `height` of the `amp-img`. Nevertheless, the emojis are supposed to be 16x16. Or rather, they are [supposed to be `1em` by `1em`](https://github.com/WordPress/wordpress-develop/blob/33caf61b8ba988da73c2e210309566f9fdb039be/src/wp-includes/formatting.php#L5350-L5351), so it should look like this:

image

So to full fix the case of inline SVG emojis we potentially need to either replace the `img.emoji` back with the underlying character (which is found in the `alt` attribute). This would not be ideal because we'd miss out on the improvements that WordPress's emoji handling was introduced for.

A full fix seems to be two-fold:

* Make emoji SVG images get sizes at `1em` somehow. This is a challenge because `amp-img` doesn't allow `em` units.
* Make an AMP-compatible version of WordPress's emoji handling, by adding a sanitizer that does in PHP what core does in JS. In the AMP plugin we disable the emoji handling in AMP:

https://github.com/ampproject/amp-wp/blob/c92e129333a8518088a37d47594079b537121d70/includes/class-amp-theme-support.php#L766-L767

These may be out of scope for this PR.

Other outstanding work:

- [ ] Fix unit test.
- [ ] Replace cURL with WP HTTP API (which would also help with mocking the uint tests).

Fixes #204.

### DIFF
--- a/third_party/fasterimage/ImageParser.php
+++ b/third_party/fasterimage/ImageParser.php
@@ -136,7 +136,7 @@ class ImageParser
                     return $this->type = 'tiff';
                 default:
                     $this->stream->resetPointer();
-                    $markup = $this->stream->read( 1024 );
+                    $markup = $this->stream->read( 1024, false );
                     if ( false !== strpos( $markup, '<svg' ) ) {
                         $this->type = 'svg';
                     } else {
@@ -361,7 +361,7 @@ class ImageParser
     protected function parseSizeForSvg()
     {
         $this->stream->resetPointer();
-        $markup = $this->stream->read( 1024 );
+        $markup = $this->stream->read( 1024, false );
         if ( ! preg_match( '#<svg.*?>#s', $markup, $matches ) ) {
             return null;
         }
--- a/third_party/fasterimage/Stream/Stream.php
+++ b/third_party/fasterimage/Stream/Stream.php
@@ -25,14 +25,15 @@ class Stream implements StreamableInterface
     /**
      * Get characters from the string but don't move the pointer
      *
-     * @param $characters
+     * @param int $characters Number of characters to read.
+     * @param bool $check_length Throw exception if there are not enough bytes left in the stream.
      *
      * @return string
      * @throws StreamBufferTooSmallException
      */
-    public function peek($characters)
+    public function peek($characters, $check_length=true)
     {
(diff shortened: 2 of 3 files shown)


westonruter:
Closing in favor of https://github.com/willwashburn/fasterimage/pull/14

PR #1809, opened 2019-01-14
Bump minimum PHP version requirement to 5.4 and include FasterImage library dynamically

felixarntz:
As pointed out in chat, we should address the unnecessary TLS verification being disabled in the FasterImage library. While doing that, a few other issues arised:

* We currently also rely on the FastImage library, solely for the sake of supporting PHP 5.3.
* We manually include the FasterImage library with custom tweaks made to it, with the TLS modification becoming just another one.

This PR implements the following changes:
* Bump minimum required PHP version to 5.4 and thus ditches the FastImage library.
* Pull in FasterImage library in the latest version via Composer, thus also getting rid of the need of bundling third-party code in the repository.
* Apply custom modifications via https://github.com/cweagans/composer-patches, making it more obvious what has been changed by us and making it seamless to update the library while keeping our changes in place.
* In addition to the previous modifications, TLS verification is enabled in the `curl_setopt()` calls made by the library.

Considerations:
* What is the current process of packaging the plugin for release? We need to ensure the `patches` folder is ignored in that process.
* As a more general observation not directly related to this change, we should ensure when packaging for release the dependencies are installed on the correct environment (PHP 5.4), otherwise we risk installing them in a higher version. Not sure whether this has been considered or tested before. Note that Composer takes the version specification in `composer.json` as minimum, not as "the dependencies _have_ to be installed in a version compatible with that".


includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -172,38 +172,7 @@ private static function determine_which_images_to_fetch( &$dimensions, &$urls_to
 	 * @param string $mode Whether image dimensions should be extracted concurrently or synchronously.
 	 */
 	private static function fetch_images( $urls_to_fetch, &$images, $mode ) {
-		// Use FasterImage when for compatible PHP versions
-		if ( 'synchronous' === $mode ||
```
westonruter: What about the `$mode` parameter? Shouldn't it be eliminated?

westonruter: There is also a `$mode` argument on `\AMP_Image_Dimension_Extractor::extract_by_downloading_images()`. Should an error be raised if it is anything other than `concurrent`? Should it be removed?

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -172,38 +172,7 @@ private static function determine_which_images_to_fetch( &$dimensions, &$urls_to
...
-			false === function_exists( 'curl_multi_exec' ) ||
```
westonruter: This `curl_multi_exec()` call gives me pause. Is it possible for this function to not exist in 5.4, say if an older version of cURL is used? We've seen something like this when an older version of ICU is present: https://github.com/ampproject/amp-wp/pull/1440

westonruter:
The build is just failing due to a PHPCS error:

```
/includes/class-amp-autoloader.php:59:55: warning - Array double arrow not aligned correctly; expected 10 space(s) between "'AMP_YouTube_Embed_Handler'" and double arrow, but found 19. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
```

westonruter:
The PHPCS fixes are opening a snowballing can of worms 😄

westonruter:
We should actually do a PHPCS cleanup task to get the entire repo up-to-date with the latest WPCS 1.2.x standard anyway.

westonruter:
Question: Why go the route of patches vs forking the repo and commit directly do it? That's what I did for PHP-CSS-Parser. Seems it could be easier to maintain the patches that way?

felixarntz:
@westonruter I opened #1810 to update and fix all coding standards errors. Let's merge that one first, then I'll pull it in here, fix merge conflicts to make it pass.

includes/utils/class-amp-image-dimension-extractor.php
felixarntz: `curl_multi_exec()` should be available on any PHP5 version as the PHP docs state, at least their is no distinction between different sets of `curl_*()` functions. In other words, if `curl_exec()` exists, we can also assume `curl_multi_exec()` exists. Note that WordPress core does that as well, in their cURL implementation they use `curl_multi_exec()` too and don't specifically check its existance. I think we should move forward with this.

felixarntz:
@westonruter 
> Why go the route of patches vs forking the repo and commit directly do it? That's what I did for PHP-CSS-Parser. Seems it could be easier to maintain the patches that way?

I think for simple tweaks the patching workflow is more pragmatic since that allows us to manage everything from this repository. Looking at the PHP-CSS-Parser, the custom tweaks there are more complex, so that justifies a fork.

I suggest we use the patching approach for now with this dependency. If and once we need to make more changes, we can use our best judgement to at some point decide the patching approach isn't viable anymore, and then we could move to a forked repository.

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -193,54 +201,15 @@ private static function determine_which_images_to_fetch( &$dimensions, &$urls_to
...
 	 *
 	 * @param array $urls_to_fetch Image src urls to fetch.
 	 * @param array $images Array to populate with results of image/dimension inspection.
 	 */
-	private static function fetch_images_via_faster_image( $urls_to_fetch, &$images ) {
+	private static function fetch_images( $urls_to_fetch, &$images ) {
 		$urls       = array_keys( $urls_to_fetch );
 		$user_agent = apply_filters( 'amp_extract_image_dimensions_get_user_agent', self::get_default_user_agent() );
-		$client     = new \FasterImage\FasterImage( $user_agent );
+		$client     = new \FasterImage\FasterImage( $user_agent ); // @todo The $user_agent is not actually able to be passed in this way to FasterImage. Needs another patch?
```
westonruter: This is something we should fix as well.

westonruter: @felixarntz What is the workflow to add a new patch?

I'm guessing we will want to do something like this here:

```diff
--- a/wp-content/plugins/amp/third_party/fasterimage/FasterImage.php
+++ b/wp-content/plugins/amp/third_party/fasterimage/FasterImage.php
@@ -25,7 +25,25 @@ class FasterImage
      */
     protected $timeout = 10;
 
-    /**
+	/**
+	 * User agent.
+	 *
+	 * @var string
+	 */
+	protected $user_agent = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_8_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/27.0.1453.110 Safari/537.36';
+
+	/**
+	 * FasterImage constructor.
+	 *
+	 * @param string $user_agent User agent.
+	 */
+	public function __construct( $user_agent = '' ) {
+		if ( $user_agent ) {
+			$this->user_agent = $user_agent;
+		}
+	}
+
+	/**
      * Get the size of each of the urls in a list
      *
      * @param array $urls
@@ -115,7 +133,7 @@ class FasterImage
         curl_setopt($ch, CURLOPT_TIMEOUT, $this->timeout);
 
         #  Some web servers require the useragent to be not a bot. So we are liars.
-        curl_setopt($ch, CURLOPT_USERAGENT, 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_8_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/27.0.1453.110 Safari/537.36');
+        curl_setopt($ch, CURLOPT_USERAGENT, $this->user_agent);
         curl_setopt($ch, CURLOPT_HTTPHEADER, [
             "Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5",
             "Cache-Control: max-age=0",
```

felixarntz: I fixed this now.

What I currently do is I have a local checkout of the FasterImage library in the tagged version that we're using. I apply the existing patch against it, change code, recreate the patch again.

Note that you can't just use the `master` branch since it contains code that isn't part of a release yet. However, for future releases, our patch file will just apply until something in that same area changes, which would naturally cause a merge conflict. Composer would notify us about that.

felixarntz:
@westonruter 
> This is a great question. This has not been a consideration that has been made when making a release. But which dependencies would be a problem here?

Generally, anything for which we don't specify a fixed version number is a problem. For example, if we say `^1` for a package, and version 1.2 relies on PHP 5.4 and 1.3 relies on PHP 5.5, and our setup is >= 5.5, it would use the later version, which is obviously a problem.

However, it is a good practice to not specify fixed versions, but rather rely on semantic versioning (I suggest we use `^8` for the PHP-CSS-Parser, `^1` for FasterImage, `^1` for WPCS, `^0.4` for the installer and so on). The specific versions are supposed to be managed via `composer.lock`, and developers should always use `composer install` unless they intend to update the dependencies for the whole project.

I added a `config --> platform` entry to `composer.json` which will ensure that Composer ignores the local setup it is run from and always installs dependencies in a way that are 5.4-compatible. This makes it bullet-proof, and furthermore helps making sure all of us contributing to the plugin are on the same page regarding dependencies as well.

westonruter:
> However, it is a good practice to not specify fixed versions, but rather rely on semantic versioning (I suggest we use `^8` for the PHP-CSS-Parser, `^1` for FasterImage, `^1` for WPCS, `^0.4` for the installer and so on). The specific versions are supposed to be managed via `composer.lock`, and developers should always use `composer install` unless they intend to update the dependencies for the whole project.

Normally yes, but we don't need to worry about that now because Renovate will take care of making sure that each dependency is updated. It will set the dependency on fixed versions so that we can be sure that a tagged version will always have the exact same versioned dependencies at the time of tagging. It's also true that the `composer.lock` would make the fixed version in `composer.json` a redundancy. So essentially we can delegate `composer update` to Renovate.

felixarntz:
@westonruter That's fine with me then. We should still keep the `config --> platform` entry in `composer.json` to ensure we don't break anything. Even for the case that Renovate doesn't know that we always need to be 5.4-compatible. :)

westonruter:
Great changes! 🎉 Good work.

includes/utils/class-amp-image-dimension-extractor.php
westonruter: Apparently this turned out to not be true: https://github.com/ampproject/amp-wp/pull/2183#issuecomment-491506514

PR #1811, opened 2019-01-15
Use JSON_UNESCAPED_UNICODE for comment state if available

### LINKED ISSUE #1791: Comment reply not working with umlauts
To reproduce:

1. Install AMP plugin
2. Activate Twenty Nineteen or any other theme that works well in native or paired mode
3. Write comment using a name with umlauts, e.g. "Max Müller".
4. Reload page, click on "Reply" to reply to that specific comment
5. Notice that "Max Müller" is displayed as "Max M\u00fcller".

Screenshot:

screenshot 2019-01-02 at 16 16 30

swissspidy:
Fixes #1791 without requiring a PHP version bump. Also means that the bug will still exist for anyone on PHP 5.3 or older, which I think is acceptable.


westonruter:
@swissspidy Ready for merging from `develop` and updating per above.

PR #1816, opened 2019-01-16
Use home_url() instead of site_url() where applicable

felixarntz:
In some locations the codebase uses `site_url()`, which however may not necessarily be the URL of the WordPress site, but rather the URL in which WordPress core is located. This is relevant for setups like https://github.com/felixarntz/wordpressdev or https://github.com/roots/bedrock which have core in a subdirectory of their webroot.

`home_url()` should be used to reference the actual website URL (which is confusing, but yeah). I hope I understood these code areas correctly and caught the instances where this should be replaced (because there are other areas where `site_url()` needs to continue to be used).


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -317,7 +317,7 @@ public function __construct( DOMDocument $dom, array $args = array() ) {
 			}
 		}
 
-		$guessurl = site_url();
+		$guessurl = home_url();
```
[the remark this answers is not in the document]
westonruter: But, maybe we need more than one `base_url` to account for registered theme directories?

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -101,7 +101,7 @@ public static function normalize_url( $url ) {
 			if ( isset( $parsed['query'] ) ) {
 				$path .= '?' . $parsed['query'];
 			}
-			$url = site_url( $path );
+			$url = home_url( $path );
```
westonruter: The `site_url()` is intended to get the URL based on the actual location.

What matters is the the host name only. So if a URL comes in like `/wp-content/uploads/foo.png` then what matters is that this resolve to an actual URL when combined with the site/home URL. Let's use WordPress.com as an example:

* `home_url()`: `https://example.com/wp-content/uploads/foo.png`
* `site_url()`: `https://example.wordpress.com/wp-content/uploads/foo.png`

Here, both will resolve since the wordpress.com subdomains are aliases for the custom domains. Where this would fail is if `site_url()` is a subdirectory of the domain (say at `https://example.com/wordpress/`), which I think is the crux of the issue here, according to your PR.

In that case, `site_url( $path )` would fail because it would result in a URL like:

`https://example.com/wordpress/wp-content/uploads/foo.png`

When in actuality the URL needed would be:

`https://example.com/wp-content/uploads/foo.png`

So we need to take just the schema and host of _either_ the home URL or site URL and combine with the full path of the URL. So that would be concatenating `https://example.com` with `/wp-content/uploads/foo.png`.

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -284,6 +284,6 @@ private static function process_fetched_images( $urls_to_fetch, $images, &$dimen
 	 * @return string
 	 */
 	public static function get_default_user_agent() {
-		return 'amp-wp, v' . AMP__VERSION . ', ' . get_site_url();
+		return 'amp-wp, v' . AMP__VERSION . ', ' . home_url();
```
westonruter: This makes sense to me.

includes/utils/class-amp-image-dimension-extractor.php
felixarntz: Correct, that was my point. By using `home_url()`, it is much more likely that the URL determined will be correct, as home URLs don't typically use a subdirectory. Alternatively (and possibly even better), we should ensure that whatever the home/site URL is, we just use the host and scheme parts from it.

PR #1810, opened 2019-01-15
Update coding standards and fix all violations

felixarntz:
~This PR updates WPCS and PHPCompatibility to its latest versions. It furthermore ensures that they stay up to date within non-breaking versions, with the concrete versions of course continuing to be managed with `composer.lock` (run `composer install` to install the tested versions and `composer update` to potentially update).~

Furthermore this PR fixes all violations that have been around in the codebase, and it removes the old `// WPCS: ... ok.` annotations that are no longer preferable and will be deprecated in the future.

PR #1822, opened 2019-01-16
Instruct wp-dev-lib to check all files not just those in patch

westonruter:
After #1810 we no longer need to limit PHPCS to report issues on the lines that were changes in a given PR, which is [the default behavior of wp-dev-lib](https://github.com/xwp/wp-dev-lib/blob/38dbb83cd63d2e537ccd4ddcc2866ab8cd57cbd9/readme.md#limiting-scope-of-checks). This was needed because the repo did not adhere to the standards, and so the patch scope limitation allowed us to make incremental improvements over time. This is no longer needed, and the patch scope limitation is going to cause us problems going forward such as when versions of WPCS are updated by Renovate (e.g. #1814), since the `composer.json` change wouldn't trigger `phpcs` by default.

### DIFF
--- a/.dev-lib
+++ b/.dev-lib
@@ -4,6 +4,7 @@ PROJECT_SLUG=amp
 SKIP_ECHO_PATHS_SCOPE=1
 README_MD_TITLE="AMP Plugin for WordPress"
 DEV_LIB_SKIP="$DEV_LIB_SKIP,jshint"
+CHECK_SCOPE=all
 
 function after_wp_install {
     echo "Installing plugins..."


westonruter:
See from Travis build that PHPCS is now being invoked even without any PHP files being touched: https://travis-ci.org/ampproject/amp-wp/jobs/480632846#L381-L382

The downside to this change is that the build will take longer, but it seems justified for the sake of consistency.

felixarntz:
LGTM. However, for an eventual pre-commit hook we should ensure PHPCS still only runs on the changed files to speed that up. For CI I agree it's not so important as consistency.

westonruter:
Agreed. For that perhaps we should do something like this:

```bash
if [ -z $TRAVIS ]; then
    CHECK_SCOPE=all
else
    CHECK_SCOPE=changed-files
fi
```

westonruter:
> Agreed. For that perhaps we should do something like this:

Done in https://github.com/ampproject/amp-wp/pull/1131/commits/f431b297be888a28b0bff0289dfb18498302e744

PR #1825, opened 2019-01-17
[WIP] Add development environment

### LINKED ISSUE #1823: Introduce a development environment
Define a local development environment that contributors can use to run and test code.

kasparsd:
Fixes #1823.

- [x] Use Docker containers defined in `docker-compose.yaml` running inside Vagrant for better hostname isolation (enables `http://ampwp.local`).

- [ ] Document system requirements and setup.


kasparsd:
I signed it!

felixarntz:
I would generally advise against putting a dev environment into a specific project's repository. IMO this plugin should provide the tools needed to develop for it (via NPM and Composer), but it should be agnostic about which dev environment to use.

_If_ there is consensus that a specific dev environment for the AMP plugin makes sense, it should be a separate repository, that could be linked to from the readme for example.

kasparsd:
Thanks for the feedback @felixarntz!

The idea behind a simplified development environment was to simplify the effort required for new developers to contribute to the project. For example, running phpunit currently requires a development machine with a full WP setup and files in particular locations. Having a minimal development environment with the suggested defaults could significantly reduce the time it takes to start contributing to the project.

A separate repo sounds like a good solution!

PR #1828, opened 2019-01-17
Fix building assets

westonruter:
* After #1781 there were introduced separate `wp-i18n` and `wp-dom-ready` scripts which are polyfills for 5.0. However, the source JS files were placed in the wrong directory so they were getting included in the build. So 645466c moves them to `assets/src` so that only their Webpack `*-compiled.js` versions are included in the `assets/js` for the build.
* Also after #1781 there was no longer a reason to include `amp-validation-tooltips.js` in a Webpack build, so cc3929e just moves it from `assets/src` to `assets/js` so that it can be loaded directly from source.
* Lastly, the build was erroneously including `assets/src` in the build, so a335f91 excludes it.


westonruter:
Here's a build of this branch for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/2771606/amp.zip)

PR #1829, opened 2019-01-17
Add double quotes around the ETag header value; recognize multiple quoted ETags in If-None-Match header

ataylorme:
`ETag` headers being served by the AMP plugin on Pantheon were not working. After some troubleshooting, adding double quotes around the `ETag` header value fixes the issue.

The [`ETag` spec](https://tools.ietf.org/html/rfc7232#section-2.3) explain the values as:

> An entity-tag consists of an opaque quoted string, possibly prefixed by a weakness indicator.

The examples use double quotes as well.

`ETag` might be working elsewhere where the provider has software that is more forgiving but our Nginx requires the quotes.

This should not be a breaking change as other, more forgiving, implementations should continue to work when the `ETag` header value is set according to the specification.

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1652,7 +1652,7 @@ class AMP_Theme_Support {
 		 * not have been set yet during the WordPress template generation is
 		 * the ETag. The AMP plugin sends a Vary header at amp_init.
 		 */
-		AMP_HTTP::send_header( 'ETag', $response_cache_key );
+		AMP_HTTP::send_header( 'ETag', '"' . $response_cache_key . '"' );
 
 		// Handle responses that are cached by the browser.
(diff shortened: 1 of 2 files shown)


westonruter:
Thank you. I've amended the PR with 788106c which ensures that clients that make requests with quoted ETags in `If-None-Match` headers will be successfully matched. This matching will be done regardless of the weak validator. 

So if WordPress returned with:

```http
ETag: "dbcf759d515248c889b3765d17fdae31"
```

Then a client request will get a `304 Not Modified` response with any of the following `If-None-Match` request headers:

```http
If-None-Match: dbcf759d515248c889b3765d17fdae31
If-None-Match: W/"dbcf759d515248c889b3765d17fdae31"
If-None-Match: "d3b07384d113edec49eaa6238ad5ff00", "dbcf759d515248c889b3765d17fdae31"
```

PR #1834, opened 2019-01-20
Update class-amp-http.php

s-a-s-k-i-a:
In PHP 7.2/7.3 calling idn_to_* functions with default arguments throws a warning. Thus we must set the variant and options explicitely.
lines 208 - 212 fixed

### DIFF
--- a/includes/class-amp-http.php
+++ b/includes/class-amp-http.php
@@ -205,7 +205,7 @@ class AMP_HTTP {
 			if ( function_exists( 'idn_to_utf8' ) ) {
 				if ( version_compare( PHP_VERSION, '5.4', '>=' ) && defined( 'INTL_IDNA_VARIANT_UTS46' ) ) {
 					$domain = idn_to_utf8( $domain, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46 ); // phpcs:ignore PHPCompatibility.FunctionUse.NewFunctionParameters.idn_to_utf8_variantFound, PHPCompatibility.Constants.NewConstants.intl_idna_variant_uts46Found
-				} elseif ( version_compare( PHP_VERSION, '7.2', '==' ) || version_compare( PHP_VERSION, '7.3', '==' ) ) {
+				} elseif ( version_compare( PHP_VERSION, '7.2', '>=' ) && version_compare( PHP_VERSION, '7.4', '<' ) ) {
 					/* In PHP 7.2/7.3 calling idn_to_* functions with default arguments throws a warning. Thus we must set the variant explicitely. This may be reverted when php 7.4 is commonly used and idn_to_utf8 can be used in its default mode without warnings. */
 					$variant = INTL_IDNA_VARIANT_UTS46;
 					$options = IDNA_DEFAULT;
--- a/includes/class-amp-http.php
+++ b/includes/class-amp-http.php
@@ -205,8 +205,8 @@ class AMP_HTTP {
 			if ( function_exists( 'idn_to_utf8' ) ) {
 				if ( version_compare( PHP_VERSION, '5.4', '>=' ) && defined( 'INTL_IDNA_VARIANT_UTS46' ) ) {
 					$domain = idn_to_utf8( $domain, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46 ); // phpcs:ignore PHPCompatibility.FunctionUse.NewFunctionParameters.idn_to_utf8_variantFound, PHPCompatibility.Constants.NewConstants.intl_idna_variant_uts46Found
-				} else if( version_compare( PHP_VERSION, '7.2', '==' ) || version_compare( PHP_VERSION, '7.3', '==' ) ){
-					/* In PHP 7.2/7.3 calling idn_to_* functions with default arguments throws a warning. Thus we must set the variant explicitely */
+				} elseif ( version_compare( PHP_VERSION, '7.2', '==' ) || version_compare( PHP_VERSION, '7.3', '==' ) ) {
+					/* In PHP 7.2/7.3 calling idn_to_* functions with default arguments throws a warning. Thus we must set the variant explicitely. This may be reverted when php 7.4 is commonly used and idn_to_utf8 can be used in its default mode without warnings. */
 					$variant = INTL_IDNA_VARIANT_UTS46;
 					$options = IDNA_DEFAULT;
 					$domain = idn_to_utf8( $domain, $options, $variant );
--- a/includes/class-amp-http.php
+++ b/includes/class-amp-http.php
@@ -205,6 +205,11 @@ class AMP_HTTP {
 			if ( function_exists( 'idn_to_utf8' ) ) {
 				if ( version_compare( PHP_VERSION, '5.4', '>=' ) && defined( 'INTL_IDNA_VARIANT_UTS46' ) ) {
 					$domain = idn_to_utf8( $domain, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46 ); // phpcs:ignore PHPCompatibility.FunctionUse.NewFunctionParameters.idn_to_utf8_variantFound, PHPCompatibility.Constants.NewConstants.intl_idna_variant_uts46Found
+				} else if( version_compare( PHP_VERSION, '7.2', '==' ) || version_compare( PHP_VERSION, '7.3', '==' ) ){
+					/* In PHP 7.2/7.3 calling idn_to_* functions with default arguments throws a warning. Thus we must set the variant explicitely */
+					$variant = INTL_IDNA_VARIANT_UTS46;
+					$options = IDNA_DEFAULT;
+					$domain = idn_to_utf8( $domain, $options, $variant );
 				} else {
 					$domain = idn_to_utf8( $domain );
 				}
--- a/includes/class-amp-http.php
+++ b/includes/class-amp-http.php
@@ -209,7 +209,7 @@ class AMP_HTTP {
 					/* In PHP 7.2/7.3 calling idn_to_* functions with default arguments throws a warning. Thus we must set the variant explicitely. This may be reverted when php 7.4 is commonly used and idn_to_utf8 can be used in its default mode without warnings. */
 					$variant = INTL_IDNA_VARIANT_UTS46;
 					$options = IDNA_DEFAULT;
-					$domain = idn_to_utf8( $domain, $options, $variant );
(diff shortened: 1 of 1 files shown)


s-a-s-k-i-a:
I signed it!

s-a-s-k-i-a:
Check again..

includes/class-amp-http.php
```diff
@@ -205,6 +205,11 @@ public static function get_amp_cache_hosts() {
 			if ( function_exists( 'idn_to_utf8' ) ) {
 				if ( version_compare( PHP_VERSION, '5.4', '>=' ) && defined( 'INTL_IDNA_VARIANT_UTS46' ) ) {
 					$domain = idn_to_utf8( $domain, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46 ); // phpcs:ignore PHPCompatibility.FunctionUse.NewFunctionParameters.idn_to_utf8_variantFound, PHPCompatibility.Constants.NewConstants.intl_idna_variant_uts46Found
+				} else if( version_compare( PHP_VERSION, '7.2', '==' ) || version_compare( PHP_VERSION, '7.3', '==' ) ){
```
westonruter: Check the Travis job for the PHPCS issues that are causing the build to fail: https://travis-ci.org/ampproject/amp-wp/jobs/481959474#L400

westonruter:
See https://github.com/ampproject/amp-wp/issues/1439 and https://github.com/ampproject/amp-wp/pull/1440 for where the existing code came from.

See also https://github.com/Automattic/jetpack/pull/10945#issuecomment-451229686 and https://github.com/Automattic/jetpack/pull/10945#issuecomment-451234961. This would fix the issue reported there?

includes/class-amp-http.php
westonruter: Wouldn't it be safer to check if `version_compare( PHP_VERSION, '7.2', '>=' )`? Is it expected to change in PHP 7.4?

s-a-s-k-i-a:
Check this: https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003 <https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003>
For current php 7.2 and 7.3its wise to explicitly set the variant to INTL_IDNA_VARIANT_UTS46 and add a comment that this change can be reverted when PHP 7.4 is the new norm and idn_to_utf8() can be used in its default mode without warnings.

includes/class-amp-http.php
```diff
@@ -205,6 +205,11 @@ public static function get_amp_cache_hosts() {
 			if ( function_exists( 'idn_to_utf8' ) ) {
 				if ( version_compare( PHP_VERSION, '5.4', '>=' ) && defined( 'INTL_IDNA_VARIANT_UTS46' ) ) {
 					$domain = idn_to_utf8( $domain, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46 ); // phpcs:ignore PHPCompatibility.FunctionUse.NewFunctionParameters.idn_to_utf8_variantFound, PHPCompatibility.Constants.NewConstants.intl_idna_variant_uts46Found
+				} elseif ( version_compare( PHP_VERSION, '7.2', '==' ) || version_compare( PHP_VERSION, '7.3', '==' ) ) {
```
westonruter: If only PHP 7.2 and 7.3 are to be targeted then this also needs to account for patch versions, yes? This would fail on PHP 7.2.1, right?

```suggestion
				} elseif ( version_compare( PHP_VERSION, '7.2', '>=' ) && version_compare( PHP_VERSION, '7.4', '<' ) ) {
```

includes/class-amp-http.php
```diff
@@ -205,6 +205,11 @@ public static function get_amp_cache_hosts() {
...
+					/* In PHP 7.2/7.3 calling idn_to_* functions with default arguments throws a warning. Thus we must set the variant explicitely. This may be reverted when php 7.4 is commonly used and idn_to_utf8 can be used in its default mode without warnings. */
```
westonruter:
```suggestion
					/*
					 * In PHP 7.2/7.3 calling idn_to_* functions with default arguments throws a warning.
					 * Thus we must set the variant explicitly. This may be reverted when PHP 7.4 is commonly used and
					 * idn_to_utf8 can be used in its default mode without warnings.
					 */
```

includes/class-amp-http.php
```diff
@@ -205,6 +205,15 @@ public static function get_amp_cache_hosts() {
 			if ( function_exists( 'idn_to_utf8' ) ) {
 				if ( version_compare( PHP_VERSION, '5.4', '>=' ) && defined( 'INTL_IDNA_VARIANT_UTS46' ) ) {
 					$domain = idn_to_utf8( $domain, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46 ); // phpcs:ignore PHPCompatibility.FunctionUse.NewFunctionParameters.idn_to_utf8_variantFound, PHPCompatibility.Constants.NewConstants.intl_idna_variant_uts46Found
+				} elseif ( version_compare( PHP_VERSION, '7.2', '>=' ) && version_compare( PHP_VERSION, '7.4', '<' ) ) {
```
felixarntz: I think this condition is never reached? Also, it probably shouldn't if `INTL_IDNA_VARIANT_UTS46` is not defined.

s-a-s-k-i-a: Yes, thats true.. I just didn't want to mess with someone else's code, because it is always possible someone else knows something I might not know :D

swissspidy: @s-a-s-k-i-a I think what he meant is that if you are using PHP 7.2 or PHP 7.3, the first condition (`version_compare( PHP_VERSION, '5.4', '>=' ) && defined( 'INTL_IDNA_VARIANT_UTS46' )`) is already `true`, so PHP will never even get to the `version_compare( PHP_VERSION, '7.2', '>=' ) && version_compare( PHP_VERSION, '7.4', '<' )` condition (unless `INTL_IDNA_VARIANT_UTS46` is not defined).

This begs the question: why is this change needed?

According to http://php.net/manual/en/function.idn-to-utf8.php, `INTL_IDNA_VARIANT_2003` (the default value for PHP 7.2 and PHP 7.3) was deprecated.

However, the current code means we always explicitly use `INTL_IDNA_VARIANT_UTS46`, no matter if you are running 5.4, 5.6, 7.2, or 7.3:

```php
if ( version_compare( PHP_VERSION, '5.4', '>=' ) && defined( 'INTL_IDNA_VARIANT_UTS46' ) ) {
    $domain = idn_to_utf8( $domain, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46 );
}
```

Thus, I don't see why you are getting any warnings.

If you perhaps could explain this to me so that I can better understand this, that would be great 🙂

s-a-s-k-i-a: Hi @swissspidy 
In one of my wp projects I was seeing this php warning while in debug mode:

`idn_to_utf8(): INTL_IDNA_VARIANT_2003 is deprecated...`
related to the edited file (class-amp-http.php) and lines of code I modified in order to fix this warning.
Because in that particular environment, obviously the variant is not set. Thus the first condition is false and my code applies.
Warning gone, all fine.

That is why I wrote this code.. not because I thought I'd add something useless to this project ;)

swissspidy: So far so clear. My only concern is that this should have been fixed by #1440 already. That change meant that if `INTL_IDNA_VARIANT_UTS46` is defined, we will use it.

So to me the only explanation for that warning you get is that you are using PHP 7.2/7.3 which was compiled with an old version of ICU where `INTL_IDNA_VARIANT_UTS46` does not exist yet.

Can you check the output of `phpinfo()` / `php -i` and paste the output here? The relevant part should look like this:

```
intl

Internationalization support => enabled
ICU version => 63.1
ICU Data version => 63.1
ICU TZData version => 2018e
ICU Unicode version => 11.0

Directive => Local Value => Master Value
intl.default_locale => no value => no value
intl.error_level => 0 => 0
intl.use_exceptions => 0 => 0
```

s-a-s-k-i-a: pls give me some time to do this. I fixed this a few weeks ago and for the life of mine, I can not remember which client's server it was, where I saw this warning, which made me fix this. 
I will have to check through all sites I implemented AMP lately 💃

swissspidy: @s-a-s-k-i-a No worries, thanks for your efforts!

s-a-s-k-i-a: I finally came across the client's server where I encountered this and fixed it using my pull requested code.

The error is of cource back after updating amp plugin, as you didnt let my fix merge.. I will fix it again manually.

This is the phpinfo:
https://www.ruhrgebietaktuell.de/info.php

Thanks

swissspidy: Thanks a lot @s-a-s-k-i-a!

Please be reminded that it's not that we don't _want_ to merge your fix, we just want to  better _understand_ the root cause of the issue and whether this fix is the _right_ approach.

I had a look at your phpinfo and this is the relevant section:

Screenshot 2019-03-17 at 13 56 09

This says that your version of PHP was compiled with ICU version **4.4.1**, which was released in **2010**. As you can imagine, this is **very old**.

Now, the problem is that PHP **7.2** (the version you are using) uses `INTL_IDNA_VARIANT_UTS46` as the default value for `idn_to_utf8()`. As the name suggests, this constant was added in ICU version 4.6, which was released in **2011**.

Now, the combination on your client's server - PHP 7.2 with a very old ICU version - is quite an edge case.

I think we can address this edge case a bit more elegantly in our plugin, and I will create a new pull request for this.

However, you should really reach out to the **hosting provider** and kindly request them to **update the PHP INTL extension** (which includes ICU). Perhaps an update to PHP 7.3 also does the trick, in case their version includes an updated ICU library.

@s-a-s-k-i-a Does that sound good to you?

Ps. I am also happy to answer questions about this in German if that's preferable.

s-a-s-k-i-a: Well the hosting provider is Mittwald and even though I see the issue here, there is still the fact, that we could simply provide a fix for websites that run on php versions compiled with older ICUs.

swissspidy: @s-a-s-k-i-a That is correct, and also what I said:

> I think we can address this edge case a bit more elegantly in our plugin, and I will create a new pull request for this.

swissspidy: New PR: #1987

swissspidy:
_New PR: #1987_

PR #1843, opened 2019-01-29
Suppress style[amp-custom] manifest HTML comment when not WP_DEBUG

westonruter:
The HTML comment that contains the manifest of what comprises the `style[amp-custom]` element can get quite large, so large in fact that Lighthouse will flag it in the performance audit. So the comment should be able to be excluded.

I suggest that we only show the comment, by default, when `WP_DEBUG` is enabled. We currently only send the `Server-Timing` headers when `WP_DEBUG` is enabled (or an admin user is logged-in).

We should still by default show the manifest when there is any CSS excluded, as this is critical for debugging in production without turning `WP_DEBUG` on. So by default we can have the manifest comment hidden unless `WP_DEBUG` is on or there is any CSS being excluded. 

So, the default value for this option is `when_css_excluded`.

If a site wants to override this and force the comment to be displayed, even when there is no excluded:

```php
    $sanitizers['AMP_Style_Sanitizer']['include_manifest_comment'] = 'always';
    return $sanitizers;
} );
```

Likewise, to always hide the comment even when `WP_DEBUG` is enabled:

```php
    return $sanitizers;
} );
```

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -59,7 +59,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 *      @type bool     $should_locate_sources      Whether to locate the sources when reporting validation errors.
 	 *      @type string   $parsed_cache_variant       Additional value by which to vary parsed cache.
 	 *      @type bool     $accept_tree_shaking        Whether to accept tree-shaking by default and bypass a validation error.
-	 *      @type bool     $include_manifest_comment   Whether to show the manifest HTML comment in the response before the style[amp-custom] element.
+	 *      @type string   $include_manifest_comment   Whether to show the manifest HTML comment in the response before the style[amp-custom] element. Can be 'always', 'never', or 'when_css_excluded'.
 	 * }
 	 */
 	protected $args;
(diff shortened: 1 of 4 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1986,9 +1985,13 @@ private function finalize_styles() {
 			}
 
 			$include_manifest_comment = (
-				'always' === $this->args['include_manifest_comment']
-				||
-				( $excluded_size > 0 && 'when_css_excluded' === $this->args['include_manifest_comment'] )
+				'never' !== $this->args['include_manifest_comment']
```
swissspidy: @westonruter Why did you add this? The value of `include_manifest_comment` is already tested in the second condition. This won't change the logic.

tests/test-amp-style-sanitizer.php
```diff
@@ -759,6 +759,7 @@ public function test_font_data_url_handling() {
 		$sanitizer   = new AMP_Style_Sanitizer(
 			$dom,
 			array(
+				'include_manifest_comment'  => AMP_Style_Sanitizer::INCLUDE_MANIFEST_COMMENT_WHEN_EXCESSIVE,
```
swissspidy: This is nowhere defined

westonruter: 😊

PR #1611, opened 2018-11-12
Fix CSS parsing by stripping BOM

kienstra:
# Steps to reproduce
1. Activate the [Essence Pro](https://my.studiopress.com/themes/essence/) Genesis theme (or use the `wp_enqueue_style()` snippet below in another theme)
2. View the source of the front page
3. Expected: the [ionicons stylesheet](https://unpkg.com/ionicons@4.1.2/dist/css/ionicons.min.css?ver=1.0.2) should have more that `0` bytes in the `<style amp-custom>`:
ionicons-expected
4. Actual: the stylesheet has 0 bytes...it seems that it's stripped in tree-shaking:
ionicons-actual

This [stylesheet](https://unpkg.com/ionicons@4.1.2/dist/css/ionicons.min.css?ver=1.0.2) is enqueued in the Essence Pro theme's `functions.php`:

```php
wp_enqueue_style(
	'ionicons',
	'//unpkg.com/ionicons@4.1.2/dist/css/ionicons.min.css',
	array(),
	CHILD_THEME_VERSION
);
```

This PR outputs the `ionicons` `@font-face`, as shown in the 'Expected' screenshot. Still, this approach probably needs more thought.

This issue doesn't look to exist on other Genesis themes that use an `ionicons` stylesheet. They use a different URL for the stylesheet, and it seems to appear fine.

PR #1844, opened 2019-01-29
Strip old-school CDATA and HTML comments from XHTML-compatible style elements

westonruter:
Reported at https://wordpress.org/support/topic/amp-issue-css-syntax-error-style-amp-custom-2/

Google Search Console was reporting a CSS parse error on an AMP page containing a style:

```css
<![CDATA[.mailpoet_hp_email_label{display:none}
```

The use of CDATA and HTML comments in styles is only for the sake of supporting XHTML back in the day. It is obsolete now. In any case, PHP-CSS-Parser doesn't strip it out, just as it doesn't strip out the BOM (#1611) so we need to do it ourselves.


westonruter:
Latest build from `develop`: [amp.zip](https://github.com/ampproject/amp-wp/files/2958312/amp.zip) (v1.1-alpha-20190312T185122Z-b15042b3)

PR #1848, opened 2019-02-01
Fix tests in environments where custom content dir is set and WP_HOME/WP_SITEURL is defined

westonruter:
Fixes tests running in the [wordpressdev Lando environment](https://github.com/felixarntz/wordpressdev):

image

In this environment, the `WP_HOME`/`WP_SITEURL` constants are set, and this causes problems for tests that tried changing the `home` and `siteurl` options. So to fix that problem, filters are now used instead.

Also, the environment registers a custom directory for the content, which broke tests related to theme path URLs. So now the core themes directory is explicitly referenced in the test.

### DIFF
--- a/tests/test-amp-style-sanitizer.php
+++ b/tests/test-amp-style-sanitizer.php
@@ -1138,7 +1138,9 @@ class AMP_Style_Sanitizer_Test extends WP_UnitTestCase {
 	public function get_stylesheet_urls() {
 
 		// Make sure core-bundled themes are registered.
-		register_theme_directory( ABSPATH . 'wp-content/themes' );
+		if ( WP_CONTENT_DIR !== ABSPATH . 'wp-content/themes' ) {
+			register_theme_directory( ABSPATH . 'wp-content/themes' );
+		}
 
 		$theme = new WP_Theme( 'twentyseventeen', ABSPATH . 'wp-content/themes' );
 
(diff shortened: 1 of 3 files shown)


tests/validation/test-class-amp-validated-url-post-type.php
```diff
@@ -311,9 +311,22 @@ public function test_get_url_from_post() {
...
 
 		// Check URL scheme.
+		add_filter(
+			'home_url',
+			function ( $url ) {
+				return set_url_scheme( $url, 'http' );
+			},
+			10
+		);
 		update_option( 'home', home_url( '/', 'http' ) );
```
felixarntz: Shouldn't this need to be ditched?

tests/test-amp-style-sanitizer.php
```diff
@@ -1136,14 +1136,20 @@ public function test_keyframe_sanitizer( $source, $expected = null, $expected_er
 	 * @returns array Stylesheet URL data.
 	 */
 	public function get_stylesheet_urls() {
+
+		// Make sure core-bundled themes are registered.
+		register_theme_directory( ABSPATH . 'wp-content/themes' );
```
felixarntz: Maybe only do this conditionally if `ABSPATH . 'wp-content'` is actually different from `WP_CONTENT_DIR`?

tests/validation/test-class-amp-validated-url-post-type.php
westonruter: Yeah. Removed in fe02465c.

westonruter:
Follow-up: https://github.com/felixarntz/wordpressdev/pull/24

PR #1852, opened 2019-02-05
Fall-back to HTTP request whenever stylesheet is not found locally or a dynamic resource

### LINKED ISSUE #1797: In multisite setup theme fonts URLs render wrong
I am converting my theme AMP compatible. Fonts URLs are working fine with single sites. But in multisite same theme fonts, URLs are wrong. It is showing fonts not found.

### LINKED ISSUE #1849: Make get_validated_url_file_path function compatible with non-static files
In some cases and projects we need to dynamically generate CSS.
If the URL is `domain.com/?action=my-theme-get-styles&foo=bar` and not a static file like `domain.com/something.css` then the validation fails due to the missing file extension and the CSS never gets parsed by AMP

westonruter:
When referencing an external stylesheet, the plugin currently is over-eager to fail the reference when it doesn't look like a CSS resource. In particular, if the URL doesn't have a `.css` extension. For such cases, it's important to fall-back to fetching the stylesheet over HTTP. So this is now implemented here. Additionally, it will very that the fetched URL is actually `text/css`.

This PR also improves handling of relative URLs in stylesheets, both for the stylesheet URLs themselves and for resources linked to in the stylesheets (e.g. background images and fonts).b

The converting of relative paths to absolute ones was not very robust. For example, a path like `foo/./bar/baz/../../../style.css` would fail to be converted `style.css`. So this is now fixed.

This should fix the following two issues, but it needs testing.

Fixes #1849.
Fixes #1797.

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -542,7 +542,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			);
 		} while ( 0 !== $count );
 
-		if ( false !== strpos( $path, './' ) ) {
+		if ( preg_match( '#(^|/)\.+/#', $path ) ) {
 			/* translators: %s is the path with the remaining relative segments. */
 			return new WP_Error( 'remaining_relativity', sprintf( __( 'There are remaining relative path segments: %s', 'amp' ), $path ) );
 		}
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -712,35 +712,26 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			return;
 		}
 
+		// @todo Try resolving relative paths like '../'.
 		$css_file_path = $this->get_validated_url_file_path( $href, array( 'css', 'less', 'scss', 'sass' ) );
-
-		if ( is_wp_error( $css_file_path ) && ( 'disallowed_file_extension' === $css_file_path->get_error_code() || 'external_file_url' === $css_file_path->get_error_code() ) ) {
+		if ( ! is_wp_error( $css_file_path ) ) {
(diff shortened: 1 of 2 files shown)


westonruter:
Here is an alpha build of the plugin from this branch for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/2830342/amp.zip)

westonruter:
Changes confirmed to work as expected in https://github.com/ampproject/amp-wp/issues/1849#issuecomment-461583340

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -513,6 +513,74 @@ public function sanitize() {
...
+			$path = preg_replace(
+				'#(?<=/)(?!\.\./)[^/]+/\.\./#',
+				'',
+				$path,
+				1,
+				$count
+			);
+		} while ( 0 !== $count );
+
+		if ( false !== strpos( $path, './' ) ) {
```
felixarntz: Is this accurate enough? Just wondering if someone named their directories weird (`mypath/my-dir-with-dot./something.css`)... edge-case, but should we rather use regex here?

westonruter: How about:

```php
if ( preg_match( '#/\.+/#', $path ) ) {
```

westonruter: Actually, I think we can remove this condition entirely as it is overkill. It shouldn't happen because of the first do/while loop. And if it fails then it will be caught below in:

https://github.com/ampproject/amp-wp/blob/9473cbd9aeb1af1ac0c3191895e3456520c4b161/includes/sanitizers/class-amp-style-sanitizer.php#L674-L677

westonruter: Meh, let's keep it as `unrelativize_path` is called in another place.

westonruter:
Here's a latest build from `develop`: [amp.zip](https://github.com/ampproject/amp-wp/files/2986526/amp.zip) (1.1-alpha-20190320T043537Z-49d5f16e)

PR #1854, opened 2019-02-05
Add Security Disclosures section from amphtml project to contributing.md

westonruter:
We don't currently have any information about what to do in the case a security issue is uncovered. So this PR incorporates the same snippet as is used in the amphtml project: https://github.com/ampproject/amphtml/#security-disclosures


swissspidy:
It's increasingly common for projects on GitHub to add a `SECURITY.md` file to the repository root.

Should we add the same snippet to such a file as well?

westonruter:
@swissspidy done, although I named it `security.md` so all markdown files are consistently lower-case.

PR #1855, opened 2019-02-05
Enqueue component scripts for attributes that require extensions (e.g. amp-lightbox-gallery for amp-img[lightbox])

### LINKED ISSUE #1808: The amp-lightbox-gallery component script is not enqueued when lightbox attribute is present on images
Hi!
amp-wp v1.0.1
wp v5.0.3

Theme setup:
```php
add_theme_support(
    'amp',
    array(
        'paired' => true,
        'template_dir' => 'templates-amp',
        'templates_supported' => 'all',
    )
);
```
I'm trying to load the [amp-lightbox-gallery](https://ampbyexample.com/components/amp-lightbox-gallery/) component:
```php
add_action(
    'amp_post_template_data',
    function ($data) {
        // Not work:
        // $data['amp_component_scripts']['amp-lightbox-gallery'] = true;
        // Not Work:
        array_merge(
            $data['amp_component_scripts'],
            array(
                'amp-lightbox-gallery' => 'https://cdn.ampproject.org/v0/amp-lightbox-gallery-0.1.js'
            )
        );

        return $data;
    }
);
```
But nothing happens. I can not understand what I'm doing wrong.

The example above should be enough to load an additional component or additional actions are needed?

Search by [project](https://github.com/ampproject/amp-wp/search?q=amp-lightbox-gallery&unscoped_q=amp-lightbox-gallery) has not brought results.

```php
add_action(
    'wp_head',
    function () {
        echo '<script async custom-element="amp-lightbox-gallery" src="https://cdn.ampproject.org/v0/amp-lightbox-gallery-0.1.js"></script>';
    }
);
```
This code works, but I'm not sure that this way is correct...

westonruter:
There seems to be a bug in the tag-and-attribute sanitizer where it is not detecting the need for the `amp-lightbox-gallery` this component when an element has the `lightbox` attribute. The problem is in the following logic:

https://github.com/ampproject/amp-wp/blob/12a0627d18e5ea37abd1838b7161a21fc389913f/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L543-L566

This is likely because almost all component scripts are related to elements, not to attributes. That is why it is likely being missed. 

This is also showing up incorrectly in unit tests. In the following example:

https://github.com/ampproject/amp-wp/blob/12a0627d18e5ea37abd1838b7161a21fc389913f/tests/test-tag-and-attribute-sanitizer.php#L237-L241

The `array()` should actually be `array( 'amp-lightbox-gallery' )`.

This is also a problem for `amp-fx-collection` where it has an `amp-fx` attribute which requires the component script to be included. The problem with `amp-fx` and `lightbox` attributes is that the validator spec has no mechanism to mark such attributes as `requires_extension`, as is the case for tags. See https://github.com/ampproject/amphtml/issues/19935. Because of this, we have to hare-code the dependencies, similar to how we currently include `amp-bind` when there is a binding attribute is present.

# TODO

- [x] Confirm that there is indeed no way in the validator to determine the required extension for a given attribute.
- [x] Determine if there are any other such attributes that need hard-coded consideration.
- [x] Require the `amp-fx-collection` extension when an element has the `amp-fx` attribute.

Fixes #1808.

### DIFF
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -553,6 +553,9 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 			if ( $node->hasAttribute( 'lightbox' ) ) {
 				$this->script_components[] = 'amp-lightbox-gallery';
 			}
+			if ( $node->hasAttribute( 'amp-fx' ) ) {
+				$this->script_components[] = 'amp-fx-collection';
+			}
 
 			// Check if element needs amp-bind component.
 			if ( $node instanceof DOMElement && ! in_array( 'amp-bind', $this->script_components, true ) ) {
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -549,6 +549,11 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 				$this->script_components = array_merge( $this->script_components, $tag_spec['requires_extension'] );
 			}
 
+			// Manually add components for attributes; this is hard-coded because attributes do not have requires_extension like tags do. See <https://github.com/ampproject/amp-wp/issues/1808>.
+			if ( $node->hasAttribute( 'lightbox' ) ) {
+				$this->script_components[] = 'amp-lightbox-gallery';
+			}
+
 			// Check if element needs amp-bind component.
 			if ( $node instanceof DOMElement && ! in_array( 'amp-bind', $this->script_components, true ) ) {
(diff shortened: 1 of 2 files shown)


westonruter:
> * Confirm that there is indeed no way in the validator to determine the required extension for a given attribute.

See https://github.com/ampproject/amphtml/issues/19935. There is currently no validation mechanism for an attribute to require an extension.

> * Determine if there are any other such attributes that need hard-coded consideration.

@aghassemi notes that `amp-fx-collection` is another example of attributes requiring an extension. The `amp-fx` is the attribute we need to look for.

westonruter:
Here's a build of the plugin from this branch for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/2837621/amp.zip)

westonruter:
Regarding this code:

```php
			if ( $node->hasAttribute( 'amp-fx' ) ) {
				$this->script_components[] = 'amp-fx-collection';
			}
```

This is irrelevant as of #1862.

PR #1858, opened 2019-02-06
Exclude duplicate stylesheets from amp-custom style manifest

### LINKED ISSUE #1714: Total included CSS size can be incorrect
I looked at a size and found a surprising line in the manifest for the `style[amp-custom]`:

```
Total included size: 50,059 bytes
```

This should not happen because any stylesheets over 50KB should get excluded. Nevertheless, the page passes AMP validation and when I copied the contents of the element the total number of bytes was actually 49,986 bytes (talk about cutting it close). So the plugin was correct in allowing that CSS, but it was displaying the incorrect number. 

The discrepancy could be due to PHP counting individual bytes (including multi-byte UTF-8 encoded characters), whereas the AMP validator is limiting the size to 50KB of Unicode characters (not UTF-8 encoded bytes). In any case, it seems the right count is being obtained when tallying up each stylesheet, but only the final sum at the end is incorrect. We may need to use `mb_strlen()` here.

westonruter:
It turns out that the stylesheets included in `style[amp-custom]` are de-duplicated, but the HTML comment with the manifest of all the stylesheets was not accounting for this deduplication. This fixes the problem to ensure that the “Total included size” is accurate and should not ever go above 50,000 bytes.

Fixes #1714.

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -1958,7 +1958,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			$excluded_original_size = 0;
 			$included_sources       = array();
 			foreach ( $stylesheet_sets['custom']['pending_stylesheets'] as $i => $pending_stylesheet ) {
-				if ( ! ( $pending_stylesheet['node'] instanceof DOMElement ) ) {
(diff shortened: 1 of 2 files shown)


westonruter:
Build of branch for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/2838669/amp.zip)

PR #1870, opened 2019-02-18
Update ESLint configuration

swissspidy:
This PR replaces #1766 since updating ESLint took a bit more effort.

In particular:

* Update `eslint`
* Update `eslint-plugin-react`
* Install and use `babel-eslint` so that ESLint is actually able to parse the JSX source files
* Specify React version used by WordPress so `eslint-plugin-react` doesn't throw a warning

As a bonus, I added the `lint` npm script so that one can just type `npm run lint`

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,15 +12,16 @@
   "@wordpress/dom-ready": "2.0.2",
   "@wordpress/i18n": "3.1.1",
   "babel-core": "6.26.3",
+  "babel-eslint": "10.0.1",
   "babel-loader": "7.1.5",
   "babel-plugin-transform-object-rest-spread": "6.26.0",
   "babel-plugin-transform-react-jsx": "6.24.1",
   "babel-plugin-transform-runtime": "6.23.0",
   "babel-preset-env": "1.7.0",
   "cross-env": "5.2.0",
-  "eslint": "4.19.1",
+  "eslint": "5.14.0",
   "eslint-config-wordpress": "2.0.0",
-  "eslint-plugin-react": "7.11.1",
+  "eslint-plugin-react": "7.12.4",
   "grunt": "1.0.2",
   "grunt-contrib-clean": "1.1.0",
   "grunt-contrib-copy": "1.0.0",
@@ -33,6 +34,7 @@
  "scripts": {
   "build": "grunt build && grunt create-build-zip",
   "deploy": "grunt deploy",
-  "dev": "cross-env BABEL_ENV=default webpack --watch"
+  "dev": "cross-env BABEL_ENV=default webpack --watch",
+  "lint": "eslint ."
  }
 }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -33,6 +34,7 @@
  "scripts": {
   "build": "grunt build && grunt create-build-zip",
   "deploy": "grunt deploy",
-  "dev": "cross-env BABEL_ENV=default webpack --watch"
+  "dev": "cross-env BABEL_ENV=default webpack --watch",
+  "lint": "eslint ."
```
westonruter: Should this also include a call to `phpcs`?

swissspidy: We could do `lint:js`, `lint:php`, and `lint`, which calls the former two, if that makes sense?

westonruter: Yeah, that sounds good. In another PHP project I have a `composer.json` with:

```
  "scripts": {
    "lint": [
      "@php-l",
      "phpcs"
    ],
    "php-l": "git ls-files -z -- '*.php' | xargs -0 -n 1 -P 4 php -l",
    "phpcs": "phpcs",
    "phpcbf": "phpcbf",
    "test": "WP_TESTS_DIR=/app/public/core-dev/tests/phpunit/ phpunit --config=phpunit-integration.xml.dist",
    "generate-readme": "@php ./vendor/xwp/wp-dev-lib/generate-markdown-readme"
  }
```

swissspidy: This begs the question whether we wanna put all scripts in `package.json` or `composer.json`. Or keep them separated, i.e. PHP things in `composer.json` and JS things in `package.json` 🤷‍♂️

swissspidy: Done now in the most recent commit.

package.json
```diff
@@ -12,28 +12,37 @@
...
   "grunt-wp-deploy": "1.2.1",
+  "npm-run-all": "^4.1.5",
   "timeago.js": "3.0.2",
   "webpack": "4.29.5",
   "webpack-cli": "3.2.3"
  },
  "main": "blocks/index.js",
  "scripts": {
-  "build": "grunt build && grunt create-build-zip",
+  "build": "npm-run-all build:*",
```
westonruter: Is this going to be stable? Could it not result in `zip` being run before `prepare`? Objects in JS do not have ordered keys, unless NPM explicitly sorts them by key when globbing like this. Do you have confirmation?

swissspidy: So far I have never had any problems with the order being mixed up by `npm-run-all` or something.

> Objects in JS do not have ordered keys

I thought this changed a bit with ES2015?

> Do you have confirmation?

According to the author, it executes tasks in definition order in `package.json`. https://github.com/mysticatea/npm-run-all/issues/116#issuecomment-348672125

I haven't looked at the code of NPM or the package though to see how order is determined.

Issue #1871, opened 2019-02-18
Remove changelog from readme.txt


swissspidy:
Background: I want to make it easier to translate this plugin so more people from around the world can learn about it.

Plugins hosted on WordPress.org can be fully translated by polyglots. This applies to the code _and_ the `readme.txt`.

Translated readmes are available on Rosetta pages like https://es.wordpress.org/plugins/amp/ and https://tw.wordpress.org/plugins/amp/.

The problem right now is that people have to translate the complete changelog of the plugin in order to reach 100% for the translation. Some people actually try to translate changelog entries, others simply copy and paste the English text.

In order to remove this burden from polyglots, the changelog section should just contain a link to the living changelog on GitHub, either https://github.com/ampproject/amp-wp/releases or preferably an actual `CHANGELOG.md` file.

Examples of plugins already doing this:

https://wordpress.org/plugins/preferred-languages/#developers
https://wordpress.org/plugins/query-monitor/#developers

While we're at it, I would also update `README.md` to point to `CHANGELOG.md`. Right now, it's basically a Markdown copy of `readme.txt`, which I think doesn't make much sense. I'd love to see a `README.md` file focused at developers contributing to the project on GitHub.

So, in summary:

* Remove changelog from `readme.txt` and point to GitHub instead
* Create separate changelog file on GitHub for easier referencing
* Update `readme.md` to make it more accessible to GitHub audience

Hope this makes sense. Happy to create a PR to kick this off.

westonruter:
> In order to remove this burden from polyglots, the changelog section should just contain a link to the living changelog on GitHub, either https://github.com/ampproject/amp-wp/releases or preferably an actual `CHANGELOG.md` file.

Sounds good to me. I like the idea of linking to the releases page on GitHub as that will eliminate maintenance of having to update `CHANGELOG.md` as well. We've been essentially putting the changelog in the release notes anyway.

> Right now, it's basically a Markdown copy of `readme.txt`, which I think doesn't make much sense.

That's right. The `readme.md` is automatically generated from `readme.txt` via `grunt readme`. The intention was to eliminate having to maintain two separate readmes.

> I'd love to see a `README.md` file focused at developers contributing to the project on GitHub.

Isn't this what the `contributing.md` is for?

swissspidy:
> That's right. The `readme.md` is automatically generated from `readme.txt` via `grunt readme`. The intention was to eliminate having to maintain two separate readmes.
[...]

PR #1872, opened 2019-02-18
Remove changelog from readme

swissspidy:
See #1871.


westonruter:
As part of this we'll also need to make sure there is GitHub release created with the changeling entries ported over.

swissspidy:
The contributing guidelines currently refer to the changelog in the readme, recommending to update it before release:

https://github.com/ampproject/amp-wp/blob/bc59039bce8600fd4e98933e2c696cc038d5dfa3/contributing.md#creating-a-release

I suppose the separate changelog is easier to scan. For example, the 1.0.0 release only lists changes since 1.0.0 RC4, whereas the changelog lists them all together. That could be an argument for a separate `changelog.md` file.

amedina:
I like the idea of separating the change log from the Readme. Wanted to do that prior to 1.0. Let's do it.

westonruter:
> I suppose the separate changelog is easier to scan. For example, the 1.0.0 release only lists changes since 1.0.0 RC4, whereas the changelog lists them all together. That could be an argument for a separate `changelog.md` file.

And it's not easy to scan changes in https://github.com/ampproject/amp-wp/releases because of pre-releases?

> I like the idea of separating the change log from the Readme. Wanted to do that prior to 1.0. Let's do it.

Yeah, I wasn't aware at that time of the translation problem.

swissspidy:
> And it's not easy to scan changes in [ampproject/amp-wp/releases](https://github.com/ampproject/amp-wp/releases) because of pre-releases?

IMO yes. A separate changelog file would be easier.

westonruter:
Be there would be more maintenance burden. How about starting with just keeping changeling in GitHub releases only and then we evaluate a separate file later?

swissspidy:
Maintenance would be the same as now, no? The changelog would just be in a different file. But yeah, let's evaluate that later.

westonruter:
I mean, more maintenance than if we just keep the changeling entry in the GitHub release notes alone. Also, in release notes we don't have to make commits to edit them, which is nice.

swissspidy:
@westonruter Updated `contributing.md` accordingly. What do you think?

Issue #1742, opened 2018-12-13
Add AMP compatibility for older core themes: 2010, 2011, 2012, 2013, and 2014


westonruter:
Even though these themes are no longer bundled with WordPress, they are still popular. We should add support for them in the plugin since they won't get built-in support in the near term.

- [x] Twenty Ten, Twenty Eleven, Twenty Twelve: #2407
- [x] Twenty Thirteen: #2476
- [x] Twenty Fourteen: #1873

PR #1873, opened 2019-02-18
Add core theme support for Twenty Fourteen

westonruter:
This PR adds AMP compatibility for Twenty Fourteen. There should be no AMP validation errors reported from the theme.

In testing, note that you will likely need to disable the admin bar in AMP as there will be too much CSS otherwise.

# Header Nav

Create a nested nav menu and assign to the top primary menu _and_ the secondary menu. You should see the submenus expand when hovering _and_ they should be keyboard accessible through tabbing:

image

image

And in a mobile viewport the hamburger menu should be tappable:

image

Which reveals the nested nav menus:

image

# Featured Content

## Slider

Create three posts with the `featured` tag and assign them each a featured image. Open the  “Featured Content” section in the Customizer and select the “Slider” layout. The homepage should then have a slider powered by `amp-carousel` as shown below:

image

## Grid

Featured content in desktop in grid mode should look like:

image

And in mobile:

image

Note that some of the heights will not be exactly the same between AMP and non-AMP, but it should be relatively unnoticeable (only a few pixels difference).

# Search bar

Clicking the Search button in the header should cause the search bar to appear _and_ the search input to be focused:

image

See #1742.

### DIFF
--- a/includes/sanitizers/class-amp-core-theme-sanitizer.php
+++ b/includes/sanitizers/class-amp-core-theme-sanitizer.php
@@ -1133,6 +1133,7 @@ class AMP_Core_Theme_Sanitizer extends AMP_Base_Sanitizer {
 				?>
 				<style>
 					/* Styles for featured content */
+					.grid #featured-content .post-thumbnail,
 					.slider #featured-content .post-thumbnail {
 						padding-top: 0; /* Override responsive hack which is handled by AMP layout. */
 						overflow: visible;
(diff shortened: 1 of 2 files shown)


westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3214576/amp.zip) (1.2-beta1-20190524T003045Z-57764074)

PR #1878, opened 2019-02-19
Eliminate comments walker in favor of comments sanitizer

### LINKED ISSUE #1877: Use own comment Walker
Hi guys,

you added a comment walker in class AMP_Theme_Support (wp_list_comments_args filter).

Unfortunately it's not possible to remove the function you added and use my own comment walker.

It would be great, if i could use a custom comment walker.

Could you add that feature please.

Thanks in advance!

Greetings

Timo

swissspidy:
With this change, a theme will be able to set their own custom comments walker without the AMP plugin interfering. The `amp-live-list` item attributes are also now only added if the `comments_live_list` flag for the `amp` theme support is present.

Fixes #1877


westonruter:
Instead of allowing the comment walker top be filterable, how about eliminating the need for an AMP-based comment walker altogether? Instead of [`AMP_Comment_Walker`](https://github.com/ampproject/amp-wp/blob/develop/includes/class-amp-comment-walker.php) it seems this logic could be moved into [`AMP_Comments_Sanitizer`](https://github.com/ampproject/amp-wp/blob/develop/includes/sanitizers/class-amp-comments-sanitizer.php).

swissspidy:
Hmm I see. I don't know about the history of this separate walker, but I guess it makes sense to move the logic.

westonruter:
I think we tried originally to avoid using sanitizers in favor of what WordPress provides for customizing output. However, in this case this causes problems since walkers cannot be composer together like filters can, so in this case I think a sanitizers is actually better. It will improve compatibility with existing plugins and themes that depend on custom walkers.

swissspidy:
@westonruter What about something like this then?

```diff
diff --git includes/class-amp-theme-support.php includes/class-amp-theme-support.php
index c127c180..7f85c713 100644
--- includes/class-amp-theme-support.php
+++ includes/class-amp-theme-support.php
@@ -880,7 +880,6 @@ class AMP_Theme_Support {
 		add_action( 'template_redirect', array( __CLASS__, 'start_output_buffering' ), $priority );
 
 		// Commenting hooks.
-		add_filter( 'wp_list_comments_args', array( __CLASS__, 'set_comments_walker' ), PHP_INT_MAX );
 		add_filter( 'comment_form_defaults', array( __CLASS__, 'filter_comment_form_defaults' ) );
 		add_filter( 'comment_reply_link', array( __CLASS__, 'filter_comment_reply_link' ), 10, 4 );
 		add_filter( 'cancel_comment_reply_link', array( __CLASS__, 'filter_cancel_comment_reply_link' ), 10, 3 );
@@ -981,29 +980,6 @@ class AMP_Theme_Support {
 		return $embed_handlers;
 	}
 
-	/**
-	 * Add the comments template placeholder marker
-	 *
-	 * @param array $args the args for the comments list..
-	 * @return array Args to return.
-	 */
-	public static function set_comments_walker( $args ) {
-		/**
-		 * Filters the AMP comments walker.
-		 *
-		 * @since 1.1.0
-		 *
-		 * @param string $walker Comments walker class name. Defaults to AMP_Comment_Walker.
-		 */
-		$amp_walker = apply_filters( 'amp_comment_walker', 'AMP_Comment_Walker' );
-
-		if ( class_exists( $amp_walker ) ) {
-			$args['walker'] = new $amp_walker();
-		}
-
-		return $args;
-	}
-
 	/**
 	 * Adds the form submit success and fail templates.
 	 */
diff --git includes/sanitizers/class-amp-comments-sanitizer.php includes/sanitizers/class-amp-comments-sanitizer.php
index 38725fcb..936f71ae 100644
--- includes/sanitizers/class-amp-comments-sanitizer.php
+++ includes/sanitizers/class-amp-comments-sanitizer.php
@@ -18,7 +18,6 @@ class AMP_Comments_Sanitizer extends AMP_Base_Sanitizer {
 	 * @since 0.7
 	 */
 	public function sanitize() {
-
 		foreach ( $this->dom->getElementsByTagName( 'form' ) as $comment_form ) {
 			/**
 			 * Comment form.
@@ -34,6 +33,13 @@ class AMP_Comments_Sanitizer extends AMP_Base_Sanitizer {
 				$this->process_comment_form( $comment_form );
 			}
 		}
+
+		$xpath = new DOMXPath( $this->dom );
+		$comments = $xpath->query('//*[starts-with(@id,\'comment-\')]' );
+
+		foreach( $comments as $comment ) {
+			$this->process_comment( $comment );
+		}
 	}
 
 	/**
@@ -156,4 +162,31 @@ class AMP_Comments_Sanitizer extends AMP_Base_Sanitizer {
 		);
 		$comment_form->setAttribute( 'on', implode( ';', $on ) );
 	}
+
+	/**
+	 * Comment.
+	 *
+	 * @param DOMElement $comment Comment.
+	 */
+	protected function process_comment( $comment ) {
+		$comment_id = str_replace( 'comment-', '', $comment->getAttribute( 'id' ) );
+
+		$comment_object = get_comment( $comment_id );
+
+		if ( ! $comment_object instanceof wp_comment ) {
+			return;
+		}
+
+		$comment->setAttribute( 'data-sort-time', esc_attr( strtotime( $comment_object->comment_date ) ) );
+
+		$update_time = $comment_object->comment_date;
+
+		$children = $comment_object->get_children( array( 'hierarchical' => 'flat', 'orderby' => 'comment_date' ) );
+
+		if ( ! empty( $children ) ) {
+			$update_time = $children[0]->comment_date;
+		}
+
+		$comment->setAttribute( 'data-update-time', esc_attr( strtotime( $update_time ) ) );
+	}
 }

```

westonruter:
Yes! 

FYI: Shouldn't use esc_attr() here since it is interacting directly with the DOM.

westonruter:
Build of plugin from this branch for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/2886259/amp.zip)

Issue #1880, opened 2019-02-19
Font selection


ThierryA:
- [x] Extend current font selection with Google Fonts
- [x] Add font preview in the list

swissspidy:
Retrieving a list of available fonts from Google Fonts requires an [API key](https://developers.google.com/fonts/docs/developer_api). Maybe we should just include a list of the most popular ones? The whole directory contains over 900 fonts, which is too much anyway.

ThierryA:
Good question? I suggest we grab the 15 top fonts [from this list](https://fonts.google.com/analytics), at least for the Beta version.

amedina:
Good idea ==> +1

swissspidy:
Is anyone aware of a React component (either in Gutenberg or third-party) we could use to preview fonts directly in the dropdown?

westonruter:
I think it would be preferable to use SVG images of the font names in their respective font faces. This would avoid the need to load all of the fonts when most will never be used.

swissspidy:
I don't think `<select>` supports SVG inside `<option>`, so we'd need a custom component there anyway.

Loading time is definitely a valid concern. Right now in #1885 I just enqueue all styles, which obviously isn't the best solution. Ideally, we'd store a list of all the different fonts used in the story to only enqueue the styles that are really necessary.

westonruter:
Yes, a custom select component that allows images for dropdown options.

swissspidy:
I found https://danmarshall.github.io/google-font-to-svg-path/ that could be used to generate SVGs for all the fonts.

When a font is selected for the first time, we can load the font stylesheet in the background. There might be FOUT while the font is being loaded, but given that most fonts probably won't be used anyway, this makes sense.

On the front end we can then check for all the unique `data-font-family` attributes and only enqueue the fonts necessary.

westonruter:
I think this can be closed now, yes?

csossi:
verified in QA

PR #1885, opened 2019-02-20
[AMP Stories] Font Selection

swissspidy:
This PR extends current font selection with ~20 popular fonts from Google Fonts.

To do:

* [x] Maybe don't make font names translatable
* [x] Live previews
* [x] Use SVGs for font previews
* [x] Only enqueue fonts in the editor when selected
* [x] Load all used fonts when opening editor
* [x] Only enqueue needed fonts on the front end

See #1880.


swissspidy:
> Only enqueue needed fonts on the front end

@westonruter Any tips on how to best do this? In the editor I simply go through all blocks with a font declaration and enqueue the stylesheet on the fly (0249e03c7ad8b60362f204aac42e271cb2903ba3). Not sure how to best approach on the front end though.

westonruter:
Instead of defining all of these font families in CSS, let's instead define them in PHP like this:

```php
$font_families = array(
    'ubuntu' => array(
        'styles' => array( 'font-family' => "'Ubuntu',sans-serif"' )
    ),
);
```

Then with this data structure in hand, when we encounter a block that has the `data-font-family` attribute, we can then just-in-time enqueue the stylesheet from Google Fonts and then when PHP add an inline style to `amp-stories.php`:

```css
[data-font-family="ubuntu"] { font-family: 'Ubuntu',sans-serif }
```

This is important because the tree-shaker does not currently shake selectors like `[data-font-family="ubuntu"]`.

As for dynamic enqueueing, there are two different approaches we could take:

1. Register dynamic block `render_callback` functions for every block that that uses a `data-font-family` attribute. Enqueue the Google Font stylesheet `link` when generating the block, and the sanitizers will automatically move them to the `head`.
2. Rely more fully on the sanitizers by introducing a new sanitizer that will look for any element that has a `data-font-family` attribute and then dynamically generate the `link` element for the Google Font stylesheet and add it to the `head`, if one hasn't already been added. This would also be responsible for adding `style` elements with the corresponding `[data-font-family="…"]` rules.

The second approach is probably better.

@swissspidy I can work on this if you want.

swissspidy:
Yes, that would be great! Thank you 🙂 

I was thinking of using `render_callback` for the page block already (#1899) so we don't have to worry about block deprecations. This might also be a good approach for the text block and easier to implement. However, I defer to you to make the decision.

westonruter:
> I was thinking of using `render_callback` for the page block already

I think `render_callback` makes sense for the page, but it wouldn't be so clean to add a `render_callback` for every single block in a story page that has font controls.

swissspidy:
True, although right now only the text block has font family controls (before it was paragraph + heading). But yeah, who knows how this might change in the future :-)

westonruter:
Good point! I'll play around with what works better.

westonruter:
@swissspidy Ready for your final review and merge!

PR #1886, opened 2019-02-20
AMP Stories: Webpack improvements

swissspidy:
I was a bit frustrated that in the code the `wp.*` globals are used directly. This way you don't get any hints about existing components by your IDE like when using the `@wordpress/` packages directly.

Changing that is rather trivial by modifying the `externals` configuration in the Webpack configuration. This way, Webpack can transform `import … from '@wordpress/components'` into `const { } = wp.components`.

This PR contains the necessary changes to support this, as well as changes to most\* of the JS source files to leverage this new feature. 

I also noticed that `@wordpress/babel-plugin-import-jsx-pragma` wasn't being used, which is the recommended way to handle the automatic imports for the WP packages.

While I was at it, I updated the Babel packages as well.

----

\* I skipped 'ampStoryEditorBlocks' somehow because that file is quite huge and probably should be overhauled a bit anyway.

**Note:** Looks like this depends on #1870. (Edit: Manually merged the branch)


westonruter:
Nice. Not sure how you learned all that, but good on ya 😄

package.json
```diff
@@ -1,40 +1,61 @@
...
+    "deploy": "grunt deploy",
+    "dev": "cross-env BABEL_ENV=default webpack --watch",
+    "lint": "npm-run-all --parallel lint:*",
+    "lint:js": "eslint .",
+    "lint:php": "vendor/bin/phpcs",
+    "test": "npm-run-all --parallel test:*",
+    "test:php": "phpunit"
+  },
+  "dependencies": {
+    "@wordpress/blocks": "6.0.7",
```
westonruter: @swissspidy Question: Is this dependency not a throw-away? If Webpack is configured to treat `@wordpress/blocks` among the `externals`, then is there any purpose in making it a dependency other than for IDEs? Why shouldn't these then be `devDependencies`?

swissspidy: Semantically they‘re still dependencies, but I get your point. I would need to dig a bit into this to see if there are any big pros or cons to it. Otherwise I don‘t have a string opinion on this.

PR #1891, opened 2019-02-20
Update amphtml from v1902072121410 to v1902191749350

westonruter:
See release notes: 

* https://github.com/ampproject/amphtml/releases/tag/1902081532110
* https://github.com/ampproject/amphtml/releases/tag/1902151859190
* https://github.com/ampproject/amphtml/releases/tag/1902191749350

Previous update: https://github.com/ampproject/amp-wp/pull/1862

Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/2886346/amp.zip)

* Add `amp-story-page-attachment` element to AMP Stories.
* Add `custom-redirect-domain` attribute to `amp-skimlinks`.
* Add `autoscroll` attribute for `amp-sidebar` component.
* Add `dock` attribute for `amp-ima-video`.
* Add recognition of `textarea[autoexpand]`.
* Re-work `bin/amphtml-update.sh` to eliminate use of Composer for amphtml dependency, and automating process for updating to latest stable version of amphtml.


contributing.md
```diff
@@ -90,10 +90,18 @@ This will create an `amp.zip` in the plugin directory which you can install. The
 ## Updating Allowed Tags And Attributes
 
 The file `class-amp-allowed-tags-generated.php` has the AMP specification's allowed tags and attributes. It's used in sanitization.
-To update that file:
+
+To update that file, perform the following steps (which need to be improved):
```
westonruter: We could automate the fetching of the latest version by getting `ampRuntimeVersion` from https://cdn.ampproject.org/rtv/metadata while chopping off the `01` release prefix.

But still, the method of including `ampproject/amphtml` repo as a Composer dependency seems hacky here. There are 4 tag references that have to be maintained. @swissspidy any thoughts on a better way to include the dependency?

swissspidy: Is there any benefit of having that repo as a Composer dependency, except for convenience and perhaps visibility of what version we're using?

Hmm on first thought, we could do this:

1. Remove dependency from `composer.json`
2. Extend `bin/amphtml-update.sh`:  
   1. Fetch version number from https://cdn.ampproject.org/rtv/metadata
   2. Download the respective ZIP file from GitHub to a temporary location
   3. Run python script
   4. Delete temporary folder again
3. Bonus: Add `AMP__AMPHTML_VERSION` constant to `amp.php` and update that with the new version number each time.
3. Stretch goal: run that script on Travis CI and fail builds (or throw warnings or something) if the current spec isn't up to date.
4. Stretch goal: have a bot run that script and open pull requests

westonruter: I suppose the main benefit of having the Composer dependency is that the plugin gets explicitly tied to a version of AMP. In the past it has been annoying to keep track of the current version of AMP that is applied, but I suppose this would be minimized by a `AMP__AMPHTML_VERSION` constant…

> 3\. Bonus: Add `AMP__AMPHTML_VERSION` constant to `amp.php` and update that with the new version number each time.

There is some of this currently in:

https://github.com/ampproject/amp-wp/blob/c9e21c766a2bf84ac48c88dbe03391abf7b62b59/includes/sanitizers/class-amp-allowed-tags-generated.php#L16-L17

But it doesn't indicate the actual AMP version. Also, these private static variables aren't even used anywhere. So the version could be added to that class instead as a const.

westonruter: I'll implement those changes.

swissspidy: Awesome! Let me know if I can help with anything.

westonruter: OK, I've now implemented this. However, I opted to not do the `AMP__AMPHTML_VERSION` constant at this time because it would not enable us to try generating the spec from a non-tagged version. 

As it stands right now, you can have a local clone of amphtml and generate the spec to whatever commit or branch you have checked out there by doing:

```basu
./bin/amphtml-update.sh amphtml/
```

Where `amphtml` is the Git repo. If you omit that, however, then it gets the current stable release.

We can revisit how we incorporate the amphtml version into the plugin later. Right now it's not strictly needed.

westonruter: Note that automating the update via Travis would not probably be worth it because the spec changes from time to time and changes are needed for the Python script to correctly generate from the spec. Also, it's important to also add unit tests to make sure that new tags/attributes are handled properly. So I think it should be a manual process when new versions come out for the forseeable future.

.gitignore
```diff
@@ -1,8 +1,10 @@
...
-build
+/vendor
+/build
 node_modules
 wiki
 amp.zip
 **/*-compiled.js
 built
+/amphtml
+.vscode
```
swissspidy: This belongs to one's personal global `.gitignore` file. Same with `.DS_Store`, actually.

bin/amphtml-update.sh
```diff
@@ -1,12 +1,21 @@
...
+set -e
 
-BIN_PATH="$(pwd)"
-PROJECT_PATH=$(dirname $PWD)
-VENDOR_PATH="$PROJECT_PATH/vendor"
+BIN_PATH="$(dirname "$0")"
+PROJECT_PATH=$(dirname $BIN_PATH)
+AMPHTML_LOCATION="$1"
 
 if ! command -v apt-get >/dev/null 2>&1; then
```
swissspidy: While we're at it, can we move this check to within the Python checks? As far as I can see, one doesn't need `apt-get` when Python is already available.

bin/amphtml-update.sh
```diff
@@ -1,12 +1,21 @@
...
-# Go to the right location.
-cd "$(dirname "$0")"
+set -e
 
-BIN_PATH="$(pwd)"
-PROJECT_PATH=$(dirname $PWD)
-VENDOR_PATH="$PROJECT_PATH/vendor"
+BIN_PATH="$(dirname "$0")"
+PROJECT_PATH=$(dirname $BIN_PATH)
+AMPHTML_LOCATION="$1"
```
swissspidy: Can we use `/amphtml` as the default here in case no argument is provided?

.gitignore
westonruter: We don't want these every to be committed, so I'd rather include them here than rely on a user to populate their global `.gitignore` (which I doubt few do).

bin/amphtml-update.sh
westonruter: When no argument is provided, then the plugin just downloads the tarball into a temp directory and then deletes the temp directory when done. We don't want a persistent location in the normal case, as then we'd not be getting the latest if `amphtml` already exists.

bin/amphtml-update.sh
westonruter: I don't think so. It's not whether `python` isn't available, but rather whether Python _and_ the protobuf packages are available.

swissspidy: Rephrase: let‘s move these checks to after the python and protobuf checks then.

In other words: this would prevent me from running the script on my Mac.

westonruter: I'm not sure I understand. The script is supposed to prevent you from running on your Mac since apt is required.

.gitignore
swissspidy: How is that change related to this PR then?

It‘s a user-specific folder that the user is responsible for excluding. If they don’t, we would catch it in the PR anway. Otherwise we‘d have to add entries for all the dozens or so popular editors out there, which is not sustainable. Or is the VSC folder in particular accidentally part of lots of PRs so far?

westonruter: It's related to this PR because I added `/amphtml` to the gitignore, and I noticed that `.vscode` was also among the uncommitted files. I'll just remove it.

bin/amphtml-update.sh
swissspidy: Looking at the code, I see that python is required, not apt specifically. Is that incorrect?

westonruter: These packages are required: `python`, `protobuf-compiler`, `python-protobuf`.

We could check to see if those Python modules are installed directly instead instead of doing so via apt, but we can do that in another PR.

westonruter: Done in https://github.com/ampproject/amp-wp/pull/1891/commits/6ae9abb7cd6948a85dc896cb35207cc457c5516f

Issue #1883, opened 2019-02-19
Text block


ThierryA:
- [x] Add text block with variations for auto, h1, h2, and paragraph; for auto variation, automatically switch the element based on an auto-smartness estimation algorithm (e.g. font-size heuristic)

ThierryA:
_Partly covers #1881_

csossi:
Verified in QA

PR #1893, opened 2019-02-21
[AMP Stories] Text Block

swissspidy:
**Work in progress**

Smart text block that replaces/combines the paragraph and heading blocks.

See #1883.


blocks/amp-story/text/edit.js
```diff
@@ -0,0 +1,84 @@
...
+import getTagName from './getTagName';
+
+const applyFallbackStyles = withFallbackStyles( ( node, ownProps ) => {
+	const { fontSize, customFontSize } = ownProps.attributes;
+	const editableNode = node.querySelector( '[contenteditable="true"]' );
+	//verify if editableNode is available, before using getComputedStyle.
+	const computedStyles = editableNode ? getComputedStyle( editableNode ) : null;
+
+	return {
+		fallbackFontSize: fontSize || customFontSize || ! computedStyles ? undefined : parseInt( computedStyles.fontSize ) || undefined
```
westonruter: Do we need to resort to getting the computed styles? I'm concerned about the performance impact.

blocks/amp-story/text/edit.js
```diff
@@ -0,0 +1,84 @@
...
+const { getComputedStyle } = window;
+
+/**
+ * Internal dependencies
+ */
+import getTagName from './getTagName';
+
+const applyFallbackStyles = withFallbackStyles( ( node, ownProps ) => {
+	const { fontSize, customFontSize } = ownProps.attributes;
+	const editableNode = node.querySelector( '[contenteditable="true"]' );
```
westonruter: Querying for `[contenteditable="true"]` seems a bit brittle. Is the node not accessible in another way?

blocks/amp-story/text/edit.js
swissspidy: FWIW I adapted this from the `core/paragraph` block, but haven't tested it properly yet.

`getComputedStyle` is meant as the fallback of the fallback.

Since the paragraph block is quite special in Gutenberg, `getComputedStyle` and the `editableNode` check might not even apply to our custom block here.

assets/src/blocks/amp-story-text/index.js
```diff
@@ -0,0 +1,85 @@
...
+			type: 'string',
+			default: 'auto',
+		},
+		fontSize: {
+			type: 'string',
+		},
+		customFontSize: {
+			type: 'number',
+		},
+		ampFontFamily: {
```
westonruter: Should the `amp` prefix be retained?

assets/src/blocks/amp-story-text/index.js
```diff
@@ -0,0 +1,85 @@
...
+
+		const styles = {
+			fontSize: fontSizeClass ? undefined : customFontSize,
+		};
+
+		return (
+			<RichText.Content
+				tagName={ tagName }
+				style={ styles }
+				className={ className ? className : undefined }
```
westonruter: Why not just:

```suggestion
				className={ className }
```

assets/src/blocks/amp-story-text/index.js
```diff
@@ -0,0 +1,85 @@
...
+	},
+
+	edit,
+
+	save( { attributes } ) {
+		const { content, fontSize, customFontSize } = attributes;
+
+		const tagName = getTagName( attributes );
+		const fontSizeClass = fontSize && `is-${ fontSize }-text`;
+		const className = fontSizeClass ? fontSizeClass : null;
```
westonruter:
```suggestion
		const className = fontSizeClass ? fontSizeClass : undefined;
```

swissspidy: That contradicts your classMame suggestion from below, no?

assets/src/blocks/amp-story-text/index.js
swissspidy: Probably, for consistency reasons. Plus this way it doesn‘t clash with a possible font family attribute in Gutenberg in the future.

assets/src/blocks/amp-story-text/index.js
westonruter: No, it's actually part of the same suggestion. If `className` is `undefined` then you could just do `className={ className }`

swissspidy:
^ eb61fc6e957c08bd36fcd93857566cafe8e8f6af contains my personal email address as a co-author, hence the googlebot warning. Should be all good now.

assets/src/blocks/amp-story-text/edit.js
```diff
@@ -0,0 +1,78 @@
...
+				className={ `${ className } ${ fontSizeClass }` }
+				placeholder={ placeholder || __( 'Write text…', 'amp' ) }
+			/>
+		</Fragment>
+	);
+}
+
+const TextEdit = compose( [
+	withFontSizes( 'fontSize' ),
+] )( TextBlock );
```
miina: Maybe we could use `withFontSizes` here without `compose`? Since it's just one HOC.

assets/src/blocks/amp-story-text/edit.js
```diff
@@ -0,0 +1,78 @@
...
+					/>
+					<SelectControl
+						label={ __( 'Select text type', 'amp' ) }
+						value={ type }
+						onChange={ ( selected ) => setAttributes( { type: selected } ) }
+						options={ [
+							{ value: 'auto', label: __( 'Automatic', 'amp' ) },
+							{ value: 'p', label: __( 'Paragraph', 'amp' ) },
+							{ value: 'h1', label: __( 'Heading 1', 'amp' ) },
+							{ value: 'h2', label: __( 'Heading 2', 'amp' ) },
```
miina: From UX perspective it could make sense to have the options in the block's toolbar as well, like in case of the heading block for example. Maybe not that important within this PR though. Thoughts?
cc @westonruter

westonruter: Good question. I suppose it depends on how frequently we expect users to be changing the setting manually. Fine to follow-up in another PR.

miina:
^ One more finding in addition to these: choosing the font family doesn't seem to be reflected in the editor at this moment.

assets/src/blocks/amp-story-text/edit.js
swissspidy: Yup. It's a leftover from when I was using two HOC's. Will address tomorrow, otherwise feel free to commit to the branch directly.

swissspidy:
> When adding an animation while having more than one Text block, sometimes the following error occurred
> Adding animation doesn't seem to get saved

Noticed that too today while working on animations. Probably a leftover from when I introduced some foundation for animation ordering in #1896. Already being addressed in #1904.

> At this moment when clicking "Return" in the Text block adds a break (`br`) instead of creating a new block -- is this intended?

I've intentionally left out key handling like this at the moment. With #1897 in mind it felt a bit odd. I think we can add it in a separate PR later if needed.

> Choosing the font family doesn't seem to be reflected in the editor at this moment.

Hmm would have to test this. At first glance I can't see why this particular PR would have broken this. But font selection is undergoing quite some changes in #1885 anyway.

miina:
> Hmm would have to test this. At first glance I can't see why this particular PR would have broken this. But font selection is undergoing quite some changes in #1885 anyway.

Yes, you're right about not being related to this PR, it doesn't seem to be working in the general AMP Stories branch either.

miina:
@swissspidy FYI: I'll fix the font-family issue in a separate PR, it's related to the `wrapperProps` not being assigned as expected and causes some other issues, too.

Issue #1879, opened 2019-02-19
Setting background image/video


ThierryA:
- [x] Add field to sidebar for the page block to allow user to select an image or video for the background (beware YouTube)
- [x] Add focal point picker for background image
~~- [ ] Auto-cropping of excessive pixels~~ moved [here](https://github.com/ampproject/amp-wp/projects/10#card-18223712)

westonruter:
@swissspidy This will depend on reworking how layers are done. So hold off on this one until I get to that.

swissspidy:
> Auto-cropping of excessive pixels

Need some UX feedback for that I think.

Using the focal point picker component would basically mean re-cropping the image on every save when the focal point changes. Or we add a "Crop" button underneath the component so users can do the cropping manually.

westonruter:
What about doing the crop at the moment the post is published or updated?

ThierryA:
> What about doing the crop at the moment the post is published or updated?
+1 and as @swissspidy just brought up during our sync (at the office), probably on preview too since it is using image positioning the admin but AMP image on preview.

csossi:
verified in QA

PR #1896, opened 2019-02-22
[AMP Stories] Remove grid layer blocks in favor of flat structure

westonruter:
This PR removes the grid layer blocks in favor of exploring a flatter block structure in which layering will eventually be done via absolute positioning, as well as via setting fill image/video via the inspector for the page as a whole.

Todo:

- [x] Remove majority of now-unused CSS.
- [x] Determine why slash command insertion does not work.
- [x] Add page-level inserter, or make it so that the global inserter properly shows all blocks when selecting a page.
- [x] Prevent insertion of non-page blocks at root.
- [x] Remove duplication in allowed blocks list.
- [ ] ~Figure out why there is block validation failed for `core/paragraph` regarding font size.~
- [x] Restore animation attributes for elements (perhaps `getAmpStoryAnimationControls`).
- [x] Make sure page block has context menu present.
- [ ] Add a `InnerBlocks` area for CTA layer. (Probably in separate PR.)
- [ ] Add dynamic `fill` layer based on page props for image/video. (Probably in separate PR.)

See #1897.


westonruter:
@swissspidy Feel free to tackle any of the todos and push to this branch when you come online.

ThierryA:
_This PR will cover the "Removing Layer" task in #1897_

swissspidy:
> Figure out why there is block validation failed for `core/paragraph` regarding font size.

Since `large` is set as the default font size, I guess Gutenberg expects there to still be a `is-large-text` class when resetting the font style selection.

I wouldn't worry too much about it at the moment since we'll add a custom text block in #1893.

assets/src/amp-story-editor-blocks.js
```diff
@@ -0,0 +1,237 @@
...
+	 * Change the default block type to amp/amp-story-page.
+	 *
+	 * This way no paragraph blocks gets inserted when pressing enter in the <PostTitle> component.
+	 *
+	 * Ideally we could override the post title so that pressing enter results in a new block inserted
+	 * inside the first page block.
+	 *
+	 * Todo: set default block name to text block when already inside
+	 */
+	setDefaultBlockName( 'amp/amp-story-page' );
```
westonruter: I just discovered that this breaks slash command insertion.

assets/src/amp-story-editor-blocks.js
```diff
@@ -0,0 +1,222 @@
...
+domReady( () => {
+	/**
+	 * Change the default block type to amp/amp-story-page.
+	 *
+	 * This way no paragraph blocks gets inserted when pressing enter in the <PostTitle> component.
+	 *
+	 * Ideally we could override the post title so that pressing enter results in a new block inserted
+	 * inside the first page block.
+	 *
+	 * Todo: set default block name to text block when already inside
```
westonruter: This is causing a problem where a page can be inserted inside of a page:

image

image

image

swissspidy: Sounds like a pressing todo then :-)

I think we can improve this by subscribing to `getSelectedBlockClientId` (using `subscribe` from `@wordpress/data`):

Whenever that value is `null`, we set the default to `amp/amp-story-page`. Otherwise we change it to a text block.

swissspidy:
190fad4d75ccb4cf28be2665e6058b014e65f756 is interesting. Curious to see where the conflicts happened exactly.

That being said, `domReady` might not be necessary anymore because of 3728f51670c82c09c61e78a072cd23a5ca1518b8

assets/src/amp-story-editor-blocks.js
swissspidy: > Add page-level inserter, or make it so that the global inserter properly shows all blocks when selecting a page.

Maybe we need `subscribe` for that too... Will have to check next week

assets/src/amp-story-editor-blocks.js
westonruter: It breaks slash commands only when `'amp/amp-story-page'`. If I change to `'core/paragraph'` then it seems all good.

assets/src/amp-story-editor-blocks.js
westonruter: That seems to do the trick! See 8cc17ea

westonruter:
There are a couple React warnings that I'm seeing:

> Warning: Unsafe lifecycle methods were found within a strict-mode tree:
>    in StrictMode (created by t)
>    in t (created by r)
>    in r (created by Context.Consumer)
>    in WithSelect(t)
>
>componentWillUpdate: Please update the following components to use componentDidUpdate instead: t

This happens even in a regular post, not a story. And then:

> Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of t which is inside StrictMode. Instead, add a ref directly to the element you want to reference.

This happens as soon as I do a slash command in a paragraph in a story.

I'm not sure what these mean.

swissspidy:
I'm gonna merge this into the branch so that I can work on the remaining tasks in separate PRs. This way I can also rebase existing ones.

> Figure out why there is block validation failed for `core/paragraph` regarding font size.

As mentioned I will look into this as part of #1893

> Add dynamic `fill` layer based on page props for image/video. (Probably in separate PR.)

Will look into this when working on #1879

PR #1899, opened 2019-02-25
[AMP Stories] Background media for story pages

swissspidy:
To do:

* [x] Add possibility to select custom poster image for videos.  
  WordPress doesn't really create poster images for videos automatically.
  For reference, in the Gutenberg video block they added a separate `<MediaUpload>` component just for this.  
  
  Thinking further, we might want to auto-generate a poster image, but this might turn out to be a bit more complex. Plus, the poster image is not required anyway.
* [x] Use poster image for the preview in the editor instead of an actual running `<video>`
* [ ] ~Improve/fix editor styling~
  **Edit:** Editor layout will be changed drastically anyway, so no need to worry about the styling of the page block for this.
* [ ] ~Actually crop background images appropriately.~
  **Edit:** As per discussion on #1879, cropping will be tackled in a separate issue.
* [ ] ~Consider using server-side rendering instead of `save()`ing the page block's HTML.~
  This way we avoid having to deal with deprecations every time the markup changes.  
  Needs some testing to see if this works when using `InnerBlocks`.
  
  **Edit:** Something that can be discussed in a separate issue.

See #1879.

### DIFF
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -16,6 +16,7 @@
 		6.1 Ghosted page inserter
 	7. Block navigator
 	8. Block mover
+	9. General block editor components
 	100. Shame
 		100.1 Gutenberg - Warning div not clickable
 
@@ -436,6 +437,18 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
 
 .post-type-amp_story .is-selected .editor-block-mover.is-visible {
 	opacity: 1;
+}
+
+/*
+ * 9. General block editor components
+ */
+
+.editor-amp-story-page-background {
+	display: block;
+}
+
+.editor-amp-story-page-background + .is-destructive {
+	margin-top: 1em;
 }
 
 /*
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -451,6 +451,20 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
 	margin-top: 1em;
 }
 
+.editor-amp-story-page-wrap {
+	position: relative;
+}
+
+.editor-amp-story-page-video {
+	position: absolute;
+	top: 0;
+	left: 0;
+	right: 0;
+	bottom: 0;
+	min-width: 100%;
+	min-height: 100%;
+}
+
 /*
  * 100. Shame
  */
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -453,6 +453,7 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
 
 .editor-amp-story-page-wrap {
 	position: relative;
+	overflow: hidden;
 }
 
 .editor-amp-story-page-video {
@@ -463,6 +464,8 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
 	bottom: 0;
 	min-width: 100%;
 	min-height: 100%;
+	object-fit: cover;
+	pointer-events: none;
 }
 
 /*
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -468,6 +468,10 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
 	pointer-events: none;
 }
 
+#editor-amp-story-page-poster__help {
+	margin-top: 0;
+}
+
 /*
  * 100. Shame
  */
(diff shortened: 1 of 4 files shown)


swissspidy:
**Initial state:**

screenshot 2019-02-28 at 11 27 43

**Image added:**

screenshot 2019-02-28 at 11 27 57

**Video added:**

screenshot 2019-02-28 at 11 28 23

Note: video currently does not autoplay or fill the whole container. Not sure how easy it is to do that with the `<video>` element.

**Video added, with poster image:**

screenshot 2019-02-28 at 11 28 37

assets/css/amp-editor-story-blocks.css
```diff
@@ -438,6 +439,32 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
...
+.editor-amp-story-page-video {
+	position: absolute;
+	top: 0;
+	left: 0;
+	right: 0;
+	bottom: 0;
+	min-width: 100%;
+	min-height: 100%;
+}
+
```
miina: It should be possible to make the video fill the content, maybe with assigning `object-fit:cover` to the `video` element and then `overflow:hidden` to the relevant parent. I think something similar was previously done (not by me) for the `AMP Background Video Layer` so it showed nicely.

assets/src/blocks/amp-story/constants.js
```diff
@@ -0,0 +1,5 @@
+
+export const IMAGE_BACKGROUND_TYPE = 'image';
+export const VIDEO_BACKGROUND_TYPE = 'video';
+export const ALLOWED_MEDIA_TYPES = [ 'image', 'video' ];
+export const POSTER_ALLOWED_MEDIA_TYPES = [ 'image' ];
```
miina: For some reason these don't seem to work for me, for example I'm still able to upload `.pdf` files, etc. Do they work as expected for you?

assets/src/blocks/amp-story/edit.js
```diff
@@ -74,33 +78,175 @@ export default class EditPage extends Component {
...
+												{ mediaUrl ? __( 'Edit Media', 'amp' ) : __( 'Upload Media', 'amp' ) }
+											</Button>
+										) }
+									/>
+								</MediaUploadCheck>
+								{ !! mediaId &&
+								<MediaUploadCheck>
+									<Button onClick={ () => setAttributes( { mediaUrl: undefined, mediaId: undefined, mediaType: undefined } ) } isLink isDestructive>
+										{ VIDEO_BACKGROUND_TYPE === mediaType ? __( 'Remove video', 'amp' ) : __( 'Remove image', 'amp' ) }
+									</Button>
```
miina: Maybe it would be good to add a suggested image / video ratio based on the [best practices](https://www.ampproject.org/docs/fundamentals/amp_story_best_practices#specify-a-poster-attribute) so that the user would have an idea which ratio to aim for the best experience. Thoughts?

assets/src/blocks/amp-story/edit.js
```diff
@@ -74,33 +78,175 @@ export default class EditPage extends Component {
...
+								<MediaUploadCheck>
+									<BaseControl
+										label={ __( 'Poster Image', 'amp' ) }
+									>
+										<MediaUpload
+											title={ __( 'Select Poster Image', 'amp' ) }
+											onSelect={ ( image ) => setAttributes( { poster: image.url } ) }
+											allowedTypes={ POSTER_ALLOWED_MEDIA_TYPES }
+											render={ ( { open } ) => (
+												<Button
```
miina: Could we make this required? Or at least let the user know that this is mandatory -- the video wouldn't show up otherwise on the AMP Story (would get sanitized by the plugin's logic). Thoughts?

assets/src/blocks/amp-story/constants.js
swissspidy: Looking at the `MediaUpload` component's implementation, you are able to upload anything, but not select non-allowed media types. Also, the `allowedTypes` property filters existing items in the media modal, so that e.g. PDFs are never shown to you in the modal.

Not sure if we can easily prevent upload of non-allowed media types. Don't feel like it's really needed though.

assets/src/blocks/amp-story/edit.js
swissspidy: Added a help text:

screenshot 2019-02-28 at 17 10 17

assets/src/blocks/amp-story/edit.js
swissspidy: Changed the label:

screenshot 2019-02-28 at 17 10 17

assets/src/blocks/amp-story/edit.js
```diff
@@ -74,33 +78,177 @@ export default class EditPage extends Component {
...
+													isDefault
+													onClick={ open }
+													className={ 'editor-amp-story-page-background' }
```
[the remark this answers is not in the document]
swissspidy: Probably just an IDE thing I need to fix :-)

PR #1901, opened 2019-02-26
Add tree shaking for :lang() selectors

### LINKED ISSUE #1900: Twenty Nineteen v1.3 has excessive CSS for AMP
Originally reported at https://wordpress.org/support/topic/twentynineteen-amp-css-broken-after-update/

In WordPress 5.1 with Twenty Nineteen 1.3, if I try loading the Gutenberg Demo post in AMP I get an `excessive_css` validation error and the theme's `style.css` is excluded. I'm also seeing this on a post published with no content at all.

The `style[amp-custom]` manifest comment shows the following:

```
     0 B: style[amp-custom=]
   417 B (95%): style#twentynineteen-style-inline-css[id=twentynineteen-style-inline-css][type=text/css]
  2372 B (97%): link#twentynineteen-print-style-css[rel=stylesheet][id=twentynineteen-print-style-css][href=https://wordpressdev.lndo.site/core-dev/src/wp-content/themes/twentynineteen/print.css?ver=1.3][type=text/css][media=print]
   122 B: style[type=text/css]
Total included size: 2,911 bytes (97% of 2,983 total after tree shaking)

The following stylesheets are too large to be included in style[amp-custom]:
 58833 B (74%): link#twentynineteen-style-css[rel=stylesheet][id=twentynineteen-style-css][href=https://wordpressdev.lndo.site/core-dev/src/wp-content/themes/twentynineteen/style.css?ver=1.3][type=text/css][media=all]
Total excluded size: 58,833 bytes (74% of 78,780 total after tree shaking)

Total combined size: 61,744 bytes (75% of 81,763 total after tree shaking)
```

Even after tree shaking, Twenty Nineteen's `style.css` is almost 60KB. The cause of the problem is in [core#45731](https://core.trac.wordpress.org/ticket/45731): Twenty Nineteen: Implement a less aggressive approach for alternate approach for non-latin font fallbacks. In particular, see these two changesets:

* https://core.trac.wordpress.org/changeset/44650
* https://core.trac.wordpress.org/changeset/44657

These changes appear to account for most of the 1,923 line insertions between 5.0.x and 5.1.

Fortunately, I believe it will be straightforward to improve the tree shaker to account the new changes since they all involve the `:lang()` selector. The AMP plugin is already accounting for this but only on the `html` element (for Twenty Seventeen):

https://github.com/ampproject/amp-wp/blob/2aa71a4fd4b42ff6487bfdbec7489b8820b7abf8/includes/sanitizers/class-amp-style-sanitizer.php#L2205-L2230

This was done in #1221. So what we need to do is extend the tree shaker to look for `:lang()` anywhere in the selector, not just after the first element.

westonruter:
Twenty Nineteen's `style.css` went from 112KB in v1.2 to an eye-dropping 208K in v1.3: almost doubling the size. When AMP has a budget of 50KB this is especially bad. As noted in #1900, the increase in size is almost all due to `:lang()` selectors being added for the various languages. This being the case, if we just tree-shake any such selectors that for all languages other than the current one, then we can recover from this large increase.

Before this PR, the tree-shaker and minifier was brought the `style.css` down to 72KB. After this PR, the total is brought down to 37KB.

Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/2903178/amp.zip) (v1.1-alpha-20190226T001832Z-bc77bcb5).

Fixes #1900.

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -922,7 +922,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	private function process_stylesheet( $stylesheet, $options = array() ) {
 		$parsed      = null;
 		$cache_key   = null;
-		$cache_group = 'amp-parsed-stylesheet-v13'; // This should be bumped whenever the PHP-CSS-Parser is updated.
+		$cache_group = 'amp-parsed-stylesheet-v14'; // This should be bumped whenever the PHP-CSS-Parser is updated.
 
 		$cache_impacting_options = array_merge(
 			wp_array_slice_assoc(
@@ -2202,7 +2202,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	private function ampify_ruleset_selectors( $ruleset ) {
 		$selectors = array();
 		$changes   = 0;
-		$language  = get_bloginfo( 'language' );
+		$language  = strtolower( get_bloginfo( 'language' ) );
 		foreach ( $ruleset->getSelectors() as $old_selector ) {
 			$selector = $old_selector->getSelector();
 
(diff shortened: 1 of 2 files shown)


westonruter:
I just realized that PR will not always have the desired effect for multi-lingual blogs. Consider, for example, if I have a site in US English but I have a post that is written in Arabic:

```html
<html lang="en-US">
...
<style>
*:lang(en) { color: red; }
*:lang(ar) { color: blue; }
</style>
...
<p>Hello, World!</p>
<p>And in Arabic, <span lang="ar">مرحبا بالعالم</span>
...
</html>
```

This PR would currently result in `*:lang(ar) { color: blue; }` being erroneously tree-shaken.

Maybe this just has to be a known issue we'll have to live with for now. I would expect it to be pretty uncommon for someone to go to the trouble of marking-up their multi-lingual content with `lang` attributes, though I'm sure there is a dedicated userbase who does. For now it seems best to account for the vast majority of cases and not have fundamentally broken styles due to the stylesheet being excluded.

swissspidy:
> I would expect it to be pretty uncommon for someone to go to the trouble of marking-up their multi-lingual content with `lang` attributes

It is quite common for multilingual content, e.g. when showing the available translations for a given post with some excerpts. However, it's mostly important for search engines and screen readers, not really for styling. Twenty Nineteen is quite an odd exception here I think.

Issue #1882, opened 2019-02-19
Set before/after animation sequence


ThierryA:
- [x] Add field for determining whether an element should be animated after another element finishes animating.

swissspidy:
Based on our discussions, in a first round this should be a simple dropdown where one could say "Start after XY finishes"

In order to do this, we need a list of all elements on the current story page with an animation.

Some `select( 'core/editor' )` functions that might be helpful: `getBlocks`, `getSelectedBlockClientId`, `getBlockHierarchyRootClientId`, and `getBlock`.

However, we probably need to create our own separate store for all animated elements. This would be easier to implement and future-proof (think full-fledge animation control in a timeline).

Speaking of that, I wonder what we could actually show in that dropdown. For each block, we have its type and its attributes/content, so maybe the dropdown could look like this:

```
Start animation after the following element:

* Immediately
* "Lorem ipsum dolor" (Text block)
* "Blue Montains" (Image)
* "Buy now" (CTA)
```

This will depend a bit on our reworking of layers, so maybe we need to hold off on this one for a bit at first.

@westonruter Thoughts?

westonruter:
Coming up with names for the elements in the drop-down could be a challenge. If you have two images in a page, how would we differentiate one from the other in the list? Rely on attachment title? What about external embeds like YouTube? The Block Navigation component does not provide helpful contextual snippets. Seems like something that should be part of core, no?

swissspidy:
Yep, it's definitely something we need to figure out along the way.

There's an issue for the Block Navigation component about this, but without much traction: https://github.com/WordPress/gutenberg/issues/11010. The following project is mentioned there as well, which shows excerpts of headings: https://github.com/sortabrilliant/guidepost

Compared to regular posts, AMP stories allow only a limited set of blocks, which might make it a bit easier for us to implement here. After that, we can refine things and then try to propose something for inclusion in Gutenberg.

swissspidy:
As suggested by @ThierryA, I had a look at how Apple's Keynote does handle animations.

At first, the UI only exposes a simple numbered dropdown, as you can see in the sidebar on this screen here:

![screenshot 2019-02-21 at 14 34 09](https://user-images.githubusercontent.com/841956/53172563-ee1a6180-35e5-11e9-989b-215d94d6a83c.png)

It works very simple: If you select **1** for an element, the current number 1 will be placed second.

For more complex arrangements, one can click on **Build Order** to see something more like we have discussed previously:

[...]

PR #1904, opened 2019-02-26
[AMP Stories] Animation Order

swissspidy:
To do:

* [x] Maybe use same `<Dropdown>` component like we already do in #1885 so we can show each block with its block type's icon.
* [x] Make sure the dropdown updates when `animatedBlocks` update.  
  Right now, content changes aren't reflected in the drop downs.
* [x] Prevent cycles in the animation order.  
  `onChange`, go through all animated blocks to ensure there are no cycles. \*  

\*  Similar checks might be needed when removing blocks. This begs the question whether it makes sense to store animation order in list instead of just having each animated block reference its predecessor. 

See #1882


swissspidy:
> Prevent cycles in the animation order.

I'm gonna implement a data store as provided by `@wordpress/data` to store animation order at a central location. From there I can subscribe to state changes (animation order changes, removed blocks, etc.) and update blocks accordingly.

This is gonna be needed for future UI iterations anyway, see https://github.com/ampproject/amp-wp/issues/1882#issuecomment-466004186 for example, where animation order is displayed in a central place.

Upon editor initialization, I'd go through all existing blocks to determine the order and set the default state.

assets/src/components/animation-controls.js
```diff
@@ -28,12 +35,16 @@ function AnimationControls( { setAttributes, attributes, animatedBlocks } ) {
 		},
 	];
 
-	animatedBlocks.map( ( block ) => {
+	getAnimatedBlocks().map( ( block ) => {
```
swissspidy: Right now, content changes aren't reflected in the dropdowns. So when I edit one text block, its excerpt should change in the dropdown of the other text block when editing animation settings.

@miina Perhaps you can see why this happens? I feel like I don't see the forest for the trees right now.

swissspidy:
The picker looks a bit better already:

screenshot 2019-03-01 at 14 12 06

assets/src/components/animation-controls.js
swissspidy: Looks like it's just so that Gutenberg doesn't update the state immediately.

When I type something in block A, and then switch to another block, using `wp.data.select( 'core/editor' ).getBlock(…)` in the console still shows the old content for block A.

assets/src/components/animation-order-picker.js
```diff
@@ -0,0 +1,110 @@
...
+	// Todo: Cover more special cases if needed.
+	switch ( block.name ) {
+		case 'core/image':
+			if ( block.attributes.url ) {
+				label = block.attributes.url.slice( block.attributes.url.lastIndexOf( '/' ) );
+			}
+
+			break;
+		case 'amp/amp-story-text':
+			const content = block.originalContent ? block.originalContent.replace( /<[^<>]+>/g, ' ' ) : '';
```
miina: Maybe this would need to make use of `block.attributes.content` instead of `block.originalContent`? Perhaps the `originalContent` is not updated and shows the original value?

swissspidy: What did I say about the forest and the trees?! That was it 🤦‍♂️ 

Thanks @miina 🎉

swissspidy:
Added some truncating of the text too:

screenshot 2019-03-01 at 15 00 52

assets/src/components/with-animation-controls.js
```diff
@@ -1,22 +1,91 @@
...
+	const { ampAnimationAfter } = props.attributes;
+	const predecessor = getBlocksByClientId( getBlockOrder( page ) ).find( ( b ) => b.attributes.anchor === ampAnimationAfter );
+
+	return {
+		parentBlock: getBlock( getBlockRootClientId( props.clientId ) ),
+		animationAfter: predecessor ? predecessor.clientId : undefined,
+		getAnimatedBlocks() {
+			const { getBlockType } = select( 'core/blocks' );
+
+			const animatedBlocks = getAnimationOrder()[ page ] || [];
```
miina: There's an error in the console that seems to refer to this line, is this happening for you as well? (Note: The line numbers in the console don't match the ones in the source.)
screen shot 2019-03-01 at 4 01 49 pm

assets/src/amp-story-editor-blocks.js
```diff
@@ -33,7 +55,23 @@ domReady( () => {
 } );
 
 subscribe( () => {
-	setDefaultBlockName( getSelectedBlockClientId() ? 'amp/amp-story-text' : 'amp/amp-story-page' );
+	const defaultBlockName = getDefaultBlockName();
+	const selectedBlockClientId = getSelectedBlockClientId();
+
+	// Ensure that the default block is page when no block is selected.
+	if ( selectedBlockClientId && 'amp/amp-story-text' !== defaultBlockName ) {
+		setDefaultBlockName( 'amp/amp-story-text' );
```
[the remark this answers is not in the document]
miina: Perhaps related but just checked that at this moment it's not possible to add a Page -- the `+` icon (inserter) does not show if a Page is selected, nor does the inserter below the Page show up if a Page is selected, at the same time it seems to be possible to add a Page inside a Page.
screen shot 2019-03-01 at 4 07 20 pm

assets/src/components/with-animation-controls.js
miina: Additional information: The error occurs when creating a few blocks, assigning animations and then refreshing the page.

assets/src/amp-story-editor-blocks.js
```diff
@@ -44,12 +82,24 @@ subscribe( () => {
...
+		anchor: {
+			type: 'string',
+			source: 'attribute',
+			attribute: 'id',
+			selector: '*',
+		},
+	};
+
+	if ( ! ALLOWED_BLOCKS.includes( name ) ) {
+		settings.attributes = {
```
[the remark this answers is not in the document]
swissspidy: Indeed. I opened #1917 to unregister disallowed blocks. This should make this part here simpler.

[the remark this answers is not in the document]
swissspidy: And yes, it doesn't make sense to add attributes to blocks that will be unregistered.

assets/src/amp-story-editor-blocks.js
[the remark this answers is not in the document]
swissspidy: > If the currently select block is an `amp/amp-story-page` then shouldn't the default block also be `amp/amp-story-page` instead of a text block?

Indeed! I might have forgotten about this scenario when optimizing the subscriber.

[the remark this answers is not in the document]
swissspidy: > At the same time it seems to be possible to add a Page inside a Page.

Looks like `ALLOWED_BLOCKS` is used as the whitelist of inner blocks, which of contains pages as well.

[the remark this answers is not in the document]
swissspidy: Let's further investigate this in #1917

assets/src/components/with-animation-controls.js
swissspidy: Looks like there is more going on here, sometimes the content just disappears now. Ugh.

swissspidy:
The two main things left to do:

1. check cycle prevention again
2. fix state error

assets/src/components/with-animation-controls.js
swissspidy: Should be fixed now with ebd26139cd7f2463803c2992f1149a45f354d02b.

swissspidy:
Cycle prevention is fixed now. Example: When block A is set to start after B, one cannot select B to start after A. Nothing happens when you try to click on such an invalid option.

To further improve messaging to the user, we could think about hiding or disabling invalid choices in the picker. Perhaps with a detailed explanation below the control or something.

assets/src/amp-story-editor-blocks.js
```diff
@@ -1,28 +1,50 @@
...
 domReady( () => {
 	setDefaultBlockName( 'amp/amp-story-page' );
 
 	// Remove all blocks that aren't whitelisted.
 	getBlockTypes().filter( ( { name } ) => ! ALLOWED_BLOCKS.includes( name ) && 'amp/amp-story-page' ).map( ( { name } ) => unregisterBlockType( name ) );
 
+	// Set initial animation order state.
+	getBlocksByClientId( getBlockOrder() )
+		.filter( ( block ) => block.name === 'amp/amp-story-page' )
+		.map( ( page ) => {
```
westonruter: Since the return value is being discarded, why not use a `for...of` loop here?

assets/src/amp-story-editor-blocks.js
```diff
@@ -36,7 +58,29 @@ domReady( () => {
...
+		} else if ( 'amp/amp-story-page' !== selectedBlock.name && 'amp/amp-story-text' !== defaultBlockName ) {
+			setDefaultBlockName( 'amp/amp-story-text' );
+		}
+	} else if ( ! selectedBlockClientId && 'amp/amp-story-page' !== defaultBlockName ) {
+		setDefaultBlockName( 'amp/amp-story-page' );
+	}
+} );
+
+store.subscribe( () => {
+	Object.keys( getAnimationOrder() ).map( ( page ) => {
```
westonruter: Per above, why no `for...of`?

```js
for ( const page of Object.keys( getAnimationOrder() ).map( ( page ) => getBlock( page ) ).filter( Boolean ) ) {
    removePage( store.getState(), page );
}
```

assets/src/amp-story-editor-blocks.js
```diff
@@ -47,11 +91,18 @@ subscribe( () => {
  * @return {Object} Settings.
  */
 const addAMPAttributes = ( settings, name ) => {
-	if ( -1 === ALLOWED_CHILD_BLOCKS.indexOf( name ) ) {
+	if ( ! ALLOWED_CHILD_BLOCKS.includes( name ) ) {
```
westonruter: Excellent

assets/src/amp-story-editor-blocks.js
```diff
@@ -100,6 +151,10 @@ const addAMPAttributes = ( settings, name ) => {
 		};
 	}
 
+	// Disable anchor support as we auto-generate IDs.
+	settings.supports = settings.supports || {};
```
westonruter: Is it safe to mutate `supports` directly? Should it not be extended:

```suggestion
	settings = Object.assign( {}, settings );
	settings.supports = Object.assign( {}, settings.supports );
```

assets/src/components/with-animation-controls.js
```diff
@@ -1,22 +1,91 @@
...
+
+const applyWithSelect = withSelect( ( select, props ) => {
+	const { getSelectedBlockClientId, getBlockRootClientId, getBlock, getBlocksByClientId, getBlockOrder } = select( 'core/editor' );
+	const { getAnimationOrder } = select( 'amp/story' );
+
+	const currentBlock = getSelectedBlockClientId();
+	const page = getBlockRootClientId( currentBlock );
+
+	const { ampAnimationAfter } = props.attributes;
+	const predecessor = getBlocksByClientId( getBlockOrder( page ) ).find( ( b ) => b.attributes.anchor === ampAnimationAfter );
```
westonruter: What is `b`?

assets/src/amp-story-editor-blocks.js
swissspidy: I might actually rewrite it a bit to use object spread instead, makes the code more consistent too.

assets/src/components/with-animation-controls.js
swissspidy: b as in block. I think I shortened it originally because the variable block was already being used in the outer scope.

assets/src/amp-story-editor-blocks.js
swissspidy: Matter of personal preference I guess :-)

In most cases I prefer using `map` vs. `for` as it's easier to write and grasp, even when not using the return value. Functional programming just works really well in this context IMO.

Looking at this code again, I think this can use `getClientIdsWithDescendants` instead in order to avoid the nested loop. I'll have another look at it.

assets/src/amp-story-editor-blocks.js
swissspidy: I like the suggestion, but I'd go all the way with the chaining. I think it is more readable that way:

```js
Object.keys( getAnimationOrder() )
    .map( ( page ) => getBlock( page ) )
    .filter( Boolean )
    .map( ( page ) => removePage( store.getState(), page ) );
```

miina:
@swissspidy One additional note about the PR from functional testing:
- At this moment when creating the story and when haven't saved yet, it's possible to choose "Animate after" for a block that doesn't have animation yet, and then this happens with the element having assigned "animate-in-after" never appearing:
screen shot 2019-03-07 at 2 12 16 pm

swissspidy:
Interesting. So a block was shown in the dropdown that actually wasn't animated? Or did you add the animation to both and then removed it from the first (which then didn't update the other one)?

miina:
The block was shown in the dropdown what wasn't actually animated.
Note that this issue doesn't seem to occur if the Story has already been Saved once.

PR #1909, opened 2019-02-27
[AMP Stories] Fix filtering wrapper props.

miina:
Fixes the filter for wrapper props  (`addWrapperProps`) and restores the following features:
- Displaying font-family in the frontend;
- Adding `data-amp-selected` attribute to the selected parent;
- Displaying / hiding image caption;

### DIFF
--- a/assets/src/amp-story-editor-blocks.js
+++ b/assets/src/amp-story-editor-blocks.js
@@ -196,14 +196,20 @@ const addWrapperProps = ( BlockListBlock ) => {
 				'amp/amp-story-page' === blockName
 			)
 		) {
-			wrapperProps = Object.assign( {}, props.wrapperProps, { 'data-amp-selected': 'parent' } );
-		} else {
-			// If we have image caption or font-family set, add these to wrapper properties.
-			wrapperProps = Object.assign( {}, props.wrapperProps, {
-				'data-amp-image-caption': ( 'core/image' === blockName && ! attributes.ampShowImageCaption ) ? 'noCaption' : undefined,
-				'data-font-family': attributes.ampFontFamily || undefined,
-			} );
+			wrapperProps = {
+				...props.wrapperProps,
+				'data-amp-selected': 'parent',
+			};
+
+			return <BlockListBlock { ...props } wrapperProps={ wrapperProps } />;
 		}
+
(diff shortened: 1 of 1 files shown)


assets/src/amp-story-editor-blocks.js
```diff
@@ -179,35 +178,37 @@ const wrapperWithSelect = compose(
  * @param {Object} BlockListBlock BlockListBlock element.
  * @return {Function} Handler.
  */
-const addWrapperProps = createHigherOrderComponent(
```
miina: @swissspidy Let me know if there was a reason for using `createHigherOrderComponent` together with `wrapperWithSelect`, I might be not aware of something.

swissspidy: Looks like an artefact from when I was shuffling around all the components.

assets/src/amp-story-editor-blocks.js
```diff
@@ -179,35 +178,37 @@ const wrapperWithSelect = compose(
...
+		let wrapperProps;
+
+		// If we have an inner block selected let's add 'data-amp-selected=parent' to the wrapper.
+		if (
+			hasSelectedInnerBlock &&
+			(
+				'amp/amp-story-page' === blockName
+			)
+		) {
+			wrapperProps = Object.assign( {}, props.wrapperProps, { 'data-amp-selected': 'parent' } );
```
swissspidy: Why using `Object.assign` instead of just using `<BlockListBlock { ...props } data-amp-selected={ 'parent' } />`? I'd want to be consistent with using JSX and object spread here.

miina: The `data-amp-selected={ 'parent' }` needs to be part of `props.wrapperProps`, not `props`.
Here the `wrapperProps` is being created of the previously existing + the new properties and then later used within `<BlockListBlock { ...props } wrapperProps={ wrapperProps } />`.

Thoughts?

swissspidy: Gotcha, makes sense. Still could use spread operator though for that :-)

miina: Good point!

miina:
@swissspidy Just wanted to follow up on your comment about the spread operator and saw you've already done, it, thank you!

PR #1916, opened 2019-03-01
AMP Latest Stories dynamic block

### LINKED ISSUE #1910: AMP Stories (carousel/list) can be embedded in WordPress posts in an excerpt/preview form
- [ ] Create a block similar to the Latest Posts block in Gutenberg that lists stories. Bundle AMP Stories with an embeddable “Latest Stories” widget (dynamic block) that shows a list of stories in preview/excerpt form

kienstra:
* Begins the AMP Latest Stories dynamic block
* Uses a lot from the Latest Posts block in Core
* Work remains, including probably improving the front-end 

Closes #1910


includes/admin/class-amp-editor-blocks.php
```diff
@@ -204,4 +212,110 @@ public function print_dirty_amp_scripts() {
...
+	 * Then, this returns the smallest featured image height.
+	 *
+	 * @param array $posts An array or WP_Post objects.
+	 * @return int $minimum_height The smallest height of a featured image of any of the posts.
+	 */
+	public function get_minimum_height( $posts ) {
+		$minimum_height = 0;
+		$height_index   = 2;
+		foreach ( $posts as $post ) {
+			$image = wp_get_attachment_image_src( get_post_thumbnail_id( $post->ID ), self::LATEST_STORIES_IMAGE_SIZE );
```
westonruter: What if, by chance, a story doesn't have a featured image set?

includes/admin/class-amp-editor-blocks.php
```diff
@@ -204,4 +212,110 @@ public function print_dirty_amp_scripts() {
...
+			'suppress_filters' => false,
+		);
+
+		$story_query = new WP_Query( $args );
+
+		ob_start();
+		foreach ( $story_query->posts as $post ) :
+			?>
+			<a href="<?php echo esc_url( get_permalink( $post ) ); ?>">
+				<?php echo get_the_post_thumbnail( $post->ID, self::LATEST_STORIES_IMAGE_SIZE ); ?>
```
westonruter: What if there is no featured image? Should a fallback be used or should such posts be skipped altogether?

westonruter: I think the underlying title of the story should be included here, perhaps as the `alt` text or some other text that is displayed alongside the image.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -11,6 +11,13 @@
  */
 class AMP_Editor_Blocks {
 
+	/**
+	 * The image size for the Latest Stories block.
+	 *
+	 * @var string
+	 */
+	const LATEST_STORIES_IMAGE_SIZE = 'large';
```
westonruter: I think this could be too big. Wouldn't `medium` be more appropriate?

kienstra: Sure, medium probably looks better, with 94b83fcd120362f75bf6d8116a95968c1bb1733a:
latest-posts-medium

includes/admin/class-amp-editor-blocks.php
kienstra: Sure, 843f993e3167949ad268b2f1b73d0aae09a17091 adds the story title as the `img[alt]`.

I'm also looking for a way to display the title.

includes/admin/class-amp-editor-blocks.php
kienstra: Good point. https://github.com/ampproject/amp-wp/pull/1916/commits/5e2c053f0eef3185900ed1c62dac0d69b295a165 continues in the loop if `get_post_thumbnail_id()` returns an empty string.

includes/admin/class-amp-editor-blocks.php
kienstra: Good questions. For now, 5e2c053f0eef3185900ed1c62dac0d69b295a165 only outputs the `<a>` with the image is `has_post_thumbnail()`. I'll think about a possible fallback.

kienstra:
Here's the Latest Stories block so far, with the title displaying below the featured image in an `<amp-fit-text>`. I'll keep looking at better ways to display the title.

latest-stories-so-far

includes/admin/class-amp-editor-blocks.php
kienstra: 246333c7e08fbdc3b7e33f9844dc64ac50633252 uses a fallback of any image attached to a post, in case there's no featured image. A more thorough solution might involve parsing the `post_content` for images, like from the image block.

kienstra: The current title display is [shown below](https://github.com/ampproject/amp-wp/pull/1916#issuecomment-469291561).

westonruter:
Note that the block could show up in non-AMP pages so using an `amp-carousel` may be a problem.

In general the design should I think follow how AMP Stories are presented in Google Search Results or the Google Discover feed.

westonruter:
In regards to accounting for there being no featured image, I recall that a featured image is _required_ for AMP Stories and that we will likely be blocking the publishing of a story until a featured image has been selected. So I think this should be less of a concern than normal. Also, the `WP_Query` could be modified to make sure it only queries posts that have a featured image, by adding `'meta_key' => '_thumbnail_id',` up front.

kienstra:
**Question About Display**

Hi @westonruter,
Thanks, good point that `<amp-carousel>` probably won't work for this.

What do you think about this display for the Latest Stories block?
(taken from https://www.seroundtable.com/google-amp-stories-25232.html)

seroundtable-amp-stories

It's similar to the display in the document [Understand how AMP looks in search results](https://developers.google.com/search/docs/guides/about-amp) (next to "Visual stories: A visual storytelling format")

Though there could be an issue if the featured image is too light, and the white text doesn't appear.

I couldn't find an AMP story in the Google search results or in the Google app feed, even on an Android phone.

westonruter:
@kienstra that looks great, yes. The featured image being too light could be accounted for by the gradient overlay, no?

If the gradient is not sufficient, a black drop shadow on the text could help.

Eventually, there could also be an opportunity here to use PHP to obtain a sampling of the pixels in the image itself, or rather to obtain the background color of the first page in the `amp-story` post itself. While the background color should be manually supplied currently, in the future it should also be automatically determined by sampling the pixels in JS, and then populating the story page prop. In either case, the background color of the first page should give a good indication of what the overall color of the featured image will be (hopefully) and that could inform whether white text or black text is used.

But for now, let's try with white text and a gradient overlay alone.

Also, I'm reconsidering what I said about `amp-carousel`. Given that AMP components will eventually be allowed outside of AMP documents, perhaps we should just go ahead and use `amp-carousel` in the block, whether the block is rendered on an AMP page or not. Note that in a non-AMP page, you'll have to manually do `wp_enqueue_script('amp-carousel')` since the sanitizer wouldn't be doing that for us. Let's try to embrace the future.

kienstra:
**Block With Gradient and amp-carousel**

Hi @westonruter,
Good idea to use the `<amp-carousel>` even on non-AMP. Here's the block. "Latest Stories" is the post title, and isn't part of the block:

showing-amp-story-2

Also, I think your idea of using a gradient works well:
gradient-demo

includes/admin/class-amp-editor-blocks.php
```diff
@@ -11,6 +11,13 @@
  */
 class AMP_Editor_Blocks {
 
+	/**
+	 * The ideal image size for the Latest Stories block.
+	 *
+	 * @var string
+	 */
+	public $image_size = array( 300, 400 );
```
kienstra: For most of the stories [here](https://www.ampproject.org/stories) that have a `poster-portrait-src` image, that image's dimensions are 600 x 800 px, like [this story](https://mic.com/stories/262/the-art-of-an-organized-refrigerator). So this is half of that.

Of course, this isn't a common dimension in WordPress.

kienstra:
**Block In The Editor**

Hi @westonruter,
What do you think of using this non-AMP alternative to an `<amp-carousel>`? It'll allow showing the block in the editor, using `<ServerSideRender>`.

The [latest commit](https://github.com/ampproject/amp-wp/pull/1916/commits/f9b06ed5ff48a380b483d7fe11b196a4cd489852) changes this, but I could easily revert it. It's a very simple CSS-only horizontal display of the images.

amp-stories-editor

westonruter:
@kienstra Can meta than the title be shown? As seen in the Google Discover feed that can be done with the author avatar, name, and then relative date.

image

westonruter:
> What do you think of using this non-AMP alternative to an `<amp-carousel>`? It'll allow showing the block in the editor, using `<ServerSideRender>`.

That would be OK, but it seems like the styling is broken:

image

Yes, I think that is fine.

includes/admin/class-amp-editor-blocks.php
westonruter: What about re-using the dimensions defined here:

https://github.com/ampproject/amp-wp/blob/3f570168fdfc9b4f0713089ad401852f52e88469/includes/class-amp-story-post-type.php#L86-L96

westonruter:
@kienstra It seems that the block in the edit post screen is missing styles:

image

kienstra:
Hi @westonruter,
>Can meta than the title be shown? As seen in the Google Discover feed that can be done with the author avatar, name, and then relative date.

Sure, that's a good idea. We'd still show the title, right? And also display the meta.

>It seems that the block in the edit post screen is missing styles

Thanks, I'll work on that.

westonruter:
> We'd still show the title, right? And also display the meta.

Correct: title, avatar, author name, relative date.

kienstra:
**Question About Styling In Block Editor**

Hi @westonruter,
Thanks for testing this. Is the Latest Stories block styling still not appearing in your environment for the block editor? Maybe [your commit](https://github.com/ampproject/amp-wp/pull/1916/commits/56967f4cea58cff48b3c3ee79cd55770959d239e) took care of that.

It looks like the [enqueue_block_assets callback](https://github.com/ampproject/amp-wp/pull/1916/files#diff-af86f8db0fb2fe24a8aea92de5ee5128R54) is running properly in my local for the block editor screen. But that could just be in my environment:
latest-stories-block-styling

westonruter:
> Thanks for testing this. Is the Latest Stories block styling still not appearing in your environment for the block editor? Maybe [your commit](https://github.com/ampproject/amp-wp/pull/1916/commits/56967f4cea58cff48b3c3ee79cd55770959d239e) took care of that.

I'm still seeing the problem. It looks like Twenty Seventeen has a style rule that is interfering:

image

I'm still seeing:

image

Note the first image is a landscape image, so that is why all of the slides are short. This normally should not happen once we properly set the featured image. Nevertheless, something should be done to prevent images from overflowing the `.slide` container.

kienstra:
**Image Overflow And list-style-type Should Be Fixed**

Hi @westonruter,
Thanks for your patience. With 665981a4b08c504ae5505998aa7a88372147d4fb, the `list-style-type` bullet and the image overflowing in the editor should be fixed. Of course, there are still styling issues remaining that I'm working on.

featured-img-stories

Do you see this also?

includes/admin/class-amp-editor-blocks.php
kienstra: Thanks, great idea. f325810d80c6b8dceb60ab913563e78fa77df121 uses that size `'amp-story-poster-portrait'`.

webpack.config.js
```diff
@@ -75,6 +75,7 @@ module.exports = {
 		'./assets/js/amp-validation-single-error-url-details-compiled': './assets/src/amp-validation-single-error-url-details.js',
 		'./assets/js/amp-story-editor-blocks-compiled': './assets/src/amp-story-editor-blocks.js',
 		'./assets/js/amp-story-blocks-compiled': './assets/src/blocks/stories.js',
+		'./assets/js/amp-agnostic-blocks-compiled': './assets/src/blocks/amp-agnostic.js',
```
kienstra: Maybe creating a new built JS file isn't needed for this. But it sounds like this block needs to be available in non-AMP, while `amp-blocks-compiled.js` is only available in Native AMP.

kienstra:

Hi @westonruter,
Could you please look at this PR again? The styling matches [your screenshot](https://github.com/ampproject/amp-wp/pull/1916#issuecomment-470207600) from the Discover feed better.

demo-stories

Thanks, Weston!

includes/admin/class-amp-editor-blocks.php
```diff
@@ -177,6 +195,14 @@ public function enqueue_block_editor_assets() {
 		}
 	}
 
+	/**
+	 * Enqueues the front-end block stylesheet.
+	 */
+	public function enqueue_block_assets() {
+		$stylesheet_base = 'amp-blocks';
+		wp_enqueue_style( $stylesheet_base . '-style', amp_get_asset_url( "/css/{$stylesheet_base}.css" ), array(), AMP__VERSION );
```
westonruter: We should not enqueue this unless it is actually needed. So what that means is that we should call `wp_enqueue_style()` _inside_ the `render_callback`. As an added benefit, browsers now will load the stylesheet asynchronously.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -123,6 +132,15 @@ public function whitelist_block_atts_in_wp_kses_allowed_html( $tags, $context )
 	 */
 	public function enqueue_block_editor_assets() {
 
+		$amp_agnostic_slug = 'amp-agnostic-blocks-compiled';
+		wp_enqueue_script(
```
westonruter: I don't think this should be in a separate script. It should rather be part of `amp-editor-blocks`, or even as part of `amp-editor-blocks-build`. The filtering out of non-AMP native blocks could be done in JS instead, as the ideal is for all AMP components to be usable in non-AMP pages this year anyway.

I don't feel strongly about this, however.

includes/admin/class-amp-editor-blocks.php
kienstra: Great idea, https://github.com/ampproject/amp-wp/pull/1916/commits/06a519c2b7b4e8d87d73ddd3fd44fdaaa35cc6b4 enqueues the stylesheet in the `render_callback`.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -204,4 +230,165 @@ public function print_dirty_amp_scripts() {
...
+		}
+
+		ob_start();
+		?>
+		<div class="<?php echo esc_attr( $class ); ?>">
+			<div class="latest-stories-carousel" style="height:<?php echo esc_attr( $min_height ); ?>px;">
+				<?php
+				foreach ( $story_query->posts as $post ) :
+					$thumbnail_id = get_post_thumbnail_id( $post );
+					if ( $thumbnail_id ) :
```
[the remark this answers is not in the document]
kienstra: Good point, https://github.com/ampproject/amp-wp/pull/1916/commits/c6da6d2f0ba3fe5a187cd7cd3fe29e3848bb8401#diff-37ed1ba86eaee9e72c2ced620309452bR602 adds your snippet to the new method.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -204,4 +230,165 @@ public function print_dirty_amp_scripts() {
...
+										esc_html( $author_display_name ),
+										esc_html( human_time_diff( get_post_time( 'U', false, $post->ID ), current_time( 'timestamp' ) ) )
+									);
+									?>
+								</span>
+							</div>
+						</a>
+						<?php
+					endif;
+				endforeach;
```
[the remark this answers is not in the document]
kienstra: Sure, https://github.com/ampproject/amp-wp/commit/c6da6d2f0ba3fe5a187cd7cd3fe29e3848bb8401 abstracts the markup in the loop into a new method.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -204,4 +230,165 @@ public function print_dirty_amp_scripts() {
...
+		$min_height  = $this->get_minimum_dimension( 'height', $story_query->posts );
+		$class       = 'amp-block-latest-stories';
+		if ( isset( $attributes['className'] ) ) {
+			$class .= ' ' . $attributes['className'];
+		}
+
+		ob_start();
+		?>
+		<div class="<?php echo esc_attr( $class ); ?>">
+			<div class="latest-stories-carousel" style="height:<?php echo esc_attr( $min_height ); ?>px;">
```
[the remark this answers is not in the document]
kienstra: Sure, I'm working on using `<ul>` now.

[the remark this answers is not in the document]
kienstra: Good idea, https://github.com/ampproject/amp-wp/pull/1916/commits/479bf16395022d10beb0b74c7ef61900253fdf28 uses `<ul>` and `<li>`.

includes/admin/class-amp-editor-blocks.php
kienstra: Good point, https://github.com/ampproject/amp-wp/pull/1916/commits/55d9d5ad2ae192a0e80504af340b3f2202872bb4 and https://github.com/ampproject/amp-wp/pull/1916/commits/018714479b4345cfce0d0870d8155dcf34a8040c remove that separate script, and [unregister](https://github.com/ampproject/amp-wp/pull/1916/commits/018714479b4345cfce0d0870d8155dcf34a8040c#diff-7ac3542e9bd216dc31cffe9828cae472R926) AMP-dependent blocks using [a blacklist approach](https://wordpress.org/gutenberg/handbook/designers-developers/developers/filters/block-filters/#using-a-blacklist) that the documentation recommends.

However if there was already an AMP-dependent block on the page, and it gets unregistered, there's an error.

I need to look at this more.

For example:
1. In Native AMP, add an `amp-mathml` block to some post
2. Change to Classic mode
3. Refresh the post's block editor
4. This appears:
>TypeError: Cannot read property 'title' of undefined at editor.min.js?ver=9.0.11:55

js-error

assets/js/amp-editor-blocks.js
```diff
@@ -113,6 +113,7 @@ const ampEditorBlocks = ( function() {
 		wp.hooks.addFilter( 'blocks.getSaveElement', 'ampEditorBlocks/filterSave', component.filterBlocksSave );
 		wp.hooks.addFilter( 'editor.BlockEdit', 'ampEditorBlocks/filterEdit', component.filterBlocksEdit );
 		wp.hooks.addFilter( 'blocks.getSaveContent.extraProps', 'ampEditorBlocks/addExtraAttributes', component.addAMPExtraProps );
+		wp.domReady( component.maybeUnregisterBlocks );
```
westonruter: Does calling this immediately (and not waiting for `domReady`) help with the error?

kienstra: Great idea, https://github.com/ampproject/amp-wp/pull/1916/commits/01bafa706c3e22c180326ba68c4f603f3d13a6a4 calls it right away, without `wp.domReady`. 

Now, the warning is as expected when converting to Classic mode:
warning-as-expected

westonruter: So there are no more outstanding issues as far as you're concerned, and it is ready for final review?

includes/admin/class-amp-editor-blocks.php
```diff
@@ -204,4 +205,131 @@ public function print_dirty_amp_scripts() {
...
+		$min_height  = $this->get_minimum_dimension( 'height', $story_query->posts );
+		$class       = 'amp-block-latest-stories';
+		if ( isset( $attributes['className'] ) ) {
+			$class .= ' ' . $attributes['className'];
+		}
+
+		ob_start();
+		?>
+		<div class="<?php echo esc_attr( $class ); ?>">
+			<ul class="latest-stories-carousel" style="height:<?php echo esc_attr( $min_height ); ?>px;">
```
westonruter: I realize that this contradicts what I said before about this being a `ul`, but isn't this actually supposed to be an `amp-carousel` outside of the editor?

assets/js/amp-editor-blocks.js
kienstra: Yes, sorry for the delay in replying.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -124,7 +125,7 @@ public function whitelist_block_atts_in_wp_kses_allowed_html( $tags, $context )
 	public function enqueue_block_editor_assets() {
 
 		// Enqueue script and style for AMP-specific blocks.
-		if ( amp_is_canonical() && AMP_Story_Post_Type::POST_TYPE_SLUG !== get_current_screen()->post_type ) {
+		if ( AMP_Story_Post_Type::POST_TYPE_SLUG !== get_current_screen()->post_type ) {
```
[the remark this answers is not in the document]
kienstra: Good question. I don't know of a reason to exclude AMP blocks like `amp-jwplayer` from AMP stories, though I don't have much experience with AMP stories yet.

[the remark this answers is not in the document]
westonruter: We can leave it as-is for now. Part of a broader discussion about which blocks go into AMP stories.

assets/js/amp-editor-blocks.js
```diff
@@ -903,6 +904,30 @@ const ampEditorBlocks = ( function() {
...
+			'amp-jwplayer',
+			'amp-mathml',
+			'amp-o2-player',
+			'amp-ooyala-player',
+			'amp-reach-player',
+			'amp-springboard-player',
+			'amp-timeago',
+		];
+
+		if ( ! component.data.hasThemeSupport ) {
```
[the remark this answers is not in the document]
kienstra: Right, that's a good catch. https://github.com/ampproject/amp-wp/pull/1916/commits/2be208de3acbebc28aaa1abe837834effc61abc2 changes the check to `amp_is_canonical()`.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -204,4 +205,131 @@ public function print_dirty_amp_scripts() {
...
+	 * This iterates through all of the posts, to find their featured image.
+	 * Then, this returns the smallest dimension (width or height).
+	 * For example, if $dimension is 'width' and the featured image widths are 100, 200 and 300,
+	 * this will return 100.
+	 *
+	 * @param string $dimension The dimension, either 'width' or 'height'.
+	 * @param array  $posts An array or WP_Post objects.
+	 * @return int $minimum_dimension The smallest dimension of a featured image.
+	 */
+	public function get_minimum_dimension( $dimension, $posts ) {
```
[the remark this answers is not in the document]
kienstra: Yeah, good idea. https://github.com/ampproject/amp-wp/pull/1916/commits/1dc1a852d4ecf172e323100a7854c9835621b2da changes the method name to `get_featured_image_minmum_height()`.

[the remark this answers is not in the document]
kienstra: Good point, https://github.com/ampproject/amp-wp/pull/1916/commits/1dc1a852d4ecf172e323100a7854c9835621b2da changes the method name to `get_featured_image_minmum_height()`.

includes/admin/class-amp-editor-blocks.php
```diff
@@ -204,4 +205,131 @@ public function print_dirty_amp_scripts() {
...
+		$class       = 'amp-block-latest-stories';
+		if ( isset( $attributes['className'] ) ) {
+			$class .= ' ' . $attributes['className'];
+		}
+
+		ob_start();
+		?>
+		<div class="<?php echo esc_attr( $class ); ?>">
+			<ul class="latest-stories-carousel" style="height:<?php echo esc_attr( $min_height ); ?>px;">
+				<?php
```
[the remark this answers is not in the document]
kienstra: Sure, good idea. https://github.com/ampproject/amp-wp/pull/1916/commits/f0ebfee589b176a41a994ad7d6cf324142ffe575 uses that suggestion.

assets/src/blocks/amp-latest-stories/index.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Internal dependencies
+ */
+import edit from './edit';
+
+export const name = 'amp/amp-latest-stories';
+
+export const settings = {
+	title: __( 'AMP Latest Stories', 'amp' ),
```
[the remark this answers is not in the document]
kienstra: Sure, https://github.com/ampproject/amp-wp/pull/1916/commits/4b7a1d239ad16e86f02a0ed5191533aeb44f3e0b changes the title to 'Latest Stories'.

includes/admin/class-amp-editor-blocks.php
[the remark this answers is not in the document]
kienstra: Sounds good

includes/admin/class-amp-editor-blocks.php
kienstra: Sure, this could be an `<amp-carousel>` only on the front-end if you'd like. The idea of having the same `<ul>` carousel on the front-end and editor was to make this more DRY.

Still, there wouldn't be too much duplication in code by having the `<amp-carousel>` on the front-end. Just slight changes in markup, and some extra editor styles. Here's what it looks like when I just changed the `<ul>` to an `<amp-carousel>`:

amp-carousel

If you'd like to have an `<amp-carousel>` on the front-end, maybe this could pass an attribute to [ServerSideRender](https://wordpress.org/gutenberg/handbook/designers-developers/developers/components/server-side-render/#attributes) like `isEditor`, so that it's different in the editor and front-end.

westonruter: Yeah, that sounds good. As for DRY, we'll eventually only include `amp-carousel` so we can live with some WET for now.

Can the carousel be `fixed-height` so it can show multiple slides at the same time?

kienstra: Sure, https://github.com/ampproject/amp-wp/pull/1916/commits/1fb85db0899da34eaf265f8827c11f59c4cb8ef7 outputs a fixed-height `<amp-carousel>` on the front end. There are still styling fixes needed for the `<amp-carousel>`. If it's alright, I'll apply those tomorrow. Sorry I've been slow replying lately.

kienstra: **Some Styling Fixes Made**
*More To Come*

https://github.com/ampproject/amp-wp/pull/1916/commits/58bfd396f282ae7092d35604a0c8005907441ac2 makes some styling improvements, but some more are needed:
current-carousel

kienstra:
**Suggestions Applied**

Hi @westonruter,
Sorry for the delay. With the latest commits, your review suggestions are now applied. The `<amp-carousel>` is now styled better:

amp-carousel

westonruter:
Oops. And I broke it.

kienstra:
Ah, sorry for forgetting to enqueue the `amp-carousel` script.

westonruter:
@kienstra thanks, this is great.

PR #1917, opened 2019-03-01
[AMP Stories] Remove all blocks that aren't whitelisted

swissspidy:
Without this, it's still possible to insert paragraph blocks and other blocks that aren't meant to be used in stories.

### DIFF
--- a/assets/src/amp-story-editor-blocks.js
+++ b/assets/src/amp-story-editor-blocks.js
@@ -11,7 +11,7 @@ import { select, subscribe } from '@wordpress/data';
  * Internal dependencies
  */
 import { withAttributes, withParentBlock, withBlockName, withHasSelectedInnerBlock, withAmpStorySettings, withAnimationControls } from './components';
-import { ALLOWED_BLOCKS, BLOCK_TAG_MAPPING } from './constants';
+import { ALLOWED_BLOCKS, ALLOWED_CHILD_BLOCKS, BLOCK_TAG_MAPPING } from './constants';
 import { maybeEnqueueFontStyle } from './helpers';
 
 const { getSelectedBlockClientId, getBlockOrder, getBlocksByClientId } = select( 'core/editor' );
@@ -19,6 +19,9 @@ const { getSelectedBlockClientId, getBlockOrder, getBlocksByClientId } = select(
 // Ensure that the default block is page when no block is selected.
 domReady( () => {
 	setDefaultBlockName( 'amp/amp-story-page' );
+
+	// Remove all blocks that aren't whitelisted.
+	getBlockTypes().filter( ( { name } ) => ! ALLOWED_BLOCKS.includes( name ) && 'amp/amp-story-page' ).map( ( { name } ) => unregisterBlockType( name ) );
 
 	// Load all needed fonts.
 	getBlocksByClientId( getBlockOrder() )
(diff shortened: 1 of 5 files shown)


assets/src/amp-story-editor-blocks.js
```diff
@@ -36,6 +36,9 @@ subscribe( () => {
 	setDefaultBlockName( getSelectedBlockClientId() ? 'amp/amp-story-text' : 'amp/amp-story-page' );
 } );
 
+// Remove all blocks that aren't whitelisted.
+getBlockTypes().filter( ( { name } ) => ! ALLOWED_BLOCKS.includes( name ) ).map( ( { name } ) => unregisterBlockType( name ) );
```
miina: Maybe this should also be inside the `domReady`, trying to get block types from console still receives all the blocks, maybe they are not registered at that point yet? Also, probably `amp/amp-story-page` should also be excluded -- it doesn't seem to be included in the ALLOWED_BLOCKS at this moment.

swissspidy: This is very well possible. Unfortunately this leads to some errors in the console.

Please have a look at the latest changes. 

Perhaps we have to do filter `allowed_block_types` in PHP instead of doing it client-side?

swissspidy:
Ugh, looks like  one can still insert pages inside pages, even though it's forbidden with `allowedBlocks`.

assets/src/amp-story-editor-blocks.js
```diff
@@ -163,7 +166,7 @@ const setBlockParent = ( props ) => {
 		};
```
miina: ^ `ALLOWED_BLOCKS` within `setBlockParent` would also need switching for `ALLOWED_CHILD_BLOCKS` -- otherwise AMP Story Page requires another Page to be its parent and it won't be possible to insert it to the root level (and probably would be possible within another Page, setting `parent` overrides the `allowedBlocks` of `InnerBlocks`).

miina: ^ That should fix it!

> Ugh, looks like one can still insert pages inside pages, even though it's forbidden with `allowedBlocks`.

swissspidy: Yay! So it was just an oversight from my side, phew :-)

swissspidy:
Works much better now!

assets/src/amp-story-editor-blocks.js
```diff
@@ -20,6 +20,9 @@ const { getSelectedBlockClientId, getBlockOrder, getBlocksByClientId } = select(
 domReady( () => {
 	setDefaultBlockName( 'amp/amp-story-page' );
 
+	// Remove all blocks that aren't whitelisted.
+	getBlockTypes().filter( ( { name } ) => ! ALLOWED_BLOCKS.includes( name ) && 'amp/amp-story-page' ).map( ( { name } ) => unregisterBlockType( name ) );
```
westonruter: Minor point, but I think it would be better to not use `filter` as a control structure. In other words, I think it would be better to do:

```js
for ( const block of getBlockTypes().filter( ( { name } ) => ! ALLOWED_BLOCKS.includes( name ) && 'amp/amp-story-page' ) )  {
    unregisterBlockType( block.name );
}
```

But take a look at the condition here: `! ALLOWED_BLOCKS.includes( name ) && 'amp/amp-story-page'`. Does it make sense? Does does it intend to say rather `! ALLOWED_BLOCKS.includes( name ) && name !== 'amp/amp-story-page'`?

swissspidy: That condition is indeed flawed. `&& 'amp/amp-story-page'` does nothing really. It would need to be `name !== 'amp/amp-story-page'`.

However, I later changed the code so that `ALLOWED_BLOCKS` includes `amp/amp-story-page` as well, which means I can remove that check without a replacement.

As for `map` vs. `for` I'll try to reply at #1904

Issue #1881, opened 2019-02-19
Semantic meaning of text block


ThierryA:
* [x] Implement semantic meaning estimation algorithm to solve multi-variate optimization problem (e.g. font size, paragraph size, element location)
* [x] Take font size into account
* [x] Take word count into account
* [x] Take element position into account

swissspidy:
Related: #1883.

Element location depends on #1897 I guess.

csossi:
Verified in QA

PR #1918, opened 2019-03-01
Take text length into account for determining semantics

swissspidy:
Uses `@wordpress/wordcount` to determine whether the text block should use `<h1>`, `<h2>`, or a `<p>` tag.

See #1881.


assets/src/blocks/amp-story-text/get-tag-name.js
```diff
@@ -1,12 +1,25 @@
...
 const H2_FONT_SIZE = 24;
+const H1_TEXT_LENGTH = 15;
+const H2_TEXT_LENGTH = 30;
+
+/*
+ * translators: If your word count is based on single characters (e.g. East Asian characters),
+ * enter 'characters_excluding_spaces' or 'characters_including_spaces'. Otherwise, enter 'words'.
+ * Do not translate into your own language.
+ */
+const wordCountType = _x( 'words', 'Word count type. Do not translate!', 'amp' );
```
miina: Are the `H1_TEXT_LENGTH` and `H2_TEXT_LENGTH` values definitely meant in words? 15 words seems quite long for a `h1` on the AMP Story Page :)

assets/src/blocks/amp-story-text/get-tag-name.js
```diff
@@ -25,5 +38,15 @@ export default function( attributes ) {
 		return 'h2';
 	}
 
+	const textLength = count( attributes.content, wordCountType, {} );
+
+	if ( H1_TEXT_LENGTH >= textLength ) {
```
miina: I'm assuming it's intentional that the font size gets priority over the word count -- maybe it would be good to add a comment about this so that no one would have the idea to merge the word count conditions with the previous conditions? Not a real issue though.

assets/src/blocks/amp-story-text/get-tag-name.js
swissspidy: Ugh, I guess I was thinking in characters.

assets/src/blocks/amp-story-text/get-tag-name.js
swissspidy: Intentional in the way that a huge font size is probably a better signal of a heading than the word count.

miina:
Just tested it and the word count seems good now. There's one issue with switching between the tags -- the element loses its focus while typing so every time the switch happens the user needs to click on the element again to continue typing. Maybe there's a way to restore the focus after switching the element? Thoughts?

swissspidy:
> Maybe there's a way to restore the focus after switching the element? Thoughts?

Good point. The easiest way is to not change the element in the editor, but only on save. It shouldn't make a difference in the editor really when the styling is properly done.

Issue #1907, opened 2019-02-27
Reordering of pages


ThierryA:
- [ ] Add reorder mode that shrinks story pages via CSS transform so that all pages can be shown in the viewport at once, and allow them to be drag-n-dropped to new positions via jQuery UI (FTW)

dawidmlynarz:
👋 Hello! I have worked on the interactions for the *reordering pages*. Again I have produced Principle prototype to show you more interactions and flows. Please take a look at the prototype attached and share your thoughts.

![amp stories - reordering](https://user-images.githubusercontent.com/8026823/53569339-42808c80-3b64-11e9-87b1-97d37a64fbf9.gif)

🙌 Looking forward to your feedback!

jwold:
This looks really good @dawidmlynarz! One thing I want to callout is that from an accessibility perspective there also needs to be a way to do this without mouse support. Great work :)

dawidmlynarz:
🙏 Thanks! Good point with accessibility! I guess that adding some simple arrows while interacting with any Page in this view could be handy. 

I’m happy to explore that but I would like to know if this could be in the scope. Snother thing is if it should be done within the same ticket or as some separate enhancement for this feature?🙂

ThierryA:
Awesome work @dawidmlynarz, will drop a few comments on Invision shortly.

> I’m happy to explore that but I would like to know if this could be in the scope. Snother thing is if it should be done within the same ticket or as some separate enhancement for this feature?🙂

Accessibility is definitely part of the scope of v1.0.0 stable but not a P0 for the beta version, so we can revisit when we move to the v.1.0.0 stable version.

csossi:
Verified in QA

Issue #1903, opened 2019-02-26
Page horizontal viewing


ThierryA:
- [x] Remove up/down movement block arrows in page controls, or replace with right/left arrows
- [x] Implement carousel animation to show adjacent pages as you cycle through

dawidmlynarz:
👋 Hello! I have been exploring possibilities for the horizontal navigation between AMP Stories. For now, I called each of the items _Slides_ as it somehow fitted more than a _Page_.

Given the fact that this is a highly interactive project, I have produced Principle prototype to show you more interactions and flows instead of a static screen that may cause a lot of questions. Please take a look at the video attached and share your thoughts.

![amp stories - horizontal](https://user-images.githubusercontent.com/8026823/53483385-ad0ccc00-3a81-11e9-812f-d02cd1272cc3.gif)

Besides that, I have also created [Invision prototype holding some static screens](https://invis.io/NSQQARUCY79) so you can leave comments inside. Please have a look at the link below and check your emails as you should get the invitations.

Looking forward to your feedback!

ThierryA:
Heya @dawidmlynarz, overall I really like it, awesome work 💥 I left a few comments on [Invision](https://invis.io/NSQQARUCY79).

One additional comment since it applies globally throughout the design is to increase contrast overall for accessibility purposes.

Looking forward to the round of reviews.

dawidmlynarz:
Thanks for the input @ThierryA - all the comments on Invision are addressed. Really good point about accessibility, I ave increased the contrast of the cards and some elements. I was mainly using colors according to the WordPress palette from Sketch UI Kit so let me know if they need to be optimized even more.

Please find the revised prototype below with the fading and slightly revised bottom navigation. Now indicators are positioned in the center below the active slide.

![amp stories - horizontal browsing - v2](https://user-images.githubusercontent.com/8026823/53489718-2dd3c400-3a92-11e9-9307-6cdedc17ef1c.gif)

If you think (from engineering perspective) that we need to refine the design, please also let me know and we will work until the feasible version. I'm open to iterate!

🙌 Looking forward to your feedback!

ThierryA:
Great, round of reviews is great.

> I called each of the items Slides as it somehow fitted more than a Page

Coming back to the comment above, AMP Stories have the concept of "Book Pages" and are referred as pages throughout the AMP Stories docs so I would suggest to stick to it.

Will take a other pass at the review with @albertomedina before it moves to engineering 😉

[...]

Issue #1902, opened 2019-02-26
[AMP Stories] Show list of templates to use when inserting a new blank page


ThierryA:
AC: User can select pre-defined single-page sample templates (for this issue, it can be PoC and not the final set of templates);
AC: User can't remove or modify the base templates in the "Template Inserter" but can modify, remove, edit blocks that are inserted by selecting a template.

- [x] Engage with Gutenberg team on status of UI of extending reusable blocks as templates
- [x] Show list of story page templates in inserter along with the blank page. Show preview of pages. The template has to be inserted to the editor as regular blocks so that they can be edited/removed, etc.

dawidmlynarz:
This part was solved for some time now in Invision but I didn't post the update so I will upload status anyway. Below you can see two cases when you can use templates. 

1. When you have blank Page and want to include a template, there is a modal that appears and you can choose whatever you need.

![ezgif com-optimize](https://user-images.githubusercontent.com/8026823/54679683-32771f80-4b08-11e9-9793-bd9696cd98c9.gif)

2. When you are already editing and you want to include one more page.

![ezgif com-optimize (1)](https://user-images.githubusercontent.com/8026823/54679954-e4165080-4b08-11e9-8d6d-b15ccf72a1dd.gif)

Looking forward to broader feedback!

amedina:
These look very good!

miina:
FI: I'm starting looking into the technical side of this issue.
cc @swissspidy

PR #1922, opened 2019-03-04
[AMP Stories] Extended Page Editor Controls

swissspidy:
**Covers:**

* [x] Carousel-like UI for navigating between pages, always keeping one page in focus
* [x] UI for reordering pages 

**Screenshots:**

Screenshot 2019-03-10 at 23 47 19
Screenshot 2019-03-10 at 23 47 28
Screenshot 2019-03-10 at 23 47 58

See #1903.
See #1907.

----

**To do after code review:**

* [x] Add Inserter for now for inserter button
* [x] Hide block navigation when reordering
* [x] Address individual notes from code review
* [x] Enable clicking on individual pages to change the current page
* [x] Improve reorder DropZones
* [x] Docs improvements

Accessibility & mobile improvements can be tackled in following PRs.


assets/src/amp-story-editor-blocks.js
```diff
@@ -52,9 +63,69 @@ domReady( () => {
 			maybeEnqueueFontStyle( block.attributes.ampFontFamily );
 		}
 	}
+
+	renderStoryComponents();
+
+	// Prevent WritingFlow component from focusing on last text field when clicking below the carousel.
+	document.getElementsByClassName( 'editor-writing-flow__click-redirect' )[ 0 ].remove();
```
westonruter: Maybe a bit better to use `querySelector` since we're only needing the first:

```suggestion
	document.querySelector( '.editor-writing-flow__click-redirect' ).remove();
```

assets/src/amp-story-editor-blocks.js
```diff
@@ -52,9 +63,69 @@ domReady( () => {
...
 } );
 
+/**
+ * Add some additional elements needed to render our custom UI controls.
+ */
+function renderStoryComponents() {
+	const editorBlockList = document.getElementsByClassName( 'editor-block-list__layout' )[ 0 ];
+	const postTitle = document.getElementsByClassName( 'editor-post-title' )[ 0 ];
```
westonruter: See note above about using `querySelector`.

swissspidy:
There might be some quirks here and there, but the reordering and horizontal editing works well so far.

swissspidy:
This is how it should look like according to the [design](https://projects.invisionapp.com/share/NSQQARUCY79#/screens):

Screenshot 2019-03-11 at 11 36 52

westonruter:
On screens that are narrower than 1500px there is overlapping of the Block Navigation with the current page:

image

westonruter:
I see the problem. Looks like we need to use `100vw/2`. Working on commit.

assets/src/blocks/amp-story/edit.js
```diff
@@ -48,36 +45,6 @@ class EditPage extends Component {
 		this.onSelectMedia = this.onSelectMedia.bind( this );
 	}
 
-	maybeAddBlockNavigation() {
-		// If no blocks are selected or if it's the current page, change the view.
-		if ( this.props.showBlockNavigation ) {
```
miina: Do we still need the determining the value of `showBlockNavigation` in the `withSelect` below or is that redundant now?

assets/src/components/story-controls.js
```diff
@@ -0,0 +1,81 @@
...
+				onClick={ startReordering }
+			/>
+		</Fragment>
+	);
+}
+
+// Todo: Move pages and moveBlockToPosition to separate reorder component.
+export default compose(
+	withSelect( ( select ) => {
+		const {	getBlocksByClientId	} = select( 'core/editor' );
```
miina: ^ Extra tab.

assets/src/helpers.js
```diff
@@ -31,3 +47,185 @@ export const maybeEnqueueFontStyle = ( name ) => {
...
+	};
+};
+
+/**
+ * Add page number label to page blocks
+ *
+ * @param {Object} BlockListBlock BlockListBlock element.
+ * @return {Function} Handler.
+ */
+export const disableBlockDropZone = createHigherOrderComponent(
```
miina: DropZone is still needed for other blocks.

westonruter:
In a future PR, work is needed on the reordering UI for mobile:

> image

> image

westonruter:
🐛 The Insert Page button doesn't do anything when I click it:

> image

I can, however, click the normal block inserter to add a new page.

westonruter:
I keep wanting to be able to click the previous page and next page to be able to navigate to them, but clicking them does nothing. I have to click on the page navigation in the bottom instead. Perhaps instead of:

```css
.editor-block-list__layout [data-type="amp/amp-story-page"].amp-page-inactive {
	opacity: .5;
	pointer-events: none;
}
```

It could do:

```css
.editor-block-list__layout [data-type="amp/amp-story-page"].amp-page-inactive {
	opacity: .5;
}
.editor-block-list__layout [data-type="amp/amp-story-page"].amp-page-inactive > div {
	pointer-events: none;
}
```

And then when another page is selected, it could automatically be made the active page?

assets/css/amp-editor-story-blocks.css
```diff
@@ -364,41 +530,114 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
...
+				var( --scrollbar-width )
+			) / 2
+			-
+			var(--block-navigation-right-margin)
+		);
 	}
 
 	.post-type-amp_story .edit-post-layout .editor-writing-flow > div > div {
-		grid-template-columns: auto minmax( 368px, calc( 100vw - 160px - 315px ) );
+		grid-template-columns: auto minmax( 368px, calc( 100vw - var(--admin-menu-expanded-sidebar-width) - 315px ) ); /* @todo Replace 368px and 315px with variables. */
```
westonruter: @swissspidy Would you please add variables for `368px` and `315px`? I wasn't sure what magic they referred to, as they aren't mentioned in the comment above. See other instances of these numbers below.

westonruter:
🐛 We should hide the `.amp-story-editor-carousel-navigation` when in the re-order mode:

image

westonruter:
- [ ] 🐛 There doesn't seem to be any way to insert new blocks other than pages presently. There is only the initial Text block that I can modify.

assets/src/blocks/amp-story/edit.js
swissspidy: Indeed unused and can be removed.

swissspidy:
Thank you both so far for your feedback!

Some quick notes:

> In a future PR, work is needed on the reordering UI for mobile
> Mobile version doesn’t seem to function properly.

Yep. I did not focus on mobile compatibility at this point. Needs some design thoughts, plus I want to improve the way we organize CSS before that anyway.

> Accessibility

Like mobile, I did not really focus  on accessibility yet, but it's an area that's dear to my heart. Something I can address in future PRs.

> The Insert Page button doesn't do anything when I click it

It's a no-op at the moment as I wanted to implement this as part of #1902 directly.

> Would be great to have some documentation for the custom Components

Totally, it's on my todo list.

> I keep wanting to be able to click the previous page and next page to be able to navigate to them, but clicking them does nothing.
> Clicking the next Page doesn’t seem to select it.

Somehow didn't occur to me, but makes sense. Will address.

assets/src/amp-story-editor-blocks.js
```diff
@@ -52,9 +63,67 @@ domReady( () => {
...
+	// Prevent WritingFlow component from focusing on last text field when clicking below the carousel.
+	document.querySelector( '.editor-writing-flow__click-redirect' ).remove();
 } );
 
+/**
+ * Add some additional elements needed to render our custom UI controls.
+ */
+function renderStoryComponents() {
+	const editorBlockList = document.querySelector( '.editor-block-list__layout' );
+	const postTitle = document.querySelector( '.editor-post-title' );
```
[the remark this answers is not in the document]
swissspidy: It's going to happen at some point, but there's been a lack of traction and not a clear decision yet on how it's gonna happen. Since we probably want to require titles for stories anyway, I don't think there's an immediate need to change this. Definitely can keep this in mind though.

miina:
> It's a no-op at the moment as I wanted to implement this as part of #1902 directly.

If the Inserter button is not functional right now and won't be within this PR then perhaps it would make sense to remove it from this PR and just implement it within #1902? It would be great if we would merge complete issues and leave out "known bugs/issues" from the main branch if possible, especially now when the branch will be on a live site for testing / trying out. Then it wouldn't create confusion and get mistaken for an unexpected bug. Thoughts?

swissspidy:
> There seems to be some empty area currently in the dropzones between two pages. The dropzones don’t seem to cover all the space between two pages when reordering. Perhaps increasing the width of the dropzones to make sure there is no "white" area could be good? 

FWIW, the same is the case for drag and drop in regular Gutenberg posts.

Currently, the drop position is indicated with blue borders. Should these now be in the middle between pages? Sounds a bit odd to me.

swissspidy:
> If the Inserter button is not functional right now and won't be within this PR then perhaps it would make sense to remove it from this PR

I'll just add it 👍

miina:
> FWIW, the same is the case for drag and drop in regular Gutenberg posts.
> Currently, the drop position is indicated with blue borders. Should these now be in the middle between pages? Sounds a bit odd to me.

It's just that I dragged a few times and it seemed it wasn't working at all and then I realized I had dragged exactly in the between of the two dropzones and that's why it wasn't working. The blue borders weren't so noticable, actually maybe they didn't show up at all since I was dragging exactly in the middle.

Just tested on a regular WP Post that the area between two blocks is very narrow and this small area seems to be completely covered -- in the AMP Stories case the Pages are not so close to each other and that leaves some uncovered room there, changing the width with CSS to expand the dropzone would work for example.

It does make sense to me that all the area between the pages would be covered, it doesn't feel like optimal user experience that if you aim to the middle of the two pages then it doesn't work as expected.

Are you thinking that it would look odd if the blue line would indicate the whole area in the middle? Thoughts?

swissspidy:
> Are you thinking that it would look odd if the blue line would indicate the whole area in the middle? 

I was more thinking that putting blue bars in the middle would look odd, but actually I don't think we would have to do that. I'll look into improving this 👍

assets/src/amp-story-editor-blocks.js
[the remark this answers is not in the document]
swissspidy: Fixed now in 7e37eedd5aad17598f808c79f1b9c64265982b03, as I had to move the component out of the post title container in order to make the inserter work properly.

assets/src/helpers.js
swissspidy: @miina The function only disables the drop zone when reordering, as can be seen further down. Will improve docs.

swissspidy:
> We should hide the .amp-story-editor-carousel-navigation when in the re-order mode

They're currently just disabled when reordering, but I agree that removing makes more sense.

swissspidy:
> There doesn't seem to be any way to insert new blocks other than pages presently. There is only the initial Text block that I can modify.

#1947 adds shortcuts. There is also the global inserter.

I also re-enabled the inserter within the page itself. However, we might want to remove it again once we have absolutely positioned elements. Otherwise it might look odd. But that's something for a future PR.

assets/css/amp-editor-story-blocks.css
swissspidy: They refer to the page individual sizes, but this was before my time. I don't think the grid here is needed anymore or even applies properly. Will see if I can remove it.

swissspidy:
@miina @westonruter Mind giving this another go?

Since it's quite a big PR, I'd like to merge it as soon as possible to make all the work that follows (e.g. #1953) easier to merge as well.

It's not perfect yet, but I can iterate on it and address issues in follow-up PRs.

westonruter:
I have verified every single line of code in this PR is flawless. 😉

miina:
Post-merge, but: LGTM! Thanks for adding the additional comments and also for splitting the reducer 👍 
Minor comment: There seems to be an empty file `assets/src/components/editor-carousel/drop-zone.js` committed.

Also, something to have in mind: looks like TwentyNineteen theme has applied their own editor styles and this (at least locally) causes the interface to break (the post in the screenshot include 3 Pages):
Screen Shot 2019-03-13 at 11 43 16 AM

miina:
Added an issue for the styling: https://github.com/ampproject/amp-wp/issues/1958.

PR #1925, opened 2019-03-05
Remove obsolete Jetpack and WPCOM files

### LINKED ISSUE #1824: Remove deprecated wpcom/shortcodes.php file
`wpcom/shortcodes.php` (containing `wpcom_amp_add_custom_embeds()`) was deprecated in 0.6.

We should consider just removing that file at some point, perhaps even in 1.1. I don't know the history of that WordPress.com / Jetpack integration though.

westonruter:
The `jetpack-helper.php` file was made obsolete in #1149 since the code was moved to Jetpack. We should be able to safely remove it now.

Also, remove `wpcom/shortcodes.php` since also deprecated/obsolete. Fixes #1824.

Lastly, per @gravityrail  in https://github.com/ampproject/amp-wp/issues/1021#issuecomment-470646607 we can also remove `wpcom-helper.php`.


swissspidy:
Related: #1824

PR #1929, opened 2019-03-05
Add first-party support for Crowdsignal poll/survey oEmbeds

westonruter:
Related: #1925.

* Promotes `WPCOM_AMP_Polldaddy_Embed` to `AMP_Crowdsignal_Embed_Handler`.
* Implements support for the Crowdsignal rebrand from PollDaddy.
* Removes shortcode handling from the AMP plugin since it is Jetpack-specific. See https://github.com/Automattic/jetpack/pull/11484 for where the code is moved to.

- [x] Ensure that https://github.com/Automattic/jetpack/pull/11484 has been released in a stable version.


kraftbj:
Confirming that since Crowdsignal/Polldaddy is a core oembed, those embeds will work in some decent way without Jetpack in AMP after this PR?

westonruter:
@kraftbj oh, I wasn't aware that it is a core embed now. Well, without support the support of the AMP plugin or Jetpack then the sample content from https://github.com/Automattic/jetpack/pull/11484 gets rendered as:

> image

In other words, the poll oEmbed successfully gets a fallback since it has `noscript` for the AMP plugin to unwrap. The `script` does cause a validation error which the AMP plugin catches:

image

The Survey oEmbed, however, renders nothing since the oEmbed response is _only_ a `script`: 

```html
<script type="text/javascript">(function(d,c,j){if(!document.getElementById(j)){var pd=d.createElement(c),s;pd.id=j;pd.src=('https:'==document.location.protocol)?'https://polldaddy.com/survey.js':'http://i0.poll.fm/survey.js';s=document.getElementsByTagName(c)[0];s.parentNode.insertBefore(pd,s);}}(document,'script','pd-embed'));</script>
```

image

This lack of `noscript` seems like a shortcoming of the Crowdsignal survey oEmbed response.

Nevertheless, now that Crowdsignal is actually directly supported by core, it seems actually that it shouldn't be removed. More specifically, the _oEmbed_ handling shouldn't be removed. The shortcode handling, however, is something that _should_ be handled rather by the Jetpack or Crowdsignal plugins since the shortcodes are _not_ defined in core. Agree?

tests/test-amp-crowdsignal-embed-handler.php
```diff
@@ -0,0 +1,106 @@
...
+				$poll_response,
+			),
+
+			/*
+			 * There seems to be a bug in WordPress core's oEmbed handling. The following patch seems necessary to
+			 * wp-includes/class-oembed.php:
+			 *  - '#https?://survey\.fm/.*#i'                    => array( 'https://api.crowdsignal.com/oembed', true ),
+			 *  + '#https?://(\w+\.)?survey\.fm/.*#i'            => array( 'https://api.crowdsignal.com/oembed', true ),
+			 */
+			'survey.fm'        => array(
```
westonruter: This test is causing the build failure. @kraftbj See my comment above as there seems to be a problem with Crowdsignal's support in core.

westonruter: https://core.trac.wordpress.org/ticket/46467

PR #1934, opened 2019-03-06
Prevent serving AMP for post embeds

westonruter:
I was testing post embeds from my native AMP site and I noticed them appearing unexpectedly in Gutenberg as:

> screen shot 2019-03-06 at 10 31 56

And when I accessed the post embed endpoint I saw it was unexpectedly an AMP page, and an [invalid one](https://gist.github.com/westonruter/503c6c8d359ce10994f306dab01c93e3) at that:

screen shot 2019-03-06 at 10 30 13

Even though it would actually be useful to allow post embeds to be AMP documents, since we could safely embed them via the AMP Shadow API, for now we should just preserve the normal behavior and fix embedding.

With the changes in this PR, the post embed in Gutenberg now looks as expected:

> screen shot 2019-03-06 at 10 32 10

And the individual embed endpoint is no longer an invalid AMP page, and it now renders properly:

> screen shot 2019-03-06 at 10 30 24

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -252,7 +252,7 @@ function post_supports_amp( $post ) {
 function is_amp_endpoint() {
 	global $pagenow, $wp_query;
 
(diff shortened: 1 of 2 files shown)


kienstra:
**Approved**

Hi @westonruter,
Good idea to prevent embeds from being AMP. Like you showed above, I also saw that this PR fixes it.

# Before
before-invalid-amp

# After
is-not-amp

PR #1937, opened 2019-03-07
[AMP Stories] Fix: restore amp-story-page ID attribute

miina:


miina:
`anchor` is used for animation order. I believe it's in use.

cc @swissspidy

westonruter:
Maybe animation order can just use `id` instead of `anchor`?

miina:
If I remember correctly then `anchor` is a general attribute name used for all the blocks, not just `amp-story-page`, however, probably @swissspidy can give more accurate information about `anchor` and its necessity for the `amp-story-page`.

swissspidy:
Basically anchor is something that already exists in Gutenberg for the ID attribute, which is why I decided to reuse that.

westonruter:
What about making the `amp-story-page` just indicate that the `anchor` prop is sourced to the `id` attribute? But wait, if Gutenberg is already using it for the `ID` attribute, then why is it not getting output? Do we just need to do:

```diff
- <amp-story-page style={ { backgroundColor: backgroundColor } } id={ id }>
+ <amp-story-page style={ { backgroundColor: backgroundColor } } id={ attributes.anchor }>
```

swissspidy:
Haven‘t tested, but yeah probably it‘s as simple as that

PR #1944, opened 2019-03-11
Enable AMP stories to be embedded in posts

### LINKED ISSUE #1931: Leverage WordPress post embeds for embedding AMP stories into posts
* Make sure Post Embed block can embed AMP Story post type. 
* Use featured image, title, and excerpt in preview.

(Issue copied from [this card](https://github.com/ampproject/amp-wp/projects/10#card-17824838))

kienstra:
* When adding a WordPress (embed) block and entering an AMP story URL, this will display a styled card
* Uses single card from the Latest Stories block carousel (#1916)

Closes #1931


kienstra:
**Current Display**

amp-story-embed

kienstra:
**Styling And Layout**

This could probably use styling and layout improvements. Also, it should probably have the sharing link that a post embed has:
amp-story-share

westonruter:
I don't think the sharing link is really that necessary. A user can just copy the URL of the link for the same effect, no? Normally a user would actually open the story and then share it from that point, I imagine.

kienstra:
>I don't think the sharing link is really that necessary. 
Sure, sounds good. Like you mentioned, someone could just copy the link.

tests/test-class-amp-story-post-type.php
```diff
@@ -40,6 +40,54 @@ public function test_the_single_story_card() {
...
+		AMP_Story_Post_Type::enqueue_embed_styling();
+		// None of the conditional is satisfied, so this should not enqueue the stylesheet.
+		$this->assertFalse( wp_style_is( AMP_Story_Post_Type::STORY_CARD_CSS_SLUG ) );
+
+		// Only the first part of the conditional is satisfied, so this again should not enqueue the stylesheet.
+		$GLOBALS['wp_query']->is_embed = true;
+		AMP_Story_Post_Type::enqueue_embed_styling();
+		$this->assertFalse( wp_style_is( AMP_Story_Post_Type::STORY_CARD_CSS_SLUG ) );
+
+		$GLOBALS['wp_query']->is_embed       = true;
```
swissspidy: Instead of manually setting the query here, I'd try to use `$this->go_to()`, which is a neat little helper method provided by the WordPress core test suite. I think it should work well for exactly this type of test.

includes/class-amp-story-post-type.php
```diff
@@ -646,4 +686,29 @@ public static function the_single_story_card( $post ) {
 		</a>
 		<?php
 	}
+
+	/**
+	 * Enqueues this post type's stylesheet for the embed endpoint.
+	 */
+	public static function enqueue_embed_styling() {
+		if ( is_embed() && is_singular( self::POST_TYPE_SLUG ) ) {
+			wp_enqueue_style(
```
westonruter: This style is being registered in two places. Can it be registered in one and then enqueued in two?

westonruter:
There are AMP validation errors related to the embed which I believe are out of scope of the changes here, but the link is showing up unexpectedly in the post editor:

image

The post embed looks good in the non-AMP page:

image

But it doesn't look right in AMP:

image

Again, this is largely due to poor post embed handling in WordPress: #809.

One idea to help deal with this is to add a `render_block` filter to then check if the URL being embedded is (1) on this site and (2) an AMP story post. If that is the case, then you could override the output to be `the_single_story_card` for that story post and avoid the iframe altogether. This would fix rendering of the post embed in AMP as well as eliminate the (known) validation errors.

includes/class-amp-story-post-type.php
kienstra: Good idea.

includes/class-amp-story-post-type.php
```diff
@@ -646,4 +686,29 @@ public static function the_single_story_card( $post ) {
...
+				self::STORY_CARD_CSS_SLUG,
+				amp_get_asset_url( '/css/' . self::STORY_CARD_CSS_SLUG . '.css' ),
+				array(),
+				AMP__VERSION
+			);
+
+			ob_start();
+			?>
+				.amp-story-embed {
```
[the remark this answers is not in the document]
kienstra: Yeah, it's a little strange to not include that in the `self::STORY_CARD_CSS_SLUG` file. It's only one style rule.

The reason was that It's specific to the AMP story embed, whereas the `self::STORY_CARD_CSS_SLUG` file applies to both the Latest Stories block (#1916) and this embed.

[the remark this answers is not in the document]
westonruter: But the one rule will get tree-shaken anyway, right? Seems fine to include it in the one CSS file then.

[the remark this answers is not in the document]
kienstra: Right, it'll be removed anyway.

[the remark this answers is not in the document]
kienstra: https://github.com/ampproject/amp-wp/pull/1944/commits/27107025dab7482f501fe41d7badedc715f3f28a moves this inline styling into the CSS file.

includes/class-amp-story-post-type.php
```diff
@@ -591,6 +604,33 @@ public static function render_text_block( $props, $content ) {
...
+	 */
+	public static function get_embed_template( $template, $type, $templates ) {
+		$old_amp_story_template = sprintf( 'embed-%s.php', self::POST_TYPE_SLUG );
+		$amp_story_embed_path   = AMP__DIR__ . '/includes/templates/embed-amp-story.php';
+		if (
+			'embed' === $type
+			&&
+			in_array( $old_amp_story_template, $templates, true )
+			&&
+			file_exists( $amp_story_embed_path )
```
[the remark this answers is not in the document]
kienstra: Thanks, https://github.com/ampproject/amp-wp/pull/1944/commits/8b0395fa9c1c20d2e81dcaa1b8351b5b771c1d4e uses your snippet above to simplify the method.

includes/class-amp-story-post-type.php
kienstra: https://github.com/ampproject/amp-wp/pull/1944/commits/a9b5fdbdfa72ef3d0d2b81bfc6645f61b9be57c1 registers the CSS file on the `wp_loaded` hook. Embed endpoints don't seem to fire `wp_enqueue_scripts`. 

It's not ideal, but `wp_loaded` was the best hook I could find that works for the embed and the Latest Stories block.

tests/test-class-amp-story-post-type.php
kienstra: Sure, https://github.com/ampproject/amp-wp/pull/1944/files#diff-694f9f17a2563b4b06a20b719fa41466R92 now uses `$this->go_to()` instead of changing `$GLOBALS['wp_query']`.

includes/class-amp-story-post-type.php
westonruter: @kienstra you can use the `wp_default_scripts` action instead, perhaps including the CSS registration in this function:

https://github.com/ampproject/amp-wp/blob/44a94ce43ac95ca853cc6502a113286cb7901a8b/includes/amp-helper-functions.php#L377-L450

kienstra:
**Embed Display In AMP**

Hi @westonruter,
Thanks for your suggestion for overriding the embed block's `render_callback` for AMP stories.

https://github.com/ampproject/amp-wp/pull/1944/commits/6a81d3c4f926d25529d030940dbd2695d78dfad1 does that, and an AMP story embed now looks better in AMP:

amp-story-embed-amp

And there aren't any more validation errors in the block editor.

kienstra:
Ah, I still need to ensure that the embed URL is from the same site.

kienstra:
**Steps To Test**

1. Create a new post (of the post type 'post')
2. Add a WordPress (embed) block:
wordpress-embed-block

3. Enter the permalink of an AMP Story post and click 'Embed'

4. Expected: the AMP Story is embedded:

embedded-amp-story

5. Expected: the front-end display:
amp-story-embed-front-end

swissspidy:
Should this also support embedding from another site in the same network?

westonruter:
> Should/Does this also support embedding from another site in the same network?

That would be nice but it wouldn't be supported presently as currently implemented via:

```php
$post = get_post( get_page_by_path( $path, OBJECT, self::POST_TYPE_SLUG ) );
```

Thinking about this some more, I wonder if overriding the `render_block` filter is actually the best approach. Perhaps actually what is needed is to override the actual HTML that is returned for oEmbed response (the `embed_html` filter and/or `oembed_response_data` filter)? If we did this, perhaps we could eliminate the need for the oEmbed iframe altogether? If the oEmbed response `html` just contained the contents of `the_single_story_card()` then that would be very cool. I am dubious, however, due to the styling required.

westonruter:
I added a few `@todo` comments.

kienstra:
Thanks, @westonruter! I'll carry out the `@todo` comments if you'd like.

westonruter:
Yep, go for it.

Something I'm surprised by actually with the WordPress embed is that there is no auto-complete for the story URL. If I create a link, I can just start typing the title of the post to get it suggested. But there is no similar facility for the WordPress Embed. This is out of scope for this PR, however. It's really something that should be supported by WordPress itself. So I've opened https://github.com/WordPress/gutenberg/issues/14393

kienstra:
**@todos applied**

Hi @westonruter,
Thanks for reviewing this again, and good idea to open https://github.com/WordPress/gutenberg/issues/14393.

The `@todos` that you added here are now applied.

includes/class-amp-story-post-type.php
```diff
@@ -222,6 +240,7 @@ public static function enqueue_block_editor_assets() {
 			return;
 		}
 
+		// @todo Move the following styles to register_styles method, and scripts to a new register_scripts method which runs at wp_default_scripts?
```
westonruter: What I meant here was that potentially there could be methods that run at `wp_default_scripts` and `wp_default_styles` which then do `$scripts->add()` and `$styles->add()` respectively. That would then leave the existing `enqueue_block_editor_assets` method to just `wp_enqueue_script($handle)` and `wp_enqueue_style($handle)`.

But this is not really necessary for any scripts and styles that are enqueued in multiple places.

includes/class-amp-story-post-type.php
```diff
@@ -646,4 +721,188 @@ public static function the_single_story_card( $post ) {
...
+		$url           = $block['attrs']['url'];
+		$url_host      = wp_parse_url( $url, PHP_URL_HOST );
+		$home_url_host = wp_parse_url( home_url(), PHP_URL_HOST );
+
+		// Exit if the URL isn't from this site.
+		if ( $url_host !== $home_url_host ) {
+			return $pre_render;
+		}
+
+		$path = str_replace( home_url( self::REWRITE_SLUG . '/' ), '', $url );
```
westonruter: This still doesn't seem very robust. What if the `$url` is an HTTP one, but the site is in HTTPS? Since you've already verified above that the host is the same, it seems this should rather do something like:

```suggestion
		$embed_url_path = wp_parse_url( $url, PHP_URL_PATH );
		$base_url_path  = wp_parse_url( trailingslashit( home_url( self::REWRITE_SLUG ) ), PHP_URL_PATH );
		if ( 0 !== strpos( $embed_url_path, $base_url_path ) ) {
			return $pre_render;
		}
		$path = substr( $embed_url_path, strlen( $base_url_path ) );
```

kienstra: Thanks, that's a good point. I [committed](https://github.com/ampproject/amp-wp/pull/1944/commits/80b3a7ea6558ec9b4c8a7bf71267846110a0d593) your code verbatim.

includes/class-amp-story-post-type.php
kienstra: Ah, that makes sense. Sorry for missing that.

It looks like you renamed some methods in https://github.com/ampproject/amp-wp/pull/1944/commits/80b3a7ea6558ec9b4c8a7bf71267846110a0d593. Let me know if you'd like me to apply your suggestion of calling `$scripts->add()` and `$styles->add` at `wp_default_scripts` and `wp_default_styles`

PR #1952, opened 2019-03-12
[AMP Stories] Update WordPress npm packages

swissspidy:
~Currently blocked by https://github.com/WordPress/gutenberg/issues/14381.~

~The following error happens:~

```
Module parse failed: Identifier 'createElement' has already been declared (2:9)
You may need an appropriate loader to handle this file type.
| import { createElement } from "@wordpress/element";
> import { createElement } from "@wordpress/element";
```


swissspidy:
Most of the package updates are just minor bumps for the latest Gutenberg release.

The Babel preset got a bit simplified, which is nice for us. Plus, the ESLint config got updated to have improved rules for inline docs. It also gets rid of some peer dependency warnings with the React ESLint configs.

PR #1953, opened 2019-03-12
[AMP Stories] Allow dragging and dropping elements within Page

### LINKED ISSUE #1897: Element positioning (drag & drop)
The goal is to allow a user to add a block to a page and then drag it wherever they want, including on top of other elements (e.g. Text block on top of an Image block)

- [x] Make all top-level elements absolutely positioned (except for background image and CTA); add draggability of elements
- [x] Automatically position newly added element when inserting a new block
- [ ] Add drag/drop reordering of elements in table of contents (contribute via Gutenberg)
- [x] Eliminate layer blocks
- [x] Ability to select and reorder elements (change z-index by virtue of document order (⬇️/⬆️ ), not changing the `z-index` CSS property). 
- [x] Also consider “bring to front” and “send to back” options. Also, consider whether user confusion for ⬇️ actually meaning to _raise_ an element in the stack, and ⬆️ to _lower_ an element in the stack.

miina:
Fixes #1897.
The goal is to allow a user to add a block to a page and then drag it wherever they want, including on top of other elements (e.g. Text block on top of an Image block)

- [x] Sets all the blocks to absolute positioning within a Page;
- Add an exception for CTA block (missing block) -- to be handled within #1963;
- [x] Add dragging and dropping within a Page;
- [x] Reflect the style in frontend;
- [x] Use "Move to back" and "Bring forward" instead of "Move Up" and "Move down";
- [x] Handle dragging blocks out of the Page limits;
- [x] Remove temporary code (such as commenting out `setBlockParent` since it's not working properly.
- [x] Position newly added element.

Other notes:
- This implementation is using already existing Gutenberg components, it replaces the original with custom ones for BlockMover (the 3 icons on each block which by default allow moving up, down, and dragging) and BlockDropZone.
- Some components that are necessary but not available via API are just copied over, such as `IgnoreNestedEvents`.
- See file comments for seeing which files are copied and which modified.
- Likely will be influenced by #1922.


assets/src/components/block-drop-zone.js
```diff
@@ -0,0 +1,89 @@
...
+import { compose } from '@wordpress/compose';
+
+/**
+ * Internal dependencies
+ */
+import {
+	MediaUploadCheck,
+} from '@wordpress/editor';
+
+class BlockDropZone extends Component {
```
miina: Would be better to rename it to something custom.

assets/src/amp-story-editor-blocks.js
```diff
@@ -217,7 +232,7 @@ const addAMPExtraProps = ( props, blockType, attributes ) => {
 const setBlockParent = ( props ) => {
 	const { name } = props;
 
-	if ( ! ALLOWED_CHILD_BLOCKS.includes( name ) ) {
+	/*if ( ! ALLOWED_CHILD_BLOCKS.includes( name ) ) {
```
miina: ^ This is temporarily commented out for being able to test normally (since adding blocks doesn't work as expected.

miina:
@westonruter Two questions:
1. I'm wondering how should we handle dragging blocks out of the page. Should we allow it?
For example:
Screen Shot 2019-03-12 at 6 30 57 PM

Here the text block is out of the borders, however, the text itself fits without issues.
In case of the image it does go over the edge and displays like this in the frontend:
Screen Shot 2019-03-12 at 6 31 11 PM

At this moment going outside of a Page is restricted from the top and left but bottom and right not (since the dropzone limits are set by the top left corner). I know that Google Slides for example does allow dragging elements out of the slide. Not sure what's the best way to go here, perhaps we could set it so that the top left corner has to be at least X pixels inside the Page but the rest can be outside. Not sure. Thoughts?

2. Another question: did you have anything in mind for the default values of a newly added element? Could it just be somewhere in the middle by default?

westonruter:
> 1. I'm wondering how should we handle dragging blocks out of the page. Should we allow it?

First of all, the text should really be sized the same in the editor as it will appear on the frontend. Otherwise it is very difficult to place it appropriately over another image.

Secondly, using pixel values for the `top`/`left` may be part of the problem here. To be suitable for responsive displays, I think it really needs to use percentages. I realize this may mean for a tricky conversion between pixel and percentage values when dragging.

And given that Google Slides and Instagram both allow dragging elements out of the viewport, I think we have to allow

> 2. Another question: did you have anything in mind for the default values of a newly added element? Could it just be somewhere in the middle by default?

What about getting the max `top` position of the other blocks in the page, and then adding another N pixels to that?

miina:
> First of all, the text should really be sized the same in the editor as it will appear on the frontend. Otherwise it is very difficult to place it appropriately over another image.

Making the editor and frontend view match is a general issue that might need addressing as a whole separate ticket. As far as I remember there were other issues, too, not, just the text size. I'll make the text match better within this PR, however, would be great to have a separate QA + fixes ticket for testing (and fixing) each allowed block.

> To be suitable for responsive displays, I think it really needs to use percentages.

Good point!

> What about getting the max top position of the other blocks in the page, and then adding another N pixels to that?

What if the max top position is already in the bottom?

westonruter:
> What if the max top position is already in the bottom?

I suppose cycle back to the top then. This is a common pattern for app windows. (Try opening 20 instances of TextEdit, for example.) Since the newly-added blocks will be at the end of the page, they will overlay any other previously-added blocks, so there won't be a concern about the newly-added block being hidden from view.

westonruter:
Merge conflicts!

assets/src/constants.js
```diff
@@ -27,6 +27,9 @@ import Slabo27 from '../images/font-names/slabo-27.svg';
 import SourceSansPro from '../images/font-names/source-sans-pro.svg';
 import Ubuntu from '../images/font-names/ubuntu.svg';
 
+export const STORY_PAGE_INNER_WIDTH = 328;
```
swissspidy: FWIW there are similar constants in the `EditorCarousel` component. We might be able to move them to this file and calculate them based on `STORY_PAGE_INNER_WIDTH`

These constants are used to calculate the `translateX` offset in the horizontal page navigation.

https://github.com/ampproject/amp-wp/blob/f5289bbd12be66ad8b53f39631e390004fd2d759/assets/src/components/editor-carousel/index.js#L16-L34

miina: Thank you for noting this. I noticed this, too, in case of positioning, the inner width is needed, it looks like in `EditorCarousel` it's together with the 5px border. I'll come back to this later, if you find anything else in the PR that needs attention then do let me know. 👍

swissspidy:
It seems that `addAMPExtraProps()` overrides any custom styles being added in the Text block's `save()` function. This clashes with #1962. Maybe it needs to check for `props.style` instead of `attributes.style` (or both)?

Also, I think the CSS changes did mess a bit with the block previews when reordering.

----

Besides that, in my limited testing it felt a bit odd to drag elements with the block mover. When I want to nicely center an element, I cannot "just move" my mouse to the middle, because the element would end up on the right of it. I therefore have to account for the block mover when dragging, and go a bit more to the left. This felt a bit counter intuitive, compared to just grabbing an element in the middle and drop it in the center of the page.

Hope that makes sense :-)

miina:
>Besides that, in my limited testing it felt a bit odd to drag elements with the block mover. When I want to nicely center an element, I cannot "just move" my mouse to the middle, because the element would end up on the right of it. I therefore have to account for the block mover when dragging, and go a bit more to the left. This felt a bit counter intuitive, compared to just grabbing an element in the middle and drop it in the center of the page.

Thanks for testing! Yes, it does make sense. It would probably feel more intuitive to just drag the whole element from anywhere but this would cause other problems, e.g. if we'd show the drag-drop icon everywhere then it would feel odd to edit the blocks (as you mentioned previously as well). Also, what if I'd like to select text within a text block (for marking italic for example). At this moment it still seems like for now using the block mover is the best option.

On the positioning -- the goal was to position the block exactly where it shows on `dragStop` -- not based on the cursor's position but on the actual block's position so that the user would visually see where it would be placed.

Maybe a solution could be to still have the block mover initially but hide it while we're dragging -- to make the cursor match the left ede of the block. Maybe this way it would be less odd? Thoughts?

assets/src/components/with-initial-positioning.js
```diff
@@ -0,0 +1,67 @@
...
+export default createHigherOrderComponent(
+	( BlockEdit ) => {
+		return withParentBlock( ( props ) => {
+			const { attributes, name, parentBlock } = props;
+
+			if ( -1 === ALLOWED_CHILD_BLOCKS.indexOf( name ) || ! parentBlock || 'amp/amp-story-page' !== parentBlock.name ) {
+				return <BlockEdit { ...props } />;
+			}
+
+			// If the positions are not set and if the block is selected.
```
miina: @swissspidy -- this is the logic for initial positioning of a newly added block. Let me know if it seems to make sense to you, I was investigating the options of doing it in a different filter and a better way to detect if a block is unmodified but there doesn't seem to be one. Thought at first of `blocks.getBlockAttributes` filter but that doesn't fire for new blocks.

swissspidy: Hmm we already have some logic in `amp-story-editor-blocks.js` to detect if there are any newly inserted pages. We can just extend this (e.g. using `getClientIdsWithDescendants()`) to check if there are _any_ newly inserted _blocks_.

If so, `updateBlockAttributes()` can be used to update the attributes of a newly inserted block. IMO that makes more sense than filtering `BlockEdit` and modifying attributes in it.

miina:
> Edit: Seems like a false alarm.
@swissspidy Made the change already, it did seem on testing that `props.style` wasn't considered properly so changed that to use `props.style` instead, however, we could look into it again when one of the PR-s is merged.

miina:
@swissspidy There is an issue with the font size not displaying accurately in the frontend (it's much smaller than it displays in the frontend), probably the AMP Story default style is causing this. Is this something you were already looking into within some other ticket?

swissspidy:
I haven't noticed _smaller_ fonts. 🤔 If anything, the font is bigger on the frontend here, because the default `h1` and `h2` styling is not yet reset to be the same as for `<p>`.

swissspidy:
Still fiddling with the page offsets in `EditorCarousel`. I'll need to double check the math there.

miina:
Would you mind sharing a screenshot of how it displays for you?

swissspidy:
_All screenshots done with Twenty Seventeen, Gutenberg 5.2, AMP in Native Mode_

Text Block, Default Font Size, Auto Text Type, Editor:

Normal-Auto-Editor

Text Block, Default Font Size, Auto Text Type, Frontend:

Normal-Auto-Frontend

Note: It's an `h1`, hence the size.

Text Block, Normal Font Size, Paragraph Text Type, Editor:

Normal-Paragraph-Frontend

Text Block, Normal Font Size, Paragraph Text Type, Frontend:

Normal-Paragraph-Editor

The font size on the frontend is indeed smaller here. That's because the AMP runtime sets the font size to 1rem for `h4, p`. We should be able to just override that in our CSS.

Text Block, Large Font Size, Paragraph Text Type, Editor:

Large-Paragraph-Frontend

Text Block, Large Font Size, Paragraph Text Type, Frontend:

Large-Paragraph-Editor

miina:
Thank you! I'll create a separate ticket for matching the editor and frontend styles, there are some other issues, too, such as the margin/padding that's added by AMP Story but isn't in the editor at this moment. That's how it displays for me.
Editor:
Screen Shot 2019-03-14 at 7 53 18 PM

Frontend:
Screen Shot 2019-03-14 at 7 52 52 PM

miina:
Added this issue for the editor vs frontend: https://github.com/ampproject/amp-wp/issues/1967

swissspidy:
I feel like blocks should have some offset from the left side initially, otherwise they're stuck to the page border:

Screenshot 2019-03-14 at 20 13 52

swissspidy:
The arrows in the block mover are odd: the top one points down and says "bring forward", the bottom one points up and says "Move to back". Screenshot:

arrow

Note: The tooltips could use some coherent naming as well, e.g. "Move forward" / "Move to back"

When dragging a text block, the blue top/bottom border of the page's dropzone are shown:

border

Speaking of which: I see that you had copy the BlockMover component from Gutenberg into the plugin. Can you perhaps document in a README in that folder what the differences are and which components perhaps are still the same?

miina:
> I feel like blocks should have some offset from the left side initially, otherwise they're stuck to the page border:

Hmm, that wasn't happening before, _looking into it_

>I see that you had copy the BlockMover component from Gutenberg into the plugin. Can you perhaps document in a README in that folder what the differences are and which components perhaps are still the same?

I've added file comments to each file to indicate which are copied and which not. Will add README, too.

> The arrows in the block mover are odd: the top one points down and says "bring forward", the bottom one points up and says "Move to back". Screenshot:

Ha. Yes, these are reversed at this moment. I guess it would be ideal to have custom icons which would reflect their goal better.

swissspidy:
Awesome work so far by the way :-)

miina:
@swissspidy Thanks for the notes, I've added the README and adjusted style / default values for the positioning a little bit. I tried to add some padding to the Page to make it match the editor better but this influenced the dropping position, too, so I'll just leave it for #1967 (at least for now).

miina:
Also, changed the labels to "Bring forward" and "Send Backward" -- same wording as Google Slides seem to have.

swissspidy:
Looks like it's flagging `bot@renovateapp.com`.

@miina Curios: Why the merge of the `develop` branch and not the `amp-stories-redux` branch into this one?

miina:
@swissspidy That was definitely not intentional, also, it's odd since I just checked the console history and I haven't merged `develop` into this PR.

miina:
Hmm, I think I can see what happened -- was planning to merge `amp-stories-redux` but instead run `git merge origin amp-story/1897-drag_drop`. So basically I tried to merge the PR branch to itself and this resulted in merging `develop` instead.

miina:
I suppose we could merge `develop` into `amp-stories-redux`.

swissspidy:
> I suppose we could merge develop into amp-stories-redux.

Done.

PR #1956, opened 2019-03-12
Fix conversion of Facebook embeds

### LINKED ISSUE #1259: Jetpack Facebook Page plugin not working with beta
**As a user, I should expect the AMP plugin, when in Native AMP mode, to convert HTML markup (including Jetpack's Facebook Page plugin widget) on render to compliant AMPHTML components.**

- [x] AC1: When both Jetpack and AMP are installed, activated and configured in Native AMP mode, the Facebook Page plugin (as activated in a widget area, screenshot below) should render in a non-error state. The AMP plugin should be able to distinguish between what embeds need an ` <amp-facebook>` or an ` <amp-facebook-page>` wrapper (and possibly `<amp-facebook-like>` and `<amp-facebook-comments>`) as described by @westonruter below.

---

I use the Jetpack Facebook Page Plugin in my sidebar.

When I activate AMP I get:
"This Facebook post is no longer available. It may have been removed or the privacy settings of the post may have changed."

### LINKED ISSUE #1955: Facebook Sanitizer Returning False Positives
The current implementation is taking any tag with a `data-href` attribute and a class name and will auto convert those to an `amp-facebook` tag.

We have a plugin that contains a div with a `data-href` tag and it has a class attribute. Because that passes the first check from AMP in the `get_embed_type()` method of `AMP_Facebook_Embed_Handler`, there is no way to get around this.

The current check is to see if there's a `fb-post` class, and then a `fb-video` class. Then it falls back to check for a video class a second time, and if that fails, returns `post`.

The only current solution is to change the `data-href` attribute to something else, but as that's a logical name for a data attribute, it may be affecting more than just us.

westonruter:
In #1955 it was reported that the Facebook embed sanitizer was erroneously processing any element that had a `data-href` attribute, as if it should be an `amp-facebook` for a post. We were not correctly checking for all of the required values of the class name.

This PR also adds support for the automatic conversion of:

* Facebook pages
* Facebook comments
* Facebook comment embeds
* Facebook likes

If there is fallback content (noted in a child element with the`fb-xfbml-parse-ignore` class) then this will be preserved as the `fallback` content for the AMP elements.

Lastly, it automatically removes the `fb-root` div and the Facebook Connect SDK, since irrelevant. This will prevent the tag-and-attribute sanitizer from complaining later.

Fixes #1955.
Fixes #1259.

Here is a build of this branch for testing: 
[amp.zip](https://github.com/ampproject/amp-wp/files/2958939/amp.zip) (v1.1-alpha-20190312T215731Z-e070d35b)

### DIFF
--- a/includes/embeds/class-amp-facebook-embed.php
+++ b/includes/embeds/class-amp-facebook-embed.php
@@ -127,6 +127,24 @@ class AMP_Facebook_Embed_Handler extends AMP_Base_Embed_Handler {
 				$this->create_amp_facebook_and_replace_node( $dom, $node, $embed_type );
 			}
 		}
+
+		/*
+		 * Remove the fb-root div and the Facebook Connect JS script since irrelevant.
+		 * <div id="fb-root"></div>
+		 * <script async defer crossorigin="anonymous" src="https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v3.2"></script>
+		 */
+		$fb_root = $dom->getElementById( 'fb-root' );
+		if ( $fb_root ) {
+			$xpath   = new DOMXPath( $dom );
+			$scripts = array();
+			foreach ( $xpath->query( '//script[ starts-with( @src, "https://connect.facebook.net" ) and contains( @src, "sdk.js" ) ]' ) as $script ) {
+				$scripts[] = $script;
+			}
+			foreach ( $scripts as $script ) {
+				$script->parentNode->removeChild( $script );
+			}
+			$fb_root->parentNode->removeChild( $fb_root );
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


Hoodlum-Z:
@westonruter do you want me to try the testing build to see if it solves my problem we are discussing here: 

https://wordpress.org/support/topic/amp-still-removing-facebook-components/#post-11390311

Thanks.

westonruter:
@Hoodlum-Z if you are using `amp-facebook` (and the like) components directly, then this PR won't change anything for you.

Hoodlum-Z:
@westonruter well I'm not going to disagree with you, but is it possible that it has indirectly affected it? In a good way, I mean. 

As you know I've had this little niggle for a while and when I installed the test version, I swapped back and forth numerous times and the code wasn't removed like it has been previously. So based on that it surely has affected it somehow?

PR #1960, opened 2019-03-13
Add ability to filter the normalized image URL used for dimension extraction

alexrah:
This filter could be used in case of multi-tier infrastructure when internal back-ends don't communicate through HTTPS while front-end servers do.

### DIFF
--- a/includes/utils/class-amp-image-dimension-extractor.php
+++ b/includes/utils/class-amp-image-dimension-extractor.php
@@ -109,7 +109,7 @@ class AMP_Image_Dimension_Extractor {
 			$url = $home . $path;
 		}
 
-		$url = apply_filters('amp_normalize_url',$url);
+		$url = apply_filters( 'amp_normalize_url', $url );
 
 		return $url;
 	}
--- a/includes/utils/class-amp-image-dimension-extractor.php
+++ b/includes/utils/class-amp-image-dimension-extractor.php
@@ -109,6 +109,8 @@ class AMP_Image_Dimension_Extractor {
 			$url = $home . $path;
 		}
 
+		$url = apply_filters('amp_normalize_url',$url);
+
 		return $url;
 	}
 
--- a/includes/utils/class-amp-image-dimension-extractor.php
+++ b/includes/utils/class-amp-image-dimension-extractor.php
@@ -120,6 +120,7 @@ class AMP_Image_Dimension_Extractor {
 		 * @param string $url            Original image URL.
 		 */
 		$normalized_url = apply_filters( 'amp_normalized_dimension_extractor_image_url', $normalized_url, $url );
+
 		return $normalized_url;
 	}
 
--- a/includes/utils/class-amp-image-dimension-extractor.php
+++ b/includes/utils/class-amp-image-dimension-extractor.php
@@ -88,30 +88,39 @@ class AMP_Image_Dimension_Extractor {
 			return false;
 		}
 
+		$normalized_url = $url;
+
 		if ( 0 === strpos( $url, '//' ) ) {
-			return set_url_scheme( $url, 'http' );
+			$normalized_url = set_url_scheme( $url, 'http' );
+		} else {
+			$parsed = wp_parse_url( $url );
+			if ( ! isset( $parsed['host'] ) ) {
+				$path = '';
+				if ( isset( $parsed['path'] ) ) {
+					$path .= $parsed['path'];
+				}
+				if ( isset( $parsed['query'] ) ) {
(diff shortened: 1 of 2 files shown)


westonruter:
@alexrah would you please provide an example showing this filter in action?

alexrah:
@westonruter here you can find an implementation:

`function custom_amp_normalize_url($url){
	$parsed = wp_parse_url( $url );
	if($parsed['host'] === 'www.example.com'){

		return set_url_scheme( $url, 'http' );

	}

	return $url;
}

function custom_amp_normalize_url_init(){
	add_filter('amp_normalize_url','edinet_amp_normalize_url',10,1);
}`

in our multi-tier infrastucture , servers exposed to the web communicate in https, while private server (where php is running) are in http; in this situation amp is unable to extract the correct image's dimensions from https urls while can perfectly works when the passed urls are in http;

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -109,6 +109,8 @@ public static function normalize_url( $url ) {
 			$url = $home . $path;
 		}
 
+		$url = apply_filters( 'amp_normalize_url', $url );
```
westonruter: Let's rename `amp_normalize_url` as `amp_normalized_dimension_extractor_image_url`. That will make it more explicit (and verbose 😄).

westonruter: I think the original URL should also be included in case the default normalization is wrong, no?

Also there seems to be a problem where this existing code could be quite wrong in other environments where HTTPS is required:

https://github.com/ampproject/amp-wp/blob/966771907ce3016c3512e2e729bd374ff35c4899/includes/utils/class-amp-image-dimension-extractor.php#L91-L93

Bringing it all together:

```php
/**
 * Normalizes the given URL.
 *
 * This method ensures the URL has a scheme and, if relative, is prepended the WordPress site URL.
 *
 * @param string $url URL to normalize.
 * @return string Normalized URL.
 */
public static function normalize_url( $url ) {
	if ( empty( $url ) ) {
		return false;
	}

	if ( 0 === strpos( $url, 'data:' ) ) {
		return false;
	}

	$normalized_url = $url;

	if ( 0 === strpos( $url, '//' ) ) {
		$normalized_url = set_url_scheme( $url, 'http' );
	} else {
		$parsed = wp_parse_url( $url );
		if ( ! isset( $parsed['host'] ) ) {
			$path = '';
			if ( isset( $parsed['path'] ) ) {
				$path .= $parsed['path'];
			}
			if ( isset( $parsed['query'] ) ) {
				$path .= '?' . $parsed['query'];
			}
			$home      = home_url();
			$home_path = wp_parse_url( $home, PHP_URL_PATH );
			if ( ! empty( $home_path ) ) {
				$home = substr( $home, 0, - strlen( $home_path ) );
			}
			$normalized_url = $home . $path;
		}
	}

	/**
	 * Apply filters on the normalized image URL for dimension extraction.
	 *
	 * @since 1.1
	 *
	 * @param string $normalized_url Normalized image URL.
	 * @param string $url            Original image URL.
	 */
	$normalized_url = apply_filters( 'amp_normalized_dimension_extractor_image_url', $normalized_url, $url );
	return $normalized_url;
}
```

includes/utils/class-amp-image-dimension-extractor.php
```diff
@@ -109,6 +109,8 @@ public static function normalize_url( $url ) {
 			$url = $home . $path;
 		}
```
westonruter: Add a phpdoc comment here that explains what the filter does. For example:

```php
/**
 * Apply filters on the normalized image URL for dimension extraction.
 *
 * @since 1.1
 *
 * @param string $normalized_url Normalized image URL.
 * @param string $url            Original image URL.
 */
```

alexrah:
Hi, thanks for all the feedback; just wondering, I have to apply all mods you suggested to a new commit by myself or are you  gonna do it? Thanks

alexrah:
Hi, I don't understand what went wrong but it seems I'm missing some of your requested changes

westonruter:
> it seems I'm missing some of your requested changes

Which changes are missing?

alexrah:
I think nothing is missing, but I see the message "_Merging is blocked: Merging can be performed automatically once the requested changes are addressed._"
So I assumed some of your requests wasn't correctly addressed.
By reviewing my last commits can you confirm everything is ok?

westonruter:
> So I assumed some of your requests wasn't correctly addressed.

No, it's just that I haven't reviewed the changes yet.

Thanks!

PR #1962, opened 2019-03-13
[AMP Stories] Add color settings to text block

### LINKED ISSUE #1961: Text Block Color Options
- [x] Allow choosing text and background color for custom text block.

Previously: #1883

swissspidy:
Note that this PR includes the `classnames` npm package for easier handling of class names. We can use it in other components too. Unfortunately, Gutenberg/WordPress does not expose this globally (see https://github.com/WordPress/gutenberg/issues/8078), which is why we have to bundle it on our own. It's only ~1KB big though, so not that big of a deal.

**To do:**

* [x] If no background color is set, derive from underlying element (could be an image or the page's background media), perhaps using `window.getComputedStyle()`
* [x] Add frontend styling for color CSS classes (e.g. `has-light-green-cyan-color`)

Fixes #1961.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,6 @@
     "@wordpress/i18n": "3.2.0",
     "@wordpress/plugins": "2.1.0",
     "@wordpress/wordcount": "2.1.0",
-    "classnames": "^2.2.6"
+    "classnames": "2.2.6"
   }
 }
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@wordpress/hooks": "2.1.0",
     "@wordpress/i18n": "3.2.0",
     "@wordpress/plugins": "2.1.0",
-    "@wordpress/wordcount": "2.1.0"
+    "@wordpress/wordcount": "2.1.0",
+    "classnames": "^2.2.6"
   }
 }
--- a/assets/css/amp-stories.css
+++ b/assets/css/amp-stories.css
@@ -1,1 +1,9 @@
 amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
+
+.has-background {
+	padding: 10px !important;
+}
+
+amp-story-grid-layer > * {
+	width: auto !important;
+}
--- a/assets/src/blocks/amp-story-text/edit.js
+++ b/assets/src/blocks/amp-story-text/edit.js
@@ -8,7 +8,7 @@ import classnames from 'classnames';
  */
 import { __ } from '@wordpress/i18n';
 import { Fragment } from '@wordpress/element';
-import { PanelBody, SelectControl } from '@wordpress/components';
+import { PanelBody, SelectControl, withFallbackStyles } from '@wordpress/components';
 import { compose } from '@wordpress/compose';
 import {
 	RichText,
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -60,6 +60,7 @@
     "@wordpress/hooks": "2.1.0",
     "@wordpress/i18n": "3.2.0",
     "@wordpress/plugins": "2.1.0",
-    "@wordpress/wordcount": "2.1.0"
+    "@wordpress/wordcount": "2.1.0",
+    "classnames": "^2.2.6"
```
westonruter: This should have the corresponding update to `package-lock.json`, yeah>? Also, it should be `"2.2.6"` exactly, not `"^2.2.6"`

swissspidy: Hmm, I thought I committed the lock file as well.

Also, you're right with the version number. By the way while we're at: is the reasoning behind this written down somewhere?

westonruter: The reason is just that Renovate will handle updating the version numbers for us. They have good docs (which I just found) about why to do it: https://renovatebot.com/docs/dependency-pinning/#why-pin-dependencies

swissspidy: Looks like the lock file was unaffected because it was already a dependency of all the WordPress packages anyway. Nothing to do there.

swissspidy:
> If no background color is set, derive from underlying element

FWIW, the fallback is only used for the contrast checker. Because the underlying color might be different in the editor than on the actual site, using `getComputedStyle()` could be inaccurate. Still better than nothing though.

swissspidy:
I wonder if we should have nicer multi-line padded text when choosing a background color: https://css-tricks.com/multi-line-padded-text/

westonruter:
The color options are incorrectly being limited by theme options:

image

Related to this, there is no color that appears when previewing because the theme's color style rules aren't being enqueued:

image

swissspidy:
> The color options are incorrectly being limited by theme options:

That is already the case for the page background color setting as well I think.

Should we try to always use the block editor [defaults](https://github.com/WordPress/gutenberg/blob/b421fc9574a7f437fb5068f8b6100db8f2926956/packages/block-editor/src/store/defaults.js) (until we add the Material color palette)?

westonruter:
Yes, we should use the defaults I think. The theme's styles shouldn't apply in the story, so the `color` should be set specifically on the elements and not use classes.

westonruter:
As expected, if I comment-out the `editor-color-palette` theme support, I get the expected default colors:

image

And it appears on the frontend (in all its glory):

image

westonruter:
For `editor-color-palette` should we just `remove_theme_support('editor-color-palette')` when editing an Story or viewing a story on the frontend?

swissspidy:
`get_theme_support( 'editor-color-palette' ) )` is only really used in the admin, shortly before `enqueue_block_editor_assets` is called.

I think we can simply filter `block_editor_settings` to remove the color palette.

On the frontend no other change should be needed, as we don't enqueue theme styles there.

westonruter:
> On the frontend no other change should be needed, as we don't enqueue theme styles there.

Good point. But we do want to allow themes to enqueue a stylesheet in the future. But we can cross that bridge when we get to:

https://github.com/ampproject/amp-wp/blob/d6cd963a617e90a125ecc4b1e4c14205586b21c0/includes/templates/single-amp_story.php#L19

PR #1970, opened 2019-03-15
Add Slide Duration Options

### LINKED ISSUE #1964: [AMP Stories] Autoplay/Slide Duration
**To do:**

* [x] Add field to the AMP Story Page block for the duration of the slide. Allow auto duration to inherit from video/audio that is in the story, as well as inheriting from sum of all elements' animations.

**Reference:**

https://www.ampproject.org/docs/reference/components/amp-story#pages:-amp-story-page

This is handled by the `auto-advance-after` attribute:

> Specifies when to auto-advance to the next page. If omitted, the page will not automatically advance. The value for auto-advance-after must be either:
>
> * A positive amount of time to wait before automatically advancing to the next page
> * An ID of an `HTMLMediaElement` or video-interface video whose completion will trigger the auto-advance

Example:

```html
<amp-story-page id="tokyo" auto-advance-after="1s">
```

swissspidy:
**What this PR does**

This PR adds a new "Advance to next page" option to the Page Settings in the block sidebar:

Screenshot 2019-03-15 at 16 19 33

Users can choose between four options:

* Manually (like today)
* Automatic
* After a certain time
* After video/audio has ended

When choosing the "Automatic" option, the total duration time of all elements on the page is calculated and rounded up to the nearest second.

When choosing "After a certain time", a range control is added to set the time:

Screenshot 2019-03-15 at 16 19 40

The handling for "After video/audio has ended" is still TBD.

**How it works**

When using "Automatic", the total duration time has to be recalculated every time the animations change (using `getTotalAnimationDuration()`). To support this, I extended the data store to include all this data as well.

This way, the `store.subscribe()` callback can listen for changes to the animation order and update the block attributes accordingly.

Similarly, the block attributes are initially scanned in `domReady` to populate the data store.

**Todo**

* [x] When choosing the 'time' option, set minimum time so that it can't be below total animation duration time
* [x] Investigate why the animation attributes aren't persisted
* [x] Investigate why sometimes block validation fails
* [x] When choosing the 'video/audio' option, automatically find video on page to set attribute accordingly.
* [x] Hide 'video/audio' option when there are no video/audio blocks.

Fixes #1964.


swissspidy:
> When choosing the 'time' option, set minimum time so that it can't be below total animation duration time

Screenshot 2019-03-15 at 16 12 36

swissspidy:
Add media option now:

Screenshot 2019-03-15 at 17 53 28

Seems to work fine, although inserting a video in an AMP story page results in a validation error somehow.

assets/src/blocks/amp-story/edit.js
```diff
@@ -186,6 +208,31 @@ class EditPage extends Component {
...
+							} }
+						/>
+						{ 'time' === autoAdvanceAfter && (
+							<RangeControl
+								label={ __( 'Time in seconds', 'amp' ) }
+								value={ autoAdvanceAfterDuration ? parseInt( autoAdvanceAfterDuration ) : 0 }
+								onChange={ ( value ) => setAttributes( { autoAdvanceAfterDuration: value } ) }
+								min={ totalAnimationDuration }
+								initialPosition={ totalAnimationDuration }
+								help={ totalAnimationDuration > 1 ? __( 'A minimum time is enforced because there are animated blocks on this page', 'amp' ) : undefined }
```
miina: For some reason enforcing a minimum time doesn't seem to work for me.
Created a new AMP Story, added Text block and Image block, both with animations, without delay and both starting Immediately, then went to Page options and chose "After a certain time", it still displayed 0. Could you confirm that it's working as expected for you?

assets/src/stores/amp-story/reducer.js
```diff
@@ -50,14 +53,42 @@ export function animationOrder( state = {}, action ) {
...
+			} else {
+				pageAnimationOrder.push( { id: item, animationType } );
+			}
+
+			return {
+				...newAnimationOrder,
+				[ page ]: pageAnimationOrder,
+			};
+
+		case 'CHANGE_ANIMATION_DURATION':
```
miina: If the duration or delay changes, would the Advance time of the relevant Page also change in case of `time` setting and in case the new values exceed the chosen value? E.g. if the user has previously chosen to advance in 1 sec but changes the animation settings so that the total animation time reaches 3 sec, would the lower value 1 sec automatically change?

swissspidy: Not currently, but I guess it makes sense

assets/src/blocks/amp-story/edit.js
swissspidy: Not at the computer anymore, but will check this again.

assets/src/stores/amp-story/reducer.js
swissspidy: Done in f25b433ad43b094f732c122d96d943e35d7fee9a

assets/src/blocks/amp-story/edit.js
swissspidy: I think I just fixed this with f25b433ad43b094f732c122d96d943e35d7fee9a as well.

swissspidy:
One thing that is not implemented now:

When choosing the "Media" option and removing the Media block on that page, the page's attribute containing that block's ID isn't updated.

To fix this, we'd need to keep a list of all media blocks on a page in our data store and update the attribute whenever the list changes. Not urgent, but it also wouldn't take long to implement I think.

swissspidy:
> When choosing the "Media" option and removing the Media block on that page, the page's attribute containing that block's ID isn't updated.

Fixed now in c9860274c9a1c2ba368af8992faf9fc218531441

swissspidy:
@miina Mind having another look at this one as well?

miina:
I'm getting the following error when trying to use Image block with animations (the editor crashes):
Screen Shot 2019-03-20 at 8 43 56 AM
Screen Shot 2019-03-20 at 8 43 33 AM

Also seeing this, not sure if it's related to this PR but I didn't see it on `amp-stories-redux` when trying animations:
Screen Shot 2019-03-20 at 8 43 18 AM

Other things noticed:
- The enforced minimum time doesn't always seem to consider the default animation value. Steps to reproduce:
1. Add new Page, select the Text block;
2. Choose "Drop" as animation with default settings, without choosing anything;
3. Go to Page's "Advance to next page" and choose "After a certain time".

- When choosing "media" for the advancing option and then removing a Media and there is no Media left the "Media" option is removed from the selection dropdown, however, the description seems to remain:
Screen Shot 2019-03-20 at 9 01 11 AM

- When leaving the value "0" for "After a certain time" then this displays:
Screen Shot 2019-03-20 at 9 09 40 AM
Probably we should have minimum seconds assigned for this value?

swissspidy:
Thanks @miina! Unfortunately I was unable to reproduce the console errors, however I spotted some things here and there in the code that could be improved / hardened, so these might not occur anymore now.

I also addressed the other things you noticed.

assets/src/blocks/amp-story/edit.js
```diff
@@ -186,6 +210,25 @@ class EditPage extends Component {
...
+							value={ autoAdvanceAfter }
+							options={ autoAdvanceAfterOptions }
+							onChange={ ( value ) => setAttributes( { autoAdvanceAfter: value } ) }
+						/>
+						{ 'time' === autoAdvanceAfter && (
+							<RangeControl
+								label={ __( 'Time in seconds', 'amp' ) }
+								value={ autoAdvanceAfterDuration ? parseInt( autoAdvanceAfterDuration ) : 0 }
+								onChange={ ( value ) => setAttributes( { autoAdvanceAfterDuration: value } ) }
+								min={ totalAnimationDuration }
```
miina: Maybe we should never allow 0 at all?

assets/src/blocks/amp-story/edit.js
```diff
@@ -186,6 +210,25 @@ class EditPage extends Component {
 							) }
 						</Fragment>
 					</PanelBody>
+					<PanelBody title={ __( 'Page Settings', 'amp' ) }>
+						<SelectControl
+							label={ __( 'Advance to next page', 'amp' ) }
+							help={ 'media' === autoAdvanceAfter ? __( 'Based on the first media block encountered on the page', 'amp' ) : undefined }
+							value={ autoAdvanceAfter }
+							options={ autoAdvanceAfterOptions }
+							onChange={ ( value ) => setAttributes( { autoAdvanceAfter: value } ) }
```
miina: Should changing from 'time' to 'auto' trigger adjusting the `autoAdvanceAfterDuration` value? If I understand correctly then "Auto" would use automatically generated value based on animations, looks like `time` and `auto` are sharing `autoAdvanceAfterDuration` for `save`, currently it looks like if switching from `auto` the previous `time` value is still used.

assets/src/blocks/amp-story/index.js
```diff
@@ -62,10 +72,27 @@ export const settings = {
...
+			mediaType,
+			poster,
+			autoAdvanceAfter,
+			autoAdvanceAfterDuration,
+			autoAdvanceAfterMedia,
+		} = attributes;
+
+		let advanceAfter;
+
+		if ( [ 'auto', 'time' ].includes( autoAdvanceAfter ) && autoAdvanceAfterDuration ) {
```
miina: Wondering if `auto` should consider Media as well?

miina:
Another note for a potential future iteration: after testing it out a little bit it started feeling it would make sense to have this setting generally for the whole story as well which can then be overridden per Page. It's likely that the user would want to set a general rule for all pages, especially `auto` or `media`. Thoughts?

swissspidy:
> Another note for a potential future iteration: after testing it out a little bit it started feeling it would make sense to have this setting generally for the whole story as well which can then be overridden per Page. 

Yeah that makes sense. This option could live in the Document Sidebar. Unfortunately it is currently not easily possible to add panels there using JS only.

It would make sense to be able to set this globally somehow so that new stories are automatically configured according to the user's preference.

assets/src/blocks/amp-story/edit.js
swissspidy: > currently it looks like if switching from auto the previous time value is still used.

Indeed. It should be trivial to fix, as we already have the total animation time available here. Will add this.

miina:
Added a follow-up issue for the global/per story setting to have it somewhere: https://github.com/ampproject/amp-wp/issues/1995.

assets/src/blocks/amp-story/index.js
swissspidy: I hesitated a bit to add this because I wasn't sure if it really made sense or not. It would be a bit limited as only non-looped video/audio blocks with autoplay could be taken into account.

At that point, I wonder whether the `media` option is even needed, as `auto` would be superior.

miina: Alternatively, we could leave the `auto` for animations only and add a description which would specify what it does (move to next slide once the animations have finished)?

> At that point, I wonder whether the media option is even needed, as auto would be superior.

It could happen in theory that the animations take longer than Media and then `media` could still have a different result being separate (if `auto` doesn't include Media).

swissspidy: New issue: #1999

Issue #1967, opened 2019-03-14
[AMP Stories] Make sure that the frontend and editor view match


miina:
AC: All the allowed blocks in the editor look visually the same (or very similar) to how they look like in the frontend. This does _not_ include animations.

- [x] Make sure that all the text containing blocks' font sizes in the editor match the sizes in frontend;
- [x] Make sure that the margins and paddings in the editor match the frontend. For example, the AMP Story Grid Layer seems to have padding but this is not reflected in the editor;
- [x] Make sure that the Text block displays similarly in the editor as in the frontend. At this moment a Text block might display 2 rows in the editor but just one in the frontend.
- [x] Verify other blocks not mentioned here separately.
- [x] Make sure the ratio of the Page is correct (currently per Weston's comment Frontend story is: 390.141/650.250 = 0.59998615917, however, Editor is: 338/553 = 0.61121157323)
- [x] Add basic editor style (e.g. max-width).

westonruter:
See also discussion about how we should include styles on the frontend: https://github.com/ampproject/amp-wp/issues/1958#issuecomment-473082141

swissspidy:
> Make sure the ratio of the Page is correct

Note that this also affects the reordering UI.

miina:
I'm wondering what would be the best approach for font sizes -- for example if the default font size is 16px in the editor and is displayed in two rows then the same font size in the frontend might still be in one row -- since the font size is not responsive. We could add some different sizes based on the screen width for the default font sizes, however, what if the user selects a specific font size? 

@swissspidy @westonruter Thoughts?

westonruter:
@miina What if the specific font sizes were stored in `vh` units instead of `px` units? That would make them responsive. They'd perhaps have to remain `px` in the editor since the the viewport is not fully contained by the story, but when rendered on the frontend they could be replaced with some computed equivalent using `vh`. This would make the font size scale based on the height of the viewport, and thus it would be responsive.

miina:
@westonruter: Just updated the PR to use `vw` units instead of `px` for the font size in the frontend, the editor still displays in `px`.

miina:
Closed within #2015.

csossi:
verified in QA

PR #1972, opened 2019-03-15
[AMP Stories] Ability to resize text block.

### LINKED ISSUE #1965: [AMP Stories] Text element sizing
AC: User can change the width and height of the Text Block by using grab handles and this is also accurately reflected in the frontend.

- [x] Add grab handles for changing the block's width/height
- [x] Eliminate amp-fit-text or integrate with block sizing

miina:
Fixes #1965.

- [x] Add resizable box around Text block.
- [x] Allow resizing and make sure it reflects on the frontend, too.
- [x] Verify and test reasonable default values.
- [x] Use percentage values for resizing width / height.
- [x] Use amp-fit-size on the frontend to match the font responsively.


assets/css/amp-stories.css
```diff
@@ -3,7 +3,3 @@ amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
 .has-background {
 	padding: 10px !important;
 }
-
-amp-story-grid-layer > * {
-	width: auto !important;
```
miina: @swissspidy I noticed that this CSS condition changes the frontend view and overrides some values needed with absolute positioning.

Would removing this break something that you worked on?

assets/src/blocks/amp-story-text/edit.js
```diff
@@ -116,26 +123,59 @@ function TextBlock( props ) {
...
+				minWidth={ minTextWidth }
+				enable={ {
+					top: true,
+					right: true,
+					bottom: true,
+					left: true,
+					topRight: false,
+					bottomRight: false,
+					bottomLeft: false,
+					topLeft: false,
```
miina: Not sure how many "dots" we should create for the Text blocks for resizing. Perhaps just each side? Or perhaps just the corners?

assets/css/amp-stories.css
swissspidy: Yeah I added it because of the background color options. The AMP runtime forces a 100% width for elements, making text blocks with a custom background color look odd.

I guess that won't be a problem anymore with this PR here.

assets/src/blocks/amp-story-text/edit.js
swissspidy: Let's test with each side (Gutenberg default I think?) and then test with corners if it doesn't feel right.

miina: Sounds good!

miina:
@swissspidy / @westonruter At this moment I ended up adding only the right and the bottom resizers -- otherwise the resizer should change the positionTop and positionBottom percentages as well to not make it move weirdly: if resizing from the left edge would just change the width while the positionLeft would remain the same, then actually the right edge would move for the user instead of the left one.

Looks like the Image block has handled for right align only (the left resizer only shows up if the image is aligned right and the right one is hidden in this case). We could follow the same pattern in case we'd want to add align to the Text block. Thoughts?

Let me know if you think it would be critical to add the resizers to all edges or if left and bottom would be OK for now.

swissspidy:
> if resizing from the left edge would just change the width while the positionLeft would remain the same, then actually the right edge would move for the user instead of the left one.

I guess in this case positionLeft _and_ the width would need to be changed.

> Let me know if you think it would be critical to add the resizers to all edges or if left and bottom would be OK for now.

I think the current status works for now.

Three things I noticed in a quick test:

**Odd initial size**

The initial width is too small, making the _Write something!_ placeholder span over two lines:

Screenshot 2019-03-18 at 10 56 13
Screenshot 2019-03-18 at 10 56 22

**Block Mover and Resize Handles always visible**

Screenshot 2019-03-18 at 10 39 23

As a user I don't want to see these controls while typing. This works quite well in the post editor: all controls disappear while typing, and the block mover only appears when hovering near the left side of the element.

**Text overflows container**

Screenshot 2019-03-18 at 10 42 11

It would be awesome if we could use `amp-fit-text` here. At the moment it's not possible to use individual AMP components on non-AMP sites though, see https://github.com/ampproject/amphtml/issues/15583

However, the code for `amp-fit-text` is rather simple: https://github.com/ampproject/amphtml/blob/e7a1b3ff97645ec0ec482192205134bd0735943c/extensions/amp-fit-text/0.1/amp-fit-text.js

I think we can easily adapt this script (mainly the `calculateFontSize_` function) for use in the text block.

In the post editor the plugin already adds a "Automatically fit text to container" control to enable this on the front end, but we might want to make this smarter for Stories.

Perhaps we can always use amp-fit-text, no matter what, or add a new "Auto" option to the "Font Size" control, which would turn on amp-fit-text.

miina:
Thanks for the comments and testing!

> I guess in this case positionLeft and the width would need to be changed.

Correct, this should be quite straightforward for the `onResizeStop`, however, it would also need doing that while reisizing.

assets/src/blocks/amp-story-text/edit.js
```diff
@@ -40,108 +43,190 @@ const applyFallbackStyles = withFallbackStyles( ( node, ownProps ) => {
...
-		setFontSize,
-		backgroundColor,
-		textColor,
-		setBackgroundColor,
-		setTextColor,
-		fallbackTextColor,
-		fallbackBackgroundColor,
-	} = props;
+class TextBlockEdit extends Component {
+	componentDidUpdate() {
```
miina: @swissspidy I've added the amp-fit-text's replication logic here. Note that I didn't make any changes to `fontSizePicker`, so I renamed the PR back to WIP.

Testing this it feels like maybe it would be good if the user could still assign a custom font size, too. Perhaps we could even have just only two values for this: Auto (amp-fit-text logic) and Custom. Thoughts?

Also, let me know if you know of a better place for adding the logic, note that it needs to check the element's changing width and height to calculate the font size so the element should be rendered.

miina:
Note that this PR has been developed on the latest `develop` branch of Gutenberg as of today (18.03) which seems to include some styling changes compared to the latest release.

swissspidy:
Regarding 5d242587e0f2b5056ae9f5f251f2b3dadf0d1868, can we use the same [`HoverArea`](https://github.com/WordPress/gutenberg/blob/7e5f1d8400a506245828e0cb415e87986dc29481/packages/block-editor/src/components/block-list/hover-area.js) component that the [`BlockListBlock`](https://github.com/WordPress/gutenberg/blob/7e5f1d8400a506245828e0cb415e87986dc29481/packages/block-editor/src/components/block-list/block.js) one uses? Perhaps we need to copy the code as it doesn't seem to be exported. This way we could only show the mover when hovering close to the left side of the block (or right side for RTL languages).

> Note that this PR has been developed on the latest develop branch of Gutenberg as of today (18.03) which seems to include some styling changes compared to the latest release.

Good to know! The next Gutenberg version (5.3) is scheduled to be released on Wednesday, so perhaps these changes are going to be included there. There are some other nice updates in there too, so I'll have a look at updating the dependencies as soon as new packages are being released.

assets/src/blocks/amp-story-text/edit.js
swissspidy: > Testing this it feels like maybe it would be good if the user could still assign a custom font size, too.
> Note that I didn't make any changes to `FontSizePicker` (yet)

At the moment I see two reasonable options here:

* Add a new "Auto" option to font size dropdown \*
* Add a "Automatically fit text to container" `ToggleControl` (like in the post editor) that is enabled by default.  
  The font size picker would only be displayed when this toggle is disabled. Screenshot:

Screenshot 2019-03-18 at 22 18 30

\* Worth noting that the font options come from the `withFontSizes` HOC (and thus editor settings), not sure how easy it is to override these.

----

Besides this, I'd probably extract `calculateFontSize` in its own class method or even move it to `helper.js` as it is static.

Also, I am not sure if this should use `setFontSize()` or use a new attribute, e.g. `autoFontSize`. But that probably makes more sense when using option 2 from above (toggle). That way you can set a custom font size, turn on the toggle, turn it off, and get back your custom font size from before.

swissspidy:
> Note that this PR has been developed on the latest develop branch of Gutenberg as of today (18.03) which seems to include some styling changes compared to the latest release.

Just noticed the styling changes now... Odd. Are they really intentional? 🤔

miina:
> This way we could only show the mover when hovering close to the left side of the block (or right side for RTL languages).

In the last iteration of AMP Stories it came out that it was unclear for users the block mover and the dragging functionality exist at all, and it was decided to display the block mover all the time when a block is selected (even when not hovering), for visibility. That's removed by now and is not necessary anymore due to not using the layers, however, maybe for discoverability, it is good to show it when hovering anywhere. Maybe that's a UX detail which we could reiterate on separately though, too. Thoughts? 

> Just noticed the styling changes now... Odd. Are they really intentional?

Which odd changes are you seeing? There was a change for block mover but this should already be addressed in the PR and should display as it was before, are you seeing something different?
Another change seems to be a general blocks' styling change (e.g. width), that should also be addressed in the PR, are you seeing something different?

Could you share a screenshot if you see something off in the PR + `develop` branch?

swissspidy:
> Which odd changes are you seeing? 

General changes in the Gutenberg editor; unrelated to this PR. Things like less rounded corners and darker border colors for the block toolbar. Needs some getting used to I guess.

assets/src/blocks/amp-story-text/edit.js
miina: > Add a "Automatically fit text to container" ToggleControl (like in the post editor) that is enabled by default.
The font size picker would only be displayed when this toggle is disabled.

Good idea, makes it consistent with other options too.

> Besides this, I'd probably extract calculateFontSize in its own class method or even move it to helper.js as it is static.

That's also a good point, especially since it would potentially be good to use it in case of other blocks at some point as well and not just with AMP Stories.

> That way you can set a custom font size, turn on the toggle, turn it off, and get back your custom font size from before.

Hmm. There are some cases where it might be unexpected to switch back to an old custom font size, however, there are more cases where it makes sense. 👍

swissspidy:
@miina I was just looking at the code again. Is the width/height here stored in pixels or percentages? For responsiveness I think we need percentages.

miina:
@swissspidy It's in pixels. Hmm, actually the font size is not accurate this way either.
So instead of assigning the pixels to the font size we should actually use the `amp-fit-text` in the frontend.

miina:
Actually, `amp-fit-text` doesn't seem to accept percentage values.

miina:
We can probably try assigning the percentage values to the wrapper and assing `fill` layout to `amp-fit-text`. _Trying it out_

miina:
@swissspidy Would you mind taking another look?

The width and height are in % now and also switched to using `amp-fit-text` in the frontend (if applicable).

swissspidy:
Very neat! The feature seems to work well.

There are certainly some styling issues we can address in #1967 to make sure the frontend experience matches the one in the editor as closely as possible. But for now I think this is good.

We can always iterate on this in future PRs :-)

Oh, and I can't wait for snapping to nicely align elements. That's gonna be 💥.

PR #1976, opened 2019-03-15
Fix warning triggered when filtering kses allowed attributes

### LINKED ISSUE #1975: Warning triggered when filtering kses allowed attributes
Amp plugin throws warning on `wp_kses_post` function.

The filter function `whitelist_block_atts_in_wp_kses_allowed_html` assumes `$tags` are always an associative array, but sometimes they could be like below:

```php
$tags = [
    'form' => [
        'action' => true,
    ],
    'strong' => true,
];
```

As a result, amp plugin triggers warnings.

```
( ! ) Warning: Cannot use a scalar value as an array in /var/www/wordpress/wp-content/plugins/amp/includes/admin/class-amp-editor-blocks.php on line 83
```

fumikito:
This commit is to fix #1975 by considering tags could not be an array.

### DIFF
--- a/includes/admin/class-amp-editor-blocks.php
+++ b/includes/admin/class-amp-editor-blocks.php
@@ -77,6 +77,9 @@ class AMP_Editor_Blocks {
 		}
 
 		foreach ( $tags as &$tag ) {
+			if ( ! is_array( $tag ) ) {
+				continue;
+			}
 			$tag['data-amp-layout']              = true;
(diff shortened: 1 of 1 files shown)


westonruter:
@fumikito please sign the CLA and I'll get this merged.

fumikito:
@westonruter I've signed it! Thanks for your review and patience.

swissspidy:
@googlebot please check again

PR #1981, opened 2019-03-16
[AMP Stories] Add maybe flushing rewrite rules.

### LINKED ISSUE #1978: [AMP Stories] Flush rewrite rules it AMP Stories rules don't exist

miina:
Fixes #1978.

Adds `maybe_flush_rewrite_rules` after registering the AMP Stories CPT which checks if rewrite rules for stories are already registered and if not, flushes the rewrite rules.


swissspidy:
This should only be a soft flush, not a hard one.

miina:
Ah, true, I'll change it to avoid `htaccess` changes.

PR #1986, opened 2019-03-18
Ensure newly supported theme support arguments do not trigger a warning

felixarntz:
Following #1745, two new theme support arguments were introduced. However, when using them you currently see an unexpected notice that they are invalid. This PR fixes the bug.


swissspidy:
Can we perhaps add tests for this?

PR #1987, opened 2019-03-18
Work around edge case combination of PHP 7.2+ with old ICU version

swissspidy:
This is an alternative, simpler approach to #1834.

Note: I also removed the PHP 5.4 check because the plugin now requires PHP 5.4.


swissspidy:
@s-a-s-k-i-a Can you please test this PR on your client's website and let me know if it fixes the issue for you? That would be awesome!

s-a-s-k-i-a:
> @s-a-s-k-i-a Can you please test this PR on your client's website and let me know if it fixes the issue for you? That would be awesome!

Fixed! I like this dryer code. Thanks!

PR #1988, opened 2019-03-18
Animation order: only show valid options

### LINKED ISSUE #1984: [AMP Stories] Extract hasCycle() helper function
The `animationOrder()` reducer uses a simple `hasCycle()` helper function to check whether a given predecessor is legit or if it would end up in a cycle (i.e. A starting after B, and B starting after A).

Since this check is only performed in the reducer, the user can still select a potentially invalid option in the UI. Selecting it just won't do anything though, since the reducer rejects it.

Ideally, this helper function gets extracted, perhaps as a selector, so that it can be used not only in the reducer, but also within the `getAnimatedBlocks` function in the `withAnimationControls` HOC.

From there, we have two options:

1. Filter out any invalid options from the list so that the user cannot see them.
2. Flag invalid options instead so that the `AnimationOrderPicker` can display them as such (e.g. make them disabled and maybe add a warning).

Option 1 is the most straightforward for now, although it might make users wonder why they don't see all options in the dropdown. If that turns out to be the case, we can always explore option 2.

Original PR: #1904

swissspidy:
Fixes #1984.

To verify:

1. Add first block (A) with animation.
2. Add second block (B) with animation, choose to begin after A.
3. Edit A, verify that you cannot select it to start after B.

### DIFF
--- a/assets/src/components/with-animation-controls.js
+++ b/assets/src/components/with-animation-controls.js
@@ -17,7 +17,7 @@ import { AnimationControls } from './';
 
 const applyWithSelect = withSelect( ( select, props ) => {
 	const { getSelectedBlockClientId, getBlockRootClientId, getBlock } = select( 'core/editor' );
-	const { getAnimationOrder } = select( 'amp/story' );
+	const { getAnimationOrder, isValidAnimationPredecessor } = select( 'amp/story' );
 
 	const currentBlock = getSelectedBlockClientId();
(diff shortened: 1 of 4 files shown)


swissspidy:
The build already passed, but somehow the status here isn't updating 🤔

miina:
I've just restarted the build, perhaps it helps.

PR #1989, opened 2019-03-18
Improve tag name determination

### LINKED ISSUE #1881: Semantic meaning of text block

swissspidy:
Tries to ensure that there’s only ever one h1 tag on a page.

Also takes into account elements positioned at the very bottom of a page. These are unlikely to be headings and will be wrapped in `<p>` tags.

Fixes #1881.


miina:
Maybe we could also reflect the tag switch within the editor by changing the font weight?

swissspidy:
> Maybe we could also reflect the tag switch within the editor by changing the font weight?

Sounds interesting, although I am not sure if the changes would be a bit unexpected for users. Perhaps worth exploring in a future PR?

To give an example of why it might be unexpected:

Let's say we have two blocks, A (content: "Foo") and B (content: "Bar"). Because of the `maybeSetTagName()` logic, A gets an `h1` tag, and B an `h2`. When I now enter some longer text in A, it gets degraded to a `p` and B gets promoted to a `h1`. -> The font weight changes for both.

miina:
> Sounds interesting, although I am not sure if the changes would be a bit unexpected for users. Perhaps worth exploring in a future PR?

Yes, probably it would be unexpected. I was thinking of this due to the difference between the frontend and the editor view -- probably what the user sees in the frontend is also unexpected (bold in the frontend, normal weight in the editor). Alternatively we could by default remove the bold font-weight from headings in the frontend, however, that might not be the best either. Thoughts?

We can definitely leave this to future PR, possibly as part of #1967.

swissspidy:
> Alternatively we could by default remove the bold font-weight from headings in the frontend, however, that might not be the best either.

That would be my suggestion for now though. The important part is the semantic meaning. Boldness can always be added via the block toolbar.

PR #1992, opened 2019-03-19
Add PHP Extension dependencies to composer.json

### LINKED ISSUE #1985: Composer: Declare PHP Extension dependencies
While looking at #1834 I noticed that we currently don't declare PHP extension and library requirements in `composer.json`.

Off the top of my head, we'd need the following platform dependencies:

```json
{
  "require": {
    "ext-intl": "*",
    "ext-libxml": "*",
    "ext-dom": "*"
  }
}
```

Perhaps even `"lib-icu": "*"`, since that's being used by `ext-intl`.

swissspidy:
* Sorts dependencies alphabetically
* Adds `ext-intl`, `ext-libxml`, `ext-dom`, and `ext-iconv` as platform dependencies
* Uses a pinned version for the `cweagans/composer-patches` requirement to match all the other entries

Fixes #1985.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,19 +1,23 @@
 {
   "name": "ampproject/amp-wp",
+  "type": "wordpress-plugin",
   "description": "WordPress plugin for adding AMP support.",
   "homepage": "https://github.com/ampproject/amp-wp",
-  "type": "wordpress-plugin",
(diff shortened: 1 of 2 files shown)


westonruter:
Love it.

PR #1993, opened 2019-03-19
[AMP Stories] Add Call to Action support

### LINKED ISSUE #1963: [AMP Stories] Call to Action (CTA) Block
**Tasks:**

* [x] Add a way to insert a button within a CTA layer.
* [x] Disable block mover for CTA block (see #1897)
* [ ] Prevent adding more than one CTA block ~(using `supports: { multiple: false }`)~
* [x] Add shortcut to add a CTA block, disable shortcut if already present

**Reference:**

https://www.ampproject.org/docs/reference/components/amp-story#amp-story-cta-layer

swissspidy:
Fixes #1963.

As per discussion on that issue, I first tried an approach that is similar to the one the Media & Text block uses. However, this caused more problems (UI wise but also development wise) than being useful and making sense.

**Overview:**

This PR adds a new `amp/amp-story-cta` block that can be inserted on the page when the following conditions are true:

* The current page is not the first page in the story
* There isn't a CTA block on the page already

The block is based on the `core/button` block, but with different labels and no support for things like the `title` attribute.

Since the AMP story page's inner blocks are wrapped in `<amp-story-grid-layer>`, this PR also adds a sanitizer that moves the CTA block outside of that as a direct descendant of the `<amp-story-page>`. At the same time, it also ensures that there's only ever one CTA block on a page. This is in addition to the client-side validation in the editor.

Since the CTA block is always in a fixed position on the frontend, the block is not visible in the editor's block navigation panel.

**To do:**

* [x] Port over editor (and perhaps frontend) CSS from button block
* [x] Fix frontend styling  
  Alignment doesn't seem to be working
* [x] Add font selection controls (font size, font family)
* [ ] Perhaps use better icon
* [ ] Make sure that dynamically changing `allowedBlocks` actually works . 
  This appears to be an upstream bug in Gutenberg: https://github.com/WordPress/gutenberg/issues/14515

**Screenshots**

Page in the editor with a CTA button in it:

Screenshot 2019-03-19 at 18 03 24

CTA shortcut in the top toolbar (when inserting is allowed):

Screenshot 2019-03-19 at 17 51 07


assets/src/blocks/amp-story-cta/edit.js
```diff
@@ -0,0 +1,148 @@
...
+		return (
+			<Fragment>
+				<div className={ className } title={ title } ref={ this.bindRef }>
+					<RichText
+						placeholder={ __( 'Add text…', 'amp' ) }
+						value={ text }
+						onChange={ ( value ) => setAttributes( { text: value } ) }
+						formattingControls={ [ 'bold', 'italic', 'strikethrough' ] }
+						className={ classnames(
+							'wp-block-button__link', {
```
swissspidy: Should change the class name here.

assets/src/blocks/amp-story-cta/edit.js
```diff
@@ -0,0 +1,148 @@
...
+									fallbackBackgroundColor,
+									fallbackTextColor,
+								} }
+							/>
+						</PanelColorSettings>
+					</InspectorControls>
+				</div>
+				{ isSelected && (
+					<form
+						className="block-library-button__inline-link"
```
swissspidy: Should change the class name here.

swissspidy:
@westonruter @miina What do you think of the current approach here so far?

westonruter:
> The current page is not the first page in the story

Where did you find this constraint in the AMP Stories spec? It does seem that the CTA grid layer fails to render properly when it is on the first page, but it doesn't cause an AMP validation error.

> * Make sure that dynamically changing `allowedBlocks` actually works .
>   This appears to be an upstream bug in Gutenberg: [WordPress/gutenberg#14515](https://github.com/WordPress/gutenberg/issues/14515)

I was able to insert a CTA block on the first page. This is due to the apparent bug in Gutenberg?

Perhaps this wouldn't be a common scenario at all, but someone could create a second page and add a CTA layer to it, and then they could re-order the pages to make the second page the first. The CTA would then just be removed without explanation. In that case should there be a warning notice displayed inside a CTA layer when it appears on the first page to account for this case?

However, I noticed that the normal post editor validation error warning notice is not being displayed in the story editor, so I need to look into that.

- [ ] Determine why validation error warning is not being displayed in story editor.

> Since the AMP story page's inner blocks are wrapped in `<amp-story-grid-layer>`, this PR also adds a sanitizer that moves the CTA block outside of that as a direct descendant of the `<amp-story-page>`.

It feels unfortunate that a sanitizer has to be used for moving an `amp-story-cta-layer` to the right spot. I don't suppose using a dynamic block would be easier.

westonruter:
There we go. AMP validation of stories is now fixed. The global warning is now restored, as well as the block-level warning which will now show inside the CTA layer when on the first page:

image

I noticed there is some oddity with the inline validation warning where it only shows up after saving and then hovering over the block. Something to do with how the store us updated 🤷‍♂️?

miina:
> Where did you find this constraint in the AMP Stories spec? It does seem that the CTA grid layer fails to render properly when it is on the first page, but it doesn't cause an AMP validation error.

This constraint was addressed in the previous version as well, however, didn't find information about it right now from the documentation, found the github issue: https://github.com/ampproject/amphtml/issues/14440

swissspidy:
> This constraint was addressed in the previous version as well, however, didn't find information about it right now from the documentation

Yeah I think I found it out when looking at the examples (which were added by that PR)

> I was able to insert a CTA block on the first page. This is due to the apparent bug in Gutenberg?

Yes. I'll try to find some time to look into this bug in Gutenberg next week.

> Perhaps this wouldn't be a common scenario at all, but someone could create a second page and add a CTA layer to it, and then they could re-order the pages to make the second page the first. The CTA would then just be removed without explanation. In that case should there be a warning notice displayed inside a CTA layer when it appears on the first page to account for this case?

Yes, it is on my todo list. I am going to add a filter similar to [`withMultipleValidation`](https://github.com/WordPress/gutenberg/blob/2ebc98eb438be408b3c83d0f0a7638d8fea5bddb/packages/edit-post/src/hooks/validate-multiple-use/index.js) for the CTA block.

> It feels unfortunate that a sanitizer has to be used for moving an amp-story-cta-layer to the right spot.

I know 😞 Perhaps better solutions emerge in the future.

miina:
>It feels unfortunate that a sanitizer has to be used for moving an amp-story-cta-layer to the right spot.

> I know 😞 Perhaps better solutions emerge in the future.

Could we filter the content instead within the AMP Stories logic? Or perhaps we could wrap each block into a separate `amp-story-grid-layer` (except for `amp-story-cta-layer`) instead of having all the blocks within it? This way we could ensure that it's last. Thoughts?
It does feel like AMP Stories shouldn't generate non-compatible code and depend on sanitization.

swissspidy:
>  Or perhaps we could wrap each block into a separate `amp-story-grid-layer`

Looking at the spec it seems that this seems doable.

> This way we could ensure that it (the CTA block) is the last

We'd still need to add some logic for that though, yes.

swissspidy:
> Or perhaps we could wrap each block into a separate `amp-story-grid-layer`

I've modified the code to wrap the blocks on save. That seems to work, except that I now get lots of block validation errors because of changing block IDs:

```
Content generated by `save` function:

<amp-story-grid-layer template="vertical"><h1 style="font-size:30px;width:76%;height:9%;position:absolute;top:10%;left:5%" class="" id="f5933f4d-425f-4173-b3c0-13ca7f685090"><amp-fit-text layout="fill" class="amp-text-content">Test</amp-fit-text></h1></amp-story-grid-layer>

Content retrieved from post body:

<amp-story-grid-layer template="vertical"><h1 style="font-size:30px;width:76%;height:9%;position:absolute;top:10%;left:5%" class="" id="c0155b96-2d39-450c-9b62-c549c2f70ca7"><amp-fit-text layout="fill" class="amp-text-content">Test</amp-fit-text></h1></amp-story-grid-layer>
```

Maybe the editor tries to get the ID from `<amp-story-grid-layer>` but fails, and then generates a new ID the actual inner element. Very strange.

----

Besides that, I now added a custom HOC for the CTA block validation:

Screenshot 2019-03-25 at 15 25 25
Screenshot 2019-03-25 at 15 29 37

Just needs some styling tweaks as the carousel navigation seems to interfere with the warning there.

Plus, the AMP validation kicks in there:

Screenshot 2019-03-25 at 15 25 36

We'd need to disable that for this particular block.

westonruter:
> We'd need to disable that for this particular block.

That can be done by replacing `remove_invalid_child` with the `removeChild` that you added at first for this specific case in the story sanitizer. Since the editor is manually showing the validation error, it makes sense that the story sanitizer would remove it silently.

swissspidy:
> Since the editor is manually showing the validation error, it makes sense that the story sanitizer would remove it silently.

I was talking about the validation error in the editor: it should not be displayed because I now added a custom error message for the CTA block. There should only be one, not both:

Screenshot 2019-03-25 at 15 25 36

I re-added the sanitizer now, but the double warning is still there. Plus, I still get validation errors despite the sanitizer:

Screenshot 2019-03-26 at 10 36 45

swissspidy:
> I re-added the sanitizer now, but the double warning is still there. Plus, I still get validation errors despite the sanitizer

It looks like we actually don't want to use `remove_invalid_child` in that case then.

swissspidy:
> Perhaps use better icon

@dawidmlynarz Do you perhaps have an idea for a good icon that we could use for the "Call to Action" block? Perhaps a combination of the current one and https://developer.wordpress.org/resource/dashicons/#external? For reference, the current one, indicating a button, is the one on the right side here: https://user-images.githubusercontent.com/841956/54626842-56882180-4a72-11e9-83f6-773b1a683482.png

> Make sure that dynamically changing allowedBlocks actually works .

I am going to try and see if this can be fixed in Gutenberg easily. Nothing we should worry about at the moment, especially since we have the sanitizer.

> It looks like we actually don't want to use remove_invalid_child in that case

@westonruter Can you confirm this?

dawidmlynarz:
@paulschreiber The only icon I can think about would be something like you have plus some "interaction marker" like cursor icon.

Bar - Call to Action

Please have a look at the rough draft above and let me know if this would work. I think it's better than a simple rectangle, which in this context doesn't have a strong meaning.

swissspidy:
Thanks @dawidmlynarz! I think that goes into the right direction. It looks a bit skewed though, maybe you could rotate the cursor a bit to the right? I'll happily use that as an SVG then.

After that, I can create another issue to consider improving the other icons there too.

paulschreiber:
@dawidmlynarz did you mention me by accident?

swissspidy:
@paulschreiber Yep, sorry for the noise! 🙂

dawidmlynarz:
Oops... Sorry for noise Paul, accident indeed. 

@swissspidy I will iterate on the icon and send over the SVG once done.

assets/src/blocks/amp-story/index.js
```diff
@@ -122,11 +122,7 @@ export const settings = {
 					)
 				}
 				<amp-story-grid-layer template="fill" style={ overlayStyle }>
```
miina: This looks like it doesn't have a closing tag. Probably something that happened with merging `amp-stories-redux`.

swissspidy: Whoops, indeed a merge conflict resolving gone wrong

miina:
Hmm, I'm seeing this error for some reason, basically an ID mismatch. Is this something that you encountered before as well? Are you seeing this locally, too?
Screen Shot 2019-04-04 at 3 21 36 PM

miina:
Also seeing this 🤔  Screen Shot 2019-04-04 at 3 24 30 PM

swissspidy:
> Hmm, I'm seeing this error for some reason, basically an ID mismatch. Is this something that you encountered before as well? 

Indeed I have, see https://github.com/ampproject/amp-wp/pull/1993#issuecomment-476224868. Although I thought it went away after a Gutenberg update or something, but apparently I remembered incorrectly.

> Also seeing this 

Ah, `assets/src/blocks/stories.js` still needs to be deleted from this branch. Maybe that is the cause.

dawidmlynarz:
@swissspidy Please have a look at the second iteration of the icon and let me know what you think.

Call2Action icon

Looking forward to your feedback!

swissspidy:
@miina The block validation errors should now be gone. Also fixed the sanitizer

@dawidmlynarz Cool, let's use that one for now!

assets/src/blocks/amp-story-cta/edit.js
```diff
@@ -0,0 +1,180 @@
...
+									maybeEnqueueFontStyle( value );
+									setAttributes( { ampFontFamily: value } );
+								} }
+							/>
+							<FontSizePicker
+								value={ fontSize.size }
+								onChange={ setFontSize }
+							/>
+						</PanelBody>
+						<PanelColorSettings
```
miina: Probably the repeating settings with Text Block we could at some point move to a separate file, maybe this could also have background opacity setting as well. Later.

assets/src/blocks/amp-story/index.js
```diff
@@ -121,12 +121,8 @@ export const settings = {
 						</amp-story-grid-layer>
 					)
 				}
-				<amp-story-grid-layer template="fill" style={ overlayStyle }>
```
miina: We should probably add this back since that's how the background overlay is displayed :)

assets/src/components/block-navigation.js
```diff
@@ -85,8 +90,10 @@ export default compose(
 		const { getCurrentPage, isReordering } = select( 'amp/story' );
 		const { getBlockOrder, getBlocksByClientId, getSelectedBlockClientId } = select( 'core/editor' );
 
+		const blocks = getCurrentPage() ? getBlocksByClientId( getBlockOrder( getCurrentPage() ) ) : [];
+
 		return {
-			elements: getCurrentPage() ? getBlocksByClientId( getBlockOrder( getCurrentPage() ) ) : [],
+			blocks: blocks.filter( ( { name } ) => ALLOWED_MOVABLE_BLOCKS.includes( name ) ),
```
miina: Curious why not display it in the Block Navigator, it's quite convenient for selecting blocks. At this moment it's not possible to use the navigator for moving blocks around anyway.

Or is it because the logic doesn't automatically force the CTA block as last? In this case that makes sense at this moment.

assets/src/blocks/amp-story/index.js
swissspidy: It's still there..?

assets/src/components/block-navigation.js
swissspidy: > Or is it because the logic doesn't automatically force the CTA block as last? In this case that makes sense at this moment.

Yup, that.

assets/src/blocks/amp-story-cta/edit.js
swissspidy: Good idea. Once both PRs are merged, we can move this into a separate `BlockColorSettings` component or something.

assets/src/blocks/amp-story/index.js
miina: Oh, right, sorry, didn't notice the green lines below are for `fill` layer.

swissspidy:
Thanks for your feedback @miina!

> Also, the style looks a little bit off right now and the block seems to be too wide, is this how you see it as well?

Indeed. Will try to add some improvements there.

swissspidy:
Current status:

Screenshot 2019-04-04 at 18 57 57

Also fixed another issue with block validation

miina:
Looks much better 👍

westonruter:
There seems to be a regression here, which may be related to the build error:

```
assets/src/blocks/amp-story/edit.js: line 4, col 8, Error - 'uuid' is defined but never used. (no-unused-vars)
```

The `amp-story-page` elements are not getting `id` attributes, which is causing them to be invalid (and thus resulting in an empty AMP story).

Additionally, there seems now to be `<br>` elements that are added before the `<amp-story-grid-layer>` and after the `</amp-story-grid-layer>`. This could be a bug with `wpautop`?

assets/src/blocks/amp-story-cta/index.js
```diff
@@ -0,0 +1,124 @@
...
+};
+
+export const name = 'amp/amp-story-cta';
+
+export const settings = {
+	title: __( 'Call to Action', 'amp' ),
+
+	description: __( 'Prompt visitors to take action with a button-style link.', 'amp' ),
+
+	icon: <SVG viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><Path fill="none" d="M0 0h24v24H0V0z" /><G><Path d="M19 6H5c-1.1 0-2 .9-2 2v8c0 1.1.9 2 2 2h14c1.1 0 2-.9 2-2V8c0-1.1-.9-2-2-2zm0 10H5V8h14v8z" /></G></SVG>,
```
westonruter: One idea for the CTA icon would be a “starburst”.

assets/src/blocks/amp-story/edit.js
```diff
@@ -42,13 +43,8 @@ const TEMPLATE = [
 ];
 
 class EditPage extends Component {
-	constructor( props ) {
-		// Call parent constructor.
-		super( props );
-
-		if ( ! props.attributes.anchor ) {
-			this.props.setAttributes( { anchor: uuid() } );
```
westonruter: This removal seems to be causing validation errors due to the page not getting an `id`.

assets/src/components/shortcuts.js
```diff
@@ -23,13 +25,26 @@ const Shortcuts = ( { insertBlock } ) => {
 					onClick={ () => insertBlock( block ) }
 					label={ blockType.title }
 					labelPosition="bottom"
+					disabled={ ! canInsertBlockType( block ) }
```
westonruter: Should the CTA button even be displayed when on the first page? It seems strange to have a button appearing by default which is disabled.

assets/src/helpers.js
```diff
@@ -240,6 +244,27 @@ export const addAMPExtraProps = ( props, blockType, attributes ) => {
...
+ *
+ * @return {Object} The element.
+ */
+export const wrapBlocksInGridLayer = ( element, blockType ) => {
+	if ( ! ALLOWED_MOVABLE_BLOCKS.includes( blockType.name ) ) {
+		return element;
+	}
+
+	return (
+		<amp-story-grid-layer template="vertical">
```
westonruter: Is it expected that every block is wrapped in an `amp-story-grid-layer` when looking at the source of an `amp-story`?

```html
<!-- wp:amp/amp-story-page {"autoAdvanceAfterDuration":0} -->
<amp-story-page style="background-color:#ffffff" class="wp-block-amp-amp-story-page">
    <amp-story-grid-layer template="fill"></amp-story-grid-layer>
    
    <!-- wp:amp/amp-story-text {"placeholder":"Write something!","tagName":"h1","autoFontSize":28,"positionTop":10} -->
    <amp-story-grid-layer template="vertical">
        <h1 style="font-size:28px;width:76%;height:9%;position:absolute;top:10%;left:5%" class=""
            id="46f4eecd-32c5-4ba0-8ec4-9e6c6b69cffb">
            <amp-fit-text layout="fill" class="amp-text-content">asdasdsad</amp-fit-text>
        </h1>
    </amp-story-grid-layer>

    <!-- wp:amp/amp-story-text {"tagName":"h1","autoFontSize":28,"positionTop":20} -->
    <amp-story-grid-layer template="vertical">
        <h1 style="font-size:28px;width:76%;height:9%;position:absolute;top:20%;left:5%" class=""
            id="dc2f83b8-d6d7-4989-9299-df8c6914dfa1">
            <amp-fit-text layout="fill" class="amp-text-content">asdasdasd</amp-fit-text>
        </h1>
    </amp-story-grid-layer>

    <!-- wp:image {"id":151,"positionTop":31,"positionLeft":12} -->
    <amp-story-grid-layer template="vertical">
        <figure class="wp-block-image" id="0da70857-1da6-4b19-b1be-bce2a7b4ac80"
                style="position:absolute;top:31%;left:12%">
            <img src="http://wordpressdev.lndo.site/content/uploads/2019/02/PIA19048_realistic_color_Europa_mosaic_alt-1024x757.jpg"
                 alt="" class="wp-image-151"/>
            <figcaption>converted PNM file</figcaption>
        </figure>
    </amp-story-grid-layer>
    <!-- /wp:image --> 
</amp-story-page>
<!-- /wp:amp/amp-story-page -->
```

westonruter: See 2d93c407483aa1d753d62c4668fa81264cc8b6b2 for this workaround.

assets/src/blocks/amp-story/edit.js
swissspidy: D'oh, wrong file 🤦‍♂️

assets/src/helpers.js
swissspidy: > Note the line break at the beginning.

Looking at the `post_content` you shared, that line break is a bit unexpected.

Regarding the `amp-story-grid-layer`, it is what I came up with in order to easily support the CTA layer within `InnerBlocks`. If you have another idea there, please let me know.

assets/src/blocks/amp-story-cta/index.js
swissspidy: Something like the 💥 emoji? you mean

westonruter: Something like this:

image

But not exactly 😄 

Actually, what you have with the rectangle does work, but what it is missing is a mouse cursor (or finger) overlaying it (e.g. 👆), to indicate that it is a button that can be clicked/touched. Since most things in AMP Stories aren't clicked, that could make its role more clear?

swissspidy: That‘s what Dawid‘s suggestion was here on the PR: a rectangle with a pointer cursor.

westonruter: > Call2Action icon

Yes! 😊 That's it.

Issue #1812, opened 2019-01-15
Add warning when featured image does not meet Google Search requirements


westonruter:
For Schema.org metadata, the featured image of a post is used as the `image` property. For this property, [Google Search requires](https://developers.google.com/search/docs/data-types/article#article_types):

> The URL to an image that is representative of the article or AMP story.
> 
> Due to format differences in search results, the following image guidelines only apply to general AMP pages, not AMP stories. AMP stories have different requirements for images.
> 
> * Only a marked-up image that directly belongs to the article should be specified.
Images should be at least 1200 pixels wide.
> * Every page must contain at least one image (whether or not you include markup). Google will pick the best image to display in Search results based on the aspect ratio and resolution.
> * Image URLs must be crawlable and indexable.
> * Images must represent the marked up content.
> * Images must be in .jpg, .png, or .gif format.
> * For best results, provide multiple high-resolution images (minimum of 800,000 pixels when multiplying width and height) with the following aspect ratios: 16x9, 4x3, and 1x1.

There is currently no guidance offered by the plugin for the featured image that is selected for a post. There should be a warning when the image is not 1200 pixels wide.

If the requirements are not met, should the image be excluded altogether? Is it better to have something rather than nothing?

We need to avoid situations where Google Search Console informs users of problems related to pages that the AMP plugin generates.

See also [plugin forum topic](https://wordpress.org/support/topic/google-search-console-image-size-smaller-than-recommended-size/).

westonruter:
Also, per:

> For best results, provide multiple high-resolution images (minimum of 800,000 pixels when multiplying width and height) with the following aspect ratios: 16x9, 4x3, and 1x1.

We should loop over all registered image sizes to find if there are others with these orientations, and then include all images that are at least 1200px wide, although it is unlikely that there will be any at that size.

robin-scott:
Confirming this is dropping onto a lot of backlogs right now - and I've been looking at a couple of solutions, one of which is a default image which complies with the specifications which could be substituted, although this "Only a marked-up image that directly belongs to the article should be specified." does appear to preclude that :|

This is "just" a warning, and Google does make a marked distinction between warning and error when it comes to indexing and Schema in my experience.

swissspidy:
> It should probably also warn when no featured image is selected at all, since this is also a GSC warning.

[...]

PR #1998, opened 2019-03-20
Set featured image if none exists, based on adding other images

### LINKED ISSUE #1954: Featured image handling and warning
From [this card](https://github.com/ampproject/amp-wp/projects/10#card-17824838):
- [x] Leverage WordPress post embeds for embedding AMP stories into posts. Make sure Post Embed block can embed AMP Story post type. Use featured image, title, and excerpt in preview. (Converted to #1931)
- [x] Set fallback default image in amp_story template when none is provided so that validation of the story is preserved. The warning should remain.
- [x] Show warning notice under the featured image control when no image is selected. Remove the global notice.
- [x] Extend Gutenberg pre-publish panel to block publishing while featured image is not set.

From [this card](https://github.com/ampproject/amp-wp/projects/10#card-18511617):
- [ ] Determine if "automatically" is the correct approach
Either or...
- [ ] Background media assigned to the first page to automatically populate featured image, if empty
- [ ] Instead of automatically being set in the postmeta, add a `get_post_metadata` filter for `_thumbnail_id` to supply image if it will be empty

kienstra:
If no featured image exists, set it based on:

* Adding a background image to a page block
* In the Image block, selecting or uploading an image via the Media Library
* In the Image block, clicking 'Upload' and uploading an image

Closes #1954


assets/src/blocks/amp-story/edit.js
```diff
@@ -79,7 +84,12 @@ class EditPage extends Component {
 		} );
 
 		if ( IMAGE_BACKGROUND_TYPE === mediaType ) {
+			const featuredMedia = select( 'core/editor' ).getEditedPostAttribute( 'featured_media' );
```
swissspidy: We're already using the `withSelect` HOC further down to pass some props. I'd move this there too.

kienstra: Good point, @swissspidy.

Also, I should have removed the logic in `amp-story/edit.js` entirely. The handler for the `MediaUpload` component conditionally updates the featured image:

upload-media


https://github.com/ampproject/amp-wp/pull/1998/commits/2237ad9c3add58b2b9c526bec55bb56016513938 removes the logic.

swissspidy:
Should this include some logic to set the featured image only when the uploaded image is large enough?

swissspidy:
For reference:

Each poster image should meet the recommended minimum size:

* Portrait: 696px x 928px
* Landscape: 928px x 696px
* Square: 928px x 928px

kienstra:
**Poster Image**

Hi @swissspidy,
Good point, this PR should probably ensure that the image meets the minimum size.

kienstra:
**Image Dimensions Now Verified**

Hi @swissspidy,
Good suggestion that this check for the minimum poster dimensions.

Based on the dimensions you shared, the following images are accepted:
https://lorempixel.com/669/928
https://lorempixel.com/928/669
https://lorempixel.com/928/928

While these dimensions won't be:
https://lorempixel.com/700/700
https://lorempixel.com/900/927

kienstra:
**Pre-publish Notice**

There's now a notice on clicking 'Publish' when an AMP Story doesn't have a featured image:
featured-image-here

assets/js/amp-block-validation.js
```diff
@@ -139,33 +138,6 @@ const ampBlockValidation = ( function() { // eslint-disable-line no-unused-vars
 			return false;
 		},
 
-		/**
-		 * Handles admin notices for AMP Stories.
-		 */
-		handleAMPStoryChange: function handleAMPStoryChange() {
```
kienstra: This notice at the top of the entire editor is replaced by a notice above the Featured Image control in `with-featured-image-notice.js`.

kienstra:
**Request For Another Review**

Hi @swissspidy,
Thanks for your suggestions. Could you please review this again?

Here's the expected behavior:

1. On adding an image in the following places, set it as the featured image if none exists:
a. The Image block's 'Media Library' or 'Upload' buttons:
image-block

b. The 'Background Media' button:

bg-media

2. Moves the notice for no featured image from the top of the block editor to above the 'Featured Image' section in the Inspector:
featured-image-notice

3. On clicking 'Publish' when there's no featured image, displays a notice in the pre-publish panel
no-feat-image

assets/src/components/with-placeholder-featured-image.js
```diff
@@ -0,0 +1,62 @@
...
+			}
+
+			delay( () => {
+				const selectedBlock = select( 'core/editor' ).getSelectedBlock();
+				if ( ! selectedBlock || ! selectedBlock.attributes.id ) {
+					return;
+				}
+
+				// Conditionally set the uploaded image as the featured image.
+				const media = wp.data.select( 'core' ).getMedia( selectedBlock.attributes.id );
```
swissspidy:
```suggestion
				const media = select( 'core' ).getMedia( selectedBlock.attributes.id );
```

You already import `select` at the top 🙂

assets/src/components/with-placeholder-featured-image.js
```diff
@@ -0,0 +1,62 @@
...
+			super( ...arguments );
+			this.componentWillUnmount = this.componentWillUnmount.bind( this );
+		}
+
+		/**
+		 * The handler for before the component unmounts.
+		 *
+		 * If there's no featured image, set this image as the featured image.
+		 * This component unmounts right after uploading an image.
+		 * But this needs to be delayed, as there's an apiFetch() call that gets the id.
```
swissspidy: Where is that `apiFetch()` call happening? Can we somehow avoid using `delay()` by `await`ing the result of that or something?

kienstra: Good question, I'll look at this more. The `apiFetch()` call happens when the `MediaUploader` calls [createMediaFromFile()](https://github.com/WordPress/gutenberg/blob/285875a1e2899b4999bb1396277be8a163354249/packages/editor/src/utils/media-upload/media-upload.js#L214).

assets/src/components/has-minimum-dimensions.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * The 3 minimum dimensions for the AMP story poster image are:
+ * 696px x 928px, 928px x 696px, or 928px x 928px.
+ *
+ * @param {Object} media A media object with width and height values.
+ * @return {boolean} Whether the media has the minimum dimensions.
+ */
+export default function( media ) {
+	const largeDimension = 928;
+	const smallDimension = 696;
```
westonruter: I think this component is overly generic in its name. It is specifically about the AMP Story poster image, due to its hard-coded values here. So either the component should be renamed the required dimensions should be supplied as a second argument, e.g.:

```js
hasMinimumDimensions( media.media_details, { smallDimension: 696, largeDimension: 923  } )
```

westonruter: This component would then be able to be re-used for general validation of the selected featured image for non-story posts as detailed in https://github.com/ampproject/amp-wp/issues/1812. In fact, if you want to extend this PR to also implement a general warning for post featured image it would probably be a good idea, as it would ensure the right abstraction.

assets/src/components/with-featured-image-notice.js
```diff
@@ -0,0 +1,50 @@
...
+
+			if ( hasFeaturedMedia ) {
+				return postFeaturedImage;
+			}
+
+			return (
+				<Fragment>
+					<Notice status="warning">
+						<span>
+							{ __( 'Featured image is used as poster-portrait-src and is mandatory', 'amp' ) }
```
westonruter: This message should be made more user-friendly, like:

> Selecting a featured image is required.

assets/src/components/has-minimum-dimensions.js
kienstra: Sounds good, Weston! I'll also extend this PR.

assets/src/components/with-featured-image-notice.js
kienstra: Sure, https://github.com/ampproject/amp-wp/pull/1998/commits/52896509b940921c2c27a445ca45cfb978d86ce0 changes the message:

selecting-featured-image-required

westonruter:
@kienstra one more thing that is needed is to make sure that there is _always_ a poster image supplied for the story, even if no featured image was selected. This will allow a story to be viewed on the frontend (as a draft) before the user selected the image. Without this, the entire `amp-story` is removed from the page and the user sees a white screen. So to ensure that this doesn't happen, a placeholder image that says something like:

> IMAGE NOT YET SUPPLIED

Should be shown.

The code in question is here: 

https://github.com/ampproject/amp-wp/blob/52896509b940921c2c27a445ca45cfb978d86ce0/includes/templates/single-amp_story.php#L34-L51

So if `$poster_portrait_src` is empty, then the placeholder should be supplied.

Also, the single template is not utilizing the other image sizes for AMP stories. In addition to `poster-portrait-src` there should be attributes provided for `poster-landscape-src` and `poster-square-src`, per the [docs](https://www.ampproject.org/docs/reference/components/amp-story#poster-guidelines-(for-poster-portrait-src,-poster-landscape-src,-and-poster-square-src)).

assets/src/amp-story-editor-blocks.js
```diff
@@ -315,11 +319,15 @@ store.subscribe( () => {
...
 } );
 
+registerPlugin( 'amp-featured-image-pre-publish', { render: withPrePublishNotice } );
+
 addFilter( 'blocks.registerBlockType', 'ampStoryEditorBlocks/setBlockParent', setBlockParent );
 addFilter( 'blocks.registerBlockType', 'ampStoryEditorBlocks/addAttributes', addAMPAttributes );
 addFilter( 'editor.BlockEdit', 'ampStoryEditorBlocks/addAnimationControls', withAnimationControls );
 addFilter( 'editor.BlockEdit', 'ampStoryEditorBlocks/addStorySettings', withAmpStorySettings );
 addFilter( 'editor.BlockEdit', 'ampStoryEditorBlocks/addPageNumber', withPageNumber );
+addFilter( 'editor.BlockEdit', 'ampStoryEditorBlocks/addPageNumber', withEditFeaturedImage );
```
westonruter: I think `addPageNumber` here is copy/pasta, right?

kienstra: Thanks, you're right. https://github.com/ampproject/amp-wp/pull/1998/commits/b9cd74b2716c354f42be76b60f77cd63eea32bcb changes the hook name.

assets/src/components/with-placeholder-featured-image.js
kienstra: https://github.com/ampproject/amp-wp/pull/1998/commits/7a3df1afbe8c35d8170d213de78611e40af3016d prevents using `delay()` by instead filtering `BlockEdit`.

assets/src/components/with-placeholder-featured-image.js
kienstra: Good idea, I [committed](https://github.com/ampproject/amp-wp/pull/1998/commits/358c63328c5f0d335b56237988e33d21ff495bc8) your suggestion.

kienstra:
**Placeholder Post Image**

Hi @westonruter,
>So to ensure that this doesn't happen, a placeholder image...should be shown

Sounds good, would you mind if I asked @da to create an image for that?

kienstra:
**Additional Poster Attributes**

Hi @westonruter,
>In addition to poster-portrait-src there should be attributes provided for poster-landscape-src and poster-square-src, per the docs.

Good point. How does https://github.com/ampproject/amp-wp/pull/1998/commits/c2381f2360d85d104bec8468bb441698e353dbc0 look? The `poster-square-src` and `poster-landscape-src` attributes are now added:

poster-src

assets/src/components/has-minimum-dimensions.js
kienstra: Hi Weston, https://github.com/ampproject/amp-wp/pull/1998/commits/22b0f54e64952c26345920bc8dfbc79b57fddb85 prepares `hasMinimumDimensions()` for use in a different context, like #1812. But I still need to extend this PR to address #1812.

kienstra:
**Post Featured Image Notice**

This adds the following to a normal post:
1. A notice above the 'Featured Image' panel if it doesn't exist or is the wrong dimension
featured-image-here

2. A notice on clicking 'Publish' if the featured image doesn't exist or has the wrong width
notice-publish

kienstra:
**Also Adds Size Validation To AMP Story**

This also adds the size notices [shown above](https://github.com/ampproject/amp-wp/pull/1998#issuecomment-476015923) to AMP stories. Before, there were only notices if the featured images didn't exist at all.

assets/src/amp-block-editor.js
```diff
@@ -6,9 +6,15 @@ import { FormToggle, Notice } from '@wordpress/components';
 import { Fragment, RawHTML } from '@wordpress/element';
```
kienstra: This might not be the best approach. I renamed `assets/src/amp-block-editor-toggle.js` to `assets/src/amp-block-editor.js`, as I added the featured image notices to it. A more general name that doesn't have 'toggle' made more sense then.

But maybe there's a better file to put this logic. I wanted to reuse `getFeaturedImageNotice` and `getPrePublishNotice` from the `components/` modules.

assets/src/amp-block-editor.js
```diff
@@ -95,6 +101,41 @@ function ComposedAMPToggle() {
...
+ * Whether the image has the minimum width for a featured image.
+ *
+ * This should have a width of at least 1200 pixels
+ * to satisfy the requirement of Google Search for Schema.org metadata.
+ *
+ * @param {Object} media A media object with width and height values.
+ * @return {boolean} Whether the media has the minimum dimensions.
+ */
+const hasMinimumFeaturedImageWidth = ( media ) => {
+	return ( media.width && media.width >= 1200 );
```
kienstra: If it's alright, I used an entirely different function for normal posts than the AMP Story featured image. The logic was quite different.

kienstra:
**Request For Another Review**

Hi @swissspidy,
Hope you're doing great.

When you have a chance, could you please review this PR again?

The main changes lately are [adding notices](https://github.com/ampproject/amp-wp/pull/1998#issuecomment-476015923) if the featured image sizes are wrong in AMP Stories and posts, and a [fallback image](https://github.com/ampproject/amp-wp/pull/1998/commits/d8add660957104e3beaff78bb8a129673b6cf8d6) if there's no featured image.

Thanks, Pascal! And congratulations on speaking at WordCamp Europe.

assets/src/amp-block-editor.js
swissspidy: wfm

kienstra:
Thanks, @swissspidy! You're right, we should document the fact that selecting a big enough image makes in the featured image (poster-portrait-src).

PR #2000, opened 2019-03-20
[AMP Stories] JS Build Tooling Optimizations

### LINKED ISSUE #1912: Streamline JS build tooling
The [@wordpress/scripts](https://www.npmjs.com/package/@wordpress/scripts) package contains some neat scripts that can be used for our Webpack / Babel setup.  

Especially with the upcoming 3.0.0 release, we can ditch lots of our custom configuration and re-use what the Gutenberg team uses. For example, that package already provides the `camelCaseDash` function and a default Webpack config we can build upon. Plus, having a good foundation to make our code testable can't be bad either :-)

This would also be a good opportunity to improve the JS folder structure a bit. I'd like to explore using at least parts of the hot module replacement solution Human Made came up with at https://humanmade.com/2018/11/26/hot-module-replacement-for-gutenberg-blocks/. Using Webpack’s `require.context` to load all the blocks would be really neat already.

In the same process, this could be a great opportunity to move the relevant CSS closer to the source. Development becomes easier when the CSS would live right next to important JS components.

During build, Webpack could extract and bundle the CSS files so the end result would be the same as now.

https://github.com/webpack-contrib/mini-css-extract-plugin is a Webpack plugin that can be leveraged for this.

swissspidy:
_More information to follow._

In #1912 I proposed streamlining our current JS build process to make it easier to develop individual components and to contribute to the AMP Stories feature as a whole. This is the first attempt at doing so.

At the core of this PR is the `@wordpress/scripts` package, which essentially is a collection of reusable scripts for WordPress development. These cover build tooling, linting, and testing (including E2E tests!).

**Main Changes**

* Change `webpack.config.js` to be built on top of the one provided by `@wordpress/scripts`
* Include `mini-css-extract-plugin` to support including CSS files in JS  
  Prerequisite for follow-up tasks
* Use webpack's `require.context` feature to load and register all block types   
  This way blocks only export their settings and don't automatically register themselves, making it easier to test them
  **Note:** This is also the reason for the high number of modified lines in this PR - lots of indentation changes :-)

**Follow-up Tasks:**

* [x] Include PostCSS and RTLCSS configs  
  Allows for proper RTL support and usage of things like variables, nested selectors, etc. in a browser-compatible way
* [ ] Move relevant styles from `amp-editor-story-blocks.css` to live next to their respective components
* [ ] Enqueue compiled AMP Stories editor CSS
* [ ] Extend `mini-css-extract-plugin` config so we can have 1 stylesheet for the editor, and 1 for the frontend
* [ ] Move `amp-editor-blocks.js` to `src` and properly use JSX
  It works well for AMP Stories so far, so this would add some consistency to the setup
* [ ] Fix all `stylelint` issues to adhere to WordPress CSS coding standards
* [ ] Make sure all the linters are run during CI

Fixes #1912.


miina:
- Not sure why but for some reason on this PR drag and drop breaks for the Text Block. Steps to reproduce:
Create a Page -> add some text, drag it to another location -> Update Post -> Refresh.
Screen Shot 2019-03-22 at 4 38 22 PM

- Another weird issue:
Screen Shot 2019-03-22 at 4 35 25 PM

The Text block is showing in the Root level Inserter.

- Another weird issue: when removing all the blocks from a Page, saving, and refreshing, this error shows, also, the post seems to have new changes right away (Update button active, etc.)
Screen Shot 2019-03-22 at 4 36 16 PM

It's weird since the Page doesn't even have an image block.

Let me know if these aren't happening for you.

swissspidy:
I think I see the same issues, but I am not sure why... Could also be related to the new Gutenberg version. I'll have to check over the weekend.

swissspidy:
@miina I changed the code so that all the filters are applied before the blocks are being registered. Otherwise it's too late for them to work for the custom blocks. This seems to have fixed all the issues.

miina:
This error still happens for some reason:
Screen Shot 2019-03-26 at 9 54 51 AM

I'm seeing actually multiple logs in the console, these specifically seem to occur when clicking on the Inserter in the Toolbar, even when there are no Image blocks on the Page.

Can you still see this issue, too?

Actually, it looks like it's coming from Reusable blocks, namely, after using `insertBlocks` -- not sure if I understand correctly but it seems to be comparing the initial source for Reusable blocks which doesn't have IDs with newly inserted blocks which do have IDs. 😕 Perhaps it's an issue with a recent update?

I've created a Reusable block which consists of two images, probably that's why I'm getting a lot of these errors in the console when clicking the Inserter.

In case you're not seeing the error right now -- would you mind testing if you see it when you add a Reusable Block including an Image, and then using the Toolbar Inserter? (This isn't related to this PR, happens on `amp-stories-redux` as well for me.)

swissspidy:
@miina Can confirm that I see this issue too on the `amp-stories-redux` branch under the same conditions.

swissspidy:
@miina I created an issue upstream: https://github.com/WordPress/gutenberg/issues/14639. The error is kinda logical, so not sure if we could really fix it. Since this is not related to this PR, do you think we can land it?

PR #2001, opened 2019-03-20
[AMP Stories] Stylesheets handling.

### LINKED ISSUE #1958: [AMP Stories] Fix styling for TwentyNineteen theme
Looks like the AMP Stories editor interface breaks with TwentyNinenteen theme due to the theme applying custom styles to the editor.
Screen Shot 2019-03-13 at 11 43 16 AM

miina:
Fixes #1958.

- Filters out enqueued stylesheets which don't have AMP Stories style marked as a dependency;
- Removes all theme stylesheets which have been added to editor via `add_editor_style`;
- Enqueues `amp-stories.css` instead of adding it inline.


includes/class-amp-story-post-type.php
```diff
@@ -200,11 +208,46 @@ public static function filter_kses_allowed_html( $allowed_tags ) {
...
+				}
+				$dep = wp_styles()->registered[ $handle ];
+
+				// If we have amp-stories as dependency, allow the style.
+				if ( is_array( $dep->deps ) && in_array( self::AMP_STORIES_STYLE_HANDLE, $dep->deps, true ) ) {
+					return true;
+				}
+
+				// Disable the active theme's style.
+				if ( false !== strpos( $dep->src, get_stylesheet_directory_uri() ) ) {
```
westonruter: We should probably include a check for `get_template_directory_uri()` as well, in the case of parent/child themes.

includes/class-amp-story-post-type.php
```diff
@@ -282,6 +336,13 @@ public static function filter_block_editor_settings( $editor_settings, $post ) {
 		if ( self::POST_TYPE_SLUG === $post->post_type ) {
 			unset( $editor_settings['colors'] );
 		}
+		if ( get_current_screen()->is_block_editor && isset( $editor_settings['styles'] ) ) {
+			foreach ( $editor_settings['styles'] as $key => $style ) {
+				if ( isset( $style['baseURL'] ) && false !== strpos( $style['baseURL'], get_stylesheet_directory_uri() ) ) {
```
westonruter: Also need to check `get_template_directory_uri()` here.

westonruter: What about theme styles that are enqueued for AMP Stories specifically? Should this check for the `deps` as well to not remove if that is the case?

includes/templates/single-amp_story.php
```diff
@@ -55,7 +55,7 @@
 
 		<?php
 		// Note that \AMP_Story_Post_Type::filter_frontend_print_styles_array() will limit which styles are printed.
-		wp_styles()->do_items();
+		print_late_styles();
```
westonruter: The `print_late_styles()` will ultimately call `wp_styles()->do_items()` so this makes sense. But I'm curious if there was a specific reason for this change?

includes/class-amp-story-post-type.php
miina: These filtered styles are added via [`add_editor_style()`](https://developer.wordpress.org/reference/functions/add_editor_style/) which only gets the stylesheet as the param, there are no dependencies to check.

I've added a whitelisting filter instead within ac852a3, thoughts?

includes/class-amp-story-post-type.php
miina: Added in ac852a3.

includes/templates/single-amp_story.php
miina: That was actually something that you originally mentioned in https://github.com/ampproject/amp-wp/issues/1958#issuecomment-473082141, it made sense  since it ultimately calls `do_items()` 😄

swissspidy: More readable this way too 👍

miina:
> Note that without any theme styles #1967 gets more important, as the editor is now lacking some basic things like a max-width. 

Yep, I'll start looking into #1967 👍

includes/class-amp-story-post-type.php
```diff
@@ -282,9 +335,46 @@ public static function filter_block_editor_settings( $editor_settings, $post ) {
...
+	 */
+	public static function is_theme_stylesheet( $url ) {
+
+		if (
+			false !== strpos( $url, get_stylesheet_directory_uri() ) ||
+			false !== strpos( $url, get_template_directory_uri() )
+		) {
+			return true;
+		}
+		return false;
```
westonruter: Minor style point, but this function could be refactored to be just:

```
return (
    0 === strpos( $url, get_stylesheet_directory_uri() )
    ||
    0 === strpos( $url, get_template_directory_uri() )
);
```

Note also that testing for `0` is better than `false` because it needs to make sure that the string is at the beginning, not somewhere later (like as a query param).

miina: Sure, I'll make that change and then merge.

Issue #1995, opened 2019-03-20
Global / per Story Slide Duration Options


miina:
AC: User can set the slide duration options either globally or per story.
AC: The global / per story setting can be overridden per slide.

Investigate options for setting the slide duration per story or globally. This way the user wouldn't have to set each slide separately in the likely use case of wanting to use `auto` or `media` for all slides.

swissspidy:
Note that it's finally possible to actually extend the document sidebar with the upcoming new Gutenberg version.

See https://github.com/WordPress/gutenberg/pull/13361 / https://github.com/WordPress/gutenberg/issues/13357

swissspidy:
@dero The PR is still set to draft – it's not clear to me whether this is now ready for review or still WIP. Can you clarify?

dero:
**Summary for QA:**

This PR introduces a concept of global settings and story settings. 

- Global settings: set on AMP plugin settings page, applied as defaults to every new story.
- Story settings: set on a story post, applied as defaults to every new page.
- Page settings: set on a page block, applied to that block only.

**Test instructions:**

1. Go to AMP plugin settings page and enable the Stories experience.
2. Change the values in the Stories Settings field and check the values persist upon saving.
3. Create a new story and notice a "Story Settings" panel in the "Document" tab.
4. Verify that values in that panel match what was set globally in step 2.
5. Click the first page and on the "Block" tab verify that the values in the "Page Settings" box match the "Story Settings".
6. Change the "Story Settings" to something else in the "Document" tab.
7. Verify that the "Page Settings" in the "Block" tab did not change.
8. Create a new page.
9. Verify that the "Page Settings" in the "Block" tab are equal to what was set in step 6.
10. Create a new page.
11. Change the "Page Settings" in the "Block" tab for that page to something else.
12. Check that other pages were not affected and their "Page Settings" stayed intact.
13. Check that the "Story Settings" in the "Document" tab remained unchanged.
14. Finish the story and publish it.
15. Verify that the advancement options are applied correctly on the front-end.

**Screencast:** https://www.youtube.com/watch?v=TppJ96XE-qU

csossi:
Verified in QA

PR #2015, opened 2019-03-22
[AMP Stories] Match FE and editor view

### LINKED ISSUE #1967: [AMP Stories] Make sure that the frontend and editor view match

miina:
Fixes #1967.

This PR does the following:
- Matches the fonts of various blocks in the FE and editor.
- Adjusts styles for each block, such as the table block, pullquote, etc.
- Changes the text style in the editor for Text block when the tagName switches.
- Enqueues AMP Stories embed style in the FE as well since it's possible to embed another AMP Story in a story.

Some issues that came out from the PR:
[RESOLVED] - Audio block seems to show validation error:
Screen Shot 2019-03-22 at 2 38 29 PM
[RESOLVED] - Fixed font sizes are not responsible, so there are still differences in how it displays in the FE.

### DIFF
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -624,35 +624,6 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
 	}
 }
 
-#amp-root-navigation .editor-block-navigation__item:hover {
-	color: #191e23;
-	border: none;
-	box-shadow: none;
-	background: #f3f4f5;
-}
-
-#amp-root-navigation .components-button.editor-block-navigation__item-button {
-	display: flex;
-	align-items: center;
-	width: 100%;
-	padding: 6px;
-	text-align: left;
-	color: #40464d;
-	border-radius: 4px;
-}
-
-#amp-root-navigation li {
-	margin-bottom: 6px;
-}
-
-#amp-root-navigation .block-editor-block-icon {
-	margin-right: 6px;
-}
-
-#amp-root-navigation .editor-block-navigation__container {
-	padding: 7px;
-}
-
 /*
  * 8. Block mover
  * Disable the block mover at all times for page blocks.
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -39,6 +39,10 @@
 	}
 }
 
+#amp-story-editor * {
+	font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
+}
+
 /*
  * 1. Style story pages wrapper.
  * - Add thicker border around the pages
@@ -251,10 +255,6 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
  * Hides the toggler and shows the navigator by default on large screens.
  */
 
-#amp-root-navigation {
-	font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
-}
-
 @media( max-width: 960px) {
 	.editor-block-list__block-edit > div > .editor-selectors {
 		display: none;
@@ -317,7 +317,6 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
 		font-size: 12px;
 		color: #666;
 		letter-spacing: 1px;
-		font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
 	}
 
 	.editor-block-list__layout div[data-type="amp/amp-story-page"] .amp-story-page-number + div {
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -34,7 +34,7 @@
 		margin: 0 auto;
 	}
 	#amp-story-editor {
-		min-width: 800px;
+		min-width: 820px;
 	}
 }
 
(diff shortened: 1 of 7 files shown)


assets/css/amp-editor-story-blocks.css
```diff
@@ -24,6 +25,18 @@
...
+ * 0. General editor layout styles.
+ */
+@media screen and (min-width: 1280px) {
+	.edit-post-visual-editor {
+		max-width: 1024px;
+		margin: 0 auto;
+	}
+	#amp-story-editor {
+		min-width: 800px;
+	}
```
miina: @swissspidy On this: https://github.com/ampproject/amp-wp/pull/2001#pullrequestreview-217233133
I'm not actually seeing a difference between how it is right now and how it was with editor styles. Note that I'm using twentynineteen theme and am not using a wide screen.

Would you mind checking if this displays normally in your screen and adjust if not? Setting these widths also moves the Block Navigation a little bit and I know there was some discussion on where it should be displayed, I might have moved it incorrectly.

westonruter:
> * Audio block seems to show validation error:

@miina: The problem seems to be:

```
The implied layout 'FIXED' is not supported by tag 'amp-story >> amp-audio'.
```

If I then supply the `fixed` layout, then I get:

```
The specified layout 'FIXED' is not supported by tag 'amp-story >> amp-audio'.
```

It seems that, interestingly, the only `layout` that is allowed is… `nodisplay`. When I supply this layout then I get another error:

```
 The mandatory attribute 'autoplay' is missing in tag 'amp-story >> amp-audio'.
```

So all of this to say, an `amp-audio` has to look like this:

```html
<amp-audio autoplay layout="nodisplay" src="https://example.com/content/uploads/2019/02/do-you-know-I-am-batman.mp3"></amp-audio>
```

Aside: The `noscript > audio` child of `amp-audio` is also causing complaints for the AMP validator.

👉  What this means is that I don't think audio should be insertable as a block. Rather, it should be treated like background media. I suggest changing the current “Background Media” to be “Background Visual” and then to add a new separate control for selecting “Background Audio”.  This would be out of scope of this PR, naturally. I'm guessing that if you select a background then any `amp-video` should automatically be made to be _muted_, as otherwise the two would play concurrently which would be bad.

swissspidy:
Few observations:

* The editor area was way to small on my 15" screen (fixed now)
* Everything is in serif 🤔 
* Block navigation styling is broken
* The editor area is a little bit too long on a 15" screen, just enough that the navigation at the bottom is not fully visible without scrolling
* Text blocks with background color are wider on frontend because they're not `inline-block`.

Screenshots:

Screenshot 2019-03-25 at 15 47 28
Screenshot 2019-03-25 at 15 52 27
Screenshot 2019-03-25 at 15 52 23

miina:
@swissspidy Thanks for testing!

> Everything is in serif 🤔

Hmm, as far as I can see then everything is in serif in the editor for normal posts as well. Is it different for you elsewhere?

> Block navigation styling is broken

Should be fixed within 0698ecb, looks like the classnames changed meanwhile.

> The editor area is a little bit too long on a 15" screen, just enough that the navigation at the bottom is not fully visible without scrolling

Not sure what could be a solution for that. In my case, I've never seen the editor area fully. Making the Page shorter would mean that it would also be narrower, that's already quite narrow now. Reducing the padding of the whole editor would make the Permalink setting hide when clicking on the Title. Suggestions/thoughts?

> Text blocks with background color are wider on frontend because they're not inline-block.

Changed the width of the inner `p` to 100% in the editor. It probably should take the full area that has been created by resizing the block. This way center text-align the user could create wider areas with background color and text in the center.

swissspidy:
> Hmm, as far as I can see then everything is in serif in the editor for normal posts as well. Is it different for you elsewhere?

Correct, except for when a theme adds some styles there. Not an issue for things like the story title, but for example the page numbers and block navigation should be sans-serif.

> Should be fixed within 0698ecb, looks like the classnames changed meanwhile.

Ah, indeed. Gutenberg moved lots of stuff to the new `block-editor` package and of course changed class names and such. There are more areas where this can be an issue for us. I'll try to investigate in a separate issue.

> Suggestions/thoughts?

Hmm let's let this sink in for a bit and revisit later.

miina:
> Correct, except for when a theme adds some styles there. Not an issue for things like the story title, but for example the page numbers and block navigation should be sans-serif.

Added `sans-serif` for these cases within 99b19c2.

swissspidy:
Since all text blocks within a story are now serif by default... I doubt that this is something many people would want.

Right now you'd have to change _every single_ text block if you want sans-serif stories. This begs the question if there should be a global setting to change the default font family (related: #1995).

miina:
@swissspidy Do you think we should set the default font generally to `sans-serif`? You probably know better about the fonts, so I'm happy to change the general font in the editor.

swissspidy:
After testing this again, yeah I think so.

Let's use `font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;` by default for now for stories, but also keep this kind of default settings in mind.

miina:
Sounds good, added within abf739c.

Also added new issue as you suggested: https://github.com/ampproject/amp-wp/issues/2028

swissspidy:
That should be changed for the frontend as well though, no?

miina:
Indeed 💡

PR #2018, opened 2019-03-23
Add support for Archives widget block when displaying as dropdown

### LINKED ISSUE #1991: Archives block with dropdown is not AMP compatible
The Archives widget has been ported to a block. When displaying as a dropdown, it is not valid AMP:

image

The same thing that was done for the Categories block needs to be done for the Archives block. That is, the `AMP_Core_Block_Handler` needs to be extended to override the `render_callback` of the `core/archives` block in the same way as it is being done for the `core/categories` block.

westonruter:
The Archives widget has joined the Categories widget as being available in blocks. These blocks both allow for the items to be displayed as a dropdown. When displayed as a dropdown, JavaScript is required to navigate once the selected option is changed. This script causes an AMP validation error and so we must use the AMP equivalent.

* Use an AMP-compatible alternative to `onchange` in the Archives widget.
* Use `render_block` filter instead of overriding the `render_callback`. 
* Eliminate duplicated code for Categories and Archives widget in favor of just applying the diff to make it valid AMP. This also fixed the alignment of the Categories block which had diverged in core.
* Always install Gutenberg during unit testing to be able to be able to test new feature plugin features. See also https://github.com/xwp/wp-dev-lib/pull/297.
* Make sure that the Archives widget does not output `uniqid()` in favor of an auto-incremented ID. See https://core.trac.wordpress.org/ticket/44883.
* Add tests for actual output of Categories and Archives widgets.

Fixes #1991.

The widget is configured like this:

> image

The widget is rendered like this on the frontend:

> image

# Archives Widget Before

> image

# Archives Widget After

> image

### DIFF
--- a/includes/embeds/class-amp-core-block-handler.php
+++ b/includes/embeds/class-amp-core-block-handler.php
@@ -13,75 +13,64 @@
 class AMP_Core_Block_Handler extends AMP_Base_Embed_Handler {
 
 	/**
-	 * Render methods for blocks.
+	 * Methods to ampify blocks.
 	 *
 	 * @var array
 	 */
-	protected $block_render_methods = array(
-		'core/categories' => 'render_categories_block',
-		'core/archives'   => 'render_archives_block',
+	protected $block_ampify_methods = array(
+		'core/categories' => 'ampify_categories_block',
+		'core/archives'   => 'ampify_archives_block',
 	);
 
 	/**
 	 * Register embed.
 	 */
 	public function register_embed() {
-		if ( ! class_exists( 'WP_Block_Type_Registry' ) ) {
-			return;
-		}
-
-		$registry = WP_Block_Type_Registry::get_instance();
-		foreach ( $this->block_render_methods as $block_name => $render_method ) {
-			$block = $registry->get_registered( $block_name );
-			if ( $block ) {
-				$block->original_render_callback = $block->render_callback;
-				$block->render_callback          = array( $this, $render_method );
-			}
-		}
+		add_filter( 'render_block', array( $this, 'filter_rendered_block' ), 0, 2 );
 	}
 
 	/**
 	 * Unregister embed.
 	 */
 	public function unregister_embed() {
-		if ( ! class_exists( 'WP_Block_Type_Registry' ) ) {
-			return;
-		}
-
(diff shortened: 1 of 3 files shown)


westonruter:
Best to review with whitespace suppressed in the diff: https://github.com/ampproject/amp-wp/pull/2018/files?w=1

westonruter:
I'm going to simplify this quite a bit more.

includes/embeds/class-amp-core-block-handler.php
```diff
@@ -45,25 +44,35 @@ public function register_embed() {
...
+	 * This excludes the disallowed JS scrips, adds <form> tags, and uses on:change for <select>.
+	 *
+	 * @see render_block_core_categories()
 	 *
 	 * @param array $attributes Attributes.
 	 * @return string Rendered.
 	 */
-	public function render( $attributes ) {
+	public function render_categories_block( $attributes ) {
+		$archives_block = WP_Block_Type_Registry::get_instance()->get_registered( 'core/categories' );
```
amedina: Should it be `core/archives`?

westonruter: Yes, but I've totally refactored it now so this is fixed.

includes/embeds/class-amp-core-block-handler.php
```diff
@@ -109,9 +112,105 @@ public function render( $attributes ) {
...
+			$class .= " align{$attributes['align']}";
+		}
+
+		if ( isset( $attributes['className'] ) ) {
+			$class .= " {$attributes['className']}";
+		}
+
+		$class .= ' wp-block-archives-dropdown';
+
+		$dropdown_id = 'wp-block-categories-dropdown-' . $block_id;
```
amedina: Is the mixing or `archives` and `categories` correct? (e.g. `wp-block-categories-dropdown-` for tthe Archives widget dropdown)

amedina:
Some good simplifications already. Still WIP?

includes/embeds/class-amp-core-block-handler.php
westonruter: You're right. This is now outdated due to the refactor.

westonruter:
@amedina Now it's good to review.

Issue #2023, opened 2019-03-26
Enhance Compatibility debugging workflow narrative


amedina:
The AMP plugin functionality is centered around interactions between `actions taken by the plugin` and `actions taken by the user`. The plugin can be configured to detect validation errors and automatically remove them (automatic sanitization). Or it can be configured to only detect errors but not removing them (manual sanitization). The user/developer has the control of what to do with detected errors. He/she can `accept` the plugin action (remove/not remove), or he/she can `reject` the plugin action. 

This dicothomy between plugin and user actions percolates to the compatibility tool, and it is a main source of confusion for users of the plugin. This manifests in the labeling of errors that are surfaced by the compatibility tool as `New Accepted` or `New Rejected`, and the labeling of errors after the user has taken an action: `Accepted` or `Rejected`. 

The problem is that the plugin actions are `Remove` and `Keep` (referring to the fate of the offending markup causing an error), while the user action is `Accept` or `Reject` (referring to the user intention with respect to the plugin action).

To remove this confusion, the compatibility tool should label the errors according to the action of the plugin, and not according to the action of the user. This entails updating the error labels this way:

- `New Accepted` to `New, Removed`
- `New Rejected` to `New, Kept`
- `Accepted` to `Removed`
- `Rejected` to `Kept`

This is only a change in labels and internal resource labeling.

westonruter:
See also https://github.com/ampproject/amp-wp/issues/1741

Does this terminology change work for "Statuses"? 

Since the underlying entity is the "validation error", does it make sense that a "validation error is removed" and a "validation error is kept"? Or does removed/kept make more sense if the underlying entity is the "invalid markup"?

amedina:
I see that point. Still we are on the right track. Let's flesh out things a little more. The ultimate goal should be two fold: 

1. Decouple `plugin actions` from `user actions`
2. Maintain consistency on the plugin side so that labels and statuses refer to either plugin actions or error-specific labels.

The status of an validation error as `removed` captures the fact that the error is not present in the generated content. That makes sense for status labels: this is an error which has been removed from the content. Similarly, the status of an error can be `Kept`, meaning it is still present in the generated content. 

There are no other statuses for errors; that is, there is not a 'fixed' status, or anything else. And this approach satisfies conditions (1) and (2) above. 

WDYT?

westonruter:
As discussed, it may be better to speak of “Retained” rather than “Kept”.

Something else to consider is that if:

[...]

PR #2024, opened 2019-03-26
[WIP] Enhance labeling of validation error to match plugin actions

amedina:
Addresses [#2023](https://github.com/ampproject/amp-wp/issues/2023)


westonruter:
As discussed, it may be better to speak of “Retained” rather than “Kept”.

Something else to consider is that if:

* ❌ Rejected ➡️ Retained/Kept
* ✅ Accepted ➡️ Removed

Do these icons still make sense?

westonruter:
Closing as this has been implemented in another PR.

Issue #2011, opened 2019-03-21
Editing/removing templates (Reusable blocks)


miina:
AC: User can save a Page as a template which will then appear with other pre-defined templates;
AC: User can remove and modify the custom templates;
AC: User can see only the templates added within the AMP Story, other Reusable blocks won't be available.

- [x] Ensure "Save as template" is in context menu for a AMP story page, and instead of it being Saving as Reusable Block (See also #2029 which might have this partly implemented)
- [ ] Make sure that Gutenberg only allows reusable blocks to be inserted at points where they are allowed.
- [ ] Make sure that the user can modify templates, the relevant blocks, settings, etc. have to exist for the edit page.
- [x] Make sure that only the AMP Stories related Reusable blocks show up in the Inserters.

westonruter:
> * instead of it being Saving as Reusable Block

I don't think we should disallow saving a page as a reusable block. I can imagine it would be useful to be able to have a page re-used across multiple stories, with changes automatically kept in sync.

Still seems like reusable blocks provide the underlying data model we need. Could we not re-use it but then mark them as templates (e.g. adding some postmeta to the `wp_block` instances) to prevent them from being deleted and/or to present them in the list of templates to insert? At the same time, filtering down the list of reusable blocks to just pages would be simple even without such a postmeta flag.

miina:
> instead of it being Saving as Reusable Block

I think that what was meant here was just the text change (this point was copied from the note added by Thierry).

Another "issue" for reusable blocks is that they lock the template meaning that the added blocks' set can't be modified once inserted to the editor.

It seems like we could use the reusable blocks' logic for creating templates, however, for displaying and inserting to the editor and displaying we would need adjustments or different logic -- instead of inserting a locked template we would just need to insert a set of blocks, and instead of displaying an icon, we would need to display the preview in the "Template Inserter".

Yes, it seems like also the base "static" templates could be using the `wp_block` post type behind the scenes with an additional post meta which determines that it's locked (the template can't be deleted and perhaps hide it from the Edit Post list).

Note that depending on how much is publicly available in the API we might need to create our own copy of the logic anyway even for small adjustments.

westonruter:
> Another "issue" for reusable blocks is that they lock the template meaning that the added blocks' set can't be modified once inserted to the editor.

They can if you convert it into a regular block:

![image](https://user-images.githubusercontent.com/134745/54843345-5fa30980-4c91-11e9-9c5e-421676048a23.png)

[...]

Issue #2010, opened 2019-03-21
[AMP Stories] Create the base templates for Travel and Fandom


miina:
AC: User can select between X (to be selected from the designs) single-page templates which are pre-populated with blocks, images and other data, so that they make sense in the context of Travel / Fandom;

- [x] Create templates in the editor based on design, use the HTML source to create templates for the "Template Inserter";
- [x] Auto-import pre-existing templates if the templates have never been imported before.

amedina:
Follow up from team meeting...

The goal here would be to **abstract the components** that would form a Fandom template. This encompasses:

- Consider the case of **multi-page templates** (used when starting a story), and **single-page templates** (used at any point during story creation, possibly with special cases for first page, middle page, last page (bookend)
- For Travel/Fandom, what would the **building blocks** commonly found in travel or fandom stories
- Which **combinations of them** would make sense from a UX POV? Each of these combinations would form a template page.
- Which **combination of template pages** would go well together? Each of these sets of pages would form a multi-page templates.

@miina @swissspidy @westonruter @dawidmlynarz @kienstra

csossi:
Any instructions re: testing for this? Not sure what I'm looking at here

miina:
**Testing Instructions**

- When clicking on the Template Inserter, 4 templates have to be available. Template Inserter:
![Screen Shot 2019-05-09 at 9 49 11 AM](https://user-images.githubusercontent.com/3294597/57436354-cb403680-723f-11e9-8edd-9a2c6ccfb17d.png)

- It has to be possible to insert a blank page from the template inserter, too.
- When clicking on any of the templates in the template inserter, the template with all the blocks that can be seen in template inserter have to be inserted to the editor.
- The placement of the blocks should be the same as in the preview within the template inserter, however, no images should be inserted, and there should be placeholders instead of the texts.

Let me know if there are any doubts. cc @csossi

csossi:
Thanks, @miina 

Can anything be done regarding these two templates? There's a purple vertical bar and copy overlapping it that looks confusing:
![image](https://user-images.githubusercontent.com/16945664/57665567-574cb800-75ca-11e9-918b-37426a1ded04.png)

![image](https://user-images.githubusercontent.com/16945664/57665588-69c6f180-75ca-11e9-9ca6-69503004ccf6.png)




- 4 templates available plus blank page:
[...]

PR #2029, opened 2019-03-26
[AMP Stories] Add template Inserter

### LINKED ISSUE #1902: [AMP Stories] Show list of templates to use when inserting a new blank page

miina:
Closes #1902.

This PR adds template inserter which at the moment displays all reusable blocks which have AMP Story Page as the root block.

Note that at this moment it is possible to create only single Page templates.

- [x] Show list of story page templates in inserter along with the blank page.
- [x] Show preview of pages.
- [x] The template has to be inserted to the editor as regular blocks so that they can be edited/removed, etc.
- [x] Allow adding reusable blocks within AMP Story.
- [x] Style properly.


miina:
@swissspidy This is still WIP but let me know if you have any initial thoughts which might influence the direction of the approach.

assets/src/components/template-inserter/block-preview.js
```diff
@@ -0,0 +1,44 @@
...
+function BlockPreview( props ) {
+	return (
+		<button onClick={ props.onClick } className="components-button editor-block-preview block-editor-block-preview">
+			<BlockPreviewContent { ...props } />
+		</button>
+	);
+}
+
+export function BlockPreviewContent( { name, attributes } ) {
+	// @todo Importing this outside of the function causes error for some reason.
```
swissspidy: This is because the new block-editor package is not yet listed in `wpDependencies` in `webpack.config.js` nor in `package.json`.

You can still use `@wordpress/editor` for now. It just proxies to the new package behind the scenes for BC.

Alternatively, since I am probably going to merge #2000 today and doing some rewriting there from `@wordpress/editor` to `@wordpress/block-editor`, you could then just merge in the latest changes from the amp-stories-redux branch.

assets/src/components/template-inserter/index.js
```diff
@@ -0,0 +1,163 @@
...
+		if ( ! this.isComponentMounted ) {
+			return;
+		}
+
+		if ( null !== this.state.reusableBlocks ) {
+			this.setState( { reusableBlocks: null } );
+		}
+
+		// @todo We only need reusable blocks that can be used by AMP Stories.
+		const blocksRequest = this.lastRequest = apiFetch( {
```
swissspidy: _Thinking out loud here_

One downside of doing the API request in the component is that we can't easily use the component twice without duplicating requests.

I was looking a bit at the `Inserter` and `InserterMenu` components in Gutenberg and how they use [`fetchReusableBlocks`](https://github.com/WordPress/gutenberg/blob/828fe40798b4d9b4a4d344b68eb0e2f62632ee8c/packages/editor/src/store/effects/reusable-blocks.js#L41-L99) and some other functions to do the data fetching. Are there perhaps some commonalities that we could re-use (maybe in combination with an [`api-fetch` middleware](https://wordpress.org/gutenberg/handbook/designers-developers/developers/packages/packages-api-fetch/#middlewares) to do the filtering) or copy over to our own data store?

> We only need reusable blocks that can be used by AMP Stories.

This goes the other way around too: In all other editors we don't want to show AMP Stories templates. `?search` doesn't work for that. An easy solution could be using a taxonomy behind the scenes. Post meta could work too, but doesn't scale well.

miina: > ?search doesn't work for that. An easy solution could be using a taxonomy behind the scenes. Post meta could work too, but doesn't scale well.

Agreed, `?search` was added more of a placeholder for preventing fetching all the reusable blocks which would then just cause errors on insertion or due to unregistered block types. I was thinking that implementing taxonomy / meta could be done either in #2010 / #2011 since those PR-s involve creating template posts.

>Are there perhaps some commonalities that we could re-use (maybe in combination with an api-fetch middleware to do the filtering) or copy over to our own data store?

Actually, we could also use the default experimental dispatch/selector for getting all reusable blocks and filter those. I thought initially that this might break the logic due to potentially including unregistered blocks, however, it looks like allowing `core/template` as an allowed block should resolve that issue. I'll make the change to test it out and then if it seems like it would make sense to manage the reusable blocks in our own data store then we can rework that later. Thoughts?

miina: This would mean though that we'd need to rework it anyway once taxonomy comes in. So yes, perhaps it does make sense to move it to our own data store.

miina: Alternatively we could just filter by having `amp-story-page` as the main block of the template, and not allow anything else, maybe just that could make sense for now, too. Thoughts?

assets/src/components/template-inserter/index.js
```diff
@@ -0,0 +1,157 @@
...
+
+		this.onToggle = this.onToggle.bind( this );
+
+		this.state = {
+			reusableBlocks: null,
+		};
+	}
+
+	componentDidMount() {
+		this.props.fetchReusableBlocks();
```
miina: @swissspidy I reworked this not to use API fetch here, however, we'll need to change it again once we get to categories.

Also looked into using the local store together with API but didn't come to a reasonable solution at this moment. Maybe for now we could actually use the API fetch as it was before and improve it later.

I'm thinking of improving the style to be acceptable and then leave the rest to the next PR-s / issues, potentially reverting to using API fetch for now. Thoughts?

miina:
@amedina @swissspidy This should be ready for testing. Note that this is the basic implementation, and search, editing, setting up the initial default templates etc. will be added within other PR-s.

The "Save as Template" option has been added to the Block settings, this replaces the "Save as Reusable Block" setting, see it in action here:

template

assets/src/constants.js
```diff
@@ -66,6 +70,31 @@ export const BLOCK_ICONS = {
 	'amp/amp-story-page': <svg id="story-page-icon" viewBox="0 0 24 24"><g id="icon" fill="#181D21"><path id="page" d="M18.4 21H5.6V3h7.8l5 4.9V21zM7.1 19.5h9.8V8.6l-4-4.1H7.1v15z" /><path id="corner" d="M11.5 5.4v4.3h4.4" /></g></svg>,
 };
 
+export const ICONS = {
```
swissspidy: Since we're adding some new icons here, can we actually move all these (including the one from `BLOCK_ICONS`) to `assets/images` and import them here like we do with the font name SVGs? 

This way we benefit from the SVG loader in the webpack config, which automatically optimizes the SVGs too.

assets/src/components/template-inserter/index.js
```diff
@@ -102,16 +102,20 @@ class TemplateInserter extends Component {
 											className="amp-stories__blank-page-inserter editor-block-preview__content block-editor-block-preview__content editor-styles-wrapper"
 										/>
 									</div>
-									{ storyTemplates && storyTemplates.map( ( item ) =>
-										<BlockPreview
+									{ storyTemplates && storyTemplates.map( ( item ) => (
+										<Button
```
miina: @swissspidy There's an issue with wrapping this inside a button -- if the block that's being previewed includes a button then this means a button within a button which is not allowed.

swissspidy: Yes, I saw your comment in the previous file. `core/button` is not in `ALLOWED_CHILD_BLOCKS` though, so I am not sure in which scenario this could happen.

miina: I think there's an inserter button or something like within a Page block which is also included in the preview.

assets/src/constants.js
miina: Saw that you already did that, thank you!

assets/src/components/template-inserter/index.js
swissspidy: Ah, I see! This reminds me of https://github.com/WordPress/gutenberg/pull/14241 where I previously suggested adding an option to disable the inserter/appender completely. Once again proof that such a feature is needed. I hope I can manage to work on a PR for that soon.

assets/src/plugins/template-menu-item.js
```diff
@@ -0,0 +1,65 @@
...
+	// @todo Display notice.
+	saveReusableBlock( reusableBlock.id );
+};
+
+/**
+ * Register plugin for adding Templates (Reusable blocks) without converting the page itself to a reusable block.
+ */
+export function registerTemplateSaveMenuItem() {
+	// @todo Change icon.
+	// @todo allowedBlockNames restriction is not working properly.
```
swissspidy: @miina FYI this is a bug in Gutenberg. The documentation says `allowedBlockNames` but it actually checks for `allowedBlocks`. I submitted a PR with the fix: https://github.com/WordPress/gutenberg/pull/14741

assets/src/components/template-inserter/index.js
swissspidy: Makes sense to me. Let's stick with the current approach and iterate in follow-up PRs.

assets/src/plugins/template-menu-item.js
miina: Great, thanks for the fix PR, too!

swissspidy:
In case anyone is wondering why the "Save as Template" menu item doesn't look so great, I reported the bug earlier and it is being fixed: https://github.com/WordPress/gutenberg/pull/14745

assets/src/components/template-inserter/index.js
miina: @swissspidy: I'm seeing these errors now when using templates, is this something you were planning working on soon? Otherwise, perhaps we should revert using the `button` element for now?
Screen Shot 2019-04-02 at 7 44 55 PM

swissspidy: Yes, it's on my list, although I don't see it as a pressing issue. Fixing it upstream takes some time anyway as we'd have to wait for the next Gutenberg release in 1-2 weeks.

A workaround could be the remove the DOM node manually, but I'm not sure that'd be worth the effort.

PR #2031, opened 2019-03-27
Allow cropping of the AMP Story featured image 

### LINKED ISSUE #2030: Auto-cropping of excessive featured images
##### *Editing features*
------
- [x] Allow cropping of image
- [x] Possibly allow cropping during preview

kienstra:
* If the featured image is larger than 696 x 928, it gives the option to crop it
* Mainly copied from the `customize-controls.js` functionality for the Site Icon

Closes #2030


assets/src/components/with-cropped-featured-image.js
```diff
@@ -0,0 +1,264 @@
...
+ * Only applies to the MediaUpload in the Featured Image component, PostFeaturedImage.
+ * Suggests cropping of the featured image if it's not 696 x 928.
+ * Mostly copied from customize-controls.js.
+ *
+ * @param {Function} InitialMediaUpload The MediaUpload component, passed from the filter.
+ * @return {Function} The wrapped component.
+ */
+export default ( InitialMediaUpload ) => {
+	/**
+	 * Mostly copied from customize-controls.js, with slight changes.
```
kienstra: This mainly copies functions from [wp.customize.CroppedImageControl](https://github.com/WordPress/wordpress-develop/blob/master/src/js/_enqueues/wp/customize/controls.js#L4495).

It would probably be better to use functions from that directly, instead of mainly copying them. Like with `wp.customize.CroppedImageControl.prototype.mustBeCropped()`.

But `wp.customize` doesn't look to be available in the block editor.

assets/src/components/with-cropped-featured-image.js
```diff
@@ -0,0 +1,264 @@
...
+	 */
+	return class FeaturedImageMediaUpload extends InitialMediaUpload {
+		/**
+		 * Constructs the class.
+		 */
+		constructor() {
+			super( ...arguments );
+
+			// This class should only be present in the MediaUpload for the Featured Image.
+			if ( this.props.modalClass && 'editor-post-featured-image__media-modal' === this.props.modalClass ) {
```
kienstra: It's not ideal to [use the modalClass](https://github.com/WordPress/gutenberg/blob/94f54892c214ebd014ec126a9575b7e902198ad9/packages/editor/src/components/post-featured-image/index.js#L50) to ensure that this `MediaUpload` is from the `PostFeaturedImage` component.

I couldn't find a better way to do this, though maybe there is one.

One alternative would be to filter the [PostFeaturedImage](https://github.com/WordPress/gutenberg/blob/94f54892c214ebd014ec126a9575b7e902198ad9/packages/editor/src/components/post-featured-image/index.js#L132) component. But that would probably violate DRY really badly.

kienstra:
**Request For Review**

Hi @swissspidy,
Could you please review this PR when you have a chance?

Expected Behavior:
1. In an AMP story, click "Set featured image"
2. Upload an image that doesn't have the exact dimensions of `696 x 928`, like http://lorempixel.com/1200/1200/technics/
3. Click "Select"
4. Expected: the Media Library displays the "Crop Image" state, and allows you to crop it
5. Expected: on clicking "Crop Image," it's cropped and added as the featured image

Even if the image is smaller, like [400 x 400](http://lorempixel.com/400/400/technics/), you can still crop it and have the full `696 x 928` image appear as the `poster-portrait-src` on the front-end.

featured-image-crop

Thanks, Pascal! Have a great weekend.

includes/class-amp-story-post-type.php
```diff
@@ -166,6 +166,9 @@ public static function register() {
 		// Register the Latest Stories block.
 		add_action( 'wp_loaded', array( __CLASS__, 'register_block_latest_stories' ), 11 );
 
+		// The AJAX handler for when an image is cropped and sent via POST.
+		add_action( 'wp_ajax_custom-header-crop', array( __CLASS__, 'crop_featured_image' ) );
```
swissspidy: Can we use a different Ajax action name here? This would help distinguish it from the one used by `Custom_Image_Header`.

assets/src/components/with-cropped-featured-image.js
swissspidy: Unfortunately I am not really an expert on `wp.media` 😕 

I think in this case it is fine to copy over the functions. Enqueueing customizer scripts in the editor would probably result in too much bloat and maybe even some unwanted side effects.

assets/src/components/with-cropped-featured-image.js
swissspidy: What about filtering `PostFeaturedImage` and only within that callback add the `MediaUpload` filter.

The `PostFeaturedImage` filter callback itself would just register that `MediaUpload` filter and return the original, unchanged component.

At least in PHP land that is possible. Not sure about the editor though.

But the `modalClass` check isn't too bad really, we could probably just keep it this way.

includes/class-amp-story-post-type.php
```diff
@@ -1115,4 +1118,110 @@ public static function get_featured_image_minimum_height( $posts ) {
...
+			unset( $error );
+		}
+
+		$image_type = ( $size ) ? $size['mime'] : 'image/jpeg';
+		$object     = array(
+			'ID'             => $parent_attachment_id,
+			'post_title'     => basename( $cropped ),
+			'post_mime_type' => $image_type,
+			'guid'           => $url,
+			'context'        => 'custom-header',
```
swissspidy: Can we change this to `amp-story-poster` or something? Where is this used?

swissspidy:
Thanks @kienstra! The cropping seems to work well so far.

Not a huge fan of upscaling low-res image, as it reduces quality and half of the work from #1998 is rendered moot (users will never see the minimum dimensions warning).

The one big thing I noticed:

1. Create new story
2. Click on "Set featured image"
3. Upload large image (e.g. [1600x1200](http://lorempixel.com/1600/1200/technics/))
4. Confirm crop suggestion
5. Preview story
6. Notice odd image sizes in the HTML output:

```html
<amp-story
  standalone
  publisher-logo-src=".../wp-admin/images/wordpress-logo.png"
  publisher="AMP"
  title=""
  poster-portrait-src=".../wp-content/uploads/2019/03/cropped-download.jpeg"
  poster-square-src=".../wp-content/uploads/2019/03/cropped-download-696x928.jpeg"
  poster-landscape-src=".../wp-content/uploads/2019/03/cropped-download-696x696.jpeg"
>...</amp-story>
```

So the square image is actually in portrait format, and the landscape image is square.

assets/src/components/with-cropped-featured-image.js
kienstra: Yeah, good point.

assets/src/components/with-cropped-featured-image.js
kienstra: OK, thanks

includes/class-amp-story-post-type.php
kienstra: Yeah, it's strange having this name that references the custom image header.

But it looks like the [wp.ajax.post call](https://github.com/WordPress/wordpress-develop/blob/6b366c6620fdd5960cedcdf80955966a715efa82/src/js/media/controllers/cropper.js#L158) that uses the `'custom-header-crop'` action is in another file, so there's not an easy way to change that.

includes/class-amp-story-post-type.php
kienstra: Sure, https://github.com/ampproject/amp-wp/pull/2031/commits/d2d69981d3464e764f5c976013af7ebcb0c11537 changes the context to `amp-story-poster`. 

I'm not sure where it's used. The only thing I could find in [customize-controls.js](https://github.com/WordPress/wordpress-develop/blob/c80325658f85d24ff82295dd2d55bfdf789f4163/src/js/_enqueues/wp/customize/controls.js#L4742) that might relate to this is:

```js
wp.ajax.post( 'crop-image', {
	nonce: attachment.nonces.edit,
	id: attachment.id,
	context: 'site-icon',
````

But that's sending a POST request, not responding to one.

kienstra:
>Not a huge fan of upscaling low-res image, as it reduces quality and half of the work from #1998 is rendered moot (users will never see the minimum dimensions warning).

Good point. I could change the `mustBeCropped()` function to avoid cropping if it's not large enough to produce a full `696 x 928` image.


>So the square image is actually in portrait format, and the landscape image is square.
Great catch. I think I figured out how to fix that. It's attempting to find sizes of the cropped image, but I could look at sizes of the source of the cropped image instead.

kienstra:
>Not a huge fan of upscaling low-res image, as it reduces quality and half of the work from #1998 is rendered moot (users will never see the minimum dimensions warning).

Right, I agree that this wasn't a great approach.

https://github.com/ampproject/amp-wp/pull/2031/commits/3d87fe6ef7ff0c6c1e234740435f8fd0e514d4a7 prevents cropping if the featured image selected is less than `696 x 928`. So it won't create a distorted image from a smaller image. And the notice will appear.

>So the square image is actually in portrait format, and the landscape image is square.

https://github.com/ampproject/amp-wp/pull/2031/commits/e87f7d44650f98589172194c975d8ee51ba9a128 should fix this. It gets the `poster-square` and `poster-landscape` images from the (larger) source of the cropped featured image, not the cropped featured image.

It wasn't possible to get a `928 x 928` image from a cropped `696 x 928` featured image.

kienstra:
**Changed The Notice**

If it's alright, I changed the notice for the featured image in both the featured image section and the pre-publish panel.

# Before
before-notice

# After
after-notice

Now that there's cropping of the image, I think it's most important to show the notice if it's less than `696 x 928`. 

The other dimensions, like for `poster-square-src`, come from the source of the cropped image. The cropped image (`696 x 928`) isn't big enough to produce a `928 x 928` image.

Issue #1845, opened 2019-01-31
Automatically extract image for Schema.org metadata from content if no featured image selected


westonruter:
This came up on https://wordpress.org/support/topic/image-amp/

When a blog post has an image added to the content but does not have any image assigned as the featured image, then GSC will complain about there not being an image supplied. We should automatically scrape the `post_content` for the first image to use in this case.

Note that Jetpack actually has some similar logic to do this in filtering `amp_post_template_metadata`: https://github.com/Automattic/jetpack/blob/c126df3d52a9e7e35cb1cefcf64281448d8becd9/3rd-party/class.jetpack-amp-support.php#L129-L174

As with #1812, we should only use an image if it satisfies the required dimensions.

PR #2038, opened 2019-03-29
Improve featured image notices

swissspidy:
This is a follow-up bug fix for #1998.

Unfortunately I only realized after merging that the "Selecting a featured image is required." was being displayed even for regular posts, not just stories.

Screenshot 2019-03-29 at 13 25 15

However, for single posts we only want a notice when an image is set but too small.

With the existing functions it didn't seem to be that clear how to change the logic without making it harder to read. So I ended up going down the rabbit hole and turning them into separate components and higher-order components at the same time. This way they fit a bit better into the rest of the code base there I think.


westonruter:
Actually, doesn't this warning make sense for regular posts as well? If you don't have a featured image set then Google Search Console issues warnings about not having an image. 

Maybe a notice would be more appropriate than a warning, since it is a hard error for AMP Stories but only a warning for other AMP pages.

A warning should be shown regardless if the selected image does not have adequate dimensions to satisfy the Google Search requirements, as again, a GSC will issue warnings.

swissspidy:
Good point. The wording should be a bit different then.

I mainly flagged this here because it's a change from previous behavior.

A notice for regular AMP pages could make sense.

westonruter:
See https://github.com/ampproject/amp-wp/issues/1812 for how the notice could apply for non-story posts.

swissspidy:
What about this:

Screenshot 2019-03-29 at 17 09 56

westonruter:
Humm. I don't think that is quite accurate. The page will still be indexed
and appear in search results. I think it's just that it won't appear in the
Search Carousel. Could you check with the Search team there as for why
specifically this is required?

But there are other reasons to encourage the featured image to be supplied,
like in social media shares. But the other reason to include the image
would be to avoid GSC issuing a warning.

All this being said, what if someone embeds the image in the post content
instead of adding as a featured image. I've seen that before and the AMP
plugin actually has some detection for that I believe. I'll check.

westonruter:
> All this being said, what if someone embeds the image in the post content instead of adding as a featured image. I've seen that before and the AMP plugin actually has some detection for that I believe. I'll check.

I got this reversed actually. The AMP plugin _in the Classic/Reader mode_ only, will skip outputting the featured image in the template if that same image exists inside the post content:

https://github.com/ampproject/amp-wp/blob/ffd66f6be06d754dbbf525d988d1a966965ed77d/includes/templates/class-amp-post-template.php#L344-L351

The related issue is then https://github.com/ampproject/amp-wp/issues/1845 which is to automatically extract image for Schema.org metadata from content if no featured image selected. This is actually similar to what is happening now for the Story editor actually. But this could be done at the PHP side of things, to rather scrape the content for a suitable image to use for Schema.org metadata if none was supplied. This is something that [Jetpack is already doing](https://github.com/Automattic/jetpack/blob/c126df3d52a9e7e35cb1cefcf64281448d8becd9/3rd-party/class.jetpack-amp-support.php#L129-L174).

In any case, let's say that the AMP plugin incorporates that Jetpack logic to scrape the content for an image to use for the Schema.org metadata. Then this notice wouldn't need to be displayed if there was such a suitable image in the content.

----

All of this being said (again), I suppose this is perhaps out of scope for what needs to be addressed first. For now, we can just assume the featured image is the only source for the Schema.org metadata. In that case, we should have a notice like you have depicted, though the text should reflect that it will help the post appear optimally when shared and when it is shown in search results.

Does Yoast have anything in it to encourage the supplying of the featured image? Wonder if we might be duplicating a warning.

miina:
> Does Yoast have anything in it to encourage the supplying of the featured image? Wonder if we might be duplicating a warning.

Yoast Glue has the option to set default featured image, however, it doesn't seem to display a warning.

swissspidy:
> The related issue is then #1845 which is to automatically extract image for Schema.org metadata from content if no featured image selected. This is actually similar to what is happening now for the Story editor actually.

Interesting. Is a PHP solution actually preferable for Stories as well?

> The text should reflect that it will help the post appear optimally when shared and when it is shown in search results.

Will improve the wording 👍

swissspidy:
I didn't want to make the notice unnecessarily long by adding specific examples. I ended up with `Selecting a featured image is recommended for an optimal user experience.`.

Screenshots:

Screenshot 2019-04-01 at 16 46 25
Screenshot 2019-04-01 at 16 46 16

assets/src/components/index.js
```diff
@@ -7,18 +7,18 @@ export { default as Shortcuts } from './shortcuts';
 export { default as StoryBlockDropZone } from './story-block-drop-zone';
 export { default as StoryBlockMover } from './block-mover';
 export { default as FontFamilyPicker } from './font-family-picker';
+export { default as PrePublishPanel } from './pre-publish-panel';
+export { default as StoryPrePublishPanel } from './story-pre-publish-panel';
```
miina: I wonder if we could have just one component which accepts different params?

assets/src/amp-story-editor-blocks.js
```diff
@@ -327,7 +327,12 @@ store.subscribe( () => {
 	}
 } );
 
-registerPlugin( 'amp-story-featured-image-pre-publish', { render: withPrePublishNotice } );
+registerPlugin(
+	'amp-story-featured-image-pre-publish',
+	{
+		render: StoryPrePublishPanel,
```
miina: Since the story actually won't display without the featured image, then I'm wondering if we should disable the `Publish` button completely if the featured image is not set? Thoughts?

swissspidy: > Since the story actually won't display without the featured image

It actually will be displayed, thanks to #1998 and #2031. When there's absolutely no image whatsoever in a story, a fallback image is being used as the poster, allowing you to still view it.

But yeah we could experiment with disabling save in this case.

miina: Oh, right! Completely forgot about the featured image related updates 💡

westonruter: > Since the story actually won't display without the featured image, then I'm wondering if we should disable the `Publish` button completely if the featured image is not set? Thoughts?

Added a new card for this: https://github.com/ampproject/amp-wp/projects/10#card-19655884

assets/src/helpers.js
```diff
@@ -472,3 +449,16 @@ export const hasMinimumStoryPosterDimensions = ( media ) => {
...
+ * Whether the image has the minimum width for a featured image.
+ *
+ * This should have a width of at least 1200 pixels
+ * to satisfy the requirement of Google Search for Schema.org metadata.
+ *
+ * @param {Object} media A media object with width and height values.
+ * @return {boolean} Whether the media has the minimum dimensions.
+ */
+export const hasMinimumFeaturedImageWidth = ( media ) => {
+	return ( media.width && media.width >= 1200 );
```
westonruter: The 1200 value is being used for the width, but there are more requirements than this we should also account for. [Namely](https://developers.google.com/search/docs/data-types/article#article_types):

> * Images should be at least 1200 pixels wide. ✅ 
> * For best results, provide multiple high-resolution images (minimum of 800,000 pixels when multiplying width and height) with the following aspect ratios: 16x9, 4x3, and 1x1.
> * Images must be in .jpg, .png, or .gif format.

So we should also address these later.

----

I see that `hasMinimumStoryPosterDimensions` is actually accounting for more of [its requirements](https://www.ampproject.org/docs/reference/components/amp-story#new-metadata-requirements), so that is good:

> Each poster image should meet the recommended minimium size:
> * Portrait: 696px x 928px
> * Landscape: 928px x 696px
> * Square: 928px x 928px

Nevertheless, it is good that we are requiring 1200 for the featured image since then it can be re-used for both the poster _and_ the Schema.org metadata. The cropping I believe is handled elsewhere.

westonruter: We should also update the Schema.org markup to output multiple images in the desired aspect ratios:

```json
{
  "@context": "https://schema.org",
  "@type": "NewsArticle",
  "image": [
    "https://example.com/photos/1x1/photo.jpg",
    "https://example.com/photos/4x3/photo.jpg",
    "https://example.com/photos/16x9/photo.jpg"
  ]
}
```

assets/src/components/pre-publish-panel.js
```diff
@@ -0,0 +1,49 @@
...
+import { withSelect } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Conditionally adds a notice to the pre-publish panel for the featured image.
+ *
+ * @return {Function} Either a plain pre-publish panel, or the panel with a featured image notice.
+ */
+const PrePublishPanel = ( { featuredMedia, validationCallback, missingMediaMessage, invalidMediaMessage, status } ) => {
+	if ( featuredMedia && validationCallback( featuredMedia ) ) {
```
westonruter: @swissspidy The `featuredMedia` here is the attachment ID, but it is getting passed into `hasMinimumFeaturedImageWidth` which expects a media object.

assets/src/helpers.js
swissspidy: > Nevertheless, it is good that we are requiring 1200 for the featured image since then it can be re-used for both the poster and the Schema.org metadata.

There are two different requirements/checks:

* 696x928 for AMP Stories
* 1200 minimum width for regular posts

Did you mean that for AMP Stories we should also require 1200 minimum width?

westonruter: Oh, yes, actually. Sorry, I rewrote that comment a couple times as I discovered some more things. So since there is a separate function specifically for stories, then the 1200 constraint isn't applying. So yes, we should also enforce AMP stories gave a featured image of at least 1200px in order to satisfy the poster requirements and the metadata requirements.

assets/src/components/pre-publish-panel.js
swissspidy: Thanks! Somehow I missed this.

assets/src/helpers.js
swissspidy: Cool, makes sense. I'll bump the width requirement in `hasMinimumStoryPosterDimensions` then.

westonruter:
The cropping tool for the featured image is now causing us problems, since it forces the image to be 696px wide, and yet the featured image is required to be 1200px wide.

swissspidy:
> The cropping tool for the featured image is now causing us problems, since it forces the image to be 696px wide, and yet the featured image is required to be 1200px wide.

Hmm in order to maintain the aspect ratio, I guess the cropping needs to be done with 1200px width as well. I can add this tomorrow.

assets/src/helpers.js
```diff
@@ -434,41 +433,62 @@ export const getPercentageFromPixels = ( axis, pixelValue ) => {
...
+ * @param {Object} media A media object with width and height values.
+ * @return {boolean} Whether the media has the minimum dimensions.
+ */
+export const hasMinimumFeaturedImageWidth = ( media ) => {
+	if ( ! media || ! media.width || ! media.height ) {
+		return false;
+	}
+	const minFeaturedImageDimensions = getMinimumFeaturedImageDimensions();
+
+	return media.width >= minFeaturedImageDimensions.width && media.height >= minFeaturedImageDimensions.height;
```
westonruter: Instead of returning a `bool` I wonder if it would be better to return a string for the error message. An empty string or null could mean there is no error. By returning the error message, it could then indicate why specifically the image is not adequate, and it can indicate what the dimensions are for the `media` supplied.

swissspidy: Besides the message itself it would be nice to return the affected data (e.g. actual width vs required width) as well, so it could be more like returning/throwing an error object.

Will there be more error messages than "image too small" and "no image set" in the future? Perhaps "wrong file format" and "file too big"? This could also affect changes to this function here.

westonruter: Yes, exactly. The other constraint not currently being accounted for is that the image must be a JPEG, PNG, or GIF. So that is something this function should check for. In the future the requirements will surely change, so if we keep it flexible then it adapt. In that regard, how about having this return an _array_ of errors? Then we can return an error regarding the file type, insufficient width, and insufficient height. The user then can fix all at once and not have to discover after fixing one error that there is yet another error to fix.

assets/src/amp-story-editor-blocks.js
```diff
@@ -331,7 +333,21 @@ store.subscribe( () => {
...
-registerPlugin( 'amp-story-featured-image-pre-publish', { render: withPrePublishNotice } );
+registerPlugin(
+	'amp-story-featured-image-pre-publish',
+	{
+		render: () => {
+			return (
+				<PrePublishPanel
+					validationCallback={ hasMinimumStoryPosterDimensions }
+					missingMediaMessage={ __( 'Selecting a featured image is required.', 'amp' ) }
+					invalidMediaMessage={ __( 'The featured image must have minimum dimensions of 696 by 928 pixels.', 'amp' ) /* @todo This is not accurate. The width is now 1200 pixels. */ }
```
westonruter: I think we should introduce a `getMinimumStoryPosterDimensions` function similar to `getMinimumFeaturedImageDimensions` and then use the returned dimensions in a string format to avoid duplicating the numbers across the codebase.

assets/src/components/with-cropped-featured-image.js
```diff
@@ -9,7 +9,7 @@ import { dispatch } from '@wordpress/data';
  */
 import { now } from 'lodash';
 
-const EXPECTED_WIDTH = 696;
+const EXPECTED_WIDTH = 696; // @todo This may be wrong as 1200px is expected for featured image?
```
westonruter: Note todo.

assets/src/components/with-story-featured-image-notice.js
```diff
@@ -2,14 +2,49 @@
...
+		return ( props ) => {
+			const { media } = props;
+
+			if ( media && hasMinimumStoryPosterDimensions( media.media_details ) ) {
+				return <PostFeaturedImage { ...props } />;
+			}
+
+			const message = ! media ?
+				__( 'Selecting a featured image is required.', 'amp' ) :
+				__( 'The featured image must have minimum dimensions of 696 by 928 pixels.', 'amp' ); // @todo This is not accurate. The width is now 1200 pixels.
```
westonruter: Note todo

assets/src/helpers.js
```diff
@@ -434,41 +433,62 @@ export const getPercentageFromPixels = ( axis, pixelValue ) => {
...
+ * to satisfy the requirement of Google Search for Schema.org metadata.
  *
  * @param {Object} media A media object with width and height values.
  * @return {boolean} Whether the media has the minimum dimensions.
  */
 export const hasMinimumStoryPosterDimensions = ( media ) => {
-	const minWidth = 696;
-	const minHeight = 928;
+	const minWidth = 696; // @todo Needs to incorporate Math.max( 696, getMinimumFeaturedImageDimensions().width );
+	const minHeight = 928; // @todo This needs to be Math.max( 928, getMinimumFeaturedImageDimensions().height ).
```
westonruter: Note todos

westonruter:
I can work to implement the changes for this tomorrow.

westonruter:
Related to this, it seems that if I select an image for the featured image and crop it, the resulting featured image is way to large for the post embed block and latest stories block.

swissspidy:
@westonruter Added some changes, let me know if that's close to what you had in mind.

westonruter:
On a story, I tried selecting a featured image from an 2700×1761 image and it offered to crop as:

image

The result was something stretched:

image

It's also unexpected that the warning remains since the image I selected is greater than 1200×928 pixels in even its minimum dimension (1761 > 1200). This appears to be because the cropped image is 696×928:

image

westonruter:
For a regular post, the notice appears as expected:

image

If I select an image that is too narrow, the message appears as expected:

image

But when selecting an image that is larger than required, the notice remains though it is empty:

image

westonruter:
# Image Size Requirements

Thinking through the image sizes that we need here…

## [Featured Image for Schema.org metadata](https://developers.google.com/search/docs/data-types/article#article_types)

> Images should be at least 1200 pixels wide.
> For best results, provide multiple high-resolution images (minimum of 800,000 pixels when multiplying width and height) with the following aspect ratios: 16x9, 4x3, and 1x1.

The minimum width is 1200px so the following image sizes are needed according to the specified aspect ratios:

* Widescreen: 1200×675 (16:9 aspect ratio)
* Square: 1200×1200 (1:1 aspect ratio)
* Fullscreen: 1200×900 (4:3 aspect ratio) 

## [AMP Story Posters](https://amp.dev/documentation/components/amp-story#poster-guidelines-(for-poster-portrait-src,-poster-landscape-src,-and-poster-square-src))

* Portrait: 696×928
* Landscape: 928×696
* Square: 928×928

## Video Poster

The recommended dimensions for an [AMP Video poster image](https://amp.dev/documentation/guides-and-tutorials/develop/amp_story_best_practices#specify-a-poster-attribute) are 720×1280. This is not the same dimensions or aspect ratio as the Portrait Poster, strangely.

## Story Fill Background Image

So given the above, what is the best dimensions for an image that fills an `amp-story-grid-layer`? The first page of the [Hello World](https://amp.dev/documentation/examples/introduction/stories_in_amp/preview/?format=websites) story has the dog image with intrinsic dimensions of 720×1279, so the same as the 720×1280 for the video poster. (The poster image for the dog running video is likewise 720 × 1276.) The third page, however, has a fill image with dimensions 960×1280 (a different aspect ratio).

----

# Image Selection Scenarios

When selecting and cropping an image, ideally we would re-use the same portrait image multiple places. However, the poster portrait image is too small to be used as as a layer fill image. And the Schema.org featured image needs to have a square or landscape orientation.

To satisfy all of the requirements it seems the source image must be at least **1200×1280**. Given such an image, resized crops can be made to satisfy all of the various dimensions. 

## Featured Image

Selecting a featured image should present a cropping tool for the Schema.org metadata _and_ another cropping tool for the portrait poster image. 

For the Schema.org image, we need a crop of 1200×900 for the 4:3 aspect ratio which can also be used for as-is for the story landscape poster, since 928×696 is the _minimum_ dimensions. The 1200×675 (16:9) image can then be a centered crop from this image.

A square 1200×1200 could be automatically cropped using the focal point of the selected 4:3 crop, either using that crop as the bounding box or if the crop was smaller than 1200px high then it would have to expand the canvas above and below the selected region.

## Featured Image for Post

When selecting a Featured Image for a non-story post, the same logic as for the Story could be used, except there would be no secondary crop for the poster dimension.

## Background Media for Story Page

When selecting an image for the background media, the media library should only present images that are at least 960×1280. Upon selection, the focal point can be chosen for the image in the block settings.

👉  _However_ there is a problem currently bytes are being wasted if a landscape image is chosen since only a portrait crop of that image is displayed. 

56011008-aab1aa80-5c9a-11e9-983c-abd4bc305074 (1)

Therefore, when selecting a background image, should this not also open a portrait cropping tool that allows for a 960×1280 image from a crop of the selected image? If this is the case, then _the focal point control would not be necessary._ ❓ 

Also, when selecting a background image and there is no featured image selected, if the selected image is selected which is 1200×1280, then this can automatically populate the featured image as well if it was not provided; in this case, no custom crops would be provided. If a user wants to pick a custom crop, then they'd need to change the image (or rather, remove the selected feature image and then re-select the same image as before? 🤔)

westonruter:
> Selecting a featured image should present a cropping tool for the Schema.org metadata _and_ another cropping tool for the portrait poster image.

I don't love the idea of having two separate cropping modals popping up.

What if instead of re-using the featured image field for the poster, we had _another_ poster image field alongside the featured image?

westonruter:
If we had separate featured image and poster image fields, the user could select two entirely separate images if they wanted to.

The poster image and the background image would be much more likely to be a 1:1 mapping, since they'd both be portrait. So selecting a poster image could automatically populate the background media, and vice-versa. In fact, perhaps we don't even need a poster image at all, come to think of it. What if the background image was the poster, period? Even in the case of the user wanting actually to select a video, AMP Stories also requires selecting a poster image:

image

The complication here would be that a user could create a story and not want any background image on the first page. So perhaps re-using the background media image (or background video poster) would not work in the general case for the story poster image. But at least these could populate each other upon selection as defaults.

The focal point picker provides more immediate feedback to tweak the position of the image, but I'm thinking that this is perhaps overkill. If a user realizes that the crop is bad, they could just select the image again and re-crop it. (Aside: When opening the media modal, the previously-selected image should ideally be re-selected so they don't have to find it again.)

westonruter:
I checked with Jon Newmuis about the recommended image sizes on the amphtml Slack and he said:

> I think we’re now recommending 9:16 aspect ratio (so, 720x1280), as phone manufacturers keep making phones taller than when we started this project.

swissspidy:
I really want to move this forward a bit as it is quite a big refactor at this point and has implications for #2176 as well due to the shared components.

Separating featured image and poster image sounds interesting, and I guess it makes sense. Would make many things easier.

assets/src/components/with-cropped-featured-image.js
```diff
@@ -1,19 +1,170 @@
...
+		if ( Math.abs( cropDetails.width - imgSelectOptions.minWidth ) < 10 ) {
+			cropDetails.width = imgSelectOptions.minWidth;
+		}
+		if ( Math.abs( cropDetails.height - imgSelectOptions.minHeight ) < 10 ) {
+			cropDetails.height = imgSelectOptions.minHeight;
+		}
+
+		// Force the resulting image to have the expected dimensions (scale down).
+		cropDetails.dst_width = imgSelectOptions.minWidth;
+		cropDetails.dst_height = imgSelectOptions.minHeight;
```
westonruter: Actually, we might as well let these be copies of the width/height. It is fine to have the featured image be _larger_ than the minimum.

westonruter:
I've fixed the issues with the featured image cropping tool, ensuring that a cropped image is not distorted. Also, the media library now prevents a user from selecting an image that is smaller than the minimum dimensions:

image

Now that the cropper is fixed, we now need to whether this featured image should be also used as the poster, and if so, whether a different crop is needed.

Also need to make sure that this tool works on non-story posts.

westonruter:
For a story I selected the Bison featured image and selected this crop:

> Screen Shot 2019-04-26 at 21 23 32

And this results in a story with this markup:

```html
<amp-story standalone
    publisher-logo-src="https://wordpressdev.lndo.site/content/uploads/2019/02/cropped-google-logo-1-3.png"
    publisher="WordPress Develop"
    title="Test"
    poster-portrait-src="https://wordpressdev.lndo.site/content/uploads/2019/04/cropped-American_bison_k5680-1-35-696x928.jpg"
    poster-square-src="https://wordpressdev.lndo.site/content/uploads/2019/04/cropped-American_bison_k5680-1-35-928x928.jpg"
    poster-landscape-src="https://wordpressdev.lndo.site/content/uploads/2019/04/cropped-American_bison_k5680-1-35-928x696.jpg">
```

So given that the originally-selected featured image was a landscape image cropped at (1200×928):

> featured-image

The `poster-portrait-src` crop (696×928) was created automatically from from the featured image:

> poster-portrait

The `poster-square-src` crop (928×928) was created automatically from from the featured image:

> poster-square

The `poster-landscape-src` crop (928×696) was created automatically from from the featured image:

> poster-landscape

In short, this all looks good for a story. Every image meets the requirements. The auto-generated crops from the featured image work well, as long as the user decides to center the image.

westonruter:
@swissspidy OK, I think this is good to merge. Please review my latest changes to the JS as you probably have a better way to organize.

westonruter:
We may need to revisit whether it is a good idea to present the cropping tool  for non-story posts.

Oh, and one more thing which may need your touch:

If I load the story editor in code editor mode, there is a JS error:

image

image

This may not be related to the changes here at all, but part of what you are working on in #2117.

swissspidy:
> We may need to revisit whether it is a good idea to present the cropping tool for non-story posts.

It was a bit unexpected, and perhaps it can be a bit unexpected for others too. Let's keep it in for now for some testing.

> This may not be related to the changes here at all, but part of what you are working on in #2117.

While I added a fix in #2117, this was there already before, I think.

assets/src/components/with-cropped-featured-image.js
```diff
@@ -255,7 +251,7 @@ export default ( InitialMediaUpload ) => {
 				data.height = height;
 			}
 
-			onSelect( data );
+			onSelect( data ); // @todo Does this do anything?
```
swissspidy: Well it passes the information to whatever `onSelect` callback prop that was passed to the component.

swissspidy:
The notices seem to work well. Let's merge this now and re-organize things when needed. Separate fields for poster image and featured image can be handled in separate PRs.

Issue #2013, opened 2019-03-21
[AMP Stories] Add any missing features necessary for templates


miina:
AC: All the features needed for the templates exist.

- [x] Look through the templates and log the features that we need to add for templates.
- [x] Implement any missing features.

Features needed:
- [x] Add border radius setting to Text block and Image block [Done, see #2055].
- [x] Add alignment options for the Text block.
- [x] Add gradient setting for Page. [Done, see #2046]

~~Add opacity Range control to allowed (child) blocks which use background color.~~ Moved to a separate issue:  #2058

miina:
Based on the template designs the following features seem to be missing:
- Adding opacity to block (this would also be useful with a Color Fill block -- a resizable block which can be just filled with color and then placed on top of an image or background image is 100% size with opacity setting);
- Gradient (would also be useful with a Color Fill block).
- Aligning text to center;
- Rotation (is already logged in #2005.)

cc @swissspidy @westonruter

swissspidy:
Color Fill block: My immediate thought was https://github.com/WordPress/gutenberg/issues/4900. The PR for that, https://github.com/WordPress/gutenberg/pull/13964, is still in the works, but according to the comments there it might be ready soon.

swissspidy:
> Aligning text to center

Should be as easy as setting `supports: { align: true }` and adding the necessary `text-align` CSS.

westonruter:
As I'm sure you're aware, also related to Section block is the Cover Image as a proto block: it allows you to add additional children to it. The Overlay has a color and opacity setting:

![image](https://user-images.githubusercontent.com/134745/54842882-5bc2b780-4c90-11e9-9787-407b5fd757cd.png)

westonruter:
> * Adding opacity to block

What about adding opacity to a block as a whole, and not just the background color? I know this was done for layers in the previous incarnation of AMP Stories. Maybe that is just asking for trouble.

miina:
> What about adding opacity to a block as a whole, and not just the background color? I know this was done for layers in the previous incarnation of AMP Stories. Maybe that is just asking for trouble.

Unless there is a reason why not to then seems like we could add it generally for other blocks, too, or at least to Image and Text Block.

Another style-related thing that occurs frequently in the templates is having rounded corners for images and Text blocks with background color. Maybe that'd also be something necessary?

westonruter:
[...]

PR #2046, opened 2019-04-02
[AMP Stories] Add overlay settings to Page

miina:
See #2013.

Adds Overlay settings to page which includes two color settings and opacity.
If one color is selected then the background is filled with the color, otherwise gradient is used.

Screen Shot 2019-04-02 at 12 41 44 PM

### DIFF
--- a/assets/src/blocks/amp-story/edit.js
+++ b/assets/src/blocks/amp-story/edit.js
@@ -6,7 +6,7 @@ import uuid from 'uuid/v4';
 /**
  * WordPress dependencies
  */
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 import {
 	InnerBlocks,
 	PanelColorSettings,
@@ -91,10 +91,10 @@ class EditPage extends Component {
 		}
 	}
 
-	removeBackgroundColors() {
+	removeBackgroundColor( index ) {
 		const { attributes, setAttributes } = this.props;
 		const backgroundColors = JSON.parse( attributes.backgroundColors );
-		backgroundColors.pop();
+		backgroundColors.splice( index, 1 );
 		setAttributes( { backgroundColors: JSON.stringify( backgroundColors ) } );
 	}
 
@@ -110,6 +110,7 @@ class EditPage extends Component {
 	getOverlayColorSettings() {
 		const { attributes } = this.props;
 		const backgroundColors = JSON.parse( attributes.backgroundColors );
+
 		if ( ! backgroundColors.length ) {
 			return [
 				{
@@ -121,16 +122,21 @@ class EditPage extends Component {
 				},
 			];
 		}
+
 		const backgroundColorSettings = [];
-		backgroundColors.forEach( ( color, key ) => {
-			backgroundColorSettings[ key ] = {
-				value: color.color,
+		const useNumberedLabels = backgroundColors.length > 1;
+
+		backgroundColors.forEach( ( color, index ) => {
+			backgroundColorSettings[ index ] = {
+				value: color ? color.color : undefined,
 				onChange: ( value ) => {
-					this.setBackgroundColors( value, key );
+					this.setBackgroundColors( value, index );
 				},
-				label: __( 'Color', 'amp' ),
+				/* translators: %s: color number */
+				label: useNumberedLabels ? sprintf( __( 'Color %s', 'amp' ), index + 1 ) :  __( 'Color', 'amp' ),
 			};
 		} );
+
 		return backgroundColorSettings;
 	}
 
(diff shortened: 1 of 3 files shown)


miina:
I'm wondering if it would be better to add a radio button for using Gradient. This way it would also be possible to use gradient with transparent colors.

At this moment if the "main" color is not selected and the second is then the gradient is created with `transparent` to the second color. However, the opposite case is not possible.

So perhaps it's better to have a radio to specifically set if the user wants to use gradient or not and then display either one or two color settings accordingly. Or maybe you have better alternatives in mind? Thoughts?

cc @swissspidy @westonruter

assets/src/blocks/amp-story/edit.js
```diff
@@ -134,20 +135,48 @@ class EditPage extends Component {
...
+			overlayStyle.backgroundImage = `linear-gradient(to bottom, ${ topColor }, ${ gradientBottomColor })`;
+			overlayStyle.opacity = overlayOpacity / 100;
+		}
+
 		return (
 			<Fragment>
 				<InspectorControls key="controls">
 					<PanelColorSettings
-						title={ __( 'Color Settings', 'amp' ) }
+						title={ __( 'Overlay Settings', 'amp' ) }
```
swissspidy:
```suggestion
						title={ __( 'Background Color', 'amp' ) }
```

If one just wants to use a solid background color without any gradient, I think this wording makes more sense.

Thanks to the opacity control further down, one will quickly discover that this can be used as an overlay for the background image.

swissspidy:
What if we want to support three or more colors in the future?

For that, I like Weston's suggestion of making this a _repeatable_ field.

By default you'll only see the one background color option with the opacity control.

Below the color option could be an "Add Color" button to add a second color. In the future we'd allow adding a third, fourth, etc. color as well.

To make this flexible and future-proof enough, there could only be a single `background` attribute that is a JSON-encoded object that looks more like this:

```json
{
  "background": {
    "opacity": 0.5,
    "colors": [
      {
        "color": "#ffffff"
      },
      {
        "color": "#000000"
      }
    ]
  }
}
```

For a regular solid background:

```json
{
  "background": {
    "opacity": 1,
    "colors": [
      {
        "color": "#ababab"
      }
    ]
  }
}
```

Eventually, we'd need to look at adding a gradient picker à la http://www.colorzilla.com/gradient-editor/, for which we'd just need to extend it like this:

```json
{
  "background": {
    "opacity": 0.5,
    "direction": 57,
    "colors": [
      {
        "color": "#ffffff",
        "step": 0
      },
      {
        "color": "#777777",
        "step": 0.5
      },
      {
        "color": "#000000",
        "step": 1
      }
    ]
  }
}
```

This way we don't have rename block attributes in the future and things like that.

Hope that makes sense.

Also, FWIW, there are a few React gradient picker components out there from which we could draw inspiration from.

assets/src/blocks/amp-story/edit.js
miina: Hmm, I was thinking that `Overlay` is a word that is valid for both cases since Overlay doesn't necessarily have to be an overlay for an image -- it can be overlay for the Page block without the image, too. Also, technically it'll be a new Fill layer (aka overlay) and not use the background color.

Since the user will not necessarily open the Panel to see what's inside then maybe Background Color could be deceiving, it would be great if the title would be self-explanatory without having to know what's inside. Maybe there's another option instead?

miina:
> This way we don't have rename block attributes in the future and things like that.

Yep, it makes sense, thought that adding a repeating field just for two colors would be too much. However, it's true that we might need it in the future and reworking this wouldn't be the best option later.

Will add the repeating field for the two colors for now, probably we can add other improvements later (considering the timeframe).

assets/src/blocks/amp-story/edit.js
swissspidy: `Background Settings` maybe? Then it could be combined with the background media panel.

Technically it's a fill layer, but for the user it's basically a background color setting with some optional opacity.

For me, "background" is easier to grasp than "overlay", and also easier to localize.

swissspidy:
> Yep, it makes sense, thought that adding a repeating field just for two colors would be too much. 

For two colors in the beginning we could label the button "Add Gradient"

assets/src/blocks/amp-story/edit.js
miina: `Background Settings` makes sense 👍

miina:
We'll also need to add remove button as well then.

assets/src/blocks/amp-story/edit.js
amedina: Since from the user perspective the net effect of this setting is providing a background color/fill, I would consider `Background Color` over `Settings. We want to make the language and UI as straightforward and intuitive as possible.

swissspidy:
`Test_AMP_Validation_Manager::test_add_block_source_comments` seems to be failing on Travis with the `latest_posts` data set. Opened an issue for that as it's unrelated to this PR: #2053

PR #2048, opened 2019-04-02
[AMP Stories] Default templates.

### LINKED ISSUE #2010: [AMP Stories] Create the base templates for Travel and Fandom

miina:
Fixes #2010.

- [x] Create system for adding default templates.
- [x] Create the actual default templates.


includes/admin/class-amp-story-templates.php
```diff
@@ -0,0 +1,197 @@
...
+	 */
+	public function init() {
+		if ( ! post_type_exists( 'amp_story' ) ) {
+			return;
+		}
+
+		add_filter( 'rest_wp_block_query', array( $this, 'filter_rest_wp_block_query' ), 10, 2 );
+
+		$this->register_taxonomy();
+		$this->maybe_import_story_templates();
```
miina: @swissspidy, @westonruter What do you think of this kind of initial approach for the default templates?

includes/admin/class-amp-story-templates.php
```diff
@@ -0,0 +1,197 @@
...
+	 * @param WP_REST_Request $request WP REST Request object.
+	 * @return array Args.
+	 */
+	public function filter_rest_wp_block_query( $args, $request ) {
+		$headers = $request->get_headers();
+		if ( ! isset( $headers['referer'][0] ) ) {
+			return $args;
+		}
+
+		$parts = wp_parse_url( $headers['referer'][0] );
```
swissspidy: Checking the referrer feels a bit hacky.

What if we passed an `?amp_story=123` param in the REST API request instead? It could be achieved using an [`apiFetch` middleware](https://wordpress.org/gutenberg/handbook/designers-developers/developers/packages/packages-api-fetch/#middlewares) that does this.

Then, here we only have to check for the existence and validity of the query param.

includes/admin/class-amp-story-templates.php
swissspidy: Sounds reasonable.

I wonder if we need some dynamic data in `get_story_templates()` as well at some point, e.g. default images, freshly generated UUIDs, etc. and how that works together with `template_exists`.

includes/admin/class-amp-story-templates.php
miina: It does feel a bit hacky, true.

The goal of this filter is to process the request that's done in the Block Inserter on Post types which are not AMP Story Post types to prevent the templates showing up in inserters. Otherwise templates show up under Reusable Blocks and just fail when inserted.

https://github.com/WordPress/gutenberg/blob/master/packages/editor/src/store/effects/reusable-blocks.js#L64

Could you elaborate on what you meant by passing `amp_story` param in the REST API? Not sure I understand what you meant.

includes/admin/class-amp-story-templates.php
miina: Hmm, maybe. Do you think that using static UUIDs might cause issues?
On images: an option could be using static URLs and not Media Library.

swissspidy: > Do you think that using static UUIDs might cause issues?

IDK. Perhaps someone exports their reusable blocks and someone imports them and then a conflict happens where two templates have the same IDs. Or maybe someone trashes a reusable block, somehow imports the templates again, and then ends up with the same situation.

> an option could be using static URLs and not Media Library.

I'd try to avoid the media library either way I think. Users wouldn't notice that an image belongs to a template and would most likely just delete it.

Static URLs means hosting them on `amp-wp.org` or a CDN, which might create some unnecessary overhead for us in maintaining that. Also, some people don't like loading assets from a CDN (e.g. emoji from WordPress.org, web fonts). Plus, it doesn't work offline.

An alternative would be bundling the images with the plugin and using `plugin_dir_url()` to reference them, making this a little bit more dynamic again.

includes/admin/class-amp-story-templates.php
swissspidy: Sure.

My train of thought was that when editing an AMP story, we would filter any REST API requests for reusable blocks and add a query param of our choice to the URL. For example `../../blocks?show_amp_story_templates_please=1`. That's what the middleware would do.

Then, on the PHP side, we could probably _always_ add the `tax_query` _unless_ that `show_amp_story_templates_please` query param is detected.

includes/admin/class-amp-story-templates.php
miina: > An alternative would be bundling the images with the plugin and using plugin_dir_url() to reference them, making this a little bit more dynamic again.

That's what I actually had in mind by "static URLs", you're right that they're not really static.

> IDK. Perhaps someone exports their reusable blocks and someone imports them and then a conflict happens where two templates have the same IDs.

Hmm, maybe what's in the content of the post wouldn't matter since the block is cloned and assigned a new ID before inserting anyway?

includes/admin/class-amp-story-templates.php
miina: Ah, okay, now I see, basically whitelisting only AMP Story instead of trying to blacklist all the others. Makes sense. This would require adjusting fetching the templates for an AMP Story instead of just using the default fetching.

There's one other moment to consider -- when the template itself is being edited (in the future). Eventually a template could consist of more than one template, and probably the default Inserter should display the Reusable blocks (templates) when editing a template, too.

includes/admin/class-amp-story-templates.php
```diff
@@ -0,0 +1,197 @@
...
+	 */
+	public function import_story_templates() {
+		$templates = self::get_story_templates();
+		foreach ( $templates as $template ) {
+			$existing_template = $this->template_exists( $template['title'], $template['content'] );
+			if ( $existing_template > 0 ) {
+				continue;
+			}
+			$post_id = wp_insert_post(
+				array(
```
westonruter: Note that this array needs to be wrapped in `wp_slash()` unfortunately, as `wp_insert_post()` expects slashed input. If this is not done backslashes will get stripped from the content, and this could be particularly bad for block attributes in JSON.

includes/admin/class-amp-story-templates.php
```diff
@@ -0,0 +1,197 @@
...
+			return 0;
+		}
+		global $wpdb;
+
+		$post_title   = wp_unslash( sanitize_post_field( 'post_title', $title, 0, 'db' ) );
+		$post_content = wp_unslash( sanitize_post_field( 'post_content', $content, 0, 'db' ) );
+
+		return (int) $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT ID FROM $wpdb->posts WHERE post_title = %s AND post_content = %s AND post_status = 'publish'",
```
westonruter: This is missing the `post_type` check. Needs to only include `wp_block` post type, right?

includes/admin/class-amp-story-templates.php
```diff
@@ -0,0 +1,197 @@
...
+	 *
+	 * Note that this is almost a copy of the default `post_exists` function except for the post status.
+	 *
+	 * @global wpdb $wpdb WordPress database abstraction object.
+	 *
+	 * @param string $title Post title.
+	 * @param string $content Post content.
+	 * @return int Post ID if post exists, 0 otherwise.
+	 */
+	private function template_exists( $title, $content ) {
```
westonruter: Instead of checking if a post exists with a given `$title` and `$content`. What about just using a slug? The `get_story_templates` method could include a `name` in addition to a `title` and `content`, and the name could be something like `amp-story-template-1` (or even a UUID). Then to check if the template exists, just check if a `wp_block` exists with that `post_name` (slug).

includes/admin/class-amp-story-templates.php
```diff
@@ -0,0 +1,197 @@
...
+	/**
+	 * Get the static list of templates, including temlate name and content.
+	 *
+	 * @return array Story templates.
+	 */
+	public static function get_story_templates() {
+		return array(
+			array(
+				'title'   => __( 'Default Template 1', 'amp' ),
+				'content' => '<!-- wp:amp/amp-story-page {"backgroundColor":"#313131","autoAdvanceAfterDuration":0} -->
```
westonruter: Instead of putting the raw HTML into the PHP, how about loading the HTML from files on disk? This will make them easier to maintain.

includes/admin/class-amp-story-templates.php
miina: Hmm, but this way re-importing (a future feature) wouldn't work as expected after editing the templates since the slugs would stay the same. The idea was that if the default template has been changed then re-importing would create new templates.

Perhaps we should check the content only instead?

westonruter: OK, good point. 

Note also that there is a `post_exists()` which could potentially be re-used?

Otherwise, the slug alone could be checked, and when a template is inserted by the AMP plugin here, a postmeta flag could be added to indicate it is `original` (or unmodified). Then, there could be a `save_post_amp_story` action which is used to remove this postmeta whenever the post is subsequently edited (by a means other than the AMP plugin here). When that postmeta is absent, then we would know that the template should _not_ be overridden since the user modified it. Otherwise, if the `original` flag _is_ present, then we could freely update the template. This would ensure that new versions of the template could be saved into the database such as when the templates themselves evolve.

miina: I think that `post_exists()` doesn't include the check for `publish` status, it would give a positive even for trashed templates.

> This would ensure that new versions of the template could be saved into the database such as when the templates themselves evolve.

That's an interesting idea. The `TEMPLATE_VERSION` constant was added for that reason -- to track if the templates have evolved meanwhile to automatically start the import (if the version is changed). However, it's true that it would just import everything instead of updating the previous ones.

One concern with updating the previous templates is that the user might actually be using them and not want an update. Thoughts?

westonruter: > I think that `post_exists()` doesn't include the check for `publish` status, it would give a positive even for trashed templates.

OK, but then you could check for the `status` after it is returned.

> One concern with updating the previous templates is that the user might actually be using them and not want an update. Thoughts?

I guess it depends on how we expect the templates to evolve over time. Would they necessarily change much? Should we just assume the overall look and feel of the template is pretty well settled and we'd make incremental changes? In other words, they could be treated like core themes: the design does not change, but the code does evolve a bit over time.

miina: > OK, but then you could check for the status after it is returned.

`post_exists()` returns only one post's ID (the first match). If we have one Trashed template and another published and the first match would be the one that has been trashed then we'd still end up importing a duplicate template. The method in the PR is actually based on `post_exists` since that was the initial thought to use, unfortunately, it doesn't allow defining the status to check against, would be useful here.

> Would they necessarily change much? Should we just assume the overall look and feel of the template is pretty well settled and we'd make incremental changes?

Probably it wouldn't change a lot, true.

On checking against the slug only: if we would have predefined names (slugs) and the template would be edited by the user, and then reimport would be run. This would mean that a new template would be created with a new slug (not the predefined one since that would already exist), would we then have to store the new slugs for being able to check later if the reimported template with a new slug has the `original` flag? Probably we should then always only check the latest "new" slug of that specific template when reimporting? Thoughts?

We could use the `original` flag in any case.

westonruter: OK, cool on not using `post_exists()`.

If a user modified the original template, then how about juts skipping it? If the user really wants to get the original template back, then they could delete the modified template and then run the import again?

miina: > If a user modified the original template, then how about just skipping it?

What if the user wants both, the original and the modified? 😄 

>  If the user really wants to get the original template back, then they could delete the modified template and then run the import again?

In this case, we should maybe display a message when running the import to indicate that some of the templates already existed and weren't imported? This could hint that deleting the previous templates would cause re-importing the templates, otherwise, it wouldn't be clear why the templates aren't being imported.

westonruter: If that is easy I suppose. But doing whatever is most straightforward to start with is probably best. Then we can iterate.

includes/admin/class-amp-story-templates.php
```diff
@@ -0,0 +1,216 @@
...
+		$this->import_story_templates();
+		// @todo This is commented out for testing purposes until the PR gets ready.
+		// AMP_Options_Manager::update_option( 'story_templates_version', self::STORY_TEMPLATES_VERSION );
+	}
+
+	/**
+	 * Import logic for story templates.
+	 */
+	public function import_story_templates() {
+		$templates = self::get_story_templates();
```
westonruter: Idea: What if each story template also included a path to a featured image which could be associated with the `wp_block` post type? This would be like a `screenshot.php` for a theme, where it could show it being populated with content. But the actual block that gets inserted would have the placeholders that users would have to provide content for? This would only apply to templates that are inserted by the AMP plugin, naturally, unless the user is able to set a featured image for `wp_block` post type generally.

miina: That's something that @amedina also suggested and that is logged as an idea under this issue:
https://github.com/ampproject/amp-wp/issues/2061

The problem mentioned with this is that the templates are not static, meaning that if the user changes the template then it wouldn't be accurate anymore.

We could stop showing it after the template is changed but this could be unexpected / feel inconsistent.

Maybe we can reiterate under #2061 which is an issue specifically for the template preview?

miina:
Editing templates removed within d268d7d. cc @westonruter

miina:
@swissspidy @westonruter 

This PR now has 4 templates:
- Tip;
- CTA;
- Quote;
- Page Title (this needs updating once the meta blocks have been merged, it's a "skeleton" template at this moment).

The template that's still missing is the[ template including rotation](https://projects.invisionapp.com/d/main#/console/17106855/355178249/preview), could be added separately later.

The template is displayed as prepopulated in the preview and inserted as "skeleton template".

Note that the preview design for the quote is a bit off, this probably still needs fixing.

miina:
Note: I'll be offline tomorrow and on Friday so feel free to make direct changes to the PR if needed.

includes/templates/story-templates/title-page.html
```diff
@@ -0,0 +1,13 @@
+<!-- wp:amp/amp-story-page {"autoAdvanceAfterDuration":0,"backgroundColors":"[{},{\u0022color\u0022:\u0022#313a43\u0022}]","overlayOpacity":40} -->
```
miina: This needs to be replaced with meta blocks instead.

swissspidy: Done in 8f1a1a166843cb860cbb0608c62a22a840683ae5

PR #2049, opened 2019-04-02
[AMP Stories] Text Block: fix <br> showing as plain text.

### LINKED ISSUE #2025: [AMP Stories] Line breaks in text block
When adding line breaks in a text block, saving the post and reloading the tab, the `<br>` tags get displayed like this in the editor and on the front end:

Screenshot 2019-03-26 at 16 43 10

There should be actual line breaks instead

miina:
Fixes #2025.

### DIFF
--- a/assets/src/blocks/amp-story-text/edit.js
+++ b/assets/src/blocks/amp-story-text/edit.js
@@ -234,8 +234,16 @@ class TextBlockEdit extends Component {
 						identifier="content"
 						wrapperClassName="wp-block-amp-story-text"
 						tagName="p"
-						value={ content }
-						onChange={ ( value ) => setAttributes( { content: value } ) }
+						// Ensure line breaks are normalised to HTML.
+						value={ content.replace( /\n/g, '<br>' ) }
+						onChange={ ( nextContent ) => {
+							setAttributes( {
+								// Ensure line breaks are normalised to characters. This
+								// saves space, is easier to read, and ensures display
+								// filters work correctly.
+								content: nextContent.replace( /<br ?\/?>/g, '\n' ),
+							} );
+						} }
 						onReplace={ this.onReplace }
 						style={ {
 							backgroundColor: backgroundColor.color,
--- a/assets/src/blocks/amp-story-text/index.js
+++ b/assets/src/blocks/amp-story-text/index.js
@@ -13,6 +13,9 @@ import {
 	getFontSize,
 } from '@wordpress/block-editor';
 import { select } from '@wordpress/data';
+import {
+	RawHTML,
+} from '@wordpress/element';
 
 /**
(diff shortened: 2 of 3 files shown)


swissspidy:
Thanks for looking into it!

Using raw HTML is prone to XSS, so I am curious: why do the paragraph and heading blocks not use this? Is there some workaround they use?

miina:
@swissspidy: just checked and it looks like `paragraph` block does use it for saving.
On headings: I actually wasn't aware that headings allow any HTML within the block.

swissspidy:
The one occasion I see in the paragraph block is in the deprecated list: https://github.com/WordPress/gutenberg/blob/fc20eeac75776fa2eb40497092551f5031dc9546/packages/block-library/src/paragraph/index.js#L202

miina:
You're right. Hmm, I guess there must be another way then.

miina:
Looks like `RichText` supports using `Shift + Enter` for adding line breaks, and that works as expected with the Text Block, too. It seems that it adds `data-rich-text-line-break` as attribute in the editor to `br` for not escaping it.

We should probably add then handling for `Enter` key as well within the Text Block, the same way as `Shift + Enter` is already handled by default.

swissspidy:
Thanks for looking into it! I assume this is the code you looked at? https://github.com/WordPress/gutenberg/blob/ba64d414eaa687ce31daab7f8fad9110dcfd7800/packages/block-editor/src/components/rich-text/index.js#L719-L751

Agreed that it makes sense to have the same handling for `Enter` in that case.

miina:
Yep, correct, that's the code I looked at.

swissspidy:
@miina I was just pointed at the preformatted block which handles this quite nicely: https://github.com/WordPress/gutenberg/blob/4905299edfb1f652229daea695fe196b55d0bf76/packages/block-library/src/preformatted/edit.js#L13-L22

miina:
Thanks for sharing, I'll get to it a little bit later today. (Feel free to take over if you'd like to)

swissspidy:
So with that `onChange` callback line breaks are only stored as `\n` in the database, which means we need to turn them back into `br`'s at some point.

I just pushed a commit with such a change to experiment with that.

swissspidy:
@miina Let me know what you think of 9e92d400a087b8fad5faa41fac270e05d3ff17f3. Not as elegant as a JS-only solution, so hopefully what you had in mind works better.

miina:
@swissspidy I think it works well for now. I was thinking of handling the key event for `Enter`, however, I didn't actually have time to look into exactly how it's done in `RichText` and how to easily bring it over.

The commit you added seems to get it done potentially in a simpler way so unless you have objections then we could use your solution. Thoughts?

PR #1861, opened 2019-02-11
Add noscript fallbacks for img, audio, and video; improve audio conversion to amp-audio; add AMP [fallback] for amp-audio/amp-video

westonruter:
Appends original `img` to `amp-img` under `noscript` for no-JS browsers. This is [recommended by AMP](https://www.ampproject.org/docs/media/amp_replacements#displaying-images-when-javascript-is-disabled):

> ## Displaying images when JavaScript is disabled
> As `<amp-img>` relies on JavaScript, if the user chooses to disable scripts, images won't display. In this case, you should provide a fallback to the image using `` and `<noscript>`, like so:
```html
<amp-img src="images/sunset.jpg"
  width="264"
  height="195">
  <noscript>
    <img src="images/sunset.jpg" width="264" height="195" />
  </noscript>
</amp-img>
```

So, given an `img` element:

```html
<img src="http://placehold.it/300x400" width="300" width="400">
```

Before this would be sanitized as:

```html
<amp-img src="http://placehold.it/300x400" width="300" width="400" class="amp-wp-enforced-sizes"></amp-img>
```

But after this change it would be:

```html
<amp-img src="http://placehold.it/300x400" width="300" width="400" class="amp-wp-enforced-sizes">
    <noscript>
        <img src="http://placehold.it/300x400" width="300" width="400">
    </noscript>
</amp-img>
```

This also short-circuits the `img`, `video`, and `audio` sanitizers for elements that are inside of `noscript` elements. Similarly, the AMP equivalent to `wp_mediaelement_fallback()` is implemented so that there are `fallback` links inside the `amp-audio` elements.

# Audio Improvements

## Audio non-AMP

> image

```html
<figure class="wp-block-audio">
    <audio controls src="https://wordpressdev.lndo.site/content/uploads/2019/02/do-you-know-I-am-batman.mp3"></audio>
    <figcaption>Caption</figcaption>
</figure>

<!--[if lt IE 9]><script>document.createElement('audio');</script><![endif]-->
<audio class="wp-audio-shortcode" id="audio-87-1" preload="none" style="width: 100%;" controls="controls">
    <source type="audio/mpeg" src="https://wordpressdev.lndo.site/content/uploads/2019/02/do-you-know-I-am-batman.mp3?_=1"/>
    <a href="https://wordpressdev.lndo.site/content/uploads/2019/02/do-you-know-I-am-batman.mp3">https://wordpressdev.lndo.site/content/uploads/2019/02/do-you-know-I-am-batman.mp3</a>
[...]

PR #2051, opened 2019-04-02
Consider element attributes when determining if an parent element is empty and can be removed

westonruter:
We discovered an issue when a site is served via HTTP instead of HTTPS. An `amp-img` [allows](https://github.com/ampproject/amphtml/blob/7b16d1cc109eb77a8e0477dd8eff9501fae638b2/validator/validator-main.protoascii#L4963-L4979) for its image to use the `http` protocol, so this is valid AMP:

```html
<amp-img src="http://example.com/bison-1024x668.jpg" srcset="http://example.com/bison-1024x668.jpg 1024w, http://example.com/bison-300x196.jpg 300w, http://example.com/bison-768x501.jpg 768w, http://example.com/bison.jpg 1200w" sizes="(max-width: 767px) 89vw, (max-width: 1000px) 54vw, (max-width: 1071px) 543px, 580px" width="1024" height="668" layout="intrinsic"></amp-img>
```

With #1861 there was introduced `noscript > img` fallbacks for serving when a user with JS disabled accesses a (canonical) AMP page. This results in markup like:

```html
<amp-img src="http://example.com/bison-1024x668.jpg" srcset="http://example.com/bison-1024x668.jpg 1024w, http://example.com/bison-300x196.jpg 300w, http://example.com/bison-768x501.jpg 768w, http://example.com/bison.jpg 1200w" sizes="(max-width: 767px) 89vw, (max-width: 1000px) 54vw, (max-width: 1071px) 543px, 580px" width="1024" height="668" layout="intrinsic">
  <noscript>
    <img src="http://example.com/bison-1024x668.jpg" srcset="http://example.com/bison-1024x668.jpg 1024w, http://example.com/bison-300x196.jpg 300w, http://example.com/bison-768x501.jpg 768w, http://example.com/bison.jpg 1200w" sizes="(max-width: 767px) 89vw, (max-width: 1000px) 54vw, (max-width: 1071px) 543px, 580px" width="1024" height="668">
  </noscript>
</amp-img>
```

It turns out that this actually is invalid AMP:

image

The `src` and `srcset` attributes are flagged as errors because—perhaps unintentionally—the `noscript > img` element [does not allow](https://github.com/ampproject/amphtml/blob/7b16d1cc109eb77a8e0477dd8eff9501fae638b2/validator/validator-main.protoascii#L2235-L2245) the `http` protocol.

The resulting behavior is the tag-and-attribute sanitizer proceeds to remove the `img`. Then the `AMP_Tag_And_Attribute_Sanitizer::remove_node()` has a `while` loop to walk up the DOM tree to examine each parent, and for each empty parent, they are removed. This is a problem because `noscript` is an empty element is removed and `amp-img` is also then an empty element and is removed. The problem is that “emptiness” is not taking into account the _attributes_ of the parent elements. So this fixes that. Nevertheless, we should also evaluate in the future if empty parents really should be removed in the first place.

An issue also needs to be opened against the AMPHTML validator to allow `http` protocol for `noscript > img` elements.

----

Also fixes tests which apparently break when testing on WordPress 4.9 and the latest Gutenberg by explicitly installing an old version of Gutenberg that works with 4.9.

### DIFF
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -1973,7 +1973,9 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 		if ( $node && $parent ) {
 			$this->remove_invalid_child( $node );
 		}
-		while ( $parent && ! $parent->hasChildNodes() && $this->root_element !== $parent ) {
+
+		// @todo Does this parent removal even make sense anymore?
+		while ( $parent && ! $parent->hasChildNodes() && ! $parent->hasAttributes() && $this->root_element !== $parent ) {
 			$node   = $parent;
 			$parent = $parent->parentNode;
 			if ( $parent ) {
--- a/.dev-lib
+++ b/.dev-lib
@@ -10,6 +10,12 @@ fi
 
 function after_wp_install {
 	echo -n "Installing Gutenberg..."
-	svn export -q https://plugins.svn.wordpress.org/gutenberg/trunk/ "$WP_CORE_DIR/src/wp-content/plugins/gutenberg"
+	if [[ "$WP_VERSION" = "4.9" ]]; then
+		# The last version that supported WordPress 4.9:
+		gutenberg_plugin_svn_url=https://plugins.svn.wordpress.org/gutenberg/tags/4.9.0/
+	else
+		gutenberg_plugin_svn_url=https://plugins.svn.wordpress.org/gutenberg/trunk/
+	fi
+	svn export -q "$gutenberg_plugin_svn_url" "$WP_CORE_DIR/src/wp-content/plugins/gutenberg"
 	echo "done"
 }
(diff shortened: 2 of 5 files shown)


includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1973,7 +1973,9 @@ private function remove_node( $node ) {
 		if ( $node && $parent ) {
 			$this->remove_invalid_child( $node );
 		}
-		while ( $parent && ! $parent->hasChildNodes() && $this->root_element !== $parent ) {
+
+		// @todo Does this parent removal even make sense anymore?
```
amedina: Are you referring to alternatively not removing empty parent nodes?

westonruter: Yes, that's right. We should evaluate why this was added in the first place, and consider discontinuing the practice. 

I believe it may have been due to having input like this:

```
<p><script>evil</script></p>
```

Normally this would sanitized as:

```
<p></p>
```

That results in what appears to be a blank area in the page. So apparently it helps in that case. But we should consider being more limited in what we consider to be suitable for removal. For example, perhaps this only makes sense for `p` elements.

PR #2052, opened 2019-04-03
Improve preview in Block Navigation

### LINKED ISSUE #1971: [AMP Stories] Improve preview in Block Navigation
For the `BlockNavigationList` component we can basically re-use `ButtonContent` from the `AnimationOrderPicker`, with the only difference being the icon position.

This would make the block navigation a bit more user friendly, as it is currently hard to distinguish individual items in it.

swissspidy:
Re-uses the "smart" preview from the animation order picker for the block navigation UI.

Instead of seeing "Text" dozens of times for every block, this now shows an excerpt of the block content, which is way more user friendly.

Screenshots:

Screenshot 2019-04-03 at 12 00 53

Screenshot 2019-04-03 at 12 01 05

Fixes #1971.

### DIFF
--- a/assets/src/components/block-preview-label.js
+++ b/assets/src/components/block-preview-label.js
@@ -1,0 +1,44 @@
+/**
+ * WordPress dependencies
+ */
+import { Fragment } from '@wordpress/element';
+import { BlockIcon } from '@wordpress/block-editor';
+
+export default ( { label: name, block, blockType, displayIcon = true, alignIcon = 'left', accessibilityText } ) => {
+	if ( ! block ) {
+		return name;
+	}
+
+	let label;
+	let content;
+
+	// Todo: Cover more special cases if needed.
+	switch ( block.name ) {
+		case 'core/image':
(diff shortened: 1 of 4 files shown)


assets/src/components/block-preview-label.js
```diff
@@ -0,0 +1,44 @@
...
+	// Todo: Cover more special cases if needed.
+	switch ( block.name ) {
+		case 'core/image':
+			if ( block.attributes.url ) {
+				content = block.attributes.url.slice( block.attributes.url.lastIndexOf( '/' ) ).slice( 0, 30 );
+
+				if ( content.length > 0 ) {
+					label = content;
+				}
+			}
```
miina: Maybe would be good to add `else` and set the label to `title` within it? Otherwise `label.length` might cause error below if the URL is not set.

swissspidy: Thanks! 38b3452e8e05fab3bdd5a7761472cf5e23201127 addresses this by always setting a label already at the beginning.

Issue #1966, opened 2019-03-14
[AMP Stories] Fix default Inserter in the Toolbar


miina:
Currently, the Inserter in the upper Toolbar doesn't seem to work as expected. It seems to allow adding a Page within a Page and not allow the allowed child blocks.
Steps to reproduce:
- Add a Page and a block (or some blocks);
- Select a block within the Page;
- Click on the default Inserter in the toolbar.

swissspidy:
There was a recent bug fix in Gutenberg for the global inserter in the toolbar: https://github.com/WordPress/gutenberg/pull/14020. It will be included in 5.3.

Let's test this with the current development version to see if anything changed for us. After all we do some on-the-fly filtering there.

swissspidy:
I just installed 5.3 RC 1 and it seems to work fine now!

1. Add page
2. Select block within page
3. Click on global inserter in the toolbar
4. See the full list of allowed child blocks (no pages or anything.

When a page or no block at all is selected, then the global inserter only allows you to insert pages.

amedina:
The flow when editing a page and I want to insert components to it does not work as expected. 

For example: I am on a blank page, click on the inserter on the top left corner, and only the page block is available. The expected scenario is for the user to see all the blocks that can be inserted, and also the possibility of adding a new page.

miina:
@amedina That's the default behavior of the Inserter of Gutenberg -- it always shows what you can insert after the currently selected block. For example, if a Page is selected then you can insert only another Page after it and that's why only a Page shows, since the other blocks can't be inserter _after_ the Page, they could only be inserter _inside_ the Page.

On a blank page by default no block is selected, that's why the inserter shows only the Page -- this would insert a new Page. If it would show other blocks there, too, then it wouldn't be clear where they could be inserter. By default to the root element, but these are not allowed there.

Perhaps ideally if a Page is selected then other blocks would also be displayed and if a non-Page block would be inserted, it would be inserted as the last block of the selected Page.
However, not sure how easy it would be to change the default behavior. Thoughts, @swissspidy?

swissspidy:
> Perhaps ideally if a Page is selected then other blocks would also be displayed and if a non-Page block would be inserted, it would be inserted as the last block of the selected Page.

Agreed. We turn Gutenberg a bit on its head with the way stories are edited, so changing the behavior here makes sense.

> However, not sure how easy it would be to change the default behavior.

[...]

PR #2054, opened 2019-04-03
[AMP Stories] Hide all block insertion points

### LINKED ISSUE #2047: [AMP Stories] Overlapping Inserter Button
when I am editing a page or a block within a page, The default inserter on top overlaps the block controls; and also it seems the inserter associated for the block appears misplaced. 
Inserter-Bug

swissspidy:
Prevents all these inserters before/after a block from showing up.

Does not address the inserter visibility on the right side of the page, as this is more related to #1966.

Fixes #2047.

### DIFF
--- a/assets/css/amp-editor-story-blocks.css
+++ b/assets/css/amp-editor-story-blocks.css
@@ -141,6 +141,10 @@ div[data-type="amp/amp-story-page"][data-amp-selected="parent"] > div > div > di
 	display: none;
 }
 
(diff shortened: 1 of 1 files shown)


miina:
I'm still seeing the inserter for some reason.

Screen Shot 2019-04-03 at 6 20 14 PM
Screen Shot 2019-04-03 at 6 20 03 PM

swissspidy:
@miina Thanks, should be fixed now with the latest commit!

I was only hiding the inserters (which are shown before a block, I think) but not the appenders (which are shown after a block, I think).

PR #2055, opened 2019-04-03
[AMP Stories] Optional rounded corners for Text and Image block

miina:
See #2013.

### DIFF
--- a/assets/css/amp-stories.css
+++ b/assets/css/amp-stories.css
@@ -4,8 +4,9 @@ amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
 	padding: 10px !important;
 }
 
-.amp-story__has-rounded-corners,
-div[data-rounded-corners="true"] figure.wp-block-image,
-.wp-block-image[data-rounded-corners="true"] img {
+.amp-story__has-rounded-corners {
+    border-radius: 5px;
+}
+[data-rounded-corners="true"] img {
 	border-radius: 8px;
 }
--- a/assets/css/amp-stories.css
+++ b/assets/css/amp-stories.css
@@ -6,6 +6,6 @@ amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
 
 .amp-story__has-rounded-corners,
 div[data-rounded-corners="true"] figure.wp-block-image,
-div[data-rounded-corners="true"] img {
+.wp-block-image[data-rounded-corners="true"] img {
 	border-radius: 8px;
 }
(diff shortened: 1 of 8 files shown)


miina:
@swissspidy `registerBlockStyle` seems to be generally for a block type, added the rounded corners Toggle by using the existing filters instead.

swissspidy:
@miina I'm not sure I follow, what exactly was the issue you were facing with that method?

miina:
@swissspidy `registerBlockStyle` seems to add a new style for all the blocks of a certain type, however, we'd only need the style for a specific single block where the user has chosen to use rounded corners. So it seems like `registerBlockStyle` is not suitable for this case. Unless I'm missing something :)

swissspidy:
> `registerBlockStyle` seems to add a new style for all the blocks of a certain type

Not really 🤔. `registerBlockStyle` provides the user alternative styles for blocks. It works by adding a class name to the block’s wrapper.

Try inserting a button block in Gutenberg and click on its icon. You can then see the different styles:

Screenshot 2019-04-03 at 16 36 00

See https://wordpress.org/gutenberg/handbook/designers-developers/developers/filters/block-filters/#block-style-variations

So if we use this:

```js
registerBlockStyle( 'amp/amp-story-text', {
    name: 'rounded',
    label: __( 'Rounded', 'amp' )
} );
```

The user can choose between the default style and the Rounded style. With Rounded, an `is-rounded` class name is added to the block.

So if the user wants rounded corner, they can simply switch the block style.

miina:
Oh, looks like I completely misunderstood this one. Thanks, will rework.

assets/css/amp-stories.css
```diff
@@ -3,3 +3,10 @@ amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
 .has-background {
 	padding: 10px !important;
 }
+
+.is-style-rounded {
+    border-radius: 5px;
+}
+.is-style-rounded img {
+	border-radius: 8px;
```
swissspidy: @miina What's the reason here for the different radius? 8px vs. 5px.

miina: That's based on the designs, the images in template designs tend to have higher value for radius and are mostly larger as well. For example:
Screen Shot 2019-04-03 at 6 37 35 PM

We could make it the same as well though, could be better for consistency. Thoughts?

swissspidy: 5px or 8px didn't really make a difference for me, so I'd rather unify them. I'll handle it.

PR #2056, opened 2019-04-03
[AMP Stories] JS Build Tooling Optimizations Part 2

swissspidy:
After #2000:

* [x] Move relevant styles from amp-editor-story-blocks.css to live next to their respective components
* [x] Enqueue compiled AMP Stories editor CSS
* [x] Remove some unused CSS along the road that was used in the beginning when there were layers

One thing I still need to figure out:

When `amp-editor-blocks.js` imports any component from `src/components/index.js` (e.g. `import { getPrePublishNotice } from './components';`, webpack extracts the CSS of _all_ the components exported by `components/index.js`. Thus, `amp-block-editor-compiled.css` contains the styles for the `PreviewPicker` component, even though it is not used.


miina:
This has now merge conflicts.

swissspidy:
Resolved conflicts now.

assets/src/components/block-navigation/index.js
```diff
@@ -10,7 +10,8 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
-import { BlockPreviewLabel } from './';
+import { BlockPreviewLabel } from '../';
+import './edit.css';
 
 /**
  * Internal dependencies
```
miina: `import { ALLOWED_MOVABLE_BLOCKS } from '../constants';` below should be changed to `import { ALLOWED_MOVABLE_BLOCKS } from '../../constants';` with the new structure.

swissspidy:
Thanks for your review @miina!

I am gonna merge this now and will follow up on the `amp-block-editor-compiled.css` issue separately. It is not urgent now as that file is not currently used yes.

PR #2059, opened 2019-04-03
Discontinue showing theme support admin pointer in 1.1

westonruter:
This admin pointer currently is shown to all users who activate the plugin:

image

With the upcoming 1.1 release, to me it seems that this can be suppressed until there is need for another admin pointer for some new feature (e.g. AMP Stories).

This is why I didn't just remove all the code in the PR, since I presume we'll use it again soon.


amedina:
I agree with this. The current admin notice is old news by now. Let's nix it.

Issue #2050, opened 2019-04-02
[AMP Stories] Formatting of Reordering UI 


amedina:
In the reordering of story pages, two things need improvement:

- The pages shown are too small, 
- The content is not scaled appropriately

They should look as the design mocks.

/cc @miina @westonruter

swissspidy:
Related: #2015

amedina:
#2015 is merged into the `redux` branch? 
![Screen Shot 2019-04-02 at 2 44 15 PM](https://user-images.githubusercontent.com/506089/55438208-d96ea900-5555-11e9-9044-c41e2ae655b9.png)
![Screen Shot 2019-04-02 at 2 44 01 PM](https://user-images.githubusercontent.com/506089/55438209-da073f80-5555-11e9-8e33-1599926a1c8a.png)

swissspidy:
Yes. It didn‘t affect the reorder view though. Just regular edit view and front end.

amedina:
Ah; makes sense.

swissspidy:
@miina The template inserter also uses BlockPreview. Is it affected by this as well?

miina:
@swissspidy Templates seems to be showing correctly, however, templates have this styling assigned (0.48 being according to the ratio of the small / original size).
```
.amp-stories__editor-inserter__results .block-editor-block-preview .wp-block {
	transform: scale(0.48) translateX(-40%) translateY(-40%);
}
```

swissspidy:
Thanks, makes sense! Then I just need to fix the default font as well.

Also, block appenders should probably not shown.

Another fun thing I noticed:

1. Select block
2. Click on reordering
3. Block is shown as selected

<img width="626" alt="Screenshot 2019-04-03 at 12 39 10" src="https://user-images.githubusercontent.com/841956/55476091-8a844a80-560d-11e9-8cf0-eea8390450c4.png">

I'm gonna fix this as well.

PR #2063, opened 2019-04-04
[AMP Stories] Reordering UI improvements

swissspidy:
* Fix default font.
* Clear any block selections upon start
* Make inner blocks smaller

See #2050.


westonruter:
image

image

The images in the page thumbnails are smaller than expected, but otherwise this seems to work OK.

PR #2066, opened 2019-04-04
[AMP Stories] Add opacity to Text block

### LINKED ISSUE #2058: [AMP Stories] Add opacity Range control to allowed (child) blocks which use background color.
This was previously in #2013, however, separated it since it might not be a blocker for default templates.

- [ ] Add opacity to image block and other allowed blocks which use background color.

miina:
Fixes #2058.

This adds an opacity control to the Text block that affects the background color only, not the whole block.

### DIFF
--- a/assets/src/blocks/amp-story-text/index.js
+++ b/assets/src/blocks/amp-story-text/index.js
@@ -18,7 +18,7 @@ import { select } from '@wordpress/data';
  * Internal dependencies
  */
 import edit from './edit';
-import { getPercentageFromPixels } from '../../helpers';
+import { getPercentageFromPixels, getRgbaFromHex } from '../../helpers';
 import { STORY_PAGE_INNER_WIDTH } from '../../constants';
 
 export const name = 'amp/amp-story-text';
@@ -72,6 +72,9 @@ const schema = {
 	customTextColor: {
 		type: 'string',
 	},
+	backgroundHexValue: {
+		type: 'string',
+	},
 	backgroundColor: {
 		type: 'string',
 	},
@@ -122,6 +125,7 @@ export const settings = {
 			ampFitText,
 			autoFontSize,
 			backgroundColor,
+			backgroundHexValue,
 			textColor,
 			customBackgroundColor,
 			customTextColor,
@@ -139,7 +143,7 @@ export const settings = {
 			'has-text-color': textColor || customTextColor,
 			'has-background': backgroundColor || customBackgroundColor,
 			[ textClass ]: textClass,
-			[ backgroundClass ]: backgroundClass,
+			[ backgroundClass ]: ( ! opacity || 100 === opacity ) ? backgroundClass : undefined,
 		} );
 
 		// Calculate fontsize using vw to make it responsive.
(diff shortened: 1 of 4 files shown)


miina:
@swissspidy We might want to change this by adding the opacity only to the background color instead. E.g. converting the color to rgba (I suppose hex should be available for the colors that use classnames, too?). Not sure if having a general opacity could be useful enough. Thoughts?

swissspidy:
Agreed. That would only make the text hard/impossible to read. If we only apply it to the background color, one could still sorta fake text opacity by using a lighter font if really needed.

We can move the range control before the `ContrastChecker` one to group it with the color settings too.

assets/src/blocks/amp-story-text/index.js
```diff
@@ -72,12 +72,19 @@ const schema = {
 	customTextColor: {
 		type: 'string',
 	},
+	backgroundHexValue: {
+		type: 'string',
+	},
```
miina: Is there a better way how to store / get the hex value?

swissspidy: I don't exactly see why we'd need `backgroundColor`, `customBackgroundColor`, `backgroundHexValue`, _and_ `opacity`. Can you elaborate a bit on the use here?

I thought that when we have `backgroundColor` (a color slug), we can get the hex value from that. When we have `customBackgroundColor` (a custom color picked by the user), that is already a hex value.

When we have `opacity`, we just convert either of those values to RGB and add the alpha channel.

miina: Do you know which method can get hex from `backgroundColor` (color slug)?

assets/src/helpers.js
```diff
@@ -498,3 +498,21 @@ export const addBackgroundColorToOverlay = ( overlayStyle, backgroundColors ) =>
...
+ */
+export const getRgbaFromHex = ( hex, opacity ) => {
+	if ( ! hex ) {
+		return undefined;
+	}
+	hex = hex.replace( '#', '' );
+	const r = parseInt( hex.substring( 0, 2 ), 16 );
+	const g = parseInt( hex.substring( 2, 4 ), 16 );
+	const b = parseInt( hex.substring( 4, 6 ), 16 );
+	return `rgba( ${ r }, ${ g }, ${ b }, ${ opacity / 100 })`;
```
swissspidy: If this just returns `[ r, g, b, opacity / 100 ]`, it would be more re-usable and also allow conversion back to hex. In the block we can then just use `const [ r, g, b, a ] = getRgbaFromHex( ... )` and `rgba( ${ r }, ${ g }, ${ b }, ${ a })\``. Alternatively, I'd probably rename this from `getRgbaFromHex` to `rgbaCssValueFromHex` or something.

miina: Sure, makes sense, I'll change it to return `[ r, g, b, opacity / 100 ]` instead.

assets/src/blocks/amp-story-text/index.js
swissspidy: You can use the `withColorContext` HOC from `@wordpress/components` to get the editor color palette passed as a prop. That's what `PanelColorSettings` uses too. Then you can use something like `getColorObjectByAttributeValues( props.colors, props.attributes.backgroundColor )` to get the color object (that contains the hex color value)

swissspidy: Added this in the recent commits.

miina: Thanks, @swissspidy, works well.

PR #2068, opened 2019-04-04
[AMP Stories] Add "Tip" block style for Text Block.

miina:
See #2013.

### DIFF
--- a/assets/css/amp-stories.css
+++ b/assets/css/amp-stories.css
@@ -10,5 +10,5 @@ amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
 }
 
 .is-style-tip {
-	border-radius: 3pt 3pt 20pt 20pt;
+	border-radius: 4px 4px 26px 26px;
 }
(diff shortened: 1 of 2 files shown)


assets/css/amp-stories.css
```diff
@@ -4,9 +4,11 @@ amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
...
-.is-style-rounded {
+.is-style-rounded,
+.is-style-rounded img {
     border-radius: 5px;
 }
-.is-style-rounded img {
-	border-radius: 5px;
+
+.is-style-tip {
+	border-radius: 3pt 3pt 20pt 20pt;
```
swissspidy: I thought `pt` is more suited for print than screen. What's the use case here? Just wanna understand it better.

miina: These values are taken from design. We can convert to pixels, too.

swissspidy: Then let's do it :-)

miina:
Changed to "Half Rounded" :D 
It is more descriptive, the word "Tip" without context is not clear.

westonruter:
Should the CSS have a `::first-line` rule that makes the first line a different style, like smaller font size?

miina:
Maybe it would be okay to use another Text Block in case of wanting to have different font size? This way the "Half Rounded" could be used more universally.

PR #2073, opened 2019-04-05
[AMP Stories] Adjust quote style to match templates

### LINKED ISSUE #2065: [AMP Stories] Style the quote block to match template designs.
Add style to the quote block based on this template design:
- https://projects.invisionapp.com/share/8QR858KUNWK#/screens/355151903

miina:
Fixes #2065.

Missing:
- [x] Quote icon;


swissspidy:
Looks like this is missing some CSS for the "Large" block style. If we don't have any style for that, then let's just unregister that one.

The "Light" style seems very odd. As a user, it was not clear that this makes everything have a white color. Why can't this be solved with a color panel instead?

Default Style:

Screenshot 2019-04-09 at 20 40 40

Large:

Screenshot 2019-04-09 at 20 41 25


Light:

Screenshot 2019-04-09 at 20 40 48

miina:
>  If we don't have any style for that, then let's just unregister that one.

Yep, let's just unregister that, good idea.

> The "Light" style seems very odd. As a user, it was not clear that this makes everything have a white color. Why can't this be solved with a color panel instead?

Hmm. Well, I guess that makes sense, we can extend the block, add additional attribute and CSS accordingly, using the svg as a mask image instead.

miina:
Actually, I'm not sure what would be a good way to change the color of the SVG with just attributes, class names, and CSS. For changing the SVG color it looks like we'd actually have to inject the SVG element into the block's content instead of using it as the background image. Not sure if that's a good idea. Any better ideas?

Alternatively, we could just change the style name to "White" instead and make the backround of the preview dark. Thoughts?
Note that the white quote is used everywhere on the templates where a quote is used, that's why created this style specifically.

miina:
For example:
Screen Shot 2019-04-10 at 12 13 58 AM

swissspidy:
> Actually, I'm not sure what would be a good way to change the color of the SVG with just attributes, class names, and CSS. For changing the SVG color it looks like we'd actually have to inject the SVG element into the block's content instead of using it as the background image.

Using `mask-image` / `-webkit-mask-image` seems to be possible. Works in [all browsers except IE](https://caniuse.com/#feat=css-masks).

CSS:

```css
amp-story-grid-layer .wp-block-quote {
    padding: 10px 10px 10px 10px;
    width: initial;
}

.wp-block-quote:before {
    -webkit-mask: url(https://amp.local/wp-content/plugins/amp/assets/images/quote-black.svg) center 0 no-repeat; /* todo: let autoprefixer do this */
    mask: url(https://amp.local/wp-content/plugins/amp/assets/images/quote-black.svg) center 0 no-repeat;
    background: rebeccapurple;
    content: "";
    display: block;
    width: 100%;
    height: 100px;
}
```

Demo:

Screenshot 2019-04-10 at 10 31 08

The only problem I see there is that pseudo elements cannot be styled via inline CSS...

So yeah, perhaps renaming & changing the preview makes more sense.

miina:
> The only problem I see there is that pseudo elements cannot be styled via inline CSS...

Yep, I also got to the same conclusion when looking into masking. I'll just rename and change the background for now.

assets/css/amp-editor-story-blocks.css
```diff
@@ -37,6 +37,42 @@
...
+}
+
+.block-editor-block-list__layout .block-editor-block-list__block [data-type="core/quote"] .block-editor-block-list__block-edit::before {
+	right: -5px;
+	left: -5px;
+	top: -5px;
+	bottom: -5px;
+}
+
+.editor-block-styles__item.block-editor-block-styles__item[aria-label="White"] .block-editor-block-styles__item-preview,
```
swissspidy: That doesn't work on localized sites. Is there any other way to target the element here?

miina: Actually, we can remove this one rule, since the inner block preview is already targeted. That made it look a bit better but it's true that in some cases it wouldn't work.

Without it, it looks like this:
Screen Shot 2019-04-10 at 3 58 22 PM

miina:
@swissspidy Let me know if you think it's OK to merge now.

PR #2083, opened 2019-04-08
[AMP Stories] Fix style for Text Block when using amp-fit-text

miina:
- Uses `flex-item` as the layout for amp-fit-text to apply padding.
- Adjust line height in the editor to match the FE.


miina:
@swissspidy Note that the line-height of `amp-fit-text` doesn't seem to be changeable so changing the editor style instead. The line height created by AMP Fit Text seems way too small value though and would look better otherwise. Not sure if there's an alternative though.

miina:
@westonruter This is the issue to address the different views for Text block in the FE and editor. Note that the editor is now copying some style from `amp-fit-text` that seems to be unchangeable in the FE:

westonruter:
Looks good with a background color supplied:

image

image

However, if I try removing the background color, then the text is not visible at all:

image

image

miina:
The invisibility should be fixed within f3e7e7d.

PR #2089, opened 2019-04-08
Improve description of the native and transitional template modes

amedina:
This change emphasizes the transient nature of the Transitional mode as a path towards a fully AMP-first site.


westonruter:
@amedina should the “development work” sentence have links back to the documentation on ampwp.org? Once the user dismisses the welcome banner, I realize there is no persistent links back to the site:

Screen Shot 2019-04-11 at 21 56 00

amedina:
Good point. We should include the link back to the site. Will do.

PR #2095, opened 2019-04-09
[AMP Stories] Fix classNames not being read from saved content.

### LINKED ISSUE #2091: [AMP Stories] Block styles cause block validation error
Steps to repeat:
1. Add a block which has multiple styles (e.g. Quote / Text block).
2. Choose an alternative style.
3. Save the story and refresh the page.

miina:
Fixes #2091.

Adds class to the wrapper Grid element to not lose the class names when reading from saved content.


miina:
@swissspidy Looks like the classNames are not being read properly from the selector element and are read from the wrapper element instead.

This is a quick fix for that since it's a blocker right now, let me know if you have a better idea.

miina:
It seems that the classNames should be read based on the selector, perhaps that's something to change in Gutenberg, not sure if there's a reason to read it from the wrapper element instead. Thoughts?

swissspidy:
Looks like `getBlockAttribute` in Gutenberg uses the `html` matcher, which in turn uses `querySelector` to get the first matching element. That seems to explain this behavior.

As an alternative, we could filter `blocks.getBlockAttributes` instead to override the attributes based on our markup.

miina:
@swissspidy Would you mind taking another look?

PR #2072, opened 2019-04-05
[WIP] Add Title block for AMP Stories

kienstra:
* Begins the work to add a title block
* Some improvements are still needed, like possibly making this a dynamic block, and implementing the layout controls

title-block

Closes #2007

Issue #1598, opened 2018-11-08
Add custom category for AMP-specific editor blocks.


miina:
It's now possible to add custom categories to editor (Gutenberg) blocks in addition to the default categories ("Common", "Widgets", etc.):

https://wordpress.org/gutenberg/handbook/extensibility/extending-blocks/#managing-block-categories

Would be useful to add a separate category for AMP blocks so that they would be grouped more appropriately within Inserter.

westonruter:
Is this still relevant?

amedina:
It seems this would this serve general AMP content creation for the case when a site is in Transitional or Native mode and the content creator wants to add only AMP compatible blocks.

miina:
Not sure if this is still relevant, most of the AMP blocks are embed blocks and these are under the Embeds Category which seems to be accurate.

There are two blocks under the Common Category: MathML and Timeago, and one under widgets (Latest Stories, which suit be good under the widgets).

If it's okay to leave MathML and Timeago under the Common Category then we could close this issue.

swissspidy:
> the case when a site is in Transitional or Native mode and the content creator wants to add only AMP compatible blocks

If you are in Transitional mode, the AMP blocks are not available for insertion at all.

If you are in Native mode, then as a user I don't care about AMP or non-AMP blocks. They're just blocks I can insert.

> If it's okay to leave MathML and Timeago under the Common Category then we could close this issue.

Agreed.

PR #2097, opened 2019-04-09
[AMP Stories] Meta Blocks (Title, Author, Date)

### LINKED ISSUE #2007: [AMP Stories] Editor includes all additional blocks to satisfy the Fandom and Travel story creation needs
AC: User can add a star rating block from 1 to 5 stars.
AC: User can add story title, byline with the author, and/or publish date which will be automatically pulled from the post data.

Updated tasks based on the comments below:
- [x] Story title
- [x] Story byline (with the author)
- [x] Story publish date

~~- Create a block for creating a Google Maps Static API~~
~~- Make sure there is a way to fill a block with color/gradient and change its opacity. Check if this can be done by Gutenberg [Section block ](https://github.com/WordPress/gutenberg/pull/13964);~~ This will be implemented as a Page setting instead, see #2013.
~~- Add a Rating block allowing simply to choose a number or stars.~~ This will be implemented separately if / when needed. See #2096 .

swissspidy:
This PR implements the last missing blocks needed for #2007. These meta blocks, as I'd like to call them, allow you to display the post title, author name, and date.

Since these are all very similar data fields, I tried to implement it without duplicating too much code. It works similar to the embed blocks in core.

For this I created a new `withMetaBlockEdit` HOC as well as a `getMetaBlockSettings` helper function that allows passing the attribute name and some other arguments.

Since these are dynamic blocks, no content\* is saved to the database, which means all the inline styles etc. needs to be handled in the PHP render callback.

\* Actually, some markup _is_ stored in the database, as we need to do some calculations for the positioning and things like that. Otherwise we'd need to duplicate this logic in PHP. However, the actual dynamic content is still generated in PHP.

This PR also moves some of the shared components (color settings, font settings, animation settings, resize, rotation) to `withAmpStorySettings`, which allows us to apply these to all movable blocks (i.e. not the CTA block).

**In a nutshell**

* New Post Title block
* New Post Author block
* New Post Date block
* Add available customization options to all movable blocks

**Screenshots**

Screenshot 2019-04-11 at 17 25 02
Screenshot 2019-04-11 at 17 22 57
Screenshot 2019-04-11 at 17 22 44
Screenshot 2019-04-11 at 17 25 53

**Possible follow-up tasks**

* [ ] Make post title block editable
* [ ] Make post date format editable
* [ ] Allow prefixing the post author's name with `By ` (to make it a byline)

Supersedes #2072.
Fixes #2007.


swissspidy:
Looks like the test issues here will be fixed by #2100.

assets/src/components/with-meta-block-edit.js
```diff
@@ -0,0 +1,178 @@
...
+
+					break;
+				case 'author':
+					const author = getAuthors().find( ( { id } ) => id === attributeValue );
+
+					blockContent = author ? author.name : __( 'Unknown', 'amp' );
+
+					break;
+				default:
+					blockContent = attributeValue;
```
miina: For the title, could we use the same placeholder as the title has?

assets/src/blocks/amp-story-post-author/index.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { getMetaBlockSettings } from '../../helpers';
+
+export const name = 'amp/amp-story-post-author';
+
+export const settings = {
+	title: __( 'Story Author', 'amp' ),
```
miina: Maybe for author it would be good to set a bit different defaults, for example, smaller font size and not using AMP Fit Text. Same for the date perhaps. Thoughts?

assets/src/components/with-meta-block-edit.js
```diff
@@ -0,0 +1,178 @@
...
+					setFontSize={ setFontSize }
+					customFontSize={ customFontSize }
+				/>
+				<ColorSettings
+					backgroundColor={ backgroundColor }
+					setBackgroundColor={ setBackgroundColor }
+					textColor={ textColor }
+					setTextColor={ setTextColor }
+					fontSize={ fontSize }
+					opacity={ opacity }
```
miina: Opacity doesn't seem to be working on the editor side.

assets/src/constants.js
```diff
@@ -51,6 +51,9 @@ export const ALLOWED_MOVABLE_BLOCKS = [
 	'core/verse',
 	'core/video',
 	'amp/amp-story-text',
+	'amp/amp-story-post-author',
+	'amp/amp-story-post-date',
+	'amp/amp-story-post-title',
```
miina: Should we register the `rounded` block style for these, too?

assets/src/blocks/amp-story-post-author/index.js
```diff
@@ -0,0 +1,27 @@
...
+	description: __( 'Displays the name of the story\'s author', 'amp' ),
+	category: 'common',
+	icon: 'admin-users',
+	keywords: [
+		__( 'post', 'amp' ),
+		__( 'author', 'amp' ),
```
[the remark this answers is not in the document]
swissspidy: Why would one search for "amp" when wanting to add the byline to an AMP story though?

For the regular post editor a new category can certainly be considered, although this block here is for AMP Stories only. Not sure if that block makes sense for regular posts as well.

assets/src/blocks/amp-story-post-author/index.js
swissspidy: Yeah adding some better defaults makes sense. I haven't added styling yet, so it should be easy.

I thought I had disabled amp-fit-text for all the meta blocks, thinking that the user might want to always have a fixed size for them as the content never really changes, which makes it easy to set the right size once.

assets/src/components/with-meta-block-edit.js
```diff
@@ -0,0 +1,178 @@
...
+			const attributeValue = getEditedPostAttribute( attribute );
+
+			let blockContent;
+
+			// Todo: Maybe pass callbacks as props instead.
+			switch ( attribute ) {
+				case 'date':
+					// Disable reason: false positive because of the two leading underscores.
+					const dateFormat = __experimentalGetSettings().formats.date; // eslint-disable-line no-restricted-syntax
+					blockContent = dateI18n( dateFormat, attributeValue );
```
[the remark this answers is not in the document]
swissspidy: If the published date hasn't been customized, then yes, WordPress should be using the current date.

assets/src/components/with-meta-block-edit.js
swissspidy: You mean `Add title`?

assets/src/constants.js
swissspidy: I think so, makes it consistent. Will add it.

assets/src/blocks/amp-story-post-author/index.js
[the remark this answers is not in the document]
miina: > Why would one search for "amp" when wanting to add the byline to an AMP story though?

To find all AMP-related blocks? It would be good to have a way how to filter out AMP related blocks only. Thoughts?

assets/src/components/with-meta-block-edit.js
miina: Yes. Otherwise the block is empty and basically invisible when not selected.

assets/src/blocks/amp-story-post-author/index.js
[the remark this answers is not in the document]
swissspidy: I see that you opened #1598 for that before. Let's use that for discussion 👍

assets/src/components/with-meta-block-edit.js
swissspidy: Ugh, indeed.

I am getting `A component is changing an uncontrolled input of type number to be controlled` erros in the console. Probably related to another issue I just saw with the styling, caused by the components now being nested within an extra div:

Screenshot 2019-04-10 at 10 56 35

Interestingly there's no console error for the text block, but the styling is off there as well.

swissspidy: Fixed now with the latest refactoring

assets/src/blocks/amp-story-post-author/index.js
swissspidy: Improved styling a bit and removed the initial font size which I copied over from the text block.

assets/src/blocks/amp-story-post-author/index.js
```diff
@@ -0,0 +1,28 @@
...
+ * Internal dependencies
+ */
+import { getMetaBlockSettings } from '../../helpers';
+import './edit.css';
+
+export const name = 'amp/amp-story-post-author';
+
+export const settings = {
+	title: __( 'Story Author', 'amp' ),
+	description: __( 'Displays the name of the story\'s author', 'amp' ),
```
westonruter: Should this not use a non-straight quote like `story‘s`?

assets/src/blocks/amp-story-post-date/index.js
```diff
@@ -0,0 +1,28 @@
...
+ * Internal dependencies
+ */
+import { getMetaBlockSettings } from '../../helpers';
+import './edit.css';
+
+export const name = 'amp/amp-story-post-date';
+
+export const settings = {
+	title: __( 'Story Date', 'amp' ),
+	description: __( 'Displays the date the story has been published on', 'amp' ),
```
westonruter: Suggestion:

> Displays the publish date of the story.

assets/src/blocks/amp-story-post-title/index.js
```diff
@@ -0,0 +1,29 @@
...
+	category: 'common',
+	icon: <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path d="M5 4v3h5.5v12h3V7H19V4z" /><path fill="none" d="M0 0h24v24H0V0z" /></svg>,
+	keywords: [
+		__( 'post', 'amp' ),
+		__( 'title', 'amp' ),
+	],
+	...getMetaBlockSettings( {
+		tagName: 'h1',
+		attribute: 'title',
+		placeholder: __( 'Add title', 'amp' ),
```
westonruter: I suggest this should say:

> (Enter story above)

Since it can't be entered here.

assets/src/blocks/amp-story-post-date/index.js
```diff
@@ -0,0 +1,28 @@
...
+	category: 'common',
+	icon: 'clock',
+	keywords: [
+		__( 'post', 'amp' ),
+		__( 'date', 'amp' ),
```
westonruter: Add `publish` as a keyword.

assets/src/blocks/amp-story-post-author/index.js
```diff
@@ -0,0 +1,28 @@
...
+	category: 'common',
+	icon: 'admin-users',
+	keywords: [
+		__( 'post', 'amp' ),
+		__( 'author', 'amp' ),
+	],
+	...getMetaBlockSettings( {
```
westonruter: Since this block could also be considered to be a “Byline” block, there may be a desire to show "By" before the name. Perhaps a toggle to include “By” before the name? This would prevent users from having to try to carefully position a separate Text block that just contains “By”. But in that case, the user can just use a Text block for the while byline manually if they want in the immediate term.

assets/src/blocks/amp-story-post-title/index.js
```diff
@@ -0,0 +1,29 @@
...
+ * Internal dependencies
+ */
+import { getMetaBlockSettings } from '../../helpers';
+import './edit.css';
+
+export const name = 'amp/amp-story-post-title';
+
+export const settings = {
+	title: __( 'Story Title', 'amp' ),
+	description: __( 'Displays the story title', 'amp' ),
```
westonruter: The descriptions from core blocks look to start with an imperative like “Display” and always end with a period.

westonruter:
Awesome work. This is so cool.

kienstra:
**Title, Author, And Date Blocks Work As Expected**
# Story Title Block
meta-title-block

# Story Author Block
meta-author-block

# Story Date Block
meta-date-block

assets/src/blocks/amp-story-post-date/index.js
swissspidy: Maybe `publish date`? Otherwise difficult to translate without context

assets/src/blocks/amp-story-post-author/index.js
swissspidy: I don't think we use these kind of quotes anywhere in the plugin nor in WordPress core. But I might use double quotes so I can get rid of the backslash.

westonruter: I did a search. They're being used in Gutenberg.

assets/src/blocks/amp-story-post-title/index.js
swissspidy: Did you mean `(Enter title above)`?

assets/src/blocks/amp-story-post-author/index.js
westonruter: For example: https://github.com/WordPress/gutenberg/blob/0b93ad47797a1ba2cea3c83a7f973d33abcd351f/packages/editor/src/components/post-locked-modal/index.js#L171

assets/src/blocks/amp-story-post-title/index.js
westonruter: Yes. That's actually what I put in my commit 😊

assets/src/blocks/amp-story-post-date/index.js
westonruter: Cool. Fixed in 277bb170802910952ee491a6beb11259716bfb63

assets/src/blocks/amp-story-post-author/index.js
westonruter: I've always been told to change straight quotes to curly quotes when I've committed such strings to core.

assets/src/blocks/amp-story-post-author/index.js
swissspidy: Good idea.

Since I already created [a card for further customizations of these blocks](https://github.com/ampproject/amp-wp/projects/10#card-20136438), I added this suggestion to the list. It requires some thinking of how to best make the meta blocks extensible in an easy way without duplicating too much code. Which is why I think we should tackle it in a separate PR.

assets/src/blocks/amp-story-post-date/index.js
```diff
@@ -18,8 +18,8 @@ export const settings = {
 	icon: 'clock',
 	keywords: [
 		__( 'post', 'amp' ),
-		__( 'publish', 'amp' ),
-		__( 'published', 'amp' ),
+		__( 'publish date', 'amp' ),
+		__( 'published date', 'amp' ),
```
swissspidy: @westonruter This is kind of redundant. Someone typing `publish` will already see the block (and probably only this block). I doubt someone will type `published` after that.

westonruter: I actually tried to enter “published” and I was surprised when the block disappeared. So I can attest to it being of some value.

westonruter: But I don't feel strongly about it.

assets/src/blocks/amp-story-post-author/index.js
westonruter: Good. Let's address later. Not critical.

assets/src/blocks/amp-story-post-date/index.js
swissspidy: Alright. Let's leave it for now.

PR #2104, opened 2019-04-10
Adjust positioning of the 'Background Media' image on the front-end

### LINKED ISSUE #2085: [AMP Stories] Focal point of the background image is not right.
Steps to reproduce:
- Add a story Page with background image and set focal point.
- Compare the results in the front-end versus editor:

Editor:
Screen Shot 2019-04-08 at 7 32 51 PM

FE: 
Screen Shot 2019-04-08 at 7 33 16 PM

kienstra:
As @miina noted in #2085, the values in the Focal Point Picker weren't appearing on the front-end for images:
fp-picker

* This applies the `focalPoint` values on the front-end
* But it uses `object-position` instead of `background-position`

images-same

Closes #2085

### DIFF
--- a/assets/css/amp-stories.css
+++ b/assets/css/amp-stories.css
@@ -50,3 +50,7 @@ amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
 .is-style-half-rounded {
 	border-radius: 4px 4px 26px 26px;
 }
+
+amp-story-grid-layer amp-img > img {
+	object-position: inherit;
+}
--- a/assets/src/blocks/amp-story-page/index.js
+++ b/assets/src/blocks/amp-story-page/index.js
@@ -82,6 +82,7 @@ export const settings = {
 	save( { attributes } ) {
 		const {
 			anchor,
+			focalPoint,
 			overlayOpacity,
 			mediaUrl,
 			mediaType,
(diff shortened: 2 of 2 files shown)


assets/css/amp-stories.css
```diff
@@ -50,3 +50,7 @@ amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
 .is-style-half-rounded {
 	border-radius: 4px 4px 26px 26px;
 }
+
+amp-story-grid-layer amp-img > img {
+	object-position: inherit;
```
kienstra: This is apparently needed because the `` didn't seem to inherit the `object-position` value without this.

assets/src/blocks/amp-story-page/index.js
```diff
@@ -107,13 +108,17 @@ export const settings = {
 			overlayStyle.opacity = overlayOpacity / 100;
 		}
 
+		const imgStyle = {
+			objectPosition: IMAGE_BACKGROUND_TYPE === mediaType && focalPoint ? `${ focalPoint.x * 100 }% ${ focalPoint.y * 100 }%` : 'initial',
```
kienstra: This is mainly taken [from edit.js](https://github.com/ampproject/amp-wp/blob/0a883ce83c15c22407bebef69db1bd5808d07247/assets/src/blocks/amp-story-page/edit.js#L162).

kienstra:
**Request For Review**

Hi @swissspidy,
Would you be available to review this?

swissspidy:
Hmm I am now getting "Received NaN for the `value` attribute. If this is expected, cast the value to a string." warnings in the console when adding a background image. Might be related to this change here.

@kienstra Mind having a look? Thanks!

kienstra:
Hi @swissspidy, thanks for bringing that up. Sure, I'll work on it.

kienstra:
It looks like this might be caused by the `<FocalPointPicker>` not having an object value sometimes. It looks like its `value` [attribute is required](https://github.com/WordPress/gutenberg/tree/4472b1bf3da10f3e382b68f6e7f8894757d7c395/packages/components/src/focal-point-picker#value). I'm looking at adding a default value for it.

kienstra:
@swissspidy, sorry for the delay. Would it be alright if I delayed looking at this until later this week? I'm probably going to focus on the demo.

swissspidy:
@kienstra Of course! That's good to know. We can have a look at some other point then.

PR #2106, opened 2019-04-10
Override inserter in header toolbar with custom implementation

### LINKED ISSUE #1966: [AMP Stories] Fix default Inserter in the Toolbar

swissspidy:
There's no real way to override the inserter in the header toolbar inserter, so I decided to just completely replace it.

This PR includes the whole `Inserter` component including sub components from Gutenberg, and customizes it a little bit to:

1. Always show all available blocks in the inserter, i.e. page and text blocks are always shown, but the CTA block is only shown when not on the first page.
2. Insert blocks in the correct place when selecting an item

The default inserter in Gutenberg shows some nice placeholders when hovering over items, to give the user an idea where it would be inserted. This is not implemented here, but could be done in the future.

For reviewing this PR, it might be easier to go through the individual commits, in order to avoid reviewing code that has just been imported 1:1 and not modified at all. The most relevant section is basically the `withSelect` and `withDispatch` part in `InserterMenu`.

Fixes #1966.


swissspidy:
> The only comment is that at this moment the Page block never seems to visible right away when opening the Inserter, even if a Page is selected or even if the story Title is selected, it's always necessary to scroll for being able to add a Page. Would it be possible to display it among the rest of the available blocks in the first view?

Turns out the default inserter displays available child blocks first if available (i.e. if there's a `rootClientId`). Since with this PR we trick the inserter into always having a `rootClientId`, these are always shown.

Removing that child blocks section moves the most used blocks to the top, which makes for a better experience:

Screenshot 2019-04-11 at 12 33 26

PR #2116, opened 2019-04-12
Add links to docs from mode descriptions

amedina:
Once the user dismisses the welcome banner there is no persistent links back to the site guiding users to documentation sources. Adding links back to development workflow guidance will help users when engaging with the template mode settings.

### DIFF
--- a/includes/options/class-amp-options-menu.php
+++ b/includes/options/class-amp-options-menu.php
@@ -140,8 +140,8 @@ class AMP_Options_Menu {
 	public function render_theme_support() {
 		$theme_support = AMP_Options_Manager::get_option( 'theme_support' );
 
-		$native_description       = __( 'Integrates AMP as the framework for your site by using the active’s theme templates and styles to render AMP responses. This means your site is <b>AMP-first</b> and your canonical URLs are AMP! Depending on your theme/plugins, a varying level of development work may be required.', 'amp' );
-		$transitional_description = __( 'Uses the active theme’s templates to generate non-AMP and AMP versions of your content, allowing for each canonical URL to have a corresponding (paired) AMP URL. This mode is useful to progressively transition towards a fully AMP-first site. Depending on your theme/plugins, a varying level of development work may be required.', 'amp' );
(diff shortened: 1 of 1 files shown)


swissspidy:
I'd suggest using `sprintf()` here and using placeholders for the actual URLs. It's a general best practice thing because this way, if the URLs to the documentation change at any point for some reason, the strings don't have to change, meaning all the translations will still be valid.

Example:

```php
/* translators: %s: URL to the documentation. */
$native_description = sprintf( __( 'Integrates AMP as the framework for your site by using the active’s theme templates and styles to render AMP responses. This means your site is <b>AMP-first</b> and your canonical URLs are AMP! Depending on your theme/plugins, a varying level of <a href="%s">development work</a> may be required.', 'amp' ), 'https://amp-wp.org/documentation/developing-wordpress-amp-sites/' );
```

includes/options/class-amp-options-menu.php
```diff
@@ -140,8 +140,8 @@ public function add_menu_items() {
...
+		$native_description       = __( 'Integrates AMP as the framework for your site by using the active’s theme templates and styles to render AMP responses. This means your site is <b>AMP-first</b> and your canonical URLs are AMP! Depending on your theme/plugins, a varying level of <a href="https://amp-wp.org/documentation/developing-wordpress-amp-sites/">development work</a> may be required.', 'amp' );
+		$transitional_description = __( 'Uses the active theme’s templates to generate non-AMP and AMP versions of your content, allowing for each canonical URL to have a corresponding (paired) AMP URL. This mode is useful to progressively transition towards a fully AMP-first site. Depending on your theme/plugins, a varying level of <a href="https://amp-wp.org/documentation/developing-wordpress-amp-sites/">development work</a> may be required.', 'amp' );
```
westonruter: @swissspidy Should the URL be a string format placeholder since translators should not be changing it, as we don't have a localized site?

westonruter: Oops, I see you commented as such in https://github.com/ampproject/amp-wp/pull/2116#issuecomment-482507052

PR #2117, opened 2019-04-12
[AMP Stories]  Re-render components when switching back from code editor

swissspidy:
When working on the default templates I was annoyed that the visual editor was kinda broken when switching back from the code editor.

The reason is that many of the custom DOM nodes get lost in that process.

This PR re-renders the components when switching back and forth between the two modes. It's admittedly a bit hacky, because it needs to wait for the editor to finish render first. I haven't found a better solution yet.


swissspidy:
Ugh. Looks like a `! document.getElementById(...)` check is also needed for some of the other components.

kienstra:
**Approved**

Hi @swissspidy,
This looks good, and switching back to the Visual Editor works well now.

visual-editor-switching

westonruter:
This is a very welcome change. I was noticing the UI getting blown away when switching to the code editor and back as well. 

@swissspidy Let us know when you've addressed the other issue you noted.

PR #2118, opened 2019-04-12
[AMP Stories] Further improve block navigation

swissspidy:
* Adds support for new meta blocks
* Uses image caption and title attribute for the label if available
* Adds some more padding

Screenshot:

Screenshot 2019-04-12 at 13 01 08

### DIFF
--- a/assets/src/components/block-preview-label.js
+++ b/assets/src/components/block-preview-label.js
@@ -2,24 +2,47 @@
  * WordPress dependencies
  */
 import { Fragment } from '@wordpress/element';
+import { getBlockType } from '@wordpress/blocks';
 import { BlockIcon } from '@wordpress/block-editor';
+import { withSelect } from '@wordpress/data';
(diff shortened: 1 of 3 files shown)


assets/src/components/block-preview-label.js
```diff
@@ -2,41 +2,79 @@
...
 					label = content;
 				}
 			}
 
+			if ( block.attributes.id ) {
+				const media = getMedia( block.attributes.id );
+
+				if ( media ) {
+					label = media.caption.raw || media.title.raw || label;
+				}
```
amedina: Perhaps we could change the order to use the title if available, otherwise use the caption, and use the id as a baseline.

westonruter:
I'll go ahead and merge this to deploy for now, with the additional tweak being a follow-up task.

assets/src/components/block-preview-label.js
swissspidy: IDs are never exposed to the user in WordPress, so I‘d leave that one out.

PR #2131, opened 2019-04-14
Add sanitizer arg to prevent noscript fallbacks; preemptively remove disallowed attributes from fallback

### LINKED ISSUE #2126: Force amp-img > noscript > img fallbacks to have HTTPS src/srcset
If an image is in post content with an `http` protocol, this will persist into the `noscript > img` fallback when it is converted to `amp-img` (see #1832). While an `amp-img` allows `http` protocol, the `noscript > img` does not. See fix to AMP spec here https://github.com/ampproject/amphtml/pull/21686. To prevent a lot of validation errors from appearing for sites that have HTTP images, we should consider automatically converting the `noscript > img` elements to HTTPS.

westonruter:
When a disallowed attribute appears on an input `img`, `video`, `audio`, or `iframe` currently the validation error will be duplicated once for the replaced AMP component and again for the original element added inside as a `noscript`. This PR prevents a duplication of the validation error by removing the disallowed attributes preemptively.

Also, in case the `noscript` fallback elements are not desired, a `add_noscript_fallback` arg is introduced for the image, audio, video, and iframe sanitizers to prevent it from being added (this is needed for AMP stories, and these commits were picked from #2107).

Lastly, as reported in https://github.com/ampproject/amphtml/pull/21686 the `amp-img` allows the HTTP protocol whereas the `amp-img > noscript > img` does not. So this PR prevents adding the `noscript > img` fallback if it lacks the HTTPS protocol. 

Fixes #2126.

### DIFF
--- a/includes/sanitizers/class-amp-audio-sanitizer.php
+++ b/includes/sanitizers/class-amp-audio-sanitizer.php
@@ -11,6 +11,7 @@
  * Converts <audio> tags to <amp-audio>
  */
 class AMP_Audio_Sanitizer extends AMP_Base_Sanitizer {
+	use AMP_Noscript_Fallback;
 
 	/**
 	 * Tag.
@@ -21,24 +22,16 @@ class AMP_Audio_Sanitizer extends AMP_Base_Sanitizer {
 	public static $tag = 'audio';
 
 	/**
-	 * Placeholder for default args.
+	 * Constructor.
 	 *
-	 * @since 1.1
-	 *
-	 * @var array
+	 * @param DOMDocument $dom  DOMDocument $dom Represents the HTML document to sanitize.
+	 * @param array       $args Optional. Sanitizer arguments. See {@see AMP_Base_Sanitizer::__construct()}.
 	 */
-	protected $DEFAULT_ARGS = array(
-		'add_noscript_fallback' => true,
-	);
+	public function __construct( $dom, $args = array() ) {
+		$this->DEFAULT_ARGS = $this->merge_default_args( $this->DEFAULT_ARGS );
 
-	/**
-	 * Attributes allowed on noscript fallback elements.
-	 *
-	 * This is used to prevent duplicated validation errors.
-	 *
-	 * @var array
-	 */
-	private $noscript_fallback_allowed_attributes = array();
+		parent::__construct( $dom, $args );
+	}
 
 	/**
 	 * Get mapping of HTML selectors to the AMP component selectors which they may be converted into.
@@ -63,19 +56,13 @@ class AMP_Audio_Sanitizer extends AMP_Base_Sanitizer {
 			return;
 		}
 
-		$this->noscript_fallback_allowed_attributes = array_fill_keys(
-			array_merge(
-				array_keys( current( AMP_Allowed_Tags_Generated::get_allowed_tag( self::$tag ) )['attr_spec_list'] ),
-				array_keys( AMP_Allowed_Tags_Generated::get_allowed_attributes() )
-			),
-			true
-		);
+		$this->initialize_allowed_attributes( self::$tag );
 
 		for ( $i = $num_nodes - 1; $i >= 0; $i-- ) {
 			$node = $nodes->item( $i );
 
 			// Skip element if already inside of an AMP element as a noscript fallback.
-			if ( 'noscript' === $node->parentNode->nodeName && $node->parentNode->parentNode && 'amp-' === substr( $node->parentNode->parentNode->nodeName, 0, 4 ) ) {
+			if ( $this->is_inside_amp_noscript( $node ) ) {
 				continue;
 			}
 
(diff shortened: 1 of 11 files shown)


includes/sanitizers/class-amp-audio-sanitizer.php
```diff
@@ -135,10 +163,24 @@ public function sanitize() {
...
+					// Remove all non-allowed attributes preemptively to prevent doubled validation errors.
+					$disallowed_attributes = array();
+					foreach ( $old_node->attributes as $attribute ) {
+						if ( ! isset( $this->noscript_fallback_allowed_attributes[ $attribute->nodeName ] ) ) {
+							$disallowed_attributes[] = $attribute->nodeName;
+						}
+					}
+					foreach ( $disallowed_attributes as $disallowed_attribute ) {
+						$old_node->removeAttribute( $disallowed_attribute );
+					}
```
felixarntz: Why not iterate only once and immediately remove the attributes instead of populating first and then doing another round of removing? If the problem is remove attributes modifying the `$old_node->attributes` property, then we could assign it to another variable to use.

westonruter: The problem is that the attributes are a live DOM list and if a node is removed while iterating then a node gets skipped. This pattern of gathering up the list of nodes and then removing them on a second iteration can be seen elsewhere in the codebase. It's not ideal.

westonruter:
We could also create an `AMP_Media_Element_Sanitizer` abstract class that these 4 sanitizers inherit from. Not sure if that is the best name, but I think that might make more sense than a trait in this case.

felixarntz:
I think the `noscript` fallback is something rather specific, so I think a trait makes sense. Also it doesn't lock us in as it kind of allows polymorphism in PHP. If we use a base class, we're tied to it, while with a trait we can go more granular.

For example, if other elements than media could benefit from that noscript fallback in the future, we could use the trait, but a media base class wouldn't fit in that case.

The latest commit includes a trait that does it. We could still easily modify it to become a base class if this is the decision.

includes/sanitizers/trait-amp-noscript-fallback.php
```diff
@@ -0,0 +1,101 @@
...
+
+	/**
+	 * Extends the default arguments array with additional defaults.
+	 *
+	 * @since 1.1
+	 *
+	 * @param array $default_args Original defaults.
+	 * @return array Extended defaults.
+	 */
+	protected function merge_default_args( array $default_args ) {
```
westonruter: Is this the right name? When we have other traits then they would potentially have their own default args to merge, so then this should be named uniquely for this trait.

westonruter: Maybe this method isn't even needed? Maybe better to just have each class that uses it define the default arg like they do for other args?

includes/sanitizers/trait-amp-noscript-fallback.php
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	/**
+	 * Initializes the internal allowed attributes array.
+	 *
+	 * @since 1.1
+	 *
+	 * @param string $tag Tag name to get allowed attributes for.
+	 */
+	protected function initialize_allowed_attributes( $tag ) {
```
westonruter: This is very specific to the noscript fallbacks, as it assumes there is it one tag spec for the given tag, which is true for the noscript fallback tag specs. In reality AMP has multiple tag specs for each tag name. So maybe this should also be renamed to be specific to the noscript fallback case.

includes/sanitizers/trait-amp-noscript-fallback.php
westonruter: Also, some sanitizers may want different defaults.

includes/sanitizers/class-amp-audio-sanitizer.php
```diff
@@ -43,11 +53,13 @@ public function sanitize() {
 			return;
 		}
 
+		$this->initialize_noscript_allowed_attributes( self::$tag );
```
westonruter: This should only happen if the `add_noscript_fallback` arg is true, no?

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -73,11 +75,13 @@ public function sanitize() {
 			return;
 		}
 
+		$this->initialize_noscript_allowed_attributes( self::$tag );
```
westonruter: See above

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -81,14 +91,16 @@ public function sanitize() {
 			return;
 		}
 
+		$this->initialize_noscript_allowed_attributes( self::$tag );
```
westonruter: See above

includes/sanitizers/class-amp-video-sanitizer.php
```diff
@@ -56,11 +66,18 @@ public function sanitize() {
 			return;
 		}
 
+		$this->initialize_noscript_allowed_attributes( self::$tag );
```
westonruter: See above

PR #2139, opened 2019-04-17
Add checkbox to enable/disable AMP Stories (disabled by default)

### LINKED ISSUE #2136: [AMP Stories] Add opt-in toggle for enabling AMP stories

westonruter:
This adds a checkbox to enable the AMP Stories feature. The feature is disabled by default.

AMP Stories toggle

Fixes #2136.

### DIFF
--- a/includes/options/class-amp-options-menu.php
+++ b/includes/options/class-amp-options-menu.php
@@ -109,6 +109,17 @@ class AMP_Options_Menu {
 			)
 		);
 
+		add_settings_field(
+			'amp_stories',
+			__( 'AMP Stories', 'amp' ),
+			array( $this, 'render_amp_stories' ),
+			AMP_Options_Manager::OPTION_NAME,
+			'general',
+			array(
+				'class' => 'amp-stories-field',
+			)
+		);
+
 		if ( wp_using_ext_object_cache() ) {
 			add_settings_field(
 				'caching',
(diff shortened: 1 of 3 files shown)


includes/options/class-amp-options-menu.php
```diff
@@ -495,6 +506,33 @@ function updateFieldsetVisibility() {
...
+				<?php esc_html_e( 'Enable experimental support for the AMP Story post type.', 'amp' ); ?>
+			</label>
+		</p>
+		<p class="description">
+			<?php
+			echo wp_kses_post(
+				sprintf(
+					/* translators: %s is the AMP Stories landing page */
+					__( 'AMP Stories is a new visual storytelling for the open web. AMP stories immerse your readers in fast-loading full-screen experiences. Easily create visual narratives, with engaging animations and tappable interactions. The AMP story format is free and part of the open web and are available for everyone to try on their websites. They can be shared and embedded across the web without being confined to a closed ecosystem or platform. Read more about <a href="%s">AMP Stories</a>.', 'amp' ),
+					esc_url( 'https://amp.dev/about/stories' )
```
amedina: Making the text shorter would be nice. WDYT about something like this: "AMP Stories is a visual storytelling format for the open web, which immerse your readers in fast-loading, full-screen, and visually rich experiences. Stories can be a great addition to your overall content strategy. Read more about <a href="%s">AMP Stories</a>."

includes/options/class-amp-options-menu.php
```diff
@@ -495,6 +506,33 @@ function updateFieldsetVisibility() {
...
+	 * AMP Stories section renderer.
+	 *
+	 * @since 1.2
+	 */
+	public function render_amp_stories() {
+		?>
+		<p>
+			<label for="enable_amp_stories">
+				<input id="enable_amp_stories" type="checkbox" name="<?php echo esc_attr( AMP_Options_Manager::OPTION_NAME . '[enable_amp_stories]' ); ?>" <?php checked( AMP_Options_Manager::get_option( 'enable_amp_stories' ) ); ?>>
+				<?php esc_html_e( 'Enable experimental support for the AMP Story post type.', 'amp' ); ?>
```
swissspidy: I'd just say "Stories" instead of "the AMP Story post type"

includes/options/class-amp-options-menu.php
```diff
@@ -495,6 +506,33 @@ function updateFieldsetVisibility() {
...
+			</label>
+		</p>
+		<p class="description">
+			<?php
+			echo wp_kses_post(
+				sprintf(
+					/* translators: %s is the AMP Stories landing page */
```
swissspidy:
```suggestion
					/* translators: %s: AMP Stories documentation URL. */
```

includes/options/class-amp-options-menu.php
swissspidy: We should also mention that this should be used in combination with WordPress 5.2 Beta or the current version of the Gutenberg plugin.

swissspidy:
I just fixed the `Test_AMP_Options_Manager::test_get_and_set_options` test, now there's only `Test_AMP_Options_Manager::test_persistent_object_caching_notice` and `AMP_Story_Post_Type_Test::test_render_block_latest_stories` failing. The former seems a bit unstable there, the latter is because the option isn't set and we don't force AMP for stories yet.

includes/options/class-amp-options-menu.php
westonruter: I think we could just hide the setting entirely if `register_block_type` is not defined. In reality we're going to drop support for WP 4.9 anyway, so I don't think this is a big concern.

westonruter: I stand corrected. WordPress 5.2 or the latest Gutenberg is required.

PR #2140, opened 2019-04-17
Fix some more i18n strings and placeholders

swissspidy:


westonruter:
Cherry-picked commits into 1.1 branch via:

```bash
git cherry-pick e059d40^..30c6ceb
```

Then squashed them prior to pushing directly onto the 1.1 branch.

PR #2143, opened 2019-04-17
Allow AMP Stories to work in Reader mode

### LINKED ISSUE #2137: [AMP Stories] Enable independent from AMP mode
We need to make sure that AMP stories can also be used in Classic (Reader) mode or perhaps even standalone.

westonruter:
Fixes #2137.

### DIFF
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -81,7 +81,7 @@ class AMP_Story_Post_Type {
 	 * @return void
 	 */
 	public static function register() {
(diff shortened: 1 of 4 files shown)


amedina:
Yes! This is super necessary.

westonruter:
Note: One side effect of the changes here (specifically 4645a4a) is that making changes to non-story posts when in Reader mode will start to show the validation error messages. Previously the validation warnings would only be shown when in Native/Transitional mode. The validation notices will likely be a source of complaints for users when in Reader mode, since they would be reported as if the Transitional mode is active.

PR #2148, opened 2019-04-18
Introduce admin pointers for AMP Stories

felixarntz:
This PR adds two new admin pointers:
* One for the newly available setting to enable AMP Stories (displayed if AMP Stories are not enabled yet)
* Another one for the AMP Stories menu item (displayed if AMP Stories enabled).

As with all admin pointers, they both can be permanently dismissed on a per-user basis. Since we now have multiple notices, two new classes are introduced and replace the previously used class for the old admin pointer (the class code was pretty much taken from Site Kit).

### DIFF
--- a/includes/admin/class-amp-admin-pointer.php
+++ b/includes/admin/class-amp-admin-pointer.php
@@ -1,149 +1,210 @@
 <?php
 /**
- * Admin pointer class.
+ * Class AMP_Admin_Pointer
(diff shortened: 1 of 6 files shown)


westonruter:
This introduced a regression. See https://github.com/ampproject/amp-wp/pull/2149#pullrequestreview-228515339

PR #2150, opened 2019-04-18
Use optimized JPEGs instead of PNGs for demo images

swissspidy:
Reduces image size quite a bit


westonruter:
Very good. This reduces the `amp.zip` archive size from 7.4MB to 2.3MB.

Issue #2161, opened 2019-04-22
[AMP Stories] Drag-drop and resizing are not accurate after the editor width/height change


miina:
The width of the Page was changed, however, dragging-dropping blocks and resizing the blocks is still considering the previous width in calculations.

Todo: Update the calculations to consider the new width and height values.

miina:
Instructions for testing:
- Drag any block, especially more to the right side.
- Verify that the block is moved to the correct position and it doesn's make a jump when finishing dragging.

csossi:
Verified in QA

Am able to drag different block types to new positions

Example:
![image](https://user-images.githubusercontent.com/16945664/57864761-7a45ba80-77ca-11e9-9bcc-1efc850655f4.png)

PR #2166, opened 2019-04-22
[AMP Stories] Make resizing responsive.

### LINKED ISSUE #2159: [AMP Stories] Resizing core blocks is not responsive
Resizing any block should be responsive and appear in the FE similarly as in the editor. Currently, when resizing an image block or a quote block for example it displays incorrectly in the FE.

In the editor:
Screen Shot 2019-04-22 at 10 04 22 AM

In the FE:
Screen Shot 2019-04-22 at 10 04 03 AM

miina:
Fixes #2159.

Includes #2161 since this, too, requires the width and height of the Page in editor to be set properly.

Note that this introduces breaking changes and thus also updates the templates.


westonruter:
Positioning looks good:

image

image

westonruter:
I'm noticing one thing unrelated to this here. After I insert an image, if I then go to resize it, then it appears stretched in the frontend:

image

image

westonruter:
> I'm noticing one thing unrelated to this here. After I insert an image, if I then go to resize it, then it appears stretched in the frontend

Fixed in 40bea8cf.

westonruter:
Oh, actually perhaps it _is_ related as I see now there is a new rule for:

```css
.wp-block-image.is-resized amp-img
```

Issue #2114, opened 2019-04-11
Image caption is centered in story editor but appears left-justified on frontend


westonruter:
![image](https://user-images.githubusercontent.com/134745/55998670-1df3f600-5c74-11e9-8338-e3bfb02cc0c3.png)

![image](https://user-images.githubusercontent.com/134745/55998679-277d5e00-5c74-11e9-8c96-f5412756dffd.png)

swissspidy:
Caption position is identical on frontend and backend 👍

PR #2167, opened 2019-04-22
[AMP Stories] Remove image caption by default.

### LINKED ISSUE #2113: [AMP Stories] Initial attachment caption for image block unexpectedly appears in story on frontend when unmodified
Given an image that has a caption populated in the media library:

image

In the editor this caption is not displayed and the caption toggle is turned off (by default):

image

Nevertheless, on the frontend the caption is being displayed:

image

If I toggle the caption on and off then it is truly removed as expected.

It seems that when adding an image from the media library, the caption's `figcaption` is present in the markup even though the toggle is off, so it is hidden. Toggling the caption on and off again seems to clear out the underlying `figcaption`.

miina:
Fixes #2113 & #2114.

Clears the image caption by default.
Style caption in the FE.

### DIFF
--- a/assets/src/components/with-amp-story-settings.js
+++ b/assets/src/components/with-amp-story-settings.js
@@ -176,7 +176,7 @@ export default createHigherOrderComponent(
 
 			// If we have an image with pre-set caption we should remove it.
 			if ( isImageBlock && ! ampShowImageCaption && attributes.caption && 0 !== attributes.caption.length ) {
-				props.setAttributes( { caption: '' } );
+				setAttributes( { caption: '' } );
 			}
 
 			const minTextHeight = 20;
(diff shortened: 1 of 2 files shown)


assets/src/components/with-amp-story-settings.js
```diff
@@ -173,6 +173,11 @@ export default createHigherOrderComponent(
 				rotationAngle,
 			} = attributes;
 
+			// If we have an image with pre-set caption we should remove it.
+			if ( isImageBlock && ! ampShowImageCaption && attributes.caption && 0 !== attributes.caption.length ) {
+				props.setAttributes( { caption: '' } );
```
swissspidy: Can just use `setAttributes` directly FWIW

PR #2174, opened 2019-04-24
[AMP Stories] Fix displaying a saved AMP Story as unsaved when loading

miina:
This PR Moves setting the initial state of animation order out of `domReady`.

Also replaces the `core/editor` with `core/block-editor` where it's still used.

### DIFF
--- a/assets/src/helpers.js
+++ b/assets/src/helpers.js
@@ -45,7 +45,7 @@ const {
 	getBlockOrder,
 	getBlock,
 	getClientIdsWithDescendants,
-} = select( 'core/editor' );
+} = select( 'core/block-editor' );
 
 const {
 	addAnimation,
@@ -58,7 +58,7 @@ const {
 	getAnimatedBlocks,
 } = select( 'amp/story' );
 
-const { updateBlockAttributes } = dispatch( 'core/editor' );
+const { updateBlockAttributes } = dispatch( 'core/block-editor' );
 
 export const maybeEnqueueFontStyle = ( name ) => {
 	if ( ! name || 'undefined' === typeof ampStoriesFonts ) {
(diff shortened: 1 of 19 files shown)


miina:
@swissspidy Looks like setting the initial animation order was causing a saved AMP Story displaying with active "Update" button due to state change on load.

Moved the logic out of `domReady` to fix that, let me know if you have a better idea for that.

assets/src/helpers.js
```diff
@@ -18,6 +18,8 @@ import { getColorClassName, getColorObjectByAttributeValues, RichText } from '@w
 /**
  * Internal dependencies
  */
+// eslint-disable-next-line no-unused-vars
+import store from './stores/amp-story';
```
miina: Importing this to ensure the store is registered before using it. Importing it within `amp-story-editor-blocks.js` even before importing the helpers doesn't seem to be enough.

swissspidy: Does `import './stores/amp-story';` alone just work to remove the need for `eslint-disable`?

miina: Ah, good point, it should work 💡

swissspidy:
The problem with `subscribe()` is that it is called _a lot_. Putting it outside of that did not work? You're using `isEqual` though, so I guess it should be fine.

I wonder though if there's a better way to handle this in general 🤔  i.e. waiting for the editor store to be initialized.

> Also replaces the `core/editor` with `core/block-editor` where it's still used.

Nice! Long overdue :-)

miina:
> Putting it outside of that did not work?

Nope, unfortunately not.

swissspidy:
Hmm I guess we'll run with this for now then.

PR #2183, opened 2019-04-24
Add curl as required extension; refactor how dependency errors are communicated

westonruter:
Someone [reported](https://wordpress.org/support/topic/getting-fatal-error-faster-image/) fatal errors due to the `curl_multi_init()` function not existing. This appears due to the `curl` extension not being loaded. So this PR adds a check for `curl` being installed.

Extensions are also now checked for being loaded, rather than feature detecting classes and functions.

Additionally, there was a lot of duplicated code in the logic for rendering the dependency errors. So this PR refactors the logic not only to remove the duplication but to ensure that all dependency errors are shown together, instead of one-by-one.

For example, if installing the plugin from source and not having done a build, while also the `curl` extension is not installed, the user is shown the following admin notice:

image

And now as well there is a warning that is shown when interacting with WP-CLI:

image

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -14,102 +14,111 @@
  */
 
 /**
- * Print admin notice regarding having an old version of PHP.
+ * Errors encountered while loading the plugin.
  *
- * @since 0.7
+ * This has to be a global for the same of PHP 5.2.
+ *
+ * @var \WP_Error $_amp_load_errors
  */
-function _amp_print_php_version_admin_notice() {
-	?>
-	<div class="notice notice-error">
-		<p>
-			<?php
-			printf(
-				/* translators: %s: required PHP version */
-				esc_html__( 'The AMP plugin requires PHP %s. Please contact your host to update your PHP version.', 'amp' ),
-				'5.4+'
-			);
-			?>
-		</p>
-	</div>
-	<?php
-}
+global $_amp_load_errors;
+
+$_amp_load_errors = new \WP_Error();
+
 if ( version_compare( phpversion(), '5.4', '<' ) ) {
-	add_action( 'admin_notices', '_amp_print_php_version_admin_notice' );
-	return;
+	$_amp_load_errors->add(
+		'insufficient_php_version',
+		sprintf(
+			/* translators: %s: required PHP version */
+			__( 'The AMP plugin requires PHP %s; please contact your host to update your PHP version.', 'amp' ),
+			'5.4+'
+		)
+	);
 }
 
-/**
- * Print admin notice regarding DOM extension is not installed.
- *
- * @since 1.0
- */
-function _amp_print_php_dom_document_notice() {
-	?>
-	<div class="notice notice-error">
-		<p>
-			<?php
-				printf(
-					/* translators: %s: PHP extension name */
-					esc_html__( 'The AMP plugin requires the %s extension in PHP. Please contact your host to install this extension.', 'amp' ),
-					'DOM'
-				);
-			?>
-		</p>
-	</div>
-	<?php
+// See composer.json for this list.
+$_amp_required_extensions = array(
+	'curl',
+	'dom',
+	'iconv',
+	'intl',
+	'libxml',
+);
+$_amp_missing_extensions  = array();
+foreach ( $_amp_required_extensions as $_amp_required_extension ) {
+	if ( ! extension_loaded( $_amp_required_extension ) ) {
+		$_amp_missing_extensions[] = $_amp_required_extension;
+	}
 }
-if ( ! class_exists( 'DOMDocument' ) ) {
-	add_action( 'admin_notices', '_amp_print_php_dom_document_notice' );
-	return;
-}
+if ( count( $_amp_missing_extensions ) > 0 ) {
+	foreach ( $_amp_missing_extensions as &$_amp_missing_extension ) {
+		$_amp_missing_extension = "<code>$_amp_missing_extension</code>"; // Not using array_map() w/ closure due to PHP 5.2 support.
+	}
 
-/**
- * Print admin notice regarding DOM extension is not installed.
- *
- * @since 1.0.1
- */
-function _amp_print_php_missing_iconv_notice() {
-	?>
-	<div class="notice notice-error">
-		<p>
-			<?php
-				printf(
-					/* translators: %s: PHP extension name */
-					esc_html__( 'The AMP plugin requires the %s extension in PHP. Please contact your host to install this extension.', 'amp' ),
-					'iconv'
-				);
-			?>
-		</p>
-	</div>
-	<?php
+	$_amp_load_errors->add(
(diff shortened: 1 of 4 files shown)


swissspidy:
> For example, if installing the plugin from source and not having done a build, while also the curl extension is not installed, the user is shown the following admin notice:

That paragraph is a bit long. Took me a second to realize these are all different issues that need to be handled individually. Perhaps we could list them on separate lines?

> And now as well there is a warning that is shown when interacting with WP-CLI

Nice!

---

Apart from that: I was wondering if we could somehow use the WP HTTP API with FasterImage, removing the absolute requirement for cURL.

westonruter:
> That paragraph is a bit long. Took me a second to realize these are all different issues that need to be handled individually. Perhaps we could list them on separate lines?

@swissspidy How about like this:

image

image

> Apart from that: I was wondering if we could somehow use the WP HTTP API with FasterImage, removing the absolute requirement for cURL.

I've thought about that, but the problem is `FasterImage` directly calls cURL. We'd need to modify `FasterImage` to allow plugging in a different HTTP API. And I'm pretty sure the WordPress HTTP API doesn't accommodate request parallelization (like Guzzle does).

westonruter:
I suppose each subsequent error line in WP-CLI should have a bullet point.

swissspidy:
Much better already, thanks!

Since the extension list is a bit repetitive, what about something like this:

```php
$extensions = array_map(
   function ( $ext ) {
      return "<code>{ $ext }</code>";
   },
   $extensions
);

sprintf(
  _n(
    'The following PHP extension is missing: %s. Please contact your host ...',
    'The following PHP extensions are missing: %s.Please contact your host ...',
    count( $extensions ),
    'amp'
  ),
  implode( ', ', $extensions )
);
```

westonruter:
Like so:

image

image

swissspidy:
Looks great! 😻

westonruter:
Cherry-picked onto 1.1 branch: 5a9c98eb

amp.php
```diff
@@ -14,102 +14,111 @@
...
+$_amp_load_errors = new \WP_Error();
+
 if ( version_compare( phpversion(), '5.4', '<' ) ) {
-	add_action( 'admin_notices', '_amp_print_php_version_admin_notice' );
-	return;
+	$_amp_load_errors->add(
+		'insufficient_php_version',
+		sprintf(
+			/* translators: %s: required PHP version */
+			__( 'The AMP plugin requires PHP %s; please contact your host to update your PHP version.', 'amp' ),
```
swissspidy: Just noticed this. We should turn this semicolon into a dot again in order to not make existing translations fuzzy.

westonruter: See https://github.com/ampproject/amp-wp/pull/2208

mehrdadabed:
Hi,
As a web-hosting company, we have blocked curl_muli_* function to prevent compromised websites from doing ddos attacks,
As far as I checked, despite the previous versions, the new one only works with curl_muli_* functions, is it possible to consider an option so your plugin also be compatible with "curl_init" funstion as well?
Thanks alot,

westonruter:
@mehrdadabed Hello, I have some questions for you:

>  we have blocked curl_muli_* function to prevent compromised websites from doing ddos attacks

How are you blocking the function? Are you using a forked version of the cURL extension or some other means to prevent access to this function?

I was curious if this function would specifically be a problem for DDoS attacks. I was thinking that the non-blocking HTTP requests that WordPress makes normally for cron could be used in almost the same way as cURL Multi. I created a [script to test](https://gist.github.com/westonruter/ff669549161809c0862caa81cf6a6850), and it seems indeed that cURL multi is able to make requests at 3⨉ the rate of non-blocking cURL.

Even so, is 3⨉ is not 30⨉. So disabling cURL Multi only seems to help marginally, not on the order of magnitudes. So is it still worthwhile?

> As far as I checked, despite the previous versions, the new one only works with curl_muli_* functions, is it possible to consider an option so your plugin also be compatible with "curl_init" funstion as well?

See also [conversation](https://github.com/ampproject/amp-wp/pull/1809/files#r247647230) on #1809. We dropped support for non-multi requests as we assumed that it was available in all PHP 5.4 installations. You are on PHP 5.4+ and yet have this function disabled? Is this a common practice among hosts? It's the first I've seen a report of it.

Are sites you host currently getting PHP server errors due to this function not existing?

At the very least, what we need to do here is check not only if the cURL extension is installed but also we need to check if the `curl_multi_*` functions exist. If both are not true, then the plugin would need to short-circuit and show an error message.

Otherwise, if this is indeed a common scenario, we may need to bring back [`FastImage`](https://github.com/tommoor/fastimage) which was removed in #1809. I'm wary of that, however, because the library is no longer being maintained, and the capabilities are different than `FasterImage` (e.g. SVG support).

/cc @felixarntz

remi-angenieux:
@westonruter you can hopefully disabled functions with `disabled_functions` in php configuration. [Documentation](https://www.php.net/manual/en/ini.core.php#ini.disable-functions)
You can found many articles which blocks this function for security reasons [like this one](https://www.cyberciti.biz/faq/linux-unix-apache-lighttpd-phpini-disable-functions/). So it can be very common to found this function blocked.

I get error 500 with function disabled and this in log : 

> PHP Warning:  curl_multi_exec() has been disabled for security reasons in .../amp/vendor/fasterimage/fasterimage/src/FasterImage/FasterImage.php on line 91

westonruter:
OK, I've opened #2319 to check for the required functions and classes in addition to the required extensions. This will at least prevent a fatal error on your server and will inform the user of what is missing. I believe the ultimate fix for you is going to be for `FasterImage` to support falling-back to only using `curl_init()` when `curl_multi_init()` is not available. I've opened https://github.com/willwashburn/fasterimage/issues/16 toward this end.

mehrdadabed:
Hi and thanks for the answer,
I do understand your concern about your script performance and speed, and certainly curl_multi functions act faster than regular "curl" ones, no doubt,
But we - as a web-hosting provider - concern about, is the number of "simultaneous connections" that could be established by a php script, as you know, it's possible to create thousands of php connections at the same time by curl_multi functions, while it's needed to wait for each connection response when regular "curl" is used, consequently monitoring/controlling outgoing connection is significantly easier in regular curl functions in compare with curl_multi ones, 
So I kindly ask you to provide such option to use curl_init when curl_multi_init is disabled,
Thank you so much,
Abed.

westonruter:
Please follow https://github.com/willwashburn/fasterimage/issues/16 the resolution.

westonruter:
FasterImage has been updated with the fallback support.

Merged update into AMP plugin: https://github.com/ampproject/amp-wp/pull/2422

So this will be part of 1.1.2, perhaps released today.

westonruter:
@mehrdadabed Please test: https://github.com/ampproject/amp-wp/pull/2423

PR #2190, opened 2019-04-24
Remove intl from required extensions

westonruter:
It turns out that `intl` isn't actually a requirement. I tried deploying changes merged in #2183 to Pantheon and upon doing so the plugin stopped initializing due to the `intl` extension not being installed. In fact, the extension is not required as we do a `function_exists` check before using it:

https://github.com/ampproject/amp-wp/blob/98c25457875a30fc18ca2b2a34d74561ab59dc88/includes/class-amp-http.php#L205-L209

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -38,9 +38,11 @@ if ( version_compare( phpversion(), '5.4', '<' ) ) {
 // See composer.json for this list.
 $_amp_required_extensions = array(
 	'curl',
+	'date',
 	'dom',
 	'iconv',
 	'libxml',
+	'spl',
 );
 $_amp_missing_extensions  = array();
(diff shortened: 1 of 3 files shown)


swissspidy:
Can we add a [`suggest`](https://getcomposer.org/doc/04-schema.md#suggest) entry for `intl` in that case?

westonruter:
Is there anything that is _not_ considered in the `composer.json`?! See c57348a.

westonruter:
Cherry-picked onto 1.1 branch: ae6d1bd321adda9936b06b5ed51a0225a5f3adc2 & d79d1f37e878dc97d6c3238169d9d3a3f26c727b

PR #2191, opened 2019-04-25
Slightly improve issue with the text block resized with background

### LINKED ISSUE #2162: Text Block with background displays incorrectly when resized very small
Text Block's resizing doesn't consider very small Text Blocks, especially with background color set.

In the editor, the fixed padding applied to any Text Block with background color causes the area to display larger than it actually is.

Editor: 
Screen Shot 2019-04-22 at 10 22 01 AM

FE:
Screen Shot 2019-04-22 at 10 22 07 AM

kienstra:
* As Miina found, when the Text block is resized small and has a background width, it appears too tall in the block editor.
* This removes the padding-top and bottom, leaving only padding-left and right.
* It doesn't completely fix the difference in the front-end and back-end display, but the text isn't cut off in the front-end.
* Still, the text on the front-end isn't quite the right size (see "After" below).

# Before
before-pr-padding

# After 
after-pr-padding

See #2162

### DIFF
--- a/assets/css/amp-stories.css
+++ b/assets/css/amp-stories.css
@@ -1,7 +1,7 @@
 amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
 
 .has-background {
-	padding: 7px !important;
+	padding: 1px 5px !important;
 	display: flex;
 }
 


kienstra:
**Question About Approach**

Hi @miina,
What do you think about this approach? It's not an ideal solution, but it looks to mainly fix the case you described in #2162.

assets/css/amp-stories.css
```diff
@@ -1,7 +1,7 @@
 amp-story-grid-layer[template="fill"] .wp-block-image { margin: 0; }
 
 .has-background {
-	padding: 7px !important;
+	padding: 1px 5px !important;
```
kienstra: padding-top and bottom are `1px`
padding-left and right are `5px`

miina:
@kienstra I'm still seeing a quite significant difference between the editor and the front-end, it looks like in the editor the displayed size is not respecting the resizing when trying to add a tiny Text block, for example here the area is much larger than in the frontend (this is resized to minimum):
Screen Shot 2019-04-25 at 12 34 19 PM

Frontend:
Screen Shot 2019-04-25 at 12 36 39 PM

This doesn't necessarily seem to do with padding only, perhaps there's something to change with the resizing, maybe either limit how small it can go or alternatively enforce the correct size.

swissspidy:
> This doesn't necessarily seem to do with padding only, perhaps there's something to change with the resizing, maybe either limit how small it can go or alternatively enforce the correct size.

+1

swissspidy:
Closing in favor of #2296 which tries to address this at a broader scope.

PR #2193, opened 2019-04-25
[AMP Stories] Ensure unit when saving animation.

### LINKED ISSUE #2188: [AMP Stories] Animation delay/duration missing units causing no effect
I set the animation delay and duration to a maximum of 5 seconds each:

> image

Nevertheless, the animation starts immediately and finishes.

The HTML generated contains:

`animate-in-delay="5000" animate-in-duration="5000"`

It turns out that this doesn't work. For it to work, there must be units supplied:

`animate-in-delay="5000ms" animate-in-duration="5000ms"`

miina:
Fixes #2188.

Adds `ms` when saving animation.

Also fixes animation for text, author, date blocks.


assets/src/constants.js
```diff
@@ -105,6 +105,9 @@ export const BLOCK_TAG_MAPPING = {
 	'core/embed': 'figure',
 	'core/image': '.wp-block-image',
 	'amp/amp-story-text': 'p,h1,h2',
+	'amp/amp-story-post-author': 'p,h1,h2',
+	'amp/amp-story-post-date': 'p,h1,h2',
+	'amp/amp-story-post-title': 'p,h1,h2',
```
swissspidy: `amp/amp-story-post-title` is `h1` only. It doesn't change tags like `amp/amp-story-text`.

Similarly, `amp/amp-story-post-date` and `amp/amp-story-post-author` just use a `div`.

miina: Oh, that's right, thanks. Fixed within 1ca5831.

PR #2196, opened 2019-04-25
Fix rotation + animation conflict.

### LINKED ISSUE #2186: Rotation conflicts with animation
I added a Text block and rotated it:

image

I then added an animation to it:

image

Even though when I deselect the block I see it rotated in the editor, when I go to the frontend after the animation completes the element is not rotated:

image

### LINKED ISSUE #2218: Animations don't always work
Added different animations to the text in story pages and most of them did not seem to work.

out

miina:
Fixes #2186. Fixes #2218.

Apparently rotation doesn't work together with animation. This PR aims to fix it.


miina:
Looks like `transform` is used for animations, which is also used for rotation and the animation `tranform` overrides the rotation. Here we wouldn't be able to use the CSS variables either since the transform assigned by animation is out of our control.

Unfortunately `amp-story-grid-layer` can't have animation properties, otherwise, we could assign animation to the wrapper grid layer instead and use the block in it for rotation.

Assigning rotation to the wrapping grid layer is possible, however, that wouldn't really be accurate at all since it rotates the grid layer.

One option could be that if animation and rotation are both configured then we'll add an additional wrapper which will take care of the rotation, added this within the PR. Thoughts @swissspidy? Any better solutions you can think of?

Also, for some reason this PR results in having the following AMP Validation error:
Screen Shot 2019-04-25 at 1 45 23 PM

cc @westonruter

westonruter:
I've identified the problem with the animation attributes. It's due to the special `AMP-STORY-GRID-LAYER default` reference point. Working on a fix…

westonruter:
> Also, for some reason this PR results in having the following AMP Validation error:

Fixed in 3d93230.

westonruter:
Note that as the PR stands currently, the element is being positioned outside the viewport:

image

image

miina:
@westonruter Thank you for adding the animation params fix.

On incorrect positioning: looks like we'd need to assign all the other params (positioning and resizing) as well to the outer wrapper for this to be positioned properly.

We can do that, however, I'm wondering if there is a better solution for mixing animation + rotation. @swissspidy Let me know if you have a "cleaner" solution in mind.

assets/src/helpers.js
```diff
@@ -396,14 +396,32 @@ export const filterBlockAttributes = ( blockAttributes, blockType, innerHTML ) =
...
  *
  * @return {Object} The element.
  */
-export const wrapBlocksInGridLayer = ( element, blockType ) => {
+export const wrapBlocksInGridLayer = ( element, blockType, attributes ) => {
 	if ( ! element || ! ALLOWED_MOVABLE_BLOCKS.includes( blockType.name ) ) {
 		return element;
 	}
 
+	// If both rotation and animation are present then we'll need to use a separate wrapper for rotation.
```
swissspidy: Can't we just always use a separate wrapper? Would make maintenance a bit easier I think.

miina: Yes, we could, I was trying to avoid adding additional wrappers to the core blocks if necessary since animation + rotation isn't necessarily a super common use case. However, if you think it's better to always add a wrapper, I can change that.

miina:
@swissspidy I've made the suggested change to always add the additional wrapper to the block within 2b467c7. Note that this breaks all the existing templates and stories.

I'm still wondering if it would be better to only use the additional wrapper in case of rotation when it's actually needed. Not sure. Do you think it's better to always have it, for consistency?

Let me know your thoughts, if it seems to be better to always add the wrapper then I'll also update the templates.

tests/test-tag-and-attribute-sanitizer.php
```diff
@@ -302,7 +302,10 @@ function () {
 									</amp-img>
 								</amp-story-grid-layer>
 								<amp-story-grid-layer template="vertical">
-									<h1>The End</h1>
+									<h1 animate-in="drop" animate-in-delay="500ms" animate-in-duration="600ms">The End</h1>
+									<div>
```
swissspidy: Wouldn't that have the `amp-story-block-wrapper` class name now?

miina: I haven't updated the tests yet, also, templates need update.

miina:
This should be ready for testing / reviewing again now.

Note that this PR breaks the previous stories. Also, the animation is rotated as well, maybe it looks a little bit odd but not sure if there's an easy way how to have the animation "straight" and then add rotation without it being even more odd.

swissspidy:
> I'm still wondering if it would be better to only use the additional wrapper in case of rotation when it's actually needed. Not sure. Do you think it's better to always have it, for consistency?
> 
> Let me know your thoughts, if it seems to be better to always add the wrapper then I'll also update the templates.

I'm leaning towards yes, for consistency.

@westonruter What's your thought on that?

westonruter:
Just to be clear I understand what we're talking about, I created a story page with 4 Text blocks:

1. Not rotated, no animation.
2. Not rotated, animated.
3. Rotated, not animated.
4. Rotated, animated.

image

The markup is:

```html
<!-- wp:amp/amp-story-page {"autoAdvanceAfterDuration":1} -->
<amp-story-page style="background-color:#ffffff" id="619fcbd2-26ba-4b2d-a732-a9960d37930c" class="wp-block-amp-amp-story-page">
	<amp-story-grid-layer template="fill"></amp-story-grid-layer>
	<!-- wp:amp/amp-story-text {"placeholder":"Write text…","tagName":"h1","autoFontSize":44,"positionTop":10} -->
	<amp-story-grid-layer template="vertical">
		<div class="amp-story-block-wrapper" style="position:absolute;top:10%;left:5%;width:76%;height:9%">
			<h1 style="font-size:44px;display:flex" class="" id="82f92984-ae89-4f7c-a8bb-395e617a6719">
				<amp-fit-text layout="flex-item" class="amp-text-content">No rotate</amp-fit-text>
			</h1>
		</div>
	</amp-story-grid-layer>

	<!-- wp:amp/amp-story-text {"tagName":"h1","autoFontSize":44,"positionTop":22,"positionLeft":3,"rotationAngle":16} -->
	<amp-story-grid-layer template="vertical">
		<div class="amp-story-block-wrapper" style="position:absolute;top:22%;left:3%;transform:rotate(16deg);width:76%;height:9%">
			<h1 style="font-size:44px;display:flex" class="" id="3544e0de-978e-420f-9269-4a9082b04496">
				<amp-fit-text layout="flex-item" class="amp-text-content">Rotated</amp-fit-text>
			</h1>
		</div>
	</amp-story-grid-layer>

	<!-- wp:amp/amp-story-text {"tagName":"h1","autoFontSize":42,"positionTop":36,"positionLeft":7} -->
	<amp-story-grid-layer template="vertical">
		<div class="amp-story-block-wrapper" style="position:absolute;top:36%;left:7%;width:76%;height:9%">
			<h1 style="font-size:42px;display:flex" class="" id="aa08b40d-b3a9-4538-a422-dde6950f0bec" animate-in="rotate-in-left" animate-in-duration="700ms">
				<amp-fit-text layout="flex-item" class="amp-text-content">Rotate in left</amp-fit-text>
			</h1>
		</div>
	</amp-story-grid-layer>

	<!-- wp:amp/amp-story-text {"tagName":"h1","positionTop":72,"positionLeft":12,"height":107,"rotationAngle":38} -->
	<amp-story-grid-layer template="vertical">
		<div class="amp-story-block-wrapper" style="position:absolute;top:72%;left:12%;transform:rotate(38deg);width:76%;height:19%">
			<h1 style="font-size:45px;display:flex" class="" id="149665f3-bb4a-45cc-b635-0dc67919e2f9" animate-in="rotate-in-left" animate-in-duration="700ms">
				<amp-fit-text layout="flex-item" class="amp-text-content">Rotated rotate in left</amp-fit-text>
			</h1>
		</div>
	</amp-story-grid-layer>
	<!-- /wp:amp/amp-story-text --></amp-story-page>
<!-- /wp:amp/amp-story-page -->
```

Positioning/rotation of the block is set on the wrapper, and the animation (`animate-in`) is only defined on the actual block root element itself.

So every block gets the `amp-story-block-wrapper`. I guess I don't feel strongly either way. If always added, there is the consistency that the position/rotation is always defined there. So that's good. What's bad is extra elements that aren't necessary. 🤷‍♂

miina:
Let's go with consistency then as Pascal is recommending. Should be ready for reviewing as it is.

PR #2197, opened 2019-04-25
Adjust page alignment in carousel

### LINKED ISSUE #2026: Current page not aligned in the middle
Looks like the offset calculation is still wrong because of some CSS width changes I presume:

Screenshot 2019-03-26 at 16 45 17

miina:
Fixes #2026.

Looks like page margin wasn't considered accurately (probably it changed at some point) and this caused the page not to be in the middle and also the location shifting with each added page.


miina:
@swissspidy Could you check if it shows as expected in your screen as well?

swissspidy:
Thanks, looks much better already.

I added 11 pages for testing and then it slowly revelead that there is still a tiny offset. You can see that the page should be just a few pixels more on the left.

Screenshot 2019-04-29 at 11 39 49

miina:
@swissspidy From what I can see then the difference is not coming when adding new pages but is actually there even with one page. It might come from the additional border of the page, not sure, considering that does seem to improve the alignment. Let me know if it seems to be well aligned now.

PR #2206, opened 2019-04-26
Change <Button> to <a>, to work around a console warning

### LINKED ISSUE #2165: Warning messages in console for templates
The templates inserter includes buttons within buttons at this moment which is forbidden.

Ideally, the `BlockEdit` would not include any buttons in the preview since they're not functional anyway. Alternatively, we should replace the `button` being the wrapper for preview.

Screen Shot 2019-04-22 at 3 07 27 PM

kienstra:
As @miina mentioned in #2165, there was a console warning: `validateDOMNesting(...): <button> cannot appear as a descendent of <button>`

So this changes the wrapping `<Button>` to `<a>`.

This isn't ideal, but it prevents the console warning.

Maybe there's a good way to selectively disable the `<Button>` components in the [BlockPreview](https://github.com/ampproject/amp-wp/blob/fc98d441277b1c83070921f8b607e53683cf0559/assets/src/components/template-inserter/index.js#L115), but I could find one.

The `<BlockPreview>` [is wrapped in](https://github.com/ampproject/amp-wp/blob/7264d585043fb863e5f144f07daff66772836230/assets/src/components/block-preview.js#L18) `<Disabled>`.

And the `<Button>` component [has a disabled property](https://github.com/WordPress/gutenberg/blob/87d7fe4be406aedb3aefa7d961ac3781ab600654/packages/components/src/button/index.js#L42) that can change the `<button>` to a `<a>`. But that property doesn't look to be `true`.

Fixes #2165

### DIFF
--- a/assets/src/components/template-inserter/index.js
+++ b/assets/src/components/template-inserter/index.js
@@ -2,7 +2,7 @@
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Dropdown, IconButton, Button } from '@wordpress/components';
+import { Dropdown, IconButton } from '@wordpress/components';
 import { Component } from '@wordpress/element';
 import { withSelect, withDispatch } from '@wordpress/data';
 import { createBlock } from '@wordpress/blocks';
--- a/assets/src/components/template-inserter/index.js
+++ b/assets/src/components/template-inserter/index.js
@@ -1,3 +1,9 @@
+
+/**
+ * External dependencies
+ */
+import { includes } from 'lodash';
+
 /**
  * WordPress dependencies
  */
@@ -7,6 +13,7 @@ import { Component } from '@wordpress/element';
 import { withSelect, withDispatch } from '@wordpress/data';
 import { createBlock } from '@wordpress/blocks';
 import { compose } from '@wordpress/compose';
+import { ENTER, SPACE } from '@wordpress/keycodes';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


kienstra:
**Request For Review**

Hi @westonruter,
Could you please review this? It's not a great solution, but it prevents the console warning.

assets/src/components/template-inserter/index.js
```diff
@@ -105,18 +105,23 @@ class TemplateInserter extends Component {
...
-										<Button
+										<a // eslint-disable-line jsx-a11y/anchor-is-valid, see https://github.com/ampproject/amp-wp/issues/2165
 											key={ `template-preview-${ item.id }` }
+											role="button"
+											tabIndex="0"
 											onClick={ () => {
 												onSelect( item );
 											} }
+											onKeyDown={ () => {
+												onSelect( item );
```
swissspidy: Haven't tested, but this should probably check for what kind of key was pressed. I guess `ENTER` and `SPACE` would be the only ones. `@wordpress/keycodes` can help with that.

swissspidy:
It's what we had before, and yes, it's not the best solution. But it really seems that it's the most feasible one.

assets/src/components/template-inserter/index.js
kienstra: Ah, good point. 

https://github.com/ampproject/amp-wp/pull/2206/commits/45c7246e79303edf90060be33bf658496a7cdeee checks that it's an `ENTER` or `SPACE` key.

PR #2208, opened 2019-04-29
Revert PHP version error message translation string from #2183

westonruter:
See https://github.com/ampproject/amp-wp/pull/2183/files#r278899906:

> We should turn this semicolon into a dot again in order to not make existing translations fuzzy.


westonruter:
Cherry-picked onto 1.1 branch: f8814cc6d577ae0930f61597712af5b080780d31

PR #2227, opened 2019-05-01
Add 'View AMP' link to admin bar in Reader mode; add admin bar item for non-admin users

### LINKED ISSUE #2203: Add "View AMP" link to admin bar on non-AMP page in Reader mode
A very common support request is confusion about why adding `/amp/` to the URL leads to a 404 instead. This is normally due to trying to access the AMP version of a Page or some other hierarchical post type, for which `?amp` is used for the AMP URL. (See also #2062 to automatically redirect from `/amp/` to `?amp` when `is_404()`.) When users do not have the AMP Validator extension installed, they will not see the AMP link in the browser extension button. If users are unfamiliar with the `amphtml` link in the head, they will have no idea how to find the AMP version or even if there is an AMP version available.

So we should consider adding the AMP link to the admin bar for AMP-enabled posts, even in reader mode:

image

Aside: We should consider eliminating the `/amp/` in favor of `?amp` everywhere. The endpoint slug is unnecessary for SEO, and it improve consistency when switching between the paired modes (transitional and reader). Additionally, there will no longer be a need to flush rewrite rules. See #2204.

westonruter:
Fixes #2203:

> A very common support request is confusion about why adding `/amp/` to the URL leads to a 404 instead. This is normally due to trying to access the AMP version of a Page or some other hierarchical post type, for which `?amp` is used for the AMP URL. […] When users do not have the AMP Validator extension installed, they will not see the AMP link in the browser extension button. If users are unfamiliar with the `amphtml` link in the head, they will have no idea how to find the AMP version or even if there is an AMP version available.
> 
> So we should consider adding the AMP link to the admin bar for AMP-enabled posts, even in reader mode:
> 
> image

The “View AMP” link on the top-level admin bar item is shortened to just “AMP”:

image

When looking at the AMP version, the order of the submenu items switch so that the first item is Validate, which corresponds with the parent icon change from 🔗 to ✅/ ⚠️:

image

image

The order changes because the first submenu item needs to be the same as the parent menu item for the sake of touch screens.

When there are rejected validation errors, the parent item is also still “AMP” but there is no child to view (since re-validation is the only option available):

image

This is similar to Native mode which used to say “Validate AMP” but also now just says “AMP”:

image

----

This PR also ensures that when a user is non-privileged (e.g. a subscriber role) that they will continue to get the AMP link added, but without the re-validation logic:

image

image

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -1067,3 +1067,60 @@ function amp_wp_kses_mustache( $markup ) {
 	$amp_mustache_allowed_html_tags = array( 'strong', 'b', 'em', 'i', 'u', 's', 'small', 'mark', 'del', 'ins', 'sup', 'sub' );
 	return wp_kses( $markup, array_fill_keys( $amp_mustache_allowed_html_tags, array() ) );
 }
+
+/**
+ * Add "View AMP" admin bar item for Transitional/Reader mode.
+ *
+ * Note that in Transitional mode, the admin bar item will be further amended by `AMP_Validation_Manager::add_admin_bar_menu_items()`.
+ *
+ * @see \AMP_Validation_Manager::add_admin_bar_menu_items()
+ *
+ * @param WP_Admin_Bar $wp_admin_bar Admin bar.
+ */
+function amp_add_admin_bar_view_link( $wp_admin_bar ) {
+	if ( is_admin() || amp_is_canonical() ) {
+		return;
+	}
+
+	if ( current_theme_supports( 'amp' ) ) {
+		$available = AMP_Theme_Support::get_template_availability()['supported'];
+	} elseif ( is_singular() ) {
+		$post      = get_queried_object();
+		$available = ( $post instanceof WP_Post ) && post_supports_amp( $post );
+	} else {
+		$available = false;
+	}
+	if ( ! $available ) {
+		// @todo Add note that AMP is not available?
+		return;
+	}
+
+	// Show nothing if there are rejected validation errors for this URL.
+	if ( ! is_amp_endpoint() && count( AMP_Validated_URL_Post_Type::get_invalid_url_validation_errors( amp_get_current_url(), array( 'ignore_accepted' => true ) ) ) > 0 ) {
+		return;
+	}
+
+	if ( is_amp_endpoint() ) {
+		$href = amp_remove_endpoint( amp_get_current_url() );
+	} else {
+		if ( is_singular() ) {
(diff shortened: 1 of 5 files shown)


includes/amp-helper-functions.php
```diff
@@ -1067,3 +1067,60 @@ function amp_wp_kses_mustache( $markup ) {
 	$amp_mustache_allowed_html_tags = array( 'strong', 'b', 'em', 'i', 'u', 's', 'small', 'mark', 'del', 'ins', 'sup', 'sub' );
 	return wp_kses( $markup, array_fill_keys( $amp_mustache_allowed_html_tags, array() ) );
 }
+
+/**
+ * Add "View AMP" admin bar item for Transitional/Reader mode.
+ *
+ * Note that in Transitional mode, the admin bar item will be further amended by `AMP_Validation_Manager::add_admin_bar_menu_items()`.
```
westonruter: Actually, `AMP_Validation_Manager::add_admin_bar_menu_items()` is responsible for both Transitional _and_ Native modes.

includes/amp-helper-functions.php
```diff
@@ -1067,3 +1067,61 @@ function amp_wp_kses_mustache( $markup ) {
...
+	if ( current_theme_supports( 'amp' ) ) {
+		$available = AMP_Theme_Support::get_template_availability()['supported'];
+	} elseif ( is_singular() ) {
+		$post      = get_queried_object();
+		$available = ( $post instanceof WP_Post ) && post_supports_amp( $post );
+	} else {
+		$available = false;
+	}
+	if ( ! $available ) {
+		// @todo Add note that AMP is not available?
```
amedina: +1 on bringing this to user's awareness (continuously)

tests/test-amp-helper-functions.php
```diff
@@ -972,6 +972,60 @@ function( $meta ) use ( $self ) {
...
+		$this->go_to( get_permalink( $post_id ) );
+		global $wp_query; // Must be here after the go_to() call.
+
+		// Check that canonical adds no link.
+		add_theme_support( 'amp' );
+		$admin_bar = new WP_Admin_Bar();
+		amp_add_admin_bar_view_link( $admin_bar );
+		$this->assertNull( $admin_bar->get_node( 'amp' ) );
+
+		// Check that paired mode does add link.
```
amedina: Nit. Change the comments when possible to reflect rebranded modes. Same above.

westonruter:
Cherry-picked onto 1.1 branch: 57e54a050a5364cb64a3912ee8b862fd497c820f and 32f726233b89eb885cbd5acfd1eb035e79f132b5

PR #2238, opened 2019-05-02
Add default value for focalPoint

### LINKED ISSUE #2112: NaN warning for focal point picker
I am now getting "Received NaN for the value attribute. If this is expected, cast the value to a string." warnings in the console when adding a background image.

Originally mentioned in https://github.com/ampproject/amp-wp/pull/2104#issuecomment-481618142, but apparently not caused by that PR.

### LINKED ISSUE #2213: Initial focal point of 50%/50% not applied to editor view
I inserted an image as background media and even though the focal point said it had a 50%/50% focal point, the image appeared off center:

image

The frontend looked correct, however:

image

To get the focal point to look correct in the editor, I had to touch the control to apparently give it a non-default value.

swissspidy:
Fixes #2213.
Also fixes #2112 I believe.

### DIFF
--- a/assets/src/blocks/amp-story-page/edit.js
+++ b/assets/src/blocks/amp-story-page/edit.js
@@ -148,7 +148,7 @@ class EditPage extends Component {
 			mediaId,
 			mediaType,
 			mediaUrl,
-			focalPoint,
+			focalPoint = { x: .5, y: .5 },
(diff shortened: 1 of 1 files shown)


westonruter:
This does fix #2213 but it does not appear to fix #2112. I'm seeing:

image

PR #2244, opened 2019-05-02
Draggable Block Navigation

### LINKED ISSUE #1974: [AMP Stories] Draggable Block Navigation
Add drag/drop reordering of elements in table of contents (contribute via Gutenberg).

This would allow us to get rid of the up/down errors in the BlockMover component. Since the drag handle in the BlockMover looks a bit odd on its own, and because it's not the best UX, I suggest instead exploring making the whole block element draggable and not just the handle.

Previously: #1897

swissspidy:
Fixes #1974.


swissspidy:
There's one thing left here:

The position of the element's clone is completely off, probably because the block navigation itself is already positioned absolutely on the screen, which throws off the calculation.

Screenshot:

assets/src/blocks/amp-story-page/edit.css
```diff
@@ -153,10 +153,6 @@ div[data-type="amp/amp-story-page"] .wp-block-image {
 	display: none;
 }
 
-#amp-story-editor .components-draggable__clone {
-	position: absolute;
```
miina: Removing this might influence other items being dragged-dropped as well, reordering and general dragging and dropping -- are these still working as expected?

assets/src/components/block-navigation/edit.css
```diff
@@ -190,3 +190,36 @@
...
 	}
 }
+
+/**
+ * Drag & Drop support
+ */
+
+.block-editor-block-navigation__list .block-editor-block-navigation__item {
+	position: relative;
+}
```
miina: Wondering if we should also change the pointer to match with dragging the blocks for example. Thoughts?

assets/src/blocks/amp-story-page/edit.css
miina: Actually, I just tested and it looks like dragging elements is a bit off now, otherwise works as expected..

swissspidy: You mean like regular elements within a page?

I thought removing that selector would cause major problems too, but I couldn't find anything. The default is `position: fixed`, which should be more precise anyway.

assets/src/components/block-navigation/edit.css
swissspidy: True. Good point!

assets/src/blocks/amp-story-page/edit.css
miina: That's what I see when dragging a Text block -- the draggable clone is quite far from the dragging icon.
Screen Shot 2019-05-03 at 5 06 17 PM

Are you seeing something else?

swissspidy: Ah, saw the issue now and fixed it by making that selector more specific to the block list.

assets/src/components/block-navigation/edit.css
swissspidy: Fixed now.

PR #2245, opened 2019-05-02
Address AMP Stories embed and Latest Stories styling issues

kienstra:
* Addresses styling issues in the AMP Stories discoverability blocks
* The inline `height` on the `.latest-stories-carousel` seemed to make the element too tall

### DIFF
--- a/assets/css/amp-story-card.css
+++ b/assets/css/amp-story-card.css
@@ -29,6 +29,11 @@ div.amp-block-latest-stories > ul.latest-stories-carousel {
 	overflow-x: hidden;
 }
 
+.amp-story-embed .latest_stories__link:hover {
+	box-shadow: none;
+	-webkit-box-shadow: none;
+}
+
 /* Applies the darker gradient background, to allow the white text to display even against a white featured image. */
 .latest_stories__link:after {
 	display: block;
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -1373,11 +1373,10 @@ class AMP_Story_Post_Type {
 	}
 
 	/**
-	 * Changes attributes of the AMP Story embed <iframe>.
+	 * Changes the height of the AMP Story embed <iframe>.
 	 *
-	 * The embed typically appears in the editor in an <iframe>, and without an <iframe> on the front-end.
+	 * In the block editor, this embed typically appears in an <iframe>, though on the front-end it's not in an <iframe>.
 	 * The height of the <iframe> isn't enough to display the full story, so this increases it.
-	 * Also, this removes attributes that AMP does not allow for the <amp-iframe>, like marginwidth.
 	 *
 	 * @param string  $output The embed output.
 	 * @param WP_Post $post The post for the embed.
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -195,6 +195,9 @@ class AMP_Story_Post_Type {
 		// In the block editor, remove the title from above the AMP Stories embed.
 		add_filter( 'embed_html', array( __CLASS__, 'remove_title_from_embed' ), 10, 2 );
 
+		// Change some attributes for the AMP story embed.
+		add_filter( 'embed_html', array( __CLASS__, 'change_embed_iframe_attributes' ), 10, 2 );
+
 		// Override the render_callback for AMP story embeds.
 		add_filter( 'pre_render_block', array( __CLASS__, 'override_story_embed_callback' ), 10, 2 );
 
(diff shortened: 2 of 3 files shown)


kienstra:
# Before 

before-not-enough-height


# After 

after-enough-height

kienstra:
**Request For Review**

Hi @swissspidy,
Could you please review this?

It fixes styling issues for the Embed block with an AMP Story, and the Latest Stories block.

* In the Latest Stories block, this removes the inline `height` on the `.latest-stories-carousel`, which seemed to make the element too tall
* In the Embed block that has an embedded AMP Story, this increases the height of the `<iframe>` in the block editor.

Thanks, and hope you had a great weekend!

assets/css/amp-story-card.css
```diff
@@ -29,6 +29,11 @@ div.amp-block-latest-stories > ul.latest-stories-carousel {
 	overflow-x: hidden;
 }
 
+.amp-story-embed .latest_stories__link:hover {
+	box-shadow: none;
+	-webkit-box-shadow: none;
```
swissspidy: Note to self: this is something for autoprefixer in the future

includes/class-amp-story-post-type.php
```diff
@@ -1368,4 +1371,48 @@ public static function remove_title_from_embed( $output, $post ) {
...
+		$original_iframe = $matches[0];
+		$new_iframe      = preg_replace(
+			'/(?<=\sheight=")\w+(?=")/',
+			strval( ( self::STORY_LARGE_IMAGE_DIMENSION / 2 ) + 4 ),
+			$original_iframe
+		);
+
+		// Remove attributes that AMP doesn't allow for the <amp-iframe>.
+		if ( is_amp_endpoint() ) {
+			$new_iframe = preg_replace(
```
miina: @kienstra In which case would this replacement be used? If `amp-iframe` isn't used in the front-end and in the editor the `iframe` element is used (if I understood it correctly).

includes/class-amp-story-post-type.php
```diff
@@ -1368,4 +1371,48 @@ public static function remove_title_from_embed( $output, $post ) {
...
+
+		preg_match( '/<iframe sandbox="allow-scripts" [^>]*>/', $output, $matches );
+		if ( ! $matches ) {
+			return $output;
+		}
+
+		$original_iframe = $matches[0];
+		$new_iframe      = preg_replace(
+			'/(?<=\sheight=")\w+(?=")/',
+			strval( ( self::STORY_LARGE_IMAGE_DIMENSION / 2 ) + 4 ),
```
swissspidy: What is this magic number `4`?

includes/class-amp-story-post-type.php
```diff
@@ -1368,4 +1371,48 @@ public static function remove_title_from_embed( $output, $post ) {
...
+		);
+
+		// Remove attributes that AMP doesn't allow for the <amp-iframe>.
+		if ( is_amp_endpoint() ) {
+			$new_iframe = preg_replace(
+				array(
+					'/\ssecurity="[\w]+"/',
+					'/\smarginwidth="[\w]+"/',
+					'/\smarginheight="[\w]+"/',
+				),
```
westonruter: This array can be replaced with a single string:

`/\s(security|marginwidth|marginheight)="[^"]*"/`

includes/class-amp-story-post-type.php
kienstra: @miina, good question. There were validation errors above the block for these before, but I can't reproduce them now.

includes/class-amp-story-post-type.php
kienstra: Hi @swissspidy, good to see you a few days ago.

The `<iframe>` apparently needs a little more height to display the entire embed `<a>`. 

Without the extra `4px`, here's what it looks like:

without-extra-4

difference-in-iframe-height

includes/class-amp-story-post-type.php
kienstra: Thanks, good idea. At the moment, it looks like the logic to remove these attributes will be removed.

includes/class-amp-story-post-type.php
kienstra: https://github.com/ampproject/amp-wp/pull/2245/commits/144e09b021911ca6f71e5690497e06310d4b071b removes the logic that removed these attributes.

kienstra:
**Ready For Another Round Of Review**

Thanks for the feedback, and sorry for the delay. With the [commit](https://github.com/ampproject/amp-wp/pull/2245/commits/144e09b021911ca6f71e5690497e06310d4b071b) above, this is ready for another round of review.

kienstra:
**Testing Steps**

1. [Go to](https://story-test-wordpress-amp.pantheonsite.io/wp-admin/edit.php?post_type=amp_story&paged=2) `/wp-admin` > AMP Stories
2. Click "View" for a published story (most stories aren't published):
front-end-url

3. Copy the URL for that:
copy-url-amp-story

4. Create a new post (not an AMP story)
5. Add an Embed block
6. Paste in the URL of the AMP Story
7. Expected: the embedded AMP Story has a full height, and shows "rounded" corners at the bottom
post-here
8. Repeat steps 5-7 for the "WordPress" block, and the results should be the same

includes/class-amp-story-post-type.php
swissspidy: Gotcha, thanks. But why exactly `4`? Sounds a bit dangerous to rely on such a magic number.

kienstra: Yeah, this calculation definitely isn't ideal. It'd be nice to avoid this completely.

But it looks like `4px` is the increase in `<iframe>` height needed to ensure the image isn't cut off, and the `border-radius` appears:

# Adding 3px

adding-3px

# Adding 4px

adding-4px

swissspidy:
See [comment above for testing steps](https://github.com/ampproject/amp-wp/pull/2245#issuecomment-491553910).

csossi:
Verified in QA

Reviewed both Embed and WordPress blocks
image

PR #2246, opened 2019-05-03
Add transform for Text block to prevent error when pasting

### LINKED ISSUE #2212: Pasting text into empty Text block raises error
After inserting a Text block, if the first thing I populate the block with comes from the clipboard, an error is then raised:

> TypeError: Cannot read property 'attributes' of undefined

image

Fortunately the editor kept working as expected after I unpaused the breakpoint.

westonruter:
When pasting into the Text block, an error is raised due to there not being any `transforms` defined on the block, and the Custom HTML block also being unregistered. This PR adds a transformation for the Text block to ensure that this doesn't happen.

This is a drafted PR because my `transform` function is not actually working, and I'm out of my depth with Gutenberg interns to figure out why.

Fixes #2212.


miina:
@swissspidy I'll continue this PR (unless you already started, let me know if that's the case).

assets/src/blocks/amp-story-text/index.js
```diff
@@ -109,4 +110,31 @@ export const settings = {
...
+				type: 'raw',
+				priority: 20,
+				selector: 'p,h1,h2',
+				transform: ( node ) => {
+					const innerHTML = node.outerHTML;
+					const blockAttributes = getBlockAttributes( name, innerHTML );
+
+					/*
+					 * When there is nothing that matches the content selector (.amp-text-content), the pasted content
+					 * lacks lacks the amp-fit-text wrapper and thus ampFitText is false.
```
swissspidy: `lacks lacks`

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -123,7 +123,9 @@ class TextBlockEdit extends Component {
 					// Ensure line breaks are normalised to HTML.
 					value={ content }
 					onChange={ ( nextContent ) => setAttributes( { content: nextContent } ) }
+					// The 2 following lines are necessary for pasting to work.
 					onReplace={ this.onReplace }
+					onSplit={ () => {} }
```
miina: Looks like for replacing to work properly `onSplit` needs to be defined, too, we're not actually using it, so it should work with an empty function.

swissspidy:
@miina Looks like the merge happened with your Gmail address. Would you mind signing the CLA with that email address as well to prevent any future googlebot issues? 🙂

westonruter:
The gmail address has signed the CLA. The problem I believe is the merge commit and that the CLA bot doesn't like them generally.

swissspidy:
Oh. Weird. Ugh. GitHub probably uses noreply... email addresses in the background for merges

PR #2249, opened 2019-05-03
Force pages to have full height when reordering

### LINKED ISSUE #2229: Story pages not showing content on reordering screen
When reordering the pages in my story, can't see the content of the pages in the reordering screen: 
Screen Shot 2019-05-01 at 5 06 27 PM

swissspidy:
Fixes #2229.

Screenshot:

Screenshot 2019-05-03 at 16 09 41

### DIFF
--- a/assets/src/components/reorderer/edit.css
+++ b/assets/src/components/reorderer/edit.css
@@ -13,9 +13,8 @@
 }
 
 .amp-story-reorderer .amp-story-reorderer-item .amp-story-reorderer-item-page {
-	max-width: 169px;
+	max-width: 166px;
 	height: 277px;
-	margin: 0 10px;
 }
 
 .amp-story-reorderer .block-editor-block-preview .block-editor-block-preview__content .components-disabled {
@@ -100,7 +99,7 @@
 	border-radius: 5px;
 	background: #fff;
 	width: 100%;
-	height: 277px;
+	height: 100%;
 	cursor: grab;
 	overflow: hidden;
 }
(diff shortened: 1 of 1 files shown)


assets/src/components/reorderer/edit.css
```diff
@@ -34,6 +34,7 @@
 .amp-story-reorderer .components-disabled div:first-of-type {
 	padding: 0;
 	width: 100%;
+	height: 100%;
```
miina: The ratio of the page preview when reordering seems to differ from the editor (the inner width/height ratio, not considering the borders). It's around 0.52 for reordering and around 0.59 for the editor. Do you think it would make sense to change that, too?

swissspidy: Totally.

Looks like I've mentioned that before in https://github.com/ampproject/amp-wp/issues/1967#issuecomment-473848183, but seems that it got lost under the pile of other comments and issues.

swissspidy: Updated it to 166x277. 166 / 277 = 0.5992779783

miina: The 166/277 seems to be the outer width and height ratio, however, in case of the editor 328/553 is the inner width and height ratio -- considering just the room available for the blocks and not the border surrounding the page. Probably the inner ratio should match instead?

Also, the `--preview-scale: 0.42;` might need adjusting with the new values.

miina: Looks like scale might be 0.47 now (155/338).

swissspidy: Yup, looks like it!

PR #2250, opened 2019-05-03
Fix not resized image appearing very small in the frontend

### LINKED ISSUE #2234: Image displayed incorrect size in the FE
Steps to reproduce:
- Add an image block with a tall image (not background media).
- Save and see the FE result.

Editor:
Screen Shot 2019-05-02 at 12 51 24 PM

FE:
Screen Shot 2019-05-02 at 1 13 19 PM

miina:
Fixes #2234.

Looks like the issue happens only when an image hasn't been resized after inserting, meaning that when width and height are missing.
This PR sets the width and height upon adding an image based on the original size.


assets/css/amp-stories-frontend.css
```diff
@@ -83,11 +84,12 @@ amp-story-grid-layer .wp-block-quote {
...
-.wp-block-image.is-resized amp-img {
+
+.wp-block-image amp-img {
     width: 100% !important;
-    height: auto !important;
+    height: 100% !important;
 }
-.wp-block-image.is-resized amp-img > img {
+.wp-block-image amp-img > img {
 	object-fit: contain;
```
miina: This enforces the correct size, however, it doesn't work the best with different ratios. Wondering if `cover` would fit better here. This, however, might cut some edges off the images. Thoughts?

swissspidy: @westonruter Maybe you have an idea here?

swissspidy:
Not sure about the `object-fit: contain;` part. But the default attribute part works well. Much better experience.

assets/css/amp-stories-frontend.css
westonruter: Let's try `cover` and see if there are any regressions.

miina: Sounds good, made the change.

PR #2240, opened 2019-05-02
Improve video poster image handling

westonruter:
- Ensure poster image is centered in editor as it will be on frontend.
- Make sure video is displayed instead of poster image.
- Add warning when no poster image is selected.
- Show preview of poster similar to how is done for the featured image.

Fixes #2214.

# No Poster Image Selected

> image

# Poster Image Selected

> image

# Video Plays in Editor

> image

PR #2255, opened 2019-05-03
Prevent video block from causing AMP validation errors in stories

### LINKED ISSUE #2109: Video blocks can cause validation errors in AMP Stories
When adding a video block to an AMP Story page and hitting Save an AMP validation error appears:

> Screen Shot 2019-04-11 at 09 52 25

This is due to the video not having:

* `autoplay`
* `poster`

When supplying these the validation error goes away:

> image

When a video is shown in an AMP Stories context, the `autoplay` toggle should be enabled by default and either disabled or hidden entirely.

Also, the `playsinline` toggle needs to be removed in the stories context since it is not allowed:

> image

Additionally, since we don't currently automatically extract a poster image from a video (see #903) there should be a warning displayed with either the poster field to indicate it is required.

westonruter:
Fixes #2109.

- [x] Set attribute defaults for `autoplay:true`. Note the toggle actually has no effect because the sanitizer will automatically supply the `autoplay` attribute since [it is `mandatory`](https://github.com/ampproject/amphtml/blob/b814e5d74cadf554c5caa1233d71e8e840788ff5/extensions/amp-video/validator-amp-video.protoascii#L142-L146) and thus it gets automatically supplied to ensure AMP validation.
- [x] Set attribute defaults for `controls:false`. This seems relevant since the video is not `interactive`, so the controls can't be manipulated anyway.
- [x] Make `loop:true` attribute default? The background video loops by default.
- [ ] ~Hide/remove the "Autoplay" toggle (since it must always be `autoplay:true`).~ Depends on https://github.com/WordPress/gutenberg/issues/6023. See https://github.com/ampproject/amp-wp/issues/2283.
- [ ] ~Hide/remove the "Play inline" toggle (and make sure it defaults to `playsinline:false`).~ Depends on https://github.com/WordPress/gutenberg/issues/6023. See https://github.com/ampproject/amp-wp/issues/2283.
- [x] Automatically populate poster image from selected video attachment's featured image. See e3d92a5 and https://github.com/ampproject/amp-wp/pull/2240#issuecomment-489125236.
- [x] Add an error message when no poster image is selected. See #2240. (Note that if a user set the `poster` and then removed it, Gutenberg leaves behind an empty `poster` attribute that technically prevents a validation error from happening; this is because of a [loose definition of `poster`](https://github.com/ampproject/amphtml/blob/b814e5d74cadf554c5caa1233d71e8e840788ff5/extensions/amp-video/validator-amp-video.protoascii#L147-L150) on the `amp-story >> amp-video` spec where the attribute is merely `mandatory`, and it leaves its type undefined; but if this in the future changes to be defined as `value_url`, then this will become a hard AMP validation error, since the plugin's sanitizer wouldn't be able to automatically supply an empty `poster` attribute to circumvent the validation error.)
image


westonruter:
> * Automatically populate poster image from selected video attachment's featured image. See [e3d92a5](https://github.com/ampproject/amp-wp/commit/e3d92a5828d3060a342796f3eb7681d61fdbe552) and [#2240 (comment)](https://github.com/ampproject/amp-wp/pull/2240#issuecomment-489125236).

@swissspidy I'm not sure how to pull this off. I'm looking at something like this:

```diff
diff --git a/assets/src/components/with-amp-story-settings.js b/assets/src/components/with-amp-story-settings.js
index 67a0240e..042c53a0 100644
--- a/assets/src/components/with-amp-story-settings.js
+++ b/assets/src/components/with-amp-story-settings.js
@@ -48,6 +48,7 @@ const applyWithSelect = withSelect( ( select, props ) => {
 	const animationOrderEntry = animatedBlocks.find( ( { id } ) => id === props.clientId );
 
 	return {
+		select,
 		parentBlock: getBlock( getBlockRootClientId( props.clientId ) ),
 		// Use parent's clientId instead of anchor attribute.
 		// The attribute will be updated via subscribers.
@@ -125,6 +126,7 @@ export default createHigherOrderComponent(
 	( BlockEdit ) => {
 		return enhance( ( props ) => {
 			const {
+				select,
 				clientId,
 				name,
 				attributes,
@@ -156,6 +158,7 @@ export default createHigherOrderComponent(
 			}
 
 			const isImageBlock = 'core/image' === name;
+			const isVideoBlock = 'core/video' === name;
 			const isTextBlock = 'amp/amp-story-text' === name;
 			const needsTextSettings = BLOCKS_WITH_TEXT_SETTINGS.includes( name );
 			const isMovableBlock = ALLOWED_MOVABLE_BLOCKS.includes( name );
@@ -179,6 +182,14 @@ export default createHigherOrderComponent(
 				setAttributes( { caption: '' } );
 			}
 
+			// If we have a video set from an attachment but there is no poster, use the featured image of the video if available.
+			if ( isVideoBlock && attributes.id && ! attributes.poster ) {
+				const media = select( 'core' ).getMedia( attributes.id );
+				if ( media && media.featuredmedia ) {
+					setAttributes( { poster: media.featuredmedia.src } );
+				}
+			}
+
 			const minTextHeight = 20;
 			const minTextWidth = 30;
```

But it seems that a `media` REST API response does not have the `featuredmedia` populated. Help?

westonruter:
Also, I'm at a loss for how to hide the “autoplay” and “plays inline” toggles in the inspector controls.

swissspidy:
> Also, I'm at a loss for how to hide the “autoplay” and “plays inline” toggles in the inspector controls.

Same. There doesn't seem to be any class name that could be used for targeting these elements. Two possible options:

a) We can try to find a way to add some class names somewhere to allow for this
b) We just roll our own copy of the video block, at least in the meantime.

Related upstream issues:

https://github.com/WordPress/gutenberg/issues/6023
https://github.com/WordPress/gutenberg/issues/15450

swissspidy:
> But it seems that a media REST API response does not have the featuredmedia populated. Help?

Indeed. Looks like that's because the attachment post type does not officially support thumbnails, yet the UI for it is exposed in the admin... 🤷‍♂ 

I tried to come up with a workaround and 2a430d4b19dc13bda3408f2ee987c485db3c4c7a seems to do the trick.

PR #2261, opened 2019-05-06
Address an issue with resized image display

### LINKED ISSUE #2235: Image not justified in preview
Using the template in the editor the image shows justified to the width of the page. When viewing in preview it shows oddly inset. I would assume if it were to inset, it would at least inset centered. 

In editor:
Screen Shot 2019-05-02 at 7 59 36 AM

In Preview: 
Screen Shot 2019-05-02 at 7 54 09 AM

kienstra:
As @MackenzieHartung reported in #2235, there looks to be an issue with unexpected "padding" on the front-end. Though it looks like it's related to `object-fit: contain`, not padding in a strict sense.

This issue seems most pronounced when resizing the image, when it gets the `is-resized` class.

# Steps to reproduce
1. Create a new AMP Story
2. Select the template with the quote (shown below)
3. Add an Image block
4. Select a large image, like http://lorempixel.com/1400/1400/technics
5. Resize the image, even a trivial amount, so it gets the `is-resized` class.
6. Click "Update" and "Preview" to see the front-end
7. Expected: the front-end appears like the editor
8. Actual: there is left and right "padding" (though this looks to be from `object-fit: contain`)

Still, I'm not sure if `object-fit: contain` serves another purpose, or removing it will cause a side-effect. https://github.com/ampproject/amp-wp/commit/cd4c9b6516e1e2ff4b62b383e41c209e0020c103 is the latest commit affecting it.

object-fit

Closes #2235


swissspidy:
@kienstra This was recently changed to use `object-fit: cover` in c7f533d637f85eeebfea671d3eb490749e1243e8 / #2250. Does that fix the issue?

kienstra:
@swissspidy,
Ah, thanks for letting me know about #2250.

It looks like the issue still remains even with that:

# Before this PR
before-this-pr

# With this PR
with-pr

kienstra:
**This PR Probably Not Needed**

Hi @westonruter,
Good point. On testing this again, it looks like this issue is fixed on the `amp-stories-redux` branch, as @swissspidy [suggested earlier](https://github.com/ampproject/amp-wp/pull/2261#issuecomment-490137031).

# Without This PR (the amp-stories-redux branch):
with-is-resized

This PR probably isn't needed anymore.

kienstra:
I'm deleting this branch, but I still have it locally in case this is reopened.

csossi:
Verified in QA:
Editor:
image

Front-End:
image

PR #2266, opened 2019-05-07
Hide the block inserter and the shortcuts when reordering pages

### LINKED ISSUE #2248: Inserting pages while reordering
When you start the reordering, we basically copy the list of all the current pages and perform the reordering on that. Once you are done with the reordering, we update the actual order with the one from the copy. That's what allows you to cancel the reordering again.

Now, when you are reordering, you are still able to insert blocks and pages via the top toolbar. However, you won't see any new pages in the reordering panel because the copy hasn't been updated.

Some possible options here:

* When adding new pages, add these to the end of the block order copy.
* Disable/Hide all the buttons in the top toolbar when reordering.
* Do nothing.

kienstra:
As Pascal mentioned, it's confusing if you can add blocks while reordering the pages.
So this disables these buttons when reordering.

shortcuts-inserter-hidden

Closes #2248

### DIFF
--- a/assets/src/stories-editor/helpers/index.js
+++ b/assets/src/stories-editor/helpers/index.js
@@ -11,7 +11,7 @@ import { each, every, isEqual } from 'lodash';
 import { render } from '@wordpress/element';
 import { count } from '@wordpress/wordcount';
 import { _x } from '@wordpress/i18n';
-import { select, dispatch, withSelect } from '@wordpress/data';
+import { select, dispatch } from '@wordpress/data';
 import { createBlock } from '@wordpress/blocks';
 import { getColorClassName, RichText } from '@wordpress/block-editor';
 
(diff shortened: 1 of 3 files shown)


assets/src/stories-editor/helpers/index.js
```diff
@@ -519,15 +519,30 @@ export const renderStoryComponents = () => {
 		}
 	}
 
+	/**
+	 * Gets a component that conditionally displays, depending on whether the pages are reordering.
+	 *
+	 * @param {Function} component The component to conditionally display.
+	 * @return {Function|null} The passed omponent, or null.
+	 */
+	const getConditionalDisplay = ( component ) => {
```
swissspidy: We already have a `withIsReordering` higher-order component.

I would suggest removing `getConditionalDisplay` in favor of using `withIsReordering` together with the [`ifCondition`](https://github.com/WordPress/gutenberg/blob/a6c36e53974ac8a9fdd6163e61b54064cfd8910f/packages/compose/src/if-condition/README.md) HOC offered by Gutenberg. This here is a perfect use case for that.

As for the `render()` part, I would probably put the `ifCondition` and `withIsReordering` within a `compose` in `assets/src/components/shortcuts/index.js` and `assets/src/components/inserter/index.js`

kienstra: Hi @swissspidy, thanks for pointing that out. I'll apply your suggestions, but it might be a while if that's alright 😄

swissspidy: I've applied the changes now in cff2ec8f6cd57588d7edb9e86ec38579aa90d0ee

kienstra:
Nice, thanks for applying this. Sorry for the delay.

PR #2267, opened 2019-05-07
Fix auto font size displaying larger than the Text block area.

### LINKED ISSUE #2233: Auto resizing text block doesn't act as intended
When resizing a text block I would assume the text would shrink to fit inside a single line as I extend the size horizontally, however, I need to massively exaggerate the size of the text box to get the auto resized text to appear on one line.

miina:
Fixes #2233.

### DIFF
--- a/assets/src/stories-editor/helpers/index.js
+++ b/assets/src/stories-editor/helpers/index.js
@@ -634,6 +634,9 @@ export const calculateFontSize = ( measurer, expectedHeight, expectedWidth, maxF
 			minFontSize = mid;
 		}
 	}
+	// Let's restore the correct font size, too.
+	measurer.style.fontSize = minFontSize + 'px';
+
 	return minFontSize;
 };
 


swissspidy:
Looking at https://github.com/ampproject/amphtml/blob/62d3b0bcdf568f2d39fe4f71d7023dcf05b7ff9d/extensions/amp-fit-text/0.1/amp-fit-text.js, the measurer there is just a hidden clone of the actual element. And instead of doing something like `measurer.style.fontSize = minFontSize + 'px';` within `calculateFontSize`, the font size of the actual element would only be changed within `TextBlockEdit.componentDidUpdate` if `autoFontSize !== fitFontSize`.

If we did this as well, would that have any benefits for the calculation, besides `calculateFontSize` being side-effect free.

miina:
I was looking into cloning the original element when implementing it initially, however, there was some reason why it didn't seem to make sense, don't remember the reason exactly though.

Perhaps the reason was that since a block renders various times all the time then a new measurer element would be created with every `componentDidUpdate`. I guess we could create a measurer per block or we could remove the measurer element after each calculation, and then add it again, not sure if that would be a good option either. If we'd have a general universal measurer element instead used by various elements at the same time then this could have some side effects of its own. Thoughts?

> If we did this as well, would that have any benefits for the calculation, besides calculateFontSize being side-effect free.

Not that I know of.

PR #2268, opened 2019-05-07
Replace `WP_DEBUG` constant with individual constants

carlalexander:
This allows for more granular control of some of the AMP plugin behaviour. For sake of backwards compatibility, the default values of new constants use the `WP_DEBUG` constant.

The reason for this change is that we ran into issues where we needed to turn on `WP_DEBUG` on production servers for AMP pages. These servers use varnish for caching and the AMP pages would often break because of excessive `Server-Timing` headers. We needed a way to disable these headers while keeping `WP_DEBUG` active.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -241,6 +241,7 @@ add_action( 'after_setup_theme', 'amp_after_setup_theme', 5 );
  * @since 0.1
  */
 function amp_init() {
+	AMP_HTTP::$server_timing = ( defined( 'WP_DEBUG' ) && WP_DEBUG ) || current_user_can( 'manage_options' );
 
 	/**
 	 * Triggers on init when AMP plugin is active.
--- a/amp.php
+++ b/amp.php
@@ -241,7 +241,6 @@ add_action( 'after_setup_theme', 'amp_after_setup_theme', 5 );
  * @since 0.1
  */
 function amp_init() {
-	AMP_HTTP::$server_timing = ( defined( 'WP_DEBUG' ) && WP_DEBUG ) || current_user_can( 'manage_options' );
 
 	/**
 	 * Triggers on init when AMP plugin is active.
--- a/includes/class-amp-http.php
+++ b/includes/class-amp-http.php
@@ -21,6 +21,13 @@ class AMP_HTTP {
 	public static $headers_sent = array();
 
 	/**
+	 * Whether Server-Timing headers are sent.
+	 *
+	 * @var bool
+	 */
+	public static $server_timing = true;
+
+	/**
 	 * AMP-specific query vars that were purged.
 	 *
 	 * @since 0.7
@@ -83,7 +90,7 @@ class AMP_HTTP {
 	 * @return bool Return value of send_header call. If WP_DEBUG is not enabled or admin user (who can manage_options) is not logged-in, this will always return false.
 	 */
 	public static function send_server_timing( $name, $duration = null, $description = null ) {
-		if ( ! WP_DEBUG && ! current_user_can( 'manage_options' ) ) {
+		if ( ! self::$server_timing ) {
 			return false;
 		}
 		$value = $name;
--- a/includes/class-amp-http.php
+++ b/includes/class-amp-http.php
@@ -23,9 +23,17 @@ class AMP_HTTP {
 	/**
 	 * Whether Server-Timing headers are sent.
 	 *
+	 * By default this is false to prevent breaking some web servers with an unexpected number of response headers. To
+	 * enable in `WP_DEBUG` mode, consider the following plugin code:
+	 *
+	 *     add_action( 'amp_init', function () {
+	 *         AMP_HTTP::$server_timing = ( ( defined( 'WP_DEBUG' ) && WP_DEBUG ) || current_user_can( 'manage_options' ) );
+	 *     } );
+	 *
+	 * @link https://gist.github.com/westonruter/053f8f47c21df51f1a081fc41b47f547
 	 * @var bool
 	 */
-	public static $server_timing = true;
+	public static $server_timing = false;
 
 	/**
 	 * AMP-specific query vars that were purged.
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1694,34 +1694,51 @@ class AMP_Theme_Support {
 			$stream_fragment = WP_Service_Worker_Navigation_Routing_Component::get_stream_fragment_query_var();
 		}
 
+		/**
+		 * Filters whether response (post-processor) caching is enabled.
+		 *
+		 * When enabled and when an external object cache is present, the output of the post-processor phase is stored in
+		 * in the object cache. When another request is made that generates the same HTML output, the previously-cached
+		 * post-processor output will then be served immediately and bypass needlessly re-running the sanitizers.
+		 * This does not apply when:
+		 *
+		 * - AMP validation is being performed.
+		 * - The response is in the Customizer preview.
+		 * - Response caching is disabled due to a high-rate of cache misses.
+		 *
+		 * @param bool $enable_response_caching Whether response caching is enabled.
+		 */
+		$enable_response_caching = apply_filters( 'amp_response_caching_enabled', ! ( defined( 'WP_DEBUG' ) && WP_DEBUG ) || ! empty( $args['enable_response_caching'] ) );
+		$enable_response_caching = (
+			$enable_response_caching
+			&&
+			! AMP_Validation_Manager::should_validate_response()
+			&&
+			! is_customize_preview()
+		);
+
+		// When response caching is enabled, determine if it should be turned off for cache misses.
+		$caches_for_url = null;
+		if ( $enable_response_caching ) {
+			list( $disable_response_caching, $caches_for_url ) = self::check_for_cache_misses();
+			$enable_response_caching                           = ! $disable_response_caching;
+		}
+
 		$args = array_merge(
 			array(
-				'content_max_width'       => ! empty( $content_width ) ? $content_width : AMP_Post_Template::CONTENT_MAX_WIDTH, // Back-compat.
-				'use_document_element'    => true,
-				'allow_dirty_styles'      => self::is_customize_preview_iframe(), // Dirty styles only needed when editing (e.g. for edit shortcodes).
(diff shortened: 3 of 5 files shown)


carlalexander:
Signed the CLA

swissspidy:
Thanks for your contribution!

I just requested a review from Weston to hear his thoughts.

Personally, I wonder whether these settings should rather be controllable via filters instead of adding additional constants. Or maybe both.

I am also curious to learn more about why `WP_DEBUG` is needed in your use case, even with this change applied.

westonruter:
> Personally, I wonder whether these settings should rather be controllable via filters instead of adding additional constants. Or maybe both.

Yes, I agree. The `include_manifest_comment` behavior is actually already filterable via https://github.com/ampproject/amp-wp/pull/1843, so you can force its behavior irrespective of `WP_DEBUG` via:

```php
    return $sanitizers;
} );
```

The `$sanitizers` here is actually later merged on top of with the `$args` that is later constructed here:

https://github.com/ampproject/amp-wp/blob/f2b270a9d6ae3f554e51a88ed3903bbd8fd274e5/includes/class-amp-theme-support.php#L1697-L1714

What's missing is a filter for manipulating these later-introduced values. This would help specifically with `enable_response_caching`, but it wouldn't help with controlling whether `Server-Timing` is sent. So something else is needed there. A constant is certainly an option, but I wonder if there isn't something else that would better.

carlalexander:
>Personally, I wonder whether these settings should rather be controllable via filters instead of adding additional constants. Or maybe both.

I don't think a filter is a bad idea at all. I went with how the code was already working, but I'm ok reworking it to make it just a filter or both. I'm more partial to making it a filter myself.

>I am also curious to learn more about why `WP_DEBUG` is needed in your use case, even with this change applied.

It's not so much that we need `WP_DEBUG`. It's that `WP_DEBUG` gets coupled to a lot of random behaviour with WordPress plugins. (This is such an example.) We turned it on temporarily to diagnose another issue, and it ended up breaking a lot of AMP pages for our editors in the process because Varnish couldn't handle all the headers.

westonruter:
> I don't think a filter is a bad idea at all. I went with how the code was already working, but I'm ok reworking it to make it just a filter or both. I'm more partial to making it a filter myself.

Note that only a subset of the `$args` should be included in what is filterable. Maybe just start with `enable_response_caching` for now. That being the case, maybe the filter should only manage whether response caching is enabled, and have it mutate a boolean value rather than an array of args.

For the `Server-Timing` case, what about a class variable like so:

```diff
diff --git a/amp.php b/amp.php
index c483eb8b..9b6c82c9 100644
--- a/amp.php
+++ b/amp.php
@@ -241,6 +241,7 @@ add_action( 'after_setup_theme', 'amp_after_setup_theme', 5 );
  * @since 0.1
  */
 function amp_init() {
+	AMP_HTTP::$server_timing = WP_DEBUG || current_user_can( 'manage_options' );
 
 	/**
 	 * Triggers on init when AMP plugin is active.
diff --git a/includes/class-amp-http.php b/includes/class-amp-http.php
index 89565f36..8eaadf88 100644
--- a/includes/class-amp-http.php
+++ b/includes/class-amp-http.php
@@ -20,6 +20,13 @@ class AMP_HTTP {
 	 */
 	public static $headers_sent = array();
 
+	/**
+	 * Whether Server-Timing headers are sent.
+	 *
+	 * @var bool
+	 */
+	public static $server_timing = true;
+
 	/**
 	 * AMP-specific query vars that were purged.
 	 *
@@ -83,7 +90,7 @@ class AMP_HTTP {
 	 * @return bool Return value of send_header call. If WP_DEBUG is not enabled or admin user (who can manage_options) is not logged-in, this will always return false.
 	 */
 	public static function send_server_timing( $name, $duration = null, $description = null ) {
-		if ( ! WP_DEBUG && ! current_user_can( 'manage_options' ) ) {
+		if ( ! self::$server_timing ) {
 			return false;
 		}
 		$value = $name;
```

Then you could override by doing:

```php
add_action( 'amp_init', function() {
    AMP_HTTP::$server_timing = false;
} );
```

Perhaps marginally better than a constant or filter?

carlalexander:
Yep, I'm ok doing it that way! 😄

carlalexander:
Just replaced with the previous commit with a new one with just the `Server-Timing` header change

westonruter:
So the remaining thing here is a filter to control whether `response_caching_enabled`. I've got this drafted patch but I'm still debugging an issue:

```diff
diff --git a/includes/class-amp-theme-support.php b/includes/class-amp-theme-support.php
index eb7b9fb3..5f6519c6 100644
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1694,35 +1694,52 @@ class AMP_Theme_Support {
 			$stream_fragment = WP_Service_Worker_Navigation_Routing_Component::get_stream_fragment_query_var();
 		}
 
+		/**
+		 * Filters whether response (post-processor) caching is enabled.
+		 *
+		 * When enabled and when an external object cache is present, the output of the post-processor phase is stored in
+		 * in the object cache. When another request is made that generates the same HTML output, the previously-cached
+		 * post-processor output will then be served immediately and bypass needlessly re-running the sanitizers.
+		 * This does not apply when:
+		 *
+		 * - AMP validation is being performed.
+		 * - The response is in the Customizer preview.
+		 * - Response caching is disabled due to a high-rate of cache misses.
+		 *
+		 * @param bool $enable_response_caching Whether response caching is enabled.
+		 */
+		$enable_response_caching = apply_filters( 'amp_response_caching_enabled', ! ( defined( 'WP_DEBUG' ) && WP_DEBUG ) );
+		$enable_response_caching = (
+			$enable_response_caching
+			&&
+			! AMP_Validation_Manager::should_validate_response()
+			&&
+			! is_customize_preview()
+		);
+
+		// When response caching is enabled, determine if it should be turned off for cache misses.
+		$caches_for_url = null;
+		if ( $enable_response_caching ) {
+			list( $disable_response_caching, $caches_for_url ) = self::check_for_cache_misses();
+			$enable_response_caching                           = ! $disable_response_caching;
+		}
+
 		$args = array_merge(
 			array(
-				'content_max_width'       => ! empty( $content_width ) ? $content_width : AMP_Post_Template::CONTENT_MAX_WIDTH, // Back-compat.
-				'use_document_element'    => true,
-				'allow_dirty_styles'      => self::is_customize_preview_iframe(), // Dirty styles only needed when editing (e.g. for edit shortcodes).
-				'allow_dirty_scripts'     => is_customize_preview(), // Scripts are always needed to inject changeset UUID.
-				'enable_response_caching' => (
-					! ( defined( 'WP_DEBUG' ) && WP_DEBUG )
-					&&
-					! AMP_Validation_Manager::should_validate_response()
-					&&
-					! is_customize_preview()
-				),
-				'user_can_validate'       => AMP_Validation_Manager::has_cap(),
-				'stream_fragment'         => $stream_fragment,
+				'content_max_width'    => ! empty( $content_width ) ? $content_width : AMP_Post_Template::CONTENT_MAX_WIDTH, // Back-compat.
+				'use_document_element' => true,
+				'allow_dirty_styles'   => self::is_customize_preview_iframe(), // Dirty styles only needed when editing (e.g. for edit shortcodes).
+				'allow_dirty_scripts'  => is_customize_preview(), // Scripts are always needed to inject changeset UUID.
+				'user_can_validate'    => AMP_Validation_Manager::has_cap(),
+				'stream_fragment'      => $stream_fragment,
 			),
-			$args
+			$args,
+			compact( 'enable_response_caching' )
 		);
 
 		$current_url = amp_get_current_url();
 		$non_amp_url = amp_remove_endpoint( $current_url );
 
-		// When response caching is enabled, determine if it should be turned off for cache misses.
-		$caches_for_url = null;
-		if ( true === $args['enable_response_caching'] ) {
-			list( $disable_response_caching, $caches_for_url ) = self::check_for_cache_misses();
-			$args['enable_response_caching']                   = ! $disable_response_caching;
-		}
-
 		/*
 		 * Set response cache hash, the data values dictates whether a new hash key should be generated or not.
 		 * This is also used as the ETag.
```

westonruter:
@calebcordry please give 21ea847 a try.

calebcordry:
@westonruter I think you meant @carlalexander :)

jerclarke:
FWIW the issue isn't that we need to have production servers with `WP_DEBUG` enabled, it's that when we tested our AMP setup on a staging server, it broke because all these unknown and almost-never-used headers made Varnish (and sometimes Nginx) choke. Ironically our live site would have been fine, but we never got there because we tested on the `WP_DEBUG` staging first.

The plugin has to work with `WP_DEBUG` enabled or you're expecting people not to test the plugin 🤷🏻‍♀️

FWIW there's an overall lesson: Don't spam headers because it can break Varnish/Nginx. If you're going to spam a ton of headers then make sure it's an explicit opt-in thing, so that people turn them on whey they are actually going to use them, rather than having to debug for days trying to find the cause. 

I see that you've all agreed on a solution that keeps the `server-timing` headers spamming constantly by default, but FWIW consider making them silent unless requested (and then document that well) in the future. 

Thanks for replying and helping Carl get it fixed.

westonruter:
@jerclarke thanks for the additional information. You make some key points. I'll change it to default to not send `Server-Timing` by default.

westonruter:
@jerclarke @carlalexander ok, this has been implemented. `Server-Timing` is now not sent unless a plugin (such as [AMP Server-Timing](https://gist.github.com/westonruter/053f8f47c21df51f1a081fc41b47f547)) opts-in to send them.

westonruter:
Build of the plugin for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3168388/amp.zip)

Please confirm if this all works as expected.

jerclarke:
Weston this is so great, thanks a million for the changes. 

Worked today on testing it but it took awhile to get the setup ready (to "break" it on our end again) and I didn't get to testing your patch yet, but will do so soon. Just wanted to reply 'cause it's been a bunch of days now.

jerclarke:
Alright it works! At least this aspect of it. 

- Got the issue caused by `server-timing` to trigger reliably on our test server with the old version of the plugin (run `new AMP_Content` several times)
- Installed the new version of the plugin and the headers went away. 

Thanks so much!

PR #2273, opened 2019-05-08
Allow dragging rotated, unselected blocks

### LINKED ISSUE #2211: Block is de-selected after dragging
Given a text block which is selected and a grab handle is present:

image

After dragging the block anywhere, the block then becomes de-selected when letting go:

image

This is a jarring experience as the sidebar contents unexpectedly changes as well as you have to re-click in the text block to make the grab handle re-appear.

This issue is even worse for a blocks like the story title, which have no editable text. For them you have to re-select the block twice. 

1. Selected: 
image

2. Moved:
image

3. Click off the element:
image

4. Click the block again:
image

Then the grab handle appears again to be able to move it. But upon moving, these steps have to be repeated.

### LINKED ISSUE #2252: Allow dragging rotated blocks without selecting them
As a follow-up to #2014 and #1974, we should revisit allowing drag & drop of rotated blocks without the need to select them first.

This way, one can more easily position them on the right side of the screen.

miina:
Fixes #2252.

- Allow dragging rotated elements.
- Keep the element rotated at all times.
- Fix position calculation considering rotation.
- Make sure that the element stays selected after dragging and also after rotating. (Fixes #2211)

Todo separately (issue #2275):
- Based on the decision about dragging functionality perhaps remove the drag icon and allow dragging element from borders (as in Google Slides for example) or from anywhere.

### DIFF
--- a/assets/src/components/story-block-drop-zone.js
+++ b/assets/src/components/story-block-drop-zone.js
@@ -53,8 +53,6 @@ class BlockDropZone extends Component {
 			positionLeft: getPercentageFromPixels( 'x', clonePosition.left - wrapperPosition.left ),
 			positionTop: getPercentageFromPixels( 'y', clonePosition.top - wrapperPosition.top ),
 		} );
-		// Make sure the block stays selected. // @todo This is not working.
-		selectBlock( srcClientId );
 	}
 
 	render() {
--- a/assets/src/components/story-block-drop-zone.js
+++ b/assets/src/components/story-block-drop-zone.js
@@ -41,9 +41,9 @@ class BlockDropZone extends Component {
 			return;
 		}
 
-		// Get the current position of the clone.
+		// We have to remove the rotation for getting accurate position.
+		clone.parentNode.style.transform = 'none';
 		const clonePosition = clone.getBoundingClientRect();
-
 		const wrapperPosition = wrapperEl.getBoundingClientRect();
 
 		// We will set the new position based on where the clone was moved to, with reference being the wrapper element.
--- a/assets/src/components/story-block-drop-zone.js
+++ b/assets/src/components/story-block-drop-zone.js
@@ -27,7 +27,7 @@ class BlockDropZone extends Component {
 	}
 
 	onDrop( event ) {
-		const { updateBlockAttributes, srcClientId } = this.props;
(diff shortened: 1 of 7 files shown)


assets/src/components/story-block-drop-zone.js
```diff
@@ -52,6 +53,8 @@ class BlockDropZone extends Component {
 			positionLeft: getPercentageFromPixels( 'x', clonePosition.left - wrapperPosition.left ),
 			positionTop: getPercentageFromPixels( 'y', clonePosition.top - wrapperPosition.top ),
 		} );
+		// Make sure the block stays selected. // @todo This is not working.
+		selectBlock( srcClientId );
```
miina: @swissspidy Quick question: do you remember if deselecting the block after dragging and after rotating was something that you purposefully added when implementing the rotation? The block gets deselected in both cases, it's not clear why (`selectBlock` here doesn't work here). I'm asking just in case you added it purposefully, otherwise, I'll continue looking for the reason.

swissspidy: Indeed it was on purpose, see https://github.com/ampproject/amp-wp/issues/2211#issuecomment-487906591.

See `clearSelectedBlock` in `stopBlockRotation` in `assets/src/components/higher-order/with-amp-story-settings.js`

miina: Good to know, thanks!

swissspidy:
The dragging and position calculation seems to work fine.

I just have some issues with the hover states of all these elements. Not necessarily related to this change.

For example, when a block is selected but not hovered on, I see its rotation handles at all times, but the resize handles only on hover. That is a bit inconsistent.

Also, when I am rotating a block that is near another block, it triggers multi-block-selection while still doing the rotating. That is very odd and unexpected. Screenshot:

Screenshot 2019-05-09 at 10 03 10

Plus, when I am rotating that block and hover over another block, that block's resize handles become visible:

Screenshot 2019-05-09 at 10 03 32

I think these handles should not be visible while rotation is in progress.

miina:
@swissspidy Yep, this behavior seems to be not caused by this change. I'll try to address it within this PR though since it's related, thanks for testing.

miina:
Actually, the multiselection issue was related to changes in the PR, restored it.

PR #2278, opened 2019-05-09
Fix background opacity.

### LINKED ISSUE #2277: Setting background opacity is not showing in the editor
Steps to reproduce:
- Add a Text block.
- Add background color.
- Try to set background opacity.

miina:
Fixes #2277.

This PR:
- Fixes displaying background opacity in the editor, for the case where the color's value is not received from the class name but is present in the chosen color's value.
- Fixes in some cases opacity with value `100` not displaying background color at all.
- Updates templates


assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -125,7 +125,7 @@ class TextBlockEdit extends Component {
 					onChange={ ( nextContent ) => setAttributes( { content: nextContent } ) }
 					// The 2 following lines are necessary for pasting to work.
 					onReplace={ this.onReplace }
-					onSplit={ () => {} }
+					unstableOnSplit={ () => {} }
```
miina: This is replaced because `onSplit` will be deprecated soon.

swissspidy: Where can I learn more about this being deprecated soon?

swissspidy:
FWIW I started adding tests for this function in `assets/src/common/helpers/test/getBackgroundColorWithOpacity.js` (well, there's only one at the moment) and it would be great if you could amend these while making changes to that function.

assets/src/stories-editor/blocks/amp-story-text/edit.js
miina: Here is the notice: https://github.com/WordPress/gutenberg/blob/master/packages/block-editor/src/components/rich-text/index.js#L117-L126

However, just saw there is a [PR to address this](https://github.com/WordPress/gutenberg/pull/14765), so, perhaps it's better to keep it `onSplit`.

miina:
> FWIW I started adding tests for this function in assets/src/common/helpers/test/getBackgroundColorWithOpacity.js

Thanks for bringing that out, I've updated the tests.

Also reverted the `onSplit` change since there seems to be a fix for that on its way making `onSplit` great again, however, one test fails now due to deprecation warning.

assets/src/stories-editor/blocks/amp-story-page/test/__snapshots__/index.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`amp/amp-story-page block edit matches snapshot 1`] = `
 <div
-  style="background-position:cover;background-repeat:no-repeat;background-size:cover"
+  style="background-position:;background-repeat:no-repeat;background-size:cover"
```
swissspidy: Hmm, that is unrelated. Should be fixed by c3908503f6401c4d26fef1f02cbf167632f9b73b.

miina: Humm, I'm getting this without it:
Screen Shot 2019-05-10 at 6 18 36 PM

swissspidy: Yes some tests are expected to fail at the moment, as outlined in the original PR. The failed snapshot should be fixed now though as I said. Might need to merge the latest changes into this branch

miina: Yep, I was referring to Snapshots fail only.

I've merged the latest changes and ran it again, now the snapshots test passes but still with an update:
https://github.com/ampproject/amp-wp/pull/2278/files#diff-3b752b7a9ffae3b6b7cef57de4b9709dL5

swissspidy: That looks OK. Thanks for double checking.

PR #2279, opened 2019-05-09
Move block positioning to block sidebar.

### LINKED ISSUE #2181: Add Move Forward/Backward options to Block Sidebar
As discussed in #2057, this is something we might want to consider.

Alternatively, there would be the drag and drop as proposed in #1974, although not sure how obvious that is for people.

### LINKED ISSUE #2231: Layering of blocks makes actioning on functions challenging
When creating a story I am adding multiple blocks to a page. They appear layered on on another, and create frustration/challenges in trying to access the blocks functions.

Screen Shot 2019-05-02 at 5 21 23 AM

Date is behind text block:
Screen Shot 2019-05-02 at 6 17 37 AM

miina:
This PR does the following:
- Remove changing the block order from block mover.
- Add changing block mover to Toolbar or to Block Sidebar
- Add proper icons.

Fixes #2181. Fixes #2231.


miina:
@swissspidy I've created this draft PR for now which removes the up/down moving from `BlockMover` and adds it to the Block Sidebar, however, it seems to me that it could make more sense in the Toolbar instead.

It feels like it requires too many clicks to reach these controls, also, they're sometimes after some other controls. Or perhaps you had something else in mind? Screenshots (ignore the icons for now):
Screen Shot 2019-05-09 at 2 47 30 PM
Screen Shot 2019-05-09 at 2 47 20 PM

Thoughts?

cc @westonruter

swissspidy:
Let's keep the discussion in #2181

miina:
@swissspidy The icons have now been updated.

assets/src/components/higher-order/with-amp-story-settings.js
```diff
@@ -261,6 +297,43 @@ export default createHigherOrderComponent(
...
+										onClick={ sendBackward }
+										icon={ sendBackwardIcon( { width: 24, height: 24 } ) }
+									>{ __( 'Backward', 'amp' ) }</IconButton>
+								) }
+								{ ! isFirst && (
+									<IconButton
+										className="amp-story-controls-send-back"
+										onClick={ moveBack }
+										icon={ sendBackIcon( { width: 24, height: 24 } ) }
+									>{ __( 'Back', 'amp' ) }</IconButton>
```
swissspidy: The `IconButton` components in the block mover had `label`, `aria-describedby` and `aria-disabled` attributes. These ones here don't.

The buttons should be properly accessible, have tooltips, and be disabled when the block cannot be moved further forward/backward (i.e. is the last or first in list)

kienstra:
**Works Well**

Hi @miina,
This looks good, and works as expected:
positioning

assets/src/components/higher-order/with-amp-story-settings.js
miina: Yes, I think that disabling makes more sense, thanks!

miina: Added within e5e58aa. Note: AFAIK the tooltip only shows on hover if there is no content in the `IconButton`.

swissspidy:
I wasn't really happy with the look and feel, plus the `aria-describedby` attribute did not link to an actual description. So I've changed that now.

Before:

block-order-before

After:

block-order-after

Issue #2210, opened 2019-04-29
Background color of story title block doesn't automatically extend to wrapped lines


westonruter:
When text wraps in the Story Title block, I would expect the background color to extend to the new line. Instead, it only appears on the first line:

![image](https://user-images.githubusercontent.com/134745/56926776-5c1c5280-6a87-11e9-81a8-cac191958ae8.png)

![image](https://user-images.githubusercontent.com/134745/56926809-6b9b9b80-6a87-11e9-831d-7ae1beebbfb2.png)

This is even before I tried resizing the height of the block. 

The issue seems to be that it has an initial `height` defined. If it had an initial `height` of `auto` then this wouldn't be a problem.

westonruter:
Note that I am able to resize the block to make the background take up both lines, but having a fixed height would have undesirable effect if, for example, the story were viewed in a display that was wider and didn't require wrapping.

westonruter:
Another symptom of this problem, given a story date block without a background color:

![image](https://user-images.githubusercontent.com/134745/56927597-6b9c9b00-6a89-11e9-8ade-09c43f6639bc.png)

Upon setting the background color, notice that the margin at the bottom is too small:

![image](https://user-images.githubusercontent.com/134745/56927633-88d16980-6a89-11e9-9447-0e5d4366aeb5.png)

Again, I believe this is due to the fact that the block gets an initial `height` value which should instead be left undefined until the user manually resizes.

swissspidy:
> Note that I am able to resize the block to make the background take up both lines, but having a fixed height would have undesirable effect if, for example, the story were viewed in a display that was wider and didn't require wrapping.

Sounds like we'd need a way to reset the height to `auto` once you started changing it?

swissspidy:
Wanted to give this a shot myself, to see how it goes.

### Step 1: Adding the title block for a story with a long title.

Editor:

<img width="385" alt="Screenshot 2019-05-10 at 11 35 52" src="https://user-images.githubusercontent.com/841956/57558104-5816ed00-7331-11e9-80f8-1f9a3e6213a3.png">

Frontend:

<img width="463" alt="Screenshot 2019-05-10 at 11 36 13" src="https://user-images.githubusercontent.com/841956/57558107-5baa7400-7331-11e9-813c-74ac3215a59d.png">

**Observation:**

The block is very small in the editor, and the text overflows the container. On the frontend the text is displayed normally.

**Expectation:**

* The initial block height is large enough to fit the whole title
[...]

PR #2282, opened 2019-05-10
[WIP] Allow amp-fit-text for meta blocks

swissspidy:
Currently a draft because the font size calculation does not seem to work yet when there is a background color, because of `display: flex` and `display: inline` discrepancies.

Screenshot:

Screenshot 2019-05-10 at 15 59 49

See #2210.


assets/src/components/higher-order/with-amp-story-settings.js
```diff
@@ -253,7 +253,7 @@ export default createHigherOrderComponent(
 										onChange={ setFontSize }
 									/>
 								) }
-								{ isTextBlock && (
+								{ needsTextSettings && (
```
swissspidy: There's already the `needsTextSettings` check above, so this one can be safely removed

swissspidy:
I decided to merge this to the #2296 feature branch as these changes go hand in hand, especially when it comes to testing. Having everything in one place makes that easier.

PR #2292, opened 2019-05-13
Speed up PHPCS

swissspidy:
Right now, running PHPCS against the plugin takes almost 2 minutes on my MacBook Pro with PHP 7.3. I am a bit impatient and think this should be improved.

Thankfully, PHPCS supports parallel processing of files. After applying the same configuration for that as in WordPress core, running PHPCS now only takes about 20-30 seconds.

Also, I figured adding the `basepath` option makes sense. This way, PHPCS warnings will not only include the full path to the file (e.g. `/Users/Pascal/.../amp/includes/...`), but just the relative one (`./includes/...`).

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -81,6 +81,12 @@
 	<arg name="extensions" value="php"/>
 	<file>.</file>
 
+	<!-- Strip the filepaths down to the relevant bit. -->
+	<arg name="basepath" value="./"/>
+
+	<!-- Check up to 20 files simultaneously. -->
+	<arg name="parallel" value="20"/>
+
 	<exclude-pattern>*/node_modules/*</exclude-pattern>
 	<exclude-pattern>*/vendor/*</exclude-pattern>
 	<exclude-pattern>build/*</exclude-pattern>


westonruter:
[Hummm](https://travis-ci.org/ampproject/amp-wp/jobs/531911572#L524):

```
git diff develop...c51fa7940e7021fdad9808c502497e31bf30f782
fatal: Path 'amp.php' exists on disk, but not in 'c51fa7940e7021fdad9808c502497e31bf30f782'.
```

phpcs.xml
```diff
@@ -81,6 +81,12 @@
 	<arg name="extensions" value="php"/>
 	<file>.</file>
 
+	<!-- Strip the filepaths down to the relevant bit. -->
+	<arg name="basepath" value="./"/>
```
westonruter: I wonder if this is causing the Travis build failure.

swissspidy: The error comes from Git, even before any PHPUnit or PHPCS is run.

It usually happens when a branch gets deleted before the tests are run. Not sure why it happens in this case.

Given that the issue only happens for the merge commit / PR build (c51fa7940e7021fdad9808c502497e31bf30f782) and the tests run fine for the push build, it's more of a glitch on Travis side than something in this PR.

PR #2295, opened 2019-05-13
Hide alignment controls and video toggles

### LINKED ISSUE #2115: [AMP Stories] Disable Alignment for Image Block
I tried changing the alignment options for an image in AMP Stories:

image

But they had no effect on the image in the frontend. This is entirely expected, since the image is not inline with other blocks, being that it is absolutely positioned. So the alignment options should just be eliminated. This goes for other blocks as well.

### LINKED ISSUE #2283: Hide Video block toggles for Autoplay and Plays Inline
Extracted from https://github.com/ampproject/amp-wp/pull/2255, where default values have been provided for `autoplay` and `playsinline`, but what is remaining:

- [x] The Autoplay toggle for the Video block needs to be hidden/removed in AMP Stories, because `autoplay` is required.
- [x] The "Play Inline" toggle for the Video block needs to be hidden/removed _always_ since it is not allowed in AMP.

These depend on https://github.com/WordPress/gutenberg/issues/6023.

image

swissspidy:
Fixes #2115 (again).
See #2283.

Video block:

Screenshot 2019-05-13 at 15 37 16

Image block:

Screenshot 2019-05-13 at 15 39 47

Pullquote block:

Screenshot 2019-05-13 at 15 39 28

Embed block:

Screenshot 2019-05-13 at 15 40 28

### DIFF
--- a/assets/css/amp-stories.css
+++ b/assets/css/amp-stories.css
@@ -74,3 +74,11 @@ amp-story-grid-layer amp-img > img {
 .is-style-white.wp-block-quote {
 	background-image: url("../images/quote-white.svg");
 }
+
+/**
+ * See https://github.com/ampproject/amp-wp/issues/2283
+ */
+.edit-post-layout[data-block-name="core/video"] .edit-post-settings-sidebar__panel-block .block-editor-block-inspector__card + div > .components-panel__body:first-child .components-toggle-control:nth-child(2),
+.edit-post-layout[data-block-name="core/video"] .edit-post-settings-sidebar__panel-block .block-editor-block-inspector__card + div > .components-panel__body:first-child .components-toggle-control:nth-child(6) {
+	display: none;
+}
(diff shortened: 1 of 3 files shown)


westonruter:
Can this be done for the non-story editor as well? We should to be able to hide the “Plays inline” toggle for the video block in non-story posts with this:

```diff
diff --git a/assets/css/amp-editor-blocks.css b/assets/css/amp-editor-blocks.css
index 976c2f6d..3abb6b44 100644
--- a/assets/css/amp-editor-blocks.css
+++ b/assets/css/amp-editor-blocks.css
@@ -18,3 +18,10 @@
 	flex: 0 0 auto;
 	margin-right: 1rem;
 }
+
+/**
+ * See https://github.com/ampproject/amp-wp/issues/2283
+ */
+.edit-post-layout[data-block-name="core/video"] .edit-post-settings-sidebar__panel-block .block-editor-block-inspector__card + div > .components-panel__body:first-child .components-toggle-control:nth-child(6) {
+	display: none;
+}
```

Currently it is displayed:

image

Nevertheless, if someone _did_ by chance have the attribute supplied in an old post, then there would be no way to turn it off. So maybe we should just leave this toggle visible in the non-story editor. What do you think?

swissspidy:
> So maybe we should just leave this toggle visible in the non-story editor. What do you think?

Perhaps for non-story posts we could extend the logic to only hide the controls when they haven't been toggled. And perhaps also only when in native mode.

We can merge this for now to fix the issue for stories, and leave #2283 open to further investigate non-story posts if we think it makes sense.

Issue #2230, opened 2019-05-02
Text Block not Displaying Properly In Preview


MackenzieHartung:
Added a text block. Within the editor the padding around the text block seems exaggerated. When viewing the story in Preview the background does not actual form a background within the text block.

In editor:
<img width="385" alt="Screen Shot 2019-05-02 at 6 39 58 AM" src="https://user-images.githubusercontent.com/24547817/57070469-85b2c500-6ca5-11e9-8a50-27f5b7dbc4eb.png">

In preview:
<img width="390" alt="Screen Shot 2019-05-02 at 6 39 34 AM" src="https://user-images.githubusercontent.com/24547817/57070468-85b2c500-6ca5-11e9-8661-183faa8a433c.png">

Same issue happening throughout the templates:

<img width="441" alt="Screen Shot 2019-05-02 at 7 54 02 AM" src="https://user-images.githubusercontent.com/24547817/57073644-a849db80-6caf-11e9-95c6-3dd9e1b3f05a.png">

<img width="392" alt="Screen Shot 2019-05-02 at 7 54 15 AM" src="https://user-images.githubusercontent.com/24547817/57073656-af70e980-6caf-11e9-9e55-0cd0bfdb90d4.png">

miina:
This is also related to #2162 and could potentially be resolved within that.

swissspidy:
I cannot seem to reproduce this at the moment.

Let's test this again, ideally after #2296 lands.

swissspidy:
See https://github.com/ampproject/amp-wp/issues/2162#issuecomment-493570554 for testing instructions.

csossi:
Rechecking when #2162 addressed

csossi:
Verified in QA (font sizing addressed in #2359 )

PR #2296, opened 2019-05-13
Improve text block size handling

### LINKED ISSUE #2162: Text Block with background displays incorrectly when resized very small

### LINKED ISSUE #2219: Setting Background Color changes the size of text?
out

swissspidy:
With background color:

Screenshot 2019-05-13 at 16 01 14

Without background color:

Screenshot 2019-05-13 at 16 01 21

Also adds `amp-fit-text` support to meta blocks:

Screenshot:

Screenshot 2019-05-10 at 15 59 49

Fixes #2219.
Fixes #2162.
See #2230.
See #2210.


swissspidy:
The build failure is unrelated to this PR; seems to be a glitch on Travis's side.

swissspidy:
I still need to do some more testing on this myself, but would love to get some more eyes on this already.

westonruter:
Placeholder is wrapping:

> image

Upon inserting a new Text block, the cursor appears to descend below the box:

> image

As soon  as text is entered, the cursor appears as expected:

image

Colors look good in editor and frontend:

> image

> image

But as soon as I uncheck `amp-fit-text` then the background color seems to bleed outside the box:

> image

Trying to resize the box doesn't cause the background color to expand to grow:

> image

assets/src/stories-editor/helpers/index.js
```diff
@@ -684,9 +684,9 @@ export const calculateFontSize = ( measurer, expectedHeight, expectedWidth, maxF
  */
 export const getPercentageFromPixels = ( axis, pixelValue ) => {
 	if ( 'x' === axis ) {
-		return Math.round( ( pixelValue / STORY_PAGE_INNER_WIDTH ) * 100 );
+		return Number( ( ( pixelValue / STORY_PAGE_INNER_WIDTH ) * 100 ).toFixed( 2 ) );
```
miina: Note that this change will likely cause validation errors for existing blocks (including templates).

swissspidy: Yes, it totally does. But I am not sure how big of an impact the rounding has anyway, so maybe I can revert it again.

swissspidy:
I just pushed some changes here that further improve height calculation and address most of @westonruter's findings.

> Trying to resize the box doesn't cause the background color to expand to grow:

I haven't found a CSS solution yet that forces a 100% height without simultaneously breaking amp-fit-text.

If anyonee has a good idea here, please feel free to chime in. I feel like we could save us a lot of trouble by not allowing the user to change the height of text blocks and instead rely on `height: auto` everywhere.

westonruter:
> I feel like we could save us a lot of trouble by not allowing the user to change the height of text blocks and instead rely on `height: auto` everywhere.

Would that be bad to do? Also referenced in #2210.

swissspidy:
@westonruter Bad in what sense? User experience? Or amount of work needed?

westonruter:
I mean, is there a good reason to not do it? I guess there would be somewhat of a chicken-and-egg problem: if the user can't change the height of the box then the text wouldn't resize to enlarge. Or are you thinking the height could be changed by the user by dragging until which point the fit-text functionality runs after which the `height:auto` is restored?

swissspidy:
> I guess there would be somewhat of a chicken-and-egg problem: if the user can't change the height of the box then the text wouldn't resize to enlarge.

Indeed. `<amp-fit-text>` needs a size. So `height: auto` wouldn't work. However, it would work when not using `amp-fit-text` though.

So basically:

| Type            | Height      |
|-----------------|-------------|
| amp-fit-text    | set by user |
| no amp-fit-text | auto        |

The drawback is that you cannot have a huge colored box with some small text in it anymore. Might be odd for the user to be able to resize the block in one scenario but not the other.

> Or are you thinking the height could be changed by the user by dragging until which point the fit-text functionality runs after which the `height: auto` is restored?

No, but I don't think that's possible because `amp-fit-text` on the frontend needs a fixed height and no font size, so `height: auto` does not work in that case. Unless we try to fake `height: auto` by constantly increasing or decreasing the `height` based on user input.

assets/src/stories-editor/helpers/index.js
miina: It actually might have a noticeable impact within #2310, will test it out.

miina: Yes, it does have a noticeable impact there, will add it within that PR.

swissspidy: works for me!

swissspidy:
FYI I've decided to consolidate this with the work on #2282. So this PR just got a little bit bigger.

swissspidy:
@miina @westonruter @kienstra Time for another round of testing!

Amongst other things I've tried to basically fake `height:auto` for text-based blocks in the last few iterations here.

Without much further ado, I'd love if you all could try creating some blocks (text and "meta" blocks) with both amp-fit-text on and off, to see how good it works.

miina:
@swissspidy Do I understand correctly that it's intended that resizing the text blocks is not possible anymore unless `amp-fit-text` is selected? This feels like a bug. If it shouldn't be possible then we should hide the resizer.

resizing

Maybe here we should check just if the changed height is smaller, not equal?
`if ( element && element.offsetHeight !== height ) {`

In case of using AMP Fit Text, the editor still seems to display a different look from the FE.

FE:
Screen Shot 2019-05-16 at 10 48 21 AM

Editor:
Screen Shot 2019-05-16 at 10 48 12 AM

Could we fill in the area in the editor as well, e.g. by assigning 100% height to the inner divs of the blocks, something like:
```
.wp-block-amp-story-text,
.wp-block-amp-story-text * {
	height: 100%;
}
.is-amp-fit-text {
	height: 100%;
}
```

For centering the text vertically in the editor to match the view in the frontend, we could do something like this (in case of the Text block):
```
.wp-block[data-type="amp/amp-story-text"] .wp-block-amp-story-text p.is-amp-fit-text {
	display: block;
	display: -webkit-box;
	display: -ms-flexbox;
	display: flex;
	-webkit-box-orient: vertical;
	-webkit-box-direction: normal;
	-ms-flex-direction: column;
	flex-direction: column;
	-ms-flex-wrap: nowrap;
	flex-wrap: nowrap;
	-webkit-box-pack: center;
	-ms-flex-pack: center;
	justify-content: center;
}
```

And then, not to break the amp fit text font calculation in the editor, we could do the following:
1. Add the following CSS rule:
```
.wp-block[data-type="amp/amp-story-text"] .wp-block-amp-story-text p.is-amp-fit-text.is-measuring-fontsize {
	height: initial !important;
}
```
2. Assign `is-measuring-fontsize` className to the measurer when the measuring starts within `calculateFontSize`, e.g.
`measurer.classList.add( 'is-measuring-fontsize' );`
3. Remove it once the font size has been determined, e.g.
`measurer.classList.remove( 'is-measuring-fontsize' );`.

Thoughts?

swissspidy:
> Do I understand correctly that it's intended that resizing the text blocks is not possible anymore unless `amp-fit-text` is selected?

No, that is not the intention. That is a bug.

> Could we fill in the area in the editor as well, e.g. by assigning 100% height to the inner divs of the blocks

> And then, not to break the amp fit text font calculation in the editor, we could do the following [...] Thoughts?

I like that idea. I'm gonna try it.

swissspidy:
Much better, but still not perfect.

Sometimes while resizing the font size is removed completely. Haven't yet figured out why.

Video: https://cloudup.com/cJivRMMiOof

Also, there's still the issue with the cursor that Weston mentioned. I suppose what we need to do there is account for the 7px padding when changing the initial block height.

miina:
> Sometimes while resizing the font size is removed completely. Haven't yet figured out why.

I just saw this on the `amp-stories-redux` branch as well, so this seems to be unrelated to the changes in the PR and to the bugs reported (the ones mentioned in the description). I'd vote for opening a separate issue for that one. Thoughts?

> Also, there's still the issue with the cursor that Weston mentioned

I actually can't see this issue right now. Should it happen when inserting a new Text block, without changing anything? Is this issue still occurring for you?

assets/src/stories-editor/blocks/amp-story-text/edit.css
```diff
@@ -47,6 +51,19 @@ div[data-type="amp/amp-story-page"] .block-editor-inner-blocks .block-editor-blo
 	display: inline-block;
 }
 
-.editor-styles-wrapper .wp-block-amp-story-text p.is-amp-fit-text {
+.wp-block[data-type="amp/amp-story-text"] .wp-block-amp-story-text .is-amp-fit-text {
```
miina: Should the same be added for the title and the author blocks as well?

swissspidy: When this is proven to be working for this block, yes.

swissspidy:
> I actually can't see this issue right now. Should it happen when inserting a new Text block, without changing anything?

To reproduce:

1. Create new story.
2. Select the text block that is automatically there
3. Disable amp-fit-text
4. Select text block again
5. Notice cursor position

Because of the padding the text (and thus the cursor) is a bit too low:

Screenshot 2019-05-16 at 15 45 33

assets/src/stories-editor/blocks/amp-story-text/edit.css
miina: Okay, makes sense. Right now it's still not possible to change the height of the title and author block unless `amp-fit-text` is used, I guess that's related? (The first comment within https://github.com/ampproject/amp-wp/pull/2296#issuecomment-492996501, it seems to be working OK now for the Text block but not for other text-like blocks).

swissspidy: Ah, the gif in that comment was for the text block. I didn't realize that is was broken for the meta blocks.

miina:
> Because of the padding the text (and thus the cursor) is a bit too low:

Aha, okay, now I can see, tried with amp-fit-text before.

> I suppose what we need to do there is account for the 7px padding when changing the initial block height.

Could we just increase the default a little bit, e.g. from 50 to 60? Is that's what you had in mind by changing the initial block height?

swissspidy:
> Could we just increase the default a little bit, e.g. from 50 to 60? Is that's what you had in mind by changing the initial block height?

Yes.

assets/src/stories-editor/blocks/amp-story-text/edit.css
swissspidy: Done now.

assets/src/stories-editor/helpers/index.js
```diff
@@ -1056,36 +1076,75 @@ export const maybeUpdateAutoAdvanceAfterMedia = ( clientId ) => {
...
+			break;
+
+		case 'amp/amp-story-post-author':
+		case 'amp/amp-story-post-date':
+		case 'amp/amp-story-post-title':
+			if ( height === getDefaultMinimumBlockHeight( name ) || ! ampFitText ) {
+				const slug = name.replace( '/', '-' );
+
+				const element = document.querySelector( `#block-${ clientId } .wp-block-${ slug }` );
+				if ( element && element.offsetHeight > height ) {
```
swissspidy: For some reason the height calculation for meta blocks is different than for other blocks, that's why the previous `element.offsetHeight !== height` condition prevented resizing these blocks. Now with `element.offsetHeight > height` you cannot make them smaller than the size that they require, which I guess is fine.

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -20,11 +20,9 @@ import { select } from '@wordpress/data';
  */
 import { calculateFontSize } from '../../helpers';
 import { getBackgroundColorWithOpacity } from '../../../common/helpers';
+import { MIN_FONT_SIZE, MAX_FONT_SIZE } from '../../constants';
```
kienstra: Good idea to store these in `../../constants`

kienstra:
Thanks, I'm reviewing this now. It looks good so far 😄

assets/src/stories-editor/helpers/index.js
```diff
@@ -912,7 +932,7 @@ export const getStylesFromBlockAttributes = ( {
 	return {
 		backgroundColor: appliedBackgroundColor,
 		color: textClass ? undefined : customTextColor,
-		fontSize: ampFitText ? autoFontSize : fontSizeResponsive,
+		fontSize: ! ampFitText ? fontSizeResponsive : undefined,
```
swissspidy: This might cause template invalidation. Will go through templates after merge.

PR #2129, opened 2019-04-14
Update amphtml to v1904021746450

westonruter:
Previously #2041.

* Add `amp-script`. Note that it is not yet uncommented in [`validator-amp-script.protoascii`](https://github.com/ampproject/amphtml/blob/bddc60cefc9c653693a13abf48ca4c25a9523cbe/extensions/amp-script/validator-amp-script.protoascii) so the AMP validator will mark pages with `amp-script` as invalid, but it is being uncommented to facilitate experimentation (as otherwise the sanitizer would strip it out).
* Allow `ol` inside `nav` like `ul` is allowed.
* Allow `track` under `amp-ima-video`.
* Require that `template` be inside the `body`.
* Deprecate `amp-story` version `0.1`.

See https://github.com/ampproject/amphtml/releases/tag/1904021746450

The `amp-script` extension is forcibly included here by patching `amphtml` as follows before running `amphtml-update.sh`:

```diff
diff --git a/extensions/amp-script/validator-amp-script.protoascii b/extensions/amp-script/validator-amp-script.protoascii
index 19e132e7b..19063d51c 100644
--- a/extensions/amp-script/validator-amp-script.protoascii
+++ b/extensions/amp-script/validator-amp-script.protoascii
@@ -15,39 +15,38 @@
 #
 
 # Invalid until launch ready.
-# tags: {
-#   tag_name: "SCRIPT"
-#   extension_spec: {
-#     name: "amp-script"
-#     version: "0.1"
-#     version: "latest"
-#   }
-#   attr_lists: "common-extension-attrs"
-# }
+tags: {
+  tag_name: "SCRIPT"
+  extension_spec: {
+    name: "amp-script"
+    version: "0.1"
+    version: "latest"
+  }
+  attr_lists: "common-extension-attrs"
+}
 
-# tags: {  # <amp-script>
-#   html_format: AMP
-#   tag_name: "AMP-SCRIPT"
-#   disallowed_ancestor: "AMP-SCRIPT"
-#   requires_extension: "amp-script"
-#   unique: true # TODO(choumx): Remove when global size limit is implemented.
-#   attrs: {
-#     name: "src"
-#     mandatory: true
-#     value_url: {
-#       protocol: "https"
-#       allow_relative: false
-#     }
-#     blacklisted_value_regex: "__amp_source_origin"
-#   }
-#   attr_lists: "extended-amp-global"
-#   amp_layout {
-#     supported_layouts: CONTAINER
-#     supported_layouts: FILL
[...]

PR #2297, opened 2019-05-13
Update amphtml to 1905021827420

westonruter:
Previously #2129.

* Add `object-fit` and `object-position` to `amp-img`, `amp-anim`, and `amp-video`.
* Add support for `amp-link-rewriter`.
* Upgrade `amp-experiment` from `0.1` to `1.0`.
* Remove requirements for `amp-addthis` data attributes.


includes/sanitizers/class-amp-allowed-tags-generated.php
```diff
@@ -11878,6 +11902,7 @@ class AMP_Allowed_Tags_Generated {
 						'requires_usage' => 2,
 						'version' => array(
 							'0.1',
+							'1.0',
```
westonruter, line 11905: This change broke a site. See https://wordpress.org/support/topic/google-optimize-integration-2/

The `1.0` version of `amp-experiment` is not stable yet (it's still itself experimental), necessitating sites to roll back to `0.1`.

westonruter: Opened issue to address: https://github.com/ampproject/amp-wp/issues/4628

Issue #2306, opened 2019-05-14
Dragging not working after resizing/rotating


amedina:
The gif looks bad, but the issue can be observed:

![out](https://user-images.githubusercontent.com/506089/57711883-74ae7000-7624-11e9-8b61-665646325413.gif)

miina:
Note: this is happening for all blocks, not just for the image block.

Steps to reproduce:
- Add a block.
- Rotate or resize it.
- Try to drag right after.

miina:
Instructions for testing:
- Add any block.
- Resize or rotate the block.
- Without clicking anything else, try to drag the block.
- Verify that dragging works.

csossi:
Verified in QA (selection not required for drag)

Issue #2275, opened 2019-05-09
Consider removing drag handle in favor of dragging the block itself


miina:
Currently, the block can only be dragged from the drag handle.

It has been suggested that removing the drag handle and allowing the user to drag the block from anywhere might be better user experience, that's also something that some other apps already allow.

This would, however, would introduce a new issue of not being able to "drag" inside the block for other reasons, such as selecting text for copying and pasting or selecting text for marking part of the text italic / bold, etc.

An alternative could be allowing dragging from the borders, in a similar way that Google Slides has, for example.

Related: #2252.

miina:
Created this issue since that's something that has been under discussion, however, couldn't find an existing issue. Let me know if an issue already existed.

See also https://github.com/ampproject/amp-wp/issues/2252#issuecomment-490502947.

cc @amedina @swissspidy @westonruter

swissspidy:
> This would, however, would introduce a new issue of not being able to "drag" inside the block for other reasons, such as selecting text for copying and pasting or selecting text for marking part of the text italic / bold, etc.

What if we just allow dragging when the block is not selected? If it's selected (i.e. you clicked on the block), then you can do all the text selection you'd want.

> An alternative could be allowing dragging from the borders, in a similar way that Google Slides has, for example.

I fear that this might add quite some complexity to the matter.

miina:
> What if we just allow dragging when the block is not selected? If it's selected (i.e. you clicked on the block), then you can do all the text selection you'd want.

Hmm, this might feel like a bug, not like a feature to the user :D It would technically make things easier for us, however, would be great if we could implement the simplest and most intuitive version for the users.

> I fear that this might add quite some complexity to the matter.

It might. One option could be adding 4 separate draggable areas around the block which would look like borders. So basically having 4 drag handles instead of one, restyled. The resizing handles would have priority. Also, perhaps have some time with this as this issue hasn't specifically brought out as in scope for the next week?

Thoughts?

swissspidy:
In a first version, let's allow dragging the whole block, without any complex logic. Then, let's iterate from there.

miina:
> In a first version, let's allow dragging the whole block, without any complex logic. Then, let's iterate from there.

This will break allowing italic and bold in a Text block, these are existing features currently. Are you suggesting that we should remove these features for now?

swissspidy:
[...]

PR #2310, opened 2019-05-14
Fix resizing for rotated blocks

### LINKED ISSUE #2252: Allow dragging rotated blocks without selecting them

miina:
Fixes #2252.


assets/src/components/resizable-box/index.js
```diff
@@ -34,13 +46,84 @@ export default ( props ) => {
...
+						left: originalPos.left - diff.left,
+						top: originalPos.top + diff.top,
+					};
+
+					blockElement.style.left = getPercentageFromPixels( 'x', updatedPos.left ) + '%';
+					blockElement.style.top = getPercentageFromPixels( 'y', updatedPos.top ) + '%';
+				}
+
+				/*if ( 0 && angle ) {
+					switch ( direction ) {
```
miina: @swissspidy If you have time, would you mind taking a look at this logic to see if you see if something seems to be off or if you have a better idea, perhaps you have a "fresher" eye. I was trying different options and for some reason, the calcuations didn't work as expected, not sure what I was missing. Even brought out a pencil and paper and everything seemed to make sense on paper, ha! I left in the previous tries and just commented it out to check again tomorrow.

Anyway, at this moment the code above kind of works, not smoothly though.

assets/src/stories-editor/blocks/amp-story-page/edit.css
```diff
@@ -126,7 +126,7 @@ div[data-type="amp/amp-story-page"] .editor-inner-blocks .editor-block-list__lay
 
 .editor-block-list__layout div[data-type="amp/amp-story-page"][data-amp-selected="parent"] .editor-block-drop-zone {
 	height: 583px;
-	width: 438px;
+	width: 558px;
```
miina: Note: this is for extending the dropzone to be able to move blocks over the right edge of a page, too.

assets/src/components/higher-order/with-amp-story-settings.js
```diff
@@ -231,6 +231,7 @@ export default createHigherOrderComponent(
 							isSelected={ isSelected }
```
miina: There's another issue here -- image block has its own resizable box, however, that does not consider rotation. Any thoughts here? Should we try to hide the original one with CSS and replace it with the custom one instead?

swissspidy: We did the same with the block movers already. I think there's no way around doing this for the resizable box as well. Go for it!

miina: Just made this change within 3e7c44f.

assets/src/stories-editor/blocks/amp-story-page/edit.css
swissspidy: So this is basically the size of the page plus some padding? Can we use some CSS variables for that to make this more clear?

assets/src/stories-editor/helpers/index.js
```diff
@@ -1143,6 +1143,63 @@ export const maybeInitializeAnimations = () => {
...
+ * @param {Object} event MouseEvent.
+ * @param {number} angle Rotation angle.
+ * @param {number} lastSeenX Starting X coordinate.
+ * @param {number} lastSeenY Starint Y coordinate.
+ * @param {string} direction Direction of resizing.
+ * @return {Object} Width and height values.
+ */
+export const getDelta = ( event, angle, lastSeenX, lastSeenY, direction ) => {
+	let deltaH = 'bottom' === direction ? Math.sqrt( Math.pow( event.clientX - lastSeenX, 2 ) + Math.pow( event.clientY - lastSeenY, 2 ) ) : 0;
+	let deltaW = 'right' === direction ? Math.sqrt( Math.pow( event.clientX - lastSeenX, 2 ) + Math.pow( event.clientY - lastSeenY, 2 ) ) : 0;
```
swissspidy: It might be useful to have small helper function for this, e.g. something like this:

```js
const getDelta = ( deltaX, deltaY ) => Math.sqrt( deltaX * deltaX + deltaY * deltaY );

let deltaH = 'bottom' === direction ? getDelta( event.clientX - lastSeenX, event.clientY - lastSeenY ) : 0;
let deltaW = 'right' === direction ? getDelta( event.clientX - lastSeenX, event.clientY - lastSeenY ) : 0;
```

It's not only DRY but also implicitly documented.

swissspidy:
A few observations:

* While resizing a rotated element it can be somewhat shaky, i.e. moving for a few sub-pixels  
  Happens with [this other React library](https://mockingbot.github.io/react-resizable-rotatable-draggable/) too, so I guess there's no way to avoid it
* Sometimes the resize handle jumps a bit, see video below for a demo
* Resizing an image crops it off instead of proportionally resizing it like in the regular posts editor
* It might make sense to have resize handlers on every side, so you can easily resize a block even when it's rotated in a weird way
* Some tests for these new helper functions would be nice, and they would also help with documentation

Video: https://cloudup.com/cgkCeM4qlHg

Editor view end result:

Screenshot 2019-05-15 at 17 47 21

Frontend view end result:

Screenshot 2019-05-15 at 17 47 26

assets/src/stories-editor/blocks/amp-story-page/edit.css
miina: Done in 33a7d6d.

miina:
@swissspidy Thank you for testing!

> While resizing a rotated element it can be somewhat shaky, i.e. moving for a few sub-pixels
Happens with this other React library too, so I guess there's no way to avoid it

Yes, it seems like it's not possible to avoid it fully.

> Sometimes the resize handle jumps a bit, see video below for a demo

Not sure exactly why that happens, it seems to happen around the spot where the resizing event starts, looking into this.

> Resizing an image crops it off instead of proportionally resizing it like in the regular posts editor

This should be fixed now.

> It might make sense to have resize handlers on every side, so you can easily resize a block even when it's rotated in a weird way

Maybe that's something that we could add once the drag handle has been removed? I'll look into that as the last thing once the rest already works. It will need some additional calculations.

> Some tests for these new helper functions would be nice, and they would also help with documentation

Good point! Once it actually works well enough, will add the tests.

miina:
@swissspidy I've added some tests and hopefully addressed most of your feedback. Could you give it another try? Note that I didn't add the `top` and `left` resizing right now, this would need reworking the calculation for the new positioning of the block and I'm not sure if I'll have enough time in time. Also, in some cases the drag handle would collide with the left resizer so we could potentially implement it together with #2275.

swissspidy:
Awesome stuff! 🎉 

A few observations for this round of testing:

Lowering the height too much results in the block's position being changed in an odd way. Maybe we can enforce some limit here?

Video: https://cloudup.com/iyj1lYp7_Ow

Image still seems to get cropped. Video: https://cloudup.com/iPfn0sY6cii

Also, the resizing isn't as smooth as in the regular post editor. Video: https://cloudup.com/ichQUmFl3mW

Not sure if fixing that one is easily possible.

swissspidy:
Pinging @westonruter @kienstra for some testing as well.

miina:
> Image still seems to get cropped.

Actually, I think it should now be the same behavior as in the FE -- it's using `cover`. For example:
FE:
Screen Shot 2019-05-17 at 12 02 39 AM

Editor:
Screen Shot 2019-05-17 at 12 02 35 AM

Original image size:
Screen Shot 2019-05-17 at 12 04 49 AM

This should be the related PR: #2250.

Thoughts?

> Lowering the height too much results in the block's position being changed in an odd way. Maybe we can enforce some limit here?

Hmm, that shouldn't probably happen, will try to fix.

> Also, the resizing isn't as smooth as in the regular post editor. Video: https://cloudup.com/ichQUmFl3mW
Not sure if fixing that one is easily possible.

At this moment I'm not sure how to make it smoother. We could maybe open a separate bug issue for that if it seems to be distracting the editing flow. Thoughts?

swissspidy:
> At this moment I'm not sure how to make it smoother. We could maybe open a separate bug issue for that if it seems to be distracting the editing flow.

Works for me. Resizing also affects the `amp-fit-text` font size calculation, so I guess we can tackle these two things together.

swissspidy:
As for the image size, I guess it makes sense then.

assets/src/components/resizable-box/index.js
```diff
@@ -34,13 +47,64 @@ export default ( props ) => {
...
+						top: resizedPosition.top - initialPosition.top,
+					};
+					// Get new position based on the difference.
+					const originalPos = {
+						left: getPixelsFromPercentage( 'x', parseInt( blockElementLeft, 10 ) ),
+						top: getPixelsFromPercentage( 'y', parseInt( blockElementTop, 10 ) ),
+					};
+
+					// @todo Figure out why calculating the new top / left position doesn't work in case of small height value.
+					// @todo Remove this temporary fix.
```
miina: @swissspidy I changed the max and min sizes for resizing based on the values in your PR #2296. Also added this check here for when height is less than 60, since that's when the shift starts, however, I'm not sure why it happens. Added this fix here for now since might not have time to figure out the reason behind it. Let me know if it seems to work a bit better now.

westonruter:
I'm getting an error when I select an image from the media library:

> RangeError: Maximum call stack size exceeded

image

Testing in Gutenberg 5.7.

miina:
@westonruter I'm getting the same error on the `amp-stories-redux` branch, this doesn't seem to be specific to this PR.

Created a separate issue for that: https://github.com/ampproject/amp-wp/issues/2350

westonruter:
@miina Thanks, the error is fixed.

Otherwise, I'm having a difficult time moving the block. It seems I have to deselect the block and re-select it in order to be able to drag the block with the handle. If I don't rotate or resize, then I can drag immediately after selecting. But if I resize or rotate the block, then the image bounces back to the original location after dragging (even though the ghost does move as I drag).

miina:
@westonruter This issue will get fixed within: https://github.com/ampproject/amp-wp/issues/2306
It's also a general and not this PR-specific issue.

westonruter:
@miina Got it. Thanks. One other thing: I noticed with the changes in this PR, an image's actual size does not change until I let go of the resize handle. Prior to this PR, the image would resize _while_ I move the resize handles. So this new behavior feels less user-friendly in how the image appears during resizing. Is that fixable?

miina:
@westonruter Good catch. We could probably create special handling for the image block to assign the current value to its wrapper while resizing, it should be possible.

We could create a separate bug issue for that, alternatively I could look into it tomorrow within this PR (won't have time tonight). Thoughts?

PR #2319, opened 2019-05-14
Check that required classes/functions have not been disabled

westonruter:
Checking that the required PHP extensions are present is not enough, as some of the functions/classes in these extensions may be disabled (e.g. `curl_multi_init()`). So we need to check that the functions and classes we use are actually present in addition to their extensions being loaded.

Amends #2183.

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -37,24 +37,68 @@ if ( version_compare( phpversion(), '5.4', '<' ) ) {
 
 // See composer.json for this list.
 $_amp_required_extensions = array(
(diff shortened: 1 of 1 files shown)


amp.php
```diff
@@ -37,24 +37,68 @@
 
 // See composer.json for this list.
 $_amp_required_extensions = array(
-	'curl',
-	'date',
```
westonruter: I removed `date` because this is actually a WordPress dependency.

PR #2308, opened 2019-05-14
Patch test_add_block_source_comments for Gutenberg 5.7 update

westonruter:
Gutenberg 5.7 adds a `wp-block-latest-posts__list` class to the Latest Posts block wrapper. This PR updates the tests to account for that change while also facilitating a transition from 5.6 to 5.7.

Issue #1820, opened 2019-01-16
Account for toggleClass in CSS tree shaker


westonruter:
Consider the following code:

```php
<style amp-custom>
#concealer {
    display:none;
}
#concealer.expanded {
    display:block;
}
</style>
...
<button on="tap:concealer.toggleClass(class='expanded')" type=button>Toggle Expanded</button>
<div id="concealer">Some Content</div>
```

The tree shaker will currently purge this CSS rule:

```css
#concealer.expanded {
    display:block;
}
```

The reason is because there is no DOM element with the `expanded` class.

Note that the tree shaker _does_ handle a related case, where amp-bind is used to toggle the class. It does this by extracting the string literals in `[class]` contained inside of it and using these as additional class names: 

https://github.com/ampproject/amp-wp/blob/732f789d5bbb73b0487ca0534ec28d2fec4e38a7/includes/sanitizers/class-amp-style-sanitizer.php#L364-L369

So with the above `style`, the CSS rule will not be shaken out in the equivalent amp-bind code:

```html
<amp-state id="expanded">false</amp-state>
<button on="tap:AMP.setState({expanded: ! expanded})" type=button>Toggle Expanded</button>
<div id="concealer" [class]="expanded ? 'expanded' : '' ">Some Content</div>
```

We need to add similar logic for the `toggleClass` case. This will require looking for `on` attributes to find ones that contain `toggleClass` and then considering its argument to be among the class names on the target element.

davisshaver:
@westonruter I think this is similar to https://github.com/ampproject/amp-wp/issues/1490?

westonruter:
It's definitely related.

PR #2327, opened 2019-05-15
Add support in nav-menu-toggle-sanitizer for nav_container to be either the body or root element

felixarntz:
It is possible to specify the `body` element as the target to modify the class for the navigation menu toggle to indicate whether the menu is currently toggled on or off. Several themes actually do this, because it allows the entire page (or more precisely everything in `body`) to be styled based on that class.

Currently this is broken because the current implementation will try to place the corresponding `amp-state` tag before the respective element, which in this case is between `head` and `body` which is obviously not valid.

This PR ensures that in case the `body` is specified the `amp-state` tag is inserted differently, as first child of `body`.


includes/sanitizers/class-amp-nav-menu-toggle-sanitizer.php
```diff
@@ -99,6 +99,11 @@ public function sanitize() {
 		$script_el->appendChild( $this->dom->createTextNode( wp_json_encode( $expanded ) ) );
 		$state_el->appendChild( $script_el );
 		$nav_el->parentNode->insertBefore( $state_el, $nav_el );
+		if ( 'body' === $nav_el->nodeName ) {
+			$nav_el->insertBefore( $state_el, $nav_el->firstChild );
```
westonruter: I like `insertBefore` rather than `appendChild` because it puts the `amp-state` closer to the element attributes that are being modified.

Just note that `amp-auto-ads` has docs that state:

> The `<amp-auto-ads>` tag should be placed as the first child of the `<body>`.

The key word is _should_. It actually does not cause a validation error to have it elsewhere than the first child:

image

This has been important for WordPress since before `wp_body_open` we've had to rely on `wp_footer` which has worked.

For `style[amp-keyframes]` it is also [specified](https://amp.dev/documentation/guides-and-tutorials/learn/spec/amphtml#keyframes-stylesheet):

> May only be placed as the last child of the document's `<body>` element.

This is an actual _hard_ error:

image

All this to say, `insertBefore` like you have here is very correct.

westonruter:
@felixarntz please rebase your commit on top of `develop` and likewise change the target to `develop`. This will fix the build error. Ref https://github.com/ampproject/amp-wp/pull/2308.

westonruter:
Please rebase to `develop`. I'll cherry-pick onto the `1.1` branch. I think you may have rebased against the `1.1` branch.

swissspidy:
Was passing by and thought I'll do it real quick. Messed it up at first, hence the force push, but now it should be good. Sorry for the noise 🙂

includes/sanitizers/class-amp-nav-menu-toggle-sanitizer.php
```diff
@@ -99,6 +99,11 @@ public function sanitize() {
 		$script_el->appendChild( $this->dom->createTextNode( wp_json_encode( $expanded ) ) );
 		$state_el->appendChild( $script_el );
 		$nav_el->parentNode->insertBefore( $state_el, $nav_el );
+		if ( 'body' === $nav_el->nodeName ) {
```
swissspidy: FWIW I've used the `html` element in the past for toggle classes. Can we test against this as well?

Speaking of testing... Are there unit tests for that?

westonruter: I was worried about that as well, but we tested it together and it didn't seem to cause any problem. 

In particular, I was worried about classes that AMP adds dynamically, including `amp-mode-mouse` or any of the classes that `amp-geo` adds. However, it appears that AMP is smart enough to persist such class names even when mutating the class via `amp-bind`. See:

https://github.com/ampproject/amphtml/blob/c2cf35069cd740822c99538830d29aedceaced12/extensions/amp-bind/0.1/bind-impl.js#L1155-L1173

A safer solution I think would actually be to make use of AMP's [`toggleClass()` action](https://amp.dev/documentation/guides-and-tutorials/learn/amp-actions-and-events#*-(all-elements)), although this would depend on (1) the container having an ID and (2) that we support `toggleClass` in the tree shaker (https://github.com/ampproject/amp-wp/issues/1820). This would ensure that multiple components could toggle classes on the `body` without stomping on each other.

felixarntz:
@swissspidy Awesome, thanks!

@westonruter 
> I like `insertBefore` rather than `appendChild` because it puts the amp-state closer to the element attributes that are being modified.

That was my thinking as well.

> Just note that `amp-auto-ads` has docs that state:

We could consider accounting for that in the future, but since it's not a hard error, I think it's okay to move forward with this.

includes/sanitizers/class-amp-nav-menu-toggle-sanitizer.php
swissspidy: So it works even when using `html` and not just `body`?

westonruter: Oh, that's a good point.

westonruter: @felixarntz if `$nav_el` is the `$this->dom->documentElement` then the `amp-state` needs to be added to `body`, not the `$nav_el`.

westonruter:
If you want to make use of `toggleClass` (perhaps in a subsequent PR), I have a PR for that: #2328.

westonruter:
> If you want to make use of `toggleClass` (perhaps in a subsequent PR), I have a PR for that: #2328.

It looks like this would actually complicate things because of the need to update the `aria-expanded` attribute on the button, and using amp-bind is easiest here. Otherwise we'd have `toggleClass` managing the state in one place and `amp-state` storing the state in another.

westonruter:
Would be good to add some tests for this as well.

westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3213298/amp.zip) (1.2-beta1-20190523T165154Z-35f9b0c6)

felixarntz:
I can get to adding tests either this Friday or early next week.

felixarntz:
@westonruter Tests are now present, this is good for merge from my end.

PR #2329, opened 2019-05-16
Greatly simplify createSkeletonTemplate

### LINKED ISSUE #2307: Improve skeleton template inserter
Related: #2289.

`createSkeletonTemplate` should be improved in a way that it resets _all_ block attributes, not just the ones in `emptyTemplateMapping`.

For example, animations, gradients, and opacity are all left untouched when inserting a skeleton template.

I think we might be able to use `getBlockType` in `createSkeletonTemplate` to get all available attributes, then reset them to their default value. Perhaps we could even use `{}` instead of all the `getSkeletonTemplateBlock` logic.

swissspidy:
Basically resets all block attributes back to their defaults when inserting a template. No more background colors that you first need to remove.

Fixes #2307.

### DIFF
--- a/assets/src/stories-editor/helpers/index.js
+++ b/assets/src/stories-editor/helpers/index.js
@@ -3,7 +3,7 @@
  */
 import uuid from 'uuid/v4';
 import classnames from 'classnames';
-import { each, every, isEqual } from 'lodash';
+import { every, isEqual } from 'lodash';
 
 /**
  * WordPress dependencies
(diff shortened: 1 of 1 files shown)


miina:
@swissspidy Shouldn't we leave the position, width, and height of the blocks? For some reason I understood that the intent was to remove the background color, font sizes, colors, etc. but still leave the positioning and size, however, I might have imagined that. Confirming just in case.

swissspidy:
Nope, you‘re right. I guess I was simplifying too much here :)

swissspidy:
Made the changes now.

assets/src/stories-editor/helpers/index.js
```diff
@@ -758,56 +758,21 @@ export const addBackgroundColorToOverlay = ( overlayStyle, backgroundColors ) =>
...
-	}
-
+const resetBlockAttributes = ( block ) => {
 	const attributes = {};
-	each( block.attributes, function( value, key ) {
-		if ( undefined === emptyTemplateMapping[ block.name ][ key ] ) {
-			attributes[ key ] = value;
-		}
-	} );
+	const attributesToKeep = [ 'positionTop', 'positionLeft', 'width', 'height', 'ampFitText', 'tagName' ];
```
miina: Wondering if it makes sense to keep `ampFitText` unless we're keeping the font size as well. Otherwise, if the area is very small and `ampFitText` is set to `false` but there is no font size assigned then it would be the default font size which would be too large. Maybe we could either remove `ampFitText` as well (basically setting it to default which is `true`) or alternatively keep the font size, too. Thoughts?

swissspidy: I am a bit on the fence with regards to where to draw the line with all these attributes. For example, font family also affects these things.

Given that `ampFitText` gives for a much better editing experience, I am inclined to remove that one here too though.

PR #2331, opened 2019-05-16
Fix assigning incorrect attributes to image block.

miina:


assets/src/stories-editor/helpers/index.js
```diff
@@ -144,9 +144,6 @@ const getDefaultMinimumBlockHeight = ( name ) => {
 		case 'core/embed':
 			return 200;
 
-		case 'core/image':
```
miina: Removed this since it looks like `getDefaultMinimumBlockHeight` is not called for image blocks anyway.

assets/src/block-editor/helpers/index.js
```diff
@@ -112,7 +112,7 @@ export const addAMPAttributes = ( settings, name ) => {
 	const isTextBlock = TEXT_BLOCKS.includes( name );
 
 	// Fit-text for text blocks.
-	if ( ! isTextBlock ) {
+	if ( isTextBlock ) {
```
miina: These attributes are for text blocks but for some reason it was with `!` sign.

assets/src/stories-editor/helpers/index.js
swissspidy: @miina How does that change to `assets/src/block-editor/helpers/index.js` affect the stories editor?

assets/src/block-editor/helpers/index.js
swissspidy: @miina How does that change to `assets/src/block-editor/helpers/index.js` affect the stories editor?

miina: Block editor loads for both, stories and the default editor.

swissspidy: I don't think it should... 😬 Let me look into that.

miina: I mean -- the block editor files load for both. For example, these controls are defined in the block editor files but load for stories, too.

Screen Shot 2019-05-16 at 5 52 16 PM

miina: You're right, it probably shouldn't, most of these features wouldn't work in AMP Stories anyway, e.g. layout and lightbox effect in Image block's case. Probably should create a ticket for reviewing which settings should be for the block editor only.

swissspidy: Ah, gotcha. These controls come from `amp-editor-blocks.js`, which loads for both screens.

For reference, `amp-block-editor.js` only loads on the regular block editor, not for stories. `amp-stories.js` only loads for the stories editor. 

The naming is a bit confusing, so I thought with "Block editor loads for both"  you meant `amp-block-editor.js`

swissspidy: Created #2333.

PR #2337, opened 2019-05-16
Improve styling for CTA warnings

### LINKED ISSUE #2294: Call to Action validation message cut off
If you happen to have a CTA block on the first page (which is invalid), you'll get a nice little error message in the editor.

However, due to some `overflow:hidden`, this message is currently partially cut off:

Screenshot 2019-05-13 at 09 14 01

swissspidy:
Fixes #2294.

Before:

Screenshot 2019-05-13 at 09 14 01

After:

Screenshot 2019-05-22 at 14 41 27


miina:
Right now when the second page is selected and the warning message is on the first page then the left arrow to navigate to the previous page doesn't seem to be working. If the arrow is not expected to work, perhaps it should be hidden under the warning message?

Screen Shot 2019-05-22 at 2 51 41 PM

miina:
Actually, I just realized that generally clicking on anywhere of the carousel navigation isn't working, probably due to `z-index: -1;`?

swissspidy:
Seems like that `-1` setting was a bit optimistic 🙃

swissspidy:
Should be improved now.

PR #2339, opened 2019-05-17
Use amp-embedly-card instead of amp-reddit to work around resizing issue

mehigh:
Before:
```html
<amp-reddit
    layout="responsive"
    data-embedtype="post"
    width="100"
    height="100" 
    data-src="https://www.reddit.com/r/me_irl/comments/52rmir/me_irl/?ref=share&amp;ref_source=embed">
</amp-reddit>
```

After:
```html
<amp-embedly-card
    data-url="https://www.reddit.com/r/me_irl/comments/52rmir/me_irl/"
    layout="responsive"
    width="100"
    height="100">
</amp-embedly-card>
```

The amp-reddit component is plagued by an issue where users have to fill in the width/height 
https://amp.dev/documentation/examples/components/amp-reddit/?referrer=ampbyexample.com
Because of that standard embeds with 

Background:
Around the area I've touched there was a todo item which references a closed ticket (although amp-reddit still is not sizing correctly):

> `@todo Sizing is not yet correct. See <https://github.com/ampproject/amphtml/issues/11869>.`

I've found there a reference to the amp-embedly-card which is functioning correctly in sizing the embed.

See also https://github.com/ampproject/amphtml/issues/11869#issuecomment-493506507

### DIFF
--- a/includes/embeds/class-amp-reddit-embed-handler.php
+++ b/includes/embeds/class-amp-reddit-embed-handler.php
@@ -61,15 +61,13 @@ class AMP_Reddit_Embed_Handler extends AMP_Base_Embed_Handler {
 			return '';
 		}
 
-		// @todo Sizing is not yet correct. See <https://github.com/ampproject/amphtml/issues/11869>.
 		return AMP_HTML_Utils::build_tag(
-			'amp-reddit',
+			'amp-embedly-card',
 			array(
-				'layout'         => 'responsive',
-				'data-embedtype' => 'post',
-				'width'          => '100',
-				'height'         => '100',
-				'data-src'       => $args['url'],
+				'layout'   => 'responsive',
+				'width'    => '100',
+				'height'   => '100',
+				'data-url' => $args['url'],
 			)
 		);
 	}


westonruter:
Could you please rebase this against `develop` and target that branch?

westonruter:
> Because of that standard embeds with

Seems something got cut off here in the description.

westonruter:
The behavior is much better as Embedly is sending this message to the parent window:

```
amp-011905091827220{"width":"600","height":707,"type":"embed-size","sentinel":"0-23875101463846451262"}
```

And that is allowing AMP to resize the `amp-embedly-card` as required. 

Reddit appears to not be sending this message directly.

westonruter:
Hold on. It appears that the `amp-reddit` component actually uses Embedly.

westonruter:
There seems to be a bug with `amp-embedly` here which can be fixed: https://github.com/ampproject/amphtml/issues/11869#issuecomment-494114923

mehigh:
Thanks for following up on this, @westonruter . 
Should I still rebase this to be against develop, or just close this awaiting for this to be fixed in AMP core?

westonruter:
@mehigh I am a bit torn. I suppose it's good to go with `amp-embedly-card` now to fix it in the immediate term, and we can always switch back to `amp-reddit` later if needed, once it is fixed.

One thing I wasn't sure about is the `amp-embedly-card`:

image

But I don't see any branding?

mehigh:
Maybe the embedly branding isn't shown for their premiers embed partners. Embedly supports many type of embeds, and some do show the embedly branding. Reddit happens not to show one as I've even read from a message written by a Reddit representative that they're using embedly behind the scenes - and that's how I arrived at this patch to go straight to the source and cut the middle-man in the process (the middle man being the amp-reddit which only wraps the embedly with another iframe).

westonruter:
@mehigh Then I'd say go ahead with your patch!

mehigh:
I've rebased this to develop and adjusted the PR to point to that branch instead. Thank you for the replies and encouragement!

PR #2340, opened 2019-05-17
Resizable iframes

### LINKED ISSUE #2301: Use Iframes for Story Embeds
For various reasons, WordPress uses iframes for embedding content from other WordPress sites (including its own content).

When embedding an AMP story, `\AMP_Story_Post_Type::override_story_embed_callback` completely turns this on its head to replace the iframe with the direct HTML output from within that iframe. From the docblock:

> WordPress post embeds are usually wrapped in an <iframe>, which can cause validation and display issues in AMP. This overrides the embed callback in that case, replacing the <iframe> with the simple AMP story card.

I would like to investigate which validation and display issues there are and find ways to get rid of them so that the embeds still work in iframes.

This would allow for a more coherent experience when embedding stories, no matter if it's your own site embedding them or someone else. Having the same content everywhere makes development and testing much easier, too.

For example, one thing that bugged me in #2245 is the use of a magic number to increase the height of the embed card. WordPress iframe embeds already have functionality in place to demand resizing via postMessage. `<amp-iframe>` support that too, just using different names. So why not use that instead?

* [ ] Document exact validation and display issues
* [ ] Use iframe everywhere
* [ ] Investigate whether logic from `\AMP_Story_Post_Type::remove_title_from_embed` can be moved to `oembed_result` / `pre_oembed_result` instead, or if `amp-script` can be used to remove that.
* [ ] In the embedded template, send AMP compatible resize postMessage requests (`embed-ready` and `embed-size`)  
  These could be something for WordPress embeds in general, FWIW
* [ ] Extend iframe sanitizer to support [resizable iframes](https://github.com/ampproject/amphtml/blob/master/extensions/amp-iframe/amp-iframe.md)

swissspidy:
Even though the AMP plugin supports [many different embed providers](https://amp-wp.org/documentation/how-the-plugin-works/amp-content-rendering-in-wordpress/native-wordpress-widget-support/), WordPress does not appear to be one of them.

This PR tries to improve this, so that Story embeds can also use iframes.

Fixes #2301.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ assets/css/*-compiled-rtl.css
 assets/css/*.map
 assets/js/*.js
 !assets/js/amp-service-worker-runtime-precaching.js
+!assets/js/wp-embed-template.js
 assets/js/*.map
 built
 /amphtml
--- a/amp.php
+++ b/amp.php
@@ -348,6 +348,7 @@ function amp_init() {
 	add_action( 'wp_loaded', 'amp_story_templates' );
 	add_action( 'wp_loaded', 'amp_add_options_menu' );
 	add_action( 'wp_loaded', 'amp_admin_pointer' );
+	add_action( 'wp_loaded', 'amp_wordpress_embed_templates' );
 	add_action( 'parse_query', 'amp_correct_query_when_is_front_page' );
 	add_action( 'admin_bar_menu', 'amp_add_admin_bar_view_link', 100 );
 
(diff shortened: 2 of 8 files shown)


westonruter:
@swissspidy If an `amp-iframe` is trying to load a URL on the same-origin as the host page, it will fail. See https://github.com/ampproject/amphtml/blob/master/spec/amp-iframe-origin-policy.md

See the §Resizing `iframe` Problem on https://github.com/ampproject/amp-wp/issues/809#issuecomment-424151058

swissspidy:
Somehow I didn't find #914, #809, nor https://github.com/ampproject/amphtml/issues/18378 🤦‍♂ 

Glad to hear this has been previously been brought up.

swissspidy:
> If an amp-iframe is trying to load a URL on the same-origin as the host page, it will fail.

Nice that #2712 is aiming to fix that.

swissspidy:
Closing for now...

PR #2343, opened 2019-05-17
Keep content in skeleton templates

### LINKED ISSUE #2307: Improve skeleton template inserter

swissspidy:
We've discussed today that the difference between skeletons and template previews is too big when there's really nothing left from the original preview. Especially because there are some drastic changes in font sizes with amp-fit-text and empty blocks.

This change addresses this by still keeping the text from the templates when inserting the skeleton templates.

Todo:

* [x] Fix font size calculation after insertion

Fixes #2307.


swissspidy:
@miina Any thoughts on the font size calculation upon inserting a template? `maybeUpdateFontSize` should already run on subscribe, but it doesn't seem to fully work yet last I checked.

miina:
@swissspidy The calculation seems to be working fine, it looks like the selector didn't get the right element to measure the font size correctly, specified the selector and it seems to be working now. Locally at least. Could you test if it seems to be better now?

swissspidy:
@miina Still seems to be the same for me. Insert template -> click on text block -> notice that font size changes. Do you also experience that?

miina:
> Still seems to be the same for me. Insert template -> click on text block -> notice that font size changes. Do you also experience that?

Hmm, I am seeing it now, I was debugging before, line by line and then it worked as expected. Without debugging it does not work as expected and changes the font to incorrect at some point.

It's weird because logging out the blocks and the final font sizes then it shows the correct font sizes as the last logged out sizes. Also, clicking on the block doesn't seem to trigger a new font size calculation, so, not sure why it changes when clicking on the block.

I'll try to debug this a bit more.

assets/src/stories-editor/index.js
```diff
@@ -201,6 +202,7 @@ subscribe( async () => {
 		maybeUpdateAutoAdvanceAfterMedia( block );
 		maybeSetTagName( block );
 		maybeSetInitialSize( block );
+		maybeUpdateFontSize( getBlock( block ) );
```
miina: I'm wondering if there's a way to use it only after inserting the template instead and only for the newly added template blocks? Otherwise, it will run for all the blocks with every change, every time each block element will be changed in size to check for the font, etc. Thoughts? 

This isn't related to the incorrect font size though.

swissspidy: The `insertBlock` action (called by the template inserter) is async, so you cannot directly call something right afterwards.

We could, however, update our own store after `insertBlock` is called to set an internal state variable ourselves. In the subscribe handler we would then only do something if that variable is set.

some pseudo code off the top of my head:

```
insertBlock(...)
dispatch('amp/stories').setHasTemplateBeenInserted( pageClientId )
...

if ( select('amp/stories').hasTemplateBeenInserted() ) {
pageClientId.innerBlocks.foreach(updateFontSize)
dispatch('amp/stories').clearHasTemplateBeenInserted( pageClientId )
}

Comment (miina) [reply assets/src/stories-editor/index.js]: I was thinking that one option could be using `await` for `insertBlocks`.

Better to use a state though.

Comment (swissspidy) [reply assets/src/stories-editor/index.js]: I don‘t think that works, IIRC it‘s not a Promise. It just dispatches an action and returns immediately. Redux in the background then handles the action

Comment (miina) [reply assets/src/stories-editor/index.js]: I actually tried meanwhile and it does seem to work, however, not sure that's the best approach anyway.

Basically adding this instead of using `insertBlock` directly. 
```
async insertTemplate( template ) {
		const { insertBlock, selectBlock } = this.props;
		await insertBlock( template );
	}
```


However, I'm thinking that maybe we could instead add the font update to `componentDidMount` of the Text block:
```
componentDidMount() {
		maybeUpdateFontSize( this.props );
	}
```
Tested very briefly that this seems to fix the font size issue as well. _Testing more thoroughly_

miina: Nope, it does not resolve that issue, it displayed correctly with just one of the templates, no clue why.

miina:
@swissspidy Could you test again if it works better now (right now for the Text block only)?

swissspidy:
@miina Seems to work fine! If you can update it for the other blocks too then I think we should be good.

miina:
@swissspidy Could you check again? I can't request for your review via GitHub since you're the author of the PR.

swissspidy:
That seems to work well, thanks @miina!

I think we are good to go with this.

Issue #1921, opened 2019-03-02
Ignore admin bar for AMP validation purposes


westonruter:
The inclusion of the admin bar was included in AMP via #1205. However, the amount of admin bar CSS is so large (20KB) that it takes up about 40% of the total allowed CSS in AMP (50KB). For this reason it very frequently causes `excessive_css` validation errors when an authenticated user is browsing the frontend. As opposed to providing this option or trying to automatically disable the admin bar when there is too much CSS (#1800), we could instead just exclude admin bar CSS (including any styles that depend on `admin-bar`) from being considered in the 50KB budget. ~This can mean that an authenticated user will often be served an invalid AMP page, but this is of little importance since the unauthenticated visitor (including crawlers) will only get the valid AMP pages for indexing. The admin bar has an indicator of AMP validity already, so users who have the AMP validator browser extension would just need to trust the admin bar status rather than the browser toolbar, or else open the URL in an incognito window to double-check.~ AMP now supports an explicit “dev mode” (https://github.com/ampproject/amphtml/issues/20974) for when invalid markup is intentionally being included in the page. The AMP validator browser extension will explicitly be aware of this mode (https://github.com/ampproject/amphtml/issues/24176) so there will not be concerns about it alerting the user of errors. Initial support for dev mode in the AMP plugin landed with https://github.com/ampproject/amp-wp/pull/3084.

👉 For full list of changes, please refer to description in PR: https://github.com/ampproject/amp-wp/pull/3187.

# Done/Obsolete

- [x] Remove checkbox from settings screen to hide admin bar in AMP. If anyone wants to do this, they can easily do so via a `show_admin_bar` filter along with a check for `! is_amp_endpoint()`. Most users will benefit from the admin bar. Decisions not options. (Done in #2346 & #2405.)
- [x] Suppress admin bar from being shown during validation requests. Done in #2500.
- [ ] ~Add note in admin bar submenu for AMP that page is not valid due to admin bar, but otherwise it is OK.~
- [ ] ~Add link to AMP admin bar menu to navigate to the current URL without the admin bar. A query param like `hide_admin_bar` can be introduced. This URL can also automatically add `#development=1`.~
- [ ] ~Consider removing the `amp` attribute from the `html` element when the admin bar is showing to prevent the AMP Validator extension from screaming at the user.~

swissspidy:
Sounds like a plan!

westonruter:
For perhaps a better alternative to this, see #2322.

[...]

Issue #1801, opened 2019-01-08
Show notice on frontend when there is excessive CSS


westonruter:
In addition to automatically disabling the admin bar when there is excessive CSS (#1800), there should be an additional message displayed to the logged-in site administrator on the frontend when there is still excessive CSS. At the moment the ways for a user to know that there is excessive CSS are:

* Notice broken styling.
* See an AMP validation error when saving a post.
* Look at the page source for the `amp-custom` style manifest to see the excessive CSS message.

In addition to these, the plugin could display a message about the excessive CSS along with some suggestions for how to fix. For developers, fixing may involve rewriting CSS. For implementors, fixing may involving switching to a different theme or using alternative plugins. For other users, the fix may be to switch back to Classic/Legacy mode until the ecosystem better supports AMP.

The admin bar would be a good location to show this message:

![image](https://user-images.githubusercontent.com/134745/57810367-72c7d800-771c-11e9-8e3e-fdcb3e814b1d.png)

westonruter:
It would also be useful to show a `progress` meter indicating the percentage of how much of the CSS budget they've used on the current page. Clicking the progress bar can then take the user to view the CSS manifest in the Validated URL screen (#2169).

pierlon:
Would the progress meter be a top-level menu item in the admin bar?

schlessera:
In my opinion, it should be a submenu under the main AMP top-level entry. Otherwise, we'll take up too much space to show a metric that will often be irrelevant.

westonruter:
> It would also be useful to show a `<meter>` element indicating the percentage of how much of the CSS budget they've used on the current page. Clicking the progress bar can then take the user to view the CSS manifest in the Validated URL screen (#2169).

I think `optimum="50" low="75" high="99"` attributes work well for `meter` works:

```html
<meter min="0" optimum="50" low="75" high="100" max="101" value="50">50%</meter>
<meter min="0" optimum="50" low="75" high="100" max="101" value="80">80%</meter>
<meter min="0" optimum="50" low="75" high="100" max="101" value="101">100%</meter>
```

![image](https://user-images.githubusercontent.com/134745/71953469-28a5d200-3197-11ea-856d-81ebe2915768.png)

* If the CSS on the page is using 75% or less of the 50KB budget used, then the color is green.
* If more than 75% is used, then the color is yellow.
* If more than 100% is used, then the color is red.

westonruter:
While I like the idea of the `meter` element:

[...]

PR #2346, opened 2019-05-18
Introduce stylesheet prioritization when determining which to concatenate

### LINKED ISSUE #2322: Prioritize stylesheets to decide which are concatenated and which are excluded
While stylesheets are being concatenated together for `style[amp-custom]`, as soon as the the total reaches 50KB then all subsequent stylesheets are excluded. Each excluded stylesheet raises an `excessive_css` validation error.

This is wrong, and it leads to a very bad user experiences. Case in point is when the admin bar is being displayed on  AMP pages, as the CSS for dashicons and admin bar get added _before_ the CSS for the theme's stylesheet. Since the `admin-bar.css` is huge, this often results in the theme's own `style.css` being excluded, leading to completely broken templates. This is a big reason for why this admin-bar toggle exists in the first place:

image

For example, on Twenty Nineteen with Jetpack running and the admin bar shown, the scripts added are:

```html
<link rel='stylesheet' id='twentynineteen-jetpack-css'  href='https://wordpressdev.lndo.site/content/plugins/jetpack/modules/theme-tools/compat/twentynineteen.css?ver=1557447387' type='text/css' media='all' />
<link rel='stylesheet' id='dashicons-css'  href='https://wordpressdev.lndo.site/core-dev/src/wp-includes/css/dashicons.css?ver=5.3-alpha-45282-src' type='text/css' media='all' />
<link rel='stylesheet' id='admin-bar-css'  href='https://wordpressdev.lndo.site/core-dev/src/wp-includes/css/admin-bar.css?ver=5.3-alpha-45282-src' type='text/css' media='all' />
<link rel='stylesheet' id='wp-block-library-css'  href='https://wordpressdev.lndo.site/content/plugins/gutenberg/build/block-library/style.css?ver=1557873353' type='text/css' media='all' />
<link rel='stylesheet' id='wp-block-library-theme-css'  href='https://wordpressdev.lndo.site/content/plugins/gutenberg/build/block-library/theme.css?ver=1557873353' type='text/css' media='all' />
<link rel='stylesheet' id='twentynineteen-style-css'  href='https://wordpressdev.lndo.site/core-dev/src/wp-content/themes/twentynineteen/style.css?ver=1.4' type='text/css' media='all' />
<link rel='stylesheet' id='twentynineteen-print-style-css'  href='https://wordpressdev.lndo.site/core-dev/src/wp-content/themes/twentynineteen/print.css?ver=1.4' type='text/css' media='print' />
```

On an AMP page, the result is:

```
     0 B: style[amp-custom=]
   106 B (2%): link#twentynineteen-jetpack-css[rel=stylesheet][id=twentynineteen-jetpack-css][href=https://wordpressdev.lndo.site/content/plugins/jetpack/modules/theme-tools/compat/twentynineteen.css?ver=1557447387][type=text/css][media=all]
   780 B (14%): link#dashicons-css[rel=stylesheet][id=dashicons-css][href=https://wordpressdev.lndo.site/core-dev/src/wp-includes/css/dashicons.css?ver=5.3-alpha-45282-src][type=text/css][media=all]
 17336 B (96%): link#admin-bar-css[rel=stylesheet][id=admin-bar-css][href=https://wordpressdev.lndo.site/content/plugins/amp/assets/css/admin-bar.css?ver=1.2-alpha1][type=text/css][media=all]
    99 B (46%): style#admin-bar-inline-css[id=admin-bar-inline-css][type=text/css]
    73 B (57%): link#amp-default-css[rel=stylesheet][id=amp-default-css][href=https://wordpressdev.lndo.site/content/plugins/amp/assets/css/amp-default.css?ver=1.2-alpha1][type=text/css][media=all]
  2168 B (18%): link#wp-block-library-css[rel=stylesheet][id=wp-block-library-css][href=https://wordpressdev.lndo.site/content/plugins/gutenberg/build/block-library/style.css?ver=1557873353][type=text/css][media=all]
   249 B (19%): link#wp-block-library-theme-css[rel=stylesheet][id=wp-block-library-theme-css][href=https://wordpressdev.lndo.site/content/plugins/gutenberg/build/block-library/theme.css?ver=1557873353][type=text/css][media=all]
  1116 B (80%): link#twentynineteen-print-style-css[rel=stylesheet][id=twentynineteen-print-style-css][href=https://wordpressdev.lndo.site/core-dev/src/wp-content/themes/twentynineteen/print.css?ver=1.4][type=text/css][media=print]
   122 B: style[type=text/css]
    39 B: style[type=text/css][media=print]
   225 B: style[type=text/css][media=screen]
    73 B: amp-img.wp-image-53 amp-wp-enforced-sizes amp-wp-b8332d9[class=wp-image-53 amp-wp-enforced-sizes amp-wp-b8332d9][src=https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1.jpg][alt=][srcset=https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1.jpg 1200w, https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1-300x196.jpg 300w, https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1-768x501.jpg 768w, https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1-1024x668.jpg 1024w][width=1200][height=783][layout=intrinsic]
    93 B: div.wp-block-media-text has-media-on-the-right is-stacked-on-mobile amp-wp-2d949e1[class=wp-block-media-text has-media-on-the-right is-stacked-on-mobile amp-wp-2d949e1]
   231 B: figure.wp-block-media-text__media amp-wp-6c8fbd0[class=wp-block-media-text__media amp-wp-6c8fbd0]
    93 B: div.wp-block-media-text alignwide amp-wp-2d42cb8[class=wp-block-media-text alignwide amp-wp-2d42cb8]
    74 B: figure#attachment_53.wp-caption alignright amp-wp-d9900c6[id=attachment_53][aria-describedby=caption-attachment-53][class=wp-caption alignright amp-wp-d9900c6]
    74 B: figure#attachment_53.wp-caption aligncenter amp-wp-343bce0[id=attachment_53][aria-describedby=caption-attachment-53][class=wp-caption aligncenter amp-wp-343bce0]
Total included size: 22,951 bytes (54% of 42,385 total after tree shaking)

The following stylesheets are too large to be included in style[amp-custom]:
 29500 B (58%): link#twentynineteen-style-css[rel=stylesheet][id=twentynineteen-style-css][href=https://wordpressdev.lndo.site/core-dev/src/wp-content/themes/twentynineteen/style.css?ver=1.4][type=text/css][media=all]
Total excluded size: 29,500 bytes (58% of 50,020 total after tree shaking)

Total combined size: 52,451 bytes (56% of 92,405 total after tree shaking)
```

The most important stylesheet of all is the one being excluded!

> image

At least the admin bar looks good! 😕 

**So what is needed is to introduce a sense of priority weights for the stylesheets as they are selected for concatenation.**

Given the stylesheets loaded above:

* https://wordpressdev.lndo.site/content/plugins/jetpack/modules/theme-tools/compat/twentynineteen.css 
* https://wordpressdev.lndo.site/core-dev/src/wp-includes/css/dashicons.css 
* https://wordpressdev.lndo.site/core-dev/src/wp-includes/css/admin-bar.css 
* https://wordpressdev.lndo.site/content/plugins/gutenberg/build/block-library/style.css 
* https://wordpressdev.lndo.site/content/plugins/gutenberg/build/block-library/theme.css 
* https://wordpressdev.lndo.site/core-dev/src/wp-content/themes/twentynineteen/style.css 
* https://wordpressdev.lndo.site/core-dev/src/wp-content/themes/twentynineteen/print.css 

The priority order should be something like:

1. https://wordpressdev.lndo.site/core-dev/src/wp-content/themes/twentynineteen/style.css
1. https://wordpressdev.lndo.site/content/plugins/gutenberg/build/block-library/style.css
1. https://wordpressdev.lndo.site/content/plugins/gutenberg/build/block-library/theme.css
1. https://wordpressdev.lndo.site/content/plugins/jetpack/modules/theme-tools/compat/twentynineteen.css
1. https://wordpressdev.lndo.site/core-dev/src/wp-includes/css/dashicons.css
1. https://wordpressdev.lndo.site/core-dev/src/wp-includes/css/admin-bar.css
1. https://wordpressdev.lndo.site/core-dev/src/wp-content/themes/twentynineteen/print.css

So the prioritized categories:

1. Parent theme stylesheets (e.g. `style.css`) which have `all`/`screen` media
1. Child theme stylesheets (e.g. `style.css`) which have `all`/`screen` media
1. Core stylesheets used by themes (e.g. `wp-block-library` and `wp-block-library-theme`), including the ones included by Gutenberg
1. Stylesheets from plugins (with `all`/`screen` media)
1. Inline `style` elements and attributes
1. Stylesheet for Dashicons
1. Stylesheet for admin-bar
1. Stylesheets with `print` media

This should greatly reduce the frequency of templates appearing completely broken when CSS is being excluded.

The prioritization could be indicated when displaying the manifest of the stylesheets in `style[amp-custom]` (including on validated URL admin screen: #2169).

Bonus: once we get to the admin bar CSS and we determine it cannot be loaded, then we can automatically remove the admin bar from the page. (Which is an alternative to #1921.)

westonruter:
Fixes #2322.

A theme's stylesheets should never be excluded in favor of the admin bar CSS. Similarly, print stylesheets should never be included at the expense of plugin stylesheets. This PR introduces prioritization when determining which stylesheets to include and which to exclude when the tree shaker is not able to reduce the total CSS to the 50KB limit. The changes here should result in a drastic reduction in the frequency of a theme appearing completely broken when viewing in Native/Transitional modes.

The stylesheet priorities are:

1. Parent theme (non-print) stylesheets
1. Child theme (non-print) stylesheets
1. Core stylesheets used by themes (for blocks)
1. Plugin (non-print) stylesheets
1. Stylesheets from `wp-includes`
1. Additional CSS from Customizer
1. Styles added by blocks and widgets
1. Dashicons
1. Print stylesheets
1. Admin bar CSS

If it so happens that the Admin Bar CSS is excluded, then the HTML element for the admin bar as well as its associated CSS are both excluded.

Since the admin bar is automatically excluded, there is now no need for the option and it is removed:

image

# TODO

- [x] Add additional tests for prioritization.
- [x] Remove admin bar when either admin-bar-css or admin-bar-inline-css is excluded.

# Other items for follow-up

- Add indication that admin bar is hidden?
- How to indicate that CSS is truncated when admin bar is hidden?
- Ensure admin bar print style is added as inline.
- Include any CSS stylesheet which has admin-bar as dependency?
- Or skip sanitizing the admin bar at all? Exempt processing of any stylesheet that targets admin bar selectors.
- Hide admin bar when doing Validation request.
- What to do about dashicons.
- Tree Shaker must exempt admin-bar.css
- Omit dashicons from tree Shaker and processing if there are no dashicon classes used in the document?
- Remove admin bar before sanitizers are run, and the restore after all run. Add dirty styles rule for admin bar CSS when authenticated.

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1299,6 +1299,24 @@ class AMP_Theme_Support {
 			41
 		);
 
+		// Override default admin bar bump to add element ID and remove obsolete IE hacks.
+		if ( has_action( 'wp_head', '_admin_bar_bump_cb' ) ) {
+			remove_action( 'wp_head', '_admin_bar_bump_cb' );
+			add_action(
+				'wp_head',
+				function () {
+					?>
+					<style id="admin-bar-bump-css" type="text/css" media="screen">
+						html { margin-top: 32px !important; }
+						@media screen and ( max-width: 782px ) {
+							html { margin-top: 46px !important; }
+						}
+					</style>
+					<?php
+				}
+			);
+		}
+
 		// Emulate customize support script in PHP, to assume Customizer.
 		add_filter(
 			'body_class',
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -931,22 +931,13 @@ class AMP_Theme_Support {
 			PHP_INT_MAX
 		);
 
+		add_action( 'admin_bar_init', array( __CLASS__, 'init_admin_bar' ) );
 		add_action( 'wp_head', 'amp_add_generator_metadata', 20 );
 		add_action( 'wp_enqueue_scripts', array( __CLASS__, 'enqueue_assets' ), 0 ); // Enqueue before theme's styles.
 		add_action( 'wp_enqueue_scripts', array( __CLASS__, 'dequeue_customize_preview_scripts' ), 1000 );
 		add_filter( 'customize_partial_render', array( __CLASS__, 'filter_customize_partial_render' ) );
 
 		add_action( 'wp_footer', 'amp_print_analytics' );
-
-		/*
-		 * Disable admin bar because admin-bar.css (28K) and Dashicons (48K) alone
-		 * combine to surpass the 50K limit imposed for the amp-custom style.
-		 */
-		if ( AMP_Options_Manager::get_option( 'disable_admin_bar' ) ) {
-			add_filter( 'show_admin_bar', '__return_false', 100 );
-		} else {
-			add_action( 'admin_bar_init', array( __CLASS__, 'init_admin_bar' ) );
-		}
 
 		/*
 		 * Start output buffering at very low priority for sake of plugins and themes that use template_redirect
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1299,22 +1299,20 @@ class AMP_Theme_Support {
 			41
 		);
 
-		// Override default admin bar bump to add element ID and remove obsolete IE hacks.
-		if ( has_action( 'wp_head', '_admin_bar_bump_cb' ) ) {
-			remove_action( 'wp_head', '_admin_bar_bump_cb' );
-			add_action(
-				'wp_head',
-				function () {
-					?>
-					<style id="admin-bar-bump-css" type="text/css" media="screen">
-						html { margin-top: 32px !important; }
-						@media screen and ( max-width: 782px ) {
-							html { margin-top: 46px !important; }
-						}
-					</style>
-					<?php
-				}
-			);
+		// Convert admin bar bump callback into an inline style for admin-bar. See \WP_Admin_Bar::initialize().
+		if ( current_theme_supports( 'admin-bar' ) ) {
+			$admin_bar_args  = get_theme_support( 'admin-bar' );
+			$header_callback = $admin_bar_args[0]['callback'];
+		} else {
+			$header_callback = '_admin_bar_bump_cb';
+		}
+		remove_action( 'wp_head', $header_callback );
+		if ( '__return_false' !== $header_callback ) {
+			ob_start();
+			call_user_func( $header_callback );
+			$style = ob_get_clean();
+			$data  = trim( preg_replace( '#<style[^>]*>(.*)</style>#is', '$1', $style ) ); // See wp_add_inline_style().
+			wp_add_inline_style( 'admin-bar', $data );
 		}
 
 		// Emulate customize support script in PHP, to assume Customizer.
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -121,16 +121,6 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	);
 
 	/**
-	 * Stylesheets.
-	 *
-	 * Values are the CSS stylesheets. Keys are MD5 hashes of the stylesheets,
-	 *
-	 * @since 0.7
-	 * @var string[]
-	 */
-	private $stylesheets = array();
-
-	/**
 	 * List of stylesheet parts prior to selector/rule removal (tree shaking).
 	 *
 	 * Keys are MD5 hashes of stylesheets.
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -2489,7 +2489,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 					$message .= '.' . $pending_stylesheet['node']->getAttribute( 'class' );
 				}
 				foreach ( $pending_stylesheet['node']->attributes as $attribute ) {
-					if ( 'id' !== $attribute->nodeName || 'class' !== $attribute->nodeName ) {
+					if ( 'id' !== $attribute->nodeName && 'class' !== $attribute->nodeName ) {
 						$message .= sprintf( '[%s=%s]', $attribute->nodeName, $attribute->nodeValue );
 					}
 				}
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -2969,7 +2969,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 		$pending_stylesheet_indices = array_keys( $this->pending_stylesheets );
 		usort(
 			$pending_stylesheet_indices,
-			function( $a, $b ) use ( $group_config ) {
+			function( $a, $b ) {
 				return $this->pending_stylesheets[ $a ]['priority'] - $this->pending_stylesheets[ $b ]['priority'];
 			}
 		);
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -854,7 +854,9 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			}
 		} elseif ( $node instanceof DOMElement && 'style' === $node->nodeName ) {
 			$element_id = (string) $node->getAttribute( 'id' );
-			if ( 'wp-custom-css' === $element_id ) {
+			if ( 'admin-bar-inline-css' === $element_id ) {
+				$priority = $admin_bar_priority;
+			} elseif ( 'wp-custom-css' === $element_id ) {
 				// Additional CSS from Customizer.
 				$priority = 60;
 			} else {
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -404,13 +404,21 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	}
 
 	/**
-	 * Get stylesheets.
+	 * Get stylesheets for amp-custom.
 	 *
 	 * @since 0.7
-	 * @returns array Values are the CSS stylesheets. Keys are MD5 hashes of the stylesheets.
+	 * @returns array Values are the CSS stylesheets.
 	 */
 	public function get_stylesheets() {
-		return $this->stylesheets;
+		return wp_list_pluck(
+			array_filter(
+				$this->pending_stylesheets,
+				function( $pending_stylesheet ) {
+					return $pending_stylesheet['included'] && 'custom' === $pending_stylesheet['group'];
+				}
+			),
+			'stylesheet'
+		);
 	}
 
 	/**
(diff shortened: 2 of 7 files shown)


postphotos:
Awesome! Excited to see this. It'll help a lot when working with excessive CSS issues.

westonruter:
I'm not sure what to do about these two points:

> * Add indication that admin bar is hidden?
> * How to indicate that CSS is truncated when admin bar is hidden?

There is an indication now but it is just an HTML comment, so it could still be disorienting to a user who expects the admin bar to be present. Otherwise, the way that a user could discover that the admin bar is being excluded by re-validating the URL via the compatibility tool:

image

On Native mode in particular, they could decide to _reject_ the validation error to cause the admin bar to appear, though with the warning as expected:

image

This would not be advisable in Transitional mode because the rejected state would then cause the `amphtml` link to be hidden. Perhaps this is all getting beyond the scope of this issue and should be addressed in #1921, #2326, and/or #1801.

westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3194793/amp.zip) (v1.2-alpha-20190519T052057Z-60e4e427)

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -2511,15 +2607,66 @@ private function finalize_styles() {
...
+			);
+			if ( $is_admin_bar_css ) {
+				$included = $pending_stylesheet['included'];
+				break;
+			}
+		}
+
+		if ( ! $included ) {
+			$admin_bar->parentNode->replaceChild(
+				$this->dom->createComment( ' ' . __( 'Admin bar (#wpadminbar) was removed to preserve AMP validity due to excessive CSS.', 'amp' ) . ' ' ),
```
swissspidy: This should use a placeholder for `#wpadminbar` and a proper translator comment.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -2767,44 +2923,102 @@ function( $id ) use ( $dom ) {
 					}
 				}
 			}
+
+			$pending_stylesheet['stylesheet']    = implode( '', $stylesheet_parts );
 			$pending_stylesheet['original_size'] = $original_size;
+			$pending_stylesheet['included']      = null; // To be determined below.
+			$pending_stylesheet['size']          = strlen( $pending_stylesheet['stylesheet'] ); // @todo Should this use mb_strlen()?
```
swissspidy: The mbstring extension might not always be available. We would need to add a requirement to `amp.php`

swissspidy:
This is looking sharp! I am not too familiar with the sanitizers yet, especially the one for stylesheets. That's why I cannot give detailed feedback on the general approach and the todos you added. But after going through the code and testing it with a few themes, this seems to work as expected.

Happy to give it a more thorough spin tomorrow though. Alternatively, I imagine that @kienstra might know this area better than I and could also provide some more valuable feedback? Correct me if I'm wrong 🙂 

> Perhaps this is all getting beyond the scope of this issue and should be addressed in #1921, #2326, and/or #1801.

Agreed, let's take this step by step.

amedina:
Regarding:

> Add indication that admin bar is hidden?
> How to indicate that CSS is truncated when admin bar is hidden?

Could we show a "watered down" admin bar? Something along the lines of an empty Admin bar with a placeholder text indicating the truncation.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: Fixed in 08356df89c13ef4fc33bf3e7c6c06c517414772e.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: Yeah, I clarified the comment to indicate that this is something to explore in the future. There are other instances of `strlen()` being used above.

Part of it is we need to clarify if the AMP validator uses _characters_ or bytes when calculating the limit. Using byte count is safer in this way. Actually, it would seem to be the latter since the spec has `max_bytes` not `max_chars`. So I think I just answered my own question 😄

westonruter:
> Could we show a "watered down" admin bar? Something along the lines of an empty Admin bar with a placeholder text indicating the truncation.

If we did this we'd still need to include the CSS for the admin bar, at least a subset of it. And the `excessive_css` is the problem in the first place. For the best user experience, I think the admin bar should be “invisible” for purposes of validation (as noted in #1921) since the admin bar would not appear for any visitors. It doesn't make sense to raise an `excessive_css` validation error for it.

I was giving it some more thought last night about what do do and I'm thinking this: if the user is authenticated, we should skip `#admin-bar-css` and `#admin-bar-inline-css` from being included in the total CSS count. When we do this, we must still keep track of the fact that we skipped this, as we should probably make a comment in the AMP admin bar explaining the validation is only due to the admin bar and it would not be seen by Google. Also, the total byte size of the admin bar CSS could should be added to the `total_bytes` for the custom CSS so that the final validating sanitizer doesn't strip them out.

I guess this is just re-hashing #1921.

westonruter:
Rebased with changes from `develop`.

amedina:
> If we did this we'd still need to include the CSS for the admin bar, at least a subset of it. And the excessive_css is the problem in the first place.

Comment was related to "how to provide an indicator" that the admin bar is hidden.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -768,6 +766,105 @@ public function sanitize() {
...
+	 * The higher the priority value, the more likely it will be that the
+	 * stylesheet will be among those excluded due to 'excessive_css' when
+	 * concatenated CSS reaches 50KB.
+	 *
+	 * @todo This will eventually need to be abstracted to not be CMS-specific, allowing for the prioritization scheme to be defined by configuration.
+	 *
+	 * @param DOMNode|DOMElement|DOMAttr $node Node.
+	 * @return int Priority.
+	 */
+	private function get_stylesheet_priority( DOMNode $node ) {
```
westonruter: @swissspidy getting your gut feeling about the prioritization order in this method would be helpful.

includes/class-amp-theme-support.php
```diff
@@ -931,23 +931,14 @@ function() {
...
 		add_action( 'wp_enqueue_scripts', array( __CLASS__, 'dequeue_customize_preview_scripts' ), 1000 );
 		add_filter( 'customize_partial_render', array( __CLASS__, 'filter_customize_partial_render' ) );
 
 		add_action( 'wp_footer', 'amp_print_analytics' );
 
-		/*
-		 * Disable admin bar because admin-bar.css (28K) and Dashicons (48K) alone
-		 * combine to surpass the 50K limit imposed for the amp-custom style.
-		 */
-		if ( AMP_Options_Manager::get_option( 'disable_admin_bar' ) ) {
```
kienstra: Nice, it'll be simpler now that users don't have to choose this option.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -768,6 +766,105 @@ public function sanitize() {
...
+				$priority = 20;
+			} elseif ( $is_plugin_asset ) {
+				// Styles from plugins are next-highest priority.
+				$priority = 30;
+			} elseif ( 0 === strpos( $schemeless_href, $remove_url_scheme( includes_url() ) ) ) {
+				// Other styles from wp-includes come next.
+				$priority = 40;
+			} else {
+				// Everything else, perhaps wp-admin styles or stylesheets from remote servers.
+				$priority = 50;
```
kienstra: This is a really nice improvement. In practice, the theme's stylesheet has been stripped before, in favor of less important stylesheets like `admin.css`. Getting the styling right is a big step in convincing people that they can run AMP.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -768,6 +766,105 @@ public function sanitize() {
...
+				$priority = $admin_bar_priority;
+			} elseif ( 'wp-custom-css' === $element_id ) {
+				// Additional CSS from Customizer.
+				$priority = 60;
+			} else {
+				// Other style elements, including from Recent Comments widget.
+				$priority = 70;
+			}
+
+			if ( 'print' === $node->getAttribute( 'media' ) ) {
```
swissspidy: The `media` attribute in `<link>` element can be either a single media type, a white-space-separated list of media types, or a complex media query.

Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#Attributes

So this simple equality check might not be enough. Although the more complex `media` declarations might not me very common, so perhaps it's not a big deal.

includes/sanitizers/class-amp-style-sanitizer.php
swissspidy: The logic in general seems to make sense, and the order is reasonable I think.

I was wondering if inline `style=""` attributes should get a higher priority (like they do in the browser). Just think about Gutenberg blocks with custom background colors and stuff. Those shouldn't be stripped out.

As for the plugin/theme detection, I remember that quite some logic was needed in `load_script_textdomain()` for a similar issue because of CDNs, different domains, etc.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: Good point. In the case of a non-scalar `media` then the stylesheet would just be handled as a `screen` priority, which isn't the worst case. Since a simple media value is going to be the vast majority of cases, I don't think we need to worry about it here. But it's good to keep in mind.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: I think the theme/plugin detection will work as-is here because it's using `plugins_url()`, `get_template_directory_uri()`/`get_stylesheet_directory_uri()`. We don't need as much granularity here.

westonruter: > I was wondering if inline `style=""` attributes should get a higher priority (like they do in the browser). Just think about Gutenberg blocks with custom background colors and stuff. Those shouldn't be stripped out.

That's a good question. I think the `link` stylesheets (e.g. `wp-block-library`) should for sure be higher priority because they have the defaults which the `style` attributes in general customize. But otherwise should a `style` element have more priority than `style` attribute?  Would those styles be more important than Additional CSS?

westonruter:
@swissspidy The remaining todos in the description are not for tackling in this PR.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -2772,44 +2934,102 @@ function( $id ) use ( $dom ) {
...
-			if ( isset( $stylesheet_set['final_stylesheets'][ $hash ] ) ) {
-				$pending_stylesheet['included']  = true;
-				$pending_stylesheet['duplicate'] = true;
+		}
+
+		// Determine which stylesheets are included based on their priorities.
+		$pending_stylesheet_indices = array_keys( $this->pending_stylesheets );
+		usort(
+			$pending_stylesheet_indices,
+			function( $a, $b ) use ( $group_config ) {
```
swissspidy: Unused `$group_config` variable?

includes/sanitizers/class-amp-style-sanitizer.php
swissspidy: So right now Additional CSS has priority of 60, and style elements and attributes both have one of 70.

I'd assume that all the CSS from `style` attributes is usually way less than from `style` elements, so making the attributes have a bit of a higher priority probably wouldn't cause any problems.

But perhaps the current solution is good enough for now. 🤷‍♂

swissspidy:
> The remaining todos in the description are not for tackling in this PR.

What about the remaining todos in the code?

westonruter:
> What about the remaining todos in the code?

There are three and they are largely for future reference, so they don't need (or can't be) done yet.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: Yeah, let's get this out for testing and we can see some real world results.

PR #2352, opened 2019-05-20
Fix dragging after resizing / rotation

### LINKED ISSUE #2306: Dragging not working after resizing/rotating

miina:
Fixes #2306.

Note that this issue happened for all the blocks, not just the image block.


swissspidy:
@miina I see that this keeps block selection after resizing and rotating, but how does this fix #2306?

miina:
@swissspidy The block gets deselected and the editor unfocused after rotating/resizing so that the user needs to click anywhere on the editor again for being able to drag again.

This PR forces the selection of the block that was resized / rotated so that the user wouldn't need to select anything.

This is before:
May-21-2019 12-12-44

This is after:
May-21-2019 12-16-10

Are you seeing something different?

swissspidy:
Ah I see. In that case it's weird that the block mover was shown for a non-selected block in the first place when it's not working 🤷‍♂

PR #2338, opened 2019-05-17
Fix Latest Stories block styling; prevent registering block when Stories not enabled

kienstra:
* Applies Weston's suggestion of checking in `componentWillMount()` for whether the stylesheet is present
* Uses `wp_localize_script` to make the URL of `amp-story-card.css` available. Maybe there's a better way to do this.

# Before 
before-no-styling

# After
after-styling-present

Fixes #2300

PR #2355, opened 2019-05-20
Add the Latest Stories data hook

### LINKED ISSUE #2300: Latest Stories Block: Missing CSS
When inserting the "Latest Stories" block, the block is completely unstyled:

Screenshot 2019-05-13 at 18 05 07

The CSS is added when saving and reloading the post:

Screenshot 2019-05-13 at 18 05 16

kienstra:
This simply uses `add_action()` to add the existing hook.

Fixes #2300

### DIFF
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -174,6 +174,8 @@ class AMP_Story_Post_Type {
 
 		add_action( 'enqueue_block_editor_assets', array( __CLASS__, 'enqueue_block_editor_scripts' ) );
 
+		add_action( 'enqueue_block_editor_assets', array( __CLASS__, 'export_latest_stories_block_editor_data' ), 100 );
+
 		add_action( 'wp_enqueue_scripts', array( __CLASS__, 'add_custom_stories_styles' ) );
 
 		// Remove unnecessary settings.


kienstra:
**Request For Review**

Hi @swissspidy,
Hope you're doing great. Could you please review this when you can?

It [looks like](https://github.com/ampproject/amp-wp/issues/2300#issuecomment-494147158) the 'Latest Stories' block isn't available anymore.

This PR essentially reverts https://github.com/ampproject/amp-wp/pull/2338/commits/b21d6debdf04e5f71c51f74bf031ea50338df6c1, but maybe there's another reason for that commit.

At the moment, it doesn't look like `export_latest_stories_block_editor_data()` is referenced anywhere else, other than in a unit test.

Maybe it's intended to be called from within another method, not added as a hook.

Thanks, Pascal!

swissspidy:
FWIW The issue with that hook/method is that it also runs on the story editor screen. There's just no output there because `amp-block-editor.js` is not enqueued in that scenario. Ideally that code would live in `AMP_Post_Meta_Box` to prevent this confusion.

westonruter:
Otherwise, `export_latest_stories_block_editor_data` could have inside of it a check to see that the current post type is not `amp_story`?

kienstra:
Sounds good, I could open a PR for that if you'd like.

swissspidy:
@kienstra Thanks, that would be great.

kienstra:
Sounds good, I'll do that now.

PR #2363, opened 2019-05-21
Update packages and fix tests

swissspidy:
The updated WP JS packages have been released today, including some of the fixes I've been working on regarding the build tooling.

Now anyone should be able to run the tests without having to run `npm link` or similar.

To verify:

1. Remove `node_modules`
1. Run `npm install`
1. Run `npm run test:js`

Expected output:

```
// ...

Test Suites: 1 skipped, 25 passed, 25 of 26 total
Tests:       6 skipped, 84 passed, 90 total
Snapshots:   6 passed, 6 total
Time:        22.123s
Ran all test suites.
```

Now that this is fixed we can look into running these tests on CI too.

Closes #2361.

PR #2364, opened 2019-05-21
Do not try to download Chromium by default

swissspidy:
This happens because the wordpress/scripts package depends on Puppeteer, which results in attempts to download Chromium when running `npm install`.

Once we use Puppeteer for end-to-end tests that can be run on Travis, we can enable the download again just for that stage. Until then, setting `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true` should make Travis jobs a bit faster.

Related: #2363, https://github.com/WordPress/gutenberg/pull/15712

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ cache:
     - $HOME/phpunit-bin
     - $HOME/deployment-targets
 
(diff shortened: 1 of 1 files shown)


westonruter:
The directory where Chromium is being downloaded to could also be added to the list of cached directories.

PR #2365, opened 2019-05-21
Fix odd behavior when transforming Text to List block

### LINKED ISSUE #2304: Odd Behavior when typing a numbered list into a text block
1. Add new text block to story page
2. Begin typing a numbered list inside text block, "1. " 
3. Upon pressing the space bar after typing "1." the block converts to a numbered list block (correct), but jumps to the top of the page (incorrect)

Screen Shot 2019-05-14 at 10 37 04 AM

Screen Shot 2019-05-14 at 10 37 26 AM

Screen Shot 2019-05-14 at 10 37 34 AM

miina:
Fixes #2304.

This PR removes the prefixed transformation from the list block, not allowing the Text block to convert into a list block automatically.


miina:
@swissspidy Do you have any ideas here how to still keep the transformations but include the attributes from the Text block?

I was originally thinking of `blocks.switchToBlockType.transformedBlock` but apparently, that's not applied in case of automatic transformation. 

Also tried to change the prefixed list transformations to include getting attributes but unfortunately the origin block isn't passed in as a param to `transform`, just `content` is passed, as far as I understood.

Right now this PR removes the prefixed transformations for list so that it doesn't convert automatically, however, not sure if there's an easy way how to keep that, too. Would be great if the transformation filter would be applied in case of automatic transformations, too. Maybe it would make sense to resolve this upstream instead.

But perhaps you have another simple solution in mind that I missed. Alternatively, we could remove the transformations for now.

Thoughts?

assets/src/stories-editor/helpers/index.js
```diff
@@ -1,3 +1,5 @@
+/* global _ */
```
swissspidy: Better to use `import { each } from 'lodash';` for that. Webpack maps that to window.lodash automatically. This way you also get autocompletion in the IDE.

Although for this use case just using `settings.transforms.from.forEach()` should be fine too.

miina:
The list block doesn't allow font/color changes so probably, for now, it's better to just remove the automated prefixed transformations, otherwise, the user won't have a choice to customize numbered/bulleted sentences since they'd automatically convert to a list instead.

swissspidy:
> Do you have any ideas here how to still keep the transformations but include the attributes from the Text block?

Not really at this moment. 😕 

> Would be great if the transformation filter would be applied in case of automatic transformations, too. Maybe it would make sense to resolve this upstream instead

Opening an issue or even PR upstream would be good 👍 

> The list block doesn't allow font/color changes so probably, for now, it's better to just remove the automated prefixed transformations

Hmm because the list blocks only has transforms form and to the default paragraph block, maybe we should add some transforms to our custom text block.

Specifically, these two:

https://github.com/WordPress/gutenberg/blob/f2cafbd39ed0e905980e6ed2d151e0747aa72151/packages/block-library/src/list/transforms.js#L37-L59
https://github.com/WordPress/gutenberg/blob/f2cafbd39ed0e905980e6ed2d151e0747aa72151/packages/block-library/src/list/transforms.js#L110-L124

Could be something for a separate issue / PR where we could review all blocks and their transforms.

Thoughts?

miina:
> Hmm because the list blocks only has transforms form and to the default paragraph block, maybe we should add some transforms to our custom text block.

Good point, we're not using the paragraph block anyway so we could replace these transformations. Makes sense in a separate PR, this one is for the bug fix specifically.

swissspidy:
Cool, created #2380 for that.

PR #2369, opened 2019-05-21
Don't export the Latest Stories data in the AMP Story editor

kienstra:
As @swissspidy [pointed out](https://github.com/ampproject/amp-wp/pull/2355#issuecomment-494325262), this data isn't needed in that case.

This uses a [similar conditional](https://github.com/ampproject/amp-wp/blob/85f8c1ffe31e154f8553e89553affdf4b2ac8bf6/includes/class-amp-story-post-type.php#L411-L413) to that in `enqueue_block_editor_styles()`.

### DIFF
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -570,6 +570,10 @@ class AMP_Story_Post_Type {
 	 * Export data used for Latest Stories block.
 	 */
 	public static function export_latest_stories_block_editor_data() {
+		if ( self::POST_TYPE_SLUG === get_current_screen()->post_type ) {
+			return;
+		}
+
 		$url = add_query_arg(
 			'ver',
 			wp_styles()->registered[ self::STORY_CARD_CSS_SLUG ]->ver,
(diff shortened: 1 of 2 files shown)


westonruter:
Build error:

```
1) Test_AMP_Post_Meta_Box::test_enqueue_block_assets
Trying to get property of non-object
/tmp/wordpress/src/wp-content/plugins/amp/includes/class-amp-story-post-type.php:573
/tmp/wordpress/src/wp-content/plugins/amp/tests/test-class-amp-meta-box.php:151
```

kienstra:
Hi @westonruter, sorry about that. The build is passing now.

westonruter:
It doesn't seem this PR has any practical effect? I checked the `amp-stories-redux` branch and I don't see `var ampLatestStoriesBlockData` on the `post-new.php?post_type=amp_story` admin page to begin with. So is this just a matter of correctness?

kienstra:
@westonruter, good point. I also see that `ampLatestStoriesBlockData` isn't present in the AMP Story editor, even without this PR. 

Maybe this could prevent a future regression, but otherwise it looks like this PR isn't very useful.

swissspidy:
As I mentioned in https://github.com/ampproject/amp-wp/pull/2355#issuecomment-494325262:

> There's just no output there because `amp-block-editor.js` is not enqueued in that scenario. Ideally that code would live in `AMP_Post_Meta_Box` to prevent this confusion.

westonruter:
> Ideally that code would live in `AMP_Post_Meta_Box` to prevent this confusion.

But I think it is preferable to keep all the AMP Stories code in this one class, as it makes it easier turn on/off.

swissspidy:
Fair enough. I guess we'd need to analyze all the code at some point anyway if we want to make it easier / bulletproof to completely turn off.

PR #2389, opened 2019-05-23
Allow dragging entire blocks

### LINKED ISSUE #2275: Consider removing drag handle in favor of dragging the block itself

kienstra:
* Removes the drag handle, and usually allows dragging the entire block
* Dragging usually doesn't work well when `pointer: cursor`, especially in the Text block

Update: dragging the image block works better now
dragging

Closes #2275


assets/src/components/block-mover/draggable.js
```diff
@@ -95,6 +97,15 @@ class Draggable extends Component {
...
 		}
 
+		/*
+		 * On dragging, the browser creates an image of the target, for example, the entire text block.
+		 * But there's already a clone below that's rotated in case the block is rotated,
+		 * and this can create a non-rotated duplicate of that.
+		 * So override this with an empty image.
+		 */
+		const dragImage = new Image();
+		event.dataTransfer.setDragImage( dragImage, 0, 0 );
```
kienstra: This workaround might need more thought, but at least prevents a duplicate, non-rotated image on dragging.

kienstra:
**Known Issues**

Pending approval of this PR's approach, I could work on these issues:

1. On clicking the image block and dragging it slowly, it sometimes engages the `<DropZone>` of the `<MediaUpload>` from the image block, not that of the AMP Story page block. This doesn't always happen, though:
image-editing

2. It's hard to drag the text block. It mainly drags on the edges of the block:
text-block-dragging

assets/src/components/block-mover/draggable.js
kienstra: Without this workaround (maybe hack):

duplicate-appears

kienstra:
**Request For Review**

Hi @swissspidy and @miina,
Hope you're doing great. Could you please review the approach in this PR?

There are [known issues](https://github.com/ampproject/amp-wp/pull/2389#issuecomment-496061775) above, and possibly more bugs.

Overall, dragging the entire block is working pretty well. The main exception I found was the text block.

Thanks!

assets/src/components/block-mover/block-drag-area.js
```diff
@@ -25,13 +14,13 @@ export const IconDragHandle = ( { isVisible, className, icon, onDragStart, onDra
...
 					<div
-						className={ dragHandleClassNames }
+						className={ className }
 						aria-hidden="true"
 						onDragStart={ onDraggableStart }
 						onDragEnd={ onDraggableEnd }
 						draggable
 					>
-						{ icon }
+						{ children }
```
kienstra: This `<BlockDragArea>` now wraps the block (children), instead of appearing alongside it. This allows dragging the entire block.

miina:
@kienstra It might be a good thing that text block is only draggable from edges, this will keep the functionality of selecting text for bold and italic. We might just need to figure out how to have the draggable area larger so that mainly only the area covered by Text wouldn't be draggable.

assets/src/components/resizable-box/edit.css
```diff
@@ -21,6 +21,11 @@
 .wp-block-image .components-resizable-box__container img {
 	height: 100%;
 	object-fit: cover;
+	/* Prevents dragging the image itself, as its container is draggable. */
+	-webkit-user-drag: none;
```
swissspidy: Pretty sure that only `-webkit-user-drag` exists nowadays. The other properties either never existed (because it was never in the spec) or are simply irrelevant.

miina:
@kienstra Looks like there is very few room between the edges of the Text block input and the edge of the block, we should probably figure out a way which would be the best option for keeping the bold and italic functionality as it is right now, and at the same time leaving more room for dragging. Perhaps just adding some kind of padding to the block (not the input) could work.

A few other notes:
- The dropzone (handled via CSS) was previously considering the drag handle only, for example, if you add an image and drag it from it's right edge out of the right side of the page, it goes out of the drag zone and thus doesn't drag. We might need to extend the drag zone.
May-27-2019 15-37-03
- We should probably display drag handle at the areas where the block is draggable, especially in case of a Text block.

Generally the approach looks good to me!

swissspidy:
> We should probably display drag handle at the areas where the block is draggable

`cursor: grab` would be good here.

----

Regarding movable text blocks, I feel like we should try a different route here.

While in a regular post it makes sense to be able to select and format some text without selecting a block first, I don't think that necessarily applies to the stories editor as well.

My immediate suggestion after comparing with other tools like Snapchat and Instagram would be this:

* If the text block is **not selected**, then I should be able to just drag it _anywhere_
* If I **select** the text block, then I should be able to just drag it _anywhere_
* If the text block is **selected**, then I should be able to click on the text block _again_ in order to write, edit, select the text in it.

This is kinda similar to what's currently being worked on upstream in https://github.com/WordPress/gutenberg/pull/15537, where an inner block is only selected after selecting the parent block first.

**tl;dr:** First click -> select block, dragging possible. Second click -> edit text, no dragging

With that being said, before we add any changes here, I want everyone to try the current state of this PR first. Then we can incorporate this feedback appropriately.

cc @ThierryA @amedina @westonruter

ThierryA:
@swissspidy I think text block dragging behaviour should as implemented in this PR with the addition of `cursor: grab`.

i.e. I don't think users should have to click twice in order to edit text as you suggested (click to select element + click to edit).

assets/src/components/resizable-box/edit.css
kienstra: Thanks, good point.

kienstra: https://github.com/ampproject/amp-wp/pull/2389/commits/cae00b77a931eb23fe87ad298d1c73e2113085e4 removes these properties. You're right, I can't find any reference to them with quick searches.

kienstra:
Hi @miina,
Thanks for reviewing this. Sorry, I accidentally replied to your comment with a confused emoji, I meant to choose the 😄 

>We should probably display drag handle at the areas where the block is draggable, especially in case of a Text block.

Good idea, https://github.com/ampproject/amp-wp/pull/2389/commits/ef75b10350fb906c65657be2a94d172c7f62271d shows the drag handle. 

I couldn't figure out a way to add padding within the Text block, so that the drag handle would display more.

For example, with:
```css
.wp-block[data-type="amp/amp-story-text"] [draggable="true"] {
	padding: 5px;
}
```

the front-end display is a little too far to the left:

text-block-here

>The dropzone (handled via CSS) was previously considering the drag handle only, for example, if you add an image and drag it from it's right edge out of the right side of the page, it goes out of the drag zone and thus doesn't drag. We might need to extend the drag zone.

Thanks, I didn't notice that before. I'll work on extending the drag zone.

kienstra:
Hi @swissspidy,
Thanks a lot for your review here.

>cursor: grab would be good here.

https://github.com/ampproject/amp-wp/pull/2389/commits/ef75b10350fb906c65657be2a94d172c7f62271d uses your idea of `cursor: grab`

cursor-grab

kienstra:
Hi @miina,

>The dropzone (handled via CSS) was previously considering the drag handle only, for example, if you add an image and drag it from it's right edge out of the right side of the page, it goes out of the drag zone and thus doesn't drag. We might need to extend the drag zone.

Thanks for bringing that up, https://github.com/ampproject/amp-wp/pull/2389/commits/2f2449c33624c890c1e53bc9c033dce5831ec570 gives the drop zone more room. Though I'm not sure that's the right value to change it to.

# Before

before-drop-zone


# After

after-drag-handle

swissspidy:
Re: making the drop zone bigger, could it theoretically take up the entire editor area? Then in JS we could block dropping when the image would be be completely off the page (is that doable?).

At some point we might want to enable dragging from one page to another, so this would need to be kept in mind as well...

Long story short, I am not sure about the best approach yet. Needs some testing I guess.

kienstra:
>Re: making the drop zone bigger, could it theoretically take up the entire editor area? Then in JS we could block dropping when the image would be be completely off the page (is that doable?).

Good question, I'll test that today if that sounds good.

swissspidy:
To give an example, we should prevent things like this from happening:

Screenshot 2019-05-29 at 11 45 28

(This image is actually on page 2)

Could be as simple as `overflow:hidden` but I guess that would be too restrictive.

miina:
> Could be as simple as overflow:hidden but I guess that would be too restrictive.

Yep, hiding the overflow would cause some other issues.

Not sure either what would be the best approach, perhaps it would make sense to allow dragging based on the size of the dragged block and never allow dragging the block so that it would be completely out of the relevant page. That wouldn't be doable by CSS and would probably require adjusting the dragging logic instead, to consider the size of the block. Not in scope for this issue though.

Maybe we could create a separate issue to re-think the dragzone.

kienstra:
**Workaround For Image Block**

https://github.com/ampproject/amp-wp/pull/2389/commits/51f43d1cc82984e7dc7f34406ccf4b1af1099608 adds a workaround for Firefox, where dragging an image engages the DropZone of the Image block, not the `DropZone` for the AMP Story page.

The expected behavior is that this `DropZone` isn't engaged, and you can instead drag the block to any part of the AMP Story page.

This isn't a great approach, but setting `-moz-user-drag: none;` for the `img` didn't work in Firefox:

dragging-results

kienstra:

Hi @swissspidy,
Thanks for all of your help with this. Could you please review this again?

#2480 captures your [points above](https://github.com/ampproject/amp-wp/pull/2389#issuecomment-496859934) about the drop zone.

One thing I couldn't figure out was [how to add more draggable space](https://github.com/ampproject/amp-wp/pull/2389#issuecomment-496477088) to the Text block. Do you have ideas for that?

Thanks, Pascal! Have a great weekend.

miina:
@kienstra A few notes:
- I'm seeing the grab mouse pointer for image block but not for the Text block nor for the meta blocks
- For the quote block, the grab mouse pointer shows on top of the text inputs as well, although it actually selects the text instead of dragging.

kienstra:
Hi @miina,
Thanks a lot for testing this. 

https://github.com/ampproject/amp-wp/pull/2389/commits/e36f1d9f0205a21e2acbf8c999c29a377ee011ba fixed your points in my local. Though there still isn't much area in the Text block that has `cursor: grab`.

grabbing

miina:
@kienstra For some reason, the background is not extending to the edges within this PR anymore:
Screen Shot 2019-06-03 at 7 18 43 PM
 
On making the area larger for dragging: one thing that we could do is assigning some padding to the the `draggable` div within the Text block, then these white edges would be draggable, too, the edge marked with the blue arrow:
Screen Shot 2019-06-03 at 7 21 01 PM

This reduces the area of the content itself though.
We could compensate that by increasing the displayed width and height of the block accordingly for the editor.
Also, instead of just leaving it transparent, we could use the same background color for the draggable element as well so that it wouldn't be noticeable, however, this would still have the actual content area reduced, it would leave the feeling as if the padding in the editor is larger than in the FE.

We could also add a different background color to create a border effect for dragging instead.

The same could be accomplished by adding border to the draggable element (and then potentially adding 10px to both height and width to compensate for the lost area):
`.wp-block[data-type="amp/amp-story-text"] div[draggable="true"] {
	border: 5px solid transparent;
}`
Perhaps having a transparent border by default and adding some color on hover?

Not sure which option would visually make the most sense to the user. Any thoughts, @kienstra / @swissspidy?

kienstra:
Hi @miina,
Thanks a lot for your ideas on this.

>@kienstra For some reason, the background is not extending to the edges within this PR anymore:

Good point, I see that also:
bg-color-here

I'll look at that, and the other solutions you mentioned above.

kienstra:
Hi @miina,
Thanks for bringing up the issue with the background color.

With https://github.com/ampproject/amp-wp/pull/2389/commits/b5cd97424bed8f6c7a115df18bcfa76b5b1163e2, it should now appear as expected:

bg-color-here

kienstra:
https://github.com/ampproject/amp-wp/pull/2389/commits/0f881359d1c8bafa477264deea1915a012d98cc3 applies Miina's idea for the Text block of using a border inside the draggable area:
border-here

I'll work on adjusting the front-end display.

kienstra:
Hi @miina,
Thanks for your great ideas here. https://github.com/ampproject/amp-wp/pull/2389/commits/6a5e5230f80bae751013c41bab68bf20d2e6377e calculates the front-end position of the Text block, taking into account the `5px` border.

The front-end display of the Text block looks good locally:
display-cursor

swissspidy:
That white border looks quite odd IMO:

Screenshot 2019-06-04 at 11 41 16

I'd just leave that space transparent.

kienstra:
Sure, https://github.com/ampproject/amp-wp/pull/2389/commits/b73b07be891ae2e6e995e59e2d57827846b8001e removes the white border.

swissspidy:
Looks like there are block invalidation errors now for the templates:

Screenshot 2019-06-04 at 15 14 50

Probably because of 6a5e5230f80bae751013c41bab68bf20d2e6377e. I assume there's no way around that? Ideally we'd have some bigger draggable area without all this complexity 😬 

----

#2494 adds some wrapper around the `RichText` component in `TextBlockEdit`. Could we perhaps add that 5px spacing to such a wrapper instead? Would that help?

assets/src/stories-editor/helpers/index.js
```diff
@@ -479,24 +479,31 @@ export const wrapBlocksInGridLayer = ( element, blockType, attributes ) => {
 	const style = {
 		style: {},
 	};
+	const isTextBlock = ( 'amp/amp-story-text' === blockType.name );
```
miina: @kienstra The border changes should be for the editor only, not sure why do we need to change anything for saving into the post content. Thoughts? The compensating would ideally influence the editor only since the dragging also influences the editor only.

miina: By influencing the editor only I'm thinking that maybe it would make more sense to adjust the positioning in the editor only, for example, to compensate in the dragging logic instead? Otherwise, if for some reason we would want to change the draggable area's size then it would break all the blocks.

Also, we should probably also compensate for the lost weight and height in the editor.

Without border:
Screen Shot 2019-06-04 at 4 37 18 PM

With border:
Screen Shot 2019-06-04 at 4 37 11 PM

As you can see the border has taken away some of the area that the block previously had. Perhaps we should add this back by displaying the block a bit larger, just in the editor?

miina:
@kienstra Font calculation seems to be considering the border area as well and is going over the actual area with background:
Screen Shot 2019-06-04 at 4 43 13 PM

> I assume there's no way around that? Ideally we'd have some bigger draggable area without all this complexity 😬

We should definitely try to implement all this within the editor only without breaking the existing blocks. We could add the compensation logic into the drag-drop logic instead, and compensate for the borders within the editor by displaying the block larger in the editor instead of changing what's being saved into the FE view. I haven't tried it but it should be possible, AFAIK. @kienstra let me know if it seems to be not doable for some reason.

Additionally, perhaps we could adjust the style of this element here, for the Text block only:
Screen Shot 2019-06-04 at 4 48 13 PM

And add it styling something like:
```
position: absolute;
top: 2px;
left: 2px;
right: 2px;
bottom: 2px;
```
This way the white area around the background would seem smaller while leaving the same size dragging area, for example:
Screen Shot 2019-06-04 at 4 51 22 PM

Thoughts?

kienstra:
**Thanks, your idea works well**

Hi @miina,
Thanks, great idea for the styling:

```css
.editor-block-list__layout .wp-block[data-type="amp/amp-story-text"] .editor-block-list__block-edit::before {
        position: absolute;
        top: 2px;
        left: 2px;
        right: 2px;
        bottom: 2px
}
```

https://github.com/ampproject/amp-wp/pull/2389/commits/70bcf0bad8bc66d5e63a46b211fbe597ba95dc18 uses this, and reverts the previous logic that changed the front-end styling:

fe-here

kienstra:
**Travis Error**

I'm not sure how to fix this [Travis Error](https://travis-ci.org/ampproject/amp-wp/jobs/541468856):

```
fatal: Path 'amp.php' exists on disk, but not in 'a17db73791f0db1b46409dc6bf482f
```
It looks like the commit it references, `a17db`, [does have amp.php](https://github.com/ampproject/amp-wp/tree/a17db73791f0db1b46409dc6bf482f).

swissspidy:
> I'm not sure how to fix this Travis Error:

Merging develop back into the branch or an empty commit (`git commit --allow-empty -m "Trigger build"`) usually helps with these hiccups.

miina:
> Thanks, your idea works well

Thanks for the updates, @kienstra.

The styling was actually suggested together with this CSS:
```
.wp-block[data-type="amp/amp-story-text"] div[draggable="true"] {
	border: 5px solid transparent;
}
```

The border was removed within 70bcf0b now, however, without that then the dragging area for the text block is again very small. Without adding the transparent border there is no reason for adding this either:
```
.editor-block-list__layout .wp-block[data-type="amp/amp-story-text"] .editor-block-list__block-edit::before {
        position: absolute;
        top: 2px;
        left: 2px;
        right: 2px;
        bottom: 2px
}
```

^ This does not make the dragging area larger, the suggestion was for styling purposes only -- to reduce the feeling of the white area between the background color and the border. However, if there is no border added then there is no need to adjust that styling either -- it will just make the borders overlap with the background color.

For example, now, dragging a Text block without amp-fit-text on (with custom font size) is very hard since the draggable area is very thin again. Does it work for you?

miina:
@kienstra For some reason, the template preview doesn't seem to be working for me, do you also see this?
Screen Shot 2019-06-05 at 12 10 58 PM

assets/src/stories-editor/blocks/amp-story-page/index.js
```diff
@@ -49,7 +49,7 @@ const schema = {
 		default: '[]',
 	},
 	overlayOpacity: {
-		default: 50,
+		default: 100,
```
miina: Note that this will likely cause invalidation issues in some templates, it needs updating in the template files, too.

miina:
@kienstra The reason for the templates not showing anymore seems to be this CSS in the `template-inserter/edit.css` file:
```
.amp-stories__editor-inserter__results .block-editor-block-preview__content > div > div:first-of-type,
.amp-stories__editor-inserter__results .block-editor-block-preview__content .rotatable-box-wrap,
.amp-stories__editor-inserter__results .block-editor-block-preview__content .reusable-block-indicator {
	display: none;
}
```
More specifically this selector:
`.amp-stories__editor-inserter__results .block-editor-block-preview__content > div > div:first-of-type`

Removing this one selector seems to fix the issue, however, would be good to verify that it doesn't cause any regression.

assets/src/stories-editor/blocks/amp-story-page/index.js
miina: Actually, it looks like for the Travel templates no validation errors appear. It would for Fandom templates though, in case these should get merged before this PR.

kienstra:
**Thanks!**

Hi @miina,
Thanks for your great ideas here. If it's alright, I cherry-picked [your commit](https://github.com/ampproject/amp-wp/pull/2389/commits/33d84b8946adc6493d0b9d5154f77fcefc70c98a) into this, understanding that it was a draft. Also, I implemented the `@todos`.

>Removing this one selector seems to fix the issue, however, would be good to verify that it doesn't cause any regression.

Thanks for finding this. It looks like style rule. that was intended to hide the previous grab handles, which are removed in this PR. So I [removed the rule](https://github.com/ampproject/amp-wp/pull/2389/commits/175a9cdba339d1995e60355b1ba4e98b449c2ec0).

westonruter:
> Thanks for your great ideas here. If it's alright, I cherry-picked [your commit](https://github.com/ampproject/amp-wp/pull/2389/commits/33d84b8946adc6493d0b9d5154f77fcefc70c98a) into this, understanding that it was a draft.

FYI: Cherry-picking others' commits is what causes googlebot to complain.

kienstra:
>FYI: Cherry-picking others' commits is what causes googlebot to complain.

Ah, sorry about that.

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -111,9 +137,20 @@ class Draggable extends Component {
...
 
 		// Position clone over the original element.
-		this.cloneWrapper.style.top = `${ getPixelsFromPercentage( 'y', parseInt( clone.style.top ) ) }px`;
-		// Add 5px adjustment for having the block mover right next to the clone.
-		this.cloneWrapper.style.left = `${ getPixelsFromPercentage( 'x', parseInt( clone.style.left ) ) }px`;
+		if ( 'amp/amp-story-text' === blockName ) {
+			// Add an adjustment for having the block mover right next to the clone.
+			const pattern = /calc\(([\d\.]+)%/;
+			const matchesTop = clone.style.top.match( pattern );
+			const percentageTop = matchesTop ? matchesTop[ 1 ] : 0;
```
kienstra: For example, if `clone.style.top` was `calc(98.53%)`, `percentageTop`, would be `98.53`

kienstra:
**Review Request**

Hi @swissspidy,
You've been busy on this project, but would you have time to review this again? 

Thanks, and have a  great weekend 😄

assets/src/stories-editor/blocks/amp-story-page/index.js
swissspidy: Updated templates in ffda3a97589671e737aca8f289b1fe8797dc2360

kienstra: Thanks!

swissspidy:
Some things I've noticed so far:

a) The tip block is a bit small in the tip template preview.

Screenshot 2019-06-07 at 12 09 30

Note: This might very well be caused by some other, unrelated change. I just wanted to track it really quick. I'm gonna test on the develop branch as well.

b) The text block's background overflows the border:

Screenshot 2019-06-07 at 12 27 08
Screenshot 2019-06-07 at 12 13 56

c) Block position on the frontend matches the block's border in the editor, not its background.

If I position my text block like this:

Screenshot 2019-06-07 at 12 31 17

It looks like this on the frontend:

Screenshot 2019-06-07 at 12 31 36

Notice the missing gap.

If you select the block in the editor, you'll see that the top border is where the block is actually located on the frontend:

Screenshot 2019-06-07 at 12 31 26

Is this intended?

swissspidy:
> a) The tip block is a bit small in the tip template preview.

So this is fine in `develop`. There must have been a change in this PR that does this.

> b) The text block's background overflows the border:

It's similar in `develop`, but not as bad/obvious as with this PR.

> c) Block position on the frontend matches the block's border in the editor, not its background.

Looks like that's the same behavior as before. So I guess it's OK 👍

kienstra:
>a) The tip block is a bit small in the tip template preview.

Good point, https://github.com/ampproject/amp-wp/pull/2389/commits/683b8f6c7e639649f6f458ff493ba04001bdd863 makes the block slightly wider:

tip-here

kienstra:
Hi @swissspidy,
Sorry for the delay.

>b) The text block's background overflows the border

Good point. I saw that also, if this is what you had in mind:

bg-overflows-border

https://github.com/ampproject/amp-wp/pull/2389/commits/4fe55887ba26c7fadde1ca03120f25e774963a28 is a workaround, at best, that seems to fix this locally.

It might be a deeper issue than this, though.

With that commit, the front-end at least looks like the editor:

# Editor
in-editor-here

# Front-end
in-front-end

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -37,6 +40,20 @@ class Draggable extends Component {
...
+	 * In the image block, ensure that the preview image itself isn't draggable.
+	 *
+	 * When the image is draggable, it engages the DropZone of the Image block,
+	 * not the DropZone that allows dragging the block on the page.
+	 * There looks to be an easier way to do this with CSS,
+	 * but -moz-user-drag: none; didn't work on Firefox.
+	 */
+	componentDidMount() {
+		document.querySelectorAll( '.block-editor-block-list__block[data-type="core/image"] img' ).forEach( ( image ) => {
+			image.setAttribute( 'draggable', 'false' );
```
miina: Not critical but I wonder if there's a better place for doing this -- it seems to do the same for every `Draggable` and each time for all the images.

miina:
Some notes from testing:
- List block: The mouse pointer is displayed as drag handle on top of the text, however, it's not actually draggable from the text, the text is selectable instead. Would be good to change the mouse pointer.
- Preformatted block displays dragging handle, however, it's almost impossible to find a place where to drag it actually, it seems to allow dragging only somewhere near the bottom border.
- The resizing handles are a bit inside the block for Text block:
Screen Shot 2019-06-10 at 1 47 06 PM

@swissspidy Do you think if any of these are critical?

swissspidy:
@miina Not critical, but also rather easy to fix/improve.

The behavior for the list and preformatted blocks is quite the same. The drag handle is always shown and in my experience always works - unless you're precisely hovering over the text itself (because the browser would start text selection). Here's what they look like in the DOM:

Screenshot 2019-06-10 at 15 47 52
Screenshot 2019-06-10 at 15 47 14

Instead of the `ul` and `pre` elements having margins by default, we could add some padding on their parents instead. Plus, they probably need the same `height: 100%` changes.

> The resizing handles are a bit inside the block for Text block:

Will update the CSS there 👍

swissspidy:
@miina Just pushed some commits that already improve the drag handling for these core blocks quite a bit.

miina:
Thanks, @swissspidy, better now! 👍 

I'm still seeing this with the list block, is it different for you? Basically it doesn't seem to be draggable from the upper part of the block. Perhaps it's OK for now since it's quite close to the Text, wondering if it would need similar handling to the Text block 🤷‍♀ 
list-test

swissspidy:
@miina Yup, that's the margin/padding situation I mentioned. Pushed a commit now that improves this a bit for the list block. We can revisit and fine-tune in the future.

Issue #2189, opened 2019-04-24
[AMP Stories] Resizing doesn't properly apply to Story blocks for Title, Author, and Date


westonruter:
I added blocks for the story title, story author, and story date and then tried resizing them to cause the text to be auto-fit like the Text block, but no resizing happened:

![image](https://user-images.githubusercontent.com/134745/56700139-77fdae00-66ad-11e9-9063-6d6b15708f0e.png)

The resized blocks don't change their height in the editor, but nevertheless their heights are changed on the frontend:

![image](https://user-images.githubusercontent.com/134745/56700154-8cda4180-66ad-11e9-8e42-8a4407dad28e.png)

So either auto-fit should be applied to these blocks like it does for the Text block, or the hight of the element should be reflected in the editor as it will be on the frontend.

swissspidy:
There's #2282 for adding amp-fit-text to these blocks

miina:
Instructions for testing:
1. Add title, author, date blocks.
2. Add background color.
3. Resize the blocks.
4. Make sure that the background color fills in the full area.
5. Make sure that the font size changes according to the new resized area if "Automatically fit text to container" setting is set.
6. Make sure that the FE and editor view match.

csossi:
Very different results on front end

Editor:
![image](https://user-images.githubusercontent.com/16945664/58037534-a1b6c300-7afb-11e9-813c-ce68130d54c0.png)

Preview:
![image](https://user-images.githubusercontent.com/16945664/58037575-b004df00-7afb-11e9-8b6b-2eec78c7f313.png)

miina:
@swissspidy This might be related to #2296, assigning this to you, let me know if it seems it's not related.

miina:
Closing as ready for re-testing for QA.

csossi:
Still seeing a diff between editor and front-end re: font size/style
Editor:
![image](https://user-images.githubusercontent.com/16945664/58114706-561d1b80-7bc6-11e9-8b23-37ed5bdf7fb6.png)

FE:
![image](https://user-images.githubusercontent.com/16945664/58114736-67662800-7bc6-11e9-9b48-0f1f15553047.png)

swissspidy:
@csossi To confirm, those are: text block, author block, date block, in that order? Or is the top one a title block?

csossi:
Correct - in that order, @swissspidy

swissspidy:
I cannot seem to reproduce the font size difference.

[...]

PR #2395, opened 2019-05-23
Update min and max font sizes to match amp-fit-text

swissspidy:
Since amp-fit-text is also used in the regular block editor, it makes sense to use the same values in both environments.

See #2189.

### DIFF
--- a/assets/src/common/constants.js
+++ b/assets/src/common/constants.js
@@ -1,2 +1,3 @@
-export const MIN_FONT_SIZE = 14;
-export const MAX_FONT_SIZE = 54;
+// See https://github.com/ampproject/amphtml/blob/e7a1b3ff97645ec0ec482192205134bd0735943c/extensions/amp-fit-text/0.1/amp-fit-text.js#L81-L85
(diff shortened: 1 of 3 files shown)


miina:
Hmm, for some reason I'm seeing 36 as the applied font size when resizing the author block large, can you confirm if you can see 72 as intended?

Screen Shot 2019-05-23 at 1 55 45 PM

swissspidy:
Hmm `36` is the default value for `autoFontSize`, suggesting that the font size calculation is not running. Are there any errors in the console or something?

Also, I added some related changes in #2396. Can you try merging that branch into this one to see if this changes anything?

miina:
Yep, looks like the font calculation is not working, for author and date block specifically.

miina:
It might be because of the 100% height assigned -- we might need to remove that for the font calculation time.

swissspidy:
It's even more trivial than that 🤦‍♂ See f07724d50e3030d0d6f2f615dbd9537a223a69f0

As for the `height: 100%`, we already remove it during calculation.

PR #2398, opened 2019-05-23
 Merge amp-editor-blocks into amp-block-editor

### LINKED ISSUE #2333: Review shared amp-editor-blocks script
`amp-editor-blocks.js` is currently enqueued for both the regular block editor as well as the stories editor.

That script is rather small, and only applies a few filters to add some extra attributes and controls to existing blocks.

As pointed out in https://github.com/ampproject/amp-wp/pull/2331#discussion_r284780247, most if not all of the controls added by  that script do not make sense in a stories context. For example the AMP layout, noloading, or lightbox controls.

We need to go through them and see whether we need to move them over to `amp-block-editor.js` (which doesn't load for stories). Perhaps we can even get rid of `amp-editor-blocks.js` so that everything is in either `block-editor.js` or `amp-stories.js`.

Tagging @westonruter @kienstra as you probably know these controls better than I.

swissspidy:
Fixes #2333.


assets/src/block-editor/helpers/index.js
```diff
@@ -133,7 +133,8 @@ export const addAMPAttributes = ( settings, name ) => {
 			attribute: 'max-font-size',
 		};
 		settings.attributes.height = {
-			default: 'core/image' === name ? 200 : 50,
+			// Needs to be higher than the maximum font size, which defaults to MAX_FONT_SIZE
+			default: 'core/image' === name ? 200 : 80,
```
miina: Should we use the constant here instead then, e.g. `MAX_FONT_SIZE + ...`?

includes/admin/class-amp-editor-blocks.php
```diff
@@ -120,49 +119,6 @@ public function whitelist_block_atts_in_wp_kses_allowed_html( $tags, $context )
...
-		wp_enqueue_script(
-			$script_handle,
-			amp_get_asset_url( 'js/amp-editor-blocks.js' ),
-			$script_dependencies,
-			AMP__VERSION,
-			true
-		);
-
-		if ( function_exists( 'wp_set_script_translations' ) ) {
-			wp_set_script_translations( $script_handle, 'amp' );
```
miina: Curious to know -- is `wp_set_script_translations` not needed anymore?

swissspidy: This `wp_set_script_translations` call here is for `amp-editor-blocks.js`, which this PR removes 🙂 

`wp_set_script_translations` for the one and only `amp-block-editor.js` script that's left is still in `AMP_Post_Meta_Box`.

miina: Ah, makes sense, thanks!

assets/src/block-editor/helpers/index.js
swissspidy: Fixed in 0e3fbe3f0ff8d66e7a78ecc7368c733189935e9a

swissspidy:
^ The PR build failure is just a Travis hiccup.

Issue #1840, opened 2019-01-26
Reliable builds and deploys


kasparsd:
Currently the build process defined in [Gruntfile.js](https://github.com/ampproject/amp-wp/blob/5aaf0005c093e39e39de7ed214de0b56807e8420/Gruntfile.js#L69-L154) relies on an unknown state of the checked out repository -- For example, it uses existing dependencies in `/vendor` and `/node_modules` directories without installing them according to the rules in the lock files.

For reliable and reproducible builds we need to isolate any local overrides and always start from a known state. I suggest to update the deploy process to the following steps:

- Create an empty `/build/src` directory.

- Copy the `.git` directory from the root to `/build/src/.git`.

- Checkout the required release branch such as `develop` or `master`. This ensures that we never accidentally deploy a feature brunch.

- Install all dependencies via `npm install` and `composer install --no-dev --optimize-autoloader`.

- Build assets `npm run build` and replace any version strings as necessary.

- Copy the release files to `/build/release` and make sure to exclude `node_modules` and other development specific files. Maybe introduce `.distignore` that defines all files and directories that shouldn't be part of the final release. Maybe use [parse-gitignore](https://www.npmjs.com/package/parse-gitignore) to generate the exclude list.

- Now deploy the /build/release to WP plugin SVN or create a release zip file.

All of these steps can be made cross-platform, too.

westonruter:
Some of this is addressed by https://github.com/ampproject/amp-wp/pull/1836 correct? Note that the release instructions in `contributing.md` include a step to run `npm install` and `composer install`. It should also include a step to make sure the right branch is checked out. 

But yeah, these are manual steps that would ideally be automated as much as possible.

kasparsd:
@westonruter Yes, some of this is covered by #1836 and for reasons similar to what you mentioned in [this comment](https://github.com/ampproject/amp-wp/pull/1836#discussion_r251527136) it would be better to introduce a completely standalone build and deploy process.

Only by moving the build process away from the working directory we can ensure that none of the work in progress is affected.

With this kind of setup we could even make Travis deploy to the WP.org repo since it would follow the exact same steps as one does locally.

swissspidy:
Thanks for opening this!

I would love to see continuous builds on Travis instead of having to do things like #1836. Happy to lead work on that front too.

Continuous deployments to WordPress.org is probably a stretch goal and not easy to do because of SVN credentials. I'd leave this out for the moment.

swissspidy:
[...]

PR #2408, opened 2019-05-24
Update Travis CI configuration

### LINKED ISSUE #2384: Run JavaScript Unit Tests on Travis
Now that the setup is there, we should run our JS test (`npm run test:js`) on Travis CI. Since those are not dependant on a PHP version or anything, we can create a new stage for only running these JS tests.

The E2E tests can follow at a later date.

swissspidy:
What this PR does:

* First and foremost, this runs JavaScript unit tests on Travis CI in a separate job.
* Disables branch builds for pull requests.  
  Travis already runs for PRs itself, running it again for the merge commit is usually pointless and just takes time.
* Creates a separate build stage for linting only.  
  This way, we don't bother running tests when code isn't properly formatted.
* Lints composer.json and package.json files in this new lint stage.
* Turns on fast finishing.  
  This way, the build will be marked as finished  as soon as one job has failed. This is done under the assumption that all the other tests are likely to fail too when one has failed already.
* Disables redundant PHP syntax and ESLint checks.  
  ESLint is already run during the lint stage, and PHP syntax only needs to be tested once per PHP version.
* Reverse order of jobs to test on newer PHP and WordPress versions first.  
  Together with the fast finishing this makes things a bit faster in case of failures.
* Adds some names to the different jobs to make it clearer what each job does.  
  See screenshots below for comparison.
* Only installs PWA plugin on WP 5.2+ and PHP 5.6+
* Runs tests against WordPress trunk again  
  This job is allowed to fail
* Makes tests more robust / fixes covers annotations

**Before** ([link](https://travis-ci.org/ampproject/amp-wp/builds/536506877))

Screenshot 2019-05-24 at 12 36 42

**After** ([link](https://travis-ci.org/ampproject/amp-wp/builds/536693345))

Screenshot 2019-05-24 at 12 36 50

**Some thoughts/notes**

* `npm run build:js` should unnecessary for the linting stage and could probably be skipped there.
* Continuous builds would be nice at some point. (see #1840)
* Is that `ampconfdemo` deploy stage still relevant?
* There are no tests against `WP_VERSION=nightly`.  
  Changing that would be useful to catch potential conflicts early, and not just after WordPress has been released.
* Always installing Gutenberg for WordPress 5.0+ seems excessive and dangerous.  
  If I am not missing something, this never tests WordPress core's block editor integration as-is, but always uses the the Gutenberg one. Currently not much of a problem, as Gutenberg does not have any new PHP stuff at the moment. But when they're going to [drop support for PHP < 5.6](https://github.com/WordPress/gutenberg/issues/15796) this might cause problems in our tests.
* Always installing the PWA plugin seems a bit excessive for just the few service worker tests there are.  
  Given that the PWA plugin only supports WordPress 5.2+, this should only be installed when testing WordPress 5.2+.

Fixes #2384.


westonruter:
> * Is that `ampconfdemo` deploy stage still relevant?

This was being used for QA last year but it has slowly discontinued being used. I'd rather see the automatic deploys be made to `*.amp-wp.org` environments now.

> * There are no tests against `WP_VERSION=nightly`. Changing that would be useful to catch potential conflicts early, and not just after WordPress has been released.

We did have `trunk` until October when it was replaced with 5.0: af3719247d6c7baa16ccd4c912d4cd5f11a932cc. This was done because of instability of core at the time, I believe. Perhaps `trunk` should be put in an `allow_failures` bucket. 

> * Always installing Gutenberg for WordPress 5.0+ seems excessive and dangerous.
>   If I am not missing something, this never tests WordPress core's block editor integration as-is, but always uses the the Gutenberg one. Currently not much of a problem, as Gutenberg does not have any new PHP stuff at the moment. 

You're right. Making Gutenberg a dependency was specifically to facilitate testing with AMP Stories. Once AMP Stories no longer requires Gutenberg to work (which may be the case as of 5.2 actually) then we should only install/activate Gutenberg on a single env.

> But when they're going to [drop support for PHP < 5.6](https://github.com/WordPress/gutenberg/issues/15796) this might cause problems in our tests.

We should also drop support for PHP<5.6 then as well.

> * Always installing the PWA plugin seems a bit excessive for just the few service worker tests there are.
>   Given that the PWA plugin only supports WordPress 5.2+, this should only be installed when testing WordPress 5.2+.

Makes sense to me.

.travis.yml
```diff
@@ -23,6 +23,11 @@ cache:
     - $HOME/phpunit-bin
     - $HOME/deployment-targets
 
+branches:
+  only:
+    - master
+    - develop
```
westonruter: Also need to add release branches here. Namely adding:

```
    - /^\d+\.\d+$/
```

.travis.yml
```diff
@@ -34,29 +39,56 @@ install:
...
-      env: WP_VERSION=latest DEV_LIB_SKIP=phpcs
-    - php: "7.1"
-      env: WP_VERSION=latest DEV_LIB_SKIP=phpcs
-    - php: "7.2"
-      env: WP_VERSION=latest DEV_LIB_SKIP=phpcs
+      env: WP_VERSION=latest DEV_LIB_SKIP=phpcs,eslint,xmllint,phpsyntax,phpunit JS_TESTS=1
+
+    - name: PHP unit tests (7.3, WordPress latest)
+      php: "7.3"
+      env: WP_VERSION=latest DEV_LIB_SKIP=phpcs,eslint,xmllint,phpsyntax
```
westonruter: Instead of `DEV_LIB_SKIP` you can instead use `DEV_LIB_ONLY` to just select the tests you want.

swissspidy:
Note: I was a bit surprised to see some of the tests pass. `\AMP_Style_Sanitizer_Test::test_prioritized_stylesheets` failed locally because it requires the Twenty Ten theme to be installed.

Turns out, wp-dev-lib downloads WP latest from https://develop.svn.wordpress.org/tags/5.2.1/, and `bin/install-wp-tests.sh` does so from https://wordpress.org/wordpress-5.2.1.zip. The former includes all themes, but the latter does only include the Themes 2016, 2017, and 2019. Hence the failure.

In ba891ce09cb2b9fea1117bf7830ed23ed1b4ac79 I skipped the test if the theme is missing.

PR #2415, opened 2019-05-24
Introduce prop types for runtime typechecking

### LINKED ISSUE #2237: Typechecking With PropTypes
While Gutenberg itself does not use [PropTypes](https://reactjs.org/docs/typechecking-with-proptypes.html) for typechecking during runtime, many other projects do, including the soon-to-be-released Site Kit plugin.

Typechecking like this is very useful for catching bugs, and is rather easy to implement.

It might be worth considering adopting this for our custom JavaScript as well, at least to some degree.

Besides that, using `defaultProps` in some cases might already be beneficial.

Related: #2121

swissspidy:
Added some prop types here and there (~ 25 components). If it's useful we can expand it to the rest of the code base.

If you run `npm run lint:js` you'll see that 145 props are currently missing prop types definitions in this PR.

Fixes #2237.


swissspidy:
To learn more about PropTypes, check out #2237, https://www.npmjs.com/package/prop-types and https://www.npmjs.com/package/prop-types.

Please let me know your thoughts on adopting runtime typechecking like this in our code.

If you want to test this PR, check out this PR, run `npm run dev` and edit stories. If there are prop type validation errors while creating a story, you would see them in the console. Nothing should happen when using `npm run build` (i.e. production mode).

miina:
I think it could be a good add-on, always good to have extra measures for catching bugs.

I guess a downside would be that we'd need to make sure to keep the prop-types updated when making changes.

Any downsides that you see for implementing this?

kienstra:
Hi @swissspidy,
This is interesting, good idea to bring this up.

I'd support this if we're committed to resolving the type errors soon after merging this.

For example, on loading a new AMP Story page, there's 1 console warning:
```
react.6025b791.js:2726 Warning: Failed prop type: Invalid prop `attributes.autoAdvanceAfterDuration` of type `number` supplied to `EditPage`, expected `string`.
```

And on adding an Image block and selecting an image, there are 2 type warnings.

This could add a lot of noise in the console if we're not going to address them soon.

But maybe this could prevent bugs.

miina:
> I'd support this if we're committed to resolving the type errors soon after merging this.

Probably would be good to do it already before merging, within this PR. I'd be happy to do it myself once all the prop types are added if it sounds like an annoying task 😄

swissspidy:
Yes, this requires some additional work during development to add prop types (otherwise ESLint will alert you) and to keep them updated (otherwise you'll get console errors).

These console errors only happen during development. So if we catch one, we're encouraged to fix them right away.

Nothing happens during production. In 6edbc475a640d1021f8bb1d68d5ea43472600815 I even mode sure the `.propTypes {}` declarations are stripped on production too, to keep builds as small as possible.

And yes, we would add missing prop types before merging, because otherwise the build would fail 🙂

swissspidy:
Alright, I added all the prop types now, and the tests are passing.

In my quick testing I couldn't see any prop types warning when fiddling with the editor. @miina Wanna give it a go too? If there are just a few issues we can fix them here.

After that, we can merge the PR I guess.

miina:
Got these two warnings when using the block Inserter in the Toolbar.

Screen Shot 2019-05-28 at 4 02 44 PM


Otherwise didn't encounter any warnings, looks good!

swissspidy:
Cool. I am going to fix these right away.

I also just now found an issue related to the changed dependencies order which I'll investigate before merging.

PR #2417, opened 2019-05-24
Improve styling of AMP admin screen and AMP admin menu item

### LINKED ISSUE #1924: AMP icon in admin menu appears black momentarily until page loads
When navigating through pages of comments I noticed the AMP icon in the admin menu was flashing black before turning light gray. For a moment I'd see:

image

And then upon DOM ready it becomes:

image

Something isn't right with how the AMP icon is added to the admin bar.

westonruter:
# AMP admin menu item

When navigating through pages of comments I noticed the AMP icon in the admin menu was flashing black before turning light gray. For a moment I'd see:

image

And then upon DOM ready it becomes:

image

The fix is to change the default fill color from black to gray. This greatly minimized the flash while waiting for core's `svg-painter` to run. The only flash then becomes going from gray to white if the AMP admin menu is selected. Otherwise, there is no flash at all. This fix is applied in this PR, along with compressing the SVG via SVGOMG.

TIL: `svg-painter`

Fixes #1924.

# AMP admin screen welcome message

I noticed the image in the welcome message on the AMP screen can overflow the container:

Screen Shot 2019-05-24 at 09 39 07

This PR ensures the container is at least 200px tall to prevent this from happening:

Screen Shot 2019-05-24 at 09 38 02

This PR also improves the layout stability of the welcome message by giving `width`/`height` dimensions to the `img`. (There is still some instability due to core's behavior of moving around admin notices on DOM ready.)


swissspidy:
Thanks for fixing these issues! I definitely noticed them too.

As for the SVG, is it possible to remove the fill completely and use CSS for that, or is that how svg-painter is supposed to work?

westonruter:
> As for the SVG, is it possible to remove the fill completely and use CSS for that, or is that how svg-painter is supposed to work?

It's how the svg-painter works, yeah. Interestingly, it changes the `fill` in the SVG dynamically:

https://github.com/WordPress/wordpress-develop/blob/6d523d504465e1118900dfb6d33f08491afd4d9e/src/js/_enqueues/wp/svg-painter.js#L215-L216

This happens dynamically on `jQuery.hover`:

https://github.com/WordPress/wordpress-develop/blob/6d523d504465e1118900dfb6d33f08491afd4d9e/src/js/_enqueues/wp/svg-painter.js#L162-L173

I just found more confirmation of this at https://stackoverflow.com/questions/30620226/custom-svg-admin-menu-icon-in-wordpress

westonruter:
These SVG icons come from MP6: https://make.wordpress.org/design/tag/icons/

<blockquote><h3>Icons</h3>
<p>We decided on recommendations for&nbsp;custom icons in plugins/themes, you can read more about this and have a look at some examples are&nbsp;<a href="https://github.com/tillkruess/MP6-Icon-Examples">in this GitHub repository</a>. Be default&nbsp;MP6 does now hide all images in admin menu item and replaces them with a generic default icon.</p>
<p>Method C, using a SVG as CSS background image, works thanks to the new JavaScript <a href="https://plugins.trac.wordpress.org/browser/mp6/trunk/js/svg-painter.js">SVG painter</a>, which re-colors base64 encoded SVGs on the fly. It might not catch all SVG color markup right away, if that’s the case, please post a link your SVG file in the comments.</p></blockquote>

Issue #2418, opened 2019-05-24
Native mode doesn't have the AMP's symbol on SERP


zJar84:
Hello, I am using the Native mode on my web and I see webs that use AMP and get a small symbol of AMP on the SERPs of Google on mobile. Results that this symbol doesn't appears on the SERPs of my web significadodelosgirasoles.ovh.

You can search `site:significadodelosgirasoles.ovh` on Google on mobile and see it.

It is because the web doesn't have more than one week?

Thanks.

westonruter:
Your site is not validating: https://validator.ampproject.org/#url=https%3A%2F%2Fsignificadodelosgirasoles.ovh

The problem is that you are outputting `amp-analytics` in the `head`. It must be inside the `body`.

Actually, I'm curious why the AMP plugin is not removing this itself since it is not valid AMP. In any case, if you move it to the `body` then the validation error will be fixed.

Desktop results don't currently include badges for AMP pages.

zJar84:
Thanks!

westonruter:
FYI: As of #2420 if you add `amp-analytics` to the `head` it will automatically be moved to the `body`. Nevertheless, it is still best to put it in the `body` to begin with.

PR #2420, opened 2019-05-24
Improve validation of the nodes in the head

### LINKED ISSUE #2419: Elements not allowed in head are not removed
As observed in #2418, if an otherwise-valid AMP element (like `amp-analytics`) appears in the `head` as with this plugin code:

```php
add_action(
	'wp_head',
	function () {
		?>
		<amp-analytics id="75a1fdc3143c" type="googleanalytics"><script type="application/json">{"vars":{"account":"UA-XXXXXX-1"},"triggers":{"trackPageview":{"on":"visible","request":"pageview"}}}</script></amp-analytics>
		<?php
	}
);
```

The result is AMP validation errors that leak onto the frontend. Such elements are not allowed in the `head`, but the AMP plugin is not removing them. 

This is a particularly confusing error to debug because the validation errors coming from the AMP validator also do not indicate that the element is incorrectly in the `head`, except for the first which alludes to it:

image

This is likely because of the HTML parser in that as soon as it encounters an element unexpected in the `head` it treats that as an implicit start to the `body`.

So either such elements need to be removed from the `head`, or they should automatically be moved to start at the beginning of the `body`.

This also goes for text nodes appearing in the `head`, not just elements.

westonruter:
This PR prevents users from invalidating their pages by accidentally putting disallowed elements in the `head`. HTML has a fixed set of elements that are allowed in the `head` and if any other element is encountered (or a text node) then this is interpreted by browsers as an implicit closing of the `head` and the opening of the `body`. Nevertheless, the PHP DOM parser does not exhibit this behavior. So by moving such nodes to the `body` automatically then the PHP handling behaves more like the browser.

* Prevent AMP validation errors from leaking to the frontend due to disallowed `head` nodes being present.
* Move invalid `head` nodes to the `body`.
* Ensure `head` and `body` are present.
* When unwrapping `noscript` elements in the `head`, move children to `body` if not allowed in `head`.

Fixes #2419.  See #2418.

### DIFF
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -184,7 +184,7 @@ class AMP_DOM_Utils {
 	 * @param DOMElement $body BODY element.
 	 */
 	private static function move_invalid_head_nodes_to_body( $head, $body ) {
-		// Walking backwards is to make it easier to prepend them to the body in the expected order.
+		// Walking backwards makes it easier to move elements in the expected order.
 		$node = $head->lastChild;
 		while ( $node ) {
 			$next_sibling = $node->previousSibling;
--- a/includes/utils/class-amp-dom-utils.php
+++ b/includes/utils/class-amp-dom-utils.php
@@ -43,6 +43,23 @@ class AMP_DOM_Utils {
 	);
 
 	/**
+	 * List of elements allowed in head.
+	 *
+	 * @link https://github.com/ampproject/amphtml/blob/445d6e3be8a5063e2738c6f90fdcd57f2b6208be/validator/engine/htmlparser.js#L83-L100
+	 * @link https://www.w3.org/TR/html5/document-metadata.html
+	 * @var array
+	 */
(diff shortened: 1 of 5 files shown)


includes/utils/class-amp-dom-utils.php
```diff
@@ -141,9 +158,62 @@ function( $noscript_matches ) {
...
+	 * Move elements not allowed in the head to the body.
+	 *
+	 * Apparently PHP's DOM is more lenient when parsing HTML to allow nodes in the HEAD which do not belong. A proper
+	 * HTML5 parser should rather prematurely short-circuit the HEAD when it finds an illegal element.
+	 *
+	 * @param DOMElement $head HEAD element.
+	 * @param DOMElement $body BODY element.
+	 */
+	private static function move_invalid_head_nodes_to_body( $head, $body ) {
+		// Walking backwards is to make it easier to prepend them to the body in the expected order.
```
amedina: Consider: "Walking backwards makes it easier to move elements in the expected order"

PR #2425, opened 2019-05-26
Improve form handling when submission handler does no redirection

westonruter:
This follows up on #911/#923.

When a plugin adds a form to the page and the handling of the form submission does not redirect to a success page, the current behavior of the AMP plugin is to just not show any indication at all that a submission was made. The `amp-form` client-side fails because the `POST` response contains an HTML page as opposed to the expected JSON. This is bad because many forms in WordPress process submissions without redirections, leaving them to appear as broken. (In contrast, when `wp_redirect()` is done, then `\AMP_HTTP::intercept_post_request_redirect()` will automatically send the `AMP-Redirect-To` response header, taking the user to the intended success/failure page.)

To improve the experience, when processing a form submission (which didn't have `action-xhr` to begin with) we need to detect when the `POST` form handler is responding with an HTML body rather than a redirect, and in this case, try to serve some JSON that will populate the `submit-success` and `submit-error` message templates, based on whether the form handler returns a 200 or 400+ status code. When a 200 status is returned then a generic success message should be displayed, and when a 400+ status code is returned then a generic error message should be displayed. A challenge here is that form processors may detect an error but they still return with a 200 code, which we can't really do anything about.

The ideal is still that a theme/plugin developer adds direct support for forms by adding `action-xhr` to the `form` elements, sending JSON responses, and adding `div[submitting]`, `div[submit-success]`, and `div[submit-error]` message templates. But this improves compatibility with themes/plugins that have not yet done this. 

To test this PR, use this test plugin that has a `[test_form_post]` shortcode: https://gist.github.com/westonruter/6c0eb253972ccdfc31f841629e2f4344

# Before

No message is displayed at all when `200` response: 

> image

Nor in an error response:

> image

# After

A new `submitting` message is displayed:

> image

And upon a 200 response, a generic success message is displayed:

> Screen Shot 2019-06-01 at 22 41 00

And when a 200 response happens but the page is redirecting, a “Redirecting…” message is displayed without the success styling since it could be redirecting to an error page:

> Screen Shot 2019-06-01 at 22 43 19

If there is a `400`+ error response, then the `submit-error` message is displayed:

> image

## Comment Form

There is a new “Submitting” message here as well:

> Screen Shot 2019-05-27 at 10 14 18

The `submit-error` is now styled in the same way as the generic form conversion:

> image

Note the `ERROR:` prefix text is [coming from WordPress itself](https://github.com/WordPress/wordpress-develop/blob/d22572a5a5c985c450893edb0f780e4e0d337b6f/src/wp-includes/comment.php#L3310).

# Todo

- [x] Should status codes be indicated in the generic responses?
- [x] Is the default styling for the form messages OK?
- [x] Improve styling for comment form error message.
- [x] Ensure no regressions for comments.
- [x] Ensure that invalid comment submission responses return with error codes. See https://core.trac.wordpress.org/ticket/47393
- [x] Do not style redirection as green success.
- [x] Test with existing form plugins.
  - [x] Jetpack Contact Forms
  - [x] Contact Form 7 (with [helper plugin](https://gist.github.com/westonruter/3501016b0d44af45d878067b1856e023))

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1078,7 +1078,7 @@ class AMP_Theme_Support {
 		</div>
 		<div submit-error>
 			<template type="amp-mustache">
-				<p class="amp-comment-submit-error">{{{error}}}</p>
+				<p class="amp-comment-submit-error">{{{message}}}</p>
 			</template>
 		</div>
 		<?php
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1720,6 +1720,8 @@ class AMP_Theme_Support {
 		$is_form_submission = (
 			isset( AMP_HTTP::$purged_amp_query_vars[ AMP_HTTP::ACTION_XHR_CONVERTED_QUERY_VAR ] ) // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 			&&
+			wp_is_json_request()
+			&&
 			isset( $_SERVER['REQUEST_METHOD'] )
 			&&
 			'POST' === $_SERVER['REQUEST_METHOD']
@@ -1727,12 +1729,8 @@ class AMP_Theme_Support {
 		if ( $is_form_submission && null === json_decode( $response ) && json_last_error() && ( ( $status_code >= 200 && $status_code < 300 ) || $status_code >= 400 ) ) {
 			return wp_json_encode(
 				array(
-					'message' => sprintf(
-						/* translators: %1$d: the HTTP response code, %2$s: the status description */
-						__( 'Server responded with status code %1$d: %2$s', 'amp' ),
-						$status_code,
-						get_status_header_desc( $status_code )
-					),
+					'status_code' => $status_code,
+					'status_text' => get_status_header_desc( $status_code ),
 				)
 			);
 		}
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1731,7 +1731,7 @@ class AMP_Theme_Support {
 						/* translators: %1$d: the HTTP response code, %2$s: the status description */
 						__( 'Server responded with status code %1$d: %2$s', 'amp' ),
 						$status_code,
-						esc_html( get_status_header_desc( $status_code ) )
+						get_status_header_desc( $status_code )
 					),
 				)
 			);
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1705,11 +1705,38 @@ class AMP_Theme_Support {
 			unset( $args['validation_error_callback'] );
 		}
 
+		$status_code = http_response_code();
+
 		/*
-		 * Check if the response starts with HTML markup.
-		 * Without this check, JSON responses will be erroneously corrupted,
-		 * being wrapped in HTML documents.
+		 * Send a JSON response when the site is failing to handle AMP form submissions with a JSON response as required
+		 * or an AMP-Redirect-To response header was not sent. This is a common scenario for plugins that handle form
+		 * submissions and show the success page via the POST request's response body instead of invoking wp_redirect(),
+		 * in which case AMP_HTTP::intercept_post_request_redirect() will automatically send the AMP-Redirect-To header.
+		 * If the POST response is an HTML document then the form submission will appear to not have worked since there
+		 * is no success or failure message shown. By catching the case where HTML is sent in the response, we can
+		 * automatically send a generic success message when a 200 status is returned or a failure message when a 400+
+		 * response code is sent.
 		 */
+		$is_form_submission = (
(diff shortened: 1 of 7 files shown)


includes/class-amp-theme-support.php
```diff
@@ -1705,11 +1705,38 @@ public static function prepare_response( $response, $args = array() ) {
...
+			'POST' === $_SERVER['REQUEST_METHOD']
+		);
+		if ( $is_form_submission && null === json_decode( $response ) && json_last_error() && ( ( $status_code >= 200 && $status_code < 300 ) || $status_code >= 400 ) ) {
+			return wp_json_encode(
+				array(
+					'message' => sprintf(
+						/* translators: %1$d: the HTTP response code, %2$s: the status description */
+						__( 'Server responded with status code %1$d: %2$s', 'amp' ),
+						$status_code,
+						esc_html( get_status_header_desc( $status_code ) )
```
westonruter: The `esc_html` here is probably not right.

includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -119,28 +119,52 @@ public function sanitize() {
...
-		$form->appendChild( $div );
+		foreach ( $elements as $attribute => $element ) {
+			if ( $element ) {
+				continue;
+			}
+			$div      = $this->dom->createElement( 'div' );
+			$template = $this->dom->createElement( 'template' );
+			$div->setAttribute( 'class', 'amp-wp-default-form-message' );
+			if ( 'submitting' === $attribute ) {
+				$mustache = $this->dom->createTextNode( esc_html__( 'Submitting…', 'amp' ) );
```
westonruter: The `esc_html__` here should just be `__`

swissspidy:
> Should status codes be indicated in the generic responses?

I don't think they're helpful for regular users if added to prominently. Maybe at the end in parentheses? e.g. `Error: XY (Error Code: 404)` or something like that

> Is the default styling for the form messages OK?

Looks reasonable and easy to override. Contrast is good too, although it relies on the theme's text color to be dark. Maybe specify color too?

westonruter:
> I don't think they're helpful for regular users if added to prominently. Maybe at the end in parentheses? e.g. `Error: XY (Error Code: 404)` or something like that

A concern I have is the `XY` isn't going to be super helpful either, as this is what `get_status_header_desc()` returns. Also, even though an 200 response is sent, a plugin or core could be sending it erroneously (IMHO) when a 400+ code is expected. 

That being the case, in the case of a 200 response, what do you think of something like:

“It appears your submission was successful.” 

(Still not very comforting!) But then include a `details` element that explains the underlying code, and a message for developers guiding them how to provide a better message, which would be sending a JSON object with a `message` property when `wp_is_json_request()`, or else to do a `wp_redirect()` to a success page.

This would be similar for a 400+ status code, except the user message could actually include the status code description (perhaps):

“Your submission failed: Bad Request.”

This could then include the same `details` element as outlined above.

Thoughts?

swissspidy:
Yeah that could work 👍

includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -92,11 +92,11 @@ public function sanitize() {
 			} elseif ( 'post' === $method ) {
 				$node->removeAttribute( 'action' );
 				if ( ! $xhr_action ) {
-					// record that action was converted tp action-xhr.
-					$action_url = add_query_arg( '_wp_amp_action_xhr_converted', 1, $action_url );
+					// Record that action was converted tp action-xhr.
```
amedina: Nit: typo 'tp'

westonruter:
Here's what I have at the moment:

image

image

includes/class-amp-theme-support.php
```diff
@@ -1705,11 +1694,36 @@ public static function prepare_response( $response, $args = array() ) {
...
+		);
+		if ( $is_form_submission && null === json_decode( $response ) && json_last_error() && ( ( $status_code >= 200 && $status_code < 300 ) || $status_code >= 400 ) ) {
+			return wp_json_encode(
+				array(
+					'status_code' => $status_code,
+					'status_text' => get_status_header_desc( $status_code ),
+				)
+			);
+		}
+
```
westonruter: This warrants a test.

includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -119,28 +122,81 @@ public function sanitize() {
...
+					$reason = __( 'Even though the server responded OK, it is possible the submission was not processed.' );
+				}
+				$reason .= ' ' . __( 'Please contact the developer of this form processor to improve this message.', 'amp' );
+
+				$p->appendChild( $this->dom->createTextNode( ' ' ) );
+				$small = $this->dom->createElement( 'small' );
+				$small->appendChild( $this->dom->createTextNode( $reason ) );
+				$small->appendChild( $this->dom->createTextNode( ' ' ) );
+				$link = $this->dom->createElement( 'a' );
+				$link->setAttribute( 'href', 'https://amp-wp.org/?p=5463' );
```
swissspidy: Just noting that this page hasn't been published yet.

westonruter: Correct. Shortlink to redirect to published page once released.

PR #2433, opened 2019-05-27
Apply animation to the outer wrapper and rotation to inner element

### LINKED ISSUE #2302: Animation - Fly In Top - Starting Position
Noted that If I select Fly In Left/Right/Bottom, the text is not visible at all and then "animates" into position. For Fly In Top, text is already appearing in display before animation starts:

image

miina:
Fixes #2302.

### DIFF
--- a/assets/src/stories-editor/helpers/index.js
+++ b/assets/src/stories-editor/helpers/index.js
@@ -389,20 +389,13 @@ export const addAMPExtraProps = ( props, blockType, attributes ) => {
 	// Always add anchor ID regardless of block support. Needed for animations.
 	newProps.id = attributes.anchor || uuid();
 
-	if ( attributes.ampAnimationType ) {
-		ampAttributes[ 'animate-in' ] = attributes.ampAnimationType;
-
-		if ( attributes.ampAnimationDelay ) {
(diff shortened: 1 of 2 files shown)


assets/src/stories-editor/helpers/index.js
```diff
@@ -502,9 +494,28 @@ export const wrapBlocksInGridLayer = ( element, blockType, attributes ) => {
 		};
 	}
 
+	const animationAtts = {};
+
+	// Add animation if necessary.
+	if ( ampAnimationType ) {
+		animationAtts[ 'animate-in' ] = ampAnimationType;
```
swissspidy: This doesn't seem to work reliably.

Everything is fine when editing a block and previewing it on the front end, but upon refreshing the page I get block validation errors because of the missing attributes:

Screenshot 2019-05-27 at 15 48 23

miina: Ugh, I had forgotten to update the attribute selectors. Should work now.

kienstra:
**Looks Good**

Hi @miina,
Sorry for the delay, but this looks good and works as expected.

"Fly In Top" animations now appear from the top:

fly-in-test

PR #2443, opened 2019-05-28
Ensure CTA block being last

### LINKED ISSUE #2358: Enforce last position for CTA block
As discussed in https://github.com/ampproject/amp-wp/issues/1974#issuecomment-494309338, the CTA is forced to always be the last layer on page on the frontend already. In the editor, this is only partially true.

We need to:

* [ ] Remove block position controls for CTA block
* [ ] Ensure last position in the editor
  * Use some z-index trickery so that the CTA is always on top of the other blocks
  * When moving around other blocks up and down, make sure they take the CTA block into account
  * Subscribe to content changes to move CTA block to the end in case the order gets messed up

miina:
Fixes #2358.

- Enforces CTA block to be the last on the page.
- Removes block position controls from the CTA block.


assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -47,6 +52,10 @@ const TEMPLATE = [
 ];
 
 class EditPage extends Component {
+	shouldComponentUpdate() {
```
miina: Added the check for the CTA block to the Page block instead. Thought that perhaps it's better to just check the relevant page changes instead of going through all the blocks in the subscribe.

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -371,25 +407,37 @@ class EditPage extends Component {
...
-	};
-} )( EditPage );
+export default compose(
+	withSelect( ( select, { clientId, attributes } ) => {
+		const { getMedia } = select( 'core' );
+		const { getBlockOrder, getBlocksByClientId, getBlockRootClientId, getBlock } = select( 'core/block-editor' );
+
+		const innerBlocks = getBlocksByClientId( getBlockOrder( clientId ) );
+		const isFirstPage = getBlockOrder().indexOf( clientId ) === 0;
+		const isCallToActionAllowed = ! isFirstPage && ! innerBlocks.some( ( { name } ) => name === 'amp/amp-story-cta' );
```
swissspidy: Since this second check here does the same as `pageHasCTABlock` and is more robust (does not only check the last item in the list), let's use that logic for `pageHasCTABlock`.

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -141,6 +150,33 @@ class EditPage extends Component {
...
+			clientId,
+		} = this.props;
+		const order = getBlockOrder( clientId );
+		if ( 1 >= order.length ) {
+			return;
+		}
+		let ctaClientID = false;
+		order.forEach( ( childId ) => {
+			const block = getBlock( childId );
+			if ( 'amp/amp-story-cta' === block.name ) {
```
swissspidy: Speaking of the `innerBlocks.some....` check and `pageHasCTABlock`,  let's use that logic for this part here too. A function that could work for all these areas could be as simple as:

```js
const getCallToActionBlock( pageClientId ) => {
  const innerBlocks = getBlocksByClientId( getBlockOrder( pageClientId ) );
  return innerBlocks.find( ( { name } ) => name === 'amp/amp-story-cta' );
}
```

We could even use that function in `withCallToActionValidation`.

assets/src/stories-editor/blocks/amp-story-page/edit.js
miina: For determining `isLast` we still need to check if the CTA block is in the last position. Since that's needed for `isLast` only, I'll add this as a separate check then.

miina: Or well, it actually shouldn't need a separate check.

miina:
Thanks for reviewing, @swissspidy, added the changes.

miina:
Will add a test as well and then should be ready for re-review.

PR #2459, opened 2019-05-29
Improve behavior for selected blocks

swissspidy:
* When selecting or rotating a block, it will always be the first and foremost block (z-index change).  
  This makes it easier to quickly edit a block without having to temporarily move it to the foreground for editing.
* Increases the area surrounding the rotation handle.  
  Makes it easier to move the cursor to the handle, without accidentally leaving the block container.
* Removes the thick left border for selected block.  
  This specific Gutenberg styling does not really apply to our movable blocks.

**Screenshots:**

Screenshot 2019-05-29 at 13 23 07
Screenshot 2019-05-29 at 13 23 10
Screenshot 2019-05-29 at 13 44 26

### DIFF
--- a/assets/src/components/rotatable-box/edit.css
+++ b/assets/src/components/rotatable-box/edit.css
@@ -16,13 +16,23 @@
 }
 
 .block-editor-block-list__block .rotatable-box-wrap {
-	box-shadow: -1px 0 rgba(66, 88, 99, .4);
-	width: 0;
-	height: 25px;
 	position: absolute;
-	bottom: -40px;
-	margin-left: 50%;
+	right: -14px;
+	bottom: -25px;
+	left: calc(50% - 15px);
+	width: 30px;
+	height: 25px;
+}
+
+.block-editor-block-list__block .rotatable-box-wrap:before {
+	display: block;
+	content: "";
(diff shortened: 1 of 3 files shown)


swissspidy:
Yeah that's a good point.

@westonruter your thoughts on that?

westonruter:
> The only thing that is now a bit more confusing is changing the order of the blocks from the sidebar since the selected block is always on top.

I think that is fine. I didn't find it particularly confusing. It makes sense that a currently-edited block would be “forced” to top temporarily.

However, I'm noticing some strange behavior when selecting and editing the block in behind. It comes forward, but only when I keep wiggling the mouse over it: https://youtu.be/DxNoDyv-aCo

This is with a fresh build of Gutenberg 5.8.

swissspidy:
Good catch! Fixed this in 0e7a87432cf8b15623779a837449b290be22e975

swissspidy:
**Testing instructions**

1. Create new story
2. Create a few blocks, ideally with some background colors to more easily distinguish them
3. Verify that the currently selected / edited block is always shown on top of the other blocks

csossi:
Verified in QA

PR #2476, opened 2019-05-30
Fix TwentyTwelve nav menu and add support for TwentyThirteen core theme

### LINKED ISSUE #1742: Add AMP compatibility for older core themes: 2010, 2011, 2012, 2013, and 2014

amedina:
Fixes #1742.

### DIFF
--- a/includes/sanitizers/class-amp-core-theme-sanitizer.php
+++ b/includes/sanitizers/class-amp-core-theme-sanitizer.php
@@ -1105,6 +1105,13 @@ class AMP_Core_Theme_Sanitizer extends AMP_Base_Sanitizer {
 						.dropdown-toggle {
 							display: none;
 						}
+						ul.nav-menu :focus-within > ul,
+						.nav-menu :focus-within > ul {
+							clip: inherit;
+							overflow: inherit;
+							height: inherit;
+							width: inherit;
+						}
 					}
 					@media (max-width: 643px) {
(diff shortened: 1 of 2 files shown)


amedina:
> I think there is one (or two) things that are left for Twenty Thirteen, and that is for expanding nested submenus

Yeah; will add.  Also Masonry and the other functionality in `functions.js` needs to be analyzed.

amedina:
@westonruter if Masonry is just avoided, the footer content would just be displayed on flexible grid?

westonruter:
The expanding nested submenus will be a bit tricky because I don't think the behavior in non-AMP should be followed, where clicking a parent item causes the submenu to appear, with a second parent click then causing the navigation to be followed. It would be better to go ahead and implement the navigation sub-menu buttons: https://amp-wp.org/documentation/playbooks/navigation-sub-menu-buttons/

This would not be 100% visual parity, but it would be an alternative approach, which I think is the defacto best practice now anyway. We'd need to add some styling.

>  if Masonry is just avoided, the footer content would just be displayed on flexible grid?

Can CSS grid do that?

PR #2478, opened 2019-05-31
Add color options to CTA block

### LINKED ISSUE #2472: CTA block's button missing color settings
To reproduce:
- Add CTA block
- Try to change the background color and text color.

swissspidy:
Fixes #2472.

* Adds background color, text color and opacity controls for CTA block.
* Improves preview of CTA block in template inserter  
  Block should be centered horizontally
* Improves template previews by enqueueing necessary Google Fonts upon loading


assets/src/components/template-inserter/index.js
```diff
@@ -162,8 +169,15 @@ export default compose(
...
+		const reusableBlocks = getReusableBlocks();
+
+		const isStoryBlock = ( clientId ) => {
+			const block = getBlock( clientId );
+			return block && 'amp/amp-story-page' === block.name;
+		};
+
 		return {
-			reusableBlocks: getReusableBlocks(),
+			reusableBlocks: reusableBlocks.filter( ( { clientId } ) => isStoryBlock( clientId ) ),
```
miina: Since now we're passing through only the filtered blocks already, then perhaps we could name it something else than `reusableBlocks` for clarity?

miina:
Some notes:
- The CTA block doesn't seem to be showing up in the inserter anymore:
Screen Shot 2019-05-31 at 3 07 22 PM

- There is a React warning in the console, perhaps it was before already:
Screen Shot 2019-05-31 at 3 07 28 PM

- The background opacity doesn't seem to affect the default color.

swissspidy:
Very odd 🤔 

> The CTA block doesn't seem to be showing up in the inserter anymore

That might be a result of that React warning in the console. However, I do not get this warning in my environment (latest Gutenberg plugin in development mode) and I still see the CTA block just fine in the inserter.

> The background opacity doesn't seem to affect the default color.

What do you mean? Opacity is not being applied at all?

Here it works just fine, no matter which color I choose...

assets/src/components/template-inserter/index.js
swissspidy: Done it 1b89e313515bb38205e801a7fe1bb8c09af9408f

miina:
> That might be a result of that React warning in the console. However, I do not get this warning in my environment (latest Gutenberg plugin in development mode) and I still see the CTA block just fine in the inserter.

I just updated the Gutenberg and now I'm also not seeing the warning, and the CTA block shows up as expected 👍 

> What do you mean? Opacity is not being applied at all?

That's in case of the default settings only. To reproduce:
- Add CTA block
- Don't change any settings and apply Opacity.

Since it has a background color by default then it feels like it should change the opacity. Thoughts?

swissspidy:
> Since it has a background color by default then it feels like it should change the opacity. Thoughts?

Gotcha. Makes sense. Should be trivial to fix.

swissspidy:
@miina Please test again 🙂

PR #2481, opened 2019-05-31
Keep rotation for skeleton template. Add vertical template.

### LINKED ISSUE #2471: Consider keeping rotation for the skeleton template
AC: Rotation is kept after inserting a template.
AC: `travel-vertical` template is available under Template Inserter.

If rotation is not kept then there is a significant difference in the looks of the template and the skeleton template.

Screen Shot 2019-05-30 at 2 49 12 PM

Screen Shot 2019-05-30 at 2 49 37 PM

miina:
Fixes #2471.


miina:
Thanks @swissspidy, fixed in c74ed07.

Issue #2380, opened 2019-05-22
Review and potentially improve block transforms


swissspidy:
As discussed in #2365, it would be good to verify existing block transforms and add some where needed.

For example, there are currently no transforms for our custom text block. However, in many cases blocks with transforms from/to the core paragraph block could also work for ours.

Issue #2376, opened 2019-05-22
Adding animation to Author block break validation


loneken79:
Using version 1.2-Beta1

Step to reproduce :
- Make a new Story 
- add a Author bloc
- (save story draft and reload page => everything is OK)
- add animation to Author bloc (fly from bottom)
- save story and reload page 
=> Author Block validation is now broken and validation proposal is inchoherent

![melic photo_wp-admin_post php_post=38046 action=edit](https://user-images.githubusercontent.com/9193295/58164720-a360c200-7c86-11e9-93ce-b77f05b2963f.png)

swissspidy:
Hi @loneken79, thanks a lot for testing our AMP Stories integration!

While I personally have not run into this issue before, it looks like @westonruter got the same issues earlier: see #2370. And now thanks to your steps to reproduce it I can also confirm the issue.

We're going to try to solve this as soon as possible and will let you know when a fix is available.

swissspidy:
I am closing this as a duplicate of #2370, just so we can keep discussion in one place.

@loneken79 Please follow that other issue for any updates. Again, we'll try to ping you whenever we have a fix ready for testing.

PR #2484, opened 2019-05-31
Prevent double wrapping elements

### LINKED ISSUE #2370: Wrapper elements causes problem for conversion in Block Resolving for Stories
I made a story page as depicted in #2368 and then when I tried switching to the branch, I got the expected:

image

When clicking “Resolve” the suggested resolution is not correct:

Screen Shot 2019-05-21 at 13 47 20

The `amp-story-grid-layer` and the `amp-story-block-wrapper` are incorrectly getting added a second time.

Fixing this will be very important once stories are in a stable release, in addition to making sure that we start maintaining `deprecated` conversions.

swissspidy:
Fixes #2370.


miina:
@swissspidy I'm still seeing the issue when Resolving the validation error, it still seems to add two warppers:
Screen Shot 2019-05-31 at 5 23 26 PM

swissspidy:
Now the duplicated wrapper is gone, but the div is still there:

Screenshot 2019-06-03 at 10 42 49

swissspidy:
Not sure where `getSaveElement()` in Gutenberg gets this extra `div` from. @miina Perhaps you have an idea off the top of your head? Otherwise I'll try to do some more debugging later.

miina:
@swissspidy I'll look into this a little bit and let you know if I find the reason.

miina:
@swissspidy I think I've found the issue. It looks like block transformation is taking place to display the "correct" HTML and since we don't have `amp-story-grid-layer` in the Text block's transformation then it assumes the block is `core/html` and converts it accordingly, adding a `div` wrapper (See `htmlToBlocks` in the `blocks` package).

I added the `amp-story-grid-layer` to the Text block's transformations, however, this would work only for the Text block and all the other blocks are still converting incorrectly (now with the new addition trying to convert to a Text block, based on the `amp-story-grid-layer` selector in the transformations). Ideally, we would pass the HTML for conversion without the wrappers ... Not sure if it's possible.

The incorrect conversion is happening within `BlockCompare` in the `render` method (`const converted = this.getConvertedContent( convertor( block ) );`.

Thoughts?

swissspidy:
> It looks like block transformation is taking place to display the "correct" HTML and since we don't have `amp-story-grid-layer` in the Text block's transformation then it assumes the block is `core/html`

When we just change `save()` without doing proper deprecation, then treating it as HTML makes sense to me.

> now with the new addition trying to convert to a Text block, based on the `amp-story-grid-layer` selector in the transformations

Makes sense too, because there now is a transform. But an invalid title block or so should be converted to HTML instead, so I'd say cabc633dd4209cc303063efb57c3b825f1088472 needs to be reverted as it's unneeded.

swissspidy:
Thinking about it again, everything in #2370 sounds as expected behavior. 🤔 

Gutenberg sees an invalid block, converts it to HTML, and then adds the wrapper elements to the HTML block because we told it to do so.**\***

All the examples in #2370 are about intentionally breaking block markup (without doing deprecation), which of course leads to block invalidation.

IMO, as long as we haven't released 1.2 stable, this is OK. But after that we need to do proper block deprecation. Learn all about it here: https://developer.wordpress.org/block-editor/developers/block-api/block-deprecation/

#2376 on the other hand is more interesting, because there we apparently seem to unintentionally break block markup, and the double wrapping is just a symptom of the issue. However, I was unable to reproduce that error there.

----

**\*** If anything, instead of bailing early like in 077df52af14fecd8892c08d5d8c05851b390cb6f, we should probably rather remove any inner `<amp-story-grid-layer>` elements in HTML block.

miina:
> Makes sense too, because there now is a transform. But an invalid title block or so should be converted to HTML instead, so I'd say cabc633 needs to be reverted as it's unneeded.

Well, yes, it does make sense that it acts this way, I agree that cabc633 needs to be reverted, I added that to show the findings from debugging. However, the unnecessary `div` is added due to unfound transformation.

Our `amp-story-grid-wrapper` interferes with the default conversion and instead of converting to the correct block (as it does by default) it always converts to `core/html` resulting in the extra wrapper `div`.

When block attributes are incorrect then normally it doesn't result converting into a HTML block, it results staying the origin block type. A block is converted into a HTML block only if a matching block is not found -- and that's what our wrappers result in.

To fix this we would either need the block source to be converted without the wrapper elements (the better option) or adding a transformation for all the blocks, which isn't a good option. 

Do you have any other ideas?

miina:
> Gutenberg sees an invalid block, converts it to HTML, and then adds the wrapper elements to the HTML block because we told it to do so.*

But it doesn't usually convert all the invalid blocks to HTML, it only converts if a block type wasn't found. E.g. if you test in the deafult editor and create a validation error to any block then you can see that it still keeps the origin block type, it never converts it to HTML block (unless a block type wasn't found at all)

swissspidy:
> When block attributes are incorrect then normally it doesn't result converting into a HTML block, it results staying the origin block type. A block is converted into a HTML block only if a matching block is not found -- and that's what our wrappers result in.

Got it!

> To fix this we would either need the block source to be converted without the wrapper elements (the better option)

^ Agreed. Now, let's find out how we can do this 🙃

miina:
> * If anything, instead of bailing early like in 077df52, we should probably rather remove any inner <amp-story-grid-layer> elements in HTML block.

Yep, this might be it -- if it already has `amp-story-grid-layer` then it's for the conversion (and not a "usual" save) and removing the wrappers should fix the issue.

swissspidy:
Maybe I am missing something but it seems like at that point it's already too late to have any effect on block conversion 🤔 Removing wrappers in `wrapBlocksInGridLayer`, no matter if by removing inner elements or not adding the second wrapper, doesn't make any difference.

What `BlockCompare.getConvertedContent` deals with is the already converted block as determined by the parser. The markup doesn't match and there's also no transform - thus there's an HTML block.

> if you test in the deafult editor and create a validation error to any block then you can see that it still keeps the origin block type,

Just tested this with the paragraph block and I think this is only because it has a raw transform for `<p>`, that's why it works. Probably the same reason for other blocks too.

> To fix this we would either need the block source to be converted without the wrapper elements (the better option) or adding a transformation for all the blocks, which isn't a good option.

So it seems like adding transforms is actually the only solution here (related: #2380)...?

miina:
> Maybe I am missing something but it seems like at that point it's already too late to have any effect on block conversion 🤔 

Looks like you're right, the transforms seems to be the only option.

swissspidy:
At that point I lean towards closing #2370 as "working as expected at the moment" and having a closer look at transforms as part of #2380.

@westonruter thoughts?

westonruter:
Is closing #2370 the right approach? It is a bug. The underlying mechanism is irrelevant for the user experience, no?

amedina:
Let's aim at fixing this.

swissspidy:
I didn't mean to intend saying not to fix this. It's just going to be quite complex as we have to add dedicated transforms to _all_ blocks that take the malformed HTML and try to create a new block based on that. And we have #2380 for dealing with transforms already where can have a closer look at that. But I guess we can leave these separated then 🤷‍♂

miina:
> It's just going to be quite complex as we have to add dedicated transforms to all blocks that take the malformed HTML and try to create a new block based on that.

It might get resolved by filtering all the block settings and adding quite basic `type: 'raw',` transform, maybe the transform method logic can even be identical for all the blocks.

I'm wondering if we might need to add another class name to the `amp-story-grid-layer` for separating the selectors for each block for the transform, we probably can't just assign `amp-story-grid-layer` as a selector for each block. If we'll need to add a class name per each block for the purposes of the selector then this change would cause block validation errors and should definitely get resolved before releasing. Thoughts?

swissspidy:
That could be an interesting solution! I'll have a look.

>  If we'll need to add a class name per each block for the purposes of the selector then this change would cause block validation errors and should definitely get resolved before releasing. Thoughts?

IIRC the block validator doesn't cause hard validation errors for `data-*` attributes. Which would make more sense than a class name anyway.

swissspidy:
@miina I just pushed some changes. Is that more or less what you had in mind?

miina:
> @miina I just pushed some changes. Is that more or less what you had in mind?

Yes :) I can still see the validation errors in the templates though, looks like it's showing it even with the `data-*` attribute. Are you also seeing that?

swissspidy:
Just found the code about the `data` attributes again: https://github.com/WordPress/gutenberg/blob/5bbda3656a530616a7a78c0a101d6ec2d8fa6a7a/packages/blocks/src/api/validation.js#L315-L324

I guess I was wrong, so the templates would need updating either way in that case.

swissspidy:
@westonruter Wanna review this one?

westonruter:
FYI: Merge conflict is happening (with great irony).

swissspidy:
Oh, yeah because of #2535

swissspidy:
Thanks @kienstra!

> Though I'm not sure of the expected behavior. Maybe the "This block contains unexpected or invalid content" box shouldn't display, or I should have tested this differently.
> 
> To test this, I pulled this PR's branch, and followed @westonruter's 5 test steps.
> 
> For the Text block, clicking 'Resolve' and 'Convert to blocks' seemed to work well. The new Text block looked the same. 

The difference is that without this PR, it would not create a new text block but instead an HTML block with excessive markup.

kienstra:
Hi @swissspidy,
>The difference is that without this PR, it would not create a new text block but instead an HTML block with excessive markup.

Sounds good. I also saw that without this PR, it created a new HTML block with markup.

This PR helps, as clicking 'Resolve' and 'Convert to blocks' creates a proper Text block.

swissspidy:
Cancelled the optional jobs on Travis

PR #2486, opened 2019-05-31
[WIP] Amp-Stories Fit Text - Swap flex vertical centering by classic inline-block style

### LINKED ISSUE #2342: Setting Text block as bold/italic causes unexpected shift
If I add a Text block and increase its height like so:

image

Upon selecting Bold or Italic, the text shifts up unexpectedly:

image

image

Removing the formatting causes the expected position to be restored.

mehigh:
Fixes #2342 
To Do:
@miina - can you add a amp-fit-text-wrapper around is-amp-fit-text to avoid applying some rules to some too generic of wrappers?

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-text/edit.css
+++ b/assets/src/stories-editor/blocks/amp-story-text/edit.css
@@ -51,12 +51,26 @@ div[data-type="amp/amp-story-page"] .block-editor-inner-blocks .block-editor-blo
 	display: inline-block;
 }
 
+.wp-block[data-type="amp/amp-story-text"] .wp-block-amp-story-text .amp-fit-text-wrapper {
+	position: relative;
+	line-height: 350px;
+}
+
 .wp-block[data-type="amp/amp-story-text"] .wp-block-amp-story-text .is-amp-fit-text {
-	line-height: 1.15;
-	display: flex;
-	flex-direction: column;
-	flex-wrap: nowrap;
-	justify-content: center;
+	display: inline-block;
+	vertical-align: middle;
+	line-height: normal;
+}
+
+.wp-block[data-type="amp/amp-story-text"] .wp-block-amp-story-text .is-amp-fit-text::before {
+	content: '';
+	position: absolute;
+	top: 0;
+	left: 0;
+	width: 100%;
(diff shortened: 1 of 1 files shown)


assets/src/stories-editor/blocks/amp-story-text/edit.css
```diff
@@ -51,12 +51,26 @@ div[data-type="amp/amp-story-page"] .block-editor-inner-blocks .block-editor-blo
 	display: inline-block;
 }
 
+.wp-block[data-type="amp/amp-story-text"] .wp-block-amp-story-text .amp-fit-text-wrapper {
+	position: relative;
+	line-height: 350px;
```
miina: Would this work in case of multiple line texts and independent of the height of the Text block itself? The Text block could also have a height of 100px for example and have 2 or more rows of text.

Does the line-height have to be dynamic here based on the Text block height? We could add that together with adding the wrapper in the editor.

mehigh: @miina I didn't consider that use-case, so it's hard-coded it in here. If we can write the inline style with the line-height attribute in there instead, that would be great and function correctly. 

So if this needs to be half-the-height of the 350px box, the wrapper would receive something like this .. `<div class="amp-fit-text-wrapper" style="line-height: 175px;">...`

The line-height is just for the wrapper, as the actual amp-fit-text receives a line-height: normal; so it doesn't matter how many rows of text are inside. 

Does this help? Or do I need to look at a different angle to fix this?

miina: Thanks, @mehigh, we can write the inline style dynamically based on the height of the wrapper, your solution should work. 👍 Will let you know if there should be any more questions.

mehigh: @miina 
I've removed the need of an extra wrapper assuming we're using some inline line-height styles on the .wp-block-amp-story-text element. I've moved the position: relative on that element instead, as we need it on a close wrapper to allow the background to extend to the full dimensions.

westonruter:
This PR was apparently closed because the commits were moved to another PR: #2494.

PR #2488, opened 2019-06-01
Fix PHPCS exclude pattern

westonruter:
Not to be merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,10 +64,7 @@ jobs:
         - source "$DEV_LIB_PATH/travis.script.sh"
         - composer validate --no-check-all
         - composer normalize --dry-run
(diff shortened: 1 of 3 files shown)


westonruter:
@swissspidy It seems the lint-checking job is not working.

swissspidy:
Oh, I think I see the issue. Will fix right away.

swissspidy:
OK, the issue wasn't as simple as I thought 🙃 

But it is fixed now! See https://travis-ci.org/ampproject/amp-wp/jobs/540032107 for proof.

Since the Travis build is running within `/home/travis/build/ampproject/amp-wp`, the `build/*` exclude pattern was applied - wrongly excluding all plugin files 😅

Issue #2430, opened 2019-05-27
Make it easier to advertise in AMP Stories


swissspidy:
AMP Stories supports an `amp-story-auto-ads` component that works like this:

```html
<amp-story>
  <amp-story-auto-ads>
    <script type="application/json">
      {
        "ad-attributes": {
          // ad server configuration
        }
      }
    </script>
  </amp-story-auto-ads>
  <amp-story-page>
  ...
</amp-story>
```

([source](https://amp.dev/documentation/guides-and-tutorials/develop/advertise_amp_stories))

Right now, the only way to insert that component is by filtering `the_content`:

https://github.com/ampproject/amp-wp/blob/1aa2afaec26781a953ca7f7442fc21973e4ce243/includes/templates/single-amp_story.php#L41-L55

I wonder if we should add a dedicated action to the template there. Perhaps we could even add a `amp_stories_auto_ads_attributes` filter and automatically print `<amp-story-auto-ads>` when the attributes are not empty.

westonruter:
I was chatting about this with @amedina last week as well. I also suggested a filter which would result in the AMP component printed when the array is not empty. 

The alternative we chatted about was to add a new ads section to the Document sidebar in the editor, but that seems very premature. The configuration would need a UI and it seems likely that the same configuration would be used across stories anyway. And we'd want an ad-provider plugin to actually inject the ads and manage the UI.

So +1 for filter.

amedina:
Sounds good; +1 for filter. 

Chatted with @jasti a bit about Ads on Stories. Should sync further to discuss enabling strategies/recommended approaches for monetization. /cc @vinaymahag

swissspidy:
> The alternative we chatted about was to add a new ads section to the Document sidebar in the editor, but that seems very premature. The configuration would need a UI and it seems likely that the same configuration would be used across stories anyway. And we'd want an ad-provider plugin to actually inject the ads and manage the UI.

Makes sense. It seems that there are a few configurations that would need to be exposed in the UI. One example would be for toggling the `next-page-no-ad` attribute on an `amp-story-page`.

amedina:
@jasti WDYT about this as a baseline?

jasti:
Yeah, there are a lot of options we can fiddle with. Lets discuss if this should all be bundled as part of the Site Kit Package or ship with the AMP plugin in the next version.

swissspidy:
> The alternative we chatted about was to add a new ads section to the Document sidebar in the editor

[...]

PR #2492, opened 2019-06-03
Add new filter for amp-story-auto-ads

swissspidy:
See #2430.

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -1130,3 +1130,34 @@ function amp_add_admin_bar_view_link( $wp_admin_bar ) {
 
 	$wp_admin_bar->add_menu( $parent );
 }
+
+/**
+ * Prints AMP Stories auto ads.
+ *
+ * @since 1.2
+ */
+function amp_print_story_auto_ads() {
+	/**
+	 * Filters the configuration data for <amp-story-auto-ads>
+	 *
+	 * This allows Dynamically inserting ads into a story.
+	 *
+	 * @param array   $data Story ads configuration data.
+	 * @param WP_Post $post The current story's post object.
+	 */
+	$data = apply_filters( 'amp_stories_auto_ads_configuration', array(), get_post() );
+
+	if ( empty( $data ) ) {
+		return;
+	}
+
+	$script_element = AMP_HTML_Utils::build_tag(
+		'script',
+		array(
+			'type' => 'application/json',
+		),
+		wp_json_encode( $data )
+	);
+
+	echo AMP_HTML_Utils::build_tag( 'amp-story-auto-ads', array(), $script_element ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+}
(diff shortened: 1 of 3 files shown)


includes/templates/single-amp_story.php
```diff
@@ -62,12 +62,17 @@
...
+			amp_print_story_auto_ads();
+			the_content();
+			?>
 		</amp-story>
 
 		<?php
 		// Note that \AMP_Story_Post_Type::filter_frontend_print_styles_array() will limit which styles are printed.
 		print_late_styles();
+
+		amp_print_analytics( '' );
```
westonruter: Good catch!

includes/amp-helper-functions.php
```diff
@@ -1130,3 +1130,34 @@ function amp_add_admin_bar_view_link( $wp_admin_bar ) {
...
+function amp_print_story_auto_ads() {
+	/**
+	 * Filters the configuration data for <amp-story-auto-ads>
+	 *
+	 * This allows Dynamically inserting ads into a story.
+	 *
+	 * @param array   $data Story ads configuration data.
+	 * @param WP_Post $post The current story's post object.
+	 */
+	$data = apply_filters( 'amp_stories_auto_ads_configuration', array(), get_post() );
```
westonruter: I think this should be `amp_story` instead of `amp_stories`, following example in #2468.

It's the filter for a single story since `get_post()` is passed in.

includes/templates/single-amp_story.php
```diff
@@ -62,12 +62,17 @@
 				poster-landscape-src="<?php echo esc_url( $poster_landscape ); ?>"
 			<?php endif; ?>
 		>
-			<?php the_content(); ?>
+			<?php
+			amp_print_story_auto_ads();
```
westonruter: Did you consider injecting the story auto-ads via a `the_content` filter? I don't know if it is particularly better to use a filter, but it came to mind.

Also, something else which may need to get filtered is the injection of a `next-page-no-ad` attribute on the `amp-story-page` components: https://github.com/ampproject/amphtml/blob/master/extensions/amp-story-auto-ads/amp-story-auto-ads.md#insertion-control

I suppose the normal `render_block` filter should be used for that?

swissspidy: > Did you consider injecting the story auto-ads via a the_content filter?

Not really. I figured there'd be too many situations where this wouldn't need to be printed (feeds, embeds, archives, etc.). So this was more straightforward.

However, I don't have a strong opinion on what's better here.

> Also, something else which may need to get filtered is the injection of a next-page-no-ad attribute on the amp-story-page components:

Yeah I mentioned that attribute in https://github.com/ampproject/amp-wp/issues/2430#issuecomment-498212546. A `render_block` filter for that sounds a bit hacky but probably the easiest way for a plugin to change the markup without causing block invalidation.

However, there would still need to be some UI to choose the pages where this should be added. So a plugin would use the `blocks.registerBlockType` filter to add a new attribute for this feature and expose a UI toggle.

Alternatively, we just add this attribute to the page block ourselves and save the block with the proper HTML attribute if it's set. However, there wouldn't be a UI by default. A plugin would be responsible for adding a toggle.

PR #2494, opened 2019-06-03
Fix italic and bold for amp-fit-text

### LINKED ISSUE #2342: Setting Text block as bold/italic causes unexpected shift

miina:
Fixes #2342.

- Adjust style so that using italic and bold doesn't break the text into multiple lines in case of using amp-fit-text (props @mehigh).
- Adjust resizing and class names accordingly.


mehigh:
👍  I'm OK to contribute this.

@miina I've sent a PR against this branch from our fork as I don't have push access to this repo:
https://github.com/ampproject/amp-wp/pull/2499

swissspidy:
Two things I just spotted:

When creating a new story, the placeholder's position of the initial text block is broken:

Screenshot 2019-06-04 at 14 43 31

Second, block styles don't work anymore when fit-text is on. See frontend vs. editor comparison here:

Screenshot 2019-06-04 at 14 45 35
Screenshot 2019-06-04 at 14 45 38

swissspidy:
When I turn off amp-fit-text on above's block, it looks like this on the frontend:

Screenshot 2019-06-04 at 14 48 03

That's because of the `.amp-story-block-wrapper > p` selector declaring `display: flex`.

swissspidy:
NB: `npm run lint:css:fix` should be able to fix the stylelint errors.

miina:
Thanks for testing, @swissspidy. The issues should be fixed now!

Props @mehigh for e9b67d3.

swissspidy:
> The bad news is that it appears that one or more commits were authored or co-authored by someone other than the pull request submitter. We need to confirm that all authors are ok with their commits being contributed to this project. Please have them confirm that here in the pull request.

Looks like that refers to the commit by @mehigh. Can you confirm that? 🙂 And perhaps double check whether you signed that CLA.

miina:
> Looks like that refers to the commit by @mehigh. Can you confirm that? 🙂 And perhaps double check whether you signed that CLA.

Yes, confirmed that it's Mike's commits and he already confirmed being OK contributing to the PR: https://github.com/ampproject/amp-wp/pull/2494#issuecomment-498421196

Also, we both have CLA-s signed. Otherwise, all my contributions would complain :D

swissspidy:
Ah, missed that comment. All good then 🙂

PR #2504, opened 2019-06-04
Fix font size oddities

### LINKED ISSUE #2366: Text font size is sometimes incorrect
QA noticed an issue which doesn't occur 100% of times and that doesn't necessarily have a way how to faithfully reproduce it, however, here are the steps where it sometimes happens:

- Add a Text block
- Make sure that amp-fit-text is on.
- Resize the block little by little
- In some cases, the font size displays very small.

image (1)

miina:
Fixes #2366.

- Uses the stored custom font size when switching off from amp-fit-text instead of assigning the amp-fit-text font size to text.
- Fixes amp-fit-text not allowing resizing smaller after it had been turned off previously.


miina:
@swissspidy Would you have time to review this today?

swissspidy:
Looking now!

PR #2505, opened 2019-06-04
Fix layout of Gallery as amp-carousel

westonruter:
Fixes https://wordpress.org/support/topic/problem-with-the-gallery-2/

The gallery was incorrectly using the `fixed-height` for the Gallery block and it wasn't obtaining the width/height properly. The Gallery shortcode was not being output with the dimensions of the images inside, but rather fixed dimensions for all embeds.

So this PR ensures that both the `gallery` shortcode and the Gallery block use the `responsive` layout, and that the `width`/`height` of the `amp-carousel` is the max `width`/`height` of the images contained therein.

The results are much improved:

# Before

image

image

# After

image

image

# Testing builds

* [amp.zip](https://github.com/ampproject/amp-wp/files/3254125/amp.zip) - 1.2-beta2-20190604T191520Z-f964e1ee
* [amp.zip](https://github.com/ampproject/amp-wp/files/3254136/amp.zip) - 1.1.4-alpha-20190604T191932Z-72acd5bb

### DIFF
--- a/includes/sanitizers/class-amp-gallery-block-sanitizer.php
+++ b/includes/sanitizers/class-amp-gallery-block-sanitizer.php
@@ -128,13 +128,16 @@ class AMP_Gallery_Block_Sanitizer extends AMP_Base_Sanitizer {
 				continue;
 			}
 
+			list( $width, $height ) = $this->get_carousel_dimensions( $node );
+
 			$amp_carousel = AMP_DOM_Utils::create_node(
 				$this->dom,
 				'amp-carousel',
 				array(
-					'height' => $this->get_carousel_height( $node ),
+					'width'  => $width,
+					'height' => $height,
 					'type'   => 'slides',
-					'layout' => 'fixed-height',
+					'layout' => 'responsive',
 				)
 			);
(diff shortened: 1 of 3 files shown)


includes/sanitizers/class-amp-gallery-block-sanitizer.php
```diff
@@ -150,15 +153,20 @@ public function sanitize() {
 	 * Get carousel height by containing images.
 	 *
 	 * @param DOMElement $element The UL element.
-	 * @return int Height.
+	 * @return array {
+	 *     Dimensions.
+	 *
+	 *     @type int $width  Width.
+	 *     @type int $heighy Height.
```
swissspidy: *height

westonruter:
Yeah, build marked as failure because of negligible negative change in coverage.

swissspidy:
> Yeah, build marked as failure because of negligible negative change in coverage.

@westonruter We can adjust the threshold for this in the codecov config. Maybe to something like 0.1% is already enough. Can open a PR.

swissspidy:
Created #2513

PR #2508, opened 2019-06-05
Always perform CSS tree shaking, regardless of total size

### LINKED ISSUE #2501: Eliminate CSS tree shaking from being optional
The AMP plugin settings screen has an option for whether or not to enable tree shaking:

image

This option is enabled by default:

https://github.com/ampproject/amp-wp/blob/b375d79c82ddb013a3842e1cf0a4b9fa7cfba7f7/includes/options/class-amp-options-manager.php#L30

And all core themes force it to be enabled:

https://github.com/ampproject/amp-wp/blob/b375d79c82ddb013a3842e1cf0a4b9fa7cfba7f7/includes/sanitizers/class-amp-core-theme-sanitizer.php#L202

When tree shaking was first introduced, the idea behind this option was to guard against scenarios where the tree shaker erroneously removed rules that actually applied to the page, specifically for themes that have been crafted to not have more than 50KB of CSS. The reality is that core + themes + plugins will almost _always_ generate more than 50KB, as evidenced by the above defaults. Also, there have been significant improvements to the tree shaker to prevent over-eager removal of rules, including:

* #2260: Account for escaped chars in CSS class names
* #2328: Account for toggleClass when obtaining used class names for tree shaking
* #2017: Prevent tree-shaking class names for amp-geo when component on page
* #1959: Fix tree shaking for dynamically-added class names
 
So since this option is not really needed anymore, I suggest we remove it entirely and _always_ perform tree shaking. Again, [decisions not options](https://wordpress.org/about/philosophy/).

westonruter:
Fixes #2501.

Remove a lot of code that is now unnecessary with removal of optional tree shaking, since it turns out that this option is obsolete.

### DIFF
--- a/includes/options/class-amp-options-manager.php
+++ b/includes/options/class-amp-options-manager.php
@@ -27,7 +27,6 @@ class AMP_Options_Manager {
 		'supported_post_types'     => array( 'post' ),
 		'analytics'                => array(),
 		'auto_accept_sanitization' => true,
-		'accept_tree_shaking'      => true,
 		'all_templates_supported'  => true,
 		'supported_templates'      => array( 'is_singular' ),
 		'enable_response_caching'  => true,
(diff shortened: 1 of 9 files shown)


westonruter:
🕺 Eliminating some technical debt!

amedina:
Test failing because of linter (?)

westonruter:
Indeed! Amended the commit to fix the `Generic.Formatting.MultipleStatementAlignment.NotSameWarning` issue.

PR #2509, opened 2019-06-05
Encourage lighter images and videos

### LINKED ISSUE #2332: Guide users to not use heavy assets when creating stories
The AMP Stories editor currently relies on users to only use pre-optimized assets (images and videos). This is not user-friendly. We should try to do this for them or else warn when we can't.

# Images

Currently the images that are added to a story can be much larger than necessary. For example, when selecting this image for background media:

image

This image is used as-is in the editor. This is a violation of the [`oversized-images` Feature Policy](https://www.chromestatus.com/feature/5953669526716416). It's true that the AMP Cache should resize the image down, but this doesn't help when viewing the image on origin. Instead of being 3024x4032 and 2MB, it should rather be 1080x1440 and ~301kB (with WordPress's default 82% JPEG compression quality).

One caveat here is that we need to eventually support landscape stories, so there will always be wasted bytes when selecting a landscape image and the focal point picker is used for selecting the right “crop” in portrait. At the very least, if the image gets a height of 1440 as opposed to 3024 then this will still be an  improvement.

So I suggest we `add_image_size()` for `amp_story_page`, with an unbounded width and a constrained height without cropping:

```php
add_image_size( 'amp_story_page', 99999, 1440, false );
```

Then when an image is picked from the media library, we should use this image size as opposed to the `full` size.

This should also apply when automatically supplying a video's featured image as its `poster`, or when selecting an image as the poster generally.

# Videos

Uploading videos taken from one's phone to the media library is a much bigger challenge, as WordPress has no native transcoding ability for videos. A 2-second video extracted from a Motion Photo taken on a Pixel 3 is 7MB:

image

The issue would be much worse if this wasn't just extracted from a Motion Photo. Imagine someone trying to use a 4K video at 60fps.

~The ideal here I think would be to allow users to embed control-less YouTube videos in stories, but this is not yet allowed. See https://github.com/ampproject/amphtml/issues/15847.~ YouTube is not going to be good to rely on, per [amphtml Slack conversation](https://amphtml.slack.com/archives/C9ARUE6Q6/p1558025296088500) with @newmuis:

> image

☝️ Edit: YouTube videos are _rarely_ portrait.

So perhaps we should add a warning to users when a video seems to be too large. I'm not sure what “too large” is, but perhaps we should target 1 megabyte per second? So we'd show no warning if there is a 7-second video that is 7MB.

The dog video in the [Hello World](https://amp.dev/documentation/examples/introduction/stories_in_amp/preview/) example is 2MB for 3 seconds:

image

kienstra:
* As Weston [suggested](https://github.com/ampproject/amp-wp/issues/2332#issue-445047113), this replaces the `'full'` image size with a custom size of `99999px` (unlimited) by `1440px`
replace-previous-size-here

* There's more work needed, including applying the rest of the points in #2332

Closes #2332


westonruter:
Make sure this new image size is used for background image.

kienstra:
Sure, I'll work on that now.

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -106,9 +109,11 @@ class PageEdit extends Component {
 
 			mediaType = media.type;
 		}
+		const fetchedMedia = select( 'core' ).getMedia( media.id );
+		const mediaUrl = has( fetchedMedia, [ 'media_details', 'sizes', MAX_IMAGE_SIZE_SLUG ] ) ? fetchedMedia.media_details.sizes[ MAX_IMAGE_SIZE_SLUG ].source_url : media.url;
```
kienstra: There might be a better way to get the image size than making a new request for images. 

The `media.sizes` array doesn't normally have the `'amp_story_image'` size. Otherwise, this would just get the size from that array.

assets/src/stories-editor/components/higher-order/with-replaced-full-size-image.js
```diff
@@ -0,0 +1,66 @@
...
+		 * This prevents the user from selecting an image in the Image block that's too big.
+		 * For example, 3200 x 4000.
+		 */
+		ampReplaceFullSizeImage() {
+			const blockEditorStore = 'core/block-editor';
+			const initialSizes = select( blockEditorStore ).getSettings( 'imageSizes' ).imageSizes;
+			const sizesWithoutFullSize = initialSizes.filter( ( size ) => 'full' !== size.slug );
+
+			// If the AMP Story slug isn't present, add it.
+			if ( ! sizesWithoutFullSize.filter( ( size ) => MAX_IMAGE_SIZE_SLUG === size.slug ).length ) {
```
kienstra: It'd be nice if this logic to add the image size didn't have to run so often. For example, if it ran only on `domready`.

But it looks like the settings update, so these settings will be overwritten if this logic only runs on `domready`.

assets/src/stories-editor/blocks/amp-story-page/edit.js
kienstra: I'll look for a way to have the `'amp_story_image'` size appear in that `sizes` object.

westonruter: It may be just that you have to re-generate image sizes? I found this: https://github.com/WP-API/WP-API/issues/779

kienstra: Nice, thanks for finding that! You'd think that would fix it, but it looks like it's still an issue in my local.

westonruter: Maybe this is not relevant, but at least in the classic Media Library's Ajax actions, there was a `image_size_names_choose` filter for ensuring the additional sizes are included in the response:

https://github.com/WordPress/wordpress-develop/blob/21b3d9031a24ce4821b5ffe613a1926fa08d7380/src/wp-includes/media.php#L3300-L3309

This I don't believe applies to the REST API, however.

kienstra: Yeah, that could have been a really helpful one. It looks like the filter doesn't run in this context, though.

kienstra: Hi @westonruter, it turns out that the `image_size_names_choose` filter that you suggested actually does work. I misspelled the image size slug, so it seemed like it didn't work.

It's used in https://github.com/ampproject/amp-wp/pull/2509/commits/78692f6764900f3a56e3e8d62b4eb97aaa5e7d75, thanks for suggesting it.

kienstra:
**Current Display For Large Video Sizes**

https://github.com/ampproject/amp-wp/pull/2509/commits/bd8d73b83b78f2ce26874c15851c67ec919336a7 adds a notice on selecting a video as the Background Media if it has more than 1 MB per second, like more than 5 MB for a 5-second video.

1. This might need better styling.
2. Maybe this could use more information, like the size of the video and its length.

large-video-file-size

To test this, you could download [this image](https://www.videezy.com/abstract/21968-rotating-shot-of-weathered-baseball-and-baseball-glove-002), which is 33 seconds and 252 MB. This should display the notice.

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -106,15 +110,56 @@ class PageEdit extends Component {
 
 			mediaType = media.type;
 		}
+		const mediaUrl = has( media, [ 'sizes', MAX_IMAGE_SIZE_SLUG, 'url' ] ) ? media.sizes[ MAX_IMAGE_SIZE_SLUG ].url : media.url;
+
+		let isExcessiveVideoSize = false;
+		if ( VIDEO_BACKGROUND_TYPE === mediaType ) {
+			isExcessiveVideoSize = this.isVideoSizeExcessive( media.filesizeInBytes, media.fileLength );
```
westonruter: Minor point, but this could be simplified to:

```js
const isExcessiveVideoSize = VIDEO_BACKGROUND_TYPE === mediaType && this.isVideoSizeExcessive( media.filesizeInBytes, media.fileLength );
```

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -275,6 +321,12 @@ class PageEdit extends Component {
 					</PanelColorSettings>
 					<PanelBody title={ __( 'Background Media', 'amp' ) }>
 						<>
+							{
+								isExcessiveVideoSize &&
+								<Notice status="error" isDismissible={ false } >
```
westonruter: Let's use `warning` instead of `error`, since it will still work.

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -275,6 +321,12 @@ class PageEdit extends Component {
...
+									{ __( 'The video size is more than 1 MB per second.', 'amp' ) }
```
westonruter: Perhaps the `1` here should be pulled from the same constant as `megabyteInBytes` above?

assets/src/stories-editor/components/higher-order/with-replaced-full-size-image.js
```diff
@@ -0,0 +1,66 @@
...
+		constructor() {
+			super( ...arguments );
+			this.componentDidUpdate = this.componentDidUpdate.bind( this );
+			this.ampReplaceFullSizeImage = this.ampReplaceFullSizeImage.bind( this );
+		}
+
+		/**
+		 * When the component updates, if it's for the Image block, replace the full size image.
+		 */
+		componentDidUpdate() {
```
westonruter: Is it not possible to intercept the media upon selection to replace the `full` with the desired image size? Then there wouldn't be a need to turnaround and update it again. Or maybe I am not understanding enough.

includes/class-amp-story-post-type.php
```diff
@@ -1500,4 +1512,26 @@ public static function change_embed_iframe_attributes( $output, $post ) {
...
+	 *     An associative array of image sizes.
+	 *
+	 *     @type string $slug Image size slug, like 'medium'.
+	 *     @type string $name Image size name, like 'Medium'.
+	 * }
+	 * @return array $image_sizes The filtered image sizes.
+	 */
+	public static function add_new_max_image_size( $image_sizes ) {
+		if ( self::POST_TYPE_SLUG === get_post_type() ) {
+			$image_sizes[ self::STORY_PAGE_IMAGE_SIZE ] = __( 'AMP Story Maximum Image', 'amp' );
```
westonruter: How about just “Story Background” or “Story Page”. Should minimize showing “AMP” to the user.

westonruter:
@kienstra I'm getting an error:

image

assets/src/stories-editor/blocks/amp-story-page/edit.js
kienstra: Good idea, https://github.com/ampproject/amp-wp/pull/2509/commits/d2f9ea1c040f24585fe582858cc8f4c4068341bf does that.

assets/src/stories-editor/blocks/amp-story-page/edit.js
kienstra: Good idea, applied with https://github.com/ampproject/amp-wp/pull/2509/commits/4f262f8a3743c53cbbf64ecd70d03dd25c11fdff.

assets/src/stories-editor/blocks/amp-story-page/edit.js
kienstra: Ah, good idea. This is done in https://github.com/ampproject/amp-wp/pull/2509/commits/972e6afba90fd324a908fc02b19d65a261e264bd.

Is this what you had in mind? It's the display with that commit:

size-is-more-than

includes/class-amp-story-post-type.php
kienstra: Good idea, https://github.com/ampproject/amp-wp/pull/2509/commits/25d105733f92d946561c8e82621f6e37f0d69d52 changes the name to `'Story Background'` if that's alright.

assets/src/stories-editor/blocks/amp-story-page/edit.js
westonruter: No, not that it shoes the number of bytes, but just that the underlying megabyte count is stored in a shared constant. In other words, the constant `1` should be defined in one place. In the translation string the constant would be used as `1 MB` but in the function it would be multiplied by 1000000.

assets/src/stories-editor/components/higher-order/with-replaced-full-size-image.js
kienstra: Good question. That entire extending JS class isn't a great approach.

https://github.com/ampproject/amp-wp/pull/2509/commits/236e2265c2cbfb442ae5f6670245df9f64ddda17 removes that entire JS class above, in favor of simply [removing the full size](https://github.com/ampproject/amp-wp/pull/2509/commits/236e2265c2cbfb442ae5f6670245df9f64ddda17#diff-37ed1ba86eaee9e72c2ced620309452bR1532) in the filter that you recommended.

kienstra:
>@kienstra I'm getting an error:

Thanks for letting me know. With https://github.com/ampproject/amp-wp/pull/2509/commits/236e2265c2cbfb442ae5f6670245df9f64ddda17 removing the class in `with-replaced-full-size-image.js`, that shouldn't be an issue anymore.

kienstra:
**Image Block**

For the Image block, this doesn't change the default size that's used. That's still the `'large'` size, which is [usually 1024 x 1024px max](https://codex.wordpress.org/Post_Thumbnails#Thumbnail_Sizes)). 

On selecting an image for the Image block, it will usually use the 'large' image (if it's available).

larg-here

But this PR removes the `'full'` size, and replaces it with this custom `99999 x 1440px` size.

kienstra:
But this does use the new custom size (99999 x 1440px) [for the Background Media](https://github.com/ampproject/amp-wp/pull/2509/files#diff-3ebaa9cdbb7e5502ab37969c8f31350eR122) and the [video's poster image](https://github.com/ampproject/amp-wp/pull/2509/files#diff-3ebaa9cdbb7e5502ab37969c8f31350eR383), if the size is available.

swissspidy:
For images that are too small we have a warning in the media modal in addition to one in the sidebar. Should we have the same for too large videos?

assets/src/stories-editor/blocks/amp-story-page/index.js
```diff
@@ -51,6 +51,10 @@ const schema = {
 	overlayOpacity: {
 		default: 50,
 	},
+	isExcessiveVideoSize: {
```
swissspidy: I don't think this should be an attribute that's saved with the block.

Whether a video is excessive or not can be determined at runtime by checking the media object.

kienstra:
>For images that are too small we have a warning in the media modal in addition to one in the sidebar. Should we have the same for too large videos?

Good question, maybe this should have a warning for videos also, for consistency.

assets/src/stories-editor/blocks/amp-story-page/index.js
kienstra: Ah, OK. So this notice doesn't need to display on reloading the page, only right after selecting the video?

kienstra: Oh, you're mentioning that whether the video size is excessive could be calculated at runtime, without saving this attribute.

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -106,15 +119,51 @@ class PageEdit extends Component {
...
+		return fileSize > fileLengthInSeconds * MEGABYTE_IN_BYTES;
+	}
+
+	/**
+	 * Gets the number of seconds in a colon-separated time string, like '01:10'.
+	 *
+	 * @param {string} time A colon-separated time, like '0:12'.
+	 * @return {number} seconds The number of seconds in the time, like 12.
+	 */
+	getSecondsFromTime( time ) {
```
swissspidy: Minor point, but since `getSecondsFromTime` and `isVideoSizeExcessive` are not really tied to this class, they could be moved to the helpers file where we could also more easily write unit tests for them. Alternatively, the methods could be declared static and also have tests more easily.

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -275,6 +325,18 @@ class PageEdit extends Component {
 					</PanelColorSettings>
 					<PanelBody title={ __( 'Background Media', 'amp' ) }>
 						<>
+							{
+								isExcessiveVideoSize &&
+								<Notice status="warning" isDismissible={ false } >
+									{
+										sprintf(
+											/* translators: %d: the number of bytes in a megabyte */
+											__( 'The video size is more than %d bytes per second.', 'amp' ),
```
swissspidy: As Weston already mentioned, this should still use MB in the string, so with something like `MEGABYTE_IN_BYTES / 1024`.

For featured image we say `Selecting a featured image is recommended for an optimal user experience.` (see `validateFeaturedImage()`). As a user I'd expect a similar user-friendly message for large videos.

assets/src/stories-editor/blocks/amp-story-page/index.js
swissspidy: Yes exactly.

assets/src/stories-editor/blocks/amp-story-page/edit.js
kienstra: Good point, https://github.com/ampproject/amp-wp/pull/2509/commits/82b827436ee0bbe723ebf9b5d8f8b9804f462cfb moves these functions to `helpers/index.js`.

assets/src/stories-editor/blocks/amp-story-page/edit.js
kienstra: Ah, thanks. How does https://github.com/ampproject/amp-wp/pull/2509/commits/c966459de356d8d4280eedb550ce65bfae71fcbe look?

new-notice-text

It doesn't use `1024` in the calculation, though. Is that alright?

assets/src/stories-editor/blocks/amp-story-page/edit.js
kienstra: Oh, thanks. How does https://github.com/ampproject/amp-wp/pull/2509/commits/c966459de356d8d4280eedb550ce65bfae71fcbe look?

new-notice-text

assets/src/stories-editor/blocks/amp-story-page/index.js
kienstra: Hm, I worked on doing this calculation inside the [render() method](https://github.com/ampproject/amp-wp/blob/b342ff33aec3da9a62ce06d33535867461a238d5/assets/src/stories-editor/blocks/amp-story-page/edit.js#L200) of `amp-story-page/edit.js`, so that `isExcessiveVideoSize` doesn't have to be stored in an attribute. 

But I can't get the `media` object of the video in that `render()` method on page load.

This:
```js
select( 'core' ).getMedia( mediaId )
```
...returns `undefined`

I'll look at this a little more.

swissspidy: That select needs to be run in a withSelect wrapper. It will return undefined at first, but as soon as the data arrives it will be there and the component will be re-rendered.

kienstra: Ah, good idea.

assets/src/stories-editor/blocks/amp-story-page/edit.js
swissspidy: Much better this way 👍

assets/src/stories-editor/blocks/amp-story-page/index.js
kienstra: Thanks, https://github.com/ampproject/amp-wp/pull/2509/commits/537fb08d7d5eb8d4465c8bbc6416acd840849cf1 uses your idea. It works locally 😄

assets/src/stories-editor/blocks/amp-story-page/edit.js
kienstra: Thanks!

includes/class-amp-story-post-type.php
```diff
@@ -1500,4 +1512,29 @@ public static function change_embed_iframe_attributes( $output, $post ) {
...
+	 * @param array $image_sizes {
+	 *     An associative array of image sizes.
+	 *
+	 *     @type string $slug Image size slug, like 'medium'.
+	 *     @type string $name Image size name, like 'Medium'.
+	 * }
+	 * @return array $image_sizes The filtered image sizes.
+	 */
+	public static function add_new_max_image_size( $image_sizes ) {
+		if ( self::POST_TYPE_SLUG === get_post_type() ) {
```
westonruter: This is not working for me. The Ajax request that obtains the media library results does not include a post as context, as far as I can see. So when I insert the background image into the story page, I do not have the expected image size. I do, however, if I make this change:

```diff
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -1530,11 +1530,7 @@ class AMP_Story_Post_Type {
 	 * @return array $image_sizes The filtered image sizes.
 	 */
 	public static function add_new_max_image_size( $image_sizes ) {
-		if ( self::POST_TYPE_SLUG === get_post_type() ) {
-			unset( $image_sizes['full'] );
-			$image_sizes[ self::MAX_IMAGE_SIZE_SLUG ] = __( 'Story Background', 'amp' );
-		}
-
+		$image_sizes[ self::MAX_IMAGE_SIZE_SLUG ] = __( 'Story Background', 'amp' );
 		return $image_sizes;
 	}
 }
```

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -275,6 +292,18 @@ class PageEdit extends Component {
 					</PanelColorSettings>
 					<PanelBody title={ __( 'Background Media', 'amp' ) }>
 						<>
+							{
+								isExcessiveVideoSize &&
+								<Notice status="warning" isDismissible={ false } >
+									{
+										sprintf(
+											/* translators: %d: the number of bytes in a megabyte */
+											__( 'A video size of less than %d MB per second is recommended.', 'amp' ),
```
westonruter: This seems to be working:

image

includes/class-amp-story-post-type.php
```diff
@@ -1500,4 +1512,29 @@ public static function change_embed_iframe_attributes( $output, $post ) {
...
+			unset( $image_sizes['full'] );
```
westonruter: Why remove the `full` size since the story background image size is being explicitly looked for?

includes/class-amp-story-post-type.php
westonruter: It's not working in that the full size image is being used, as opposed to an image filename ending in `-1080x1440.jpg`.

assets/src/stories-editor/blocks/amp-story-page/edit.js
kienstra: Nice!

includes/class-amp-story-post-type.php
kienstra: Yeah, the `unset()` statement has no effect on the background image, as it's not using the `full` size.

This filter applies to the story background image *and* the Image block.

That `unset()` statement prevents the user from selecting the `'full'` size image in the Image block:

larg-here

For the Image block, this PR doesn't change the default size that's used. That's still the `'large'` size, which is [usually 1024 x 1024px max](https://codex.wordpress.org/Post_Thumbnails#Thumbnail_Sizes)). Though I could look at changing that.

If this PR didn't remove the `full` size, it wouldn't help much for the Image block. 

It would add an intermediate size that users could choose, which is slightly bigger than the `large` size. But they could still choose the full image.

includes/class-amp-story-post-type.php
kienstra: Interesting, thanks for testing this. 

What was the original size of the image that you selected as the 'Background Media'?

These steps seem to work well locally:
1. In 'Background Media,' click 'Select Media' or 'Edit Media' 
2. Upload a [1900 x 1900 image](http://lorempixel.com/1900/1900/technics/)
3. Click 'Select'
4. Observe that the URL of the background image ends in 1440-1440.jpeg, meaning it used the `amp_story_page` size:

```
https://loc.test/wp-content/uploads/2019/06/story-1900-1900-3-1440x1440.jpeg
```
bck-image-here

These 4 steps also worked with these image sizes:
[1000 x 1500](http://lorempixel.com/1000/1500/technics/)
[1200 x 1500](http://lorempixel.com/1200/1500/technics/)
[1500 x 1600](http://lorempixel.com/1500/1600/technics/)
[1800 x 1800](http://lorempixel.com/1800/1800/technics/)

Of course, this didn't work with images less than 1440 in height, as they're not big enough to have an `amp_story_page` sized image (99999 x 1440).

assets/src/stories-editor/blocks/amp-story-page/edit.js
swissspidy: Just the translator comment is incorrect (bytes vs megabytes)

swissspidy: Fixed in bb9e2d0de4e90707fde00621ed4f4338e8e6ddf0

swissspidy:
My understanding of #2332 was that these warnings for excessive videos should be shown anywhere a video can be inserted: either as a background video or in a video block. Right now, this warning is only shown for background media.

A suggestion of mine was to also show the warnings when selecting the video in the media modal, just like we do for the featured image.

So I would say this only partially solves #2332 for now. We should definitely address these other points too, but IMO after release.

kienstra:
Hi @westonruter,
Thanks for your patience here.

https://github.com/ampproject/amp-wp/pull/2509/commits/0bd76bc652288cb82448d6134448323d5d853778#diff-37ed1ba86eaee9e72c2ced620309452bR1545 adds a conditional in the PHP filter to check that `'query-attachments' === $_POST['action']`. 

This allows it to apply to the background image, but not to the Image block's 'Image Size' `<select>` element (or an Image block in a non-story editor).

I also brought back [the JS filter](https://github.com/ampproject/amp-wp/pull/2509/files#diff-31d524f8dd926f661a7ba1db7ca21718R20), like earlier in this PR, which [replaces](https://github.com/ampproject/amp-wp/pull/2509#issuecomment-499790081) the Image block's `Full` size with the `amp_story_page` size.

This time, it didn't see the `Uncaught TypeError: Cannot call class as a function` error that [you found earlier](https://github.com/ampproject/amp-wp/pull/2509#issuecomment-499611747). But it'd be interesting to know if you see that error.

kienstra:
Hi @swissspidy,
Thanks a lot for your help on #2389.

Would you be available to review this again? Thanks, Pascal!

westonruter:
@kienstra I'm getting a JS error upon loading the editor:

image

kienstra:
>@kienstra I'm getting a JS error upon loading the editor:

OK, thanks. The [JS filter](https://github.com/ampproject/amp-wp/pull/2509/files#diff-31d524f8dd926f661a7ba1db7ca21718R20) is probably responsible for that.

kienstra:
Nice, great idea: https://github.com/ampproject/amp-wp/pull/2509/commits/fb29ca8a227f5825ab4519c13e98aa71e4f5ef81

assets/src/stories-editor/components/higher-order/with-replaced-full-size-image.js
```diff
@@ -0,0 +1,73 @@
...
+			 * When the component updates, if it's for the Image block, replace the full size image.
+			 */
+			componentDidUpdate() {
+				const { name } = this.props;
+				if ( 'core/image' === name ) {
+					this.ampReplaceFullSizeImage();
+				}
+
+				if ( super.componentDidUpdate ) {
+					super.componentDidUpdate();
```
westonruter: I'm no React expert, but shouldn't this be passing the args passed into `componentDidUpdate`? In other words:

```js
super.componentDidUpdate( ...arguments );
```

westonruter:
I don't see any other  example in the codebase of `BlockEdit` extended, so perhaps this is not the right approach. Composition should be used instead?

assets/src/stories-editor/components/higher-order/with-replaced-full-size-image.js
kienstra: Ah, good point.

kienstra:
>I don't see any other example in the codebase of BlockEdit extended, so perhaps this is not the right approach. Composition should be used instead?

Good point, inheritance isn't a best practice in React, it seems. The only issue is that this needs to 'hook' into the Image edit component's `componentDidUpdate()`.

But if this extended it with a higher-order-component, the HOC probably wouldn't call its `componentDidUpdate` method when the Image component's `componentDidUpdate` method runs.

Maybe there's another approach, though.

kienstra:
I'll keep looking at this.

kienstra:
Hi @westonruter,
Thanks for your help here.

Does https://github.com/ampproject/amp-wp/pull/2509/commits/c330ce77a06d1213afca08b33b01e84c306db47c work for you? It uses the PHP filter that you had suggested originally. It works for me locally.

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -275,6 +292,18 @@ class PageEdit extends Component {
 					</PanelColorSettings>
 					<PanelBody title={ __( 'Background Media', 'amp' ) }>
 						<>
+							{
+								isExcessiveVideoSize &&
+								<Notice status="warning" isDismissible={ false } >
+									{
+										sprintf(
+											/* translators: %d: the number of recommended megabytes per second */
+											__( 'A video size of less than %d MB per second is recommended.', 'amp' ),
```
westonruter: Minor thing but it would be useful to know how many MB/second the selected video is, to know how far from the target the video is.

kienstra:
@westonruter, thanks a lot for your help and patience!

PR #2535, opened 2019-06-06
 Use same font-weight for text block regardless of type

### LINKED ISSUE #2519: Automatic text type should not change formatting
Text blocks have the text type (i.e. whether it's a heading or a paragraph) set to _automatic_ by default. Among other things, the text type changes depending on text length and position of the block. When the text type changes, so does the formatting, e.g. a heading is more bold than a paragraph.

For users these changes can be very unclear. Ideally the text type change does not affect formatting.

swissspidy:
Fixes #2519.


assets/css/amp-stories.css
```diff
@@ -42,6 +42,10 @@ amp-story-grid-layer[template="fill"] .wp-block-image {
 	overflow-wrap: break-word;
 }
 
+.wp-block-amp-amp-story-text {
```
westonruter: If one AMP is good, two is better! AMP^2

PR #2540, opened 2019-06-06
Improve display of duplicated pages

### LINKED ISSUE #2510: Story not working on duplicate action
It has been observed that whenever we use block duplicate action and then publish the story, the story doesn't work and shows something like this
image

This problem is still persisting with beta 2. Turns out when pages are duplicated they also **duplicate the story page ID** which turns amp story engine to play weirdly and then we can not move around at all. I don't if the pageID is the actual reason to this but this really doesn't let us use duplicate action at all

kienstra:
# Before
The `amp-story-page[id]` was the same in both the duplicated page and the original page (thanks to @swissspidy for pointing to the `anchor` attribute in #2510).

before-here

This [is set from the anchor attribute](https://github.com/ampproject/amp-wp/blob/bf82129e1636ecb945f0675f951112b1d74e5cdd/assets/src/stories-editor/blocks/amp-story-page/save.js#L51) in the block.

# After
So if the `anchor` attribute is the same for two pages, this resets it on the duplicated one to the same as its `clientId`:

duplicating-blocks

Fixes #2510

### DIFF
--- a/assets/src/stories-editor/index.js
+++ b/assets/src/stories-editor/index.js
@@ -280,16 +280,13 @@ store.subscribe( () => {
 	 * Prevent an issue where duplicated page blocks had the same anchor attribute, and didn't look right on the front-end.
 	 * @see https://github.com/ampproject/amp-wp/issues/2510
 	 */
-	const usedAnchors = [];
 	const conditionallyUpdateAnchor = ( block ) => {
 		if ( ! has( block, [ 'attributes', 'anchor' ] ) ) {
 			return;
 		}
 
-		if ( usedAnchors.includes( block.attributes.anchor ) ) {
+		if ( block.attributes.anchor !== block.clientId ) {
 			updateBlockAttributes( block.clientId, { anchor: block.clientId } );
-		} else {
-			usedAnchors.push( block.attributes.anchor );
 		}
 	};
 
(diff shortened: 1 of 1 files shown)


kienstra:
**Request For Review**

Hi @westonruter and @swissspidy,
Could you please review this PR? Thanks, and hope you're doing great.

assets/src/stories-editor/index.js
```diff
@@ -269,6 +275,28 @@ store.subscribe( () => {
...
+	 * @see https://github.com/ampproject/amp-wp/issues/2510
+	 */
+	const usedAnchors = [];
+	const conditionallyUpdateAnchor = ( block ) => {
+		if ( ! has( block, [ 'attributes', 'anchor' ] ) ) {
+			return;
+		}
+
+		if ( usedAnchors.includes( block.attributes.anchor ) ) {
+			updateBlockAttributes( block.clientId, { anchor: block.clientId } );
```
swissspidy: This doesn't work when `block.attributes.anchor === block.clientId`. Probably easier to just do something like this:

```js
if ( block.attributes.anchor !=== block.clientId ) {
  updateBlockAttributes( block.clientId, { anchor: block.clientId } );
}
```

kienstra: Sure, good idea. https://github.com/ampproject/amp-wp/pull/2540/commits/ba4f1558a650eb29a1046645074d820e0eb3a1e8 uses your snippet above.

PR #2550, opened 2019-06-08
Refactor AMP modes

### LINKED ISSUE #2311: Restore suppression of validation error warnings appearing in Reader mode
Introduced in #2143 and pointed out in https://github.com/ampproject/amp-wp/pull/2143#issuecomment-483997810:

> Note: One side effect of the changes here (specifically 4645a4a) is that making changes to non-story posts when in Reader mode will start to show the validation error messages. Previously the validation warnings would only be shown when in Native/Transitional mode. The validation notices will likely be a source of complaints for users when in Reader mode, since they would be reported as if the Transitional mode is active.

### LINKED ISSUE #2312: Always allow switching from Transitional to Native
If `add_theme_support('amp', [ 'paired' => true ])` is present, then a user should still be able to still switch from Transitional to Native via the AMP admin screen, and back from Native to Transitional. If a site works in Transitional mode then by definition it should work in Native mode. But if `paired => false` (or `paired` is absent), then Native would be the only option, as the reverse is not true: if a theme only works in Native mode then it may be using AMP components in its templates (e.g. `amp-sidebar`) without any non-AMP version also implemented.

This will allow all Paired themes (e.g. Neve) to be also used in Native mode.

### LINKED ISSUE #2470: Add functionality for users to chose AMP + Stories, AMP-only, Stories-only
As long as the Stories editor is part of the AMP Plugin, it needs to provide the controls for users to chose how the plugin would be used. The options could include these (or a variation):

- <code>AMP + Stories</code>: Serve AMP content, and enable Stories
- <code>AMP Only</code>: Use AMP only, hide Stories editor
- <code>Stories Only</code>: Only use the Stories feature, do not serve AMP pages

One possibility (discussed with @westonruter) could be to add a new "Operational Mode" complementing the Native, Transitional, Reader trifecta, with a "Stories" mode. The user could check only the Stories checkbox (Stories only), or check the Stories checkbox and one of the AMP template modes (Stories + AMP).

### LINKED ISSUE #2507: Phase out “Native” terminology in favor of “AMP-first”
The AMP settings screen describes the three modes: Reader, Transitional, and Native.

image

Originally Reader was called “Classic” and Transitional was called “Paired”, but these were renamed in #2034. The “Native” name was not changed, however. While it used to be called “canonical AMP” it is now termed “AMP first” as referenced in [AMP as your web framework](https://blog.amp.dev/2019/05/01/amp-as-your-web-framework/). While the Native mode description does reference AMP-first, the mode itself is not termed that yet.  This should change.

Changing the mode name may need to be done as part of a redesign of the whole screen, as with #1288. This especially true as we hope to be able to default sites to AMP-first in the future (as opposed to Reader mode), per #1384.

westonruter:
This PR intends to refactor the AMP modes in the plugin in order to resolve three issues:

- [x] Fixes #2312: Always allow switching from Transitional to Native on sites that are defined with `add_theme_support('amp', ['paired' => true])`.
- [x] Fixes #2311: Restore suppression of validation error warnings appearing in Reader mode
- [x] Fixes #2470: Add functionality for users to chose AMP + Stories, AMP-only, Stories-only 
  - [x] Ensure availability of Latest Stories block and Story post embed.
  - [x] Include enabled experiences in generator tag, as well as website mode.
- [x] Fixes #2507: Phase out “Native” terminology in favor of “AMP-first”
  - [x] Rename historical references of `native` to `standard`, and `paired` to `transitional`.

# Always allow switching themes with built-in Transitional support to Native

When a theme had `add_theme_support( 'amp', array( 'paired' => true ) )` they would see:

> image

Now, they see this instead:

> image

This allows a theme which declares support for both AMP and non-AMP, to just opt to only serve AMP pages instead of being paired mode. This allows a theme like [Neve](https://wordpress.org/themes/neve/) to be used in either mode, even though it is defined as being `paired => true`.

Additionally, if they had `add_theme_support( 'amp' )` then before they would see:

> image

Now they see a slightly tweaked message:

> image

Notice the link to the ecosystem page is moved to the bottom.

# Generator Tag

Given a site in transitional mode with both website and stories experiences enabled, the generator tag is now:

```html
<meta name="generator" content="AMP Plugin v1.2-beta2; mode=transitional; experiences=website,stories">
```

Note the new `experiences` param. If switching to standard (formerly native) mode, this is also reflected:

```html
<meta name="generator" content="AMP Plugin v1.2-beta2; mode=standard; experiences=website,stories">
```

If stories is disabled, then this is shown as:

```html
<meta name="generator" content="AMP Plugin v1.2-beta2; mode=standard; experiences=website">
html

If website is disabled but stories is enabled, then:

```html
<meta name="generator" content="AMP Plugin v1.2-beta2; mode=none; experiences=stories">
```

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -517,7 +517,7 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  *
  *      add_theme_support( AMP_Theme_Support::SLUG );
  *
- * This will serve templates in native AMP, allowing you to use AMP components in your theme templates.
+ * This will serve templates in AMP-first, allowing you to use AMP components in your theme templates.
  * If you want to make available in transitional mode, where templates are served in AMP or non-AMP documents, do:
  *
  *      add_theme_support( AMP_Theme_Support::SLUG, array(
@@ -530,7 +530,7 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  *          'template_dir' => 'amp',
  *      ) );
  *
- * If you want to have AMP-specific templates in addition to serving native AMP, do:
+ * If you want to have AMP-specific templates in addition to serving AMP-first, do:
  *
  *      add_theme_support( AMP_Theme_Support::SLUG, array(
  *          'paired'       => false,
@@ -553,7 +553,7 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  *      ) );
  *
  * @see AMP_Theme_Support::read_theme_support()
- * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is native and there is not separate AMP URL current URL.
+ * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is AMP-first and there is not separate AMP URL current URL.
  */
 function amp_is_canonical() {
 	if ( ! current_theme_supports( AMP_Theme_Support::SLUG ) ) {
--- a/amp.php
+++ b/amp.php
@@ -278,7 +278,7 @@ function amp_deactivate() {
 		}
 	}
 
-	flush_rewrite_rules();
+	flush_rewrite_rules( false );
 }
 
 /*
--- a/amp.php
+++ b/amp.php
@@ -343,6 +343,7 @@ function amp_init() {
 	add_action( 'admin_init', 'AMP_Options_Manager::register_settings' );
 	add_action( 'wp_loaded', 'amp_add_options_menu' );
 	add_action( 'wp_loaded', 'amp_admin_pointer' );
+	add_action( 'wp_loaded', 'amp_post_meta_box' ); // Used in both Website and Stories experiences.
 
 	if ( AMP_Options_Manager::is_website_experience_enabled() ) {
 		add_rewrite_endpoint( amp_get_slug(), EP_PERMALINK );
@@ -351,7 +352,6 @@ function amp_init() {
 		add_action( 'parse_query', 'amp_correct_query_when_is_front_page' );
 		add_action( 'admin_bar_menu', 'amp_add_admin_bar_view_link', 100 );
 		add_action( 'wp_loaded', 'amp_editor_core_blocks' );
-		add_action( 'wp_loaded', 'amp_post_meta_box' );
 		add_filter( 'request', 'amp_force_query_var_value' );
 
 		// Add actions for reader mode templates.
--- a/amp.php
+++ b/amp.php
@@ -333,38 +333,42 @@ function amp_init() {
 	 */
 	do_action( 'amp_init' );
 
-	add_rewrite_endpoint( amp_get_slug(), EP_PERMALINK );
-
 	add_filter( 'allowed_redirect_hosts', array( 'AMP_HTTP', 'filter_allowed_redirect_hosts' ) );
 	AMP_HTTP::purge_amp_query_vars();
 	AMP_HTTP::send_cors_headers();
 	AMP_HTTP::handle_xhr_request();
 	AMP_Theme_Support::init();
 	AMP_Validation_Manager::init();
-	AMP_Post_Type_Support::add_post_type_support();
-	AMP_Story_Post_Type::register();
 	AMP_Service_Worker::init();
-	add_action( 'init', array( 'AMP_Post_Type_Support', 'add_post_type_support' ), 1000 ); // After post types have been defined.
+	add_action( 'admin_init', 'AMP_Options_Manager::register_settings' );
+	add_action( 'wp_loaded', 'amp_add_options_menu' );
+	add_action( 'wp_loaded', 'amp_admin_pointer' );
+
+	if ( AMP_Options_Manager::is_website_experience_enabled() ) {
+		add_rewrite_endpoint( amp_get_slug(), EP_PERMALINK );
+		AMP_Post_Type_Support::add_post_type_support();
+		add_action( 'init', array( 'AMP_Post_Type_Support', 'add_post_type_support' ), 1000 ); // After post types have been defined.
+		add_action( 'parse_query', 'amp_correct_query_when_is_front_page' );
+		add_action( 'admin_bar_menu', 'amp_add_admin_bar_view_link', 100 );
+		add_action( 'wp_loaded', 'amp_editor_core_blocks' );
+		add_action( 'wp_loaded', 'amp_post_meta_box' );
+		add_filter( 'request', 'amp_force_query_var_value' );
+
+		// Add actions for reader mode templates.
+		add_action( 'wp', 'amp_maybe_add_actions' );
+
+		// Redirect the old url of amp page to the updated url.
+		add_filter( 'old_slug_redirect_url', 'amp_redirect_old_slug_to_new_url' );
+	}
+
+	if ( AMP_Options_Manager::is_stories_experience_enabled() ) {
+		AMP_Story_Post_Type::register();
+		add_action( 'wp_loaded', 'amp_story_templates' );
+	}
 
 	if ( defined( 'WP_CLI' ) && WP_CLI ) {
 		WP_CLI::add_command( 'amp', new AMP_CLI() );
 	}
-
-	add_filter( 'request', 'amp_force_query_var_value' );
-	add_action( 'admin_init', 'AMP_Options_Manager::register_settings' );
-	add_action( 'wp_loaded', 'amp_editor_core_blocks' );
-	add_action( 'wp_loaded', 'amp_post_meta_box' );
-	add_action( 'wp_loaded', 'amp_story_templates' );
-	add_action( 'wp_loaded', 'amp_add_options_menu' );
-	add_action( 'wp_loaded', 'amp_admin_pointer' );
-	add_action( 'parse_query', 'amp_correct_query_when_is_front_page' );
-	add_action( 'admin_bar_menu', 'amp_add_admin_bar_view_link', 100 );
-
-	// Redirect the old url of amp page to the updated url.
-	add_filter( 'old_slug_redirect_url', 'amp_redirect_old_slug_to_new_url' );
-
-	// Add actions for legacy post templates.
-	add_action( 'wp', 'amp_maybe_add_actions' );
 
 	/*
 	 * Broadcast plugin updates.
--- a/amp.php
+++ b/amp.php
@@ -561,8 +561,8 @@ function amp_is_canonical() {
 	}
 
 	$args = AMP_Theme_Support::get_theme_support_args();
-	if ( isset( $args['paired'] ) ) {
-		return empty( $args['paired'] );
+	if ( isset( $args[ AMP_Theme_Support::PAIRED_FLAG ] ) ) {
+		return empty( $args[ AMP_Theme_Support::PAIRED_FLAG ] );
 	}
 
 	// If there is a template_dir, then transitional mode is implied.
--- a/amp.php
+++ b/amp.php
@@ -553,7 +553,7 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  *      ) );
  *
  * @see AMP_Theme_Support::read_theme_support()
- * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is AMP-first and there is not separate AMP URL current URL.
+ * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is AMP-first and there is not a separate (paired) AMP URL.
  */
 function amp_is_canonical() {
 	if ( ! current_theme_supports( AMP_Theme_Support::SLUG ) ) {
--- a/assets/src/block-editor/index.js
+++ b/assets/src/block-editor/index.js
@@ -16,20 +16,23 @@ import './store';
 
 const { ampLatestStoriesBlockData } = window;
 
-addFilter( 'blocks.registerBlockType', 'ampEditorBlocks/addAttributes', addAMPAttributes );
-addFilter( 'blocks.getSaveElement', 'ampEditorBlocks/filterSave', filterBlocksSave );
-addFilter( 'editor.BlockEdit', 'ampEditorBlocks/filterEdit', filterBlocksEdit, 20 );
-addFilter( 'blocks.getSaveContent.extraProps', 'ampEditorBlocks/addExtraAttributes', addAMPExtraProps );
-addFilter( 'editor.PostFeaturedImage', 'ampEditorBlocks/withFeaturedImageNotice', withFeaturedImageNotice );
-addFilter( 'editor.MediaUpload', 'ampEditorBlocks/addCroppedFeaturedImage', ( InitialMediaUpload ) => withCroppedFeaturedImage( InitialMediaUpload, getMinimumFeaturedImageDimensions() ) );
+// Add filters if AMP for Website experience is enabled.
+if ( select( 'amp/block-editor' ).isWebsiteEnabled() ) {
+	addFilter( 'blocks.registerBlockType', 'ampEditorBlocks/addAttributes', addAMPAttributes );
+	addFilter( 'blocks.getSaveElement', 'ampEditorBlocks/filterSave', filterBlocksSave );
+	addFilter( 'editor.BlockEdit', 'ampEditorBlocks/filterEdit', filterBlocksEdit, 20 );
+	addFilter( 'blocks.getSaveContent.extraProps', 'ampEditorBlocks/addExtraAttributes', addAMPExtraProps );
+	addFilter( 'editor.PostFeaturedImage', 'ampEditorBlocks/withFeaturedImageNotice', withFeaturedImageNotice );
+	addFilter( 'editor.MediaUpload', 'ampEditorBlocks/addCroppedFeaturedImage', ( InitialMediaUpload ) => withCroppedFeaturedImage( InitialMediaUpload, getMinimumFeaturedImageDimensions() ) );
 
-const plugins = require.context( './plugins', true, /.*\.js$/ );
+	const plugins = require.context( './plugins', true, /.*\.js$/ );
 
-plugins.keys().forEach( ( modulePath ) => {
-	const { name, render, icon } = plugins( modulePath );
+	plugins.keys().forEach( ( modulePath ) => {
+		const { name, render, icon } = plugins( modulePath );
 
-	registerPlugin( name, { icon, render } );
-} );
+		registerPlugin( name, { icon, render } );
+	} );
+}
 
 /*
  * If there's no theme support, unregister blocks that are only meant for AMP.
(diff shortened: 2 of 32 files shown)


amedina:
Excellent. These changes increase both user friendliness and flexibility. Kudos on removing the option for Reader mode when Transitional mode is added via theme support.

amedina:
> This allows a theme which declares support for both AMP and non-AMP, to just opt to only serve AMP pages instead of being paired mode.

This means users can go from Transitional to Native also in all cases. It would be good to allow users in Transitional mode to chose Native AMP for some pages. This would enable sites to track progress of the "AMPlification" of their content ~ `Your site is 86% Native AMP, 14% Transitional` or `Your site is 86% Native AMP, 10% Transitional`, `4% no-AMP`.

amedina:
And when `Bento AMP` is a thing, the scores can be detailed at an even higher granular level.

westonruter:
Chosing AMP-first for some URLs but paired for othets is somewhat described under the “Mixed-Some-Only-AMP” mode in https://github.com/ampproject/amp-wp/issues/934

IMO that adds a lot of complexity that I highly doubt users would be able to figure out or use. Possibly some complicated changes to validation logic as well. Unsure how scores would be calculated.

Decisions not options.

I think we should help users get to AMP-first on URLs that support it. For other templates/URLs, non-AMP would be served unless the theme is using Bento AMP components.

amedina:
I don't think the complexity will be much higher; and the benefit will be significant especially if weighed by the benefit of being able to measure the progression of a site towards AMP first.

amedina:
> This means users can go from Transitional to Native also in all cases. It would be good to allow users in Transitional mode to chose Native AMP for some pages. This would enable sites to track progress of the "AMPlification" of their content ~ Your site is 86% Native AMP, 14% Transitional or Your site is 86% Native AMP, 10% Transitional, 4% no-AMP.

Closing the loop on this. We agree that a more fluid model for the Transitional mode would facilitate the transition of sites towards AMP-first experiences by avoiding the current all-or-nothing approach. However, we will be able to achieve this in a much easier way once [Bento AMP](https://blog.amp.dev/2019/05/01/amp-as-your-web-framework/) is available. With Bento AMP a fully fluid transitional mode can be achieved without a lot of the complexity.

amp.php
```diff
@@ -278,7 +278,7 @@ function amp_deactivate() {
 		}
 	}
 
-	flush_rewrite_rules();
+	flush_rewrite_rules( false );
```
westonruter: The addition of `false` here is something which should have been here from the start, as there is no need to update `.htaccess`.

amp.php
```diff
@@ -549,16 +553,16 @@ function amp_correct_query_when_is_front_page( WP_Query $query ) {
  *      ) );
  *
  * @see AMP_Theme_Support::read_theme_support()
- * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is native and there is not separate AMP URL current URL.
+ * @return boolean Whether this is in AMP 'canonical' mode, that is whether it is AMP-first and there is not separate AMP URL current URL.
```
amedina: >  @return boolean Whether this is in AMP 'canonical' mode, that is whether it is AMP-first and there is not separate AMP URL current URL.

Remove `current URL` at the end

assets/src/block-editor/store/selectors.js
```diff
@@ -22,11 +22,13 @@ export function hasThemeSupport( state ) {
...
 /**
- * Returns whether the current site uses native AMP.
+ * Returns whether the current site uses AMP first (as opposed to paired).
+ *
+ * @todo Rename to isStandardAMP or isAMPFirst?
  *
  * @param {Object} state Editor state.
  *
- * @return {boolean} Whether the current site uses native AMP.
+ * @return {boolean} Whether the current site uses AMP first.
```
amedina: >  @return {boolean} Whether the current site uses AMP first.

Whether the current site **is** AMP first

includes/admin/class-amp-editor-blocks.php
```diff
@@ -44,11 +44,11 @@ public function init() {
 			add_filter( 'wp_kses_allowed_html', array( $this, 'whitelist_block_atts_in_wp_kses_allowed_html' ), 10, 2 );
 
 			/*
-			 * Dirty AMP is required when a site is in native mode but not all templates are being served
+			 * Dirty AMP is required when a site is in AMP first mode but not all templates are being served
```
amedina: AMP-first

includes/admin/class-amp-editor-blocks.php
```diff
@@ -44,11 +44,11 @@ public function init() {
...
 			 * as AMP. In particular, if a single post is using AMP-specific Gutenberg Blocks which make
 			 * use of AMP components, and the singular template is served as AMP but the blog page is not,
 			 * then the non-AMP blog page need to load the AMP runtime scripts so that the AMP components
-			 * in the posts displayed there will be rendered properly. This is only relevant on native AMP
+			 * in the posts displayed there will be rendered properly. This is only relevant on AMP first
```
amedina: AMP-first mode

includes/amp-helper-functions.php
```diff
@@ -318,7 +318,7 @@ function is_amp_endpoint() {
 		return $has_amp_query_var;
 	}
 
-	// When there is no query var and AMP is not canonical/native, then this is definitely not an AMP endpoint.
+	// When there is no query var and AMP is not canonical (AMP first), then this is definitely not an AMP endpoint.
```
amedina: AMP-first

includes/class-amp-theme-support.php
```diff
@@ -54,6 +54,30 @@ class AMP_Theme_Support {
...
+	 */
+	const TRANSITIONAL_MODE_SLUG = 'transitional';
+
+	/**
+	 * Flag used in args passed to add_theme_support('amp') to indicate transitional mode supported.
+	 *
+	 * @since 1.2
+	 * @var string
+	 */
+	const PAIRED_FLAG = 'paired';
```
kienstra: Good idea to have this in a constant.

includes/class-amp-theme-support.php
```diff
@@ -209,16 +300,33 @@ public static function read_theme_support() {
...
 				);
 			}
-			self::$support_added_via_option = false;
+
+			$is_paired = ! empty( $args[ self::PAIRED_FLAG ] );
+
+			self::$support_added_via_theme = $is_paired ? self::TRANSITIONAL_MODE_SLUG : self::STANDARD_MODE_SLUG;
+
+			/*
+			 * If the theme has transitional support, allow the user to opt for AMP first mode via an option, since a theme
```
amedina: AMP-first

includes/class-amp-theme-support.php
```diff
@@ -303,7 +411,7 @@ public static function ensure_proper_amp_location( $exit = true ) {
 
 		if ( amp_is_canonical() || is_singular( AMP_Story_Post_Type::POST_TYPE_SLUG ) ) {
 			/*
-			 * When AMP native/canonical, then when there is an /amp/ endpoint or ?amp URL param,
+			 * When AMP first/canonical, then when there is an /amp/ endpoint or ?amp URL param,
```
amedina: AMP-first

includes/class-amp-theme-support.php
```diff
@@ -2011,7 +2119,7 @@ public static function prepare_response( $response, $args = array() ) {
 
 		if ( $blocking_error_count > 0 && ! AMP_Validation_Manager::should_validate_response() ) {
 			/*
-			 * In native AMP, strip html@amp attribute to prevent GSC from complaining about a validation error
+			 * In AMP first, strip html@amp attribute to prevent GSC from complaining about a validation error
```
amedina: AMP-first

includes/options/class-amp-options-menu.php
```diff
@@ -143,112 +158,216 @@ public function add_menu_items() {
...
+							websiteExperienceInput.setCustomValidity( '' );
+						}
+
+						checkboxInputs.each( function() {
+							document.body.classList.toggle( 'amp-experience-' + this.value, this.checked );
+						} );
+					};
+
+					checkboxInputs.on( 'change', handleExperiencesUpdate );
+					handleExperiencesUpdate();
```
kienstra: It's nice how the 'Website' options only display when 'Website' is checked:

website-mode

includes/options/class-amp-options-manager.php
```diff
@@ -17,20 +17,32 @@ class AMP_Options_Manager {
...
+	 */
+	const STORIES_EXPERIENCE = 'stories';
+
 	/**
 	 * Default option values.
 	 *
 	 * @var array
 	 */
 	protected static $defaults = array(
+		'experiences'              => array( self::WEBSITE_EXPERIENCE ),
```
amedina: Should we enable both experiences by default?

includes/options/class-amp-options-menu.php
```diff
@@ -314,26 +433,26 @@ public function render_validation_handling() {
 			<?php endif; ?>
 
 			<script>
-			(function( $ ) {
-				var getThemeSupportMode = function() {
-					var checkedInput = $( 'input[type=radio][name="amp-options[theme_support]"]:checked' );
+			(function( $, standardModeSlug ) {
+				const getThemeSupportMode = () => {
```
kienstra: Nice, good idea to replace `var` with `const`

includes/options/class-amp-options-manager.php
westonruter: Good question. Not sure. Part of me says no because historically it would be adding something that people didn't ask for. Jetpack's custom content types are disabled by default. 

I'd say leave it off by default now, and then we need to revamp the `readme.txt` to indicate that Stories is a key feature alongside Websites. (That's something that should be done prior to the 1.2 release.)

Then in 1.3 we can make it enabled by default, especially when Gutenberg is no longer a dependency, and when the Stories editor is more mature.

includes/validation/class-amp-validation-manager.php
```diff
@@ -220,6 +222,49 @@ function( $query_vars ) {
...
+			'auto-draft' !== $post->post_status
+			&&
+			'trash' !== $post->post_status
+		);
+		if ( ! $supports_validation ) {
+			return false;
+		}
+
+		// Story post type always supports validation.
+		if ( AMP_Story_Post_Type::POST_TYPE_SLUG === $post->post_type ) {
```
amedina: Don't the conditions determining `$supports_validation` apply for Stories as well? Wondering which one doesn't and therefore if this point is reached.

includes/validation/class-amp-validation-manager.php
```diff
@@ -236,7 +281,7 @@ public static function is_theme_support_forced() {
 	}
 
 	/**
-	 * Return whether sanitization is forcibly accepted, whether because in native mode or via user option.
+	 * Return whether sanitization is forcibly accepted, whether because in AMP first mode or via user option.
```
amedina: AMP-first

includes/validation/class-amp-validation-manager.php
```diff
@@ -258,7 +303,7 @@ public static function is_sanitization_auto_accepted() {
 	 * - Parent admin item and first submenu item: link to non-AMP version.
 	 * - Second submenu item: link to validate the URL.
 	 *
-	 * When on native AMP response:
+	 * When on AMP first response:
```
amedina: AMP-first

includes/validation/class-amp-validation-manager.php
```diff
@@ -841,16 +877,16 @@ public static function print_edit_form_validation_status( $post ) {
 		esc_html_e( 'There is content which fails AMP validation.', 'amp' );
 		echo ' ';
 
-		// Auto-acceptance is from either checking 'Automatically accept sanitization...' or from being in Native mode.
+		// Auto-acceptance is from either checking 'Automatically accept sanitization...' or from being in AMP first (standard) mode.
```
amedina: AMP-first

includes/validation/class-amp-validation-manager.php
```diff
@@ -841,16 +877,16 @@ public static function print_edit_form_validation_status( $post ) {
...
 		if ( self::is_sanitization_auto_accepted() ) {
 			if ( ! $has_rejected_error ) {
 				esc_html_e( 'However, your site is configured to automatically accept sanitization of the offending markup. You should review the issues to confirm whether or not sanitization should be accepted or rejected.', 'amp' );
 			} else {
 				/*
-				 * Even if the 'auto_accept_sanitization' option is true, if there are non-accepted errors in non-Native mode, it will redirect to a non-AMP page.
+				 * Even if the 'auto_accept_sanitization' option is true, if there are non-accepted errors in non-standard mode, it will redirect to a non-AMP page.
```
amedina: non-Standard

includes/validation/class-amp-validation-manager.php
```diff
@@ -1955,6 +1991,25 @@ public static function print_plugin_notice() {
 	 * @return void
 	 */
 	public static function enqueue_block_validation() {
+		/*
+		 * The AMP_Validation_Manager::post_supports_validation() method is not being used here because
+		 * a post's status for validation checking can change during the life of the editor, such as when
+		 * the user to toggle AMP back on after having turned it off, and then get the validation
```
amedina: > the user to toggle AMP back on after having turned it off, and then get the validation

the user **toggles** AMP back on after having urned it off, and then **gets** the validation

tests/test-class-amp-cli.php
```diff
@@ -120,15 +120,15 @@ public function test_get_posts_that_support_amp() {
 		$this->assertEquals( $ids, AMP_CLI::get_posts_that_support_amp( $ids ) );
 		AMP_CLI::$force_crawl_urls = false;
 
-		// In Native AMP, the IDs should include all of the newly-created posts.
+		// In AMP first, the IDs should include all of the newly-created posts.
```
amedina: AMP-first

tests/test-class-amp-meta-box.php
```diff
@@ -183,7 +183,7 @@ public function test_render_status() {
 		$this->assertContains( $amp_status_markup, $output );
 		$this->assertContains( $checkbox_enabled, $output );
 
-		// This is in AMP native mode with a template that can be rendered.
+		// This is in AMP first mode with a template that can be rendered.
```
amedina: AMP-first

tests/test-class-amp-meta-box.php
```diff
@@ -242,7 +242,7 @@ public function test_get_status_and_errors() {
 			$this->instance->get_status_and_errors( $post )
 		);
 
-		// In Native AMP, there also shouldn't be errors.
+		// In AMP first, there also shouldn't be errors.
```
amedina: AMP-first

tests/test-class-amp-theme-support.php
```diff
@@ -917,7 +947,7 @@ public function test_amend_comment_form() {
 		$this->go_to( get_permalink( $post_id ) );
 		$this->assertTrue( is_singular() );
 
-		// Test native AMP.
+		// Test AMP first.
```
amedina: AMP-first

tests/validation/test-class-amp-validation-manager.php
```diff
@@ -200,8 +276,8 @@ public function test_add_admin_bar_menu_items() {
 		remove_filter( 'amp_supportable_templates', '__return_empty_array' );
 		AMP_Options_Manager::update_option( 'all_templates_supported', true );
 
-		// Admin bar item available in native mode.
-		add_theme_support( AMP_Theme_Support::SLUG, array( 'paired' => false ) );
+		// Admin bar item available in AMP first mode.
```
amedina: AMP-first

tests/validation/test-class-amp-validation-manager.php
```diff
@@ -340,8 +420,9 @@ public function test_add_rest_api_fields() {
 		);
 		$this->assert_rest_api_field_present( $post_types_non_canonical );
 
-		// Test in a Native AMP (canonical) context.
+		// Test in a AMP first (canonical) context.
```
amedina: AMP-first

includes/validation/class-amp-validation-manager.php
westonruter: Yes they do, so they're checked above:

https://github.com/ampproject/amp-wp/blob/6e12cf1cf80d9f5df9181ae37854c8d3143c9318/includes/validation/class-amp-validation-manager.php#L239-L253

Maybe I don't understand?

westonruter:
@amedina FYI, in 9737cd7 I've made it so that stories _always_ have auto-accepting sanitization, and along with that I've shortened the warning to remove wording that is not relevant in non-website experience:

image

includes/options/class-amp-options-manager.php
amedina: Agree. Makes perfect sense. Let's follow that approach.

assets/src/block-editor/index.js
```diff
@@ -16,20 +16,23 @@ import './store';
...
 const { ampLatestStoriesBlockData } = window;
 
-addFilter( 'blocks.registerBlockType', 'ampEditorBlocks/addAttributes', addAMPAttributes );
-addFilter( 'blocks.getSaveElement', 'ampEditorBlocks/filterSave', filterBlocksSave );
-addFilter( 'editor.BlockEdit', 'ampEditorBlocks/filterEdit', filterBlocksEdit, 20 );
-addFilter( 'blocks.getSaveContent.extraProps', 'ampEditorBlocks/addExtraAttributes', addAMPExtraProps );
-addFilter( 'editor.PostFeaturedImage', 'ampEditorBlocks/withFeaturedImageNotice', withFeaturedImageNotice );
-addFilter( 'editor.MediaUpload', 'ampEditorBlocks/addCroppedFeaturedImage', ( InitialMediaUpload ) => withCroppedFeaturedImage( InitialMediaUpload, getMinimumFeaturedImageDimensions() ) );
+// Add filters if AMP for Website experience is enabled.
+if ( select( 'amp/block-editor' ).isWebsiteEnabled() ) {
```
westonruter: @swissspidy How do these changes in 05dda18f7b2914c259c8c2c2fa9e525e36822ab9 look?

assets/src/block-editor/index.js
```diff
@@ -52,14 +55,24 @@ const blocks = require.context( './blocks', true, /(?<!test\/)index\.js$/ );
...
-		return;
-	}
+	const isLatestStoriesBlock = 'amp/amp-latest-stories' === name;
 
-	const blockRequiresAmp = AMP_DEPENDENT_BLOCKS.includes( name );
+	const shouldRegister = (
+		(
+			select( 'amp/block-editor' ).isWebsiteEnabled() &&
+			(
+				( ! isLatestStoriesBlock && select( 'amp/block-editor' ).isStandardMode() && AMP_DEPENDENT_BLOCKS.includes( name ) ) ||
```
swissspidy: The `! isLatestStoriesBlock` check is unnecessary here because `AMP_DEPENDENT_BLOCKS` does not include `amp/amp-latest-stories` anyway.

assets/src/block-editor/index.js
```diff
@@ -52,14 +55,24 @@ const blocks = require.context( './blocks', true, /(?<!test\/)index\.js$/ );
...
+				( isLatestStoriesBlock && select( 'amp/block-editor' ).isStoriesEnabled() )
```
swissspidy: We already have the exact same check a few lines down, so this one here seems redundant.

assets/src/block-editor/index.js
swissspidy: Good at first glance 👍 Just made some minor edits.

assets/src/block-editor/index.js
swissspidy: Just made some minor edits.

assets/src/block-editor/index.js
swissspidy: Simplified it a bit now.

includes/options/class-amp-options-manager.php
westonruter: Created #2567 for updating the readme.

PR #2555, opened 2019-06-10
Add requirement that Gutenberg 5.8 be active for Stories

westonruter:
There are bugs fixed in the block editor only available in the Gutenberg plugin. Since it's not yet in core, we need to check for the specific Gutenberg version to allow Stories to be active.


westonruter:
Failing in the WordPress 4.9 test. Since Gutenberg now requires WordPress 5.0 and greater, this needs to be part of the check.

PR #2557, opened 2019-06-11
Prevent block invalidation due to missing width/height

### LINKED ISSUE #2556: Resizing text block causes validation error
Steps to reproduce:
1. Checkout the `develop` branch, create a new Text block and add some text
2. Resize it
3. Click 'Save Draft'
4. Reload the page
5. Expected: there's no validation error
6. Actual: there's a validation error:

text-block-here

The validation error:
validation-error-here

swissspidy:
Fixes #2556

### DIFF
--- a/assets/src/stories-editor/helpers/index.js
+++ b/assets/src/stories-editor/helpers/index.js
@@ -508,25 +508,23 @@ export const wrapBlocksInGridLayer = ( element, blockType, attributes ) => {
 		height,
 	} = attributes;
 
-	const style = {
-		style: {},
-	};
+	let style = {};
 
 	if ( 'undefined' !== typeof positionTop && 'undefined' !== typeof positionLeft ) {
-		style.style = {
-			...style.style,
+		style = {
+			...style,
 			position: 'absolute',
-			top: `${ positionTop }%`,
-			left: `${ positionLeft }%`,
+			top: `${ positionTop || 0 }%`,
+			left: `${ positionLeft || 0 }%`,
 		};
 	}
 
 	// If the block has width and height set, set responsive values. Exclude text blocks since these already have it handled.
-	if ( width && height ) {
-		style.style = {
-			...style.style,
-			width: `${ getPercentageFromPixels( 'x', width ) }%`,
-			height: `${ getPercentageFromPixels( 'y', height ) }%`,
+	if ( 'undefined' !== typeof width && 'undefined' !== typeof height ) {
+		style = {
+			...style,
+			width: width ? `${ getPercentageFromPixels( 'x', width ) }%` : '0%',
+			height: height ? `${ getPercentageFromPixels( 'y', height ) }%` : '0%',
 		};
 	}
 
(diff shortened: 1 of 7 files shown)


miina:
I'm still seeing a validation error after resizing a Text block and then reloading the page:

Screen Shot 2019-06-11 at 11 00 50 AM

swissspidy:
Looks like the `onResizeStop()` call in `ResizableBox` tries to get the position from the wrong element, so `positionTop` is something like `calc(10% - 5px)` and `positionLeft` is something like `calc(5% - 5px)`.

swissspidy:
@kienstra since this whole position adjustment for text blocks is something you have implemented, you are probably better suited for looking at this.

kienstra:
Sure, I'll work on this, though I'm not sure if I'll be able to fix it 😄

miina:
I can a look as well.

miina:
@swissspidy Could you test if it works OK now?

swissspidy:
@miina If I position a block somewhere, and then increase its height like 20 times, it moves a bit to the the top every time. Demo: https://cloudup.com/c_kA2ebgcre

swissspidy:
@miina You added some `Math.round()` calls there - are these the reason? If so, are they needed?

miina:
> You added some Math.round() calls there - are these the reason? If so, are they needed?

I added the Math.round calls since the top and left attributes were saved without decimals. Since the resizing itself was using the decimals then the difference between the saved value and the displayed value was causing the position to shift when resizing (even without rotation). Now the displayed position reflects the saved value accurately and doesn't jump anymore when resizing. However, having an angle requires the top and left also to be changed during resizing and now it's more "jumpy" since the displayed value isn't using decimals anymore.

I will look more into the jumpiness issue, not sure how much time it'll take. Hopefully, it'll be quite straightforward. If it's OK not to merge this fix yet for the Text block then I'll just continue within this PR.

miina:
Alright, rounding is removed now. Would you mind testing again, @swissspidy?

assets/src/stories-editor/helpers/index.js
```diff
@@ -243,12 +243,10 @@ export const addAMPAttributes = ( settings, name ) => {
 
 	if ( isMovableBlock ) {
 		addedAttributes.positionTop = {
-			type: 'number',
```
swissspidy: Is number only for integers? Or why was this removed?

miina: It looks like it was saving it as a string. Having a type shouldn't be mandatory anyway, at some point adding a type even caused a warning, probably it has changed by now but AFAIK it's still not necessarily needed. Thoughts?

miina:
> Left a comment about the position block attributes.

Just approved the PR, if removing the type from position attributes is OK for you then let's merge.

assets/src/stories-editor/helpers/index.js
swissspidy: Hmm, according to the docs "the framework does not automatically perform type casting of meta", so if you expect numbers you'd need to do that yourselves. Would have to check the actual code to see what it does though.

Just wanted to point it out here though as it seemed to be a bit of an exception.

miina: Hmm, the logic is using `.toFixed()`, probably that's why it's saving as a `string`.
For saving as a number we'd have to convert it again with `parseFloat` then perhaps.

swissspidy: Yeah that would need a type cast again.

miina: OK, let's cast it to a number then.

miina: Added within 9bc8981.
Also added back the type for readability.

PR #2562, opened 2019-06-11
Fix resizing for formatting blocks

### LINKED ISSUE #2553: Resizing going smaller than the block for less common blocks
This issue seems to happen at least for the table, pullquote, verse, code block.

Steps to reproduce:
1. Add any of the blocks mentioned above.
2. Add some content.
3. Try resizing smaller.

Screenshots based on pullquote.

Editor:
Screen Shot 2019-06-10 at 4 08 57 PM

FE:
Screen Shot 2019-06-10 at 4 09 14 PM

miina:
Fixes #2553.

Does the following:
- Verse and Preformatted blocks: Removes resizing, based on how it displays in the frontend and the purpose of the blocks, it might not really make sense to resize these blocks -- it won't have an effect in the FE anyway. It could make sense if these blocks would have background color, we could try to implement it together with that later.
- Table block: adjusts the minimum allowed height for resizing based on rows of the table.
- Code block: Restricts resizing the height smaller than the text allows. Also hides the overflow to match the FE and editor.


assets/src/stories-editor/constants.js
```diff
@@ -31,7 +31,12 @@ export const STORY_PAGE_INNER_WIDTH = 328;
 export const STORY_PAGE_INNER_HEIGHT = 553;
 
 export const MIN_BLOCK_WIDTH = 40;
-export const MIN_BLOCK_HEIGHT = 30;
+export const MIN_BLOCK_HEIGHTS = {
```
swissspidy: Can we somehow consolidate this with `getDefaultMinimumBlockHeight`?

assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
```diff
@@ -37,7 +37,8 @@ import {
 	BLOCKS_WITH_TEXT_SETTINGS,
 	BLOCKS_WITH_COLOR_SETTINGS,
 	MIN_BLOCK_WIDTH,
-	MIN_BLOCK_HEIGHT,
+	MIN_BLOCK_HEIGHTS,
+	BLOCKS_WITHOUT_RESIZING,
```
swissspidy: Minor naming nitpick: the other constants are named `BLOCKS_WITH_...` - can we use the same here? In other words, whitelist instead of blacklist.

assets/src/stories-editor/constants.js
miina: Hmm, I'm not sure because one of these is used for the initial default height and the other one is for the minimum allowed resizing. The minimum allowed heights don't necessarily make sense as the default heights. Thoughts?

assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
miina: Changed within 90039dd.

assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
```diff
@@ -300,6 +317,34 @@ export default createHigherOrderComponent(
 							</RotatableBox>
 						</ResizableBox>
 					) }
+					{ ! needsResizing && (
+						<RotatableBox
+							blockElementId={ `block-${ clientId }` }
+							initialAngle={ rotationAngle }
+							className="amp-story-editor__rotate-container"
+							angle={ rotationAngle }
```
swissspidy: Looks like this causes prop type invalidation when `rotationAngle` is undefined. I guess we need to make the prop optional in the prop types definition.

Will have a look.

assets/src/stories-editor/components/resizable-box/index.js
```diff
@@ -54,7 +54,7 @@ const EnhancedResizableBox = ( props ) => {
 	} = props;
 
 	const isImage = 'core/image' === blockName;
-	const isBlockWithText = BLOCKS_WITH_TEXT_SETTINGS.includes( blockName );
+	const isBlockWithText = BLOCKS_WITH_TEXT_SETTINGS.includes( blockName ) || 'core/code' === blockName;
```
swissspidy: Are the verse and preformatted blocks also relevant here?

assets/src/stories-editor/constants.js
swissspidy: Fair enough. A mention in the docs would be helpful though.

assets/src/stories-editor/components/resizable-box/index.js
miina: Not at this moment since these two blocks don't have the resizing anymore (see the first point in the description). If at some point we'll add background color or any other feature that would make resizing have a point then we could add it back.

swissspidy:
@miina The test for the `amp/amp-story-cta` is currently failing because apparently `BlockEdit` is rendered twice in `withAmpStorySettings`. You can easily verify this:

1. Create new story
2. Create a second page
3. Insert CTA block
4. Notice that the edit component is rendered twice.

The `( ! isMovableBlock || isEmptyImageBlock )` check and the `! needsResizing` checks are both true for the CTA block.

miina:
> BlockEdit is rendered twice in withAmpStorySettings

Thanks! Added `isMovableBlock` condition within 03bd647.

swissspidy:
@miina Does the table block also look like this for you?

Screenshot 2019-06-11 at 17 19 26

miina:
Hmm, it doesn't look exactly like this but it's definitely off, the styling. _Checking shortly_

miina:
> @miina Does the table block also look like this for you?

Should look better now.

miina:
@kienstra If you have time, would you mind testing this PR a bit as well?

kienstra:
Sure, @miina! I'll test it now.

westonruter:
There's a merge conflict that should be resolved prior to testing.

swissspidy:
Fixed conflicts now

miina:
Thank you for testing, @kienstra.

PR #2589, opened 2019-06-12
Move amp-analytics under amp-story from body

westonruter:
Turns out that while `amp-analytics` is allowed in Stories, it must be a child of `amp-story` or `amp-story-page`, instead of the `body`. 

Interestingly, the validating sanitizer does not catch this problem so the validation error leaks through to the frontend. I'm not sure how that constraint is being specified in the validator. But that's something to look at later.

### DIFF
--- a/includes/templates/single-amp_story.php
+++ b/includes/templates/single-amp_story.php
@@ -65,14 +65,13 @@ the_post();
 			<?php
 			amp_print_story_auto_ads();
 			the_content();
+			amp_print_analytics( '' );
 			?>
 		</amp-story>
 
 		<?php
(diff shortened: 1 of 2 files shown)


westonruter:
# Testing instructions

1. Populate the Analytics screen with sample data, for example: `{"vars":{"account":"UA-XXXXXX-1"},"triggers":{"trackPageview":{"on":"visible","request":"pageview"}}}`
2. Publish a story.
3. View the story on the frontend
4. Ensure there are no AMP validation errors, as `amp-analytics` should appear right before `</amp-story>` instead of `</body>`.

swissspidy:
Verified in technical QA

PR #2591, opened 2019-06-12
Fix double opacity for text block

swissspidy:
Before:

Screenshot 2019-06-12 at 20 27 33

After:

Screenshot 2019-06-12 at 20 27 37


kienstra:
Like @westonruter mentioned, this looks good:

# Before

double-opacity

# After

no-double-opacity

And other combinations of text blocks looked good, like with no background, rotated, and resized

miina:
Unfortunately, this seems to cause an issue for amp-fit-text with little text and a long block. For example:
Screenshot 2019-06-12 at 21 47 01

The `inherit` was used for extending the background color in the editor to the full extent and also for the block styles (if I remember correctly).

swissspidy:
Removing from the project since it was reverted in #2601.

Issue #2560, opened 2019-06-11
Display a notice if a video in the video block is over 1 MB per second


kienstra:
* Splitting off a requirement from #2509 into this issue
* There should be a warning when adding a video to the Video block that is over 1 MB per second:
<img width="520" alt="video-block" src="https://user-images.githubusercontent.com/4063887/59262697-588ef600-8c05-11e9-9c45-6184af3821ee.png">

As @swissspidy mentioned:
>A suggestion of mine was to also show the warnings when selecting the video in the media modal, just like we do for the featured image.

<img width="718" alt="featured-im" src="https://user-images.githubusercontent.com/4063887/59262987-dfdc6980-8c05-11e9-979b-2d635210af68.png">

westonruter:
Also, once a video has been selected, both for the video block and for the video background, let's have the warning message indicate what the current MB/second is for the selected video. This will tell them how far from the target they are.

kienstra:
Sure, sounds good 😄

westonruter:
# Testing instructions

Given an uploaded MP4 that is relatively large given its duration, such as:

![image](https://user-images.githubusercontent.com/134745/59387904-39848700-8d1f-11e9-8579-7b21637f650d.png)

Select the video as a page background and you should see the warning amended with text like “The selected video is 10 MB per second”:

![image](https://user-images.githubusercontent.com/134745/59387938-4e611a80-8d1f-11e9-874e-e5cd834bc092.png)

westonruter:
> Also, once a video has been selected, both for the video block and for the video background, let's have the warning message indicate what the current MB/second is for the selected video. This will tell them how far from the target they are.

Fixed in #2607.

kienstra:
**Request For Testing**

Hi @csossi,
Could you please this, using the [steps here](https://github.com/ampproject/amp-wp/pull/2755#issuecomment-509507198)?

csossi:
Verified in QA

PR #2607, opened 2019-06-12
Improve selection of background video

westonruter:
* Limit video filetype to MP4
* Update “Edit Media“ to “Change Media” in button.
* Update warning to indicate the MB per second of the selected video. See https://github.com/ampproject/amp-wp/issues/2560#issuecomment-501117194.

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -43,6 +43,7 @@ import {
 	addBackgroundColorToOverlay,
 	getCallToActionBlock,
 	isVideoSizeExcessive,
+	getVideoBytesPerSecond,
 } from '../../helpers';
 import {
 	ALLOWED_CHILD_BLOCKS,
@@ -53,6 +54,7 @@ import {
 	POSTER_ALLOWED_MEDIA_TYPES,
 	MAX_IMAGE_SIZE_SLUG,
 	VIDEO_ALLOWED_MEGABYTES_PER_SECOND,
+	MEGABYTE_IN_BYTES,
 } from '../../constants';
 import './edit.css';
 
@@ -254,6 +256,7 @@ class PageEdit extends Component {
 
 		const colorSettings = this.getOverlayColorSettings();
 		const isExcessiveVideoSize = VIDEO_BACKGROUND_TYPE === mediaType && isVideoSizeExcessive( media );
+		const videoBytesPerSecond = VIDEO_BACKGROUND_TYPE === mediaType ? getVideoBytesPerSecond( media ) : -1;
 
 		return (
 			<>
@@ -300,6 +303,12 @@ class PageEdit extends Component {
 											/* translators: %d: the number of recommended megabytes per second */
 											__( 'A video size of less than %d MB per second is recommended.', 'amp' ),
 											VIDEO_ALLOWED_MEGABYTES_PER_SECOND
+										)
+									}
+									{
+										videoBytesPerSecond > 0 && ' ' + sprintf(
+											__( 'The selected video is %d MB per second.', 'amp' ),
+											Math.round( videoBytesPerSecond / MEGABYTE_IN_BYTES )
 										)
 									}
 								</Notice>
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -307,6 +307,7 @@ class PageEdit extends Component {
 									}
 									{
 										videoBytesPerSecond > 0 && ' ' + sprintf(
+											/* translators: %d: the number of actual megabytes per second */
 											__( 'The selected video is %d MB per second.', 'amp' ),
 											Math.round( videoBytesPerSecond / MEGABYTE_IN_BYTES )
 										)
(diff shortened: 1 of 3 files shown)


assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -302,6 +305,12 @@ class PageEdit extends Component {
 											VIDEO_ALLOWED_MEGABYTES_PER_SECOND
 										)
 									}
+									{
+										videoBytesPerSecond > 0 && ' ' + sprintf(
+											__( 'The selected video is %d MB per second.', 'amp' ),
```
swissspidy: Missing translator comment

westonruter:
# Testing instructions

Given an uploaded MP4 that is relatively large given its duration, such as:

image

Select the video as a page background and you should see the warning amended with text like “The selected video is 10 MB per second”:

image

If selecting an optimized video, such as:

image

Then no warning should appear:

image

assets/src/stories-editor/blocks/amp-story-page/edit.js
westonruter: Fixed in 1d20f42bc7bcb9d7e4adfca1c1ef2ffd60b4b86b

assets/src/stories-editor/helpers/index.js
```diff
@@ -1425,6 +1425,19 @@ export const getCallToActionBlock = ( pageClientId ) => {
 	return innerBlocks.find( ( { name } ) => name === 'amp/amp-story-cta' );
 };
 
+/**
+ * Gets the number of megabytes per second for the video.
+ *
+ * @param {Object} media The media object of the video.
+ * @return {number} Number of megabytes per second, or -1 if media details unavailable.
```
swissspidy: Why not just return `null` instead in that case? -1 is easy to miss and we don't use such a return type anywhere else.

westonruter: Like so?

```diff
diff --git a/assets/src/stories-editor/blocks/amp-story-page/edit.js b/assets/src/stories-editor/blocks/amp-story-page/edit.js
index e5793fda..167f5a53 100644
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -256,7 +256,7 @@ class PageEdit extends Component {
 
 		const colorSettings = this.getOverlayColorSettings();
 		const isExcessiveVideoSize = VIDEO_BACKGROUND_TYPE === mediaType && isVideoSizeExcessive( media );
-		const videoBytesPerSecond = VIDEO_BACKGROUND_TYPE === mediaType ? getVideoBytesPerSecond( media ) : -1;
+		const videoBytesPerSecond = VIDEO_BACKGROUND_TYPE === mediaType ? getVideoBytesPerSecond( media ) : null;
 
 		return (
 			<>
@@ -306,7 +306,7 @@ class PageEdit extends Component {
 										)
 									}
 									{
-										videoBytesPerSecond > 0 && ' ' + sprintf(
+										videoBytesPerSecond && ' ' + sprintf(
 											/* translators: %d: the number of actual megabytes per second */
 											__( 'The selected video is %d MB per second.', 'amp' ),
 											Math.round( videoBytesPerSecond / MEGABYTE_IN_BYTES )
diff --git a/assets/src/stories-editor/helpers/index.js b/assets/src/stories-editor/helpers/index.js
index 7d74cf03..669bc813 100644
--- a/assets/src/stories-editor/helpers/index.js
+++ b/assets/src/stories-editor/helpers/index.js
@@ -1429,11 +1429,11 @@ export const getCallToActionBlock = ( pageClientId ) => {
  * Gets the number of megabytes per second for the video.
  *
  * @param {Object} media The media object of the video.
- * @return {number} Number of megabytes per second, or -1 if media details unavailable.
+ * @return {number|null} Number of megabytes per second, or null if media details unavailable.
  */
 export const getVideoBytesPerSecond = ( media ) => {
 	if ( ! has( media, [ 'media_details', 'filesize' ] ) || ! has( media, [ 'media_details', 'length' ] ) ) {
-		return -1;
+		return null;
 	}
 	return media.media_details.filesize / media.media_details.length;
 };
```

kienstra:
**Maybe possible to upload and select a .mov file**

Steps to reproduce
1. Click 'Background Media' > 'Select Media'
2. Upload a `.mov` file, [like this](https://drive.google.com/file/d/1Re9M8TJI_xvZ6zlWiMGe11B5Bbck9_Z9/view?usp=sharing)
3. Expected: it's not possible to select it, and there's no notice
4. Actual: it's possible so click 'Select' in the Media Library, and the notice appears (A video size of less than...)


notice-appears

kienstra:
Ah, sorry. Not great timing to report that.

swissspidy:
@kienstra So you're saying it (the mime type limitation) is not working?

westonruter:
That seems like a bug in the media library. It should be respecting `allowedTypes` for both browsing and uploading.

kienstra:
>@kienstra So you're saying it (the mime type limitation) is not working?

Yes, that seems to be the case. Though it's hard to say if it's an issue with this plugin.

I'm using Core version `5.2.2-alpha-45482` and Gutenberg version `5.9.0-rc.1`.

westonruter:
It is a bug in core actually.

On the post editor I just tried adding an Image block. Then I went through the same flow, selecting a MOV file to upload and then immediately select:

image

I was then able to use the MOV as image, even though normally I would not be able to find MOV files by browsing the media modal.

miina:
@csossi See https://github.com/ampproject/amp-wp/pull/2607#issuecomment-501465226 for testing instructions.

csossi:
Verified in QA

PR #2610, opened 2019-06-12
Fix double opacity

### LINKED ISSUE #2606: Text block with background opacity is doubled
When adding a Text block with a background of say 50%, the result is an opacity of ~75% because they are getting doubled up:

Screenshot 2019-06-12 at 20 27 33

It should rather look like this:

Screenshot 2019-06-12 at 20 27 37

See #2591 for what was thought to be a fix, but it had to be reverted (#2601) because it caused problem with `amp-fit-text`.

miina:
Fixes #2606.


miina:
This seems to fix the double opacity issue. Not sure if it could cause any other issues, testing didn't show at this moment but I might be missing something.

westonruter:
Good:

image

image

image

PR #2611, opened 2019-06-12
Fix text block height when fit-text is disabled

swissspidy:
Screenshot:

Screenshot 2019-06-13 at 00 04 56

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-text/edit.css
+++ b/assets/src/stories-editor/blocks/amp-story-text/edit.css
@@ -19,7 +19,7 @@
 }
 
 .wp-block[data-type="amp/amp-story-text"] .amp-story-resize-container .components-resizable-box__handle-bottom {
-	bottom: -10px;
+	bottom: -8px;
 	width: 50px;
 	left: calc(50% - 25px);
 }
@@ -75,7 +75,7 @@ div[data-type="amp/amp-story-page"] .block-editor-inner-blocks .block-editor-blo
 	top: 2px;
 	left: 2px;
 	right: 2px;
-	bottom: 0;
+	bottom: 2px;
 }
 
 .wp-block[data-type="amp/amp-story-text"] div[draggable="true"] {
--- a/assets/src/stories-editor/blocks/amp-story-text/edit.css
+++ b/assets/src/stories-editor/blocks/amp-story-text/edit.css
@@ -19,7 +19,7 @@
 }
 
 .wp-block[data-type="amp/amp-story-text"] .amp-story-resize-container .components-resizable-box__handle-bottom {
-	bottom: -19px;
+	bottom: -10px;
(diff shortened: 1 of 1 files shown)


miina:
@swissspidy Is there a way how to assign this for non amp-fit-text only? It looks like it changes the amp-fit-text option a little bit. Will post a screenshot shortly.

miina:
This is the result with amp-fit-text (see the padding in the bottom between the block line and the background color)
Screenshot 2019-06-13 at 00 13 54


This is without this PR (ignore the height change of the block, that's my doing, just the padding is the issue)
Screenshot 2019-06-13 at 00 12 36

Thoughts?

swissspidy:
> Is there a way how to assign this for non amp-fit-text only?

Not with just CSS I think, as the targeted elements here are the parent elements of the ones with the necessary classes.

Same for the border (`.editor-block-list__layout .wp-block[data-type="amp/amp-story-text"] .editor-block-list__block-edit::before`) it seems like. That seems a bit of a fragile styling there anyway, so IMO it could just as well be fixed in 1.2.1

miina:
It might get fixed within https://github.com/ampproject/amp-wp/pull/2610, too, which removes the `::before` for amp-fit-text for the background color.

miina:
Looks like changing this
```.editor-block-list__layout .wp-block[data-type="amp/amp-story-text"] .editor-block-list__block-edit::before```

to have `bottom: 2px;` instead of `bottom: -9px;` solves the extra padding issue.

**Comment by swissspidy** [comment]
@miina @westonruter Just pushed a change that fixes the gap at the bottom.

File: assets/src/stories-editor/blocks/amp-story-text/edit.css [hunk 1]
```diff
@@ -75,9 +75,15 @@ div[data-type="amp/amp-story-page"] .block-editor-inner-blocks .block-editor-blo
 	top: 2px;
 	left: 2px;
 	right: 2px;
-	bottom: -9px;
+	bottom: 0;
```
miina: Much better! Curious why not assign `2px` as the other edges have?

westonruter:
Much better, but it looks like the bottom gap is  1px-2px more than the other edges:

image

swissspidy:
Alright. _Now_ it should be fixed.

miina:
Let's merge!

miina:
Instructions for testing:
1. Add a Text block.
2. Assign background color.
3. Unset the "Automatically fit text to container" setting.
4. Verify that background color extends to the lines of the block (there should be a tiny 2px area between the line and the block only).

For context: previously the background color was added behind the text only, even if the block's height was increased quite a lot.

csossi:
Verified in QA

Autofit on:
image

Autofit turned off:
image

PR #2613, opened 2019-06-12
Fix editing of draggable blocks in Safari

### LINKED ISSUE #2612: Unable to type in text blocks in Safari
When in Safari - all text blocks do not display text when typing, and do not update from the helper text.

This is not impacting the "code" block, however.

kienstra:
# Before
As @MackenzieHartung pointed out in #2612, it wasn't possible to type in any `[draggable]` element in Safari, meaning the Text block wasn't usable.

# After
This uses @miina's idea of `-webkit-user-select` for the `[draggable]` element that wraps the blocks.

This fixed the issue in Safari, at least in brief testing. Dragging also still works in Chrome and Firefox, after quick testing.

Fixes #2612

### DIFF
--- a/assets/src/stories-editor/components/block-mover/draggable.js
+++ b/assets/src/stories-editor/components/block-mover/draggable.js
@@ -116,9 +116,10 @@ class Draggable extends Component {
 		 * On dragging, the browser creates an image of the target, for example, the entire text block.
 		 * But there's already a clone below that's rotated in case the block is rotated,
 		 * and this can create a non-rotated duplicate of that.
-		 * So override this with an empty image.
+		 * So override this with a transparent image.
 		 */
 		const dragImage = new Image();
+		dragImage.src = 'data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=';
 		event.dataTransfer.setDragImage( dragImage, 0, 0 );
 
 		event.dataTransfer.setData( 'text', JSON.stringify( transferData ) );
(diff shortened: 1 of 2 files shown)


kienstra:
**Request For Review**

Hi @swissspidy,
Could you please review this? Thanks, I know it's late.

kienstra:
Hm, it looks like in Safari, the Text block doesn't drag if the text isn't highlighted.

swissspidy:
Travis failure is an unrelated hiccup, can be ignored.

kienstra:
This might be related to this:
https://stackoverflow.com/questions/21680363/prevent-drag-event-to-interfere-with-input-elements-in-firefox-using-html5-drag

kienstra:
This might be related:
https://github.com/react-dnd/react-dnd/issues/178#issuecomment-192021227

miina:
Not sure if we should merge this and then look into the dragging issue separately? Thoughts, @swissspidy?

@kienstra Did you have any new findings for the dragging issue meanwhile?

swissspidy:
Looking at the link that Ryan sent, we might need to test possible differences between using `-webkit-user-select: auto;` and `user-select: text;`.

Getting this one into 1.2 is better than having no fix at all.

miina:
> Looking at the link that Ryan sent, we might need to test possible differences between using -webkit-user-select: auto; and user-select: text;.

Seems to have the same result, using either of these.

miina:
@swissspidy Could you test if dragging and selecting both work for you now in Safari?

swissspidy:
It does work like a charm 😍

Tested all text-based blocks with no issues.

PR #2619, opened 2019-06-13
Add height for Video elements.

miina:
Fixes resizing for Video block by adding height to `.block-mover` and it's descendants.


swissspidy:
@csossi Still pending a new deployment on the test environment, but please test the video block resizing again after that.

csossi:
Verified in QA

PR #2622, opened 2019-06-13
Always allow switching to Reader mode, but default to mode defined by theme

westonruter:
This is a follow-up on #2550.

When using a core theme, a notice is displayed encouraging a user to switch to Standard or Transitional mode:

image

If using a non-core theme without any special theme support, it looks like:

image

However, when using a theme like [Neve](https://wordpress.org/themes/neve/) which has built-in paired AMP support, the Reader mode option is removed:

image

The user can still switch to Transitional mode (see #2312), but the Reader mode is gone. 

Additionally, if a theme is active which is AMP-first (non-paired `amp` theme support declared), then they see no options at all:

image

This can be undesirable because while the theme itself has `amp` support, the plugins being used by the theme may very well not. For such sites, having the option to stay in Reader mode is important until AMP-compatible plugins can be activated. As it stands today, the lack of Reader mode availability could cause theme makers hesitate from adding AMP compatibility, since it limits the options available to their users. It could cause users to either switch to a different theme if they want AMP, or turn off AMP if they really want the theme. 

So we should let Reader mode always be an available option. Nevertheless, upon activating the AMP plugin and the theme has `amp` support, then the default option should be Standard when `paired`  flag is present and Transitional when `paired` flag is present. As explained in #2312, if `paired` is absent then the Transitional mode should still not be made available.

This is related to #1384: Default to transitional mode with auto-sanitization for new activations, and to native mode when compatible theme/plugins active.

# Proposed Changes

No change for when a core theme is being used, continuing to default to Reader mode or when using a theme that lacks `amp` theme support.

However, when switching to a theme like Neve which has `paired` theme support, then for _existing_ plugin activations (which had previously been on Reader) then the Reader mode remains active and a green notice is displayed encouraging them to switch:

image

For _new_ AMP plugin activations, the default mode is Transitional, and again the Reader mode is still presented as an option (the notice is removed because the user switched):

image

Similarly, if the user has activated a theme that has `amp` theme support without the `paired` flag (i.e. an AMP-first theme), then upon switching to such a theme from an _existing_ Reader mode install they would see defaulting to Reader with a notice:

image

And _brand new_ AMP plugin activations would default to Standard for themes that have non-paried `amp` theme support, but with the ability to switch to Reader:

image

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -311,21 +311,18 @@ class AMP_Theme_Support {
 
 			$is_paired = ! empty( $args[ self::PAIRED_FLAG ] );
 
-			self::$support_added_via_theme = $is_paired ? self::TRANSITIONAL_MODE_SLUG : self::STANDARD_MODE_SLUG;
+			self::$support_added_via_theme  = $is_paired ? self::TRANSITIONAL_MODE_SLUG : self::STANDARD_MODE_SLUG;
+			self::$support_added_via_option = $theme_support_option;
 
-			/*
-			 * If the theme has transitional support, allow the user to opt for AMP-first mode via an option, since a theme
-			 * in transitional mode entails that it supports serving templates as both AMP and non-AMP, and this it is
-			 * able to serve AMP-first pages just as well as paired pages. Otherwise, consider that the the mode was
-			 * not set at all via option.
-			 */
+			// Make sure the user option can override what the theme has specified.
 			if ( $is_paired && self::STANDARD_MODE_SLUG === $theme_support_option ) {
 				$args[ self::PAIRED_FLAG ] = false;
 				add_theme_support( self::SLUG, $args );
-				self::$support_added_via_option = $theme_support_option;
+			} elseif ( ! $is_paired && self::TRANSITIONAL_MODE_SLUG === $theme_support_option ) {
+				$args[ self::PAIRED_FLAG ] = true;
+				add_theme_support( self::SLUG, $args );
 			} elseif ( self::READER_MODE_SLUG === $theme_support_option ) {
 				remove_theme_support( self::SLUG );
-				self::$support_added_via_option = $theme_support_option;
 			}
 		} elseif ( self::READER_MODE_SLUG !== $theme_support_option ) {
 			$is_paired = ( self::TRANSITIONAL_MODE_SLUG === $theme_support_option );
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -71,6 +71,14 @@ class AMP_Theme_Support {
 	const TRANSITIONAL_MODE_SLUG = 'transitional';
 
 	/**
+	 * Slug identifying reader website mode.
+	 *
+	 * @since 1.2
+	 * @var string
+	 */
+	const READER_MODE_SLUG = 'reader';
+
+	/**
 	 * Flag used in args passed to add_theme_support('amp') to indicate transitional mode supported.
 	 *
 	 * @since 1.2
@@ -248,7 +256,7 @@ class AMP_Theme_Support {
 			$theme_support = self::get_support_mode_added_via_theme();
 		}
 		if ( ! $theme_support ) {
-			$theme_support = 'disabled';
+			$theme_support = self::READER_MODE_SLUG;
 		}
 		return $theme_support;
 	}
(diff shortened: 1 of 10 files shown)


westonruter:
Best to suppress whitespace changes when reviewing: https://github.com/ampproject/amp-wp/pull/2622/files?w=1

amedina:
Regarding the last scenario where `... a user activates a theme with AMP-first support`:

There may be cases where the user decides to move from `Standard` to `Transitional` because, for example, a plugin was added which invalidates some of the pages/templates. In such a case, the user may switch mode and either disable AMP for those templates (which could be done without switching modes) or have two versions for those pages. 

This case is not desirable but possible.

kienstra:
Hi @westonruter,
Hope your day's off to a great start.

What do you think about moving this to 'Ready for Merging,' without more functional testing? I could test this also, if that would help.

Thanks, Weston!

westonruter:
I did test various scenarios, including variables:

* Upgrading from 1.0 to 1.2
* Testing with DB that had `theme_support` option saved for `native`, `paired`, and `disabled`.
* Activating a theme with theme support in 1.0 and ensuring the same mode is selected upon upgrade.
* On fresh installs, activating the plugin should have default mode that corresponds to mode defined in the theme.

Having your set of eyes on testing to make sure it works as expected would be helpful.

includes/class-amp-theme-support.php
```diff
@@ -303,20 +311,20 @@ public static function read_theme_support() {
...
 
-			/*
-			 * If the theme has transitional support, allow the user to opt for AMP-first mode via an option, since a theme
-			 * in transitional mode entails that it supports serving templates as both AMP and non-AMP, and this it is
-			 * able to serve AMP-first pages just as well as paired pages. Otherwise, consider that the the mode was
-			 * not set at all via option.
-			 */
-			self::$support_added_via_option = ( $is_paired && self::STANDARD_MODE_SLUG === $theme_support_option ) ? self::STANDARD_MODE_SLUG : null;
-			if ( self::STANDARD_MODE_SLUG === self::$support_added_via_option ) {
+			// Make sure the user option can override what the theme has specified.
```
kienstra: Good idea. Like you mentioned, the theme might have support, but plugins might not.

kienstra:
**Upgrading from 1.0 to 1.2**
*theme_support option looks good*

1. 'Classic' mode to 'Reader' mode:
classic-mode

2. 'Paired' mode to 'Transitional' mode:
paired-mode

3. 'Native' mode to 'Standard' mode:
native-mode

Steps Taken
1. Activate Twenty Nineteen
2. Run `wp option delete amp-options`
2. Use the [1.0.0-built](https://github.com/ampproject/amp-wp/tree/1.0.0-built) tag
3. In the AMP Options menu, observe that 'Classic' mode is selected
4. Use the `[1.2-RC1-built](https://github.com/ampproject/amp-wp/tree/1.2-RC1-built) tag
5. Verify that the right template options are present, like 'Reader' mode
6. Repeat steps 2-5 for 'Paired' and 'Native' modes

kienstra:
**Theme with declared AMP support**

>Activating a theme with theme support in 1.0 and ensuring the same mode is selected upon upgrade.

1. Activate Twenty Fifteen child them that has `add_theme_support( 'amp' )`
2. `wp option delete amp-options`
3. `git fetch --tags && git checkout 1.0.0-built`
4. Observe the AMP Settings page:

1 0-native

5. `git checkout 1.2-RC1-built`
6. In the AMP Options menu, observe that 'Standard' mode is selected, as expected:
1 2-native

7. Repeat steps 2-5, but in the child theme change `add_theme_support( 'amp' )` to `add_theme_support( 'amp', array( 'paired' => true ) )`

8. As expected, on checking out `1.2-RC1-built`, it's in 'Transitional' mode:

transitional-1 2

9. 'Transitional' mode works as expected, with a non-AMP URL and an AMP URL

kienstra:
**Testing on (simulated) fresh install**
*Also worked as expected*

>On fresh installs, activating the plugin should have default mode that corresponds to mode defined in the theme.


**Steps Taken**


1. `wp option delete amp-options # simulate a fresh install`
2. `git checkout 1.2-RC1-built`
3. Activate a Twenty Fifteen child them that has `add_theme_support( 'amp' )`
4. As expected, 'Standard' mode is selected:

1 2-rc1-native

5. `wp option delete amp-options`
6. In the child theme, change `add_theme_support( ‘amp’ )` to `add_theme_support( ‘amp’, array( ‘paired’ => true ) )`
7. As expected, 'Transitional' is selected, but the user can change it:

theme-paired

kienstra:
**Looks Good!**

Hi @westonruter,
Sorry for the delay. All of the variables that you [mentioned above](https://github.com/ampproject/amp-wp/pull/2622#issuecomment-502145863) also looked good in my testing.

See you soon!

westonruter:
Thank you for confirming!

Issue #2621, opened 2019-06-13
Text block with fit text can wrap on load and overflow container until interacted with


westonruter:
I created a Text block like so:

![image](https://user-images.githubusercontent.com/134745/59455711-9fc5e400-8dc9-11e9-8c3d-e6dadf93f8cc.png)

Upon reloading the editor, it appears as:

![image](https://user-images.githubusercontent.com/134745/59455750-ac4a3c80-8dc9-11e9-8cea-63af06074729.png)

But when I click to edit the text, then it reflows as expected:

![image](https://user-images.githubusercontent.com/134745/59455795-c552ed80-8dc9-11e9-854c-01bd1a0b17ca.png)

Reloading the page brings back the original overflow appearance.

The underlying block markup is:

```html
<!-- wp:amp/amp-story-text {"placeholder":"Write text…","tagName":"h1","align":"center","textColor":"very-dark-gray","backgroundColor":"vivid-cyan-blue","opacity":70,"autoFontSize":37,"positionTop":12.66,"positionLeft":11.59} -->
<amp-story-grid-layer template="vertical" data-block-name="amp/amp-story-text"><div class="amp-story-block-wrapper" style="position:absolute;top:12.66%;left:11.59%;width:76.22%;height:10.85%" animate-in="rotate-in-left" animate-in-delay="1801ms" animate-in-duration="700ms"><h1 style="background-color:rgba(6, 147, 227, 0.7);text-align:center;display:flex" class="wp-block-amp-amp-story-text has-text-color has-background has-very-dark-gray-color" id="aeecffc1-2be7-469b-ba5c-a64a71d03298"><amp-fit-text layout="flex-item" class="amp-text-content">Union Station!</amp-fit-text></h1></div></amp-story-grid-layer>
<!-- /wp:amp/amp-story-text -->
```

westonruter:
When removing `vertical-align:middle` from:

https://github.com/ampproject/amp-wp/blob/0172fbaabea895ce58b1b75bb41ed1e3310994c3/assets/src/stories-editor/blocks/amp-story-text/edit.css#L56-L60

This improves things _marginally_ going from:

![image](https://user-images.githubusercontent.com/134745/59456662-b3724a00-8dcb-11e9-84d7-295c913822a2.png)

To:

![image](https://user-images.githubusercontent.com/134745/59456719-d0a71880-8dcb-11e9-85d1-2b339122f516.png)

[...]

PR #2624, opened 2019-06-14
Fix Text block's padding and related resizing issues

### LINKED ISSUE #2620: Lack of bottom padding on text block
Adding a Text block with a background color shows that there is an apparent lack of bottom padding:

image

Selecting text shows it going outside of its container:

image

This is not exhibited on the frontend:

image

This is the same whether fit-text is enabled or not.

### LINKED ISSUE #2629: Alignment not working for meta blocks when amp-fit-text Off
The alignment doesn't seem to be working for the following blocks:
- Author
- Title
- Date

Note that this happens only when amp-fit-text is turned off.

To reproduce:
- Create any of the blocks mentioned above.
- Turn "Automatically fit text to container" setting off.
- Try to align.

Screenshot 2019-06-14 at 12 40 57

miina:
Fixes #2620.

Aims to fix the following issues:
- Text content going over the edge of its wrapper.
- Resizing rotated Text block not working properly.

Also fixes #2629.

- Fixes aligning meta blocks in both the editor and on the frontend

### DIFF
--- a/assets/css/amp-stories-editor.css
+++ b/assets/css/amp-stories-editor.css
@@ -105,6 +105,20 @@
 	height: 100%;
 }
 
+.wp-block[data-type="core/video"] .block-mover,
+.wp-block[data-type="core/video"] .block-mover div {
+	height: 100%;
+}
+
+.wp-block[data-type="core/video"] .block-mover .editor-rich-text.block-editor-rich-text,
+.wp-block[data-type="core/video"] .block-mover .editor-rich-text.block-editor-rich-text div {
+	height: initial;
+}
+
+.wp-block[data-type="core/video"] .wp-block-video {
+	margin: 0;
+}
+
 .wp-block-code textarea {
 	white-space: nowrap;
 }
--- a/assets/css/amp-stories.css
+++ b/assets/css/amp-stories.css
@@ -11,10 +11,18 @@ amp-story-grid-layer[template="fill"] .wp-block-image {
 	padding: 7px !important;
 }
 
+/* with amp-fit-text */
 .wp-block-amp-amp-story-post-author,
 .wp-block-amp-amp-story-post-title,
 .wp-block-amp-amp-story-post-date {
 	display: flex;
+}
+
+/* no amp-fit-text */
+.wp-block-amp-amp-story-post-author.amp-text-content,
+.wp-block-amp-amp-story-post-title.amp-text-content,
+.wp-block-amp-amp-story-post-date.amp-text-content {
+	display: block;
 }
 
 .wp-block-amp-amp-story-cta.aligncenter {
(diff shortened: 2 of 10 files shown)


miina:
Note: Looks like quite a few regressions have occurred for the Text block, would probably be good to take a look afterward and see if there's a way how to simplify all this padding and border consideration.

westonruter:
Just to note that this isn't fixing the issue yet in #2621. Upon loading the editor I see:

image

And when clicking to edit the block it adjusts:

image

kienstra:
**Approved**

Hi @miina,
Nice work, as usual. This looks good.

It looks to fix the 2 points you [mentioned above](https://github.com/ampproject/amp-wp/pull/2624#issue-288145077):

>1. Text content going over the edge of its wrapper.

## Before
before-textblock


## After
after-textblock

>2. Resizing rotated Text block not working properly.

## Before

non-amp-fit-text

## After

after-no-empty-area


3. As @westonruter mentioned, I saw that the issue from #2621 still exists. That would have been a nice added bonus to fix that, but of course this doesn't say that it does 😄 

4. The 'Title,' 'Author,' and 'Date' blocks also rotate, resize, and drag well:

## Editor

editor-here

## Front-end

fe-here

miina:
Unfortunately, just found a regression from this PR -- the text alignment doesn't seem to be working anymore for amp-fit-text... Continuing to look into this.

swissspidy:
@miina Quick testing indicates adding the alignment on the wrapper would solve this.

miina:
@swissspidy I tried that but for some reason this influences the font size -- if the alignment is to the right, the font size is much smaller. Does this happen for you as well?

swissspidy:
I haven't noticed that, no.

But just to be clear: you're saying text alignment does not work when amp-fit-text is ON? In my testing it wasn't working when amp-fit-text is OFF.

miina:
@swissspidy In my testing the alignment wasn't working in either of the cases eventually. Just added a fix that should fix that. I'll try again with the wrapper alignment, too.

On the font size changing: I could see that specifically with a very narrow (low height value) block -- the font size changed depending on the alignment. Not that this happened when testing modifying code, not within console.

miina:
@swissspidy Looks like it happens when there is more than one word:
font-change

miina:
@swissspidy Could you test again if it works better now?

swissspidy:
I was testing with "Hello World" so far and haven't run into font size oddities when changing alignment.

swissspidy:
While the meta blocks are now aligned in the editor, they don't appear to be aligned on the frontend.

Looks like they'd also need `display:block`, or at least not `display:flex` like `.wp-block-amp-amp-story-post-title` currently has by default.

swissspidy:
Let's get this in and see if it holds up in testing 👍

PR #2625, opened 2019-06-14
Export Stories Feature

### LINKED ISSUE #2464: Enable Export Option within Editor
As a user I want to see a CTA to allow me to export my AMP Story from within the editor.

AC1: The CTA should appear within the right panel of the AMP Stories editor
AC2: A user should not see the ability to export (perhaps the option is disabled) until the AMP Story is "saved".

### LINKED ISSUE #2466: Export AMP Story as Zip File
As a user, when exporting an AMP Story, I want the exported story to be in the format of a zip file.

AC1: Selecting the "export" (or like) button within the editor the initiate a request to the server to create a zip file of the AMP Story
AC2: The zip file will include all assets referenced within the story
AC3: Once the zip file is ready the user will be prompted to download the zipped story
AC4: The zip file will contain index.html as well as all assets (images, videos, etc.)
AC5: The zip file will be named the same as the title of the AMP Story
AC6: The style of the exported AMP Story matches the source AMP Story

valendesigns:
The architectural summary for this feature can be found here https://docs.google.com/document/d/1z7DPpsCB63lq7QgbEMKr23ks6ZqgZ7cEYXQ5W3sxmj0/edit#

closes #2466, closes #2464


valendesigns:
I signed it!

valendesigns:
@westonruter as promised earlier in Slack here is the work so far on this feature. There are as couple failing tests which are easy enough to fix, but the bulk of the functionality is implemented. The remaining work is sanitizing the output, getting the nonce to work when hitting the preview URL to copy the HTML, and finally adding tests.

includes/class-amp-story-post-type.php
```diff
@@ -682,6 +702,18 @@ public static function enqueue_block_editor_scripts() {
 			'ampStoriesFonts',
 			self::get_fonts()
 		);
+
+		wp_localize_script(
+			self::AMP_STORIES_SCRIPT_HANDLE,
+			'ampStoriesExport',
+			array(
+				'action'  => 'amp-story-export',
```
valendesigns: I'm thinking maybe we can put this in a property since it gets repeated all over the place.

includes/class-amp-story-post-type.php
```diff
@@ -1490,6 +1522,232 @@ public static function insert_attachment( $object, $cropped ) {
...
+		if ( 1 !== count( $posts ) ) {
+			return $posts;
+		}
+
+		$is_preview = (
+			$query->is_preview
+			&& self::POST_TYPE_SLUG === $posts[0]->post_type
+			&& isset( $_GET['story_export'] ) // phpcs:ignore
+			&& isset( $_GET['_wpnonce'] ) // phpcs:ignore
+			// phpcs:ignore && wp_verify_nonce( wp_unslash( $_GET['_wpnonce'] ), 'amp-story-export-' . $posts[0]->ID )
```
valendesigns: For some reason this is failing to validate the nonce. Even if I change the action so it doesn't use a post ID it fails. It's created above in the `$query_args` array that's passed to `get_preview_post_link()` — it should work. Thought?

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,125 @@
...
+							! empty( $this->args['slug'] )
+						);
+
+						// Used to generate the path with implode.
+						$args = array(
+							$this->args['base_url'],
+							$this->args['slug'],
+						);
+
+						if ( 'link' === $name && 'href' === $attritbute ) { // @todo this is not working.
```
valendesigns: This is not able to modify the canonical href for some reason. Maybe due to it being in the `head`?

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,125 @@
...
+
+							// Replace the asset.
+							if ( $asset && $update_path ) {
+								$args[] = 'assets';
+								$args[] = basename( $asset );
+								$node->setAttribute( $attritbute, implode( '/', $args ) );
+							}
+
+							// Add to assets array.
+							$this->assets[] = $asset;
```
valendesigns: Probably should check if there is a value for asset before adding it.

assets/src/stories-editor/plugins/export.js
```diff
@@ -0,0 +1,57 @@
...
+					} );
+			}
+		} );
+};
+
+export const name = 'amp-story-export';
+
+export const render = () => (
+	<PluginMoreMenuItem
+		icon={ ampStoriesExport.icon }
```
swissspidy: No need to pass this via `wp_localize_script()`. I'd just add this as a string here.

assets/src/stories-editor/plugins/export.js
```diff
@@ -0,0 +1,57 @@
...
+		onClick={ handleExport }
+	>
+		{ ampStoriesExport.label }
```
swissspidy: No need to pass this via `wp_localize_script()`. Can use `__()` from the `@wordpress/i18n` package.

assets/src/stories-editor/plugins/export.js
```diff
@@ -0,0 +1,57 @@
...
+import { PluginMoreMenuItem } from '@wordpress/edit-post';
+const { ampStoriesExport, FormData, URL } = window;
+
+const handleExport = () => {
+	const fromData = new FormData();
+
+	// Add the form data.
+	fromData.append( 'action', ampStoriesExport.action );
+	fromData.append( '_wpnonce', ampStoriesExport.nonce );
+	fromData.append( 'post_ID', document.getElementById( 'post_ID' ).value );
```
swissspidy: A more safe approach for getting the post ID would be this:

```js
/**
 * WordPress dependencies
 */
import { select } from '@wordpress/data';

const { getCurrentPostId } = select;

fromData.append( 'post_ID', getCurrentPostId() );
```

includes/class-amp-story-post-type.php
```diff
@@ -1490,6 +1522,232 @@ public static function insert_attachment( $object, $cropped ) {
...
+					$msg = esc_html__( 'Zip archive read error.', 'amp' );
+					break;
+				case ZipArchive::ER_SEEK:
+					$msg = esc_html__( 'Zip archive seek error.', 'amp' );
+					break;
+				default:
+					$msg = esc_html__( 'Unknown zip archive error', 'amp' );
+					break;
+			}
+			return new WP_Error( 'amp_story_zip_archive_error', $msg );
```
swissspidy: Given that these error messages are not really actionable for the user, I'd rather do something like this:

```php
/* translators: %s: error code */
$msg = sprintf( __( 'There was an error generating the ZIP archive. Error code: %s', 'amp' ), $res );
```

Good side effect of that: polyglots don't have to translate 10 new strings.

assets/src/stories-editor/plugins/export.js
valendesigns: That's how it was originally, planned on removing some of these variables before finishing 👍

assets/src/stories-editor/plugins/export.js
valendesigns: I wasn't aware of that package, thanks!

includes/class-amp-story-post-type.php
ThierryA: +1 for a class constant (or else if more suitable)

includes/sanitizers/class-amp-story-export-sanitizer.php
ThierryA: Agreed, not a big deal but could be moved to below line 114.

includes/class-amp-story-post-type.php
```diff
@@ -323,6 +329,20 @@ function( $sanitizers ) {
 			}
 		);
 
+		add_filter(
+			'amp_content_sanitizers',
+			function( $sanitizers ) {
+				if ( is_singular( self::POST_TYPE_SLUG ) && isset( $_GET['story_export'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+					$sanitizers['AMP_Story_Export_Sanitizer'] = array(
+						'base_url' => untrailingslashit( AMP_Options_Manager::get_option( 'story_export_base_url' ) ),
+						'slug'     => wp_unslash( $_GET['story_export'] ), // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
westonruter: Is passing `$_GET['story_export']` necessary here? Couldn't this be replaced with `get_post()->post_name`?

includes/class-amp-story-post-type.php
```diff
@@ -323,6 +329,20 @@ function( $sanitizers ) {
 			}
 		);
 
+		add_filter(
+			'amp_content_sanitizers',
+			function( $sanitizers ) {
+				if ( is_singular( self::POST_TYPE_SLUG ) && isset( $_GET['story_export'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
westonruter: Should this include a cap check as well? Otherwise, anyone would be able to make this request. 

Actually, since the request may not be authenticated, perhaps doing the nonce check here is what is more appropriate.

westonruter: In other words, add this as a condition:

```php
check_ajax_referer( 'amp-story-export', 'nonce', false )
```

So:

```diff
- if ( is_singular( self::POST_TYPE_SLUG ) && isset( $_GET['story_export'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+ if ( is_singular( self::POST_TYPE_SLUG ) && isset( $_GET['story_export'] ) && check_ajax_referer( 'amp-story-export', 'nonce', false ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```

includes/class-amp-story-post-type.php
```diff
@@ -1490,6 +1522,232 @@ public static function insert_attachment( $object, $cropped ) {
...
+			);
+
+			// Passed to `wp_remote_get()`.
+			$args = array(
+				'timeout'   => 30,
+				'sslverify' => false,
+			);
+
+			// Get the preview URL.
+			$response = wp_remote_get( get_preview_post_link( $post, $query_args ), $args );
```
westonruter: Just thinking that the cookies in the current request may need to be passed along here, so that draft stories and private stories can be exported. You can refer to the validation manager for example code here:

https://github.com/ampproject/amp-wp/blob/9342ddab702c06ad496cd36ff1f21f30ffc12014/includes/validation/class-amp-validation-manager.php#L1833-L1844

westonruter: Additionally, for doing nonce checking for unauthenticated users, refer also to:

https://github.com/ampproject/amp-wp/blob/9342ddab702c06ad496cd36ff1f21f30ffc12014/includes/validation/class-amp-validation-manager.php#L1633-L1634

https://github.com/ampproject/amp-wp/blob/9342ddab702c06ad496cd36ff1f21f30ffc12014/includes/validation/class-amp-validation-manager.php#L1822-L1826

includes/class-amp-story-post-type.php
```diff
@@ -1490,6 +1522,232 @@ public static function insert_attachment( $object, $cropped ) {
...
+	 * @return WP_Error
+	 */
+	private static function generate_export( $post_id ) {
+		$post = get_post( $post_id );
+
+		if ( ! $post ) {
+			return new WP_Error( 'amp_story_export_invalid_post', esc_html__( 'The AMP Story does not exist.', 'amp' ) );
+		}
+
+		$slug = sanitize_title( $post->post_title, $post->ID );
```
westonruter: Instead of `post_title`, why not `$post->post_name`?

includes/class-amp-story-post-type.php
```diff
@@ -1490,6 +1522,232 @@ public static function insert_attachment( $object, $cropped ) {
...
+
+			// Create the zip directory.
+			$zip->addEmptyDir( $slug );
+
+			// Add README.txt file.
+			$zip->addFromString( $slug . '/README.txt', 'temporary content...' );
+
+			// Passed to `get_preview_post_link()` for nonce access and to sanitize the output.
+			$query_args = array(
+				'story_export' => $slug,
```
westonruter: Per above, I don't think the `$slug` need be passed here.

includes/class-amp-story-post-type.php
```diff
@@ -1490,6 +1522,232 @@ public static function insert_attachment( $object, $cropped ) {
...
+
+		if ( ! $post ) {
+			return new WP_Error( 'amp_story_export_invalid_post', esc_html__( 'The AMP Story does not exist.', 'amp' ) );
+		}
+
+		$slug = sanitize_title( $post->post_title, $post->ID );
+		$file = sprintf( '%s.zip', $slug );
+
+		$zip = new ZipArchive();
+		$res = $zip->open( $file, ZipArchive::CREATE | ZipArchive::OVERWRITE );
```
westonruter: What directory is this being added to? What if two people are trying to export the same story at the same time? I think instead of `$file` this should rather use `tempnam()`

includes/class-amp-story-post-type.php
westonruter: Probably because the user is not authenticated for the subsequent request. See above about passing the $_COOKIE` through in the request.

includes/class-amp-story-post-type.php
```diff
@@ -1490,6 +1522,232 @@ public static function insert_attachment( $object, $cropped ) {
...
+	/**
+	 * Replaces the array of retrieved posts after they've been fetched and
+	 * internally processed with the exported story.
+	 *
+	 * @param WP_Post[] $posts Array of post objects.
+	 * @param WP_Query  $query The WP_Query instance (passed by reference).
+	 *
+	 * @return WP_Post[] The post results array.
+	 */
+	public static function display_export_preview( $posts, $query ) {
```
westonruter: Instead of this I suggest passing the user's `$_COOKIE` through in the call to `wp_remote_get()`

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,125 @@
...
+		if ( 0 < $num_nodes ) {
+			for ( $i = $num_nodes - 1; $i >= 0; $i-- ) {
+				$node = $nodes->item( $i );
+
+				if ( ! $node instanceof DOMElement ) {
+					continue;
+				}
+
+				if ( ! empty( $attritbutes ) ) {
+					foreach ( $attritbutes as $attritbute ) {
```
westonruter: Typo in var names.

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,125 @@
...
+			'publisher-logo-src',
+			'publisher',
+			'poster-portrait-src',
+			'poster-square-src',
+			'poster-landscape-src',
+		);
+		$this->sanitize_assets( 'amp-story', $attrs );
+
+		// AMP Image Nodes.
+		$this->sanitize_assets( 'amp-img', array( 'src' ) );
```
westonruter: Also need to check `srcset` attribute.

westonruter: Also need to look for `amp-video` and its `poster` attribute.

includes/class-amp-story-post-type.php
valendesigns: If memory serves when the title has not been set then `$post->post_name` is empty.

includes/sanitizers/class-amp-story-export-sanitizer.php
valendesigns: `$update_path` is used to replace asset URLs where `$this->assets[]` is used to store the original URL and we only do a replacement if there is a need for it. Adding after line 114 would cause the array to be empty unless we were also replacing URLs.

includes/class-amp-story-post-type.php
```diff
@@ -1490,6 +1522,232 @@ public static function insert_attachment( $object, $cropped ) {
...
+			wp_send_json_error(
+				array(
+					'errorMessage' => esc_html__( 'You do not have the required permissions to export AMP stories.', 'amp' ),
+				),
+				403
+			);
+		}
+
+		// We need the ZipArchive class to make this work.
+		if ( ! class_exists( 'ZipArchive', false ) ) {
```
swissspidy: Might be a good thing to add to the `suggest` list in `composer.json`

assets/src/stories-editor/plugins/export.js
```diff
@@ -0,0 +1,57 @@
...
+
+	// Request the export.
+	fetch( ampStoriesExport.ajaxUrl, {
+		method: 'POST',
+		body: fromData,
+	} )
+		.then( ( response ) => {
+			if ( response.ok ) {
+				// Handle the returned blob data.
```
swissspidy: Might be worth adding a snackbar notification upon successful download.

https://developer.wordpress.org/block-editor/components/snackbar/

assets/src/stories-editor/plugins/export.js
```diff
@@ -0,0 +1,57 @@
...
+							document.body.appendChild( a );
+							a.click();
+							a.remove();
+						}
+					} );
+			} else {
+				// Handle the returned JSON error.
+				response.json()
+					.then( ( error ) => {
+						// @todo should we show this in a different way?
```
swissspidy: I think an error notice makes most sense:

https://developer.wordpress.org/block-editor/components/notice/

includes/class-amp-story-post-type.php
```diff
@@ -1519,13 +1571,105 @@ public static function insert_attachment( $object, $cropped ) {
...
+		if ( isset( $_POST['action'] ) && ( 'query-attachments' === $_POST['action'] || 'upload-attachment' === $_POST['action'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Missing
+			$image_sizes[ self::MAX_IMAGE_SIZE_SLUG ] = $full_size_name;
+		} elseif ( get_post_type() && self::POST_TYPE_SLUG === get_post_type() ) {
+			$image_sizes[ self::MAX_IMAGE_SIZE_SLUG ] = $full_size_name;
+			unset( $image_sizes['full'] );
+		}
+
+		return $image_sizes;
+	}
+
```
valendesigns: Not sure why it's adding the three methods above but will look into what's happening in the morning.

valendesigns: Never mind, it's moved it from lower in the code. Disregard the last message.

assets/src/stories-editor/plugins/export.js
valendesigns: @swissspidy How would we know if the download was successful or failed?

valendesigns: Or if the user canceled the download?

includes/class-amp-story-post-type.php
valendesigns: I tested this and `$post->post_name` is empty when the title is blank.

includes/class-amp-story-post-type.php
valendesigns: I've used `wp_tempnam` and then replace the filename with the header.

includes/class-amp-story-post-type.php
valendesigns: Thank you for the direction here. I was able to rework this.

assets/src/stories-editor/plugins/export.js
swissspidy: Le me rephrase: after clicking on the button, a snackbar notification can be added so the user gets some feedback.

Just like when you click on "Copy All Content" in the Gutenberg block editor.

assets/src/stories-editor/plugins/export.js
```diff
@@ -0,0 +1,73 @@
+/* global fetch */
```
swissspidy: We're more and more getting away from using `/* global ... */` annotations in favor of a more explicit way:

```js
const { fetch } = window;
```

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+				],
+				403
+			);
+		}
+
+		// Bail if the user has not saved the story yet.
+		if ( 'auto-draft' === get_post_status( wp_unslash( $_POST['post_ID'] ) ) ) {
+			wp_send_json_error(
+				[
+					'errorMessage' => esc_html__( 'Save the AMP story before exporting.', 'amp' ),
```
swissspidy: Are we able to disable the export button in the UI already when the story hasn't been saved yet? Or how does one get into this scenario?

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+		$res = $zip->open( $file, ZipArchive::CREATE | ZipArchive::OVERWRITE );
+
+		// Generate ZIP and send headers.
+		if ( true === $res ) {
+
+			// Create the zip directory.
+			$zip->addEmptyDir( $slug );
+
+			// Add README.txt file.
+			$zip->addFromString( $slug . '/README.txt', 'temporary content...' );
```
swissspidy: **To do:** We need to define the content here.

Also, it might be easier to have this as an actual file in the plugin that we can just include here. Makes changing the text easier.

swissspidy:
I just quickly exported a story I've created a while ago that contains some images.

Since I didn't change the export base URL, the images still referenced their original location (as expected). However, there was still an empty assets folder in the ZIP, which was a bit odd.

After that, I changed the base URL, but I got the same result (empty folder).

includes/options/class-amp-options-menu.php
```diff
@@ -633,6 +647,40 @@ public function render_caching() {
...
+				position: relative;
+				top: -2px;
+				font-weight: 400;
+				line-height: 1;
+			}
+		</style>
+		<fieldset <?php disabled( ! current_user_can( 'edit_files' ) ); ?>>
+			<p>
+				<label for="story_export_base_url">
+					<strong><?php echo wp_kses_post( __( 'Base URL for exported stories. <span>Beta</span>', 'amp' ) ); ?></strong>
```
kienstra: Like @swissspidy [mentioned](https://github.com/ampproject/amp-wp/pull/2625#issuecomment-509644664), it looks like if this isn't populated:
not-populated

...the exported file looks to have an empty `assets/` directory, and I didn't see a notice or warning of this:

assets-dir-empty

includes/class-amp-story-post-type.php
valendesigns: When the user clicks the button and has not yet saved the story this error message will be returned. However, we need to also handle times when the user has made changes but also has not saved the story, which I'm not sure exactly how to handle that at the moment. Maybe we could initiate an auto save?

includes/class-amp-story-post-type.php
valendesigns: I think this is probably the last piece to get finished. But I agree, a file would make sense.

valendesigns:
@kienstra Not sure what's happening with the images. It's working perfectly with my VVV setup. Is this a new story or an older one?

kienstra:
>@kienstra Not sure what's happening with the images. It's working perfectly with my VVV setup. Is this a new story or an older one?

Hi @valendesigns, it's with a new story created to test this. I'll test it again with an existing story.

kienstra:
Hi @valendesigns,
It looks like the `assets/` directory is still empty in my local when I:

1. Checked out `develop`
2. Created a story with an image block
3. Checked out this PR's branch, `add/2466-export`
4. `composer install && npm install && npm audit fix && npm run dev`
5. Went back to the same story
6. Exported it:
exported-story

7. Expected: the `.zip` file has an image in the `assets/` directory when it's extracted
8. Actual: it didn't have an image

This happened both with 'Base URL for exported stories.' populate with `https://foobar.com`, and with it empty.

Here's the `.zip` file:

[test-landscape-story-image-300-193.zip](https://github.com/ampproject/amp-wp/files/3375073/test-landscape-story-image-300-193.zip)

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,168 @@
...
+		 * @param string $asset The original URL.
+		 *
+		 * @return bool|string Returns false when $update_path is false, else the new URL.
+		 */
+		$get_asset_path = function( $asset ) use ( $update_path ) {
+			if ( $asset && $update_path ) {
+				$args = [
+					$update_path,
+					'assets',
+					basename( $asset ),
```
westonruter: What happens if there are two files uploaded (or linked to) with the same filename? I'm thinking there could be one at like `/wp-content/uploads/2018/12/christmas.jpg` and another at `/wp-content/uploads/2019/12/christmas.jpg`. So perhaps there should be a some some logging of which asset URLs have been processed and what their filenames are. If a new asset URL is processed and it would end up getting the same filename as a previous, then it should get some auto-increment added to the filename.

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,168 @@
...
+				return implode( '/', $args );
+			}
+			return false;
+		};
+
+		if ( 0 < $num_nodes ) {
+			for ( $i = $num_nodes - 1; $i >= 0; $i-- ) {
+				$node = $nodes->item( $i );
+
+				if ( ! $node instanceof DOMElement ) {
```
westonruter: We already know that each `$node` is going to be an element, so this can be removed.

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,168 @@
...
+					'assets',
+					basename( $asset ),
+				];
+				return implode( '/', $args );
+			}
+			return false;
+		};
+
+		if ( 0 < $num_nodes ) {
+			for ( $i = $num_nodes - 1; $i >= 0; $i-- ) {
```
westonruter: This could be made flatter by doing:

```php
foreach ( $nodes as $node ) {
```

In other sanitizers we do this dance with walking backwards over the array in order to prevent causing problems when a node gets removed, since the node list is live. But since none of the nodes are getting removed, there's no need for the dance.

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,168 @@
...
+				$node = $nodes->item( $i );
+
+				if ( ! $node instanceof DOMElement ) {
+					continue;
+				}
+
+				if ( ! empty( $attributes ) ) {
```
westonruter: To reduce nesting, this could be:

```
if ( empty( $attributes ) ) {
    return;
}
```

But actually, shouldn't that be moved up to the top of the method? But is it even actually needed since it is always going to be an array? And looping over an empty array just won't do anything.

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,168 @@
...
+		}
+	}
+
+	/**
+	 * Add and sanitize the export assets for the AMP Story.
+	 *
+	 * @param string $name       The DOMElement name.
+	 * @param array  $attributes The DOMElement attributes.
+	 */
+	public function sanitize_assets( $name, $attributes ) {
```
westonruter: Suggest adding `array` as type specifier:

```diff
- public function sanitize_assets( $name, $attributes ) {
+ public function sanitize_assets( $name, array $attributes ) {
```

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,168 @@
...
+		if ( ! empty( $this->assets ) && ! headers_sent() ) {
+			header( 'X-AMP-Export-Assets: ' . implode( ',', $this->assets ) );
+		}
+	}
+
+	/**
+	 * Add and sanitize the export assets for the AMP Story.
+	 *
+	 * @param string $name       The DOMElement name.
+	 * @param array  $attributes The DOMElement attributes.
```
westonruter: Instead of `array` this could be `string[]`

includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -0,0 +1,168 @@
...
+	 * @var array
+	 */
+	protected $DEFAULT_ARGS = [
+		'base_url' => '',
+	];
+
+	/**
+	 * Default assets.
+	 *
+	 * @var array
```
westonruter: Could be `string[]` instead of `array` for more specificity.

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+
+				// Replace the Canonical URL in the document head.
+				$html = preg_replace( '/(?<=<link rel="canonical" href=")(.*)(?=">)/', $export_args['canonical_url'], $html );
+			}
+
+			// Add index.html file.
+			$zip->addFromString( $slug . '/index.html', $html );
+
+			// Get the assets.
+			$assets = explode( ',', $response['headers']['x-amp-export-assets'] );
```
westonruter: See note below for how the list of assets could actually be extracted from an HTML comment in the body of the response.

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+			$export_args = self::get_export_args( $slug );
+
+			if ( $export_args['canonical_url'] ) {
+
+				// Clean up the source.
+				$html = preg_replace( '/(?=<!--)([\s\S]*?-->)/', '', $html );
+				$html = str_replace( '/*# sourceURL=amp-custom.css */', '', $html );
+
+				// Replace the Canonical URL in the document head.
+				$html = preg_replace( '/(?<=<link rel="canonical" href=")(.*)(?=">)/', $export_args['canonical_url'], $html );
```
westonruter: Would this be better done in the sanitizer? In other words, should the `$export_args['canonical_url']` be passed among the `$query_args`?

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+			// Get the HTML from the response body.
+			$html = $response['body'];
+
+			$export_args = self::get_export_args( $slug );
+
+			if ( $export_args['canonical_url'] ) {
+
+				// Clean up the source.
+				$html = preg_replace( '/(?=<!--)([\s\S]*?-->)/', '', $html );
+				$html = str_replace( '/*# sourceURL=amp-custom.css */', '', $html );
```
westonruter: This can be retained actually. It's not actually pointing to any actual URL. It's merely here for the sake of code coverage in DevTools so that there is a filename to identify a coverage for.

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+
+			// Get the HTML from the response body.
+			$html = $response['body'];
+
+			$export_args = self::get_export_args( $slug );
+
+			if ( $export_args['canonical_url'] ) {
+
+				// Clean up the source.
+				$html = preg_replace( '/(?=<!--)([\s\S]*?-->)/', '', $html );
```
westonruter: What is this doing?

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+			// Passed to `get_preview_post_link()` for nonce access and to sanitize the output.
+			$query_args = [
+				'story_export' => true,
+				'_wpnonce'     => wp_create_nonce( self::AMP_STORIES_AJAX_ACTION ),
+			];
+
+			// Passed to `wp_remote_get()`.
+			$args = [
+				'cookies'     => wp_unslash( $_COOKIE ), // Pass along cookies so private pages and drafts can be accessed.
+				'timeout'     => 15, // Increase from default of 5 to give extra time for the plugin to identify the sources for any given validation errors; also, response caching is disabled when validating.
```
westonruter: I  think you can remove this: `also, response caching is disabled when validating`, since a validation request actually is not being performed. And the post-processor cache should actually be working for this request.

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+		}
+
+		$slug = sanitize_title( $post->post_title, $post->ID );
+		$file = wp_tempnam( $slug );
+
+		$zip = new ZipArchive();
+		$res = $zip->open( $file, ZipArchive::CREATE | ZipArchive::OVERWRITE );
+
+		// Generate ZIP and send headers.
+		if ( true === $res ) {
```
westonruter: Suggestion: since this `if` statement is long, it could be made easier to read if the `else` condition were put up here, which would also reduce nesting. In other words:

```php
if ( true !== $res ) {
	/* translators: %s is the ZipArchive error code. */
	return new WP_Error( 'amp_story_zip_archive_error', sprintf( esc_html__( 'There was an error generating the ZIP archive. Error code: %s', 'amp' ), $res ) );
}
```

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+			);
+		}
+
+		// We need the ZipArchive class to make this work.
+		if ( ! class_exists( 'ZipArchive', false ) ) {
+			wp_send_json_error(
+				[
+					'errorMessage' => esc_html__( 'The ZipArchive class is required to export AMP stories.', 'amp' ),
+				],
+				403
```
westonruter: Trivial point, but I think this should be `400`  rather than `403`.

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+	public static function handle_export() {
+		check_ajax_referer( self::AMP_STORIES_AJAX_ACTION, 'nonce' );
+
+		// The user must have the correct permissions.
+		if ( ! current_user_can( 'edit_files' ) ) {
+			wp_send_json_error(
+				[
+					'errorMessage' => esc_html__( 'You do not have the required permissions to export AMP stories.', 'amp' ),
+				],
+				403
```
westonruter: Technically this should be `401` since the user is unauthorized.

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+		// Generate and export the archive.
+		$export = self::generate_export( wp_unslash( $_POST['post_ID'] ) );
+
+		// Export failed.
+		if ( is_wp_error( $export ) ) {
+			wp_send_json_error(
+				[
+					'errorMessage' => $export->get_error_message(),
+				],
+				403
```
westonruter: I think this should default to `400` but it should try to get the error code from the `WP_Error` similar to what is done in `\WP_REST_Server::error_to_response()`.  In other words:

```php
		$error_data = $error->get_error_data();

		if ( is_array( $error_data ) && isset( $error_data['status'] ) ) {
			$status = $error_data['status'];
		} else {
			$status = 500;
		}
```

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1637,204 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+			);
+		}
+
+		// Failed to export for an unknown reason not related to generating the archive.
+		wp_send_json_error(
+			[
+				'errorMessage' => esc_html__( 'Could not generate the AMP story archive.', 'amp' ),
+			],
+			403
```
westonruter: Here `500` is the most appropriate.

includes/sanitizers/class-amp-story-export-sanitizer.php
valendesigns: @westonruter Perhaps instead of trying to track and ensure all the images are unique we convert `/wp-content/uploads/2019/12/christmas.jpg` to `2019-12-christmas.jpg` when moving it to the assets directory. This would avoid collisions.

westonruter: Replacing all slashes with hyphens is a great idea.

includes/class-amp-story-post-type.php
valendesigns: It was not working in the sanitizer.

includes/class-amp-story-post-type.php
valendesigns: Removing the HTML comments

westonruter: I don't think this is necessary, is it?

westonruter: If this were needed, it would be better done in the sanitizer anyway since it could easily locate all the comment nodes via XPath. But we aren't removing comments elsewhere for optimization,  so I think they should be left in place.

valendesigns: Well, not necessary but probably good to do since it references the original WordPress install and related URLs. For example, it removes this block.

```
<!--
    35 B (0%): link#wp-block-library-css[rel=stylesheet][href=http://local.wordpress.test/wp-content/plugins/gutenberg/build/block-library/style.css?ver=1562098564][type=text/css][media=all]
   219 B (25%): link#amp-stories-css[rel=stylesheet][href=http://local.wordpress.test/wp-content/plugins/amp/assets/css/amp-stories.css?ver=1.2.1-alpha][type=text/css][media=all]
     0 B: style#amp-stories-inline-css[type=text/css]
   289 B (25%): link#amp-stories-frontend-css[rel=stylesheet][href=http://local.wordpress.test/wp-content/plugins/amp/assets/css/amp-stories-frontend.css?ver=1.2.1-alpha][type=text/css][media=]
    84 B: amp-story-page#127ad995-f0c1-47a2-81aa-02fef257f373.wp-block-amp-amp-story-page amp-wp-74885c3
    86 B: amp-img.amp-wp-6ed12b8[layout=fill][src=https://localhost/assets/Nature.jpg]
   123 B: div.amp-story-block-wrapper amp-wp-41b864c
    75 B: h1#ab02e26f-4a27-4da4-a00a-6fb09c27a11c.wp-block-amp-amp-story-text amp-wp-f7dbba7
Total included size: 911 bytes (6% of 13,850 total after tree shaking)
-->
```

westonruter: I see. That comment should instead be removed in another way (via #1843). In particular:

```php
    return $sanitizers;
} );
```

That could be added to the same logic that adds the exporter sanitizer in the first place.

miina:
@kienstra and @swissspidy, on the empty assets folder:
Note that the assets are received using `wp_remote_get` which fails if the SSL certificate is not valid (cURL error 60). Could this be the issue where you're testing?

cc @valendesigns

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1589,222 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+		// Add the assets.
+		if ( ! empty( $assets ) ) {
+
+			// Create the empty assets directory.
+			$zip->addEmptyDir( $slug . '/assets' );
+
+			foreach ( $assets as $asset ) {
+				$response = wp_remote_get( $asset );
+				if ( is_array( $response ) && ! empty( $response['body'] ) ) {
+					$zip->addFromString( $slug . '/assets/' . self::export_image_basename( $asset ), $response['body'] );
```
miina: Perhaps it would make sense to add an additional notice if some of the assets failed to load? E.g. if `$response` is an error, display a message to express that some of the assets failed to load; and potentially logging the errors somewhere, too. Would avoid debugging / support cases in case of server issues. It could fail due to SSL issues, timing out, etc. Thoughts?

cc @westonruter

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1589,222 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+	/**
+	 * Ajax handler to export the story ZIP archive.
+	 *
+	 * This method returns an error as JSON and the binary data on success.
+	 */
+	public static function handle_export() {
+		check_ajax_referer( self::AMP_STORIES_AJAX_ACTION, 'nonce' );
+
+		// The user must have the correct permissions.
+		if ( ! current_user_can( 'edit_files' ) ) {
```
miina: This would allow export for only admins, probably editors and authors should also be able to export Stories?

Also, it looks like `edit_files` is not actually a used capability anymore, it's assigned to the admins but doesn't seem to be used for anything.

Looks like the tmp folder is used for generating the files before downloading, perhaps we should remove this condition and allow export for anyone who can edit and create AMP Stories. Thoughts?

cc @swissspidy @westonruter

includes/class-amp-story-post-type.php
swissspidy: As for SSL issues, I would probably just disable SSL verification for the `wp_remote_get` call.

includes/class-amp-story-post-type.php
swissspidy: > allow export for anyone who can edit and create AMP Stories

Agreed. If the current user can edit the current story, they should be able to export it.

kienstra:
**SSL**

Hi @miina,
>@kienstra and @swissspidy, on the empty assets folder:
Note that the assets are received using wp_remote_get which fails if the SSL certificate is not valid (cURL error 60). Could this be the issue where you're testing?

Good idea. On adding `[ 'sslverify' => false ]`, the images appeared in the exported file's `assets/` directory.

```diff
diff --git a/includes/class-amp-story-post-type.php b/includes/class-amp-story-post-type.php
index 5d5f7937..983cf2fd 100644
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -1783,7 +1783,7 @@ class AMP_Story_Post_Type {
                        $zip->addEmptyDir( $slug . '/assets' );
 
                        foreach ( $assets as $asset ) {
-                               $response = wp_remote_get( $asset );
+                               $response = wp_remote_get( $asset, [ 'sslverify' => false ] ) );
                                if ( is_array( $response ) && ! empty( $response['body'] ) ) {
                                        $zip->addFromString( $slug . '/assets/' . self::export_image_basename( $asset ), $response['body'] );
                                }
```

includes/class-amp-story-post-type.php
valendesigns: If we return an error at this stage the whole process fails. It's kind of an all or nothing scenario.

includes/class-amp-story-post-type.php
valendesigns: It's been updated to `publish_posts`.

includes/class-amp-story-post-type.php
valendesigns: I've updated the code to address the SSL issue.

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1589,222 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+	/**
+	 * Ajax handler to export the story ZIP archive.
+	 *
+	 * This method returns an error as JSON and the binary data on success.
+	 */
+	public static function handle_export() {
+		check_ajax_referer( self::AMP_STORIES_AJAX_ACTION, 'nonce' );
+
+		// The user must have the correct permissions.
+		if ( ! current_user_can( 'publish_posts' ) ) {
```
swissspidy: Better to use `current_user_can( 'publish_post', wp_unslash( $_POST['post_ID'] ) )` for more granular control.

Of course only after an initial `isset( $_POST['post_ID'] )` check.

Also, `wp_unslash( $_POST['post_ID'] )` is used multiple times, so maybe something like `$post_id = wp_unslash( $_POST['post_ID'] )` can help make this more DRY.

assets/src/stories-editor/plugins/export.js
```diff
@@ -0,0 +1,80 @@
...
+						} );
+					} );
+			}
+		} );
+};
+
+export const name = 'amp-story-export';
+
+export const render = () => (
+	<PluginMoreMenuItem
```
swissspidy: Gutenberg in some places uses `wp.data.select( 'core/editor' ).getCurrentPost()._links['wp:action-publish' ]` to determine whether the user can publish a post.

See https://github.com/WordPress/gutenberg/blob/94f54892c214ebd014ec126a9575b7e902198ad9/packages/editor/src/components/post-schedule/check.js for example.

With `withSelect` and `ifCondition` HOCs that is quite simple to achieve here, in order to hide this for users who don't have the permission.

assets/src/stories-editor/plugins/export.js
```diff
@@ -0,0 +1,80 @@
...
+		icon={ 'media-archive' }
+		onClick={ handleExport }
+	>
+		{ __( 'AMP Story Export', 'amp' ) }
```
swissspidy: Let's rename this to `Export Story`.

includes/class-amp-story-post-type.php
```diff
@@ -1553,4 +1589,222 @@ public static function add_new_max_image_size( $image_sizes ) {
...
+				],
+				403
+			);
+		}
+
+		// We need the ZipArchive class to make this work.
+		if ( ! class_exists( 'ZipArchive', false ) ) {
+			wp_send_json_error(
+				[
+					'errorMessage' => esc_html__( 'The ZipArchive class is required to export AMP stories.', 'amp' ),
```
swissspidy: Let's use a placeholder for `ZipArchive` and add a translator comment accordingly.

includes/options/class-amp-options-menu.php
```diff
@@ -633,6 +647,40 @@ public function render_caching() {
...
 
+	/**
+	 * Render the stories export settings section.
+	 *
+	 * @since 1.2
+	 */
+	public function render_stories_export() {
+		?>
+		<style>
+			label[for="story_export_base_url"] span {
```
swissspidy: Not sure if we need the "Beta" label here, but can we combine this selector with the existing one used earlier on?

valendesigns: I'll just remove the label.

PR #2626, opened 2019-06-14
Prevent text block overflowing with amp-fit-text

### LINKED ISSUE #2621: Text block with fit text can wrap on load and overflow container until interacted with

kienstra:
* Removes the Text block's `componentDidMount()`, which called `maybeUpdateFontSizes()` 
* It looks like this [incorrectly calculated](https://github.com/ampproject/amp-wp/issues/2621#issuecomment-501899584) the font size, assigning a higher size to `autoFontSize`, and therefore the `style` attribute.
* For example, on page load, the Text block's `autoFontSize` attribute was originally (and correctly) `37`.
* Then, the `componentDidMount()` method called `maybeUpdateFontSizes()`. This increased that to `39`, which caused the text to overflow.
* Still, this might cause a regression, and it could use more testing.

Fixes #2621


kienstra:
I'm testing this with the templates now. There could be complications there.

kienstra:
https://github.com/ampproject/amp-wp/pull/2626/commits/9f259cc86a756d7286b44cc185707fec5ce155f8 causes a regression when selecting a template, in that it sometimes doesn't size the text correctly.

swissspidy:
Regarding 9f259cc86a756d7286b44cc185707fec5ce155f8, maybe it can be handled in `componentDidUpdate`? Only when related props/attributes have changed of course.

kienstra:
>Regarding 9f259cc, maybe it can be handled in componentDidUpdate? Only when related props/attributes have changed of course.

Good point. I think the `maybeUpdateFontSizes()` in [componenteDidUpdate](https://github.com/ampproject/amp-wp/blob/aa49fd5d3e45f99de72b7933b5b67a62d7b1f6a0/assets/src/stories-editor/blocks/amp-story-text/edit.js#L53) might be enough.

There was an issue with the `fandom-title.html` template's 'Travel Tips' text block not displaying correctly on selecting it. But I'm working on that now.

kienstra:
https://github.com/ampproject/amp-wp/pull/2626/commits/4af71a0feed35e8d637b423ea0c126355f8e2e0a addresses the problem of the 'Travel Tips' text block in the `title-page.html` template having the wrong size on selecting the template.

It changes that template to be `amp-fit-text`. Even though it wasn't `amp-fit-text` in its `title-page.html` file, it looks like on selecting the template, the block became `amp-fit-text`. 

So this keeps it uniform, and should prevent having to run `maybeUpdateFontSizes()` on the `componentDidMount()` for that block.

kienstra:
**Request For Review**

Hi @swissspidy,
Could you please review this? Here's a [test story](https://story-test-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=2039&action=edit) with the issue.

includes/templates/story-templates/fandom-title.html
```diff
@@ -1,6 +1,6 @@
...
-    <!-- wp:amp/amp-story-text {"type":"p","className":"is-style-rounded","customTextColor":"#ffffff","customBackgroundColor":"#e44240","autoFontSize":11,"positionTop":52.26,"positionLeft":6.1,"height":30,"width":111} -->
+    <!-- wp:amp/amp-story-text {"type":"p","className":"is-style-rounded","customTextColor":"#ffffff","customBackgroundColor":"#e44240","autoFontSize":10,"positionTop":52.26,"positionLeft":6.1,"height":30,"width":111} -->
     <amp-story-grid-layer data-block-name="amp/amp-story-text" template="vertical"><div class="amp-story-block-wrapper" style="position:absolute;top:52.26%;left:6.1%;width:33.84%;height:5.42%"><p style="background-color:rgba(228, 66, 64, 1);color:#ffffff;display:flex" class="wp-block-amp-amp-story-text is-style-rounded has-text-color has-background is-style-rounded" id="1bc21884-841c-4218-9e4e-a1568d3e47ae"><amp-fit-text layout="flex-item" class="amp-text-content">SPOILERS ALERT</amp-fit-text></p></div></amp-story-grid-layer>
```
kienstra: This is a very minor change to prevent the 'SPOILER ALERTS' text from being on 2 lines instead of 1.

kienstra:
Thanks a lot for reviewing this. Feel free to commit directly to this, I'll probably be off for several hours.

miina:
@kienstra I'm still seeing incorrect text sizes when inserting the Template.

Updating the font size in `componendDidMount` was actually added specifically to fix that issue -- to have correct font size when inserting templates.

Perhaps there's another way to fix this, maybe there is something not being considered for the font size calculation. Will try to debug.

miina:
Looks like when the initial calculation is happening, the width of the content is full width by default, meaning that the font size changing is not really changing the width dynamically during the calculation. However, next times, the width changes according to the font size assigned.

Perhaps that's related to the actual cause of the font difference 🤷‍♀

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -30,10 +30,6 @@ class TextBlockEdit extends Component {
 		this.onReplace = this.onReplace.bind( this );
 	}
 
-	componentDidMount() {
-		maybeUpdateFontSize( this.props );
```
miina: We should probably try find the cause for the font size difference instead of removing the initial calculation. As mentioned, this seems to cause issues with templates.

swissspidy: Reverted it for now.

PR #2633, opened 2019-06-14
Only add text block template for the default page in a story

### LINKED ISSUE #2541: Insert blank pages in a fully 'blank' state
Currently when a new page is created, it is inserted into the story as a blank page with a text box in it. 

Usually the creation flow is to add an asset to the page first and then add the text. By having a text box inserted from the start, it forces the user to either type some text, then add a background image/color/video, and then go back to the text to edit it and configure it. It would be better to insert blank pages totally blank, and let the user decide on the order she wants to insert the components of the page. 

/cc @pbakaus

### LINKED ISSUE #2632: Block error when using "/" shortcut
This might be an issue in Gutenberg 5.9.

1. Create a new page
2. Type `/title` or any other block name in the already added text block
3. See the broken page.

The error happens within `insideSelectWithDataAlign` in Gutenberg. That HOC filters the `BlockListBlock`. For some reason, it gets a weird set of props:

Screenshot 2019-06-14 at 16 25 46

The `clientId` in that screenshot seems to be the one from the existing text block.

Interestingly, this crash only seems to happen with the pre-existing text block on the page. Once you delete that text block and add a new one, `/` works fine.

swissspidy:
When creating a new story, the first page has an empty text block in it.

When adding new pages, no text block is added to the page.

Fixes #2541.
Fixes #2632. Or at least almost.

The #2632 bug still happens for the default text block on the first page when creating a new story. This is an upstream issue in Gutenberg with templates.

I think this PR is a good compromise in the meantime until this issue gets fixed in Gutenberg.

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -460,7 +460,7 @@ class PageEdit extends Component {
 					{ backgroundColors.length > 0 && (
 						<div style={ overlayStyle } />
 					) }
-					<InnerBlocks template={ TEMPLATE } allowedBlocks={ allowedBlocks } />
+					<InnerBlocks allowedBlocks={ allowedBlocks } />
 				</div>
 			</>
 		);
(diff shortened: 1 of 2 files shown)


kienstra:
**Approved**

Hi @swissspidy,
This is a nice improvement.

The default page looks good. As expected, it doesn't have a Text block anymore.

Also, the `/` shortcut works when adding a Text block to a new page.

block-shortcut

PR #2634, opened 2019-06-14
Make sure block anchor IDs don’t start with a number

### LINKED ISSUE #2631: Block IDs sometimes start with a number
I just ran into an issue where a block's ID on the frontend was starting with a number. This is not allowed in HTML and specifically in AMP Stories breaks things like animations.

We use [`uuid()`](https://www.npmjs.com/package/uuid) to automatically generate these IDs.

Possible fix:

Create new `getUniqueId()` helper function. This function calls `uuid()` in a loop until it does not start with a number.

Then, we'd use that helper function everywhere we currently import `uuid/v4` directly (3 instances at the moment).

swissspidy:
Fixes #2631.

### DIFF
--- a/assets/src/stories-editor/helpers/index.js
+++ b/assets/src/stories-editor/helpers/index.js
@@ -439,7 +439,7 @@ export const addAMPExtraProps = ( props, blockType, attributes ) => {
 	const newProps = { ...props };
 
 	// Always add anchor ID regardless of block support. Needed for animations.
-	newProps.id = attributes.anchor || uuid();
+	newProps.id = attributes.anchor || getUniqueId();
 
 	if ( attributes.rotationAngle ) {
 		let style = ! newProps.style ? {} : newProps.style;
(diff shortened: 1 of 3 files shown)


assets/src/stories-editor/helpers/index.js
```diff
@@ -1452,3 +1452,20 @@ export const isVideoSizeExcessive = ( media ) => {
...
+ * Returns a unique ID that is guaranteed to not start with a number.
+ *
+ * Useful for using in HTML attributes.
+ *
+ * @return {string} Unique ID.
+ */
+export const getUniqueId = () => {
+	let uniqueId = uuid();
+
+	while ( /^\d/.test( uniqueId ) ) {
```
westonruter: I suggest a performance improvement here. Instead of a `while` to generate `uuid` until it starts with a letter, do this instead:

```js
return uuid().replace( /^\d/, 'a' );
```

swissspidy: Done in d63850e9347767269b8d44f062a0c06884392bae

PR #2635, opened 2019-06-14
Re-calculate font sizes of text block in more cases

swissspidy:
This way it also runs when the content changes, ampFitText is toggled, or the font family is changed.

**UNTESTED:** I am not sure if these changes even make sense.


swissspidy:
Looks like this is obsolete.

PR #2641, opened 2019-06-15
Add RSS feed link for Stories

westonruter:
At the moment in order for a search engine to discover the AMP Stories published on a site, a user must either:

0. Link to a story with regular hyperlink
1. Link to a story via post embed
2. Use the Latest Stories block
3. Install an XML Sitemap plugin (until [part of core](https://make.wordpress.org/core/2019/06/12/xml-sitemaps-feature-project-proposal/))

If a user hasn't done these things, in order to ensure stories are discoverable by search engines, the AMP plugin should add an RSS feed specific for stories.

This PR adds an RSS link for Stories like the following:

```html
<link rel="alternate" type="application/rss+xml" title="Stories" href="https://example.com/feed/?post_type=amp_story">
```

----

This PR also eliminates `AMP_Story_Post_Type::maybe_flush_rewrite_rules()` by adding story-awareness to `AMP_Options_Manager::maybe_flush_rewrite_rules()`, which eliminates the inefficiency of checking rewrite rules for every request. Now it will just flush rewrite rules upon toggling the Stories experience.

### DIFF
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -324,8 +324,6 @@ class AMP_Story_Post_Type {
 		);
 
 		add_action( 'wp_head', array( __CLASS__, 'print_feed_link' ) );
-
-		self::maybe_flush_rewrite_rules();
 	}
 
 	/**
@@ -1361,29 +1359,6 @@ class AMP_Story_Post_Type {
 		}
 
 		return ob_get_clean();
-	}
-
-	/**
-	 * Flushes rewrite rules if it hasn't been done yet after having AMP Stories Post type.
-	 */
-	public static function maybe_flush_rewrite_rules() {
-		$current_rules = get_option( 'rewrite_rules' );
-
-		// If we're not using permalinks.
-		if ( empty( $current_rules ) ) {
-			return;
-		}
-
-		// Check if the rewrite rule for showing preview exists for different permalink settings.
-		$story_rules = array_filter(
-			array_keys( $current_rules ),
-			function( $rule ) {
-				return 0 === strpos( $rule, self::REWRITE_SLUG ) || false !== strpos( $rule, '/' . self::REWRITE_SLUG . '/' );
-			}
-		);
-		if ( empty( $story_rules ) ) {
-			flush_rewrite_rules( false );
-		}
 	}
 
 	/**
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -1387,15 +1387,11 @@ class AMP_Story_Post_Type {
 	}
 
 	/**
-	 * Add RSS feed link for stories, if theme supports automatic-feed-links.
+	 * Add RSS feed link for stories.
 	 *
 	 * @since 1.2
 	 */
 	public static function print_feed_link() {
-		if ( ! current_theme_supports( 'automatic-feed-links' ) ) {
-			return;
-		}
-
 		$post_type_object = get_post_type_object( self::POST_TYPE_SLUG );
 		$feed_url         = add_query_arg(
 			'post_type',
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -323,6 +323,8 @@ class AMP_Story_Post_Type {
 			}
 		);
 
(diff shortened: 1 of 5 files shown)


includes/class-amp-story-post-type.php
```diff
@@ -1384,6 +1386,30 @@ function( $rule ) {
 		}
 	}
 
+	/**
+	 * Add RSS feed link for stories, if theme supports automatic-feed-links.
+	 *
+	 * @since 1.2
+	 */
+	public static function print_feed_link() {
+		if ( ! current_theme_supports( 'automatic-feed-links' ) ) {
```
westonruter: This condition is being used in core's `feed_links()` function, but I'm not sure it makes sense here.

Thinking about it more, I'm thinking that the theme support was added to prevent double-printing the feed links for themes that used to manually add it to the `head`, similar to the story for the `title-tag` theme support.

So this condition should probably be removed. Agreed?

includes/class-amp-story-post-type.php
```diff
@@ -1384,6 +1386,30 @@ function( $rule ) {
...
+		$post_type_object = get_post_type_object( self::POST_TYPE_SLUG );
+		$feed_url         = add_query_arg(
+			'post_type',
+			self::POST_TYPE_SLUG,
+			get_feed_link()
+		);
+		printf(
+			'<link rel="alternate" type="%s" title="%s" href="%s">',
+			esc_attr( feed_content_type() ),
+			esc_attr( $post_type_object->labels->name ),
```
westonruter: This seems like a good default feed title.

includes/class-amp-story-post-type.php
```diff
@@ -1362,26 +1362,23 @@ public static function render_block_latest_stories( $attributes ) {
 	}
 
 	/**
-	 * Flushes rewrite rules if it hasn't been done yet after having AMP Stories Post type.
+	 * Add RSS feed link for stories.
+	 *
+	 * @since 1.2
 	 */
-	public static function maybe_flush_rewrite_rules() {
```
westonruter: I realized this was unnecessary as it should have been merged into the existing `\AMP_Options_Manager::maybe_flush_rewrite_rules()`. This also ensures that the story rewrite rules actually get removed upon deactivation of the Stories experience.

includes/class-amp-story-post-type.php
```diff
@@ -1362,26 +1362,23 @@ public static function render_block_latest_stories( $attributes ) {
...
-			return;
-		}
-
-		// Check if the rewrite rule for showing preview exists for different permalink settings.
-		$story_rules = array_filter(
-			array_keys( $current_rules ),
-			function( $rule ) {
-				return 0 === strpos( $rule, self::REWRITE_SLUG ) || false !== strpos( $rule, '/' . self::REWRITE_SLUG . '/' );
-			}
+	public static function print_feed_link() {
```
swissspidy: Should this respect `current_theme_supports( 'automatic-feed-links' )`?

westonruter: @swissspidy Originally I did have this, but I reasoned myself out of it in https://github.com/ampproject/amp-wp/pull/2641#discussion_r294061433

Do you agree with my reasoning there?

westonruter: Another reason for why this theme support check doesn't make sense here is that it is not used by https://developer.wordpress.org/reference/functions/feed_links_extra/

swissspidy: Ah, I see. Yes, makes sense to not have it then.

westonruter:
# Testing Instructions

1. Enable the Stories experience
2. On the homepage and other pages of the site, you should see `<link rel="alternate" type="application/rss+xml" title="Stories" href="https://story-test-wordpress-amp.pantheonsite.io/feed/?post_type=amp_story">` in the `head`.
3. The link should take you to a valid RSS feed listing the latest published stories: https://story-test-wordpress-amp.pantheonsite.io/feed/?post_type=amp_story
4. Deactivate the Stories experience, and the link should go away.

csossi:
<link rel="alternate" type="application/rss+xml" title="Stories" href="https://story-test-wordpress-amp.pantheonsite.io/feed/?post_type=amp_story"> 

Verified in QA

PR #2646, opened 2019-06-17
Add caption toggle for video block.

### LINKED ISSUE #2393: Consider making video caption optional
Video caption always displays in the editor, consider making it optional via Toggle as it is for Image block.

miina:
Fixes #2393.

Adds toggle for video block's caption for making it optional, as it is for the image block.

### DIFF
--- a/assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
+++ b/assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
@@ -238,8 +238,6 @@ export default createHigherOrderComponent(
 				width,
 				opacity,
 				type: textBlockTextType,
-				ampShowImageCaption,
-				ampShowCaption,
 				ampAnimationType,
 				ampAnimationDuration,
 				ampAnimationDelay,
@@ -268,6 +266,8 @@ export default createHigherOrderComponent(
 			} else {
 				minHeight = MIN_BLOCK_HEIGHTS[ name ] || MIN_BLOCK_HEIGHTS.default;
 			}
+
+			const captionAttribute = isVideoBlock ? 'ampShowCaption' : 'ampShowImageCaption';
 
 			return (
 				<>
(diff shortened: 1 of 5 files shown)


assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
```diff
@@ -239,6 +239,7 @@ export default createHigherOrderComponent(
 				opacity,
 				type: textBlockTextType,
 				ampShowImageCaption,
+				ampShowCaption,
```
miina: I was originally thinking of using just `ampShowCapion` for both of these instead of separating since the functionality is really the same for both attributes, however, since we released already then the previously saved `ampShowImageCaption` would get lost when changing the attribute's name.

@swissspidy Any thoughts if there's a better way to do it without losing the previous attribute while not duplicating the logic for just a different attribute name?

swissspidy: Hmm normally this would be a use case for [deprecated blocks](https://developer.wordpress.org/block-editor/developers/block-api/block-deprecation/), but since this extends existing blocks, I don't have an optimal solution off the top of my head.

If it makes things easier you could probably do something like this:

```js
const captionAttribute = isVideoBlock ? 'ampShowCaption' : 'ampShowImageCaption;
// ...
checked={ attributes[ captionAttribute ] }
// ...
props.setAttributes( { [ captionAttribute ]: ! attributes[ captionAttribute ] } );
```

miina: Changed, thanks!

assets/src/stories-editor/components/with-wrapper-props.js
```diff
@@ -48,10 +48,13 @@ const withWrapperProps = ( BlockListBlock ) => {
...
 		}
 
+		const noCaption = ( 'core/image' === blockName && ! attributes.ampShowImageCaption ) ||
+			( 'core/video' === blockName && ! attributes.ampShowCaption );
+
 		// If we have image caption or font-family set, add these to wrapper properties.
 		wrapperProps = {
 			...props.wrapperProps,
-			'data-amp-image-caption': ( 'core/image' === blockName && ! attributes.ampShowImageCaption ) ? 'noCaption' : undefined,
+			'data-amp-caption': noCaption ? 'noCaption' : undefined,
```
swissspidy: Does this affect block validation?

miina: Nope, this is in the editor only. In the FE an empty caption isn't displayed anyway.

Issue #2523, opened 2019-06-06
Consider making Block Navigation toggleable


swissspidy:
Sometimes the block navigation can get a bit into the way, and it might be irrelevant when you only have 1 or 2 blocks on a page anyway.

We should consider making it possible to collapse the block navigation, or perhaps even hide it completely (like you can close the sidebar).

That would also be a good thing especially on smaller screens.

swissspidy:
The default block navigation in the post editor is in the top toolbar and looks like this:

![Screenshot 2019-06-12 at 16 29 57](https://user-images.githubusercontent.com/841956/59359845-62147d00-8d2f-11e9-9666-7641e7ae6943.png)

So my suggestion would be to have such a button in the top toolbar as well, but use it to just hide/show our custom block navigation. Still at the same place.

amedina:
That is a good idea. Would it be possible to `undock` the box that pops up? I am thinking about the scenario where the creator clicks on that toolbar button, and the block navigation opens around the place where it is located now; and then the user can click on a `close` section of it and it would disappear.  This would have the advantage of keeping the navigation closer to the pages when open.

swissspidy:
That was my suggestion. Our custom component stays in the same location, but we would add a toggle in the toolbar for it.

amedina:
Got it; that would be a good  UX enhancement.

PR #2656, opened 2019-06-18
Adjust styling for the block navigation.

### LINKED ISSUE #2578: Block navigation goes too narrow when resizing the window small
Steps to reproduce:
1. Resize the window to quite narrow.
2. Observe the results:
Screenshot 2019-06-12 at 16 06 12

Expected:
- The block navigation shouldn't look broken. Not clear what's the best approach here though.

miina:
Fixes #2578.

Adjusts styling for block navigation to avoid it from looking "broken".

### DIFF
--- a/assets/src/stories-editor/components/block-navigation/edit.css
+++ b/assets/src/stories-editor/components/block-navigation/edit.css
@@ -18,7 +18,7 @@
 	}
 }
 
-@media (max-width: 1100px) and (min-width: 1001px) {
+@media (min-width: 1001px) and (max-width: 1100px) {
 
 	.is-sidebar-opened .editor-block-list__layout {
 		margin-right: -100px;
--- a/assets/src/stories-editor/components/block-navigation/edit.css
+++ b/assets/src/stories-editor/components/block-navigation/edit.css
@@ -12,20 +12,30 @@
 }
 
 @media (max-width: 1200px) {
+
 	.is-sidebar-opened #amp-story-block-navigation {
 		left: 0;
 	}
 }
 
-@media (max-width: 1100px) and (min-width: 961px) {
+@media (max-width: 1100px) and (min-width: 1001px) {
+
 	.is-sidebar-opened .editor-block-list__layout {
 		margin-right: -100px;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


miina:
This PR basically does the 3 following changes:
- Hides the block navigation when the screen width is less than 1000 since it doesn't fit then.
- Moves the block navigation more to the left to give it more room. This will make the title and block navigation not be on the same line for some specific width but perhaps it's better than showing it on top of a Page?
- Shifts the Page more to the right in some specific case to give more room to the block navigation.
- Does all this for when the sidebar is opened only.
- Sets minimum width to block navigation.

This will improve the style and won't show the navigation broken anymore but will cause some shifts for some specific widths. An alternative could be just hiding the block already before (e.g. width 1100) and leaving everything else the same, reverting the changes from this PR.

Thoughts, @swissspidy?

swissspidy:
Looks reasonable at first glance.

I think this fits very well together with #2523. On small sites, instead of just hiding the block navigation, we could display it like the original block navigation with a popover:

Screenshot 2019-06-18 at 17 13 08

On bigger screens, we wouldn't use a popover, but just keep displaying it at the current position.

Of course that can be done in a separate PR specifically for #2523.

miina:
Let's implement #2523 within a separate PR, I could pick that up after the bug fixes.

swissspidy:
Sounds like a plan.

assets/src/stories-editor/components/block-navigation/edit.css
```diff
@@ -4,19 +4,41 @@
...
 }
 
+@media (max-width: 1200px) {
+
+	.is-sidebar-opened #amp-story-block-navigation {
+		left: 0;
+	}
+}
+
+@media (max-width: 1100px) and (min-width: 1001px) {
```
kienstra: This isn't wrong, but do you think having the `min-width` before the `max-width` would make it easier to read?

miina: Sure, I put it this way thinking of desktop-first, but it's not following this generally. Will change.

PR #2657, opened 2019-06-19
Updating PHP Codesniffer

koconder:
Dependency update, should pass Travis CI


westonruter:
Changes like this should normally come automatically via Renovate.

koconder:
> Changes like this should normally come automatically via Renovate.

I'm happy to close if thats the case

westonruter:
It seems you were updating dependencies to old versions, anyway.

The [current version](https://github.com/sabberworm/PHP-CSS-Parser/releases) of PHP-CSS-Parser is 8.3.0, and that's what is in the repo (as opposed to your  update to 8.2.1):

https://github.com/ampproject/amp-wp/blob/05d9d178f6ee33018f3290deb431a3222cc85614/composer.json#L17

Same goes for phpcodesniffer-composer-installer.

PR #2664, opened 2019-06-19
Show CTA block in Block Navigation

### LINKED ISSUE #2564: Block Navigation label is incorrect when there is only a CTA block
**To reproduce:**

1. Create a few pages
2. Add a CTA block on a page
3. Notice that the block navigation says "No elements added to this page yet."

**Screenshot:**

Screenshot 2019-06-11 at 17 10 41

**Reason**

This happens because we don't show the CTA block in the block navigation as it cannot be moved anyway.

**Possible Solution(s)**

Since #2443 we enforce the CTA block to be last in the list. So perhaps now we could display it in the block navigation again.

However, we would need to make sure that the block is not draggable there and that it doesn't have a drop zone, i.e. no block can be moved past it.

Alternatively, we could try to show a different text in that case. Or perhaps this can be considered as `won't fix` under the stance of the CTA block not being a regular element like the other blocks — a call to action.

miina:
Fixes #2564.

- Displays CTA block in the block navigation.
- Excludes CTA block from DropZone.

Note that there is space between the CTA block and the rest of the blocks in the block navigation. This comes from the default margin of the list, however, left it like this since it could actually make sense, to separate the non-draggable CTA from others and to clearly position it as the last block. Thoughts welcome.

UI:
Screenshot 2019-06-19 at 17 01 19

### DIFF
--- a/assets/src/stories-editor/components/block-navigation/index.js
+++ b/assets/src/stories-editor/components/block-navigation/index.js
@@ -93,10 +93,15 @@ export default compose(
 		const { getCurrentPage, isReordering } = select( 'amp/story' );
 		const { getBlockOrder, getBlocksByClientId, getSelectedBlockClientId } = select( 'core/block-editor' );
 
-		const blocks = getCurrentPage() ? getBlocksByClientId( getBlockOrder( getCurrentPage() ) ) : [];
-
+		let blocks = getCurrentPage() ? getBlocksByClientId( getBlockOrder( getCurrentPage() ) ) : [];
+		const callToActionBlock = blocks.find( ( { name } ) => name === 'amp/amp-story-cta' );
+		blocks = blocks.filter( ( { name } ) => ALLOWED_MOVABLE_BLOCKS.includes( name ) ).reverse();
+		// If CTA block was found, let's add it to the end.
+		if ( callToActionBlock ) {
+			blocks.push( callToActionBlock );
+		}
 		return {
-			blocks: blocks.filter( ( { name } ) => ALLOWED_MOVABLE_BLOCKS.includes( name ) ).reverse(),
+			blocks,
 			selectedBlockClientId: getSelectedBlockClientId(),
 			isReordering: isReordering(),
 		};
--- a/assets/src/stories-editor/components/block-navigation/index.js
+++ b/assets/src/stories-editor/components/block-navigation/index.js
@@ -18,30 +18,44 @@ import BlockNavigationItem from './item';
 import { ALLOWED_MOVABLE_BLOCKS } from '../../constants';
 import './edit.css';
 
-function BlockNavigationList( { blocks,	selectedBlockClientId, selectBlock } ) {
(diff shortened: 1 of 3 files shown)


kienstra:
**Approved**

Hi @miina,
Nice work! The CTA now appears in the block navigation:
cta-block-in-nav

And as expected, the CTA block can't be dragged in the block navigation:
cta-block-cant-drag

assets/src/stories-editor/components/block-navigation/index.js
```diff
@@ -70,7 +70,15 @@ function BlockNavigation( { blocks, selectBlock, selectedBlockClientId } ) {
 					selectBlock={ selectBlock }
 				/>
 			) }
-			{ ! hasBlocks && (
+			{ /* Add CTA block as a separate item to exclude it from DropZone. */ }
+			{ callToActionBlock && (
+				<BlockNavigationItem
```
swissspidy: So while the other blocks are in a nice `ul` list, this item here is just a simple `div`? Seems a bit odd from a semantics point of view.

assets/src/stories-editor/components/block-navigation/index.js
```diff
@@ -93,10 +104,13 @@ export default compose(
 		const { getCurrentPage, isReordering } = select( 'amp/story' );
 		const { getBlockOrder, getBlocksByClientId, getSelectedBlockClientId } = select( 'core/block-editor' );
 
-		const blocks = getCurrentPage() ? getBlocksByClientId( getBlockOrder( getCurrentPage() ) ) : [];
-
+		let blocks = getCurrentPage() ? getBlocksByClientId( getBlockOrder( getCurrentPage() ) ) : [];
+		// Let's get the CTA block to handle it separately.
```
swissspidy: I wonder if this could be handled in `BlockNavigationList` instead

assets/src/stories-editor/components/block-navigation/index.js
miina: Maybe, however, the items in the list are draggable and within dropzone, so it doesn't feel like it should be in the same list either -- it behaves very differently. 🤷‍♀Do you think it would be better to create a separate list for that?

miina: Or are you thinking that it should still be in the same list, just behave differently?

swissspidy: > Or are you thinking that it should still be in the same list, just behave differently?

I was thinking this (with the necessary logic being in `BlockNavigationList`), however just having a separate list might be more feasible.

miina: Alright, will move it to `BlockNavigationList` and add a separate list.

miina: Changes made within 0b14eea.

Note that there is some room left between the lists to separate the draggable items from the static CTA link, seems to make sense when testing, however, let me know if you have objections.

miina:
Ah, right, didn't consider that when moving the logic to `BlockNavigationList`. Should be fixed within 1c48463.

PR #2680, opened 2019-06-24
Fix hidden editor on zooming.

### LINKED ISSUE #2662: Editor halfway hidden when using browser zoom
As reported by Alberto, the editor area is somehow halfway hidden on his site.

Video: https://cloudup.com/c6ZtNTi-6zL

The site uses a standard set of plugins like Jetpack, Atomic Blocks, Site Kit, Gutenberg, and the Hemingway.

We had a similar issue once, see #2119. This one seems to be different.

Edit:  Looks like this is related to zooming controls in the browser (Cmd+Plus, Cmd+Minus)

miina:
Fixes #2662.

Applies the CSS for `__resizable_base__` for all the blocks, not just the image block since it looks like other blocks are causing the same issue.

### DIFF
--- a/assets/css/amp-stories-editor.css
+++ b/assets/css/amp-stories-editor.css
@@ -89,7 +89,7 @@
  * @see https://github.com/WordPress/gutenberg/issues/7180
  */
 
-.wp-block-image .components-resizable-box__container + .__resizable_base__ { /* stylelint-disable-line selector-class-pattern */
+.wp-block .components-resizable-box__container + .__resizable_base__ { /* stylelint-disable-line selector-class-pattern */
 	left: 0 !important;
 }
 


kienstra:
**Approved**

Hi @miina,
This looks good. It fixes #2662 in my local.

# Before

page-block

# After

after-here

miina:
@swissspidy To confirm -- was this Travis failure something to ignore, what was the issue with this?
Screenshot 2019-06-25 at 23 46 54

miina:
Thanks for testing, @kienstra!

swissspidy:
@miina A Travis error that can be ignored. AFAIK happens when there‘s a commit to develop while the build starts.

PR #2688, opened 2019-06-26
Fix 4.9 JS polyfills

### LINKED ISSUE #2679: i18n polyfill is incomplete
I got issues with Yoast SEO on WordPress 4.9 on loading the first panel.

After a bit of JS debugging I discovered that the problem it is the polyfill of this plugin that is loaded by yoast (because of the script name `wp-i18n`) that is the same of WP 5.x and used by that plugin.

Replacing the content of the js file of this plugin with the official from WP 5.2.1 is fixing the issue.

Ref: https://wordpress.org/support/topic/first-tab-traffic-lights-is-empty/

swissspidy:
Fixes #2679.

ZIP file for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3330034/amp.zip)

### DIFF
--- a/assets/src/polyfills/wp-dom-ready.js
+++ b/assets/src/polyfills/wp-dom-ready.js
@@ -3,8 +3,6 @@
  */
 import domReady from '@wordpress/dom-ready';
 
-if ( ! window.wp ) {
-	window.wp = {};
-}
+window.wp = window.wp || {};
 
-wp.domReady = domReady;
+window.wp.domReady = domReady;
--- a/assets/src/polyfills/wp-i18n.js
+++ b/assets/src/polyfills/wp-i18n.js
@@ -3,8 +3,6 @@
  */
 import * as i18n from '@wordpress/i18n';
 
-if ( ! window.wp ) {
-	window.wp = {};
-}
+window.wp = window.wp || {};
 
-wp.i18n = i18n;
+window.wp.i18n = i18n;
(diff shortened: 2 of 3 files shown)


Mte90:
The problem still persist...

The file compiled wp-i18n.js is more tiny of the counterpart of amp 1.1.3

swissspidy:
Oh, I think I found the issue then. Be right back with a fix...

swissspidy:
@Mte90 Please test this new build: [amp.zip](https://github.com/ampproject/amp-wp/files/3330112/amp.zip)

Mte90:
Now works :-D

swissspidy:
Excellent, thank you! We'll include this fix in the next bugfix release, version 1.2.1.

planetahuevo:
It does not work for me. 
Which are the files that have changed?

swissspidy:
@planetahuevo Did you use the file from https://github.com/ampproject/amp-wp/pull/2688#issuecomment-505867570? The JS files should have changed.

Generally speaking, you should consider updating to WordPress 5.2 and install the Classic Editor plugin if you need it.

planetahuevo:
Yes and no. I am planning to evaluate this by the end of the year but not before. If the only issues are AMP and Yoast I will use other SEO plugin and won´t use AMP at all.
What I am trying to say is that the decision won´t be made based on 2 plugins but more the overall WordPress differences. 
So far, what WP 5.2 offer is not worth the change, but that can change in the future.
I would love to have a LTS for WordPress that I can trust with updates and not having this issues every single month. Dreaming... :)

planetahuevo:
I will try the plugin again tomorrow or next week and let you know.

PR #2689, opened 2019-06-26
Fix lagging Template Inserter

### LINKED ISSUE #2577: TemplateInserter extremely slow / lagging
When clicking on the button to open the template inserter in the stories editor, it takes more than a second until something happens.

When the template inserter finally opens, you cannot immediately scroll through the template list as some assets (background images, etc.) are still being loaded.

At this point, you'll see incomplete template previews because some images are still missing and fonts are being enqueued.

After that, finally, the template inserter is fully functional!

-- 

A possible solution would be to how a spinner while the templates are being fetched in order to not delay any UI interactions.

For something like #2012 we'd need such a spinner anyway.

miina:
Fixes #2577.

Creates `TemplatePreview` component and loads it with a small latency to not cause a lag for loading all the previews at the same time.

### DIFF
--- a/assets/src/stories-editor/components/template-inserter/index.js
+++ b/assets/src/stories-editor/components/template-inserter/index.js
@@ -118,7 +118,7 @@ class TemplateInserter extends Component {
 											className="amp-stories__blank-page-inserter editor-block-preview__content block-editor-block-preview__content editor-styles-wrapper"
 										/>
 									</div>
-									{ this.state.storyTemplates.map( ( item ) => (
+									{ this.state.storyTemplates.map( ( item, index ) => (
 										<a // eslint-disable-line jsx-a11y/anchor-is-valid, see https://github.com/ampproject/amp-wp/issues/2165
 											key={ `template-preview-${ item.id }` }
 											role="button"
@@ -135,6 +135,7 @@ class TemplateInserter extends Component {
 										>
 											<TemplatePreview
 												item={ item }
+												isFirst={ 0 === index }
 											/>
 										</a>
 									) ) }
--- a/assets/src/stories-editor/components/template-inserter/index.js
+++ b/assets/src/stories-editor/components/template-inserter/index.js
@@ -8,7 +8,7 @@ import PropTypes from 'prop-types';
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Dropdown, IconButton, Spinner } from '@wordpress/components';
+import { Dropdown, IconButton } from '@wordpress/components';
 import { Component } from '@wordpress/element';
 import { withSelect, withDispatch } from '@wordpress/data';
 import { createBlock } from '@wordpress/blocks';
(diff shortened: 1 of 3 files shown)


assets/src/stories-editor/components/template-inserter/index.js
```diff
@@ -24,6 +24,44 @@ import addTemplateIcon from '../../../../images/add-template.svg';
...
+
+		this.state = {
+			parentLoaded: false,
+		};
+	}
+
+	componentDidMount() {
+		if ( ! this.state.parentLoaded ) {
+			// Set timeout to cause a small latency between loading the templates, otherwise they all try to load instantly and cause a lag.
+			setTimeout( () => {
```
miina: @swissspidy Let me know if you have a better solution for this issue.

Basically, it looks like `react-window` and its lazy-loading is also not usable since it's assuming that each list item which needs to get lazy-loaded is being fetched. However, in our case there is nothing to fetch, it's rendering itself which takes time and we can't really add a promise for rendering and component.

Added this option which loads the templates with a small latency so it displays the inserter with spinners at first and then loads the templates with small time differences.

It loads at first just with a spinner and once loaded the first time, changes the state to load the actual preview.

Tried a bunch of different options (including injecting the previews after loading the frame) but this seemed to work most reliably.

Let me know if you have a better idea.

swissspidy: At first glance I fear that using `setTimeout` is in fact the best solution here. I've seen similar patterns in the past too, which also indicates that this is the way to go.

Just a few minor notes / questions:

1. This should use the `withSafeTimeout` HOC so that timeouts are properly cleared when components are unmounted.
2. Maybe rename `parentLoaded` to `shouldLoad`
3. Is the timeout of `100` enough? Or could it even be `0` / omitted in order to just run during the next tick in the event loop?
4. Haven't tested yet, but how busy does the UI feel with all these spinners appearing for a short time? Should it perhaps just `return null` instead?

miina: 1. 👍 
2. Yep, that's more accurate actually.
3. It seemed to be enough when testing, omitting or setting to `0` didn't work, however. We can test with some alternative values but `100` seemed to work well.
4. Well, the visible window is quite small, so not all the spinners are really visible, locally testing the spinners displayed just for a moment and then the first one was replaced already. One thing that we could do would be to also pass the index of the template to the `TemplatePreview` and display the spinner only if it's the first Template, this way we would show only one spinner but would still indicate that something is happening. Thoughts?

swissspidy:
Finally got around to test this. It feels like a very good improvement. Opening and closing the inserter a few times still felt sluggish, but after clearing the cache it seems smoother.

Still not smooth though when throttling network requests to use 3G. Not sure what can be done in that case 🤷‍♂ 

What I would definitely change is the spinner position. Ideally it should be centered within a `.block-editor-block-preview` div.

As for only showing one spinner, I am not yet sure what's best. But if they're all positioned in the container's center, you would indeed not see multiple spinners at once anyway.

miina:
> Still not smooth though when throttling network requests to use 3G. Not sure what can be done in that case 🤷‍♂

Yep, I tried with 3G as well, the editor itself took ages to load already as well, at least locally, even on "Fast 3G". Thought that perhaps it's OK since we're not focusing on the mobile editing at this point. Not sure.

> What I would definitely change is the spinner position. Ideally it should be centered within a .block-editor-block-preview div.

Hmm, that's a good point, I'll center that.

miina:
The spinner should be centered now.

swissspidy:
Spinner position looks better now 👍 

With the centered spinners it makes more sense to see spinners everywhere instead of whitespace when scrolling. Just changed that.

swissspidy:
Gonna merge regardless of the unrelated intermittent build failures due to HTTP timeouts.

PR #2691, opened 2019-06-26
Fix issues for Author role

### LINKED ISSUE #2685: Breaking editing experience for user role 'Author'
Users with role of 'Author' seem unable to preview or publish stories. See snapshots for observed results. 
Screen Shot 2019-06-26 at 11 59 44 AM
Screen Shot 2019-06-26 at 11 59 52 AM

miina:
Fixes #2685 
Adds missing allowed CSS rules and missing attributes.

Todo:
- [x]  Figure out how to keep `background-color` attribute with `rgba` values (and other values that contain parenthesis).

- [x] Fix setting background Media.


miina:
@westonruter Looks like some CSS rules were not allowed, these have been added now, however, there is still one issue. Looks like any CSS rule which includes parenthesis is getting removed within `safecss_filter_attr`:
https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/kses.php#L2045-L2234

More specifically this part here:
https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/kses.php#L2224

This means that background-color with `rgba` value is being removed and this is used quite a lot within the editor.

Not sure what would be a good approach within the plugin, perhaps it would make sense to allow `rgba` and `rgb` values in core instead. I could create a patch for that if it seems to make sense, however, would still be good to fix it ASAP for the plugin, too. Thoughts?

miina:
Looks like there are some related trac issues:
- https://core.trac.wordpress.org/ticket/24157
- https://core.trac.wordpress.org/ticket/37134

miina:
Hmm, not sure if there is another option except for filtering the user caps to add `unfiltered_html` to `author` for AMP Stories.

includes/class-amp-story-post-type.php
```diff
@@ -206,6 +206,8 @@ public static function register() {
 
 		add_filter( 'wp_kses_allowed_html', array( __CLASS__, 'filter_kses_allowed_html' ), 10, 2 );
 
+		add_filter( 'safe_style_css', array( __CLASS__, 'filter_safe_style_css' ), 10, 1 );
```
westonruter: Can this be limited to the story post type?

miina: We could check the `rest_route` from `$_REQUEST` to detect if it's for the story post type.

miina: Added the check within c1755bf.

includes/class-amp-story-post-type.php
```diff
@@ -350,6 +350,18 @@ public static function remove_classic_editor_link( $actions, $post ) {
 	 * @return string[] Filtered array of allowed CSS attributes.
 	 */
 	public static function filter_safe_style_css( $attr ) {
+
+		// Only continue if it's a REST request for amp_story post type.
+		if ( ! isset( $_REQUEST['rest_route'] ) || empty( $_REQUEST['rest_route'] ) ) {
```
swissspidy: Untested, but this might cause issues with WP-CLI because `$_REQUEST` does not exist in a CLI context.

`rest_api_loaded()` in core uses `$GLOBALS['wp']->query_vars['rest_route']` to get the route. Maybe that can be used too?

miina: Good point, will check now.

miina: Changed within 528a32c.

miina:
@swissspidy Any thoughts on what we should do related to this issue: https://core.trac.wordpress.org/ticket/37134

I'm thinking of adding `unfiltered_html` caps for `author` for AMP Stories post type until this gets improved in core, any alternatives?

includes/class-amp-story-post-type.php
```diff
@@ -340,6 +344,83 @@ public static function remove_classic_editor_link( $actions, $post ) {
...
+		if ( ! isset( $args[0] ) || 'unfiltered_html' !== $args[0] ) {
+			return $allcaps;
+		}
+
+		if ( ! current_user_can( 'edit_posts' ) ) {
+			return $allcaps;
+		}
+
+		// Let's get the post's ID from the request and see if it's an AMP Story.
+		if ( ! isset( $_REQUEST['rest_route'] ) ) { // phpcs:ignore WordPress.CSRF.NonceVerification.NoNonceVerification
```
miina: @swissspidy Here the `GLOBALS` is not always available when needed. Not sure what's a good approach here.

miina:
Added `unfiltered_html` caps within 0f5827b. I guess this also influences the Multisite users.

swissspidy:
KSES is 😫

> Figure out how to keep background-color attribute with rgba values (and other values that contain parenthesis).

The biggest one that comes to mind is `transform()`, which we use for rotation. So if you currently rotate a block as an author, it's gonna break.

includes/class-amp-story-post-type.php
swissspidy: `$_REQUEST['rest_route']` doesn't work when using pretty permalinks. So I guess we need to parse the request URI ourselves, because this runs before WP does its init...

swissspidy:
Given that KSES is simply applied using a filter on `init`, using the `user_has_cap` filter here felt too low-level.

The new approach filters the REST response before the request is processed to remove the KSES filters from post content.

Still, this removes KSES as a whole, which is quite a dangerous approach since we only really care about the style attribute. So I wouldn't consider this as the final solution just yet.

swissspidy:
I tried to come up with a solution that doesn't require to turn off KSES, see 68941db7e145ed115ee91fe235481c12d681c552.

@westonruter What do you think of the current approach?

If it makes sense, we can flesh it out and add some tests.

includes/class-amp-story-post-type.php
```diff
@@ -341,7 +345,109 @@ public static function remove_classic_editor_link( $actions, $post ) {
...
+		add_filter(
+			'pre_post_content',
+			static function ( $post_content ) use ( &$style_attr_values ) {
+				$post_content = preg_replace_callback(
+					'|style=\\\"([^"]*)\\\"|', // Because the post data is slashed.
+					static function ( $matches ) use ( &$style_attr_values ) {
+						$hash                       = md5( $matches[1] );
+						$style_attr_values[ $hash ] = $matches[1];
+
+						return str_replace( $matches[1], $hash, $matches[0] );
```
westonruter: I'm not sure what I'm looking at here. 

Does this not result in `style=\"...\"` being replaced with `style=\"ed20d984b757ad5291963389fc209864\"`? Does this work because `safecss_filter_attr()` skips considering an unrecognized CSS property when it lacks a value? 

In other words, input such as:

```html
<span style="position:absolute"></span>
```

Is sanitized as:

```html
<span></span>
```

Whereas:

```html
<span style="position"></span>
```

Is left intact.

That seems somewhat prone to breakage if `safecss_filter_attr()` gets made more robust, no?

includes/class-amp-story-post-type.php
```diff
@@ -341,7 +345,109 @@ public static function remove_classic_editor_link( $actions, $post ) {
...
+			}
+		);
+
+		// ...And bring it back afterwards.
+		add_filter(
+			'content_save_pre',
+			static function ( $post_content ) use ( &$style_attr_values ) {
+				return str_replace( array_keys( $style_attr_values ), array_values( $style_attr_values ), $post_content );
+			},
+			20
```
westonruter: Since Kses runs on  `content_save_pre` at priority `10`, what if instead of using the `pre_post_content` filter, this could just use two `content_save_pre` filters: the first on running at priority `0` and the second one at priority `20` to restore?

includes/class-amp-story-post-type.php
```diff
@@ -341,7 +345,109 @@ public static function remove_classic_editor_link( $actions, $post ) {
...
+
+		// Replace inline styles before KSES...
+		add_filter(
+			'pre_post_content',
+			static function ( $post_content ) use ( &$style_attr_values ) {
+				$post_content = preg_replace_callback(
+					'|style=\\\"([^"]*)\\\"|', // Because the post data is slashed.
+					static function ( $matches ) use ( &$style_attr_values ) {
+						$hash                       = md5( $matches[1] );
+						$style_attr_values[ $hash ] = $matches[1];
```
westonruter: There should still be some validation of the properties here, no? The PHP-CSS-Parser could be used, though perhaps that would be overkill.

includes/class-amp-story-post-type.php
swissspidy: I considered that as well. Doesn‘t really make a difference I guess.

includes/class-amp-story-post-type.php
swissspidy: Good point. Something that we could catch with tests though I think.

includes/class-amp-story-post-type.php
swissspidy: Ideally core would have such a proper validation itself. Currently no validation takes place besides removing non-whitelisted properties and removing any properties whose value contains parentheses (and some other characters). The latter which is causing problems for us. Hence this hashing approach here.

includes/class-amp-story-post-type.php
```diff
@@ -341,7 +345,109 @@ public static function remove_classic_editor_link( $actions, $post ) {
...
+		}
+
+		$style_attr_values = [];
+
+		// Replace inline styles before KSES...
+		add_filter(
+			'pre_post_content',
+			static function ( $post_content ) use ( &$style_attr_values ) {
+				$post_content = preg_replace_callback(
+					'|style=\\\"([^"]*)\\\"|', // Because the post data is slashed.
```
westonruter: Just recalled that that this should be sure to include the `<` and `>` of the containing tag as otherwise this could match text content of elements, like if I write a story about `style` attributes.

So I suggest a pattern like:

```php
'#<\w+(?:-\w+)*\s[^>]*?style=\\\"([^"]*)\\\"[^>]+?>#'
```

In this way it will match the `style` attribute in the entire start tag.

includes/class-amp-story-post-type.php
swissspidy: I guess ideally we'd use `safecss_filter_attr()` before the hashing, but without the `Remove any CSS containing containing \ ( & } = or comments, except for url() useage checked above` part in order to not harm security.

includes/class-amp-story-post-type.php
swissspidy: Added back core-style KSES functionality now, so there's no regression.

swissspidy:
Let's do another round of review here to see if this is on the right track :-)

includes/class-amp-story-post-type.php
westonruter: Resolved by f781b0c.

includes/class-amp-story-post-type.php
```diff
@@ -341,7 +343,271 @@ public static function remove_classic_editor_link( $actions, $post ) {
...
+		}
+
+		$style_attr_values = [];
+
+		// Replace inline styles before KSES...
+		add_filter(
+			'content_save_pre',
+			static function ( $post_content ) use ( &$style_attr_values ) {
+				$post_content = preg_replace_callback(
+					'|<\w+(?:-\w+)*\s[^>]*?style=\\\"([^"]*)\\\"[^>]+?>|', // Because the post data is slashed.
```
westonruter: Suggested:

```diff
- // Because the post data is slashed.
+ // Extra slashes appear here because $post_content is pre-slashed.
```

includes/class-amp-story-post-type.php
```diff
@@ -341,7 +343,271 @@ public static function remove_classic_editor_link( $actions, $post ) {
...
+
+				return $post_content;
+			},
+			0
+		);
+
+		// ...And bring it back afterwards.
+		add_filter(
+			'content_save_pre',
+			static function ( $post_content ) use ( &$style_attr_values ) {
```
westonruter: Minor point, but `&$style_attr_values` doesn't strictly need to be passed by reference.

includes/class-amp-story-post-type.php
```diff
@@ -341,7 +343,271 @@ public static function remove_classic_editor_link( $actions, $post ) {
...
+				$post_content = preg_replace_callback(
+					'|<\w+(?:-\w+)*\s[^>]*?style=\\\"([^"]*)\\\"[^>]+?>|', // Because the post data is slashed.
+					static function ( $matches ) use ( &$style_attr_values ) {
+						$hash                       = md5( $matches[1] );
+						$style_attr_values[ $hash ] = self::safecss_filter_attr( $matches[1] );
+
+						// Replaces the complete style attribute value with its hashed version.
+						return str_replace( $matches[1], $hash, $matches[0] );
```
westonruter: To be a little more robust, I think the pattern should be tweaked a bit more as:

```php
'|(?P<before><\w+(?:-\w+)*\s[^>]*?)style=\\\"(?P<styles>[^"]*)\\\"(?P<after>[^>]+?>)|'
```

Then instead of doing a replacement in the string of `$matches` it can just concatenate:

```php
return $matches['before'] . sprintf( ' style="%s" ', $hash ) . $matches['after'];
```

What's more: to guard against a future `safecss_filter_attr()` which could sanitize away the md5 hash, what if the hash gets placed inside of a Kses-allowed data attribute like `data-temp-style-hash`. Then this could be:

```php
return $matches['before'] . sprintf( ' data-temp-style-hash="%s" ', $hash ) . $matches['after'];
```

And the replacement below could be:

```php
return preg_replace_callback(
    '/ data-temp-style-hash="(?P<hash>[0-9a-f]+)" /',
    function ( $matches ) use ( $style_attr_values ) {
        return isset( $style_attr_values[ $matches['hash'] ] ) ? sprintf( ' style="%s" ', esc_attr( wp_slash( $style_attr_values[ $matches['hash'] ] ) ) ) : '';
    },
    $post_content
);
return str_replace( array_keys( $style_attr_values ), array_values( $style_attr_values ), $post_content );
```

includes/class-amp-story-post-type.php
```diff
@@ -341,7 +343,271 @@ public static function remove_classic_editor_link( $actions, $post ) {
...
+
+		// Replace inline styles before KSES...
+		add_filter(
+			'content_save_pre',
+			static function ( $post_content ) use ( &$style_attr_values ) {
+				$post_content = preg_replace_callback(
+					'|<\w+(?:-\w+)*\s[^>]*?style=\\\"([^"]*)\\\"[^>]+?>|', // Because the post data is slashed.
+					static function ( $matches ) use ( &$style_attr_values ) {
+						$hash                       = md5( $matches[1] );
+						$style_attr_values[ $hash ] = self::safecss_filter_attr( $matches[1] );
```
westonruter: Should probably do `wp_unslash( $matches[1] )`

miina:
Made some changes based on @westonruter's suggestions, however, there are still some oddities happening with the `amp-story-page` style attribute, will have to debug this more (tomorrow). Marked it WIP for now.

includes/class-amp-story-post-type.php
miina: Wouldn't it always show the originally defined empty array if not passed by reference?

westonruter: Oh, that's a good point! Yes. Disregard my comment.

includes/class-amp-story-post-type.php
```diff
@@ -341,7 +343,280 @@ public static function remove_classic_editor_link( $actions, $post ) {
...
+	 * @link https://core.trac.wordpress.org/ticket/37134
+	 *
+	 * @param WP_HTTP_Response|WP_Error $response Result to send to the client. Usually a WP_REST_Response or WP_Error.
+	 * @param array                     $handler  Route handler used for the request.
+	 * @param WP_REST_Request           $request  Request used to generate the response.
+	 *
+	 * @return WP_HTTP_Response|WP_Error The filtered response.
+	 */
+	public static function filter_rest_request_for_kses( $response, $handler, $request ) {
+		if ( is_wp_error( $response ) ) {
```
miina: Is there a reason to filter this for admin users? Perhaps we should just short-circuit in case of admins. Thoughts?

includes/class-amp-story-post-type.php
```diff
@@ -351,6 +626,8 @@ public static function filter_kses_allowed_html( $allowed_tags ) {
 			'amp-story-page',
 			'amp-story-grid-layer',
 			'amp-story-cta-layer',
+			'amp-img',
+			'amp-video',
```
miina: These seem to be necessary for the background media.

includes/class-amp-story-post-type.php
```diff
@@ -363,11 +640,14 @@ public static function filter_kses_allowed_html( $allowed_tags ) {
...
-			$allowed_tag['animate-in-delay']    = true;
-			$allowed_tag['animate-in-after']    = true;
-			$allowed_tag['data-font-family']    = true;
+			$allowed_tag['animate-in']           = true;
+			$allowed_tag['animate-in-duration']  = true;
+			$allowed_tag['animate-in-delay']     = true;
+			$allowed_tag['animate-in-after']     = true;
+			$allowed_tag['data-font-family']     = true;
+			$allowed_tag['data-block-name']      = true;
+			$allowed_tag['data-temp-style-hash'] = true;
```
miina: Note: `data-temp-style-hash` is being removed for `amp-page` between the two added filters unless it's added as an allowed tag.

includes/class-amp-story-post-type.php
swissspidy: Yeah we could short-circuit if `current_user_can( 'unfiltered_html' )`

miina: Yep, this way we'd include Super Admins, too. Will add.

swissspidy:
@miina Just to clarify, is the `Fix setting background Media.` todo still open or already fully addressed by f27cc870388177707b9a9918149481b4760775ad?

miina:
> Just to clarify, is the Fix setting background Media. todo still open or already fully addressed by f27cc87?

Sorry for that, should be resolved, marked it "checked" now.

PR #2692, opened 2019-06-26
only show status on supported post types

paulschreiber:
Fixes the post metabox so AMP Status is not show on unsupported post types.

### DIFF
--- a/includes/admin/class-amp-post-meta-box.php
+++ b/includes/admin/class-amp-post-meta-box.php
@@ -263,10 +263,14 @@ class AMP_Post_Meta_Box {
 	 * @param WP_Post $post Post.
 	 */
 	public function render_status( $post ) {
+		$supported_post_types = AMP_Options_Manager::get_option( 'supported_post_types', array() );
+
 		$verify = (
 			isset( $post->ID )
 			&&
 			is_post_type_viewable( $post->post_type )
+			&&
+			in_array( $post->post_type, $supported_post_types, true )
 			&&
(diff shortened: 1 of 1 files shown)


westonruter:
Originally this was intentional in order to provide discovery for enabling the post types. But this was perhaps too “in your face”.

paulschreiber:
Without this patch, you see the AMP enabled/disabled radio button on CPTs that aren't AMP-enabled (or aren't logically posts, such as custom page layouts). Clicking the "edit" link won't make that content available; it will instead show you a warning:
image

westonruter:
Yes, that is what I meant was intentional.

swissspidy:
I can see how this is confusing though. Is this shown only to admins or also other authors?

westonruter:
Yeah, if we show it then it should be limited to admins.

swissspidy:
Just checked, even contributors currently see that message.

@paulschreiber Would showing the message just for admins (users with `manage_options` capability) work for your environment?

paulschreiber:
@swissspidy Showing the message only for admins is an improvement.

swissspidy:
In addition to only showing for admins, I think it also should just be an info notice, not a warning. Or ideally, really not show anything.

I opened #2724 to discuss this more broadly.

westonruter:
This will need to be re-factored to implement in the block editor. 

The PR also needs to target `develop`. So I'm going to close this PR in favor of opening a new issue to capture the full scope: #4593.

PR #2693, opened 2019-06-26
Avoid undefined variable warning for $path_args which is populated by wp_parse_str()

paulschreiber:


westonruter:
Can you provide details about how to reproduce the issue? What are the testing instructions?

paulschreiber:
The WPVIP scanner displayed a warning about `$path_args` being undefined when we submitted AMP 1.2.0.

By code inspection, I see this is correct as `$path_args` isn't defined before use in `wp_parse_str()`.

westonruter:
This is normal PHP, however, as the variable is passed by reference. For example, you will use a `$matches` variable when calling `preg_match()` without defining it beforehand.

paulschreiber:
I define `$matches` beforehand. PHP's loose typing is letting you get away with something here, but it would be good to be more intentional / clearer about the code.

swissspidy:
I think there are usually two types of positions in the PHP community about this. Personally I think it does not hurt at all to be explicit and to make the code more readable.

westonruter:
@paulschreiber Is this the only place that WPVIP scanner raises a warning?

swissspidy:
@westonruter Shouldn't this be merged into `develop` instead?

westonruter:
Oops. I didn't see that. Yes.

westonruter:
Bah. For simplicity I'll just cherry-pick the commit into `develop`.

westonruter:
Or rather, a new cherry-pick PR into `develop`.

PR #2695, opened 2019-06-27
Fix and consolidate some of the i18n strings

### LINKED ISSUE #2198: I18N String review
I did some translation work this week and am happy to say that the AMP plugin is now almost fully translated to German (formal + informal) and German Switzerland (formal + informal) 💪

Admittedly, it was kinda hard to translate the plugin as the majority of the strings is very long and technical. Often times there is also a lack of context.

Because of this, I am opening this issue as some sort of tracking ticket for improving the i18n strings over time.

swissspidy:
Fixes #2198.


assets/src/stories-editor/blocks/amp-story-post-date/index.js
```diff
@@ -17,9 +17,8 @@ export const settings = {
 	category: 'common',
 	icon: 'clock',
 	keywords: [
-		__( 'post', 'amp' ),
+		__( 'story', 'amp' ),
 		__( 'publish date', 'amp' ),
-		__( 'published date', 'amp' ),
 		__( 'date', 'amp' ),
```
miina: ^ This is also redundant since `publish date` already covers the `date` keyword.

assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
```diff
@@ -513,7 +513,7 @@ export default createHigherOrderComponent(
 									} }
 								/>
 								<RangeControl
-									label={ __( 'Background Opacity', 'amp' ) }
+									label={ __( 'Opacity', 'amp' ) }
```
miina: Are you removing this for translation purposes or generally as well? Wondering if it's clear without it that it's for the background only? I guess it should be.

includes/admin/class-amp-admin-pointers.php
```diff
@@ -96,7 +96,7 @@ private function get_pointers() {
 				array(
 					'selector'        => '#menu-posts-' . AMP_Story_Post_Type::POST_TYPE_SLUG,
 					'heading'         => __( 'AMP', 'amp' ),
-					'description'     => __( 'Head over here to create your first AMP story.', 'amp' ),
+					'description'     => __( 'Head over here to create your first story.', 'amp' ),
```
miina: Looks like you've removed `.` from some of the descriptions, should it be removed from here as well?

assets/src/stories-editor/components/inserter/index.js
```diff
@@ -112,7 +112,7 @@ class Inserter extends Component {
 				position={ position }
 				onToggle={ this.onToggle }
 				expandOnMobile
-				headerTitle={ __( 'Add a block', 'amp' ) }
+				headerTitle={ __( 'Add block', 'amp' ) }
```
miina: I'm not an English expert at all but wouldn't `Add block` be used if it's for a specific block and `Add a block` more accurate here?

includes/admin/class-amp-admin-pointers.php
miina: Or I guess this is a different kind of description?

assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
swissspidy: Mainly because it should be clear enough and because we already use `__( 'Opacity', 'amp' )` as a string in the editor. Makes things consistent and results in one string less to translate.

includes/admin/class-amp-admin-pointers.php
swissspidy: This is the text within an admin pointer, which is quite a different area from the stories editor.

assets/src/stories-editor/components/inserter/index.js
swissspidy: Actually, I think we should use `element` instead of `block` here anyway.

Mainly removed the `a` to consolidate strings.

assets/src/stories-editor/blocks/amp-story-post-date/index.js
swissspidy: Makes sense!

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -338,7 +338,7 @@ class PageEdit extends Component {
 								{ !! mediaId &&
 								<MediaUploadCheck>
 									<Button onClick={ () => setAttributes( { mediaUrl: undefined, mediaId: undefined, mediaType: undefined } ) } isLink isDestructive>
-										{ VIDEO_BACKGROUND_TYPE === mediaType ? __( 'Remove Video', 'amp' ) : __( 'Remove image', 'amp' ) }
+										{ _x( 'Remove', 'background media', 'amp' ) }
```
kienstra: Good idea to simplify this to 'Remove.' It's inside the 'Background Media' section, so it's clear enough what it's removing:

remove-here

assets/src/stories-editor/components/inserter/index.js
kienstra: Yeah, this would probably be fine with or without `a`.

PR #2711, opened 2019-06-30
Add support for portrait featured images in non-AMP Story posts

### LINKED ISSUE #2684: How to redefine featured image sizes?
In non-story custom posts, I sometimes use vertical featured images. This means that the crop for 1200x600 is incorrect and should use 9x16 instead of 16x9 as the math (1200x2134)

Is there any way I can override that for selected post types? Otherwise, can I disable the prompt to crop so my editors stop making images that don't work right from the NON-amp layouts?

kienstra:
* Only applies to non-AMP Story posts, like posts and pages
* If the featured image is a portrait, this suggests a crop of `1200 x 2133` if the featured image has at least those dimensions
* Before, it suggested a landscape crop, even for portrait images. For example, a `1600 x 2200` image would have a suggested crop of `1200 x 675`.

See #2684.

### DIFF
--- a/assets/src/common/components/featured-image-select-media-frame.js
+++ b/assets/src/common/components/featured-image-select-media-frame.js
@@ -72,7 +72,7 @@ const FeaturedImageToolbarSelect = wp.media.view.Toolbar.Select.extend( {
 		const minWidth = state.collection.get( 'library' ).get( 'suggestedWidth' );
 		const minHeight = state.collection.get( 'library' ).get( 'suggestedHeight' );
 
-		if ( ! attachment || ( attachment.get( 'width' ) >= minWidth && attachment.get( 'height' ) >= minHeight ) ) {
+		if ( ! attachment || ! attachment.get( 'width' ) || ( attachment.get( 'width' ) >= minWidth && attachment.get( 'height' ) >= minHeight ) ) {
 			this.secondary.unset( 'select-error' );
 		} else {
 			this.secondary.set(
--- a/assets/src/block-editor/index.js
+++ b/assets/src/block-editor/index.js
@@ -11,7 +11,7 @@ import { select } from '@wordpress/data';
  */
 import { withCroppedFeaturedImage, withFeaturedImageNotice } from '../common/components';
 import { addAMPAttributes, addAMPExtraProps, filterBlocksEdit, filterBlocksSave } from './helpers';
-import { getMinimumFeaturedImageDimensions } from '../common/helpers';
+import { getMinimumFeaturedImageDimensions, getMinimumPortraitFeaturedImageDimensions } from '../common/helpers';
 import './store';
 
 const {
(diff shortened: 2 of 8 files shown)


kienstra:
**Suggested Size**

Now that this also accepts a portrait size, should the message in the Media Library also include that?

This is from a post (not an AMP Story), on clicking 'Set featured image':
here-is-an-image

kienstra:
| Image Size        | Expected behavior on selecting as featured image | 
| ------------- |:-------------|
| [1200 x 2133](https://ipsumimage.appspot.com/1200x2133)  | Different in this PR, no crop suggested because it's already the ideal 9 x 16 portrait size |
| [1200 x 2200](https://ipsumimage.appspot.com/1200x2200)  | Different in this PR, suggests portrait crop of `1200 x 2133` |
| [1400 x 1000](https://ipsumimage.appspot.com/1400x1000)  | Same as before, suggests landscape crop |
| [1400 x 2000](https://ipsumimage.appspot.com/1400x2000)  | Same as before, suggests landscape crop. Would have been a portrait, but doesn't have the needed height of `2133`. |

kienstra:
**Request For Review**

Hi @swissspidy and @miina,
Could you please review this? 

As [shown above](https://github.com/ampproject/amp-wp/pull/2711#issuecomment-507071020), this allows cropping portrait featured images in non-AMP story posts.

Thanks, and hope you had a great weekend.

kienstra:
**Notice In Media Library**

Before, there was a notice for about a second on uploading a [1200 x 675 image](http://lorempixel.com/1200/675/technics/) as a featured image. But it only appeared in the Media Library:

image-notice

https://github.com/ampproject/amp-wp/pull/2711/commits/bc10ae5417257e04c9e901d17fc2f3fef6215542 should fix this.

assets/src/common/components/featured-image-select-media-frame.js
```diff
@@ -72,7 +72,7 @@ const FeaturedImageToolbarSelect = wp.media.view.Toolbar.Select.extend( {
 		const minWidth = state.collection.get( 'library' ).get( 'suggestedWidth' );
 		const minHeight = state.collection.get( 'library' ).get( 'suggestedHeight' );
 
-		if ( ! attachment || ( attachment.get( 'width' ) >= minWidth && attachment.get( 'height' ) >= minHeight ) ) {
+		if ( ! attachment || ! attachment.get( 'width' ) || ( attachment.get( 'width' ) >= minWidth && attachment.get( 'height' ) >= minHeight ) ) {
```
kienstra: This should fix the issue in https://github.com/ampproject/amp-wp/pull/2711#issuecomment-507504228

This could be a problem if `attachment.get( 'width' )` is `0`, but I'm not sure if or how often an image would have a width of `0`.

PR #2209, opened 2019-04-29
Improve handling of layout for images

westonruter:
* Prevent inline styles on images from overriding required display for layout. This prevents images with `style="display:inline"` from failing to render. See #1803.
* Eliminate fetching image dimensions for layouts that don't require them. If image has a `fixed-height` layout and it only has a `height` defined, there is no need to fetch the image for the `width`. Other layouts (like `fill`) also do not require dimensions.
* Allow special 'auto' value for width for `fixed-height` layout.

Given this `post_content`:

```html
An inline image: <img src="https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1-1024x668.jpg" width="40" height="40" style="display: inline"> Do you like it?

Here is a responsive image:

<img src="https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1-1024x668.jpg" height="668" width="1024" data-amp-layout="responsive" style="display: inline">

Image with fixed height which should look very stretched:

<img src="https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1-1024x668.jpg" height="30" data-amp-layout="fixed-height" style="display: inline">
```

# Before 👎 

> image

```html
<p>An inline image: <amp-img src="https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1-1024x668.jpg" width="40" height="40" class="amp-wp-enforced-sizes amp-wp-2d150e5" layout="intrinsic"><noscript><img src="https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1-1024x668.jpg" width="40" height="40" class="amp-wp-2d150e5"></noscript></amp-img> Do you like it?</p>
<p>Here is a responsive image:</p>
<p><amp-img src="https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1-1024x668.jpg" height="668" width="1024" layout="responsive" class="amp-wp-enforced-sizes amp-wp-2d150e5"><noscript><img src="https://wordpressdev.lndo.site/content/uploads/2019/02/1200px-American_bison_k5680-1-1024x668.jpg" height="668" width="1024" class="amp-wp-2d150e5"></noscript></amp-img></p>
<p>Image with fixed height which should look very stretched:</p>
[...]

Issue #2146, opened 2019-04-18
Percentage units and bogus values are erroneously permitted for width/height attributes


westonruter:
Reported on the [support forum](https://wordpress.org/support/topic/pages-not-converting-to-amp-after-last-update/), when someone adds the normal Facebook Comments embed:

```html
<div class="fb-comments" data-href="https://developers.facebook.com/docs/plugins/comments#configurator" data-width="100%" data-numposts="5"></div>
```

The AMP plugin currently outputs this as:

```html
<amp-facebook-comments layout="responsive" width="100%" height="400" data-href="https://developers.facebook.com/docs/plugins/comments#configurator" data-numposts="5"></amp-facebook-comments>
```

The AMP validator then flags this as a validation error:

> The attribute 'width' in tag 'amp-facebook-comments' is set to the invalid value '100%'.

It's surprising that this is leaking through the sanitizers, but when looking at the PHP-generated spec data it's not surprising:

https://github.com/ampproject/amp-wp/blob/621a0502f56c8d8e00b4ddbd61bc5ff013307a40/includes/sanitizers/class-amp-allowed-tags-generated.php#L15997-L16005

There are no constraints on the `width` and `height` attributes as defined in the [Validator protoascii](https://github.com/ampproject/amphtml/blob/568e45a548dd8b4aead0ca2088b238247670820c/validator/validator-main.protoascii#L5183-L5197):

```
# AMP Layout attributes: implied for TagSpecs with amp_layout field
# set. Note that while these attributes are defined as optional here,
# depending on amp layout the validator will examine these fields and
# their values.
attr_lists: {
  name: "$AMP_LAYOUT_ATTRS"
  attrs: { name: "height" }
  attrs: { name: "heights" }
  attrs: { name: "layout" }
  attrs: { name: "sizes" }
  attrs: { name: "width" }
  # <amp-bind>
  attrs: { name: "[height]" }
  attrs: { name: "[width]" }
}
```

It appears that this is done because the validation constraints require conditional logic that is not feasible for the protoascii format. Instead, the [`validator.js`](https://github.com/ampproject/amphtml/blob/568e45a548dd8b4aead0ca2088b238247670820c/validator/engine/validator.js) has imperative logic in its [`CssLength`](https://github.com/ampproject/amphtml/blob/568e45a548dd8b4aead0ca2088b238247670820c/validator/engine/validator.js#L3246-L3314). In particular, it has this logic:

```js
[...]

PR #2712, opened 2019-06-30
Improve handling of iframes: handle same-origin URLs and convert 100% width to fixed-height

westonruter:
This PR improves a couple aspects for how iframes are converted to AMP. Spurred on by a [support topic](https://wordpress.org/support/topic/shortcodes-for-geo-mashup-not-working/#post-11659045).

# Absolutize iframe URLs

AMP [does not allow relative URLs](https://github.com/ampproject/amphtml/blob/2708f15903cfa60f8a70a2dce330bfbf18302aee/validator/validator-main.protoascii#L2290) for `iframe` elements and it [may eventually not allow relative URLs](https://github.com/ampproject/amphtml/blob/2708f15903cfa60f8a70a2dce330bfbf18302aee/extensions/amp-iframe/validator-amp-iframe.protoascii#L69) for `amp-iframe`. So this PR adds the scheme and host to a URL that only includes a path.

Input:

```html
<iframe src="/core-dev/src/readme.html" width="100%" height="123"></iframe>
```

Before:

```html
<amp-iframe src="/core-dev/src/readme.html" width="525" height="123" sandbox="allow-scripts allow-same-origin" layout="intrinsic" class="amp-wp-enforced-sizes">
    <span placeholder="" class="amp-wp-iframe-placeholder"></span>
    <noscript>
        <iframe width="100%" height="123"></iframe>
    </noscript>
</amp-iframe>
```

After:

```html
<amp-iframe src="https://wordpressdev.lndo.site/core-dev/src/readme.html" width="auto" height="123" layout="fixed-height" sandbox="allow-scripts">
    <span placeholder="" class="amp-wp-iframe-placeholder"></span>
    <noscript>
        <iframe src="https://wordpressdev.lndo.site/core-dev/src/readme.html" width="100%" height="123"></iframe>
    </noscript>
</amp-iframe>
```

This also fixes an AMP validation error where the `src` of the fallback noscript `iframe` was getting stripped. Note that the `noscript > iframe` should be removed by the validating sanitizer once this is implemented:

https://github.com/ampproject/amp-wp/blob/61ed6bf9cb425b2fbff621a92f1a874a9a6579e5/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L23-L24

# Remove `allow-same-origin` sandbox from same-origin iframes

One very common issue with iframes in WordPress is issues due to using a same-origin source. This is not allowed in AMP when `allow-same-origin` is in the `sandbox` attribute. In this case, an error is shown:

image

See [same origin policy](https://github.com/ampproject/amphtml/blob/master/spec/amp-iframe-origin-policy.md).

Instead of the iframe having a hard failure due to violating the same-origin policy, an alternative is to simply remove the `allow-same-origin` from the iframe. While the iframe won't be able to interact with the parent window, at least the content can load instead of being blank. 

- [x] Still need to confirm what other side-effects will be experienced here.
- [x] Test with same-site post embeds.

For before/after, see above.

# Allow alias origin to be supplied

In a case where a site has a domain alias set up (which does not automatically redirect), the iframe sanitizer now supports converting same-origin URL to use the alias origin when encountering such an iframe. No alias is provided by default, but it can be done via a filter:

```php
add_filter(
	'amp_content_sanitizers',
	function ( $sanitizers ) {
		$sanitizers['AMP_Iframe_Sanitizer']['alias_origin'] = 'https://alias.' . wp_parse_url( get_home_url(), PHP_URL_HOST );
		return $sanitizers;
	}
);
```

Without this filter, see the example output above under absolutized iframe URLs. With the filter present, the output for a site that has an  origin of `https://wordpressdev.lndo.site`:

```html
<amp-iframe src="https://alias.wordpressdev.lndo.site/core-dev/src/readme.html" width="auto" height="123" layout="fixed-height" sandbox="allow-scripts allow-same-origin">
     <span placeholder="" class="amp-wp-iframe-placeholder"></span>
    <noscript>
        <iframe src="https://alias.wordpressdev.lndo.site/core-dev/src/readme.html" width="100%" height="123"></iframe>
    </noscript>
</amp-iframe>
```

# Convert 100% width to fixed-layout

AMP does not support the percentage unit in the `width` attribute. Instead, the `fixed-height` layout is supposed to be used when 100% width is desired.

Input:

```html
<iframe src="https://example.com/" width="100%" height="123"></iframe>
```

Before:

```html
<amp-iframe src="https://example.com/" width="525" height="123" sandbox="allow-scripts allow-same-origin" layout="intrinsic" class="amp-wp-enforced-sizes"></amp-iframe>
```

After:

```html
<amp-iframe src="https://example.com/" width="auto" height="123" layout="fixed-height" sandbox="allow-scripts allow-same-origin"></amp-iframe>
```

Support for the `auto` value for the `width` attribute was added in #2209.

----

Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3343396/amp.zip) (v1.2.1-alpha-20190630T221839Z-737f8ca5)

### DIFF
--- a/includes/sanitizers/class-amp-iframe-sanitizer.php
+++ b/includes/sanitizers/class-amp-iframe-sanitizer.php
@@ -34,11 +34,20 @@ class AMP_Iframe_Sanitizer extends AMP_Base_Sanitizer {
 	/**
 	 * Default args.
 	 *
-	 * @var array
+	 * @var array {
+	 *     Default args.
+	 *
+	 *     @type bool   $add_placeholder       Whether to add a placeholder element.
+	 *     @type bool   $add_noscript_fallback Whether to add a noscript fallback.
+	 *     @type string $current_origin        The current origin serving the page. Normally this will be the $_SERVER[HTTP_HOST].
+	 *     @type string $alias_origin          An alternative origin which can be supplied which is used when encountering same-origin iframes.
+	 * }
 	 */
 	protected $DEFAULT_ARGS = array(
 		'add_placeholder'       => false,
 		'add_noscript_fallback' => true,
+		'current_origin'        => null,
+		'alias_origin'          => null,
 	);
 
 	/**
@@ -70,7 +79,18 @@ class AMP_Iframe_Sanitizer extends AMP_Base_Sanitizer {
 			$this->initialize_noscript_allowed_attributes( self::$tag );
 		}
 
+		// Ensure origins are normalized.
+		$this->args['current_origin'] = $this->get_origin_from_url( $this->args['current_origin'] );
+		if ( ! empty( $this->args['alias_origin'] ) ) {
+			$this->args['alias_origin'] = $this->get_origin_from_url( $this->args['alias_origin'] );
+		}
+
 		for ( $i = $num_nodes - 1; $i >= 0; $i-- ) {
+			/**
+			 * Iframe element.
+			 *
+			 * @var DOMElement $node
+			 */
 			$node = $nodes->item( $i );
 
 			// Skip element if already inside of an AMP element as a noscript fallback.
@@ -86,7 +106,7 @@ class AMP_Iframe_Sanitizer extends AMP_Base_Sanitizer {
 			 * If the src doesn't exist, remove the node. Either it never
 			 * existed or was invalidated while filtering attributes above.
 			 *
-			 * @todo: add a filter to allow for a fallback element in this instance.
+			 * @todo: add an arg to allow for a fallback element in this instance (note that filter cannot be used inside a sanitizer).
 			 * @see: https://github.com/ampproject/amphtml/issues/2261
 			 */
 			if ( empty( $normalized_attributes['src'] ) ) {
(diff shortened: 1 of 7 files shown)


includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -139,10 +160,27 @@ public function sanitize() {
 	private function normalize_attributes( $attributes ) {
 		$out = array();
 
+		$remove_allow_same_origin = false;
```
westonruter: Actually, now I wonder why `allow-same-origin` is even the default `sandbox`. AMP never allows iframes to mutate the parent DOM, so why is this ever useful?

westonruter: One key reason is the document would not be able to access cookies:

> Uncaught DOMException: Failed to read the 'cookie' property from 'Document': The document is sandboxed and lacks the 'allow-same-origin' flag.

westonruter: Another side effect of this is the `origin` in `postMessage` will be `null`.

Still, it seems far preferable to render an iframes with these sandbox restrictions than to render a blank box with an amp-iframe error.

westonruter:
The 100% width conversion is part of this underlying issue: https://github.com/ampproject/amp-wp/issues/2146

PR #2737, opened 2019-07-03
Enforce the file type of the featured image, background media, and video block

### LINKED ISSUE #2727: Video upload does not work
Repro steps:

1. Upload a video, the video thumbnail does not appear in the media library, but if you upload a single video, it is selected in the pane on the right. 
1. Tapping "Select" takes you to the "Crop image" screen; however the image is broken, so you cannot hit "Select" and you are stuck on this screen.

image
image

kienstra:
* If the selected featured image, like a `.mov` file, is the wrong type, it displays a notice and disables the 'Select' button.
* Applies to the featured image in the normal post block editor and the AMP Story editor. Though maybe it should only apply to the AMP Story editor.
* Uses the `allowedTypes` property of the `MediaUpload` component. 

featured-image-select

Here's the notice:

new-notice

Fixes #2727


westonruter:
Does this also apply to the video block? Users should be disallowed from selecting or uploading anything but MP4 videos.

kienstra:
>Does this also apply to the video block? Users should be disallowed from selecting or uploading anything but MP4 videos.

Hi @westonruter, no, it doesn't apply to the video block or to the 'Background Media' control. I still need to work on that.

assets/src/common/helpers/index.js
```diff
@@ -199,3 +204,83 @@ export const getAspectRatioType = ( width, height ) => {
...
 	}
 };
+
+/**
+ * Gets the compiled template for a given notice message.
+ *
+ * @param {string} message The message to display in the template.
+ * @return {Function} compiledTemplate A function accepting the data, which creates a compiled template.
+ */
+export const getNoticeTemplate = ( message ) => {
```
kienstra: This function is simply extracted from the previous logic, and almost not changed at all.

assets/src/stories-editor/constants.js
```diff
@@ -106,7 +106,7 @@ export const ALLOWED_BLOCKS = [
 
 export const IMAGE_BACKGROUND_TYPE = 'image';
 export const VIDEO_BACKGROUND_TYPE = 'video';
-export const ALLOWED_MEDIA_TYPES = [ 'image', 'video/mp4' ];
+export const ALLOWED_MEDIA_TYPES = [ 'image', 'video' ];
```
kienstra: This constant is only used for the Background Media and its poster image.

`video/mp4` didn't seem to prevent other video mime types, like `video/quicktime`.

I think these are intended to be types, like `attachment.get( 'type' )` would return. The only time I saw `video/mp4` was with `attachment.get( 'mime' )`

kienstra: The new function `isFileTypeAllowed` ensures that videos have the right mime type.

kienstra:
**Mainly Working**
*I Still Need To Test More*

This is mainly working for the 'Background Media,' Core Video block, and the featured image. 

If any of them have the wrong mime type, like `'video/quicktime'` for a `.mov` file, it displays a notice and disables the 'Select' button:

uploading-video

I still need to test this more, and I'm working on a unit test for `enforceFileType()`.

assets/src/common/components/featured-image-select-media-frame.js
```diff
@@ -1,125 +0,0 @@
-/**
```
kienstra: This is mainly moved to [select-media-frame.js](https://github.com/ampproject/amp-wp/pull/2737/files#diff-507fa4419fd99ea17660e963ced76788).

assets/src/common/components/select-media-frame.js
```diff
@@ -0,0 +1,182 @@
+/**
```
kienstra: Much of this is simply copied and slightly refactored from [featured-image-select-media-frame.js](https://github.com/ampproject/amp-wp/pull/2737/files#diff-094dcdc5d37691ad94173f75bc4df3f0).

kienstra: The name is changed because it doesn't only apply to featured images now, it also applies enforces the file type.

assets/src/common/components/with-enforced-file-type.js
```diff
@@ -0,0 +1,90 @@
...
+	 */
+	return class EnforcedFileTypeMediaUpload extends InitialMediaUpload {
+		/**
+		 * Constructs the class.
+		 */
+		constructor() {
+			super( ...arguments );
+
+			// This class should only be present in the MediaUpload for the AMP Story 'Background Media' or when only 'video' types are allowed, like in the Core Video block.
+			if ( 'story-background-media' === this.props.id || isEqual( [ 'video' ], this.props.allowedTypes ) ) {
```
kienstra: `isEqual( [ 'video' ], this.props.allowedTypes` is the only way I could find to select the Core 'Video' block. Though this will of course also apply to any other `MediaUpload` where that happens to be true.

kienstra:
| Scenario   |  Expected behavior |
|---|---|
|  1. AMP Story or normal post: upload as featured image a .txt or .mov file | 'Select' button disabled, notice appears: "The selected file mime type ... is not allowed."  | 
| 2. AMP Story Background Media: upload a .txt or .mov file  | Same as above  | 
| 3. Core Video block in AMP Story and normal post editor: upload as featured image anything other than an `.mp4` video, like an `.mov` file  |  Same as above | 
| 4. AMP Story editor: select or upload as featured image an image with dimensions smaller than`1200 x 1600`  |  Shouldn't change in this PR: should display the notice "The selected image is too small..." | 
| 5. AMP Story editor: select or upload as featured image an image with dimensions larger than `1200 x 1600`, like [1900 x 1900](http://lorempixel.com/1900/1900/technics/)  |  Shouldn't change in this PR: this should still display the cropper | 

The first 3 scenarios are new in this PR, the last 2 are to test regressions.

kienstra:
**Video Block**

In the Core Video block, this PR only applies restricts the files types for the 'Media Library' button, not the 'Upload' button:

media-library

kienstra:
**Request For Review**

Hi @swissspidy and @westonruter,
Could you please review this when you can? 

In the [first 3 scenarios above](https://github.com/ampproject/amp-wp/pull/2737#issuecomment-508886732), this displays a notice in the Media Library and disables the 'Select' button:

notice-select-disabled

assets/src/common/components/with-enforced-file-type.js
westonruter: That seems fine because MP4 is the only format supported in stories.

westonruter:
Is applying the restriction to the Video block's Upload button too difficult to include here?

kienstra:
Hi @westonruter,

>Is applying the restriction to the Video block's Upload button too difficult to include here?

Good question. I'll look at that tomorrow if that's alright.

It looks like if [the accept prop](https://github.com/WordPress/gutenberg/blob/84543ff908fab4f88c45e42e656b553afa4fdd48/packages/block-library/src/video/edit.js#L191) were `video/mp4` instead of `video/*`, that might restrict uploading to `mp4` files.

Maybe there's a way to change that with the `editor.BlockEdit` filter.

kienstra:
**Video Upload Button**

Hi @westonruter,

>Is applying the restriction to the Video block's Upload button too difficult to include here?

Sure, https://github.com/ampproject/amp-wp/pull/2737/commits/a72b3e51943900ab4853beaf8b075649f3f09ddd restricts the upload type of the Video block to `video/mp4':

only-allows-uploading-mp4

Sorry for the delay.

kienstra:
Thanks, @westonruter!

kienstra:
**Request For Review**

Hi @swissspidy,
Could you please review this?

Here's the [expected behavior](https://github.com/ampproject/amp-wp/pull/2737#issuecomment-508886732). 

Thanks, and hope you have a great weekend.

PR #2738, opened 2019-07-03
[WIP] Update dev-lib / phpunit setup

swissspidy:


swissspidy:
Doesn't really work this way because that always installs PHPUnit 4 due to the platform constraints.

PR #2740, opened 2019-07-04
Latest Stories: Filter out stories without featured image

### LINKED ISSUE #2704: Latest Stories block Edit and ServerSideRender requests are inconsistent
Tested with the latest Gutenberg from the repo as of today.

When trying to use the Latest Stories the following happens:
Screenshot 2019-06-28 at 12 14 55
Screenshot 2019-06-28 at 12 14 48

To reproduce:
- Create AMP Stories without featured image (make sure there are no other Stories or the others are drafts)
- Open any post.
- Add Latest Stories Post.

swissspidy:
You'll now see the correct "No Stories found" message if you have stories without featured images, instead of an empty box.

Fixes #2704.

### DIFF
--- a/assets/src/block-editor/blocks/amp-latest-stories/edit.js
+++ b/assets/src/block-editor/blocks/amp-latest-stories/edit.js
@@ -50,47 +50,45 @@ class LatestStoriesEdit extends Component {
 		const { attributes, setAttributes, latestStories } = this.props;
 		const { order, orderBy, storiesToShow } = attributes;
 
-		const inspectorControls = (
-			<InspectorControls>
-				<PanelBody title={ __( 'Latest Stories Settings', 'amp' ) }>
-					<QueryControls
-						{ ...{ order, orderBy } }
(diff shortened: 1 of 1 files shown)


assets/src/block-editor/blocks/amp-latest-stories/edit.js
```diff
@@ -50,47 +50,45 @@ class LatestStoriesEdit extends Component {
 		const { attributes, setAttributes, latestStories } = this.props;
 		const { order, orderBy, storiesToShow } = attributes;
 
-		const inspectorControls = (
```
kienstra: Good idea to simply output this in the return statement instead of storing it in a `const`.

kienstra:
**Approved**

Hi @swissspidy,
Nice work. As you mentioned, this has the 'No stories found.' message if the only stories available don't have featured images.

# Before
before-latest-stories

# After
after-amp-stories

PR #2742, opened 2019-07-04
Determined tag name should not influence user font size

### LINKED ISSUE #2720: Normal font is larger than Medium font
When looking at the list of font sizes, it appears that Normal font size should be smaller than Medium font, but when selected Normal font size is larger than Medium.

swissspidy:
Fixes #2720.


kienstra:
**Approved**

Hi @swissspidy,
This looks good.

Testing this branch vs. the `develop` branch, I couldn't find a case where the removed logic here actually made a difference on the front-end. 

Maybe I missed something. Still, this PR makes the codebase cleaner, and the Text block worked the same in my testing.

PR #2744, opened 2019-07-04
Change block dimensions when text outgrows current one

### LINKED ISSUE #2719: Background color does not extend with the text when resized
When autosizing text is toggled off, if a user changes the font larger than the block size the font extends past the block, but the background color does not extend with the text.

Similar to: https://github.com/ampproject/amp-wp/issues/2210

swissspidy:
Fixes #2719.

### DIFF
--- a/assets/src/stories-editor/helpers/index.js
+++ b/assets/src/stories-editor/helpers/index.js
@@ -768,7 +768,7 @@ export const calculateFontSize = ( measurer, expectedHeight, expectedWidth, maxF
 	if ( ! measurer.offsetHeight || ! measurer.offsetWidth ) {
 		return false;
 	}
-	measurer.classList.toggle( 'is-measuring-fontsize' );
+	measurer.classList.toggle( 'is-measuring' );
 
 	maxFontSize++;
 
@@ -788,7 +788,7 @@ export const calculateFontSize = ( measurer, expectedHeight, expectedWidth, maxF
 	// Let's restore the correct font size, too.
 	measurer.style.fontSize = minFontSize + 'px';
 
-	measurer.classList.toggle( 'is-measuring-fontsize' );
+	measurer.classList.toggle( 'is-measuring' );
 
 	return minFontSize;
 };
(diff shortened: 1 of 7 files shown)


swissspidy:
Hmm we might even be able to write a nice E2E test for that.

assets/src/stories-editor/helpers/index.js
```diff
@@ -1215,6 +1215,40 @@ export const maybeUpdateFontSize = ( block ) => {
...
+ */
+export const maybeUpdateBlockDimensions = ( block ) => {
+	const { name, clientId, attributes } = block;
+	const { width, height, ampFitText, content } = attributes;
+
+	if ( ampFitText ) {
+		return;
+	}
+
+	switch ( name ) {
```
kienstra: It's just a personal preference, but I might have used an `if` block instead of a `switch`. Not that this is wrong 😄

assets/src/stories-editor/helpers/index.js
```diff
@@ -1215,6 +1215,40 @@ export const maybeUpdateFontSize = ( block ) => {
...
+ * Updates a block's width and height in case it doesn't use amp-fit-text and the font size has changed.
+ *
+ * @param {Object}  block                         Block object.
+ * @param {string}  block.clientId                Block client ID.
+ * @param {Object}  block.attributes              Block attributes.
+ * @param {number}  block.attributes.width        Block width in pixels.
+ * @param {number}  block.attributes.height       Block height in pixels.
+ * @param {string}  block.attributes.content      Block inner content.
+ * @param {boolean} block.attributes.ampFitText   Whether amp-fit-text should be used or not.
+ * @param {number}  block.attributes.autoFontSize Automatically determined font size for amp-fit-text blocks.
```
kienstra: Wow, great documentation of `block`

assets/src/stories-editor/helpers/index.js
```diff
@@ -1215,6 +1215,40 @@ export const maybeUpdateFontSize = ( block ) => {
...
+ */
+export const maybeUpdateBlockDimensions = ( block ) => {
```
kienstra: It's nice how this is in `helpers.js`.

assets/src/stories-editor/helpers/index.js
swissspidy: Well I copied over from the other helper function 🙃

assets/src/stories-editor/helpers/index.js
swissspidy: True, because it‘s only checking one block. I copied it over in the expectation that this might apply to other blocks in the future as well.
Actually, maybe it already does today. I‘ll do some testing.

assets/src/stories-editor/helpers/index.js
kienstra: Ah, haha

PR #2755, opened 2019-07-08
Add video block size validation in the Media Library

### LINKED ISSUE #2560: Display a notice if a video in the video block is over 1 MB per second

kienstra:
* In the Media Library, on selecting a video with the wrong size, this displays a notice
* Applies to the Media Library for the Video block and 'Background Media' 
* Doesn't disable the 'Select' button
* Branched off #2737, as that refactored the Media Library logic
* If/when that PR is merged, I'll probably rebase this PR to `develop` and target this PR to `develop`

in-video

Closes #2560


assets/src/common/helpers/test/getVideoBytesPerSecond.js
```diff
@@ -0,0 +1,37 @@
+/**
```
kienstra: This is mainly moved from `stories-editor/helpers/test/getVideoBytesPerSecond.js`, without many changes. Likewise for `isVideoSizeExcessive.js`.

kienstra:
**Steps To Test**

1. Create a new story
2. For the 'Background Media' of a page, upload or select from the 'Media Library' tab a video with more than 1 MB per second, [like this](https://drive.google.com/file/d/1gM3_6iarZcrjeXo2lOePVSPxlUEVKiiU/view?usp=sharing)
3. Excepted: there is a notice like:

>A video size of less than 1 MB per second is recommended. The selected video is 10 MB per second.

notice-appears

kienstra:
This isn't quite ready for review, as it's still targeting the `fix/upload-file-type` branch. If and when that branch is merged into `develop`, I'll probably rebase this to `develop` and target that branch.

westonruter:
@kienstra I merged `develop` and changes the base branch, but then I got confused, so I reset it back!

kienstra:
OK, all good!

swissspidy:
Just rebased the branch and fixed merge conflicts

kienstra:
>Just rebased the branch and fixed merge conflicts

Nice, thanks!

kienstra:
**Thanks for unblocking this**

Hi @swissspidy,
Thanks a lot for rebasing this branch and handling the merge conflicts. 

It's Saturday, so no hurry 😄 

But I just tested this, and it should be ready for review.

miina:
That's not related to the Media Library message directly, however, should there be a warning message outside of Media Library as well, as it is for the Background Media?
1. When inserting the video from URL then there is no warning displayed.
2. When first selecting a video -> closing the Media Library -> clicking the Edit Video icon -> Choosing "Upload" the Media Library won't open and thus the warning isn't displayed.

Thoughts?

swissspidy:
> When inserting the video from URL then there is no warning displayed.

I don't think we can check the video length _and_ file size without downloading the video.

miina:
> I don't think we can check the video length and file size without downloading the video.

I was actually testing with local URL-s where we'd probably have the image data but it's true that it's much more likely that the URL would be used for external images where getting the size and length needs downloading.

I guess the only place where the warning is missing then is when a previous video is changed via  "Edit Video" and "Upload".

westonruter:
> I don't think we can check the video length _and_ file size without downloading the video.

One option would be to do a `HEAD` request for the video URL. That would at least get the file size in `Content-Length`.

swissspidy:
A `HEAD` request is also what came to find. For the video length it might be actually quite easy:

When inserting a video using a URL, we can just access the `<video>` element's duration property, which gives the duration in seconds.

westonruter:
> When inserting a video using a URL, we can just access the `<video>` element's duration property, which gives the duration in seconds.

Great idea. I believe this just means waiting for the [`loadedmetadata`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/loadedmetadata_event) event.

kienstra:
I merged in `develop`, as there is a new custom Video block edit component from #2799 that was just merged.

swissspidy:
Added a warning now when pasting a video URL:

Screenshot 2019-07-16 at 11 26 18

miina:
I'm seeing `0` at the place of the notice if there is no error (see below "Replace Image")

Screenshot 2019-07-16 at 13 05 35

miina:
Likely not the most common use case but looks like displaying the warning based on URL stops showing when I do the following:
- Add a video from URL that exceeds the limits.
- Click "Edit Video" icon.
- Choose "Insert from URL".
- Don't change anything, leave the same URL, submit.

miina:
> I guess the only place where the warning is missing then is when a previous video is changed via "Edit Video" and "Upload".

Actually, there is no need to change a previous video, just using "Upload" directly and choosing a video exceeding the limits won't show the warning since the Media Library doesn't open then.

swissspidy:
> Likely not the most common use case but looks like displaying the warning based on URL stops showing when I do the following

Cannot reproduce this.

> I'm seeing 0 at the place of the notice if there is no error (see below "Replace Image")

Cannot reproduce either.

> Since 4 is used for every test, maybe for readability we could use a general descriptive variable for the seconds

Made the tests much more readable now in 3ce5e76b6ee8868c814a0a248e3ac0b7e679f8df.

swissspidy:
> Actually, there is no need to change a previous video, just using "Upload" directly and choosing a video exceeding the limits won't show the warning since the Media Library doesn't open then.

I guess we can then just show the warning in the sidebar.

swissspidy:
@miina Mind checking again?

miina:
Awesome, the issues mentioned before are gone now!

One more oddity that I just saw, the same video shows different values in the Media Library vs in the sidebar.
Sidebar:
Screenshot 2019-07-16 at 15 10 24

Media Library:
Screenshot 2019-07-16 at 15 10 13

Thoughts?

swissspidy:
> One more oddity that I just saw, the same video shows different values in the Media Library vs in the sidebar.

Interesting. Will try to reproduce.

miina:
> Interesting. Will try to reproduce.

Note that I was using [this video](https://drive.google.com/file/d/1gM3_6iarZcrjeXo2lOePVSPxlUEVKiiU/view?usp=sharing).

swissspidy:
Looks like it was a rounding issue.

PR #2766, opened 2019-07-08
Make main plugin file parseable on PHP 5.2+

swissspidy:
`__DIR__` is PHP 5.3+ only, and for the time being ideally the main plugin file should be parseable with PHP 5.2 as well.


amp.php
```diff
@@ -145,7 +149,7 @@
 
 unset( $_amp_required_extensions, $_amp_missing_extensions, $_amp_required_constructs, $_amp_missing_classes, $_amp_missing_functions, $_amp_required_extension, $_amp_construct_type, $_amp_construct, $_amp_constructs );
 
-if ( ! file_exists( __DIR__ . '/vendor/autoload.php' ) || ! file_exists( __DIR__ . '/vendor/sabberworm/php-css-parser' ) || ! file_exists( __DIR__ . '/assets/js/amp-block-editor.js' ) ) {
+if ( ! file_exists( AMP__DIR__ . '/vendor/autoload.php' ) || ! file_exists( AMP__DIR__ . '/vendor/sabberworm/php-css-parser' ) || ! file_exists( AMP__DIR__ . '/assets/js/amp-block-editor.js' ) ) {
```
swissspidy: If there are arguments against moving the constant definitions to the top, an alternative would be to use `dirname( __FILE__ )` here.

westonruter: The only reason I can think of to not move the constants above would be for any plugins that are using `defined('AMP__VERSION')` checks to see if the plugin is active.

swissspidy: Yup, that was my thinking as well.

Did a quick search through the plugin directory here: https://wpdirectory.net/search/01DF9H0EM4J37ZWVC6XV48QRWP

I mostly see version comparisons or combinations with other checks there.

PR #2493, opened 2019-06-03
Code Readability & Maintainability Improvements

swissspidy:
I began fixing some annotations related to code coverage. While doing so, I encountered a few inconsistencies and even bugs in the tests code. Many of these things probably also apply to `/includes`, so I might look at that later as well.

Most notable changes:

* Fix invalid covers annotations for functions (9b7135463e5d2d4b22ba4c0f39f5ca0fd8677375)
* Use static access for factories (45b1e8168b1991a84ee1db0d09c24bd8888062f5)  
  Since core.trac.wordpress.org/changeset/35242 the factory has been a static method of `WP_UnitTestCase`. Non-static access is only there for BC.
* Use get_echo helper function in tests (7065ace9d87a62fa2c1c4c290ee69ff3715b8052)  
  Makes things a bit more DRY and easier to read 
* Improves types of assertions being used (c3b9f86d91e0f40a043904a27cafe2ad7a434087 b2d0151220910bcc9edb7685d6a01460fe84c1b5)
* Declare some closures as static if possible (076406d93f41c974033cfcf8ec8b0c19bcd47738 3a9c5d838820fe1814d5866958f935bd0f801759 dd219fb5518124a959b8974442dafc2fb52a93b0 f8b8ac5d09fb71b9a2cf136b2082c75707c54ca6)
* Fixes an invalid assertion that shouldn't have been passing in the first place (5407916d5539bf8dd57243e1a2b74314e53088a0)
* Remove some unused code (bf3fa54dca25d7bebe7432ca7ea611bf754cf6a4 7bf39a11d70dd7780aeeea1151989808e4b57d18 2717d7ac04e4e4e002a367f4880bb46be0f1cea1)

As for code review, checking the individual commits is probably easiest.

PR #2767, opened 2019-07-08
Use short array syntax throughout the project

swissspidy:
This is a bit of a follow-up to #2493.

The main plugin file, `amp.php`, is left untouched as it still should be parseable on PHP 5.2.

Also does not change `includes/sanitizers/class-amp-allowed-tags-generated.php` as it is auto-generated.

### DIFF
--- a/.dev-lib
+++ b/.dev-lib
@@ -3,12 +3,9 @@
 DEFAULT_BASE_BRANCH=develop
 ASSETS_DIR=wp-assets
 PROJECT_SLUG=amp
-SKIP_ECHO_PATHS_SCOPE=1
 README_MD_TITLE="AMP Plugin for WordPress"
 DEV_LIB_SKIP="$DEV_LIB_SKIP,jshint"
-if [[ ! -z $TRAVIS ]]; then
-	CHECK_SCOPE=all
-fi
+CHECK_SCOPE=all
 
 function after_wp_install {
 	if [[ "$WP_VERSION" != "4.9" ]]; then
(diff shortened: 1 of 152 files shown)


westonruter:
As part of this, the PHPCS ruleset should be updated for force the preferred syntax for consistency.

swissspidy:
Of course, good call!

swissspidy:
The file is currently excluded from phpcs, and I left it out on purpose because it's auto-generated.

Fun Fact: PHPCS actually crashes locally for me because the file is so huge.

> I can update the generator to output the bracketed syntax as well.

Sounds good to me!

westonruter:
> Fun Fact: PHPCS actually crashes locally for me because the file is so huge.

As part the plugin v2.0 effort, I'd like to revisit that file entirely, look at ways it can be made smaller and/or broken up into smaller files.

swissspidy:
I was curious:

```
$ php -d memory_limit=4096M vendor/bin/phpcbf includes/sanitizers/class-amp-allowed-tags-generated.php

PHPCBF RESULT SUMMARY
-----------------------------------------------------------------------------
FILE                                                         FIXED  REMAINING
-----------------------------------------------------------------------------
includes/sanitizers/class-amp-allowed-tags-generated.php     15247  9
-----------------------------------------------------------------------------
A TOTAL OF 15247 ERRORS WERE FIXED IN 1 FILE
-----------------------------------------------------------------------------

Time: 1 mins, 60 secs; Memory: 12MB

```

😅😅😅

westonruter:
/me shakes fist at Travis…

```
git diff develop...6bfa9746c7fb6c5c09a56667ff5c1c919c984901
fatal: Path 'amp.php' exists on disk, but not in '6bfa9746c7fb6c5
```

PR #2798, opened 2019-07-11
Adjust CTA style considering potential image.

### LINKED ISSUE #2706: Consider removing inline image feature from CTA block (or improve styling)
Apparently, the CTA block allows adding an inline image, this feature might be unnecessary for AMP Stories, we might want to consider removing it. Currently, this looks more like a bug and not a feature. We should either remove it or fix the styling to make BE and FE match and align the text better.

To reproduce:
- Add CTA block
- Add an inline image from the Toolbar's "More RichText Controls" menu.
- Add some Text, too
- Check how it looks like in the FE.
- Compare the FE and BE, it's quite different.

Screenshot 2019-06-28 at 12 55 31
Screenshot 2019-06-28 at 12 55 19

miina:
Fixes #2706

### DIFF
--- a/assets/css/amp-stories-frontend.css
+++ b/assets/css/amp-stories-frontend.css
@@ -16,6 +16,10 @@ amp-story-grid-layer {
 	align-content: unset;
 }
 
+.amp-block-story-cta__link amp-img * {
+	object-fit: contain;
+}
+
 .wp-block-pullquote {
 	border-top: 4px solid #555d66;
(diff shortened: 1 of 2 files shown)


kienstra:
**Approved**

Hi @miina,
Good idea to fix this. Before, this was most obvious when the image was a portrait.

# Editor

editor-here

# Front-end, Before

fe-before

# Front-end, After

fe-after

Issue #2455, opened 2019-05-29
Error when pasting YouTube URL in Video block


swissspidy:
When creating a Video block in Gutenberg and pasting a YouTube URL in the "Insert from URL" field, Gutenberg automatically creates a YouTube block for it.

In the AMP Stories editor, this does not work, because YouTube videos are currently not allowed in Stories and thus we disabled the YouTube embed block. There's an error in the console:

Here's where this happens: 

https://github.com/WordPress/gutenberg/blob/ee7c19acf998d99455939dc2501a66179d82cd97/packages/block-library/src/video/edit.js#L95-L114

And here's the function in question:

https://github.com/WordPress/gutenberg/blob/ee7c19acf998d99455939dc2501a66179d82cd97/packages/block-library/src/embed/util.js#L61-L118

It's probably the same issue when pasting video URLs from some other providers whose blocks we have disabled.

It looks like `createUpgradedEmbedBlock` does not at all check whether the block type is registered, so I think this is something we should report upstream.

----

First reported in https://github.com/ampproject/amp-wp/issues/2394#issuecomment-496634928.

swissspidy:
Created an upstream issue: https://github.com/WordPress/gutenberg/issues/15883

swissspidy:
Upstream pull request: https://github.com/WordPress/gutenberg/pull/15885

swissspidy:
Merged the upstream fix. It will hopefully be part of Gutenberg 5.9.

_**Note:** Nothing to do for QA here until Gutenberg 5.9 is released!_

PR #2799, opened 2019-07-12
Enable playing the Core Video block's video in the editor

### LINKED ISSUE #2530: Enable video previews in editor
In AMP story context, videos must autoplay and loop to be valid. That's why we hide these controls already.

Now, it would be even better when the videos would autoplay in the editor already. Just like we already do for the page background video.

For that we probably have to extend the video block's edit component as the video tag is hardcoded (and disabled) there:

https://github.com/WordPress/gutenberg/blob/ff70de60086e8f505de51a9f9883d77bf5eda119/packages/block-library/src/video/edit.js#L280-L291

### LINKED ISSUE #2808: Video poster image warning shown in placeholder
**Steps to reproduce**

1. Add new video block
2. Select video without a poster image
3. Verify that the missing poster image warning appears in the block sidebar.
4. Click on pencil icon in the block toolbar to replace the video
5. See the poster image warning in the placeholder

**Screenshot**

Screenshot 2019-07-15 at 14 22 09

kienstra:
* Allows the video in the Core Video block to play in the editor
* This forks the Core Video block's edit component
* Ideally, this would just extend that component, and override the `render()` method.
* But that caused a console warning, that suggested that the class extend `Component` instead of the Video `BlockEdit`
* Also incorporates logic to automatically set the poster image upon video selection
* Also prevents the poster warning to be shown in edit mode by inlining the `withVideoPosterImageNotice` logic within the component itself.
* Prevents poster image to be removed as it is required. It can only be replaced.

Closes #2530.
Fixes #2808.


assets/css/amp-stories.css
```diff
@@ -89,14 +89,6 @@ table.wp-block-table {
...
 }
 
-/**
- * See https://github.com/ampproject/amp-wp/issues/2283
- */
-.edit-post-layout[data-block-name="core/video"] .edit-post-settings-sidebar__panel-block .block-editor-block-inspector__card + div > .components-panel__body:first-child .components-toggle-control:nth-child(2),
-.edit-post-layout[data-block-name="core/video"] .edit-post-settings-sidebar__panel-block .block-editor-block-inspector__card + div > .components-panel__body:first-child .components-toggle-control:nth-child(6) {
-	display: none;
-}
-
```
kienstra: This hid the 'Autoplay' and 'Play inline' toggles, but the edit component doesn't have them anymore.

assets/src/stories-editor/components/video-block-edit-with-preview.js
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	onSelectURL( newSrc ) {
+		const { attributes, setAttributes } = this.props;
+		const { src } = attributes;
+
+		// Set the block's src from the edit component's state, and switch off
+		// the editing UI.
+		if ( newSrc !== src ) {
+			// Omit the embed block logic, as that didn't seem to work.
```
kienstra: Before, entering an embed URL below 'Insert from URL' was supposed to create an Embed block:

insert-from-url

This is [intentionally disabled](https://github.com/ampproject/amp-wp/issues/2455) for YouTube, but even other embeds [like Vimeo](https://vimeo.com/111847786) don't seem to display.

**Is it OK that I removed the logic do convert to embeds?**

https://github.com/WordPress/gutenberg/blob/cb718ffdff57e35f05f4239ff03ceb476ae69a86/packages/block-library/src/video/edit.js#L111-L117

Entering a non-embed URL still works. But embed support will probably require copying a lot of other dependencies, like [createUpgradedEmbedBlock()](https://github.com/WordPress/gutenberg/blob/cb718ffdff57e35f05f4239ff03ceb476ae69a86/packages/block-library/src/embed/util.js#L61-L122), making this file much longer.

assets/src/stories-editor/components/video-block-edit-with-preview.js
```diff
@@ -0,0 +1,246 @@
...
+
+/**
+ * Mainly forked from the Core Video block Edit component, but allows the <video> to play instead of being disabled.
+ *
+ * There are very few changes from the Core Video block's components.
+ * The main change is that in render(), the <video> is not wrapped in <Disabled>, so it can play.
+ *
+ * @class
+ */
+class VideoBlockEditWithPreview extends Component {
```
kienstra: Much of this is simply copied from the Video block's [edit.js](https://github.com/WordPress/gutenberg/blob/cb718ffdff57e35f05f4239ff03ceb476ae69a86/packages/block-library/src/video/edit.js).

assets/src/stories-editor/components/video-block-edit-with-preview.js
```diff
@@ -0,0 +1,246 @@
...
+						loop
+						controls={ controls }
+						poster={ poster }
+						ref={ this.videoPlayer }
+						src={ src }
+					/>
+					{ ( ! RichText.isEmpty( caption ) || isSelected ) && (
+						<RichText
+							tagName="figcaption"
+							placeholder={ __( 'Write caption…' ) }
```
kienstra: The `__()` calls don't have the AMP textdomain. If it works like PHP, ideally they'll just use the Gutenberg translations.

assets/src/stories-editor/index.js
```diff
@@ -312,6 +313,7 @@ addFilter( 'blocks.registerBlockType', 'ampStoryEditorBlocks/filterBlockTransfor
 addFilter( 'editor.BlockEdit', 'ampStoryEditorBlocks/addStorySettings', withAmpStorySettings );
 addFilter( 'editor.BlockEdit', 'ampStoryEditorBlocks/addPageNumber', withPageNumber );
 addFilter( 'editor.BlockEdit', 'ampStoryEditorBlocks/addEditFeaturedImage', withEditFeaturedImage );
+addFilter( 'editor.BlockEdit', 'ampEditorBlocks/addVideoBlockPreview', withVideoBlockPreview, 9 );
```
kienstra: The priority of `9` allows the other filters to run, which add 'Block Position,' 'Animation,' and 'Story Settings' to the Inspector.

kienstra:
**Request For Review**
Hi @swissspidy and @miina,
Could you please review this?

It allows the video in the Core Video block to play in the editor, where it was wrapped in `<Disabled>` before:

video-autoplay

Though this might be overkill, as it forks most of the Core Vide block's edit component.

I could figure out a way around this. I tried creating a new class that [extends the Video edit component](https://github.com/ampproject/amp-wp/pull/2799/files#diff-4e3e66d8cc7f5207a7265f813182a2a3R20), like:

```js
class VideoEditWithPreview extends InitialBlockEdit {
        render() {
               /* implements render */
        }
}
```

Ideally, this would simply override the `render()` method, and inherit the rest of the parent class's methods.

This caused a console warning that the component should instead extend `Component`. 

Maybe there's another way around this that I haven't thought of.

assets/src/stories-editor/components/video-block-edit-with-preview.js
swissspidy: Can you amend the docblock to mention the removed player controls?

assets/src/stories-editor/components/index.js
```diff
@@ -24,6 +24,8 @@ export { default as withHasSelectedInnerBlock } from './higher-order/with-has-se
 export { default as withPageNumber } from './higher-order/with-page-number';
 export { default as withStoryFeaturedImageNotice } from './higher-order/with-story-featured-image-notice';
 export { default as withEditFeaturedImage } from './with-edit-featured-image';
+export { default as withVideoBlockPreview } from './with-video-block-preview';
+export { default as VideoBlockEditWithPreview } from './video-block-edit-with-preview';
```
swissspidy: Given that it does more than just enabling autoplay/previews for the video, I suggest renaming `VideoBlockEditWithPreview` to `CustomVideoBlockEdit` and `withVideoBlockPreview` to `withCustomVideoBlockEdit` to make it more clear what it does.

assets/src/stories-editor/components/video-block-edit-with-preview.js
swissspidy: > Is it OK that I removed the logic to convert to embeds?

I suppose so. AMP Stories doesn't really support any external providers AFAIK.

cc @westonruter

assets/src/stories-editor/components/video-block-edit-with-preview.js
swissspidy: Well, it depends. If a string is ever removed from Gutenberg without us knowing, it won't get translated at all.

It's usually best to include the text domain, even if it means translators have to do a bit more work.

assets/src/stories-editor/components/video-block-edit-with-preview.js
```diff
@@ -0,0 +1,296 @@
...
+							label={ __( 'Edit video' ) }
+							onClick={ switchToEditing }
+							icon="edit"
+						/>
+					</Toolbar>
+				</BlockControls>
+				<InspectorControls>
+					<PanelBody title={ __( 'Video Settings' ) }>
+						<ToggleControl
+							label={ __( 'Playback Controls' ) }
```
swissspidy: Uhm, it doesn't seem like controls are actually displayed in AMP Stories? The runtime seems to hide them no matter what.

@westonruter Can you confirm this? Should we just remove this option?

assets/src/stories-editor/components/video-block-edit-with-preview.js
kienstra: Ah, good point.

assets/src/stories-editor/components/video-block-edit-with-preview.js
kienstra: Sure!

kienstra: Ah, thanks for amending the DocBlock.

assets/src/stories-editor/components/video-block-edit-with-preview.js
miina: AMP Story should support Twitter embed in theory.

swissspidy: Pasting a tweet URL in a video block and converting it to a twitter embed would be a bit unexpected, no?

miina: True 😄 My comment was a bit out of context.

assets/src/stories-editor/components/video-block-edit-with-preview.js
miina: AFAIK this was already removed at some point.

Found also this: https://github.com/ampproject/amp-wp/pull/2255 where `controls` was set to `false` by default.

swissspidy: Hmm so we should remove the `controls` setting, but keep the `loop` (default to true) and `mute` settings?

Even though `mute` doesn't seem to have any effect. When opening an AMP Story, sound is muted by default for the whole story (see upper right corner), and when I enable sound, the video is not muted.

miina: Yep, the `mute` control doesn't seem to have any effect anyway, looks like "Loop" is the only control that actually does anything.

Seems like we could probably just add the `loop` and leave the rest out for now.

swissspidy: Added back the `loop` setting and removed the one for `controls`.

If `loop` is disabled, `controls` are still shown since one might want to re-start the video during editing.

swissspidy:
As pointed out in #2808, I think we can now easily move the logic from `withVideoPosterImageNotice` into our custom block edit component. This would allow fixing #2808 _and_ simplify the code there.

kienstra:
>As pointed out in #2808, I think we can now easily move the logic from withVideoPosterImageNotice into our custom block edit component. This would allow fixing #2808 and simplify the code there.

Ah, good point.

swissspidy:
Since I added some changes, I'd like another pair of eyes do the final review here 🙂

assets/src/stories-editor/components/video-block-edit-with-preview.js
westonruter: External video providers aren't supported yet, but support is planned soon: https://github.com/ampproject/amphtml/issues/15847

kienstra:
@swissspidy, would it help if I reviewed the latest changes?

assets/src/stories-editor/components/custom-video-block-edit.js
```diff
@@ -0,0 +1,317 @@
...
+							</BaseControl>
+						</MediaUploadCheck>
+					</PanelBody>
+				</InspectorControls>
+				<figure className="wp-block-video">
+					<video
+						autoPlay
+						muted
+						loop={ loop }
+						controls={ ! loop }
```
miina: Wondering if instead of always displaying the controls when the loop is off we could instead add a `ToggleControl` for `controls` which would be visible only if `loop` is turned off? Thoughts?

swissspidy: Wasn‘t the conclusion that controls are not visible on the front end? Adding a ToggleControl for an editor-only thing doesn‘t make much sense then I think.

swissspidy:
@kienstra Sure, but you cannot approve your own PR :-)

assets/src/stories-editor/components/custom-video-block-edit.js
miina: Yep, but if they're not visible on the front end then why not have `controls={ ! loop }` just `false` instead?

swissspidy: Explained my thinking here: https://github.com/ampproject/amp-wp/pull/2799#discussion_r303392747

miina: > If loop is disabled, controls are still shown since one might want to re-start the video during editing.

swissspidy: No, they‘re shown in the editor to quickly replay the video. That change to the edit component has no effect on the frontend.

miina: Right, that's the edit component 🤦‍♀🤦‍♀🤦‍♀Ignore, please.

kienstra:
**Request For Testing**

Hi @csossi,
Could you please test this?

1. Create a new AMP Story
2. Add a Video block
3. Select a video for it
4. Expected: the video [plays in the editor](https://github.com/ampproject/amp-wp/pull/2799#issuecomment-510728653) (it didn't before)
video-block-editor


5. Expected: there should be toggles for 'Display Caption' and 'Loop,' and they should work

PR #2800, opened 2019-07-13
Use custom readme file for story export archive

### LINKED ISSUE #2623: Write the Readme content specific to Export Feature
Write content for README.txt to include instructions, and a link to the documentation.

swissspidy:
Fixes #2623.


includes/story-export/readme.txt
```diff
@@ -0,0 +1,9 @@
+=== AMP Stories Export Archive ===
```
ThierryA: I would suggest to rephrase to something like that:

> _This folder contains the HTML as well as all the assets for a story exported by the WordPress AMP Stories editor (https://wordpress.org/plugins/amp/). The base URL is set according to the AMP plugin's story base URL setting._
> 
> _You can use this folder to host the generated story on another server. Note that using HTTPS is recommended._

@MackenzieHartung could you perhaps proof read the above?

swissspidy: Sounds good to me.

We should also create a documentation page on amp-wp.org and point to that URL in this readme.

There we can add screenshots, instructions, etc.

MackenzieHartung: The revisions @ThierryA made look good to me. I am assuming, however, that we will be moving the bulk of this content to the amp-wp.org site, and point to there from the readme as @swissspidy mentioned. Correct?

swissspidy: Yes. I think we'll keep like the first sentence or so, and then add the link to the docs.

swissspidy:
Updated the readme and moved the content to https://amp-wp.org/documentation/amp-stories/ / https://amp-wp.org/documentation/amp-stories/exporting-stories/

PR #2809, opened 2019-07-15
Fix FE and editor difference for text block

### LINKED ISSUE #2721: Text does not display in Preview/Publish the same way it does in the editor
Text in preview is a different size than text in editor

Editor:
Screen Shot 2019-07-01 at 4 42 23 PM

Preview:
Screen Shot 2019-07-01 at 4 40 50 PM

Text wrapped in editor is not wrapped in preview

Editor: 
Screen Shot 2019-07-01 at 4 43 49 PM

Preview:
Screen Shot 2019-07-01 at 4 44 17 PM

miina:
Fixes #2721 

- Fixes incorrect width and height saved for Text blocks.
- Fixes font size calculation due to font family not applied during font calculation.

AC: If the aspect ratio is exactly the same, the text should match exactly in the FE and in the editor -- there has to be the same amount of rows and the same words have to be on the same rows.
AC: If the text is saved in one font size when amp-fit-text is on, it should not change upon refresh.


swissspidy:
```
/home/travis/build/ampproject/amp-wp/assets/src/stories-editor/helpers/index.js
  42:2  error  'TEXT_BLOCK_BORDER' is defined but never used  no-unused-vars
```

miina:
OK, now the `npm run lint:js` ran successfully, sorry for that.

swissspidy:
@miina How can I best test this?

Issue #2795, opened 2019-07-11
Adding E2E Tests


swissspidy:
Now that we have the E2E testing infrastructure set up (see #2696), we should look into how we can best leverage that to further improve code quality and catch bugs in the AMP Stories editor.

While we of course can add E2E tests for all aspects of the AMP plugin, we should focus on AMP Stories first. Still, I am mentioning some of the other areas in the plugin here.

#2779 is related, but I'd want to focus on E2E tests first instead of the unit tests, as I think here we have the bigger opportunity to solidify the AMP Stories editor.

Ideally, going forward we add E2E tests whenever we work on an area that is a good fit for such a test or when we introduce a new feature. Then, later we can go through pre-existing areas to add them.

**Testable Areas**

The following list is not necessarily complete, but should give a good idea about what we can and should test.

- [ ] Block Editor
    - [ ] Reader Mode
        - [ ] Verify that customizer controls and toggles are visible and working
    - [ ] Transitional Mode
        - [ ] Verify that AMP blocks are not available in the editor
    - [ ] Standard Mode
        - [ ] Verify that AMP blocks are available in the editor
    - [ ] Block validation
        - [ ] Verify that block validation happens when AMP is enabled for the current post.
        - [ ] Verify that block validation is not happening when AMP is disabled for the current post.
- [ ] Settings
    - [ ] Verify visibility of post type selection area depending on whether Website Mode is active.
    - [ ] Verify that Story Export URL is only visible when Stories experience is toggled
- [ ] Validation Screen
    - [ ] Verify status changes
    - [ ] Verify alter when there are unsaved changes
- [ ] Integrations
    - [ ] PWA (there is nothing user-facing yet)
    - [ ] Site Kit (maybe in the future if not already done in Site Kit)

**Infrastructure Changes**

* [ ] Create additional helper plugins that we might need during E2E tests, and set this up accordingly.
  * One example could be a plugin that makes sure pointers can not be dismissed permanently, makes testing of these easier
* [ ] Possibly use smarter APIs like `expect(instance).toMatchElement(selector[, options])` instead of XPath -> #2888.
* [ ] Possibly also run E2E tests against an older WordPress version

miina:
FI: I'm adding some CTA-related e2e tests within #2824.

westonruter:
For testing mobile redirection, see https://github.com/ampproject/amp-wp/issues/5268.

PR #2810, opened 2019-07-15
Improve CI configuration

swissspidy:
* Fix definition of global env variable
* Remove incorrect docker volumes
* Stop local environment after running e2e tests
* Add demo plugin that demonstrates how plugins can be used during e2e tests
* Add new CI step to run e2e tests as author

Related: #2795.

### DIFF
--- a/tests/e2e/plugins/demo.php
+++ b/tests/e2e/plugins/demo.php
@@ -1,0 +1,10 @@
+<?php
+/**
+ * Plugin Name: E2E Tests Demo Plugin
+ * Plugin URI:  https://github.com/ampproject/amp-wp
+ * Description: Demo Plugin that can be installed during E2E tests.
+ * Author:      AMP Project Contributors
+ * Author URI:  https://github.com/ampproject/amp-wp/graphs/contributors
+ */
+
+// Silence is golden.
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,6 +93,7 @@ jobs:
         - npm run build:js # after npm dependencies have been installed
         - npm run env:reset-site
         - npm run test:e2e
+        - npm run env:stop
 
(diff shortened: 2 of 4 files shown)


westonruter:
Why is the demo plugin needed here?

swissspidy:
@westonruter It basically acts as documentation to show how adding plugins for E2E tests works.

Alternatively we can just wait until we actually need some of these plugins when adding the first bunch of tests.

westonruter:
I'm confused though: these tests are for the AMP plugin. Why would adding plugins need to be tested in this project?

westonruter:
Unless the intention is to test that AMP validation errors are getting properly attributed?

swissspidy:
@westonruter It's not about testing plugins. This change allows for adding some utility plugins that can be activated during E2E tests. 

For example, such a plugin could create a custom post type and the accompanying E2E test would verify that block validation errors are shown for this custom post type too.

tests/e2e/plugins/demo.php
```diff
@@ -0,0 +1,10 @@
+<?php
+/**
+ * Plugin Name: E2E Tests Demo Plugin
+ * Plugin URI:  https://github.com/ampproject/amp-wp
+ * Description: Demo Plugin that can be installed during E2E tests.
```
miina: > For example, such a plugin could create a custom post type and the accompanying E2E test would verify that block validation errors are shown for this custom post type too.

@swissspidy Could we add this clarifying example to the description here, too?

swissspidy: @miina Sure. Maybe in `contributing.md` though?

miina: Yep, makes sense 👍

swissspidy: Done in 2fdebe1ed017294154c015864e656c4a94b9f037

PR #2811, opened 2019-07-15
Make the asset path relative to the story

valendesigns:
Also fix updating of `canonical` link.

### DIFF
--- a/assets/src/stories-editor/plugins/export.js
+++ b/assets/src/stories-editor/plugins/export.js
@@ -14,7 +14,7 @@ import { __ } from '@wordpress/i18n';
 
 const { ampStoriesExport, fetch, FormData, URL } = window;
 
-const handleExport = ( { postId, createErrorNotice, createSuccessNotice } ) => {
+const handleExport = ( { postId, createErrorNotice, createSuccessNotice, removeNotice } ) => {
 	const formData = new FormData();
 	const errorMsg = __( 'Could not generate story archive.', 'amp' );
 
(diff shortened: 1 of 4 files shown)


includes/sanitizers/class-amp-story-export-sanitizer.php
```diff
@@ -108,19 +108,32 @@ public function sanitize() {
...
-					'href' => $this->args['canonical_url'],
-				]
-			);
-
-			$head = $this->dom->getElementsByTagName( 'head' )->item( 0 );
-			$head->appendChild( $rel_canonical );
+			$head          = $this->dom->getElementsByTagName( 'head' )->item( 0 );
+			$links         = [];
+			$link_elements = $head->getElementsByTagName( 'link' );
+			$rel_canonical = null;
```
valendesigns: I can remove this, it's not needed.

assets/src/stories-editor/plugins/export.js
```diff
@@ -23,11 +23,30 @@ const handleExport = ( { postId, createErrorNotice, createSuccessNotice } ) => {
...
+	 * Show error notice.
+	 *
+	 * @param {?Error} error
+	 */
+	const showErrorNotice = ( error = null ) => {
+		removeNotice( progressNoticeId );
+		createErrorNotice( error ? error.message : errorMsg,
+			{
+				id: errorNoticeId,
+				type: 'snackbar',
```
valendesigns: The error shouldn't be a snackbar.

assets/src/stories-editor/plugins/export.js
```diff
@@ -56,21 +78,19 @@ const handleExport = ( { postId, createErrorNotice, createSuccessNotice } ) => {
...
 				// Handle the returned JSON error.
 				response.json()
 					.then( ( error ) => {
-						createErrorNotice( ( error.data && error.data.errorMessage ) ? error.data.errorMessage : errorMsg, {
-							id: 'amp-story-export__error-notice',
-						} );
-					} );
+						showErrorNotice( ( error.data && error.data.errorMessage ) ? new Error( error.data.errorMessage ) : null );
+					} )
+					.catch( showErrorNotice );
```
valendesigns: Will writing it this way pass the error message to the function as a param automatically?

westonruter: Yes. The callback is passed an `Error` object which then has a `message` property.

assets/src/stories-editor/plugins/export.js
westonruter: Fixed in cd68cb73

PR #2814, opened 2019-07-15
Add webpackbar plugin to have nicer output

swissspidy:
I was a bit jealous about the Site Kit plugin's nice output when running webpack, so I quickly installed the same `webpackbar` package.

Result:

Screenshot 2019-07-15 at 20 18 51


webpack.config.js
```diff
@@ -130,6 +149,13 @@ const classicEditor = {
 	entry: {
 		'amp-post-meta-box': './assets/src/classic-editor/amp-post-meta-box.js',
 	},
+	plugins: [
+		...defaultConfig.plugins,
+		new WebpackBar( {
+			name: 'Block Editor',
```
miina: @swissspidy  Just in case -- is it intended for this to be `Block Editor` and not `Classic Editor`?

swissspidy: D‘oh, yes. Copy/Paste mistake.

PR #2642, opened 2019-06-16
Add looping for video headers; extend tree-shaking for amp-video-* class names

westonruter:
Theme support for the video header is lacking in the plugin. In particular, the video header is failing to loop. This fixes that.

To test, use a theme that supports video headers (e.g. Twenty Seventeen). Make sure you select a short MP4 video and a short YouTube video, for example https://www.youtube.com/watch?v=qbGbDNYArTU

Note that in order for the video header to appear in AMP, you _must_ currently also select a video header _image_. This is because the AMP video is injected via the `get_header_image_tag` filter which only applies if a header image is assigned.

The ability to loop YouTube videos recently was made more seamless via https://github.com/ampproject/amphtml/pull/22858. Note there is still a momentary flicker of the title: https://github.com/ampproject/amphtml/issues/22855#issuecomment-511674431.

# Changes

- Add missing `loop` attribute on MP4 video header loops in AMP.
- Ensure YouTube video header loops in AMP (see https://github.com/ampproject/amphtml/issues/22855); this mostly works, but the looping is not perfect as can be seen in this comparison video: https://youtu.be/8PZGJf3dtwM
- Update AMP validator spec once https://github.com/ampproject/amphtml/pull/22858 live and use new `loop` attribute instead of `data-param-loop` and `data-param-playlist`. See https://github.com/ampproject/amp-wp/pull/2816.
- Hide video autoplay equalizer animation in bottom-right corner of video header in AMP.
- Update tree-shaker to prevent removing style rules with `amp-video-*` class names when any AMP component using the video-loader is present, and thus any styles in [`video-autoplay.css`](https://github.com/ampproject/amphtml/blob/568c032d0c5bf6538c61dd546d43be0969c0ceb4/css/video-autoplay.css).

# Out of scope

- Fix video header so not dependent on header image being selected.

PR #2819, opened 2019-07-16
Add pause button for video header

westonruter:
As noted by @swissspidy in #2642:

> In the non-AMP version, `wp-includes/js/wp-custom-header.js` adds a button with the ID `wp-custom-header-video-button` that allows to pause/play the video header, to make custom video headers more accessible.
>
> It would be nice if we could keep this accessibility level. Perhaps in a separate PR?

This PR looks to add that button.

There is one issue, however: `amp-bind` is not set up to allow toggling the `paused` state of an `amp-video`. Neither is there a `togglePlay` action exposed on `amp-video`. There are only `play` and `pause` actions. This means that the singular `wp-custom-header-video-button` will not work as-is in how it toggles pause/play. There will need to be two separate buttons, one for `pause` and another for `play` and only one can be shown at a time. For accessibility, we'll need to ensure that `focus` is put on the alternate button when toggling.

We should also consider adding the play/pause button for the YouTube videos.

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -2422,6 +2422,28 @@ class AMP_Theme_Support {
 					]
 				)
 			);
+
+			// @todo Also add this for amp-youtube?
+			$video_markup .= '<amp-state id="headerVideoPaused"><script type="application/json">false</script></amp-state>';
+			$video_markup .= sprintf(
+				'<button type="button" id="wp-custom-header-video-button" class="wp-custom-header-video-button wp-custom-header-video-play" [class]="%s" on="%s">',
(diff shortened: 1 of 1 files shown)


westonruter:
> There is one issue, however: `amp-bind` is not set up to allow toggling the `paused` state of an `amp-video`. Neither is there a `togglePlay` action exposed on `amp-video`. There are only `play` and `pause` actions. This means that the singular `wp-custom-header-video-button` will not work as-is in how it toggles pause/play. There will need to be two separate buttons, one for `pause` and another for `play` and only one can be shown at a time. For accessibility, we'll need to ensure that `focus` is put on the alternate button when toggling.

Todo: Raise this with AMP core.

schlessera:
> Todo: Raise this with AMP core.

=> https://github.com/ampproject/amphtml/issues/25669

PR #2823, opened 2019-07-17
Automatically set poster images for videos

### LINKED ISSUE #903: Automatically provide poster for videos
Often when adding a video to a post in testing, the post ends up being invalid AMP with an:

> No "poster" attribute has been provided for amp-video.

To me it seems a fallback poster should be supplied to prevent such a ~validation~ error in the console. _It isn't actually a validation error._

kienstra:
* So far, on selecting a video, this looks to see if it has an image in the `media` object.
* If so, it uses that image as the poster.
* This applies to the Video block and the AMP Story block 'Background Media'.
* So far, this only implements the beginning of the [requirements](https://github.com/ampproject/amp-wp/issues/903#issuecomment-498914903).
* Even this might need refinement, like seeing if that image is the proper size. There's a lot more work needed on this PR.

Closes #903

### DIFF
--- a/assets/src/common/helpers/index.js
+++ b/assets/src/common/helpers/index.js
@@ -392,3 +392,27 @@ export const getContentLengthFromUrl = async ( url ) => {
 	} );
 	return Number( response.headers.get( 'content-length' ) );
 };
+
+/**
+ * On selecting a video, ensure that there is a poster image.
+ *
+ * Not an arrow function so that it can be called like maybeSupplyPoster.call( this, media ).
+ * @todo: if the image matches the defaultImagePattern or it doesn't exist,
+ * get an image using canvas, like Weston suggested.
+ * Then POST the image to an endpoint, and use that image if it succeeds.
+ *
+ * @param {Object} media The selected media.
+ */
+export const maybeSupplyPoster = function( media ) {
+	if ( ! media || ! media.image ) {
+		return;
+	}
+
+	// Videos often have a 'default' image from WP core as media.image, which isn't useful as a poster.
+	const defaultImagePattern = /wp-includes\/images\/media\/video\.png$/;
+	const { image } = media;
+	if ( ! image.src.match( defaultImagePattern ) ) {
+		const { setAttributes } = this.props;
+		setAttributes( { poster: image.src } );
+	}
+};
--- a/assets/src/stories-editor/components/custom-video-block-edit.js
+++ b/assets/src/stories-editor/components/custom-video-block-edit.js
@@ -37,7 +37,7 @@ import { withSelect } from '@wordpress/data';
 /**
  * Internal dependencies
  */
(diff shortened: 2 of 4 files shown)


kienstra:
**Sets Poster If Featured Image Happens To Exist**

This PR is still in an early stage. But so far, it uses a video's featured image as the poster, if one exists.

f-here

You might have to manually set a video's featured image in order to see this work. I don't think videos usually have them.

1. Go to the Media Library and click a video:

click-a-video

2. Click 'Edit more details':
edit-de

3. Click 'Set featured image'

Of course, this PR will eventually avoid these manual steps.

assets/src/common/helpers/index.js
```diff
@@ -392,3 +392,27 @@ export const getContentLengthFromUrl = async ( url ) => {
...
+ *
+ * @param {Object} media The selected media.
+ */
+export const maybeSupplyPoster = function( media ) {
+	if ( ! media || ! media.image ) {
+		return;
+	}
+
+	// Videos often have a 'default' image from WP core as media.image, which isn't useful as a poster.
+	const defaultImagePattern = /wp-includes\/images\/media\/video\.png$/;
```
swissspidy: Why not use `String.prototype.endsWith()` instead of regex?

assets/src/stories-editor/components/custom-video-block-edit.js
```diff
@@ -188,6 +188,7 @@ class CustomVideoBlockEdit extends Component {
 			// selected media, then switches off the editing UI
 			setAttributes( { src: media.url, id: media.id } );
 			this.setState( { src: media.url, editing: false, duration: null, videoSize: null } );
+			maybeSupplyPoster.call( this, media );
```
swissspidy: We're already providing the poster image in `componentDidUpdate` if the video has a featured image (`videoFeaturedImage`). So what's the difference to that approach?

swissspidy:
I'm a bit confused because we're already doing all of this. 🤔

My understanding was that we need to add poster images for videos who don't already have one, by extracting the first frame of the video in a canvas. This was outlined in https://github.com/ampproject/amp-wp/issues/903#issuecomment-498914903

swissspidy:
See #2826 for an initial proof-of-concept for the originally thought-out implementation.

assets/src/stories-editor/components/custom-video-block-edit.js
kienstra: Good point, that's already provided.

kienstra:
Hi @swissspidy,
>I'm a bit confused because we're already doing all of this. 🤔

>My understanding was that we need to add poster images for videos who don't already have one, by extracting the first frame of the video in a canvas. This was outlined in #903 (comment)

You're right, I missed the fact that the poster [is already provided](https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/blocks/amp-story-page/edit.js#L135) if it exists. This PR isn't helpful.

PR #2826, opened 2019-07-17
Extract video frames to automatically set poster images

swissspidy:
To do:

* [x] Set extracted poster image as featured image of the video attachment.
* [x] Extract poster image for page background video as well.
* [x] Hide poster image attachments in media library?
* [x] Perhaps move REST API code elsewhere
* [x] Add tests for REST API field addition.
* [x] Hide poster image section during video frame extraction

See #903.


assets/src/stories-editor/helpers/index.js
```diff
@@ -1508,3 +1508,30 @@ export const getCallToActionBlock = ( pageClientId ) => {
...
+		video.src = src;
+		video.addEventListener( 'loadeddata', () => {
+			const canvas = document.createElement( 'canvas' );
+			canvas.width = video.videoWidth;
+			canvas.height = video.videoHeight;
+
+			const ctx = canvas.getContext( '2d' );
+			ctx.drawImage( video, 0, 0, canvas.width, canvas.height );
+
+			canvas.toBlob( resolve );
```
miina: ^ Does this work for any URL-s, external included?

assets/src/stories-editor/components/custom-video-block-edit.js
```diff
@@ -341,25 +357,62 @@ CustomVideoBlockEdit.propTypes = {
...
 
-		if ( props.attributes.id && ! props.attributes.poster ) {
-			const media = getMedia( props.attributes.id );
-			const featuredImage = media && get( media, [ '_links', 'wp:featuredmedia', 0, 'href' ], null );
-			videoFeaturedImage = featuredImage && getMedia( Number( featuredImage.split( '/' ).pop() ) );
+		const { id, poster } = attributes;
+
+		if ( id && ! poster ) {
+			const media = getMedia( id );
+			videoFeaturedImage = media && media.featured_media > 0 && getMedia( media.featured_media );
```
miina: It looks like by the time when `videoFeaturedImage` is received, getting the first frame of the video has already run, meaning that a new image is created even though a previous image exists already. Is it using the existing image properly when testing?

assets/src/stories-editor/components/custom-video-block-edit.js
```diff
@@ -101,10 +103,19 @@ class CustomVideoBlockEdit extends Component {
 			getContentLengthFromUrl( src ).then( ( videoSize ) => {
 				this.setState( { videoSize } );
 			} );
+
+			if ( ! poster ) {
+				if ( videoFeaturedImage ) {
+					setAttributes( { poster: videoFeaturedImage.source_url } );
+				} else {
+					uploadVideoFrame( src );
```
miina: Would be good to display a thumbnail, too, it's unclear right now which image is being used, it just shows "Replace Image".

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -340,14 +362,12 @@ class PageEdit extends Component {
 										) }
 										id="story-background-media"
 									/>
+									{ mediaUrl && (
+										<Button onClick={ () => setAttributes( { mediaUrl: undefined, mediaId: undefined, mediaType: undefined } ) } isLink isDestructive>
+											{ _x( 'Remove', 'background media', 'amp' ) }
```
miina: Maybe we shouldn't even show the `Remove` button anymore? Since it's set automatically anyway if missing, maybe just "Replace" is enough?

assets/src/stories-editor/helpers/index.js
```diff
@@ -1508,3 +1508,30 @@ export const getCallToActionBlock = ( pageClientId ) => {
...
+export const getFirstFrameOfVideo = async ( src ) => {
+	const video = document.createElement( 'video' );
+	video.setAttribute( 'muted', 'muted' );
+
+	return new Promise( ( resolve ) => {
+		video.src = src;
+		video.addEventListener( 'loadeddata', () => {
+			const canvas = document.createElement( 'canvas' );
+			canvas.width = video.videoWidth;
+			canvas.height = video.videoHeight;
```
miina: Should we also consider the recommended media limits?

assets/src/stories-editor/helpers/index.js
swissspidy: There can be CORS issues with external URLs: https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image#Security_and_tainted_canvases

assets/src/stories-editor/components/custom-video-block-edit.js
swissspidy: True. This could be more aligned with how the page background poster is displayed.

assets/src/stories-editor/blocks/amp-story-page/edit.js
swissspidy: Yup! Already got rid of that link in the video block.

assets/src/stories-editor/helpers/index.js
swissspidy: Set the `crossOrigin` attribute now, which seems to resolve any issues with this.

assets/src/stories-editor/blocks/amp-story-page/edit.js
swissspidy: Thought: With this change we're likely always going to have a poster image - why even show the replacement option?

miina: Hmm. Probably some users would want to choose a poster image they like, the first frame isn't necessarily a good representation of the video. True that most users probably wouldn't change the setting but there are probably some who would want to choose that.

assets/src/stories-editor/helpers/index.js
swissspidy: I guess we can upload the original one and then let WordPress crop the image. I just adjusted `uploadVideoFrame` in the component to check for `MAX_IMAGE_SIZE_SLUG`.

assets/src/stories-editor/components/custom-video-block-edit.js
swissspidy: You're right, it does upload it again. 😞 I'll look into it.

swissspidy:
Added another set of improvements that could benefit from some more reviewing.

I should be able to address the last remaining todo tomorrow.

miina:
Sorry for the delay, _checking now_.

swissspidy:
> Hide poster image section during video frame extraction

Not sure how feasible that is, will have to check it out. Perhaps it's not that big of an issue though.

includes/class-amp-story-media.php
```diff
@@ -0,0 +1,329 @@
...
+		// The AJAX handler for when an image is cropped and sent via POST.
+		add_action( 'wp_ajax_custom-header-crop', [ __CLASS__, 'crop_featured_image' ] );
+
+		add_action( 'pre_get_posts', [ __CLASS__, 'filter_poster_attachments' ] );
+
+		add_action( 'rest_api_init', [ __CLASS__, 'rest_api_init' ] );
+	}
+
+	/**
+	 * If there's no featured image for the poster-portrait-src, this adds a fallback.
```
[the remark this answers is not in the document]
swissspidy: That‘s just a method I moved from the post type class to this one. It‘s not related to this PR‘s changes.

assets/src/stories-editor/components/custom-video-block-edit.js
```diff
@@ -232,45 +238,49 @@ class CustomVideoBlockEdit extends Component {
...
 								className="editor-video-poster-control"
 							>
-								<BaseControl.VisualLabel>
-									{ __( 'Poster Image', 'amp' ) }
-								</BaseControl.VisualLabel>
+								{
+									! poster &&
+									<Notice status="error" isDismissible={ false } >
+										{ __( 'A poster image must be set.', 'amp' ) }
+									</Notice>
```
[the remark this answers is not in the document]
swissspidy: Hence the open todo item in this PR to hide the notice during extraction.

includes/class-amp-story-media.php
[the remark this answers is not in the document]
miina: Oh, right, that's for the general poster image, not for videos 👍

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -492,37 +497,87 @@ PageEdit.propTypes = {
...
+		source_url: PropTypes.string,
+	} ),
+	uploadVideoFrame: PropTypes.func,
 };
 
 export default compose(
-	withSelect( ( select, { clientId, attributes } ) => {
+	withDispatch( () => {
+		const { saveMedia } = dispatch( 'core' );
+		const { moveBlockToPosition	} = dispatch( 'core/block-editor' );
```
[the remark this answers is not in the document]
swissspidy: Hmm, I'll see if there's an ESLint rule that can catch these kind of things

assets/src/stories-editor/components/custom-video-block-edit.js
```diff
@@ -105,21 +105,29 @@ class CustomVideoBlockEdit extends Component {
...
+
+		if ( ! poster && videoFeaturedImage ) {
+			setAttributes( { poster: videoFeaturedImage.source_url } );
+		}
+
+		if ( ! poster && ( prevProps.attributes.src !== src || prevProps.media !== media ) ) {
+			if ( ! id || ( media && ! media.featured_media ) ) {
+				uploadVideoFrame( src );
+			}
+		}
```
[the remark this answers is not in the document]
swissspidy: I'll add a comment as an explainer and add the poster condition as a wrapper.

kienstra:
**Thanks, reviewing now**

Hi @swissspidy,
Thanks, I'm also reviewing this now.

kienstra:
**Approved**

Hi @swissspidy,
This looks really good. It extracts the posters for the Background Media and Video block:

# Background Media
poster-image-appears

# Video Block
video-poster-image

As expected, the poster image doesn't appear in the Media Library when clicking 'Background Media' > 'Select Media,' for example:

poster-im-not-here

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -144,16 +148,21 @@ class PageEdit extends Component {
...
-				}
+				this.videoPlayer.current.load();
 			}
 
 			if ( ! poster && ( prevProps.attributes.mediaUrl !== mediaUrl || prevProps.media !== media ) ) {
 				if ( videoFeaturedImage ) {
 					setAttributes( { poster: videoFeaturedImage.source_url } );
 				} else if ( ! mediaId || ( media && ! media.featured_media ) ) {
-					uploadVideoFrame( mediaUrl );
+					/*
```
miina: I wonder if we should generate the poster image for already existing videos as well, not only when the media gets updated? E.g. if there's a saved video without the poster image and the Story gets loaded, should we also set the poster image? Perhaps it's OK not to 🤷‍♀ Thoughts?

miina:
One thing just noticed:
In case of Video block, the generated poster image does seem to show up in the Media Library.

assets/src/stories-editor/blocks/amp-story-page/edit.js
swissspidy: That should actually work already, because `componentDidUpdate` gets called after initialization too.

You can try by checking out `develop`, setting a video, then checking out this branch, building, and reloading the editor.

swissspidy:
> In case of Video block, the generated poster image does seem to show up in the Media Library.

Will test. Perhaps there was a timeout or so when updating the attachment. Was anything logged in the console?

If you run `wp.data.select( 'core' ).getMedia( 123456 )` for the poster image, do you see any `meta` entry for it?

assets/src/stories-editor/blocks/amp-story-page/edit.js
miina: Hmm, maybe, but is shoudln't satisfy this condition :`prevProps.attributes.mediaUrl !== mediaUrl || prevProps.media !== media`?

Just tried with background media and it doesn't seem to generate the poster image.

miina:
> Will test. Perhaps there was a timeout or so when updating the attachment. Was anything logged in the console?

Just tried again. Console:
Screenshot 2019-07-23 at 20 58 00

Yep, meta shows, looks like not as expected:
Screenshot 2019-07-23 at 21 01 59

Now, the first time when I tried, it didn't show the generated image in the Media Library, however, when I edited the Video block by uploading another Video then this happened.

swissspidy:
Oh, it tries to send it to the not-yet-uploaded image (not the blob). Should be an easy fix. One second.

assets/src/stories-editor/blocks/amp-story-page/edit.js
swissspidy: Refactored the code a bit now and removed the logic for "upgrading" existing videos.

swissspidy:
Apparently there's one last thing that was caught in review by @miina:

> There seems to be one issue — when switching to a new video then the frame extraction is triggered instead of using the already set featured image of the video.
> To specify — in the beginning the correct image is set.
> And then it’s replaced by the new extracted image.

swissspidy:
So far it seems to work well. Going to add E2E tests in a new PR.

PR #2857, opened 2019-07-20
Adjust CTA style considering potential image & fix spacings.

mehigh:
Fixes QA feedback received in #2706

The issue which was present on https://story-test-wordpress-amp.pantheonsite.io/stories/title-here/ and highlighted by Claudio was happening because there was no height applied on the wrapper. 
As I dived deeper into this I noticed some paddings were not respected when an image was added inside, so I fixed that too. 
Another glitch was happening when the CTA was the only block inside of the story and aligning it to the left or right inherited some 34px top margin from the Gutenberg core styles which I had to overwrite to keep the CTA in place and not move up/down depending on the alignment.
Screen recording of how this looks: https://dsh.re/8d464

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-cta/edit.css
+++ b/assets/src/stories-editor/blocks/amp-story-cta/edit.css
@@ -13,9 +13,6 @@
 
 	/* Remove the top/bottom padding from the available space. */
 	max-height: calc(110px - 24px);
-
-	/* Counteract the height increase caused by the inline-block. */
-	margin-bottom: -9px;
 }
 
 .amp-block-story-cta__link img {
--- a/assets/src/stories-editor/blocks/amp-story-cta/edit.css
+++ b/assets/src/stories-editor/blocks/amp-story-cta/edit.css
@@ -5,10 +5,17 @@
 	height: 20%;
 }
 
-.editor-block-list__layout div[data-type="amp/amp-story-cta"] * {
+.block-editor-block-list__layout div[data-type="amp/amp-story-cta"]:not(#id) {
+	margin-top: 0;
+}
 
-	/* 20% of the full height of a page in the editor */
-	max-height: 110px;
+.editor-block-list__layout div[data-type="amp/amp-story-cta"] img {
+
+	/* Remove the top/bottom padding from the available space. */
+	max-height: calc(110px - 24px);
+
+	/* Counteract the height increase caused by the inline-block. */
+	margin-bottom: -9px;
 }
 
 .amp-block-story-cta__link img {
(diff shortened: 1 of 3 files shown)


mehigh:
I'm OK to contribute this to open source :) (remember needing to send such a message on the last contribution).

miina:
@mehigh Just tested the PR, looks like in the FE the button is not still not displaying correctly, the editor view looks good.

The button in the FE should adjust the style based on the size of the image, however, not it has a fixed height. For example:
Editor view (correct):
Screenshot 2019-07-22 at 11 51 26

FE -- image is much smaller but the button is still with max available height.
Screenshot 2019-07-22 at 11 51 36

Same happens without having an image:
Editor (correct):
Screenshot 2019-07-22 at 11 54 07

FE:
Screenshot 2019-07-22 at 11 54 13

swissspidy:
@mehigh Are you able to address that early this week?

mehigh:
I intend to work on it in 3-4h today. So, yeah, the plan is to turn it
around quickly.

mehigh:
I'm working on this now.

miina:
@swissspidy Would you mind giving it a quick test as well to see if you notice any issues -- last time it was working OK for me locally but not in the test environment, so would be good to double-check just in case.

swissspidy:
For me it looks like this with 24x18 image:

Backend:

Screenshot 2019-07-22 at 19 50 03

Frontend:

Screenshot 2019-07-22 at 19 49 18

FE is alright, but the vertical alignment in the editor is definitely off. Am I missing something?

assets/src/stories-editor/blocks/amp-story-cta/edit.css
```diff
@@ -5,10 +5,17 @@
...
+
+.editor-block-list__layout div[data-type="amp/amp-story-cta"] img {
+
+	/* Remove the top/bottom padding from the available space. */
+	max-height: calc(110px - 24px);
 
-	/* 20% of the full height of a page in the editor */
-	max-height: 110px;
+	/* Counteract the height increase caused by the inline-block. */
+	margin-bottom: -9px;
```
miina: @mehigh Looks like the margin-bottom here is causing the vertical alignment display incorrectly in the editor, this can be seen in case of tiny images specifically, for example:
Screenshot 2019-07-22 at 21 22 19

Without this margin:
Screenshot 2019-07-22 at 21 22 27

Perhaps it's OK to remove this afterall?

miina:
Not missing anything, you're right, @swissspidy, I was testing with images with ratio where it didn't show this clearly. Thanks!

Added a comment to the related change: https://github.com/ampproject/amp-wp/pull/2857/files#r305968032

assets/src/stories-editor/blocks/amp-story-cta/edit.css
mehigh: Yeah, let's remove it.

miina:
Works as expected now, thanks, Mike!

PR #2824, opened 2019-07-17
Allow dragging CTA button within the CTA block

miina:
Closes #2781. 
Closes #2899.

Left todo:
- [x] Look into why sometimes top and left attribute are assigned to CTA block. (Needs testing).
- [x] Adjust CTA block styling to match the borders exactly to make the drop logic more precise.
- [x] Remove alignment options.
- [x] Maybe don't display the link when dragging.
- [x] Make the adjustments to `save` and FE view, too (see how to do that without breaking existing blocks).
- [x] See if can make more DRY. 
- [x] Tests (e2e).
- [ ] Update the logic once #2863 gets merged.

Issue #2787, opened 2019-07-10
Allow selecting of one or more elements upon command-click, select


MackenzieHartung:
As a user I want to be able to select more than on block at a time.

AC1: Upon holding the command key down, and selecting individual blocks a user will be able to select one or more blocks.
AC2: Blocks will remain selected, until user clicks without holding command

swissspidy:
So.... In Gutenberg multi-selection is really designed for selecting a range of blocks from start to end, not a loose list of blocks that could be all over the place.

Examples in the code:

https://github.com/WordPress/gutenberg/blob/1b5fc6a593581bfe1ee75128f0d05c154c4b8fdf/packages/block-editor/src/store/selectors.js#L538-L598

https://github.com/WordPress/gutenberg/blob/1b5fc6a593581bfe1ee75128f0d05c154c4b8fdf/packages/block-editor/src/store/selectors.js#L313-L337

https://github.com/WordPress/gutenberg/blob/1b5fc6a593581bfe1ee75128f0d05c154c4b8fdf/packages/block-editor/src/store/actions.js#L157-L193

This is what you currently get when selecting two blocks in a row in Gutenberg:

![Screenshot 2019-07-22 at 15 21 55](https://user-images.githubusercontent.com/841956/61635776-91a29780-ac94-11e9-8cc9-ef71a5d9c1e4.png)

As you can see, one can easily (un)group or remove a list of selected blocks. There's also a selection indicator in the sidebar. These features directly rely on these selectors, e.g. `getMultiSelectedBlocks()`, `getMultiSelectedBlockClientIds()`, and `getSelectedBlockClientIds()`.

**Ideally** we would make a contribution to Gutenberg to change the way multi-selection works internally, to allow multi-selection without having to specify a range.

Next, we could add a change that allows multi-selecting blocks by using Cmd+Click. (In the same PR or a new one)

That means the change would make it into an upcoming version of the Gutenberg plugin, and then into WordPress 5.3.

**Alternatively** we could try to roll our own, which would mean we need to override a bunch of components. Not sure how feasible it is, but given the additional maintenance burden, I don't think we can get around contributing this directly upstream.

swissspidy:
Upstream issue created: https://github.com/WordPress/gutenberg/issues/16797

^ cc @miina

swissspidy:
Upstream PR work in progress: https://github.com/WordPress/gutenberg/pull/16811

swissspidy:
[...]

PR #2863, opened 2019-07-22
Require double click (or just two clicks) for editing.

### LINKED ISSUE #2783: Enable users to click and drag anywhere within a block, if block contains text require a double click to edit text.
As a user I want to be able to click anywhere on a block, and drag/place it elsewhere on a page.

AC1: Single clicking anywhere on a block, and dragging will drag a block.
AC2: Upon release block will remain in new place.
AC3: Double clicking within a text block will enable edit-mode.

miina:
Closes #2783. 

Requires double click for editing the Text block.

Todo:

- [x] Display edit mode only on double click.
- [x] Remove dragging area border and related logic.
- [x] Don't allow dragging in edit mode.
- [x] Better focus when going into edit mode.
- [x]  e2e tests.


assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -67,6 +78,20 @@ class TextBlockEdit extends Component {
...
+		if ( this.state.isEditing && ! prevState.isEditing ) {
+			const textInput = document.querySelector( '.is-selected .wp-block-amp-amp-story-text' );
+			if ( textInput ) {
+				// Create selection, collapse it in the end of the content.
+				const range = document.createRange();
+				range.selectNodeContents( textInput );
+				range.collapse( false );
+				const selection = window.getSelection();
+				selection.removeAllRanges();
+				selection.addRange( range );
```
miina: Couldn't find a better way how to focus after going to editable state. See also https://github.com/WordPress/gutenberg/issues/9740.

swissspidy: Maybe we can ask Ella or generally in #core-editor to find out if there's a better way.

miina: 👍 Will do.

assets/src/stories-editor/components/resizable-box/index.js
```diff
@@ -88,19 +80,24 @@ const EnhancedResizableBox = ( props ) => {
...
+
 				// Ensure the measures not crossing limits.
 				appliedWidth = appliedWidth < lastWidth ? lastWidth : appliedWidth;
 				appliedHeight = appliedHeight < lastHeight ? lastHeight : appliedHeight;
 
 				const elementTop = parseFloat( blockElement.style.top );
 				const elementLeft = parseFloat( blockElement.style.left );
 
-				const positionTop = ! isText ? Number( elementTop.toFixed( 2 ) ) : Number( ( elementTop + textBlockBorderInPercentageTop ).toFixed( 2 ) );
-				const positionLeft = ! isText ? Number( elementLeft.toFixed( 2 ) ) : Number( ( elementLeft + textBlockBorderInPercentageLeft ).toFixed( 2 ) );
```
kienstra: It's great to see these calculations simplified.

assets/src/stories-editor/components/resizable-box/index.js
```diff
@@ -232,8 +229,12 @@ const EnhancedResizableBox = ( props ) => {
...
 				lastHeight = appliedHeight;
 
-				if ( textBlockWrapper && ampFitText ) {
-					textBlockWrapper.style.lineHeight = isText ? appliedHeight - ( TEXT_BLOCK_PADDING * 2 ) + 'px' : appliedHeight + 'px';
+				if ( textBlockWrapper ) {
+					if ( ampFitText ) {
+						textBlockWrapper.style.lineHeight = appliedHeight + 'px';
+					}
+					// Also add the height to the wrapper since the background color is set to the wrapper.
+					textBlockWrapper.style.height = appliedHeight + 'px';
```
kienstra: Nice refactoring of this conditional block. It's easier to understand.

assets/src/stories-editor/constants.js
```diff
@@ -234,5 +234,4 @@ export const AMP_STORY_FONT_IMAGES = {
 	Ubuntu,
 };
 
-export const TEXT_BLOCK_BORDER = 5;
```
kienstra: It's great that this workaround isn't needed anymore.

kienstra:
Hi @miina,
This looks good. Editing and dragging of the Text block works well:

dragging-editing

In my opinion, double-clicking in order to edit might be a little too much. 

For example, Preview from macOS also allows dragging the entire text element. But editing text only requires one click, once the text is focused.

Though that's a requirement from #2783, and this PR applies it well.

editing-here

swissspidy:
> In my opinion, requiring a double-click to edit might be a little too much.

Agreed. When the block is already selected, just requiring a single click would be nice. Not sure if that causes conflicts with dragging though.

miina:
> Agreed. When the block is already selected, just requiring a single click would be nice. Not sure if that causes conflicts with dragging though.

Will try it out.

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -146,30 +212,77 @@ class TextBlockEdit extends Component {
...
+								className="is-not-editing editor-rich-text block-editor-rich-text wp-block-amp-story-text"
+								onClick={ () => {
+									if ( isSelected ) {
+										this.toggleIsEditing( true );
+									}
+								} }
+								onKeyDown={ ( e ) => {
+									if ( ENTER === e.keyCode && isSelected ) {
+										this.toggleIsEditing( true );
+									}
```
miina: This isn't really working.

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -146,30 +212,77 @@ class TextBlockEdit extends Component {
...
+								{ hasOverlay && (
+									<div
+										role="textbox"
+										tabIndex="-1"
+										className="amp-overlay"
+										onClick={ ( e ) => {
+											this.toggleOverlay( false );
+											e.stopPropagation();
+										} }
+										onKeyDown={ () => {} }
```
miina: That's here for linting purposes only, is there anything useful we could actually add here? `ENTER` could be useful, however, it doesn't really seem to be working.

miina:
@kienstra & @swissspidy Made some changes, could you test if it feels easier to use now?

miina:
Note: will add e2e once the behavior has been confirmed to be OK.

kienstra:
**Editing Looks Good**

Hi @miina,
The editing experience looks better. It's nice how you allow toggling the `isEditing` state on a single click once the block is selected, not a double click:

editing-now

From my perspective, this is ready, other than your 2 points above on `ENTER`.

swissspidy:
When I select a text block via the block navigation, I still have to click twice to edit the text. However, when click on a block directly to select it, I only have to click once to edit the text. Makes sense from a technical POV, just not sure if it's intuitive. But maybe it's just me.

assets/src/stories-editor/blocks/amp-story-text/edit.js
swissspidy: Is the callback even called? Or is the condition not working? Or the toggling? Need `event.preventDefault();`?

assets/src/stories-editor/blocks/amp-story-text/edit.js
swissspidy: Which linter rule is causing this then? An a11y one?

Working `ENTER` support would be nice.

Also: `ESC` to remove editing mode again

assets/src/stories-editor/blocks/amp-story-text/edit.js
miina: It's not even called.

I read that it might be for inputs mainly or when `contenteditable` is true, however, it does look like there are some places in Gutenberg where these are used without these conditions. Not sure.

We can't set `contenteditable` either since then the overlay would lose its purpose.

Any thoughts?
Added the eslint exception for now, the rule is `jsx-a11y/click-events-have-key-events`.

miina:
> When I select a text block via the block navigation, I still have to click twice to edit the text. However, when click on a block directly to select it, I only have to click once to edit the text. Makes sense from a technical POV, just not sure if it's intuitive. But maybe it's just me.

This actually feels more consistent experience at this moment, to me at least. Perhaps we can leave it as it is for now and reconsider if more similar feedback should come? Should be quite simple to change if needed.

tests/e2e/specs/stories-editor/text-block.test.js
```diff
@@ -0,0 +1,56 @@
...
+		await page.click( `.${ textBlockClass }` );
+		await page.keyboard.type( textToWrite );
+		const content = await page.$eval( '.block-editor-block-list__layout .block-editor-block-list__block .wp-block-amp-amp-story-text', ( node ) => node.textContent );
+		expect( content ).toStrictEqual( textToWrite );
+	} );
+
+	// @todo Dragging is not working for some reason.
+	it.skip( 'should allow dragging the Text block from anywhere when not editing', async () => { // eslint-disable-line jest/no-disabled-tests
+		const textBlock = await page.$( '.wp-block-amp-story-text-wrapper' );
+		await dragAndResize( textBlock, { x: 0, y: 25 } );
```
miina: Can't get the dragging to work, out of ideas for this moment. Perhaps you have any ideas, @swissspidy?
If not then maybe we could create a separate issue for investigating dragging + e2e.

swissspidy: Have never tested dragging with Puppeteer before, but I can have a look.

swissspidy: There's an overlay over the page block by default which might cause problems (`.block-editor-inner-blocks.has-overlay::after).

In my testing the text block move once I interacted with the page while debugging the test. So _something_ happens.

miina: I also saw something happening but what made it hard to debug was that sometimes nothing happened and sometimes something happened. For example, once a new tab opened with the block attributes displayed in the search bar (e.g. `{"type":"block","srcIndex":0,"srcRootClientId":"","srcClientId":"6d9d675d-768c-4e27-85a9-d3520739e243"}` 🤷‍♀

miina: > There's an overlay over the page block by default which might cause problems (`.block-editor-inner-blocks.has-overlay::after).

In theory, clicking on the block to make sure that the block is selected first should get rid of the overlay. But perhaps it's not happening as it should.

It could be worthwhile trying if the dragging works on a regular post, although that might be not easy to do / debug since the dropzone is quite narrow.

swissspidy: > For example, once a new tab opened with the block attributes displayed in the search bar

Had that too once or twice. Maybe an issue in the upstream e2e utils?

---

I'll try to look at it again tonight to see if I can find out more.

swissspidy: No luck so far. Will indeed have to try with something else for starters (reorder panel, regular block mover in the block editor, etc.).

Maybe I am missing something, but `dragAndResize` should work in theory (despite the name)

miina: It should work for both dragging and resizing (since that's also dragging basically). Not sure why it's not working. It's used for e2e tests upstream as well, however, not 100% sure if it actually works there either since that test is just using the Snapshots and snapshots assume that the test works OK in the first place.

Maybe we should create a separate issue for figuring out the dragging? This will be necessary for other e2e tests as well, such as the snap + rotation; resizing, reordering, etc.

Thoughts?

swissspidy: The upstream snapshots looked OK though when I checked these today.

Will give it one more try, then we can create a new issue.

miina: Just checked that the shapshot does seem to look OK indeed.

swissspidy: Opened a new issue to explore this: #2975.

assets/src/stories-editor/blocks/amp-story-text/edit.js
swissspidy: ^ @miina thoughts on supporting this? Would be good to have keyboard support here.

assets/src/stories-editor/blocks/amp-story-text/edit.js
swissspidy: I think you need a `tabindex` in order to be able to listen to `onKeyDown` on a `div`.

The `tabindex` could only be added when the block is selected.

miina: It actually does have a `tabindex` already. It had `tabindex` `0` before but since the key events didn't seem to be working anyway then set it to `-1`.

Also tried removing the `BlockMover` to see if that might be interfering with the events but still didn't work.

Also tested that `onKeyDown` within `block.js` is not triggered at all, it might be related to the issue here: https://github.com/ampproject/amp-wp/issues/2787#issuecomment-516988322

Thoughts?

assets/src/stories-editor/blocks/amp-story-text/edit.js
miina: Same issue here as above -- the events are not triggered.

assets/src/stories-editor/blocks/amp-story-text/edit.js
miina: Just tested it again and discovered that in some cases it does work. Haven't understood yet in which cases it works and which not. 

I'll implement the keydown events for the working times 👍

assets/src/stories-editor/blocks/amp-story-text/edit.js
miina: Added `ENTER` support, `ESC` support wouldn't make sense here since this element is not displayed in editing mode at all. `ESCAPE` support for `RichText` doesn't work either since it already has some kind of handling for that.

swissspidy:
Just merged #2824, so this needs some updates.

That PR mentions:

> Update the logic once #2863 gets merged.

miina:
Looks like `core-editor.test.js` was failing due to `switchEditorModeTo` not waiting for the selector being available.

At least now it works locally, hopefully, fixes it here, too.

miina:
@swissspidy Mind giving it another review?

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -146,30 +211,41 @@ class TextBlockEdit extends Component {
...
-						className={ classnames( className, {
-							'has-text-color': textColor.color,
-							[ textColor.class ]: textColor.class,
-							[ fontSize.class ]: ampFitText ? undefined : fontSize.class,
-							'is-amp-fit-text': ampFitText,
-						} ) }
-						placeholder={ placeholder || __( 'Write text…', 'amp' ) }
-					/>
+					{ isEditing &&
+						<RichText
```
miina: Maybe it would make sense to move this to `DraggableText` as well actually.

miina: Could rename it to `DraggableRichText`. Would make a more complete component. Thoughts?

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -67,6 +81,20 @@ class TextBlockEdit extends Component {
 		if ( checkBlockDimensions ) {
 			maybeUpdateBlockDimensions( this.props );
 		}
+
+		// If the state changed to editing, focus on the text.
+		if ( this.state.isEditing && ! prevState.isEditing ) {
+			const textInput = document.querySelector( '.is-selected .wp-block-amp-amp-story-text' );
+			if ( textInput ) {
```
miina: Just realized that this could also be a helper function, it's a duplicate with the CTA block.

assets/src/stories-editor/blocks/amp-story-text/edit.js
miina: Although maybe it would make the component confusing with that many attributes.

miina:
Not sure why the CTA warning message test is failing here, not happening locally. Looking into this ...

miina:
Looks like the tests are OK now finally.

assets/src/stories-editor/blocks/amp-story-text/edit.js
swissspidy: I think for now we can keep it as-is.

PR #2871, opened 2019-07-24
Fix issue with column block

### LINKED ISSUE #2870: Conflict With Gutenberg Columns Block
### Description Of Problem

With Gutenberg plugin (6.1.1) installed and AMP in **Transitional** or **Standard** mode, the Gutenberg columns block is unusable — when a column's block picker is invoked, it appears and then closes immediately.

Problem has been observed in:
- Chrome (Version 75.0.3770.100 (Official Build) (64-bit))
- Firefox (68.0.1)
- Safari (Version 12.1.1 (14607.2.6.1.1)).

### Video

https://cloudup.com/cjq9BdDF5TN

### Steps to Reproduce

1) Spin up a clean WordPress instance (e.g. using https://jurassic.ninja/)
2) Install Gutenberg plugin (6.1.1)
3) Create a post, add columns block, select "Two Columns, Equally Split" layout. 
4) Verify blocks clicking the + icons in either column correctly displays the block picker, and blocks can be selected normally.
5) Install AMP plugin and set it to "Reader Mode" (it will default to this)
6) Return to post and verify that column block continues to work normally. 
7) Switch AMP to "Transitional Mode" or "Standard Mode."
8) Return to post and attempt to add a block to the Columns block. Observe that the block picker appears and disappears immediately, which makes it unusable.

swissspidy:
This changes the `withSelect` call in the `withValidationErrorNotice` HOC to not return empty arrays.

Since it seems to generate a new array (reference) every time, strict equal fails, triggering a new component update.

To avoid this, return `undefined` when there are no block validation errors anyway.

Kind of related: https://github.com/WordPress/gutenberg/issues/16612

Fixes #2870.

### DIFF
--- a/assets/src/block-validation/components/higher-order/with-validation-error-notice.js
+++ b/assets/src/block-validation/components/higher-order/with-validation-error-notice.js
@@ -14,8 +14,10 @@ import { ValidationErrorMessage } from '../';
 const applyWithSelect = withSelect( ( select, { clientId } ) => {
 	const { getBlockValidationErrors } = select( 'amp/block-validation' );
 
+	const blockValidationErrors = getBlockValidationErrors( clientId );
+
 	return {
-		blockValidationErrors: getBlockValidationErrors( clientId ),
+		blockValidationErrors: blockValidationErrors.length ? blockValidationErrors : undefined,
 	};
 } );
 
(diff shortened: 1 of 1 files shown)


swissspidy:
**Testing instructions**

1. Create new regular post
2. Add columns block
3. Click on inserter/appender to add a new paragraph to the column
4. See that it works as expected, and the appender is not suddenly closing or anything

PR #2874, opened 2019-07-24
Add E2E tests for video poster image extraction

swissspidy:
**To do**

* [ ] Consider running tests on Chrome all the time and how to achieve that.


tests/e2e/specs/stories-editor/video-poster-image-extraction.test.js
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * WordPress dependencies
+ */
+import { createNewPost, getAllBlocks, selectBlockByClientId } from '@wordpress/e2e-test-utils';
+
+/**
+ * Internal dependencies
+ */
+import { activateExperience, deactivateExperience } from '../../utils';
```
miina: Maybe we could have just `toggleExperience` instead? E.g. like `toggleScreenOption` in Gutenberg.

Then we could do just `if ( isChecked !== shouldBeChecked ) {` check to see if click or not.

swissspidy:
So the test would theoretically be correct, but after uploading the video there seems to be an issue with media playback.

Turns out, Chromium does not support the proprietary MP4 codecs:

https://github.com/GoogleChrome/puppeteer#q-what-features-does-puppeteer-not-support

So... basically we cannot test this.

tests/e2e/specs/stories-editor/video-poster-image-extraction.test.js
swissspidy: I see that you did this in #2824.

For me it felt more explicit to have two separate methods, like `activatePlugin` and `deactivatePlugin`.

Don't have a strong opinion about it though. 🤷‍♂

swissspidy:
I briefly changed the Puppeteer config like this:

```diff
module.exports = {
	launch: {
		dumpio: true,
		headless: process.env.HEADLESS !== 'false',
+		executablePath: '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome',
	},
	browserContext: 'default',
};
```

And it worked!

Travis CI seems to [support Google Chrome](https://docs.travis-ci.com/user/chrome) too, so maybe we could just always test against Chrome 🤷‍♂ 
Worth noting though that Puppeteer officially does not support anything other than Chromium.

tests/e2e/specs/stories-editor/video-poster-image-extraction.test.js
miina: For the plugin (de)activation two separate methods seem to make more sense to me as well.

Anyway -- we can also keep it separately, I also don't have a strong opinion about it, was mainly following how it was in Gutenberg for the settings.

swissspidy:
As for the Google Chrome support, I am not sure how feasible it is, and thus recommend exploring that separately.

PR #2879, opened 2019-07-24
Restrict the templates to only the AMP Story editor

### LINKED ISSUE #2873: Story templates queried on regular post
Steps to reproduce:
- Create a new post (doesn't happen with existing post).
- Start typing `/template` into the existing paragraph block.
- See the console.

AC: The story templates should be queried only in the context of an AMP Story post.

kienstra:
As Miina pointed out in #2873, templates had appeared in the editors on creating new posts:

new-post

This PR restricts templates to **only** the AMP story editor. **Are there any other contexts that need to query for templates?**

Maybe this isn't the best approach.

The previous approach of [filter_pre_get_posts()](https://github.com/ampproject/amp-wp/blob/7feaf9648a3cd23f7565f2c7a098b1fd87766076/includes/admin/class-amp-story-templates.php#L73), as I understand it, was to blacklist the template from appearing in the reusable block editor:

reusable-block-editor

I think that's what this conditional checks for:
https://github.com/ampproject/amp-wp/blob/7feaf9648a3cd23f7565f2c7a098b1fd87766076/includes/admin/class-amp-story-templates.php#L76

But this PR is more of a **whitelist** approach. It blocks the templates in any query outside of the AMP Story editor for new or existing stories.

It's possible to simply blacklist non-AMP stories posts. But that would probably make [this conditional](https://github.com/ampproject/amp-wp/compare/fix/template-appearing-post?expand=1#diff-00dc6914ea2368cc5f87b97dc2c4f636R82) even more complex:

```php
if ( isset( $referer['query'] ) ) {
	if ( 'post_type=' . AMP_Story_Post_Type::POST_TYPE_SLUG === $referer['query'] ) {
	...
```

Closes #2873


kienstra:
**Request For Review**

Hi @miina and @swissspidy,

It's late, so no pressure to do this now.

Thanks!

includes/admin/class-amp-story-templates.php
```diff
@@ -65,20 +65,28 @@ public function filter_user_has_cap( $allcaps, $caps, $args ) {
...
-		if ( 'edit.php' !== $pagenow || ! $query->is_admin ) {
+		if ( 'wp_block' !== $query->get( 'post_type' ) ) {
 			return $query;
 		}
 
-		if ( 'wp_block' !== $query->get( 'post_type' ) ) {
-			return $query;
+		$referer = wp_parse_url( wp_get_referer() );
+		if ( isset( $referer['query'] ) ) {
+			if ( 'post_type=' . AMP_Story_Post_Type::POST_TYPE_SLUG === $referer['query'] ) {
```
swissspidy: I'd use `$parsed_args = wp_parse_args( $referer['query'] );` already here, and then compare with `$parsed_args['post_type']` directly. Otherwise this check fails when some middleware adds additional query params to the request.

swissspidy:
Might be a simple thing to add an E2E test for while we're at it.

@miina Any objections to merging #2874 (despite the currently skipped tests) so that we can use those new E2E utility functions here?

miina:
> @miina Any objections to merging #2874 (despite the currently skipped tests) so that we can use those new E2E utility functions here?

Nope, that's fine. An alternative could be merging only the utility functions in a separate PR and leaving the skipped tests into the open PR to investigate further.

But it's OK either way.

swissspidy:
Found one issue during review that might have been slipped through if it weren't for e2e tests:

The story templates were all visible in the block editor when the stories experience is turned off. That's because the code for the templates wasn't run at all in that case.

I fixed that now and added some e2e tests for these scenarios. So IMO that's good to go 👍 

However, there's one test that is currently skipped because it throws Puppeteer off. The reason seems to be the click action in the `openTemplateInserter()` utility.

@miina Perhaps you have an idea there?

miina:
@googlebot I consent.

miina:
> I fixed that now and added some e2e tests for these scenarios

Awesome, good catch!

The e2e tests should be OK now, looks like the templates had an issue so another proof in the same PR that we need e2e tests!

swissspidy:
Awesome, thanks a lot!

PR #2906, opened 2019-07-26
Fix setting animation order.

### LINKED ISSUE #2880: "Begin after" setting not working when setting for an unsaved Story
Steps to reproduce:
- Add a block, set animation.
- Add another block, set animation "Begin after", and choose the first block.
- Save draft now and refresh.
- Check the "Begin After" setting of the second block -- it's "Immediately" instead of the previously set value.

Note: when doing this with a Story that was previously already saved then it works.

### LINKED ISSUE #2886: E2E Tests: Animation Order
#2880 makes it clear that we need an E2E test that verifies the integrity of the animation order picker.

### LINKED ISSUE #2923: Custom HTML block causing validation error
To reproduce:
- Add Custom HTML block
- Add any content, e.g. "Hello World"
- Save
- Refresh

AC: Custom HTML block should either work as expected or not be visible in the Inserter.

miina:
Fixes #2880.
Closes #2886 
Fixes #2923.

Todo:
- [x] Fix setting `ampAnimationAfter`.
- [x] Fix templates after changed markup.
- [x] Deprecate and migrate all blocks allowed in AMP Stories which have wrapper set (core and custom).
- [x] Moves `wrapBlocksInGridLayer` filter to PHP to avoid similar issues in the future.
- [x] E2E tests.


assets/src/stories-editor/helpers/index.js
```diff
@@ -1354,10 +1368,8 @@ export const maybeInitializeAnimations = () => {
 			const page = getBlockRootClientId( block.clientId );
 
 			if ( page ) {
-				const { ampAnimationType, ampAnimationAfter, ampAnimationDuration, ampAnimationDelay } = block.attributes;
-				const predecessor = allBlocks.find( ( b ) => b.attributes.anchor === ampAnimationAfter );
-
-				addAnimation( page, block.clientId, predecessor ? predecessor.clientId : undefined );
```
miina: @swissspidy FYI: the issue seemed to come from the animation of predecessor not being initialized before adding the animation -- this made `isValidAnimationPredecessor` always fail. Well, not always, depending on the order of the blocks.

Haven't tested thoroughly yet but seems to work now.

assets/src/stories-editor/helpers/index.js
```diff
@@ -1343,6 +1334,29 @@ export const maybeSetTagName = ( clientId ) => {
...
+	const { ampAnimationAfter } = block.attributes;
+	const predecessor = allBlocks.find( ( b ) => b.attributes.anchor === ampAnimationAfter );
+	if ( predecessor ) {
+		const animations = getAnimatedBlocks();
+		const pageAnimationOrder = animations[ page ] || [];
+		const predecessorEntry = pageAnimationOrder.find( ( { id } ) => id === predecessor.clientId );
+
+		// We need to initialize the predecessor first.
+		if ( ! predecessorEntry ) {
+			initializeAnimation( predecessor, page, allBlocks );
```
miina: Note: should make sure that this won't cause a cycle.

miina:
Was testing with deprecations and there's an issue with it -- we're making the changes in filters and not in the `save` function itself -- the filters are used for deprecations as well so looks like the usual deprecating wouldn't work for this case at all.

@swissspidy Do you have any thoughts here? If not then I'll also ask in core-editor as well to see if that's something that has been considered.

Saw this deprecation-related issue: https://github.com/WordPress/gutenberg/issues/7604

miina:
Might have found an option:

We could assign an extra attribute to the deprecated block, e.g. `deprecated: 'v1.2'` and then, inside the filter, check if this setting exists and then run the previous filter logic according to the version number.

Nor sure exactly yet how the migration would work in this case, will have to test.

assets/src/stories-editor/components/story-controls/index.js
```diff
@@ -48,7 +48,7 @@ function StoryControls( { isReordering, startReordering, saveOrder, resetOrder }
 
 	return (
 		<>
-			<TemplateInserter />
+			{ /* <TemplateInserter /> */ }
```
miina: Commented out for debugging purposes.

assets/src/stories-editor/deprecations/filters/index.js
```diff
@@ -0,0 +1,3 @@
+export { default as addAMPExtraPropsDeprecations } from './add-amp-extra-props';
+export { default as wrapBlockInGridLayerDeprecations } from './wrap-blocks-in-grid-layer';
+export { default as addAMPAttributesDeprecations } from './add-amp-attributes';
```
miina: ^ The changes in these filter files aren't actually that big, however, following how the deprecation is done in Gutenberg and also for not making the logic very confusing with different conditions within `helpers.js` file filters, thought that perhaps it makes sense to copy the full filter as it was. Not sure if that's the best approach though, thoughts?

miina:
@swissspidy Would have time for a quick sanity check for the deprecations before I continue?

assets/src/stories-editor/deprecations/filters/index.js
miina: Also, applying only some changes conditionally in the filter would mean that the future changes might influence the deprecation logic, too. This would need testing if it's OK in case there's a future deprecation for example.

Anyway -- reducing the repeated code here would definitely be a place for improvement if we can do it considering potential future deprecations + not influencing the deprecation logic with future changes.

miina:
@kienstra If you have time, would you mind testing if the PR seems to work as expected?
- On this branch, setting the animation order should work (displaying animation for one block once the other finishes should work as expected in the FE and also display correct values in the editor)
- Block deprecations should work, the best way to test this is to check out `develop` branch, create some blocks with different settings, save, check out this PR, and then refresh the page -- there should be no validation errors in the editor (errors in console are expected).

kienstra:
>@kienstra If you have time, would you mind testing if the PR seems to work as expected?

Sure, I'll test this!

kienstra:
**Animation Order**
*Works As Expected Locally*

animation-order-here

kienstra:
**Deprecation Worked Well Locally**

Hi @miina,
The deprecation seemed to work well for the Core blocks I tested:
* Image
* Code
* Embed
* List
* Preformatted
* Quote
* Pullquote

I set an animation for each of them, and an animation order for many. 

dep-test

There was a validation error for the Custom HTML block, but that doesn't look to be related to this PR. It also appeared when adding the block with the `develop` branch, and reloading using the `develop` branch.

miina:
Thank you for testing, @kienstra!

The HTML block issue has a separate ticket: #2923.

assets/src/stories-editor/helpers/index.js
```diff
@@ -1582,3 +1607,18 @@ export const uploadVideoFrame = async ( { id: videoId, src } ) => {
 		} );
 	} );
 };
+
+/**
+ * Add anchor for a block if it's missing.
+ *
+ * @param {string} clientId Block ID.
+ */
+export const maybeAddMissingAnchor = ( clientId ) => {
```
swissspidy: Isn't that a duplicate of `conditionallyUpdateAnchor`? I think we should remove all usage of `conditionallyUpdateAnchor` with this new function as it has a better name :-)

swissspidy: Actually that does something different. 🤦‍♂ 

Nevermind.

miina:
@kienstra When you have time, would you mind testing/reviewing again? It has had some changes meanwhile (such as moving one of the filters to PHP), would be good to test again.

A quick way how to add some blocks for testing is using templates, too.

kienstra:
>@kienstra When you have time, would you mind testing/reviewing again? It has had some changes meanwhile (such as moving one of the filters to PHP), would be good to test again.

Sure, sorry for the delay. I'm testing this now.

kienstra:
**Approved**

Hi @miina,
This worked well. When I added the following blocks with the `develop` branch, and reloaded the page with this PR's branch, there was no validation error:

* Image
* Code
* Embed
* HTML
* Pullquote
* Quote
* Table
* Verse
* Video

testing-deprecations

I added animations to most of the blocks.

The Custom HTML block didn't have the same content when reloading the page on the `fix/2880-animations` branch. 

Though this is an improvement, as it didn't have a validation error. Before, even when only using the `develop` branch, the block had a validation error on reloading.

When adding the Custom HTML block on the`develop` branch, it had this content:

```
This is a Custom HTML block
```

...and on reloading the page with the `fix/2880-animations` branch, it had this content:

```
<amp-story-grid-layer template="vertical" data-block-name="core/html"><div class="amp-story-block-wrapper" style="position:absolute;top:38.93%;left:5%" animate-in="whoosh-in-right" animate-in-duration="500ms"><div id="a5d971a1-fc6d-45d0-b33c-165c9556043d">This is a Custom HTML block</div></div></amp-story-grid-layer>
```

Like we talked about, and you captured in #2923, the problem with this block was present in the `develop` branch before this PR.

miina:
> ...and on reloading the page with the fix/2880-animations branch, it had this content:

That's an expected change in the content due to moving the filter. Since HTML block was not working before anyway then it shouldn't be an issue either.

miina:
There is some Travis issue happening with one Job right now (have restarted it a few times already but times out), once that gets sorted, let's merge the PR.

I will ask Claudio to prepare a Story which can be used for testing after this gets deployed.

Does that sound OK to you, @swissspidy?

swissspidy:
What's the Travis issue? A failing test or something else?

> I will ask Claudio to prepare a Story which can be used for testing after this gets deployed.

Sounds good to me. Make sure the story is a bit more complex and uses at least one custom HTML block, uses animations, etc.

miina:
> What's the Travis issue? A failing test or something else?

One [e2e Tavis Job](https://travis-ci.org/ampproject/amp-wp/jobs/568778005) wasn't able to set up the environment for testing. Looks like it's finally running now 👍

PR #2913, opened 2019-07-29
Fix specification of text_domain for WordPress.WP.I18n sniff

westonruter:
I found that the `WordPress.WP.I18n` sniff was failing to catch a missing text domain. This resulted in a support forum report: https://wordpress.org/support/topic/two-l10n-issues/

I found the format needs to be as specified on https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#internationalization-setting-your-text-domain

So this fixes the text_domain property and adds missing text domains for translation functions.

### DIFF
--- a/includes/class-amp-story-media.php
+++ b/includes/class-amp-story-media.php
@@ -319,7 +319,7 @@ class AMP_Story_Media {
 			'featured_media',
 			[
 				'schema' => [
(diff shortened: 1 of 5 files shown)


swissspidy:
Ah, the other format was deprecated at some point IIRC.

PR #2919, opened 2019-07-30
Remove deprecated formattingControls.

miina:
`formattingControls` will be deprecated with https://github.com/WordPress/gutenberg/pull/14542.

This PR updates the code in CTA block to match that.

Updates style for `Inline Code` and removes the deprecated param.

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-cta/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-cta/edit.js
@@ -73,7 +73,7 @@ class CallToActionEdit extends Component {
 						placeholder={ __( 'Add text…', 'amp' ) }
 						value={ text }
 						onChange={ ( value ) => setAttributes( { text: value } ) }
-						allowedFormats={ [ 'bold', 'italic', 'strikethrough' ] }
+						allowedFormats={ [ 'core/bold', 'core/italic', 'core/strikethrough' ] }
 						className={ classnames(
 							'amp-block-story-cta__link', {
 								'has-background': backgroundColor.color,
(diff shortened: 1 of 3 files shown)


assets/src/stories-editor/blocks/amp-story-cta/edit.js
```diff
@@ -73,7 +73,7 @@ class CallToActionEdit extends Component {
 						placeholder={ __( 'Add text…', 'amp' ) }
 						value={ text }
 						onChange={ ( value ) => setAttributes( { text: value } ) }
-						formattingControls={ [ 'bold', 'italic', 'strikethrough' ] }
+						allowedFormats={ [ 'core/bold', 'core/italic', 'core/strikethrough' ] }
```
miina: @swissspidy This change would actually remove the Inline Image option from the CTA block button, looks like it wasn't working as expected before and still allowed it. Should we still keep it? Perhaps yes, e.g. if someone wants to add a small icon or sth similar there. Thoughts?

swissspidy: Hmm good question. In the previous issue about the CTA block's inline image support we fixed the styling. So now there's no need to remove support for it, no?

miina: I suppose so, we had this under discussion briefly before when it seemed that it's not removable anyway, so wanted a second opinion first.

Will add the explicit support here then. Or well, by default everything is allowed so will check if there's something that shouldn't be.

miina: Looks like previously everything was allowed.

Adjust the style of Inline Code a bit instead of removing.

swissspidy: Excellent!

kienstra:
**Approved**

Hi @miina,
Good idea to remove the `formattingControls`, with their deprecation in Gutenberg that you mentioned. 

This PR looks good.

PR #2920, opened 2019-07-31
Change dragging and dropping of layers

### LINKED ISSUE #2784: Improve drag & drop reordering of elements in layer panel
As a user I want the drag/drop reordering of elements in the layer panel to be easier, and more intuitive, to use

kienstra:
On dragging layers, this adds a space between them, like in [jQuery UI](https://jqueryui.com/sortable/).

# Before

current-dragging

# After 

dragging-example

Closes #2784

### DIFF
--- a/assets/src/stories-editor/components/block-navigation/edit.css
+++ b/assets/src/stories-editor/components/block-navigation/edit.css
@@ -229,18 +229,27 @@
 	display: none;
 }
 
-.block-editor-block-navigation__list .block-editor-block-navigation__item .components-drop-zone.is-close-to-bottom {
+.block-editor-block-navigation__list li .block-editor-block-navigation__item .components-drop-zone {
 	background: none;
-	border-bottom: 3px solid #0071a1;
 }
 
-.block-editor-block-navigation__list li .block-editor-block-navigation__item .components-drop-zone.is-close-to-top {
-	background: none;
-	border: none;
+.block-editor-block-navigation__list li .block-navigation__placeholder {
+	display: none;
 }
 
-.block-editor-block-navigation__list li:first-child .block-editor-block-navigation__item .components-drop-zone.is-close-to-top {
-	background: none;
-	border-top: 3px solid #0071a1;
-	border-bottom: none;
+/* When dragging near the top of a drop zone, display the placeholder so there is a space between this and the button above. */
+.block-editor-block-navigation__list li .block-editor-block-navigation__item .components-drop-zone.is-close-to-top + .block-navigation__placeholder {
+	display: block;
+	visibility: hidden;
+	height: 36px;
+}
+
+/* Allow dropping below the last <button> by doubling the height of the drop zone. */
+.block-editor-block-navigation__list li:nth-last-child(1) .components-drop-zone:not(.is-close-to-top) {
+	height: 72px;
+}
+
+/* Hide the dragged <button>, as there will be a clone. */
+.block-editor-block-navigation__item-is-dragging > .editor-block-navigation__item-button {
+	display: none;
 }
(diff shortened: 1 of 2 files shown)


assets/src/stories-editor/components/block-navigation/edit.css
```diff
@@ -229,18 +229,27 @@
...
+.block-editor-block-navigation__list li .block-navigation__placeholder {
+	display: none;
 }
 
-.block-editor-block-navigation__list li:first-child .block-editor-block-navigation__item .components-drop-zone.is-close-to-top {
-	background: none;
-	border-top: 3px solid #0071a1;
-	border-bottom: none;
+/* When dragging near the top of a drop zone, display the placeholder so there is a space between this and the button above. */
+.block-editor-block-navigation__list li .block-editor-block-navigation__item .components-drop-zone.is-close-to-top + .block-navigation__placeholder {
```
kienstra: It'd be nice if the  `div.block-navigation__placeholder` wasn't necessary. Maybe there's a way to avoid using it.

Its purpose is to add space between the buttons when dragging.

But when I tried to remove this style rule for the placeholder, and replace it with something like:

```css
.block-editor-block-navigation__list li .block-editor-block-navigation__item .components-drop-zone.is-close-to-top {
	position: static;
	height: 36px;
}
```

...there was a 'toggling' when dragging over the bottom button:

last-drop

kienstra:
**Request For Review**

Hi @swissspidy and @miina,
Could you please review this? 

Thanks, have a great day :smile:

swissspidy:
My first thought when looking at the GIF of the new behavior: this could use some smooth CSS transition.

kienstra:
>My first thought when looking at the GIF of the new behavior: this could use some smooth CSS transition.

Thanks, I'll work on that.

kienstra:
**Transition**

Hi @swissspidy,
What do you think about this transition, applied with https://github.com/ampproject/amp-wp/pull/2920/commits/4853101bf97c2bba89f15e0e73ffd68e28f8b131?
css-tran

assets/src/stories-editor/components/block-navigation/edit.css
miina: Probably the solution with the extra div is OK. If I'm not mistaken then this flickering might come from dragleave firing when moving over containers, however, not sure if this can be handled in a better way. You could look into `onDragLeave` event just in case or search about `dragleave` causing flickering, the placeholder might be a better option though :)

(Not 100% sure that the issue comes from that.)

miina:
I think I'd even set the transition to `0.3s`, I think there are some other transitions where we have `0.3s` set, such as toggling the resizing and rotation handles, that's a different thing though, so no strong feelings about it.

kienstra:
>I think I'd even set the transition to 0.3s, I think there are some other transitions where we have 0.3s set, such as toggling the resizing and rotation handles, that's a different thing though, so no strong feelings about it.

Good idea, `0.3s` looks better:

dragging-transition

assets/src/stories-editor/components/block-navigation/edit.css
kienstra: Good idea to look at the `onDragLeave` possibly causing the 'toggling.' I'll look at that.

kienstra: Good idea to check out the `dragleave` event.

I'm not sure if that's responsible for the 'flickering' above.

As a crude example, disabling it via:

```
document.querySelector( '#amp-story-block-navigation > div > div > ul > li:nth-last-of-type(1) > div > div.components-drop-zone' )
        .addEventListener(
                'dragleave', 
                function( event ) { event.preventDefault();
        }
);
```

...didn't seem to fix this. Though maybe disabling it isn't what you had in mind.

I think it might be caused by my [CSS snippet above](https://github.com/ampproject/amp-wp/pull/2920#discussion_r309025541) not being very good 😄 Though there might not be a way around it.

miina: Thanks for looking into this, it works well as it is, too.

PR #2928, opened 2019-08-01
Try: Move wrapBlockInGridLayer filter to PHP

miina:
This is a potential addition to #2906, this PR,  instead of adding wrappers to the markup, saves attributes to post content and uses those to add wrappers dynamically.

This is a result of discussion with @swissspidy to see if there are ways how to minimize having deprecating all the blocks after a markup change. See the parent PR for seeing what's necessary for deprecating all the blocks due to filter change.


miina:
@westonruter When you have a moment, could you check if you have any concerns with this PR moving adding wrappers around blocks to PHP instead of filtering the blocks in JS as it is done right now?

Note that this PR is based on #2880 which is a quite simple bug fix, in theory, however, due to markup change in a filter needed deprecating _all the blocks_.

This is currently mainly a try to see how it works -- so far it seems to work OK however this kind of change would need extensive testing (as does #2880 later).

We could then try moving `addAMPExtraProps` to PHP as well.

miina:
Note: this would also fix #2923.

includes/class-amp-story-post-type.php
```diff
@@ -1270,6 +1287,122 @@ public static function render_block_with_google_fonts( $block_content, $block )
...
+		}
+
+		if ( isset( $atts['anchor'] ) ) {
+			$wrapper_atts['id'] = $atts['anchor'];
+		}
+
+		$before = "<amp-story-grid-layer template='vertical' data-block-name='$name'>
+			<div class='amp-story-block-wrapper'";
+		foreach ( $wrapper_atts as $att => $value ) {
+			$before .= " $att='$value'";
```
westonruter: The `$value` should be wrapped in `esc_attr()`

includes/class-amp-story-post-type.php
```diff
@@ -1270,6 +1287,122 @@ public static function render_block_with_google_fonts( $block_content, $block )
...
+			$style['height'] = self::get_percentage_from_pixels( 'y', $atts['height'] ) . '%';
+		}
+
+		$wrapper_style = isset( $atts['style'] ) ? $atts['style'] : '';
+
+		foreach ( $style as $att => $value ) {
+			$wrapper_style .= "$att:$value;";
+		}
+
+		$wrapper_atts['style'] = $wrapper_style;
```
westonruter: Should this be conditional based on whether `! empty( $wrapper_atts['style'] )`?

includes/templates/story-templates/fandom-cta.html
```diff
@@ -1,7 +1,7 @@
...
     <!-- wp:amp/amp-story-text {"tagName":"h2","align":"center","customTextColor":"#ffffff","autoFontSize":23,"positionTop":46,"positionLeft":8,"height":94,"width":275} -->
-    <amp-story-grid-layer data-block-name="amp/amp-story-text" template="vertical"><div class="amp-story-block-wrapper" id="8d2f7f12-1f26-4f71-a3d5-d20b7dd20ea0" style="position:absolute;top:46%;left:8%;width:83.84%;height:17%"><h2 style="color:#ffffff;text-align:center;display:flex" class="wp-block-amp-amp-story-text has-text-color"><amp-fit-text layout="flex-item" class="amp-text-content">Check your knowledge in Game of Thrones fan-made Trivia!</amp-fit-text></h2></div></amp-story-grid-layer>
+    <h2 style="color:#ffffff;text-align:center;display:flex" class="wp-block-amp-amp-story-text has-text-color"><amp-fit-text layout="flex-item" class="amp-text-content">Check your knowledge in Game of Thrones fan-made Trivia!</amp-fit-text></h2>
```
westonruter: One thing that gives me pause is that now the blocks are direct children of `amp-story-page`. If this were rendered straight out to the frontend this would be an AMP validation error because the elements need to be inside of `amp-story-grid-layer`. This is more of a theoretical concern because the sanitizers have to run on the markup anyway for it to be valid, so it's not a big deal.

Nevertheless, in the future when it is possible for there to be multiple `InnerBlocks` in a single block, then this would seem like a good future improvement to have two separate `InnerBlocks` for the `amp-story-page`: one `InnerBlocks` which is wrapped by `amp-story-grid-layer` to contain all the blocks like this, and another `InnerBlocks` which is wrapped by `amp-story-cta-layer`.

Cf. #1993.

miina:
> So this is essentially revisiting #2196 (comment)?

Well, yes.

> It should also prevent issues with double wrapping (#2484) as well as fix issues with resolving block validation errors, right?

Yes, we might be able to revert changes from #2484 -- needs testing for previously existing blocks.

includes/templates/story-templates/fandom-cta.html
miina: This probably won't be rendered straight out to the frontend at any moment, otherwise, any dynamic parts (e.g. potential dynamic blocks) wouldn't really display anything.

Worst case scenario we could add the wrappers back in the future and then create deprecation for that logic, however, fairly certain that this won't happen :)

> Nevertheless, in the future when it is possible for there to be multiple InnerBlocks in a single block, then this would seem like a good future improvement

Good point.

miina:
Two things:
1. It looks like it might not make sense to move `addAMPExtraProps` to PHP -- the main addition there is `style` attribute and it's using many JS-based utilities to set the style, for example getting the background color with opacity from the color slug, etc. Not sure how feasible it would be to bring the used utilities to PHP. An alternative could be putting the style together in JS side and then saving it already "ready" as an attribute. This, however, also seems a bit excess since all the style attributes are already saved as params.

Thoughts?

cc @swissspidy @westonruter 

2. I tested what happens if we revert the changes of #2484, it generally works as expected since we're not using the `amp-story-grid-layer` as a wrapper anymore, however, it does not work well in the following case:
- If the user has created a block before this PR and thus has the wrapper saved to DB.
- If this block already includes a validation error from before and is saved with it.
- If now, after updating, the user tried to resolve the block validation errors.

In this case, the block validation error just assumes it's a Custom HTML Block and converts it to that, keeping the grid layer wrapper, etc.

This seems to be quite uncommon case, but in theory, could happen. So looks like we could keep the transformations for now just in case. (We can, however, remove the `data-block-name` from displaying dynamically in the FE.)

Issue #2888, opened 2019-07-26
E2E Tests: Avoid using XPath


swissspidy:
XPath is usually hard to grasp. Let's make our lives easier by using APIs like `toMatchElement` everywhere, like we already do for newer tests.

swissspidy:
Stale, closing

PR #2937, opened 2019-08-05
Hide the Block Manager and add an e2e test for it

### LINKED ISSUE #2904: E2E Tests: Block Manager disabled
Originally disabled in #2180, it seems that the Block Manager option is now visible again.

Let's fix this and add a proper test to verify that the option is not visible.

kienstra:
# Before
block-manager-appears

# After
block-manager

Closes #2904

### DIFF
--- a/tests/e2e/specs/stories-editor/block-manager-disabled.test.js
+++ b/tests/e2e/specs/stories-editor/block-manager-disabled.test.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { createNewPost, getAllBlocks, selectBlockByClientId } from '@wordpress/e2e-test-utils';
+import { createNewPost } from '@wordpress/e2e-test-utils';
 
 /**
  * Internal dependencies
@@ -20,14 +20,11 @@ describe( 'Block Manager', () => {
 	it( 'the button should not appear, due to styling of display: none', async () => {
 		await createNewPost( { postType: 'amp_story' } );
 
-		// Select the page block.
-		await selectBlockByClientId(
-			( await getAllBlocks() )[ 0 ].clientId
-		);
-
 		// Click the 'More tools & options' button to expand the popover.
-		await page.click( '.components-button.components-icon-button.components-dropdown-menu__toggle' );
+		await page.click( '.edit-post-more-menu .components-dropdown-menu__toggle' );
 
-		await expect( page ).not.toMatch( 'Block Manager' );
+		// The text 'Block Manager' should not be present.
+		const blockManagerButton = await page.waitForXPath( '//button[contains(text(), "Block Manager")]' );
+		expect( await blockManagerButton.isIntersectingViewport() ).toStrictEqual( false );
 	} );
 } );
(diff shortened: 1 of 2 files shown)


kienstra:
**Request For Review**

Hi @swissspidy,
Hope you had a great weekend.

Thanks, Pascal!

tests/e2e/specs/stories-editor/block-manager-disabled.test.js
```diff
@@ -0,0 +1,33 @@
...
+
+	afterAll( async () => {
+		await deactivateExperience( 'stories' );
+	} );
+
+	it( 'the button should not appear, due to styling of display: none', async () => {
+		await createNewPost( { postType: 'amp_story' } );
+
+		// Select the page block.
+		await selectBlockByClientId(
```
swissspidy: Why is that needed? The 'More tools & options' button should work without selecting a block.

kienstra: Good question. The test doesn't pass without that, but I'll look for alternatives.

swissspidy: `npm run test:e2e:interactive` helps with debugging.

In a manual quick test in the browser I did not need to select a block for this to work.

Perhaps you just need an `await page.waitForSelector(...)` here instead.

swissspidy: Actually, there's a nice `clickBlockToolbarButton` utility in Gutenberg that can be used for this.

Example use case: `await clickBlockToolbarButton( 'More options' );`

kienstra: Thanks, `npm run test:e2e:interactive` was really helpful! 

Thanks for bringing up `clickBlockToolbarButton()`. It looks like [it queries inside](https://github.com/WordPress/gutenberg/blob/2f757fa8012edb5cd684a26adef9cb4e8c2158db/packages/e2e-test-utils/src/click-block-toolbar-button.js#L7) `'.block-editor-block-toolbar'`, and didn't seem to find the right button.

https://github.com/ampproject/amp-wp/pull/2937/commits/79e5d057ba6f610fcbfe84aa4cbf17de3730272c removes the `selectBlockByClientId()` call. 

I think the problem here was that the previous `toMatch()` call simply looked for the text of `'Block Manager'` in the DOM, whether it was visible or not.

So this uses [isIntersectingViewport()](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#elementhandleisintersectingviewport). 

It also uses XPath, though I know [you prefer to avoid it](https://github.com/ampproject/amp-wp/issues/2888). This is to target the text of `'Block Manager'` inside the `<button>`. Otherwise, if this just targeted the CSS selector of the `<button>`, the markup might change, and this test wouldn't catch if the text of 'Block Manager' displayed.

swissspidy: Ugh, sorry, I meant `clickOnMoreMenuItem`.

kienstra: Oh, all good.

PR #2960, opened 2019-08-05
Remove unused `eslint-disable` directives

swissspidy:
Verified by running `npm run lint:js -- --report-unused-disable-directives` before and after making these changes.


swissspidy:
Looks like [`eslint-plugin-eslint-comments`](https://www.npmjs.com/package/eslint-plugin-eslint-comments) could be used that does the same thing as part of the regular linting (and more). See https://mysticatea.github.io/eslint-plugin-eslint-comments/

PR #2967, opened 2019-08-05
First pass at rotation snapping

### LINKED ISSUE #2876: Allow users to see blocks snap to position (vertical and horizontal) when rotating.
As a user I want a block to snap at quarter angles when I am rotating it.

AC1: When rotating a block, at each quarter angle it should snap into place to allow the user to achieve perfect alignment with the angle.

swissspidy:
Fixes #2876.

Demo: https://cloudup.com/c2FrGxfBbr0

Tomorrow I'll attempt to add e2e tests for it if possible.


assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
```diff
@@ -307,6 +309,8 @@ export default createHigherOrderComponent(
 
 								stopBlockActions();
 							} }
+							snap={ [ -165, -135, -120, -105, -90, -75, -60, -45, -30, -15, 0, 15, 30, 45, 60, 75, 90, 105, 120, 135, 150, 165, 180 ] }
```
miina: Could we create a constant for this, or move this to `constants.js`? Seems to be identical for both lines where used.

Or potentially even create a function which given the difference of the degrees (e.g. 15deg as it is here) returns an array. In case we'd want to change it to 10 or 20 for example. Although that's likely an overkill until we actually need it :)

swissspidy: I was thinking about passing a function as well.

Both options work for me though.

assets/src/stories-editor/components/rotatable-box/index.js
```diff
@@ -107,6 +108,28 @@ class RotatableBox extends Component {
...
+	 * Given a rotation angle, finds the closest angle to snap to.
+	 *
+	 * Inspired by the implementation in re-resizable.
+	 *
+	 * @see https://github.com/bokuweb/re-resizable
+	 *
+	 * @param {number} angle
+	 * @return {number} New angle.
+	 */
+	findClosestSnap( angle ) {
```
swissspidy: @miina should we perhaps use `memize` here to cache results?

miina:
Works well, makes things much easier!

One oddity: when the block is rotated near 180 degrees then for some reason rotating it on the right side, the first 2-3 "snapping points" aren't snapping right away, that's the only place where it happens. It's not a blocker or not necessarily a bug / noticeable difference at all, just wondering if that's expected.

For example:
snapping

assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
miina: We might later consider having a different degree step for tiny blocks versus larger blocks for example.

assets/src/stories-editor/components/rotatable-box/index.js
miina: I'm not certain. Not sure if we'd need to use the same angle frequently enough / if this calculation is heavy enough for it being worthwhile. WDYT? Or could we move this function out of this component to helpers so that all the components could use the same memorized values? Still not sure. WDYT?

swissspidy:
> One oddity: when the block is rotated near 180 degrees then for some reason rotating it on the right side, the first 2-3 "snapping points" aren't snapping right away, that's the only place where it happens.

Looks like I missed one snapping point at -150.

assets/src/stories-editor/components/higher-order/with-amp-story-settings.js
swissspidy: Added constants for these.

assets/src/stories-editor/components/rotatable-box/index.js
swissspidy: I think it cannot hurt to optimize this. `re-resizable` also memoizes the function.

Other components likely have different snap targets though, so it's mostly useful on a per-component basis.

And yes, extracting makes sense. Did that now.

assets/src/stories-editor/constants.js
```diff
@@ -239,3 +239,6 @@ export const AMP_STORY_FONT_IMAGES = {
 
 export const TEXT_BLOCK_BORDER = 5;
 export const TEXT_BLOCK_PADDING = 7;
+
+export const BLOCK_ROTATION_SNAPS = [ -165, -150, -135, -120, -105, -90, -75, -60, -45, -30, -15, 0, 15, 30, 45, 60, 75, 90, 105, 120, 135, 150, 165, 180 ];
```
miina: Still something missing. It should maybe have `-180` as well to complete the round.

swissspidy: Added in 4d68cfe10a98ae3fafecb59f174d34b7a1913cef. Does that work better now?

kienstra:
**Looks Good!**

Hi @swissspidy,
This works really well:

block-snaps-as-expected

It also makes it much easier to align two diagonal blocks, for example:

easier-align-di

swissspidy:
Opened #2975 to investigate e2e testing for drag & drop.

PR #2972, opened 2019-08-06
Add RTL admin-bar.css and update fork to WP 5.3-alpha@160fc055da

westonruter:
# Add RTL admin-bar

This PR includes an `admin-bar-rtl.css` and ensures it gets loaded when `is_rtl()`.

## Before

Screen Shot 2019-08-06 at 13 58 57

## After

Screen Shot 2019-08-06 at 14 15 01

# Update Forks

The `admin-bar.css` in core has been updated since we forked it from WP 4.9. This updates it to the latest: https://github.com/WordPress/wordpress-develop/blob/160fc055da156248277513dd5256fbcc66faa5a7/src/wp-includes/css/admin-bar.css

Changes applied to fork:

```diff
0a1,10
> /*
> This is forked from core's admin-bar.css from WP 5.3-alpha at 160fc055da156248277513dd5256fbcc66faa5a7.
> 
> https://github.com/WordPress/wordpress-develop/blob/160fc055da156248277513dd5256fbcc66faa5a7/src/wp-includes/css/admin-bar.css
> 
> - References to .hover have been replaced with :focus-within (which is not supported in IE11).
> - Universal selector properties have been removed which interferes with AMP shadow elements.
> */
> 
> 
6c16
< 	position: static;
---
> 	/* Removed because interferes with amp-img>img: position: static; */
187c197
< #wpadminbar .quicklinks .menupop.hover ul li .ab-item,
---
> #wpadminbar .quicklinks .menupop:focus-within ul li .ab-item,
201c211
< #wpadminbar li.hover > .ab-sub-wrapper {
---
> #wpadminbar li:focus-within > .ab-sub-wrapper {
206c216
< #wpadminbar .menupop li.hover > .ab-sub-wrapper {
---
> #wpadminbar .menupop li:focus-within > .ab-sub-wrapper {
212c222
< #wpadminbar .ab-top-secondary .menupop li.hover > .ab-sub-wrapper {
---
> #wpadminbar .ab-top-secondary .menupop li:focus-within > .ab-sub-wrapper {
221c231,235
< #wpadminbar .ab-top-menu > li.hover > .ab-item {
---
> #wpadminbar .ab-top-menu > li:focus-within > .ab-item {
> 	background: #32373c;
> 	color: #00b9eb;
> }
> #wpadminbar .ab-top-menu > li:focus-within > .ab-item {
227c241
< #wpadminbar > #wp-toolbar li.hover span.ab-label,
---
> #wpadminbar > #wp-toolbar li:focus-within span.ab-label,
273c287
< #wpadminbar .quicklinks .menupop.hover ul li a,
---
> #wpadminbar .quicklinks .menupop:focus-within ul li a,
283,287c297,301
< #wpadminbar .quicklinks .ab-sub-wrapper .menupop.hover > a,
< #wpadminbar .quicklinks .menupop.hover ul li a:hover,
< #wpadminbar .quicklinks .menupop.hover ul li a:focus,
< #wpadminbar .quicklinks .menupop.hover ul li div[tabindex]:hover,
< #wpadminbar .quicklinks .menupop.hover ul li div[tabindex]:focus,
---
> #wpadminbar .quicklinks .ab-sub-wrapper .menupop:focus-within > a,
> #wpadminbar .quicklinks .menupop:focus-within ul li a:hover,
> #wpadminbar .quicklinks .menupop:focus-within ul li a:focus,
> #wpadminbar .quicklinks .menupop:focus-within ul li div[tabindex]:hover,
> #wpadminbar .quicklinks .menupop:focus-within ul li div[tabindex]:focus,
295,296c309,310
< #wpadminbar li.hover .ab-icon:before,
< #wpadminbar li.hover .ab-item:before,
---
> #wpadminbar li:focus-within .ab-icon:before,
> #wpadminbar li:focus-within .ab-item:before,
307,308c321,322
< #wpadminbar.mobile .quicklinks .hover .ab-icon:before,
< #wpadminbar.mobile .quicklinks .hover .ab-item:before {
---
> #wpadminbar.mobile .quicklinks :focus-within .ab-icon:before,
> #wpadminbar.mobile .quicklinks :focus-within .ab-item:before {
395c409
< #wpadminbar .ab-top-menu > #wp-admin-bar-recovery-mode.hover >.ab-item,
---
> #wpadminbar .ab-top-menu > #wp-admin-bar-recovery-mode:focus-within >.ab-item,
482c496
< 	width: auto;
---
> 	width: 16px; /* Was auto. */
491c505
< 	display: inline;
---
> 	display: inline-block; /* Was inline. */
496c510
< 	width: auto;
---
> 	width: 16px; /* Was auto. */
532c546
< #wpadminbar .quicklinks .ab-sub-wrapper .menupop.hover > a .blavatar {
---
> #wpadminbar .quicklinks .ab-sub-wrapper .menupop:focus-within > a .blavatar {
855c869
< 	#wpadminbar .quicklinks .menupop.hover ul li .ab-item,
---
> 	#wpadminbar .quicklinks .menupop:focus-within ul li .ab-item,
866c880
< 	#wpadminbar .menupop li.hover > .ab-sub-wrapper {
---
> 	#wpadminbar .menupop li:focus-within > .ab-sub-wrapper {
1060c1074
< 	#wpadminbar li.hover ul li,
---
> 	#wpadminbar li:focus-within ul li,
1109,1111c1123
< 	#wpadminbar {
< 		position: absolute;
< 	}
---
> 	/* Removed: #wpadminbar { position: absolute; } */
```

### DIFF
--- a/assets/css/admin-bar-rtl.css
+++ b/assets/css/admin-bar-rtl.css
@@ -87,7 +87,7 @@ html:lang(he-il) .rtl #wpadminbar * {
 }
 
 #wpadminbar {
-	direction: ltr;
+	direction: rtl;
 	color: #ccc;
 	font-size: 13px;
 	font-weight: 400;
@@ -96,7 +96,7 @@ html:lang(he-il) .rtl #wpadminbar * {
 	height: 32px;
 	position: fixed;
 	top: 0;
-	left: 0;
+	right: 0;
 	width: 100%;
 	min-width: 600px; /* match the min-width of the body in wp-admin/css/common.css */
 	z-index: 99999;
@@ -117,15 +117,15 @@ html:lang(he-il) .rtl #wpadminbar * {
 }
 
 #wpadminbar ul#wp-admin-bar-root-default>li {
-	margin-right: 0;
+	margin-left: 0;
 }
 
 #wpadminbar .quicklinks ul {
-	text-align: left;
+	text-align: right;
 }
 
 #wpadminbar li {
-	float: left;
+	float: right;
 }
 
 #wpadminbar .ab-empty-item {
@@ -133,7 +133,7 @@ html:lang(he-il) .rtl #wpadminbar * {
 }
 
 #wpadminbar .quicklinks .ab-top-secondary > li {
-	float: right;
+	float: left;
 }
 
 #wpadminbar .quicklinks a,
@@ -146,7 +146,7 @@ html:lang(he-il) .rtl #wpadminbar * {
 }
 
 #wpadminbar .quicklinks > ul > li > a {
-	padding: 0 8px 0 7px;
+	padding: 0 7px 0 8px;
 }
 
 #wpadminbar .menupop .ab-sub-wrapper,
@@ -163,7 +163,7 @@ html:lang(he-il) .rtl #wpadminbar * {
 #wpadminbar.ie7 .menupop .ab-sub-wrapper,
 #wpadminbar.ie7 .shortlink-input {
 	top: 32px;
-	left: 0;
+	right: 0;
 }
 
 #wpadminbar .ab-top-menu > .menupop > .ab-sub-wrapper {
@@ -171,8 +171,8 @@ html:lang(he-il) .rtl #wpadminbar * {
 }
 
 #wpadminbar .ab-top-secondary .menupop .ab-sub-wrapper {
-	right: 0;
-	left: auto;
+	left: 0;
+	right: auto;
 }
 
 #wpadminbar .ab-submenu {
@@ -213,15 +213,15 @@ html:lang(he-il) .rtl #wpadminbar * {
 
 #wpadminbar .menupop li:hover > .ab-sub-wrapper,
 #wpadminbar .menupop li:focus-within > .ab-sub-wrapper {
-	margin-left: 100%;
+	margin-right: 100%;
 	margin-top: -32px;
 }
 
 #wpadminbar .ab-top-secondary .menupop li:hover > .ab-sub-wrapper,
 #wpadminbar .ab-top-secondary .menupop li:focus-within > .ab-sub-wrapper {
-	margin-left: 0;
-	left: inherit;
-	right: 100%;
+	margin-right: 0;
+	right: inherit;
+	left: 100%;
 }
 
 #wpadminbar:not(.mobile) .ab-top-menu > li > .ab-item:focus,
@@ -231,6 +231,7 @@ html:lang(he-il) .rtl #wpadminbar * {
 	background: #32373c;
 	color: #00b9eb;
 }
+
 #wpadminbar .ab-top-menu > li:focus-within > .ab-item {
 	background: #32373c;
 	color: #00b9eb;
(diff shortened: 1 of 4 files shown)


swissspidy:
Can we automate this? If I‘m not miataken we already use rtlcss for other styles.

westonruter:
Yeah, we should. I don't know how to configure that, however. Something like this?

```js
// ...
const adminBar = {
	...defaultConfig,
	module: {
		rules: [
			{
				test: /admin-bar\.css$/,
				use: [
					MiniCssExtractPlugin.loader,
					'css-loader',
					'postcss-loader',
				],
			},
		],
	},
	plugins: [
		new MiniCssExtractPlugin( {
			filename: '../css/[name]-compiled.css',
		} ),
		new RtlCssPlugin( {
			filename: '../css/[name]-compiled-rtl.css',
		} ),
		new WebpackBar( {
			name: 'Admin Bar',
			color: '#67b255',
		} ),
	],
};
// ...
```

swissspidy:
At first glance that seems like a step in the right direction 👍 I can have a closer look tomorrow.

westonruter:
Yeah, except it doesn't work 😄

assets/css/admin-bar-rtl.css
```diff
@@ -0,0 +1,1182 @@
...
+#wpadminbar .menupop .menupop > .ab-item {
+	display: block;
+	padding-left: 2em;
+}
+
+#wpadminbar .menupop .menupop > .ab-item .wp-admin-bar-arrow:before {
+	top: 1px;
+	left: 10px;
+	padding: 4px 0;
+	content: "\f141";
```
swissspidy: This should be `content: "\f139";` in RTL.

assets/css/admin-bar-rtl.css
```diff
@@ -0,0 +1,1182 @@
...
+
+#wpadminbar .ab-top-secondary .menupop .menupop > .ab-item {
+	padding-right: 2em;
+	padding-left: 1em;
+}
+
+#wpadminbar .ab-top-secondary .menupop .menupop > .ab-item .wp-admin-bar-arrow:before {
+	top: 1px;
+	right: 6px;
+	content: "\f139";
```
swissspidy: This should be `content: "\f141";` in RTL (opposite of the one in the default CSS)

swissspidy:
As for the webpack config, `entry` and `output` were missing. But even then, I had issues with the CSS loader processing it.

Given that webpack does not yet support standalone CSS processing very well (it always builds a JS chunk), it's probably easiest to just use RTLCSS standalone for this.

Since we need to add RTL versions for all other stylesheets anyway, I suggest exploring this in a separate issue. Then we can also look into using PostCSS for all these stylesheets, which allows us to use things like `postcss-themes` as well.

One thing that I already noticed: In order to not override the wrong files or create too many files, we should separate `assets/css` into a `src` and `dist` folder, like we already have for JS.
Right now we have kind of a mix in this folder (see `*-compiled.css` files).

assets/css/admin-bar-rtl.css
swissspidy: (FWIW I ran `admin-bar.css` through `RTLCSS` locally, and this was changed)

assets/css/admin-bar-rtl.css
swissspidy: (FWIW I ran `admin-bar.css` through `RTLCSS` locally, and this was changed)

assets/css/admin-bar-rtl.css
westonruter: Yeah, I see you're right: https://github.com/WordPress/WordPress/blob/b8fd2b631b890d3a97a803db7b021218e27ae1e8/wp-includes/css/admin-bar-rtl.css#L339-L343

Issue #1017, opened 2018-03-13
Discovery around adding performance profiling/testing


westonruter:
In light of #993 it is clear we need more performance profiling in place for the plugin.

- [ ] Using the scripts used to populate a site with test content, generate a post fully populated with embeds, widgets, and comments, with enough lorem ipsum text to count as a long-form post. Then with the [“AMP Conf” theme](https://github.com/xwp/ampconf) active, capture the source of the page. Then save the output to `tests` directory for use in testing.
- [ ] Run Xdebug profiler and do a sanity check for performance bottlenecks.
- [ ] Add a PHPUnit test for performance, loading up the captured HTML output and making sure the processing of the HTML does not take more than a reasonable amount of time.
- [ ] Ensure that external HTTP requests are mocked to prevent network requests from causing periodic failures. (Aside: Make sure all other requests in tests are also mocked to prevent any external requests from hitting the network when running tests.)
- [ ] Add a new test suite that includes the performance testing and also the slow `runInSeparateProcess` tests. Let this test suite be run with all the others on Travis CI, but exclude it by default on the local environment. Or rather, as in the case of Gutenberg, use environment variable for this same purpose. A separate testsuite/group seems better.

See also #990 for sending back server timings via HTTP response headers.

westonruter:
For one approach, consider: https://github.com/WordPress/gutenberg/blob/master/phpunit/class-performance-test.php

schlessera:
I talked about this with @amedina and suggested to look into the [Blackfire profiler](https://blackfire.io/) I had already mentioned. This has built-in support for CI integration, and although this is only provided in the paid edition, it is probably still cheaper than building such a system from scratch using Xdebug.

archon810:
https://github.com/ampproject/amp-wp/issues/2069#issuecomment-530590423 shows just how bad validation could get. 3s normal WP functions after pressing Update, then a 9s AMP validation.

schlessera:
Added #3361 as a result of this discovery ticket. Closing this for now. If we decide not to proceed with Blackfire, we can reopen again.

PR #2974, opened 2019-08-07
Normalize amp-bind syntax to non-bracketed data-amp-bind-* attributes

### LINKED ISSUE #1162: Utilize new data-amp-bind-* alternative syntax for amp-bind attributes
There is a new XHTML-compatible amp-bind attribute syntax: https://github.com/ampproject/amphtml/issues/11115 via https://github.com/ampproject/amphtml/pull/15408

Instead of:

```html
<span class="foo" [class]="foo ? 'foo' : 'bar'">
```

We can now do:

```html
<span class="foo" data-amp-bind-class="foo ? 'foo' : 'bar'">
```

What this means is that we now have an official way of representing amp-bind attributes in PHP's DOM internals. We can now eliminate `AMP_DOM_Utils::get_amp_bind_placeholder_prefix()` and `AMP_DOM_Utils::restore_amp_bind_attributes()` entirely, and [`AMP_DOM_Utils::convert_amp_bind_attributes()`](https://github.com/Automattic/amp-wp/blob/927b07ca926582fd0a2be585bf8f12a16fae0a49/includes/utils/class-amp-dom-utils.php#L191-L269) can be modified to just convert the bracketed syntax into the `data-amp-bind-*` syntax. There would be no need to restore the bracketed syntax when re-serializing. The result will be improved performance.

westonruter:
Fixes #1162.

The libxml in PHP does not understand the bracketed syntax of bound attributes in AMP-bind. So before loading an HTML document into the DOM, we first have to convert the bracketed syntax into something that is not munged via `\AMP_DOM_Utils::convert_amp_bind_attributes()`. Originally this replacement was a placeholder that was replaced with the bracketed-syntax upon serialization (via `\AMP_DOM_Utils::restore_amp_bind_attributes()`). 

However, AMP now supports an alternative syntax for bound attributes which is XML-compatible: instead of `[foo]` one does `data-amp-bind-foo`. Now that this is equivalent in AMP, there is no need to restore the bracketed syntax upon serialization. We can normalize to the XML-compatible syntax. A couple added benefits to this are:

1. Less processing time doing search/replace.
2. Consumers of the AMP documents don't themselves have to worry about the bracketed syntax (see [support forum topic](https://wordpress.org/support/topic/hambuger-menu-no-opening-on-static-version-of-website/)).
3. We no longer have to handle both syntaxes when doing validation. At the time of spec ingestion, all bracketed-attributes can be converted to `data-amp-bind-*` attributes. 

Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3494197/amp.zip) - v1.2.1-beta1-20190812T202306Z-b41861fe


includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -597,6 +596,8 @@ private function process_node( $node ) {
 					$is_bind_attribute = (
 						'[' === $name[0]
 						||
+						'data-amp-bind-' === substr( $name, 0, 40 )
```
ataylorme: Should this use `AMP_BIND_DATA_ATTR_PREFIX` instead of hard-coding `data-amp-bind-`?

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -597,6 +596,8 @@ private function process_node( $node ) {
 					$is_bind_attribute = (
 						'[' === $name[0]
 						||
+						AMP_DOM_Utils::AMP_BIND_DATA_ATTR_PREFIX === substr( $name, 0, 40 )
```
swissspidy: Shouldn't that be `14` instead of `40`?

westonruter: Yes it should! But it turned out to still pass the unit test because the next condition did pass:

```php
( isset( $this->rev_alternate_attr_name_lookup[ $name ] ) && '[' === $this->rev_alternate_attr_name_lookup[ $name ][0] )
```

Where `$this->rev_alternate_attr_name_lookup` maps all `data-amp-bind-*` attributes to their bracketed versions. All of that is just a waste now, because otherwise that array should just contain:

```json
{
    "srcset": "src",
    "href": "xlink:href"
}
```

But as it stands right now it has 110 other entries for redundant `data-amp-bind-*` mappings. I'm going to eliminate it along with my proposed follow-ups.

swissspidy:
> Less processing time doing search/replace.

Perhaps we could have some dedicated performance tests at some point?

westonruter:
> Perhaps we could have some dedicated performance tests at some point?

Yes. See #1017.

PR #2978, opened 2019-08-07
Video poster image size improvements

### LINKED ISSUE #2881: Extracted video poster image doesn't use correct image size
When adding a video block, it always uses the full size image URL instead of a more suitable one.

Same for the page background video, where we already have a suitable image size.

Originally added in: #903.

swissspidy:
Fixes #2881.

To test:

1. Get a demo video for testing, e.g. https://www.pexels.com/video/blue-sky-with-white-clouds-above-a-body-of-water-1860684/
2. Create a new page
3. Add background media
4. Upload video
5. Notice that poster image is extracted and saved
6. Save and preview story.
7. Poster file name should have suffix `-2560x1440.png`


assets/src/stories-editor/components/custom-video-block-edit.js
```diff
@@ -137,8 +137,8 @@ class CustomVideoBlockEdit extends Component {
 			this.setState( { extractingPoster: true } );
 
 			uploadVideoFrame( { id, src } )
-				.then( ( posterUrl ) => {
-					setAttributes( { poster: posterUrl } );
+				.then( ( fileObj ) => {
+					setAttributes( { poster: getPosterImageFromFileObj( fileObj ) } );
```
miina: Actually... Should this be applied for the Video block as well? Or just the background video?

swissspidy: @miina you wrote this comment on the video block, soo... 🤔

miina: I know :) I wanted to ask with this if this actually should be applied to the Video block as well? In other words, is it intended that the poster image of a video block has the size based on the Page's poster image's recommended size? Shouldn't the poster image be based on the video's size instead?

miina: I guess the goal was to limit the size. So feel free to ignore!

Issue #2732, opened 2019-07-02
More functionality in a text box / text element


vinaymahag:
It would be nice if you could hit enter and start a new line in a text box, and if you could have different fonts (sizes, colors) on the new line.

swissspidy:
#2979 addresses the `Enter` part. For the other part (different font sizes and colors) we could draw some inspiration from https://wordpress.org/plugins/advanced-rich-text-tools/, but it might not be quite straightforward and needs some UX considerations.

swissspidy:
Related: #2982 will add support for changing background color for different texts using the `data-text-background-color` attribute.

PR #2979, opened 2019-08-07
Text block: create new paragraph on enter

swissspidy:
See #2732.

This has come up a few times in user testing, as it's rather annoying that nothing happens when pressing enter.

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-text/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-text/edit.js
@@ -153,6 +153,7 @@ class TextBlockEdit extends Component {
 						// Ensure line breaks are normalised to HTML.
 						value={ content }
 						onChange={ ( nextContent ) => setAttributes( { content: nextContent } ) }
+						multiline={ true }
 						// The 2 following lines are necessary for pasting to work.
 						onReplace={ this.onReplace }
 						onSplit={ () => {} }


kienstra:
**Looks Good**

Hi @swissspidy,
Though I didn't review this this soon enough, this PR works well.

In the Text block, clicking 'enter' now creates a new line:

text-block

swissspidy:
No worries @kienstra, thanks a lot for testing!

miina:
@swissspidy Looks like this change is causing a bug with the Text block: #2997 

Mind if I revert the change for now? Looks like adding `multiline` might need changes in templates and potentially also deprecation (since none of the previously saved text blocks display any content, neither the templates). 

Perhaps we can fix it by reverting the change for now and then look into `multiline` again?

swissspidy:
Huh. Not so cool. Sure, feel free to revert. But would be good to address this soon-ish.

schlessera:
The problem here is not directly related to the templates, but rather to the fact that, when `multiline` is enabled, the `RichText` component expects the content to be one or more HTML tags, (not sure whether it is restricted to paragraph tags). So, RichText then expects `<p>Text field content</p>` (note the paragraph tags), but actually gets `Text field content` (which would be fine for non-multiline `RichText`...), which it does not accept as valid HTML (I assume) and causes the `content` to be empty instead (which in turn causes the placeholder to be displayed).

I can see two immediate solutions:
a) Adapt the templates so they all include the required markup to work properly. This would not take care of any existing stories/templates that are already stored in the database.
b) Add a filtering function to the `content` that is being passed to the `RichText`'s `value` prop, to conditionally wrap the `content` in paragraph texts if they are only untagged text.

I'd prefer the second option, as it transparently fixes the issue while keeping all preexisting persisted content intact. I discussed this with @kienstra, and we agreed that's the better way to go and that we could look into the `RichText` implementation to find out how it defines what is valid or not, and use that exact same code to decide whether to wrap in `<p>` tags.

miina:
Thanks for looking into this. Filtering makes sense to me!

On `a)`: we can adapt the templates as well, independent of the other solution. The templates are stored as posts and have a version number -- if the version number changes, the template posts are updated as well. Updating the templates won't cause any issues for the user nor will break the previous templates, however, it's good to have them up to date with the current logic.

On `b)` -- there is also the `render_block` filter that filters each block separately, that might be easier for filtering.

miina:
For reference, when updating template files (in `includes/templates/story-templates`) then here is the version of the templates which triggers posts update when changed:
https://github.com/ampproject/amp-wp/blob/186c83ee4d8b0fc7d383ae27fd2891c9bb094ec7/includes/admin/class-amp-story-templates.php#L19

schlessera:
@miina Thanks! I will update it in the PR. Additionally, during development, I found you can force an update of the templates through WP-CLI:
```
wp eval '(new AMP_Story_Templates )->import_story_templates();'
```

schlessera:
Trying to fix the templates with `multiline={ true }` results in about half of the templates working, and half of them getting invalidated, as the content is empty.

I'm not 100% but the distinction between these seems to be that some have a `type="p"` attribute while others have a `tagName="p"` attribute. Could this be relevant?

As far as I can tell, `type` is a setting in the AMP Text component, whereas `tagName` is part of the WP RichText component... Wouldn't both be needed all the time?

I guess the main question is: is the template markup inconsistent and represents different versions of the code in time, leading to these problems now? In that case, it would be good to know what an actual reference template markup would currently look like? I guess it would just be what is generated when you create an actual Story page and then look at the markup?

kienstra:
>wp eval '(new AMP_Story_Templates )->import_story_templates();'

Wow, that's great! I didn't know you could do that.

schlessera:
Oh, I think I found the culprit. On `"ampFitText":false`, the content is passed through `RawHtml`, but when `"ampFitText":true`, that's not the case, so all HTML tags are being stripped.

schlessera:
Hmm, I thought it had something to do with https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/blocks/amp-story-text/save.js#L27 , but at that point, the `content` attribute is already empty...

miina:
> I guess the main question is: is the template markup inconsistent and represents different versions of the code in time, leading to these problems now? In that case, it would be good to know what an actual reference template markup would currently look like? I guess it would just be what is generated when you create an actual Story page and then look at the markup?

On templates: the Story template source files are all in the `includes/templates/story-templates` folder -- this includes the base templates as they are displayed in the Template Inserter preview. Based on those files the templates are generated as posts (`wp_block` post type). The templates that get inserted to a Story when clicking on a template are stripped out of most of its attribute values. Not sure if this answers any of the questions? For fixing the templates we'd probably need to update the markup within `includes/templates/story-templates` so that the Text blocks would include `p` tags surrounding the content.

On fixing the previously existing blocks:
We could perhaps check in the `edit.js` of Text block if the [`content`](https://github.com/ampproject/amp-wp/blob/186c83ee4d8b0fc7d383ae27fd2891c9bb094ec7/assets/src/stories-editor/blocks/amp-story-text/edit.js#L104) includes `p` tags and if not then wrap the `content` into `p` tags before [passing it as `value`](https://github.com/ampproject/amp-wp/blob/186c83ee4d8b0fc7d383ae27fd2891c9bb094ec7/assets/src/stories-editor/blocks/amp-story-text/edit.js#L154) to the `RichText`? Or is the `content` attribute empty in `render` in `edit.js` already? If it's empty already then, then we could use the `blocks.getBlockAttributes` filter since this includes the `innerHTML` of the block as well. Although in theory the `edit.js` `render` should still have the `content` attribute set.

schlessera:
I finally figured it out, after coming full-circle...

This commit was breaking the multiline functionality: https://github.com/ampproject/amp-wp/commit/7aa7de5c2fccab669bc03ed24fac4ade1de2a93a

Namely, forcefully setting the `tagName` to `"p"`. As the other part of the commit is not being used anymore, I assume for now that this is jus a left-over...

I'll create a new PR with a fix.

PR #2980, opened 2019-08-07
Decouple post meta box and block validation assets

### LINKED ISSUE #2953: Post Meta Box script erroneously loaded on block editor screen
While looking at #2934 I noticed that `AMP_Post_Meta_Box` enqueues the `amp-post-meta-box` JS & CSS even for the block editor. However, those do only apply for the classic editor.

Also, it looks like `amp-post-meta-box.css` contains some CSS rules specific to the AMP block validation in the block editor. These should be decoupled from that CSS and instead moved to a separate `amp-block-validation.css` stylesheet that complements the existing `amp-block-validation.js` script.

swissspidy:
Fixes #2953.

### DIFF
--- a/assets/css/amp-post-meta-box.css
+++ b/assets/css/amp-post-meta-box.css
@@ -65,18 +65,3 @@
 		line-height: 2.8;
 	}
 }
-
-.amp-block-validation-errors {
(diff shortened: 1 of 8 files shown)


swissspidy:
Tests need to be updated, but otherwise this is what I had in mind for fixing this.

PR #2986, opened 2019-08-07
Improve handling of unlisted Vimeo videos

westonruter:
Vimeo has the ability to create unlisted videos: https://vimeo.com/blog/post/share-unlisted-videos-with-private-links/

Such videos have URLs like: `https://vimeo.com/172355597/abcdef0123`

The AMP plugin was erroneously turning this into the following:

```html
<p><amp-vimeo data-videoid="abcdef0123" layout="responsive" width="600" height="338"></amp-vimeo></p>
```

And this was causing an AMP validation error, because the video ID has to consist of digits only. But here the secret video key was erroneously being provided as the ID.

Now, it's likely that this will still result in the Vimeo unlisted video being unviewable, since the unlisted key is just dropped. Nevertheless, it's somewhat better than having an AMP validation error.

### DIFF
--- a/tests/php/test-amp-vimeo-embed.php
+++ b/tests/php/test-amp-vimeo-embed.php
@@ -8,7 +8,7 @@
 /**
  * Class AMP_Vimeo_Embed_Test
  *
- * @covers AMP_SoundCloud_Embed_Handler
+ * @covers AMP_Vimeo_Embed_Handler
  */
 class AMP_Vimeo_Embed_Test extends WP_UnitTestCase {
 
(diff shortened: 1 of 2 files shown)


westonruter:
I've reported the issue upstream as well: https://github.com/ampproject/amphtml/issues/23804

tests/php/test-amp-vimeo-embed.php
```diff
@@ -1,6 +1,22 @@
...
+/**
+ * Class AMP_Vimeo_Embed_Test
+ *
+ * @package AMP
+ */
 
+/**
+ * Class AMP_Vimeo_Embed_Test
+ *
+ * @covers AMP_SoundCloud_Embed_Handler
```
swissspidy:
```diff
- * @covers AMP_SoundCloud_Embed_Handler
+ * @covers AMP_Vimeo_Embed_Handler
```

PR #2987, opened 2019-08-08
Optimize svg loading

swissspidy:
* Use different svg loader
* Optimize all svgs using svgo

`babel-plugin-inline-react-svg` is not super well maintained and was flagged by `npm audit` because it uses an old version of `svgo`.

There's no need to optimize svgs everytime they are loaded. Instead, we can optimize them once.


swissspidy:
Looks like it still needs some Babel plugin...

swissspidy:
Closing in favor of #2999.

PR #2988, opened 2019-08-08
Hide tooltips that should be hidden

swissspidy:
Hides tooltip containers that don't seem to be actually used themselves. Only their content is extracted in `amp-validation-tooltips.js`.

Before:

Screenshot 2019-08-08 at 16 17 55

After:

Screenshot 2019-08-08 at 16 26 45

### DIFF
--- a/assets/css/admin-tables.css
+++ b/assets/css/admin-tables.css
@@ -69,3 +69,7 @@ td.column-found_elements_and_attributes div {
 .wrap .wp-heading-inline + .page-title-action {
 	margin-left: 1rem;
 }
+
+.tooltip[hidden] {
+	visibility: hidden;
+}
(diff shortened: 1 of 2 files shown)


westonruter:
Where are you seeing these containers being shown? I don't see them at all on any of the compatibility tool admin screens.

swissspidy:
On the Error Index screen, the Validated URLs screen, and on the Validated URLs details screen.

Example:

Screenshot 2019-08-09 at 11 29 46

Happens on my personal site as well as locally on my AMP plugin dev environment.

swissspidy:
Indeed seems to come from an injected stylesheet from an extension. I think it makes sense to explicitly hide the tooltip element to avoid such a conflict.

PR #2991, opened 2019-08-08
E2E tests for Page block.

### LINKED ISSUE #2903: E2E Tests: Page Block
* [ ] Verify background color functionality
* [ ] Verify gradients work as expected
* [ ] Verify opacity setting works as expected (Related: #2902)
* [ ] Test background media setting
* [ ] Test page advancement setting

miina:
Closes #2903.


tests/e2e/specs/stories-editor/page.test.js
```diff
@@ -0,0 +1,183 @@
...
+		await createNewPost( { postType: 'amp_story' } );
+		// Select the default page block.
+		await selectBlockByClientId(
+			( await getAllBlocks() )[ 0 ].clientId
+		);
+	} );
+
+	it( 'should be possible to add background color to Page', async () => {
+		const panelTitle = 'Background Color';
+		const colorToggle = await getSidebarPanelToggleByTitle( panelTitle );
```
miina: Note: the upstream `findSidebarPanelToggleButtonWithTitle` does not seem to work here.

swissspidy: [`findSidebarPanelToggleButtonWithTitle`](https://github.com/WordPress/gutenberg/blob/a779901c4fba1d8cd32475d8489708c2a134bb9f/packages/e2e-test-utils/src/find-sidebar-panel-toggle-button-with-title.js) is quite a long name 😄 

Looks like the upstream function doesn't work because the `Background Color` label is wrapped in an additional `.editor-panel-color-settings__panel-title` `<span>`, which it doesn't catch.

Sounds like something that could be reported upstream... Added a note here: https://github.com/WordPress/gutenberg/issues/16675

tests/e2e/specs/stories-editor/page.test.js
```diff
@@ -0,0 +1,183 @@
...
+
+		const editorPage = page;
+		const previewPage = await openPreviewPage( editorPage, 'amp-story' );
+		await previewPage.waitForSelector( '.amp-story-block-wrapper' );
+
+		const [ elementHandle ] = await previewPage.$x( '//amp-story-page/amp-story-grid-layer/amp-img/@src' );
+		const srcHandle = await elementHandle.getProperty( 'value' );
+		const src = await srcHandle.jsonValue();
+		expect( src ).toContain( '.jpg' );
+		expect( src ).toContain( 'http://localhost:8890/wp-content/uploads/2019/08/' );
```
swissspidy: That doesn't work next month... or when testing against a different site.

miina: Right! Of course it doesn't 💡 

I guess just `wp-content/uploads` should be enough. Or maybe this `expect` is not necessary at all. Will change.

PR #2999, opened 2019-08-08
Optimize all svgs using svgo

swissspidy:
Even though `babel-plugin-inline-react-svg` should care of svg optimization, it's better to make sure these are optimized from the get-go.

Also, we might switch to another solution at some point, as that Babel plugin isn't super well maintained.


swissspidy:
Odd build failure: https://travis-ci.org/ampproject/amp-wp/jobs/569443178#L827-L860

swissspidy:
svgo doesn't seem to like already optimized SVGS 🤷‍♂ 

Can revisit at some point in the future.

Issue #2673, opened 2019-06-21
Allow Developer Tools to be turned off to minimize (even hide) validation error info


westonruter:
AMP validation error notices are concerning to non-technical users:

> ![Screen Shot 2019-06-21 at 17 02 31 (1)](https://user-images.githubusercontent.com/134745/59931999-7f8bca00-9446-11e9-8313-88c3035713ee.png)

Rejection of validation errors should be limited to higher-privileged users, with Authors and Contributors (or other non-Admin role) shown validation errors warnings that just prompt them either to remove the invalid block (#2285) or to escalate the error to an administrator for support.

There should perhaps be a filter to configure the verbosity of the warnings for authors. There's a tough balance here because at one hand the warnings could be hidden to not concern users, but then at the same time the errors mean that markup is being removed. So if the validation error notice is not presented to the user, there needs to be a way for those validation errors to be escalated to an administrator.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Whether developer tools are disabled should no longer be determined by the template mode. Developer tools should be able to be enabled in Reader mode the same way as they are in Transitional/Standard modes.
* A new user preference (exposed on the profile screen) should be exposed for whether developer tools are enabled. (This may also be exposed in the wizard per #4705.)
* Only administrators have access to the developer tools, although this should be gated behind a new `amp_validate` capability which by default is mapped by the `manage_options` capability (in the same way as `customize` is by default granted by `edit_theme_options`). 
* When developer tools are turned off, the validation screens are not displayed in the admin menu (although they can still be accessed by users who can `amp_validate`).
* When developer tools are turned off, synchronous validation is not performed in the editor when saving a post. Instead, saving a _published_ post should schedule an event to validate that URL. (To be done as part of #1756.)
* When there are unreviewed validation errors, a Site Health test should be make this known. See https://github.com/ampproject/amp-wp/issues/1756#issuecomment-623864593.

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

archon810:
I'd love to see this brilliant feature, combined with a way to run AMP validation async https://github.com/ampproject/amp-wp/issues/2069.

westonruter:
See also this plugin: https://gist.github.com/westonruter/31ac0e056b8b1278c98f8a9f548fcc1a

[...]

Issue #2316, opened 2019-05-14
Redesign Validation Screens (Developer Tools)


westonruter:
The admin screens for the Validated URLs and the Validation Errors are implemented in a classic WordPress way™, using post list tables and other traditional admin hooks. The end result is that much of the operations require full page loads, and the interfaces are dated. The UI needs to be revamped using modern WordPress interfaces pioneered with Gutenberg (e.g. using React). In the same way, we should eliminate synchronous validation as much as possible: https://github.com/ampproject/amp-wp/issues/2069.

Additionally, more of the management of the validation errors should be surfaced inside of the validation error notices on the edit post screen. Instead of merely indicating the number of validation errors as well as the invalid markup with each block, there should be a way to access full error details and source information from the editor itself; authors should just be able to remove the block (#2285) and/or escalate the validation error to an administrator. Non-administrators should not be able to change the validation error state, and they should not be able to access the compatibility tool screens. In short, redesigning the compatibility tool screens should involve components which can be re-used in the context of the validation warning notices.

As part of this, the site validation functionality needs to be fully implemented in the REST API. Currently we have a bare minimum implemented for Gutenberg integration, the read-only `amp_validity` field added to the post entity. In addition to that, there needs to be endpoints for:

* Listing all validated URLs.
* Requesting a URL to be validated.
* Listing the validation errors for a given URL.
* Listing validation errors.
* Listing URLs containing a given validation error.
* Changing the validation status for one or more errors.

amedina:
- Allowing for advance information and options to be hidden (toggled), similarly to how advanced settings are dealt with in Chrome.

amedina:
- Revise and improve all messages/copy in the plugin
- Change language from "Accepted" and "Rejected" to reflect the status of the error rather than the action taken by the user: i.e. "Markup Removed" and "Markup not Removed" (this will align the debugging workflow and language with the Sanitization aspects/approaches)
- Redesign information exposure approach (i.e. expose baseline information, and make more complex/advance information accessible on demand)

/cc @pbakaus

pbakaus:
OMG I'm so excited!

amedina:
* Enable contextual strict mode, where users on non-admin roles (e.g. editors, authors, contributors) would not have access to the compatibility tool and will not be allowed to control the transition status of plugin actions (e.g. can’t change a validation error action from “Accepted” to “Rejected”

amedina:
* Provide programmatic access to hide compatibility tool

[...]

PR #3005, opened 2019-08-09
Always hide AMP admin menu item and compatibility tool menu items for non-admins role

### LINKED ISSUE #2702: Inconsistent capability checks on settings page
Looking at #2692, I realized that even contributors have access to the AMP settings page, even though you cannot actually change the settings.

There's a message saying:

```
You do not have permission to modify these settings. They are shown here for your reference. Please contact your administrator to make changes.
```

Screenshot:

Screenshot 2019-06-28 at 11 19 06

There are a few problems I see on that page:

1. The Experiences checkboxes are not readonly.
1. There are links to AMP Validated URLs and AMP Validation Error Index.  
    1. The user can access the AMP Validated URLs screen, but cannot perform any action.
    1. The user has no access to the AMP Validation Error Index screen.
1. This settings page is totally irrelevant for contributors and basically any user who is not an admin.  
  Users on my site should not need to see this screen. They don't even know what AMP is, nor should they need to know.

Suggested changes:

1. Make checkboxes readonly
2. Prevent access to AMP Validated URLs screen for users without the necessary permission
3. Hide "View current site compatibility results for standard and transitional modes" link if user lacks permission
4. Prevent access to settings screen if user lacks permission

westonruter:
See  https://wordpress.org/support/topic/hide-amp-in-side-bar-for-specific-roles-authors/

When a user is not an administrator, they still have access to the AMP settings screen but all of the fields on the screen are disabled and they can't modify anything. This was allowed because the Validated URLs and Validation Error screens are are admin submenu items under the top-level AMP menu page. Thus the top-level  page was given `edit_posts` capability, though all of the settings required `manage_options` to change.

The thinking here was to allow the users to access the admin screens the compatibility tool even if they cannot `manage_options`. In reality, this is just noise and non-admins should not be concerned with site-level validation errors. See https://github.com/ampproject/amp-wp/issues/2316#issuecomment-517482714 and https://github.com/ampproject/amp-wp/issues/2673. 

Nevertheless, a case can be made to continue allowing a user to access the Validated URL screen individually for posts that they can edit. This is what this PR does. Non-admin users never see the top-level AMP admin menu item, and they never see the admin menu items for Validated URLs and Validation Errors. The only way they can get to these screens is by causing a validation error, at which point they will see the warning notice in Gutenberg, allowing them to access the screen via the “Review Issues” link.

This PR also hides the Validated URLs from the “At a Glance” dashboard widget, if the user is not an administrator.

# Before

Screen Shot 2019-08-09 at 13 21 14

# After

Screen Shot 2019-08-09 at 13 20 08

----

Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3488014/amp.zip) - 1.2.1-beta1-20190809T205628Z-303e81ea

Fixes #2702.

### DIFF
--- a/tests/php/validation/test-class-amp-validation-error-taxonomy.php
+++ b/tests/php/validation/test-class-amp-validation-error-taxonomy.php
@@ -84,27 +84,6 @@ class Test_AMP_Validation_Error_Taxonomy extends WP_UnitTestCase {
 	}
 
 	/**
-	 * Test should_show_in_menu.
-	 *
-	 * @covers AMP_Validation_Error_Taxonomy::should_show_in_menu()
-	 */
-	public function test_should_show_in_menu() {
-		global $pagenow;
-		add_theme_support( AMP_Theme_Support::SLUG );
-		$this->assertTrue( AMP_Validation_Error_Taxonomy::should_show_in_menu() );
-
-		remove_theme_support( AMP_Theme_Support::SLUG );
-		$this->assertFalse( AMP_Validation_Error_Taxonomy::should_show_in_menu() );
-
-		$pagenow          = 'edit-tags.php';
-		$_GET['taxonomy'] = 'post_tag';
-		$this->assertFalse( AMP_Validation_Error_Taxonomy::should_show_in_menu() );
-
-		$_GET['taxonomy'] = AMP_Validation_Error_Taxonomy::TAXONOMY_SLUG;
-		$this->assertTrue( AMP_Validation_Error_Taxonomy::should_show_in_menu() );
-	}
-
-	/**
 	 * Test get_term.
 	 *
 	 * @covers AMP_Validation_Error_Taxonomy::get_term()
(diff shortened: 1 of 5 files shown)


tests/php/validation/test-class-amp-validation-error-taxonomy.php
```diff
@@ -994,6 +975,9 @@ public function test_add_admin_notices() {
 	 * @covers \AMP_Validation_Error_Taxonomy::filter_tag_row_actions()
 	 */
 	public function test_filter_tag_row_actions() {
+		wp_set_current_user( $this->factory()->user->create( [ 'role' => 'administrator' ] ) );
+		global $pagenow;
+		$pagenow = 'edit-tags.php';
```
westonruter: I'm confused as to why this is needed now. Perhaps global state had set it before.

westonruter: Yes, this seems to be the case.

PR #3006, opened 2019-08-11
Improve story page background media with image srcset, reduced image size, and a11y text

### LINKED ISSUE #2835: Re-evaluate maximum image size in stories
_Based on a [discussion in the AMP Slack workspace](https://amphtml.slack.com/archives/C9ARUE6Q6/p1561810803118000) (link needs [registration](https://bit.ly/amp-slack-signup))_

In the Stories editor we currently limit images to a maximum height of 1440px, while the official recommendation is 720p for image and video assets.

The conclusion from the discussion was that it might be worth experimenting with those values where possible and seeing how they impact performance in practice.

westonruter:
Fixes #2835. Revisits #2332, #2509.

* Changes `amp_story_page` image size to 1280px high instead of 1440px high.
* Instead of adding an `amp-img` for the `amp-story-page` background image element, this PR adds an `img` similar to how the `core/image` block does, including the important `wp-image-{attachment_id}` class. This ensures that `wp_make_content_images_responsive()` is able to supply a `srcset` for the image (whereas currently there is none), potentially allowing reduced bandwidth when more appropriate image sizes are available; this allows for plugins like Jetpack to supply dynamically-generated image sizes. The image sanitizer converts the `img` to an `amp-img`. The inclusion of `srcset` also means this logic can be replaced with just `const mediaUrl = media.url`:

https://github.com/ampproject/amp-wp/blob/186c83ee4d8b0fc7d383ae27fd2891c9bb094ec7/assets/src/stories-editor/blocks/amp-story-page/edit.js#L137

* Improves accessibility by adding an `alt` attribute to the background image, and an `aria-label` to the background video.
* Uses the `object-position` attribute instead of the `object-position` style. This reduces the amount of CSS included `style[amp-custom]` and avoids the opaque `amp-wp-...` class name.

Still needing to be done:

- [x] Add block deprecation logic.
- [ ] Add `1800px`-tall image size for desktop stories?
- [x] Simplify logic in `\AMP_Story_Media::add_new_max_image_size()`? See https://github.com/ampproject/amp-wp/pull/2509/commits/c330ce77a06d1213afca08b33b01e84c306db47c.

# Rendered Output Before
```html
...
<amp-story-grid-layer template="fill">
	<amp-img
		src="https://wordpressdev.lndo.site/content/uploads/2019/08/American_bison_k5680-1-2-2208x1280.jpg"
		layout="fill"
		class="amp-wp-896316c"
	></amp-img>
</amp-story-grid-layer>
...
```

## Rendered Output After
```html
...
<amp-story-grid-layer template="fill">
	<amp-img
		src="https://wordpressdev.lndo.site/content/uploads/2019/08/American_bison_k5680-1-2.jpg"
		layout="fill"
		alt="American Bison"
		class="wp-image-2911 amp-wp-enforced-sizes"
		object-position="85% 50%"
		srcset="https://wordpressdev.lndo.site/content/uploads/2019/08/American_bison_k5680-1-2-1963x1280.jpg 1963w, https://wordpressdev.lndo.site/content/uploads/2019/08/American_bison_k5680-1-2-300x196.jpg 300w, https://wordpressdev.lndo.site/content/uploads/2019/08/American_bison_k5680-1-2-1024x668.jpg 1024w, https://wordpressdev.lndo.site/content/uploads/2019/08/American_bison_k5680-1-2-768x501.jpg 768w, https://wordpressdev.lndo.site/content/uploads/2019/08/American_bison_k5680-1-2-1200x783.jpg 1200w, https://wordpressdev.lndo.site/content/uploads/2019/08/American_bison_k5680-1-2-1400x913.jpg 1400w"
	></amp-img>
</amp-story-grid-layer>
...
```


westonruter:
I'm somewhat out of my depth for the `deprecated` handling.

/cc @swissspidy

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -134,13 +135,15 @@ class PageEdit extends Component {
 			mediaType = media.type;
 		}
 
-		const mediaUrl = has( media, [ 'sizes', MAX_IMAGE_SIZE_SLUG, 'url' ] ) ? media.sizes[ MAX_IMAGE_SIZE_SLUG ].url : media.url;
```
westonruter: @kienstra This appears no longer to be necessary when the `srcset` is present. 

Originated in https://github.com/ampproject/amp-wp/pull/2509/commits/0bd76bc652288cb82448d6134448323d5d853778

swissspidy:
@miina Mind taking a look at adding the block deprecation for the changed img tag?

miina:
> Mind taking a look at adding the block deprecation for the changed img tag?

Sure, checking now.

miina:
@googlebot I consent.

miina:
Added the deprecation,

- [ ] Templates need an update as well,

will check that, too, to be able to test the deprecation better.

assets/src/stories-editor/blocks/amp-story-page/edit.js
kienstra: Ah, that makes sense.

westonruter:
> * Eliminate now-unnecessary unsetting of `$image_sizes['full']` in `\AMP_Story_Media::add_new_max_image_size()`? See [c330ce7](https://github.com/ampproject/amp-wp/commit/c330ce77a06d1213afca08b33b01e84c306db47c).

@kienstra Thoughts on how much of this is still relevant now?

kienstra:
Hi @westonruter,

>Eliminate now-unnecessary unsetting of $image_sizes['full'] in \AMP_Story_Media::add_new_max_image_size()? See c330ce7.

>@kienstra Thoughts on how much of this is still relevant now?

Good question.

I think the [elseif block that unsets the full image](https://github.com/ampproject/amp-wp/blob/f17f59747387af9706b49a78001b819859e65dbb/includes/class-amp-story-media.php#L225-L227) may still be relevant, but the [first if block](https://github.com/ampproject/amp-wp/blob/f17f59747387af9706b49a78001b819859e65dbb/includes/class-amp-story-media.php#L224) probably isn't.

### 1. First `if` block

https://github.com/ampproject/amp-wp/blob/f17f59747387af9706b49a78001b819859e65dbb/includes/class-amp-story-media.php#L223-L224

^ This probably isn't needed anymore. It had 2 purposes:

a) On clicking 'Background Media' > 'Select Media' and uploading an image, it adds `self:: MAX_IMAGE_SIZE_SLUG` to the available image sizes

b) In the Image block, on clicking 'Media Library,' it adds the `self:: MAX_IMAGE_SIZE_SLUG` to the available image sizes in the Media Library

But removing that `if` block doesn't seem to have an effect. Like you mentioned, in this PR the Background Media [doesn't look at](https://github.com/ampproject/amp-wp/pull/3006/files#diff-3ebaa9cdbb7e5502ab37969c8f31350eL137) `media.sizes`, it uses the `media.url`.

### 2. elseif block

https://github.com/ampproject/amp-wp/blob/f17f59747387af9706b49a78001b819859e65dbb/includes/class-amp-story-media.php#L225-L228
Deleting this conditional did have an effect, and I'm not sure if it'd be good to delete it now.

To recap what it does, in the Image block's 'Image Size' `<select>`, it removes the 'Full Size', and adds 'Story Max Size'.

Without that code above, a user can select a full size image. And it looks like the `srcset` doesn't reduce the image size enough.

Though the user would have to select the full size image, large is still the default.

For example:

1. Checkout this PR's branch and delete this line: 
https://github.com/ampproject/amp-wp/blob/f17f59747387af9706b49a78001b819859e65dbb/includes/class-amp-story-media.php#L99
1. Create an Image block
2. Click 'Media Library'
3. Upload a [4000 x 4000 image](https://picsum.photos/id/374/4000/4000)
4. In the 'Image Size' `<select>`, choose 'Full Size'
5. Publish the story, and go to the front-end
6. Expected: the `srcset` produces a `currentSrc` at least a little smaller than the full `4000 x 4000` size
7. Actual: the `currentSrc` is `4000 x 4000`

full-size-image

This seems to be the case even with smaller image sizes, like [2000 x 2000](https://picsum.photos/id/379/2000/2000).

With this PR's branch as is, without deleting [that line](https://github.com/ampproject/amp-wp/blob/f17f59747387af9706b49a78001b819859e65dbb/includes/class-amp-story-media.php#L99), it's a little better:

story-here

westonruter:
@kienstra Thanks a lot for the thorough response. I did some testing as well and it turns out that when doing device emulation (simulating Pixel 2, for example) the `srcset` causes the full size to be reduced to the dimensions of the story page automatically. The `srcset` is not properly calculated when looking at a non-landscape story on a desktop viewport. So this being the case, I think it's fine actually to leave the `full` size. Done in 9e96dfe.

westonruter:
I think we can defer adding a new image size for desktop for a future release.

westonruter:
@miina Could you resolve the conflicts with the templates?

westonruter:
Humm, E2E failures?

miina:
Hmph, there's a test that doesn't pass reliably, will restart the tests, should look into improving this test again. Thought that it got fixed already but apparently not.

miina:
Or actually, this time it looks like it's another test. Looking into this now.

includes/templates/story-templates/travel-vertical.html
```diff
@@ -1,5 +1,5 @@
 <!-- wp:amp/amp-story-page {"mediaType":"image","focalPoint":{"x":0.6161616161616161,"y":1},"autoAdvanceAfterDuration":0} -->
-<amp-story-page style="background-color:#ffffff" id="cd109143-b365-4e99-9c44-f5bd707cf9c1" class="wp-block-amp-amp-story-page"><amp-story-grid-layer template="fill"><amp-img layout="fill" src="AMP_STORY_TEMPLATES_URL/quote-bg.jpg" style="object-position:61.61616161616161% 100%"></amp-img></amp-story-grid-layer><amp-story-grid-layer template="fill"></amp-story-grid-layer>
+<amp-story-page style="background-color:#ffffff" id="cd109143-b365-4e99-9c44-f5bd707cf9c1" class="wp-block-amp-amp-story-page"><amp-story-grid-layer template="fill"><img layout="fill" src="AMP_STORY_TEMPLATES_URL/quote-bg.jpg" object-position="61.61616161616161% 100%"/></amp-story-grid-layer><amp-story-grid-layer template="fill"></amp-story-grid-layer>
```
swissspidy: Template version needs to be bumped

miina: Looks like the template version bump gets lost with merging :D

Changing.

miina:
A different test is failing every time. Hard to debug as well since locally not failing.

It might be related to the issue in this version of Puppeteer that `page.click` doesn't seem to work reliably and instead `page.evaluate` should be used instead for consistent results, e.g.:
```
        await page.evaluate( ( selector ) => {
		document.querySelector( selector ).click();
	}, btnSelector );
```
The upstream helpers use the `page.click`, perhaps we should try replacing those and then try. @swissspidy Any other ideas?

swissspidy:
Looking at the tests now (`tests/e2e/specs/stories-editor/cta.test.js` failed), but what I can see is that there are block validation errors. Mind checking that, @miina?

swissspidy:
Skipped that flaky test for now; can be investigated separately.

swissspidy:
**Testing instructions**

1. Create new story
2. Upload large image as page background
3. Preview story
4. Verify that in the source code the img has a `srcset` and an `alt` attribute
5. Verify that template insertion still works as usual

miina:
Will look into that CTA test.

westonruter:
Opened Jetpack PR to reduce Photon-sized images from 1440px to 1280px: https://github.com/Automattic/jetpack/pull/13278

PR #3007, opened 2019-08-11
Fix AMP Story editor compatibility with code editor

### LINKED ISSUE #3010: Disable toolbar options in code editor
Gutenberg disables the global block inserter when code editing is on:

Screenshot 2019-08-12 at 13 46 57

Since we override that inserter with our custom one, that doesn't work in the Stories editor.

We should align behavior here so that the inserter is disabled as well.

Also, the shortcuts in the toolbar should be disabled and the current block should be de-selected.

Gutenberg's behavior is documented in an e2e test:

https://github.com/WordPress/gutenberg/blob/2f757fa8012edb5cd684a26adef9cb4e8c2158db/packages/e2e-tests/specs/editor-modes.test.js#L88-L118

westonruter:
I encountered a couple bugs when attempting the story editor after having been in code editor mode. First of all:

Screen Shot 2019-08-11 at 14 50 04

There was an apparent missing check for whether the `.block-editor-writing-flow__click-redirect` element exists before removing it. So this is fixed in ad8e115.

After fixing this, the editor was able to load without causing an error, but when switching to visual editor I saw duplicated controls:

Screen Shot 2019-08-11 at 14 52 02

This is due to not checking if the `#amp-story-editor` element already exists in the page. Fixed by 4c67fbf.

Also fixes #3010.


swissspidy:
Looks good! Might be a nice thing to add a small e2e test for. Will have a look.

tests/e2e/specs/stories-editor/code-editor.test.js
```diff
@@ -0,0 +1,46 @@
...
+
+	it( 'allows switching to code editor and back', async () => {
+		await createNewPost( { postType: 'amp_story' } );
+
+		await page.waitForSelector( '#amp-story-block-navigation' );
+
+		await expect( page ).not.toMatchElement( '.block-editor-writing-flow__click-redirect' );
+
+		await switchEditorModeTo( 'Code' );
+		await page.click( '.edit-post-more-menu [aria-label="More tools & options"]' );
```
miina: Could also use `clickButtonByLabel` here which also waits until the selector becomes available, although its availability is not really an issue here.

PR #3013, opened 2019-08-12
Text block: create new paragraph on enter

schlessera:
See https://github.com/ampproject/amp-wp/issues/2732

This is the second attempt to fix this, previous PR at https://github.com/ampproject/amp-wp/pull/2979 was reverted due to a regression.

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-text/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-text/edit.js
@@ -149,9 +149,8 @@ class TextBlockEdit extends Component {
 				} ) } style={ wrapperStyle } >
 					<RichText
 						wrapperClassName="wp-block-amp-story-text"
-						tagName="p"
-						// Ensure line breaks are normalised to HTML.
 						value={ content }
+						multiline="p"
 						onChange={ ( nextContent ) => setAttributes( { content: nextContent } ) }
 						// The 2 following lines are necessary for pasting to work.
 						onReplace={ this.onReplace }
--- a/assets/src/stories-editor/blocks/amp-story-text/save.js
+++ b/assets/src/stories-editor/blocks/amp-story-text/save.js
@@ -31,6 +31,7 @@ const TextBlockSave = ( { attributes } ) => {
 				style={ styles }
 				className={ className }
 				value={ content }
+				multiline="p"
 				format="string"
 			/>
 		);


miina:
@schlessera  I'm still seeing the previously existing Text blocks and the Text blocks in templates empty.

Is it working as expected for you?

Screenshot 2019-08-12 at 18 12 19

schlessera:
Sorry about that, I was changing back and forth, and seem to have mis-tested.

There's two different factors I simply cannot get to work at the same time. The main issue is that the `p` tag from the `paragraph` type in the text box and the `p` tag from the multiline feature are being confused, so the wrong one gets eradicated.

So, basically, when you want to have paragraph-level multiline, paragraph-elements in the templates get stripped. The text block currently supports 3 types: `p`, `h1` & `h2`. The `p` one gets stripped if multiline with paragraphs is enabled.

I can think of 3 solutions:

1.) Change the `p` type for the text block to something else. Probably not a good idea.
2.) Use only `br` tags for multiline, and add additional behavior so that `Enter` does the same as `Shift+Enter`. You would then only have line breaks in multiline text blocks, no paragraphs.
3.) Modify the underlying `RichText` component so that it can target more precisely, probably using classes. That might be opening a big can of worms, though.

@miina Is one of the above solutions acceptable? Can you think of something else?

schlessera:
> The `p` one gets stripped if multiline with paragraphs is enabled.

That's not well phrased. When the `p` one is present, the code to fetch the Richtext multiline content is mistargeted at the AMP textbox content and returns something that is discarded as not valid.

miina:
> The main issue is that the p tag from the paragraph type in the text box and the p tag from the multiline feature are being confused, so the wrong one gets eradicated.

The selector for the inner content is the `.amp-text-content` class. In theory, having multiple `p` elements shouldn't be an issue since it's pulling the content based on the classname. However, we're trying to wrap all the content with multiple `p` elements into a wrapper `p` element in here:
https://github.com/ampproject/amp-wp/blob/1cc5cf9ea5ee9004a554d6228876d046cb0c363d/assets/src/stories-editor/blocks/amp-story-text/save.js#L29-L34

Having `p` inside `p` is not allowed and results in the following markup instead:
Screenshot 2019-08-12 at 20 09 17

An alternative could be wrapping the content into `div` instead when saving (if it's `! ampFitText`), e.g.
```
                         <RichText.Content
				tagName="div"
				style={ styles }
				className={ className }
				value={ content }
				format="html"
			/>
```

Would result in the following markup:
Screenshot 2019-08-12 at 20 17 27

Would need some extra handling on the editing side as well not to cause validation errors and also deprecation for the previous blocks to work. Probably not the best option.

(Maybe that's what you were thinking of with option 1)? )

> 2.) Use only br tags for multiline, and add additional behavior so that Enter does the same as Shift+Enter. You would then only have line breaks in multiline text blocks, no paragraphs.

This seems to be the best option of the 3. I just checked and looks like `amp-fit-text` doesn't seem to calculate the font size correctly when it has multiple `p` elements in the front-end -- it goes over the limits of the width and height set to it.

Editor, correctly changing the font size based on the room it has:
Screenshot 2019-08-12 at 19 48 17

FE where the text does not adjust the size correctly:
Screenshot 2019-08-12 at 19 48 22

In conclusion -- multiple `p` elements don't seem to be working well with `amp-fit-text` anyway, the `br` tags seem to work as expected. Also doesn't break the previous blocks. I'd vote for that.

Thoughts?

@swissspidy Tagging you as well in case you have an alternative suggestion.

swissspidy:
Can we try the `br` approach here? Does `RichText` easily support adding that on <kbd>Enter</kbd>?

schlessera:
I'm currently looking into using the `<br>` approach here.

Using `multiline="br"` does not work as expected, it actually uses paragraph tags all the way, then, even if you use shift+enter, only that the shift-enter paragraph tag still has the class that is normally appended to the br tags.

From within the wrapper, it also seems to be non-trivial to intercept the enter and add a br tag, as we don't necessarily now where we are within the content. From what I gather in the documentation, this is not exposed.

The plan is now to intercept the enter key and turn it into a shift+enter instead. I can successfully wrap the richtext to intercept the enter key, but I don't know yet how to forward a shift+enter to the richtext control then.

Adapting the event does not work, the `shiftKey` property is read-only. So I think we need to `preventDefault()` and trigger a new event instead. I'm unsure how to trigger this event, so that it reaches the proper Richtext control. Any pointers on how that can be reliably achieved?

swissspidy:
Is simply calling `setAttributes` on enter a possibility? Otherwise that sounds a bit too complicated 🤔

schlessera:
I've tried many different things by now, but I can't get a working constellation that allows for multiple lines without breaking something else.

I tried:

- using multiple variations of `tagName` and `multiline` to find one that just works (multiline does only really support `p` and `li` tags, though, something like `br` basically produces random results)
- changing the wrapping tags of either the text within the textblocks or the blocks themselves
- using `onChange` to directly modify the contents
- using `onChange` to `setAttributes` on the props
- using `onEnter` (doesn't seem to trigger on the RichText)
- using `onKeyDown` to change the key (cannot be done as the `shiftKey` is read-only)
- ... probably other combinations as well...

Either this is non-trivial to get working, or I'm missing something obvious. I don't rule the latter out, as I'm not up to speed with Gutenberg and the AMP Stories yet. But by now, I don't think this is a good ticket for me to get into the flow, as I'm just trying random things at this point.

I suggest I either leave this for now (as I need to switch to the Plugin project anyway), or I pair up with @miina, who's more experienced, to at least make this an educational opportunity. Otherwise I feel I'm just wasting time on this particular ticket.

@miina, @kienstra Thoughts on the above?

miina:
@schlessera Thanks for trying all these things.

I think it makes sense to leave this ticket for now, based on your comments and the things you've tried so far it looks like there is something preventing the events that might be out of our control. E.g. something that requires an upstream change and thus would need debugging Gutenberg instead.

Looked into this issue more and found the following:

- `RichText` is already handling the `onEnter` event by default and basically ignores the custom `onEnter` added:
https://github.com/WordPress/gutenberg/blob/989c224ce81008d3448932cfb03aa0a71922e6be/packages/block-editor/src/components/rich-text/index.js#L55
- This said, one of the custom attributes which `onEnter` is using in its handling seems to be `onSplit` which in the Text block's case is set empty right now.
https://github.com/WordPress/gutenberg/blob/989c224ce81008d3448932cfb03aa0a71922e6be/packages/block-editor/src/components/rich-text/index.js#L84

Empty in Text block:
https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/blocks/amp-story-text/edit.js#L158

Maybe one option could be checking if we could do the handling within `onSplit`.

miina:
Accidentally sent the comment.

Wanted to add also that this was again just an idea so it's completely OK to leave this ticket for now and move on to something that makes more sense.

It's true that this is probably not the best first ticket for getting to know the plugin generally and it came out to be much more tricky than it seemed at first.

kienstra:
Yeah, like @miina said, it's fine to hand this issue off 😄

schlessera:
I'm closing this PR for now to clean the board.

PR #2977, opened 2019-08-07
RTLCSS all the things

swissspidy:
See #2976.

To do:

* [x] Make sure RTL stylesheets are actually loaded by WordPress

PR #3015, opened 2019-08-12
Fix handling of polyfill dependencies

swissspidy:
This aims to fix the issue discovered in #2977 by:

* Adding a polyfill for the new `wp.serverSideRender` global  
  The old `wp.components.serverSideRender` has been deprecated and [will throw notices](https://github.com/WordPress/gutenberg/pull/16133).
* Adjusted the `DependencyExtractionWebpackPlugin` config so that polyfills will not bundle their dependencies, but instead will reference the existing `wp.*` globals - except for themselves.  
  This way the polyfills are only as small as they need to be. Actually only affects the `server-side-render` polyfill and not the others.
* Adds a missing version param to the polyfills.

For testing:

* Verify that the "Enable AMP" toggle in the block editor is shown even when not using the Gutenberg plugin.

### DIFF
--- a/assets/src/polyfills/wp-server-side-render.js
+++ b/assets/src/polyfills/wp-server-side-render.js
@@ -1,0 +1,8 @@
+/**
+ * WordPress dependencies
+ */
+import * as serverSideRender from '@wordpress/server-side-render';
+
+window.wp = window.wp || {};
+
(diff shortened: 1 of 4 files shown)


swissspidy:
Great!

I‘m going to add a small e2e test for it tomorrow and then merge.

PR #3021, opened 2019-08-13
Fix poster image extraction

### LINKED ISSUE #2881: Extracted video poster image doesn't use correct image size

swissspidy:
* Force-loads video to prevent black image to be saved
* Changes mime type to jpeg

Fixes #2881 (again), where this regression was reported.


swissspidy:
Will investigate the failing build...

swissspidy:
Looks like `media-file-size-warnings.test.js` is failing for the same reason the poster extraction test is failing:

Since `getFirstFrameOfVideo()` now tries to play the video (to make sure the first frame is loaded), and Chromium doesn't support that, there's an uncaught `DOMException`.

I tried using [`canPlayType`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType) to detect whether the browser can play the video, but both Chrome and Chromium just return `maybe`. So that doesn't work.

Now I see the following options:

1. Try to see if `video.play()` can be avoided.
1. Alternatively, find alternative to `canPlayType` check.
1. Skip these tests (as some sort of temporary quick fix).

Thoughts?

swissspidy:
Found a way to avoid playing the video by using the `canplay` event, but that is not fired on Chromium... Will try to figure out a way there.

assets/src/stories-editor/helpers/index.js
```diff
@@ -1553,28 +1553,30 @@ export const getUniqueId = () => {
...
 			canvas.height = video.videoHeight;
 
 			const ctx = canvas.getContext( '2d' );
 			ctx.drawImage( video, 0, 0, canvas.width, canvas.height );
 
-			canvas.toBlob( resolve );
+			canvas.toBlob( resolve, 'image/jpeg' );
 		} );
+
+		video.src = `${ src }#t=0.5`;
```
westonruter: Why the `#t=0.5`? If intending to advance to half a second into the video, I believe you rather need to use `video.src.currentTime = 0.5`.

swissspidy: This is to seek forward a little bit so the poster image isn't just a black rectangle (because we're not playing the video). But it didn't occur to me that I can just use `video.currentTime` (not `video.src.currentTime`)

westonruter: Yes, `video.src.currentTime` was a typo. `video.currentTime` is what I meant.

Issue #2992, opened 2019-08-08
Snapping: Resizing blocks in relation to the page


swissspidy:
See #2875 for details.

Example: if the block is 200px from the left side then resizing the right size should allow snapping to 200px from the right side as well.

swissspidy:
Screencast: https://cloudup.com/cqBfxMlZFGF

Shows functionality from  this, #2993, #2994, and #2995.

swissspidy:
**Testing Instructions**

1. Create new story
1. Move the pre-existing block around on the page
1. For each of the block's sides (left, right, top, bottom) and centers (vertical and horizontal), verify that red snapping guidelines are shown in the following scenarios:  
    1. When resizing the block to the sides of the page (left, right, top, bottom).
    1. When resizing the block to the center (vertical and horizontal) of the page.

Basically, snapping guidelines should be shown to allow easily aligning blocks to the sides and center of the page while doing the resising. No actual _snapping_ is done.

Next, verify that no guidelines are shown when resizing and at the same time holding down the alt/option key.

csossi:
Verified in QA

PR #3023, opened 2019-08-13
Add more resizing handles

### LINKED ISSUE #2985: Allow user to resize blocks from corners
As a user I want an easier way to resize blocks.

AC1: Selecting the corner of a block, should allow the user to resize the block from the corner.

swissspidy:
**To do**

* [x] Consolidate styling for resizing handles (currently split across 2 files)
* [x] Fix styling for resizing handles (some look different on hover)
* [x] Change block position when dragging from top or left (side or corner)
* [x] E2E Tests

Fixes #2985.


swissspidy:
Current WIP screenshots:

Screenshot 2019-08-13 at 16 30 57
Screenshot 2019-08-13 at 16 30 52

swissspidy:
@miina I think this is a good opportunity to extract the `onResize` callbacks into a bunch of smaller helper functions. Those are currently quite large. Thoughts?

miina:
>  I think this is a good opportunity to extract the onResize callbacks into a bunch of smaller helper functions. Those are currently quite large.

Sounds good to me! It is getting quite large, especially when we'll add more calculations for additional handles.

By "Rotation handles" in the title, do you mean "resizing handles"? Or is the plan to add more rotation handles as well?

swissspidy:
> By "Rotation handles" in the title, do you mean "resizing handles"? Or is the plan to add more rotation handles as well?

Imagine additional rotation handles at every blue dot 😄 Yes, I meant resizing handles.

swissspidy:
> Change block position when dragging from top or left (side or corner)

@miina In case you wanna tackle this one, that would be greatly appreciated 😄 It would probably take me way longer to dig through the logic there.

miina:
I can try 😄

miina:
Note:
Would be good to get #2863 merged first since it reduces the complexity around the Text block by removing the area added for dragging only.

swissspidy:
Alright! That one's on my list for reviewing today anyway 🙂

miina:
@googlebot I consent.

miina:
The corner handles are still not working when the block is rotated, will continue looking into it.

miina:
The corners are working OK now as well, however, now there's a bug with the block still continuing to change the position when moving the mouse even when it's not resizing anymore due to the minimum size limit.

Looking into this now.

miina:
The issue mentioned above is fixed now as well, will add e2e tests and should be done.

Actually, will check also if this can be fixed:
https://github.com/ampproject/amp-wp/blob/cc10a5d322eb61fa8530965cb16d5e6acaaf9b97/assets/src/stories-editor/components/resizable-box/index.js#L214-L219

^ It's a bit more noticeable now with the corner handles. If it seems to remain a mystery then will create a separate issue for it.

assets/src/stories-editor/components/resizable-box/tests/getResizedWidthAndHeight.js
```diff
@@ -1,7 +1,7 @@
 /**
  * Internal dependencies
  */
-import { getResizedWidthAndHeight } from '../';
+import { getResizedWidthAndHeight } from '../helpers';
```
miina: Doesn't look like this test is running, at least not locally with `npm run test:js`.

miina:
Just for information that dragging from the resizing handles seems to work as expected in e2e tests, started adding those, will finish tomorrow.

assets/src/stories-editor/components/resizable-box/tests/getResizedWidthAndHeight.js
swissspidy: File ending needs to be `.test.js`

miina: Hmm, none of the `js` tests seem to have `.test.js` ending, only the `e2e` ones.

I just copied the `getBlockPositioning.js` file over from the general `helpers.js` tests, as you did with `getResizedWidthAndHeight.js` file, didn't change the ending.

Will check the config tomorrow.

miina:
E2e tests are added now. The resizing part should be ready for testing/reviewing.

assets/src/stories-editor/components/resizable-box/tests/getResizedWidthAndHeight.js
swissspidy: The test folder name should be `test`, not `tests`. Will fix.

swissspidy:
@miina Please review this when you get a chance 🙂 
It affects #2992 a bit, so would be nice if this one's in before that.

swissspidy:
Ah, will need to fix the merge conflict first. Doing that tomorrow.

And for #2992 I'll just rebase it on this branch. That seems to work fine.

PR #3031, opened 2019-08-14
Update e2e test setup

swissspidy:
Over at the [Site Kit repository](https://github.com/google/site-kit-wp) there was a really nice addition to the e2e setup. See https://github.com/google/site-kit-wp/pull/400 for details.

With this PR I am trying to port these over almost 1:1.

The biggest changes:

* Makes all the Bash scripts more DRY und thus easier to read.
* Creates new users out of the box, not relying on a `E2E_ROLE` env var or similar.
* Makes it easier to use a specific WordPress version other than `latest` in the future.

Contrary to Gutenberg, where we got the inspiration for this setup from, we don't need to run all the tests as a separate user. Instead, we only need to run a certain amount of tests as non-admins. It doesn't make sense to use a second build job for this on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,7 +92,7 @@ jobs:
       install:
         - nvm install
         - composer install
-        - npm run install
+        - npm install
       script:
         - npm run build:js
         - npm run build:css
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,11 +92,11 @@ jobs:
       install:
         - nvm install
         - composer install
-        # npm is handled by `npm run env:start`
+        - npm run install
       script:
-        - npm run env:start # also install npm dependencies
         - npm run build:js
         - npm run build:css
+        - npm run env:start
         - npm run env:reset-site
         - npm run test:e2e:ci
         - npm run env:stop
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,21 +101,6 @@ jobs:
         - npm run test:e2e:ci
         - npm run env:stop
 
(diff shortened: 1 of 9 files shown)


swissspidy:
@miina Can you quickly reply to googlebot here? I cherry-picked your commit 1188a8dedc4acb6e0993042a0cd2d8c40fb828dd to this branch, hence the complaining.

.travis.yml
```diff
@@ -92,26 +92,11 @@ jobs:
...
-      install:
-        - nvm install
-        - composer install
-        # npm is handled by `npm run env:start`
-      script:
-        - npm run env:start # also install npm dependencies
         - npm run build:js
         - npm run build:css
+        - npm run env:start
         - npm run env:reset-site
```
kienstra: It's nice how much simpler this is.

bin/local-env/launch-containers.sh
```diff
@@ -23,12 +20,12 @@ fi
...
 
 # Download image updates.
 echo -e $(status_message "Downloading Docker image updates...")
-docker-compose $DOCKER_COMPOSE_FILE_OPTIONS pull
+dc pull
 
 # Launch the containers.
 echo -e $(status_message "Starting Docker containers...")
-docker-compose $DOCKER_COMPOSE_FILE_OPTIONS up -d >/dev/null
+dc up -d >/dev/null
```
kienstra: Good idea to decompose this into `dc()`

miina:
@googlebot I consent.

PR #3032, opened 2019-08-14
Add an e2e test file for the Global Inserter

### LINKED ISSUE #2669: image data uri don´t show
Hi
I have a site with version 1.1.3 that have a share plugin with data:image/svg images, when upgraded to version 1.2.0 these images don´t show.
In other hand, stories don´t recognize gutenberg is installed, the radio button to enable stories is grey.

Wordpress 5.2.2
Amp 1.2.0
Generatepress 2.3
Sassy social share 3.2.26 modified to use svg data uri in sustitution of svg files
PHP 7.3.6

Rollback to the 1.1.3

### LINKED ISSUE #2894: E2E Tests: Global Inserter
* [ ] Verify that the global inserter shows _all_ possible blocks at all times, including the page block, even when an inner block is selected.
* [ ] Verify that inserter does only show the CTA block when not on the first page.
* [ ] Verify that inserter inserts elements always as children of the current page.
* [ ] Verify that inserter inserts pages always as top-level blocks.

kienstra:
* This tests that the Global Inserter has all of the expected blocks
* It also tests the cases in https://github.com/ampproject/amp-wp/issues/2894#issue-473407229

stories-gi

Closes #2894


kienstra:
**Request For Review**

Hi @swissspidy and @miina,

tests/e2e/specs/stories-editor/global-inserter.test.js
```diff
@@ -0,0 +1,138 @@
...
+const CTA_BLOCK_NAME = 'Call to Action';
+
+/**
+ * Tests that a given block is present in the inserter.
+ *
+ * @param {string} block The name of the block, like 'Image'.
+ */
+const testBlockPresentInInserter = async ( block ) => {
+	// It should be possible to insert the block.
+	await insertBlock( block );
```
miina: `insertBlock` should be already inserting the block using the global inserter, if that succeeds then this would already mean that the block is present in the inserter. Maybe it would make more sense to test if the block was actually added to the page after inserting? Alternatively, if you're testing if everything works without using `insertBlock` then maybe inserting the block after checking that there is a match would make more sense?

 Or perhaps I'm missing the intent of the test. Is there a case where `insertBlock` would succeed but the rest of the testing steps would get a different result?

tests/e2e/specs/stories-editor/global-inserter.test.js
```diff
@@ -0,0 +1,138 @@
...
+			await createNewPost( { postType: 'amp_story' } );
+			await testBlockPresentInInserter( block );
+
+			// It should still be possible to select the block, even with an Image innerBlock inserted and selected.
+			await insertBlock( 'Image' );
+			await testBlockPresentInInserter( block );
+		}
+	);
+
+	it( 'should only show the CTA block when not on the first page', async () => {
```
miina: This is already covered in the CTA tests.

miina: Or well, it is testing the presence in the Inserter, so I guess it's a bit different.

tests/e2e/specs/stories-editor/global-inserter.js
```diff
@@ -0,0 +1,138 @@
...
+
+const INSERTER_SELECTOR = '#amp-story-inserter';
+const CTA_BLOCK_NAME = 'Call to Action';
+
+/**
+ * Tests that a given block is present in the inserter.
+ *
+ * @param {string} block The name of the block, like 'Image'.
+ */
+const testBlockPresentInInserter = async ( block ) => {
```
swissspidy: That could be a use case for a [custom matcher](https://jestjs.io/docs/en/expect#custom-matchers-api).

swissspidy:
Closing for now

PR #3033, opened 2019-08-14
Inline color support

### LINKED ISSUE #2982: Implement flat text background scrim (support “data-text-background-color” attribute)
_Edit by @swissspidy:_

Documentation: https://amp.dev/documentation/components/amp-story/#data-text-background-color

> The `data-text-background-color` attribute highlights the text of the element with a specified color. To highlight the entire block, add this attribute directly to the text element. To only highlight the text, add the attribute and text to an inner . Note that works anywhere inside an `<amp-story-page>`, not just in `<amp-story-grid-layer>`.

[Example 1](https://github.com/ampproject/amphtml/blob/ddd07fdbba8985d968b1c2d495500a8c1d4607a9/extensions/amp-story/img/text-background-color-ex-1.png):

text-background-color-ex-1

[Example 2](https://github.com/ampproject/amphtml/blob/ddd07fdbba8985d968b1c2d495500a8c1d4607a9/extensions/amp-story/img/text-background-color-ex-2.png):

text-background-color-ex-2

I think example 2 is already possible with our existing text background color support.

Example 1 can be achieved by using Gutenberg's [Format API](https://developer.wordpress.org/block-editor/tutorials/format-api/), which makes it possible for developers to add custom buttons to the formatting toolbar and have them apply a _format_ to a text selection (e.g. bold, italic, or in this case background color).

https://wordpress.org/plugins/advanced-rich-text-tools/ is an example plugin that already allows adding an inline background color. We could do the same, just with different markup that uses the `data-text-background-color` attribute.

That attribute can have any color that `background-color` in CSS supports ([source](https://github.com/ampproject/amphtml/blob/2a2055e2857cf5eec7b9e53bdf44004adfdfb4c7/extensions/amp-story/1.0/utils.js#L289-L316)).

swissspidy:
Fixes #2982.

**To do**

* [x] Move color controls to block toolbar
* [x] Verify whether `data-text-background-color` attribute can actually be used or not
* [x] Ensure that the preview in the editor matches the actual result on the frontend


swissspidy:
Current status:

Screenshot 2019-08-14 at 20 01 04
Screenshot 2019-08-14 at 19 59 55

It seems like we cannot easily use the `data-text-background-color` attribute on the frontend and some inline style in the backend. But I need to get confirmation about that.

As for the UI, the color pickers should probably be added to the existing color panel within `withAmpStorySettings`.

swissspidy:
**Current status:**

Screenshot 2019-08-15 at 12 36 02
Screenshot 2019-08-15 at 12 35 58
Screenshot 2019-08-15 at 12 27 34
Screenshot 2019-08-15 at 12 17 30
Screenshot 2019-08-15 at 11 02 17

kienstra:
**Approved**

Hi @swissspidy,
This is a nice featured.

Changing the Text block's color and background color via the toolbar worked as expected:

rich-text-bg

text-r-controls

miina:
One note: The opacity control doesn't have an effect on the highlight. It actually makes more sense to me like this, however, wondering what are your thoughts on this / if it might be unexpected. Maybe naming "Opacity" to "Background opacity" could be an option. Thoughts?

swissspidy:
You mean the opacity control in the color panel in the block sidebar?

miina:
> You mean the opacity control in the color panel in the block sidebar?

Yes

swissspidy:
Hmm I'm -1 on adding some premature changes there. Let's add this and see what kind of feedback we get.

Issue #2782, opened 2019-07-10
Discovery of what/where/how of edit-ability related to supporting AMP Story Page attachments


MackenzieHartung:

miina:
Starting to look into this.

westonruter:
As we chatted about last week, the MVP for this could be to just allow a given page to link to another post/page at the story page level as an attachment, and that when the story page is viewed the content from the linked post/page could then be displayed as a story attachment. This would avoid the user from having to leave the context of the story to access the desired content. If this were done, then the embedded post/page content would probably need to be explicitly embedded inside of a `blockquote` with a `cite` attribute and link at the bottom pointing to the original location. This would avoid duplicate content warnings, and an added benefit would be that it would more easily allow direct linking to the attached content.

westonruter:
If we take the subsequent step of allowing attachment content to live inline with the story page itself, then this could be done in addition to a previous ability to link to an existing post/page as attachment content.

miina:
Maybe it would be good to make the Page Attachment already live within the page itself since that's likely something we'd like to provide in the future as well, considering that Page Attachment allows all kind of different 3rd party video embeds and social media embeds, etc. Note that both the CTA Layer and the Page Attachment have to be the last elements of the amp-story-page so we could do the following:

- Add a new Page Attachment block which could be available only if the CTA block hasn't been added.
- This block could have the available options for the Page Attachment as sidebar settings: title, theme, CTA text
- We could replicate the looks of the Page Attachment in the editor, display it in the bottom of the Page with the Attachment's CTA visible. If it's clicked on, the Page Attachment opens and covers the relevant page.
- Additionally, the Page Attachment block would have an inner block for embedding other posts/pages from the site, this could, for example, appear when the Page Attachment is open.

We'll likely want to allow adding other inner blocks for the Page Attachment in the future as well, this way we'd already create the basis for that, however, we could start with just this one embed option as @westonruter suggested.

Thoughts?

cc @swissspidy

miina:
This is currently a discovery task only. Is there a preference for creating a separate issue for the actual implementation or could we just adjust the same issue for that once the what-where-how has been confirmed?

swissspidy:
Some links for reference:

Documentation: https://amp.dev/documentation/components/amp-story/#amp-story-page-attachment
Best Practices: https://amp.dev/documentation/guides-and-tutorials/develop/amp_story_best_practices/#add-additional-content-with-attachments

----

[...]

Issue #1973, opened 2019-03-15
Implement support for `<amp-story-page-attachment>`


amedina:
AMP Stories allow you to attach additional content to a page as explained [here](https://www.ampproject.org/docs/reference/components/amp-story#page-attachments).

csossi:
- Changed "Swipe Up" copy to "Click Here":
![image](https://user-images.githubusercontent.com/16945664/64871122-90e20e00-d612-11e9-846e-d0a53b1923d9.png)

On previewing, it still has "Swipe Up":
![image](https://user-images.githubusercontent.com/16945664/64871163-a7886500-d612-11e9-8eed-74a35d2f9d31.png)

csossi:
Verified in QA

swissspidy:
Screencast: https://cloudup.com/cWMBzAeg8gh

PR #3035, opened 2019-08-14
[WIP] Page Attachment block

miina:
WIP PR for implenting Page Attachment.

Adds Page Attachment block, replicates the component's FE behavior.

Todo.
- [x] Add Page Attachment block.
- [x] Display the block in Inserter only when allowed.
- [x] Ensure being the last block.
- [x] Add CTA button which will reveal the attachment.
- [x] Display content based on the selected post ID.
- [ ] Add search form for getting an actual post to display (currently static 1).
- [ ] Decide what to display from the post.
- [ ] Add background color, font, text color settings?
- [ ] Add link to the original post.
- [ ] Restyle based on desktop design.
- [ ] Add proper block icon.
- [ ] E2E


miina:
Started creating the Page Attachment block, while creating it, a few questions arose:
- It seems like it would make sense to additionally allow choosing at least background color and font size in addition to just pulling in the content. Otherwise, it's either just on a dark or light background and might not look the best. An alternative would be to always display light text if using the dark theme and dark text when using the light theme. Thoughts?
- Any other settings that might be necessary?
- Currently, the editor design is based on how it would look like on a mobile phone. Desktop design in the FE differs quite a bit. Perhaps we should implement the desktop design instead in the editor, that should cover the landscape as well then. Thoughts?

This is how it looks like now. Light theme:
attachment

Dark theme:
Screenshot 2019-08-15 at 00 02 47



Additionally, I was searching around a bit about duplicate content and quoting and citing and only found information about quote and cite _not_ being treated differently, meaning that even if we'd wrap all the content inside those it would still be considered duplicate. Is there any resource which would state otherwise?

Originally I suggested displaying the excerpt but now testing it out it looks like that would be too short to look good.

If the post content would be considered duplicate anyway then perhaps it could make sense to implement Inner blocks instead then? E.g. Text block and Image block to start with.

Or do you think it's OK to just pull the content even if it'll be considered duplicate?

cc @swissspidy @westonruter

miina:
Actually, I didn't try how the attachment would look like with all these together:
- Title
- Category
- Featured Image
- Excerpt (or perhaps X first paragraphs or X first words?)
- Link to the Post.

Perhaps that could still work as well.

Would be great to hear your thoughts as well.

swissspidy:
Love the WYSIWYG edit functionality already!

> It seems like it would make sense to additionally allow choosing at least background color and font size in addition to just pulling in the content. Otherwise, it's either just on a dark or light background and might not look the best. An alternative would be to always display light text if using the dark theme and dark text when using the light theme. Thoughts?

Adding text & background color options sounds reasonable.

> Currently, the editor design is based on how it would look like on a mobile phone. Desktop design in the FE differs quite a bit. Perhaps we should implement the desktop design instead in the editor, that should cover the landscape as well then. Thoughts?

Since we don't have landscape support in the editor yet, using the mobile design in the editor makes sense.

> Additionally, I was searching around a bit about duplicate content and quoting and citing and only found information about quote and cite not being treated differently, meaning that even if we'd wrap all the content inside those it would still be considered duplicate. Is there any resource which would state otherwise?
> 
> Or do you think it's OK to just pull the content even if it'll be considered duplicate?

No idea off the top of my head. I wouldn't worry too much about that though. If I select a post to be displayed in that page attachment I'd write the content specific to that attachment anyway.

> Originally I suggested displaying the excerpt in attachment but now testing it out it looks like that would be too short to look good.

I'd assume one would want to tweak the template that is used for that anyway.

Or perhaps there could be options like "display excerpt", "display full content", "display  featured image", "display read more link", etc.

> If the post content would be considered duplicate anyway then perhaps it could make sense to implement Inner blocks instead then? E.g. Text block and Image block to start with.

I'm not sure how big of a hassle that would be. It would require us to go back to a linear edit flow for the attachment, showing block movers, disallow dragging, etc.

Long term we'd probably want to support that anyway though, soo... 🤷‍♂

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
```diff
@@ -0,0 +1,301 @@
...
+
+	componentDidMount() {
+		this.fetchSelectedPost();
+	}
+
+	componentWillUnmount() {
+		this.isStillMounted = false;
+	}
+
+	componentDidUpdate( prevProps ) {
```
miina: Initially when the block got unselected then the attachment was also closed, however, this caused the attachment to lose focus sometimes in some weird moments and it was quite disturbing. Now the attachment stays open until it's closed from `X` even if not selected.

Not sure what would be the best approach here. You can test the previous version out when adding this:
```
                if ( ! attributes.isSelected && prevProps.isSelected ) {
			this.toggleAttachment( false );
		}
```

Thoughts?

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
```diff
@@ -0,0 +1,301 @@
...
+								{ selectedPost && selectedPost.content && (
+									<RawHTML>{ selectedPost.content.rendered }</RawHTML>
+								) }
+								{ ( ! postId || failedToFetch ) && (
+									<>
+										{ failedToFetch && (
+											<span>{ __( 'The selected post failed to load, please select a new post', 'amp' ) }</span>
+										) }
+										<PostSelector
+											placeholder={ __( 'Search & select a post to embed content.', 'amp' ) }
```
miina: Not sure if there should be a more detailed description when nothing is selected.

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
```diff
@@ -0,0 +1,301 @@
...
+			setAttributes( newAttributes );
+		}
+	}
+
+	fetchSelectedPost() {
+		const { postId } = this.props.attributes;
+		this.isStillMounted = true;
+		if ( postId ) {
+			const fetchRequest = this.fetchRequest = apiFetch( {
+				path: `/wp/v2/posts/${ postId }`,
```
miina: Should we allow other post types?

miina:
@swissspidy Had some questions/uncertainties, added these as comments within the changed files, would you mind checking when you have time?

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
swissspidy: That sounds reasonable to me, yes. For example, personally I would rather want to embed a page here than a post.

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
swissspidy: We can always change it if it doesn't feel right. It's just a text :-)

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
swissspidy: Makes sense to me. There are quite a few ways to accidentally unselect a block.

swissspidy:
Just realized that when embedding a post in an attachment, we'd also need to load the default block library CSS in the editor so that things like the column block work properly.

westonruter:
FYI: As of #3084 you'll be able to add `amp-video` to an `amp-story-page-attachment`.

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
Skimm3r: Hello @miina @swissspidy 
This is Gary from Taiwan, I am a marketer. 
If you please consider adding embed function, that way for some of the people who have product page can insert a link directly without extra works.

miina: > If you please consider adding embed function, that way for some of the people who have product page can insert a link directly without extra works.

Note that embedding will be possible from the same site only for now. The post can be easily searched by title or content and then selected with just one click, the content will be pulled in without extra work.

Skimm3r:
I'll give it a try.
Thanks

Miina <notifications@github.com> 於 2019年8月26日 週一 下午11:17 寫道：

miina:
> we'd also need to load the default block library CSS in the editor so that things like the column block work properly.

We have some custom styling so it looks we need to add the relevant CSS styles for child elements of Page Attachment only ( `.amp-page-attachment-content .wp-block-XYZ`). Added the column blocks for now.

Also tested a bunch of other blocks and these seem to work as expected (including Archives, Image, Gallery, Cover Image).

Code block causes validation error and gets sanitized out.

Will test tomorrow with the remaining blocks as well.

swissspidy:
> Thoughts?

Can we perhaps keep the theme block attribute, but just leave out the UI for now? This way one could use a dark theme using a filter if they want to, and we don't have to invalidate the block once we add it back.

miina:
> and we don't have to invalidate the block once we add it back.

Keeping the block attribute wouldn't probably make a difference since the Page Attachment is a dynamic block, we wouldn't need to deprecate it later even if we'd add it back. Or did you mean something else by invalidating the block?

Perhaps we could create a separate issue for tracking the dark theme and AMP not switching the theme reliably in the FE instead?

swissspidy:
Ah I forgot it's a dynamic block.

Maybe @westonruter has an idea about the dark theme CSS? If it's a bug in AMP Stories itself we should report this upstream.

miina:
Hmm, the odd thing is that sometimes it worked in the FE (mostly not though). The main problem is the header of the attachment (which is wrapped in `shadow-root`) -- it works only partially by setting the title and the `X` font color for closing the attachment white (as it should) but leaves the background white as well 🤷‍♀ 

@swissspidy Are you OK with creating a separate issue for that and investigating it there? I could look at it either before time off if I get ready with other assigned issues or when I get back from time off. Thoughts?

swissspidy:
Sure, that works

miina:
Issue for tracking the theme option: https://github.com/ampproject/amp-wp/issues/3116

miina:
Note: I'll refactor this a little bit, will ask for review again once ready.

assets/src/stories-editor/blocks/amp-story-page/background-color-settings.js
```diff
@@ -0,0 +1,104 @@
+/**
```
miina: Note: moved this part out of the Page block since the file was growing huge. Same for Page Settings. Would be good to perhaps separate Media handling as well but since that's worked on in another PR, left it out.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -0,0 +1,180 @@
...
+	const {
+		attachmentClass,
+		postId,
+		postType,
+		title,
+		wrapperStyle,
+	} = attributes;
+
+	const fetchSelectedPost = () => {
+		isStillMounted = true;
```
swissspidy: Looks like isMounted checks are considered an [anti-pattern](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html).

A possible alternative using a custom hook: https://github.com/rajeshnaroth/react-cancelable-promise-hook

I also stumbled upon https://github.com/dai-shi/react-hooks-async, but that seems a bit overkill for this use case.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -0,0 +1,180 @@
...
+							{ `<h2>${ selectedPost.title.rendered }</h2>${ selectedPost.content.rendered }` }
+						</RawHTML>
+					) }
+					{ ( ! postId || failedToFetch ) && (
+						<>
+							{ failedToFetch && (
+								<span>{ __( 'The selected post failed to load, please select a new post', 'amp' ) }</span>
+							) }
+							<PostSelector
+								placeholder={ __( 'Search & select a post or page to embed content.', 'amp' ) }
```
swissspidy: This placeholder is quite long and doesn't fit within the input field:

Screenshot 2019-09-02 at 11 32 19

Suggestion:

* Add a proper label, e.g. "Attachment Content"
* Move this placeholder text as a description below the label
* Change placeholder to something like "Select post or page..."

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -0,0 +1,180 @@
...
+				<div className={ attachmentClass } style={ wrapperStyle }>
+					{ selectedPost && selectedPost.content && (
+						<RawHTML>
+							{ `<h2>${ selectedPost.title.rendered }</h2>${ selectedPost.content.rendered }` }
+						</RawHTML>
+					) }
+					{ ( ! postId || failedToFetch ) && (
+						<>
+							{ failedToFetch && (
+								<span>{ __( 'The selected post failed to load, please select a new post', 'amp' ) }</span>
```
swissspidy: What if it was just an intermittent issue and the user doesn't want to select any other post? Is a "Try Again" button an option?

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
miina: >Looks like isMounted checks are considered an anti-pattern.

It looks to be that's specifically for `isMounted()` and it actually mentions below to use your own check for if the component is still mounted:

> An easy migration strategy for anyone upgrading their code to avoid isMounted() is to track the mounted status yourself. Just set a _isMounted property to true in componentDidMount and set it to false in componentWillUnmount, and use this variable to check your component’s status.

In this component's case we do need to ensure that the component is still mounted since the fetch request might take longer time and the component might already get unmounted meanwhile, it's not necessarily a faulty logic within the component that creates setting state when the component is already unmounted (as mentioned in the post). Thoughts?

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
swissspidy: Forgot to mention the lack of styling on this input field too

assets/src/stories-editor/blocks/amp-story-page/background-color-settings.js
swissspidy: Nice! Long overdue :-)

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -0,0 +1,180 @@
...
+
+	useEffect( () => {
+		fetchSelectedPost();
+		return () => {
+			isStillMounted = false;
+		};
+	}, [] );
+
+	useEffect( () => {
+		fetchSelectedPost();
```
barklund: Wouldn't this (combined with L66) call this fetch function twice on initial component render? Why do you need both?

includes/class-amp-story-post-type.php
```diff
@@ -1746,6 +1747,91 @@ public static function render_block_latest_stories( $attributes ) {
...
+	 * Registers the Page Attachment block.
+	 */
+	public static function register_block_page_attachment() {
+		register_block_type(
+			'amp/amp-story-page-attachment',
+			[
+				'attributes'      => [
+					'postId'          => [
+						'type'    => 'number',
+						'default' => 1, // For testing only.
```
swissspidy: Might wanna remove this again :-)

assets/src/stories-editor/components/block-navigation/item.js
```diff
@@ -85,11 +85,11 @@ class BlockNavigationItem extends Component {
 
 	render() {
 		const { block, getBlockIndex, isSelected, onClick } = this.props;
-		const isCallToActionBlock = 'amp/amp-story-cta' === block.name;
+		const isUnMovableBlock = 'amp/amp-story-cta' === block.name || 'amp/amp-story-page-attachment' === block.name;
```
swissspidy: Since `BlockNavigationList` is already doing this check, we could just pass down the information.

assets/src/stories-editor/components/block-navigation/index.js
```diff
@@ -117,12 +117,12 @@ export default compose(
 		const { getBlockOrder, getBlocksByClientId, getSelectedBlockClientId } = select( 'core/block-editor' );
 
 		let blocks = getCurrentPage() ? getBlocksByClientId( getBlockOrder( getCurrentPage() ) ) : [];
-		// Let's get the CTA block to handle it separately.
-		const callToActionBlock = blocks.find( ( { name } ) => name === 'amp/amp-story-cta' );
+		// Let's get the CTA/Attachment block to handle it separately.
+		const unMovableBlock = blocks.find( ( { name } ) => name === 'amp/amp-story-cta' || name === 'amp/amp-story-page-attachment' );
```
swissspidy: Should we leverage `ALLOWED_MOVABLE_BLOCKS` here?

```js
const unMovableBlock = blocks.find( ( { name } ) => ! ALLOWED_MOVABLE_BLOCKS.includes( name ) );
```

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-opener.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+import { __ } from '@wordpress/i18n';
+import { RichText } from '@wordpress/block-editor';
+import { ENTER, SPACE } from '@wordpress/keycodes';
+
+const AttachmentOpener = ( { setAttributes, toggleAttachment, openText } ) => {
+	return (
+		<div className="open-attachment-wrapper">
+			<span
+				role="button"
```
swissspidy: Curious: why did a regular `button` not work?

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
barklund: In this instance this pattern remembering "mounted" state is definitely acceptable, as it is the only way to know whether the callback still needs to be run. 

The `react-cancelable-promise-hook` does the exact same thing internally. The module however does not work.

Using this module, you cannot tell the difference between the promise cancelling because the original promise failed (bad fetch) or because it has been cancelled. Not easily anyway.

They do attempt to [pass an object as the error message](https://github.com/rajeshnaroth/react-cancelable-promise-hook/blob/master/index.js#L8-L9) to be able to distinguish based off this property, but that fails as the JS engine (at least Chrome) internally converts the message to a string:

```
assert(new Error({foo: 1}).message === '[object Object]')
```

Creating a helper is definitely a good idea, but it has to be done correctly. Creating a custom error object to throw adding an `isCancelled` attribute. If you even need to throw an error and not just silently ignore the cancelled promise return.

I see different ways of creating such a helper, but thinking deeper about it, it's actually quite a bit tricky. What's the proper API for such a hook? Let's wait until we see a few more use-cases for such behaviour before trying to refactor.

I'd just go for the simple property that you have now, and worry about refactoring later. I don't event think it's worthy of a tech debt ticket.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-opener.js
```diff
@@ -0,0 +1,50 @@
...
+				tabIndex="0"
+				onClick={ () => {
+					toggleAttachment( true );
+				} }
+				onKeyDown={ ( event ) => {
+					if ( ENTER === event.keyCode || SPACE === event.keyCode ) {
```
swissspidy: Nice detail, was very handy in my testing!

I wonder if there could be similar functionality when the attachment is already expanded. I kept pressing <kbd>ESC</kbd>, but all that happened was that the attachment block received focus:

Screenshot 2019-09-02 at 12 07 01

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
swissspidy: > I'd just go for the simple property that you have now, and worry about refactoring later. I don't event think it's worthy of a tech debt ticket.

Works for me. Thanks to both of you!

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-opener.js
miina: It might have, this was a replication of the front-end markup how the Page Attachment displays it. Do you have a preference for trying it out with `button` instead?

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-opener.js
miina: Hmm. We could try adding the `ESC` handling when Page Attachment is selected, however, this might create some confusion since if the block loses focus the Attachment will still stay open, and then `ESC` would not close the Attachment, the user might not notice that the block lost focus though. Probably catching any `ESC` event to close the attachment even if it's not selecting wouldn't be that great either.

Thoughts?

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
miina: Not intentional, thank you!

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
miina: Screenshot 2019-09-02 at 16 35 12

Let me know if it's better now.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -129,15 +133,34 @@ const AttachmentContent = ( props ) => {
...
+												{ __( 'The selected post failed to load, please select a new post or ', 'amp' ) }
+												<Button
+													className="refetch-attachment-post"
+													onClick={ ( event ) => {
+														event.stopPropagation();
+														fetchSelectedPost();
+													} }
+													isLink
+												>
+													{ __( 'try again', 'amp' ) }
```
swissspidy: Unfortunately the tooling doesn't support string interpolation for translatable strings yet 😞 

However, string concatenation is not great for translators. All they would see is `The selected post failed to load, please select a new post or ` and a separate string `try again`, which makes it impossible to translate in every language.

Suggestion: use a full sentence `The selected post failed to load, please select a new post or  try again` and add a separate `Try Again` button after the sentence.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -0,0 +1,203 @@
...
+					{ ! isFetching && selectedPost && selectedPost.content && (
+						<RawHTML>
+							{ `<h2>${ selectedPost.title.rendered }</h2>${ selectedPost.content.rendered }` }
+						</RawHTML>
+					) }
+					{ ! isFetching && ( ! postId || failedToFetch ) && (
+						<>
+							{ failedToFetch && (
+								<div>
+									{ ! isFetching && (
```
swissspidy: Isn't that redundant because of the `! isFetching` check above?

miina: Yep! Was reorganizing things and left that in. Will fix.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
miina: Changes within 999e4fd.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-opener.js
swissspidy: Good points. Let's leave it as-is for now and see how it works during testing/beta.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-opener.js
swissspidy: Not urgent to change it I suppose.

tests/e2e/specs/stories-editor/page-attachment.js
```diff
@@ -0,0 +1,167 @@
...
+	insertBlock,
+	searchForBlock, openPreviewPage,
+} from '../../utils';
+
+/**
+ * Opens the first found Page Attachment.
+ *
+ * @return {Promise<void>} Promise.
+ */
+async function openPageAttachment() {
```
spacedmonkey: Shouldn't these functions be utils?

includes/class-amp-story-post-type.php
```diff
@@ -1746,6 +1747,90 @@ public static function render_block_latest_stories( $attributes ) {
...
+
+	/**
+	 * Renders the dynamic block Page Attachment.
+	 *
+	 * @param array $attributes The block attributes.
+	 * @return string $markup The rendered block markup.
+	 */
+	public static function render_block_page_attachment( $attributes ) {
+
+		if ( ! isset( $attributes['postId'] ) ) {
```
spacedmonkey: Also check if is empty.

includes/class-amp-story-post-type.php
```diff
@@ -1746,6 +1747,90 @@ public static function render_block_latest_stories( $attributes ) {
...
+		}
+		if ( isset( $attributes['wrapperStyle']['color'] ) ) {
+			$style .= 'color:' . $attributes['wrapperStyle']['color'] . ';';
+		}
+
+		ob_start();
+		?>
+		<amp-story-page-attachment layout="nodisplay" theme="light" data-cta-text="<?php echo esc_attr( $attributes['text'] ); ?>" data-title="<?php echo esc_attr( $attributes['title'] ); ?>">
+			<div class="<?php echo esc_attr( $attributes['attachmentClass'] ); ?>" style="<?php echo esc_attr( $style ); ?>">
+				<h2><?php echo esc_html( $content_post->post_title ); ?></h2>
```
spacedmonkey: Why not use `<?php the_title(); ?>` here.

swissspidy: Doesn't look like `$content_post` is the `global $post` here. Although then `the_content()` wouldn't work. 🤔

tests/e2e/specs/stories-editor/page-attachment.js
swissspidy: Should they be used outside this test suite at some point we can make them top level utils.

includes/class-amp-story-post-type.php
```diff
@@ -1871,7 +1875,7 @@ public static function render_block_page_attachment( $attributes ) {
 		?>
 		<amp-story-page-attachment layout="nodisplay" theme="light" data-cta-text="<?php echo esc_attr( $attributes['text'] ); ?>" data-title="<?php echo esc_attr( $attributes['title'] ); ?>">
 			<div class="<?php echo esc_attr( $attributes['attachmentClass'] ); ?>" style="<?php echo esc_attr( $style ); ?>">
-				<h2><?php echo esc_html( $content_post->post_title ); ?></h2>
+				<h2><?php echo the_title(); ?></h2>
```
spacedmonkey: echo not required here.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-opener.js
```diff
@@ -0,0 +1,50 @@
...
+						toggleAttachment( true );
+						event.stopPropagation();
+					}
+				} }
+				className="amp-story-page-open-attachment-icon"
+			>
+				<span className="amp-story-page-open-attachment-bar amp-story-page-open-attachment-bar-left" />
+				<span className="amp-story-page-open-attachment-bar amp-story-page-open-attachment-bar-right" />
+			</span>
+			<RichText
```
spacedmonkey: Should this attachment text have option to style the font?

includes/class-amp-story-post-type.php
swissspidy: This just crossed with my fix in 5e47558c097059e00eb145586a1595c3ee16899f, so consider it resolved 🙂

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-opener.js
swissspidy: I don't think it makes sense to have a font selection option for this kind of embedded content. It wasn't mentioned in #1973 nor #2782, so I consider it out of scope for this PR.

includes/class-amp-story-post-type.php
```diff
@@ -1798,6 +1799,99 @@ public static function render_block_latest_stories( $attributes ) {
...
+		if ( isset( $attributes['wrapperStyle']['color'] ) ) {
+			$style .= 'color:' . $attributes['wrapperStyle']['color'] . ';';
+		}
+
+		ob_start();
+		?>
+		<amp-story-page-attachment layout="nodisplay" theme="light" data-cta-text="<?php echo esc_attr( $attributes['text'] ); ?>" data-title="<?php echo esc_attr( $attributes['title'] ); ?>">
+			<div class="<?php echo esc_attr( $attributes['attachmentClass'] ); ?>" style="<?php echo esc_attr( $style ); ?>">
+				<h2><?php the_title(); ?></h2>
+				<?php the_content(); ?>
```
spacedmonkey: Should this having a wrapping div here?

swissspidy: You mean specifically the content?

Perhaps it makes more sense to iterate on this more broadly in a follow-up PR to allow overriding the markup via filters / theme templates.

spacedmonkey: It just pretty normal to wrap `<div>` around content for styling reasons.

swissspidy: Added one now.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -0,0 +1,199 @@
...
+import { RawHTML, useEffect, useRef, useState } from '@wordpress/element';
+import { ENTER, SPACE } from '@wordpress/keycodes';
+
+/**
+ * Internal dependencies
+ */
+import { PostSelector } from '../../components';
+import useElementClickDetector from './element-click-detector';
+
+const AttachmentContent = ( props ) => {
```
spacedmonkey: A comment explaining what this code does would be useful here.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -0,0 +1,199 @@
...
+		title,
+		wrapperStyle,
+	} = attributes;
+
+	const fetchSelectedPost = () => {
+		isStillMounted = true;
+		if ( postId ) {
+			setIsFetching( true );
+			const currentFetchRequest = fetchRequest = apiFetch( {
+				path: `/wp/v2/${ postType }s/${ postId }`,
```
spacedmonkey: Not all post types work this way. The path of the rest api doesn't have to match the post type here.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -0,0 +1,199 @@
...
+		toggleAttachment( false );
+	} );
+
+	useElementClickDetector( removeLinkRef, null, true );
+
+	return (
+		<div className="attachment-container">
+			<div className="attachment-wrapper">
+				<div className="attachment-header">
+					<span
```
spacedmonkey: Why not use a iconbutton here?

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
swissspidy: FWIW the post selector is currently hardcoded to only support `page` and `post`, for which this assumption here is true.

Is your recommendation to make this future-proof by fetching the respective REST base from `wp-json/wp/v2/types`?

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
swissspidy: This is the same close button that is used on the frontend by AMP, to make sure they look the same.

spacedmonkey: Might worth putting a comment on the code to state that.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
spacedmonkey: I would just pass the values of post types / path to rest api in a `wp_localize_script`. I would also add a filter to allow users to add custom post types.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-opener.js
```diff
@@ -0,0 +1,50 @@
...
+import PropTypes from 'prop-types';
+
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { RichText } from '@wordpress/block-editor';
+import { ENTER, SPACE } from '@wordpress/keycodes';
+
+const AttachmentOpener = ( { setAttributes, toggleAttachment, openText } ) => {
```
spacedmonkey: Comment explaining what this function does would be useful here.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-opener.js
spacedmonkey: I believe that publishers are going to want to style all of text on a story with they font. It would extremely strange if some part do not have the brand styling. I have created a ticket regarding this. https://github.com/ampproject/amp-wp/issues/3224

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Internal dependencies
+ */
+import './edit.css';
+import { getBackgroundColorWithOpacity } from '../../../common/helpers';
+import AttachmentOpener from './attachment-opener';
+import AttachmentContent from './attachment-content';
+
+class PageAttachmentEdit extends Component {
```
spacedmonkey: Comment explain this

spacedmonkey: Does this need to be a class here?

assets/src/stories-editor/blocks/amp-story-page/background-color-settings.js
```diff
@@ -0,0 +1,104 @@
...
+import { __, sprintf } from '@wordpress/i18n';
+import {
+	PanelColorSettings,
+} from '@wordpress/block-editor';
+import {
+	Button,
+	RangeControl,
+} from '@wordpress/components';
+
+const BackgroundColorSettings = ( { backgroundColors, setAttributes, overlayOpacity } ) => {
```
spacedmonkey: Comment on what this does.

assets/src/stories-editor/components/post-selector/index.js
```diff
@@ -0,0 +1,339 @@
...
+		this.props.onChange( inputValue );
+		this.updateSuggestions( inputValue );
+	}
+
+	onKeyDown( event ) {
+		const { showSuggestions, selectedSuggestion, suggestions, loading } = this.state;
+		// If the suggestions are not shown or loading, we shouldn't handle the arrow keys.
+		if ( ! showSuggestions || ! suggestions.length || loading ) {
+			switch ( event.keyCode ) {
+				// When UP is pressed, if the caret is at the start of the text, move it to the 0
```
spacedmonkey: For accessiblity, esc key key press should be handled here, exiting the dropdown. I would also look at tabbing...

assets/src/stories-editor/components/post-selector/index.js
```diff
@@ -0,0 +1,339 @@
...
+					break;
+			}
+
+			return;
+		}
+
+		const suggestion = this.state.suggestions[ this.state.selectedSuggestion ];
+
+		switch ( event.keyCode ) {
+			case UP: {
```
spacedmonkey: For accessiblity, esc key key press should be handled here, exiting the dropdown.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
swissspidy: Makes sense. We have the system in place for that already 👍

assets/src/stories-editor/blocks/amp-story-page/page-settings.js
```diff
@@ -0,0 +1,79 @@
...
+	RangeControl,
+} from '@wordpress/components';
+import { useSelect } from '@wordpress/data';
+
+/**
+ * Internal dependencies
+ */
+import { getTotalAnimationDuration } from '../../helpers';
+
+const PageSettings = ( { autoAdvanceAfter, autoAdvanceAfterDuration, setAttributes, clientId } ) => {
```
spacedmonkey: Comment function

assets/src/stories-editor/blocks/amp-story-page/page-settings.js
```diff
@@ -0,0 +1,79 @@
...
+	let autoAdvanceAfterHelp;
+
+	if ( 'media' === autoAdvanceAfter ) {
```
spacedmonkey: Comment explaining code.

assets/src/stories-editor/components/block-navigation/index.js
```diff
@@ -18,7 +18,7 @@ import BlockNavigationItem from './item';
 import { ALLOWED_MOVABLE_BLOCKS } from '../../constants';
 import './edit.css';
 
-function BlockNavigationList( { blocks,	selectedBlockClientId, selectBlock, callToActionBlock } ) {
+function BlockNavigationList( { blocks, selectedBlockClientId, selectBlock, unMovableBlock } ) {
```
spacedmonkey: Should `unMovableBlock` be a helper function? Seems like it might be useful else where in the code..

assets/src/stories-editor/blocks/amp-story-page/page-settings.js
swissspidy: The component itself or specifically this condition here?

FWIW this is not newly introduced code, it just has been extracted into its own file.

assets/src/stories-editor/components/post-selector/index.js
```diff
@@ -0,0 +1,339 @@
...
+import PropTypes from 'prop-types';
+
+/**
+ * Internal dependencies
+ */
+import './edit.css';
+
+const stopEventPropagation = ( event ) => event.stopPropagation();
+
+class PostSelector extends Component {
```
spacedmonkey: Add a comment here.

assets/src/stories-editor/components/block-navigation/index.js
swissspidy: Can you perhaps elaborate what you had in mind?

Worth noting that it's not really a `boolean` value, but the result of `const unMovableBlock = blocks.find( ( { name } ) => ! ALLOWED_MOVABLE_BLOCKS.includes( name ) );` in the `withSelect` below.

assets/src/stories-editor/components/post-selector/index.js
swissspidy: There is already one at the top of the file FWIW

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -0,0 +1,222 @@
...
+											isLink
+										>
+											{ __( ' Try again', 'amp' ) }
+										</Button>
+									</p>
+								</div>
+							) }
+							<PostSelector
+								labelText={ __( 'Attachment Content', 'amp' ) }
+								placeholder={ __( 'Search for a post or page ...', 'amp' ) }
```
spacedmonkey: Make this more generic, in case someone has filtered the support post types to include custom post types.

assets/src/stories-editor/components/post-selector/index.js
```diff
@@ -0,0 +1,339 @@
+/**
```
spacedmonkey: Instead of building a custom component why not reuse the autocomplete used by the font picker? This would keep accessibility and styling inline.

swissspidy: Worth noting that development on this was started way before the font picker was implemented and merged.

Also, I don't think that component allows for a dynamic list fetched via an API.

assets/src/stories-editor/components/block-navigation/index.js
spacedmonkey: Couldn't we make this a reuable function like this. 

`const unMovableBlock = ( name ) => => ! ALLOWED_MOVABLE_BLOCKS.includes( name ) ;`

includes/class-amp-story-post-type.php
```diff
@@ -956,11 +957,29 @@ public static function enqueue_block_editor_scripts() {
...
 		$allowed_video_mime_types = array_values( array_intersect( $allowed_video_mime_types, wp_get_mime_types() ) );
 
+		/**
+		 * Filters the list of allowed post types for use in page attachments.
+		 *
+		 * @since 1.3
+		 *
+		 * @param array Allowed post types.
+		 */
+		$page_attachment_post_types = apply_filters( 'amp_story_allowed_page_attachment_post_types', [ 'page', 'post' ] );
```
spacedmonkey: :+1:

assets/src/stories-editor/components/post-selector/index.js
spacedmonkey: I believe you are wrong here. Check this example from gutenberg that you originally sent over. https://github.com/WordPress/gutenberg/pull/7385/files#diff-8267665c4a351e7321fa6763c14922d3R30 This is doing a api requests to get a list of users. 

[Source](https://github.com/alphagov/accessible-autocomplete#source) is a function, you can put whatever logic you want in there, so a call to the api could be put in there.

assets/src/stories-editor/components/post-selector/index.js
```diff
@@ -219,6 +219,16 @@ class PostSelector extends Component {
 				}
 				break;
 			}
+			case ESCAPE: {
```
spacedmonkey: :tada:

assets/src/stories-editor/components/post-selector/index.js
swissspidy: While this is worthwhile to explore, I don't have the capacity to replace the component with the Autocomplete one in time for tomorrow. Would you have time to do that?

Otherwise I suggest exploring this between beta and RC. Thoughts?

PR #3039, opened 2019-08-17
Improve generation of AMP pages in post-processor

### LINKED ISSUE #1778: HTML fragment responses are erroneously converted into incomplete AMP documents
_Update from @westonruter:_

I've created a sample plugin that can be used to easily reproduce the issue: https://gist.github.com/westonruter/ca17e51c1fe7fe25efc95f04e98b3fbc

After activating the plugin and enabling native/paired mode, going to a URL such as https://example.com/?test_fragment&amp will show:

image

The problems with this are:

1. An HTML fragment is converted into an HTML document, unexpectedly.
2. The character encoding is all messed up. Potentially because it lacked a `<meta charset=utf-8>` to begin with.
3. The resulting AMP document is not even valid. This is likely due to the lack of the `wp_head` action not being run.

We should at the very least prevent starting the post-processor when the response does not contain `<html`. This can be done by extending the logic here:

https://github.com/ampproject/amp-wp/blob/c92e129333a8518088a37d47594079b537121d70/includes/class-amp-theme-support.php#L1522-L1530

However, we also need to validate the other two points to see if there isn't some bug to fix.

_Original description is as follows._

----

Hello there
I'm using your plugin. There is nothing wrong. But I see a character error in the search bar on the main page. In fact, everything was good for a few days. I didn't understand why.

When you put your plugin inactive, everything returns to normal. I've looked at the link below, but there are already these codes.

https://github.com/ampproject/amp-wp/blob/e60cb152a50e2ffbf5504d41aee859ad1d0e0baa/includes/class-amp-theme-support.php#L192

Example
https://www.odevbul.org (search pls)
karakter

### LINKED ISSUE #2968: Allow plugins to add metadata to AMP Stories
- I am optimizing my Story for **Search** & for **Social** using the Yoast plugin

- When I publish a Story, it IS being crawled and being added to Search results when I optimize with the Yoast plugin. ✅ 
- However, when I add my story to Slack or Social (Social/Share), it is not fetching the image or description. We think this is a bug. 😐

westonruter:
* Introduce `amp_story_head` action which plugins (e.g. Jetpack and Yoast) can use to inject additional metadata. Note that `amp_story_footer` action is not included because there is no clear use case for it yet in Stories. Fixes #2968.
* Prevent post-processing responses that do not contain well-formed HTML documents; they must have `<html>` and `<head>`. Pass-through HTML fragments unmodified, similar to passing-through JSON. Fixes #1778.
* Use post-processor phase to inject the AMP boilerplate and `v0.js`, rather than relying on `wp_head` action to do so. This eliminates duplicate code and makes AMP template generation more robust. Note that the logic in `\AMP_Theme_Support::ensure_required_markup()` should eventually be broken into a separate sanitizer, especially as part of the optimized AMP work (https://github.com/ampproject/amp-wp/issues/958).


includes/class-amp-theme-support.php
```diff
@@ -1657,41 +1642,87 @@ public static function ensure_required_markup( DOMDocument $dom, $script_handles
...
 
-		/*
-		 * "Specify the <script> tags for render-delaying extensions (e.g., amp-experiment, amp-dynamic-css-classes, and amp-story)."
-		 * "Specify the <script> tags for remaining extensions (e.g., amp-bind, ...). These extensions are not render-delaying and therefore
-		 * should not be preloaded because they might take away important bandwidth for the initial render."
-		 * {@link https://docs.google.com/document/d/169XUxtSSEJb16NfkrCr9y5lqhUR7vxXEAsNxBzg07fM/edit AMP Hosting Guide}
-		 */
+		// "5. Load the AMP runtime."
 		if ( isset( $amp_scripts['amp-runtime'] ) ) {
 			$ordered_scripts['amp-runtime'] = $amp_scripts['amp-runtime'];
```
schlessera: Is this missing an `unset`? It looks like it will be merged in a second time a bit further down below otherwise.
```suggestion
			$ordered_scripts['amp-runtime'] = $amp_scripts['amp-runtime'];
			unset( $amp_scripts['amp-runtime'] );
```

includes/class-amp-theme-support.php
```diff
@@ -1657,41 +1642,87 @@ public static function ensure_required_markup( DOMDocument $dom, $script_handles
...
+		 * "8. Specify any custom styles by using the <style amp-custom> tag."
 		 */
-		if ( isset( $links['icon'] ) ) {
-			foreach ( $links['icon'] as $link ) {
-				$link->parentNode->removeChild( $link ); // So we can move it.
-				$head->insertBefore( $link, $previous_node->nextSibling );
-				$previous_node = $link;
+		$style = $xpath->query( './style[ @amp-custom ]', $head )->item( 0 );
+		if ( $style ) {
+			// Ensure the CSS manifest comment remains before style[amp-custom].
```
schlessera: I think this could potentially be an unrelated comment here. However, it would not break anything in that case anyway, so it's probably a non-issue.

includes/class-amp-theme-support.php
```diff
@@ -1840,8 +1871,11 @@ public static function prepare_response( $response, $args = [] ) {
 			);
 		}
 
-		// Abort if the response was not HTML.
-		if ( 'text/html' !== substr( AMP_HTTP::get_response_content_type(), 0, 9 ) || '<' !== substr( ltrim( $response ), 0, 1 ) ) {
+		/*
+		 * Abort if the response was not HTML. To be post-processed as an AMP page, the output-buffered document must
+		 * have the HTML mime type and it must start <html> followed by <head> tag (with whitespace, doctype, and comments optionally interspersed).
```
schlessera:
```suggestion
		 * have the HTML mime type and it must start with <html> followed by <head> tag (with whitespace, doctype, and comments optionally interspersed).
```

includes/class-amp-theme-support.php
```diff
@@ -1840,8 +1871,11 @@ public static function prepare_response( $response, $args = [] ) {
...
+		 */
+		if ( 'text/html' !== substr( AMP_HTTP::get_response_content_type(), 0, 9 ) || ! preg_match( '#^(?:<!.*?>|\s+)*<html.*?>(?:<!.*?>|\s+)*<head.*?>#is', $response ) ) {
```
schlessera: The regexp here seems shaky. For example, an HTML fragment without a `<head>`, but containing a `<header>` element would still pass.

includes/class-amp-theme-support.php
```diff
@@ -2025,11 +2059,14 @@ public static function prepare_response( $response, $args = [] ) {
 		 * See <https://www.w3.org/International/questions/qa-html-encoding-declarations>.
 		 */
 		if ( ! preg_match( '#<meta[^>]+charset=#i', substr( $response, 0, 1024 ) ) ) {
+			$meta_charset = sprintf( '<meta charset="%s">', esc_attr( get_bloginfo( 'charset' ) ) );
+
 			$response = preg_replace(
-				'/(<head[^>]*>)/i',
-				'$1' . sprintf( '<meta charset="%s">', esc_attr( get_bloginfo( 'charset' ) ) ),
+				'/(<head.*?>)/is',
```
schlessera: As the filtering regexp up above will let `<header>` elements count as well, this will also happily append the meta charset to the first `<header>` element in that case.

includes/amp-helper-functions.php
```diff
@@ -361,8 +361,23 @@ function amp_get_asset_url( $file ) {
...
+ * Get AMP boilerplate stylesheets.
+ *
+ * @since 1.3
+ * @link https://www.ampproject.org/docs/reference/spec#boilerplate
+ *
+ * @return string[] Stylesheets, where first is contained in style[amp-boilerplate] and the second in noscript>style[amp-boilerplate].
+ */
+function amp_get_boilerplate_stylesheets() {
+	return [
+		'body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}',
```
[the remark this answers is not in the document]
westonruter: That's true but I was thinking by using a function we could have more flexibility later when refactoring. The `amp_get_boilerplate_stylesheets()` function could be made into a stub that is marked as `@deprecated`. I'd rather do this for now and then use constants once we have our `AMP` namespace up and running. Also, the `amp_get_boilerplate_code()` function could similarly have a return value that is a constant.

In any case, this function would be called like 2 times during the life of a page being generated, so I think this may be a premature optimization to turn it into constants. Do you think?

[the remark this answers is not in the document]
schlessera: Yes, makes sense.

includes/class-amp-theme-support.php
schlessera: Yes, that was understood. However, you're just fetching "any comment before this element". There's no check to make sure you'er actually fetching the correct comment.

However, I think this is a non-issue here mostly.

schlessera:
E2E tests are failing, probably for unrelated reasons.

Issue #2431, opened 2019-05-27
Eventually drop support for older WordPress versions


swissspidy:
This originated from #2408 / https://github.com/ampproject/amp-wp/pull/2408#issuecomment-495721510.

The AMP plugin currently requires WP 4.9+ and PHP 5.4+ for a few reasons:

* The plugin should be able to run on as many sites as possible.
* 4.9 is the last version without Gutenberg support, so it was assumed that transition period is a bit longer
* There were no drastic benefits in updating the WordPress version requirements
* There were no drastic benefits in updating the PHP version requirements

That means we currently need to support the following scenarios:

* WP 4.9
* WP 4.9 with Gutenberg plugin 4.9
* WP 5.0+ (Gutenberg built-in)
* WP 5.0+ with Gutenberg plugin

Gutenberg will eventually drop support for PHP 5.2 and require WordPress 5.2+ (which requires PHP 5.6):

* https://github.com/WordPress/gutenberg/issues/15796
* https://github.com/WordPress/gutenberg/pull/15809

That makes everything a bit more complex.

In related news, Yoast SEO is going back to their policy of only supporting WordPress latest and latest - 1: https://yoast.com/supporting-older-versions-of-wordpress/

One possibility to reduce our workload would be to bump requirements for PHP and WordPress as well. Perhaps with the AMP plugin 2.0 we could move to PHP 5.6 and WordPress latest and latest - 1 as well.

cathibosco:
One possibility to reduce our workload would be to bump requirements for PHP and WordPress as well. Perhaps with the AMP plugin 2.0 we could move to PHP 5.6 and WordPress latest and latest - 1 as well. 

I feel like this is reasonable from the perspective agencies and individual users. If fact it’s an opportunity to educate users once again the importance of maintenance.

westonruter:
Should we go ahead and bump to 5.6 as the minimum required version of PHP now?

Thoughts on when/whether we should do the bump to 7.0? /cc @schlessera

swissspidy:
I think a bump to 5.6 in anticipation of version 2.0 makes sense. Would be great to outline a short plan / todo list for that. Some items off the top of my head:

* Use newer version(s) of PHPUnit
* Use `phpdbg` for code coverage reports
* Update build environment on Travis (#3044)
* Unblock issues like #3252 and potentially others
* Remove legacy code in anticipation of #2315

7.0 could then happen around the time when core bumps its requirement too?

westonruter:
[...]

PR #3044, opened 2019-08-19
Use Xenial build environment on Travis

swissspidy:


swissspidy:
Totally forgot the fact that this env only supports PHP 5.6+ 🤦‍♂

westonruter:
We can do this as soon as we bump the minimum to PHP 5.6 in #2431.

PR #3052, opened 2019-08-20
Wait until content loaded before calculating font size.

### LINKED ISSUE #3047: Font size changing after refresh
Steps to reproduce:
- Add an author block.
- Resize the block to be quite long and quite narrow.
- Save
- Refresh

The result before refresh:
Screenshot 2019-08-19 at 21 31 33

The result after refresh:
Screenshot 2019-08-19 at 21 31 44

miina:
Fixes #3047 

Waits until the author block's content is loaded before calculating the font size.

Looks like the content switched quickly from `Anonymous` to the actual username, resulting in updating block attributes so that the update for `anonymous` font-size was applied later than the real font size.

This PR checks if the author block's content is loaded before calculation.

### DIFF
--- a/assets/src/stories-editor/components/with-meta-block-edit.js
+++ b/assets/src/stories-editor/components/with-meta-block-edit.js
@@ -28,7 +28,9 @@ import { maybeUpdateFontSize, maybeUpdateBlockDimensions } from '../helpers';
 // @todo: Allow individual blocks to add custom controls.
 class MetaBlockEdit extends Component {
 	componentDidMount() {
-		maybeUpdateFontSize( this.props );
+		if ( this.contentHasLoaded() ) {
+			maybeUpdateFontSize( this.props );
+		}
 	}
 
 	componentDidUpdate( prevProps ) {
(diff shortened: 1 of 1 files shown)


assets/src/stories-editor/components/with-meta-block-edit.js
```diff
@@ -64,6 +66,21 @@ class MetaBlockEdit extends Component {
...
+	 * @return {boolean} If content has loaded.
+	 */
+	contentHasLoaded() {
+		const { name, blockContent } = this.props;
+		if ( 'amp/amp-story-post-author' !== name ) {
+			return true;
+		}
+
+		const anonymousUserName = __( 'Anonymous', 'amp' );
+		return anonymousUserName.toLowerCase() !== blockContent.toLowerCase();
```
miina: Not sure if this might cause issues in some case, perhaps there's a better way to check if the content has loaded. Thoughts?

swissspidy: That seems a bit overly complicated.

Why not just return an `isLoading` prop or so in the `withSelect` part further down?
The component itself shouldn't be concerned with deciding that.

I suppose `isLoading` could mean either `getEditedPostAttribute()` or `getAuthors()` is returning `undefined`.

miina: Yep, that makes 100x more sense. 🤦‍♀

miina:
@swissspidy Would you mind reviewing again to get it to 1.2.1?

Issue #1815, opened 2019-01-16
Discovery: Improve CLI integration


swissspidy:
The plugin currently implements a pair of WP-CLI commands for AMP validation: `wp amp validate-site`  and `wp amp reset-site-validation`.

As a WP-CLI maintainer, there are a few things bugging me about this command. For example:

* The `AMP_CLI` class does not extend the `WP_CLI_Command` class as one should.
* Everything is `public static`, just for the tests. Instead, this functionality should be extracted from CLI and/or the CLI command should be tested using WP-CLI Behat tests
* It's not easy to add more CLI commands because everything is in one huge class
* Not following best practices like naming of commands (WP-CLI usually uses `wp <noun> <verb>`), adding debug statements, and others.

westonruter:
> * Not following best practices like naming of commands (WP-CLI usually uses `wp <noun> <verb>`)

Don't `wp amp validate-site` and `wp amp reset-site-validation` follow this practice?

swissspidy:
If you leave the `amp` part out of the picture, it's just two verbs, but they're actually about the same thing — validation. When you compare that to `wp post create` and `wp post delete`, something like `wp amp validation run` and `wp amp validation reset` would fit this more closely. This way you could extend this more easily too, e.g. with something like `wp amp validation info`.

swissspidy:
> The CLI command should be tested using WP-CLI Behat tests

See https://github.com/wearerequired/traduttore/pull/124 for precedence. It's not so easy at the moment so needs testing and perhaps even upstream changes.

swissspidy:
Sort of related: [AMP-Toolbox CLI](https://github.com/ampproject/amp-toolbox/tree/master/packages/cli)

schlessera:
There is duplication of business logic in the WP-CLI commands, like doing database requests and filtering them, for example.

I would suggest creating a repository for dealing with the validation custom post type and the validation custom taxonomy, so that the exact queries and their provenance is abstracted away and deduplicated. This repository would have the following benefits:
- Deduplication of business logic
- Abstraction decouples consuming code from actual implementation
- Easy to change technical details later on
- Easy to swap from CPT to custom DB table if needed
- Outside code can reuse the repository without requiring inside knowledge
- Performance optimizations can be added transparently after the fact

Is this something that makes sense, or are abstractions of that kind not wanted? If this makes sense, I'll create a new issue to discuss the actual interface.

schlessera:
[...]

PR #3056, opened 2019-08-20
See #1815

schlessera:
- Move them into their own subfolder `cli`.
- Separate into files per subcommand, with overarching namespace.
- Avoid static access and turn into actual objects, with dynamic access.
- Make classes final, as there are no plans for extensibility yet.
- Rename validation subcommands into `validation run` and `validation reset`.
- Some low-hanging fruit simplifications.
- Adapt tests for all of the above.


westonruter:
Changes look good to me! Only issue is that short array syntax should be used. Should be a quick fix with `phpcbf`.

https://travis-ci.org/ampproject/amp-wp/jobs/574484063#L767-L889

schlessera:
Fixed, so tests will probably pass now.

I think to remember though that there was a blog post by Gary Pendergast about the WPCS requiring `array()` notation because of preferred legibility...

westonruter:
> I think to remember though that there was a blog post by Gary Pendergast about the WPCS requiring `array()` notation because of preferred legibility...

Here's that post: https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/

> image

I'm not sure I agree. Parentheses are often abundant and square brackets make arrays more distinct. Also, brackets are very familiar as they are used in JS/JSON.

swissspidy:
Switching to short array syntax in this project was a deliberate choice a short while ago, and I don't think we should follow core's stance on it just because it's core. :-) 

-----

FWIW, plugin activation currently fails because of this:

```

Error: Callable "AMP_CLI_Validation" does not exist, and cannot be registered as `wp amp validation`.
```

This makes the E2E tests fail.

schlessera:
> I'm not sure I agree. Parentheses are often abundant and square brackets make arrays more distinct. Also, brackets are very familiar as they are used in JS/JSON.

Oh, I'm glad to use short array notation. Just wanted to raise that point, in case we want to stick with absolute Core standards adherence.

> FWIW, plugin activation currently fails because of this:

Sorry about that, with the back-and-forth of cherry-picking and a merge conflict, the rename of the command (`AMP_CLI_Validation` => `AMP_CLI_Validation_Command`) was stuck in a partial state. Fixed now.

schlessera:
Rebased and refreshed the lock file.

PR #3060, opened 2019-08-21
Snapping Guidelines

### LINKED ISSUE #2992: Snapping: Resizing blocks in relation to the page

### LINKED ISSUE #2993: Snapping: Moving blocks in relation to the page
See #2875 for details.

Example: if moving around the center, snapping should enable positioning exactly to the center of the page.

### LINKED ISSUE #2994: Snapping: Resizing blocks in relation to other blocks
See #2875 for details.

Example: when resizing a block it should be easy to make it the same size as other blocks, as indicated by snapping indicators.

### LINKED ISSUE #2995: Snapping: Moving blocks in relation to other blocks
See #2875 for details.

Example: when moving a block, it should be easy to align it on the same X or Y position as another block.

swissspidy:
## Summary

Fixes #2992 (resizing blocks in relation to the page).
Fixes #2994 (resizing blocks in relation to other blocks).
Fixes #2993 (moving blocks in relation to the page).
Fixes #2995 (moving blocks in relation to other blocks).

As previously discussed, this PR addresses the snapping _guidelines_ part, but not currently the actual snapping.

**Resizing**

* [ ] What if the resizing block is rotated?
* [x] What if a block other than the resizing block (aka one of its siblings) is rotated?
* [x] Clear snap lines when too far away from snap line? (would trigger another state change)
* [x] Allow disabling snapping when dragging, but still shown helper lines
* [x] Add snapping points for vertical and horizontal center too?
* [x] Snap line is flickering quite often
* [ ] ~Address block multi-selection interference~ see #3092.

**Dragging**

* [x] What if the dragging block is rotated?
* [x] What if a block other than the dragging block (aka one of its siblings) is rotated?
* [x] Clear snap lines when too far away from snap line? (would trigger another state change)
* [x] Make proof-of-concept implementation usable
* [x] Allow disabling snapping when dragging, but still shown helper lines
* [x] Prevent block from being stuck in snapping position

**General**
* [x] Tidy up code
* [x] Allow disabling snapping when rotating (unrelated, but seemed worthwhile)

## Checklist


swissspidy:
> Address block multi-selection interference

When I use `toggleSelection` in `onDragStart` it prevents dragging completely. So that doesn't sound like a viable option.

swissspidy:
Besides cleaning up the code there are basically two things left:

* Snapping when the to-be-resized block is rotated.
* Fixing the sluggish dragging

Especially the second one bothers me. I'm sure there's plenty of potential for performance improvements here. Any suggestions welcome.

miina:
Just tested again after the changes. The dragging lags less now, however, for some reason, it gets completely stuck quite often, seems to happen at snapping lines position. E.g. dragging a block and once it reaches a snapping line position such as in the middle of the page, the block "gets stuck" there:
stuck_snapping

Also, the snapping lines seem to be flickering a little bit sometimes. For example when doing the following:
- Resize the block so that it's right edge will match a snapping line of the Page.
- Move the mouse a tiny bit -- so that the snapping line match would remain (the resizing doesn't change the width of the block) -- quite often then the snapping line disappears and comes back again:
snapping

Are you also seeing these?

swissspidy:
@miina Yes. The being-stuck issue is an open todo, see above. Will add one for the flickering too.

-----

The following example use cases are not yet covered:

1. if the block is 200px from the left side then resizing the right size should allow snapping to 200px from the right side as well.
1. when resizing a block it should be easy to make it the same size as other blocks, as indicated by snapping indicators.

The second one currently only works when the blocks have the same x/y position. For making it work generally, a snap target needs to be dynamically added based on the siblings' widths.

Both snap cases only apply to resizing, not dragging. So this needs some special handling.

swissspidy:
> when resizing a block it should be easy to make it the same size as other blocks, as indicated by snapping indicators.

This one is bit of an exceptional case, as it means the snap target will be different from the indicators. Perhaps better to do implement that separately in the future as this isn't supported yet by the current logic.

swissspidy:
I'm currently out of ideas for fixing the being-stuck issue.

@miina Any chance you could have a quick look?

miina:
> @miina Any chance you could have a quick look?

Sure, I'll finish with functional testing for Page Attachment and then.

miina:
Added a potential fix for left position snapping, also added 0350b01 for "flickering" fix (for just horizontal right snapping for now).

Left the top positioning and "flickering" fixes for other snapping positions out for now, feel free to add those if this approach seems OK.

miina:
@googlebot I consent.

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -158,12 +162,19 @@ class Draggable extends Component {
...
+		// Where the original block would be positioned based on that.
+		const leftToCompareWith = initialBlockX + leftDiff;
+
 		if ( this.horizontalSnaps.includes( horizontalLeftSnap ) ) {
 			const snapLine = [ [ horizontalLeftSnap, 0 ], [ horizontalLeftSnap, STORY_PAGE_INNER_HEIGHT ] ];
 			newSnapLines.push( snapLine );
 
 			if ( snappingEnabled ) {
-				left = horizontalLeftSnap - rotatedWidthDiff;
+				if ( BLOCK_DRAGGING_SNAP_GAP > Math.abs( leftToCompareWith - horizontalLeftSnap ) ) {
```
miina: I guess these should actually be `>=`.

swissspidy:
@miina Awesome! This seems to do the trick.

swissspidy:
> What if the resizing block is rotated?

Still haven't gotten to that one but I am not sure if it's even worth it. What would be the use case for snapping a resized block?

miina:
> Still haven't gotten to that one but I am not sure if it's even worth it. What would be the use case for snapping a resized block?

You mean rotated, I assume, not resized.

Maybe if the block is rotated 90 or 180 degrees it makes the most sense?
Screenshot 2019-08-27 at 20 21 17

Also, it might be that the user wants to resize so that the block would be the same distance from the edges, in case of rotated blocks, too?

Also for matching a neighbor block to be exactly the same size, e.g.
Screenshot 2019-08-27 at 20 25 15

WDYT?

swissspidy:
Ah, I forgot about that template. That‘s a good example. Let‘s see if I can undust my geometry skills tomorrow 😄

swissspidy:
Snap lines are now displayed when resizing a rotated block.

What I don't like is that for doing the actual snapping in `ResizableBox`'s `onResize`, I basically have to do everything twice (rotated and non-rotated). @miina Can we combine these two cases somehow in `getBlockPositioning`? This would greatly simplify things.

Another thing I've noticed: When resizing a block, the text is not centered anymore in the block unless I reload the page or edit the block:

Screenshot 2019-08-28 at 13 36 05

But perhaps that's also already an issue in `develop`.

miina:
> But perhaps that's also already an issue in develop.

Yep, looks like it, added an issue: #3113 

> What I don't like is that for doing the actual snapping in ResizableBox's onResize, I basically have to do everything twice (rotated and non-rotated).

In theory, the condition of `if ( ! angle )` shouldn't be necessary at all since the rotated logic should work with `0` degrees exactly the same way (needs testing out just in case). It's just making a bit more calculations, not a huge difference anyway really, you could probably just remove this:
https://github.com/ampproject/amp-wp/blob/0caf34f1126a126ff8b05aa1aa1e38331dfcf197/assets/src/stories-editor/components/resizable-box/index.js#L207-L217

Would that help with doing everything twice?

barklund:
@googlebot I consent.

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -161,10 +161,7 @@ class Draggable extends Component {
 		}
 
 		if ( newSnapLines.length ) {
-			const hasSnapLine = ( item ) => snapLines.find( ( snapLine ) => isShallowEqual( item[ 0 ], snapLine[ 0 ] ) && isShallowEqual( item[ 1 ], snapLine[ 1 ] ) );
-			if ( ! newSnapLines.every( hasSnapLine ) ) {
-				setSnapLines( ...newSnapLines );
-			}
+			setSnapLines( newSnapLines );
```
barklund: All these minor optimizations are completely irrelevant at this point. Removed them all to just keep the most obvious approach.

assets/src/stories-editor/components/contexts/Snapping.js
```diff
@@ -0,0 +1,98 @@
+/**
```
barklund: I don't know if this is the right place for this component. Maybe it could get its own folder in stead of grouping it with future contexts? It might make sense to split the HoC from this component for readability.

assets/src/stories-editor/components/block-mover/draggable.js
swissspidy: Makes sense!

assets/src/stories-editor/components/contexts/Snapping.js
swissspidy: Don't have a strong opinion on that. Since this is the first and only context for now there's no urgency in organizing the code differently.

swissspidy:
Looks like there was some regression in this PR at some point, as now the snap lines for a rotated block are displayed as if the block wasn't rotated — or perhaps there is just wrong line being drawn:

Screenshot 2019-09-17 at 15 28 33 (3)

tests/shared/test-utils/wp-element-mock.js
```diff
@@ -0,0 +1,112 @@
+/**
```
barklund: This goes in a new `shared` folder, as it will also be needed for the integration tests, that I'm working on elsewhere.

assets/src/stories-editor/components/resizable-box/index.js
```diff
@@ -215,40 +249,84 @@ class EnhancedResizableBox extends Component {
 						lastDeltaW = deltaW;
 					}
 
-					if ( ! angle ) {
-						// If the resizing is to left or top then we have to compensate
-						if ( REVERSE_WIDTH_CALCULATIONS.includes( direction ) ) {
-							const leftInPx = getPixelsFromPercentage( 'x', parseFloat( blockElementLeft ) );
-							blockElement.style.left = getPercentageFromPixels( 'x', leftInPx - lastDeltaW ) + '%';
+					const dimensions = blockElement.getBoundingClientRect();
```
miina: Hmm, what if the block is rotated?
`blockElement = element.closest( '.wp-block' ).parentNode;` meaning that it's not considering the rotation since the parent node isn't rotated, only `.wp-block` is. If that's the case then this would mean that `actualRight` and `actualLeft` would probably be incorrect. Thoughts?

assets/src/stories-editor/components/resizable-box/index.js
```diff
@@ -215,40 +249,84 @@ class EnhancedResizableBox extends Component {
...
+					actualRight -= parentBlockOffsetLeft;
+					actualBottom -= parentBlockOffsetTop;
+					actualLeft -= parentBlockOffsetLeft;
+
+					const horizontalCenter = actualLeft + ( ( actualRight - actualLeft ) / 2 );
+					const verticalCenter = actualTop + ( ( actualBottom - actualTop ) / 2 );
+
+					const snappingEnabled = ! event.getModifierState( 'Alt' );
+
+					if ( snappingEnabled ) {
```
miina: Could this be moved to a helper instead since it seems to be an exact duplicate of the logic of populating the `newSnapLines` in `draggable.js`? Perhaps there was a reason to keep it separately in both, mentioning just in case anyway.

swissspidy: It was originally separated because this section included also the actual snapping (which was different for the two scenarios), and not just adding the guidelines to the array. But I guess this can be just as well done _after_ this section.

spacedmonkey:
This PR should be rebased. After rebase, we need to make sure AXE accessibility tests still pass.

assets/src/stories-editor/components/resizable-box/index.js
swissspidy: Fixed in 9e42965816ec3a77e4b1469983b2290f2104ccef.

miina:
From functional testing.
Looks like there is something odd happening with the lines:
glitching

This kind of flickering seems to happen for a rotated block only. Are you also seeing this?

miina:
Perhaps the CTA button should also have a snapping line for centering?

swissspidy:
> This kind of flickering seems to happen for a rotated block only. Are you also seeing this?

Yes indeed, which seems very odd. Not sure yet why that is.

miina:
> @miina For the vertical centering there already is a snapping line. Did you mean for the horizontal middle of the whole CTA area (lower part of the page)?

Hmm, I actually didn't see the vertical centering line -- did mean that. Will check again soon.

assets/src/stories-editor/components/resizable-box/index.js
```diff
@@ -13,10 +13,11 @@ import { ResizableBox } from '@wordpress/components';
 /**
  * Internal dependencies
  */
+import withSnapTargets from '../higher-order/with-snap-targets';
 import './edit.css';
 import {
 	getPercentageFromPixels,
-	getPixelsFromPercentage,
+	findClosestSnap, getRelativeElementPosition,
```
barklund:
```suggestion
	findClosestSnap,
	getRelativeElementPosition,
```

barklund:
I've had another look at this, and it looks good with all the most recent optimisations.

I have some comments/suggestions. I don't know if they're blockers or just nice-to-have's.

Two comments regarding available snap lines:

- No snap lines are calculated for the center of other objects. I can align the center of the dragged object to something and see a snap line, but I can't see if this element aligns (on any edge/center) with the **center of another element**.
- Snap lines aren't calculated at all for CTA buttons. They might be calculated for the overall CTA block, but as it's in a fixed position it's not that relevant. No snap lines are calculated for the button inside the CTA block, when moving other elements on the page and trying to align them to the CTA button.

Two comments regarding UX:

- I often see more than one snap line at the same time when there are many objects on the page. I don't think this is normal UX. In most other programs there's always only 1 snap line visible at a time (which has its own pros and cons, but it seems to be the established norm).
- I can't see which element I'm snapping to. It's often not clear if a snap line is related to an object or e.g. the page center. This is done in different ways in other programs - e.g. by only having the snap lines extend to far edge of the object you're snapping to - so only if it's the page itself, the line will be full-height. Another option is to highlight the "snapped" element with a border or slight box shadow. This latter issue is of course made even more complex by the right one above.

swissspidy:
Thanks, that's some very good feedback! I'll try to wrap my head around it today.

swissspidy:
> No snap lines are calculated for the button inside the CTA block, when moving other elements on the page and trying to align them to the CTA button.

This was actually a bug, fixed in eaef3d4a826dd4f822bb5a56a6d9f8f0fce5fb2f

----

The other points have been addressed in 1198cbb3b4de8b08961d240379e1857fe5ea89c1.

assets/src/stories-editor/components/higher-order/with-snap-targets.js
```diff
@@ -0,0 +1,183 @@
...
+
+	const parentBlockElement = getBlockInnerElement( getBlock( parentBlock ) );
+
+	if ( ! parentBlockElement ) {
+		return defaultData;
+	}
+
+	const siblings = getBlocksByClientId( getBlockOrder( parentBlock ) )
+		.filter( ( { clientId: blockId } ) => blockId !== clientId )
+		.filter( getBlockInnerElement );
```
barklund: You extract the inner element in the loop anyway, so this could be extended to:

```suggestion
		.map( getBlockInnerElement )
		.filter( Boolean );
```

And then simplify lines 115-116 and 151-152.

swissspidy: True. I think I already did that at some point, not sure why I didn't keep it 🤔

swissspidy: Fixed in dc8f7d18c58e887881a0cb405b14a48a7928771c

assets/src/stories-editor/components/higher-order/with-snap-targets.js
```diff
@@ -0,0 +1,182 @@
...
+				// Bottom page border.
+				[ STORY_PAGE_INNER_HEIGHT ]: [ getHorizontalLine( STORY_PAGE_INNER_HEIGHT ) ],
+			},
+			{
+				set: getSetter( 0, STORY_PAGE_INNER_HEIGHT ),
+			} );
+
+			for ( const blockElement of siblings ) {
+				const { top, right, bottom, left } = getRelativeElementPosition( blockElement, parentBlockElement );
+				const center = top + ( ( bottom - top ) / 2 );
```
barklund: Less obvious, but simpler - it is just finding the average 🤷‍♂

```suggestion
				const center = ( top + bottom ) / 2;
```

assets/src/stories-editor/components/higher-order/with-snap-targets.js
```diff
@@ -0,0 +1,182 @@
...
+
+				const start = targetLeft < left ? targetLeft : left;
```
barklund: Use `Math.min`:

```suggestion
				const start = Math.min( targetLeft, left );
```

assets/src/stories-editor/components/higher-order/with-snap-targets.js
swissspidy: Not sure what I was thinking there 🤔 

Will update all the occurrences.

assets/src/stories-editor/components/higher-order/with-snap-targets.js
swissspidy: so obvious in hindsight 🤦‍♂ Thanks!

assets/src/stories-editor/components/contexts/snapping.js
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Internal dependencies
+ */
+import { STORY_PAGE_INNER_HEIGHT, STORY_PAGE_INNER_WIDTH } from '../../constants';
+
+const SnapContext = createContext();
+
+const Snapping = ( { children } ) => {
+	const [ snapLines, setSnapLines ] = useState( [] );
+	const [ hasSnapLines, setHasSnapLines ] = useState( false );
```
barklund: Thinking a bit more about it, is this variable even necessary? Can't we stick to the array only and if empty, show nothing? We always `hide` and `clear` at the same time anyway I think. Is there any occasion where we don't show the lines but remember them anyway?

swissspidy: Hmm I can't think of any such situation at the moment.

One thing that comes to mind: When we implement the actual snapping, we _could_ still show the snapping guidelines, but disable the actual snapping when pressing <kbd>Alt</kbd>. But that does not affect this case here.

So I think it's fine to remove.

swissspidy: Removed it now

barklund: Perfect, that makes the API a lot cleaner.

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -75,8 +89,81 @@ class Draggable extends Component {
...
+		} = this.props;
+
 		const top = parseInt( this.cloneWrapper.style.top ) + event.clientY - this.cursorTop;
+		const left = parseInt( this.cloneWrapper.style.left ) + event.clientX - this.cursorLeft;
+
+		if ( top === lastY && left === lastX ) {
+			return;
+		}
+
+		const isCTABlock = 'amp/amp-story-cta' === blockName;
```
spacedmonkey: There is a helper for this in https://github.com/ampproject/amp-wp/pull/3312

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -75,8 +89,81 @@ class Draggable extends Component {
...
+			bottom: actualBottom,
+			left: actualLeft,
+		} = getRelativeElementPosition( blockElement, parentBlockElement );
+
+		const horizontalCenter = ( actualRight + actualLeft ) / 2;
+		const verticalCenter = ( actualTop + actualBottom ) / 2;
+
+		const newSnapLines = [];
+
+		const snappingEnabled = ! event.getModifierState( 'Alt' );
```
spacedmonkey: Why have a button that disables the snaps? Is this documented anywhere?

spacedmonkey:
Should they be a way to see snap lines with moving a block? Maybe with some kind of keyboard shortcut. Maybe holding the ctrl key when a block should reveal the snaps lines when a block is selected. This should maybe followed up in another ticket.

swissspidy:
> Should they be a way to see snap lines with moving a block? Maybe with some kind of keyboard shortcut. Maybe holding the ctrl key when a block should reveal the snaps lines when a block is selected. This should maybe followed up in another ticket.

This PR by default displays snap lines both when dragging and resizing a block.

Snapping guidelines (and the actual snapping once implemented) will be **disabled** when pressing the <kbd>Alt</kbd> / <kbd>Option</kbd> key.

assets/src/stories-editor/helpers/getSnapCalculatorByDimension.js
```diff
@@ -20,7 +54,7 @@ import createSnapList from './createSnapList';
  * @param {number} maxValue Maximum value of page in this direction.
  * @param {Array.<number>} primary Coordinate property names in primary direction.
  * @param {Array.<number>} secondary Coordinate property names in secondary direction.
- * @return {Function} Function mapping an actual object's position to all possible snap targets.
+ * @return {SnapTargetsEnhancer} Function mapping an actual object's position to all possible snap targets.
```
barklund: Nice! 🎉

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -75,8 +89,81 @@ class Draggable extends Component {
...
+
+		if ( snappingEnabled ) {
+			// Go through all snap targets and find the one that is closest.
+			const findSnap = ( snapKeys, ...values ) => {
+				return values
+					.map( ( value ) => {
+						const snap = findClosestSnap( value, snapKeys, BLOCK_DRAGGING_SNAP_GAP );
+						return [ value, snap ];
+					} )
+					.filter( ( arr ) => arr[ 1 ] !== null )
```
barklund: What does this filter do again? When would it be `null`?

assets/src/stories-editor/components/block-mover/draggable.js
swissspidy: Will rewrite once that PR is merged 👍

assets/src/stories-editor/components/block-mover/draggable.js
barklund: Ah of course, if there is no closest snap to the given value, it will be null.

swissspidy: `arr[ 1 ]` is the `snap` variable from the `[ value, snap ]` array above, i.e. the result of `findClosestSnap`. It is `null` when no snap target was found for the given `value`.

The thinking here was that I need a map of value -> snap target to find the most suitable snap target. However, when I look at this code again now it doesn't even use `arr[ 0 ]` 🤔

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -75,8 +89,81 @@ class Draggable extends Component {
...
+
+		const snappingEnabled = ! event.getModifierState( 'Alt' );
+
+		if ( snappingEnabled ) {
+			// Go through all snap targets and find the one that is closest.
+			const findSnap = ( snapKeys, ...values ) => {
+				return values
+					.map( ( value ) => {
+						const snap = findClosestSnap( value, snapKeys, BLOCK_DRAGGING_SNAP_GAP );
+						return [ value, snap ];
```
barklund: Hm, why are you returning the value at all here? You're only ever using `a[ 1 ]` below, aren't you? Also, this entire logic is duplicated here and in `resisable`. Should be refactored out of here and tested. It'll take a look at it.

swissspidy: Yeah that's what I just noticed while writing https://github.com/ampproject/amp-wp/pull/3060#discussion_r331541948.

miina:
Two notes from functional testing, both can be for future consideration if not blockers right now.

1. In some cases the snapping lines feel a little bit confusing, that's especially the case if the lines are shown related to another block which doesn't have a background color, for example:
Screenshot 2019-10-07 at 18 43 18

In this case the Title block is quite large but has white background and it's quite confusing to understand what are the snapping lines for. Perhaps it would be a good idea to display the border around the block that's being snapped against?

2. When the background is red or some other similar color then the snapping lines aren't very visible and strain the eye a bit. AI could be helpful here perhaps.

swissspidy:
@barklund I'll create one after merging this. We need to flesh out the acceptance criteria there carefully as the last version in this PR was not 100% satisfying.

@miina Some very good points! We should definitely follow up on these.

For the first one, Morten mentioned something similar in https://github.com/ampproject/amp-wp/pull/3060#issuecomment-537714742::

> I can't see which element I'm snapping to. It's often not clear if a snap line is related to an object or e.g. the page center. This is done in different ways in other programs - e.g. by only having the snap lines extend to far edge of the object you're snapping to - so only if it's the page itself, the line will be full-height. Another option is to highlight the "snapped" element with a border or slight box shadow. This latter issue is of course made even more complex by the right one above.

I implemented the first idea (only having the snap lines extend to far edge of the object you're snapping to), but not the second one with highlighting the snapped element. I don't know any program that does this (just like adapting the snap line color), so it felt premature to add it.

miina:
Hmm, actually, seeing some other oddities, could you confirm if you also see this?
1. It seems to start lagging after adding a few blocks, and generally behaves so that the usability is not the best, it became not possible to drag.
dragging

2. CTA block is showing tons of lines and it's quite hard to understand which line is the correct line for centering, maybe for the CTA block only the Page limits + the centering line would make sense for now? Perhaps other blocks are not that relevant?
cta-lines

barklund:
> 1. It seems to start lagging after adding a few blocks, and generally behaves so that the usability is not the best, it became not possible to drag.

I have seen that for a while actually. It's in develop too and have been for about a week - I haven't tracked it back to where it originally comes from.

But when there's multiple elements on the page, you can't select-and-drag in one click (not always at least). You have to click to select first, then drag.

I wasn't sure if it was a bug locally for me only as nobody else reported it though, but it seems it may not be.

> 2. CTA block is showing tons of lines and it's quite hard to understand which line is the correct line for centering, maybe for the CTA block only the Page limits + the centering line would make sense for now? Perhaps other blocks are not that relevant?

Hm, the opacity definitely makes it harder to see what's being snapped to at least. Tough one. UX?

swissspidy:
Created #3458 for the lagging issue.

As for the CTA block: could we change it so that the lines are shown on top? Definitely shouldn't be hidden like that. Regarding the amount of snap lines it would be good to get some real world testing I think. It's not just for the CTA block where we need to find a good balance.

Issue #2315, opened 2019-05-14
Extract post-processor logic into CMS-agnostic PHP library


westonruter:
There is need in Drupal, Joomla!, Typo3 among other PHP-based CMSes for an official library to do what the AMP WordPress plugin does in its post-processor phase. The Drupal AMP module is currently using [`amp-library`](https://github.com/Lullabot/amp-library/) for this purpose. but it is very out of date and it needs to be overhauled (see https://github.com/Lullabot/amp-library/issues/231). A library which is maintained by the AMP project is needed to ensure it is kept up to date.

Such a library should include:

* Converting HTML elements to their equivalent AMPHTML ones (e.g. `img` to `amp-img`).
* Processing stylesheets (concatenating, minifying, tree-shaking).
* Removing invalid markup and reporting what was removed. 

This would essentially be extracting the entirety of the [`AMP_Theme_Support::prepare_response()`](https://github.com/ampproject/amp-wp/blob/2984fc9315bbd7c430598c9c75c0148b6527fe27/includes/class-amp-theme-support.php#L1663-L2050) method, minus the parts that are WordPress-specific. The extracted functionality would need to be configurable to plug in to any CMS, including the way it handles script dependencies and object caching.

As part of this effort, we should take the opportunity to refactor the sanitizers to clean up some legacy cruft and embrace a more modern PHP 5.6 world (which is WordPress's new minimum version). Likewise, the [`AMP_Theme_Support::ensure_required_markup()`](https://github.com/ampproject/amp-wp/blob/2984fc9315bbd7c430598c9c75c0148b6527fe27/includes/class-amp-theme-support.php#L1298-L1566) should be converted into a sanitizer, and a new sanitizer for implementing optimized/transformed AMP can be introduced (#958).

While not part of the post-processor phase, there would also be value in extracting the the oEmbed handling, defining a mapping of oEmbed URLs to how they should be converted into AMP.

bmack:
Hi Weston,

I think we should go with a "amp-php-migrator-engine" package or so. This could be ideally prepared for PHP 7+ but still support PHP 5.6 for WP - for now I'm fine coding the PHP 5.6-way.

We could then create "adapters" for each CMS as separate packages IMHO.

Do you have more docs on the amp-wp package, otherwise I can start migrating and look into a proper API so WP works with it.

westonruter:
[...]

PR #3064, opened 2019-08-21
Normalize frameborder=no|yes to frameborder=0|1 in iframe sanitizer

### LINKED ISSUE #2335: Normalize frameborder=no|yes to frameborder=0|1 in iframe sanitizer
When adding an `iframe` such as:

```html
<iframe frameborder="no" height="200" scrolling="no" src="https://player.megaphone.fm/DEM1119155060?" width="100%"></iframe>
```

It gets converted into an `amp-iframe` successfully as:

```html
<amp-iframe frameborder="0" height="200" scrolling="no" src="https://player.megaphone.fm/DEM1119155060?" width="640" sandbox="allow-scripts allow-same-origin" layout="intrinsic" class="amp-wp-enforced-sizes">
    <span placeholder="" class="amp-wp-iframe-placeholder"></span>
    <noscript>
        <iframe height="200" scrolling="no" src="https://player.megaphone.fm/DEM1119155060?" width="100%"></iframe>
    </noscript>
</amp-iframe>
```

Nevertheless, one AMP validation error is raised in the process:

image

This validation error is pointless, however, because the `iframe` that `amp-iframe` component generates already provides `frameborder=0`:

image

It seems that while HTML allows `no` as a synonym for `0` (and `yes` for `1`), this is [not supported by AMP](https://github.com/ampproject/amphtml/blob/bf504ab750f64b9343c228a1756d38d95bbba7c8/extensions/amp-iframe/validator-amp-iframe.protoascii#L46-L50):

```
  attrs: {
    name: "frameborder"
    value: "0"
    value: "1"
  }
```

Apparently this is [not even allowed in HTML4](https://www.w3.org/TR/html4/present/frames.html#h-16.5), so it could just be a common author mistake (which browsers gracefully handle, and it has been [deprecated/removed in HTML5](https://www.w3.org/TR/2010/WD-html5-20100304/obsolete.html#non-conforming-features)):

```
  frameborder (1|0)          1         -- request frame borders? --
```

The AMP plugin makes the default `frameborder=0` as seen in the generated example above. What is missing is when the `$value` is `no` it should be normalized to `0`, and when `yes` normalized to `1`:

https://github.com/ampproject/amp-wp/blob/437c335b84106a8099fb3929c8d6faeaa6d7fb7b/includes/sanitizers/class-amp-iframe-sanitizer.php#L153-L158

This will prevent a validation error from being raised, and it will allow `iframe[frameborder=yes]` to actually result in `amp-iframe[frameborder=1]` as the user intends.

schlessera:
Fixes #2335

The logic to normalize the frameborder attribute is based on the text from [https://html.spec.whatwg.org/#frames-and-framesets](https://html.spec.whatwg.org/#frames-and-framesets):

> A frameset or frame element has a border if the following algorithm returns true:
> 
> If the element has a frameborder attribute whose value is not the empty string and whose first character is either a U+0031 DIGIT ONE (1) character, a U+0079 LATIN SMALL LETTER Y character (y), or a U+0059 LATIN CAPITAL LETTER Y character (Y), then return true.

Open questions:

- [x] The actual normalization logic is a method called `sanitize_boolean_digit()` in the `AMP_Base_Sanitizer` parent class, to mirror the behavior of `sanitize_dimension()` which is in there, too. Should the base class regroup such logic that might be needed across sanitizer? Or would it make sense to extract them into either separate helper objects or traits?
- [x] The name of the method is `sanitize_boolean_digit()`... is there a name that is clearer here? Maybe an official HTML "attribute type" name I don't know about?

### DIFF
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -255,6 +255,34 @@ abstract class AMP_Base_Sanitizer {
 	}
 
 	/**
+	 * Sanitizes a boolean character (or string) into a '0' or '1' character.
+	 *
+	 * @param string $value A boolean character to sanitize. If a string containing more than a single
+	 *                      character is provided, only the first character is taken into account.
+	 *
+	 * @return string Returns either '0' or '1'.
+	 */
+	public function sanitize_boolean_digit( $value ) {
+
+		if ( empty( $value ) ) {
+			return '0';
+		}
+
+		if ( ! is_string( $value ) && ! is_numeric( $value ) ) {
+			return '0';
+		}
+
+		switch ( substr( (string) $value, 0, 1 ) ) {
+			case '1':
+			case 'y':
+			case 'Y':
+				return '1';
+		}
+
+		return '0';
+	}
+
+	/**
 	 * Sets the layout, and possibly the 'height' and 'width' attributes.
 	 *
 	 * @param string[] $attributes {
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -264,14 +264,17 @@ abstract class AMP_Base_Sanitizer {
 	 */
 	public function sanitize_boolean_digit( $value ) {
 
+		// Default to false if the value was forgotten.
 		if ( empty( $value ) ) {
 			return '0';
 		}
 
+		// Default to false if the value has an unexpected type.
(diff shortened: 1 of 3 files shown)


westonruter:
@schlessera Since `sanitize_boolean_digit()` is only being used in the `AMP_Iframe_Sanitizer`, perhaps best to just make it a `private` method on that class for now? Figuring out the best place for these methods and whether traits should be used will be part of the larger refactoring effort in #2315.

PR #3065, opened 2019-08-21
Add Roave Security Advisories package as develoment dependency

### LINKED ISSUE #3053: Add Roave Security Advisories package to Composer dependencies
I recommend adding the [Roave Security Advisories](https://github.com/Roave/SecurityAdvisories) package as a development dependency.

It adds additional constraints to a Composer project that exclude any versions of packages that are known to have security vulnerabilities. This is an additional security precaution that is generally recommended for any Composer-based project.

schlessera:
Fixes #3053

Adds [Roave Security Advisories](https://github.com/Roave/SecurityAdvisories) as a development dependency, which adds constraints to Composer that automatically exclude any version of particular packages that are known to have security vulnerabilities.


schlessera:
I opted to do this immediately as it was a 2-minute change and is security-related. I hope that was ok.

PR #3075, opened 2019-08-22
Adjust block inserter style.

### LINKED ISSUE #3062: Global Inserter looks broken on upstream develop branch
Looks like some changes to the inserter style were implemented which look broken on AMP Stories:

Screenshot 2019-08-21 at 18 03 09

This change hasn't been released yet but we'll likely need to address this thinking of the release.

miina:
Fixes #3062.

Before:
Screenshot 2019-08-21 at 18 03 09

After:
Screenshot 2019-08-22 at 10 35 21

### DIFF
--- a/assets/src/stories-editor/components/inserter/edit.css
+++ b/assets/src/stories-editor/components/inserter/edit.css
@@ -1,11 +1,3 @@
 .amp__block-editor-inserter__menu.block-editor-inserter__menu {
-	display: block;
-}
-
-.amp__block-editor-inserter__search.block-editor-inserter__search {
-	width: 90%;
-}
-
-.amp__block-editor-inserter__results.block-editor-inserter__results {
-	height: 360px;
+	flex-direction: column;
 }
--- a/assets/src/stories-editor/components/inserter/edit.css
+++ b/assets/src/stories-editor/components/inserter/edit.css
@@ -1,0 +1,11 @@
+.amp__block-editor-inserter__menu.block-editor-inserter__menu {
+    display: block;
+}
+
+.amp__block-editor-inserter__search.block-editor-inserter__search {
+    width: 90%;
+}
+
+.amp__block-editor-inserter__results.block-editor-inserter__results {
+    height: 360px;
+}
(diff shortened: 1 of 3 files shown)


swissspidy:
I wonder which Gutenberg change caused this. I only found https://github.com/WordPress/gutenberg/pull/16813 so far that could be relevant. That PR also makes me wonder how/if we're going to implement that same feature in the future... 🤔

assets/src/stories-editor/components/inserter/edit.css
```diff
@@ -0,0 +1,11 @@
+.amp__block-editor-inserter__menu.block-editor-inserter__menu {
```
swissspidy: Upstream seems to use flexbox for this and the results. Does it make much of a difference?

https://github.com/WordPress/gutenberg/blob/47dfd82ce25e18458d6349cf9c40d94f8fdf2c9f/packages/block-editor/src/components/inserter/style.scss#L38-L41
https://github.com/WordPress/gutenberg/blob/47dfd82ce25e18458d6349cf9c40d94f8fdf2c9f/packages/block-editor/src/components/inserter/style.scss#L91-L110

assets/src/stories-editor/components/inserter/edit.css
```diff
@@ -0,0 +1,11 @@
...
+    display: block;
```
swissspidy: Shouldn't these be tabs? 🤔

miina:
> I wonder which Gutenberg change caused this. I only found WordPress/gutenberg#16813 so far that could be relevant. That PR also makes me wonder how/if we're going to implement that same feature in the future...

Yep, that's the PR that caused the change.

I'm thinking that since right now we will release before the new feature comes out then the Inserter should look OK in both cases, that's why maybe it wouldn't make sense to move over to the new style and markup right now fully. However, for 1.2.2 we could apply the changes, we could also then consider changing the classnames so that we wouldn't be influenced by upstream changes. Thoughts?

assets/src/stories-editor/components/inserter/edit.css
miina: Hmm, will try an alternative with flexbox.

Issue #1316, opened 2018-08-04
Add support for the <picture> element, converting to <amp-img>


westonruter:
The `AMP_Img_Sanitizer` needs to be extended to add support for the `<picture>` element. For example, on https://fivethirtyeight.com/features/the-eternal-question-how-much-do-these-apricots-weigh/ there is:

```html
<picture class="featured-picture">
	<source media="(min-width: 768px)" srcset="https://fivethirtyeight.com/wp-content/uploads/2016/07/riddler_4x3_default.gif?strip=info&amp;ssl=1">
	<source srcset="https://fivethirtyeight.com/wp-content/uploads/2016/07/riddler_4x3_default.gif?strip=info&amp;ssl=1">
	<img alt="riddler_4x3_default" src="https://fivethirtyeight.com/wp-content/uploads/2016/07/riddler_4x3_default.gif?strip=info&#038;ssl=1" />
</picture>
```

I'm not sure in this case why there are multiple `source` elements that have the same URL. But normally a `type` would be defined on the `source` so that the user agent can skip a `source` that is not supported. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture

This should be converted into the appropriate `amp-img`. Note that `amp-img` allows for fallbacks to be defined as children. So that is how `amp-img` can be used to implement multiple `source`s: https://www.ampproject.org/docs/reference/components/amp-img#example:-specifying-a-fallback-image

The `picture` element allows for an `img` to be nested inside of it as a fallback, so it's not critical that the plugin recognize a `picture` to work.

schlessera:
According to the [docs](https://amp.dev/documentation/guides-and-tutorials/develop/style_and_layout/control_layout/?format=websites), the `<amp-img>` element does directly support a `srcset` attribute, so I assume it is not needed to nest `<amp-img>` elements to simulate the `<picture>` element.

schlessera:
So far I assume the expected output should be as follows...
1.)
```html
<picture><img></picture>
=> <amp-img><noscript><img></noscript></amp-img>
```
2.)
```html
<picture><source srcset /><img></picture>
=> <amp-img srcset><noscript><img></noscript></amp-img>
```
3.)
```html
<picture><source srcset /><source srcset /><img></picture>
=> <amp-img srcset><amp-img srcset><noscript><img></noscript></amp-img></amp-img>
```

schlessera:
[...]

PR #3082, opened 2019-08-23
WIP - Add support for <picture> element

schlessera:
> This is still a partial draft only, I'm pushing to make sure the code won't be lost.

See #1316

### DIFF
--- a/includes/class-amp-autoloader.php
+++ b/includes/class-amp-autoloader.php
@@ -87,6 +87,7 @@ class AMP_Autoloader {
 		'AMP_Video_Sanitizer'                => 'includes/sanitizers/class-amp-video-sanitizer',
 		'AMP_Core_Theme_Sanitizer'           => 'includes/sanitizers/class-amp-core-theme-sanitizer',
 		'AMP_Noscript_Fallback'              => 'includes/sanitizers/trait-amp-noscript-fallback',
+		'AMP_Picture_Sanitizer'              => 'includes/sanitizers/class-amp-picture-sanitizer',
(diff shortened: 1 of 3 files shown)


schlessera:
Still reading through specs and docs to figure out what the expected end result should actually be.

From what I can gather, the nested syntax only triggers the fallback when the browser fails to load the image (unsupported format, for example). So, to map multiple media queries in `<source>` tags, this is not usable.

My current best guess is the following mapping:
```
<picture>
    <source media="(min-width: 1201px)" srcset="[image large] 1x">
    <source media="(min-width: 993px)" srcset="[image medium] 1x">
    <source media="(min-width: 601px)" srcset="[image small] 1x">
    <img src="[image small]" alt="some image">
</picture>
```
   vvv
```
<amp-img media="(min-width: 1201px)" src="[image large] 1x" layout="responsive" alt="some image">
<amp-img media="(min-width: 993px) and (max-width: 1200px)" src="[image medium] 1x" layout="responsive" alt="some image">
<amp-img media="(min-width: 601px) and (max-width: 992px)" src="[image small] 1x" layout="responsive" alt="some image">
<amp-img media="(max-width: 600px)" src="[image small]" alt="some image">
```

The `<amp-img>` elements are siblings and don't exclude each other like the `<source>` elements do. So, this would in effect mean we need to interpret the media queries and make transform them into mutually exclusive conditions.

Is that correct? If so, then the code to do this transformation will be hugely complex...

schlessera:
Alright. Will you collect feedback internally, or shall I contact people to try to find something out?

Maybe there's something obvious I'm missing, but right now this looks like a problem we cannot easily solve for 100% of cases.

Issue #2990, opened 2019-08-08
Right-Click: Display block settings menu


swissspidy:
Following discovery #2785, let's intercept right-click attempts to display the `BlockSettingsMenu`. This can be done by filtering `editor.BlockEdit` to wrap all blocks in a new container where we can add this event listener.

Note: Considerations need to be put into accessibility, making sure that keyboard navigation and focus handling works as expected.

miina:
Some thoughts so far after testing this out:
- Since there can be only one right-click menu open at a time then it seems to make sense to have just one general menu which is updated based on the block which was clicked on.
- It would close when anywhere else is being clicked.
- Be replaced on each right-click, even if it's the same block, since the position would change depending on the click position.

I guess we would keep the default menu in the Toolbar as well, and add the more relevant buttons to the new menu? We'll probably need a custom component for that anyway since it looks like `BlockSettingsMenu` can't be added opened by default. Which options should we have in the menu?
- Remove block;
- Copy;
- Cut;
- Duplicate;

Thoughts?

cc @swissspidy

swissspidy:
That sounds like a good summary and first step! 👍

csossi:
Verified in QA

spacedmonkey:
Screencast can be found here - https://youtu.be/95GjenV63i8

Issue #2989, opened 2019-08-08
Right-Click: Add options to cut/copy/paste blocks


swissspidy:
As outlined in #2785, let's extend the block settings menu to offer cut + copy options, where the block content is added to the clipboard.

Cases where Gutenberg is adding blocks to the clipboard:

https://github.com/WordPress/gutenberg/blob/b95e865de32ee62d722be3e3e71b334a4b639032/packages/block-editor/src/components/copy-handler/index.js#L26-L57

https://github.com/WordPress/gutenberg/blob/b95e865de32ee62d722be3e3e71b334a4b639032/packages/components/src/clipboard-button/index.js#L27-L67

Pasting would be handled by the `RichText` component which we use in all the text-based blocks and has some complex logic for handling the clipboard data properly:

https://github.com/WordPress/gutenberg/blob/03121c5e59c63d4c79a293f71bbb7b9613fd311c/packages/rich-text/src/component/index.js#L187-L260

We probably need to copy most of that component's logic.

westonruter:
Special care will be needed for paste since access is restricted for security/privacy reasons. See https://developers.google.com/web/updates/2018/03/clipboardapi

csossi:
Verified in QA

Issue #2538, opened 2019-06-06
Use React Hooks where appropriate


swissspidy:
In the latest version of Gutenberg, a bunch of handy [React hooks](https://reactjs.org/docs/hooks-intro.html) have been added, most notably `useSelect` and `useDispatch`. ([see documentation](https://github.com/WordPress/gutenberg/tree/2564c2122f75541cbdf4fcfdebdfc839dadeeb7a/packages/data)). It also exposes existing hooks like `useState`, for example.

Specifically for AMP Stories we have added many higher-order components that over time can make code quite hard to read. We also heavily leverage the pre-existing `withSelect` and `withDispatch` HOCs.

With React hooks we have the opportunity to greatly simplify many of our custom React components to make them easier to comprehend and maintain in the long term.

Note: `withSelect` internally already uses that new `useSelect` hook, so switching to hooks directly would remove an unnecessary layer in-between.

swissspidy:
* We currently require Gutenberg 5.8 or higher
* React hooks are available since Gutenberg 5.3
* `useSelect` has been introduced in Gutenberg 5.9
* A bug with scalar return types values in `useSelect` has been fixed in Gutenberg 6.3

swissspidy:
Quick update: #3167 bumped the Gutenberg requirement to 5.9, so we can now freely leverage `useSelect and `useDispatch`.

spacedmonkey:
List of remaining components 

- `InserterMenu`
- `EnhancedResizableBox`
- `IgnoreNestedEvents`
- `Draggable`
- `PostSelector`

swissspidy:
#3412 covered the stories editor, but as Jonny notes there are still some remaining class components. For various reasons:

Some components are copied from Gutenberg (e.g. `IgnoreNestedEvents`), and keeping them that way makes it easier to merge any upstream changes. Others are quite large and could benefit from their own PR (e.g. `Draggable`), others have some uncertainty around their future (e.g. `PostSelector`)

As for the non-stories part we have to wait with hook usage due to backward compatibility (see #3359).

westonruter:
@swissspidy Is this still relevant?

PR #3083, opened 2019-08-23
Keyboard Copy + Paste

### LINKED ISSUE #2996: Keyboard Copy + Paste
See the [discovery ticket](https://github.com/ampproject/amp-wp/issues/2786) for more details.

AC: User can Copy a block by using <kbd><kbd>Ctrl</kbd>+<kbd>C</kbd></kbd> (while a block is selected), select any Page and paste the selected block by using <kbd><kbd>Ctrl</kbd>+<kbd>V</kbd></kbd>.
AC: The block should be copied with all the relevant settings brought over as well, e.g. same positioning, etc.

We could achieve this for by adding `onPaste` handler to the selected Page block which would then process the `clipboardData` and create new blocks accordingly.

See how it's done in [`RichText`](https://github.com/WordPress/gutenberg/blob/0214d79f059f75f49e9d7014243b78eb839a3a25/packages/block-editor/src/components/rich-text/index.js#L104), see also the [RichText package](https://github.com/WordPress/gutenberg/blob/0214d79f059f75f49e9d7014243b78eb839a3a25/packages/rich-text/src/component/index.js#L196).

Note that this is very closely related to #2989 and should share the pasting logic.

miina:
Fixes #2996 
Potentially will include Right Click copy-paste issues as well (#2990 #2989).

Todo:

- [x] Handle onPaste in Page
- [ ] Ensure only allowed blocks are copied
- [ ]  Verify that all the blocks work
- [ ] E2E tests.

Potentially:
- [ ] Include Right-click menu copy-paste, too

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -86,7 +86,6 @@ class PageEdit extends Component {
 		this.videoPlayer = createRef();
 		this.onSelectMedia = this.onSelectMedia.bind( this );
 		this.onPaste = this.onPaste.bind( this );
-		this.handlePaste = this.handlePaste.bind( this );
 	}
 
 	/**
@@ -187,28 +186,15 @@ class PageEdit extends Component {
 		}
 	}
 
-	handlePaste( { html, plainText } ) {
-		debugger;
-		const { onReplace, tagName, canUserUseUnfilteredHTML } = this.props; // @todo check about the canUseruse ... prop
-
-		const mode = 'BLOCKS';
-
-		const content = pasteHandler( {
-			HTML: html,
-			plainText,
-			mode,
+	onPaste( event ) {
+		const {
+			clientId,
+			insertBlocks,
+			isSelected,
 			tagName,
 			canUserUseUnfilteredHTML,
-		} );
+		} = this.props;
 
-		if ( content.length > 0 ) {
-			// Insert inner blocks @todo
-			onReplace( content );
-		}
-	}
-
-	onPaste( event ) {
-		const { isSelected } = this.props;
 		// Ignore if the Page is not the selected page.
 		if ( ! isSelected ) {
 			return;
@@ -237,14 +223,19 @@ class PageEdit extends Component {
 
 		event.preventDefault();
 
-		// Allows us to ask for this information when we get a report.
-		window.console.log( 'Received HTML:\n\n', html );
-		window.console.log( 'Received plain text:\n\n', plainText );
+		const mode = 'BLOCKS';
 
-		this.handlePaste( {
-			html,
+		const content = pasteHandler( {
+			HTML: html,
 			plainText,
+			mode,
+			tagName,
+			canUserUseUnfilteredHTML,
 		} );
+
+		if ( content.length > 0 ) {
+			insertBlocks( content, null, clientId );
+		}
 	}
 
 	removeBackgroundColor( index ) {
@@ -583,12 +574,16 @@ PageEdit.propTypes = {
 		autoAdvanceAfter: PropTypes.string,
 		autoAdvanceAfterDuration: PropTypes.number,
 	} ).isRequired,
+	canUserUseUnfilteredHTML: PropTypes.bool,
+	insertBlocks: PropTypes.func.isRequired,
+	isSelected: PropTypes.bool.isRequired,
 	setAttributes: PropTypes.func.isRequired,
 	media: PropTypes.object,
 	allowedBlocks: PropTypes.arrayOf( PropTypes.string ).isRequired,
 	totalAnimationDuration: PropTypes.number.isRequired,
 	getBlockOrder: PropTypes.func.isRequired,
 	moveBlockToPosition: PropTypes.func.isRequired,
+	tagName: PropTypes.string.func,
 	videoFeaturedImage: PropTypes.shape( {
 		source_url: PropTypes.string,
 	} ),
@@ -596,18 +591,21 @@ PageEdit.propTypes = {
 
 export default compose(
 	withDispatch( () => {
-		const { moveBlockToPosition } = dispatch( 'core/block-editor' );
+		const { insertBlocks, moveBlockToPosition } = dispatch( 'core/block-editor' );
 		return {
+			insertBlocks,
 			moveBlockToPosition,
 		};
 	} ),
 	withSelect( ( select, { clientId, attributes } ) => {
 		const { getMedia } = select( 'core' );
-		const { getBlockOrder, getBlockRootClientId } = select( 'core/block-editor' );
+		const { getBlockOrder, getBlockRootClientId, getSettings } = select( 'core/block-editor' );
 		const { getAnimatedBlocks } = select( 'amp/story' );
 
 		const isFirstPage = getBlockOrder().indexOf( clientId ) === 0;
 		const isCallToActionAllowed = ! isFirstPage && ! getCallToActionBlock( clientId );
+
+		const { __experimentalCanUserUseUnfilteredHTML } = getSettings();
 
 		const { mediaType, mediaId, poster } = attributes;
 
@@ -630,6 +628,7 @@ export default compose(
 			allowedBlocks: isCallToActionAllowed ? ALLOWED_CHILD_BLOCKS : ALLOWED_MOVABLE_BLOCKS,
 			totalAnimationDuration: totalAnimationDurationInSeconds,
 			getBlockOrder,
+			canUserUseUnfilteredHTM: __experimentalCanUserUseUnfilteredHTML,
 		};
 	} ),
 )( PageEdit );
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -33,6 +33,7 @@ import {
 	dispatch,
 } from '@wordpress/data';
 import { compose } from '@wordpress/compose';
+import { pasteHandler } from '@wordpress/blocks';
 
 /**
  * Internal dependencies
@@ -84,6 +85,8 @@ class PageEdit extends Component {
 
 		this.videoPlayer = createRef();
 		this.onSelectMedia = this.onSelectMedia.bind( this );
+		this.onPaste = this.onPaste.bind( this );
+		this.handlePaste = this.handlePaste.bind( this );
 	}
 
 	/**
@@ -184,6 +187,66 @@ class PageEdit extends Component {
 		}
 	}
 
+	handlePaste( { html, plainText } ) {
+		debugger;
+		const { onReplace, tagName, canUserUseUnfilteredHTML } = this.props; // @todo check about the canUseruse ... prop
+
+		const mode = 'BLOCKS';
+
+		const content = pasteHandler( {
+			HTML: html,
+			plainText,
+			mode,
+			tagName,
+			canUserUseUnfilteredHTML,
+		} );
+
+		if ( content.length > 0 ) {
+			// Insert inner blocks @todo
+			onReplace( content );
+		}
+	}
+
+	onPaste( event ) {
+		const { isSelected } = this.props;
+		// Ignore if the Page is not the selected page.
+		if ( ! isSelected ) {
+			return;
+		}
+		const clipboardData = event.clipboardData;
+
+		let plainText = '';
+		let html = '';
+
+		// IE11 only supports `Text` as an argument for `getData` and will
+		// otherwise throw an invalid argument error, so we try the standard
+		// arguments first, then fallback to `Text` if they fail.
+		try {
+			plainText = clipboardData.getData( 'text/plain' );
+			html = clipboardData.getData( 'text/html' );
+		} catch ( error1 ) {
+			try {
+				html = clipboardData.getData( 'Text' );
+			} catch ( error2 ) {
+				// Some browsers like UC Browser paste plain text by default and
+				// don't support clipboardData at all, so allow default
+				// behaviour.
+				return;
+			}
+		}
+
+		event.preventDefault();
+
+		// Allows us to ask for this information when we get a report.
+		window.console.log( 'Received HTML:\n\n', html );
+		window.console.log( 'Received plain text:\n\n', plainText );
+
+		this.handlePaste( {
+			html,
+			plainText,
+		} );
+	}
+
 	removeBackgroundColor( index ) {
 		const { attributes, setAttributes } = this.props;
 		const backgroundColors = JSON.parse( attributes.backgroundColors );
@@ -253,7 +316,13 @@ class PageEdit extends Component {
 	}
 
 	render() { // eslint-disable-line complexity
-		const { attributes, media, setAttributes, totalAnimationDuration, allowedBlocks } = this.props;
+		const {
+			attributes,
+			media,
+			setAttributes,
+			totalAnimationDuration,
+			allowedBlocks,
+		} = this.props;
 
 		const {
 			mediaId,
@@ -475,7 +544,10 @@ class PageEdit extends Component {
 						) }
 					</PanelBody>
 				</InspectorControls>
-				<div style={ style }>
+				<div
+					style={ style }
+					onPaste={ this.onPaste }
+				>
 					{ /* todo: show poster image as background-image instead */ }
 					{ VIDEO_BACKGROUND_TYPE === mediaType && media && (
 						<div className="editor-amp-story-page-video-wrap">


swissspidy:
Nice, can‘t wait to test this!

miina:
There is an issue with keyboard copying, namely, the `CopyHandler` doesn't always trigger for many blocks, looks like it triggers only if the block is selected in some specific way (so far first dragging the block and then copying seems to work the best). This seems to be an issue with regular post and the default block editor as well, e.g. adding an image block, clicking on it, and then pasting into empty paragraph sometimes copies the block into clipboard but not always.

The following blocks seem to work OK:
- text
- title
- author
- date
- Embed

The rest of the blocks are not reliably copied to clipboard. _If_ the block gets copied then everything works as expected.

Debugged the upstream `CopyHandler` to verify it's not reliably triggering. Also tried adding a custom copy handler by using `onCopyCapture` instead, also tried adding the ` handler directly wrapping each block, still not triggering for most of the blocks.

@swissspidy Any thoughts on this? 

Otherwise, looks like the Right-Click menu might be the way to reliably implement this feature.

swissspidy:
Haven't gotten around to test this myself yet, but what you describe sounds like an upstream bug (or perhaps even a browser issue) that we can't work around ourselves. Is there any related issue in the Gutenberg repository about the copy handler?

> Otherwise, looks like the Right-Click menu might be the way to reliably implement this feature.

👍

miina:
> Is there any related issue in the Gutenberg repository about the copy handler?

Couldn't find it as much as I checked at least, can look some more. I'll try testing it again later with clean WP Install (with AMP Plugin completely disabled) to see if that's still the case and then open an issue if yes.

assets/src/stories-editor/components/right-click-menu/index.js
```diff
@@ -0,0 +1,271 @@
...
+		clientY,
+		copyBlock,
+		cutBlock,
+		getBlock,
+		removeBlock,
+		duplicateBlock,
+		pasteBlock,
+	} = props;
+	const [ isOpen, setIsOpen ] = useState( true );
+	const [ copiedBlock, setCopiedBlock ] = useState( '' );
```
miina: As @westonruter brought up over the last grooming call then it's not possible to just read the clipboard data without and actual paste event.

WDYT of this workaround? Overview:
- When a block is either copied or cut, the block's markup is saved to the state as well.
- When something is saved to the state, and the right-click menu is opened, the Paste button will show up as well.
- When state is empty, the button is not displayed.

Shortcomings:
- When something else (e.g. a text or something from a different browser tab etc.) is copied meanwhile with the default browser (or keyboard) Copy then the state will still display the previously copied block. This could have a workaround of saving the copied block to a store instead of state, and we catch all the Copy events and empty the store where relevant.
- After refresh, the state is empty -- this means that pasting is not possible anymore with the button.
- Pasting to other Pages is not possible.

Note: Using <kbd><kbd>Ctrl</kbd>+<kbd>V</kbd></kbd> would still work as normally.

RightClick menu for a Block with sth saved to state:
Screenshot 2019-08-29 at 14 05 31


RightClick menu for Page:
Screenshot 2019-08-29 at 14 05 23

Thoughts?

cc @swissspidy

assets/src/stories-editor/components/right-click-menu/index.js
```diff
@@ -0,0 +1,271 @@
...
+		clientX,
+		clientY,
+		copyBlock,
+		cutBlock,
+		getBlock,
+		removeBlock,
+		duplicateBlock,
+		pasteBlock,
+	} = props;
+	const [ isOpen, setIsOpen ] = useState( true );
```
swissspidy: Yay for using hooks! ❤️  We should do that [more](https://github.com/ampproject/amp-wp/issues/2538).

assets/src/stories-editor/components/right-click-menu/index.js
swissspidy: Nice idea! As you say, there are some shortcomings, but I think it's a good compromise and better than not having that option at all.

I'd probably use a store rather than state though for this. Otherwise 👍

swissspidy: > it's not possible to just read the clipboard data without an actual paste event.

Looks like one just needs to use the Permissions API with the `clipboard-read` permission (Chrome only unfortunately)

https://developers.google.com/web/updates/2018/03/clipboardapi

The first time the user clicks on the paste button, we could ask for permission to directly use the clipboard:

https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read#Example

If permission is denied, we fall back to the data in the store/state.

miina: Yep, saw that it's possible in Chrome only.

Thoughts that this would cause some inconsistency in behavior though. E.g. in Chrome being able to copy from one tab and paste to the other and in other browsers not.

If that's acceptable then let's go for it. Thoughts?

swissspidy: To me that sounds like progressive enhancement, so we should do it :-)

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
```diff
@@ -36,17 +36,27 @@ const applyWithDispatch = withDispatch( ( dispatch, ownProps, { select } ) => {
...
 		if ( ! document.getElementById( 'amp-story-right-click-menu' ) ) {
-			const editLayout = document.querySelector( '.edit-post-layout' );
-
 			const menuWrapper = document.createElement( 'div' );
 			menuWrapper.id = 'amp-story-right-click-menu';
 
 			editLayout.appendChild( menuWrapper );
 		}
 
+		// Calculate the position to display the right click menu.
```
swissspidy: If it helps, the `Popover` component has utils to do similar position calculation:

https://github.com/WordPress/gutenberg/blob/02c5f3950dda05ffbba532d285e58be9232c0acf/packages/components/src/popover/utils.js

miina: Thanks, will check again tomorrow.

miina:
@swissspidy When you have time would you mind testing/reviewing to see if it seems to work as expected?

assets/src/stories-editor/components/right-click-menu/index.js
```diff
@@ -0,0 +1,322 @@
...
+		getCopiedMarkup,
+		removeBlock,
+		duplicateBlock,
+		pasteBlock,
+	} = props;
+	const [ isOpen, setIsOpen ] = useState( true );
+
+	useEffect( () => {
+		setIsOpen( true );
+	}, [ clientIds, clientX, clientY ] );
```
miina: I wonder if the full `event` object should be passed here instead of these params, theoretically it's possibly that the user tries to open the menu exactly in the same point and it wouldn't open then.

assets/src/stories-editor/components/right-click-menu/index.js
```diff
@@ -0,0 +1,322 @@
...
+		removeBlock( firstBlockClientId );
+	};
+
+	const onDuplicate = () => {
+		onClose();
+		duplicateBlock( firstBlockClientId );
+	};
+
+	const onPaste = () => {
+		pasteBlock( firstBlockClientId );
```
barklund: I assume there's an excellent reason for the changed ordering here. Could you add a comment explaining it, so someone doesn't come along and changes this in some attempt of generalising these callbacks?

assets/src/stories-editor/components/right-click-menu/index.js
barklund: The event object will be a `SyntheticEvent`, and as those are pooled, they might match by reference, but not by contents - so using them directly as guards for an effect would probably give weird results (but most times "work").

Events have a `timeStamp` (in milliseconds), that you can use to distinguish them – `SyntheticEvent` correctly copies this attribute as well.

I am however not sure it is needed here.

assets/src/stories-editor/store/reducer.js
```diff
@@ -159,4 +159,22 @@ export function blocks( state = {}, action ) {
...
 
-export default combineReducers( { animations, currentPage, blocks } );
+export function copiedMarkup( state = {}, action ) {
+	switch ( action.type ) {
+		case 'SET_COPIED_MARKUP':
+			const { markup } = action;
+			return markup;
+
+		case 'CLEAR_COPIED_MARKUP':
+			if ( ! state.length ) {
```
barklund: What is this guarding against? If someone sets `true` or `6` as the copied markup, then this means it can't be cleared, or?

swissspidy:
So far it seems to work really well!

Two things I noticed in the UI:

* The very first time the context menu is opened, the popover doesn't have its correct position yet, so it jumps quite a bit. Could this be avoided somehow, e.g. by visually hiding it for a split second?

* I can continue to paste a block, even if my computer's actual clipboard is empty. Is there a way for me to trigger `clearCopiedMarkup` as well?

miina:
> I can continue to paste a block, even if my computer's actual clipboard is empty. Is there a way for me to trigger clearCopiedMarkup as well?

Hmm, perhaps we should empty the store "clipboard" every time a block is Pasted?
There is an `onCopy` handler for clearing it up if something else is copied, however, this does not catch the copy event most of the times (same issue as here: https://github.com/ampproject/amp-wp/pull/3083#issuecomment-525690762). Thoughts?

miina:
> The very first time the context menu is opened, the popover doesn't have its correct position yet, so it jumps quite a bit. Could this be avoided somehow, e.g. by visually hiding it for a split second?

Thanks for catching that, will check.

assets/src/stories-editor/store/reducer.js
miina: It's only supposed to save strings as markup, changed the logic a little bit.

assets/src/stories-editor/components/right-click-menu/index.js
miina: Ha! Re-organized this, there was no reason for changing the order here, except for "not seeing the forest through the trees".

miina:
> I can continue to paste a block, even if my computer's actual clipboard is empty. Is there a way for me to trigger clearCopiedMarkup as well?

Adding clearing the stored text to paste afte it has been pasted once already.

That's one of the issues that come with allowing Paste use the actual clipboard as well, otherwise, we could rename the "Paste" to sth like "Paste Last Copied Block" and only use the saved markup. 
Well, the root cause is actually the Copy event not being caught reliably.

Let me know if allowing Pasting just once with the right-click menu makes sense to you

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -479,15 +494,18 @@ PageEdit.propTypes = {
 };
 
 export default compose(
-	withDispatch( () => {
+	withDispatch( ( dispatch ) => {
```
swissspidy: Good catch!

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
```diff
@@ -0,0 +1,105 @@
...
+	const onContextMenu = ( event ) => {
+		const selectedBlockClientIds = getSelectedBlockClientIds();
+
+		if ( selectedBlockClientIds.length === 0 ) {
+			return;
+		}
+
+		// Let's ignore multi-selection for now.
+		if ( hasMultiSelection() ) {
+			return;
```
swissspidy: Can we also return early when text within a block has selected and not just that block as a whole?

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
```diff
@@ -0,0 +1,105 @@
...
+				return <BlockEdit { ...props } />;
+			}
+
+			// Not relevant for reordering.
+			if ( isReordering ) {
+				return <BlockEdit { ...props } />;
+			}
+
+			return (
+				<div onContextMenu={ onContextMenu }>
```
swissspidy: What bugs me a bit is that, given an empty clipboard, nothing happens when right-clicking on a page. Can we display the browser's context menu in that case?

assets/src/stories-editor/components/right-click-menu/index.js
```diff
@@ -0,0 +1,284 @@
+/* global navigator */
```
swissspidy: Should not be needed anymore after #3150 I believe. Otherwise it sounds like a bug, and it should be added manually added to `.eslintrc`.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
miina: Added within c247c19.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
miina: Added within c247c19

Note that this depends on the markup saved to the store.

miina:
Looks like the Media Inserter test is not reliable now 😞

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -79,12 +80,13 @@ class PageEdit extends Component {
 			this.props.setAttributes( { anchor: getUniqueId() } );
 		}
 
+		this.onSelectMedia = this.onSelectMedia.bind( this );
```
spacedmonkey: Why was this moved?

miina: I was originally trying to get rid of the binding and use class properties instead (see #3152), and then it came out that class properties are not configured at this moment, so added it back to a different place, no difference otherwise.

spacedmonkey:
Screenshot from 2019-09-03 17-22-03
I see this message on the first time I paste. Not sure what we can do about this.

miina:
> I see this message on the first time I paste. Not sure what we can do about this.

@spacedmonkey See the conversation here for an explanation for this: https://github.com/ampproject/amp-wp/pull/3083#discussion_r319032200

spacedmonkey:
In firefox paste doesn't work and get the following error. 
```
navigator.clipboard.readText is not a function
```

miina:
@spacedmonkey Thanks for cathing that!

> `navigator.clipboard.readText is not a function`

Should be fixed now.

> `Warning: validateDOMNesting(...): <button> cannot appear as a descendant of <button>.
    in button (created by ForwardRef)`

This happens on `develop` as well as far as I'm testing, not sure why it's not happening on Chrome though.

> `MouseEvent.mozPressure is deprecated. Use PointerEvent.pressure instead.`

This seems to be something related to Firefox itself, for example, it happens when copying and pasting any random text as well (also on `develop`), also happens on other sites, e.g. GitHub.

assets/src/stories-editor/helpers/index.js
```diff
@@ -1718,3 +1718,58 @@ export const processMedia = ( media ) => {
...
+ * Ensure that only allowed blocks are pasted.
+ *
+ * @param {[]}      blocks Array of blocks.
+ * @param {string}  clientId Page ID.
+ * @param {boolean} isFirstPage If is first page.
+ * @return {[]} Filtered blocks.
+ */
+export const ensureAllowedBlocksOnPaste = ( blocks, clientId, isFirstPage ) => {
+	const allowedBlocks = [];
+	// @todo This will need handling for Page Attachment once it's available.
```
miina: ^ @swissspidy Bringing your attention here just in case -- if/when #3035 gets merged, this would need an update as well to ensure that Page Attachment wouldn't get copied if a CTA block or Attachment already exists on the same Page (no restriction for the first Page).

tests/e2e/specs/stories-editor/right-click-menu.js
```diff
@@ -0,0 +1,146 @@
...
+		const callToActionSelector = '.wp-block-amp-amp-story-cta';
+		const ctaBlock = await page.waitForSelector( callToActionSelector );
+		await openRightClickMenu( ctaBlock );
+
+		await clickButton( 'Copy Block' );
+
+		await goToPreviousPage();
+		await selectBlockByClientId( firstPageClientId );
+		const pageBlock = await page.$( `#block-${ firstPageClientId }` );
+		// Wait for transition time 300ms.
```
spacedmonkey: Why this a better way of doing this? Waiting around for 300ms feels a little weird.

spacedmonkey:
After doing some testing, I noticed that a copied block can not be pasted multiple times. This is different from the normal context menu that comes with the OS. If I have something in the clipboard, I expect to be able to paste it many times. It is a confusing behaviour that doesn't work here. If it is not possible paste multiple times, then this behaviour should be explained to users or the behavior should be changed.

spacedmonkey:
When pasting a text block, it pastes on top of the existing block. This makes it extremely unclear if the text block as pasted or not. I had to check the sidebar to make sure they was a new block added. Would it be better to paste the object below the existing block so that it is clear that paste has worked.

spacedmonkey:
Pasting a block, should not pasting at the point of the original position, it should paste at the point the cursor position.

assets/src/stories-editor/blocks/amp-story-page/copy-paste-handler.js
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Internal dependencies
+ */
+import { ensureAllowedBlocksOnPaste } from '../../helpers';
+
+function CopyPasteHandler( { children, onCopy, clientId, isSelected } ) {
+	const {
+		isFirstPage,
+		canUserUseUnfilteredHTML,
+	} = useSelect(
```
spacedmonkey: Should `useSelect` be used? We cannot reliably use it just yet. See #2538 (comment)

assets/src/stories-editor/blocks/amp-story-page/copy-paste-handler.js
```diff
@@ -0,0 +1,128 @@
...
+		// otherwise throw an invalid argument error, so we try the standard
+		// arguments first, then fallback to `Text` if they fail.
+		try {
+			plainText = clipboardData.getData( 'text/plain' );
+			html = clipboardData.getData( 'text/html' );
+		} catch ( error1 ) {
+			try {
+				html = clipboardData.getData( 'Text' );
+			} catch ( error2 ) {
+				// Some browsers like UC Browser paste plain text by default and
```
spacedmonkey: Should this error state, return a message to the user to explain that something went wrong?

spacedmonkey:
Control + C and Control + V doesn't work.

swissspidy:
> Control + C and Control + V doesn't work.

See relevant discussion at https://github.com/ampproject/amp-wp/pull/3083#issuecomment-525690762

assets/src/stories-editor/blocks/amp-story-page/copy-paste-handler.js
swissspidy: We can now use `useSelect` since #3167, see https://github.com/ampproject/amp-wp/issues/2538#issuecomment-527228741. Or do you mean because of the issue with scalar values?

swissspidy:
> When pasting a text block, it pastes on top of the existing block. This makes it extremely unclear if the text block as pasted or not. I had to check the sidebar to make sure they was a new block added. Would it be better to paste the object below the existing block so that it is clear that paste has worked.

I noticed that too. We already do that for newly inserted blocks, see [`maybeSetInitialPositioning()`](https://github.com/ampproject/amp-wp/blob/8bc8077300f225d3192a76b7b4db1bd477707657/assets/src/stories-editor/helpers/index.js#L1081-L1118). We should try to adapt this logic for the pasting as well.

tests/e2e/specs/stories-editor/right-click-menu.js
swissspidy: Indeed. Let's find a more suitable indicator, perhaps we can just wait for the `transitionend` event to fire.

spacedmonkey:
Should I start working some of the bugs that I have noticed here? Are they valid?

swissspidy:
@spacedmonkey Optimizing pasting to use the cursor position would be nice!

> I noticed that a copied block can not be pasted multiple times

IIRC it was changed to work like this on purpose, but we can try to change it back 👍

assets/src/stories-editor/blocks/amp-story-page/copy-paste-handler.js
spacedmonkey: Yeah, we can use `useSelect`

spacedmonkey:
> **Edit:** just saw your suggestion about the cursor position

I am going to take a look into this now.

spacedmonkey:
Fixed issue that wouldn't let you paste multiple times. https://github.com/ampproject/amp-wp/pull/3083/commits/d5f4250d5dabf8f9a337c1858697dab417bc7612

spacedmonkey:
As it is a big change, I broken off my code fix for cursor position to another PR. 
https://github.com/ampproject/amp-wp/pull/3220

spacedmonkey:
In my testing today, the keyboard copy and paste was working. I will text on different browsers.

spacedmonkey:
@googlebot I consent.

spacedmonkey:
Screenshot from 2019-09-10 15-10-07

The right click menu doesn't show on screens less than 781px wide :(

PR #2925, opened 2019-07-31
Custom attributes support on analytics component

ethanclevenger91:
For things like `data-includes="credentials"` or remote config.

PR #3095, opened 2019-08-26
Allow custom attributes on amp-analytics component

ethanclevenger91:
Continuation of #2925 , sorry for hiccup w/ git.

Solves things like data-includes="credentials" or remote config.


tests/php/test-amp-analytics-options.php
```diff
@@ -240,12 +240,16 @@ public function test_amp_get_analytics() {
...
-				$analytics[ $key ]['type'] = 'test';
+				$analytics[ $key ]['type']                                = 'test';
+				$analytics[ $key ]['attributes']['data-include']          = '_till_responded';
+				$analytics[ $key ]['attributes']['data-block-on-consent'] = 'credentials';
 				return $analytics;
 			}
 		);
 		$analytics = amp_get_analytics();
 		$this->assertEquals( 'test', $analytics[ $key ]['type'] );
+		$this->assertEquals( '_till_responded', $analytics[ $key ]['data-include'] );
```
swissspidy: This should be `$analytics[ $key ]['attributes']['data-include']` it seems?

tests/php/test-amp-analytics-options.php
```diff
@@ -265,6 +269,20 @@ public function test_amp_print_analytics() {
 
 		$this->assertStringStartsWith( '<amp-analytics', $output );
 		$this->assertContains( 'type="googleanalytics"><script type="application/json">{"requests":{"event":', $output );
+
+		add_filter(
+			'amp_analytics_entries',
+			static function( $analytics ) use ( $key ) {
```
swissspidy: `$key` is undefined here.

tests/php/test-amp-analytics-options.php
```diff
@@ -240,12 +240,16 @@ public function test_amp_get_analytics() {
...
+				$analytics[ $key ]['type']                                = 'test';
+				$analytics[ $key ]['attributes']['data-include']          = '_till_responded';
+				$analytics[ $key ]['attributes']['data-block-on-consent'] = 'credentials';
 				return $analytics;
 			}
 		);
 		$analytics = amp_get_analytics();
 		$this->assertEquals( 'test', $analytics[ $key ]['type'] );
+		$this->assertEquals( '_till_responded', $analytics[ $key ]['data-include'] );
+		$this->assertEquals( 'credentials', $analytics[ $key ]['data-block-on-consent'] );
```
swissspidy: This should be `$analytics[ $key ]['attributes']['data-block-on-consent']` it seems?

felixarntz:
> It should then be possible to replace the hard-coded markup with a `amp_analytics_entries` filter.

As discussed in person earlier, I'd prefer to have Site Kit continue to generate the markup as it's more decoupled from specifics of how the AMP plugin works. There _is_ a dependency for sure, but this way it's less coupled.

Nonetheless, the suggested change makes sense from my POV to provide the flexibility to those using the AMP plugin's filters.

tests/php/test-amp-analytics-options.php
ethanclevenger91: @swissspidy Thanks for these. I believe I've resolved the issues. Unable to get the phpunit tests to run on my machine. Getting:

```
PHP Fatal error:  Unable to locate wordpress-tests-lib in C:\Users\ethan\dev-archive\amp-wp\vendor\xwp\wp-dev-lib\sample-config\phpunit-plugin-bootstrap.php on line 40
```

And that's running `phpunit` directly rather than `npm run test:php` which can't find `phpunit`, even though it's executable on my path.

westonruter:
Alas, a merge conflict! Due to #3106.

ethanclevenger91:
@westonruter this should be resolved

tests/php/test-amp-analytics-options.php
```diff
@@ -269,13 +275,27 @@ function ( $entries ) use ( $analytics, &$trigger_count ) {
 				$trigger_count++;
 			}
 		);
-
+
```
westonruter: This whitespace is causing a linting error: https://travis-ci.org/ampproject/amp-wp/jobs/593274574#L777-L787

tests/php/test-amp-analytics-options.php
```diff
@@ -261,6 +265,8 @@ public function test_amp_print_analytics() {
 
 		$analytics = amp_get_analytics();
 
+		$key = key( $analytics );
+
```
westonruter: Whitespace here is causing linting error: https://travis-ci.org/ampproject/amp-wp/jobs/593274574#L777-L787

westonruter:
😕 Humm. Tests are still failing: https://travis-ci.org/ampproject/amp-wp/jobs/593324798#L1037-L1049

swissspidy:
@westonruter I think this comes [from the hook added to `amp_print_analytics` above](https://github.com/ampproject/amp-wp/pull/3095/files#diff-4c692e4a86d5fee52c878cd27a3a4129R271-R277):

```php
add_action(
	'amp_print_analytics',
	function ( $entries ) use ( $analytics, &$trigger_count ) {
		$this->assertEquals( $analytics, $entries );
		$trigger_count++;
	}
);
```

In the test `amp_analytics_entries` is filtered, but the hook still compares the output with the old `$analytics` value. IMO that `$this->assertEquals( $analytics, $entries );` call should be moved outside the hook callback. Or just removed entirely, as its `$output` is already tested below.

ethanclevenger91:
@swissspidy Outside the hook callback we don't have access to the raw array, which is what this is testing.

We can remove the action once it's been run and tested. To be thorough, you might re-add it using the fresh result of `amp_get_analytics()` while testing the `amp_analytics_entries` filter. Or, like you said, you could drop the test altogether and so long as the output is the same at the end, assume the rest was as expected.

Side-note - I'd love to run these tests locally before pushing but `phpunit` currently gives me ` Unable to locate wordpress-tests-lib`. I assume I'm missing a step?

westonruter:
> Side-note - I'd love to run these tests locally before pushing but `phpunit` currently gives me ` Unable to locate wordpress-tests-lib`. I assume I'm missing a step?

It assumes that you're running the site in a checkout of the [wordpress-develop](https://github.com/wordpress/wordpress-develop) codebase, which includes all the core tests. 

If you have WordPress's core tests set up elsewhere on your system, then you can define the path to the [`tests/phpunit`](https://github.com/WordPress/wordpress-develop/tree/master/tests/phpunit) directory via the `WP_TESTS_DIR` environment variable.

Here's where that code is coming from:

https://github.com/xwp/wp-dev-lib/blob/e0205d733c2cabccfd05ec281ae43872fc42c221/sample-config/phpunit-plugin-bootstrap.php#L28-L42

PR #3097, opened 2019-08-27
Fix lightbox on file-linked galleries

### LINKED ISSUE #2850: Add lightbox effect doesn't work with gallery shortcode
I have a gallery shortcode inserted with the shortcode block like [gallery ..... link="file"] so each image should be linked to its source image. And without any amp settings selected it works correctly. However, if in the AMP settings in the shortcode block editor sidebar I have selected to "Add lightbox effect", then the the links to the source images disappear and the lightbox doesn’t work. It’s is the same as if link=”file” is not in the shortcode. 

Originally https://wordpress.org/support/topic/amp-carousel-doesnt-have-alt-tags/

schlessera:
Add the `lightbox` attribute to gallery shortcode images when they are being linked to by file and the AMP lightobx setting is enabled.

There was no test suite yet for the gallery embed handler, so I added a test suite to make sure this works as expected.

Fixes #2850

### DIFF
--- a/includes/embeds/class-amp-gallery-embed.php
+++ b/includes/embeds/class-amp-gallery-embed.php
@@ -166,11 +166,18 @@ class AMP_Gallery_Embed_Handler extends AMP_Base_Embed_Handler {
 		$is_lightbox = isset( $attributes['amp-lightbox'] ) && true === filter_var( $attributes['amp-lightbox'], FILTER_VALIDATE_BOOLEAN );
 		if ( isset( $attributes['amp-carousel'] ) && false === filter_var( $attributes['amp-carousel'], FILTER_VALIDATE_BOOLEAN ) ) {
 			if ( true === $is_lightbox ) {
+				$add_lightbox_attribute = static function ( $attr ) {
+					$attr['lightbox'] = '';
+					return $attr;
+				};
+
 				remove_filter( 'post_gallery', [ $this, 'maybe_override_gallery' ], 10 );
-				add_filter( 'wp_get_attachment_image_attributes', [ $this, 'add_lightbox_attribute' ], 10, 3 );
+				add_filter( 'wp_get_attachment_image_attributes', $add_lightbox_attribute );
+
 				$attributes['link'] = 'none';
 				$html               = gallery_shortcode( $attributes );
-				remove_filter( 'wp_get_attachment_image_attributes', [ $this, 'add_lightbox_attribute' ], 10 );
+
+				remove_filter( 'wp_get_attachment_image_attributes', $add_lightbox_attribute );
 				add_filter( 'post_gallery', [ $this, 'maybe_override_gallery' ], 10, 2 );
 			}
 
@@ -290,20 +297,5 @@ class AMP_Gallery_Embed_Handler extends AMP_Base_Embed_Handler {
 			}
 		</style>
 		<?php
-	}
-
-	/**
-	 * Add lightbox attribute to a provided attachment.
-	 *
-	 * @param array        $attr       Attributes for the image markup.
-	 * @param WP_Post      $attachment Image attachment post.
-	 * @param string|array $size       Requested size. Image size or array of
-	 *                                 width and height values (in that order).
-	 *                                 Default 'thumbnail'.
-	 * @return array Modified attributes for the image markup.
-	 */
-	public function add_lightbox_attribute( $attr, $attachment, $size ) {
-		$attr['lightbox'] = '';
-		return $attr;
 	}
 }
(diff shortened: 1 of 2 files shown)


schlessera:
Right now, the test suite has 2 failures: `shortcode_with_lightbox` and `shortcode_with_lightbox_linking_to_file`, so both of the cases where the lightbox switch is on but the carousel switch is off.

It looks like these cases are not yet handled by the gallery embed handler, as it spits out non-AMP HTML (probably the unmodified original shortcode handler output).

Also, if you don't specifiy the AMP carousel switch value at all, it seems to default to using a carousel nevertheless. Is that the intended behavior?

schlessera:
I haven't tried this on the frontend, this is purely from running the sanitizer within tests and setting its flags.

The gallery embed code does not contain any reference of the mode, so I doubt that is the case here, unless one of the two scenarios (reader vs standard/transitional) hasn't been implemented yet.

For the non-AMP output, you can see the result here: https://travis-ci.org/ampproject/amp-wp/jobs/577377621#L1208-L1224

So, here's what we need to decide upon:
- Should a gallery use `<amp-carousel>` if the carousel switch is off? If this is conditional, what are those conditions?
- Is it a bug that the new tests have uncovered that the gallery spits out non-AMP images when the carousel is off and lightbox is on?

westonruter:
> * Should a gallery use `<amp-carousel>` if the carousel switch is off? If this is conditional, what are those conditions?

In Reader mode, a gallery uses an `amp-carousel` by default. In Standard/Transitional modes, a gallery should only use `amp-carousel` if it is toggled on. 

I just tried adding a Gallery block and I'm not seeing the carousel toggle:

> image

But in the Reader-mode template, it is shown as a carousel:

> image

If I switch to Transitional mode, then I _do_  see a toggle but it is disabled by default:

> image

And on the frontend:

> image

And when enabling the toggle, then I see a carousel on the frontend:

> image

Perhaps the confusion is that there is actually a gallery _sanitizer_ as well: [`AMP_Gallery_Block_Sanitizer`](https://github.com/ampproject/amp-wp/blob/7515ad11d4006f7716d916b5f44ddacd4fd748d5/includes/sanitizers/class-amp-gallery-block-sanitizer.php). It has a `carousel_required` argument which is set to `true` if in Reader mode:

https://github.com/ampproject/amp-wp/blob/93370466991242cb6105a3bfde8c1c63bfe522ee/includes/amp-helper-functions.php#L842-L844

Yes, the gallery handling is somewhat convoluted!

westonruter:
> * Is it a bug that the new tests have uncovered that the gallery spits out non-AMP images when the carousel is off and lightbox is on?

No, it's not a bug because the `AMP_Gallery_Embed` runs during template generation. During post-processing, the `AMP_Img_Sanitizer` then ensures any remaining `img` elements get converted to `amp-img`. So during template generation (e.g. in embed handlers) it's fine to output `img` elements.

schlessera:
Thanks for all the context. Based on the above, I assume that the current tests should then compare to what should be expected. They can certainly be extended, but I'd suggest doing so in subsequent tickets to get this one moving.

includes/embeds/class-amp-gallery-embed.php
```diff
@@ -289,4 +291,19 @@ public function print_styles() {
...
+	 * Add lightbox attribute to a provided attachment.
+	 *
+	 * @param array        $attr       Attributes for the image markup.
+	 * @param WP_Post      $attachment Image attachment post.
+	 * @param string|array $size       Requested size. Image size or array of
+	 *                                 width and height values (in that order).
+	 *                                 Default 'thumbnail'.
+	 * @return array Modified attributes for the image markup.
+	 */
+	public function add_lightbox_attribute( $attr, $attachment, $size ) {
```
westonruter: Do you think this would be better as a closure? The above could then be:

```php
$add_lightbox_attribute = static function( $attr ) {
		$attr['lightbox'] = '';
		return $attr;
};
add_filter( 'wp_get_attachment_image_attributes', $add_lightbox_attribute );
$attributes['link'] = 'none';
$html               = gallery_shortcode( $attributes );
remove_filter( 'wp_get_attachment_image_attributes', $add_lightbox_attribute );
```

This would avoid adding to the public API space.

Not a big deal, however.

Also, the `$attachment` and `$size` attributes are not used so they could be left off.

schlessera: Solved in https://github.com/ampproject/amp-wp/pull/3097/commits/a9e18f9cd0f1a7a37977400f964933b3b7a6cfe0

PR #3101, opened 2019-08-27
Discard invalid hosts in URL attributes

### LINKED ISSUE #2671: Invalid URL like `foo@` passed through without sanitization causing AMP validation error

image image

image

----

schlessera:
The reason why the bogus value of `@foo` could even pass in the first place is that `wp_parse_url()` does not return a `$host` for that value, and this inadvertently bypasses the check for invalid characters in the host name.

I fixed it by falling back to `$path` when `$host` is not detected, and checking `$path` for invalid characters in that case.

I am not 100% sure though whether that might not discard another case where `$path` might contain valid characters that are turned into invalid characters when interpreted as a host name. However, as this is for the `value_url` type of attributes only, I couldn't immediately think of such a case.

Fixes #2671

### DIFF
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -1329,8 +1329,19 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 					$url = substr( $url, strlen( $protocol ) + 1 );
 				}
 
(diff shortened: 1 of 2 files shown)


schlessera:
Ok, running the entire test suite reveals that this breaks more than it solves, back to the drawing board...

schlessera:
Closing this PR, as I don't think the targeted bug can be solved at the attribute sanitizer level: https://github.com/ampproject/amp-wp/issues/2671#issuecomment-525340332

PR #3103, opened 2019-08-27
Add comment website URL sanitization

### LINKED ISSUE #2671: Invalid URL like `foo@` passed through without sanitization causing AMP validation error

schlessera:
This is the second attempt at fixing #2671, after I had to find out that it couldn't possibly be done at the attribute sanitization level.

This now introduces an additional step that iterates over comment entries and checks their (final) website URL targets.

If an invalid URL is detected, the `href` attribute of that link is removed. This should keep the visual output intact, while getting rid of the validation error.

Fixes #2671

### DIFF
--- a/includes/sanitizers/class-amp-comments-sanitizer.php
+++ b/includes/sanitizers/class-amp-comments-sanitizer.php
@@ -8,7 +8,7 @@
 /**
  * Class AMP_Comments_Sanitizer
  *
- * Strips and corrects attributes in forms.
+ * Strips and corrects attributes in comments.
  */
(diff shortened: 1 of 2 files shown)


swissspidy:
Curious: how does a removed href attribute affect keyboard navigation?

schlessera:
Working on a different PR to make this more generalized, so it also catches invalid hrefs for post content. See https://github.com/ampproject/amp-wp/issues/2671#issuecomment-525394936

westonruter:
Closing in favor of #3151.

PR #3104, opened 2019-08-27
Single Animation Previews

### LINKED ISSUE #3070: Enable user to preview only one animation within a page
Dependent on Issue #3069 

As a user I want to be able to preview only one animation within the page I am editing.

AC1 - Add a CTA in the editing experience that allows a user to view just on animation on the page.
AC2 - If there are multiple animations, only the chosen animation will preview.

swissspidy:
Fixes #3070.


miina:
This is awesome! It (almost 😄) works! 🎉 

 "Almost" works since rotation doesn't seem to work anymore, probably we'll need to add another wrapper for the animations so that it wouldn't collide with the rotation wrapper.

One additional comment from functional testing:
In some animation cases, for example in case of Fly In Bottom, the resizing handles are visible while the animation is happening, could we remove those for the animation time?

miina:
Rotation issue:
rotation

Resizing handles issue:
visible-handles

miina:
I'm actually seeing odd behavior for resizing as well, didn't happen on `develop`, could you confirm if you see the same? (Resizing from one edge seems to move the block from other edges, too:
resizing

assets/src/stories-editor/helpers/index.js
```diff
@@ -1715,3 +1734,121 @@ export const processMedia = ( media ) => {
...
+			break;
+		case 'pan-left':
+		case 'pan-right':
+			offsetX = STORY_PAGE_INNER_WIDTH - ( width * scalingFactor );
+			offsetY = ( STORY_PAGE_INNER_HEIGHT - ( height * scalingFactor ) ) / 2;
+			blockElement.style.setProperty( '--animation-scale-start', scalingFactor );
+			blockElement.style.setProperty( '--animation-scale-end', scalingFactor );
+			break;
+		case 'pan-down':
+		case 'pan-up':
```
swissspidy: Since images with a `pan-*` animation are always scaled on the frontend (even after animation has finished), I wonder whether we should do the same in the editor as well for a better WYSIWYG experience.

miina: Yes, would be great! Are there any downsides?

swissspidy:
Thanks for testing @miina!

> "Almost" works since rotation doesn't seem to work anymore, probably we'll need to add another wrapper for the animations so that it wouldn't collide with the rotation wrapper.
>
> I'm actually seeing odd behavior for resizing as well, didn't happen on develop, could you confirm if you see the same? (Resizing from one edge seems to move the block from other edges, too

The issue wasn't a conflict between the `transform` properties, but that the `:root` selector was wrongly removed by `postcss-preset-env`, which broke lots of things.

Resizing and rotation now works again as expected.

> In some animation cases, for example in case of Fly In Bottom, the resizing handles are visible while the animation is happening, could we remove those for the animation time?

As long as the animation is playing, resize handles should actually be hidden:

https://github.com/ampproject/amp-wp/blob/4353af3c55c81ceb5181de70b156d2a13071db71/assets/src/stories-editor/style.css#L246-L251

I'll try to improve that though 👍

miina:
Thanks for the fixes, will test again shortly.

> As long as the animation is playing, resize handles should actually be hidden:

Maybe it's the set transition time why it takes some time to hide?

swissspidy:
Uhm, I think I just found a bug related to animation 🚨

When a block is saved, the `animate-in-delay` and `animate-in-duration` attributes in the resulting attributes contain only numbers, instead of numbers with an `ms` suffix... That means AMP just ignores them and there's no delay or anything on the frontend

assets/src/stories-editor/helpers/index.js
swissspidy: It might be a bit unexpected for the user perhaps. I'll do a quick check to see how feasible it is.

swissspidy: We couldn't just use `scale()` in CSS, as that would scale _everything_, including things like resize handles. Also, the image must not overflow the page in that case.

We'd need to actually increase the width and height in the editor by the determined scaling factor, but only in the editor and when not playing the animation preview.

Since this is a bit out of scope for this issue/PR here, I'll open a new one to explore this.

swissspidy: Issue opened: #3118.

miina:
Discovered another issue, unfortunately:
rotated

Looks like the rotation has an effect on the animation direction as well.

assets/src/stories-editor/constants.js
```diff
@@ -116,17 +116,17 @@ export const MAX_IMAGE_SIZE_SLUG = 'amp_story_page';
...
+	'fly-in-top': 400,
+	pulse: 400,
+	'rotate-in-left': 600,
+	'rotate-in-right': 600,
 	'twirl-in': 1000,
-	'whoosh-in-left': 500,
-	'whoosh-in-right': 500,
+	'whoosh-in-left': 400,
+	'whoosh-in-right': 400,
 	'pan-left': 1000,
```
miina: ^ These were originally taken from the [AMP defaults](https://amp.dev/documentation/guides-and-tutorials/start/visual_story/animating_elements/), any reason to change that?

swissspidy: The defaults changed. See commit https://github.com/ampproject/amp-wp/pull/3104/commits/814f609c0b825f11d38e2bcce4891779480ae885 and upstream change https://github.com/ampproject/amphtml/pull/21446.

miina:
One more finding when adding a delay: looks like instead of waiting out of the Page's borders, the delay is happening in the original position:
late-animation

assets/src/stories-editor/constants.js
miina: Ah, good to know, thanks!

swissspidy:
Thanks for testing! Great findings.

> Looks like the rotation has an effect on the animation direction as well.

Sounds like the animation transform needs to be outside the rotation transform. Currently it's within. Not sure if there's an element there I could use though 🤔 

> One more finding when adding a delay: looks like instead of waiting out of the Page's borders, the delay is happening in the original position:

Good point. Basically need to apply the first keyframe already before animation starts. Should be doable.

miina:
Rotation works well now and so does the delay for most of the blocks 👍 

There is still something odd happening with the Pan animation and delaying:
pan

Also, it looks quite different on the FE, in the editor the animation doesn't cover the full Page, whereas in the FE it does:

Screenshot 2019-08-30 at 19 20 11

Thoughts?

swissspidy:
Which pan animation was that?

miina:
"Pan Left"

swissspidy:
@miina The pan animations should work much better now!

swissspidy:
Decided to implement the animation preview for all blocks on a given page in this PR as well, which led to some refactoring and IMO some general code improvements as well. But it also means it needs another round of review.

Besides this enhancement I also improved the animations themselves, like the pan animation mentioned above.

miina:
Alright. Will test again.

miina:
CTA block is misbehaving quite noticeably, had the suspicion since its dragging logic is a bit different, tested that first :D  

cta

swissspidy:
Welp, I should have tested with that block too I guess 😄

miina:
I'm still (or again actually) seeing the delay being visibly on the page instead of waiting outside (Single Animation Preview):

delay

Is it still working OK for you?

miina:
Same for Playing all:
delay

swissspidy:
Ah, I think this regressed with 53c917c5dc8ac78bb37de7db6f9b33b0e9cefe23 🤦‍♂

miina:
Otherwise seems to work well 👍 Will try to check the code tomorrow as well.

swissspidy:
Both issues should be fixed now.

assets/src/stories-editor/components/with-active-page-state.js
```diff
@@ -32,6 +35,7 @@ const withActivePageState = ( BlockListBlock ) => {
 				...props.className,
 				'amp-page-active': isTopLevelBlock && isActivePage,
 				'amp-page-inactive': isTopLevelBlock && ! isActivePage,
+				'amp-page-is-animating': isPlayingAnimation,
```
miina: If this is only for active page then what happens if the user moves to another page while the animation is playing?

miina: Looks like this happens:
test

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -114,10 +114,17 @@ class Draggable extends Component {
 	 */
 	onDragStart( event ) {
 		const { blockName, elementId, transferData, onDragStart = noop } = this.props;
-		const element = document.getElementById( elementId );
 		const isCTABlock = 'amp/amp-story-cta' === blockName;
+		const element = isCTABlock ? document.getElementById( elementId ) : document.getElementById( elementId ).parentNode;
```
miina: Would be good to leave a comment here on why the CTA block has an exception.

assets/src/stories-editor/components/with-active-page-state.js
swissspidy: That looks fun! Let's leave it like this 😄 

Seems like I need to move all of the animation playback stuff to the data store, so that we know exactly which animations are currently playing where.

Then, when doing an action like changing the current page we can easily cancel them via the data store.

spacedmonkey:
Feedback after playing with this PR 

- When play animation and play all animations button is pressed, the button icon should change to stop. Clicking again, pause should stop the animation. 

- Above the Play all animations, there should be list of all animations that appear on this page. It makes it much easier to know is going on if a user has lots of blocks with animations.

swissspidy:
Just pushed a ton of changes!

The actual animations in the browser are now mere side effects of the changes in the Redux store. That means animations can be handled via one central location.

In other words:

* The play button turns into a pause/stop button after clicking
* Animations are stopped immediately when switching pages

swissspidy:
> Above the Play all animations, there should be list of all animations that appear on this page. It makes it much easier to know is going on if a user has lots of blocks with animations.

Hmm I am not sure whether this is the right place for this. If there are lots of animated blocks, this list would be quite long, and make the block sidebar more busy than it already is.

My first thought was to have some kind of indicator in the block navigation for animated blocks.

spacedmonkey:
Icon should be a stop icon, not pause icon.

swissspidy:
I don't think there is a dashicon stop icon: https://developer.wordpress.org/resource/dashicons/#admin-media

spacedmonkey:
> I don't think there is a dashicon stop icon: https://developer.wordpress.org/resource/dashicons/#admin-media

That is a pain :( But a stop icon should be pretty easy to find. It is just a square. I do think the is an important difference here in pause vs stop. If the icon / wording says pause, that is untrue, as it is stopping the animations and restarting them.

swissspidy:
Sure, I can look into it. In the meantime, let's focus on reviewing the rest :-)

swissspidy:
Demo Video: https://cloudup.com/cgpRSkJoa84

Includes playback of a single animation and all animations on a page, as well as a  quick comparison with the animations on the frontend.

barklund:
Should the page animation inspector panel perhaps include a list of the animations on the page? Or at least a count of them?

Currently even if there are no animations, the button "Play all animations" is still there and active even though nothing happens on click. Should the button perhaps be disabled if there are no animations on the page?

Maybe the button text could simply be `play ${n} animation(s)` and then it would be obvious why the button is disabled if the count is zero?

swissspidy:
> Should the page animation inspector panel perhaps include a list of the animations on the page? Or at least a count of them?

This was suggested by Jonny as well above, but I had some reservations about this being the right place for such a list. See https://github.com/ampproject/amp-wp/pull/3104#issuecomment-528868119. I do like the suggestion of using a count in the button label though.

> Currently even if there are no animations, the button "Play all animations" is still there and active even though nothing happens on click. Should the button perhaps be disabled if there are no animations on the page?

Hmm I already implemented this in 41138b0caa417aaa5087762330b6bbf64ff91dac 🤔 There's even a test for it. But perhaps there is some edge case. Will look into it.

barklund:
> > Currently even if there are no animations, the button "Play all animations" is still there and active even though nothing happens on click. Should the button perhaps be disabled if there are no animations on the page?
> 
> Hmm I already implemented this in [41138b0](https://github.com/ampproject/amp-wp/commit/41138b0caa417aaa5087762330b6bbf64ff91dac) 🤔 There's even a test for it. But perhaps there is some edge case. Will look into it.

If I add an element with an animation, the "play all" button appears on the page. If I remove the animation from the element, the "play all" button is correctly removed. But if I delete the element outright without first setting the animation to "none", the "play all" button sticks around still thinking there's an animation on the page.

barklund:
If I create a cascading sequence of animations, so element 1 plays immediately, element 2 plays after element 1 and element 3 plays after element 2, the "play all animations" action gets stuck after the second animation - the third one never starts. This does work correctly in "preview", it's only in the editor with the "play all animations", that third level animations don't play.

I also managed to set up a longer chain, where one of the chained animations played immediately even though it was set to play after an element, that never actually played. I can't recreate it though.

barklund:
> If I create a cascading sequence of animations, so element 1 plays immediately, element 2 plays after element 1 and element 3 plays after element 2, the "play all animations" action gets stuck after the second animation - the third one never starts. This does work correctly in "preview", it's only in the editor with the "play all animations", that third level animations don't play.

While the page animation playback is in progress (or stuck as in this case and will never complete until I press the "stop" button), I can still select other blocks on the page, change them, move tem, delete them, etc.

It might be a good idea to abort the animation playback as soon as the page loses focus/selection.

assets/src/stories-editor/blocks/amp-story-page/animation-settings.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Displays the animation settings for the page block.
+ *
+ * It allows the user to preview all animations on the current page.
+ *
+ * @param {Object} props Component props.
+ * @param {string} props.clientId Client ID.
+ */
+const AnimationSettings = ( { clientId } ) => {
```
barklund: This component `AnimationSettings ` is almost identical to the `AnimationControls` button. Could we somehow reuse both the logic and the presentation and generalise the few differences?

This will also have the added benefit of rewriting `AnimationControls` to use hooks instead of HoC's.

assets/src/stories-editor/store/middlewares.js
```diff
@@ -0,0 +1,45 @@
...
+		);
+	};
+	let chain = [];
+
+	const middlewareAPI = {
+		getState: store.getState,
+		dispatch: ( ...args ) => enhancedDispatch( ...args ),
+	};
+	chain = middlewares.map( ( middleware ) => middleware( middlewareAPI ) );
+	enhancedDispatch = flowRight( ...chain )( store.dispatch );
```
barklund: Why is the same variable `enhancedDispatch` used twice here, as the two definitions of it are wildly different and serve very different purposes? This code is pretty complex already, reusing this variable doesn't make it easier.

```suggestion
	const enhancedDispatch = flowRight( ...chain )( store.dispatch );
```

assets/src/stories-editor/store/middlewares.js
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @return {Object} Update Store Object.
+ */
+function applyMiddlewares( store ) {
+	const middlewares = [
+		refx( effects ),
+		multi,
+	];
+
+	let enhancedDispatch = () => {
```
barklund:
```suggestion
	const disabledDispatch = () => {
```

assets/src/stories-editor/store/middlewares.js
```diff
@@ -0,0 +1,45 @@
...
+		throw new Error(
+			'Dispatching while constructing your middleware is not allowed. ' +
+			'Other middleware would not be applied to this dispatch.'
+		);
+	};
+	let chain = [];
+
+	const middlewareAPI = {
+		getState: store.getState,
+		dispatch: ( ...args ) => enhancedDispatch( ...args ),
```
barklund:
```suggestion
		dispatch: ( ...args ) => disabledDispatch( ...args ),
```

assets/src/stories-editor/store/middlewares.js
swissspidy: I borrowed this from Gutenberg's middleware implementation for the most part 🤷‍♂ But your suggestion makes sense. I think it can be simplified in general.

assets/src/stories-editor/store/reducer.js
```diff
@@ -19,10 +20,10 @@ import { isValidAnimationPredecessor } from './selectors';
  *
  * @return {Object} Updated state.
  */
-export function animations( state = {}, action ) {
-	const newAnimationOrder = { ...state };
+export function animations( state = {}, action ) { // eslint-disable-line complexity
```
barklund: Cyclomatic complexity hits 22 for this function, which definitely hints that a rewrite or reorganisation would be a good idea - if for nothing else to break out the different cases into smaller functions. It's a 1000 sloc function!

Redux has [a good guide](https://redux.js.org/recipes/structuring-reducers/refactoring-reducer-example) for breaking reducers into smaller components.

assets/src/stories-editor/store/middlewares.js
swissspidy: Hmm applying the suggestions actually breaks things, throwing `Dispatching while constructing your middleware is not allowed` errors in `assets/src/stories-editor/store/index.js`.

assets/src/stories-editor/helpers/index.js
```diff
@@ -1719,3 +1744,222 @@ export const processMedia = ( media ) => {
...
+	const targetFitsWithinPage = width <= STORY_PAGE_INNER_WIDTH || height <= STORY_PAGE_INNER_HEIGHT;
+
+	if ( targetFitsWithinPage ) {
+		const scalingFactor = 1.25;
+
+		const widthFactor = STORY_PAGE_INNER_WIDTH > width ? STORY_PAGE_INNER_WIDTH / width : 1;
+		const heightFactor = STORY_PAGE_INNER_HEIGHT > height ? STORY_PAGE_INNER_HEIGHT / height : 1;
+
+		return Math.max( widthFactor, heightFactor ) * scalingFactor;
+	}
```
barklund: But what if the element is already larger than the page, but only by a tiny amount, say 1%? Then this function would suggest not to scale the element, even thought it's smaller than the required size, no?

Maybe some unit tests would be helpful for this simple utility function.

assets/src/stories-editor/store/middlewares.js
barklund: Hm, that sounds weird. It shouldn't change anything?

Well, you can of course just ignore my suggestions. If it ain't broken and all that.

But maybe a comment would be helpful.

assets/src/stories-editor/helpers/index.js
swissspidy: FWIW, this method is basically a straight copy from the AMP framework to make sure the logic is the same: https://github.com/ampproject/amphtml/blob/13b3b6d92ee0565c54ec34732e88f01847aa8a91/extensions/amp-story/1.0/animation-presets-utils.js#L91-L111

assets/src/stories-editor/blocks/amp-story-page/animation-settings.js
swissspidy: I thought about it, but it seemed like an unnecessary abstraction for such a small component.

swissspidy:
> It might be a good idea to abort the animation playback as soon as the page loses focus/selection.

Gutenberg has a [`withFocusOutside`](https://github.com/WordPress/gutenberg/tree/50fb29bb2181e56a4a2d4c5407ba23c8bba2c4da/packages/components/src/higher-order/with-focus-outside) HOC to detect when focus leaves an element, but it only works for class components. Maybe I can replicate that with `useEffect`...

swissspidy:
> But if I delete the element outright without first setting the animation to "none", the "play all" button sticks around still thinking there's an animation on the page.

Ah! Since the animation store doesn't really do garbage collection for deleted elements, the list included invalid items. Can circumvent this by filtering the list directly in the component.

assets/src/stories-editor/store/reducer.js
swissspidy: Extracted the individual cases into their own functions now. Not 100% DRY still, but it's a start.

swissspidy:
> If I create a cascading sequence of animations, so element 1 plays immediately, element 2 plays after element 1 and element 3 plays after element 2, the "play all animations" action gets stuck after the second animation - the third one never starts.

This should be fixed now

swissspidy:
> It might be a good idea to abort the animation playback as soon as the page loses focus/selection.

Haven't found an easy way to do this. I am inclined to leave this extra hardening for later so we can get this in for now.

Issue #1949, opened 2019-03-11
AMP consent for analytics configuration


p2er:
## What's the issue?

AMP Consent can not be bound to analytics config defined via plugin page. Should supply interface as described here: https://www.ampproject.org/docs/reference/components/amp-consent#promptifunknownforgeogroup

## How do we reproduce the issue?

1. Open analytics config page (wp-admin/admin.php?page=amp-analytics-options)
2. Enter config
3. Should provide field for consent configuration

## Info

AMP may support more advanced pre-defined blocking behaviors in the future. Because of this, the value of data-block-on-consent is reserved only for the following supported pre-defined attributes:

**_till_responded** : Unblock the component until the user has responded to the consent prompt, or the consent prompt has been skipped.
**_till_accepted** : Default basic blocking behavior, expect that when _till_accepted is explicitly added, individual components cannot override the blocking behavior.
**_auto_reject** : Always reject the consent automatically if consent is required but unknown. The reject consent decision will not be stored. It is recommended not to specify a consent prompt UI when auto rejecting consent for every components.

westonruter:
In other words, you're looking for a UI to supply the `data-block-on-consent` attribute of the `amp-analytics` element, correct?

Related: https://github.com/ampproject/amp-wp/issues/775#issuecomment-458752732

p2er:
Yes, correct. Right now tracking starts with and without consent unless I implement the analytics configuration template wise.

westonruter:
Adding the `<amp-analytics>` element in the template is a good approach in the mean time.

p2er:
Is there a way to override the function hook and use the configuration passed through including the JSON and analytics type? That would also be helpful to parse the document and remove config parts that use unavailable selectors on that rendered page. For instance: Admin configures hook to amp image in order to check if teaser has been seen by the user. But some templates do and some do not contain said teaser. Adding the selector in the configured JSON would cause JS Errors on pages where the image is missing. Maybe there is a way for you guys to integrate this in the css sanitizer at some point.

westonruter:
Do you mean the [`amp_print_analytics`](https://github.com/ampproject/amp-wp/blob/756c35475982e36eeb28511e96e82a4f68941372/includes/amp-helper-functions.php#L647) function? It is not currently overridable. 

[...]

Issue #2202, opened 2019-04-26
Send entire templates in Reader/Classic mode through the sanitizers


westonruter:
Currently only content (and the featured image) are sent through the sanitizers. This is the classic approach, and it is why there's the _content_ in `amp_content_sanitizers`. However, limiting the sanitizers to the content has some problems which would be fixed by sanitizing the entire page: 

1. Markup that is output outside the content (such as in `amp_post_template_head` or `amp_post_template_footer`) is not sent through the sanitizers, resulting in the possibility of invalid AMP pages when plugins don't implement AMP properly. This has come up frequently in the support forums. By sanitizing the entire document, this would no longer happen frequently.
2. The CSS in the Reader mode templates is output in a `style.php` template and the `amp_post_template_css` action. The style rules are not sent through the CSS sanitizer. This means that they are neither minified nor tree-shaken. If the entire template were sent through the sanitizers, then this would be a key benefit (and would resolve #688).
3. By sending the entire reader mode templates through the sanitizers, this closer aligns with traditional theme templates and will facilitate transitioning from those classic post templates to standard theme templates (e.g. where `$this->load_parts()` is not used).
4. There are currently two entirely different code-paths for Reader vs non-Reader modes. There are opportunities to better harmonize the two, if we can use the code in `AMP_Theme_Support` for both. 

----

This should make Reader mode templates available for Polylang to translate ([topic](wordpress.org/support/topic/static-home-page-with-polylang-2/)). 

By making Reader mode templates regular WordPress templates, we'll also be able to avoid situations where users copy Reader mode templates to be used in Transitional mode, only to find that they start getting fatal errors due to `$this`. (Examples: [1](https://wordpress.org/support/topic/cant-customize-templates-uncaught-error-using-this-when-not-in-object-conte/), [2](https://wordpress.org/support/topic/childlike-plugin-mods/), [3](https://wordpress.org/support/topic/page-unable-to-reach/)). 

**Update:** This is probably a bit out of scope here. I think we should take the existing Reader mode post templates and port them to regular WordPress theme templates, but when we do so we need to introduce generic `amp_head` and `amp_footer` actions (and the like) which run instead of `amp_post_template_head` (à la `wp_head`) and `amp_post_template_footer` (à la `wp_footer`), similar to what we did for stories with `amp_story_head`. The `amp_post_template_*` actions should be deprecated because they are passed a `AMP_Post_Template` instance which is not relevant when templates can render any WordPress query, including index pages.

[...]

PR #3106, opened 2019-08-27
Add amp_print_analytics action triggering before printing entries as amp-analytics

westonruter:
This is useful for printing additional `amp-analytics` tags to the page without having to refactor any existing markup generation logic to use the data structure mutated by the `amp_analytics_entries` filter. For such cases, this action should be used for printing `amp-analytics` tags as opposed to using the `wp_footer` and `amp_post_template_footer` actions; this will ensure analytics will also be included on AMP Stories.

See:

* https://github.com/ampproject/amp-wp/pull/3095#issuecomment-525402333
* https://github.com/google/site-kit-wp/issues/435
* https://github.com/ampproject/amp-wp/issues/1949

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -673,7 +673,21 @@ function amp_print_analytics( $analytics ) {
 	if ( '' === $analytics ) {
 		$analytics = [];
 	}
+
 	$analytics_entries = amp_get_analytics( $analytics );
+
+	/**
+	 * Triggers before analytics entries are printed as amp-analytics tags.
+	 *
+	 * This is useful for printing additional `amp-analytics` tags to the page without having to refactor any existing
+	 * markup generation logic to use the data structure mutated by the `amp_analytics_entries` filter. For such cases,
+	 * this action should be used for printing `amp-analytics` tags as opposed to using the `wp_footer` and
+	 * `amp_post_template_footer` actions; this will ensure analytics will also be included on AMP Stories.
+	 *
+	 * @since 1.3
+	 * @param array $analytics_entries Analytics entries, already potentially modified by the amp_analytics_entries filter.
+	 */
+	do_action( 'amp_print_analytics', $analytics_entries );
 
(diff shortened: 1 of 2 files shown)


westonruter:
> Question regarding Reader mode: Does the `amp-analytics` extension get loaded there automatically? I was under the assumption that, so far, it only was if there were actual Analytics entries in the AMP plugin's option. However, now that there is the hook, there might be a need for the extension even without entries from the option.

@felixarntz Excellent point. Currently if the `$entries` is empty then the `amp-analytics` extension will not be included by default:

https://github.com/ampproject/amp-wp/blob/7feaf9648a3cd23f7565f2c7a098b1fd87766076/includes/templates/class-amp-post-template.php#L139

https://github.com/ampproject/amp-wp/blob/7feaf9648a3cd23f7565f2c7a098b1fd87766076/includes/admin/functions.php#L154-L171

https://github.com/ampproject/amp-wp/blob/7feaf9648a3cd23f7565f2c7a098b1fd87766076/includes/amp-post-template-functions.php#L121-L132

So you will need to make sure that Site Kit does:

```php
add_filter(
	'amp_post_template_data',
	function( $data ) {
		$data['amp_component_scripts'] = array_merge(
			$data['amp_component_scripts'],
			array(
				'amp-analytics' => true,
			)
		);
		return $data;
	}
);
```

This will be needed until the entire Reader mode templates are sent through the sanitizers, at which point the required components will be automatically detected. See https://github.com/ampproject/amp-wp/issues/2202 is landed

felixarntz:
@westonruter 
> Currently if the `$entries` is empty then the `amp-analytics` extension will not be included by default:

For reader mode, Site Kit already covers that. How about AMP Stories though? How do I make sure the extension gets loaded in the AMP Stories template? That happens by the sanitizer too I assume?

westonruter:
Correct. The AMP Stories single template is post-processed in its entirety, so any component script requirements are auto-detected.

Issue #2699, opened 2019-06-27
Move adding background media to the toolbar, or a readily visible place.


miina:
Based on testing, most of the users struggle with adding background media to a Page, most of the users try to do it via adding an image block. Also, the place for adding Background Media doesn't seem to be obvious.

Some potential improvement ideas so far from discussion with @swissspidy:
- Adding functionality to an Image block for converting it into the background image.
- Adding a button to the Page's Toolbar for adding a background image.
- Alternatively, we could use an icon for the Image / Background Image and this could potentially prompt the user to choose which to add.

**Acceptance criteria based on backlog grooming, added by @MackenzieHartung on Aug 21:**

As a user, I want an easier way to add background media to a page in my story.

AC1 - The existing image icon in the top tool bar will be the access point for a user to add media (video or image) to the page.
AC2 - If it is necessary for clarity, the icon can change to represent a broader media (video and image)
AC3 - When selecting the icon in the top tool bar the user will be presented with the option to add background image, inline image, background video and inline video.
AC4 - Selecting a specific option will take the user to the same location they would go today when making the selection to add background media or inline media today.
AC5 - Have the Background Color Panel open by default for the Page block.

spacedmonkey:
Could the background image use the featured image? We could filter and change the name of the featured image to background image. Not sure if this would cause other issues with meta data for example. 

This issue reminds me a little of this - https://github.com/WordPress/gutenberg/issues/13795

miina:
The Background Media is set per Page as the background of a single AMP Story Page, it can be either an image or a video.

For example 2 Pages, different background images:
<img width="678" alt="Screenshot 2019-08-21 at 17 16 01" src="https://user-images.githubusercontent.com/3294597/63439796-65b33700-c437-11e9-9fb3-6e14179816d9.png">


It's not mandatory either, the reason for the issue was that the users did not find the place where to set the background image/video for a Page, they tried to do it via clicking on the icon which adds a regular Image block.

spacedmonkey:
Put some thought into this one. Here is my recommendation. 

- Add a new Amp Story Background block. 
   - User are only able to unable to insert one background block. 
   - After background block is added and selected, all other blocks are not displayed. Background block will take the full page. 
   - Block will have existing options from sidebar and will live reload.  
   - Once background is deselected, UI for block is hidden. 
[...]

PR #3109, opened 2019-08-28
Change wording in media dropdown, so that if background is set, the change …

spacedmonkey:
Change wording in media dropdown, so that if background is set, the change from insert to update. Add e2e tests.

See feedback in #2699


assets/src/stories-editor/components/media-inserter/index.js
```diff
@@ -21,21 +21,35 @@ const POPOVER_PROPS = {
...
 	const blocks = [
 		'core/video',
 		'core/image',
 	];
+	const currentBlock = getCurrentPageAttributes();
+
+	let imageTitle = __( 'Insert Background Image', 'amp' );
+	let videoTitle = __( 'Insert Background Video', 'amp' );
+	if ( currentBlock.mediaId ) {
+		switch ( currentBlock.mediaType ) {
```
swissspidy: ESLint complains about a missing default case.

I'd probably just write something like this anyway:

```js
const imageTitle = 'image' === currentBlock.mediaType ? __( 'Update Background Image', 'amp' ) : __( 'Insert Background Image', 'amp' );
const videoTitle = 'video' === currentBlock.mediaType ? __( 'Update Background Video', 'amp' ) : __( 'Insert Background Video', 'amp' );
```

assets/src/stories-editor/components/media-inserter/index.js
```diff
@@ -118,6 +133,14 @@ const applyWithSelect = withSelect( ( select ) => {
 		},
 		// As used in <HeaderToolbar> component
 		showInserter: select( 'core/edit-post' ).getEditorMode() === 'visual' && select( 'core/editor' ).getEditorSettings().richEditingEnabled,
+		getCurrentPageAttributes: () => {
```
swissspidy: Can't we just directly pass on the `mediaType` and `mediaId` attributes here instead of a function that returns all the attributes? `mediaType` only might even be enough.

assets/src/stories-editor/components/media-inserter/index.js
spacedmonkey: Resolved in https://github.com/ampproject/amp-wp/pull/3109/commits/ba5a34c8189deeeede2de789246adbde189d7886

spacedmonkey:
Feedback actioned in https://github.com/ampproject/amp-wp/pull/3109/commits/ba5a34c8189deeeede2de789246adbde189d7886

assets/src/stories-editor/components/media-inserter/index.js
```diff
@@ -21,21 +21,25 @@ const POPOVER_PROPS = {
 	position: 'bottom right',
 };
 
-const MediaInserter = ( { insertBlock, updateBlock, canInsertBlockType, showInserter } ) => {
+const MediaInserter = ( { insertBlock, updateBlock, canInsertBlockType, showInserter, getCurrentPageBackgroundType } ) => {
 	const blocks = [
 		'core/video',
 		'core/image',
 	];
+	const mediaType = getCurrentPageBackgroundType();
```
barklund: Would it not be better to get this information through the `useSelect` hook? As it is now, I don't think this component will re-render if the media type changes.

And if `useSelect` isn't available yet (I'm not sure we're on the right version), at least use the `withSelect` HoC.

barklund: I think you can remove the prop (and the passing of the prop) and replace this line with:

```suggestion
	const mediaType = useSelect(
		( select ) => {
			const clientId = select( 'amp/story' );
			const blocks = select( 'core/block-editor' ).getBlocksByClientId( clientId );
			return (blocks && blocks.shift().attributes.mediaType) || '';
		}
	);
```

Of course you need to import `useSelect` somewhere.

spacedmonkey: It does use select here https://github.com/ampproject/amp-wp/pull/3109/files#diff-d718b69a2e2ea3f9c4ccfd8999fa59f3R126-R134 .

barklund: But not `useSelect` which is a hook forcing the component to re-render if the (relevant part of the) store changes.

spacedmonkey: What issue would that solve here?

barklund: The code as it is has the problem, that if the store updates outside of this component, this component would not update to reflect the correct state – e.g. if the background image is removed via another component, this component would still display "Update Background Image" even though there is no current background image.

Using the `useSelect` hook (or the `withSelect` HoC), React will force this component to re-render with the correct state, regardless of what component or method updates the media type of the current background element.

swissspidy: @barklund But this component **does** use `withSelect`.

See the `applyWithSelect` part at the bottom:

https://github.com/ampproject/amp-wp/blob/7451bcc7024fcf0385bdff3d2e13e6bccf6f31aa/assets/src/stories-editor/components/media-inserter/index.js#L112-L137

swissspidy: As for `useSelect`, we cannot reliably use it just yet. See https://github.com/ampproject/amp-wp/issues/2538#issuecomment-526165456

barklund: > @barklund But this component **does** use `withSelect`.

Ah, I did not catch that. Thanks for clarifying. I misunderstood because the argument was given as a callback, not as a direct value.

Is there any reason `mediaType` isn't directly provided as a prop to the component?

swissspidy: For consistency reasons I'd also say directly passing `mediaType` makes sense 👍

spacedmonkey: > For consistency reasons I'd also say directly passing `mediaType` makes sense

Do you mean changing from passing a function to passing a variable?

swissspidy: Yes

spacedmonkey: Based on your feedback. https://github.com/ampproject/amp-wp/pull/3109/commits/c77650087ac7e82a45f85d85fd8a7a340b266b6f

The feedback about `useSelect` is very valid, as the background can be updated from the sidebar still. Not sure what we can do about this apart from use `useSelect`

swissspidy: Since `withSelect` uses `useSelect` under the hood, I don't see an issue here.

spacedmonkey: Screenshot from 2019-08-30 16-18-19
Screenshot is when I update background from the sidebar. The side should have updated :thinking:

spacedmonkey:
Something changed upstream with `getBlocksByClientId` which broke my code, after a rebase. I have changed to use `getBlock`, which is cleaner and works now. Becuase I using selects, the title updates when something is updated from either place.

spacedmonkey:
Can someone re-review this please.

tests/e2e/specs/stories-editor/media-inserter.js
```diff
@@ -68,6 +68,28 @@ describe( 'Stories Editor Screen', () => {
 		expect( src ).toContain( 'wp-content/uploads' );
 	} );
 
+	it( 'should be possible to update Background Image', async () => {
+		// Click the media selection button.
+		await page.waitForSelector( '.amp-story-media-inserter-dropdown' );
+		await page.click( '.amp-story-media-inserter-dropdown' );
```
miina: Looks like these two lines might deserve a little helper such as "openMediaInserter" or something like that? Seems to be used in most of the tests, sometimes more than once (like here).

Also, could we add a test which would verify that when inserting Media from the Sidebar, the text changes correctly in the Inserter as well?

assets/src/stories-editor/components/media-inserter/index.js
```diff
@@ -109,6 +113,17 @@ const applyWithSelect = withSelect( ( select ) => {
...
 
+	const getCurrentPageBackgroundType = () => {
+		const clientId = getCurrentPage();
+		const block = select( 'core/block-editor' ).getBlock( clientId );
+		if ( ! block ) {
+			return '';
+		}
+	
+		const mediaType = ( block.attributes.mediaType ) ? block.attributes.mediaType : '';
+		return mediaType;
```
miina: This could just be `return ( block.attributes.mediaType ) ? block.attributes.mediaType : '';`

spacedmonkey: I personally find it easy to read as is. I don't love having logic in the return statement, for readability. But I don't feel strongly about this.

tests/e2e/specs/stories-editor/media-inserter.js
spacedmonkey: Resolved in https://github.com/ampproject/amp-wp/pull/3109/commits/968ebfd96edacebefecfcdf572a72598806248be

assets/src/stories-editor/components/media-inserter/index.js
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3109/commits/c053210919f8d02ec8da3b1a5f65ada5d6129891

assets/src/stories-editor/components/media-inserter/index.js
```diff
@@ -118,6 +132,7 @@ const applyWithSelect = withSelect( ( select ) => {
 		},
 		// As used in <HeaderToolbar> component
 		showInserter: select( 'core/edit-post' ).getEditorMode() === 'visual' && select( 'core/editor' ).getEditorSettings().richEditingEnabled,
+		mediaType: getCurrentPageBackgroundType(),
```
miina: Just realized that the `Shortcuts` component is using almost identical `withSelect` logic except for the added `mediaType`.

Is there a way how to combine those?
For example:
These are both functional components so we could do the following instead:
- Create a helper hook for the `withSelect` logic that matches and reuse it in both components directly, remove the `withSelect`.
- Use an additional `useSelect` for the `mediaType` within `MediaInserter` directly.

Not sure how much that makes sense here. Thoughts?

spacedmonkey: I know that the logic is repeated here. But I am not sure of the benefit of a helper. At some point the code might change the logic might need to differ. Shared code might not be useful here. With mediaType, there is enough of a different to overlook a very small amount of repeated code here.

barklund: Generalising this can be tricky, as it can be hard to find the right separation, but I think the common redux way would be to move the logic to selectors, and then reuse those. So this function would be:

```
import {
	selectIsReordering,
	selectCanInsertBlockType,
	selectShowInserter,
	selectMediaType,
} from './some/selector';

...

const applyWithSelect = withSelect( ( select ) => {
	const isReordering = selectIsReordering( select );
	const canInsertBlockType = selectCanInsertBlockType( select );
	const showInserter = selectShowInserter( select );
	const mediaType = selectMediaType( select );

	return { isReordering, canInsertBlockType, showInserter, mediaType };
} );
```

This does help with unit testing (as the applySelect invocation is completely trivial now and can be ignored for testing purposes, wheres the selectors are testable units).

However if it's not common practice to move the selector logic to external helpers, it makes the coding practice inconsistent.

miina: Thanks for chiming in, @barklund, makes sense!

PR #3110, opened 2019-08-28
Font Picker: Bundle list of Google Fonts 

### LINKED ISSUE #3026: Font Picker: Bundle list of Google Fonts
As outlined in #2877, we need to download a JSON file of Google Fonts that we can then display in the font family picker.

* [ ] Download list of fonts to bundle in the plugin.
* [ ] Document process of how the JSON file can be downloaded and kept up-to-date.
* [ ] E2E (#2900 )

spacedmonkey:
Fixes #3026


includes/class-amp-story-post-type.php
```diff
@@ -986,184 +986,21 @@ public static function get_fonts() {
...
-				'fallbacks' => [ 'sans-serif' ],
-			],
-			[
-				'name'      => 'Verdana',
-				'fallbacks' => [ 'Geneva', 'sans-serif' ],
-			],
-		];
+		$file_content = file_get_contents( __DIR__ . '/data/fonts.json' );  // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+		$google_fonts = json_decode( $file_content, true );
+		$fonts        = [];
```
swissspidy: `$fonts` should still include default fonts from before like Arial, Verdana, etc.

Then, after adding the Google Fonts to the list, `$fonts` would need to be sorted by font name.

includes/class-amp-story-post-type.php
```diff
@@ -1190,7 +1027,7 @@ public static function get_fonts() {
 			static function ( $font ) use ( $fonts_url, $subsets ) {
 				$font['slug'] = sanitize_title( $font['name'] );
 
-				if ( isset( $font['gfont'] ) ) {
+				if ( isset( $font['gfont'] ) && ! empty( $font['gfont'] ) ) {
```
swissspidy: When using `empty()`, the `isset` check is redundant.

spacedmonkey: No it's not, my new code sets an empty `gfont`

includes/class-amp-story-post-type.php
spacedmonkey: This was not clear from the original issue.

includes/class-amp-story-post-type.php
```diff
@@ -986,184 +986,21 @@ public static function get_fonts() {
...
-				'name'      => 'Verdana',
-				'fallbacks' => [ 'Geneva', 'sans-serif' ],
-			],
-		];
+		$file_content = file_get_contents( __DIR__ . '/data/fonts.json' );  // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+		$google_fonts = json_decode( $file_content, true );
+		$fonts        = [];
+		foreach ( $google_fonts['items'] as $font ) {
+			$gfont    = '';
+			$variants = array_diff( $font['variants'], [ 'regular' ] );
```
swissspidy: I think `regular` should not be excluded, but instead replaced with `400`.

Example:

When I want to download the Ubuntu font from Google Fonts in regular and bold variants, I get a URL like this:

```
https://fonts.googleapis.com/css?family=Ubuntu:400,700&display=swap
```

includes/class-amp-story-post-type.php
swissspidy: What I meant is that this:

```php
if ( isset( $font['gfont'] ) && ! empty( $font['gfont'] ) ) {}
```

is equal to this:

```php
if ( ! empty( $font['gfont'] ) ) {}
```

`empty()` does not generate a warning if the variable does not exist, so it's safe to remove the `isset` check.

includes/class-amp-story-post-type.php
swissspidy: Sorry for the confusion then! Shall I update the issue to be more precise in general?

includes/class-amp-story-post-type.php
spacedmonkey: Can you explain this?

includes/class-amp-story-post-type.php
spacedmonkey: I can remove the isset check if you want.

includes/class-amp-story-post-type.php
swissspidy: Taking the `Ubuntu` font here as an example:

In the `develop` branch, the `gfont` entry for that font is `Ubuntu:400,700`.

In `fonts.json`, the following variants are listed for `Ubuntu`:

```
[
	"300",
	"300italic",
	"regular",
	"italic",
	"500",
	"500italic",
	"700",
	"700italic"
]
```

In this PR, this is turnt into `Ubuntu:300,300italic,italic,500,500italic,700,700italic`, whereas the correct entry ([as output by the Google Fonts UI](https://fonts.google.com/specimen/Ubuntu?selection.family=Ubuntu:300,300i,400,400i,500,500i,700,700i)) would be `Ubuntu:300,300i,400,400i,500,500i,700,700i`.

So `regular` is turned into `400`, and `300italic` is turned into `300i`, `700italic` into `700i`, etc.

Given that we only support regular, italic, and bold in the stories editor UI, I would strip out everything else, so that we get `Ubuntu:400,400i,700,700i` in the end.

Hope that makes sense.

If not, happy to amend a test case to demonstrate this.

spacedmonkey: Regular should excluded here. By loading any font, it will just automatically load the regular. Take this for example https://fonts.googleapis.com/css?family=Ubuntu . Just loads the regular (400). So it is not required. Which is why it was ignored.  

Good spot, `700italic` -> `700i`. But this is a WIP, so I hadn't tests the fonts loading yet.

spacedmonkey:
> Worth noting that this only affects the stories editor, and not the rest of the admin. However, is there any chance we can reduce the size of the JSON file after the download? There seem to be some fields in it that we don't need that could be removed.

In https://github.com/ampproject/amp-wp/pull/3110/commits/1a1796494ebfa646ae67eaee5f33e9a51f07e939 I wrote a script to script out unneed fields. This take the json file from 512kb -> 72kb.

spacedmonkey:
> So far it looks good, but some tests for the new font list assembling would be good.
@swissspidy I don't that e2e are a good idea. I believe the focus should of testing should be on #3155 as the autocomplete is the help behaviour change here. Other kinds of tests here also don't really work. Unit tests in grunt code doesn't make a lot of sense. 

Did anyone have any other kind of tests in mind?

includes/class-amp-story-post-type.php
spacedmonkey: Resolved https://github.com/ampproject/amp-wp/pull/3110/commits/e725b9fc55bc16fa7e0614f36b28a900a6d07431

includes/class-amp-story-post-type.php
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3110/commits/3a16fadde9a3858d339de32d187df9501063b69d

swissspidy:
> This take the json file from 512kb -> 72kb.

Nice!

> Did anyone have any other kind of tests in mind?

Unit tests for `AMP_Story_Post_Type::get_fonts()` and `AMP_Story_Post_Type::get_fallback()` would be nice. I know we haven't had some before, but now with the added complexity I think it makes sense.

.gitignore
```diff
@@ -12,3 +12,5 @@ assets/js/*.deps.json
 assets/js/*.map
 built
 /amphtml
+.env
+.idea/
```
swissspidy: Might be something for your global `.gitignore` file :-)

.gitignore
```diff
@@ -12,3 +12,5 @@ assets/js/*.deps.json
 assets/js/*.map
 built
 /amphtml
+.env
```
swissspidy: Should we add a `.env.example` file with `GOOGLE_FONTS_API_KEY=` as the content? (yes I would rename it). Then we can document it in `contributing.md`

.gitignore
spacedmonkey: I think, it is helpful for devs that join the project.

contributing.md
```diff
@@ -172,6 +175,18 @@ To get a production build, run:
 npm run build:js
 ```
 
+### Updating Google fonts 
+
+Before updating Google Fonts, an `.env` is required to add the Google Fonts API key. 
+Details of how to get an API key can be found on in the [Google fonts docs](https://developers.google.com/fonts/docs/developer_api). 
+To add an API. 
+
+1. Copy `.env.example` to `.env`.
```
swissspidy:
```suggestion
1. Copy `example.env` to `.env`. 
```

contributing.md
```diff
@@ -62,6 +62,9 @@ If you have an incompatible version of Node in your development environment, you
 nvm install
 ```
 
+A [Google Fonts API Key](https://developers.google.com/fonts/docs/developer_api)  is required to build or update Google font list. May not be require for many contributors.
```
swissspidy:
```suggestion
A [Google Fonts API Key](https://developers.google.com/fonts/docs/developer_api)  is required to build or update Google font list. May not be required for many contributors.  
```

contributing.md
```diff
@@ -172,6 +175,18 @@ To get a production build, run:
...
 
+### Updating Google fonts 
+
+Before updating Google Fonts, an `.env` is required to add the Google Fonts API key. 
+Details of how to get an API key can be found on in the [Google fonts docs](https://developers.google.com/fonts/docs/developer_api). 
+To add an API. 
+
+1. Copy `.env.example` to `.env`. 
+1. Replace `replacemewithrealkey` with API key. 
+1. Run `Grunt download-fonts`
```
swissspidy: Should we perhaps rather recommend running `npm run build:download-fonts`? This is currently the point of entry for most of the tasks.

contributing.md
spacedmonkey: FIxed in https://github.com/ampproject/amp-wp/pull/3110/commits/1c057e221ad70b2f4c3cf421f7faee74a000331a

contributing.md
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3110/commits/1c057e221ad70b2f4c3cf421f7faee74a000331a

contributing.md
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3110/commits/1c057e221ad70b2f4c3cf421f7faee74a000331a

.gitignore
spacedmonkey: Done

spacedmonkey:
> Unit tests for `AMP_Story_Post_Type::get_fonts()` and `AMP_Story_Post_Type::get_fallback()` would be nice. I know we haven't had some before, but now with the added complexity I think it makes sense.

@swissspidy Unit tested for these functions have been added. Through I have refactored the code to make it easier to test. 

I have also updated the docs as requested.

tests/php/test-class-amp-story-post-type.php
```diff
@@ -542,4 +542,91 @@ public function test_render_block_with_grid_layer_with_not_movable_block() {
...
+		$this->assertArrayHasKey( 'gfont', $fonts[0] );
+		$this->assertArrayHasKey( 'extra', $fonts[0] );
+	}
+
+	/**
+	 * Test fallback font.
+	 *
+	 * @covers AMP_Story_Post_Type::get_fallback
+	 */
+	public function test_fallback_font() {
```
swissspidy: This only really tests the `default` `case` in that method, no?

I'd use a quick data provider here to test all cases + the default fallback.

includes/class-amp-story-post-type.php
```diff
@@ -1209,6 +1113,94 @@ static function ( $font ) use ( $fonts_url, $subsets ) {
...
+	}
+
+	/**
+	 * Helper method to lookup fallback font.
+	 *
+	 * @param string $category Google font category.
+	 *
+	 * @return string $fallback Fallback font.
+	 */
+	public static function get_fallback( $category ) {
```
swissspidy: Hmm we should rename that to `get_font_fallback` to make it more clear what it is about.

tests/php/test-class-amp-story-post-type.php
```diff
@@ -542,4 +542,91 @@ public function test_render_block_with_grid_layer_with_not_movable_block() {
...
+				'name'      => 'ABeeZee',
+				'fallbacks' => [ 'sans-serif' ],
+				'gfont'     => 'ABeeZee:400,400i',
+				'extra'     => 'extra field for test',
+			],
+		];
+		$fonts          = AMP_Story_Post_Type::get_google_fonts( $original_fonts, $file );
+		$this->assertTrue( is_array( $fonts ) );
+		$this->assertEquals( 952, count( $fonts ) );
+		$this->assertArrayHasKey( 'name', $fonts[0] );
```
swissspidy: Doesn't this just test the `ABeeZee` font entry?

It'd be good to have a test for one of the fonts from the JSON file. Especially for the font variants logic in the `gfont` entry. Not just asserting that the array key exists.

spacedmonkey: There are 952 fonts in the file, I just picked on test. It doesnt make any sense to test them all.

includes/class-amp-story-post-type.php
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3110/commits/014f71fefe9eee5533d4a6cc7ad3bb5f370572dd

tests/php/test-class-amp-story-post-type.php
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3110/commits/014f71fefe9eee5533d4a6cc7ad3bb5f370572dd

tests/php/test-class-amp-story-post-type.php
swissspidy: Not all of them, but at least one, and not only `assertArrayHasKey`.

Also, isn't `$fonts[0]` the `ABeeZee` entry? Or is it something else?

spacedmonkey:
:tada:

PR #3117, opened 2019-08-29
Add href attribute sanitizer to catch invalid URLs

### LINKED ISSUE #2671: Invalid URL like `foo@` passed through without sanitization causing AMP validation error

schlessera:
Some invalid URLs are being passed over from WordPress and rendered into clickable links.

This new sanitizer targets the `href` attribute and makes sure it is a valid URL. If not, it remove the `href` attribute. This change leaves the visual output intact, but makes the perceived link not clickable and avoids AMP validation errors.

Fixes #2671


schlessera:
I tried to do this via the Tag & Attribute sanitizer first:
Image 2019-08-29 at 10 28 04 AM

But this leaves an empty `href` attribute in place, which I assume can cause side-effects in terms of accessibility:
Image 2019-08-29 at 10 27 22 AM

This new sanitizer completely removes the offending `href` attribute instead, which seems preferable:
Image 2019-08-29 at 10 28 36 AM

swissspidy:
> But this leaves an empty href attribute in place, which I assume can cause side-effects in terms of accessibility

Indeed a missing `href` has a different meaning than an empty attribute, [according to the spec](https://www.w3.org/TR/2016/REC-html51-20161101/textlevel-semantics.html#the-a-element).

> If the a element has no href attribute, then the element represents a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element’s contents.

> The target, download, rel, rev, hreflang, and type attributes must be omitted if the href attribute is not present.

----

The sanitizer currently strips out too many valid things like `<link rel="dns-prefetch" href="//cdn.ampproject.org">`. or `<a href="#foo">`.

`filter_var( '//cdn.ampproject.org', FILTER_VALIDATE_URL )` returns `false`.
`filter_var( '#hello', FILTER_VALIDATE_URL )` returns `false` too.

Test output:

```
There were 3 failures:
1) AMP_Style_Sanitizer_Test::test_amp_selector_conversion with data set "attribute_selectors" ('<div id="content" tabindex="-.../span>', '[type="button"], [type="reset...red; }', '[type="button"],[type="reset"...x red}')
--- Expected
+++ Actual
@@ @@
-'[type="button"],[type="reset"],[type^="submit"]{color:red}a[href^=http]:after,a[href^="#"]:after{color:blue}span[hidden]{display:none}#content[tabindex="-1"]:focus{outline:solid 1px red}'
+'[type="button"],[type="reset"],[type^="submit"]{color:red}span[hidden]{display:none}#content[tabindex="-1"]:focus{outline:solid 1px red}'
/tmp/wordpress/src/wp-content/plugins/amp/tests/php/test-amp-style-sanitizer.php:712
2) AMP_Style_Sanitizer_Test::test_attribute_selector with data set "href_attribute" ('<a href="foo">Foo</a>', array(true, true))
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'a[href^="#"]:after'
-    1 => 'a[href^=http]:after'
 )
/tmp/wordpress/tests/phpunit/includes/abstract-testcase.php:658
/tmp/wordpress/src/wp-content/plugins/amp/tests/php/test-amp-style-sanitizer.php:862
3) Test_AMP_Theme_Support::test_prepare_response
Failed to find: <link rel="dns-prefetch" href="//cdn.ampproject.org">
Failed asserting that false is not false.
```

swissspidy:
> I tried to do this via the Tag & Attribute sanitizer first:
> But this leaves an empty href attribute in place

Could this still be part of the Tag & Attribute sanitizer, and then this sanitizer would just remove empty `href` attributes? Or perhaps another sanitizer could already be used for that.

schlessera:
> Could this still be part of the Tag & Attribute sanitizer, and then this sanitizer would just remove empty href attributes? Or perhaps another sanitizer could already be used for that.

The main issue remains that of "how to check wether a URL is valid". If we do that in the Tag & Attribute sanitizer, we still need an additional pass to remove empty hrefs. So we're not saving anything from moving the logic into the Tag & Attribute sanitizer, it would just be more convoluted.

In the above example, where you have the empty href attribute, this would still invalidate false-positives just like the current solution.

schlessera:
I tried a different approach now, basically using `wp_parse_url( esc_url( $url ) )`.

schlessera:
I'll add this logic as well:
> The target, download, rel, rev, hreflang, and type attributes must be omitted if the href attribute is not present.

schlessera:
The code now also removes left-over attributes that should be omitted according to the spec.

Before:
Image 2019-08-29 at 10 28 36 AM

After:
Image 2019-08-29 at 1 10 48 PM

includes/sanitizers/class-amp-href-sanitizer.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @throws LogicException If the xpath query produced invalid results.
+	 */
+	public function sanitize() {
+		$xpath = new DOMXPath( $this->dom );
+
+		foreach ( $xpath->query( self::XPATH_SELECTOR ) as $node ) {
+			// This should not happen, something is wrong with our query.
+			if ( ! $node instanceof DOMElement
+				|| ! $node->hasAttribute( 'href' ) ) {
+				throw new LogicException(
```
schlessera: Do we use asserts and logic exceptions for defensive development to make the code more resilient (and document it at the same time)?

kienstra:
**Travis Build**

Hi @schlessera,
Here's a diff that made the [failed e2e test in Travis](https://travis-ci.org/ampproject/amp-wp/jobs/578385019#L3505) pass locally:

```diff
diff --git a/tests/e2e/specs/stories-editor/story-templates.js b/tests/e2e/specs/stories-editor/story-templates.js
index 0715779d..cd06b711 100644
--- a/tests/e2e/specs/stories-editor/story-templates.js
+++ b/tests/e2e/specs/stories-editor/story-templates.js
@@ -73,11 +73,14 @@ describe( 'Story Templates', () => {
                        } );
 
                        it( 'should display non-template reusable blocks in the reusable blocks management screen', async () => {
+                               const titleSelector = '.page-title .row-title';
+
                                await visitAdminPage( 'edit.php', 'post_type=wp_block' );
+                               await page.waitForSelector( titleSelector );
 
                                // Check that it is untitled
                                const title = await page.$eval(
-                                       '.page-title .row-title',
+                                       titleSelector,
                                        ( element ) => element.innerText
                                );
                                expect( title ).toBe( 'Untitled Reusable Block' );
```

I'm not sure it'll make the build pass on Travis, though.

includes/sanitizers/class-amp-href-sanitizer.php
westonruter: As discussed in our chat today, the XPath for `*[@href]` will never return anything but elements containing `href` attributes, so this logic is unnecessary/redundant.

includes/sanitizers/class-amp-href-sanitizer.php
```diff
@@ -0,0 +1,75 @@
...
+					'Href sanitizer xpath query returned invalid node'
+				);
+			}
+
+			$url = esc_url( $node->getAttribute( 'href' ) );
```
westonruter: Should this be `esc_url_raw()`? We wouldn't want the HTML entities to trip things up.

includes/sanitizers/class-amp-href-sanitizer.php
```diff
@@ -0,0 +1,75 @@
...
+ * Crawls all href attributes to ensure their URLs are valid.
+ */
+final class AMP_Href_Sanitizer extends AMP_Base_Sanitizer {
+
+	/**
+	 * Xpath query selector to target all elements with a href attribute.
+	 *
+	 * @var string
+	 */
+	const XPATH_SELECTOR = '//*[@href]';
```
westonruter: I think this selector may not be complete, since there are more attributes with URL values than `href`.

Essentially, this needs to apply to every attribute that has a `value_url` property in the [`AMP_Allowed_Tags_Generated`](https://github.com/ampproject/amp-wp/blob/75674dfab5ac7f0fb59c35828d2501326240bbbc/includes/sanitizers/class-amp-allowed-tags-generated.php), for example:

https://github.com/ampproject/amp-wp/blob/75674dfab5ac7f0fb59c35828d2501326240bbbc/includes/sanitizers/class-amp-allowed-tags-generated.php#L17466-L17474

There are many examples of this, such as `xlink:href`, `src`, `srcset`, `data-tooltip-icon`, `data-url`, `data-share-media`, `data-share-url`, `config`, `data-href`, `data-src`, `data-tag`, and so on. 

Therefore, I think the best place to put this code is actually going to be in [`AMP_Tag_And_Attribute_Sanitizer`](https://github.com/ampproject/amp-wp/blob/75674dfab5ac7f0fb59c35828d2501326240bbbc/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php), in particular somewhere that `AMP_Rule_Spec::VALUE_URL` is being looked for:

https://github.com/ampproject/amp-wp/blob/75674dfab5ac7f0fb59c35828d2501326240bbbc/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L837-L887

Then there won't be a need to manually maintain this list, and the validation would be done at a point where it is already walking over the tree.

includes/sanitizers/class-amp-href-sanitizer.php
```diff
@@ -0,0 +1,75 @@
...
+
+			// Href URL is invalid, so we remove the entire attribute.
+			$node->removeAttribute( 'href' );
+
+			/*
+			 * "The target, download, rel, rev, hreflang, and type attributes must be omitted
+			 * if the href attribute is not present."
+			 * See: https://www.w3.org/TR/2016/REC-html51-20161101/textlevel-semantics.html#the-a-element
+			 */
+			foreach ( self::$attributes_tied_to_href as $attribute ) {
```
westonruter: Per above, when an `href`  attribute is removed, this code could be added to:

https://github.com/ampproject/amp-wp/blob/75674dfab5ac7f0fb59c35828d2501326240bbbc/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L559-L561

Nevertheless, this code may not be strictly necessary because it's not an AMP validation error, just a HTML semantics error.

swissspidy:
@kienstra Very weird, I am seeing these failures elsewhere too. Will open a PR with your suggested change!

includes/sanitizers/class-amp-href-sanitizer.php
schlessera: > not an AMP validation error, just a HTML semantics error.
Yes, but enforcing the AMP validation has introduced this HTML semantic error in the first place, so I don't think we should just leave it in.

includes/sanitizers/class-amp-href-sanitizer.php
schlessera: I'll try this in a separate PR again, to see what the result will be. We can later close whatever we don't want to use.

schlessera: I created a new PR #3151 with this approach. If that PR is preferable, we can close this one here.

includes/sanitizers/class-amp-href-sanitizer.php
schlessera: https://github.com/ampproject/amp-wp/pull/3117/commits/15fa5154e1ab5e9199d997e5a1819cc7342e299a

westonruter:
Closing in favor of #3151.

PR #3122, opened 2019-08-29
Update assertions for failing resizing tests

### LINKED ISSUE #3119: E2E Tests: Failing Resizing Test
Since yesterday the test for resizing a rotated text block is failing. Something in the environment has changed in a way that negatively affects the test, presumably the new Gutenberg release that also happened yesterday.

swissspidy:
Fixes #3119.


miina:
Maybe we should add 2 more tests then: for non-rotated block to ensure it's not changing its position when resizing from right and bottom? WDYT?

swissspidy:
Sure, that would be good.

PR #3149, opened 2019-08-30
Update block editor package

swissspidy:
Supersedes #3123.


assets/src/stories-editor/helpers/index.js
```diff
@@ -9,6 +9,7 @@ import memize from 'memize';
 /**
  * WordPress dependencies
  */
+import '@wordpress/core-data';
```
barklund: I don't understand this. What is the current problem and how does this fix it?

package.json
```diff
@@ -31,11 +31,12 @@
     "@babel/plugin-transform-react-jsx": "7.3.0",
     "@wordpress/babel-preset-default": "4.5.0",
     "@wordpress/blob": "2.5.0",
-    "@wordpress/block-editor": "3.0.0",
+    "@wordpress/block-editor": "3.1.0",
```
barklund: There are [no release notes for this version](https://github.com/WordPress/gutenberg/blob/master/packages/block-editor/CHANGELOG.md), so hard to tell what the consequences are. But if all tests pass, I guess we're good.

swissspidy: The tests were failing in #3123, but are passing here due to the change in the helpers file.

assets/src/stories-editor/helpers/index.js
swissspidy: The JS unit tests were failing without this change because the `core` data store was not initialized. See #3123 for the failed build info.

It is used referenced here:

https://github.com/ampproject/amp-wp/blob/e33ce8787ea08851843a2bfdc2abb600dfe43023/assets/src/stories-editor/helpers/index.js#L81

Because `dispatch( 'core' )` was `undefined`, this causes errors.

Now there were two options:

1. Move `const { saveMedia } = dispatch( 'core' );` to within `uploadVideoFrame`, the only place where it is used.
2. Import `@wordpress/core-data` so the data store is initialized in the JS unit tests

Where option 2 felt more robust to me.

Not sure why exactly this suddenly happened though. Perhaps it was the change at https://github.com/WordPress/gutenberg/pull/16932.

miina:
@googlebot I consent.

miina:
Merging `develop` once more for the tests to run just in case.

PR #3150, opened 2019-08-30
Update WP ESLint plugin

swissspidy:


assets/src/common/components/featured-image-cropper.js
```diff
@@ -5,14 +5,13 @@ const { wp } = window;
...
  * @constructs FeaturedImageCropper
- * @inheritDoc
  */
 const FeaturedImageCropper = wp.media.controller.Cropper.extend( {
 	/**
 	 * Creates an object with the image attachment and crop properties.
 	 *
 	 * @param {wp.media.model.Attachment} attachment The attachment to crop.
-	 * @return {jQuery.promise} A jQuery promise that represents the crop image request.
+	 * @return {Function} A jQuery promise that represents the crop image request.
```
barklund: Shouldn't this be a `Promise`? Or is a jquery Promise too different from a _real_ one to warrant that type?

assets/src/stories-editor/helpers/index.js
```diff
@@ -526,14 +526,14 @@ export const filterBlockAttributes = ( blockAttributes, blockType, innerHTML ) =
...
  *
- * @param {WPElement} element                  Block element.
- * @param {Object}    blockType                Block type object.
- * @param {Object}    attributes               Block attributes.
- * @param {number}    attributes.positionTop   Top offset in pixel.
- * @param {number}    attributes.positionLeft  Left offset in pixel.
- * @param {number}    attributes.rotationAngle Rotation angle in degrees.
- * @param {number}    attributes.width         Block width in pixels.
- * @param {number}    attributes.height        Block height in pixels.
+ * @param {Element} element                  Block element.
```
barklund: So this isn't a React element (with that weird `{import('react').ReactElement}` syntax), but a... DOM element?

swissspidy: Indeed looks like it should be annotated as a React element:

https://github.com/WordPress/gutenberg/blob/403fa51163f2122017625ca94f852fb4c863d145/packages/blocks/src/api/serializer.js#L104-L111

As for the syntax, I was worried that a normal import would result in unused variable warnings. But it looks like that isn't the case, so I'll update that.

assets/src/common/components/featured-image-cropper.js
swissspidy: It's not a real Promise, but an object with a similar API. Kinda like a poor man's implementation of a Promise.

According to https://api.jquery.com/Types/#Promise it doesn't have a `catch` method for example.

But since we don't really rely on the return value in our code, I think we can just use `Promise` with some updated comment to clarify it's a jQuery Promise

assets/src/common/components/pre-publish-panel.js
```diff
@@ -19,9 +19,10 @@ import { validateFeaturedImage } from '../helpers';
 /**
  * Conditionally adds a notice to the pre-publish panel for the featured image.
  *
- * @param {Object}  featuredMedia Media object.
- * @param {Array}   dimensions    Required image dimensions.
- * @param {boolean} required      Whether selecting a featured image is required.
+ * @param {Object} props Component props.
+ * @param {Object} props.featuredMedia Media object.
+ * @param {Array} props.dimensions Required image dimensions.
```
barklund: Indentation alignment for the parameter name is lost here. It's not important to me in any way, but I don't know if it is according to some coding standard / style guide.

barklund: Similarly in a few other places:

https://github.com/ampproject/amp-wp/pull/3150/files#diff-772941062b46bc04ec5e6d20e2dcd9afR15-R19
https://github.com/ampproject/amp-wp/pull/3150/files#diff-ed917cb11654e9f341941fb5ed4e2361R16-R18

swissspidy: Yeah we are not really consistent with that 😞 
The thing is, there is no ESLint rule that checks for alignment here and could enforce either style. So if you see some alignment, it has probably been added manually in the past.

I think we should move away from this vertical alignment anyway, because - other than in the PHP tooling - it cannot be enforced. Also, type definitions can be quite long for complex objects, making it hard to read with this alignment.

PR #3151, opened 2019-08-30
Remove value_url attributes with an invalid URL

### LINKED ISSUE #2671: Invalid URL like `foo@` passed through without sanitization causing AMP validation error

schlessera:
Adds logic to the `value_url` checks to invalidate URLs that cannot be properly parsed by `wp_parse_url()`, like `http://foo@`.

Also adds logic to remove left-over attributes after the removal of an invalid attribute. This is used to check specifically for the `href` attribute being removed, and apply the following HTML5 spec detail:

> The target, download, rel, rev, hreflang, and type attributes must be omitted if the href attribute is not present.

## Before

The invalid link successfully passes sanitization and causes validation errors on the frontend.

Image 2019-08-30 at 6 02 41 PM

Image 2019-08-30 at 4 49 44 PM

## After

The invalid link is caught. As a result, the `href` attribute is removed, as well as all attributes that directly relate to it.

Other attributes not directly related to the `href` attribute stay intact, in order to preserve styling.

Image 2019-08-30 at 4 49 12 PM

Fixes #2671
Previous attempts at fixing this: #3103 & #3117

### DIFF
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -433,6 +433,7 @@ abstract class AMP_Base_Sanitizer {
 		$should_remove = $this->should_sanitize_validation_error( $validation_error, compact( 'node' ) );
 		if ( $should_remove ) {
 			$element->removeAttributeNode( $node );
+			$this->clean_up_after_attribute_removal( $element, $node, $validation_error );
 		}
 		return $should_remove;
 	}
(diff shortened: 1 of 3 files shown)


swissspidy:
Sorry for the noise with these failing e2e tests. Already looking into it.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -433,6 +433,7 @@ public function remove_invalid_attribute( $element, $attribute, $validation_erro
 		$should_remove = $this->should_sanitize_validation_error( $validation_error, compact( 'node' ) );
 		if ( $should_remove ) {
 			$element->removeAttributeNode( $node );
+			$this->clean_up_after_attribute_removal( $element, $node, $validation_error );
```
westonruter: Placing this here ended up causing an exception in some cases. See https://github.com/ampproject/amp-wp/pull/3682

PR #3155, opened 2019-08-30
Use autocomplete for font selection.

### LINKED ISSUE #3027: Font Picker: Use searchable dropdown
See #2877 for details.

When we have all these new fonts, we need a better font picker that easily allows scrolling through the hundreds of fonts.

[`accessible-autocomplete`](https://github.com/alphagov/accessible-autocomplete) seems like a good candidate for me, but perhaps there is also a Gutenberg-originating component that we can use.

spacedmonkey:
Fixes #3027


miina:
Notes from functional testing:
- The fonts don't seem to be actually applied to the block (not in editor nor in the FE)
- After deselecting the blocks which had font assigned (e.g. by clicking somewhere else in the editor) and then choosing the block again which had the font assigned, the following displayed in the Font selector:
Screenshot 2019-09-02 at 12 43 28

- After refreshing, the font selector is empty, not displaying the chosen value
- Would be good to have a placeholder such as "Type to search font" or something similar.

assets/src/stories-editor/components/font-family-picker/index.js
```diff
@@ -24,44 +30,60 @@ function FontFamilyPicker( {
...
+			[];
+		syncResults( searchResults );
 	};
 
-	const options = fonts.map( ( font ) => ( {
-		value: font.name,
-		label: font.name,
-	} ) );
+	const suggestionTemplate = ( result ) => {
+		maybeEnqueueFontStyle( result.name );
```
swissspidy: This loads the whole font stylesheet, whereas for performance reasons it might be beneficial to only load the glyphs actually needed to display the font name.

See https://github.com/ampproject/amp-wp/issues/3027#issuecomment-526531210 for an example.

Then, when the font has been selected, the whole font stylesheet could be loaded.

assets/src/stories-editor/components/font-family-picker/index.js
```diff
@@ -6,13 +6,19 @@ import PropTypes from 'prop-types';
 /**
  * WordPress dependencies
  */
-import { __, sprintf } from '@wordpress/i18n';
-
+import { __, _n, sprintf } from '@wordpress/i18n';
+import { BaseControl } from '@wordpress/components';
+/**
+ * Internal dependencies
```
swissspidy: Thee grouping of the dependencies in this file got a bit mixed up / duplicated

assets/src/stories-editor/components/font-family-picker/index.js
spacedmonkey: I did some profiling here, the font is only 24kb, which is the size of a small image. As this is only loaded once the font is being displayed, it also has the benefit the preload the font before when displayed in the content of the amp page. Even 10 fonts loaded is only 240kb, which is the size of a small image.

spacedmonkey:
> Notes from functional testing:
> 
> * The fonts don't seem to be actually applied to the block (not in editor nor in the FE)
> * After deselecting the blocks which had font assigned (e.g. by clicking somewhere else in the editor) and then choosing the block again which had the font assigned, the following displayed in the Font selector:
> 
> Screenshot 2019-09-02 at 12 43 28
> 
> * After refreshing, the font selector is empty, not displaying the chosen value
> * Would be good to have a placeholder such as "Type to search font" or something similar.

This tweak this morning broke everything. https://github.com/ampproject/amp-wp/pull/3155/commits/aab610eaed2759c81d7425d8036a9016d6eb0d44 and are fixed in https://github.com/ampproject/amp-wp/pull/3155/commits/6fbec707af655a8400069ad2725a187c6cb63877

assets/src/stories-editor/components/font-family-picker/index.js
spacedmonkey: Resolved in https://github.com/ampproject/amp-wp/pull/3155/commits/f958f54816f234df0c749998e06e3931682abf31

miina:
Probably we could already add some e2e tests for the existing fonts, and just add some more once the Google fonts PR gets merged.

swissspidy:
The existing tests for `FontFamilyPicker` in `assets/src/stories-editor/components/font-family-picker/test/index.js` don't really apply anymore with this change. Since the component is now merely a wrapper around `Autocomplete`, we can probably remove those tests and concentrate on E2E tests instead.

I'd say the most important part to test there is the accurate preview / stylesheet enqueuing.

assets/src/stories-editor/components/font-family-picker/index.js
swissspidy: Hmm that sounds promising. I'll check it out myself while reviewing.

assets/src/stories-editor/components/font-family-picker/index.js
```diff
@@ -65,14 +70,16 @@ function FontFamilyPicker( {
...
 				preserveNullOptions={ true }
 				placeholder={ __( 'None', 'amp' ) }
-				showNoOptionsFound={ false }
 				displayMenu="overlay"
+				showNoOptionsFound={ false }
 				tStatusQueryTooShort={ ( minQueryLength ) =>
 					// translators: %d: the number characters required to initiate an author search.
 					sprintf( __( 'Type in %d or more characters for results', 'amp' ), minQueryLength )
 				}
 				// translators: 1: the index of thre selected result. 2: The total number of results.
```
barklund: There's a typo in this line and I think you need to move it one line down now that you multiline'd the next attribute.

spacedmonkey: What is the typo?

spacedmonkey:
E2E tested added. I don't know how an e2e is going to test that preview works, as this is visual.

assets/src/stories-editor/components/font-family-picker/index.js
swissspidy:
```diff
-// translators: 1: the index of thre selected result. 2: The total number of results.
+// translators: 1: the index of the selected result. 2: The total number of results.

**Comment by swissspidy** [comment]
> I don't know how an e2e is going to test that preview works, as this is visual.

Testing that the stylesheets are enqueued would be good

File: assets/src/stories-editor/components/font-family-picker/index.js (hunk 3 above)
Comment (spacedmonkey) [reply assets/src/stories-editor/components/font-family-picker/index.js]: Fixed https://github.com/ampproject/amp-wp/pull/3155/commits/2b91bb5ba99035055bcbb2b79dba598b46b38858

**Comment by spacedmonkey** [comment]
> Testing that the stylesheets are enqueued would be good

There is now a test to make sure that google fonts are loaded. https://github.com/ampproject/amp-wp/pull/3155/commits/a27364f34ba93a184207d3a70dc6697100cb341a

File: assets/src/stories-editor/components/font-family-picker/index.js [hunk 4]
```diff
@@ -23,50 +26,76 @@ function FontFamilyPicker( {
...
+				placeholder={ __( 'None', 'amp' ) }
+				displayMenu="overlay"
+				tNoResults={ () =>
+					__( 'No font found', 'amp' )
+				}
+				tStatusQueryTooShort={ ( minQueryLength ) =>
+					// translators: %d: the number characters required to initiate an author search.
+					sprintf( __( 'Type in %d or more characters for results', 'amp' ), minQueryLength )
+				}
+				// translators: 1: the index of the selected result. 2: The total number of results.
```
barklund: This has to be moved one line down for the translation extractor to pick this hint up.

The JS i18n extractor is different from the PHP i18n extractor:

- in PHP the translator hint just has to be the most recent comment before the line in question
- in JS the translator hint has to be a comment in the line **exactly above** the line in question

assets/src/stories-editor/components/font-family-picker/index.js
```diff
@@ -23,50 +26,76 @@ function FontFamilyPicker( {
...
+				defaultValue={ value }
+				dropdownArrow={ () => '' }
+				preserveNullOptions={ true }
+				placeholder={ __( 'None', 'amp' ) }
+				displayMenu="overlay"
+				tNoResults={ () =>
+					__( 'No font found', 'amp' )
+				}
+				tStatusQueryTooShort={ ( minQueryLength ) =>
+					// translators: %d: the number characters required to initiate an author search.
```
barklund: This translator hint is a bit misleading, as it's a font search, not an author search. It's not super important, just weird.

spacedmonkey:
I noticed a bug now that all the google fonts are loaded. See screen below.
Screenshot from 2019-09-05 10-59-16

When searching for a popular word in a list of fonts you get a lot of results. Sans in the example gets you many results. This gives you a list of results with a scroll bar. Selecting a font using the keyboard or using the scroll wheel + click works. However, using the scroll bars / arrow on the list, dismisses the element. 

A list that long isn't specially usable or nice to use. What should the bahavior be when the list of results is extremely long.

assets/src/stories-editor/components/font-family-picker/index.js
```diff
@@ -26,6 +27,7 @@ function FontFamilyPicker( {
 	fonts = [],
 	onChange = () => {},
 	value = '',
+	instanceId,
```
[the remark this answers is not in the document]
swissspidy: This way the component always has a unique  id/label pair, without the parent having to pass an ID to it.

swissspidy:
> When searching for a popular word in a list of fonts you get a lot of results. Sans in the example gets you many results.

This loads dozens of stylesheets at once too. 😕

swissspidy:
Took the liberty to make some small changes, esp. around naming.

I currently struggle with clearing the input field. After choosing a font, how can I clear the current selection so that the text block uses no custom font anymore?

spacedmonkey:
@swissspidy After some testing and work on accessibility yesterday, I fixed the scrolling issue with the override seen in code. I also change the input from overlay and inline. This does two things, it makes the font selector much easier to use and scroll inside. It also brings it inline with the tag selection already found on the page. See 

Screenshot from 2019-09-06 09-07-59

I believe some sort of 'Clear font' button should be added. This would mean you could do the following. Add a font to a piece of text. Search for another font, not find anything, clear font. Readd original font. 

Not for if we need anything else that would dismiss the existing search term. This would mean override much of the existing compound, basically making it custom and loosing the benefit of using an open compound in the first place.

swissspidy:
Thanks for calling me out. Definitely didn't intend to break all the work you've been doing, I was more fiddling with the different options this component provides. Will address in a bit.

spacedmonkey:
I have been playing around clear font button. Here is what I have locally. 

Screenshot from 2019-09-06 10-29-32

But I am wonder if this is a out of scope of this ticket and should be another PR.

swissspidy:
> But I am wonder if this is out of scope of this ticket and should be another PR.

The old font picker allows you to clear the font:

Screenshot 2019-09-06 at 12 28 53

Removing that now would be a regression IMHO.

spacedmonkey:
When trying to add a 'Clear font' button, I noticed that there is no way to change the value of the autocomplete programmatically. The default value can be set but this value changing doesn't seem to redraw the input. Set is handle within the compound and can not be accessed externally. Again, I might be missing something here...

spacedmonkey:
Created an issue with the core library regarding the clear issue https://github.com/alphagov/accessible-autocomplete/issues/354

assets/src/stories-editor/components/font-family-picker/status.js
```diff
@@ -0,0 +1,90 @@
...
+				marginRight: '-1px',
+				overflow: 'hidden',
+				padding: '0',
+				position: 'absolute',
+				whiteSpace: 'nowrap',
+				width: '1px',
+			} }
+		>
+			{ content }
+			<span>{ bump ? ',' : ',,' }</span>
```
barklund: What does this `bump` do exactly?

assets/src/stories-editor/components/font-family-picker/status.js
```diff
@@ -0,0 +1,90 @@
...
+
+export default class Status extends Component {
+	constructor( props ) {
+		super( props );
+		this.state = {
+			bump: false,
+		};
+	}
+
+	componentWillReceiveProps( { queryLength } ) {
```
barklund: This is [a deprecated method](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops). We should most definitely try to avoid it.

Why is it necessary?

swissspidy: Looks like it was just copied over from upstream: https://github.com/alphagov/accessible-autocomplete/blob/baee8582ae28ec499289eded99053e0c27ebf33d/src/status.js#L22-L27

assets/src/stories-editor/components/font-family-picker/status.js
swissspidy: Copied from upstream: https://github.com/alphagov/accessible-autocomplete/blob/baee8582ae28ec499289eded99053e0c27ebf33d/src/status.js#L77

assets/src/stories-editor/components/font-family-picker/status.js
barklund: Yeah, I figured out the reasoning (a11y), however it's a bit flawed as the aria live component only works if on screen at first HTML render - which it obviously won't be. So I don't think it works anyway.

It did rewrite the component in 37a3355 to avoid these silly work-arounds.

assets/src/stories-editor/components/font-family-picker/status.js
```diff
@@ -4,78 +4,54 @@
  */
 import PropTypes from 'prop-types';
 
-/**
- * WordPress dependencies
- */
-import { Component } from '@wordpress/element';
+const Status = ( {
```
spacedmonkey: :+1:

swissspidy:
@barklund @spacedmonkey Regarding thee `Status` component, I feel like we could just ditch it in a future PR and use `withSpokenMessages` / `@wordpress/a11y` instead for the `aria-live` region.

PR #3159, opened 2019-08-30
Fix notice thrown for checking nonexistent is_block_editor property

### LINKED ISSUE #3154: Undefined property: WP_Screen::$is_block_editor
### Description

I'm getting this notice after installing version `1.2.2` on a Wordpress `4.9.10`.  The reason seems to be because `is_block_editor` was introduced [only on version 5](https://developer.wordpress.org/reference/classes/wp_screen/is_block_editor/#changelog).

```
( ! ) Notice: Undefined property: WP_Screen::$is_block_editor in /content/plugins/amp/includes/admin/class-amp-post-meta-box.php on line 136
```

### To reproduce:

- Make sure you have `define('WP_DEBUG', true);` on your config file.
- Use Wordpress `4.9.10` and AMP plugin `1.2.2 `.

### Recommended solution:

Check if the property exists first

```php
(isset($screen->is_block_editor) && ! $screen->is_block_editor)
```

schlessera:
We need to check first whether the `$screen->is_block_editor` property actually exists.

Otherwise, we get a notice on WP < 5.0.
Image 2019-08-30 at 9 20 32 PM

Fixes #3154

### DIFF
--- a/includes/admin/class-amp-post-meta-box.php
+++ b/includes/admin/class-amp-post-meta-box.php
@@ -133,7 +133,7 @@ class AMP_Post_Meta_Box {
 		$validate = (
 			isset( $screen->base ) &&
 			'post' === $screen->base &&
-			! $screen->is_block_editor &&
+			( ! isset( $screen->is_block_editor ) || ! $screen->is_block_editor ) &&
 			is_post_type_viewable( $post->post_type ) &&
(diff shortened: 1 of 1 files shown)


westonruter:
The `is_block_editor` check was added in #2980 via 2fcc5b9891d8f91f24a0cd02d71f6280706820e5.

@swissspidy Would you sanity check this?

PR #3165, opened 2019-09-02
Use Jest silent reporter to reduce noise in output

swissspidy:


swissspidy:
Test output for E2E tests: https://travis-ci.org/ampproject/amp-wp/jobs/579706272#L916-L927

Test output for JS unit tests: https://travis-ci.org/ampproject/amp-wp/jobs/579706271#L539-L546

So clean 😻

spacedmonkey:
Does this get rid of all these warning in the log would be amazing. Is there a worry about not seeing warning here for debugging.

swissspidy:
It does, yes. And when there is a failing test, you'll still see the full error output, kinda like with PHPUnit. Here are some screenshots: https://github.com/rickhanlonii/jest-silent-reporter#screenshots

For debugging locally you could run `npm run test:js -- --reporters=
"default"` to use the default reporter again with all log messages.

Issue #3168, opened 2019-09-03
[AMP Stories] Add filter for allowing more video extensions


pradeep910:
AMP story background media select option only allows mp4 videos for selection. 

We'd need to add a filter to allow more extensions like `.mov`, `.mkv`, `.flv`, `.avi` etc. to make it ready for Video Transcoding.

thrijith:
I am currently editing these files and building the zip for temporarily adding other extensions.

https://github.com/ampproject/amp-wp/blob/f5cbc6f6a40d79edf4f245c0c6b031d273263c80/assets/src/stories-editor/constants.js#L110
https://github.com/ampproject/amp-wp/blob/8ff2ce74650322a0d2b1f6de60f92db9ad123ced/assets/src/common/helpers/index.js#L252-L254

I am working on adding a filter for the same, will update here once I am ready with the changes.

PR #3171, opened 2019-09-03
Add filter to allow more video types in AMP story background selection

thrijith:
Fixes https://github.com/ampproject/amp-wp/issues/3168


thrijith:
@googlebot I signed it!

assets/src/common/helpers/index.js
```diff
@@ -18,6 +18,8 @@ import {
 	VIDEO_ALLOWED_MEGABYTES_PER_SECOND,
 } from '../constants';
 
+const { ampAllowedVideoTypes } = window;
```
swissspidy: Files in `common` can be loaded also in non-story editor context, so this is not guaranteed to exist.

assets/src/stories-editor/constants.js
```diff
@@ -107,8 +109,8 @@ export const ALLOWED_BLOCKS = [
 export const IMAGE_BACKGROUND_TYPE = 'image';
 export const VIDEO_BACKGROUND_TYPE = 'video';
 
-export const ALLOWED_VIDEO_TYPES = [ 'video/mp4' ];
-export const ALLOWED_BACKGROUND_MEDIA_TYPES = [ IMAGE_BACKGROUND_TYPE, ...ALLOWED_VIDEO_TYPES ];
+export const ALLOWED_VIDEO_TYPES = ampAllowedVideoTypes;
```
swissspidy: This should rather use the data store vs. just assigning it to a constant.

includes/class-amp-story-post-type.php
```diff
@@ -900,6 +900,27 @@ public static function enqueue_block_editor_scripts() {
...
+
+		// If `$allowed_video_mime_types` doesn't have valid data or is empty add default supported type.
+		if ( ! is_array( $allowed_video_mime_types ) || empty( $allowed_video_mime_types ) ) {
+			$allowed_video_mime_types = [ 'video/mp4' ];
+		}
+
+		// Pass final list of allowed video types to script.
+		wp_localize_script(
+			self::AMP_STORIES_SCRIPT_HANDLE,
+			'ampAllowedVideoTypes',
```
swissspidy: Let's rename this to `ampStoriesEditorSettings` and pass an array of settings instead of just the video mime types. Otherwise this doesn't scale.

swissspidy:
@thrijith I took the liberty to make some changes directly so you can better see the direction I think this should take. Let me know what you think.

I also saw that we still have many areas where `video/mp4` is hardcoded. For example in our custom video block edit component.

thrijith:
> I took the liberty to make some changes directly so you can better see the direction I think this should take. Let me know what you think.

Thanks @swissspidy 🙂, I was not sure, how to go about adding changes needed for store.

> I also saw that we still have many areas where `video/mp4` is hardcoded. For example in our custom video block edit component.

Are you referring the accept part? https://github.com/ampproject/amp-wp/blob/217d5929b15bd48e2eaac431d94b03c198ca5466/assets/src/stories-editor/components/custom-video-block-edit.js#L234-L237

It doesn't allow selection for other extensions when using `Upload` button in the Video block.

thrijith:
```diff --git a/assets/src/stories-editor/components/custom-video-block-edit.js b/assets/src/stories-editor/components/custom-video-block-edit.js
index 412fabac..d45a5580 100644
--- a/assets/src/stories-editor/components/custom-video-block-edit.js
+++ b/assets/src/stories-editor/components/custom-video-block-edit.js
@@ -235,7 +235,7 @@ class CustomVideoBlockEdit extends Component {
 					className={ className }
 					onSelect={ onSelectVideo }
 					onSelectURL={ this.onSelectURL }
-					accept="video/mp4"
+					accept={ allowedVideoMimeTypes.join( ',' ) }
 					allowedTypes={ allowedVideoMimeTypes }
 					value={ this.props.attributes }
 					notices={ noticeUI }
diff --git a/assets/src/stories-editor/components/with-enforced-video-upload-type.js b/assets/src/stories-editor/components/with-enforced-video-upload-type.js
index 39c49239..7520580c 100644
--- a/assets/src/stories-editor/components/with-enforced-video-upload-type.js
+++ b/assets/src/stories-editor/components/with-enforced-video-upload-type.js
@@ -14,11 +14,11 @@ import PropTypes from 'prop-types';
  */
 export default ( InitialMediaPlaceholder ) => {
 	const withEnforcedVideoUploadType = ( props ) => {
-		const { accept, className } = props;
+		const { accept, className, allowedTypes } = props;
 		let newProps = { ...props };
 
 		if ( 'wp-block-video' === className && 'video/*' === accept ) {
-			newProps = { ...props, accept: 'video/mp4' };
+			newProps = { ...props, accept: allowedTypes.join( ',' ) };
 		}
 
 		return <InitialMediaPlaceholder { ...newProps } />;
@@ -27,6 +27,7 @@ export default ( InitialMediaPlaceholder ) => {
 	withEnforcedVideoUploadType.propTypes = {
 		accept: PropTypes.string,
 		className: PropTypes.string,
+		allowedTypes: PropTypes.arrayOf( PropTypes.string ).isRequired,
 	};
 
 	return withEnforcedVideoUploadType;
```

Since allowed video type is available in props, we could do something like this to remove usage of hardcoded mime type from custom video block element.

swissspidy:
`with-enforced-video-upload-type.js` is probably going away soon as we don't really need that anymore now that we have `custom-video-block-edit.js` - see #3189.

Otherwise your suggested change looks good 👍

thrijith:
I have updated the `accept` attribute value for custom video edit block, I think the PR can be merged, as it allows selection of other mime type in `Background Media` as well as video block. Please let me know if it needs any other change.

includes/class-amp-story-post-type.php
```diff
@@ -936,6 +936,31 @@ public static function enqueue_block_editor_scripts() {
...
+		 * Filter list of allowed video mime types.
+		 *
+		 * This can be used to add additionally supported formats, for example by plugins
+		 * that do video transcoding.
+		 *
+		 * @since 1.2
+		 *
+		 * @param array Supported video mime types.
+		 */
+		$allowed_video_mime_types = apply_filters( 'amp_supported_video_types', [ 'video/mp4' ] );
```
swissspidy:
```suggestion
		$allowed_video_mime_types = apply_filters( 'amp_story_supported_video_types', [ 'video/mp4' ] );
```

Let's make it clear that this is for stories only.

includes/class-amp-story-post-type.php
```diff
@@ -936,6 +936,31 @@ public static function enqueue_block_editor_scripts() {
...
+		 * Filter list of allowed video mime types.
+		 *
+		 * This can be used to add additionally supported formats, for example by plugins
+		 * that do video transcoding.
+		 *
+		 * @since 1.3
+		 *
+		 * @param array Allowed video mime types.
+		 */
+		$allowed_video_mime_types = apply_filters( 'amp_story_allowed_video_types', [ 'video/mp4' ] );
```
spacedmonkey: There is another filter in core that is very similar `wp_video_extensions` could someone filtering that value effect anything?

spacedmonkey: Should the values in this array be compared to values in the array returned by `wp_get_mime_types()`

thrijith: > There is another filter in core that is very similar `wp_video_extensions` could someone filtering that value effect anything?

Story is using static mime type `video/mp4` right now, `wp_video_extensions` only adds extensions so it should not cause any affect here.

> Should the values in this array be compared to values in the array returned by wp_get_mime_types()

Compared to see if the values received by `amp_story_allowed_video_types` are valid or not?

spacedmonkey: I am worrying if this array should be compared against values found in `wp_get_mime_types()`. What is the point of adding a mime type that core itself doesn't support? So maybe using `array_intersect` of the two arrays?

thrijith: Since `wp_get_mime_types` uses `mime_types` filter internally which can be modified, relying on that won't be full proof either, as the function contains any other file mime types, comparing received data isn't good as it can contain any type passed by services using that filter.

Only full proof way would be checking against static valid video formats list, which would have to be updated in some way.

swissspidy: I don't think we need a bulletproof sanity check here when someone's `doing_it_wrong™`. `array_intersect` with `wp_get_mime_types` is good enough IMO.

thrijith: @spacedmonkey I've made the changes as suggested, please check and let me know if it's ok.

I've kept the previous validation to add `video/mp4` by default, if received data is empty or not an array, let me know if that needs change as well. Thanks.

PR #3177, opened 2019-09-04
Remove featured image cropping from the non-AMP Story editor

### LINKED ISSUE #2684: How to redefine featured image sizes?

kienstra:
# Before
In the block editor for posts (though also for AMP Stories), on selecting many [image sizes](https://ipsumimage.appspot.com/1400x1000) from the Media Library, there was a suggested crop: 
feat-crop

# After
There is no cropping for the featured image on non-AMP story posts. This PR shouldn't change the behavior for AMP Stories. There can [still be notices](https://github.com/ampproject/amp-wp/issues/2684#issuecomment-526791367) about small featured image sizes for posts, if applicable.

Closes #2684


assets/src/block-editor/components/with-media-library-notice.js
```diff
@@ -0,0 +1,109 @@
+/**
```
kienstra: This is copied almost verbatim from [with-cropped-featured-image.js](https://github.com/ampproject/amp-wp/blob/bcda6d639920c8fe5c5e5fa5af83ef15d815a72f/assets/src/common/components/with-cropped-featured-image.js#L1).

assets/src/common/helpers/index.js
```diff
@@ -392,3 +392,35 @@ export const getContentLengthFromUrl = async ( url ) => {
...
+ * Sets the featured image, on selecting it in the Media Library.
+ *
+ * @param {string} url             Image URL.
+ * @param {number} id    Attachment ID.
+ * @param {number} width           Image width.
+ * @param {number} height          Image height.
+ * @param {Function} onSelect      A function in the MediaUpload component called on selecting the image.
+ * @param {Function} dispatchImage A function to dispatch the change in image to the store.
+ */
+export const setImageFromURL = ( { url, id, width, height, onSelect, dispatchImage } ) => {
```
kienstra: This is copied [almost verbatim](https://github.com/ampproject/amp-wp/blob/bcda6d639920c8fe5c5e5fa5af83ef15d815a72f/assets/src/common/components/with-cropped-featured-image.js#L259). It's in a separate function so 2 classes can use it.

kienstra:
**Request For Review**

If you have a chance, could you please review this PR?

Thanks!

assets/src/common/helpers/index.js
```diff
@@ -392,3 +392,35 @@ export const getContentLengthFromUrl = async ( url ) => {
 	} );
 	return Number( response.headers.get( 'content-length' ) );
 };
+
+/**
+ * Sets the featured image, on selecting it in the Media Library.
+ *
+ * @param {string} url             Image URL.
+ * @param {number} id    Attachment ID.
```
schlessera: Alignment is off.
```suggestion
 * @param {number} id              Attachment ID.
```

assets/src/block-editor/components/with-media-library-notice.js
```diff
@@ -0,0 +1,104 @@
...
+import { setImageFromURL } from '../../common/helpers';
+import { dispatch } from '@wordpress/data';
+
+const { wp } = window;
+
+/**
+ * Gets a wrapped version of MediaUpload to display a notice for small images.
+ *
+ * Only applies to the MediaUpload in the Featured Image component, PostFeaturedImage.
+ * Mostly copied from customize-controls.js.
```
schlessera: For copied files, it makes sense to record the version that was copied, for easier checking if upstream changes need to be merged in the future.

schlessera:
I did a review of the code, but I have to admit I can't say much about whether the approach is the best one I can imagine without digging much deeper into the code.

I looks like you're using a helper function to deduplicate code. Generally, I consider anything that lands in a "Helper" object or collection to be a design smell, but at first glance it looks like the actual object that would receive this functionality in a better design is not extensible enough here as it is provided by Core. Am I interpreting this correctly?

kienstra:
Thanks, @schlessera! Just getting back from time off now, but I'll address your points soon.

assets/src/block-editor/components/with-media-library-notice.js
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @param {Function} InitialMediaUpload          The MediaUpload component, passed from the filter.
+ * @param {Object}   minImageDimensions          Minimum required image dimensions.
+ * @param {Object}   alternateMinImageDimensions Alternate required image dimensions, like portrait dimensions (optional).
```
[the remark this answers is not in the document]
kienstra: Good catch, https://github.com/ampproject/amp-wp/pull/3177/commits/46401a9433a8c7899144926dd5ee9c510dfeebec fixes this.

assets/src/block-editor/components/with-media-library-notice.js
kienstra: Great idea, https://github.com/ampproject/amp-wp/pull/3177/commits/c971ce903b7ce08cd810a763c79751338f0185e9 adds a `@link` tag.

kienstra:
Hi @schlessera,
Sorry for the delay. Thanks for reviewing this, and for your great feedback.

>I looks like you're using a helper function to deduplicate code. Generally, I consider anything that lands in a "Helper" object or collection to be a design smell, but at first glance it looks like the actual object that would receive this functionality in a better design is not extensible enough here as it is provided by Core. Am I interpreting this correctly?

Yeah, I think you're right that it's not best practice 😄 

Though I couldn't think of a way around it.

As you probably saw, these files have classes that extend the `MediaUpload` component:
* [with-cropped-featured-image.js](https://github.com/ampproject/amp-wp/pull/3177/files#diff-1fa51d59be09326b2b38671d04a1a9e6R50) (AMP Stories)
* [with-media-library-notice.js](https://github.com/ampproject/amp-wp/pull/3177/files#diff-538d50f210178fdcaa9c81076259f96aR33) (Block Editor)

And those classes have a common method, [setImageFromURL](https://github.com/ampproject/amp-wp/pull/3177/files#diff-2602739dd54c92264c54236ad23b5599R412). Like you mentioned, I extracted it into `helpers/index.js`.

Maybe there's a better way to make that method available to both of those extending classes, instead of in a [helpers file](https://github.com/ampproject/amp-wp/pull/3177/files#diff-2602739dd54c92264c54236ad23b5599R412). But I couldn't think of one so far.

assets/src/common/helpers/index.js
kienstra: Thanks, I committed your suggestion.

westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3602879/amp.zip) - v1.3-alpha-20190911T214802Z-10763143

westonruter:
@kienstra please resolve the conflict and then merge.

kienstra:
Hi @westonruter,
Sorry for the delay. I'll merge this when Travis passes.

PR #3187, opened 2019-09-05
Integrate admin bar with AMP dev mode

### LINKED ISSUE #1921: Ignore admin bar for AMP validation purposes

westonruter:
* Skip sanitizing admin bar (and removing it) in favor of ignoring (in dev mode).
* Introduce `amp_is_dev_mode()` function along with an `amp_dev_mode_enabled` filter; this returns `true` by default when the user is logged-in and the admin bar is showing.
* Introduce an `AMP_Dev_Mode_Sanitizer` filter which is registered among the sanitizers when `amp_is_dev_mode()` returns true. Takes as an argument an array of XPath expressions for elements to add `data-ampdevmode` to; this array can be filtered via the `amp_dev_mode_element_xpaths` hook. When the admin bar is shown, it will at least include the XPath expressions for excluding admin bar.
* Eliminate need for keeping forked version of `admin-bar.css` in plugin.
* Automatically add `data-ampdevmode` attributes to scripts and styles that depend on the admin bar, excluding them from being sanitized out of the document. This includes Dashicons, when the stylesheet is not being enqueued as part of the admin bar.
* Skip processing stylesheets and converting elements that have `data-ampdevmode` attributes.

Fixes #1921.

Reverts or partially-undoes #2346, #2972, #2500, #2405.

# Before

When themes have too much CSS for the admin bar to fit, the admin bar was removed from the page:

> image

When there was not too much CSS, the admin bar was included in post-processing, converting all elements in the admin bar to AMP (including the Gravatar to `amp-img`) and processing/tree-shaking the `admin-bar.css` and `dashicons.css`; the `admin-bar.js` was dequeued and any plugins that tried to add JS to put interactivity in the admin bar were blocked as validation errors:

> image

# After

The admin bar element tree, the admin bar stylesheets (`admin-bar.css`, its inline style, and `dashicons.css` if not directly enqueued by a theme/plugin) will all be marked with the `dev-ampdevmode` attribute, in addition to this the `data-ampdevmode` attribute being added to the root element. The AMP validator extension will show one single error related to dev mode, but the extension will soon be updated to change this from an error badge to some other indicator specific dev mode:

> image

The root HTML element, the admin bar element tree, the styles for the admin bar (`admin-bar.css`, inline style, `dashicons.css`), and now even the `admin-bar.js` all get the `data-ampdevmode` attribute added and thus are excluded from sanitization: 

> image

Now plugins can exempt JS administrative functionality in the admin bar from AMP validation. For example, with https://github.com/Automattic/jetpack/pull/13450 the Notifications module in Jetpack now works on AMP pages:

> image

The required scripts and styles all get the `data-ampdevmode` attribute:

> image

Even when there is excessive CSS on the page, the admin bar will still be included:

> image

# Todo

- [x] When admin bar is showing, add `data-ampdevmode` to the `html` element.
- [x] Ensure that `data-ampdevmode` is added to all admin bar assets, both scripts and styles,  and any related dependencies.
- [x] Exclude processing stylesheets that have `data-ampdevmode` attributes.
- [x] Remove `#wpadminbar` from the page prior to running the sanitizers, and then restore it at the end, but with `data-ampdevmode` attributes being added to every element in the tree.
- [x] Add `#wpadminbar` to `dynamic_element_selectors` when admin bar is shown. ~May require #1478.~
- [x] Add filter for controlling whether dev mode is enabled.
- [x] Add filter for the XPath queries for elements to decorate with `data-ampdevmode` attributes.
- [x] Make dev-mode for admin bar conditional based on user being authenticated, since the admin bar can technically be forced for unauthenticated users? `( is_user_logged_in() && is_admin_bar_showing() )` But then we'd still need the forked admin bar CSS. In reality, only a very very small number of sites would force the admin bar to be shown for unauthenticated users, so this is probably not worth worrying about. But perhaps we should still keep in place the stylesheet prioritization logic.
- [x] ~Add support for declaring certain script/style handles as being for dev mode?~ (Maybe later.)
- [x] ~Make use of `data-ampdevmode` to allow Customizer scripts and styles in the preview, as opposed to the sanitizer args `allow_dirty_styles` and `allow_dirty_scripts`.~ (To be investigated later; Customizer integration is less important than the admin bar, and the current approach of just allowing all scripts in the Customizer preview works fine.)
- [x] Skip converting elements that have `data-ampdevmode`?
- [x] Skip processing `style` attributes on elements that have `data-ampdevmode`?
- [x] Add tests.
- [x] Try [implementing](https://github.com/google/site-kit-wp/pull/505) support in Site Kit.
- [x] Try [implementing](https://github.com/Automattic/jetpack/pull/13450) support in Jetpack.
- [x] [Implement](https://github.com/pantheon-systems/pantheon-hud/pull/44) support in Pantheon HUD.

----

Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3597735/amp.zip) - v1.3-alpha-20190910T210727Z-e15bbe33


westonruter:
I'm going to add some more tests, but the scope of this I believe is complete.

westonruter:
I'm also going to open PRs for Jetpack and Site Kit to ensure the API allows for the plugins to successfully exempt admin bar integrations from AMP validation.

includes/class-amp-theme-support.php
```diff
@@ -2103,10 +2264,18 @@ public static function prepare_response( $response, $args = [] ) {
 			$dom->documentElement->setAttribute( 'amp', '' );
 		}
 
-		AMP_HTTP::send_server_timing( 'amp_dom_parse', -$dom_parse_start, 'AMP DOM Parse' );
+		// Apply dev mode mutations when in dev mode.
+		if ( $dev_mode_enabled ) {
+			list( $original_admin_bar_element, $placeholder_admin_bar_element ) = self::apply_dev_mode_mutations( $dom, $dev_mode_xpaths );
```
westonruter: It would be somewhat better if this code were placed into a new `AMP_Dev_Mode_Sanitizer`. The challenge here is that the sanitizer needs to process elements and then take some more actions later on (replacing the placeholder with the original admin bar). Nevertheless, since the conversion sanitizers (e.g. `AMP_Img_Sanitizer) are all now configured to skip over elements that have `data-ampdevmode` attributes, then in practice there isn't _really_ a need to temporarily remove the `#wpadminbar` element from the document while sanitization is occuring. We aren't removing other such elements outside of this tree (e.g. scripts and stylesheets). So perhaps better to just go with a new sanitizer which is forcibly set to run at the very beginning (in the same way that the style sanitizer and tag/attribute sanitizer run at the end).

includes/class-amp-theme-support.php
```diff
@@ -1822,6 +1877,106 @@ public static function filter_customize_partial_render( $partial ) {
...
+		 * Filters the XPath queries for elements that should be enabled for dev mode.
+		 *
+		 * By supplying XPath queries to this filter, the data-ampdevmode attribute will automatically be added to the
+		 * root HTML element as well as to any elements that match the expressions. The attribute is added to the
+		 * elements prior to running any of the sanitizers.
+		 *
+		 * @since 1.3
+		 * @param string[] XPath element queries. Context is the root element.
+		 */
+		$dev_mode_xpaths = apply_filters( 'amp_dev_mode_element_xpaths', [] );
```
schlessera: I would cast to an array here, as you can never really trust plugins/themes to properly use filters.
```suggestion
		$dev_mode_xpaths = (array) apply_filters( 'amp_dev_mode_element_xpaths', [] );
```
This avoids the AMP plugin throwing errors in lines 1936-1937, and just make the offending plugin/theme not work as expected.

includes/class-amp-story-post-type.php
```diff
@@ -233,6 +233,17 @@ function() {
 		add_action( 'amp_story_head', 'wp_site_icon', 99 );
 		add_action( 'amp_story_head', 'wp_oembed_add_discovery_links' );
 
+		// Disable admin bar from even trying to be output, since wp_head and wp_footer hooks are not on the template.
```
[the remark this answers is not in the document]
westonruter: That's a good question. I'm not sure if the styles used in AMP Stories would even be compatible with the styles used for the admin bar. Since AMP Stories are a full-screen experience, I suspect that the admin bar would clash. But it would be something useful to investigate at a later time, yes.

includes/class-amp-theme-support.php
westonruter: Done in d1ac1642707faf4ca28157bf974c7c8aaaf2458f

includes/amp-helper-functions.php
```diff
@@ -899,6 +935,33 @@ function amp_get_content_sanitizers( $post = null ) {
 	 */
 	$sanitizers = apply_filters( 'amp_content_sanitizers', $sanitizers, $post );
 
+	if ( amp_is_dev_mode() ) {
```
schlessera: I think it would be preferable to have this piece of logic be part of the `AMP_Dev_Mode_Sanitizer` class instead, and then make that sanitizer decide when it acts or not, instead of hardcoding it in here.

It should be the responsibility of the `AMP_Dev_Mode_Sanitizer` to know when to do something and what to do, not some helper function collection.

tests/php/test-amp-helper-functions.php
```diff
@@ -628,6 +629,33 @@ public function test_amp_get_content_embed_handlers() {
...
+
+		// Test authenticated user with admin bar showing.
+		add_filter( 'show_admin_bar', '__return_true' );
+		wp_set_current_user( self::factory()->user->create( [ 'role' => 'administrator' ] ) );
+		$this->assertTrue( is_admin_bar_showing() );
+		$this->assertTrue( is_user_logged_in() );
+		$this->assertTrue( amp_is_dev_mode() );
+
+		// Test unauthenticated user with admin bar forced.
+		add_filter( 'show_admin_bar', '__return_true' );
```
schlessera: Theoretical issue: The previous `'show_admin_bar'` filter was not removed. As it is the same one, this is a non-issue, but if it would be a different one, or something would be interjected in-between, it could change the test results.

tests/php/test-amp-helper-functions.php
```diff
@@ -703,10 +703,61 @@ static function( $classes ) {
...
+		$this->assertEquals( 'AMP_Style_Sanitizer', $ordered_sanitizers[ count( $ordered_sanitizers ) - 2 ] );
+		$this->assertEquals( 'AMP_Tag_And_Attribute_Sanitizer', $ordered_sanitizers[ count( $ordered_sanitizers ) - 1 ] );
+	}
+
+	/**
+	 * Test amp_get_content_sanitizers().
+	 *
+	 * @covers ::amp_get_content_sanitizers()
+	 */
+	public function test_amp_get_content_sanitizers_with_dev_mode() {
```
schlessera: As stated previously, I would suggest always having the `AMP_Dev_Mode_Sanitizer` around instead, but letting it decide itself what to act upon. It can include an early bail in case it doesn't make sense to run, so there's no point in having special logic to add it conditionally.

This would mean some of the tests below are not needed.

includes/amp-helper-functions.php
westonruter: I did think about that, but I think it's preferable here because this function is inherently connected directly with WordPress as it applies a filter. There are two reasons for why this should not be in the sanitizer:

1. It is important that the array of sanitizer args returned by `amp_get_content_sanitizers()` be deterministic in what the resulting sanitized document be. In other words, we cannot apply filters _inside_ of a sanitizer because then we cannot use the sanitizer args as a source of truth for computing a cache key for the post-processed output, as we can now.
1. We should be trending the sanitizers toward being pure PHP classes that are do not connect with WordPress APIs (especially the plugin API per above) as this is the direction needed for abstracting the sanitizers onto a CMS-agnostic package.

Additionally, this `amp_is_dev_mode()` function would be useful outside of the context of a sanitizer. For example, a theme/plugin may want to call this function during template generation to determine whether it can output some dev code (e.g. Query Monitor).

The `amp_is_dev_mode()` is gatekeeping whether or not the `AMP_Dev_Mode_Sanitizer` should be used in the first place. So it is operating at a higher level.

Aside: should it rather be named `is_amp_dev_mode()`?

tests/php/test-amp-helper-functions.php
westonruter: See https://github.com/ampproject/amp-wp/pull/3187#discussion_r322769113

westonruter:
What comes to mind is that we could mark Dashicons as being in dev mode only if it was not directly enqueued or a dependency of any enqueued stylesheet other than `admin-bar`.

includes/amp-helper-functions.php
schlessera: > Aside: should it rather be named is_amp_dev_mode()?

I'm seeing the `amp_` as a prefix here, as this could just as well be used outside of the AMP plugin, so I think the name is fine.

tests/php/test-amp-helper-functions.php
westonruter: Removed prior filter in d79be8df.

includes/class-amp-theme-support.php
```diff
@@ -1422,7 +1422,31 @@ protected static function has_dependency( WP_Dependencies $dependencies, $curren
 	 * @return string Tag.
 	 */
 	public static function filter_admin_bar_style_loader_tag( $tag, $handle ) {
-		if ( self::has_dependency( wp_styles(), $handle, 'admin-bar' ) ) {
+		if ( 'dashicons' === $handle ) {
+			// Conditionally include Dashicons in dev mode only if was included because it is a dependency of admin-bar.
+			$needs_dev_mode = true;
+			foreach ( wp_styles()->queue as $queued_handle ) {
+				$is_dashicons_without_admin_bar = (
```
schlessera: The variable `$is_dashicons_without_admin_bar` doesn't fulfil a function here, you can just replace it with `if` and immediately set the dev mode:
```php
				if (
					// If a theme or plugin directly enqueued dashicons, then it is not added via admin-bar dependency and it is not part of dev mode.
					'dashicons' === $queued_handle
					||
					// If a stylesheet has dashicons as a dependency without also having admin-bar as a dependency, then no dev mode.
					(
						self::has_dependency( wp_styles(), $queued_handle, 'dashicons' )
						&&
						! self::has_dependency( wp_styles(), $queued_handle, 'admin-bar' )
					)
				) {
					$needs_dev_mode = false;
					break;
				}
```

tests/php/test-class-amp-theme-support.php
```diff
@@ -1203,9 +1206,56 @@ function() {
...
+
+		// Enqueued directly.
+		wp_enqueue_style( 'dashicons' );
+
+		ob_start();
+		wp_print_styles();
+		$output = ob_get_clean();
+
+		$this->assertContains( '<link rel=\'stylesheet\' id=\'dashicons-css\'', $output ); // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedStylesheet
+		$this->assertContains( '<link data-ampdevmode rel=\'stylesheet\' id=\'admin-bar-css\'', $output ); // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedStylesheet
```
schlessera: Maybe add another assert here to make sure you don't have both the dev-mode and the non-dev-mode dashicons enqueue here?

tests/php/test-class-amp-theme-support.php
```diff
@@ -1203,9 +1206,56 @@ function() {
...
+		AMP_Theme_Support::init_admin_bar();
+
+		// Enqueued indirectly.
+		wp_enqueue_style( 'my-font-pack', 'https://example.com/fonts', [ 'dashicons' ], '0.1' );
+
+		ob_start();
+		wp_print_styles();
+		$output = ob_get_clean();
+
+		$this->assertContains( '<link rel=\'stylesheet\' id=\'dashicons-css\'', $output ); // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedStylesheet
```
schlessera: Maybe add another assert here to make sure you don't have both the dev-mode and the non-dev-mode dashicons enqueue here?

includes/class-amp-theme-support.php
schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3187/commits/5de22ffafdbb65ea87b5aeb97635472447a6a194

tests/php/test-class-amp-theme-support.php
schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3187/commits/290adf71350fb758be61c03bef9c94c208ee6761

schlessera:
@googlebot I consent.

Issue #2313, opened 2019-05-14
Highlight themes and plugins that are AMP-compatible inside WordPress


westonruter:
Currently the plugin has a short paragraph directing users to the [ecosystem page](https://amp-wp.org/ecosystem/) on amp-wp.org before the list of template modes (#2135):

![image](https://user-images.githubusercontent.com/134745/57722816-2a38ed80-763c-11e9-8371-72b7f0efa691.png)

This is a good start but it's not integrated into the admin screens for installing themes and plugins.

Add new default tabs on the “Add New” theme/plugin screens for “AMP-compatible”, when in 
Native/Transitional modes. This should selected instead of “Featured” because it's important to surface themes/plugins that actually work with AMP.

![image](https://user-images.githubusercontent.com/134745/57723237-0fb34400-763d-11e9-822c-6b585426a3e2.png)

![image](https://user-images.githubusercontent.com/134745/57723174-ee525800-763c-11e9-87c3-f62bc6562e65.png)

In addition, when searching for themes/plugins we could intercept the installation/activation action to show a warning whether or not it is known to be AMP-compatible.

Lastly, themes/plugins being listed could have an AMP badge for those known to be compatible. Also, for those which we know are _not_ compatible, there could similarly be a warning badge, but this would require for us to maintain a list of incompatible plugins. For other themes and plugins which have an unknown compatibility status, there should be some message to indicate this.

----

## Implementation Brief

1. Add a build step which downloads the AMP-compatible themes and plugins from WordPress.org and adds it as a JSON file to the plugin for us to use. (In the future we may download the list dynamically at runtime, but for now we can have a static copy which may make things simpler. Or maybe not.)
2. Create a new tab in the Add Plugins screen (left of Featured) that lists out the AMP-compatible plugins using the data obtained in the first step. Only plugins on WordPress.org would have an Install button available. Other plugins will have just an "Learn More" button.
3. Create a new tab in the Add Themes screen (left of Popular) that lists out the AMP-compatible themes using the data obtained in the first step. Only themes on WordPress.org would have an Install button available. Other themes will have just an "Learn More" button.
4. When doing a search for plugins, decorate the resulting plugins with PX badges which are included in the list of AMP-compatible plugins. This will naturally be limited to the plugins on WordPress.org.
4. When doing a search for themes, decorate the resulting plugins with PX badges which are included in the list of AMP-compatible plugins. This will naturally be limited to the themes on WordPress.org.

amedina:
[...]

PR #3188, opened 2019-09-05
If a theme supports Reader Mode, display a notice

### LINKED ISSUE #2931: Indicate when active theme is designed for Reader mode
Hello,

is there a way to disable Standard and Transitional mode for a theme? 

When you add support for Standard it shows a message "Your active theme is known to work well in standard or transitional mode.". I'd like to show a message ~"Your active theme only works in Reader mode." and have the Standard and Transitional mode radio inputs disabled or hidden. I've had multiple users switch to these modes and then get confused why everything is broken.

Thanks!

kienstra:
The following notice displays when all of the following are true:
* The active theme doesn't call `add_theme_support( 'amp' )`
* The theme or child theme have an `amp/` directory
* 'Standard' or 'Transitional' mode are selected

As mentioned in #2931, this is for themes that have templates for 'Reader Mode,' and aren't intended for 'Standard' or 'Transitional Mode':

new-notice-reader-here

Also, the link to the Ecosystem page is removed in that case:

another-lnink

The notice above is modeled after an existing notice:

2015-notice

Closes #2931


tests/php/test-class-amp-theme-support.php
```diff
@@ -223,6 +233,33 @@ public function test_read_theme_support_and_support_args() {
 		$this->assertTrue( get_theme_support( AMP_Theme_Support::SLUG ) );
 	}
 
+	/**
+	 * Test theme_supports_reader_mode.
+	 *
+	 * @covers \AMP_Theme_Support::theme_supports_reader_mode.
+	 */
+	public function test_theme_supports_reader_mode() {
+		remove_all_filters( 'theme_root' );
```
kienstra: It looks like this [theme_root filter](https://github.com/ampproject/amp-wp/blob/0303912777b69b1857f2ab61bfcdd25d4361163a/tests/php/test-amp-style-sanitizer.php#L2241) changes the result of `get_template_directory()`, causing this test to fail. 

Maybe there's a better place to remove that filter, though.

Strangely, calling it in [AMP_Style_Sanitizer_Test::tearDown()](https://github.com/ampproject/amp-wp/blob/0303912777b69b1857f2ab61bfcdd25d4361163a/tests/php/test-amp-style-sanitizer.php#L29) didn't work.

kienstra:
**Request For Review**

Thanks, and have a great day.

kienstra:
Here's a [screencast for this](https://youtu.be/ZrKCpqBT5Mc) (unlisted on YouTube). Of course, I could record another one if the notice text, or anything else changes.

tests/php/test-class-amp-theme-support.php
```diff
@@ -223,6 +233,33 @@ public function test_read_theme_support_and_support_args() {
...
+		$this->assertFalse( AMP_Theme_Support::theme_supports_reader_mode() );
+
+		remove_theme_support( 'amp' );
+		AMP_Theme_Support::read_theme_support();
+
+		// The mode is Reader, but there is no /amp directory in the theme.
+		$this->assertFalse( AMP_Theme_Support::theme_supports_reader_mode() );
+
+		$this->amp_directory_in_theme = get_template_directory() . '/amp';
+		mkdir( $this->amp_directory_in_theme );
```
westonruter: This makes me wary. It would be preferable for filesystem changes made during tests to be done in the temp directory. 

Could there perhaps instead be new mock themes created  in `tests/data/themes` similar to how [there is in core](https://github.com/WordPress/wordpress-develop/tree/master/tests/phpunit/data/themedir1)? There could be a theme in standard mode, a theme in transitional mode, and a theme in reader mode.

Activating such a theme could be done with something like this:

```php
register_theme_directory( __DIR___ . '/data/themes' );
search_theme_directories( true ); // Regenerate the transient.
switch_theme( 'theme-with-reader-mode' );
```

Having such test data themes in the plugin will be helpful for the future to reduce current dependencies we have on core themes being installed in the test suite.

@schlessera Thoughts?

ktmn:
Works for me, thanks!

includes/options/class-amp-options-menu.php
```diff
@@ -318,9 +319,13 @@ public function render_theme_support() {
 						<p><?php esc_html_e( 'Your active theme is known to work well in standard or transitional mode.', 'amp' ); ?></p>
 					</div>
 				<?php endif; ?>
+			<?php elseif ( AMP_Theme_Support::theme_supports_reader_mode() ) : ?>
+				<div class="notice notice-success notice-alt inline">
+					<p><?php echo wp_kses_post( $reader_mode_support ); ?></p>
```
schlessera: `wp_kses_post()` might be overly generous here, how about being specific for the expected translation:
```suggestion
					<p><?php echo wp_kses( $reader_mode_support, [ 'strong' => [] ] ); ?></p>
```

schlessera: `<?php echo` can be simplified to `<?=`:
```suggestion
					<p><?= wp_kses_post( $reader_mode_support ) ?></p>
```
This is safe to use starting from PHP 5.4, as it was decoupled from the [`short_open_tag`](https://www.php.net/manual/en/ini.core.php#ini.short-open-tag) setting that caused conflicts with XML.

includes/class-amp-theme-support.php
```diff
@@ -364,6 +364,23 @@ public static function get_theme_support_args() {
...
 
+	/**
+	 * Gets whether the parent or child theme supports Reader Mode.
+	 *
+	 * True if the theme does not call add_theme_support( 'amp' ) at all,
+	 * and it has an amp/ directory for templates.
+	 *
+	 * @return bool Whether the theme supports Reader Mode.
+	 */
+	public static function theme_supports_reader_mode() {
```
schlessera: The class is called `AMP_Theme_Support`, and this is part of the fully qualified name. So the method name here is:
```php
AMP_Theme_Support::theme_supports_reader_mode()
```
which seems unnecessarily verbose. The entire class is about what the theme supports, so that prefix does not provide any added value.

Some alternatives for more natural reading would be:
```php
AMP_Theme_Support::supports_reader_mode()
```
or even
```php
AMP_Theme_Support::for_reader_mode()
```

The end goal should always be to have the code read as naturally as possible (while still being clear and precise, of course).

tests/php/test-class-amp-theme-support.php
schlessera: Yes, that could work and would be preferable to writing directly into a theme here.

Ideally, though, we wouldn't need the filesystem at all for unit tests. A cleaner approach would be to use a virtual filesystem that we fully control and can reset whenever we want using vfsStream: http://vfs.bovigo.org/

This has the following added benefits:
- Faster, everything happens in memory.
- No thrashing of the disk with temp files, so better for the disk health.
- Perfectly safe, as no operations hit the actual disk, so you don't risk a rogue delete operation to cause harm.
- No permissions issues, as you control the entire disk including the permissions, so more reliable across environments.
- Possibility to set up multiple differing filesystems, and switch as needed.

schlessera: Actually, as we don't have unit tests here, I'm not sure if we can make this work transparently with WordPress Core, so my above comment might not apply. We'd have to try if you can actually register a custom theme folder within that virtual file system without hitting issues with WordPress.

includes/class-amp-theme-support.php
kienstra: Great point. I'll change the name of this.

kienstra:
>Actually, as we don't have unit tests here, I'm not sure if we can make this work transparently with WordPress Core, so my above comment might not apply.

Good point, though. I'll look into this.

includes/options/class-amp-options-menu.php
kienstra: Ah, great idea.

kienstra: Ah, good point.

kienstra:
Thanks for the great feedback here. I'll apply it on (my) Monday if that's alright.

includes/class-amp-theme-support.php
kienstra: https://github.com/ampproject/amp-wp/pull/3188/commits/2bec6cc7e84615d6a5d151f4b606302304fed9f8 changes the name to `AMP_Theme_Support::supports_reader_mode()`

includes/options/class-amp-options-menu.php
kienstra: Ah, it would have been better if I simply committed your suggestion here. But I forgot it was there.

https://github.com/ampproject/amp-wp/pull/3188/commits/50c7a14a69663505f6b05fd6f7a7808d19c9f679 applies your suggestion.

westonruter: We don't currently use this shorter syntax anywhere else in the plugin, and it's currently causing a linting failure because we actually currently disallow it it: https://travis-ci.org/ampproject/amp-wp/jobs/582745144#L762-L771

So either we should revert to `<?php echo` or decide to do a mass update to `<?=`. I think we should do the former, and do the latter for the next release, all in one go, and add a PHPCS check to force `<?=` instead of `<?php echo` (old habits die hard!)

schlessera: Yes, agree, let's revert for now.

kienstra: Sure, https://github.com/ampproject/amp-wp/pull/3188/commits/3f2e83d5c1b5e191d08d3d8831fd353e4974b9a8 reverts it. Sorry for the delay.

includes/class-amp-theme-support.php
```diff
@@ -364,6 +364,23 @@ public static function get_theme_support_args() {
...
+		$amp_dir = '/amp';
```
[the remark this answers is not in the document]
kienstra: Good ideas. https://github.com/ampproject/amp-wp/pull/3188/commits/3ce024edadf1205c74c881904687ddeffee4fe21 uses `trailingslashit()`, and stores the `'amp'` directory name in `AMP_Theme_Support::READER_MODE_TEMPLATE_DIRECTORY`. What do you think about that?

tests/php/test-class-amp-theme-support.php
kienstra: Thanks, vfsStream actually seemed to work really well, though maybe my implementation could use some changes. How does https://github.com/ampproject/amp-wp/pull/3188/commits/634e5b09535c1deec5599570614ffbee6ce2c022 look?

kienstra:

Hi @schlessera and @westonruter,
Sorry for the delay. All the review points are now addressed. Could you please do another round of review?

Thanks, and hope you're having a great week so far.

includes/options/class-amp-options-menu.php
```diff
@@ -318,9 +319,13 @@ public function render_theme_support() {
 						<p><?php esc_html_e( 'Your active theme is known to work well in standard or transitional mode.', 'amp' ); ?></p>
 					</div>
 				<?php endif; ?>
+			<?php elseif ( AMP_Theme_Support::supports_reader_mode() ) : ?>
+				<div class="notice notice-success notice-alt inline">
+					<p><?php echo wp_kses( $reader_mode_support, [ 'strong' => [] ] ); ?></p>
+				</div>
 			<?php endif; ?>
 
-			<?php if ( ! AMP_Theme_Support::get_support_mode_added_via_theme() ) : ?>
```
westonruter: For consistency, I'm also including `! $builtin_support` in the conditions here. See ffee40fd. In reality `AMP_Theme_Support::get_support_mode_added_via_theme()` should actually return `true` when `in_array( get_template(), AMP_Core_Theme_Sanitizer::get_supported_themes(), true )`.

We'll want to revisit how we help users discover AMP-compatible themes in https://github.com/ampproject/amp-wp/issues/2313.

PR #3193, opened 2019-09-05
Fix vertical centering of fitted text block content

### LINKED ISSUE #3113: Text is not centered when not in edit mode
Steps to reproduce:
- Add a text block (leave Auto fit text setting on.)
- Write "Hello" (or something else) in it.
- Resize the block to be quite long.

Results:
Screenshot 2019-08-28 at 21 04 39

Expected results:
The text should be vertically aligned in the middle.

barklund:
Fixed #3113

**Background**
The error occurs because the `.textbox` further in uses `vertical-align: middle` to align itself relative vertically centered relative to its parent, but this only works if the text box doesn't inherit this 100% height.

Even with the 100% height declaration, the text box sometimes (in particular on first load) receives the proper less-than-100% height, but as soon as the outer block is resized, the text box would switch back to top-aligned regardless of fit.

**Fix**
With this change, the inner text box will always be vertically centered. This has been confirmed across Chrome, Firefox, Safari and MSEdge - all able to display the error before, all not showing the error after.

**Tests**
This PR also adds two tests for proper vertical alignment of fitted text blocks contents within their block as a whole.

The first test validates that (once at max font size) a text box does not change size when the surrounding block is resized. This failed before this fix was in place.

The second test validates that the text box is (_more or less_) vertically centered within the surrounding block. Combined with the first test, this ensures that the text must be vertically aligned within the block.

Neither test verifies, that the text changes font size correctly when the block is resized. However, the second test happens to validate that the block is correctly aligned after it happens to change font size when resized (but works regardless of the font resizing).


swissspidy:
Resizing a text block when the "Automatically fit text to container" option (aka amp-fit-text) is **on** works well.

However, when I turn off that option, I cannot drag the block anywhere anymore. I can only drag the area where the text is:

Screenshot 2019-09-05 at 16 55 14

barklund:
> However, when I turn off that option, I cannot drag the block anywhere anymore. I can only drag the area where the text is:

Ah yes, then both of the two first `<div>`'s need to be 100% height. And the first descendant has zero classes or attributes I can recognise it by, nor do I expect that it's possible to add any.

There are two possible solutions, neither looking good, but the latter is probably slightly cleaner:

**Direct and direct-direct**
```
.wp-block[data-type="amp/amp-story-text"] .block-mover > div,
.wp-block[data-type="amp/amp-story-text"] .block-mover > div > div {
```

**Direct and draggable**
```
.wp-block[data-type="amp/amp-story-text"] .block-mover > div,
.wp-block[data-type="amp/amp-story-text"] .block-mover div[draggable] {
```

Thoughts?

swissspidy:
Both is fine as long as there is an inline comment explaining it 👍

spacedmonkey:
Why are text blocks with 'Automatically fit text to container' disabled not aligned to middle? That is really confusing as a user if I turn that setting off for my text to move.

barklund:
> Why are text blocks with 'Automatically fit text to container' disabled not aligned to middle? That is really confusing as a user if I turn that setting off for my text to move.

That's a good question. I will forward it to the channel.

swissspidy:
@barklund I still cannot drag the text block anywhere below the text when fit-text is turned off. Example:

Screenshot 2019-09-06 at 16 01 26

I can drag when hovering over "Hello World", but not when dragging anywhere on that rectangle.

Please test this scenario as well.

swissspidy:
As mentioned on Slack, vertical centering behavior is intentional.

We could, however, create a new issue to enable support for vertical centering for non-amp-fit-text blocks. Perhaps by using the `BlockVerticalAlignmentToolbar` component

Issue #1915, opened 2019-02-28
Turn off display_errors during the post-processing phase (sanitization)


westonruter:
A user reported a two fatal errors showing up in the log:

```
[28-Feb-2019 17:51:38 UTC] PHP Fatal error:  Allowed memory size of 33554432 bytes exhausted (tried to allocate 72 bytes) in .../wp-content/plugins/amp/vendor/sabberworm/php-css-parser/lib/Sabberworm/CSS/Parser.php on line 784
[28-Feb-2019 17:51:38 UTC] PHP Fatal error:  Unknown: Cannot use output buffering in output buffering display handlers in Unknown on line 0
```

The second error is a symptom of the first (for that see https://github.com/ampproject/amp-wp/issues/1914). The site has `display_errors` turned on and so the printing of the first error is causing the second error. To prevent the second error from happening, we need to turn off `display_errors` during post-processing.

This actually identifies an improvement to be made in the plugin: we need to turn off display_errors during post-processing so that this doesn't happen. This should be done in the `AMP_Theme_Support::prepare_response()` method:

https://github.com/ampproject/amp-wp/blob/616d7a3397b3b3fe3545efbf2559eb21437c5aaa/includes/class-amp-theme-support.php#L1616-L1632

westonruter:
I was able to reproduce this by forcing my install to have very low memory. I added this to my `wp-config.php`:

```php
if ( php_sapi_name() !== 'cli' ) {
	define( 'WP_MEMORY_LIMIT', '8M' );
	define( 'WP_MAX_MEMORY_LIMIT', WP_MEMORY_LIMIT );
	ini_set( 'memory_limit', WP_MEMORY_LIMIT );
}
```

And then I ran:

```bash
wp transient delete --all
```

And then when I tried accessing an AMP page, I got a white screen of death with this in the error log:

```
[04-Apr-2019 18:26:21 UTC] PHP Fatal error:  Allowed memory size of 8388608 bytes exhausted (tried to allocate 2097160 bytes) in /app/public/content/plugins/amp/vendor/sabberworm/php-css-parser/lib/Sabberworm/CSS/Parsing/ParserState.php on line 285
[04-Apr-2019 18:26:21 UTC] PHP Fatal error:  Unknown: Cannot use output buffering in output buffering display handlers in Unknown on line 0
```

westonruter:
[...]

PR #3198, opened 2019-09-05
Add notice if Xdebug extension is loaded

aaemnnosttv:
As discovered in https://github.com/ampproject/amp-wp/issues/1915#issuecomment-528117914, xdebug can cause AMP stylesheet parsing to take much longer than normal and potentially crash the site due to maximum execution limit timeout.

This PR adds a notice to inform the user of the potential risk.

I would expect perhaps the wording should maybe change, if so, just let me know!

### DIFF
--- a/amp.php
+++ b/amp.php
@@ -251,8 +251,9 @@ function _amp_xdebug_admin_notice() {
 	<div class="notice notice-warning">
 		<p>
 			<?php
-			echo wp_kses_post(
-				__( 'Your server currently has the Xdebug PHP extension loaded. This can cause some of the AMP plugin\'s processes to timeout depending on your system resources and configuration. Please deactivate Xdebug for the best experience.', 'amp' )
+			echo esc_html_e(
+				'Your server currently has the Xdebug PHP extension loaded. This can cause some of the AMP plugin\'s processes to timeout depending on your system resources and configuration. Please deactivate Xdebug for the best experience.',
+				'amp'
 			);
 			?>
 		</p>
--- a/amp.php
+++ b/amp.php
@@ -252,7 +252,7 @@ function _amp_xdebug_admin_notice() {
 		<p>
 			<?php
 			echo wp_kses_post(
-				__( 'We\'ve detected that your server currently has the Xdebug PHP extension loaded. This can cause some of the AMP plugin\'s processes to timeout depending on your system resources and configuration. We recommend deactivating Xdebug for the best experience.', 'amp' )
+				__( 'Your server currently has the Xdebug PHP extension loaded. This can cause some of the AMP plugin\'s processes to timeout depending on your system resources and configuration. Please deactivate Xdebug for the best experience.', 'amp' )
 			);
 			?>
 		</p>
--- a/amp.php
+++ b/amp.php
@@ -241,6 +241,28 @@ if ( 'amp' !== basename( AMP__DIR__ ) ) {
 	add_action( 'admin_notices', '_amp_incorrect_plugin_slug_admin_notice' );
 }
 
+/**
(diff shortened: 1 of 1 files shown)


amp.php
```diff
@@ -241,6 +241,28 @@ function _amp_incorrect_plugin_slug_admin_notice() {
 	add_action( 'admin_notices', '_amp_incorrect_plugin_slug_admin_notice' );
 }
 
+/**
+ * Print admin notice if the xdebug extension is loaded.
+ *
+ * @since T.B.D
```
westonruter: This can be set to `1.3`.

amp.php
```diff
@@ -241,6 +241,28 @@ function _amp_incorrect_plugin_slug_admin_notice() {
...
+ */
+function _amp_xdebug_admin_notice() {
+	?>
+	<div class="notice notice-warning">
+		<p>
+			<?php
+			echo wp_kses_post(
+				__( 'We\'ve detected that your server currently has the Xdebug PHP extension loaded. This can cause some of the AMP plugin\'s processes to timeout depending on your system resources and configuration. We recommend deactivating Xdebug for the best experience.', 'amp' )
```
westonruter:
```suggestion
				__( 'Your server currently has the Xdebug PHP extension loaded. This can cause some of the AMP plugin\'s processes to timeout depending on your system resources and configuration. Please deactivate Xdebug for the best experience.', 'amp' )
```

amp.php
```diff
@@ -241,6 +241,28 @@ function _amp_incorrect_plugin_slug_admin_notice() {
...
+ * Print admin notice if the xdebug extension is loaded.
+ *
+ * @since T.B.D
+ */
+function _amp_xdebug_admin_notice() {
+	?>
+	<div class="notice notice-warning">
+		<p>
+			<?php
+			echo wp_kses_post(
```
westonruter: Since there are no tags appearing in the translation string, this could rather use `esc_html_e()`.

amp.php
```diff
@@ -241,6 +241,28 @@ function _amp_incorrect_plugin_slug_admin_notice() {
 	add_action( 'admin_notices', '_amp_incorrect_plugin_slug_admin_notice' );
 }
 
+/**
+ * Print admin notice if the xdebug extension is loaded.
```
westonruter:
```suggestion
 * Print admin notice if the Xdebug extension is loaded.
```

aaemnnosttv:
Updated per your feedback @westonruter.

amp.php
```diff
@@ -241,6 +241,29 @@ function _amp_incorrect_plugin_slug_admin_notice() {
...
+ * Print admin notice if the Xdebug extension is loaded.
+ *
+ * @since 1.3
+ */
+function _amp_xdebug_admin_notice() {
+	?>
+	<div class="notice notice-warning">
+		<p>
+			<?php
+			echo esc_html_e(
```
swissspidy: Double echo

aaemnnosttv: It was late 🤦‍♂

leecollings:
This should never have been added. If something in your plugin is making websites crash when xdebug is enabled, then surely shouldn’t the issue in your plugin be fixed rather than advising the user to disable the extension?

It’s not the users fault for having xdebug enabled, and most of the time it’s intentionally enabled so there should be no reason to force users to have to see a notice saying they’re doing something wrong.

There’s obviously an issue in the coding of your plugin that needs fixing.

westonruter:
@leecollings if it bothers you so much, you can remove it yourself in another plugin with just this:

```php
remove_action( 'admin_notices', '_amp_xdebug_admin_notice' );
```

swissspidy:
I've created a new issue to discuss the placement of the warning: #3499.

PR #3201, opened 2019-09-06
Introduce default story settings

### LINKED ISSUE #1995: Global / per Story Slide Duration Options

dero:
Fixes #1995 (Global and Per story slide duration options)

**Scopes**

- settings (global): a set of global settings to be managed on the AMP > General page.
- settings (story): the same set of per story settings to be managed in the Document > Story Settings.
- settings (page): the original Page settings.

**Tasks**

- [x] Add settings (global) and use `AMP_Options_Manager` to store them.
- [x] Make sure the settings (global) inputs only display when AMP Stories are enabled.
- [x] Use `PluginDocumentSettingPanel` to create a panel to hold the settings (story).
- [x] When a new **story** is created, make sure the current settings (global) are copied into new story's meta and that settings (story) are connected to that meta.
- [x] When a **page** is created, make sure the settings (story) are applied as defaults.
- [x] ~Implement sanitizer and serialization / deserialization handler for `register_meta` in order to be able to pass an array of settings to the editor instead of passing them individually.~
- [x] Add unit test
- [x] Add e2e tests.


assets/src/stories-editor/plugins/story-settings-panel.js
```diff
@@ -0,0 +1,116 @@
...
+import { metaToAttributeNames } from '../helpers';
+
+/**
+ * WordPress dependencies
+ */
+import { PluginDocumentSettingPanel } from '@wordpress/edit-post';
+import { __ } from '@wordpress/i18n';
+import { compose } from '@wordpress/compose';
+
+import {
```
swissspidy: Minor nitpick: Unnecessary empty lines before and after this import

assets/src/stories-editor/plugins/story-settings-panel.js
```diff
@@ -0,0 +1,116 @@
...
+	withSelect,
+	withDispatch,
+} from '@wordpress/data';
+
+import {
+	SelectControl,
+	RangeControl,
+} from '@wordpress/components';
+
+let MetaFields = ( props ) => {
```
swissspidy: Instead of using `let` I'd name a "dumb" component like this "MetaFields" and then the enhanced one "EnhancedMetaFields` or so.

assets/src/stories-editor/plugins/story-settings-panel.js
```diff
@@ -0,0 +1,116 @@
...
+
+	if ( 'media' === autoAdvanceAfter ) {
+		autoAdvanceAfterHelp = __( 'Based on the first media block encountered on the page', 'amp' );
+	} else if ( 'auto' === autoAdvanceAfter ) {
+		autoAdvanceAfterHelp = __( 'Based on the duration of all animated blocks on the page', 'amp' );
+	}
+
+	return (
+		<>
+			<SelectControl
```
swissspidy: Can we add an explanation that this changes the setting globally for the whole story? I don't think it's clear otherwise.

includes/class-amp-story-post-type.php
```diff
@@ -2053,4 +2079,107 @@ private static function generate_export( $post_id ) {
...
+	 * @since 1.3
+	 *
+	 * @return array
+	 *
+	 * - meta_args array Arguments passed to `register_meta`; sanitize_callback is required.
+	 * - data      array Any additional data.
+	 */
+	public static function get_stories_settings_meta_definitions() {
+		return apply_filters(
+			'amp_stories_settings_meta_definitions',
```
swissspidy: Is a filter here really needed? If so, it would need a docblock. But I think we could remove that filter.

includes/class-amp-story-post-type.php
```diff
@@ -2053,4 +2079,107 @@ private static function generate_export( $post_id ) {
...
+						},
+					],
+					'data'      => [],
+				],
+			]
+		);
+	}
+
+	/**
+	 * Adds stories global settings as post meta to all new Stories.
```
swissspidy: To clarify: the intention here is that the global setting is applied during post creation, meaning that changing the global setting won't apply to pre-existing stories after the fact?

includes/options/class-amp-options-manager.php
```diff
@@ -35,17 +35,19 @@ class AMP_Options_Manager {
...
+		'analytics'                                    => [],
+		'auto_accept_sanitization'                     => true,
+		'all_templates_supported'                      => true,
+		'supported_templates'                          => [ 'is_singular' ],
+		'enable_response_caching'                      => true,
+		'version'                                      => AMP__VERSION,
+		'story_templates_version'                      => false,
+		'story_export_base_url'                        => '',
+		'stories_settings_auto_advance_after'          => '',
+		'stories_settings_auto_advance_after_duration' => 0,
```
swissspidy: Separate option entries for every story default is not really scalable. I'd use one entry for all the story default settings. Thoughts?

Also, all the other options here are prefixed `story_`, not `_stories`.

includes/options/class-amp-options-menu.php
```diff
@@ -670,6 +681,88 @@ public function render_stories_export() {
...
+		 * There should be a single source for this data to be shared in the application.
+		 *
+		 * @todo Figure out the best way to keep these settings DRY.
+		 */
+		$auto_advance_after_options = $stories_settings_definitions['stories_settings_auto_advance_after']['data']['options'];
+
+		$auto_advance_after          = AMP_Options_Manager::get_option( 'stories_settings_auto_advance_after', '' );
+		$auto_advance_after_duration = AMP_Options_Manager::get_option( 'stories_settings_auto_advance_after_duration', 1 );
+		?>
+		<fieldset <?php disabled( ! current_user_can( 'publish_posts' ) ); ?>>
```
swissspidy: I'd not print these settings in the first place when the user cannot edit them.

To see the settings screen one needs the `manage_options` cap anyway, no? So there is not much value in this check here.

includes/options/class-amp-options-menu.php
```diff
@@ -670,6 +681,88 @@ public function render_stories_export() {
...
+	 * @since 1.3
+	 */
+	private function render_stories_settings_page_advance() {
+		$stories_settings_definitions = AMP_Story_Post_Type::get_stories_settings_meta_definitions();
+
+		/**
+		 * Currently the list of options with values and descriptions is hardcoded in PHP and JS.
+		 * There should be a single source for this data to be shared in the application.
+		 *
+		 * @todo Figure out the best way to keep these settings DRY.
```
swissspidy: Since #3171 we have this new `window.ampStoriesEditorSettings` global and a `getSettings()` selector. Could we use that to pass these defaults to the editor?

includes/class-amp-story-post-type.php
dero: @swissspidy Correct.

includes/options/class-amp-options-manager.php
dero: @swissspidy Agreed - and I'll also fix the nomenclature.

One issue I ran into is with `register_meta`. The meta field holding all settings needs to be registered in order for the editor to be able to pick it up. However, it requires a `type` value to be specified and the option to specify `array` or `object` theme will be added in WordPress 5.3.

I'm working on adding logic to serialize and deserialize the settings as well as on an updated sanitizer to be able to handle the meta contents.

swissspidy: The meta fields can still be individual and not arrays. But in the Options  Manager class an array makes more sense.

swissspidy: Serialized meta fields prevent things like easy filtering via `WP_Query` too :/

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -97,7 +99,7 @@ class PageEdit extends Component {
 	 * @param {Object} media.image      Media image object.
 	 * @param {string} media.image.src  Media image URL
 	 */
-	onSelectMedia = ( media ) => {
+	onSelectMedia( media ) {
```
swissspidy: This should be changed back to use class properties

```suggestion
	onSelectMedia = ( media ) => {
```

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -83,6 +84,7 @@ class PageEdit extends Component {
 		};
 
 		this.videoPlayer = createRef();
+		this.onSelectMedia = this.onSelectMedia.bind( this );
```
swissspidy: This should be changed back to use class properties

```suggestion

```

dero:
**Summary for QA:**

This PR introduces a concept of global settings and story settings. 

- Global settings: set on AMP plugin settings page, applied as defaults to every new story.
- Story settings: set on a story post, applied as defaults to every new page.
- Page settings: set on a page block, applied to that block only.

**Test instructions:**

1. Go to AMP plugin settings page and enable the Stories experience.
2. Change the values in the Stories Settings field and check the values persist upon saving.
3. Create a new story and notice a "Story Settings" panel in the "Document" tab.
4. Verify that values in that panel match what was set globally in step 2.
5. Click the first page and on the "Block" tab verify that the values in the "Page Settings" box match the "Story Settings".
6. Change the "Story Settings" to something else in the "Document" tab.
7. Verify that the "Page Settings" in the "Block" tab did not change.
8. Create a new page.
9. Verify that the "Page Settings" in the "Block" tab are equal to what was set in step 6.
10. Create a new page.
11. Change the "Page Settings" in the "Block" tab for that page to something else.
12. Check that other pages were not affected and their "Page Settings" stayed intact.
13. Check that the "Story Settings" in the "Document" tab remained unchanged.
14. Finish the story and publish it.
15. Verify that the advancement options are applied correctly on the front-end.

**Screencast:** https://www.youtube.com/watch?v=TppJ96XE-qU

includes/class-amp-story-post-type.php
```diff
@@ -956,11 +992,19 @@ public static function enqueue_block_editor_scripts() {
...
+		$meta_definitions         = self::get_stories_settings_definitions();
+		$auto_advancement_options = $meta_definitions['auto_advance_after']['data']['options'];
+
 		wp_localize_script(
 			self::AMP_STORIES_SCRIPT_HANDLE,
 			'ampStoriesEditorSettings',
 			[
 				'allowedVideoMimeTypes' => $allowed_video_mime_types,
+				'storySettings'         => [
+					'metaKey'                 => self::STORY_SETTINGS_OPTION,
```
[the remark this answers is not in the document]
dero: No, that's a relict from a previous implementation - thanks for noticing. :+1:

assets/src/stories-editor/plugins/story-settings-panel.js
```diff
@@ -0,0 +1,108 @@
...
+	} ),
+)( MetaFields );
+
+export const name = 'amp-story-settings-panel';
+
+export const render = () => {
+	return (
+		<PluginDocumentSettingPanel
+			name={ name }
+			title={ __( 'Story Settings', 'amp' ) }
```
swissspidy: Is it expected that this adds an icon next to the title? Looks a bit odd:

Screenshot 2019-09-11 at 19 18 47

assets/src/stories-editor/plugins/story-settings-panel.js
```diff
@@ -0,0 +1,108 @@
...
+	const {
+		autoAdvanceAfter,
+		autoAdvanceAfterDuration,
+	} = metaToAttributeNames( meta );
+
+	const currentOption = autoAdvanceAfterOptions.find( ( i ) => i.value === autoAdvanceAfter ) || {};
+
+	return (
+		<>
+			<p>{ __( 'These settings are applied to new pages.', 'amp' ) }</p>
```
swissspidy: Why only new pages?

When I create a new story and immediately change this setting, I'd expect this to change the value for the existing first page as well.

includes/options/class-amp-options-menu.php
```diff
@@ -670,6 +681,81 @@ public function render_stories_export() {
 		<?php
 	}
 
+	/**
+	 * Render the stories settings section.
+	 *
+	 * @since 1.3
+	 */
+	public function render_stories_settings() {
+		$this->render_stories_settings_page_advance();
```
swissspidy: When the stories experienced is **disabled** on the AMP settings screen, the "Stories Settings" section should be **hidden**. This is currently the case for the "Stories Export" section.

There is some inline JS on the settings page that handles this.

Screenshots:

Screenshot 2019-09-11 at 19 22 45
Screenshot 2019-09-11 at 19 22 38

Speaking of which, it makes sense to incorporate the "Stories Export" setting into this "Stories Settings" section so both settings (export  and page advance) are under the "Stories Settings" section.

Mockup:

Screenshot 2019-09-11 at 19 27 05

includes/options/class-amp-options-menu.php
```diff
@@ -670,6 +681,81 @@ public function render_stories_export() {
...
+		$definitions                = AMP_Story_Post_Type::get_stories_settings_definitions();
+		$auto_advance_after_options = $definitions['auto_advance_after']['data']['options'];
+
+		$story_settings            = AMP_Options_Manager::get_option( 'story_settings' );
+		$story_settings_field_name = sprintf( '%s[%s]', AMP_Options_Manager::OPTION_NAME, AMP_Story_Post_Type::STORY_SETTINGS_OPTION );
+		?>
+		<fieldset>
+			<p>
+				<label for="stories_settings_auto_advance_after">
+					<strong><?php echo esc_html__( 'Advance to next page', 'amp' ); ?></strong>
```
swissspidy: In the story editor, there is the "These settings are applied to new pages." help text.

Here, I'd expect a similar help text: "These settings are applied to new stories only."

assets/src/stories-editor/plugins/story-settings-panel.js
dero: @swissspidy Because it seems more consistent than any alternative where changing a story-wide settings may or may not affect any of the existing pages without proper explanation.

Ideally, I believe, it should be possible to _see_ a link between page settings to story settings - and the UI should clearly show which page settings are linked to story settings and which are changed and applied to that particular page only.

We might consider creating a UX ticket to investigate the various options here and enhance the experience in a later iteration.

I'm ready to defer to your experience and implement a change, just please let me know when you think should story settings _stop_ immediately affecting the page settings - is it after one of the page settings is changed by the user _or_ when a new page is created?

assets/src/stories-editor/plugins/story-settings-panel.js
dero: @swissspidy Agreed, but it is both expected and somewhat required.

- `registerPlugin` adds this [default icon](https://github.com/WordPress/gutenberg/blob/9fef2e82720b2ddf4d7d8c8bd3a42a24ae721840/packages/plugins/src/api/index.js#L131) to its context.
- `PluginDocumentSettingsPanel` [picks it up](https://github.com/WordPress/gutenberg/blob/aff692edeb78499df84f897f77befb240c5f88e8/packages/edit-post/src/components/sidebar/plugin-document-setting-panel/index.js#L97) as a fallback.

Seems like the simplest way to get rid of the icon there is to provide an incorrect icon name or a blank SVG icon. Neither seems like a great solution. What do you propose?

includes/options/class-amp-options-menu.php
dero: @swissspidy Incorporating the export setting into Stories Settings may be confusing in tandem with the additional request you made:

> Here, I'd expect a similar help text: "These settings are applied to new stories only."

The logical place for this text is on top of all the settings, but it's not true the export setting is applied to new stories only. That's a universal setting. It's only these newly added settings that only apply to new stories.

I'd suggest we keep them separate - at least for the time being. And maybe explore this concept of "defaults" vs. "settings" more from the UX perspective later on.

assets/src/stories-editor/plugins/export.js
```diff
@@ -92,6 +92,8 @@ const handleExport = ( { postId, createErrorNotice, createSuccessNotice, removeN
 
 export const name = 'amp-story-export';
 
+export const icon = 'media-archive';
```
spacedmonkey: What does this do?

swissspidy: Define the icon that is used when registering the plugin here:

https://github.com/ampproject/amp-wp/blob/50bc094b83764fa3409b60694606c1660f24f994/assets/src/stories-editor/index.js#L306-L312

spacedmonkey:
What happens, if you set the defaults. Create 3 stories.Then change the defaults again? What is the expected behaviour here?

tests/e2e/specs/stories-editor/page.js
```diff
@@ -218,4 +248,102 @@ describe( 'Story Page', () => {
 		const previewPage = await openPreviewPage( editorPage, 'amp-story' );
 		await previewPage.waitForXPath( '//amp-video[contains(@aria-label, "Hello World")]' );
 	} );
+
+	it( 'should pass global story settings to a new story', async () => {
+		const documentPanelSelector = 'button.edit-post-sidebar__panel-tab[data-label="Document"]';
+		const storySettingsPanelSelector = '.amp-story-settings-panel';
+		const storySettingsAdvanceAfterSelector = '.components-panel__body.amp-story-settings-panel .components-select-control__input:nth-of-type(1)';
```
spacedmonkey: You could add a custom class to the select menu here, to avoid kind of expensive selector.

tests/e2e/specs/stories-editor/page.js
```diff
@@ -218,4 +248,102 @@ describe( 'Story Page', () => {
...
+		const storySettingsAdvanceAfterDurationSelector = '.components-panel__body.amp-story-settings-panel [aria-label="Time in seconds"]';
```
spacedmonkey: Custom class might be better again.

tests/e2e/specs/stories-editor/page.js
```diff
@@ -218,4 +248,102 @@ describe( 'Story Page', () => {
...
+
+		expect( await page.evaluate( ( selector ) => {
+			return document.querySelector( selector ).value;
+		}, storySettingsAdvanceAfterDurationSelector ) ).toBe( '2' );
+	} );
+
+	it( 'should save the story settings correctly', async () => {
+		const documentPanelSelector = 'button.edit-post-sidebar__panel-tab[data-label="Document"]';
+		const storySettingsPanelSelector = '.amp-story-settings-panel';
+		const storySettingsAdvanceAfterSelector = '.amp-story-settings-panel .components-select-control__input:nth-of-type(1)';
```
spacedmonkey: This same selector seems to be reused. Could this be a const defined globally?

swissspidy:
> What happens, if you set the defaults. Create 3 stories.Then change the defaults again? What is the expected behaviour here?

According to https://github.com/ampproject/amp-wp/pull/3201#discussion_r322448841 the existing stories will be unaffected by this.

Might be good to have a test case covering that scenario if not already present.

tests/e2e/specs/stories-editor/page.js
```diff
@@ -218,4 +248,102 @@ describe( 'Story Page', () => {
...
+		await page.click( storySettingsPanelSelector );
+		await page.waitForSelector( storySettingsAdvanceAfterSelector );
+
+		expect( await page.evaluate( ( selector ) => {
+			return document.querySelector( selector ).value;
+		}, storySettingsAdvanceAfterSelector ) ).toBe( 'auto' );
+	} );
+
+	it( 'should apply story settings to newly created pages', async () => {
+		await page.click( '.block-editor-inserter .editor-inserter__amp-inserter' );
```
spacedmonkey: Why not just use `		await insertBlock( 'Page' );`?

tests/e2e/specs/stories-editor/page.js
```diff
@@ -218,4 +248,102 @@ describe( 'Story Page', () => {
...
+		const storySettingsAdvanceAfterSelector = '.components-panel__body.amp-story-settings-panel .components-select-control__input:nth-of-type(1)';
+		const storySettingsAdvanceAfterDurationSelector = '.components-panel__body.amp-story-settings-panel [aria-label="Time in seconds"]';
+
+		await page.click( documentPanelSelector );
+		await page.waitForSelector( storySettingsPanelSelector );
+
+		await page.click( storySettingsPanelSelector );
+		await page.waitForSelector( storySettingsAdvanceAfterSelector );
+
+		expect( await page.evaluate( ( selector ) => {
```
spacedmonkey: Could we make the following into the helper function?
```
const getValueBySelet = ( el ) => {
         return await page.evaluate( ( selector ) => {
	           return document.querySelector( selector ).value;
	}, el ) );
}
```
Same code seems to be being reused a couple of times.

tests/e2e/specs/stories-editor/page.js
```diff
@@ -218,4 +248,102 @@ describe( 'Story Page', () => {
...
+			return document.querySelector( selector ).value;
+		}, pageAdvancementSelector ) ).toBe( 'time' );
+
+		expect( await page.evaluate( ( selector ) => {
+			return document.querySelector( selector ).value;
+		}, secondsSelector ) ).toBe( '2' );
+	} );
+
+	it( 'should not affect story settings when individual page settings are changed', async () => {
+		await page.click( '.block-editor-inserter .editor-inserter__amp-inserter' );
```
spacedmonkey: Why not just use await insertBlock( 'Page' );?

tests/e2e/specs/stories-editor/page.js
spacedmonkey: You could do the same with 
```
		await page.evaluate( ( selector ) => {
			document.querySelector( selector ).value = '';
		}, secondsSelector );
```

tests/php/test-class-amp-story-post-type.php
```diff
@@ -696,4 +721,98 @@ public function get_gfont_data() {
...
+		$definitions = AMP_Story_Post_Type::get_stories_settings_definitions();
+		$this->assertEquals( $default_definitions, $definitions );
+	}
+
+	/**
+	 * Test the definitions return value
+	 *
+	 * @covers AMP_Story_Post_Type::add_story_settings_meta_to_new_story
+	 */
+	public function test_add_story_settings_meta_to_new_story() {
```
spacedmonkey: Should we also test an existing story (post), so that we can check existing stories were unaffected.

tests/php/test-class-amp-story-post-type.php
```diff
@@ -696,4 +721,98 @@ public function get_gfont_data() {
...
+		];
+	}
+
+	/**
+	 * Test the definitions return value
+	 *
+	 * @dataProvider get_default_settings_definitions
+	 * @covers AMP_Story_Post_Type::get_stories_settings_meta_definitions
+	 */
+	public function test_get_stories_settings_meta_definitions( $default_definitions ) {
```
spacedmonkey: :100:

assets/src/stories-editor/plugins/story-settings-panel.js
```diff
@@ -0,0 +1,111 @@
...
+			<SelectControl
+				label={ __( 'Advance to next page', 'amp' ) }
+				help={ currentOption.description || '' }
+				value={ autoAdvanceAfter }
+				options={ autoAdvanceAfterOptions }
+				onChange={ ( value ) => updateMeta( { amp_story_auto_advance_after: value } ) }
+			/>
+			{ 'time' === autoAdvanceAfter && (
+				<RangeControl
```
spacedmonkey: Add a min and max here.

assets/src/stories-editor/plugins/story-settings-panel.js
```diff
@@ -0,0 +1,111 @@
...
+
+		return {
+			meta: getEditedPostAttribute( 'meta' ),
+			autoAdvanceAfterOptions,
+		};
+	} ),
+	withDispatch( ( dispatch, { meta } ) => {
+		const { editPost } = dispatch( 'core/editor' );
+		return {
+			updateMeta( newMeta ) {
```
spacedmonkey: :+1:

assets/src/stories-editor/plugins/story-settings-panel.js
spacedmonkey: I agree with @swissspidy global settings should allow to everything, as the name implies. Other this is really confusing for users.

tests/e2e/specs/stories-editor/page.js
```diff
@@ -218,4 +248,102 @@ describe( 'Story Page', () => {
...
+		await page.click( documentPanelSelector );
+		await page.waitForSelector( storySettingsPanelSelector );
+
+		await page.click( storySettingsPanelSelector );
+		const durationSettingHandle = await page.waitForXPath( storySettingsAdvanceAfterDurationXPathSelector );
+
+		expect( await page.evaluate( ( node ) => {
+			return node.value;
+		}, durationSettingHandle ) ).toBe( '2' );
+	} );
```
spacedmonkey: There should be a test for following. 

Set defaults. 
Create story. 
Save story. 
check values. 
Set defaults to something new. 
Create story. 
Save story. 
check values. 
Load original story. 
Check values.

spacedmonkey: https://github.com/ampproject/amp-wp/pull/3201#issuecomment-531178001

tests/php/test-class-amp-story-post-type.php
dero: Added a test for this.

tests/e2e/specs/stories-editor/page.js
dero: @spacedmonkey Because for some reason I completely ignored its existence. :)

tests/e2e/specs/stories-editor/page.js
dero: Done and added to the test file. Doesn't seem to be universal enough use-case to warrant its own helper.

tests/e2e/specs/stories-editor/page.js
dero: Good idea, added a few `const`s for some of the repeated selectors.

assets/src/stories-editor/plugins/story-settings-panel.js
dero: @swissspidy Created a ticket to figure out which UI updates does this feature need and linked both your relevant comments on this PR: https://github.com/ampproject/amp-wp/issues/3238

spacedmonkey:
I am seeing more tests failing locally. 
```
  ● Story Page › should save the story settings correctly

    TimeoutError: waiting for selector ".amp-story-settings-advance-after select" failed: timeout 30000ms exceeded

      271 |     it( 'should save the story settings correctly', async () => {
      272 |             await expandSettingsPanel();
    > 273 |             await page.waitForSelector( SETTINGS_ADVANCE_AFTER );
          |                        ^
      274 | 
      275 |             await page.select( SETTINGS_ADVANCE_AFTER, 'auto' );
      276 | 

      at new WaitTask (../../node_modules/puppeteer/lib/DOMWorld.js:561:28)
      at DOMWorld._waitForSelectorOrXPath (../../node_modules/puppeteer/lib/DOMWorld.js:490:22)
      at DOMWorld.waitForSelector (../../node_modules/puppeteer/lib/DOMWorld.js:444:17)
      at Frame.waitForSelector (../../node_modules/puppeteer/lib/FrameManager.js:628:47)
      at Frame.<anonymous> (../../node_modules/puppeteer/lib/helper.js:112:23)
      at Page.waitForSelector (../../node_modules/puppeteer/lib/Page.js:1089:29)
      at Object.waitForSelector (specs/stories-editor/page.js:273:14)

  ● Story Page › should not affect story settings when individual page settings are changed

    Evaluation failed: TypeError: Cannot read property 'value' of null

      at ../../__puppeteer_evaluation_script__:2:44
      at ExecutionContext._evaluateInternal (../../node_modules/puppeteer/lib/ExecutionContext.js:122:13)
        -- ASYNC --
      at ExecutionContext.<anonymous> (../../node_modules/puppeteer/lib/helper.js:111:15)
      at DOMWorld.evaluate (../../node_modules/puppeteer/lib/DOMWorld.js:112:20)
        -- ASYNC --
      at Frame.<anonymous> (../../node_modules/puppeteer/lib/helper.js:111:15)
      at Page.evaluate (../../node_modules/puppeteer/lib/Page.js:827:43)
      at Page.<anonymous> (../../node_modules/puppeteer/lib/helper.js:112:23)
      at evaluate (specs/stories-editor/page.js:40:27)
      at Object.getInputValue (specs/stories-editor/page.js:313:17)

  ● Story Page › should not affect existing stories when story defaults are changed

    TimeoutError: waiting for selector ".amp-story-settings-advance-after select" failed: timeout 30000ms exceeded

      319 | 
      320 |             await expandSettingsPanel();
    > 321 |             await page.waitForSelector( SETTINGS_ADVANCE_AFTER );
          |                        ^
      322 | 
      323 |             const originalStoryAdvancement = await getInputValue( SETTINGS_ADVANCE_AFTER );
      324 |             const originalStoryAdvancementDuration = await getInputValue( SETTINGS_ADVANCE_AFTER_DURATION );

      at new WaitTask (../../node_modules/puppeteer/lib/DOMWorld.js:561:28)
      at DOMWorld._waitForSelectorOrXPath (../../node_modules/puppeteer/lib/DOMWorld.js:490:22)
      at DOMWorld.waitForSelector (../../node_modules/puppeteer/lib/DOMWorld.js:444:17)
      at Frame.waitForSelector (../../node_modules/puppeteer/lib/FrameManager.js:628:47)
      at Frame.<anonymous> (../../node_modules/puppeteer/lib/helper.js:112:23)
      at Page.waitForSelector (../../node_modules/puppeteer/lib/Page.js:1089:29)
      at Object.waitForSelector (specs/stories-editor/page.js:321:14)
```

spacedmonkey:
@googlebot I consent

swissspidy:
Regarding 5897656e4fd663c900f50c724f991eea76a6c823:

[`toMatchElement`](https://github.com/smooth-code/jest-puppeteer/tree/09703eaeeeab553e13142153b55030db05611f7c/packages/expect-puppeteer#expectinstancetomatchelementselector-options) by default waits for 30 seconds for the expected element, so this `waitForSelector` call is redundant.

So if `toMatchElement` is failing here after the 30 second wait, there probably are some reusable blocks listed on this page that haven't been cleaned up.

swissspidy:
The PHPUnit failure is unrelated and not a blocker, see https://github.com/ampproject/amp-wp/pull/3243#issuecomment-531652972

swissspidy:
The PHPUnit failure is unrelated and not a blocker, see https://github.com/ampproject/amp-wp/pull/3243#issuecomment-531652972

swissspidy:
FYI I created #3247 to investigate the flaky e2e tests

PR #3205, opened 2019-09-06
Add accessibility label for video blocks

### LINKED ISSUE #3186: Add accessibility label for video blocks
_Spawned from #2983_

As a user I want to be able to see and edit the accessibility label for video elements directly in the block editor for a given video placed in a story.

The value should be pre-populated with the value from the `title` attribute from the media editor.

The field should be a simple single-line text field placed in the _Video Settings_ section of the block element editor with a descriptive label.

The value should be saved with the individual page with the given media element and not on the media element itself. If multiple pages/stories use the same media element, they can have different accessibility labels.

The field should be rendered as the `aria-label` attribute on the `<amp-video>` node  as part of the AMP story output.

barklund:
This fixes #3186 by adding a new attribute to AMP video blocks, `ampAriaLabel`, and giving the editor access to edit this variable in the "Video Settings" inspector panel.

Furthermore this attribute is pre-populated from the media object selected by looking at `alt_text` or `title` of existing media object, if either exists. This works for both existing media elements and newly created ones.

Finally, this value is used in the output on the AMP story by manipulating the output of core video blocks and attaching the `aria-label` attribute to the `<amp-video />` element if applicable.

### Missing elements
- [ ] Add unit tests for `addVideoAriaLabel`.
- [ ] Add e2e tests for the whole process.


assets/src/stories-editor/components/custom-video-block-edit.js
```diff
@@ -266,6 +277,12 @@ class CustomVideoBlockEdit extends Component {
 							onChange={ this.toggleAttribute( 'loop' ) }
 							checked={ loop }
 						/>
+						<TextControl
+							label={ __( 'Accessibility label', 'amp' ) }
```
swissspidy:
```suggestion
							label={ __( 'Accessibility Label', 'amp' ) }
```

To match `Video Settings` and other labels here. `Assistive Text` could work too.

swissspidy: Updated the wording now to align with #3214.

swissspidy:
For the e2e test you could draw inspiration from #3214.

spacedmonkey:
@barklund How does this ticket relate to https://github.com/ampproject/amp-wp/pull/3214 . They seem very similar.

swissspidy:
One is for the page background media, this is for the video block. Split up on purpose.

swissspidy:
@barklund Added an e2e test, but it is skipped like all the other video-related tests at the moment. The reason is that Chromium (used by Puppeteer) does not support playing MP4 videos (see links in the docblock for details).

PR #3206, opened 2019-09-07
Add contextual awareness to Tag & Attribute sanitizer

### LINKED ISSUE #3137: AMP_Allowed_Tags_Generated class is stripping amp-date-display {{iso}} format from amp-timeago component
When trying to convert a timestamp to an iso datetime format using an amp-date-display component with amp-mustache template the {{iso}} format is being stripped from the amp-timeago component by the datetime regex value in the **AMP_Allowed_Tags_Generated** class on line **5417**

It would be good if the below could be allowed to work as it would give a degree of flexibility with regards to the type of datetime we can pass to amp-timeago (sometimes it's not possible to change the format of the date being consumed via an external api call etc)

```html
<amp-date-display timestamp-seconds="1567062005" layout="fixed" width="360" height="20">
        <amp-timeago class="m1"
        width="160"
        height="20"
        datetime="{{iso}}">&nbsp;</amp-timeago>
    </template>
</amp-date-display>
```

schlessera:
This PR changes the way nodes are processed in the Tag & Attribute sanitizer from breadth-first to depth-first. This is mainly achieved by replacing the stack additions with an array_splice to replace the currently processed node.

It also adds a mechanism to track contextual information ("Are we currently within a template tag?") to make the processing aware of such context, and add the ability to conditionally act based on such context.

Change breakdown:

* Adds `AMP_Context` class as a container to track the current context.
* Adds `AMP_Contextual_Node` class as stack marker to track enter, toggle and leave operations for such context.
* Switches node traversal from breadth-first to depth-first.
* Wraps node children in contextual nodes if context information is needed.
* Adds first contextual knowledge about whether we are within a template tag or not.
* Disables attribute value sanitization when within a template tag and a template placeholder is detected.

Fixes #3137


schlessera:
As we currently don't have a single consistent iteration over the DOM for all sanitizers combined, we cannot easily share the contextual information across sanitizers.

For solving something like a `<template>` tag within a `<template>` tag (which seems to be invalid for AMP, as there's no unambiguous way of knowing what template tag is handling which placeholders), we should either wait for the v2 redesign, or extract the context handling into a trait that can be shared amongst sanitizers to deduplicate the code at least.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -258,29 +277,125 @@ public function sanitize() {
...
+		if ( ! $node->hasAttributes() ) {
+			return false;
+		}
+		$type = $node->attributes->getNamedItem( 'type' );
+		if ( null === $type ) {
+			return false;
+		}
+
+		// The type attribute of a template tag defines what templating engine to use.
+		switch ( $type->nodeValue ) {
```
schlessera: This `switch` only has one case so far, but this properly communicates the intent that there are multiple possible values we need to consider here. Turning it into an `if` for the short-term would obfuscate this intent and probably lead to problems later on when extending.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -258,29 +277,125 @@ public function sanitize() {
...
+	 * keep track of said context. This is done because we linearize the stack
+	 * and lack hierarchical information when processing the nodes. So we turn
+	 * one parent node into a pair of starting and ending contextual nodes.
+	 *
+	 * @param DOMNode $node Node to examine.
+	 * @return DOMNode[]|false Pair of contextual nodes, or false if none to be
+	 *                         used.
+	 */
+	protected function needs_contextual_nodes( DOMNode $node ) {
+		switch ( $node->nodeName ) {
```
schlessera: This `switch` only has one case so far, but this properly communicates the intent that there are multiple possible values we need to consider here. Turning it into an `if` for the short-term would obfuscate this intent and probably lead to problems later on when extending.

westonruter:
Suggesting we close this PR in favor of #3243.

PR #3208, opened 2019-09-07
Remove deprecated AMP_WP_Utils class

### LINKED ISSUE #2872: Remove deprecated code
As a follow-up to #876, where `AMP_WP_Utils` was deprecated, I think it is now time to fully remove that code.

A quick search on https://wpdirectory.net/ revealed exactly **0** plugins and themes that use any of these utils. There are only plugins (e.g. AMP for WP) that also have this class, but that's because they include this plugin here as a Composer dependency.

We could remove the code in the next bigger release, e.g. 1.3.0 or 2.0.0 the latest.

Similarly, we could remove other code that was deprecated in < 1.0, e.g. `amp_load_classes()`

schlessera:
* Removes files `includes/utils/class-amp-wp-utils.php`
* Removes reference to `AMP_WP_Utils` in file `includes/class-amp-autoloader.php`

I couldn't find any tests for it and a string search for `AMP_WP_Utils` didn't reveal any other occurrences.

Fixes #2872


kienstra:
Sorry for missing the review here. It's good to have the plugin cleaner 😄

PR #3209, opened 2019-09-09
Add special-case handling for layout=fill interpretation

### LINKED ISSUE #3142: Fallback height overriding inline style height on iframes
### Description

The following sort of `iframe`:
```
<iframe 
src="{SRC}" 
style="position:absolute;top:0;left:0;width:100%;height:100%;border:1px solid #aaa;border-bottom:0;box-sizing:border-box;"
></iframe>
```
gets converted to the following `amp-iframe`:
```
<amp-iframe 
src="{SRC}" 
height="400" 
layout="fixed-height" class="{...}" i-amphtml-layout="fixed-height" style="height: 400px;">{...}</amp-iframe>
```

Since the height is done through an inline style and not the actual `height` attribute, the `AMP_Base_Sanitizer` assigns it a height of `FALLBACK_HEIGHT` (400). This can potentially lead to weird layout issues, like the bunch of extra padding seen here:

Screen Shot 2019-08-28 at 12 09 34 PM

### To reproduce:

1. Install and activate [DocumentCloud plugin](https://wordpress.org/plugins/documentcloud/).
2. Add the following shortcode to a post:
```
[documentcloud url="http://www.documentcloud.org/documents/3117065-Obama-Library-South-Side-Community-Benefits.html"]
````
3. View post on frontend in AMP mode.

### Recommended solution:

I realize that this is somewhat like expecting you to handle other people using things in weird ways, but I think the best solution would be in `AMP_Base_Sanitizer::set_layout` to follow this priority:
1.  Check and use `height`/`width` attributes if possible.
2. If not found, check and use `height`/`width` inline styles if possible.
3. If none of those are found, set `FALLBACK_HEIGHT`.

### PR DESCRIPTION
**Pull request by schlessera**
## Changes
This PR adds the following to the `AMP_Base_Sanitizer` class:
* `protected function parse_style_string( $style_string )` to parse the value of a `style` attribute into an associative array of `property => value` entries.
* `protected function reassemble_style_string( $styles )` to reassemble an associative array of `property => value` entries into a string that can be used as a `style` attribute.
* Special-case handling in `set_layout()` method to cover the following cases:
	* `position:absolute;top:0;left:0;bottom:0;right:0` => `layout=fill`
	* `position:absolute;top:0;left:0;width:100%;height:100%` => `layout=fill`
* Tests for the above.

## Before

Image 2019-09-09 at 11 18 01 AM

## After

Image 2019-09-09 at 11 17 23 AM

Fixes #3142

### DIFF
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -280,10 +280,10 @@ abstract class AMP_Base_Sanitizer {
 			// top, left, bottom, right are used.
 			if ( isset( $styles['position'], $styles['top'], $styles['left'], $styles['bottom'], $styles['right'] )
 				&& 'absolute' === $styles['position']
-				&& '0' === (string) $styles['top']
-				&& '0' === (string) $styles['left']
-				&& '0' === (string) $styles['bottom']
-				&& '0' === (string) $styles['right']
+				&& 0 === (int) $styles['top']
+				&& 0 === (int) $styles['left']
+				&& 0 === (int) $styles['bottom']
+				&& 0 === (int) $styles['right']
 			) {
 				unset( $attributes['style'], $styles['position'], $styles['top'], $styles['left'], $styles['bottom'], $styles['right'] );
 				if ( ! empty( $styles ) ) {
@@ -296,8 +296,8 @@ abstract class AMP_Base_Sanitizer {
 			// top, left, width, height are used.
 			if ( isset( $styles['position'], $styles['top'], $styles['left'], $styles['width'], $styles['height'] )
 				&& 'absolute' === $styles['position']
-				&& '0' === (string) $styles['top']
-				&& '0' === (string) $styles['left']
+				&& 0 === (int) $styles['top']
+				&& 0 === (int) $styles['left']
 				&& '100%' === (string) $styles['width']
 				&& '100%' === (string) $styles['height']
 			) {
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -704,7 +704,7 @@ abstract class AMP_Base_Sanitizer {
 	 * Parse a style string into an associative array of style attributes.
 	 *
 	 * @param string $style_string Style string to parse.
-	 * @return array Associative array of style attributes.
+	 * @return string[] Associative array of style attributes.
 	 */
 	protected function parse_style_string( $style_string ) {
 		// We need to turn the style string into an associative array of styles first.
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -277,7 +277,7 @@ abstract class AMP_Base_Sanitizer {
 		if ( ! empty( $attributes['style'] ) ) {
 			$styles = $this->parse_style_string( $attributes['style'] );
 
-			// top, left, bottom, right are used.
+			// Apply fill layout if top, left, bottom, right are used.
 			if ( isset( $styles['position'], $styles['top'], $styles['left'], $styles['bottom'], $styles['right'] )
 				&& 'absolute' === $styles['position']
 				&& 0 === (int) $styles['top']
@@ -293,7 +293,7 @@ abstract class AMP_Base_Sanitizer {
 				return $attributes;
 			}
 
-			// top, left, width, height are used.
+			// Apply fill layout if top, left, width, height are used.
 			if ( isset( $styles['position'], $styles['top'], $styles['left'], $styles['width'], $styles['height'] )
 				&& 'absolute' === $styles['position']
 				&& 0 === (int) $styles['top']
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -717,6 +717,8 @@ abstract class AMP_Base_Sanitizer {
 		}
 
 		$chunks = array_chunk( $elements, 2 );
+
+		// phpcs:ignore PHPCompatibility.FunctionUse.NewFunctions.array_columnFound -- WP Core provides a polyfill.
 		return array_combine( array_column( $chunks, 0 ), array_column( $chunks, 1 ) );
 	}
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by schlessera** [comment]
The `parse_style_string( $style_string )` method currently uses a custom algorithm to do this operation. It would probably be more robust to reuse the Sabberworm CSS parser for this, but it doesn't seem to have proper built-in support for inline styles, so we'd probably have to fake a selector or something along these lines.

@westonruter Thoughts about current (hacky) custom algorithm vs (hacky) PHP CSS Parser reuse?

File: includes/sanitizers/class-amp-base-sanitizer.php [hunk 1]
```diff
@@ -257,21 +257,59 @@ public function sanitize_dimension( $value, $dimension ) {
...
+		if ( ! empty( $attributes['style'] ) ) {
+			$styles = $this->parse_style_string( $attributes['style'] );
+
+			// top, left, bottom, right are used.
+			if ( isset( $styles['position'], $styles['top'], $styles['left'], $styles['bottom'], $styles['right'] )
+				&& 'absolute' === $styles['position']
+				&& '0' === (string) $styles['top']
+				&& '0' === (string) $styles['left']
+				&& '0' === (string) $styles['bottom']
+				&& '0' === (string) $styles['right']
```
westonruter: Perhaps it would be good to `intval()` for each of these instead? Sometimes people do `0px` even though they shouldn't. And `0 === intval( '0px' )`

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -257,21 +257,59 @@ public function sanitize_dimension( $value, $dimension ) {
...
+				}
+				$attributes['layout'] = 'fill';
+				return $attributes;
+			}
+
+			// top, left, width, height are used.
+			if ( isset( $styles['position'], $styles['top'], $styles['left'], $styles['width'], $styles['height'] )
+				&& 'absolute' === $styles['position']
+				&& '0' === (string) $styles['top']
+				&& '0' === (string) $styles['left']
```
westonruter: See note above about using `intval()` instead.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -257,21 +257,59 @@ public function sanitize_dimension( $value, $dimension ) {
...
 	public function set_layout( $attributes ) {
 		if ( isset( $attributes['layout'] ) && ( 'fill' === $attributes['layout'] || 'flex-item' !== $attributes['layout'] ) ) {
 			return $attributes;
 		}
+
+		// Special-case handling for inline style that should be transformed into layout=fill.
+		if ( ! empty( $attributes['style'] ) ) {
+			$styles = $this->parse_style_string( $attributes['style'] );
+
+			// top, left, bottom, right are used.
```
westonruter:
```suggestion
			// Apply fill layout if top, left, bottom, right are used.
```

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -661,4 +699,53 @@ public function maybe_add_amp_image_lightbox_node() {
 		);
 		$body_node->appendChild( $amp_image_lightbox );
 	}
+
+	/**
+	 * Parse a style string into an associative array of style attributes.
+	 *
+	 * @param string $style_string Style string to parse.
+	 * @return array Associative array of style attributes.
```
westonruter:
```suggestion
	 * @return string[] Associative array of style attributes.
```

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -661,4 +699,53 @@ public function maybe_add_amp_image_lightbox_node() {
...
+	 */
+	protected function parse_style_string( $style_string ) {
```
westonruter: Beautiful function. 😍

includes/sanitizers/class-amp-base-sanitizer.php
schlessera: Yes, makes sense. I'll use a cast `(int)` instead, though. `intval()` is more of a remnant from PHP 4 when actual type-casting was not yet part of the language.

includes/sanitizers/class-amp-base-sanitizer.php
schlessera: I don't trust it and am not sure this is a good way to go about. I would prefer to use actual parsing like with PHP-CSS-Parser, but that seems to only work through a hack, too.

schlessera: I don't think it is actually a problem here, as we're only looking for 2 very specific patterns, and not using malformed CSS in any other way anyway, so we're good for now. But the function shouldn't be used for anything more involved than that.

includes/sanitizers/class-amp-base-sanitizer.php
schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3209/commits/47f0f331655388c57c45c5acafc097d07330383c

includes/sanitizers/class-amp-base-sanitizer.php
schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3209/commits/94156e5c02bf5a2d6764f0ef3853348e62747844

includes/sanitizers/class-amp-base-sanitizer.php
schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3209/commits/94156e5c02bf5a2d6764f0ef3853348e62747844

includes/sanitizers/class-amp-base-sanitizer.php
westonruter: Yes, better to not use the PHP-CSS-Parser because it is too expensive for this runtime operation.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -661,4 +699,53 @@ public function maybe_add_amp_image_lightbox_node() {
...
+
+		if ( 0 !== count( $elements ) % 2 ) {
+			// Style string was malformed, try to process as good as possible by stripping the last element.
+			array_pop( $elements );
+		}
+
+		$chunks = array_chunk( $elements, 2 );
+
+		// phpcs:ignore PHPCompatibility.FunctionUse.NewFunctions.array_columnFound -- WP Core provides a polyfill.
+		return array_combine( array_column( $chunks, 0 ), array_column( $chunks, 1 ) );
```
westonruter: The polyfill is apparently not being loaded because this is causing a test error in PHP 5.4

schlessera: Ha, it was WP-CLI that has the polyfill, not WordPress Core. We have 3 options here:
1.) Rewrite the algorithm to not require `array_column()`.
2.) Pull in `ramsey/array_column` as a dependency.
3.) Copy the code in https://github.com/ramsey/array_column/blob/master/src/array_column.php into our helper functions file.

@westonruter Thoughts?

westonruter: Since we plan to upgrade to PHP 5.6 soon, feel free to do whatever is fastest as a short term solution. I suppose that means copying the code into the helper functions file.

schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3209/commits/8794360b3562cad6b6e6887311320b4edc7c08f0

westonruter:
One more follow-up was needed for this: #3295.

It turns out that such fill elements also sometimes redundantly set `width=100% height=100%`: #3277.

PR #2017, opened 2019-03-22
Prevent tree-shaking class names for amp-geo when component on page

westonruter:
Conditional tree shaking for dynamic class names was added in #1959, but it missed handling class names added for `amp-geo`. See [support topic](https://wordpress.org/support/topic/filter-for-tree-shaking/).

This PR will prevent tree shaking the classes when the `amp-geo` component is used on the page:

* `amp-geo-pending`
* `amp-geo-no-group`
* `amp-geo-group-`…
* `amp-iso-country-`…

See #1490.

Build of plugin for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/2998238/amp.zip) (v1.1-alpha-20190322T203831Z-f45def68)

PR #1959, opened 2019-03-13
Fix tree shaking for dynamically-added class names

westonruter:
Prevent tree shaking CSS rules that use class names which are dynamically added, including those for:

* [`amp-dynamic-css-classes`](https://www.ampproject.org/docs/reference/components/amp-dynamic-css-classes) (naturally)
* [`amp-carousel`](https://www.ampproject.org/docs/reference/components/amp-carousel#styling)
* [`amp-form`](https://www.ampproject.org/docs/reference/components/amp-form#classes-and-css-hooks)
* [`amp-user-notification`](https://www.ampproject.org/docs/reference/components/amp-user-notification#styling)
* [`amp-access`](https://www.ampproject.org/docs/reference/components/amp-access)/[`amp-access-laterpay`](https://www.ampproject.org/docs/reference/components/amp-access-laterpay#styling)
* [`amp-image-lightbox`](https://www.ampproject.org/docs/reference/components/amp-image-lightbox#styling)
* [`amp-live-list`](https://www.ampproject.org/docs/reference/components/amp-live-list#styling)
* [`amp-sidebar`](https://www.ampproject.org/docs/reference/components/amp-sidebar#styling-toolbar)
* [`amp-sticky-ad`](https://www.ampproject.org/docs/reference/components/amp-sticky-ad#styling)
* [`amp-video-docking`](https://github.com/ampproject/amphtml/blob/master/extensions/amp-video-docking/amp-video-docking.md#styling)

Fixes #1946.
Fixes #1490.

----

Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/2962979/amp.zip) (1.1-alpha-20190313T180459Z-d5495203)

PR #3212, opened 2019-09-09
Prevent tree-shaking CSS selectors for dynamic attributes

westonruter:
This fixes an issue [reported](https://wordpress.org/support/topic/plugin-is-stripping-css-attribute-selector/) on the support forum:

> I’m only trying to style an expanded menu inside of a sidebar. I’m using standard AMP elements, including amp-accordion. The inline style that gets added by the plugin is automatically removing the styles that would allow this, for example:
> 
> ```css
> section.menu-section[expanded] {
> background: green;
> }
> ```
>
> The plugin isn’t stripping out any other content. 

In #1959 (and #2017, #2642, #2710) tree shaking for CSS selectors was made aware of the class names for AMP components which can be dynamically added/removed from a page. Even though the selector with such a class name may not exist at the initial time of rendering the page, the selector may very well apply later. The same logic needed to be extended to attributes which are added/removed to AMP components. In particular:

* `amp-sidebar`: The `open` attribute is dynamically added when the sidebar is opened.
* `amp-accordion`: The `expanded` attribute is added when a section  is opened by the user.
* `amp-live-list`: The `data-tombstone` attribute can appear in an `amp-live-list`'s contents during the life of a page.

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -188,6 +188,12 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 * This is initially populated with boolean attributes which can be mutated by AMP at runtime,
 	 * since they can by dynamically added at any time.
 	 *
+	 * @todo With the exception of 'hidden' (which can be on any element), the values here could be removed in favor of
+	 *       checking to see if any of the related elements exist in the page in `\AMP_Style_Sanitizer::has_used_attributes()`.
+	 *       Nevertheless, selectors mentioning these attributes are very numerous, so tree-shaking improvements will be marginal.
+	 *
+	 * @see \AMP_Style_Sanitizer::has_used_attributes()
+	 *
 	 * @since 1.1
 	 * @var array
 	 */
@@ -199,7 +205,6 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 		'hidden'    => true,
 		'loop'      => true,
 		'multiple'  => true,
-		'open'      => true,
 		'readonly'  => true,
 		'required'  => true,
 		'selected'  => true,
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -667,6 +672,32 @@ private function has_used_attributes( $attribute_names ) {
 
 				$this->used_attributes[ $attribute_name ] = ( 0 !== $this->xpath->query( $expression )->length );
 			}
+
+			// Attributes for amp-accordion, see <https://amp.dev/documentation/components/amp-accordion/#styling>.
```
schlessera: We'll probably have more coming later on, so it might be better to write this as a mapping array with generic logic:
```php
	private function has_used_attributes( $attribute_names ) {
		$dynamically_added_attributes = [
			// Attributes for amp-accordion, see <https://amp.dev/documentation/components/amp-accordion/#styling>.
			'expanded'       => [ 'amp-accordion' ],
			// Attributes for amp-sidebar, see <https://amp.dev/documentation/components/amp-sidebar/#styling>.
			// The 'open' attribute is also used by the HTML5 <details> attribute.
			'open'           => [ 'amp-sidebar', 'details' ],
			// Attributes for amp-live-list, see <https://amp.dev/documentation/components/amp-live-list/#styling>.
			'data-tombstone' => [ 'amp-live-list' ],
		];

		foreach ( $attribute_names as $attribute_name ) {

			// [...]

			foreach ( $dynamically_added_attributes as $attribute => $tags ) {
				if ( $attribute !== $attribute_name ) {
					continue;
				}

				foreach ( $tags as $tag ) {
					if ( $this->has_used_tag_names( [ $tag ] ) ) {
						// This is a dynamically added attribute, so skip removal.
						continue 3;
					}
				}

				return false;
			}

			// [...]
		}
		return true;
	}
```
The list with the mappings can even be extracted into a place that is more manageable as needed.

PR #3214, opened 2019-09-09
Add text control for background media alt text

### LINKED ISSUE #3185: Add direct access to accessibility label for background elements
_Spawned from #2983_

As a user I want to be able to see and edit the accessibility label for background elements directly in the block editor for a given story page.

The value should be prepopulated from the media editor:
- For background images, the field should be pre-populated with the value from the `alt` attribute from the media editor.
- For background videos, the field should be pre-populated with the value from the `title` attribute from the media editor.

The field should be a simple single-line text field placed near the _Change Media_ button for the background block element in the _Background Media_ section with a descriptive label.

The value should be saved with the individual page with the given media element and not on the media element itself. If multiple pages/stories use the same media element, they can have different accessibility labels.

The value should be rendered as part of the AMP story output as follows:
- For background images, the field should be rendered as the `alt` attribute on the `<amp-img>` node.
- For background images, the field should be rendered as the `aria-label` attribute on the `<amp-video>` node.

swissspidy:
Fixes #3185.

Screenshot:

Screenshot 2019-09-09 at 23 22 58


swissspidy:
@barklund Let's align on a wording for the label that we can use both in #3205 and here.

swissspidy:
Reduced margin now:

Screenshot 2019-09-11 at 11 24 49

swissspidy:
@spacedmonkey Mind checking again now?

PR #3216, opened 2019-09-10
Sort the list of active plugins

### LINKED ISSUE #3172: Normalize the validated environment
The code in `AMP_Validated_URL_Post_Type::get_validated_environment()` uses the `'active_plugins'` option as-is.

As there are plugins that actively manipulate the ordering of this option to put themselves in first place, and as this might lead to situations where multiple plugins go back-and-forth and constantly alternate in positions, we should normalize the array of `'active_plugins'` first before using it, to avoid unnecessary cache invalidations and make the mechanism less brittle.

NOTE: a different order of plugins might technically result in a different output, so we might not invalidate when this would actually be needed. However, chances are very low that this is the case, whereas I've seen conflicts in "who goes first" on multiple sites already.

schlessera:
As there are plugins that actively manipulate the ordering of this option to put themselves in first place, and as this might lead to situations where multiple plugins go back-and-forth and constantly alternate in positions, we normalize the array of 'active_plugins' by sorting it, to avoid unnecessary cache invalidations and make the mechanism less brittle.

Fixes #3172

### DIFF
--- a/includes/validation/class-amp-validated-url-post-type.php
+++ b/includes/validation/class-amp-validated-url-post-type.php
@@ -797,9 +797,14 @@ class AMP_Validated_URL_Post_Type {
 	 * @return array Environment.
 	 */
 	public static function get_validated_environment() {
+		// We want to sort the list of plugins to avoid fluctuations due to plugins fighting for first spot
+		// to constantly invalidate our cache.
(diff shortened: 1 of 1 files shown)


westonruter:
Where's the demo video? 😉

PR #3218, opened 2019-09-10
Make sure the value passed to defaultvalue is never null.

### LINKED ISSUE #3027: Font Picker: Use searchable dropdown

spacedmonkey:
Fixes #3027

### DIFF
--- a/assets/src/stories-editor/components/font-family-picker/autocomplete.js
+++ b/assets/src/stories-editor/components/font-family-picker/autocomplete.js
@@ -32,7 +32,7 @@ class Autocomplete extends OriginalAutocomplete {
 	 * Override render method, to add clear font button.
 	 *
 	 */
-	render() {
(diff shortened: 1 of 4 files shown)


swissspidy:
@spacedmonkey The `expect( textBlockBefore.attributes.ampFontFamily ).toBeNull();` assertion now fails in the e2e tests

swissspidy:
This fixes the editor crash reported in https://github.com/ampproject/amp-wp/issues/3027#issuecomment-529845105 for me 👍

Issue #3217, opened 2019-09-10
Missing escaping in AMP settings screen code


schlessera:
The AMP settings screen code in `AMP_Options_Manager` has a few instances of unescaped output:
* The translated strings in the warning messages that contain links are unescaped. They should be passed through `wp_kses( $translation, [ 'a' => [ 'href' => true ] ] )`.
* The settings errors are stored without sanitization in the database and are then displayed without escaping in the admin backend. They should be passed through `esc_html()` before storing them in the database.

PR #3219, opened 2019-09-10
Add warning notice to AMP settings screen when not using HTTPS

### LINKED ISSUE #2006: Add warning when a site is not served via HTTPS
Many AMP components require HTTPS to function. When a page is not served as HTTPS, an AMP component can just fail to work (e.g. forms) leaving the user in confusion. See https://github.com/Automattic/jetpack/issues/11595. While we originally thought in https://github.com/ampproject/amp-wp/issues/969 to just handle the warning in the PWA plugin, since it is so important to serve as HTTPS we should show a warning on the AMP settings screen similar to how we show a warning when not using an object caching plugin.

This would be unnecessary if core starts to show a non-HTTPS warning as part of Health Check before we get around to it.

schlessera:
This PR adds a notice to the AMP Settings screen to warn users when their site is not being served over HTTPS.

To check this, it doesn't solely rely on `is_ssl()` but also checks the `url` and `wpurl` options, similar to what Site Health does, to make sure we get a value for the frontend as well.

The warning links to the official HTTPS support page from WP: [https://wordpress.org/support/article/why-should-i-use-https/](https://wordpress.org/support/article/why-should-i-use-https/)

Image 2019-09-10 at 12 01 53 PM

Fixes #2006

### DIFF
--- a/includes/options/class-amp-options-manager.php
+++ b/includes/options/class-amp-options-manager.php
@@ -637,9 +637,9 @@ class AMP_Options_Manager {
 		$uses_ssl = (
 			is_ssl()
 			&&
-			( strpos( get_bloginfo( 'wpurl' ), 'https' ) !== 0 )
+			( strpos( get_bloginfo( 'wpurl' ), 'https' ) === 0 )
 			&&
-			( strpos( get_bloginfo( 'url' ), 'https' ) !== 0 )
+			( strpos( get_bloginfo( 'url' ), 'https' ) === 0 )
 		);
 
 		if ( ! $uses_ssl && 'toplevel_page_' . self::OPTION_NAME === get_current_screen()->id ) {
--- a/includes/options/class-amp-options-manager.php
+++ b/includes/options/class-amp-options-manager.php
@@ -645,15 +645,16 @@ class AMP_Options_Manager {
 		if ( ! $uses_ssl && 'toplevel_page_' . self::OPTION_NAME === get_current_screen()->id ) {
 			printf(
 				'<div class="notice notice-warning"><p>%s</p></div>',
-				sprintf(
-					/* translators: %s: "Why should I use HTTPS" support URL */
-					wp_kses(
-						__( 'Your site is not being fully served over a secure connection (using HTTPS).<br>' .
-						    'As some AMP functionality requires a secure connection, you might experience degraded performance or broken components.<br>' .
-						    '<a href="%s">More details</a>', 'amp' ),
-						[ 'br' => [], 'a' => [ 'href' => true ] ]
+				wp_kses(
+					sprintf(
+						/* translators: %s: "Why should I use HTTPS" support URL */
+						__( 'Your site is not being fully served over a secure connection (using HTTPS).<br>As some AMP functionality requires a secure connection, you might experience degraded performance or broken components.<br><a href="%s">More details</a>', 'amp' ),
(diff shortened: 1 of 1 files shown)


schlessera:
While writing this PR I noticed that there is some escaping missing in this class and opened a new issue for this: #3217

includes/options/class-amp-options-manager.php
```diff
@@ -623,6 +624,41 @@ public static function show_response_cache_disabled_notice() {
...
+			( strpos( get_bloginfo( 'url' ), 'https' ) !== 0 )
+		);
+
+		if ( ! $uses_ssl && 'toplevel_page_' . self::OPTION_NAME === get_current_screen()->id ) {
+			printf(
+				'<div class="notice notice-warning"><p>%s</p></div>',
+				sprintf(
+					/* translators: %s: "Why should I use HTTPS" support URL */
+					wp_kses(
+						__( 'Your site is not being fully served over a secure connection (using HTTPS).<br>' .
```
swissspidy: String concatenation doesn't work for translatable strings. Given that there are line breaks, perhaps there should be three `__()` calls here.

includes/options/class-amp-options-manager.php
```diff
@@ -623,6 +624,41 @@ public static function show_response_cache_disabled_notice() {
...
+	 * @since 1.3
+	 *
+	 * @return void
+	 */
+	public static function insecure_connection_notice() {
+		// is_ssl() only tells us whether the admin backend uses HTTPS here, so we add a few more sanity checks.
+		$uses_ssl = (
+			is_ssl()
+			&&
+			( strpos( get_bloginfo( 'wpurl' ), 'https' ) !== 0 )
```
swissspidy: Shouldn't this be `===`? 🤔

schlessera: Oh, yes, indeed. Good catch, that's what I get for copying from the site health code... :)

includes/options/class-amp-options-manager.php
schlessera: Erm, I already changed that... I think ... 🤔

schlessera: Ah, I changed it and then went on to add escaping to the other calls as well, but finally decided against doing that in the same PR and scrapped the changes.

includes/options/class-amp-options-manager.php
schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3219/commits/243320364c10ea0e0e79af0e6a36900a15a6f7ea

includes/options/class-amp-options-manager.php
schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3219/commits/d842fa8083805504c12bcab01524e6e35261ef56

includes/options/class-amp-options-manager.php
westonruter: TIL: `get_bloginfo( 'wpurl' )` and `get_bloginfo( 'url' )` are wrappers around `site_url()` and `home_url()` respectively (which are themselves wrappers around `get_option()`, 🐢🐢🐢).

Maybe it would be preferable to do `'https' !=== wp_parse_url( get_bloginfo( 'wpurl' ), PHP_URL_SCHEME )` here and below? A tiny bit more robust perhaps.

PR #3222, opened 2019-09-10
Use a transient to cache the potentially slow "Pending URLs Count" query

### LINKED ISSUE #2736: Cache query result in add_admin_menu_new_invalid_url_count
Query Monitor just warned me about a slow query in WordPress admin coming from `AMP_Validated_URL_Post_Type::add_admin_menu_new_invalid_url_count()`:

https://github.com/ampproject/amp-wp/blob/13114be415be0a7096e97b577482cf3ea4602ae7/includes/validation/class-amp-validated-url-post-type.php#L370-L380

I think that post count result can easily be cached in a transient and updated/flushed whenever validation happens.

schlessera:
Extracts the pending URL count query into a separate method and then wraps usage of that method into a transient.

The key name is `amp_pending_urls_count` and the expiry time is set to 24h.

Fixes #2736

### DIFF
--- a/includes/validation/class-amp-validated-url-post-type.php
+++ b/includes/validation/class-amp-validated-url-post-type.php
@@ -83,11 +83,11 @@ class AMP_Validated_URL_Post_Type {
 	const VALIDATION_ERRORS_META_BOX = 'amp_validation_errors';
 
 	/**
-	 * The transient key to use for caching the pending URLs count.
+	 * The transient key to use for caching the number of URLs with new validation errors.
 	 *
 	 * @var string
 	 */
-	const PENDING_URLS_COUNT_TRANSIENT = 'amp_pending_urls_count';
+	const NEW_VALIDATION_ERROR_URLS_COUNT_TRANSIENT = 'amp_new_validation_error_urls_count';
 
 	/**
 	 * The total number of errors associated with a URL, regardless of the maximum that can display.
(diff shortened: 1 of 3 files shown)


includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -82,6 +82,13 @@ class AMP_Validated_URL_Post_Type {
 	 */
 	const VALIDATION_ERRORS_META_BOX = 'amp_validation_errors';
 
+	/**
+	 * The transient key to use for caching the pending URLs count.
```
westonruter: Instead of "pending" it would be more accurate to use "new" for the changes here. The URLs being queried are those with _new_ accepted/rejected validation errors:

```
					AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_NEW_REJECTED_STATUS,
					AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_NEW_ACCEPTED_STATUS,
```

As opposed to _ack_ rejected/accepted errors.

Using `new` instead of `pending` would be more consistent in regards to terminology.

But given the  resulting containing element  is `pending-count`, I don't feel strongly about this.

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -788,6 +814,8 @@ public static function store_validation_errors( $validation_errors, $url, $args
 			update_post_meta( $post_id, '_amp_queried_object', $args['queried_object'] );
 		}
 
+		delete_transient( static::PENDING_URLS_COUNT_TRANSIENT );
```
westonruter: I believe this is not the only place where this is needed. In particular, anywhere that `wp_update_term()` is called for the `amp_validation_error` taxonomy (like `\AMP_Validation_Error_Taxonomy::handle_validation_error_update()`) should also flush this transient, right?

schlessera: Ah, good point, I was only looking for changes in the post type, and thought this central method would cover all cases. But the query is indeed relying on taxonomies as well.

includes/validation/class-amp-validated-url-post-type.php
westonruter: The `pending-count` class name is only appearing here because the Comments admin menu item also uses it, and the markup for it was copied:

image

I thought that `pending-count` had some styles associated with it, but this is not the case. Feel free to rename the class name as you see fit as well.

includes/validation/class-amp-validated-url-post-type.php
schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3222/commits/12733e0809844b239affcb8e8808eadb86d61f66

schlessera: I think this should cover it properly now, but not 100% certain.

westonruter: @schlessera Looks like it is needed in one more place: `\AMP_Validated_URL_Post_Type::handle_validation_error_status_update()`

includes/validation/class-amp-validated-url-post-type.php
schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3222/commits/b887ee5c4470334d0a03c6fd24b5fc04b8f7569e

includes/validation/class-amp-validated-url-post-type.php
schlessera: Fixed in https://github.com/ampproject/amp-wp/pull/3222/commits/19521867428cff20d01786c2822bc7730457c48f

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -360,6 +367,33 @@ public static function add_admin_menu_new_invalid_url_count() {
...
 		}
 
+		$new_validation_error_urls = get_transient( static::NEW_VALIDATION_ERROR_URLS_COUNT_TRANSIENT );
+
+		if ( false === $new_validation_error_urls ) {
+			$new_validation_error_urls = static::get_validation_error_urls_count();
+			set_transient( static::NEW_VALIDATION_ERROR_URLS_COUNT_TRANSIENT, $new_validation_error_urls, DAY_IN_SECONDS );
+		}
+
+		if ( 0 === $new_validation_error_urls ) {
```
westonruter: FYI: The strict equality here caused a regression since integers stored in transients when no persistent object cache is used will get cast to strings. See #3478.

PR #3236, opened 2019-09-13
Fix cut handler by shortcut

barklund:
_Fixes https://github.com/ampproject/amp-wp/issues/2989#issuecomment-531245896_

In order to allow the browser to correctly pick up what content is being cut, we need to keep it on-stage for another cycle before visually removing it from the page.

This is done by wrapping the block removal in a `setTimeout()`.

### TODO

- [ ] Add e2e tests for copy-pasting with keyboard shortcuts.


dero:
The solution looks good. Approved. We have decided to postpone adding the e2e tests for this update, correct?

PR #3237, opened 2019-09-13
Fix custom CTA text for page attachment

### LINKED ISSUE #1973: Implement support for `<amp-story-page-attachment>`

barklund:
Fixes #1973.

FE saves the CTA text in [a variable named `openText`](https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/blocks/amp-story-page-attachment/index.js#L22-L25), but the backend used `text` to access it. With a simple rename it works as expected with and without CTA text specified.


dero:
@barklund Just wondering about the failing e2e tests. Have you looked into the test run and verified the issues are not related to this change?

PR #3243, opened 2019-09-15
Keep track of the currently-open elements while walking over DOM and use depth-first post-order traversal to improve processing

### LINKED ISSUE #1493: Unexpected validation Handling of `<details>` tag

### LINKED ISSUE #3137: AMP_Allowed_Tags_Generated class is stripping amp-date-display {{iso}} format from amp-timeago component

westonruter:
This PR builds upon #3206 to simplify the logic for the `AMP_Tag_And_Attribute_Sanitizer`. Instead of introducing a new context construct which is added to the stack/queue, this PR replaces the contexts with a true recursive walking over the DOM tree, and keeping track of the current number of open elements by type at any given point. The number of open elements of given names can then be used not only to determine the context for whether or not attribute validation is currently happening inside of a `template`, but also it can be used to speed up checks for whether a given element has a particular ancestor.

* Walk over tree with depth-first search and post-order node traversal. This post-order traversal also fixes #1493 (thus superseding #1511), as the original parent node will be reported in a validation error since the deepest node is processed first.
* Instead of stack array with contexts, recursively walk through DOM tree and keep track of the number of open ancestor elements.
* Use open ancestor template elements to determine whether to skip validating attribute values.
* Use open ancestor elements as a shortcut for the logic in `AMP_Tag_And_Attribute_Sanitizer::get_ancestor_with_matching_spec_name()`.
* Add DOMElement and array types for arguments to `AMP_Tag_And_Attribute_Sanitizer` methods.
* Remove obsolete checks for arguments being `DOMElement`.
* Eliminate unnecessary `AMP_Tag_And_Attribute_Sanitizer::delegated_sanitize_disallowed_attribute_values_in_node()`.
* Use data provider for `test_replace_node_with_children_validation_errors`.

Fixes #3137. Closes #3206.


westonruter:
There is currently an unrelated fatal error happening when run PHPUnit tests against WordPress `trunk`:

> `Fatal error: Cannot declare class WP_Block_Styles_Registry, because the name is already in use in /app/public/content/plugins/gutenberg/lib/class-wp-block-styles-registry.php on line 14`

Also reported in Gutenberg: https://github.com/WordPress/gutenberg/issues/17444. 

The issue is resolved if I check out the current `master` branch of Gutenberg, so this appears to be a temporary problem that will be fixed with the next release of Gutenberg, so this may warrant merging without that job passing. The E2E failure also seems unrelated.

westonruter:
In regards to the the [CLA complaint](https://github.com/ampproject/amp-wp/pull/3243#issuecomment-531642507), it seems like a bug with the bot because all of the commits from @schlessera were OK in the other PR: https://github.com/ampproject/amp-wp/pull/3206. So I'm happy to forcibly flip the bit to `cla: yes`.

schlessera:
@googlebot I consent.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -234,7 +237,7 @@ public function get_scripts() {
 	 * @param array $attr_spec_list Attribute spec list.
 	 * @return array Modified attribute spec list.
 	 */
-	private function process_alternate_names( $attr_spec_list ) {
+	private function process_alternate_names( array $attr_spec_list ) {
```
schlessera: I strongly suggest not using actual typehints for `array`, as they introduce all sorts of problems with PHP. The main issue that any object that behaves like an array will not pass this type check, so you're forced to stay with dumb arrays and can never add any optimization later on.
```suggestion
	private function process_alternate_names( $attr_spec_list ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -252,35 +255,65 @@ private function process_alternate_names( $attr_spec_list ) {
...
+	private function sanitize_element( DOMElement $element ) {
+		if ( ! isset( $this->open_elements[ $element->nodeName ] ) ) {
+			$this->open_elements[ $element->nodeName ] = 0;
+		}
+		$this->open_elements[ $element->nodeName ]++;
+
+		$script_components = [];
+
+		// First recurse into children to sanitize descendants.
+		// The check for $element->parentNode at each iteration is to make sure an invalid child didn't bubble up removal
```
schlessera: Something off with the phrase here.
Maybe like this:
```suggestion
		// The check for $element->parentNode at each iteration is to make sure an invalid child didn't bubble up removed
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -292,7 +325,7 @@ public function sanitize() {
 	 * @param array      $rule_spec Rule spec.
 	 * @return array Augmented rule spec.
 	 */
-	private function get_rule_spec_list_to_validate( $node, $rule_spec ) {
+	private function get_rule_spec_list_to_validate( DOMElement $node, array $rule_spec ) {
```
schlessera: Again, would suggest not typehinting against actual arrays.
```suggestion
	private function get_rule_spec_list_to_validate( DOMElement $node, $rule_spec ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
schlessera: That being said, this is a private method, so there's no problem in changing the typehint later if needed. However, it could just as well happen that the method is turned public later without reconsidering the `array` typehint, that's why I just generally avoid it.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -603,7 +630,7 @@ private function process_node( $node ) {
 	 * @param DOMElement $node      The DOMElement of the node to check.
 	 * @return boolean $is_missing boolean Whether a required attribute is missing.
 	 */
-	public function is_missing_mandatory_attribute( $attr_spec, $node ) {
+	public function is_missing_mandatory_attribute( array $attr_spec, DOMElement $node ) {
```
schlessera:
```suggestion
	public function is_missing_mandatory_attribute( $attr_spec, DOMElement $node ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -640,7 +667,7 @@ public function is_missing_mandatory_attribute( $attr_spec, $node ) {
 	 * @param array      $cdata_spec CDATA.
 	 * @return true|WP_Error True when valid or error when invalid.
 	 */
-	private function validate_cdata_for_node( $element, $cdata_spec ) {
+	private function validate_cdata_for_node( DOMElement $element, array $cdata_spec ) {
```
schlessera:
```suggestion
	private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -667,11 +694,11 @@ private function validate_cdata_for_node( $element, $cdata_spec ) {
...
 	 * @since 0.5
 	 *
-	 * @param DOMNode $node The node to validate.
-	 * @param array   $tag_spec The specification.
+	 * @param DOMElement $node     The node to validate.
+	 * @param array      $tag_spec The specification.
 	 * @return boolean $valid Whether the node's placement is valid.
 	 */
-	private function validate_tag_spec_for_node( $node, $tag_spec ) {
+	private function validate_tag_spec_for_node( DOMElement $node, array $tag_spec ) {
```
schlessera:
```suggestion
	private function validate_tag_spec_for_node( DOMElement $node, $tag_spec ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -711,12 +738,12 @@ private function validate_tag_spec_for_node( $node, $tag_spec ) {
...
 	 *
-	 * @param DOMNode $node           Node.
-	 * @param array[] $attr_spec_list Attribute Spec list.
+	 * @param DOMElement $node           Node.
+	 * @param array[]    $attr_spec_list Attribute Spec list.
 	 *
 	 * @return float Number of times the attribute spec list matched. If there was a mismatch, then 0 is returned. 0.5 is returned if there is an implicit match.
 	 */
-	private function validate_attr_spec_list_for_node( $node, $attr_spec_list ) {
+	private function validate_attr_spec_list_for_node( DOMElement $node, array $attr_spec_list ) {
```
schlessera:
```suggestion
	private function validate_attr_spec_list_for_node( DOMElement $node, $attr_spec_list ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -936,20 +955,11 @@ private function validate_attr_spec_list_for_node( $node, $attr_spec_list ) {
...
-
-		if ( ! $node instanceof DOMElement ) {
-			/**
-			 * If $node is only a DOMNode and not a DOMElement we can't
-			 * remove an attribute from it anyway.  So bail out now.
-			 */
-			return [];
-		}
-
+	private function get_disallowed_attributes_in_node( DOMElement $node, array $attr_spec_list ) {
```
schlessera:
```suggestion
	private function get_disallowed_attributes_in_node( DOMElement $node, $attr_spec_list ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -970,42 +980,12 @@ private function get_disallowed_attributes_in_node( $node, $attr_spec_list ) {
...
-	 *
-	 * @see $this->sanitize_disallowed_attribute_values_in_node() which delegates to this method
-	 *
 	 * @param DOMElement $node                       Node.
 	 * @param array[][]  $attr_spec_list             Attribute spec list.
 	 * @param DOMAttr[]  $attributes_pending_removal Attributes pending removal.
 	 * @return DOMAttr[]|false Attributes to remove, or false if the element itself should be removed.
 	 */
-	private function delegated_sanitize_disallowed_attribute_values_in_node( $node, $attr_spec_list, $attributes_pending_removal ) {
+	private function sanitize_disallowed_attribute_values_in_node( DOMElement $node, array $attr_spec_list, array $attributes_pending_removal ) {
```
schlessera:
```suggestion
	private function sanitize_disallowed_attribute_values_in_node( DOMElement $node, $attr_spec_list, $attributes_pending_removal ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1022,6 +1002,12 @@ private function delegated_sanitize_disallowed_attribute_values_in_node( $node,
 				continue;
 			}
 
+			// Check the context to see if we are currently within a template tag.
+			// If this is the case and the attribute value contains a template placeholder, we skip sanitization.
+			if ( ! empty( $this->open_elements['template'] ) && preg_match( '/^.*{{.*}}/', $attr_node->nodeValue ) ) {
```
schlessera: The logic is not correct here. The `preg_match()` would only be correct for `type=amp-mustache`, any other type of template would probably have other placeholder syntax.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1102,7 +1088,7 @@ private function delegated_sanitize_disallowed_attribute_values_in_node( $node,
 	 *      - AMP_Rule_Spec::FAIL - $attr_name is mandatory, but doesn't exist
 	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name is not mandatory
 	 */
-	private function check_attr_spec_rule_mandatory( $node, $attr_name, $attr_spec_rule ) {
+	private function check_attr_spec_rule_mandatory( DOMElement $node, $attr_name, array $attr_spec_rule ) {
```
schlessera:
```suggestion
	private function check_attr_spec_rule_mandatory( DOMElement $node, $attr_name, $attr_spec_rule ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1136,7 +1122,7 @@ private function check_attr_spec_rule_mandatory( $node, $attr_name, $attr_spec_r
 	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
 	 *                                        is no rule for this attribute.
 	 */
-	private function check_attr_spec_rule_value( $node, $attr_name, $attr_spec_rule ) {
+	private function check_attr_spec_rule_value( DOMElement $node, $attr_name, array $attr_spec_rule ) {
```
schlessera:
```suggestion
	private function check_attr_spec_rule_value( DOMElement $node, $attr_name, $attr_spec_rule ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1290,10 +1276,8 @@ private function check_attr_spec_rule_value_regex( $node, $attr_name, $attr_spec
 	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
 	 *                                        is no rule for this attribute.
 	 */
-	private function check_attr_spec_rule_value_regex_casei( $node, $attr_name, $attr_spec_rule ) {
-		/**
-		 * Check 'value_regex_casei' - case insensitive regex match
-		 */
+	private function check_attr_spec_rule_value_regex_casei( DOMElement $node, $attr_name, array $attr_spec_rule ) {
```
schlessera:
```suggestion
	private function check_attr_spec_rule_value_regex_casei( DOMElement $node, $attr_name, $attr_spec_rule ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1323,7 +1307,7 @@ private function check_attr_spec_rule_value_regex_casei( $node, $attr_name, $att
 	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
 	 *                                        is no rule for this attribute.
 	 */
-	private function check_attr_spec_rule_valid_url( $node, $attr_name, $attr_spec_rule ) {
+	private function check_attr_spec_rule_valid_url( DOMElement $node, $attr_name, array $attr_spec_rule ) {
```
schlessera:
```suggestion
	private function check_attr_spec_rule_valid_url( DOMElement $node, $attr_name, $attr_spec_rule ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1385,7 +1369,7 @@ private function parse_protocol( $url ) {
 	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
 	 *                                        is no rule for this attribute.
 	 */
-	private function check_attr_spec_rule_allowed_protocol( $node, $attr_name, $attr_spec_rule ) {
+	private function check_attr_spec_rule_allowed_protocol( DOMElement $node, $attr_name, array $attr_spec_rule ) {
```
schlessera:
```suggestion
	private function check_attr_spec_rule_allowed_protocol( DOMElement $node, $attr_name, $attr_spec_rule ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1456,7 +1440,7 @@ static function ( $srcset_part ) {
 	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
 	 *                                        is no rule for this attribute.
 	 */
-	private function check_attr_spec_rule_disallowed_relative( $node, $attr_name, $attr_spec_rule ) {
+	private function check_attr_spec_rule_disallowed_relative( DOMElement $node, $attr_name, array $attr_spec_rule ) {
```
schlessera:
```suggestion
	private function check_attr_spec_rule_disallowed_relative( DOMElement $node, $attr_name, $attr_spec_rule ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1506,7 +1490,7 @@ private function check_attr_spec_rule_disallowed_relative( $node, $attr_name, $a
 	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
 	 *                                        is no rule for this attribute.
 	 */
-	private function check_attr_spec_rule_disallowed_empty( $node, $attr_name, $attr_spec_rule ) {
+	private function check_attr_spec_rule_disallowed_empty( DOMElement $node, $attr_name, array $attr_spec_rule ) {
```
schlessera:
```suggestion
	private function check_attr_spec_rule_disallowed_empty( DOMElement $node, $attr_name, $attr_spec_rule ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1530,7 +1514,7 @@ private function check_attr_spec_rule_disallowed_empty( $node, $attr_name, $attr
 	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
 	 *                                        is no rule for this attribute.
 	 */
-	private function check_attr_spec_rule_disallowed_domain( $node, $attr_name, $attr_spec_rule ) {
+	private function check_attr_spec_rule_disallowed_domain( DOMElement $node, $attr_name, array $attr_spec_rule ) {
```
schlessera:
```suggestion
	private function check_attr_spec_rule_disallowed_domain( DOMElement $node, $attr_name, $attr_spec_rule ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1563,7 +1547,7 @@ private function check_attr_spec_rule_disallowed_domain( $node, $attr_name, $att
 	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
 	 *                                        is no rule for this attribute.
 	 */
-	private function check_attr_spec_rule_blacklisted_value_regex( $node, $attr_name, $attr_spec_rule ) {
+	private function check_attr_spec_rule_blacklisted_value_regex( DOMElement $node, $attr_name, array $attr_spec_rule ) {
```
schlessera:
```suggestion
	private function check_attr_spec_rule_blacklisted_value_regex( DOMElement $node, $attr_name, $attr_spec_rule ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1605,7 +1589,7 @@ private function check_attr_spec_rule_blacklisted_value_regex( $node, $attr_name
 	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name does not exist or there
 	 *                                        is no rule for this attribute.
 	 */
-	private function check_attr_spec_rule_value_properties( $node, $attr_name, $attr_spec_rule ) {
+	private function check_attr_spec_rule_value_properties( DOMElement $node, $attr_name, array $attr_spec_rule ) {
```
schlessera:
```suggestion
	private function check_attr_spec_rule_value_properties( DOMElement $node, $attr_name, $attr_spec_rule ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1660,7 +1644,7 @@ private function check_attr_spec_rule_value_properties( $node, $attr_name, $attr
 	 * @param array[]|string[] $attr_spec_list Attribute spec list.
 	 * @return bool Return true if attribute name is valid for this attr_spec_list, false otherwise.
 	 */
-	private function is_amp_allowed_attribute( $attr_node, $attr_spec_list ) {
+	private function is_amp_allowed_attribute( DOMAttr $attr_node, array $attr_spec_list ) {
```
schlessera:
```suggestion
	private function is_amp_allowed_attribute( DOMAttr $attr_node, $attr_spec_list ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1863,7 +1833,7 @@ private function remove_disallowed_descendants( $node, $allowed_descendants ) {
 	 * }
 	 * @return bool Whether the element satisfies the requirements, or else it should be removed.
 	 */
-	private function check_valid_children( $node, $child_tags ) {
+	private function check_valid_children( DOMElement $node, array $child_tags ) {
```
schlessera:
```suggestion
	private function check_valid_children( DOMElement $node, $child_tags ) {
```

schlessera:
Regarding the `array` typehints: I'm all for adding stricter typehinting and you've noticed already that I've been adding them. But using `array` is broken and shouldn't be used, because it makes it impossible to replace dumb arrays by smarter iterables constructs later on. PHP 7 solved this by adding the `iterable` type, but we can't use that yet.

schlessera:
Regarding the general approach of using `open_templates[]` with recursion:
I thought that the entire point of introducing the `$stack` in this sanitizer was to get away from recursion. PHP does not support tail call optimization (TCO), so there is not proper way of using recursion without introducing a non-scalable limitation. I think that the limitation might be high enough here to not be an actual issue, but that is hard to guess and can never be guaranteed.
So the code is generally fine with that simplification, if we keep this caveat in mind.
However, the template-specific logic should be changed, as the fact that there is currently only one single type of template is not something we should rely on.

westonruter:
> Regarding the `array` typehints: I'm all for adding stricter typehinting and you've noticed already that I've been adding them. But using `array` is broken and shouldn't be used, because it makes it impossible to replace dumb arrays by smarter iterables constructs later on. PHP 7 solved this by adding the `iterable` type, but we can't use that yet.

@schlessera And here I was hoping you would be so proud of me 😄 But that makes perfect sense. Excellent point.

westonruter:
> However, the template-specific logic should be changed, as the fact that there is currently only one single type of template is not something we should rely on.

We can actually rely on that currently because we will know the moment that non-Mustache templates are allowed because the change will correspond with spec updates. So we'll see it when performing a spec update, and we can make the required changes at that time (if ever).

westonruter:
> I thought that the entire point of introducing the `$stack` in this sanitizer was to get away from recursion. PHP does not support tail call optimization (TCO), so there is not proper way of using recursion without introducing a non-scalable limitation. I think that the limitation might be high enough here to not be an actual issue, but that is hard to guess and can never be guaranteed.

@schlessera Interesting point. The tag-and-attribute sanitizer was introduced a long time ago (before my time) in https://github.com/ampproject/amp-wp/pull/600, specifically in https://github.com/delputnam/amp-wp/commit/e7e51f06364fd65bc84830f19baeb8d7ed6ba4ff. So I don't know if this was in mind when opting for a non-recursive solution. I kinda think that it may not have been on the radar, since the sanitizer was not yet concerned with validating constraints based on the location of a node in a hierarchy.

The concern here would be if an HTML page has a DOM tree with extreme nesting of elements. In HTML4 it appears the max nesting depth was 100 elements: https://bugzilla.mozilla.org/show_bug.cgi?id=256180#c36

This limit was lifted in HTML5, but it seems unusual to have documents with such deeply-nested elements (though I know they do exist). I can't seem to find any data on the frequency.

Perhaps this should be merged and then we can look at refactoring recursion to re-introduce iteration with a stack?

westonruter:
> The E2E failure also seems unrelated.

It just passed! 

image

schlessera:
I'm okay with merging this and reconsidering for v2 (or until we hit bug reports).

westonruter:
@schlessera I just tested and it appears that libxml has a maximum nesting depth of 256.

> Warning: DOMDocument::loadHTML(): Excessive depth in document: 256 use XML_PARSE_HUGE option

image

https://3v4l.org/2hpqn

So as long as PHP doesn't complain about recursing 256+ times, then this shouldn't be a problem faced by a recursive `sanitize_element` method. I added a5a27c4 to test this assumption.

schlessera:
> So as long as PHP doesn't complain about recursing 256+ times

PHP should be fine about that, but the default recursion limit if Xdebug is active is 100 (it can be raised, though).

swissspidy:
This seems to have broken code coverage collection on Travis CI, see #3337.

PR #3248, opened 2019-09-16
Detect unused variables using PHPCS

swissspidy:
Saw this [over at Gutenberg](https://github.com/WordPress/gutenberg/pull/17300) and thought it might be a nice little addition to our linting setup.


includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -1432,6 +1432,7 @@ public static function handle_validate_request() {
 
 		$post = null;
 		$url  = null;
+		$args = [];
```
westonruter: Interesting we didn't see a PHP notice for this before.

PR #3253, opened 2019-09-16
Add rel=preconnect link for AMP CDN

westonruter:
In a [support topic](https://wordpress.org/support/topic/add-google-font-in-the-footer/#post-11937934) I found out that Lighthouse audits are flagging the lack of `rel=preconnect` links for https://cdn.ampproject.org:

> image

This currently is not part of the [AMP optimization guide](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/optimize_amp/), and I thought it was unnecessary because there is already more-specific `preload` links, like:

```html
<link rel="preload" as="script" href="https://cdn.ampproject.org/v0.js">
```

I checked with @sebastianbenz and he pointed out that not all browsers support `rel=preload`, so adding a `rel=preconnect` link would be a good thing to do.

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1549,15 +1549,29 @@ class AMP_Theme_Support {
 			$head->appendChild( $script );
 		}
 
-		// Ensure rel=canonical link.
-		$links         = [];
+		// Gather all links.
+		$links         = [
+			'preconnect' => [
+				// Include preconnect link for AMP CDN for browsers that don't support preload.
+				AMP_DOM_Utils::create_node(
+					$dom,
+					'link',
+					[
+						'rel'  => 'preconnect',
+						'href' => 'https://cdn.ampproject.org',
+					]
+				),
+			],
+		];
 		$link_elements = $head->getElementsByTagName( 'link' );
-		$rel_canonical = null;
 		foreach ( $link_elements as $link ) {
 			if ( $link->hasAttribute( 'rel' ) ) {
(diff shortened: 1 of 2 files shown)


includes/class-amp-theme-support.php
```diff
@@ -1542,15 +1542,29 @@ public static function ensure_required_markup( DOMDocument $dom, $script_handles
 			$head->appendChild( $script );
 		}
 
-		// Ensure rel=canonical link.
-		$links         = [];
+		// Gather all links.
+		$links         = [
+			'preconnect' => [
+				// Include preconnect link for AMP CDN for browsers that don't support preload.
+				AMP_DOM_Utils::create_node(
```
westonruter: Only room for improvement here is that I'd like it if the page already had this link that it would be prevented from being duplicated.

schlessera: That would then come from a third-party plugin/theme instead? It would be difficult to detect, it might be easier to sanitize the duplicate away instead.

westonruter: Potentially. We actually can detect because we gather up a list of all the links right here. So we could just skip adding it if it already exists.

Aside: There is also a need to remove component scripts that do not exist among `$script_handles`, to remove scripts that are not relevant to the page (in case someone manually adds a script which turns out to not be relevant to the current page).

westonruter: In practice there should not be duplicates. After thinking about it more, it seems wasteful to detect for a duplicate `link[rel=preconnect]` for AMP, and it doesn't hurt to have it duplicated in the first place.

PR #3259, opened 2019-09-16
Fix resize for non-fit text box

### LINKED ISSUE #3199: Non-fitted text blocks cannot be resized smaller
A non-fitted text block (a block with the default-enabled "Adjust font size automatically" turned off) cannot be resized to a smaller size.

**Steps to recreate:**

- Create a text block.
- Type some text.
- Switch off "automatically fit text".
- Resize the block.

**Expected result:**
- You can make the box any size you want (down to the configured minimum size).

**Actual result:**
- You can only ever make the box bigger than it was when you started resizing or keep it the same size.
- If you make the box bigger, this new size will be the new "minimum size" for future resizing.

barklund:
*Fixes #3199.*

Reducing the size of non-fit text box was broken recently - possibly due changes upstream. The fixed height of text box is required for other section of the code to work correctly, however it broke reducing size here.

By setting the text box height to `auto` before calculating `scrollHeight`, the true height can once again be established - original height will be reset after change.

Furthermore, this commit also changes the `<=` to `<` in the limiting condition - to allow one-directional resize.

However there are still weird issues for rotated text boxes when hitting the minimum size, but that have probably existed for some time.

**Please note**, if you are reviewing this, play around with both fitted and non-fitted text boxes as well as other types of elements and try to resize them both larger and smaller in all directions. Furthermore, do this when rotated as well.


swissspidy:
This fixes #3199 for me at first glance. Will test it a bit more though.

> However there are still weird issues for rotated text boxes when hitting the minimum size, but that have probably existed for some time.

Do you mean this? https://cloudup.com/cXICc_z3pah

barklund:
> Do you mean this? https://cloudup.com/cXICc_z3pah

Exactly that. Equally, but differently, weird at all angles.

spacedmonkey:
This might be related, https://github.com/WordPress/gutenberg/pull/17011 . Upstream package was updated. This might have effected something here.

spacedmonkey:
Screenshot from 2019-09-18 10-07-54
I can still see this weird sizing when I paste text into the editor.

swissspidy:
> I can still see this weird sizing when I paste text into the editor.

Hmm this should be caught in the text block's edit component's `componentDidUpdate` method:

https://github.com/ampproject/amp-wp/blob/93276f713e78abf546c1277ee34564a9c4bdd123/assets/src/stories-editor/blocks/amp-story-text/edit.js#L70-L79

https://github.com/ampproject/amp-wp/blob/93276f713e78abf546c1277ee34564a9c4bdd123/assets/src/stories-editor/helpers/index.js#L1262-L1315

Perhaps it needs the same `height` fix there so the `element.offsetHeight` is correct.

spacedmonkey:
@swissspidy Should this be part of this PR or another one? Is it related?

swissspidy:
Semi-related, but can be looked at in a separate PR.

swissspidy:
Created two new issues for these bugs: #3290 and #3291.

PR #3274, opened 2019-09-17
Fix allow_failures setting in Travis config

swissspidy:


.travis.yml
```diff
@@ -59,8 +59,12 @@ after_script:
 
 jobs:
   fast_finish: true
+  # These need to be exact matches, including whitespace!
```
schlessera: Are you sure about the whitespace? There's nothing in the docs mentioning this, just that the exact same keys need to be used...

schlessera: Alright, I can see that it failed without whitespace already, so that's that.

swissspidy: It didn't work before without the whitespace, so yeah. Someone else experienced the same [on Stack Overflow](https://stackoverflow.com/questions/39563597/travis-ci-allow-failures-when-multiple-environment-variables-are-set) too.

Issue #2670, opened 2019-06-21
Remove unnecessary file_exists() checks from plugin builds


westonruter:
In order to help guide users to properly install the plugin when it has been cloned from GitHub, the plugin includes this code:

https://github.com/ampproject/amp-wp/blob/cb0c7e324f92ce5523378fee923fb7677f9acd1b/amp.php#L148-L157

As [potentially discovered](https://wordpress.org/support/topic/error-on-updating-to-the-latest-version-of-plugin/page/2/#post-11658421) on the support forum, the `file_exists()` check may return stale results that necessitate calling `clearstatcache()`. 

Nevertheless, this entire `if` statement is irrelevant from being included in plugin builds distributed from WordPress.org because we already _know_ that those files exist at build time. So we can potentially prevent a bug from happening and improve performance a _tiiiiiiny_ bit by just removing it entirely from `amp.php` during a build.

swissspidy:
Related core ticket: https://core.trac.wordpress.org/ticket/36455

westonruter:
It can be bumped.

westonruter:
I prototyped this yesterday in another plugin:

https://github.com/westonruter/syntax-highlighting-code-block/blob/47d78a41084f5625f349b3de785caef8941d997d/syntax-highlighting-code-block.php#L21

https://github.com/westonruter/syntax-highlighting-code-block/blob/47d78a41084f5625f349b3de785caef8941d997d/Gruntfile.js#L85

PR #3278, opened 2019-09-17
Update wordpress/scripts package to 5.0.0

swissspidy:
Replaces `deps.json` files with `asset.php` files.

Uses the script hashes instead of the plugin versions for enqueueing the scripts.

Supersedes #3272.


includes/admin/class-amp-post-meta-box.php
```diff
@@ -151,17 +151,20 @@ public function enqueue_admin_assets() {
 
 		wp_styles()->add_data( self::ASSETS_HANDLE, 'rtl', 'replace' );
 
-		$script_deps_path    = AMP__DIR__ . '/assets/js/' . self::ASSETS_HANDLE . '.deps.json';
-		$script_dependencies = file_exists( $script_deps_path )
-			? json_decode( file_get_contents( $script_deps_path ), false ) // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+		$asset_file = AMP__DIR__ . '/assets/js/' . self::ASSETS_HANDLE . '.asset.php';
+		$asset      = file_exists( $asset_file )
```
westonruter: The `file_exists()` check here could cause this issue to raise its head again: https://github.com/ampproject/amp-wp/issues/2670

Perhaps what is needed is something like https://github.com/ampproject/amp-wp/issues/2670 where the build step ensures that the file gets loaded unconditionally if running in a plugin build?

swissspidy: That sounds like the `file_exists()` check is not needed at all, since the plugin fails way earlier when running from source.

westonruter:
Should this go ahead and be part of 1.3?

swissspidy:
I hesitated a bit right before RC, and there‘s no user facing benefit in this build tooling change anyway. So... 🤷‍♂️

schlessera:
I strongly discourage putting PHP files into an `assets/js` folder, that is all kinds of misleading and could potentially even cause security issues.

westonruter:
Why did wordpress/scripts move from JSON to PHP for this data?

swissspidy:
The relevant change here in Gutenberg is https://github.com/WordPress/gutenberg/pull/17298.

It has been made in preparation for an eventual change to an easier and more automated asset registration in WordPress. In the future it should be possible to just do something like `wp_register_asset( 'my/path/to/script.js' );` and let WordPress figure out the rest because of all the metadata in the PHP files.

Using PHP as the output format apparently showed to be [up to 10x faster](https://github.com/WordPress/gutenberg/pull/17298#issuecomment-531813087) than using JSON.

However, I agree that hiding PHP files in `assets/js` is misleading and potentially dangerous.

Some options off the top of my head:

* Go back to using JSON.  
  Apparently there is an `outputFormat` option that can be used for that.
* Move asset PHP files to different location.  
  This would probably require some upstream change.
* Move the PHP code inline during build.  
  This would mean replacing the `require` call with the actual array from that file.
* Change folder structure.

.gitignore
```diff
@@ -7,7 +7,7 @@ amp.zip
 /assets/css/*
 !/assets/css/src/
 assets/js/*.js
-assets/js/*.deps.json
+assets/js/*.asset.php
```
westonruter: This update was missed in `Gruntfile.js`. See #3366.

schlessera:
Related discussion in Core: https://core.trac.wordpress.org/ticket/48154

PR #3286, opened 2019-09-18
Use template_dir consistently as signal for Transitional mode

westonruter:
An issue reported by @archon810 was that if if a theme contains a theme support declaration like the following:

```php
add_theme_support( 'amp', [
	'template_dir' => 'amp',
] );
```

The theme support option in the admin will default to Standard when first activating the plugin, as opposed to the Transitional mode, which is what is implied by supplying `template_dir` without `paired`. This is shown in `amp_is_canonical()`:

https://github.com/ampproject/amp-wp/blob/93276f713e78abf546c1277ee34564a9c4bdd123/amp.php#L600-L601

The issue is that this same logic is not being used `AMP_Theme_Support::read_theme_support()`:

https://github.com/ampproject/amp-wp/blob/93276f713e78abf546c1277ee34564a9c4bdd123/includes/class-amp-theme-support.php#L321

This is something that was likely missed in #2550.

So this PR ensures that `AMP_Theme_Support::read_theme_support()` results in the same theme support being derived as is indicated by `amp_is_canonical()`.

To test this PR:

1. Add the above `add_theme_support()` to a theme's `functions.php`
2. Run `wp option delete amp-options`.
3. Load the AMP settings screen.
4. You should see Transitional as the selected mode as opposed to Standard.

### DIFF
--- a/tests/php/test-class-amp-options-manager.php
+++ b/tests/php/test-class-amp-options-manager.php
@@ -269,6 +269,95 @@ class Test_AMP_Options_Manager extends WP_UnitTestCase {
 		}
 	}
 
+
+	public function get_test_get_options_defaults_data() {
+		return [
+			'reader'                               => [
+				null,
+				AMP_Theme_Support::READER_MODE_SLUG,
+			],
+			'transitional_without_template_dir'    => [
+				[
+					'paired' => true,
+				],
+				AMP_Theme_Support::TRANSITIONAL_MODE_SLUG,
+			],
+			'transitional_implied_by_template_dir' => [
+				[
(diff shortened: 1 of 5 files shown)


tests/php/test-class-amp-options-manager.php
```diff
@@ -269,6 +269,95 @@ public function test_get_and_set_options() {
 		}
 	}
 
+
+	public function get_test_get_options_defaults_data() {
```
kienstra: Great testing of this.

westonruter:
Moved directly to Ready for Merging.

Issue #3249, opened 2019-09-16
Review Testing Strategy


MackenzieHartung:
## Feature description

As a contributor, I want to have a better testing strategy. 

Our test strategy will be mainly informed by the ideas of Kent C Dodds.

The most important aspects of his thinking is covered in these posts:
- [Write tests. Not too many. Mostly integration.](https://kentcdodds.com/blog/write-tests)
- [Write fewer, longer tests.](https://kentcdodds.com/blog/write-fewer-longer-tests)

Preliminary considerations include the following stack:
- jest for overall test running.
- react-testing-library for setup, manipulation and validation of React components.
- puppeteer for running a limited set of e2e tests.
- axe for accessibility testing.
- storybook for visual unit testing.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* AC1: This ticket is to begin the review and discussion surrounding the existing testing strategy (e.g. e2e tests).
* AC2: Additionally, we should be looking into the automation surrounding accessibility testing, as we examine our testing strategy as a whole.
* AC3: Testing should consider the new editor being developed
 
## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

PR #3294, opened 2019-09-18
Add AXE into e2e testing

spacedmonkey:
First pass at adding AXE into e2e testing. 

Related: #3249

### DIFF
--- a/tests/e2e/config/bootstrap.js
+++ b/tests/e2e/config/bootstrap.js
@@ -123,6 +123,40 @@ function observeConsoleLogging() {
 }
 
 /**
+ * Runs Axe tests when the block editor is found on the current page.
+ *
+ * @return {?Promise} Promise resolving once Axe texts are finished.
+ */
+async function runAxeTestsForBlockEditor() {
+	if ( ! await page.$( '.block-editor' ) ) {
+		return;
+	}
+
+	await expect( page ).toPassAxeTests( {
+		// Temporary disabled rules to enable initial integration.
+		// See: https://github.com/WordPress/gutenberg/pull/15018.
+		disabledRules: [
+			'aria-allowed-role',
+			'aria-valid-attr-value',
+			'button-name',
+			'color-contrast',
+			'dlitem',
+			'duplicate-id',
+			'label',
+			'link-name',
+			'listitem',
+			'region',
+		],
+		exclude: [
+			// Ignores elements created by metaboxes.
+			'.edit-post-layout__metaboxes',
+			// Ignores elements created by TinyMCE.
+			'.mce-container',
+		],
+	} );
+}
+
+/**
  * Before every test suite run, delete all content created by the test. This ensures
  * other posts/comments/etc. aren't dirtying tests and tests don't depend on
  * each other's side-effects.
@@ -140,6 +174,7 @@ beforeAll( async () => {
 // eslint-disable-next-line jest/require-top-level-describe
 afterEach( async () => {
 	await clearLocalStorage();
+	await runAxeTestsForBlockEditor();
 	await setBrowserViewport( 'large' );
 } );
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -4077,6 +4077,15 @@
         "enzyme-to-json": "^3.3.5"
       }
     },
+    "@wordpress/jest-puppeteer-axe": {
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@wordpress/jest-puppeteer-axe/-/jest-puppeteer-axe-1.3.0.tgz",
+      "integrity": "sha512-rQg88X4cgN3UKSUazAlI4q6+nQfTIBe8I3l31GUGlUOeTNjmvJOO6+sFJDGsP4/h8XLYhOZe4j2Y9m0OMa02eQ==",
+      "dev": true,
+      "requires": {
+        "axe-puppeteer": "^1.0.0"
+      }
+    },
     "@wordpress/keycodes": {
       "version": "2.6.0",
       "resolved": "https://registry.npmjs.org/@wordpress/keycodes/-/keycodes-2.6.0.tgz",
@@ -5606,6 +5615,21 @@
       "resolved": "https://registry.npmjs.org/aws4/-/aws4-1.8.0.tgz",
       "integrity": "sha512-ReZxvNHIOv88FlT7rxcXIIC0fPt4KZqZbOlivyWtXLt8ESx84zd3kMC6iK5jVeS2qt+g7ftS7ye4fi06X5rtRQ==",
       "dev": true
+    },
+    "axe-core": {
+      "version": "3.3.2",
+      "resolved": "https://registry.npmjs.org/axe-core/-/axe-core-3.3.2.tgz",
+      "integrity": "sha512-lRdxsRt7yNhqpcXQk1ao1BL73OZDzmFCWOG0mC4tGR/r14ohH2payjHwCMQjHGbBKm924eDlmG7utAGHiX/A6g==",
+      "dev": true
+    },
+    "axe-puppeteer": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/axe-puppeteer/-/axe-puppeteer-1.0.0.tgz",
+      "integrity": "sha512-hTF3u4mtatgTN7fsLVyVgbRdNc15ngjDcTEuqhn9A7ugqLhLCryJWp9fzqZkNlrW8awPcxugyTwLPR7mRdPZmA==",
+      "dev": true,
+      "requires": {
+        "axe-core": "^3.1.2"
+      }
     },
     "axobject-query": {
       "version": "2.0.2",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "@wordpress/hooks": "2.6.0",
     "@wordpress/html-entities": "2.5.0",
     "@wordpress/i18n": "3.6.0",
+    "@wordpress/jest-puppeteer-axe": "1.3.0",
     "@wordpress/keycodes": "2.6.0",
     "@wordpress/plugins": "2.7.0",
     "@wordpress/postcss-themes": "2.2.0",
--- a/tests/e2e/jest.config.js
+++ b/tests/e2e/jest.config.js
@@ -8,6 +8,7 @@ module.exports = {
 	],
 	setupFilesAfterEnv: [
 		'<rootDir>/config/bootstrap.js',
+		'@wordpress/jest-puppeteer-axe',
 		'expect-puppeteer',
 	],
 	testPathIgnorePatterns: [


tests/e2e/config/bootstrap.js
```diff
@@ -122,6 +122,40 @@ function observeConsoleLogging() {
...
+ *
+ * @return {?Promise} Promise resolving once Axe texts are finished.
+ */
+async function runAxeTestsForBlockEditor() {
+	if ( ! await page.$( '.block-editor' ) ) {
+		return;
+	}
+
+	await expect( page ).toPassAxeTests( {
+		// Temporary disabled rules to enable initial integration.
```
swissspidy: Too bad this hasn't got traction so far: https://github.com/WordPress/gutenberg/issues/15452

tests/e2e/config/bootstrap.js
```diff
@@ -122,6 +122,40 @@ function observeConsoleLogging() {
 	} );
 }
 
+/**
+ * Runs Axe tests when the block editor is found on the current page.
+ *
+ * @return {?Promise} Promise resolving once Axe texts are finished.
+ */
+async function runAxeTestsForBlockEditor() {
+	if ( ! await page.$( '.block-editor' ) ) {
```
swissspidy: Already looking forward to apply these tests to other parts of the AMP plugin as well 💥

tests/e2e/config/bootstrap.js
spacedmonkey: I am not sure what to do here. I was going to remove some of these rules, however part of the gutenberg editor fail, which gives a false positive. Unfortunately we have to stick with what gutenberg is doing, unless we want our tests to fail everytime we update upstream. :(

spacedmonkey:
Even having these tests in has already highlight some issues we have. It will also help with future work. 

I would recommend fixing the issues that are found in the PR in another PR. Once that is complete, merge this as is and we can tweak it later.

swissspidy:
@spacedmonkey Can you create individual issues for all the things flagged here so we can better track them?

assets/src/stories-editor/components/autocomplete/index.js
```diff
@@ -47,6 +47,7 @@ class Autocomplete extends OriginalAutocomplete {
 			tStatusQueryTooShort,
 			tStatusSelectedOption,
 			tStatusResults,
+			ariaLabel,
```
spacedmonkey: Sneaked in some very small accessibility fixes here to make the tests pass.

tests/e2e/config/bootstrap.js
```diff
@@ -122,6 +122,38 @@ function observeConsoleLogging() {
...
+		return;
+	}
+
+	await expect( page ).toPassAxeTests( {
+		// Temporary disabled rules to enable initial integration.
+		// See: https://github.com/WordPress/gutenberg/pull/15018.
+		disabledRules: [
+			'aria-allowed-role',
```
spacedmonkey: These rules are disabled because of the errors in gutenberg. I would love for this ignored rules to be removed.

tests/e2e/specs/stories-editor/block-position-controls.js
```diff
@@ -104,7 +104,7 @@ describe( 'Block Position Controls', () => {
 
 			// Since the selected block is already at the back, the 'Back' and 'Backward' buttons should be disabled.
 			expect( page ).toMatchElement( '.amp-story-controls-send-back[aria-disabled="true"]' );
-			expect( page ).toMatchElement( '.amp-story-controls-send-backward[aria-disabled="true"]' );
+			expect( page ).toMatchElement( '.amp-story-controls-send-backwards[aria-disabled="true"]' );
```
spacedmonkey: This is a fix for e2e to pass. Not sure how this was passing before.

tests/e2e/config/bootstrap.js
```diff
@@ -122,6 +122,38 @@ function observeConsoleLogging() {
 	} );
 }
 
+/**
+ * Runs Axe tests when the block editor is found on the current page.
+ *
+ * @return {?Promise} Promise resolving once Axe texts are finished.
+ */
+async function runAxeTestsForBlockEditor() {
```
swissspidy: Can we rename this to `runAxeTestsForStoriesEditor` and use a different selector to detect it?

tests/e2e/config/bootstrap.js
```diff
@@ -122,6 +122,38 @@ function observeConsoleLogging() {
...
+			'color-contrast',
+			'duplicate-id',
+			'region',
+		],
+		exclude: [
+			// Ignores elements created by metaboxes.
+			'.edit-post-layout__metaboxes',
+			// Ignores elements created by TinyMCE.
+			'.mce-container',
+			// Ignore attachent close button.
```
swissspidy:
```suggestion
			// Ignore attachment close button.
```

tests/e2e/config/bootstrap.js
```diff
@@ -122,6 +122,38 @@ function observeConsoleLogging() {
...
+ * @return {?Promise} Promise resolving once Axe texts are finished.
+ */
+async function runAxeTestsForBlockEditor() {
+	if ( ! await page.$( '.block-editor' ) ) {
+		return;
+	}
+
+	await expect( page ).toPassAxeTests( {
+		// Temporary disabled rules to enable initial integration.
+		// See: https://github.com/WordPress/gutenberg/pull/15018.
```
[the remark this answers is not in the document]
spacedmonkey: This code was copied and pasted from this PR. The reason the tests are disabled are because of this PR. For reference, it makes more sense to link to the original PR here.

spacedmonkey:
> I'd like to see a reference to which exact rule(s), file and line, that should be enabled for each of the "child" issues from this one (#3301, #3305, #3307).

Good point @barklund . 

Regarding these other issues. 
- #3301 Will be fixed this PR. 
- #3305 The role for this is ignored see [this](https://github.com/ampproject/amp-wp/pull/3294/files#diff-062b4845cbce5614a6b8635e5f1e0a80R140) line
- #3307 The role for this is ignored see [this](https://github.com/ampproject/amp-wp/pull/3294/files#diff-062b4845cbce5614a6b8635e5f1e0a80R141) line

As suggested by @westonruter , existing fails were ignored where possible and other that couldn't be ignored were fixed. Those two issues will go into the backlog and will like not be looked at until after 2.0 release.

tests/e2e/config/bootstrap.js
swissspidy: The block editor and stories editor will have different configurations though. The `exclude` config already has a stories-specific config in it as well.

Also, eventually we should eventually add these tests for the plugin settings screen as well, where we're going to have yet another set of configuration options.

Having three separate sets would make maintenance easier.

tests/e2e/config/bootstrap.js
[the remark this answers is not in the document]
swissspidy: Then I'd rather link to https://github.com/WordPress/gutenberg/issues/15452 where progress on that is being tracked, _and_ this PR.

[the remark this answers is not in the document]
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3294/commits/78161f9d42a87c9ca636467daaa96705ee4540aa

tests/e2e/config/bootstrap.js
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3294/commits/78161f9d42a87c9ca636467daaa96705ee4540aa

assets/src/stories-editor/components/font-family-picker/index.js
```diff
@@ -63,6 +63,7 @@ function FontFamilyPicker( {
 						inputValue: inputValueTemplate,
 					}
 				}
+				ariaLabel={ __( 'Font Family', 'amp' ) }
```
swissspidy: Isn't this redundant when the `BaseControl` already displays a label?

The bug seems to be that it displays the label for the element with ID `id`, but the `Autocomplete` component internally uses `${ id }__listbox` as the ID, and not just `id` alone.

So the correct fix would be passing `id={ `${ id }__listbox` }` to the `BaseControl` so that the label actually references an existing element.

tests/e2e/config/bootstrap.js
swissspidy: The selector in `runAxeTestsForStoriesEditor` is still the same though.

What I meant was having separate functions with separate configurations:

* `runAxeTestsForBlockEditor` with the current `.block-editor` selector
* `runAxeTestsForStoriesEditor` with a `#amp-story-editor` selector
* Eventually: `runAxeTestsForSettingsScreen` with a `#amp-settings` selector
* Eventually: `runAxeTestsForValidationScreen` with a `.post-type-amp_validated_url` selector

assets/src/stories-editor/components/font-family-picker/index.js
spacedmonkey: That solution didn't work ^ . I tried something similar in development.

This solution did work - https://github.com/ampproject/amp-wp/pull/3294/commits/8fee020a2503f13eedb5f32a8868bc648fd50a94

tests/e2e/config/bootstrap.js
spacedmonkey: > The selector in `runAxeTestsForStoriesEditor` is still the same though.
> 
> What I meant was having separate functions with separate configurations:
> 
> * `runAxeTestsForBlockEditor` with the current `.block-editor` selector
> * `runAxeTestsForStoriesEditor` with a `#amp-story-editor` selector
> * Eventually: `runAxeTestsForSettingsScreen` with a `#amp-settings` selector
> * Eventually: `runAxeTestsForValidationScreen` with a `.post-type-amp_validated_url` selector

What are you asking to change here? Just change the selector, like this `runAxeTestsForStoriesEditor` with a `#amp-story-editor` selector

assets/src/stories-editor/components/font-family-picker/index.js
barklund: I don't fully follow with this is trying to fix.

I understand that AXE is complaining, that the rendered `<input />` inside the autocomplete doesn't have a label (a `<label />` or an `aria-label`), right? And this fix ads a label (via `aria-label`), though it's kind of a duplicate.

But it does have a `<label />` **and** the label's `for` attribute correctly references the `id` of the `<input />`:
label_for_input

So I don't really follow neither the problem nor the fix?

swissspidy: Oh, I see. The issue was with the `listbox` which, other than the `input`, didn't have a label associated with it.

If there's a visible label, then `aria-labelledby` should be used to refer to that label. 8fee020a2503f13eedb5f32a8868bc648fd50a94 did this.

tests/e2e/config/bootstrap.js
swissspidy: Yes. **And** have a **second function**, `runAxeTestsForBlockEditor`, with a different selector that only targets the block editor but **not** the stories editor. That function would then not `exclude` `'.amp-story-page-attachment-close-button'` as it's **irrelevant** in the standalone block editor.

If it's unclear I can also change this myself.

swissspidy: And of course, let me know if you think it doesn't make sense. Then we can just do the `#amp-story-editor` selector change for now.

assets/src/stories-editor/components/font-family-picker/index.js
swissspidy: So to summarize this:

* The `input` field has "Font Family" as its label (due to the `for` attribute).
* The list of items (`listbox`) has "Type to search for fonts` as its label (due to  the `aria-labelledby`).

In my testing with VoiceOver the help text was not read aloud at all, but perhaps it's done by other screen readers.

In any case, even if this passes the AXE test it would be good to have some manual tests for this independent of this PR, once it's merged.

spacedmonkey: The list box is NOT the input. It is the `<ul>` that contains the autocomplete options.

swissspidy: So much is clear yes

Issue #2092, opened 2019-04-08
Detect high cache miss rate for obtaining parsed CSS from variable stylesheets


westonruter:
When PHP is generated with random content (e.g. https://github.com/ampproject/amp-wp/issues/1239), the post-processor cache gets disabled as of https://github.com/ampproject/amp-wp/pull/1325 to prevent polluting the object cache with garbage cache values. This only matters when a persistent object cache is present, since that is when the post-processor cache is used.

When the style sanitizer comes across a stylesheet, it parses it and stores the cached result in a transient. This is done regardless of whether there is a persistent object cache present, since parsing CSS is expensive. If, however, the CSS has variable content in any way then this can result in a site with out a persistent object cache to have its `wp_options` table balloon to hundreds of thousands of garbage transients, as seen on this [support topic](https://wordpress.org/support/topic/database-is-being-spammed-very-hardly-by-amp-plugin/).

I believe such variable stylesheets to be an uncommon occurrence, but when it does happen, we should warn the user and stop storing the parsed CSS in transients. There could also be a parameter to the style sanitizer to disable the transient cache, as well as provide a threshold for what constitutes an unacceptable rate of cache misses.

schlessera:
I suggest solving this like you avoid filling disks with log information: by using rotation.

We should define a limited number of slots to store cached CSS results in and use the content hash to check whether we need to use a new slot or not. So we could for example cache the last 10 versions of the CSS stylesheet.

This would avoid the worst case scenario where you just fill the DB until the server is taken down. We could also track the velocity of how fast we go through these to show a warning to the user if caching does not make sense.

westonruter:
That sounds great.

Note that each individual stylesheet (`link`, `style` element, `style` attribute) is stored in a separate transient. So the rotation should be expanded to include a large enough slots for all the  variable number of stylesheets on a site. There could be quite a few. Perhaps default to 1000? Maybe it should be configurable?

schlessera:
With #3300, the main problem should be resolved: we won't be able to completely fill up the database anymore.

To now detect high velocity cycling so that we can show a warning to the user and/or skip caching, I can think of two separate approaches:

* After each request, check the distance that the index has travelled and compare it to a threshold. This is a simplistic approach and would probably detect most instances where we do useless caching, provided we get the threshold right. However, as we don't know any specifics, we can only disable the entire caching system for stylesheets. This would require us to add one more DB write to store information that allows us to calculate this delta.
[...]

Issue #1313, opened 2018-08-03
Add support for inline styles as long as they aren't !important


westonruter:
As noted in #1312, [AMP now allows inline `style` attributes](https://www.ampproject.org/docs/design/responsive_amp#add-styles-to-a-page) as long as they do not include `!important` qualifiers. Nevertheless, the AMP plugin continues to have the style sanitizer extract inline styles to combine into `style[amp-custom]`. The reason is that we need to preserve the `!important` styles, which we have to continue extracting to create the higher-specificity selector rules. And if we selectively move such non-important inline styles to `style[amp-custom]` there could be unintended side effects with the cascade. This is something to look into after the 1.0 release.

westonruter:
> The reason is that we need to preserve the `!important` styles, which we have to continue extracting to create the higher-specificity selector rules.

Consider an input HTML document that looks like this:

```html
<style>
p#para {
    color: red !important;
    font-weight: normal !important;
    text-decoration: underline;
}
</style>
<p id="para" style="color: green !important; font-weight: bold; text-decoration: overline;">I should still be green, not bold, and overlined!</p>
```

At the moment, this would get converted into AMP as follows:

```html
<style amp-custom>
:root:not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_) .amp-wp-27ad0ad {
    color: green;
}
:root:not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_):not(#_) p#para {
    color: red;
    font-weight: normal;
}
:root:not(#_):not(#_):not(#_):not(#_):not(#_) .amp-wp-27ad0ad {
    font-weight: bold;
    text-decoration: overline;
}
p#para {
    text-decoration: underline;
}
</style>
<p id="para" class="amp-wp-27ad0ad">I should still be green, not bold, and overlined!</p>
```

This renders in AMP identically as non-AMP:

![image](https://user-images.githubusercontent.com/134745/65732715-fce75c00-e07f-11e9-8640-7788a82d75ae.png)

However, what if instead of extracting the inline style, we just removed the `!important` qualifier, since the property in a `style` attribute should override any non-important rule, regardless of the selector specificity? So consider this output instead:

```html
<style amp-custom>
[...]

PR #3300, opened 2019-09-19
Add cache pool abstraction to cache stylesheets

schlessera:
This PR adds a cache pool abstraction that abstracts away the differences between the object cache and transients when caching stylesheets.

The transient cache implementation provides a fixed-size pool of cache entries that are rotated to enforce a maximum number of transients and avoid filling up the database.

Note: For the architecture rethink, this should be an interface with 2 separate implementations that will get injected, instead of a single class that contains all implementations.

See #2092

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -77,6 +77,18 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	const SELECTOR_EXTRACTED_ATTRIBUTES = 3;
 
 	/**
+	 * Cache group to use for caching stylesheets.
+	 *
+	 * This should be bumped whenever the PHP-CSS-Parser is updated or parsed format is updated.
+	 *
+	 * @var string
+	 * @private
+	 * @since 1.4
+	 * @see \AMP_Style_Sanitizer::prepare_stylesheet()
+	 */
+	const CACHE_GROUP = 'amp-parsed-stylesheet-v19';
+
+	/**
 	 * Array of flags used to control sanitization.
 	 *
 	 * @var array {
@@ -305,6 +317,13 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	];
 
 	/**
+	 * Cache pool implementation to use if no object cache is available.
+	 *
+	 * @var AMP_Cache_Pool;
+	 */
+	private $cache_pool;
+
+	/**
 	 * Get error codes that can be raised during parsing of CSS.
 	 *
 	 * This is used to determine which validation errors should be taken into account
@@ -399,6 +418,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 		$this->base_url    = untrailingslashit( $guessurl );
 		$this->content_url = WP_CONTENT_URL;
 		$this->xpath       = new DOMXPath( $dom );
+		$this->cache_pool  = new AMP_Cache_Pool( self::CACHE_GROUP );
 	}
 
 	/**
@@ -1350,7 +1370,6 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	private function process_stylesheet( $stylesheet, $options = [] ) {
 		$parsed      = null;
 		$cache_key   = null;
-		$cache_group = 'amp-parsed-stylesheet-v19'; // This should be bumped whenever the PHP-CSS-Parser is updated or parsed format is updated.
 
 		$cache_impacting_options = array_merge(
 			wp_array_slice_assoc(
@@ -1367,12 +1386,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 		);
 
 		$cache_key = md5( $stylesheet . wp_json_encode( $cache_impacting_options ) );
-
-		if ( wp_using_ext_object_cache() ) {
-			$parsed = wp_cache_get( $cache_key, $cache_group );
-		} else {
-			$parsed = get_transient( $cache_group . '-' . $cache_key );
-		}
+		$parsed    = $this->cache_pool->get( $cache_key );
 
 		/*
 		 * Make sure that the parsed stylesheet was cached with current sanitizations.
@@ -1390,18 +1404,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 
 		if ( ! $parsed || ! isset( $parsed['stylesheet'] ) || ! is_array( $parsed['stylesheet'] ) ) {
 			$parsed = $this->prepare_stylesheet( $stylesheet, $options );
-
-			/*
-			 * When an object cache is not available, we cache with an expiration to prevent the options table from
-			 * getting filled infinitely. On the other hand, if an external object cache is available then we don't
-			 * set an expiration because it should implement LRU cache expulsion policy.
-			 */
-			if ( wp_using_ext_object_cache() ) {
-				wp_cache_set( $cache_key, $parsed, $cache_group );
-			} else {
-				// The expiration is to ensure transient doesn't stick around forever since no LRU flushing like with external object cache.
-				set_transient( $cache_group . '-' . $cache_key, $parsed, MONTH_IN_SECONDS );
-			}
+			$this->cache_pool->set( $cache_key, $parsed );
 		}
 
 		return $parsed;
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -1368,8 +1368,8 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 * }
 	 */
 	private function process_stylesheet( $stylesheet, $options = [] ) {
-		$parsed      = null;
-		$cache_key   = null;
+		$parsed    = null;
+		$cache_key = null;
 
 		$cache_impacting_options = array_merge(
 			wp_array_slice_assoc(
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -418,7 +418,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 		$this->base_url    = untrailingslashit( $guessurl );
 		$this->content_url = WP_CONTENT_URL;
 		$this->xpath       = new DOMXPath( $dom );
-		$this->cache_pool  = new AMP_Cache_Pool( self::CACHE_GROUP );
+		$this->cache_pool  = new AMP_Cache_Pool( self::CACHE_GROUP, isset( $this->args['cache_pool_size'] ) ? $this->args['cache_pool_size'] : AMP_Cache_Pool::DEFAULT_POOL_SIZE );
 	}
 
 	/**
--- a/includes/utils/class-amp-cache-pool.php
+++ b/includes/utils/class-amp-cache-pool.php
@@ -31,7 +31,7 @@ final class AMP_Cache_Pool {
 	 *
 	 * @var int
 	 */
-	private $pool_index;
+	private $pool_index = -1;
 
 	/**
 	 * Cache group to use.
@@ -107,7 +107,7 @@ final class AMP_Cache_Pool {
 	private function get_rotated_transient( $key ) {
 		$pool_index = array_search( $key, $this->pool_map, true );
 
-		if ( ! $pool_index ) {
+		if ( false === $pool_index ) {
 			return false;
 		}
 
@@ -129,7 +129,7 @@ final class AMP_Cache_Pool {
 		}
 
 		// As we didn't find the key, we create a new pool slot to store it.
-		if ( false === $pool_index ) {
+		if ( false === $pool_index || -1 === $this->pool_index ) {
 			$this->advance_pool_index();
 			$this->pool_map[ $this->pool_index ] = $key;
 		}
@@ -146,7 +146,7 @@ final class AMP_Cache_Pool {
 	 */
 	private function read_pool_meta() {
 		$this->pool_map   = get_transient( "{$this->group}-pool-map" ) ?: [];
-		$this->pool_index = get_transient( "{$this->group}-pool-index" ) ?: 0;
+		$this->pool_index = get_transient( "{$this->group}-pool-index" ) ?: -1;
 	}
 
 	/**
--- a/includes/utils/class-amp-cache-pool.php
+++ b/includes/utils/class-amp-cache-pool.php
@@ -13,6 +13,13 @@
 final class AMP_Cache_Pool {
 
 	/**
+	 * Default size of the cache pool.
+	 *
+	 * @var int
+	 */
+	const DEFAULT_POOL_SIZE = 1000;
+
+	/**
 	 * Pool map of cached entries.
 	 *
 	 * @var array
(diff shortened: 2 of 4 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1367,12 +1386,7 @@ private function process_stylesheet( $stylesheet, $options = [] ) {
 		);
 
 		$cache_key = md5( $stylesheet . wp_json_encode( $cache_impacting_options ) );
-
-		if ( wp_using_ext_object_cache() ) {
-			$parsed = wp_cache_get( $cache_key, $cache_group );
-		} else {
-			$parsed = get_transient( $cache_group . '-' . $cache_key );
-		}
+		$parsed    = $this->cache_pool->get( $cache_key );
```
kienstra: This looks to work well.

Once the stylesheet is stored in a transient. this `->get()` method retrieves it.

sanitizer-here

kienstra:
**Approved**

Hi @schlessera,
This looks really good. 

Once a parsed stylesheet is stored in a transient, this retrieves it from the transient, instead of parsing it again.

I set an Xdebug breakpoint and stepped through `$this->cache_pool->set( $cache_key, $parsed )`, and it looked to work as excepted.

includes/utils/class-amp-cache-pool.php
```diff
@@ -0,0 +1,170 @@
...
+	 */
+	private $is_object_cache_available;
+
+	/**
+	 * Instantiate an AMP_Cache_Pool object.
+	 *
+	 * @param string $group Optional. Group to use. Defaults to an empty string.
+	 * @param int    $size  Optional. Size of the cache pool.
+	 */
+	public function __construct( $group = '', $size = self::DEFAULT_POOL_SIZE ) {
```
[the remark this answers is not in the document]
schlessera: The size is needed, and it felt obvious to allow this to be set through instantiation. This way, you could also create several pools with differing groups that might have different sizes as well.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -399,6 +418,7 @@ public function __construct( DOMDocument $dom, array $args = [] ) {
 		$this->base_url    = untrailingslashit( $guessurl );
 		$this->content_url = WP_CONTENT_URL;
 		$this->xpath       = new DOMXPath( $dom );
+		$this->cache_pool  = new AMP_Cache_Pool( self::CACHE_GROUP );
```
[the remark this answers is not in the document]
schlessera: This is not being injected because I don't think the architecture is currently set up well for injecting dependencies.
For now, I wanted to keep it simple and just directly instantiated here. Hopefully, we'll have injection for v2.

includes/utils/class-amp-cache-pool.php
```diff
@@ -0,0 +1,170 @@
...
+		$this->group = $group;
+		$this->size  = $size;
+
+		$this->is_object_cache_available = wp_using_ext_object_cache();
```
westonruter: If using an external object cache, then essentially the cache pool is not used, is that right? This assumes that that the persistent object cache has implemented some algorithm for garbage collection. 

Aside: Shouldn't WordPress transient system implement some cache replacement policy? If there was such a cache pool in core like you've implemented here, then there wouldn't be a need for this class, right?

includes/utils/class-amp-cache-pool.php
```diff
@@ -0,0 +1,170 @@
...
+
+	/**
+	 * Default size of the cache pool.
+	 *
+	 * @var int
+	 */
+	const DEFAULT_POOL_SIZE = 1000;
+
+	/**
+	 * Pool map of cached entries.
```
westonruter:
```suggestion
	 * Pool map of cached entry keys.
```

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: Perhaps the pool size should be passed as an argument from `$this->args`? If a default were supplied in `$DEFAULT_ARGS` then this would allow for a site to override the default if the default is not suitable.

```suggestion
		$this->cache_pool  = new AMP_Cache_Pool( self::CACHE_GROUP, isset( $this->args['cache_pool_size'] ? $this->args['cache_pool_size'] : AMP_Cache_Pool::DEFAULT_POOL_SIZE ) );
```

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1390,18 +1404,7 @@ private function process_stylesheet( $stylesheet, $options = [] ) {
...
-			 * getting filled infinitely. On the other hand, if an external object cache is available then we don't
-			 * set an expiration because it should implement LRU cache expulsion policy.
-			 */
-			if ( wp_using_ext_object_cache() ) {
-				wp_cache_set( $cache_key, $parsed, $cache_group );
-			} else {
-				// The expiration is to ensure transient doesn't stick around forever since no LRU flushing like with external object cache.
-				set_transient( $cache_group . '-' . $cache_key, $parsed, MONTH_IN_SECONDS );
-			}
+			$this->cache_pool->set( $cache_key, $parsed );
```
westonruter: This is great. 👍

includes/utils/class-amp-cache-pool.php
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	/**
+	 * Advance the index into the pool cache.
+	 */
+	private function advance_pool_index() {
+		self::$pool_indexes[ $this->group ]++;
+
+		if ( self::$pool_indexes[ $this->group ] >= $this->size ) {
+			self::$pool_indexes[ $this->group ] = 0;
```
westonruter: In regards to https://github.com/ampproject/amp-wp/issues/2092#issuecomment-533004679, this would be where we could potentially show an error message to the user, warning the that stylesheet cache pool is overflowing, yes? This should issue a PHP warning to the error log?

includes/sanitizers/class-amp-style-sanitizer.php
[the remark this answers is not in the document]
kienstra: Ah, OK. I thought this was dependency injection.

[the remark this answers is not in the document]
schlessera: Dependency injection would be if exterior code would inject "some cache" and this code here would not need to know which exact implementation is being used. This would then decouple this code from the cache, and the cahe could be changed/replaced without needing to adapt this code here.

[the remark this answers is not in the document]
kienstra: OK, thanks!

includes/utils/class-amp-cache-pool.php
schlessera: Yes, the manual pooling/rotation is only used with transients. Proper object caches have built-in mechanisms for this.

schlessera: I think WordPress should use something like this with a separate table, transients as they are now are a hack with several performance issues.

includes/utils/class-amp-cache-pool.php
schlessera: No, this is only expressing the "distance travelled". Velocity is distance over time, we don't know anything about the time here. If we hit the end of the pool several times a day, we have a very high velocity. If we only hit it every 3 months, it might just be a site with lots of edits.

includes/sanitizers/class-amp-style-sanitizer.php
schlessera: Yes, that might come in handy. I can't commit your suggestion as is, though, it is broken (parentheses in the wrong position).

schlessera:
> This may make sense to also use in the AMP_Image_Dimension_Extractor since images can also end up filling up the wp_options table.

Yes, it was already build to be usable in multiple locations, with differing groups and sizes, and the tests already make sure this works as expected.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: Fixed 😊

schlessera:
I wonder whether the default size of 1000 pool entries (per group) is not too high. The main issue I can see is that we serialize and unserialize the entire pool map. For 1000 entries, this might have a non-negligible performance impact.

westonruter:
It's hard to know. We somewhat need data to help inform that.

This provides the count:

```bash
wp transient list --fields=name --search='amp-parsed-stylesheet-*' --format=count
```

I've asked someone to run that to get a number from a production site that is not running an object cache.

westonruter:
Note that a the parsed stylesheets being stored here can come from:

1. Stylesheets referenced by `link` elements.
2. Stylesheets inline in `style` elements.
3. Style properties stored in inline `style` attributes.

A transient will be created for each 👉 _unique_ 👈 instance of the above to contain the parsed stylesheet information.

schlessera:
On hold for now as we are awaiting an export of production transient data from an actual site to evaluate.

schlessera:
Here's my initial observation from looking through the export we got from a bigger site:
* Every value is a serialization of a serialization. Given that most of our memory usage is located in `unserialize()` according to initial profiling, I recommend we change this so that we only ever have one level of serialization.
* The format is not very space-efficient, there's certainly improvements we can do here.
* A large majority of the entries are every simple, often just a single property, like `flex:0.3456` or `text-align:center`. I think we should discuss how to store this in a more optimized and/or introducing utility classes. So, instead of having `.amp-wp-4263548 { text-align: center; } .amp-wp-9263742 { text-align: center; } ...`, we could have `.amp-wp-tacenter { text-align: center; }` instead which would deduplicate this. Similarly, we could have "calculated utility classes" that need to be calculated and tracked, but don't need to be stored. So, instead of `.amp-wp-3856964 { flex: 2.564; }` we could have `.amp-wp-flex-2-564 { flex: 2.564; }`. This would not deduplicate these (as they are quite individualized), but we would not need to store them in the database.
* The fields that are almost always empty should not even be part of the serialization, like `'validation_results'` or `'imported_font_urls'`.

schlessera:
Also, what is the reasoning behind extracting all the inline styles into `<style amp-custom>`? As far as I can tell, AMP does support inline styles...
Image 2019-09-26 at 4 45 29 PM

kienstra:
>Also, what is the reasoning behind extracting all the inline styles into <style amp-custom>? As far as I can tell, AMP does support inline styles...

Ah, interesting. Maybe that's because AMP used to not support inline styles.

westonruter:
> Ah, interesting. Maybe that's because AMP used to not support inline styles.

That's mostly why, but there are other reasons. See https://github.com/ampproject/amp-wp/issues/1313 which suggests the elimination of moving inline `style` attributes to `style` rules, but this is problematic currently when a style rule contains an `!important` property, as that is how a style rule can override an inline style, and there's no way to override it otherwise, as far as I know. Please continue this discussion following https://github.com/ampproject/amp-wp/issues/1313#issuecomment-535731938.

westonruter:
> * A large majority of the entries are every simple, often just a single property, like `flex:0.3456` or `text-align:center`. I think we should discuss how to store this in a more optimized way and/or introducing utility classes. So, instead of having `.amp-wp-4263548 { text-align: center; } .amp-wp-9263742 { text-align: center; } ...`, we could have `.amp-wp-tacenter { text-align: center; }` instead which would deduplicate this. Similarly, we could have "calculated utility classes" that need to be calculated and tracked, but don't need to be stored. So, instead of `.amp-wp-3856964 { flex: 2.564; }` we could have `.amp-wp-flex-2-564 { flex: 2.564; }`. This would not deduplicate these (as they are quite individualized), but we would not need to store them in the database.

These are all coming from inline styles (in `style` attributes). This being the case, it may be better to parse the style rules at runtime rather than fire up the PHP CSS Parser. This was done recently in #3209 when converting elements for `layout=fill`. Perhaps this could be done here as well, and this would eliminate a lot of transients. We'd want to ensure the same sanitizations are done on the properties, however.

schlessera:
> These are all coming from inline styles (in style attributes). This being the case, it may be better to parse the style rules at runtime rather than fire up the PHP CSS Parser. This was done recently in #3209 when converting elements for layout=fill. Perhaps this could be done here as well, and this would eliminate a lot of transients. We'd want to ensure the same sanitizations are done on the properties, however.

I'm not quite sure I understand what you mean by the above, specifically the suggestion to "parse the style rules at runtime rather than fire up the PHP CSS Parser". Would that mean they are neither validated/sanitized, not tree-shaken anymore? And what would you then actually do with them at runtime? I seem to be missing something here.

westonruter:
The inline styles don't need to be tree-shaken, because we know they are going to exist in the page. What I mean is that in [`AMP_Style_Sanitizer::collect_inline_styles()`](https://github.com/ampproject/amp-wp/blob/72bd4418f2e0862156b13af24caf6cc78af0fe6b/includes/sanitizers/class-amp-style-sanitizer.php#L2417-L2470) it is currently constructing a style rule dynamically like so:

https://github.com/ampproject/amp-wp/blob/72bd4418f2e0862156b13af24caf6cc78af0fe6b/includes/sanitizers/class-amp-style-sanitizer.php#L2437-L2439

And then the rule gets processed:

https://github.com/ampproject/amp-wp/blob/72bd4418f2e0862156b13af24caf6cc78af0fe6b/includes/sanitizers/class-amp-style-sanitizer.php#L2443-L2449

This then causes the PHP CSS Parser to be fired up. The `AMP_Style_Sanitizer::process_stylesheet()` takes as input a _stylesheet_, meaning it has to handle arbitrary style rules. However, the data being processed here is simple style properties. So it seems this could do something faster by skipping the full parsing and instead extract the properties to then do the work in the [`AMP_Style_Sanitizer::process_css_delcaration_block()`](https://github.com/ampproject/amp-wp/blob/72bd4418f2e0862156b13af24caf6cc78af0fe6b/includes/sanitizers/class-amp-style-sanitizer.php#L2025-L2102), which would be limited to just needing to do:

* [Remove disallowed properties](https://github.com/ampproject/amp-wp/blob/72bd4418f2e0862156b13af24caf6cc78af0fe6b/includes/sanitizers/class-amp-style-sanitizer.php#L2049-L2067).
* [Transform `!important` qualifiers](https://github.com/ampproject/amp-wp/blob/72bd4418f2e0862156b13af24caf6cc78af0fe6b/includes/sanitizers/class-amp-style-sanitizer.php#L2093)

And I think that's all that needs to be done for properties in a `style` attribute.

schlessera:
Okay, I understood now, the term "runtime" was confusing me. I'll open a new issue for that.

What is needed now to take this PR across the finish line?

- [ ] Add size argument to `$DEFAULT_ARGS` so sites can override our default.
- [ ] Add documentation about the transient pool and how to adapt its size.

Would this be enough to merge this for now?

westonruter:
Given that a large majority of the transients are for inline styles, if the PHP CSS Parser were bypassed in favor of a more lightweight parser (e.g. using regex, though this is less robust!) then we could avoid needing to store so many transients and there will be less need to increase the size of the pool.

westonruter:
I'm just concerned that if a site has 5000 transients and the pool size is 1000, but they don't know they have so many, then there will be very frequent cache misses and they wouldn't know it is happening.

archon810:
This 👆.

westonruter:
I wonder if the PHP CSS Parser is actually sufficiently fast for the simple case if CSS properties in a `style` attribute.

westonruter:
Also because there is no need to parse the selector for tree shaking.

westonruter:
Closing in favor of #4177

Issue #3311, opened 2019-09-20
Allow drag element to different story page


barklund:
_Result of #3180._

As a story editor, I want to be able to drag an element from one story page to another.

This should work for all elements.

This should allow me to drag either to a direct neighbor or (if my viewport shows it) to an even further neighbor in either direction.

Auto-scrolling is outside the scope of this ticket. If you want to drag to a page much further "down the line" (e.g. from page 1 to page 10), you have to drop on intermediate pages that you can "reach" and then continue drag from there.

barklund:
### Testing instructions

#### Test drag of regular block
1. Create new empty story.
2. Add a new page 2.
3. Add a text block 2 page 2.
4. Drag new text block on page 2 to page 1, and see that it drops there and the editor switches to have page 1 in focus.

#### Test drag of CTA block
1. Create new empty story.
2. Add a new page 2.
3. Add a new page 3.
4. Add a CTA block to page 3.
5. Add a new page 4.
6. Switch to page 2.
7. Add a CTA block to page 2.
8. Attempt to drag CTA block from page 2 to page 1, and see that **it's not allowed** *(because it's the first page in the story)*.
9. Attempt to drag CTA block from page 2 to page 3, and see that **it's not allowed** *(because it already has a CTA block)*.
10 Attempt to drag CTA block from page 2 to page 4, and see that **it is allowed**.

Note: to see all the way from page 2 to page 4, you might need to reduce content size several times and/or use a very wide browser.

csossi:
Verified in QA

barklund:
Demo created here: https://youtu.be/xUzGdwFeP-w

PR #3312, opened 2019-09-20
Allow drag to different page

barklund:
_Fixes #3311._

This PR allows editors to drag elements to other pages.

See early demo here: https://www.youtube.com/watch?v=ONFOvRF2jog

This is done by while dragging, detecting where the actual mouse cursor is relative to the current page. A highlight callback is invoked while dragging with the offset of the page currently under the cursor. If dropped above another page, a drop callback is invoked.

The callbacks then check if there even is a page in this direction and then either highlight that page or clone the element to that page respectively.

### TODO

- [x] Allow dragging an element to another page.
- [x] Fix dragging for CTAs.
- [x] Figure out if we need dragging for Page Attachments (**resolution:** we don't for now).
- [x] Maintain proper Separation of Concerns.
- [x] Make sure CTA can't be moved to a page which already has a CTA.
- [x] Tests? Unit testing is almost impossible, but an e2e test is feasible.


assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -10,28 +10,87 @@ import PropTypes from 'prop-types';
...
+
+		// Do we even have a neighbor in that direction? (offset=0 is not a neighbor)
+		if ( offset === 0 || newPageIndex < 0 || newPageIndex >= pages.length ) {
+			return null;
+		}
+
+		return pages[ newPageIndex ];
+	};
+
+	const onNeighborDrop = ( offset, finalPosition ) => {
```
swissspidy: At first glance, I'd probably use `withDispatch` and put `onNeighborDrop` and `onNeighborHover`  there.

barklund: Hm, that's not really doable due to the `currentHoverElement` as it's created with a hook not available in a `withSelect` handler (I think?). But maybe it's not even necessary to create this element with a hook, as the BlockDraggable is never re-rendered while the drag is ongoing (at least I don't think it is). Will try to experiment with this.

barklund: The hook isn't necessary in the component, but it definitely can't go in a `withDispatch` as it's re-invoked for every drag of the element (as the register updates I guess).

I'll put another component in-between and see if I like the result.

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -54,13 +90,108 @@ BlockDraggable.propTypes = {
...
+			rootClientId,
+			block: getBlock( clientId ),
+			getBlockOrder,
+		};
+	} ),
+	withDispatch( ( dispatch, { getBlockOrder, rootClientId, clientId, block, blockName } ) => {
+		const { setCurrentPage } = dispatch( 'amp/story' );
+		const { selectBlock, removeBlock, insertBlock, updateBlockAttributes } = dispatch( 'core/block-editor' );
+
+		const isCTABlock = 'amp/amp-story-cta' === blockName;
```
spacedmonkey: Attachment and CTA, they are only one allowed per page. I can't see the validation here doesn't allow you to move these blocks to a page that already has one.

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -54,13 +90,108 @@ BlockDraggable.propTypes = {
...
+			index: getBlockIndex( clientId, rootClientId ),
+			rootClientId,
+			block: getBlock( clientId ),
+			getBlockOrder,
+		};
+	} ),
+	withDispatch( ( dispatch, { getBlockOrder, rootClientId, clientId, block, blockName } ) => {
+		const { setCurrentPage } = dispatch( 'amp/story' );
+		const { selectBlock, removeBlock, insertBlock, updateBlockAttributes } = dispatch( 'core/block-editor' );
+
```
spacedmonkey: Why not use `ALLOWED_MOVABLE_BLOCKS` here and if not a movable block, then don't do the move.

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -54,13 +90,108 @@ BlockDraggable.propTypes = {
...
+		/**
+		 * Is the current element allowed on a given page?
+		 *
+		 * This is currently only used to filter out instances, where CTA's aren't allowed.
+		 *
+		 * @param {number} targetPageIndex  Zero-based index of new page
+		 * @param {string} targetPageId  Block id of new page
+		 * @return {boolean} Returns true if this element is allowed on the given page, false otherwise.
+		 */
+		const isElementAllowedOnPage = ( targetPageIndex, targetPageId ) => {
```
spacedmonkey: Might be worth comparing this against [ensureAllowedBlocksOnPaste](https://github.com/ampproject/amp-wp/blob/06ff84fec4357359ce556c889aacecfb06db66cb/assets/src/stories-editor/helpers/index.js#L1867-L1898).

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -54,13 +90,108 @@ BlockDraggable.propTypes = {
...
+		 * @param {Object} newAttributes  Object with attributes to update on element on new page.
+		 */
+		const onNeighborDrop = ( offset, newAttributes ) => {
+			const newPageId = getNeighborPageId( offset );
+			if ( ! newPageId ) {
+				return;
+			}
+
+			// Remove block and add cloned block to new page.
+			removeBlock( clientId );
```
spacedmonkey: Would be safer to remove the block last thing here?

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -54,13 +90,108 @@ BlockDraggable.propTypes = {
...
+			const clonedBlock = cloneBlock( block );
+			insertBlock( clonedBlock, null, newPageId );
+			updateBlockAttributes( clonedBlock.clientId, newAttributes );
+
+			// Switch to new page.
+			setCurrentPage( newPageId );
```
spacedmonkey: :+1:

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -61,13 +64,33 @@ class Draggable extends Component {
...
 			event.preventDefault();
 		}
 
+		// Attempt drop on neighbor if offset
+		if ( this.pageOffset !== 0 ) {
+			// All this is about calculating the position of the (correct) element on the new page.
+			const currentElementTop = parseInt( this.cloneWrapper.style.top );
+			const currentElementLeft = parseInt( this.cloneWrapper.style.left );
+			const newLeft = currentElementLeft - ( this.pageOffset * PAGE_AND_BORDER );
+			const isCTABlock = 'amp/amp-story-cta' === blockName;
```
spacedmonkey: `isCTABlock` should be a helper function. It referenced 3 times.

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -61,13 +64,33 @@ class Draggable extends Component {
...
+			const baseHeight = isCTABlock ? STORY_PAGE_INNER_HEIGHT / 5 : STORY_PAGE_INNER_HEIGHT;
```
spacedmonkey: Should this also check for attachment blocks.

spacedmonkey:
This PR really needs e2e tests. 

I would also look rebasing and running those e2e tests against AXE for accessibility. I would look [aria-grabbed](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-grabbed_header) and  [aria-dropeffect](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-dropeffect) accessibility attributes. These attributes will likely need to be updated dynamically blocks move.

spacedmonkey:
We might want to update using the screen reader feature using [speak](https://github.com/WordPress/gutenberg/tree/f91e66ca2bcd9e1e7b97aa39c01f2888fc751712/packages/a11y#speak). We might want to update these users that the drag was / wasn't successful.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: Unmovable blocks aren't wrapped in `Draggable`, so this code will never be run for those - AFAIK.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: Attachment isn't wrapped in Draggable, so it's not relevant here.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: It does overlap, but it also handles blocks, that aren't draggable.

(also, it's a weird piece of code - why isn't it an `array.filter` expression?)

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: What could happen if done in reverse?

assets/src/stories-editor/components/block-mover/draggable.js
barklund: Attachment blocks aren't draggable.

assets/src/stories-editor/components/block-mover/draggable.js
barklund: Makes sense.

barklund:
> This PR really needs e2e tests.

I'm having problems getting the e2e setup to properly test these. It might be because those tests are just too power-intensive for my machine though (which is bad!).

Also, e2e's aren't meant for exhaustive testing in any way. It's an anti-pattern if e2e's are used for this. e2e should only be used for some very overall happy-path testing to ensure, that overall dragging works as expected. Unit and integration testing should be done to validate the finer things.

swissspidy:
FWIW we've had issues in the past with testing dragging as Chromium/Puppeteer doesn't have the best support there.

barklund:
> We might want to update using the screen reader feature using [speak](https://github.com/WordPress/gutenberg/tree/f91e66ca2bcd9e1e7b97aa39c01f2888fc751712/packages/a11y#speak). We might want to update these users that the drag was / wasn't successful.

Uh, that's a simple yet interesting package. It could definitely be used here. But maybe the overall a11y of dragging should be considered in some strategy rather than hotfixes I can attempt with `speak()`. I don't have the knowledge required to make sure this is actually accessible and doesn't just feel like it.

spacedmonkey:
> I don't have the knowledge required to make sure this is actually accessible and doesn't just feel like it.

I agree here, my understanding of `speak` in this use case is limited. Not sure what exactly the messaging should be here. Maybe we could loop in an accessibility expect for some help here. 

However the  aria-grabbed and aria-dropeffect accessibility attributes seem simple enough to implement here and would be a good start. 

This feature is not an accessibility friendly one in the first place, as most people with accessibility needs will be using screen readers or keyboard anyway. Which is why #3330 would be better for those users.

assets/src/stories-editor/components/block-mover/block-draggable.js
spacedmonkey: CTA isn't not draggable either it seems. So why is this check required?

assets/src/stories-editor/components/block-mover/block-draggable.js
spacedmonkey: Good point. Your check below for CTA confused me, as that is a none draggable block.

assets/src/stories-editor/components/block-mover/block-draggable.js
swissspidy: Worth noting that the button within the CTA block is draggable. In case that makes a difference here.

assets/src/stories-editor/components/block-mover/block-draggable.js
spacedmonkey: Why not move this to helper functions file and then add some unit / integration tests?

( I didn't write that function, but I find switch statement easy to read. )

assets/src/stories-editor/components/block-mover/block-draggable.js
miina: (Not sure if it's relevant here but FI that the button inside CTA block is draggable and it does use the `StoryBlockMover` component.)

barklund: Ah, I see. The button inside the CTA block is the draggable one, but for the purposes of the draggable component, it kind of appears to be the block that is draggable. There are some checks around this making sure that it's the right part of the block that moves of course.

assets/src/stories-editor/components/block-mover/block-draggable.js
spacedmonkey: This code reminds of code used for cutting blocks. There was an issue that was removed block before copied block was ready. This might not be an issue here, just came to mind when reviewing code. 

Wonder if this could / should reuse code from cut and paste? They basically do the same thing..

assets/src/stories-editor/components/block-mover/draggable.js
spacedmonkey: There already seems to be number of places in code where `isCTABlock` is referenced.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: The function internally use the `select`'ed function `getCallToActionBlock`, but that could of course be supplied to the helper. I'll look into it. But yes, it is a good candidate for unit test.

I don't really see an easy (or even a hard) combination with cut and paste. Functionality is overlapping, but far from similar.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: The function internally use the `select`'ed function `getCallToActionBlock`, but that could of course be supplied to the helper. I'll look into it. But yes, it is a good candidate for unit test.

barklund: Actually there's a lot of this logic existing elsewhere – e.g. used in the inserter, that only shows allowed elements and already hides CTA on first page or on pages with an existing CTA. I will try to combined all of these restrictions.

barklund: This rewrite even caused me to remove a test, but I think it's okay as the heavy-lifting has been offloaded to Gutenberg core functionality in stead.

Also it's much cleaner now, thanks for the heads up!

barklund:
The recent refactor of the _"is block A allowed on page B"_ functionality touches on a lot of extra components, that will need to be QA'ed as well: Media Inserter, Shortcuts, Paste Block.

Just FYI for future reference.

assets/src/stories-editor/components/inserter/menu.js
```diff
@@ -409,9 +408,7 @@ export default compose(
 				return true;
 			}
 
-			// canInsertBlockType() alone is not enough, see https://github.com/WordPress/gutenberg/issues/14515
-			const destinationBlockListSettings = getBlockListSettings( destinationRootClientId );
-			return canInsertBlockType( name, getCurrentPage() ) && destinationBlockListSettings && destinationBlockListSettings.allowedBlocks.includes( name );
```
barklund: Hm, why does this old code reference `getCurrentPage()` in one instance and `destinationRootClientId` in the other? Should they both be `destinationRootClientId` and thus this should be used in the refactored code as well?

barklund: @swissspidy it seems you [made the original implementation here](https://github.com/ampproject/amp-wp/commit/bb4a35c71fb673e77b2bcbd5167a15ae1c1dc2a2#diff-e633a19ee56c05aab9455a179f61a078R387-R395). Can you help me resolve this? Is my refactor correct or should it reference `destinationRootClientId `.

assets/src/stories-editor/helpers/index.js
```diff
@@ -1868,33 +1870,23 @@ export const copyTextToClipBoard = ( text ) => {
...
+	blocks.filter( ( block ) => isBlockAllowedOnPage( block.name, pageId ) );
+
+/**
+ * Is the given block allowed on the given page?
+ *
+ * @param {Object}  name The name of the block to test.
+ * @param {string}  pageId Page ID.
+ * @return {boolean} Returns true if the element is allowed on the page, false otherwise.
+ */
+export const isBlockAllowedOnPage = ( name, pageId ) => {
```
barklund: Most invocations of this function have `pageId` set to current page. Should it just default to this perhaps (if null, set to `getCurrentPage()`), reducing the complexity of those calls?

assets/src/stories-editor/helpers/index.js
```diff
@@ -1868,33 +1870,23 @@ export const copyTextToClipBoard = ( text ) => {
...
+	// canInsertBlockType() alone is not enough, see https://github.com/WordPress/gutenberg/issues/14515
+	const blockSettings = getBlockListSettings( pageId );
+	return canInsertBlockType( name, pageId ) && blockSettings && blockSettings.allowedBlocks.includes( name );
```
barklund: I don't actually know why this works, why this disables multiple instances of Attachments/CTA's and CTA's on the first page, but it does. Testing that the function actually does that is probably best left to where those settings have been set – wherever that might be.

barklund: Ah, [here it is](https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/blocks/amp-story-page/edit.js#L260-L261). I didn't even know it checked that you could only have one of the two things on page, not just one of each.

assets/src/stories-editor/components/block-mover/draggable.js
barklund: Done in 6109bec.

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -54,13 +90,79 @@ BlockDraggable.propTypes = {
 	children: PropTypes.any.isRequired,
 	onDragStart: PropTypes.func,
 	onDragEnd: PropTypes.func,
+	onNeighborHover: PropTypes.func,
+	onNeighborDrop: PropTypes.func,
+	getNeighborPageId: PropTypes.func,
```
swissspidy: These prop types here can be marked `isRequired` as they should be guaranteed to exist

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -17,12 +17,20 @@ import { withSafeTimeout } from '@wordpress/compose';
...
+	getPercentageFromPixels,
+	getRelativeElementPosition,
+	isCTABlock,
+} from '../../helpers';
 import {
 	STORY_PAGE_INNER_WIDTH,
 	STORY_PAGE_INNER_HEIGHT,
 } from '../../constants';
 
+const PAGE_BORDER = 50;
```
swissspidy: `PAGE_BORDER` might be a bit misleading. In `EditorCarousel` there is also a `PAGE_BORDER` but it refers to the actual element's border. Here it is the space between elements. But where is that defined? Does it match some CSS somewhere? Would be good to have that mental connection documented here.

barklund: I spent waaaaaaay too much time searching for this one only to find it was a CSS only thing. It's defined [here](https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/blocks/amp-story-page/edit.css#L31) plus [here](https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/blocks/amp-story-page/edit.css#L72).

And yes, it would be a good idea to document this origin. And rename it to `PAGE_SPACING` or similar.

spacedmonkey:
This PR needs a rebase and conflicting file fixed.

assets/src/stories-editor/helpers/test/ensureAllowedBlocksOnPaste.js
```diff
@@ -1,28 +0,0 @@
-/**
```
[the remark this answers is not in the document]
barklund: The function  only sends the argument to another function. A proper test of this function alone would then only confirm, that it calls the inner function correctly. I actually tried to do that, but mocking a function inside a module while testing another function inside the same module is... nign-on impossible (too many weird workarounds and stubs required)! And testing that the function correctly loops over its array argument seems trivial.

The actual return values aren't determined by this function. The old test does still pass, but it's no longer determined by this function.

It's not a unit test, if it's testing this function **and** 3+ other functions, 2 of the from Gutenberg core.

assets/src/stories-editor/helpers/index.js
barklund: I was considering it, but would then have to mock the `select`'ed functions making it a bit tricky, but of course possible.

I'll take a swing at it and see how cumbersome it is.

assets/src/stories-editor/components/block-mover/draggable.js
swissspidy: `STORY_PAGE_BORDER` is just as misleading as `PAGE_BORDER` IMO. "spacing" or "margin" makes more sense.

spacedmonkey: > `STORY_PAGE_BORDER` is just as misleading as `PAGE_BORDER` IMO. "spacing" or "margin" makes more sense.

`STORY_PAGE_MARGIN` works for me.

barklund: Fixed in b3be9f5.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: Fixed in 04fffea.

assets/src/stories-editor/helpers/index.js
barklund: It required reorganising things a bit, but it's a lot cleaner now: 3c7afcf.

barklund: I did not actually default `pageId` to `getCurrentPage()`. It's of course still possible. Thoughts?

assets/src/stories-editor/constants.js
```diff
@@ -5,6 +5,7 @@ import { __ } from '@wordpress/i18n';
 
 export const STORY_PAGE_INNER_WIDTH = 328;
 export const STORY_PAGE_INNER_HEIGHT = 553;
+export const STORY_PAGE_MARGIN = 50;
```
swissspidy: Just realized there is `TOTAL_PAGE_MARGIN` in the `EditorCarousel` component:

https://github.com/ampproject/amp-wp/blob/86370162531938d30d518ae9d552d6ba48a3b8ff/assets/src/stories-editor/components/editor-carousel/index.js#L23-L24

Let's use `STORY_PAGE_MARGIN` there too 🎉

barklund: Fixed in 64f341d.

spacedmonkey:
Code is looking like it's nearly there. Accessilibity is still a worry here, what are your plans to make this accessible @barklund ?

barklund:
> Code is looking like it's nearly there. Accessibility is still a worry here, what are your plans to make this accessible @barklund ?

So, I did this: 7ee41f3.

But it doesn't really work as far as I can tell. It seems the `aria-grabbed` attribute is stripped from the resulting HTML.

I could also add speech to when the mouse is over another page and when the element is dropped on another page. Adding `aria-dropeffect` is a lot more complex though. I don't want to start doing too much like this until I actually know if any of it is working for end users.

In order to implement accessible drag and drop, I'd prefer to defer it to #3325 and implement keyboard-based navigation and not bother with making the mouse-based navigation accessible. I wouldn't mind going straight to this ticket now, but it's not the top priority as per milestones.

barklund:
> > Code is looking like it's nearly there. Accessibility is still a worry here, what are your plans to make this accessible @barklund ?
> 
> So, I did this: [7ee41f3](https://github.com/ampproject/amp-wp/commit/7ee41f35de3feaea2255f867485d2af473f47238).
> 
> But it doesn't really work as far as I can tell. It seems the `aria-grabbed` attribute is stripped from the resulting HTML.

`aria-grabbed` is [deprecated in WAI-ARIA 1.1+](https://w3c.github.io/aria/#aria-grabbed). And [this means that _"it is recommended that authors do not use it for new content"_](https://w3c.github.io/aria/#deprecated).

So there's no reason to introduce it here, as it's against recommendations. I will revert.

My recommendation to defer drag accessibility to #3325 still stands.

spacedmonkey:
> `aria-grabbed` is [deprecated in WAI-ARIA 1.1+](https://w3c.github.io/aria/#aria-grabbed). And [this means that _"it is recommended that authors do not use it for new content"_](https://w3c.github.io/aria/#deprecated).
> 
> So there's no reason to introduce it here, as it's against recommendations. I will revert.
> 
> My recommendation to defer drag accessibility to #3325 still stands.

Sorry that was bad advice on my part as I only quickly looked these up. It seems that both [aria-grabbed](https://w3c.github.io/aria/#aria-grabbed) and [aria-dropeffect](https://w3c.github.io/aria/#aria-dropeffect) are deprecated as WAI-ARIA 1.1+. However, do a little research, they is nothing else that has replaced it from the looks of it. I think it is safe to keep it in, it works. 

But I do think we should put our efforts into #3325. This work can be helped by abstracting as much of the code the blocks a blocks, out. Then implementing other interfaces like a new option in the context menu should be extremely simple.

assets/src/stories-editor/components/inserter/menu.js
swissspidy: This component an almost 1:1 copy of the InserterMenu component in `@wordpress/block-editor`. The inserter in the regular block editor is not only shown in the top toolbar, but also when using the appenders within a columns block or between paragraphs.

In our use case, the component is only used in the top toolbar though, and as such actually doesn't need the whole `destinationRootClientId` logic — the destination will always be the current page block.

So your changes look correct 👍

assets/src/stories-editor/helpers/index.js
swissspidy: I think it's fine to keep that as-is, there'd be only a little benefit.

Nice work on the tests!

swissspidy:
I don't think we should have deprecated ARIA stuff in the code. Especially because no assistive technology really supports it ([demo](https://codepen.io/aminimalanimal/pen/zravbo)).

> But I do think we should put our efforts into #3330.

> In order to implement accessible drag and drop, I'd prefer to defer it to #3325 and implement keyboard-based navigation and not bother with making the mouse-based navigation accessible.

Both sounds good to me.

assets/src/stories-editor/helpers/test/isBlockAllowedOnPage.js
```diff
@@ -0,0 +1,68 @@
...
+	it( 'should return false if block list does not contain element', () => {
+		mockCanInsertBlockType.mockImplementationOnce( () => true );
+		mockGetBlockListSettings.mockImplementationOnce( () => ( { allowedBlocks: [] } ) );
+
+		const result = isBlockAllowedOnPage( BLOCK_NAME, PAGE_ID );
+
+		expect( result ).toBe( false );
+	} );
+
+	it( 'should return true only iff element is allowed by both type and block list', () => {
```
swissspidy:
```suggestion
	it( 'should return true only if element is allowed by both type and block list', () => {
```

Typo or did you mean `if and only if`? 🙂

barklund: I did mean `iff`, yes.

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -76,10 +104,11 @@ class Draggable extends Component {
 	 * @param  {Object} event The non-custom DragEvent.
 	 */
 	onDragOver = ( event ) => {
+		const { onNeighborHover, blockName } = this.props;
 		const top = parseInt( this.cloneWrapper.style.top ) + event.clientY - this.cursorTop;
 
 		// Don't allow the CTA button to go over its top limit.
-		if ( 'amp/amp-story-cta' === this.props.blockName ) {
+		if ( isCTABlock( blockName ) ) {
```
swissspidy: It seems a bit confusing that this limitation is also in place when dragging over another page.

Could this be changed so that the limitation is lifted when dragging over a neighbor? Although this would mean there could be a jump when going back to the current page... So not sure if that would make it better.

spacedmonkey: Is this limitation because a if you might drag a CTA onto another page that already has one?

swissspidy: The height limitation is because CTA blocks are only allowed to be placed in the bottom part of a page.

swissspidy: Here's a demo video explaining the dragging situation with the CTA block: https://cloudup.com/cUthP9YKEst

spacedmonkey: Here is a screencast of me dragging between pages. 
https://www.youtube.com/watch?v=Jocbj5tNa8I

This behaviours as I would expect. One thing I noticed that when you hover over the first page, it doesn't show any sign the CTA to isn't going to dragged. Maybe make the border of that page red. Just an idea..

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -10,28 +10,76 @@ import PropTypes from 'prop-types';
...
+		hoverElement.current.pageId = newPageId;
+
+		// Unhighlight old highlighted page.
+		if ( hoverElement.current.element ) {
+			hoverElement.current.element.classList.remove( ...hoverElement.current.classes );
+		}
+
+		// Highlight neigboring page.
+		if ( newPageId ) {
+			const isLegal = offset === 0 || isBlockAllowedOnPage( blockName, newPageId );
```
spacedmonkey: Instead of `isLegal` why not `isAllowed`

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -61,13 +69,35 @@ class Draggable extends Component {
...
+		onNeighborHover( 0, true );
+
+		// Attempt drop on neighbor if offset
+		if ( this.pageOffset !== 0 ) {
+			// All this is about calculating the position of the (correct) element on the new page.
+			const currentElementTop = parseInt( this.cloneWrapper.style.top );
+			const currentElementLeft = parseInt( this.cloneWrapper.style.left );
+			const newLeft = currentElementLeft - ( this.pageOffset * PAGE_AND_MARGIN );
+			const blockIsCTA = isCTABlock( blockName );
+			const baseHeight = blockIsCTA ? STORY_PAGE_INNER_HEIGHT / 5 : STORY_PAGE_INNER_HEIGHT;
```
spacedmonkey: Instead of 3 `blockIsCTA` statement, why not why if statement?

swissspidy: Then you couldn't use `const`, no?

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -47,20 +95,85 @@ const BlockDraggable = ( { children, clientId, blockName, rootClientId, blockEle
...
+
+		/**
+		 * Drop this element on the page `offset` away from the current page (if possible). If dropped, update the position of the element on the new page.
+		 *
+		 * Currently this function removes the old element and creates a clone on the new page.
+		 *
+		 * @param {number} offset  Integer specifying offset from current page - e.g. -2 on page 3 will return id of page 1. Page indices are zero-based.
+		 * @param {Object} newAttributes  Object with attributes to update on element on new page.
+		 */
+		const onNeighborDrop = ( offset, newAttributes ) => {
```
spacedmonkey: Please move this into it's own file / helpers. It would be extremely useful to reuse this code for #3330

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -47,20 +95,85 @@ const BlockDraggable = ( { children, clientId, blockName, rootClientId, blockEle
...
+
+		/**
+		 * Get id of neighbor page that is `offset` away from the current page.
+		 *
+		 * If no page exists in that direction, null will be returned.
+		 *
+		 * @param {number} offset  Integer specifying offset from current page - e.g. -2 on page 3 will return id of page 1. Page indices are zero-based.
+		 * @return {string} Returns id of target page or null if no page exists there.
+		 */
+		const getNeighborPageId = ( offset ) => {
```
spacedmonkey: Please move this into it's own file / helpers. It would be extremely useful to reuse this code for #3330

assets/src/stories-editor/components/block-mover/draggable.js
spacedmonkey: They would change to lets yes.

assets/src/stories-editor/components/block-mover/block-draggable.js
spacedmonkey: This PR https://github.com/ampproject/amp-wp/pull/3417 relays on this functions. They should be moved.

assets/src/stories-editor/components/block-mover/block-draggable.js
spacedmonkey: This PR #3417 relays on this functions. They should be moved.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: Moved to a new helper hook in 1cde15c.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: Moved to a new helper hook in 1cde15c.

assets/src/stories-editor/components/block-mover/draggable.js
barklund: I reorganised the code in 1cde15c.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: Fixed in 1cde15c.

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -10,28 +10,108 @@ import PropTypes from 'prop-types';
...
+	 * Clear highlighting of pages.
+	 */
+	const clearHighlight = () => {
+		// Unhighlight old highlighted page.
+		if ( hoverElement.current.element ) {
+			hoverElement.current.element.classList.remove( ...hoverElement.current.classes );
+		}
+
+		// Reset current hover element
+		hoverElement.current = { pageId: null, element: null, classes: [] };
```
spacedmonkey: This clears out all classes. Are they not classes on the element we might want to keep?

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -10,28 +10,108 @@ import PropTypes from 'prop-types';
...
+
+	/**
+	 * Drop this element on the page `offset` away from the current page (if possible). If dropped, update the position of the element on the new page.
+	 *
+	 * Currently this function removes the old element and creates a clone on the new page.
+	 *
+	 * @param {number} offset  Integer specifying offset from current page - e.g. -2 on page 3 will return id of page 1. Offset must be non-zero.
+	 * @param {Object} newAttributes  Object with attributes to update on element on new page.
+	 */
+	const dropElementByOffset = ( offset, newAttributes ) => {
```
spacedmonkey: This is super clean :+1:

assets/src/stories-editor/components/block-mover/draggable.js
```diff
@@ -61,13 +69,41 @@ class Draggable extends Component {
...
+		// Attempt drop on neighbor if offset
+		if ( this.pageOffset !== 0 ) {
+			// All this is about calculating the position of the (correct) element on the new page.
+			const currentElementTop = parseInt( this.cloneWrapper.style.top );
+			const currentElementLeft = parseInt( this.cloneWrapper.style.left );
+			const newLeft = currentElementLeft - ( this.pageOffset * PAGE_AND_MARGIN );
+
+			let baseHeight, xAttribute, yAttribute;
+			if ( isCTABlock( blockName ) ) {
+				baseHeight = STORY_PAGE_INNER_HEIGHT / 5;
```
spacedmonkey: What is the / 5 do here. If this should it's own constant on this line needs to be documented.

assets/src/stories-editor/helpers/isBlockAllowedOnPage.js
```diff
@@ -0,0 +1,24 @@
...
+} = select( 'core/block-editor' );
+
+/**
+ * Is the given block allowed on the given page?
+ *
+ * @param {Object}  name The name of the block to test.
+ * @param {string}  pageId Page ID.
+ * @return {boolean} Returns true if the element is allowed on the page, false otherwise.
+ */
+export default ( name, pageId ) => {
```
spacedmonkey: Should this be a custom hook similar to `useMoveBlockToPage`, as you are using a state data?

assets/src/stories-editor/helpers/useMoveBlockToPage.js
```diff
@@ -0,0 +1,69 @@
+/**
```
spacedmonkey: I love this use of hooks :+1: Good work

assets/src/stories-editor/components/block-mover/block-draggable.js
```diff
@@ -10,28 +10,108 @@ import PropTypes from 'prop-types';
...
+			return;
+		}
+
+		clearHighlight();
+
+		// Highlight page and mark whether drop is allowed or not.
+		if ( pageId ) {
+			// Drop is always allowed on the initial page (offset=0)
+			const isAllowed = offset === 0 || isBlockAllowedOnPage( blockName, pageId );
+			const classes = [
```
spacedmonkey: Should this use / import the `classnames` function?

assets/src/stories-editor/helpers/useMoveBlockToPage.js
```diff
@@ -0,0 +1,69 @@
...
+ * Hook that exposes functions relevant for moving an element to a neigboring page.
+ *
+ * @param {string} blockId  Id of block to move
+ * @return {Object} Returns two functions: `getPageByOffset` and `moveBlockToPage`.
+ */
+const useMoveBlockToPage = ( blockId ) => {
+	const pages = useSelect( ( select ) => select( 'core/block-editor' ).getBlockOrder() );
+	const block = useSelect( ( select ) => select( 'core/block-editor' ).getBlock( blockId ) );
+	const currentPageId = useSelect( ( select ) => select( 'amp/story' ).getCurrentPage() );
```
spacedmonkey: I think it would be clearer to do 
```
const {
		pages,
		block,
		currentPageId,
} = useSelect(
		( select ) => {
			const {
				getBlockOrder,
				getBlock,
			} = select( 'core/block-editor' );
			const { getCurrentPage } = select( 'amp/story' );
			return {
				pages: getBlockOrder(),
	                        block: getBlock( blockId ),
		                currentPageId: getCurrentPage()
			};
		}, [ blockId ]
      );
```

spacedmonkey: See example from https://github.com/ampproject/amp-wp/blob/8ae3c05f0f525ed85305527777d5ef07d40a3ae0/assets/src/stories-editor/blocks/amp-story-page/copy-paste-handler.js#L19-L37

assets/src/stories-editor/helpers/useMoveBlockToPage.js
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * Move the element to the given page. Also update the element with the given properties.
+	 *
+	 * Currently this function removes the old element and creates a clone on the new page.
+	 *
+	 * @param {string} pageId  Id of page to move element to
+	 * @param {Object} attributes  Object with attributes to update on element on new page.
+	 */
+	const moveBlockToPage = ( pageId, attributes ) => {
```
spacedmonkey: Work noting that for https://github.com/ampproject/amp-wp/pull/3417 , I will need to tweak this function. I would do something like this. 

```
const moveBlockToPage = ( pageId, attributes = null ) => {
```

```
if( null !== attributes ) {
     updateBlockAttributes( clonedBlock.clientId, attributes );
}
```
Updating attributes are not necessary when moving a block via context menu.

assets/src/stories-editor/components/block-mover/draggable.js
swissspidy: Agreed.

The same logic is already used a bit below:

```js
// 20% of the full value in case of CTA block.
const baseHeight = blockIsCTA ? STORY_PAGE_INNER_HEIGHT / 5 : STORY_PAGE_INNER_HEIGHT;
```

So extracting it makes sense.

The reason is that the CTA layer in an AMP story can only be placed i n the lower 20% of the page, hence the division by 5.

spacedmonkey: So if this logic is repeated then it should be a const for sure. It will make it easier refactor later. 

So maybe `STORY_PAGE_INNER_HEIGHT_FOR_CTA` as a name?

assets/src/stories-editor/helpers/useMoveBlockToPage.js
barklund: Ah of course, forgot that part. Will do!

assets/src/stories-editor/helpers/useMoveBlockToPage.js
barklund: Clearer how? I don't think it's any more clear.

I will add the dependencies to the `useSelect`'s though - forgot it took those.

But I actually don't think that the more elaborate single `useSelect` is an improvement. With the dependencies it will actually also be less optimal, as two of the variables don't need to update just because `clientId` does.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: I did that initially, but as I need an array and not a merged string (`classList.add/.remove` doesn't accept `("foo bar")`, only `("foo", "bar")`), I would need to split it anyway - and then `classNames` was doing nothing essentially.

assets/src/stories-editor/helpers/isBlockAllowedOnPage.js
barklund: Yes, you're right. It should. It would be beneficial to update all helpers using `select` or `dispatch` to be hooks in order for them to properly update components that use them. I will update this one.

assets/src/stories-editor/components/block-mover/block-draggable.js
barklund: This doesn't actually manipulate any element. The `classes` property here only remembers which classes were set on the last `element` in use in order to remove them again.

assets/src/stories-editor/helpers/isBlockAllowedOnPage.js
barklund: Argh, it can't easily be as it's used in a class component, `InserterMenu`. And that's a pretty complex component I don't feel like rewriting right now.

I will split it into two copies and deprecate the non-hook one.

I don't know if there's a better transition method, I'm open for suggestions.

assets/src/stories-editor/helpers/useMoveBlockToPage.js
spacedmonkey: It is easier to read as you don't repeat `useSelect( ( select ) => select(`. Long lines of code are harder to scan. That is why it was done that way in the other code snippet I sent over. 

It is just also how I have seen hooks used elsewhere in other projects. I don't feel super strongly about this one, functionality the code is fine, just don't like the style of it.

assets/src/stories-editor/helpers/isBlockAllowedOnPage.js
spacedmonkey: There are 15 selects and 11 dispatches in the helper functions. These should all be convert to custom hooks. We should follow up with another ticket.

spacedmonkey: > Argh, it can't easily be as it's used in a class component, InserterMenu. And that's a pretty complex component I don't feel like rewriting right now.

I think this is related to what you are saying. https://github.com/ampproject/amp-wp/pull/3412 . Might be useful.

swissspidy: Not all of them are used in actual components though FWIW

barklund: Attempted fixed in e963844.

assets/src/stories-editor/helpers/useMoveBlockToPage.js
barklund: Added dependencies in 3241e3b.

assets/src/stories-editor/components/block-mover/draggable.js
barklund: Actual placement of the element is done in CSS:

https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/blocks/amp-story-cta/edit.css#L1-L6

I have moved this to a new constant in 005bac0.

assets/src/stories-editor/helpers/useMoveBlockToPage.js
barklund: Done in 354ee52 and added unit tests as well!

assets/src/stories-editor/helpers/useMoveBlockToPage.js
spacedmonkey: As I don't feel super strongly about this, I am going to mark this as resolved.

swissspidy:
@spacedmonkey Could you re-review this?

assets/src/stories-editor/components/inserter/menu.js
```diff
@@ -52,6 +52,7 @@ import { addQueryArgs } from '@wordpress/url';
 import BlockPreview from '../block-preview';
 import BlockTypesList from '../block-types-list';
 import { ALLOWED_TOP_LEVEL_BLOCKS } from '../../constants';
+import { isBlockAllowedOnPage } from '../../helpers';
```
spacedmonkey: Why can't the hook be used here?

assets/src/stories-editor/components/media-inserter/index.js
```diff
@@ -12,10 +12,11 @@ import { withDispatch, withSelect } from '@wordpress/data';
 import { DropdownMenu } from '@wordpress/components';
 import { compose, ifCondition } from '@wordpress/compose';
 import { __, sprintf } from '@wordpress/i18n';
+
 /**
  * Internal dependencies
  */
-import { processMedia } from '../../helpers';
+import { processMedia, isBlockAllowedOnPage } from '../../helpers';
```
spacedmonkey: Why isn't the hook used here?

swissspidy: This component hasn't been migrated to use hooks yet.

assets/src/stories-editor/components/inserter/menu.js
swissspidy: Because this is still a class component, not a functional one. See #2538 for explanation.

spacedmonkey: I am just worried about bug coming from not using select correctly here and getting stale data.

spacedmonkey:
This bug is still happening for me on the latest update. 
See screencast. https://www.youtube.com/watch?v=0ZQDFQVPRYw&feature=youtu.be

assets/src/stories-editor/components/inserter/menu.js
swissspidy: I'll take a stab at it 👍

swissspidy:
> This bug is still happening for me on the latest update.

Hmm that seems to be because the drop zone around a page is a bit bigger than the page itself, to allow for better positioning of elements.

assets/src/stories-editor/components/media-inserter/index.js
swissspidy: I'll take a stab at rewriting it 👍

assets/src/stories-editor/components/inserter/menu.js
spacedmonkey: Isn't the benefit of hooks if that you put them everyone and mix and match?

assets/src/stories-editor/components/media-inserter/index.js
swissspidy: Addressed in 144c430993400716171afc71b47df7bdfddfbc30

assets/src/stories-editor/components/media-inserter/index.js
```diff
@@ -1,31 +1,71 @@
-/**
```
spacedmonkey: :+1:

assets/src/stories-editor/components/media-inserter/index.js
spacedmonkey: :+1:

assets/src/stories-editor/components/inserter/menu.js
spacedmonkey: How that `isBlockAllowedOnPage` is only used in `Menu`, I think we should get rid of the helper and move the logic to `withSelect`

barklund:
> This bug is still happening for me on the latest update.
> See screencast. https://www.youtube.com/watch?v=0ZQDFQVPRYw&feature=youtu.be

So, the thing is, you're allowed to drop items off-page if you feel like it – but only within a certain boundary off-page. The general item drop is handled in `story-block-drop-zone`, and I actually think the size of the boundary is determined by the size of an element, thus via CSS - but not sure about that.

However, in this case it gets extra weird, because off-page is technically on top of another page, that you just can't move the item to. I don't know if there's an easy way to handle this, that doesn't impact too many other things and doesn't feel like a workaround and special-case-handling.

I'll look into this drop zone, and maybe limit the drop zone so it never overlaps a neighbouring page - seems like a fair restriction.

assets/src/stories-editor/components/media-inserter/index.js
```diff
@@ -114,56 +145,14 @@ const mediaPicker = ( dialogTitle, mediaType, updateBlock ) => {
...
-			updateBlockAttributes( clientId, processed );
-			selectBlock( clientId );
-		},
 	};
 } );
 
 export default compose(
 	applyWithSelect,
-	applyWithDispatch,
 	ifCondition( ( { isReordering } ) => ! isReordering ),
```
barklund: What's this weird thing doing here? Can we get rid of it and then get rid of the whole `compose` thing? What even is this?

swissspidy: `ifCondition` is a HOC that only renders the component if the given condition is satisfied. Basically a shortcut for doing `if ( isReordering ) { return null; }` in the component.

swissspidy: https://github.com/WordPress/gutenberg/blob/f68a107c777169cd80b3ce56ec4f7c8360f2178c/packages/compose/src/higher-order/if-condition/README.md

assets/src/stories-editor/components/inserter/menu.js
swissspidy: That could work yeah 👍

assets/src/stories-editor/components/media-inserter/index.js
swissspidy: And yes, we can get rid of it

swissspidy:
Given that the situation with the drop zone mentioned above has been around before, it can be addressed in a separate PR

PR #3320, opened 2019-09-21
Added dark color scheme

multiformeingegno:
It activates when the OS/browser supports it and "dark" is the color-scheme declared by the browser.

Background color and text pass WCAG AA/AAA accessibility test.

### DIFF
--- a/templates/style.php
+++ b/templates/style.php
@@ -414,3 +414,31 @@ amp-carousel > amp-img > img {
 	position: absolute;
 	<?php echo is_rtl() ? 'left' : 'right'; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>: 16px;
 }
+
+/* Dark Theme */
+
+:root {
+	color-scheme: light dark;
+}
+
+@media (prefers-color-scheme: dark) {
+	body,
+	.amp-wp-article {
+		background: #202020;
+		color: #fff;
+	}
+
+	.amp-wp-tax-category,
+	.amp-wp-tax-tag,
+	.amp-wp-meta,
+	span.amp-site-title,
+	.amp-wp-title {
+		color: #fff;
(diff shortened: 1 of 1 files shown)


multiformeingegno:
@googlebot I signed it!

templates/style.php
```diff
@@ -414,3 +414,31 @@
...
+	.amp-wp-title {
+		color: #fff;
+	}
+
+	a,
+	a:visited,
+	.amp-wp-comments-link a {
+		color: #93B3DB;
+	}
+}
```
westonruter: The build is failing because the file doesn't end in a newline: https://travis-ci.org/ampproject/amp-wp/jobs/587871984#L782-L783

multiformeingegno: Pushed another commit to add the newline

swissspidy:
In the Customizer we already have the option to switch between light and dark modes:

https://github.com/ampproject/amp-wp/blob/b44ac6ac687008e4d1c5dd2ef78b2a5d3d9ecd23/includes/settings/class-amp-customizer-design-settings.php#L321-L338

How does this relate to the browser's dark mode preference?

At the very least, I'd say this change should use the same dark colors as the Customizer.

Perhaps we should add a checkbox like "Automatically use Dark Mode if requested by the users" to the Customizer for this? 

Or remove the color scheme option entirely from the customizer?

westonruter:
> Perhaps we should add a checkbox like "Automatically use Dark Mode if requested by the users" to the Customizer for this?

I think this is best.

swissspidy:
@multiformeingegno Would you be up for helping extend the customizer option as mentioned above?

westonruter:
Closing as @amedina and I are going to propose a new Reader mode direction that will rely on existing AMP-compatible themes.

Issue #3321, opened 2019-09-22
Markup prepended via the_content filter causes amp-story to be removed from page


honeyrbhandari:
Hi,

Please help me. The amp stories are not working in my WP site. Here is one of the story, I created - https://wisdom2shine.com/stories/listen-to-your-true-calling/ It goes into a blank screen. It was working fine earlier. However, now it fails to load and shows the following error message.

![image](https://user-images.githubusercontent.com/55599364/65385831-e031e800-dd50-11e9-8a5e-888d0baaffce.png)

Even when I try to edit it, it does not allow me to do so because of the following validation error mesage:
![image](https://user-images.githubusercontent.com/55599364/65385912-23408b00-dd52-11e9-9439-4e61e2c99d71.png)

Please help as I need to create content in the form of stories and I am struggling hard to get it resolved :(...

Will look forward to a resolution. 
Best Regards,
Honey

westonruter:
The rejected validation error for the `script` is likely the problem. Switch that to accepted. 

We need to prevent this endless redirect from happening.

I'm confused why a `script` is getting added to the Story template in the first place, however. Please expand the details for both of those validation errors.

westonruter:
The error for `amp-story` is also going to be a problem, because it will show the story as blank when removed. So yeah, pleas click details for each and share screenshots.

honeyrbhandari:
Hi,

Thanks a lot for your prompt response! Here are the screenshots:
![image](https://user-images.githubusercontent.com/55599364/65447859-1a84ad80-de55-11e9-9575-2711714f390c.png)

![image](https://user-images.githubusercontent.com/55599364/65447918-4142e400-de55-11e9-8cf7-22f71b0eb025.png)

Best Regards,
Honey

westonruter:
@honeyrbhandari Do you know which plugin is responsible for adding the `WebFontConfig`? In any case, that validation error should be marked as “Accepted” because the AMP Stories editor handles the fonts already, and custom JS is not allowed.

The second validation error regarding `amp-story` is more confusing. Can you share a screenshot of the editor for that story? 

Please also do this:

1. Mark that `amp-story` error as “rejected”.
2. Access the URL of the story but add `?amp_validate` to the URL. This will prevent that endless redirect bug.
3. Share the HTML source of the the page (via browser “View Page Source”). You can paste that into a [Gist](https://gist.github.com/).

My suspicion is that

westonruter:
[...]

PR #3329, opened 2019-09-23
Improve handling of Stories in Compatibility Tool

westonruter:
In #3321 it was discovered that rejecting the sanitization of a validation error on an AMP Story results in an infinite redirect when a site is not in Standard mode. When processing a Story, it should behave as if the site is in Standard mode. No redirection should be performed when there are rejected validation errors, but rather the `amp` attribute should be removed from the page to prevent it from being validated as a valid AMP page.

In the same way, when a Story is listed among the Validated URLs in the admin, there should be no `?amp` appearing in the displayed URLs.

Testing instructions:

1. Set mode to Transitional.
2. Create a Story and add a Custom HTML block with some invalid AMP markup, like `<script>bad</script>`
3. Publish Story.
4. A warning message should appear about validation errors. Click “Review Issues”.
5. With this ifx applied, the Validated URL screen should not show `?amp` in the heading.
5. On the Validated URL screen, set the validation error status as “Rejected” instead of “New Accepted”. Press Update.
6. Try navigating to the permalink of the post, and the Story should render but not as a valid AMP page; the Custom HTML markup should appear in the page, and the `html` element should have the `amp` attribute omitted.

# Before

image

Viewing the story:

image

# After

image

Viewing the story:

image

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -2115,7 +2115,7 @@ class AMP_Theme_Support {
 				AMP_HTTP::send_server_timing( 'amp_processor_cache_hit', -$prepare_response_start );
 
 				// Redirect to non-AMP version.
-				if ( ! amp_is_canonical() && $blocking_error_count > 0 ) {
+				if ( ! amp_is_canonical() && ! is_singular( AMP_Story_Post_Type::POST_TYPE_SLUG ) && $blocking_error_count > 0 ) {
 					if ( AMP_Validation_Manager::has_cap() ) {
 						$non_amp_url = add_query_arg( AMP_Validation_Manager::VALIDATION_ERRORS_QUERY_VAR, $blocking_error_count, $non_amp_url );
 					}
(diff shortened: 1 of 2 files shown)


kienstra:
**Approved**

Hi @westonruter,
This looks good! Great debugging on #3321.

Like you mentioned, I also saw that there were too many redirects when a validation error for a story was rejected:

too-many-redirects

But with this PR, the AMP story with the rejected error appears as expected:
amp-story-without-error

swissspidy:
See issue description above for testing instructions

csossi:
Warning message isn't appearing after "Publish"
image

But if I leave the story and return, then it appears:
image

westonruter:
@csossi I was seeing this as well. However, I just tried in the latest 6.6.0-RC1 of Gutenberg, and I'm not seeing it. When creating a post or a story with invalid markup, it is showing me the warning upon the first save:

image

image

swissspidy:
Verified this as per the testing instructions 👍

PR #3334, opened 2019-09-23
Ensure all AMP scripts (including v0.js) get moved to the head

westonruter:
When using a Latest Stories block on a page, this results on `amp-carousel` and `amp-runtime` scripts being enqueued on the page, and printed in the footer. While AMP component scripts are getting moved to the `head`, this was not being done for `v0.js`. So this ensures that no validation error from happening regarding `v0.js` being in the `body`:

image

Steps to reproduce:

1. Create a story.
2. Create a page containing a Latest Stories block.
3. View the page as AMP.
4. Without the fix, a validation error should be reported (above); with the fix, no error should be reported.

### DIFF
--- a/tests/php/test-class-amp-theme-support.php
+++ b/tests/php/test-class-amp-theme-support.php
@@ -1352,6 +1352,42 @@ class Test_AMP_Theme_Support extends WP_UnitTestCase {
 	}
 
 	/**
+	 * Test moving AMP scripts from body to head.
+	 *
+	 * @covers AMP_Theme_Support::prepare_response()
+	 * @covers AMP_Theme_Support::ensure_required_markup()
+	 */
+	public function test_scripts_get_moved_to_head() {
+		ob_start();
+		?>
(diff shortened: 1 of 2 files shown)


tests/php/test-class-amp-theme-support.php
```diff
@@ -1351,6 +1351,42 @@ public function test_ensure_required_markup_schemaorg( $script, $expected ) {
 		$this->assertEquals( $expected, substr_count( $dom->saveHTML(), 'schema.org' ) );
 	}
 
+	/**
+	 * Test moving AMP scripts from body to head.
+	 *
+	 * @covers AMP_Theme_Support::prepare_response()
+	 * @covers AMP_Theme_Support::ensure_required_markup()
+	 */
+	public function test_scripts_get_moved_to_head() {
```
kienstra: Nice testing of this!

swissspidy:
Verified in QA

Issue #1420, opened 2018-09-10
Expose reason for why a validation error is raised


westonruter:
When there is a validation error causing an element such as the following to be removed:

```php
<link rel="manifest" href="http://example.com/wp-json/app/v1/web-manifest">
```

There is no corresponding reason _why_ it has been removed. The developer currently has to figure it out on their own. This us unfortunate because the plugin has the reason as part of the validation logic. So in the above example, the validation error is raised due to an illegal protocol being used. The validation error would be detected here:

https://github.com/Automattic/amp-wp/blob/b3e1b4d07dda70c5070d30f7426f849eb3a39823/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L983-L985

And this is due to the spec mandating the `https` protocol for `manifest` links:

https://github.com/Automattic/amp-wp/blob/b3e1b4d07dda70c5070d30f7426f849eb3a39823/includes/sanitizers/class-amp-allowed-tags-generated.php#L7111-L7119

We need to surface this reason when a validation error happens.

We should be reusing the same error codes (and [messages](https://github.com/ampproject/amphtml/blob/5348681bd1a75916bed605495538dd996348133b/validator/validator-main.protoascii#L7061-L7550)) that the validator itself uses.

Compare/contrast with what the plugin is using today:

https://github.com/ampproject/amp-wp/blob/980375ca2eaabad974b1602ab0291645eb0e0f78/includes/validation/class-amp-validation-error-taxonomy.php#L2872-L2949

# Validation Error Codes

Status | Code
---------|------
? | ATTR_DISALLOWED_BY_IMPLIED_LAYOUT
? | ATTR_DISALLOWED_BY_SPECIFIED_LAYOUT
? | ATTR_MISSING_REQUIRED_EXTENSION
? | ATTR_REQUIRED_BUT_MISSING
? | ATTR_VALUE_REQUIRED_BY_LAYOUT
? | BASE_TAG_MUST_PRECEED_ALL_URLS
? | CDATA_VIOLATES_BLACKLIST
? | CHILD_TAG_DOES_NOT_SATISFY_REFERENCE_POINT
? | CHILD_TAG_DOES_NOT_SATISFY_REFERENCE_POINT_SINGULAR
? | CSS_EXCESSIVELY_NESTED
? | CSS_SYNTAX_BAD_URL
? | CSS_SYNTAX_DISALLOWED_DOMAIN
[...]

PR #3336, opened 2019-09-24
Sanitize invalid children of amp-story and amp-story-page elements to prevent white story of death

### LINKED ISSUE #3321: Markup prepended via the_content filter causes amp-story to be removed from page

westonruter:
A compatibility issue was discovered in #3321 with the [Reading Time WP](https://wordpress.org/plugins/reading-time-wp/) plugin, but it is likely going to happen with other plugins as well. The Reading Time WP plugin filters `the_content` to inject this at the beginning:

```html
<span class="rt-reading-time" style="display: block;"><span class="rt-label">Reading Time: </span> <span class="rt-time">1</span> <span class="rt-label rt-postfix">minute</span></span>
```

This results in invalid `amp-story` which [restricts its children](https://github.com/ampproject/amphtml/blob/1c1e722ad99fc5352e5519b28cdb0db99854e2fd/extensions/amp-story/validator-amp-story.protoascii#L100-L111) to elements like `amp-story-page`. When this `span` is a direct child and the `child_tag_name_oneof` constraint is violated, the result is the entire `amp-story` being invalid and a white story of death (where the `body` has no children). The validation error is not helpful at all:

image

This problem was actually “prophesied” in #2926:

> 👉 A side-effect of the change here is the sanitization of AMP components with invalid children will be more draconian. For example, if an `amp-story` has an invalid child element, then the entire `amp-story` element will be removed, as opposed to the invalid child alone being removed. Nevertheless, since only AMP components have such validation constraints for children, it should not so common for this to occur in WordPress sites that are just outputting normal HTML. It only will become an issue when starting to use AMP components, and this makes https://github.com/ampproject/amp-wp/issues/1420 much more important as it will be needed to explain _why_ the element was removed.

So this PR fixes the problem by extending the `AMP_Story_Sanitizer` to preemptively remove AMP Story elements under `amp-story` and `amp-story-page` which are invalid. These are the two elements which have the `child_tag_name_oneof` constraint. This special case sanitizer is especially important for AMP Stories since all of the markup for a story is in `post_content` and is prone to be mutated with `the_content` filters to add elements like word counts, sharing buttons, and related posts. This PR prevents such elements from being seen by the tag-and-attribute sanitizer, thus preventing the `amp-story` and `amp-story-page` as a whole from being removed.

In the case of the `span` which the Reading Time WP plugin adds to `the_content`, the validation error now becomes much more helpful:

image

And no white story of death occurs.

Fixes #3321.

### DIFF
--- a/includes/sanitizers/class-amp-story-sanitizer.php
+++ b/includes/sanitizers/class-amp-story-sanitizer.php
@@ -21,18 +21,14 @@ class AMP_Story_Sanitizer extends AMP_Base_Sanitizer {
 	public static $tag = 'amp-story-page';
 
 	/**
-	 * AMP Story tag spec.
+	 * Allowed children for given tags.
 	 *
 	 * @var array
 	 */
-	private $amp_story_tag_spec;
-
-	/**
-	 * AMP Story Page tag spec.
-	 *
-	 * @var array
-	 */
-	private $amp_story_page_tag_spec;
+	private $tag_allowed_children = [
+		'amp-story'      => [],
+		'amp-story-page' => [],
+	];
 
 	/**
 	 * Sanitize the AMP elements contained by <amp-story-page> element where necessary.
@@ -40,8 +36,23 @@ class AMP_Story_Sanitizer extends AMP_Base_Sanitizer {
 	 * @since 0.2
 	 */
 	public function sanitize() {
-		$this->amp_story_tag_spec      = AMP_Allowed_Tags_Generated::get_allowed_tag( 'amp-story' )[0];
-		$this->amp_story_page_tag_spec = AMP_Allowed_Tags_Generated::get_allowed_tag( 'amp-story-page' )[0];
+
+		// Gather the allowed children for amp-story and amp-story-page elements.
+		foreach ( array_keys( $this->tag_allowed_children ) as $tag_name ) {
+			$rule_specs = AMP_Allowed_Tags_Generated::get_allowed_tag( $tag_name );
+			if ( ! isset( $rule_specs ) ) {
+				continue;
+			}
+			foreach ( $rule_specs as $rule_spec ) {
+				if ( ! isset( $rule_spec['tag_spec']['child_tags']['child_tag_name_oneof'] ) ) {
+					continue;
+				}
+				$this->tag_allowed_children[ $tag_name ] = array_merge(
+					$this->tag_allowed_children[ $tag_name ],
+					$rule_spec['tag_spec']['child_tags']['child_tag_name_oneof']
+				);
+			}
+		}
 
 		$amp_story_element = $this->dom->getElementsByTagName( 'amp-story' )->item( 0 );
 		if ( $amp_story_element instanceof DOMElement ) {
@@ -64,8 +75,8 @@ class AMP_Story_Sanitizer extends AMP_Base_Sanitizer {
 				if ( 'amp-story-page' === $node->nodeName ) {
 					$page_number++;
 					$this->sanitize_story_page_element( $node, $page_number );
-				} elseif ( ! in_array( $node->nodeName, $this->amp_story_tag_spec['tag_spec']['child_tags']['child_tag_name_oneof'], true ) ) {
-					$this->remove_invalid_child( $node );
+				} else {
+					$this->maybe_remove_disallowed_child( $node );
 				}
 			}
 			$node = $next_node;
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-story-sanitizer.php
```diff
@@ -20,50 +20,89 @@ class AMP_Story_Sanitizer extends AMP_Base_Sanitizer {
...
+	 */
+	private function sanitize_story_element( DOMElement $element ) {
+		$page_number = 0;
 
-			if ( ! $node ) {
-				continue;
+		$node = $element->firstChild;
+		while ( $node ) {
+			$next_node = $node->nextSibling;
+			if ( $node instanceof DOMElement ) {
```
schlessera: Same logic inversion here, I would prefer an early return (`continue` in this case) instead of making the main logic look like an edge case.

includes/sanitizers/class-amp-story-sanitizer.php
```diff
@@ -20,50 +20,89 @@ class AMP_Story_Sanitizer extends AMP_Base_Sanitizer {
...
+	private function sanitize_story_page_element( DOMElement $element, $page_number ) {
+		$cta_layer_count = 0;
 
-				if ( $j > 0 ) {
-					// There can only be one CTA layer.
-					$cta_layer_node->parentNode->removeChild( $cta_layer_node );
+		$node = $element->firstChild;
+		while ( $node ) {
+			$next_node = $node->nextSibling;
+			if ( $node instanceof DOMElement ) {
```
schlessera: I would also prefer an early return/continue here instead.

tests/php/test-amp-story-sanitizer.php
```diff
@@ -46,6 +46,14 @@ public function get_data() {
...
 			],
+			'story_with_invalid_root_elements' => [
+				'<p>Word count: 4</p><amp-story-page><amp-story-grid-layer><p>Lorem Ipsum Demet Delorit.</p></amp-story-grid-layer></amp-story-page><p>Related posts: <a href="https://example.com/"><strong>Example</strong></a></p>',
+				'<amp-story-page><amp-story-grid-layer><p>Lorem Ipsum Demet Delorit.</p></amp-story-grid-layer></amp-story-page>',
+			],
+			'story_with_invalid_layer_siblings' => [
+				'<amp-story-page><p>Before layer</p><amp-story-grid-layer><p>Lorem Ipsum Demet Delorit.</p></amp-story-grid-layer><p>After layer</p></amp-story-page</p>',
+				'<amp-story-page><amp-story-grid-layer><p>Lorem Ipsum Demet Delorit.</p></amp-story-grid-layer></amp-story-page>',
+			],
 		];
```
schlessera: Test for CTA removal is missing...?

includes/sanitizers/class-amp-story-sanitizer.php
```diff
@@ -20,50 +20,89 @@ class AMP_Story_Sanitizer extends AMP_Base_Sanitizer {
...
+
 	/**
 	 * Sanitize the AMP elements contained by <amp-story-page> element where necessary.
 	 *
 	 * @since 0.2
 	 */
 	public function sanitize() {
-		$nodes     = $this->dom->getElementsByTagName( self::$tag );
-		$num_nodes = $nodes->length;
+		$this->amp_story_tag_spec      = AMP_Allowed_Tags_Generated::get_allowed_tag( 'amp-story' )[0];
```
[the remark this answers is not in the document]
westonruter: That's correct. If we update the Validator spec and it results in these tag specs being `null` then we'd catch it in the unit test.

includes/sanitizers/class-amp-story-sanitizer.php
westonruter: True, but the reason why I did it this way was because of `$node = $next_node` needing to run below. Otherwise, I'd have added:

```php
    $node = $next_node;
    continue;
}
```

But that seems worse because the logic is duplicated.

includes/sanitizers/class-amp-story-sanitizer.php
westonruter: See reasoning above.

tests/php/test-amp-story-sanitizer.php
westonruter: That should be covered above by `story_with_cta_on_first_page` and `story_with_multiple_cta_on_second_page`.

includes/sanitizers/class-amp-story-sanitizer.php
schlessera: How about something like this:
```php
$node = $element->firstChild;
do {
	$next_node = $node->nextSibling;
		continue;
	}
	if ( 'amp-story-page' === $node->nodeName ) {
		$page_number++;
		$this->sanitize_story_page_element( $node, $page_number );
	} elseif ( ! in_array( $node->nodeName, $this->amp_story_tag_spec['tag_spec']['child_tags']['child_tag_name_oneof'], true ) ) {
		$this->remove_invalid_child( $node );
	}
} while ( $node = $next_node );
```

schlessera: Note: This is mostly just preference here. I'll approve the changes and let you decide whether you want to make changes or not.

westonruter: Thanks, I like that. However, I tried it and then there is a PHPCS compliant: `WordPress.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition`. We can revisit later.

includes/sanitizers/class-amp-story-sanitizer.php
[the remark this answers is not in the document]
westonruter: @schlessera I hardened this in 685cfa13.

includes/sanitizers/class-amp-story-sanitizer.php
```diff
@@ -21,27 +21,38 @@ class AMP_Story_Sanitizer extends AMP_Base_Sanitizer {
...
-		$this->amp_story_tag_spec      = AMP_Allowed_Tags_Generated::get_allowed_tag( 'amp-story' )[0];
-		$this->amp_story_page_tag_spec = AMP_Allowed_Tags_Generated::get_allowed_tag( 'amp-story-page' )[0];
+
+		// Gather the allowed children for amp-story and amp-story-page elements.
+		foreach ( array_keys( $this->tag_allowed_children ) as $tag_name ) {
+			$rule_specs = AMP_Allowed_Tags_Generated::get_allowed_tag( $tag_name );
+			if ( ! isset( $rule_specs ) ) {
+				continue;
+			}
+			foreach ( $rule_specs as $rule_spec ) {
```
westonruter: Note that the `$rule_specs` array only has one item in it.

PR #3338, opened 2019-09-24
Add PHP 7.4 to build matrix

swissspidy:
This lets us catch potential compatibility issues in advance while not failing the whole build.


swissspidy:
Looks like this even caught one faulty test already:

```
1) AMP_Story_Post_Type_Test::test_override_story_embed_callback
Trying to access array offset on value of type int

/tmp/wordpress/src/wp-content/plugins/amp/tests/php/test-class-amp-story-post-type.php:402
/tmp/wordpress/src/wp-content/plugins/amp/tests/php/test-class-amp-story-post-type.php:222
```

Issue #2084, opened 2019-04-08
Add ability to opt-in to a beta release channel


westonruter:
In order to facilitate testing beta versions of the plugin, we should consider allowing a user to opt-in to a beta channel which would pull updates from pre-releases on GitHub.

swissspidy:
Downloading from GitHub is against the [plugin directory guidelines](https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/#8-plugins-may-not-send-executable-code-via-third-party-systems). Quote:

> Executing outside code within a plugin when not acting as a service is not allowed, for example:
> 
>  * Serving updates or otherwise installing plugins, themes, or add-ons from servers other than WordPress.org’s

We should probably just use `trunk` for pre-releases, which is commonly done anyway. Plus, this way people can already translate pre-release versions, which is quite useful too.

westonruter:
Good point. But we could have the plugin updater configured to pull from `trunk` on WordPress.org? Just wanting to avoid manual ZIP file wrangling.

swissspidy:
We'd have to check what http://api.wordpress.org/plugins/update-check/1.1/ supports.

Most probably we'd just have to filter some transients to point to the trunk ZIP file, and then probably provide a way to downgrade.

westonruter:
I recently used the [Jetpack Beta Tester](https://jetpack.com/download-jetpack-beta/) plugin which was really slick, but it is perhaps overkill for the AMP plugin. It allows selecting a feature branch from which to automatically install build.

jdelia:
Are they any instructions on how we should update the plugin on our installs when testing the beta and RC versions? So far, I've just been deleting the existing plugin, installing the new from the built zip and then renaming the plugin folder to `amp`.

swissspidy:
At the moment that's the recommended and most common approach. Although there should be no need to rename the plugin because a) you deleted the old one and b) the new one should be named `amp` already.

We want to make this process easier though, which this issue here is for.

jdelia:
Thank you. The beta and RC zips when installed are not creating an `amp` folder. You'll see this message in the dashboard for example: `You appear to have installed the AMP plugin incorrectly. It is currently installed in the amp-wp-1.2.1-RC1-built directory, but it needs to be placed in a directory named amp. Please rename the directory. This is important for WordPress plugin auto-updates.`

swissspidy:
[...]

PR #3340, opened 2019-09-24
Remove unused Travis deploy stage

westonruter:
We're currently not using the `ampconfdemo` environment for testing, which is what the current `deploy` stage is targeting. In any case, the limit of multidev environments on that site has been overrun doing deploys when pushing to the 1.3 branch currently causes a [failure](https://travis-ci.org/ampproject/amp-wp/builds/588979057):

image

So we should just eliminate the `deploy` stage and re-focus on getting automatic deployments to a testing environment set up. See #1840, #2084.


schlessera:
The file `./bin/deploy-travis-pantheon.sh` is probably not needed anymore: https://github.com/ampproject/amp-wp/blob/0c627dc4be5085e00309f10970a38b0132fa40b1/bin/deploy-travis-pantheon.sh

Also, this cache folder referenced in `.travis.yml` can probably also be removed: https://github.com/ampproject/amp-wp/blob/0c627dc4be5085e00309f10970a38b0132fa40b1/.travis.yml#L25

westonruter:
Good catch. Fixed in 1cc08e02.

schlessera:
Seems good to go!

westonruter:
Failed due to unrelated E2E test failure, so merging anyway.

PR #3084, opened 2019-08-23
Update allowed tags/attributes from spec in amphtml 1908162134430

westonruter:
Previously: #3003.

- [x] Run `./bin/amphtml-update.sh`
- [ ] Examine diff for changelog
- [ ] Update spec generator as needed based on spec format changes.
- [ ] Modify validating sanitizer based on changes to spec, if needed.
- [ ] Add tests for key changes

# Changelog

..

# Details

```bash
(
    PREV_VERSION=1907301630320;
    THIS_VERSION=1908162134430; 
    git diff $PREV_VERSION...$THIS_VERSION -w -- $( git ls-files $THIS_VERSION -- . | grep '.protoascii' )
)
```

<details>
<summary>Compare 1907301630320...1908162134430</summary>

```diff
diff --git a/extensions/amp-autocomplete/validator-amp-autocomplete.protoascii b/extensions/amp-autocomplete/validator-amp-autocomplete.protoascii
index 6c0485078..03ec7cfb0 100644
--- a/extensions/amp-autocomplete/validator-amp-autocomplete.protoascii
+++ b/extensions/amp-autocomplete/validator-amp-autocomplete.protoascii
@@ -29,7 +29,6 @@ tags: {  # <amp-autocomplete>
   tag_name: "AMP-AUTOCOMPLETE"
   spec_name: "amp-autocomplete"
   requires_extension: "amp-autocomplete"
-  mandatory_ancestor: "FORM"
   attrs: {
     name: "filter"
     value_casei: "custom"
diff --git a/extensions/amp-carousel/validator-amp-carousel.protoascii b/extensions/amp-carousel/validator-amp-carousel.protoascii
index 1b1f31e9c..1984908e9 100644
--- a/extensions/amp-carousel/validator-amp-carousel.protoascii
+++ b/extensions/amp-carousel/validator-amp-carousel.protoascii
@@ -23,6 +23,7 @@ tags: {  # amp-carousel
   extension_spec: {
     name: "amp-carousel"
     version: "0.1"
+    version: "0.2"
     version: "latest"
     requires_usage: GRANDFATHERED
     deprecated_allow_duplicates: true
@@ -53,6 +54,10 @@ attr_lists: {
   attrs: {
     name: "loop"
     value: ""
+  }
+    attrs: {
+    name: "slide"
+    value_regex: "[0-9]+"
   }
   # <amp-bind>
   attrs: { name: "[slide]" }
[...]

PR #3346, opened 2019-09-25
Auto-reject excessive CSS validation errors

### LINKED ISSUE #2326: Discontinue excluding excessive CSS by default on Standard mode sites
Currently when there is too much CSS, the first stylesheet that exceeds the 50KB limit is excluded from being included on the page. This leads to pages that appear broken. While prioritizing the stylesheets to concatenate will help a lot (#2322), we should also consider whether we should exclude stylesheets by default. This is particularly important when activating AMP on a theme that has more than 50KB of CSS even without the admin bar, or when a certain combination of blocks causes more than 50KB of CSS to be added to the page after tree shaking.

We currently have this toggle to automatically-accept validation errors, which results in excessive CSS being excluded:

image

What if an `excessive_css` validation error was a special case which instead defaulted to “New Rejected” status? On a Native mode site, this would result in the `amp` attribute being removed from the `html` element, and essentially serving an unmarked AMP page which would be able to take advantage of all the performance benefits of AMP except for being served from the AMP cache.

Should this auto-reject for `excessive_css` apply to Transitional mode as well or just Native mode? I'm thinking no, as the non-AMP version would be a better default that includes all the CSS. The primary problem today is Native mode sites that exclude CSS. 

If we do this, it will be important to show a notice that too much CSS is on the page: #1801.

schlessera:
The `AMP_Validation_Manager::is_sanitization_auto_accepted()` now accepts an `$error` as argument and auto-rejects the specific case of an excessive CSS validation error.

In addition, the Tag & Attribute sanitizer specifically ignores the `<style amp-custom>` tag when it comes to the `max_bytes` rule in the spec (introduced in #3084).

Finally, to ensure that the resulting `<style amp-custom>` is not stripped itself for excessive size, the tag & attribute sanitizer treats this tag as a special case for enforcing the `'max_bytes'` spec rule.

Fixes #2326

### DIFF
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -671,7 +671,7 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 		if ( isset( $cdata_spec['max_bytes'] ) && strlen( $element->textContent ) > $cdata_spec['max_bytes'] ) {
 			// Skip the <style amp-custom> tag, as we want to display it even with an excessive size if it passed the style sanitizer.
 			// This would mean that AMP was disabled to not break the styling.
-			if ( $element->nodeName !== 'style' || ! $element->hasAttributes() || $element->attributes->item(0)->nodeName !== 'amp-custom' ) {
+			if ( 'style' !== $element->nodeName || ! $element->hasAttributes() || 'amp-custom' !== $element->attributes->item( 0 )->nodeName ) {
 				return new WP_Error( 'excessive_bytes' );
 			}
 		}
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -668,12 +668,14 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 	 * @return true|WP_Error True when valid or error when invalid.
 	 */
 	private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
-		if ( isset( $cdata_spec['max_bytes'] ) && strlen( $element->textContent ) > $cdata_spec['max_bytes'] ) {
+		if (
+			isset( $cdata_spec['max_bytes'] ) && strlen( $element->textContent ) > $cdata_spec['max_bytes']
+			&&
 			// Skip the <style amp-custom> tag, as we want to display it even with an excessive size if it passed the style sanitizer.
 			// This would mean that AMP was disabled to not break the styling.
-			if ( 'style' !== $element->nodeName || ! $element->hasAttributes() || 'amp-custom' !== $element->attributes->item( 0 )->nodeName ) {
-				return new WP_Error( 'excessive_bytes' );
-			}
+			! ( 'style' === $element->nodeName && $element->hasAttributes() && 'amp-custom' === $element->attributes->item( 0 )->nodeName )
+		) {
+			return new WP_Error( 'excessive_bytes' );
 		}
 		if ( isset( $cdata_spec['blacklisted_cdata_regex'] ) ) {
 			if ( preg_match( '@' . $cdata_spec['blacklisted_cdata_regex']['regex'] . '@u', $element->textContent ) ) {
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -673,7 +673,7 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 			&&
 			// Skip the <style amp-custom> tag, as we want to display it even with an excessive size if it passed the style sanitizer.
 			// This would mean that AMP was disabled to not break the styling.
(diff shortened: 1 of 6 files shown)


includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -669,7 +669,11 @@ public function is_missing_mandatory_attribute( $attr_spec, DOMElement $node ) {
 	 */
 	private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
 		if ( isset( $cdata_spec['max_bytes'] ) && strlen( $element->textContent ) > $cdata_spec['max_bytes'] ) {
-			return new WP_Error( 'excessive_bytes' );
+			// Skip the <style amp-custom> tag, as we want to display it even with an excessive size if it passed the style sanitizer.
+			// This would mean that AMP was disabled to not break the styling.
+			if ( 'style' !== $element->nodeName || ! $element->hasAttributes() || 'amp-custom' !== $element->attributes->item( 0 )->nodeName ) {
```
westonruter:
```suggestion
			if ( 'style' !== $element->nodeName || ! $element->hasAttribute( 'amp-custom' ) ) {
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -668,7 +668,13 @@ public function is_missing_mandatory_attribute( $attr_spec, DOMElement $node ) {
 	 * @return true|WP_Error True when valid or error when invalid.
 	 */
 	private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
-		if ( isset( $cdata_spec['max_bytes'] ) && strlen( $element->textContent ) > $cdata_spec['max_bytes'] ) {
+		if (
+			isset( $cdata_spec['max_bytes'] ) && strlen( $element->textContent ) > $cdata_spec['max_bytes']
+			&&
+			// Skip the <style amp-custom> tag, as we want to display it even with an excessive size if it passed the style sanitizer.
+			// This would mean that AMP was disabled to not break the styling.
+			! ( 'style' === $element->nodeName && $element->hasAttributes() && 'amp-custom' === $element->attributes->item( 0 )->nodeName )
```
westonruter:
```suggestion
			! ( 'style' === $element->nodeName && $element->hasAttribute( 'amp-custom' ) )
```

schlessera:
Lots of broken tests for this change, will have to go through them and see what exact Pandora's box I actually opened here... ;)

westonruter:
> Lots of broken tests for this change, will have to go through them and see what exact Pandora's box I actually opened here... ;)

It may be just a matter of replacing `add_theme_support('amp')` with `add_theme_support('amp', ['paired' => true])`.

westonruter:
Alternatively, if the return value of `is_sanitization_auto_accepted` can be filtered, then the tests could could be updated to just `__return_true` for that filter.

schlessera:
> if the return value of `is_sanitization_auto_accepted` can be filtered

Is that something we should add anyway?

westonruter:
I think so. I think it makes sense to be able to give the ability to filter whether a given validation error is auto-accepted. This would be needed when removing the auto-accept checkbox, ultimately.

schlessera:
There's only one failing test now: https://travis-ci.org/ampproject/amp-wp/jobs/589550895#L1035-L1044

The `'invalid_element'` that is not being stripped anymore is very specifically is the `<style amp-custom>` in here: https://github.com/ampproject/amp-wp/blob/5888398e9bb7821e084320bb6d70d671f9243a6b/tests/php/test-tag-and-attribute-sanitizer.php#L1818-L1836

The current logic is ignoring `'max_bytes'` for `<style amp-custom>` tags so they don't get removed: https://github.com/ampproject/amp-wp/pull/3346/files#diff-ee41b32e64c90445a71328ca93cfa799R676

I could of course change the test now, but I wonder if there's no way to make the edge case I'm overriding more specific so that the test wouldn't break here. I can't think of a way though.

@westonruter Thoughts?

schlessera:
The only way I can think of here without modifying the tests is to check from within the Tag & Attribute sanitizer whether we have NEW_REJECTED validation errors in the current document. However:
- I'm not yet sure this would reliably work in terms of timing.
- It would tightly couple the Tag & Attribute sanitizer to the current WordPress data model, which we want to avoid.

So for now I'm going ahead and changing that test instead, but let's discuss further as needed.

schlessera:
> I think so. I think it makes sense to be able to give the ability to filter whether a given validation error is auto-accepted. This would be needed when removing the auto-accept checkbox, ultimately.

As I don't actually need it here, I'll open a separate issue for that.

schlessera:
I added tests for:
- asserting that `excessive_style` is rejected by default (validation manager)
- <style amp-custom> is not removed if it exceeds `'max_bytes'` (tag & attribute sanitizer)

I think this should cover the current behavior. The style sanitizer did not require any changes. However, right now, the CSS manifest comment does not contain any added information. I'm not sure if we should add something there or not.

kienstra:
**Looks Good**

Hi @schlessera,
This looks good.

On testing this with *Query Monitor* to ensure the CSS is above 50 KB, the `excessive_css` errors were auto-rejected in Standard mode:

new-r


...and the `<amp-custom>` didn't have any styling removed:

amp-attribute-not-present

Before this PR, the Query Monitor CSS was removed, as it put the CSS over the 50 KB limit.

PR #3347, opened 2019-09-25
Fix for pasted box size

### LINKED ISSUE #3291: Text block does not update height after pasting
## Bug Description

As reported in #3259:

See https://github.com/ampproject/amp-wp/pull/3259#issuecomment-532610932 for where this might be happening.

## Expected Behaviour

Pasting arbitrary text should create a new text block with correct height

## Steps to reproduce

1. Copy text from somewhere
1. Click on story page
1. Paste via keyboard
1. Notice how newly added text block does not update its height

## Screenshots

Screenshot from 2019-09-18 10-07-54

---------------

## Acceptance criteria

* Pasting arbitrary text should create a new text block with correct height

## Implementation brief

* Update `maybeUpdateBlockDimensions` so it correctly handles pasted blocks

## QA testing instructions

1. Copy text from somewhere
1. Click on story page
1. Paste via keyboard ( CTRL + V)
1. Verify height of text block is the height of the text.

## Demo

* https://youtu.be/X8i0Tb71QLc

## Changelog entry

* Improve calculating block dimensions when pasting text

spacedmonkey:
Fixes #3291

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-text/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-text/edit.js
@@ -40,13 +40,14 @@ class TextBlockEdit extends Component {
 	}
 
 	componentDidUpdate( prevProps, prevState ) {
-		const { attributes, fontSize, isSelected } = this.props;
+		const { attributes, fontSize, isSelected, setAttributes } = this.props;
 		const {
 			height,
 			width,
 			content,
 			ampFitText,
 			ampFontFamily,
+			isPasted,
 		} = attributes;
 
 		// If the block was unselected, make sure that it's not editing anymore.
@@ -73,6 +74,12 @@ class TextBlockEdit extends Component {
 			prevProps.attributes.ampFontFamily !== ampFontFamily ||
 			prevProps.attributes.content !== content
 		);
+
+		// If block has been pasted, then regenerate height and width.
+		if ( isPasted ) {
+			setAttributes( { isPasted: false } );
+			maybeUpdateBlockDimensions( this.props );
+		}
 
 		if ( checkBlockDimensions ) {
 			maybeUpdateBlockDimensions( this.props );
@@ -252,6 +259,7 @@ TextBlockEdit.propTypes = {
 		opacity: PropTypes.number,
 		className: PropTypes.string,
 		ampFontFamily: PropTypes.string,
+		isPasted: PropTypes.bool,
 	} ).isRequired,
 	isSelected: PropTypes.bool.isRequired,
 	clientId: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


assets/src/stories-editor/blocks/amp-story-text/index.js
```diff
@@ -40,6 +40,10 @@ const schema = {
 	align: {
 		type: 'string',
 	},
+	isPasted: {
```
barklund: Does including it here mean that this attribute will be stored in WP for every text block? Or is it still only an internal attribute in the frontend initialised to `false` for all blocks?

Just want to make sure we aren't polluting the database with unnecessary attributes.

barklund: It doesn't seem to go to the backend at all, everything looks good here.

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -74,6 +75,12 @@ class TextBlockEdit extends Component {
 			prevProps.attributes.content !== content
 		);
 
+		// If block has been pasted, then regenerate height and width.
+		if ( isPasted ) {
+			setAttributes( { isPasted: false } );
+			maybeUpdateBlockDimensions( this.props );
+		}
+
```
miina: Should `if ( checkBlockDimensions ) {` be with `else` now? To ensure it's not happening the second time after it's already done with `isPasted`? Not sure if that would ever be the case though.

spacedmonkey: We could do this 

```javascript
if ( checkBlockDimensions || isPasted ) {
	maybeUpdateBlockDimensions( this.props );
}

if ( isPasted ) {
	setAttributes( { isPasted: false } );
}
```

miina: 👍 
Alternatively, we could add `isPasted` to the `checkBlockDimension` conditions so that if  `isPasted`, `checkBlockDimensions` would be `true`. And then just add the `if ( isPasted )` for setting it `false` as in your previous comment.

barklund: @miina's latter suggestion also "fixes" that case that stuff pasted into a fitted text box doesn't require block dimension updates. But that doesn't actually seem to be a problem anyway.

spacedmonkey: The above code snippet feels the safest, as it is much more human readable here. I will update my PR.

spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3347/commits/4460c2e171e129a41b20d6a9cb475d29cf2bae64

PR #3349, opened 2019-09-26
Fix bug in style sanitizer not freeing up memory because of wrong index

schlessera:
In the style sanitizer, when the `@import` statements are being processed, there is an `unset()` operation that works against an invalid index. It looks like the iterating indexes were mixed up in the initial code and further broken in subsequent changes.

As this is within an `unset()`, it doesn't throw an immediate error, but trying to actually look at what is being unset returns an invalid index error.


westonruter:
Cool. How did you find this and what is the memory impact?

schlessera:
I was going through the code to see how the calculcations for the CSS manifest are made, and whether there would be an obvious need to adapt for #3346.

While trying to understand the logic, this just didn't make sense to me.

I don't think the memory impact will be substantial for most sites, as this seems to only include small strings per entry.

PR #3353, opened 2019-09-26
Use PHPCompatibilityWP for compat checks

swissspidy:
This WordPress specific ruleset prevents false positives from the PHPCompatibility standard by excluding back-fills and poly-fills which are provided by WordPress.

It doesn't replace PHPCompatibility, but builds on top of it instead.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -78,7 +78,7 @@
 
 	<!-- Include sniffs for PHP cross-version compatibility. -->
 	<config name="testVersion" value="5.4-"/>
-	<rule ref="PHPCompatibility">
+	<rule ref="PHPCompatibilityWP">
 		<exclude-pattern>bin/*</exclude-pattern>
 	</rule>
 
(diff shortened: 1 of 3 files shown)


phpcs.xml
```diff
@@ -77,7 +77,7 @@
 	</rule>
 
 	<!-- Include sniffs for PHP cross-version compatibility. -->
-	<config name="testVersion" value="5.4-99.0"/>
+	<config name="testVersion" value="5.4-"/>
 	<rule ref="PHPCompatibility">
```
schlessera: You need to use the replacement ruleset for it to take effect:
```suggestion
	<rule ref="PHPCompatibilityWP">
```

swissspidy: D'oh, I thought I did that.

schlessera:
Right now, the ruleset is not active in the tests. Once you've made the suggested change and it is being enforced, we'll have to see how many failures we'll get. From my experience, the cleaner you want to decouple code from WordPress, the worse this standard becomes, as it really assumes "the WordPress Way", all the way.
So we might end up needing to generally exclude entire areas like the sanitizers if it gets too bad.

schlessera:
Oh, nevermind my previous comment, I'm now thinking about the WP PHPCS rules, not the PHP compat rules.

Issue #3247, opened 2019-09-16
Investigate flaky E2E tests


swissspidy:
There were quite a few flaky e2e tests recently. Many were uncovered in #3201 specifically and temporarily skipped.

Let's audit these tests to make them more robust.

PR #3355, opened 2019-09-26
Flaky tests investigation

miina:
Restores tests for investigating the failure (#3247)


swissspidy:
FWIW testing with the `--no-cache` flag might also be useful here. It helped uncover a few issues in the Site Kit tests.

miina:
OK, I've run the e2e tests for at least 6 (or more, lost the count) times and it's passing every time.

It looks like the issue comes from the verification happening too soon after pasting. the `toMatch` waits for 500ms, however, that doesn't always seem to be enough time. `waitForSelector` works since it waits for more time.

Thoughts / doubts?

swissspidy:
Let's give it a try 👍 

Not sure if there's a good way to detect these kind of things during development. Don't want to run into such issues every couple of weeks.

Issue #3309, opened 2019-09-19
Migrate Jetpack shortcode handling out of AMP plugin and into Jetpack


chvillanuevap:
The AMP plugin includes shortcodes for several various embed handlers, including:

* `tweet`
* `vimeo`
* `soundcloud`
* `youtube`
* `dailymotion`
* `instagram`
 
For example, the AMP plugin includes a `vimeo` shortcode as part of its `AMP_Vimeo_Embed_Handler` for historical reasons, but over time more and more Jetpack code has been moved from the AMP plugin to… Jetpack. See #1021. This code needs to be extracted from the AMP plugin to move over to Jetpack. All such shortcode handling needs to be moved to Jetpack since the shortcodes are only being registered for AMP responses, leaving non-AMP pages without any shortcode defined.

----
Hi there!

I am trying to convert my website's theme to be fully AMP compatible. 

In the blog posts from my website, we use the Jetpack Vimeo shortcode. As the [documentation](https://en.support.wordpress.com/videos/vimeo/) states, we can use the shortcode by providing the URL or the video ID. We have used the shortcode exclusively using the [vimeo 123456] format, and it has worked as intended so far.

While testing the AMP version of my site, I noticed that the Vimeo videos were not showing. Further debugging took me to the AMP plugin, to the file `includes/embeds/class-amp-vimeo-embed.php`, to the function `shortcode` in line 71.

```php
public function shortcode( $attr ) {
	$video_id = false;

	if ( isset( $attr['id'] ) ) {
		$video_id = $attr['id'];
	} elseif ( isset( $attr['url'] ) ) {
		$video_id = $this->get_video_id_from_url( $attr['url'] );
	} elseif ( isset( $attr[0] ) ) {
		$video_id = $this->get_video_id_from_url( $attr[0] );
	} elseif ( function_exists( 'shortcode_new_to_old_params' ) ) {
		$video_id = shortcode_new_to_old_params( $attr );
	}

	if ( empty( $video_id ) ) {
		return '';
	}

	return $this->render(
		[
			'video_id' => $video_id,
		]
	);
}
```
The issue here is that the way the Jetpack documentation says it's possible to use the shortcode there is no need to use an `id` or `url` attribute, so the first two if statements do not apply to my case. Now, the third statement applies because I have submitted the Vimeo ID to the shortcode, and therefore `$attr[0]` is set. However, the statement assumes that I have passed a URL, as it calls `get_video_id_from_url`, where as I have passed an ID. 

[...]

PR #2722, opened 2019-07-01
Improve Soundcloud embed: support playlists, preserve visual/height params, include fallback content

westonruter:
Fixes #2715.

* Add support for Soundcloud playlists
* Preserve `visual` param to match the non-AMP version.
* Preserve `height` param so the height is largely the same as the non-AMP version. Note that there is not a 1:1 match yet because `amp-soundcloud` does not support the `responsive` layout. See https://github.com/ampproject/amphtml/issues/23144.
* Add `fallback` content with a link to the original Soundcloud link with its title; this improves accessibility and indexability.

Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3347837/amp.zip) (1.2.1-alpha-20190702T011358Z-e864ea9d)

# Non-AMP version

Track embed: https://soundcloud.com/jack-villano-villano/mozart-requiem-in-d-minor

image

Playlist embed: https://soundcloud.com/classical-music-playlist/sets/classical-music-essential-collection

image

# Old AMP Version 

Track embed:

image

```html
<figure class="wp-block-embed-soundcloud wp-block-embed is-type-rich is-provider-soundcloud  wp-has-aspect-ratio">
	<div class="wp-block-embed__wrapper">
		<amp-soundcloud data-trackid="90097394" layout="fixed-height" height="200"></amp-soundcloud>
	</div>
</figure>
```

Playlist embed:

(Fails to render)

```html
<p><a href="" class="amp-wp-embed-fallback"></a></p>
```

# New AMP Version

Track:

image

```html
<figure class="wp-block-embed-soundcloud wp-block-embed is-type-rich is-provider-soundcloud  wp-has-aspect-ratio">
	<div class="wp-block-embed__wrapper">
		<amp-soundcloud data-trackid="90097394" data-visual="true" height="400" layout="fixed-height">
			<a fallback href="https://soundcloud.com/jack-villano-villano/mozart-requiem-in-d-minor">Mozart – Requiem in D minor Complete Full by Jack Villano Villano</a>
		</amp-soundcloud>
	</div>
</figure>
```

Playlist:

image

```html
<figure class="wp-block-embed-soundcloud wp-block-embed is-type-rich is-provider-soundcloud  wp-has-aspect-ratio">
	<div class="wp-block-embed__wrapper">
		<amp-soundcloud data-playlistid="40936190" data-visual="true" height="450" layout="fixed-height">
[...]

PR #3358, opened 2019-09-27
Improve handling of YouTube embeds

### LINKED ISSUE #3313: iFrame Title Attribute
Currently I have a filter setup on `the_content` to add a title to all iFrames for accessibility purposes on non-AMP pages. However, AMP-WP strips out that title. I assume because the iFrame titles are not on the actual content.

Is there a way the AMP plugin can pull the iFrame title through to the AMP version if it exists on the non-AMP version? Or a way I can extend `AMP_Iframe_Sanitizer` to add in a title?

### LINKED ISSUE #3348: Undefined index: query (class-amp-youtube-embed.php in AMP_YouTube_Embed_Handler::get_video_id_from_url)
If YouTube is embedded in this format: https://www.youtube.com/embed/xS3yL5kR02 and not the usual ?v= method, the following notice will be triggered:

```
ErrorException: Notice: Undefined index: query
#12 /wp-content/plugins/amp/includes/embeds/class-amp-youtube-embed.php(166): AMP_YouTube_Embed_Handler::get_video_id_from_url
#11 /wp-content/plugins/amp/includes/embeds/class-amp-youtube-embed.php(84): AMP_YouTube_Embed_Handler::shortcode
#10 /wp-content/plugins/amp/includes/embeds/class-amp-youtube-embed.php(106): AMP_YouTube_Embed_Handler::oembed
#9 /wordpress/wp-includes/class-wp-embed.php(173): WP_Embed::shortcode
#8 /wordpress/wp-includes/class-wp-embed.php(427): WP_Embed::autoembed_callback
#7 [internal](0): preg_replace_callback
#6 /wordpress/wp-includes/class-wp-embed.php(409): WP_Embed::autoembed
#5 /wordpress/wp-includes/class-wp-hook.php(286): WP_Hook::apply_filters
#4 /wordpress/wp-includes/plugin.php(208): apply_filters
#3 /wp-content/themes/ap2/amp/single.php(12): include
#2 /wordpress/wp-includes/template-loader.php(78): require_once
#1 /wordpress/wp-blog-header.php(19): require
#0 /index.php(23): null
```

I'm using v1.2.2 of the AMP WP plugin.

westonruter:
This pull request aims to clean up handling of YouTube embed handling, in large part to apply similar improvements as #2722 for SoundCloud.

- [x] Preserve original YouTube iframe `title` as `placeholder` content: `<a placeholder href="https://youtube.com/watch?v=tiQcA7cwEMo">AMP for JavaScripters</a>`. See https://github.com/ampproject/amp-wp/pull/3358#discussion_r344363294.
- [x] Add the `title` attribute to the `amp-youtube` element for a11y, in addition to the `fallback` content. See https://github.com/ampproject/amp-wp/issues/3313#issuecomment-535743947.
- [x] Merge `test-amp-youtube-embed.php` and `test-class-amp-youtube-embed-handler.php`, update tests following model of #2722.
- [x] Fix video ID parsing logic. See #3348.

Fixes #3348. Fixes #3313.

Relates to https://github.com/ampproject/amp-wp/issues/3309 in that the shortcode logic needs to be removed/moved to Jetpack.


westonruter:
This PR is free for anyone to take over.

westonruter:
@googlebot I consent.

pierlon:
@googlebot I consent.

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -83,69 +141,72 @@ public function shortcode( $attr ) {
...
 	 * @see \WP_Embed::shortcode()
+	 * @deprecated This is no longer being used.
 	 *
 	 * @param array  $matches URL pattern matches.
 	 * @param array  $attr    Shortcode attribues.
 	 * @param string $url     URL.
 	 * @return string Rendered oEmbed.
 	 */
 	public function oembed( $matches, $attr, $url ) {
+		_deprecated_function( __METHOD__, '1.3.1' );
```
westonruter:
```suggestion
		_deprecated_function( __METHOD__, '1.5.0' );
```

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -83,69 +141,72 @@ public function shortcode( $attr ) {
...
+				'fallback' => '',
 			]
 		);
 
 		if ( empty( $args['video_id'] ) ) {
 			return AMP_HTML_Utils::build_tag(
 				'a',
 				[
-					'href'  => esc_url( $args['url'] ),
+					'href'  => esc_url( $url ),
```
westonruter: Just realized that this needs to rather be `esc_url_raw()` since the HTML entity encoding will be handled by the DOM. If there are other places where `esc_url()` is being used in this way, it should be updated to be `esc_url_raw()` as well.

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -162,7 +223,7 @@ private function get_video_id_from_url( $url ) {
 		} else {
 			/* phpcs:ignore Squiz.PHP.CommentedOutCode.Found
 			The query looks like ?v={id} or ?list={id} */
-			parse_str( $parsed_url['query'], $query_args );
+			parse_str( $parsed_url['query'], $query_args ); // @todo Bug! See <https://github.com/ampproject/amp-wp/issues/3348>.
```
westonruter: To be done in this PR or another?

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -83,69 +141,72 @@ public function shortcode( $attr ) {
...
+				wp_array_slice_assoc( $args, [ 'layout', 'width', 'height', 'title' ] )
+			),
+			$args['fallback']
 		);
 	}
 
 	/**
 	 * Determine the video ID from the URL.
 	 *
+	 * @todo Needs to be totally refactored.
```
westonruter: In this PR or another?

tests/php/test-class-amp-youtube-embed-handler.php
```diff
@@ -66,4 +124,96 @@ public function test_video_override() {
...
+				'<p><amp-youtube data-videoid="kfVsfOSbJY0" layout="responsive" width="480" height="270" title="Rebecca Black - Friday"><a fallback href="https://youtube.com/watch?v=kfVsfOSbJY0">Rebecca Black &#8211; Friday</a></amp-youtube></p>' . PHP_EOL,
+				'<p><amp-youtube data-videoid="kfVsfOSbJY0" layout="responsive" width="480" height="270"></amp-youtube></p>' . PHP_EOL,
+			],
+		];
+	}
+
+	/**
+	 * @dataProvider get_conversion_data
+	 */
+	public function test__conversion( $source, $expected, $fallback = null ) {
```
westonruter: Could you rename `$fallback` to be something else here? It gets confusing with `fallback` content in AMP. Perhaps `$expected_without_fallback` would be better.

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -49,22 +50,79 @@ public function __construct( $args = [] ) {
...
+	 * @param string|null $url  Embed URL, for fallback purposes.
+	 * @return array Props for rendering the component.
+	 */
+	private function parse_props( $html, $url ) {
+		$props = [];
+
+		if ( preg_match( '#<iframe[^>]*?title="(?P<title>[^"]+)"#s', $html, $matches ) ) {
+			$props['title']    = $matches['title'];
+			$props['fallback'] = sprintf(
+				'<a fallback href="%s">%s</a>',
```
westonruter: I keep arguing with myself whether this should be a `fallback` or a `placeholder`: https://amp.dev/documentation/guides-and-tutorials/develop/style_and_layout/placeholders/

Actually, I think the answer is this _should_ be changed to rather be an `amp-img` as a `placeholder` as shown on the [`amp-youtube`](https://amp.dev/documentation/components/amp-youtube/) docs:

```html
  <amp-youtube
      id="myLiveChannel"
      data-live-channelid="UCB8Kb4pxYzsDsHxzBfnid4Q"
      width="358"
      height="204"
      layout="responsive">
    <amp-img
      src="https://i.ytimg.com/vi/Wm1fWz-7nLQ/hqdefault_live.jpg"
      placeholder
      layout="fill"
      />
  </amp-youtube>
```

The example there is incomplete however. In particular, the `amp-img` should have an `alt` attribute provided which is the `title` of the video.

Also, instead of actually generating an `amp-img` here, just output a regular `img` as it will get converted into an `amp-img` later by the image sanitizer.

Also, the `amp-img` placeholder would be suitable even when there is no `title` available, it's just that the `alt` would have to be blank.

Also, keeping the link to the original YouTube video is good. 

So what to do:

0. Unconditionally construct the link regardless of whether the `iframe` has a `title`.
1. Change this to be `placeholder` instead of `fallback`.
2. Parse construct the HQ video thumbnail from the video ID (e.g. `https://i.ytimg.com/vi/$video_id/hqdefault.jpg`).
3. Supply an `img` element as the child of the `a` link.
4. Supply the `title` of the `iframe` as the `img[alt]`, if available.

So in the end, it should look like this:

```html
<p>
    <amp-youtube data-videoid="kfVsfOSbJY0" layout="responsive" width="480" height="270" title="Rebecca Black - Friday">
        <a placeholder href="https://www.youtube.com/watch?v=kfVsfOSbJY0">
            <img src="https://i.ytimg.com/vi/kfVsfOSbJY0/hqdefault.jpg" alt="Rebecca Black &#8211; Friday" data-amp-layout="fill">
        </a>
    </amp-youtube>
</p>
```

And in WordPress<5.2, the only difference would be no `alt` attribute.

includes/embeds/class-amp-youtube-embed-handler.php
pierlon: That needs to be removed as the [`embed_oembed_html`](https://github.com/ampproject/amp-wp/pull/3358/commits/35260fd6b5068f75b987ffc631ab87366d9f24a9#diff-c0516153a81ae9941d43b32041e16347R53) filter correctly handles the URL and transforms it into an embeddable one.

includes/embeds/class-amp-youtube-embed-handler.php
pierlon: Will be done in this PR.

tests/php/test-class-amp-youtube-embed-handler.php
pierlon: Ah, OK. `$expected_without_fallback` seems a bit confusing as well, how about something direct like `$fallback_for_expected`?

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -49,22 +50,93 @@ public function __construct( $args = [] ) {
...
+		$props['placeholder'] = AMP_HTML_Utils::build_tag(
+			'a',
+			[
+				'placeholder' => '',
+				'href'        => esc_url_raw( $url ),
+			],
+			$img
+		);
+
+		if ( preg_match( '#<iframe[^>]*?height="(?P<height>\d+)"#s', $html, $matches ) ) {
```
schlessera: Instead of having three separate regexes here that parse the iframe, it would be preferable to build a single one that parses all three attributes in one go.

See https://regex101.com/r/zErkQE/1 for an example of how to do that (with attributes in arbitrary order).

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -150,35 +223,11 @@ public function render( $args ) {
 	 * @return integer Video ID.
```
schlessera:
```suggestion
	 * @return integer|false Video ID, or false if none could be retrieved.
```

includes/embeds/class-amp-youtube-embed-handler.php
pierlon: That's a much better solution, thanks @schlessera :pray:

schlessera:
@googlebot I consent.

pierlon:
Hey @schlessera, you're needed to give consent over here again :smile:

schlessera:
@pierlon It's already done.

pierlon:
Ah, thanks!

includes/embeds/class-amp-youtube-embed-handler.php
westonruter: I'm impressed. I didn't realize this could be done with a single pattern.

pierlon:
@westonruter on WP <5.2, the `title` attribute on `amp-youtube` and the `alt` attribute on `img` lingers without a value. Would that be a concern?

westonruter:
> on WP <5.2, the `title` attribute on `amp-youtube` and the `alt` attribute on `img` lingers without a value. Would that be of concern?

@pierlon Yeah, it would probably be best if the attributes were removed on PHP 5.2. I've adjusted them in 97e8db5.

includes/embeds/class-amp-youtube-embed-handler.php
westonruter: @pierlon @schlessera I've applied the same change to the Soundcloud embed handler, and extracted the common logic to the base class. Please take a look.

westonruter:
I've improved the display of the placeholder image in fc0e9e0 by using `object-fit=cover`.

Before:

Screen Shot 2019-11-09 at 16 27 33

After:

Screen Shot 2019-11-09 at 16 27 46

westonruter:
Aside: I'm curious why the `amp-youtube` element doesn't provide such a `placeholder` by default.

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -77,4 +77,29 @@ public function __construct( $args = [] ) {
...
+	 * @since 1.5.0
+	 *
+	 * @param string   $tag_name        Tag name.
+	 * @param string[] $attribute_names Attribute names.
+	 * @return string Regular expression pattern.
+	 */
+	protected function get_html_attribute_pattern( $tag_name, $attribute_names ) {
+		return sprintf(
+			'/<%s%s/',
+			$tag_name,
```
schlessera: Although unlikely to be a real issue, the `$tag_name` should be put through `preg_quote()`.

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -77,4 +77,29 @@ public function __construct( $args = [] ) {
...
+			implode(
+				'',
+				array_map(
+					function ( $attr_name ) {
+						return sprintf( '(?=[^>]*?%1$s="(?P<%1$s>[^"]+)")?', $attr_name );
```
schlessera: Although unlikely to be a real issue, the `$attr_name` should be put through `preg_quote()`.

includes/embeds/class-amp-base-embed-handler.php
westonruter: Done in 62e7ee1b93f2baa6092e33893060f7aac16b1dfd

includes/embeds/class-amp-base-embed-handler.php
westonruter: Done in 62e7ee1b93f2baa6092e33893060f7aac16b1dfd

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -77,4 +77,29 @@ public function __construct( $args = [] ) {
...
+	/**
+	 * Get regex pattern for matching HTML attributes from a given tag name.
+	 *
+	 * @since 1.5.0
+	 *
+	 * @param string   $tag_name        Tag name.
+	 * @param string[] $attribute_names Attribute names.
+	 * @return string Regular expression pattern.
+	 */
+	protected function get_html_attribute_pattern( $tag_name, $attribute_names ) {
```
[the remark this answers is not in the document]
westonruter: Excellent point. I didn't refactor far enough. See d45fd592a35b1dc18c17e1d6c30df558fc34c4b2.

westonruter:
I'm excited about the changes here. Huge improvements to YouTube embed handling. Great fallback even for when JS is turned off in the browser:

image

csossi:
Verified in QA

PR #3359, opened 2019-09-27
Experiments with React hooks

swissspidy:
⚠️ **Work in progress - do not merge** ⚠️

Includes some experimental refactoring of dozens of components to a) make them functional ones and b) to leverage [React hooks](https://reactjs.org/docs/hooks-intro.html). It is largely untested, even though I haven't spot any issues so far.

This PR likely has to be split up as hooks cannot be used pre-WordPress 5.3 unless Gutenberg is active, which means they're currently guaranteed to work in the Stories editor, but not elsewhere.


assets/src/stories-editor/components/block-navigation/item.js
```diff
@@ -44,28 +43,47 @@ const parseDropEvent = ( event ) => {
...
+		return getBlockRootClientId( clientId );
+	}, [ clientId ] );
+	const blockOrder = useSelect( ( select ) => {
+		const { getBlockOrder } = select( 'core/block-editor' );
+		return getBlockOrder( rootClientId );
+	}, [ rootClientId ] );
+	const blocks = useSelect( ( select ) => {
+		const { getBlocksByClientId } = select( 'core/block-editor' );
+		return getBlocksByClientId( blockOrder ).map( ( { clientId: id } ) => id ).reverse();
+	}, [ blockOrder ] );
```
barklund: Why not select all three in a single `useSelect`? I know that technically `rootClientId` might not change just because `clientId` does (it might be a different block on the same page) and then your version is a bit quicker, but the overhead is so tiny, that I don't think it warrants this more complex setup.

I'd just go for:

```
	const {
		rootClientId,
		blockOrder,
		blocks,
	} = useSelect( ( select ) => {
		const { getBlockRootClientId, getBlockOrder, getBlocksByClientId } = select( 'core/block-editor' );
		const rootClientId = getBlockRootClientId( clientId );
		const blockOrder = getBlockOrder( rootClientId );
		const blocks = getBlocksByClientId( blockOrder ).map( ( { clientId: id } ) => id ).reverse();
		return { rootClientId, blockOrder, blocks };
	}, [ clientId ] );
```

assets/src/stories-editor/components/editor-carousel/index.js
```diff
@@ -24,74 +23,87 @@ import './edit.css';
...
-class EditorCarousel extends Component {
-	translateWrapper = () => {
+const EditorCarousel = () => {
+	const { pages, currentPage, currentIndex, previousPage, nextPage, isReordering } = useSelect( ( select ) => {
+		const {
+			getBlockOrder,
+			getBlocksByClientId,
+			getAdjacentBlockClientId,
+		} = select( 'core/block-editor' );
+		const { getCurrentPage, isReordering: _isReordering } = select( 'amp/story' );
```
barklund: Yeah, this is a common annoying problem. What's a good name for a boolean-returning function that doesn't clash with the name of the variable holding the result of said function?

assets/src/stories-editor/components/editor-carousel/index.js
```diff
@@ -24,74 +23,87 @@ import './edit.css';
...
+
+		const _currentPage = getCurrentPage();
+		const _pages = getBlocksByClientId( getBlockOrder() );
+
+		const index = pages.findIndex( ( { clientId } ) => clientId === currentPage );
```
barklund: Surely this should be `_pages`, right?

```suggestion
		const index = _pages.findIndex( ( { clientId } ) => clientId === currentPage );
```

assets/src/stories-editor/components/editor-carousel/index.js
```diff
@@ -24,74 +23,87 @@ import './edit.css';
...
+			nextPage: getCurrentPage() ? getAdjacentBlockClientId( _currentPage, 1 ) : null,
+			isReordering: _isReordering(),
+		};
+	}, [] );
+
+	useEffect( () => {
 		const wrapper = document.querySelector( '#amp-story-controls + .block-editor-block-list__layout' );
 
 		if ( ! wrapper ) {
 			return;
```
barklund: Under which circumstances does this return and could this perhaps be triggered by other props then the two used in the guard for the `useEffect`? The old code re-ran on every component update regardless of which prop changed and I could fear that this part here might be related to another prop?

assets/src/stories-editor/components/rotatable-box/index.js
```diff
@@ -19,22 +18,28 @@ import { __, sprintf } from '@wordpress/i18n';
...
+const RotatableBox = ( { angle, initialAngle, blockElementId, className, speak, onRotateStart, onRotate, onRotateStop, snap, snapGap } ) => {
+	const [ isRotating, setIsRotating ] = useState( false );
+	const [ currentAngle, setAngle ] = useState( angle );
 
-		this.state = {
-			isRotating: false,
-			angle: props.angle,
-		};
-	}
+	let elementRef = null;
```
barklund: Maybe `useRef` here?

assets/src/stories-editor/components/rotatable-box/index.js
```diff
@@ -19,22 +18,28 @@ import { __, sprintf } from '@wordpress/i18n';
...
+
+	useEffect( () => {
+		elementRef = document.getElementById( this.props.blockElementId );
```
barklund:
```suggestion
		elementRef = document.getElementById( blockElementId );
```

assets/src/stories-editor/components/rotatable-box/index.js
```diff
@@ -43,45 +48,45 @@ class RotatableBox extends Component {
...
-				angle += 360;
+			let newAngle = LEFT === keyCode ? currentAngle - 30 : currentAngle + 30;
+			if ( newAngle > 360 ) {
+				newAngle -= 360;
+			} else if ( newAngle <= -360 ) {
+				newAngle += 360;
 			}
 
-			this.elementRef.style.transform = `rotate(${ angle }deg)`;
+			elementRef.style.transform = `rotate(${ newAngle }deg)`;
```
barklund: I'd do this in a `useEffect` with `angle` as a guard. Then you don't need it both here and in the `onMouseMove`.

assets/src/stories-editor/components/rotatable-box/index.js
```diff
@@ -43,45 +48,45 @@ class RotatableBox extends Component {
...
-			this.elementRef.style.transform = `rotate(${ this.props.initialAngle }deg)`;
-
-			this.setState(
-				{
-					isRotating: false,
-					angle: this.props.initialAngle,
-				},
-				() => this.props.onRotateStop && this.props.onRotateStop( e, this.props.initialAngle )
-			);
+			elementRef.classList.remove( 'is-rotating' );
```
barklund: Similarly, I'd do this in a `useEffect` utilising that `classList` has a toggle method:

```
useEffect( () => {
	elementRef.classList.toggle( 'is-rotating', isRotating );
}, [ isRotating ] );
```

assets/src/stories-editor/components/rotatable-box/index.js
```diff
@@ -121,66 +125,60 @@ class RotatableBox extends Component {
...
+				>
+					<span className="screen-reader-text">
+						{ __( 'Rotate Block', 'amp' ) }
+					</span>
+				</Button>
 			</div>
-		);
-	}
-}
+			{ this.props.children }
```
barklund:
```suggestion
			{ children }
```

assets/src/stories-editor/components/rotatable-box/index.js
```diff
@@ -43,45 +48,45 @@ class RotatableBox extends Component {
...
-
-			this.setState(
-				{
-					isRotating: false,
-					angle: this.props.initialAngle,
-				},
-				() => this.props.onRotateStop && this.props.onRotateStop( e, this.props.initialAngle )
-			);
+			elementRef.classList.remove( 'is-rotating' );
+			elementRef.style.transform = `rotate(${ initialAngle }deg)`;
```
barklund: This one would also be superfluous by the way.

assets/src/stories-editor/components/block-navigation/item.js
swissspidy: Sure that also works 👍

assets/src/stories-editor/components/rotatable-box/index.js
swissspidy: Makes sense 👍

assets/src/stories-editor/components/editor-carousel/index.js
swissspidy: Good question. To avoid it we could also just use `select( 'amp/story' ).isReordering()`.

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -41,55 +37,113 @@ import PageSettings from './page-settings';
...
 
-	ensureCorrectBlockOrder() {
-		const {
-			getBlockOrder,
-			moveBlockToPosition,
-			clientId,
-		} = this.props;
-		const order = getBlockOrder( clientId );
-		if ( 1 >= order.length ) {
+	const videoPlayer = createRef();
```
barklund: `createRef` can be replaced with `useRef`.

assets/src/stories-editor/components/editor-carousel/index.js
swissspidy: Good question. Will have to test this and perhaps search a bit in git history.

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
```diff
@@ -21,56 +21,34 @@ import AttachmentContent from './attachment-content';
...
-			textColor !== prevProps.textColor
-		) {
-			const { style, attachmentClass } = this.getWrapperAttributes();
-			const newAttributes = { wrapperStyle: style };
-			if ( textColor !== prevProps.textColor ) {
-				newAttributes.attachmentClass = attachmentClass;
-			}
-			setAttributes( newAttributes );
+	const toggleAttachment = ( open ) => {
+		if ( open !== isOpen ) {
```
barklund: [`useState` only re-renders if the value has changed](https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update), so this entire function can just be replaced with `setIsOpen`.

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
```diff
@@ -81,42 +59,39 @@ class PageAttachmentEdit extends Component {
...
-		const {
-			attributes,
-			setAttributes,
-		} = this.props;
+	useEffect( () => {
+		const { style, attachmentClass } = getWrapperAttributes();
+		const newAttributes = { wrapperStyle: style };
+		newAttributes.attachmentClass = attachmentClass;
+		setAttributes( newAttributes );
+	}, [ backgroundColor, customBackgroundColor, textColor ] );
```
barklund: You need everything used by `getWrapperAttributes()` in the guard as well - it's complex, but you can end up with a situation where the `useEffect` references and "older" `getWrapperAttributes`, that uses an "older" `colors` variable and thus is out of sync with the component as a whole. Using multiple hooks, especially `useEffect` can give some really weird side-effects when some of the are reused and some aren't and they reference different callbacks, etc.

Either add `colors` to this guard or use a `useRef` for `colors` to make sure that you always access the most recent value.

Looking at it more closely, why not inline the functionality of `getWrapperAttributes` in the effect directly? It's only used here as far as I can see?

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
```diff
@@ -81,42 +59,39 @@ class PageAttachmentEdit extends Component {
...
-						openText={ openText }
-					/>
-				}
-			</>
-		);
-	}
-}
+	return (
+		<>
+			{ isOpen &&
```
barklund: Convert this to a ternary - that will also get rid of the fragment and reduce the complexity of the return.

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -25,73 +24,82 @@ import { getBackgroundColorWithOpacity } from '../../../common/helpers';
...
+		autoFontSize,
+		width,
+		height,
+		opacity,
+	} = attributes;
+
+	const [ isEditing, setIsEditing ] = useState( false );
+	const [ hasOverlay, setHasOverlay ] = useState( true );
+
+	const toggleIsEditing = ( enable ) => {
```
barklund: These two wrappers for the state setters are also irrelevant. States only cause re-render if the value changes (compared with `Object.is`).

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -25,73 +24,82 @@ import { getBackgroundColorWithOpacity } from '../../../common/helpers';
...
+		}
+	}, [ ampFitText, ampFontFamily, width, height, content ] );
 
-		if ( checkBlockDimensions ) {
-			maybeUpdateBlockDimensions( this.props );
+	useEffect( () => {
+		if ( ! ampFitText ) {
+			maybeUpdateBlockDimensions( props );
 		}
+	}, [ ampFitText, fontSize, ampFitText, ampFontFamily, content ] );
```
barklund: Copy-paste error:

```suggestion
	}, [ ampFitText, fontSize , ampFontFamily, content ] );
```

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -25,73 +24,82 @@ import { getBackgroundColorWithOpacity } from '../../../common/helpers';
...
-			! isEqual( prevProps.fontSize, fontSize ) ||
-			prevProps.attributes.ampFitText !== ampFitText ||
-			prevProps.attributes.ampFontFamily !== ampFontFamily ||
-			prevProps.attributes.content !== content
-		);
+	useEffect( () => {
+		if ( ampFitText ) {
+			maybeUpdateFontSize( props );
+		}
+	}, [ ampFitText, ampFontFamily, width, height, content ] );
```
barklund: This is super messy, because the "helper" is stateful and uses external information making it hard to correctly guard for when it needs to re-run - also it has side-effects. What if this method changes internally and this guard isn't updated?

The helper should received all arguments directly, be stateless and be side-effect free for this to be "proper".

I'm not a big fan of this, but it's probably too huge a task to rewrite these helpers too.

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -25,73 +24,82 @@ import { getBackgroundColorWithOpacity } from '../../../common/helpers';
...
+	const onReplace = ( blocks ) => {
 		// Make sure that 'undefined' values aren't passed into onReplace.
 		blocks = blocks.filter( ( block ) => 'undefined' !== typeof block );
+
 		if ( ! blocks.length ) {
 			return;
 		}
-		onReplace( blocks.map( ( block, index ) => (
+
+		props.onReplace( blocks.map( ( block, index ) => (
```
barklund: Shouldn't this be recursion? `props.onReplace` doesn't exist, does it?

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -101,143 +109,99 @@ class TextBlockEdit extends Component {
...
-			clientId,
-			fontSize,
-			isPartOfMultiSelection,
-			isSelected,
-			backgroundColor,
-			customBackgroundColor,
-			textColor,
-			name,
-		} = this.props;
+	const userFontSize = fontSize && fontSize.size ? fontSize.size + 'px' : undefined;
```
barklund:
```suggestion
	const userFontSize = fontSize && fontSize.size ? `${ fontSize.size }px` : undefined;
```

assets/src/stories-editor/blocks/amp-story-text/edit.js
swissspidy: Nope. A block's edit component gets passed a `onReplace` prop.

See https://github.com/WordPress/gutenberg/blob/fea63772b6e2f71e7422c999e436c44bd3914021/packages/block-editor/src/components/block-list/block.js#L459-L470

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -101,143 +109,99 @@ class TextBlockEdit extends Component {
...
-		const {
-			placeholder,
-			content,
-			align,
-			ampFitText,
-			autoFontSize,
-			height,
-			opacity,
-		} = attributes;
+	const { colors } = select( 'core/block-editor' ).getSettings();
```
barklund: No `useSelect` here? One reason would be to force a re-render if the store changes.

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -101,143 +109,99 @@ class TextBlockEdit extends Component {
...
+	const textWrapperClassName = 'wp-block-amp-story-text';
+	const textClassNames = {
+		'has-text-color': textColor.color,
+		[ textColor.class ]: textColor.class,
+		[ fontSize.class ]: ampFitText ? undefined : fontSize.class,
+		'is-amp-fit-text': ampFitText,
+	};
+	const textStyle = {
+		color: textColor.color,
+		fontSize: ampFitText ? autoFontSize + 'px' : userFontSize,
```
barklund:
```suggestion
		fontSize: ampFitText ? `${ autoFontSize }px` : userFontSize,
```

assets/src/stories-editor/blocks/amp-story-text/edit.js
swissspidy: I just missed that one I guess :-)

assets/src/stories-editor/blocks/amp-story-text/edit.js
```diff
@@ -101,143 +109,99 @@ class TextBlockEdit extends Component {
...
+				<AlignmentToolbar
+					value={ align }
+					onChange={ ( value ) => setAttributes( { align: value } ) }
+				/>
+			</BlockControls>
+			<div className={ classnames( wrapperClass, {
+				'with-line-height': ampFitText,
+				'is-empty-draggable-text': ! isEditing && ! content.length,
+			} ) } style={ wrapperStyle } >
+				{ isEditing &&
```
barklund: Convert to ternary

assets/src/stories-editor/components/inserter/index.js
```diff
@@ -127,12 +126,8 @@ Inserter.propTypes = {
 const applyWithSelect = withSelect( ( select ) => {
 	const { isReordering } = select( 'amp/story' );
```
barklund: Why is this still down here and not inside the component?

swissspidy: For use by the `ifCondition` HOC. Felt a bit easier with that, but can also use hooks instead.

assets/src/stories-editor/components/story-block-drop-zone.js
```diff
@@ -88,30 +88,19 @@ class BlockDropZone extends Component {
...
-	updateBlockAttributes: PropTypes.func,
 	srcClientId: PropTypes.string,
 	srcBlockName: PropTypes.string,
 };
 
-export default withDispatch( ( dispatch ) => {
-	const { updateBlockAttributes } = dispatch( 'core/block-editor' );
-
-	return {
-		updateBlockAttributes( ...args ) {
```
barklund: This old code is really weird? Is there a purpose here?

assets/src/stories-editor/components/template-inserter/template-preview.js
```diff
@@ -8,47 +8,37 @@ import PropTypes from 'prop-types';
...
 		}
-	}
-
-	render() {
-		if ( ! this.state.shouldLoad ) {
-			return <Spinner />;
-		}
-
-		const { item } = this.props;
+	}, [ shouldLoad ] );
```
barklund: This whole effect can be replaced with:

```
	useEffect( () => setTimeout( setShouldLoad, 100, true ), [] );
```

Of course the comment should be kept somewhere. React Concurrent is still not in stable, but will probably come out in the next major release.

assets/src/stories-editor/components/template-inserter/template-preview.js
```diff
@@ -8,47 +8,37 @@ import PropTypes from 'prop-types';
...
 
 /**
  * Internal dependencies
  */
 import { BlockPreview } from '../';
 
-class TemplatePreview extends Component {
-	constructor( ...args ) {
-		super( ...args );
+const TemplatePreview = ( { item } ) => {
```
barklund:
```suggestion
const TemplatePreview = ( { item, setTimeout } ) => {
```

assets/src/stories-editor/components/story-block-drop-zone.js
swissspidy: I don‘t think so. It was just more complicated than needed.

assets/src/stories-editor/components/editor-carousel/index.js
swissspidy: Did a slight refactor here and removed that condition.

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
swissspidy: Fixed in 3ccbe2bec262efacbc5652911a87112b4f0555c6

assets/src/stories-editor/blocks/amp-story-text/edit.js
swissspidy: Fixed now

assets/src/stories-editor/blocks/amp-story-text/edit.js
swissspidy: You mean the helper should receive all arguments individually instead of just  `props`?

swissspidy:
Closing in favor of #3412. The block editor part can be extracted from this PR at some point in the future.

PR #3360, opened 2019-09-27
Fix notice error for featured image in Reader Mode template when using Multisite-Global-Media plugin

### LINKED ISSUE #2869: Notice error for featured image in Reader Mode template when using Multisite-Global-Media plugin
I am getting the following notice error 
```
Notice: Trying to get property 'post_excerpt' of non-object in /var/www/html/wordpress/wp-content/plugins/amp/includes/templates/class-amp-post-template.php on line 378
```

spacedmonkey:
Do some error handling for missing posts. Also make the excerpt is run through the get_the_excerpt filter.

Fixes #2869

### DIFF
--- a/includes/templates/class-amp-post-template.php
+++ b/includes/templates/class-amp-post-template.php
@@ -371,7 +371,7 @@ class AMP_Post_Template {
 		if ( ! $featured_image ) {
 			return;
 		}
-		
+
 		$sanitized_html = AMP_DOM_Utils::get_content_from_dom( $dom );
 
 		$this->add_data_by_key(
--- a/includes/templates/class-amp-post-template.php
+++ b/includes/templates/class-amp-post-template.php
@@ -347,7 +347,11 @@ class AMP_Post_Template {
 			return;
 		}
 
-		$featured_id = get_post_thumbnail_id( $post_id );
+		$featured_id    = get_post_thumbnail_id( $post_id );
+		$featured_image = get_post( $featured_id );
+		if ( ! $featured_image ) {
+			return;
+		}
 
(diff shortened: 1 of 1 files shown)


includes/templates/class-amp-post-template.php
```diff
@@ -369,13 +367,18 @@ private function build_post_featured_image() {
 			]
 		);
 
+		$featured_image = get_post( $featured_id );
+		if ( ! $featured_image ) {
+			return;
+		}
```
westonruter: Shouldn't this condition to short-circuit rather be moved above? No need to construct `$dom` or `$assets` if it's just going to `return`.

includes/templates/class-amp-post-template.php
```diff
@@ -369,13 +367,18 @@ private function build_post_featured_image() {
...
+
 		$sanitized_html = AMP_DOM_Utils::get_content_from_dom( $dom );
 
 		$this->add_data_by_key(
 			'featured_image',
 			[
 				'amp_html' => $sanitized_html,
-				'caption'  => $featured_image->post_excerpt,
+				'caption'  => get_the_excerpt( $featured_image ),
```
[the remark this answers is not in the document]
spacedmonkey: I personally thought that was a benefit here. But you maybe right. I have remove it here https://github.com/ampproject/amp-wp/pull/3360/commits/e30638fa8fb6b576d6c812234c50dd03d3034581

includes/templates/class-amp-post-template.php
spacedmonkey: Fair enough https://github.com/ampproject/amp-wp/pull/3360/commits/e30638fa8fb6b576d6c812234c50dd03d3034581

PR #3366, opened 2019-09-27
Include *.assets.php files in build

westonruter:
This follows up on #3278, updating the build process to include `*.asset.php` files rather than `*.deps.json`.

This is just to fix something broken and is not considering whether or not it is a good idea to use PHP vs JSON. See https://github.com/ampproject/amp-wp/pull/3278#issuecomment-533152735.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -46,7 +46,7 @@ module.exports = function( grunt ) {
 		// Clean up the build.
 		clean: {
 			compiled: {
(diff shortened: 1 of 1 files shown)


schlessera:
> This is just to fix something broken and is not considering whether or not it is a good idea to use PHP vs JSON.

Related discussion in Core happens within https://core.trac.wordpress.org/ticket/48154

Issue #3397, opened 2019-10-01
Resizing Text block: Trying to resize smaller than possible moves the position


miina:
Steps to reproduce:
- Create a Text block.
- Turn off amp-fit-text.
- Resize it from the left handle to as narrow as possible.
- See the results.
![resizing-2](https://user-images.githubusercontent.com/3294597/65973325-6cd45880-e46b-11e9-9bef-b169fd7ac0eb.gif)

Expected: The block should stay in place when resizing and reaching limits.

Note: Same issue for non-rotated Text block, too.

Todo:
- [x] Fix the issue
- [x] Add tests for resizing for when amp fit text is turned off.

miina:
Note that this happens in case of other resizing handles, too. Not all but most.

swissspidy:
I think that's the same as #3290.

miina:
Ah, looks like it! Thanks for noting that.

swissspidy:
@miina Could you add testing instructions here for Claudio? Thanks in advance!

csossi:
Verified in QA

Issue #3326, opened 2019-09-23
Unskip resizing test after bugfix


swissspidy:
Since #3199 got resolved with #3259, we need to unskip the relevant e2e test for it:

https://github.com/ampproject/amp-wp/blob/b44ac6ac687008e4d1c5dd2ef78b2a5d3d9ecd23/tests/e2e/specs/stories-editor/resizing.js#L275-L299

In my quick test it actually didn't quite work out of the box, so perhaps the issue is not yet fully resolved.

spacedmonkey:
After unskipping this test, I am still getting this error in the e2e tests. 
```
  ● Resizing › Non-Fitted Text block › should not resize smaller than the set minimum width and height

    expect(received).toStrictEqual(expected) // deep equality

    Expected: 40
    Received: 250

      288 | 			await dragAndResize( resizableHandleLeft, { x: 300, y: 0 } );
      289 | 			dimensions = await getSelectedBlockDimensions();
    > 290 | 			expect( dimensions.width ).toStrictEqual( textBlockMinWidth );
          | 			                           ^
      291 | 
      292 | 			const resizableHandleRight = await page.$( '.wp-block.is-selected .components-resizable-box__handle-right' );
      293 | 			await dragAndResize( resizableHandleRight, { x: -300, y: 0 } );

      at Object.toStrictEqual (specs/stories-editor/resizing.js:290:31)

```

@barklund Should this tests now work?

swissspidy:
The test was probably flawed before, so it needs some investigation into what is actually expected. The text block in that test doesn't have any content in it either, just the placeholder, which might make a different as well. Would be good to test both cases.

spacedmonkey:
I believe this video displays what the issue is 

https://www.youtube.com/watch?v=9nWmkmaUv5Q&feature=youtu.be

In short, the block isn't resized but instead the block is being dragging around. I am unable to workout why this happening.

spacedmonkey:
I believe the issue is somewhere in the test suite itself. It took me a long time to track this down. Not sure it is the best use of time right now to focus on this.

swissspidy:
Let's focus on other items first that aren't in the backlog. We can get to this later.

miina:
@spacedmonkey If you're OK with me taking this issue over then I'm thinking of looking into this within #3397 since it's needed there.

spacedmonkey:
[...]

PR #3399, opened 2019-10-01
Fix reducing width for a Text block

### LINKED ISSUE #3396: Resizing smaller not possible than initial size for a rotated block (without amp-fit-text)
Steps to reproduce:
- Create a text block and turn amp-fit-text off.
- Rotate the block a little bit.
- Add some text inside it so that it would be on two lines at least.
- Resize the text from the bottom so that there would be enough room to resize it smaller from the side.
- Try resizing smaller from the side (e.g. right side).

resizing

Expected: As long as there is room for the text, it should be possible to resize it smaller from any side.

miina:
Fixes #3396


miina:
Waiting for the tests to run to see if this might cause any regressions.

miina:
@barklund Since it looks like you worked on a related issue (#3199) -- would you mind reviewing?

barklund:
@googlebot I consent.

swissspidy:
We have some e2e tests for resizing text blocks. One of them is actually skipped at the moment, see #3326.

Sounds like a good opportunity to add a test for this scenario here too.

miina:
I was planning to add the tests (including looking into the skipped tests which are relevant) within #3397.

swissspidy:
Works for me

Issue #3325, opened 2019-09-23
Unable to move AMP story block using keyboard


spacedmonkey:
## Bug Description

When using keyboard navigation, users are unable to move movable blocks in the AMP Story interface. For accessible reasons, this should be possible. 

## Additional context

 - WordPress version:
 - Plugin version:
 - Gutenberg plugin version (if applicable):
 - AMP plugin template mode:
 - PHP version:
 - OS:
 - Browser: [e.g. chrome, safari]

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* After selecting a block, I should be able to move the block up, down, left and right via the direction keys. (QUESTION: HOW MUCH SHOULD IT MOVE ON EACH KEY STROKE?)

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

barklund:
It's not trivial, but here's a [great guide to best practice expected behaviour](https://www.w3.org/wiki/PF/ARIA/BestPractices/DragDrop).

swissspidy:
Previously suggested in #2454. Will close it as a duplicate.

spacedmonkey:
As a talking point, I have created a WIP pull request in #3400. It is still discovery stages at this point, but it is a workable solution.

spacedmonkey:
Screencast of the tabbing and arrows in use. 
https://www.youtube.com/watch?v=u280hGEgCd0&feature=youtu.be

spacedmonkey:
@MackenzieHartung @pbakaus A lot of work and discussion has taken place on #3400 . This ticket is not just a simple bug, it has lots of moving parts and is a fundamental change on how blocks are moved. This PR will effect other on going such as #3060 #3311  #3416 . This ticket needs to go through the new process, so we can capture all requirements

jauyong:
Hey @jasti I only added the one AC for now. Based on @spacedmonkey's comments I'm not sure if that is too high level and if more is needed. I will wait for you to move this to the IB column...

swissspidy:
In #3400 I left a comment for a new feature that could be its own AC:

> It would be nice if one could move in bigger steps when holding down an additional key, e.g. ctrl, alt, or cmd. If I am not mistaken this kind of behavior is common in other applications when moving elements with the keyboard.

So just arrow keys: move by single pixels
Meta key + arrow key: move in steps of N pixels

Issue #2788, opened 2019-07-10
Allow select-mode to select one or more elements


MackenzieHartung:
As a user I want to be able to engage "select" mode to select one or more elements at a time.

AC1: A CTA will be added to the editor that when selected will put the user in select mode.
AC2: When in select mode a user can click to select a block.
AC3: When in select mode a user can drag from an empty area on the page and select one or multiple elements by dragging the selection area over the element.
AC4: A user will remain in select mode until they double-click within a text block (related to issue #2783  )

swissspidy:
This depends on #2787, as otherwise it would not be possible to select a random list of blocks in this way.

swissspidy:
Gutenberg issue to follow: https://github.com/WordPress/gutenberg/issues/17088

swissspidy:
I don't think this 100% depends on #2787, as we can first start with single selection only, and then when #2787 is resolved we support selecting multiple elements.

PR #3400, opened 2019-10-01
Add keyboard options to blocks

### LINKED ISSUE #3280: Unexpected behavior when pressing the delete / backspace key in AMP Story editor
When a block is selected and delete key is pressed, sometimes a block is not remove / deleted. This leads to press the delete key multiple times, that then deletes the page block. Generally the bahavior of the delete key in the AMP Story editor, should be handled better for accessibility reasons.

spacedmonkey:
Add options to blocks to move and delete blocks. 

Fixes #3280 #3325

### DIFF
--- a/assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
+++ b/assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
@@ -1,0 +1,86 @@
+/**
+ * WordPress dependencies
+ */
+import { withDispatch } from '@wordpress/data';
+import { createHigherOrderComponent } from '@wordpress/compose';
+import { UP, DOWN, RIGHT, LEFT, DELETE } from '@wordpress/keycodes';
+/**
+ * Internal dependencies
+ */
+import { ALLOWED_CHILD_BLOCKS } from '../../constants';
+
+const applyWithDispatch = withDispatch( ( dispatch, props, { select } ) => {
+	const {	isReordering } = select( 'amp/story' );
+	const { getSelectedBlock } = select( 'core/block-editor' );
+	const { updateBlockAttributes, removeBlock } = dispatch( 'core/block-editor' );
+
+	const onKeyPress = ( event ) => {
+		const { keyCode } = event;
+		const selectedBlock = getSelectedBlock();
+		if(!selectedBlock){
+			return;
+		}
+
+		let top = 0;
+		let left = 0;
+		switch(keyCode){
+			case UP:
+				top = -1;
+				break;
+			case DOWN:
+				top = 1;
+				break;
+			case RIGHT:
+				left = 1;
+				break;
+			case LEFT:
+				left = -1;
+				break;
+			case DELETE:
+				removeBlock( selectedBlock.clientId );
+				return;
+			default:
+				return;
+		}
+		event.preventDefault();
+		const newPositionTop = selectedBlock.attributes.positionTop + top;
+		const newPositionLeft = selectedBlock.attributes.positionLeft + left;
+		updateBlockAttributes( selectedBlock.clientId, { positionTop: newPositionTop, positionLeft: newPositionLeft } );
+
+	};
+
+	return {
+		isReordering,
+		onKeyPress,
+	};
+} );
+
+/**
+ * Higher-order component that adds right click handler to each inner block.
+ *
+ * @return {Function} Higher-order component.
+ */
+export default createHigherOrderComponent(
+	( BlockEdit ) => {
+		return applyWithDispatch( ( props ) => {
+			const { name, onKeyPress, isReordering } = props;
+			const isPageBlock = 'amp/amp-story-page' === name;
+
+			// Add for page block and inner blocks.
+			if ( ! isPageBlock && ! ALLOWED_CHILD_BLOCKS.includes( name ) ) {
+				return <BlockEdit { ...props } />;
+			}
+
+			// Not relevant for reordering.
+			if ( isReordering() ) {
+				return <BlockEdit { ...props } />;
+			}
+			return (
+				<div className='KeyPress'  onKeyDown={ onKeyPress }>
+					<BlockEdit { ...props } />
+				</div>
+			);
+		} );
+	},
+	'withKeyboardNavigationHandler'
+);
--- a/assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
+++ b/assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
@@ -15,9 +15,15 @@ const applyWithDispatch = withDispatch( ( dispatch, props, { select } ) => {
 	const { updateBlockAttributes, removeBlock } = dispatch( 'core/block-editor' );
 
 	const onKeyPress = ( event ) => {
-		const { keyCode } = event;
+		const { keyCode, target } = event;
+		const { classList } = target;
 		const selectedBlock = getSelectedBlock();
+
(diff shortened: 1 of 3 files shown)


spacedmonkey:
This PR has the following issues. 
- Doesn't work on AMP text blocks
- Block with text editing and selected text, currently move. 
- Better handling for when arrow keys are held down.

spacedmonkey:
This PR is far from complete. But I have added for review as I would like some thoughts on this. 

The biggest issue, is not moving a block left and right when editing text. I have a very dirty work around for this, but I want something a little more full proof way of knowing a user is editing. This is hard to do, when it is a mixture of core and amp story blocks. I was thinking of having a `isEditing` field in the redux store. But this would require hook into core blocks and changing all amp story blocks. Also feel like this will not be full proof.

assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
```diff
@@ -0,0 +1,96 @@
...
+	const onKeyPress = ( event ) => {
+		const { keyCode, target } = event;
+		const { classList } = target;
+		const selectedBlock = getSelectedBlock();
+
+		if ( ! selectedBlock ) {
+			return;
+		}
+
+		if ( classList.contains( 'editor-rich-text__editable' ) && classList.contains( 'is-selected' ) ) {
```
swissspidy: Isn't there a `isSelected` prop?

assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
```diff
@@ -0,0 +1,96 @@
...
+				top = 1;
+				break;
+			case RIGHT:
+				left = 1;
+				break;
+			case LEFT:
+				left = -1;
+				break;
+			case DELETE:
+				removeBlock( selectedBlock.clientId );
```
swissspidy: What about the [existing keyboard shortcut](https://github.com/WordPress/gutenberg/blob/4202ab3c7bd150810f9f589eaae39b2a73fb2138/packages/block-editor/src/components/block-editor-keyboard-shortcuts/index.js) for removing blocks?

assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
spacedmonkey: Not that I can see 
```
attributes: {ordered: false, values: "", addedAttributes: 0, opacity: 100, positionTop: 30, …}
clientId: "b1540913-690e-4e23-8a0a-8f531c77301e"
innerBlocks: []
isValid: true
name: "core/list"
```

swissspidy:
> The biggest issue, is not moving a block left and right when editing text.

When I'm typing in a text block, I wouldn't expect arrow keys to move the block 🤔

assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
spacedmonkey: See https://github.com/ampproject/amp-wp/issues/3280

spacedmonkey:
> When I'm typing in a text block, I wouldn't expect arrow keys to move the block
Yes, that is what I am saying. When in text block is selected and the cursor is blinking in an editable field, left and right arrows, should move the cursor and not the block.

miina:
>  I was thinking of having a isEditing field in the redux store. But this would require hook into core blocks and changing all amp story blocks.

Hmm, maybe that's something we might possibly want to do anyway -- to implement edit vs drag-mode for all the text blocks (which would require extending all the core blocks anyway)? Currently, only the CTA block and Text block have the feature of being able to drag from anywhere while not editing and being able to edit after two clicks (and losing the dragging ability).

Wondering if the core blocks allowing text (quote, list, etc.) might require the same.

Also, I wonder if instead of `isEditing` field in the Redux store using React Context might be an option here. Not sure how much additional code re-rewriting this might require though.

Thoughts?

swissspidy:
That reminds me of #2788 and https://github.com/WordPress/gutenberg/issues/17088.

barklund:
I just noticed that the [`KeyboardShortcuts` components of `@wordpress/components`](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/keyboard-shortcuts) already handles text input focus and makes sure not to invoke the events if focus is inside a text field. You might be able to use it too (as I'm trying to use it for the context menu).

spacedmonkey:
A good example of Keyboard shortcuts usable for something similar in the gutenberg can be found [here](https://github.com/WordPress/gutenberg/blob/cdfb97c8fa61a43431e4c38fa5be9939ed052822/packages/components/src/color-picker/hue.js#L121-L137).

spacedmonkey:
I have updated the PR to use the `KeyboardShortcuts` component. I am still having the same issue editting text based blocks. The workaround I put in place doesn't work for many of the core blocks, as the classnames do not match. We are going to have to think of a new system of selecting blocks, maybe with a double click..

assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
```diff
@@ -0,0 +1,118 @@
...
+
+	return {
+		isReordering,
+		onMoveBlock,
+		deleteSelectedBlocks,
+	};
+} );
+
+/**
+ * Higher-order component that adds right click handler to each inner block.
```
swissspidy:
```suggestion
 * Higher-order component that adds a keyboard navigation handler to each inner block.
```

assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @return {Function} Higher-order component.
+ */
+export default createHigherOrderComponent(
+	( BlockEdit ) => {
+		return applyWithDispatch( ( props ) => {
```
swissspidy: Could use hooks here

assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
```diff
@@ -0,0 +1,118 @@
...
+			return;
+		}
+
+		if ( classList.contains( 'editor-rich-text__editable' ) && classList.contains( 'is-selected' ) ) {
+			return;
+		}
+
+		let top = 0;
+		let left = 0;
+		switch ( keyCode ) {
```
swissspidy: Needs some testing, but it would be nice if one could move in bigger steps when holding down an additional key, e.g. ctrl, alt, or cmd. If I am not mistaken this kind of behavior is common in other applications when moving elements with the keyboard

assets/src/stories-editor/components/higher-order/with-keyboard-navigation-handler.js
swissspidy: Odd 🤔 

I thought it does:

https://github.com/WordPress/gutenberg/blob/29ad4d8740b04b5bf094b6901cccd71552767cb3/packages/block-editor/src/components/block-list/block.js#L459-L470

https://github.com/WordPress/gutenberg/blob/29ad4d8740b04b5bf094b6901cccd71552767cb3/packages/block-editor/src/components/block-edit/index.js#L34-L43

https://github.com/WordPress/gutenberg/blob/29ad4d8740b04b5bf094b6901cccd71552767cb3/packages/block-editor/src/components/block-edit/edit.js#L12-L34

Issue #2102, opened 2019-04-09
Explore use of Symfony's CssSelector to do deeper tree shaking


westonruter:
Symfony has a [`CssSelector` component](https://symfony.com/doc/current/components/css_selector.html) which can be used to convert a CSS selector to an XPath expression. This in theory could be used to do evaluate a given selector against the document to determine whether it applies. This could shake quite a bit of CSS that we currently don't consider, for example selectors with multiple class names. Special care would be needed to account for selectors that interact with elements in an AMP document that are dynamically modified at runtime.

westonruter:
Alternatively, we could still use the [CSS selector parser](https://github.com/symfony/css-selector) to construct the XPath expressions with an awareness of which parts are dynamic in an AMP document. Note also:

https://github.com/ampproject/amp-wp/blob/4a78a2df53b00b9794b66406c9908d94a47f610b/includes/sanitizers/class-amp-style-sanitizer.php#L1510-L1513

westonruter:
This is not going to be so straightforward because the library requires PHP 7.1. The plugin currently requires PHP 5.4, though we should soon update to 5.6 like core. But this won't be feasible until PHP 7.1 is the minimum supported version. This could happen in core by the end of 2019, so we might be able to do it at that time.

Issue #1507, opened 2018-10-12
Sanitizer doesn't catch non-focusable element with `on` attribute for missing role & tabindex


westonruter:
Create a post with the content containing:

```html
<span on="tap:AMP.setState({clicked: !clicked})">Click Here</span>
```

When viewing the post on the frontend, the plugin's sanitizer doesn't complain at all about this. The AMP validator, however, does complain:

![image](https://user-images.githubusercontent.com/134745/46883821-05f80280-ce08-11e8-84e2-a77d5c04b49a.png)

Generally the plugin removes markup that is invalid rather than supply markup that is missing. But I'm curious why the plugin's sanitizer isn't removing the elements entirely when they lack the required attributes. Otherwise, the AMP plugin could add the missing `tabindex` and `role` attributes for such elements.

The reason why this isn't currently caught by the sanitizer may be because the rule is not captured in protoascii but rather it could be an ad hoc rule.

westonruter:
Closing in favor of #4447.

This issue also will rarely (if ever) occur now due to #4528.

PR #3403, opened 2019-10-01
Add compat support for twentytwenty core theme

### LINKED ISSUE #3342: Add theme support for Twenty Twenty (and WordPress 5.3)
WordPress 5.3 will have a new default core theme: [Twenty Twenty](https://make.wordpress.org/core/2019/09/06/introducing-twenty-twenty/). We need to make sure that it works as well as possible in AMP. 

Note that Twenty Nineteen support (#1587) did not have 100% feature parity in AMP specifically due to the navigation+ menu on desktop and the sidebar menu on mobile. Ideally we'd be able to achieve complete parity. Note that this may involve submitting patches to Twenty Twenty itself to reduce dependence on JavaScript for implementing the desired functionality.

schlessera:
Adds support for the new `twentytwenty` core theme that will be part of WordPress 5.3.

Changes and todo:
- [x] Dequeue custom JS with handle `'twentytwenty-js'`
- [x] Remove JS/No-JS script from `'wp_head'`
- [x] Reimplement cover modals functionality
- [x] Add back smooth scrolling
- [x] Reimplement primary menu keyboard focus
- [x] Reimplement modal menu(s)
- [x] Reimplement custom toggle code (is this needed after the above is fixed?)
- [x] Fix search modal
- [x] Fix spacing discrepancies ([screenshot](https://cl.ly/8e34a06a293e/Image%202019-10-01%20at%207.05.44%20PM.png)) - ~seems to be an upstream bug: https://github.com/ampproject/amphtml/issues/25032~ - not a bug, can be fixed by adding `display: block` to the image, see https://github.com/ampproject/amp-wp/issues/3493#issuecomment-542791803
- [x] ~Fix video scaling discrepancy ([screenshot](https://cl.ly/976b1294cd37/Image%202019-10-01%20at%205.43.19%20PM.png))~ - unable to replicate with latest theme code
- [x] ~Fix full-width embeds ([screenshot](https://cl.ly/40bfe2be05ac/Image%202019-10-01%20at%207.14.32%20PM.png))~ => fixed via https://github.com/ampproject/amp-wp/issues/3506
- [x] ~Fix broken `<pre>` tag ([screenshot](https://cl.ly/dfc6f15be0d6/Image%202019-10-01%20at%207.16.54%20PM.png))~ - see https://github.com/ampproject/amp-wp/pull/3403#issuecomment-541550931
- [x] Fix featured image ([screenshot](https://cl.ly/a60e35288d2f/Image%202019-10-01%20at%207.18.16%20PM.png))
- [x] ~Fix footer area ([screenshot](https://cl.ly/9ba2d6e01d4b/Image%202019-10-01%20at%207.22.40%20PM.png))~ - seems to have been an issue with the non-AMP version
- [x] ~Fix WordPress.TV embed adding videopress iframe custom JS script ([screenshot](https://cl.ly/d9451874509e/Image%202019-10-03%20at%204.05.07%20PM.png))~ => fixed via https://github.com/ampproject/amp-wp/issues/3508
- [x] ~Fix admin bar offset on mobile ([screenshot](https://cl.ly/e2d32b25b2b2/Image%202019-10-03%20at%204.16.23%20PM.png))~ => fixed via https://github.com/ampproject/amp-wp/issues/3436
- [x] Fix focus while opening search modal
- [x] Automatically keep current page visible in submenus
- [x] Adapt `aria-expanded` to toggle state
- [x] ~Twitter embed looks different, but seems broken in non-AMP instead (?) ([screenshot](https://cl.ly/71df55482f66/Image%202019-10-12%20at%2011.31.32%20AM.png))~ - see https://github.com/ampproject/amp-wp/pull/3403#issuecomment-541664134
- [x] Fix styling issue with mobile menu "Close" button ([screenshot](https://cl.ly/836d41a4abd9/Image%202019-10-14%20at%209.46.25%20AM.png))
- [x] Fix keyboard accessibility for dropdown menus (e.g. using `:focus-within`)
- [x] Write tests for the above 😓 (only reusable code is tested, see https://github.com/ampproject/amp-wp/pull/3403#issuecomment-543833277)

Fixes #3342.


includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1515,4 +1545,293 @@ public function add_twentyfourteen_search() {
...
+		$css_selector = trim( $css_selector );
+		$css_selector = preg_replace( '/\s+/', ' ', $css_selector );
+
+		/**
+		 * Provide a manual conversion from CSS selector to XPath query.
+		 *
+		 * @param string CSS selector that needs to be converted to XPath.
+		 * @return string XPath to use. Return the unchanged CSS selector if no conversion provided.
+		 */
+		$xpath = apply_filters( 'amp_xpath_from_css_selector', $css_selector );
```
westonruter: Using a filter here will be unreliable because it will make the post-processed output no longer deterministic based on the args being passed to the sanitizers.

Do we need to use XPath? See also #2102.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -456,8 +456,29 @@ private function get_used_class_names() {
...
+					$state_data = json_decode( $state_element->textContent, true );
+
+					if ( isset( $state_data[ $properties[ $index ] ] ) ) {
+						$classes .= ' ' . implode( ' ', $state_data[ $properties[ $index ] ] );
+					}
+				}
+
+				continue;
+			}
+
```
westonruter: Could you provide an example for what this is solving for?

includes/sanitizers/class-amp-core-theme-sanitizer.php
schlessera: Ah, yes, of course. I'll remove the filter again.

Also, great find with that Symfony package, that would come in handy. Would you consider it an option to use a quick 5.6 fork until we can use the real thing? Getting from 7.1 to 5.6 shouldn't be much more than stripping the return types for most of the code. Maybe it can even be done automatically with https://github.com/rectorphp/rector

If we have that fork available, we could also start looking into improving the tree shaker.

includes/sanitizers/class-amp-style-sanitizer.php
schlessera: I don't think it is solving anything right now.

Here's the current state of the code: I'm pretty much done with providing a generic replacement of the toggles and modals functionality of Twentytwenty. This means I'm not only using the markup of the theme to know where to use amp-state, amp-lightbox, etc, but it would also just work for people using twentytwenty as a base and extending it. If they add a new toggle or modal, it is automatically supported by AMP as well.

The one remaining issue is the styling right now, as I correctly add and remove the proper classes to the toggles, but these are still hidden because of a `display: none` style. I thought I could solve it by making sure the `active` class is not tree-shaken, but the behavior is more complex than that.

So the above code does not do much right now, and I will probably remove it again, because I need to solve this in a different way.

schlessera: Well, it does indeed solve the problem once the bugs are out... :)

schlessera:
I've finished building generic replacements for the "toggles" and "modals" that the theme provides. This now makes it functional and usable.

There are still lots of fine-tuning issues left, but the bulk of the work seems to be done.

schlessera:
The "spacing discrepancies" mentioned above seem to be related to the `<amp-image>` element or its parent `<figure>` element, and the way they size themselves. However, I couldn't find out the specific problem yet.

Image 2019-10-12 at 9 31 18 AM

The `<amp-img>` contains an image at 580x300px and is also explicitly set to these dimensions. The computed height is 300px.

For the surrounding `<figure>`, though, the computed height is 308px. It shows a small gap at the bottom of the image where it leaves 8px of space. I couldn't find any CSS that would account for this gap.

I assume for now that this spacing is incorrectly calculated inside of `<amp-img>`'s JS code.

schlessera:
It looks like the "broken `<pre>` tag" issue above points to a bigger problem of parsing/escaping HTML.

The issue is caused by the cat picture in the `<pre>` tag (not the opening and closing brackets around the neck):
Image 2019-10-12 at 11 42 05 AM
This results in the following broken HTML:
Image 2019-10-12 at 11 42 12 AM

So part of the cat is output as actual HTML tags to the browser, breaking the markup.

schlessera:
Regarding the broken `<pre>` tag:

I could trace the issue back to `$dom->loadHTML( $document );` here:
https://github.com/ampproject/amp-wp/blob/1.3/includes/utils/class-amp-dom-utils.php#L153

I could replicate it here: https://3v4l.org/kSM6Q

This is unrelated to the theme compat, so I will open a new bug report and ignore the issue here.

schlessera:
Regarding the Twitter embed issue raised above, I don't think this is something we should fix here. The "test" is to display the Twitter embed in a "wide block" layout. As the Twitter embed is not responsive, I don't think this actually makes sense to use in the first place. Also, if you use it nevertheless, right now AMP looks less broken than non-AMP. Therefore, I'm skipping work on this for now.

schlessera:
Regarding the remaining issue with `aria-expanded`:

Right now, this area of the theme is being actively worked on, with a general [a11y review](https://github.com/WordPress/twentytwenty/issues/785) being done. It might make sense to wait for this to be finished until we spend time developing AMP-analogs. I suggest doing so in a separate PR, so that we can merge this PR sooner rather than later.

Also, once we start working on adding `aria-expanded` logic, I'm afraid we need to switch back to using `<amp-state>` again instead of `toggleClass` for the toggle functionality. I don't see a way of toggling attributes other than the class, so this seems to only be achievable via `<amp-bind>`.

@westonruter thoughts?

schlessera:
Tests needed:
- [ ] Wrapping modals in `<amp-lightbox>` in `AMP_Core_Theme_Sanitizer::wrap_modal_in_lightbox()`
- [ ] Implement modals in `AMP_Core_Theme_Sanitizer::add_twentytwenty_modals()`
- [ ] Implement toggles in `AMP_Core_Theme_Sanitizer::add_twentytwenty_toggles()`
- [ ] Setting iframe layout in `alignfull` `<figure>` to `responsive`
- [ ] Video in cover block is set to `layout=fill`
- [ ] `AMP_Core_Theme_Sanitizer::add_amp_action()`
- [ ] `AMP_Core_Theme_Sanitizer::has_class()`
- [ ] WordPress.TV embed handler
- [ ] Automatically expanding submenus to make current page visible

schlessera:
@westonruter Before I start writing tests here, it would be good to know whether I can move some functionality into a more centralized location, namely `AMP_DOM_Utils`:

* [`AMP_Core_Theme_Sanitizer::get_element_id()`](https://github.com/ampproject/amp-wp/pull/3403/files#diff-5d05e505f93786061c2078486784c46dR1821-R1850)
* [`AMP_Core_Theme_Sanitizer::xpath_from_css_selector()`](https://github.com/ampproject/amp-wp/pull/3403/files#diff-5d05e505f93786061c2078486784c46dR1852-R1905)
* [`AMP_Core_Theme_Sanitizer::add_amp_action()`](https://github.com/ampproject/amp-wp/pull/3403/files#diff-5d05e505f93786061c2078486784c46dR1907-R1950)
* [`AMP_Core_Theme_Sanitizer::has_class()`](https://github.com/ampproject/amp-wp/pull/3403/files#diff-5d05e505f93786061c2078486784c46dR1952-R1967)

For the first one, this would allow communication of "generated IDs" across sanitizers, otherwise we might generate duplicate Ids once we use something similar in a different sanitizer.

For the other 3, there's no added benefit except for making it easier to test. Right now, these are protected methods within `AMP_Core_Theme_Sanitizer`, as they are not really the main responsibility of that class.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+			return $prefix;
+		}
+
+		$id = "{$prefix}-{$index_counter[ $prefix ]}";
+		$index_counter[ $prefix ] ++;
+
+		$element->setAttribute( 'id', $id );
+
+		return $id;
+	}
```
westonruter: Why not use/polyfill [`wp_unique_id()`](https://developer.wordpress.org/reference/functions/wp_unique_id/)?

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+	 *
+	 * This is a very simplistic conversion and will only work for very basic
+	 * CSS selectors. Therefore, it provides a filter for themes & plugins to
+	 * hook into to provide custom replacements.
+	 *
+	 * @param string $css_selector CSS selector to convert.
+	 * @return string|null XPath that closely mirrors the provided CSS
+	 *                     selector, or null if an error occurred.
+	 */
+	protected function xpath_from_css_selector( $css_selector ) {
```
westonruter: Yes, this makes sense to move to `AMP_DOM_Utils`.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+
+			$events[ $index ] = "{$event}:{$actions}";
+			$found            = true;
+		}
+
+		if ( ! $found ) {
+			$events[] = "{$event}:${action}";
+		}
+
+		$element->setAttribute( 'on', implode( ';', $events ) );
```
westonruter: This may warrant adding `tabindex=0` and `role=button` if the `$element` is not an `a` or a `button`. See #1507. Probably fine to consider later.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+	 * Register an AMP action to an event on a given element.
+	 *
+	 * If the element already contains one or more events or actions, the method
+	 * will assemble them in a smart way.
+	 *
+	 * @param DOMElement $element Element to add an action to.
+	 * @param string     $event   Event to trigger the action on.
+	 * @param string     $action  Action to add.
+	 */
+	protected function add_amp_action( DOMElement $element, $event, $action ) {
```
westonruter: Yes, this makes sense to add to `AMP_DOM_Utils`, with the below caveat comment.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+	 * @return bool Whether the element has the requested class.
+	 */
+	protected function has_class( DOMElement $element, $class ) {
+		if ( ! $element->hasAttribute( 'class' ) ) {
+			return false;
+		}
+
+		$classes = $element->getAttribute( 'class' );
+
+		return (bool) preg_match( "/\b{$class}\b/", $classes );
```
westonruter: To make this more robust:

```suggestion
		return (bool) preg_match( "/(^|\s){$class}(\s|$)/", $classes );
```

The `\b` will match hyphenated class names erroneously.

See also: 

https://github.com/ampproject/amp-wp/blob/cd1f8ae8bb5fcc89458767dde721797fe1158324/includes/sanitizers/class-amp-core-theme-sanitizer.php#L576-L582

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+	}
+
+	/**
+	 * Check whether a given element has a specific class.
+	 *
+	 * @param DOMElement $element Element to check the classes of.
+	 * @param string     $class Class to check for.
+	 * @return bool Whether the element has the requested class.
+	 */
+	protected function has_class( DOMElement $element, $class ) {
```
westonruter: Yes, makes sense to move to `AMP_DOM_Utils`.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+		$buttons = [
+			"{$modal_id}.open"  => $open_xpaths,
+			"{$body_id}.toggleClass(class=showing-modal,force=true)" => $open_xpaths,
+			"{$modal_id}.close" => $close_xpaths,
+			"{$body_id}.toggleClass(class=showing-modal,force=false)" => $close_xpaths,
+		];
+
+		foreach ( $buttons as $action => $button_xpaths ) {
+			foreach ( $button_xpaths as $button_xpath ) {
+				foreach ( $this->xpath->query( $button_xpath ) as $node ) {
```
westonruter: This `O(n^3)` looks scary. Would be good to add a comment explaining the expectation for the total number of iterations in the innermost loop.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+		$menu_item = $element;
+
+		while ( ! $this->has_class( $menu_item, 'menu-item' ) ) {
+			$menu_item = $menu_item->parentNode;
+			if ( ! $menu_item ) {
+				return $element;
+			}
+		}
+
+		$sub_menu = $this->xpath->query( "//*[ @class and contains( concat( ' ', normalize-space( @class ), ' ' ), ' sub-menu ' ) ]", $menu_item )->item( 0 );
```
westonruter: Nice use of `normalize-space()`. TIL!

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -117,6 +117,19 @@ public function sanitize() {
...
+				if ( $node->parentNode instanceof DOMElement && 'figure' === $node->parentNode->tagName ) {
+					$figure_node = $node->parentNode;
+				}
+				if ( $node->parentNode->parentNode instanceof DOMElement && 'figure' === $node->parentNode->parentNode->tagName ) {
+					$figure_node = $node->parentNode->parentNode;
+				}
+				if ( $figure_node && $figure_node->hasAttribute( 'class' ) && in_array( 'alignfull', explode( ' ', $figure_node->getAttribute( 'class' ) ), true ) ) {
+					$normalized_attributes['layout'] = 'responsive';
+				}
+
```
westonruter: Would this change be good to cherry-pick into a separate PR with its own issue?

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -461,8 +461,30 @@ private function get_used_class_names() {
 			$classes .= ' ' . $class_attribute->nodeValue;
 		}
 
-		// Find all [class] attributes and capture the contents of any single- or double-quoted strings.
+		// Find all [class] attributes and capture the contents of any single- or double-quoted strings,
+		// as well as <amp-state> content that might be used.
 		foreach ( $this->xpath->query( '//*/@' . AMP_DOM_Utils::AMP_BIND_DATA_ATTR_PREFIX . 'class' ) as $bound_class_attribute ) {
+			if ( preg_match_all( '/([a-zA-Z0-9_]+?)\[\s*([\'"])([^\2]*?)\2\s*\]/', $bound_class_attribute->nodeValue, $matches ) ) {
```
westonruter: Please provide some examples above for what this pattern will be matching.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: What are the classes that this is adding to `$classes`?

Perhaps this would be good to cherry-pick into a separate PR?

includes/sanitizers/class-amp-video-sanitizer.php
```diff
@@ -134,6 +134,11 @@ public function sanitize() {
 			$new_attributes = $this->filter_attachment_layout_attributes( $node, $new_attributes, $layout );
 			if ( empty( $new_attributes['layout'] ) && ! empty( $new_attributes['width'] ) && ! empty( $new_attributes['height'] ) ) {
 				$new_attributes['layout'] = 'responsive';
+
+				// The wp-block-cover__video-background class applies the 'object-fit: cover' CSS property, so we need to set the layout to 'fill'.
+				if ( $node->hasAttribute( 'class' ) && in_array( 'wp-block-cover__video-background', explode( ' ', $node->getAttribute( 'class' ) ), true ) ) {
+					$new_attributes['layout'] = 'fill';
+				}
```
westonruter: Perhaps good to cherry-pick into another PR?

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+			// Element.
+			if ( preg_match( '/^(?<element>[^.][a-zA-Z0-9-_]*)$/', $token, $matches ) ) {
+				$descendant        = $direct_descendant ? '/' : '//';
+				$xpath            .= "{$descendant}{$matches['element']}";
+				$direct_descendant = false;
+				$token             = strtok( ' ' );
+				continue;
+			}
+
+			$token = strtok( ' ' );
```
westonruter: Magic!

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+	 * @param string     $event   Event to trigger the action on.
+	 * @param string     $action  Action to add.
+	 */
+	protected function add_amp_action( DOMElement $element, $event, $action ) {
+		if ( ! $element->hasAttribute( 'on' ) ) {
+			$element->setAttribute( 'on', "{$event}:{$action}" );
+			return;
+		}
+
+		$events = explode( ';', $element->getAttribute( 'on' ) );
```
westonruter: I think this will be somewhat brittle because it's possible for `;` to appear inside of a method argument, though granted this is unlikely. Instead of exploding perhaps it would be better to inject `$action` into the attribute value right after matching `"$event:"`? Or if it doesn't exist, then just amend `";$event:$action"`? Since this doesn't have a proper parser of the action syntax, I'm wary of giving a false sense of security about using it generally.

In this case perhaps it is better not to include in `AMP_DOM_Utils`.

includes/sanitizers/class-amp-style-sanitizer.php
schlessera: Right now, this is not being used anymore. But it would be used again if we swicth back to using `<amp-bind>` for toggling classes, as we'll need it for making the `aria-expanded` toggling work (see https://github.com/ampproject/amp-wp/pull/3403#issuecomment-541666915).

In that case, this would be responsible for keeping the `active` class around, as it will not be visible in the initial markup but added via `setState()`.
I suggest removing this from here, and doing a separate a11y PR for `aria-expanded` once the a11y stuff for twentytwenty has settled.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1522,4 +1569,446 @@ public function add_twentyfourteen_search() {
...
+	public function add_twentytwenty_modals() {
+		$modals = $this->xpath->query( "//*[ @class and contains( concat( ' ', normalize-space( @class ), ' ' ), ' cover-modal ' ) ]" );
+
+		if ( false === $modals || 0 === $modals->count() ) {
+			return;
+		}
+
+		foreach ( $modals as $modal ) {
+
+			if ( ! $modal instanceof DOMElement ) {
```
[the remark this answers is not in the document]
schlessera: The syntax you've linked to is for the `element()` node-type test, i.e. "from a given set of nodes, only return the ones of type element".

The asterisk I'm using in my XPath is actually this one in the syntax: https://www.w3.org/TR/xpath-31/#doc-xpath31-Wildcard
It is then followed by a predicate: https://www.w3.org/TR/xpath-31/#doc-xpath31-Predicate to denote which nodes to retrieve.

I'm only getting nodes of type element because the full query (in unabreviated syntax) would be:
```
child:://*[ attribute::class and contains( concat( ' ', normalize-space( attribute::class ), ' ' ), ' cover-modal ' ) ]
```
So, generally, if no axis provided, the `child` axis is the default... most of the time:
```
If the axis name is omitted from an axis step, the default axis is child, with
two exceptions: (1) if the NodeTest in an axis step contains an AttributeTest or
SchemaAttributeTest then the default axis is attribute; (2) if the NodeTest in
an axis step is a NamespaceNodeTest then the default axis is namespace - in an
implementation that does not support the namespace axis, an error is raised
[err:XQST0134].
```
This means, in this specific case, I'm operating on the `child` axis, and the principal node type of that axis is the `element` (see https://www.w3.org/TR/xpath-31/#dt-axis-step).

This means I do indeed get only element nodes, but it is more complex than just having the `*` wildcard.

Also, the alternative (for proper IDE support) is to have this instead:
Image 2019-10-15 at 9 29 02 AM

I'm changing the checks into this docblock then (which is forced to be unnecessarily cumbersome by PHPCS), as you can argue that the actual check incurs a performance overhead. However, I'm not sure that is a big deal, and given the complexity of the issue, and the obscene alternative, this could just be considered defensive coding.

includes/sanitizers/class-amp-core-theme-sanitizer.php
schlessera: ... could you please "rephrase"? Not sure if this is good or bad here...

includes/sanitizers/class-amp-core-theme-sanitizer.php
schlessera: Because I didn't know about it... :)

Also, it looks broken to me. We'll end up with `wp-amp-toggle-357` and `wp-amp-toggle-523`... It should keep count of the index per prefix to build clean IDs.

includes/embeds/class-amp-wordpress-tv-embed-handler.php
```diff
@@ -0,0 +1,43 @@
...
+	 * @return string Filtered block content.
+	 */
+	public function filter_rendered_block( $block_content, $block ) {
+		if ( ! isset( $block['blockName'] ) || 'core-embed/wordpress-tv' !== $block['blockName'] ) {
+			return $block_content;
+		}
+
+		$modified_block_content = preg_replace( '#<script(?:\s.*?)?>.*?</script>#s', '', $block_content );
+
+		return null !== $modified_block_content ? $modified_block_content : $block_content;
```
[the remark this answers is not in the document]
schlessera: Just general defensive coding? Regex engines are changing between PHP versions, so this would ensure that, no matter what, we don't take the server down if something unexpected happens...

includes/sanitizers/class-amp-core-theme-sanitizer.php
schlessera: How about making it more robust then? The further we progress on "automating" the transformation into AMP code, the more we'll need to rely on events and actions.

includes/sanitizers/class-amp-iframe-sanitizer.php
schlessera: => https://github.com/ampproject/amp-wp/issues/3506

includes/sanitizers/class-amp-video-sanitizer.php
schlessera: I'm having trouble replicating this issue with latest `develop`. Will revisit later, it might be solved through a different PR already...

includes/sanitizers/class-amp-core-theme-sanitizer.php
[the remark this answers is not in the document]
westonruter: OK, I relent 😄 Carry on.

includes/embeds/class-amp-wordpress-tv-embed-handler.php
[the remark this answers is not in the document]
westonruter: OK, fair enough.

includes/sanitizers/class-amp-core-theme-sanitizer.php
westonruter: OK, please make it more robust then.

includes/sanitizers/class-amp-core-theme-sanitizer.php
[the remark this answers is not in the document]
schlessera: The checks were already replaced by docblock atrocities in https://github.com/ampproject/amp-wp/pull/3403/commits/6ce943b3875c3348963be2a8c056e298dbf278a3

I hope you're happy with yourself! 😉

includes/sanitizers/class-amp-core-theme-sanitizer.php
schlessera: @westonruter: Can you spot an issue with the following regex: https://regex101.com/r/PvNt9y/1 ?

I think this should be set up so that it only looks after the important bits and completely ignores anything within parentheses.

schlessera:
@westonruter I couldn't find any tests for theme compat for previous themes? Do you want me to still invest the time to write tests for the remaining items (see https://github.com/ampproject/amp-wp/pull/3403#issuecomment-541682635)?

It will be quite a bit of work to build such a theme test from scratch, and it will probably not test much more then my current assumptions anyway, as I don't have visual regression testing at my disposal yet. So it might not be worth the effort for now, and would be better to get this into QA as quickly as possible?

includes/sanitizers/class-amp-core-theme-sanitizer.php
schlessera: For the `aria-expanding` code, I need to refactor back into using `<amp-state>`, so this will be gone again.

schlessera: I made this clearer in https://github.com/ampproject/amp-wp/pull/3403/commits/abe0b0691a0a5ceabbf305602c7de02a386d4097

includes/sanitizers/class-amp-style-sanitizer.php
schlessera: This was removed as not needed anymore.

includes/sanitizers/class-amp-core-theme-sanitizer.php
schlessera: The problematic case you mentioned above is handled and tested.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -641,6 +662,32 @@ protected static function get_twentyseventeen_navigation_outer_height() {
...
+			static function() {
+				ob_start();
+				?>
+				<style>
+				.featured-media amp-img {
+					position: static;
+				}
+
+				.wp-block-image img {
+						display: block;
```
westonruter: Extra indentation here.
```suggestion
					display: block;
```

includes/utils/class-amp-dom-utils.php
```diff
@@ -711,4 +719,174 @@ public static function recursive_force_closing_tags( $dom, $node = null ) {
...
+	 * @return bool Whether the element has the requested class.
+	 */
+	public static function has_class( DOMElement $element, $class ) {
+		if ( ! $element->hasAttribute( 'class' ) ) {
+			return false;
+		}
+
+		$classes = $element->getAttribute( 'class' );
+
+		return (bool) preg_match( "/\b{$class}\b/", $classes );
```
westonruter: Still needs to be made more robust since `\b` matches in hyphens. This should do it:

```suggestion
		return in_array( $class, preg_split( '/\s/', $classes ), true );
```

includes/utils/class-amp-dom-utils.php
```diff
@@ -711,4 +719,174 @@ public static function recursive_force_closing_tags( $dom, $node = null ) {
...
+		return $id;
+	}
+
+	/**
+	 * Provides a "best guess" as to what XPath would mirror a given CSS
+	 * selector.
+	 *
+	 * This is a very simplistic conversion and will only work for very basic
+	 * CSS selectors. Therefore, it provides a filter for themes & plugins to
+	 * hook into to provide custom replacements.
```
westonruter: The filter no longer is present, so this can be removed.

includes/utils/class-amp-dom-utils.php
```diff
@@ -711,4 +719,174 @@ public static function recursive_force_closing_tags( $dom, $node = null ) {
...
+	 *
+	 * @since 1.4.0
+	 *
+	 * @param string $css_selector CSS selector to convert.
+	 * @return string|null XPath that closely mirrors the provided CSS selector,
+	 *                     or null if an error occurred.
+	 */
+	public static function xpath_from_css_selector( $css_selector ) {
```
westonruter: Given that this is very simplistic, and only used in the core theme sanitizer, how about moving it back into the core theme sanitizer and keep it `private` for now?

westonruter: Or else, rename it in a way that indicates it only supports basic selectors.

includes/utils/class-amp-dom-utils.php
schlessera: Fixed via https://github.com/ampproject/amp-wp/pull/3403/commits/5d7d8595155f80bdbd4771ea4033df40306a1469

includes/utils/class-amp-dom-utils.php
schlessera: Fixed via https://github.com/ampproject/amp-wp/pull/3403/commits/edafd6f6678c98a6cc213bd20daf169a5dcf15b9

includes/sanitizers/class-amp-core-theme-sanitizer.php
westonruter: Just that this is a special function that magically has internal state.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1023,6 +1023,18 @@ static function() use ( $args ) {
 						right: calc(100% + 2rem);
 					}
 
+					.admin-bar .menu-modal {
+						/* Use padding to shift down modal because amp-lightbox has top:0 !important. */
+						padding-top: 32px;
```
schlessera: Oh, I just noticed I stored this in my browser, but not in the PR...
Have to watch out for this next time, I thought it was fixed.

westonruter: Oh, how did you store it in your browser?

schlessera: @westonruter See https://developers.google.com/web/updates/2018/01/devtools#overrides

Issue #3407, opened 2019-10-01
Unable to open context menu using the keyboard


barklund:
As a user, I want to be able to open the context menu for blocks (both pages and elements) using the keyboard only.

It should be possible to open the context menu when a page or element is selected by pressing Shift+F10 (which [appears to be the reigning standard](https://www.google.com/search?q=accessibility%20context%20menu%20%22shift%20f10%22), though there's no hard specification for this).

### Testing instructions

1. Create a new story
2. Tab until the default empty text block is in focus
3. Hold down Shift and press F10.
4. A context menu should appear near the top left corner of the element

spacedmonkey:
This screencast is from develop on chrome + linux. 
https://youtu.be/8XRz_WWTiUo

barklund:
> This screencast is from develop on chrome + linux.
> https://youtu.be/8XRz_WWTiUo

@spacedmonkey, you tabbed to the rotation handle and opened the context menu. What is the expected behaviour and how did the function perform relative to that?

spacedmonkey:
The expected behaviour is what is seen in google slides. Which is what we have. 
https://youtu.be/C2dVJ-_5l2o

I would say, you tab into the block, once selected, should display when context when is triggered then menu is displayed.

csossi:
Verified in QA:
![image](https://user-images.githubusercontent.com/16945664/66576679-24d9c380-eb46-11e9-8b7d-cfb8c28a602c.png)

PR #3408, opened 2019-10-01
Enable access to context menu through keyboard

barklund:
_Fixes #3407._

This allows the user to open the context menu either through right-clicking (as before) or by pressing <kbd>Shift</kbd>+<kbd>F10</kbd>.


spacedmonkey:
This PR could do with a simple e2e.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
```diff
@@ -107,7 +135,7 @@ export default createHigherOrderComponent(
 			}
 
 			return (
-				<div onContextMenu={ onContextMenu }>
+				<div tabIndex="0" role="button" onContextMenu={ handleEvent } onKeyDown={ handleEvent }>
```
spacedmonkey: Is button the best role here?

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
```diff
@@ -55,20 +69,34 @@ const applyWithSelect = withSelect( ( select, props ) => {
...
-		const relativePositionX = event.clientX - wrapperDimensions.left;
-		const relativePositionY = event.clientY - wrapperDimensions.top - toolBarHeight;
+
+		let eventX = 0;
+		let eventY = 0;
+
+		if ( isContextmenu ) {
+			eventX = event.clientX;
+			eventY = event.clientY;
+		} else if ( isKeydown ) {
```
spacedmonkey: Why is this a else if? Is there a chance that at this point in the code, it would not be a keydown event?

spacedmonkey:
When doing some testing locally, I noticed that shift+f10 already works to open the context menu. What is the issue that we are trying to fix here?

miina:
In my testing (Chrome) I'm not seeing the context menu opening with <kbd><kbd>Shift</kbd>+<kbd>F10</kbd></kbd> unless on this PR.

Are you certain that it works on `develop`, too? I'm not seeing it working on `develop`.

Double checked but will check again.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
```diff
@@ -38,6 +42,16 @@ const applyWithSelect = withSelect( ( select, props ) => {
 			return;
 		}
 
+		// Ignore if it's a keydown event and the correct combo hasn't been pressed.
```
swissspidy: Can we document the right combo here in a comment? Just so it's more clear.

From what I've seen in a quick search, <kbd><kbd>Shift</kbd>+<kbd>F10</kbd></kbd> seems to be a Windows thing, and macOS doesn't even support any keyboard shortcut for that. Bummer.

swissspidy:
@spacedmonkey @miina It's an OS / browser thing. For example macOS doesn't support any keyboard shortcuts for the context menu at all. This PR now seems to make it available universally.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
barklund: The best I could find. It has to be an interactive role (otherwise linter complains about key handlers on non-interactive elements) and [the list of interactive roles](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-noninteractive-element-interactions.md#how-do-i-resolve-this-error) is very limited.

I'm open for other suggestions.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
barklund: Not currently. But I like to be on the safe side and it reads more cleanly – you know when reading the code what situation is handled by this branch – could also be solved by a comment of course. Should I change it to a `switch` perhaps?

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
barklund: Of course.

Should we try to feature detect and only add the key handler in case it's not supported natively? Also, how to check for that? I'll investigate...

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
swissspidy: Could we use the [`KeyboardShortcuts`](https://github.com/WordPress/gutenberg/tree/fec4a44ce4111bc3dfc628ae5ea4f58be1ebe6a5/packages/components/src/keyboard-shortcuts) components from the `@wordpress/components` package for this feature? It uses [Mousetrap](https://www.npmjs.com/package/mousetrap) under the hood. [Here is one usage example](https://github.com/WordPress/gutenberg/blob/fec4a44ce4111bc3dfc628ae5ea4f58be1ebe6a5/packages/block-editor/src/components/block-editor-keyboard-shortcuts/index.js) in Gutenberg.

It might then not be necessary to do any custom `role` stuff here.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
barklund: I don't think we can detect this. Sending simulated keyboard events won't trigger the context menu and you can't query about potential keys on the keyboard (most win keyboards have the "menu" key, that opens this menu just like Shift+F10.

We can detect OS, but that's a bit crude.

spacedmonkey:
It works for me on Linux / Ubuntu. Maybe we should check on Windows as well.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
swissspidy: Don't think we need to feature detect it. If the OS already supports it, we're not changing behavior, just re-implementing it.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
barklund: It's... tricky... but it might work.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
barklund: Fixed in 3fa7dd1.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
barklund: Fixed in 3fa7dd1. It actually wasn't that hard, but there is a pretty big limitation though. I don't know if it's an issue.

If you select an element with the mouse, you cannot press Shift+F10 (on Mac at least) to open the context menu (the focus is wrong and the `KeyboardShortcuts` handler doesn't trigger at all). However, if you tab to an element, then it works perfectly.

I think it's a bit unusual to use the mouse to navigate to the element but the keyboard to open the context menu, so I don't think it's an issue, but it might be. Will require real-world testing.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
barklund: Made irrelevant in 3fa7dd1.

assets/src/stories-editor/components/higher-order/with-right-click-handler.js
swissspidy: Hmm in my testing it also only worked when tabbing, but only when the block's drag handle was selected (because it's a button for a11y reason).

I was not able at all to use it on the CTA block.

swissspidy:
Quick summary:

* Doesn't work on CTA blocks
* Missing tests (unless there's a good reason for that)
* Doesn't work reliably when tabbing (works only when the drag handle is focused)

barklund:
> * Doesn't work on CTA blocks
> * Doesn't work reliably when tabbing (works only when the drag handle is focused)

Fixed in b64b91a.

> * Missing tests (unless there's a good reason for that)

Done in 9bcf822 and cfe0e12 (that latter one is pretty **D'oh!**).

tests/e2e/specs/stories-editor/context-menu.js
```diff
@@ -0,0 +1,177 @@
...
+			await insertBlock( 'Page Attachment' );
+			const callToActionSelector = '.wp-block[data-type="amp/amp-story-page-attachment"]';
+			const ctaBlock = await page.waitForSelector( callToActionSelector );
+			await makeRightClick( ctaBlock );
+			const duplicateSelector = 'right-click-duplicate';
+			expect( page ).not.toMatchElement( duplicateSelector );
+		} );
+	} );
+
+	describe( 'using Keyboard', () => {
```
spacedmonkey: Should we also add some other tests to check if the first element is selected and so users can navigate elements in our context menu using the down / up arrows.

barklund: Again I feel a strong need to point out, that e2e's aren't meant for exhaustive covering-all-the-bases-testing - not by a long stretch. It's meant for a testing a few happy paths. Please bear in mind, that every test added also adds 5 seconds to the build time of **every single future PR**. I don't think it's worth it.

Testing that the `ContextMenu` works, should be done in a test for the `ContextMenu`. This is merely a test that the ContextMenu shows up when the correct keyboard combination is pressed.

assets/src/stories-editor/components/higher-order/with-context-menu.js
```diff
@@ -56,49 +58,64 @@ const applyWithSelect = withSelect( ( select, props ) => {
...
+
+		let eventX = 0;
+		let eventY = 0;
+
+		if ( isRightClick ) {
+			// Use coordinates of actual click to place context menu.
+			eventX = event.clientX;
+			eventY = event.clientY;
+		} else {
+			// Place menu in the upper left corner of the element - but slightly inside.
```
swissspidy: Would it perhaps make sense to use the center of the element instead?

swissspidy: Asking because the current behavior looks odd for rotated blocks.

barklund: Great suggestion, I will check the possibilities.

assets/src/stories-editor/components/higher-order/with-context-menu.js
```diff
@@ -56,49 +58,64 @@ const applyWithSelect = withSelect( ( select, props ) => {
...
+			const elementPosition = event.target.getBoundingClientRect();
+			eventX = elementPosition.left + 20;
```
spacedmonkey:
```suggestion
		eventX = elementPosition.left + ( elementPosition.width / 2 );
```

assets/src/stories-editor/components/higher-order/with-context-menu.js
```diff
@@ -56,49 +58,64 @@ const applyWithSelect = withSelect( ( select, props ) => {
...
+			eventY = elementPosition.top + 20;
```
spacedmonkey:
```suggestion
			eventY = elementPosition.top + ( elementPosition.height / 2 );
```

assets/src/stories-editor/components/higher-order/with-context-menu.js
spacedmonkey: Added some Suggested to that works on the center point of element.

swissspidy:
After merging #3417 there are some merge conflicts that need to be resolved.

Once that is done, together with the updated position calculation, I think this is good to go.

PR #3410, opened 2019-10-02
Extract stories helpers into individual files

swissspidy:
## Summary

Addresses issue #3431.

The `helpers.js` file has grown quite a bit and it has become harder to maintain.

This PR is an attempt at splitting up all these helpers into individual files, which should make it easier to find, maintain, and test them.

## Checklist


swissspidy:
@miina @spacedmonkey @barklund What are your thoughts on doing this? Does it make sense or do you see it causing unnecessary overhead?

miina:
Does make sense to me to make it more manageable. I've thought about the `helpers` file going huge almost every time when adding a new helper so unless there is a good reason why keep it in one file then separate files make sense to me.

spacedmonkey:
In https://github.com/ampproject/amp-wp/pull/3312 Morten is already doing this. 

I say go ahead. Hopefully it will mean each helper functions will have a unit test to go with it.

swissspidy:
Can be merged once there's approval.

In a next step, we can look at tests for these utilities.

Also, if a helper function is only really used by one component, it should be moved to live next to it and not under `helpers`.

The helpers under `common` and `block-editor` are an entirely different beast to tackle separately as part of a larger effort.

PR #3411, opened 2019-10-02
Enforce stricter rules when linting React components

swissspidy:
* Enforce self-closing tags when there are no children
* Prevent unnecessary fragments
* Improve codee indentation
* Prevent unneeded prop types
* Prevent usage of unwrapped string literals

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -40,6 +40,15 @@
     "require-await": "error",
     "rest-spread-spacing": [ "error", "never" ],
     "react/prop-types": "error",
+    "react/jsx-closing-tag-location": "error",
+    "react/jsx-fragments": "error",
+    "react/jsx-first-prop-new-line": "error",
+    "react/jsx-max-props-per-line": [ "error", { "when":  "multiline" } ],
+    "react/jsx-no-literals": "error",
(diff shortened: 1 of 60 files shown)


miina:
Needs merging `develop` now.

Q: What does the `Prevent unneeded prop types` do? Does it mean that if the proptype is not actually present, it will display an error?

swissspidy:
Yes, that's exactly what the [`react/no-unused-prop-types`](https://github.com/yannickcr/eslint-plugin-react/blob/c153a86f092c65e9fb01ee5ce544302b67f040de/docs/rules/no-unused-prop-types.md) rule does.

We could even go further and even [forbid vague prop types](https://github.com/yannickcr/eslint-plugin-react/blob/c153a86f092c65e9fb01ee5ce544302b67f040de/docs/rules/forbid-prop-types.md) with more specific alternatives available (`any`, `array`, `object`). However, that doesn't seem that useful at the moment.

miina:
> Yes, that's exactly what the react/no-unused-prop-types rule does.

OK, that's great 👍

swissspidy:
Resolved merge conflicts now

PR #3412, opened 2019-10-02
Leverage React hooks in most areas of the stories editor

swissspidy:
## Summary

Addresses issue #2538.

Split out from ampproject/amp-wp#3359 to focus on the stories editor only for now.

Hooks cannot be used pre-WordPress 5.3 unless Gutenberg is active, which means they're currently only guaranteed to work in the Stories editor.

Doing this migration was very helpful for better seeing side effects in components and to simplify logic in them.

## Checklist


spacedmonkey:
Let mark this as draft and only merge it a week or two after 5.3 goes out.

swissspidy:
The Stories editor currently has a hard requirement to use latest version of the Gutenberg plugin (or soon WordPress 5.3), so I don't see a reason for waiting.

assets/src/stories-editor/blocks/amp-story-cta/edit.js
```diff
@@ -8,128 +8,97 @@ import PropTypes from 'prop-types';
...
-import {
-	Dashicon,
-	IconButton,
-} from '@wordpress/components';
-import {
-	URLInput,
-	RichText,
-} from '@wordpress/block-editor';
+import { useState, useEffect } from '@wordpress/element';
+import { Dashicon, IconButton } from '@wordpress/components';
```
spacedmonkey: Why were these changed a single line?

swissspidy: Didn't really make much sense to have these on separate lines as these are only two imports. So while I was it, I tidied up the imports here a bit.

I'm planning on configuring ESLint to enforce these kind of things if possible.

assets/src/stories-editor/blocks/amp-story-cta/edit.js
```diff
@@ -8,128 +8,97 @@ import PropTypes from 'prop-types';
...
+
+	const [ isEditing, setIsEditing ] = useState( false );
+	const [ hasOverlay, setHasOverlay ] = useState( true );
+
+	useEffect( () => {
+		if ( ! anchor ) {
+			setAttributes( { anchor: getUniqueId() } );
 		}
-	}
+	}, [ anchor, setAttributes ] );
```
miina: Hmm, why is `setAttributes` tracked here? Isn't `setAttributes` always the same?

assets/src/stories-editor/blocks/amp-story-cta/edit.js
```diff
@@ -8,128 +8,97 @@ import PropTypes from 'prop-types';
...
 		}
-		if ( this.state.isEditing && ! prevState.isEditing ) {
+	}, [ isSelected ] );
+
+	useEffect( () => {
+		if ( isEditing ) {
 			setInputSelectionToEnd( '.is-selected .amp-block-story-cta__link' );
 		}
-	}
+	}, [ isEditing ] );
```
miina: Curious if there's a reason for having a separate `useEffect` for `isEditing` and `isSelecting` instead of passing `[ isEditing, isSelected ]` together.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -61,45 +62,45 @@ const AttachmentContent = ( props ) => {
...
 		return () => {
-			isStillMounted = false;
+			isStillMounted.current = false;
 		};
 	}, [] );
 
 	useEffect( () => {
 		fetchSelectedPost();
-	}, [ postId ] );
+	}, [ fetchSelectedPost ] );
```
miina: ^ Why this change? Shouldn't we track the change of `postId` here instead?

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
```diff
@@ -61,45 +62,45 @@ const AttachmentContent = ( props ) => {
...
+					if ( isStillMounted.current && fetchRequest.current === currentFetchRequest ) {
 						setSelectedPost( null );
 						setFailedToFetch( true );
 						setIsFetching( false );
 					}
 				}
 			);
 		}
-	};
+	}, [ postId, postType, allowedPostTypes ] );
```
miina: Since `allowedPostTypes` is not something that the user can actually change, is it necessary to track the change here?

assets/src/stories-editor/blocks/amp-story-page-attachment/edit.js
```diff
@@ -81,42 +53,28 @@ class PageAttachmentEdit extends Component {
...
-			setAttributes,
-		} = this.props;
 
-		const {
-			openText,
-		} = attributes;
+		const newAttributes = { wrapperStyle: attachmentStyle };
+		newAttributes.attachmentClass = attachmentClass;
+		setAttributes( newAttributes );
+	}, [ colors, backgroundColor, customBackgroundColor, opacity, textColor, setAttributes ] );
```
miina: ^ Here as well it looks like there are some unnecessary values tracked. E.g. `colors` (since it's called only once with `useSelect`, also, it doesn't really change either probably dynamically) and `setAttributes` which is also a static function.

assets/src/stories-editor/blocks/amp-story-cta/edit.js
swissspidy: The ESLint rule for React hooks will complain if it's missing. See https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies.

We can't really guarantee that `setAttributes` is always the same as it's outside our control, but adding it as a dependency here doesn't hurt either way.

assets/src/stories-editor/blocks/amp-story-cta/edit.js
swissspidy: Do you mean because there are `isEditing` and `isSelected` checks within the effects anyway?

IMO it's a bit more explicit this way, so they only really fire when necessary.

assets/src/stories-editor/blocks/amp-story-page/animation-settings.js
```diff
@@ -27,24 +28,35 @@ import stopIcon from '../../../../images/stories-editor/stop.svg';
...
+			isPlayingAnimation: select( 'amp/story' ).isPlayingAnimation( clientId ),
+			animatedBlocks: getAnimatedBlocksPerPage( clientId ).filter( ( { id, animationType } ) => animationType && getBlock( id ) ),
+		};
+	}, [ clientId ] );
+
+	const { playAnimation, stopAnimation } = useDispatch( 'amp/story' );
+
+	const onAnimationStart = useCallback(
+		() => playAnimation( clientId ),
+		[ clientId, playAnimation ]
```
miina: ^ Here, too, we don't really need to track `playAnimation`

assets/src/stories-editor/blocks/amp-story-page/animation-settings.js
```diff
@@ -27,24 +28,35 @@ import stopIcon from '../../../../images/stories-editor/stop.svg';
...
+	);
+	const onAnimationStop = useCallback(
+		() => stopAnimation( clientId ),
+		[ clientId, stopAnimation ]
```
miina: Same here for `stopAnimation`.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
swissspidy: Functions that are used within an effect should be listed as dependencies, see https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

`fetchSelectedPost` itself is wrapped in `useCallback` which has `postId` as a dependency, so this is never out of sync

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
swissspidy: There's no benefit in omitting it, plus this helps to prevent issues in case this should ever change in the future.

assets/src/stories-editor/blocks/amp-story-page/background-media-settings.js
```diff
@@ -121,7 +121,7 @@ const BackgroundMediaSettings = ( {
 				} )
 				.catch( () => setExtractingPoster( false ) );
 		}
-	}, [ media, mediaType, videoFeaturedImage ] );
+	}, [ media, mediaId, mediaType, mediaUrl, poster, videoFeaturedImage, isExtractingPoster, setAttributes ] );
```
miina: ^ Here as well at least `setAttributes` is redundant, not sure about the others. Curious to know why were the other variables added here for change tracking, too. E.g. `poster`, `mediaUrl`, etc.

assets/src/stories-editor/blocks/amp-story-page-attachment/attachment-content.js
miina: Okay, good to know! 👍

assets/src/stories-editor/blocks/amp-story-page/background-media-settings.js
swissspidy: These variables are all used within the effect. If they change, they change the effect's logic too.

assets/src/stories-editor/blocks/amp-story-cta/edit.js
miina: Okay, good to know that there is a ESLint rule that complains about it! 

It looked odd to me to add the function as well especially since `anchor` is already being tracked.
In this case most of the other comments (if not all) I had can be omitted as well!

assets/src/stories-editor/blocks/amp-story-cta/edit.js
miina: Yep, it makes more sense here probably anyway since `isSelected` change might trigger `isEditing` change etc.

assets/src/stories-editor/blocks/amp-story-page/animation-settings.js
miina: In theory, we wouldn't need to add `playAnimation` here since `clientId` already has been added, based on https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

However, I guess there's no harm doing it either, as you mentioned. It's just weird to read for me (will get used to it).

assets/src/stories-editor/blocks/amp-story-page/background-media-settings.js
miina: Yep, I can see that these are all used. Wondering if something wasn't working before due to the variables not being tracked.

barklund: There could definitely have been times where it should have had an effect but didn't, but they might have been unreachable states due to other circumstances.

E.g. if `isExtractingPoster` was `true` and it was the only thing holding up this block `if ( media && ! media.featured_media && ! isExtractingPoster ) {`, changing that flag to `false` wouldn't have caused the effect to run again and upload the video frame. But that probably wasn't a possible state.

PR #3433, opened 2019-10-03
Fix sliding when resizing with minimum limits reached

### LINKED ISSUE #3326: Unskip resizing test after bugfix

### LINKED ISSUE #3397: Resizing Text block: Trying to resize smaller than possible moves the position

miina:
## Summary

Fixes #3397
Fixes #3326 

- Unskips the resizing tests.
- Prevents position change when minimum limits have been reached when reducing the width/height.
- Prevents `scrollHeight` calculations to limit resizing if the Text block is empty anyway.

## Checklist

- [x] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/issues/3397).
- [x] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md).


miina:
OK, found the reason for tests being unreliable: there was an `await` missing 🤦‍♀ 

However, this brought out some actual additional bug when fit-text is turned off, so looking into that now.

tests/e2e/specs/stories-editor/resizing.js
```diff
@@ -269,34 +269,78 @@ describe( 'Resizing', () => {
...
+			// Get the initial position.
+			const { positionTop: positionTopBefore } = await getSelectedBlockPosition();
+			// Try resizing again.
+			await dragAndResize( resizableHandleTop, { x: 0, y: 250 } );
+			const { positionTop } = await getSelectedBlockPosition();
+			expect( positionTop ).toStrictEqual( positionTopBefore );
+
+			// Rotate the block and try again.
+			await rotateSelectedBlock();
+			await dragAndResize( resizableHandleTop, { x: 300, y: 0 } );
```
barklund: Here you switch to dragging in `x` direction in stead of `y` direction after rotation. But in the above test, you drag in `x` both before and after rotation. I don't know which is the right one, but it seems to be this one, perhaps?

I guess that depends on how the item is rotated of course 😝.

barklund:
If you resize "quickly", then the item jumps when you reach minimum position, as it jumps to the last "legal" top position and minimum height from there in stead of minimum height based off the "static" bottom position.

This is particularly obvious with a rotated block, but happens for non-rotated ones as well.

I tried to see, if I could fix this, but I just can't wrap my head around it. Maybe for non-rotated blocks, but definitely couldn't figure it out for the rotated ones.

Screen Recording 2019-10-03 at 17 07 27

miina:
Thanks for testing! I think I remember seeing this "jumpiness" on quick resizing from before as well, it's not easy to test on `develop` though since currently the block slides away instead (without this PR).

Will try to look into this jumpiness as well in this PR, if it starts taking too much time then will create a separate issue instead.

tests/e2e/specs/stories-editor/resizing.js
miina: Hah! Good catch, thank you. The rotation is not enough for the initial direction not to work -- will leave a comment with degrees.

spacedmonkey:
I am going to wait to code review until the  "jumpiness"  is fixed.

miina:
Added a new issue for the jumpiness. It's unrelated to the original issue anyway and possibly lower priority, too.
https://github.com/ampproject/amp-wp/issues/3444

swissspidy:
Still seeing the issue: https://cloudup.com/cRZMrlUPdy8

miina:
> Still seeing the issue: https://cloudup.com/cRZMrlUPdy8

Hmm, that seems to be a different issue -- the original issue for the position continuing to change even though the minimum limits have been reached and it's not possible to reduce anymore (in case of amp-fit-text turned off).

I'll debug this new finding as well to see if it seems to be an easy fix. For testing/QA purposes might be good to track that as a different issue though.

Looks like that's for when "amp-fit-text" is turned on, right?

swissspidy:
> Looks like that's for when "amp-fit-text" is turned on, right?

Correct. A separate issue is fine for me. Thanks for having a look!

miina:
OK, looks like the tests have passed a few times in a row now.

Issue #1294, opened 2018-07-31
Introduce and expose query arguments for debugging AMP content generation


westonruter:
There is currently a query that can be added to an AMP URL to facilitate debugging:

* `amp_preserve_source_comments`: This prevents source comments from being removed when performing an `amp_validate` request.

This isn't exposed anywhere in the UI to discover, and the developer just has to know it exists. Maybe this is fine, but there should be a Wiki page with debugging information to inform of its use. ~Alternatively, when `WP_DEBUG` is enabled there could be a link added to the AMP admin bar menu to link to it.~ (Update: UI to be addressed in a subsequent issue.)

But beyond `amp_preserve_source_comments` there is probably other flags that would be useful for debugging:

* Disable output buffering and short-circuit the sanitizers. This would facilitate inspection of the response from WordPress prior to the post-processors kicking in. This could be particularly helpful debugging embed handlers as well as figuring out where a fatal error may be happening.
* Disable response cache. This would be useful to help with performance debugging since there is a response cache to prevent post-processors from re-processing the same HTML repeatedly.
* Exclude sanitizers from being run. This could help with identifying which sanitizer is causing problems.
* Prevent redirection from AMP URL to non-AMP URL when there are unaccepted validation errors. The `amp_validate` param does this already, but it also has the behavior of adding validation error source comments, so having another param may be helpful.
* Automatically reject all validation errors on a given page.
* Others?

These debug parameters should only be made available when an appropriately-authorized user is authenticated.

westonruter:
The `amp_validate` query param can also be used to preview what a site in classic mode will look like when theme support is enabled. The user must be an admin for this to be honored.

kienstra:
**Question About Working On This**

Hi @westonruter,
Hope your week's off to a great start.

It looks like this is assigned to you. Could I work on it? Otherwise, no problem 😄 

Thanks, Weston!

westonruter:
@kienstra Yes, please go for it.

kienstra:
Thanks, Weston!

kienstra:
**Question about a query arg**

Hi @westonruter,
Happy friday!

>Exclude sanitizers from being run. This could help with identifying which sanitizer is causing problems.

Should this exclude only a single sanitizer, like `?exclude_sanitizer=nav-menu-dropdown` 

Or would this somehow be different from the other query arg:

[...]

PR #3448, opened 2019-10-04
Add debugging query arguments

kienstra:
## Summary
Adds query arguments for debugging under the `amp_flags` query arg. For example:
`https://example.com/baz-post/?amp&amp_flags[disable_post_processing]=1`

| Query argument        | Description    | 
| ------------- |-------------|
|  `disable_post_processing`  | Prevents output buffering and sanitizers from running  |
| `disable_response_cache`  | Prevents the response from being cached      | 
| `prevent_redirect` | Prevents redirecting from an AMP URL to a non-AMP URL, like if there are unaccepted validation errors in Transitional mode  | 
| `reject_all_errors` | Automatically sets all errors to 'ack rejected'    | 
| `accept_excessive_css` |  Accepts all `excessive_css` errors. Tree shaking still takes place, but no styling is removed because it's over the limit.  | 
| `disable_amp` | Disables AMP for the URL | 
| `disable_tree_shaking ` | Disables tree shaking |

@todo: 
- [x] Exclude sanitizers (already covered in `disable_post_processing`)
- [x] Add a UI or wiki page, documenting the query arguments (Update: add a UI [like this](https://github.com/ampproject/amp-wp/pull/3448#issuecomment-538582596))

Addresses issue #1294

## Checklist


westonruter:
> * Add a UI or wiki page, documenting the query arguments

I think these would work well as toggles in an submenu item of the AMP admin bar menu item, if the user has the capability to use them. Clicking on a nav menu item can toggle the option on/off as it reloads the page.

includes/class-amp-theme-support.php
```diff
@@ -1872,6 +1893,10 @@ public static function start_output_buffering() {
 			newrelic_disable_autorum();
 		}
 
+		if ( isset( $_GET[ self::DISABLE_POST_PROCESSING_QUERY_VAR ] ) && AMP_Validation_Manager::has_cap() ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
westonruter: Instead of introducing these as top-level query parameters, how about putting them all under some `amp_flags` query var? This would make it easier to scrub them from the URL and it would reduce opportunities for collision. So So here it could be checking `isset( $_GET['amp_flags'][ self::DISABLE_POST_PROCESSING_QUERY_VAR ] )`.

includes/class-amp-theme-support.php
```diff
@@ -2278,6 +2315,15 @@ public static function prepare_response( $response, $args = [] ) {
 					$non_amp_url = add_query_arg( AMP_Validation_Manager::VALIDATION_ERRORS_QUERY_VAR, $blocking_error_count, $non_amp_url );
 				}
 
+				// Remove debugging query args.
+				$non_amp_url = remove_query_arg(
+					[
+						self::DISABLE_RESPONSE_CACHE_QUERY_VAR,
+						AMP_Validation_Error_Taxonomy::REJECT_ALL_VALIDATION_ERRORS_QUERY_VAR,
+					],
```
westonruter: Per above, this could just remove `amp_flags`.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -465,6 +472,11 @@ public static function get_validation_error_sanitization( $error ) {
 			$forced = 'with_preview';
 		}
 
+		if ( isset( $_GET[ self::REJECT_ALL_VALIDATION_ERRORS_QUERY_VAR ] ) && AMP_Validation_Manager::has_cap() ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
westonruter: As opposed to adding a new condition here, I suggest using the `amp_validation_error_sanitized` filter below, since these debug scenarios are outside the normal band of validation.

kienstra:
Hi @westonruter,

>I think these would work well as toggles in an submenu item of the AMP admin bar menu item, if the user has the capability to use them. Clicking on a nav menu item can toggle the option on/off as it reloads the page.

Ah, interesting. Something like this?

maybe-submenu-item

westonruter:
Yes, something like that. I realize that some of the toggles are mutually-exclusive, so that complicates it a bit.

includes/class-amp-theme-support.php
kienstra: That's a really good idea.

includes/class-amp-theme-support.php
kienstra: Sure, I'll apply this also.

kienstra:
Ah, the [commit above](https://github.com/ampproject/amp-wp/pull/3448/commits/e738bcfcb568cf5786d96fb68c6b3ed78f691a86) needs reworking.

includes/validation/class-amp-validation-error-taxonomy.php
kienstra: Good idea, how does [this look]( https://github.com/ampproject/amp-wp/pull/3448/files#diff-ff6d70c0505290e11be8ca83967c2af0R285)?

includes/class-amp-theme-support.php
kienstra: https://github.com/ampproject/amp-wp/pull/3448/commits/19363fc23939a38173ecfe0bd4d4baad3de2583a adds an [amp_flags query var](https://github.com/ampproject/amp-wp/pull/3448/commits/19363fc23939a38173ecfe0bd4d4baad3de2583a#diff-2585472f207548f626a43a7ff3cab922R103), to access these query vars like you mentioned.

For example:
`https://loc.test/your-post/?amp&amp_flags[reject_all_errors]`

includes/class-amp-theme-support.php
kienstra: Good idea, it's [applied here](https://github.com/ampproject/amp-wp/pull/3448/commits/19363fc23939a38173ecfe0bd4d4baad3de2583a#diff-2585472f207548f626a43a7ff3cab922R2326).

swissspidy:
Should the query vars be prefixed to not cause conflicts with other plugins?

kienstra:
Hi @swissspidy,
>Should the query vars be prefixed to not cause conflicts with other plugins?


Good question. https://github.com/ampproject/amp-wp/pull/3448/commits/19363fc23939a38173ecfe0bd4d4baad3de2583a adds an [amp_flags](https://github.com/ampproject/amp-wp/pull/3448/commits/19363fc23939a38173ecfe0bd4d4baad3de2583a#diff-2585472f207548f626a43a7ff3cab922R103) query var, to access these query vars like:

`https://loc.test/your-post/?amp&amp_flags[reject_all_errors]`

...which is accessed like:

```php
$_GET[ AMP_Theme_Support::AMP_FLAGS_QUERY_VAR ][ self::REJECT_ALL_VALIDATION_ERRORS_QUERY_VAR ] )
```

kienstra:
The UI for these query vars still needs a lot of work, but here's its current state:

query-vars-debugging

westonruter:
> The UI for these query vars still needs a lot of work, but here's its current state:

@kienstra I think the submenu could be made a bit simpler by removing the checkboxes. In other words, each of the nav menu items would be a link which would be preceded by some on/off indicator. Clicking the link would cause the page to reload with the query var update(s), and opening the submenu would then show the toggle indicator had been flipped.

kienstra:
Hi @westonruter, that's a really good idea. It'll make this much simpler.

kienstra:
Here's the PHP-rendered submenu. Each node, like 'Disable post-processing', is a link to the AMP URL with that query var.

new-submenu-here

includes/validation/class-amp-validation-manager.php
```diff
@@ -519,6 +522,63 @@ function() {
...
+	 * For example, if the URL has the 'Prevent redirect' query var present,
+	 * the title in the Admin Bar will be '✅ Prevent redirect'.
+	 *
+	 * @param string $title_text The title of the option, like 'Prevent redirect'.
+	 * @param string $query_var The query var of the option.
+	 * @return string The debugging option title, possible with a 'check mark' emoji at the beginning.
+	 */
+	public static function get_debugging_option_title( $title_text, $query_var ) {
+		if ( isset( $_GET[ AMP_Theme_Support::AMP_FLAGS_QUERY_VAR ][ $query_var ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			return '✅ ' . $title_text;
```
westonruter: Instead of this emoji, let's perhaps use these:

* “☑” (U+2611 Ballot Box with Check)
* “☐” (U+2610 Ballot Box)

When the option is enabled, the first should be used, and when the option is not enabled the second should be used. This will help with communicating that clicking on the link has the effect of toggling the option.

There will be a need to set the `font-family` explicitly to prevent the first one from being rendered as an emoji. So it can be wrapped in a `span`. They should end up looking like:

image

kienstra: Thanks, those look good!

kienstra: I'll change this to use those.

kienstra: Hi @westonruter, how does this look, with https://github.com/ampproject/amp-wp/pull/3448/commits/ef0720cdfaa78fa38623fe7ff39a9b43344290dc?
new-emojis

kienstra:
>There will be a need to set the font-family explicitly to prevent the first one from being rendered as an emoji. So it can be wrapped in a span.

Also, did you have a `font-family` in mind? This [uses](https://github.com/ampproject/amp-wp/pull/3448/commits/ef0720cdfaa78fa38623fe7ff39a9b43344290dc#diff-7054f237c4bba5d39484cc1990dd8810R576) some fonts from [admin-bar.css](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/css/admin-bar.css#L12), but doesn't use `-apple-system, BlinkMacSystemFont`.

schlessera:
As debugging is obviously a big and important part of using the AMP plugin, I think sprinkling debugging code throughout the other pieces of logic too much is a problem in terms o technical debt.

I suggest adding a separate class that deals with this responsibility and abstracts away the details (abbreviated, untested example):
```php
final class AMP_Debug {

	/**
	 * A top-level query var for AMP debug flags.
	 *
	 * @var string
	 */
	const QUERY_VAR = 'amp_flags';

	/**
	 * A query var to disable post processing.
	 *
	 * @var string
	 */
	const DISABLE_POST_PROCESSING = 'disable_post_processing';

	/**
	 * A query var to disable the response cache.
	 *
	 * @var string
	 */
	const DISABLE_RESPONSE_CACHE = 'disable_response_cache';

	/**
	 * A query var to prevent a redirect to a non-AMP URL.
	 *
	 * @var string
	 */
	const PREVENT_REDIRECT_TO_NON_AMP = 'prevent_redirect';

	/**
	 * Check whether a given debug flag is currently set.
	 *
	 * @param string $flag Flag to check.
	 * @return bool Whether the given debug flag is set.
	 */
	public static function has_flag( $flag ) {
		static $debug_flags = null;

		if ( null === $debug_flags ) {
			$debug_flags = filter_input_array(
				INPUT_GET,
				[
					AMP_Debug::QUERY_VAR => [
						AMP_Debug::DISABLE_POST_PROCESSING     => FILTER_VALIDATE_BOOLEAN,
						AMP_Debug::DISABLE_RESPONSE_CACHE      => FILTER_VALIDATE_BOOLEAN,
						AMP_Debug::PREVENT_REDIRECT_TO_NON_AMP => FILTER_VALIDATE_BOOLEAN,
						// [...]
					],
				]
			);
		}

		return array_key_exists( $flag, $debug_flags ) ? (bool) $debug_flags[ $flag ] : false;
	}
}
```

This allows us to keep everything in one place that is directly related to dealing with that (like adding the debugging menu to the admin bar), and also lets us streamline the naming (as we have the class itself as the context.

Then, within the other logic, a check turns from this:
```php
isset( $_GET[ AMP_Theme_Support::AMP_FLAGS_QUERY_VAR ][ AMP_Theme_Support::DISABLE_POST_PROCESSING_QUERY_VAR ] )
```
into this:
```php
AMP_Debug::has_flag( AMP_Debug::DISABLE_POST_PROCESSING )
```

Also the above now properly respects the expectations we have for booleans, like returning false if the URL variable is `?amp_flags[disable_post_processing=false]`.

kienstra:
Hi @schlessera,
That's a really good idea to use `AMP_Debug`, prevent the logic from being mixed into several classes.

includes/validation/class-amp-debug.php
```diff
@@ -0,0 +1,107 @@
...
+	 */
+	public static function has_flag( $flag ) {
+		if (
+			isset( $_GET[ self::AMP_FLAGS_QUERY_VAR ][ $flag ] ) // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			&&
+			array_key_exists( $flag, self::get_all_query_vars() )
+			&&
+			AMP_Validation_Manager::has_cap()
+		) {
+			return filter_var( $_GET[ self::AMP_FLAGS_QUERY_VAR ][ $flag ], FILTER_VALIDATE_BOOLEAN ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
kienstra: Using `filter_input_array()` would have been better than having to pass `$_GET` to `filter_var()`, but I couldn't figure out how to test this, without something like:

```php
Monkey\Functions\expect( 'filter_input_array' ) 
...
```

Monkey [isn't available](https://github.com/ampproject/amp-wp/blob/65208999a6e817ff7c74afc31b5e622c971326b0/composer.json#L27) in `composer.json`, though we could propose adding it if was important enough. Also, maybe there's another way to mock `filter_input_array()`.

kienstra:
With https://github.com/ampproject/amp-wp/pull/3448/commits/996157a242412354099945d448d5a17b72f10b1a, the behavior of the debugging query vars is a little different

For example, here's the return value of `AMP_Debug::has_flag( 'reject_all_errors' )`:

##  True
`&amp_flags[reject_all_errors]=1`
`&amp_flags[reject_all_errors]=true`

## False
`&amp_flags[reject_all_errors]=0`
`&amp_flags[reject_all_errors]=false`
`&amp_flags[reject_all_errors]=made-up`

includes/class-amp-autoloader.php
```diff
@@ -101,6 +101,7 @@ class AMP_Autoloader {
 		'AMP_Validation_Callback_Wrapper'    => 'includes/validation/class-amp-validation-callback-wrapper',
 		'AMP_Validated_URL_Post_Type'        => 'includes/validation/class-amp-validated-url-post-type',
 		'AMP_Validation_Error_Taxonomy'      => 'includes/validation/class-amp-validation-error-taxonomy',
+		'AMP_Debug'                          => 'includes/validation/class-amp-debug',
```
kienstra: It's probably debatable whether `class-amp-debug.php` should be in the `validation/` directory. Most of the debugging flags have to do with validation, but some don't, like `disable_amp`.

schlessera:
> With 996157a, the behavior of the debugging query vars is a little different.

What about the flag without value?
```
&amp_flags[reject_all_errors]
```

kienstra:
>What about the flag without value?

Ah, good question. That'll be `false`.

kienstra:

Thanks for your help on this. 

When you have a chance, could you please do another round of review?

The [table of flags above](https://github.com/ampproject/amp-wp/pull/3448#issue-324859822) is still up-to-date, but the [query var now needs a value](https://github.com/ampproject/amp-wp/pull/3448#issuecomment-545987328), like:

`&amp_flags[reject_all_errors]=true`

schlessera:
> Ah, good question. That'll be false.

@kienstra Would it make sense to change this so that no flag value means `true`?

kienstra:
>@kienstra Would it make sense to change this so that no flag value means true?

Sure, that's the behavior for the `amp` query var.

includes/validation/class-amp-validation-manager.php
pierlon: Any reason why this can't be a checkbox input?

includes/validation/class-amp-validation-manager.php
```diff
@@ -519,6 +522,50 @@ function() {
...
+	 * For example, if the URL has the 'Prevent redirect' query var present,
+	 * the title in the Admin Bar will be '☑ Prevent redirect'.
+	 *
+	 * @param string $title_text The title of the option, like 'Prevent redirect'.
+	 * @param string $query_var The query var of the option.
+	 * @return string The debugging option title, possible with a 'check mark' emoji at the beginning.
+	 */
+	public static function get_debugging_option_title( $title_text, $query_var ) {
+		$checkbox = AMP_Debug::has_flag( $query_var ) ? '☑' : '☐';
+		return sprintf( '<span class="amp-debug-option" style="font-family:Roboto,Oxygen-Sans,sans-serif;font-size:17px">%s</span> %s', $checkbox, esc_html( $title_text ) );
```
pierlon: Any reason why this can't be a checkbox input?

kienstra: Yeah, a checkbox could be good, it'd just add some complexity 😄 

There used to be checkboxes in this PR, here are [some](https://github.com/ampproject/amp-wp/pull/3448#issuecomment-541987039) [comments](https://github.com/ampproject/amp-wp/pull/3448#issuecomment-542776417) about that.

kienstra:
>Excellent PR so far @kienstra. The only concern I have right now is that with the use of checkboxes it gives the impression that multiple flags can be assigned at once, which is not so. I would suggest using something similar to the radio input.

Hi @pierlon,
Thanks a lot for reviewing this.

That's an interesting suggestion. It's possible, though not likely, that 2 debug flags would be present at once.

Still, one would have to manually add those query vars, it's not possible to get 2 debugging flags by simply clicking in the admin bar.

For example:
`
https://example.com/?amp&amp_flags[reject_all_errors]&amp_flags[prevent_redirect]
`

amp-flags-prevent-redirect

westonruter:
> That's an interesting suggestion. It's possible, though not likely, that 2 debug flags would be present at once.

Ah, this actually is the intention here, to allow multiple debug flags to be checked at once. A tricky bit, however, is that various debug flags are mutually-exclusively (https://github.com/ampproject/amp-wp/pull/3448#issuecomment-538585575), so there would need to be some logic that removes/disables some of the options based on the state of the others.

westonruter:
@kienstra Let's put this on pause until we can flesh out more of the acceptance criteria and determine how the different options relate to each other, and now those options apply in Standard vs Transitional modes.

kienstra:
Sounds good, thanks

PR #3453, opened 2019-10-07
CTA button responsiveness

### LINKED ISSUE #3343: CTA button is only initially centered, needing re-centering after adding text
## Bug Description

Something doesn't seem quite right about the CTA block: the button is apparently aligned to the center only prior to text populating it, meaning that as soon as soon as you supply text the button is off center. It doesn't seem ideal to require the user to manually re-center after supplying text. There may also be issues here with responsiveness.

Also, the URL field needs some border when not focused as right now it is dangling in the phantom zone.

## Steps to reproduce

image

---------------

## Acceptance criteria

1. After positioning the CTA in the editor, it should be in the same position in the preview and live mode
2. By default, the button should be positioned 5% left as the other blocks.
3. The button should look similar in the FE and in the editor, e.g. if it goes over the edge in the editor, then the part over the edge shouldn't be visible in the FE either.

## Implementation brief

* 

## QA testing instructions

* Add a CTA block
* Add some text inside
* Drag the block to different positions, e.g. left / center / right
* With each position and text verify that the editor view and FE view would look similar.

## Demo

* 

## Changelog entry

*

miina:
## Summary

- Calculates responsive width and height for the CTA button to accurately position in in the FE.
- Sets max limits for CTA button (for not having a huge button in landscape)
- Adjusts CTA button's font size.
- Adds border around URL Input for visibility.

Note: Enhancing the CTA button by converting it to a button block and making it resizable will be done in a separate follow-up PR since it's out of scope for fixing the bug.

Addresses issue #3343

## Checklist


assets/src/stories-editor/blocks/amp-story-cta/save.js
```diff
@@ -27,16 +29,20 @@ const CallToActionSave = ( { attributes } ) => {
...
 		<amp-story-cta-layer id={ anchor ? anchor : getUniqueId() }>
-			<RichText.Content
-				tagName="a"
-				className={ className }
-				href={ url }
-				style={ styles }
-				value={ text }
-			/>
+			<div className="amp-cta-button-wrapper">
+				<RichText.Content
```
miina: I'm wondering if it would make sense to use `amp-fit-text` here instead, for responsiveness. Thoughts?

swissspidy: Makes sense to me.

Just using that by default could work. Can't think of a scenario where it makes sense being able to disable `amp-fit-text`.

assets/src/stories-editor/blocks/amp-story-cta/edit.js
```diff
@@ -94,7 +95,17 @@ const CallToActionEdit = ( {
 					<RichText
 						placeholder={ placeholder }
 						value={ text }
-						onChange={ ( value ) => setAttributes( { text: value } ) }
+						onChange={ ( value ) => {
+							setAttributes( { text: value } );
+							// Also update width and height based on the room that the CTA button takes.
+							const element = document.querySelector( `#amp-story-cta-button-${ clientId } .amp-block-story-cta__link` );
```
swissspidy: This seems to be the element from the `edit` component, which can have different dimensions than the saved block. Example:

When editing:

Screenshot 2019-10-07 at 14 50 31

Not editing:

Screenshot 2019-10-07 at 14 50 34

Frontend:

Screenshot 2019-10-07 at 14 50 39

miina: Thank you for testing. Looks like there was a CSS rule overwritten that made the edit and view look different for the CTA block, now it should look the same.

However, will check into amp-fit-text anyway so that might change things a bit again.

spacedmonkey: I am still seeing issues around editing and previewing. 

Screenshot from 2019-10-09 11-49-26
Screenshot from 2019-10-09 11-50-10

spacedmonkey:
Screenshot from 2019-10-09 11-52-58
Screenshot from 2019-10-09 11-50-10

There are lots of differences between admin preview and FE. Line height and padding seem very different.

assets/src/stories-editor/blocks/amp-story-cta/edit.js
miina: Thanks for testing! Just in case: did you build the CSS before testing? There were some FE changes that don't get built automatically.

Line height was different indeed, adjusted that.

spacedmonkey:
> * Sets max limits for CTA button (for not having a huge button in landscape)
> * Adjusts CTA button's font size.
> * Adds border around URL Input for visibility.

> Note: Enhancing the CTA button by converting it to a button block and making it resizable will be done in a separate follow-up PR since it's out of scope for fixing the bug.

I am not sure I understand the reasoning behind converting to a button in another PR. Surely that means we are doubling the work, as fixes are applied to this that will have to be reviewed, tested and QAed. Which would then be undone by converting to button. Specially since this already deprecated the current style button. Fewer deprecations the better.

assets/src/stories-editor/blocks/amp-story-cta/edit.js
spacedmonkey: CSS build fixed the fix for me, thanks for the tip.

Shouldn't the button have a max width of the container that force to the next line. I can't see a good reason why you would want a button that runs on forever like that. To make text go onto a new line, , I had to press enter, which left really weird.

miina: Maybe, that's out of scope for this bug though. We generally allow (at this point) the blocks to be "over the edge" so these two things might need some UX thought together. E.g. if the CTA button is in the middle and you start typing and we allow 100% width then it can still go over the edge.

Quite certain it's a lower priority compared to some other issues.

miina:
> I am not sure I understand the reasoning behind converting to a button in another PR. Surely that means we are doubling the work, as fixes are applied to this that will have to be reviewed, tested and QAed. Which would then be undone by converting to button. Specially since this already deprecated the current style button. Fewer deprecations the better.

Being mindful of the deadlines then that's out of scope for this specific bug right now. I definitely do understand the point of converting to button right away, however, this would arguably be a new feature (especially if it would include resizing etc., which would make sense to me), not a bug fix addressing this bug. The deprecation needed for the current PR is quite simple, fortunately, no migration needed. Converting to a button, however, would likely require much more compared to this bug fix.

In conclusion -- being mindful of the few time we have left, converting to a button would be better to address separately. I'll create a new issue for that and add it to the definition board.

miina:
Follow-up issue: https://github.com/ampproject/amp-wp/issues/3476

assets/src/stories-editor/blocks/amp-story-cta/deprecated.js
```diff
@@ -73,6 +91,50 @@ CallToActionSaveV121.propTypes = {
...
+	} = attributes;
+
+	const className = getClassNameFromBlockAttributes( { ...attributes, className: 'amp-block-story-cta__link' } );
+	const styles = getStylesFromBlockAttributes( attributes );
+
+	styles.top = btnPositionTop ? `${ btnPositionTop }%` : undefined;
+	styles.left = btnPositionLeft ? `${ btnPositionLeft }%` : undefined;
+
+	return (
+		<amp-story-cta-layer id={ anchor ? anchor : getUniqueId() }>
```
spacedmonkey: I don't love conditional logic here. Can we just make a const the line above?

miina: Note that this is a deprecation which is an exact copy of the `save` function from the plugin version `1.3.0`.

Let's keep it as it is and not make changes "to the past".

spacedmonkey: Fair enough

Issue #3444, opened 2019-10-04
Blocks "jumping" when resizing to minimum limits fast


miina:
## Bug Description

When resizing a block (any block _except for_ Text block without amp-fit-text) to small quickly, it jumps.

## Expected Behaviour

Block should stay in place even when resizing quickly.

## Steps to reproduce

1. Add a block (e.g. Text / Image)
2. Resize the block to be quite tall.
3. Rotate the block a bit.
4. Use the top resizing handle and reduce the block's height quickly.
5. See the results.

## Screenshots

![Screen Recording 2019-10-03 at 17 07 27](https://user-images.githubusercontent.com/637548/66164382-907cd780-e600-11e9-88b2-87b9cca1b31f.gif)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* AC1: When a block is placed on a page, regardless of size or rotational orientation, it should have a centered focal point, so when the block is rotated, or resized, it does not "jump" visually on the page.

Related to #3461 

## Implementation brief

Fixed in #3529. 

## QA testing instructions

1. Add a block (e.g. Text / Image)
1. Resize the block to be quite tall.
1. Rotate the block a bit.
1. Use the top resizing handle and reduce the block's height quickly.
1. Block should not move / jump. 

## Demo

* 

![junmping2](https://user-images.githubusercontent.com/3294597/67093306-9dc5c480-f1b1-11e9-8b8a-e334e8ad8c07.gif)
![jumping](https://user-images.githubusercontent.com/3294597/67093308-9dc5c480-f1b1-11e9-999d-eeefcc2b3426.gif)

## Changelog entry

*

spacedmonkey:
Looking into this issue, seems a little a simple fix to apply CSS fix. I found a property called [object-position](https://css-tricks.com/almanac/properties/o/object-position/).
```
.wp-block-image .components-resizable-box__container img {
       object-position: 50% 50%;
}
```

miina:
Note that this issue occurs with other blocks as well, at least with the Text block and Meta blocks.

spacedmonkey:
The calculation for top  is incorrect because of the angel and min height being met. Not sure why this is yet, I am going to continue to look into this.

miina:
@spacedmonkey FYI: here is the original comment based on which the issue was created if that's helpful somehow: https://github.com/ampproject/amp-wp/pull/3433#issuecomment-538130369

[...]

Issue #3398, opened 2019-10-01
Text size not adjusting correctly when resizing a rotated block


miina:
Steps to reproduce:
- Add a Text block
- Write some text in it.
- Rotate the block.
- Resize it larger step by step, e.g. from the right top corner.

![resizing-3](https://user-images.githubusercontent.com/3294597/65974507-57602e00-e46d-11e9-8ef2-c4926eb3fc0d.gif)

Expected:
- The font size should always adjust according to the room it has.

miina:
Note: Seems to happen for a rotated block only when testing.

jauyong:
Hey @miina after I resize my rotated text block, it looks like the text is adjusting appropriately. I never got the small text in your gif after my resizes. 

![3398](https://user-images.githubusercontent.com/5874851/67884892-f8014680-fb1c-11e9-8169-ec63daf37980.gif)

miina:
Hey @jauyong, it still happens but it's not something that happens all the time, I had to resize quite a few times for this to happen. Tested on staging:
![font-size](https://user-images.githubusercontent.com/3294597/67904250-c8ae0200-fb3b-11e9-961d-05fb661dde35.gif)

PR #3454, opened 2019-10-07
Fix sliding when reducing size.

### LINKED ISSUE #3447: Text block with amp-fit-text "sliding" incorrectly when close to minimum limit
## Bug Description

Text block with amp-fit-text "sliding" incorrectly when close to minimum limit when reducing the size. The positioning changes incorrectly.

## Expected Behaviour

Reducing a block's size should not make it "slide".

## Steps to reproduce

1. Create a Text block.
2. Add some text.
3. Rotate the block and make it quite wide.
4. Reduce the size to minimum width from the right handle.
5. See the block either sliding a bit / making a jump.

## Screenshots

jumpiness
https://cloudup.com/cRZMrlUPdy8

See the original conversation in #3397

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* Create a Text block.
* Add some text.
* Rotate the block and make it quite wide.
* Reduce the size to minimum width from the right handle.
* Verify that the block is not "sliding" when reaching minimum limit.

## Demo

* 

## Changelog entry

*

miina:
## Summary

From investigation:

The issue here is coming from the font size going over the limits of the block while reducing the size and this making the block slide. It does _not_ have anything to do with positioning.

Addresses issue #3447 

- [x] Fix sliding for Text block and Meta blocks with amp-fit-text.

## Checklist


assets/src/stories-editor/components/resizable-box/index.js
```diff
@@ -190,14 +192,26 @@ class EnhancedResizableBox extends Component {
...
+						let scrollHeight = textElement.scrollHeight;
+
 						if ( appliedWidth < scrollWidth || appliedHeight < scrollHeight ) {
-							reachedMinLimit = true;
-							appliedWidth = lastWidth;
-							appliedHeight = lastHeight;
+							// If we have amp-fit-text on, then reduce the font size to prevent the block
+							// shifting due to font size not fitting.
+							if ( ampFitTextElement ) {
+								while ( appliedWidth < scrollWidth || appliedHeight < scrollHeight ) {
```
miina: Looks like the sliding issue came from the initial pre-resize amp-fit-text font size being too large for the container and this causing the container to "slide". After resizing stopping and the font calculation happened, the block "jumped" to its correct position. Note that this did not influence the top/left positioning in any way.

WDYT of this approach? Reducing the font size already while resizing.

resizing

barklund: If it's performant, it's a great idea and better UX than the original. Does it work well for even long texts and weird fonts (with many ligatures)?

swissspidy: It only works when making the block smaller, not when making it larger, which is confusing. Also doesn't seem to respect `MAX_FONT_SIZE` and `MIN_FONT_SIZE`. I just made a block with lots of text very small and the text basically disappeared because the font size was set to `0px`.

miina: It definitely should have font limits. Also, if we go that way we should also implement the same logic for the post meta blocks as well to make it consistent.

On performance: seems to work well with weird fonts as well, at least as far as I tested.

On working when making the block smaller only: yep, I was doubting that it might be confusing. When turning off amp-fit-text and the previously saved font size is too large then we automatically increase the size of the block so that the text would fit inside. However, we do not reduce the block if there's a lot of room. Meaning that elsewhere we do care about the font not going over the lines, not so much about there being extra room left. Because of that, I was 50-50 that perhaps it's OK just to make sure that the text fits the block but ignore if there's more room. It's a bit different case for amp-fit-text though.

Anyway -- I'll add the size increasing as well now that you brought it out as well.

barklund:
I see quite a few problems with this though. I have the default fitted text box on a new story, paste a lot of text in there and start resizing. I can't make the box shorter, when I make it narrower the content jumps weirdly around, when I make it wider the content font size is suddenly tiny. All this is without rotating.

In the end I ended up with this weird monstrosity - where the content doesn't take up the full width while at the same time taking up too much length.
Screenshot 2019-10-08 at 16 27 22

miina:
Thanks for testing. Yes, it's not final yet.

Was working on this but wasn't able to figure out yet why some of the things were happening as they were happening so will continue tomorrow morning.

Just pushed a fix for the specific issue that you mentioned, however, there are still a bunch of oddities happening.

miina:
Related to some of the issues I'm seeing: https://github.com/ampproject/amp-wp/issues/3398

Not sure exactly why that happens.

miina:
An alternative for fixing the "sliding" specifically would be to assign `absolute` value to the element which contains the text for the time of resizing. Perhaps it might make more sense to just do that and leave adjusting the font size for later 🤷‍♀

Using `absolute` on the text element (or `hidden` overflow on the parent element) causes other kind of jumpiness though:

jumpiness


Anyway, will give it another try in the morning.

miina:
After looking into this more then it looks like it does not make sense right now to add the font size reducing -- mostly the bug #3398 which means that while resizing the text is quite often switching between tiny and huge font so it looks like some odd flickering.

That would need fixing first first.

Also tried reducing/increasing the font by 1 instead of trying to find the "ideal" font size but this causes some additional issues when resizing very fast.

barklund:
@googlebot I consent.

barklund:
@miina, this is once again due to the slightly weird decision of displaying the inner box as `inline-block`. Inline block elements observe `line-height` and thus has some extra "padding" due to line height being added under some (but not all) circumstances.

When the overflow changes to `hidden` (or position to `absolute`), the parent element suddenly gets the extra height from the line height and this causes a slight shift for rotated elements.

I have now set the `line-height` to `0` for the outer block element (and re-set it to `1.15` for the inner one) and it seems to work for all the elements, but please text extensively - and remember to test them all with both fit and non-fit plus single- and multi-line texts.

We really should place that inner block correctly with `display:flex` on the contaner rather than `display:inline-block` on the element itself. But that's a future concern.

miina:
Thanks for this!

Tested with Text block and post meta blocks (title, author, date), with both amp-fit-text on and without; with multiple lines and single lines; also with bold and italic in the middle in case of the Text block.

Didn't detect any regressions.

miina:
@spacedmonkey As you haven't touched the code of this PR, would you mind reviewing?

spacedmonkey:
I can confirm that I am still seeing the slide as well :(

miina:
Uh, that's unexpected.

Will debug some more then.

miina:
Weird but I'm not able to replicate it, tried for some time with different rotations etc.
Tested on Chrome and also Safari and Firefox. 🤷‍♀ 

not-sliding

What I did:
- Add Text block
- Leave amp-fit-text on.
- Add some background color for visibility.
- Rotate the block
- Write "Hello there and there as content
- Resize wider
- Try resizing smaller.

Is there anything else you're doing when testing? Not sure what am I doing differently.

swissspidy:
@miina Hmm not seeing the exact same issue as in the video anymore, but it still happens when making the block very small or very big.

See new video: https://cloudup.com/cn74fKgoQZL

Comparing with your demo, it doesn't go as far.

miina:
@swissspidy Thanks for checking again.

On the block sliding when going very large: that seems to be a different issue, looks like it's related to the resizing logic + positioning and not the font size. We might want to report a separate issue for that for prioritization. Potentially it's less important since it's unlikely to resize the block that much over the limits of the Page. Thoughts? So many issues with resizing have appeared lately.

On the block sliding when going very small: 
Could it be a caching issue and CSS not updating? I'm seeing that in your video the text goes over the block, however, since it has `overflow:hidden` assigned (which is the fix in this PR) then it should not be visible outside of the block.

spacedmonkey:
Two possiblity related issues https://github.com/ampproject/amp-wp/issues/3461 & https://github.com/ampproject/amp-wp/issues/3444

miina:
> Two possiblity related issues #3461 & #3444

These two are related in terms of being related to resizing but have a different cause. #3444 still needs investigation and Implementation Brief but it's likely related to positioning the block while resizing. #3461 has Implementation Brief where the source of the issue is mentioned (see Cause of the bug and possible fix).

In this issue's case, it's the font size that causes it.

Note that #3444 might be related to sliding when making a block very big since both are related to positioning the block when resizing. We could combine those if you prefer!

swissspidy:
Thanks for the clarification. I'm fine with merging this for now as it's already an important improvement.

It would be good to take a more holistic approach for the remaining resizing issues, otherwise we keep chasing edge cases.

miina:
Most of the resizing bugs so far haven't been the quickest to resolve, and somehow each bug has brought out new bugs. Initially, I combined the newly found related bugs as well but then it started taking too much time since the issues (causes) were actually different + each one of those took some debugging time. And then one bug was fixed but waiting for another bug getting fixed, too. Also, some cases are edge cases and some more common cases, so prioritization would be good, too.

Not sure what's the best approach here, not sure if putting all the issues together would make sense in terms of having different causes + taking too much time to get anything fixed. 🤷‍♀At least that's how I felt when actually working on those.

spacedmonkey:
> Not sure what's the best approach here, not sure if putting all the issues together would make sense in terms of having different causes

I am not sure I agree here. A fix for one issue, may effect other work another developer is working on. 
I would also question, why do we have some many bugs with resizing? I see we are using [re-resizable](https://github.com/bokuweb/re-resizable), I wonder if they are upstream bugs in this library and maybe we should replace this with another library...

swissspidy:
> I would also question, why do we have some many bugs with resizing?  I see we are using re-resizable, I wonder if they are upstream bugs in this library and maybe we should replace this with another library...

The library is just fine. The problem lies in the complexity of resizing elements that are rotated, as we have to do re-calculate the block's top and left position constantly while resizing.

miina:
> I am not sure I agree here. A fix for one issue, may effect other work another developer is working on.

In my case, I had been debugging one issue for a few days. Then another issue came out which had a completely different cause. That took quite a long time to figure out + there was some feedback rounds and it delayed the first (related but not the same) bug for more than a week. If we gather all the different bugs together then it might take quite a long time for those getting fixed although another bug got fixed already. Also -- some of these bugs are edge cases (such as #3444) and being mindful of the timeline then we should prioritize.

> I would also question, why do we have some many bugs with resizing? I see we are using re-resizable, I wonder if they are upstream bugs in this library and maybe we should replace this with another library...

Nope, not upstream bugs, so far there have been a lot of regressions since resizing is quite influenced by a lot of stuff and almost every block has resizing + we've added dragging and rotating to it. So it's fully our own bugs :)

miina:
Just realized that there's a bug related to this PR: the fix should be for amp-fit-text only! Reopening

Issue #3450, opened 2019-10-05
Add lightbox effect doesn't work for image blocks


rsmith4321:
## Bug Description

If an image is inserted using the block editor image block, and under AMP Settings the Add lightbox effect is selected as on.  The image is not loaded in a lightbox when clicked. 

## Expected Behaviour

The image should be loaded in an amp lightbox when clicked. 

## Steps to reproduce

The first 3 images on this page have Add lightbox effect on with no effect. https://www.ryansmithphotography.com/myrtle-beach-wedding-photography-prices/

## Screenshots

<img width="2108" alt="Screen Shot 2019-10-04 at 11 14 46 PM" src="https://user-images.githubusercontent.com/7704223/66249116-f2bb0280-e6fc-11e9-8f9e-8347f4a39e50.png">

## Additional context

 - WordPress version: 5.2.3
 - Plugin version: 1.3.0
 - Gutenberg plugin version (if applicable): 
 - AMP plugin template mode: standard
 - PHP version: 7.3
 - OS:
 - Browser: chrome
 - Device: MacOS Desktop

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

westonruter:
@rsmith4321 Do you have any filters on images like you had for galleries in that other issue?

rsmith4321:
As far as I know there are no filters or special settings on my site for images. Does this work on other sites?  I just assumed it was something that hasn't been looked at yet similar to the gallery lightbox issue.

westonruter:
@kienstra Would you please investigate?

kienstra:
Sure, @westonruter!

kienstra:
**Thanks, I also saw this**

Hi @rsmith4321,
Thanks for bringing this up. It looks like this is a valid bug.

As a quick workaround, does removing the link in the Image blocks help?

<img width="1435" alt="image-block-link" src="https://user-images.githubusercontent.com/4063887/66275556-4473a200-e84f-11e9-8139-17e23e826f49.png">

I could reproduce this issue when the Image block has a link, but the lightbox looks good when it doesn't.

I'll look at a code fix in the plugin, but maybe that workaround will help in the meantime.

rsmith4321:
[...]

PR #3460, opened 2019-10-08
Add lightbox support to Core Image blocks with links

kienstra:
## Summary

As reported in #3450, there's an issue where images that are wrapped in `<a>` didn't support the lightbox feature. Clicking on them simply caused the browser to go to that link.

So this recognizes when an image is wrapped in `<a>`, and runs the lighbox logic for it.

Steps to reproduce:
1. Create a post
2. Add an Image block
3. Add any image to it
4. Click the 'link' icon, and click 'Media File' or 'Attachment Page':
image-add-link

5. Click 'Add lightbox effect'
6. Click 'Update,' and go to the front-end
7. Expected: Clicking that image causes the lightbox to display
8. Actual: it goes to the URL of the image:
image-attachment-url

Addresses issue #3450

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -372,8 +372,18 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 			return $attributes;
 		}
 
-		$is_node_wrapped_in_anchor = 'a' === $parent_node->tagName && $parent_node->parentNode instanceof DOMElement && 'figure' === $parent_node->parentNode->tagName;
-		if ( 'figure' !== $parent_node->tagName && ! $is_node_wrapped_in_anchor ) {
+		// This should be true for links to the actual media file, but not for links to the attachment page.
+		$is_node_wrapped_in_media_file_link = (
+			'a' === $parent_node->tagName
+			&&
+			$parent_node->parentNode instanceof DOMElement
+			&&
+			'figure' === $parent_node->parentNode->tagName
+			&&
+			attachment_url_to_postid( $parent_node->getAttribute( 'href' ) )
+		);
+
+		if ( 'figure' !== $parent_node->tagName && ! $is_node_wrapped_in_media_file_link ) {
 			return $attributes;
 		}
 
@@ -396,10 +406,11 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 			$attributes['lightbox']          = '';
 
 			/*
-			 * Remove the <a> if the image is wrapped in one, as it can prevent the lightbox from working.
+			 * Removes the <a> if the image is wrapped in one, as it can prevent the lightbox from working.
+			 * But this only removes the <a> if it links to the media file, not the attachment page.
 			 * If this feature is added, this can probably be removed: https://github.com/ampproject/amphtml/issues/25021
 			 */
-			if ( $is_node_wrapped_in_anchor ) {
+			if ( $is_node_wrapped_in_media_file_link ) {
 				$node->parentNode->parentNode->replaceChild( $node, $node->parentNode );
 			}
 		}
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -368,12 +368,17 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 	 */
 	private function maybe_add_lightbox_attributes( $attributes, $node ) {
 		$parent_node = $node->parentNode;
-		if ( ! ( $parent_node instanceof DOMElement ) || 'figure' !== $parent_node->tagName ) {
+		if ( ! ( $parent_node instanceof DOMElement ) ) {
 			return $attributes;
 		}
 
-		// Account for blocks that include alignment.
-		// In that case, the structure changes from figure.wp-block-image > img
+		$is_node_wrapped_in_anchor = 'a' === $parent_node->tagName && $parent_node->parentNode instanceof DOMElement && 'figure' === $parent_node->parentNode->tagName;
+		if ( 'figure' !== $parent_node->tagName && ! $is_node_wrapped_in_anchor ) {
+			return $attributes;
+		}
+
+		// Account for blocks that include alignment or images that are wrapped in <a>.
+		// With alignment, the structure changes from figure.wp-block-image > img
 		// to div.wp-block-image > figure > img and the amp-lightbox attribute
 		// can be found on the wrapping div instead of the figure element.
 		$grand_parent = $parent_node->parentNode;
@@ -393,13 +398,18 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 			$attributes['tabindex']          = 0;
 
 			$this->maybe_add_amp_image_lightbox_node();
+
+			if ( $is_node_wrapped_in_anchor ) {
+				// Remove the <a> that the image is wrapped in, as it can prevent the lightbox from working.
+				$node->parentNode->parentNode->replaceChild( $node, $node->parentNode );
+			}
 		}
 
 		return $attributes;
 	}
 
 	/**
-	 * Determines is a URL is considered a GIF URL
+	 * Determines if a URL is considered a GIF URL
 	 *
 	 * @since 0.2
 	 *
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -405,7 +405,6 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 			/*
 			 * Removes the <a> if the image is wrapped in one, as it can prevent the lightbox from working.
 			 * But this only removes the <a> if it links to the media file, not the attachment page.
-			 * If this feature is added, this can probably be removed: https://github.com/ampproject/amphtml/issues/25021
 			 */
 			if ( $is_node_wrapped_in_media_file_link ) {
 				$node->parentNode->parentNode->replaceChild( $node, $node->parentNode );
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -393,11 +393,7 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 
 		if ( isset( $parent_attributes['data-amp-lightbox'] ) && true === filter_var( $parent_attributes['data-amp-lightbox'], FILTER_VALIDATE_BOOLEAN ) ) {
 			$attributes['data-amp-lightbox'] = '';
-			$attributes['on']                = 'tap:' . self::AMP_IMAGE_LIGHTBOX_ID;
-			$attributes['role']              = 'button';
-			$attributes['tabindex']          = 0;
-
-			$this->maybe_add_amp_image_lightbox_node();
+			$attributes['lightbox']          = '';
 
 			if ( $is_node_wrapped_in_anchor ) {
 				// Remove the <a> that the image is wrapped in, as it can prevent the lightbox from working.
--- a/includes/sanitizers/class-amp-img-sanitizer.php
+++ b/includes/sanitizers/class-amp-img-sanitizer.php
@@ -368,7 +368,7 @@ class AMP_Img_Sanitizer extends AMP_Base_Sanitizer {
 	 */
 	private function maybe_add_lightbox_attributes( $attributes, $node ) {
 		$parent_node = $node->parentNode;
-		if ( ! ( $parent_node instanceof DOMElement ) ) {
+		if ( ! ( $parent_node instanceof DOMElement ) || ! ( $parent_node->parentNode instanceof DOMElement ) ) {
 			return $attributes;
 		}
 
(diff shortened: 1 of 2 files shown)


kienstra:
**Request To Test**

Hi @rsmith4321,
Hope you've had a great week.

When you have a chance, could you please test this PR?

It should fix your Issue #3450, and shouldn't require the [workaround](https://github.com/ampproject/amp-wp/issues/3450#issuecomment-538785885) of removing the Image block's link.

Thanks, Ryan!

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -393,13 +398,18 @@ private function maybe_add_lightbox_attributes( $attributes, $node ) {
 			$attributes['tabindex']          = 0;
 
 			$this->maybe_add_amp_image_lightbox_node();
+
+			if ( $is_node_wrapped_in_anchor ) {
+				// Remove the <a> that the image is wrapped in, as it can prevent the lightbox from working.
```
westonruter: Did you check if this issue has been reported in AMP? Perhaps it is a bug in AMP itself that should be fixed?

kienstra: Ah, that's a good point. It is a little strange that the plugin has to remove the `<a>`.

kienstra: I couldn't find an issue in github.com/ampproject/amphtml for this, but I'll keep looking.

kienstra: I also might open an issue in `amphtml` for this.

kienstra: I opened an issue in `amphtml`: https://github.com/ampproject/amphtml/issues/25021

schlessera:
Code looks good to me. However, we should wait for feedback on ampproject/amphtml#25021 before proceeding.

kienstra:
>Code looks good to me. However, we should wait for feedback on ampproject/amphtml#25021 before proceeding.

Nice, thanks!

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -388,18 +393,19 @@ private function maybe_add_lightbox_attributes( $attributes, $node ) {
...
-			$attributes['on']                = 'tap:' . self::AMP_IMAGE_LIGHTBOX_ID;
-			$attributes['role']              = 'button';
-			$attributes['tabindex']          = 0;
+			$attributes['lightbox']          = '';
 
-			$this->maybe_add_amp_image_lightbox_node();
+			if ( $is_node_wrapped_in_anchor ) {
+				// Remove the <a> that the image is wrapped in, as it can prevent the lightbox from working.
+				$node->parentNode->parentNode->replaceChild( $node, $node->parentNode );
+			}
```
kienstra: If `amphtml` later supports wrapping `<amp-img lightbox>` in `<a>`, this can be removed.

kienstra:
This PR is now changing the Core Image block from using the `<amp-image-lightbox>` to the recommended `<amp-lightbox-gallery>`. It [looks like](https://github.com/ampproject/amphtml/issues/25021#issuecomment-541846264) `<amp-image-lightbox>` will eventually be deprecated.

It looks like this works well so far. The only change I noticed is that all `<amp-img lightbox>` elements will be visible in the lightbox:

amp-gallery-lightbox

westonruter:
@rsmith4321 Would you test this please? Here's a build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3731846/amp.zip)

rsmith4321:
I tested it and it works well. I think it's kind of nice to be able to
cycle to other images on the page in the lightbox even though they aren't
contained in a gallery. It still has the same issue of the lightbox not
opening if it's linked to the media file surrounded by an <a> tag.

westonruter:
> It still has the same issue of the lightbox not opening if it's linked to the media file surrounded by an <a> tag.

I'm confused. @kienstra I thought this is what the `$is_node_wrapped_in_anchor` was for?

kienstra:
@westonruter, right, this PR should allow the image to be wrapped in a `<a>`. Let me look at this more.

westonruter:
If the link is for the media file itself, then I think it is correct to remove the hyperlink in favor of the lightbox alone, since clicking the link would just cause the image to be loaded on a separate page; here a lightbox is a better experience since you don't leave the page context.

If, however, the link is for the attachment _page_, then in this case I think it is better to retain the link.

kienstra:
>If the link is for the media file itself, then I think it is correct to remove the hyperlink in favor of the lightbox alone, since clicking the link would just cause the image to be loaded on a separate page; here a lightbox is a better experience since you don't leave the page context.

>If, however, the link is for the attachment page, then in this case I think it is better to retain the link.

Sounds good, I'll work on applying that.

kienstra:
Hi @rsmith4321,
>It still has the same issue of the lightbox not opening if it's linked to the media file surrounded by an <a> tag.

Thanks for testing this!

Is that on a live site, by chance, and could you share a link to it?

That might help to reproduce it.

Locally, when linking to the Media File and selecting the lightbox, the lightbox works:

ligthbox-media

Thanks again, Ryan! You've been really helpful in testing everything.

rsmith4321:
Yes, on this page
https://www.ryansmithphotography.com/myrtle-beach-photo-booths/ the three
images at the top have the lightbox selected on but I removed the link to
the media files. The images further down the page I switched the lightbox
on but they still have the link to the media file and they don't open in a
lightbox.

kienstra:
Thanks, I also see the issue with the lightbox not working.

In the [link provided](https://www.ryansmithphotography.com/myrtle-beach-photo-booths/), it looks like the lightbox doesn't work when it's for an Image block that has a link and alignment:

has-alignment-and-a

Image blocks with alignment are wrapped in an extra `<div>`. The logic accounts for [that extra div](https://github.com/ampproject/amp-wp/pull/3460/files#diff-20fe23f4cf9210c1bde24280f116244aR390), or an extra `<a>` from the link, but not both.

I'll work on this.

kienstra:
**Request To Test**

Hi @rsmith4321,
Thanks for your patience. Could you please test this [amp.zip](https://github.com/ampproject/amp-wp/files/3741812/amp.zip), which should fix the issue with lightboxes on [the URL you provided](https://www.ryansmithphotography.com/myrtle-beach-photo-booths/)?

This shouldn't be an issue on that URL you shared, but the lightbox will only work if the link is to the 'Media File,' not to the 'Attachment Page':
link-to-attachment-page

...though on that [URL you provided](https://www.ryansmithphotography.com/myrtle-beach-photo-booths/), it looks like the URLs are to the 'Media File,' so they should be fine.

Thanks, Ryan! You've been really helpful in reporting and testing these issues.

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -368,38 +368,71 @@ private function adjust_and_replace_node( $node ) {
...
 
-		// Account for blocks that include alignment.
-		// In that case, the structure changes from figure.wp-block-image > img
+		// This should be true for links to the actual media file, but not for links to the attachment page.
+		$is_node_wrapped_in_media_file_link = (
+			'a' === $parent_node->tagName
+			&&
+			( 'figure' === $parent_node->tagName || 'figure' === $parent_node->parentNode->tagName )
+			&&
+			attachment_url_to_postid( $parent_node->getAttribute( 'href' ) )
```
kienstra: This ensures the link is to the 'Media File,' not the 'Attachment Page':

link-to-attachment-page

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -368,38 +368,71 @@ private function adjust_and_replace_node( $node ) {
...
+		// Account for blocks that include alignment or images that are wrapped in <a>.
+		// With alignment, the structure changes from figure.wp-block-image > img
 		// to div.wp-block-image > figure > img and the amp-lightbox attribute
 		// can be found on the wrapping div instead of the figure element.
 		$grand_parent = $parent_node->parentNode;
-		if ( $grand_parent instanceof DOMElement ) {
-			$classes = preg_split( '/\s+/', $grand_parent->getAttribute( 'class' ) );
-			if ( in_array( 'wp-block-image', $classes, true ) ) {
-				$parent_node = $grand_parent;
-			}
```
kienstra: The logic above is mainly copied into `does_node_have_block_class()`.

rsmith4321:
No problem. Yes you wouldn't want a lightbox if the image is an link to
another page and not to the media file so that seems like the correct
behavior. I tried the latest build you provided and I'm still seeing the
same results on that page.

kienstra:
Thanks, Ryan! Right, it does look like the issue still exists on [that URL](https://www.ryansmithphotography.com/myrtle-beach-photo-booths/). I'll look at that.

kienstra:
Maybe [attachment_url_to_postid](https://github.com/ampproject/amp-wp/pull/3460/files#diff-20fe23f4cf9210c1bde24280f116244aR381) isn't returning the ID, as expected. Though I couldn't reproduce this locally.

includes/sanitizers/class-amp-img-sanitizer.php
westonruter: From what I  recall, `attachment_url_to_postid()` is not super reliable. I think it would be better to look at the URL to see if it looks like a file instead. For example:

```html
$is_file_url = preg_match( '/\.\w+$/', wp_parse_url( $parent_node->getAttribute( 'href' ), PHP_URL_PATH ) );
```

kienstra: Good idea!

kienstra: It's applied in https://github.com/ampproject/amp-wp/pull/3460/commits/783a8d8e239657f3f2eb265fe6a43b9313a28c84

kienstra:
Hi @rsmith4321,
Sorry to keep sending you plugin builds to test.

But here's another [amp.zip](https://github.com/ampproject/amp-wp/files/3745901/amp.zip) when you have a chance. 

Thanks a lot for your help.

rsmith4321:
Yes the latest build seems to fix the issue for me, you can check out that
same page now I still have some of the images linked to the media file and
it seems to strip that out for the lightbox.

rsmith4321:
The only thing I notice on a mouse hover the cursor is not changed to a
pointer because it's not a link anymore. I would recommend if you can add
that so on desktop there is an indication the images can be opened in the
lightbox.  I'm not sure how to target an image with the
data-amp-lightbox="true" because it's not a regular class so I can't just
add this to the css.

swissspidy:
Glad to hear it's working!

----

Personally I don't think the AMP plugin should change the cursor as it's the theme's responsibility.

You should be able to use the following CSS in your theme to change the cursor:

```css
[data-amp-lightbox="true"] { cursor: pointer; }
```

kienstra:
Hi @rsmith4321,
Thanks a lot for testing this so many times. It's good to hear it's working.

>The only thing I notice on a mouse hover the cursor is not changed to a
pointer because it's not a link anymore.

Sorry, it would have been ideal if it could have still been wrapped in a link. We ended up having to strip.

The [styling above](https://github.com/ampproject/amp-wp/pull/3460#issuecomment-544454365) should work, let me know if there's an issue.

includes/sanitizers/class-amp-img-sanitizer.php
```diff
@@ -368,38 +368,71 @@ private function adjust_and_replace_node( $node ) {
...
 	private function maybe_add_lightbox_attributes( $attributes, $node ) {
 		$parent_node = $node->parentNode;
-		if ( ! ( $parent_node instanceof DOMElement ) || 'figure' !== $parent_node->tagName ) {
+		if ( ! ( $parent_node instanceof DOMElement ) || ! ( $parent_node->parentNode instanceof DOMElement ) ) {
 			return $attributes;
 		}
 
-		// Account for blocks that include alignment.
-		// In that case, the structure changes from figure.wp-block-image > img
+		$is_file_url                        = preg_match( '/\.\w+$/', wp_parse_url( $parent_node->getAttribute( 'href' ), PHP_URL_PATH ) );
```
westonruter: A future improvement would be to prevent the lightbox toggle from even being made available when the image block has a link to the attachment page. Otherwise, the lightbox toggle has no effect when an attachment page is linked.

kienstra:
**Request For Final Review**

Hi @westonruter,
Happy Monday!

When you have a chance, could you please review this PR again?

In addition to the [testing steps above](https://github.com/ampproject/amp-wp/pull/3460#issue-325574237), this should also work when adding alignment to the image, like `alignright`, after testing step 4.

Thanks, Weston!

includes/sanitizers/class-amp-img-sanitizer.php
kienstra: Yeah, that's a good idea. It's a little deceptive now.

swissspidy: Created an issue for it: #3579.

kienstra: Nice!

rsmith4321:
> Glad to hear it's working!
> 
> Personally I don't think the AMP plugin should change the cursor as it's the theme's responsibility.
> 
> You should be able to use the following CSS in your theme to change the cursor:
> 
> ```css
> [data-amp-lightbox="true"] { cursor: pointer; }
> ```

Thanks I'll use that. I would agree except you are doing something unusual by stripping the link.  Normally a lightbox would open from a link to the image so the cursor would change. So I don't think any theme is going to ever know to add that css.

kienstra:
>Thanks I'll use that. I would agree except you are doing something unusual by stripping the link.

Yeah, that's a good point. We tried to get around stripping the link, but the `amphtml` project couldn't change the behavior, due to backwards-compatibility.

PR #3470, opened 2019-10-08
Always request the `force_svg_support` theme feature

### LINKED ISSUE #3009: Always replace `no-svg` class with `svg` if exists
The `AMP_Core_Theme_Sanitizer` includes a [method](https://github.com/ampproject/amp-wp/blob/develop/includes/sanitizers/class-amp-core-theme-sanitizer.php#L564-L581) to replace the `no-svg` class with a `svg` class on the document element. While this is used in the Twenty Seventeen theme, I think it would be useful to generically run this for every theme.

* Many themes take a core theme for inspiration when building certain features, and this is one of them (Twenty Seventeen is a good example for how to use SVG icons instead of the "old" way with an icon font).
* A theme adding the `no-svg` class would have its SVGs broken today without applying this same fix manually.

I think we can run this sanitizer snippet unconditionally because probably anyone that had a `no-svg` would want to have it replaced with `svg` if SVGs work (which we force in Twenty Seventeen already). Alternatively, if there's reason to suspect this as too "aggressive", we could also hide it behind an opt-in via theme support.

pierlon:
## Summary

This PR allows `force_svg_support` theme feature to always be requested.

Fixes #3009.

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-core-theme-sanitizer.php
+++ b/includes/sanitizers/class-amp-core-theme-sanitizer.php
@@ -82,7 +82,6 @@ class AMP_Core_Theme_Sanitizer extends AMP_Base_Sanitizer {
 					'twentyseventeen_javascript_detection', // AMP is essentially no-js, with any interactively added explicitly via amp-bind.
 				],
 			],
-			'force_svg_support'                   => [],
 			'force_fixed_background_support'      => [],
 			'add_twentyseventeen_masthead_styles' => [],
 			'add_twentyseventeen_image_styles'    => [],
@@ -570,14 +569,18 @@ class AMP_Core_Theme_Sanitizer extends AMP_Base_Sanitizer {
 	 * @link https://caniuse.com/#feat=svg
 	 */
 	public function force_svg_support() {
-		$this->dom->documentElement->setAttribute(
-			'class',
-			preg_replace(
-				'/(^|\s)no-svg(\s|$)/',
-				' svg ',
-				$this->dom->documentElement->getAttribute( 'class' )
-			)
-		);
+		$class = $this->dom->documentElement->getAttribute( 'class' );
+
+		if ( strpos( $class, 'no-svg' ) ) {
+			$this->dom->documentElement->setAttribute(
+				'class',
+				preg_replace(
+					'/(^|\s)no-svg(\s|$)/',
+					' svg ',
+					$this->dom->documentElement->getAttribute( 'class' )
+				)
+			);
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


pierlon:
@googlebot I signed it!

swissspidy:
Thanks a lot for your contribution @pierlon!

Just tagging a couple of people here to get the necessary eyes on it :-)

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -570,14 +569,18 @@ public function add_smooth_scrolling( $link_xpaths ) {
...
-			'class',
-			preg_replace(
-				'/(^|\s)no-svg(\s|$)/',
-				' svg ',
-				$this->dom->documentElement->getAttribute( 'class' )
-			)
-		);
+		$class = $this->dom->documentElement->getAttribute( 'class' );
+
+		if ( strpos( $class, 'no-svg' ) ) {
```
westonruter: This will fail with `class="no-svg"` since the position will be `0`.

But why the need for wrapping the condition? If the string isn't found then just no change will be made.

pierlon: > This will fail with class="no-svg" since the position will be 0.
Made the fix just as you commented :dash:.

> But why the need for wrapping the condition?
An empty class attribute (`class=""`) will be set if there was no class attribute.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -570,14 +569,18 @@ public function add_smooth_scrolling( $link_xpaths ) {
...
+		if ( strpos( $class, 'no-svg' ) !== false ) {
+			$this->dom->documentElement->setAttribute(
+				'class',
+				preg_replace(
+					'/(^|\s)no-svg(\s|$)/',
+					' svg ',
+					$this->dom->documentElement->getAttribute( 'class' )
+				)
+			);
+		}
```
westonruter: I suggest something like this to optimize/harden this a bit more:

```php
$class = $this->dom->documentElement->getAttribute( 'class' );
if ( $class ) {
	$count = 0;
	$class = preg_replace(
		'/(^|\s)no-svg(\s|$)/',
		' svg ',
		$class,
		$count
	);
	if ( $replaced_count > 0 ) {
		$this->dom->documentElement->setAttribute( 'class', $class );
	}
}
```

pierlon: Done, thanks for the suggestion. You missed the argument for the limit, so I've set it to the default (`-1`).

PR #3473, opened 2019-10-08
Add direct support for Scribd embeds

### LINKED ISSUE #3424: Scribd embeds are not fully supported
## Bug Description

Scribd is a core-supported oEmbed provider. However, when adding a Scribd embed with an AMP validation error is sanitized due to a `script` being included in the oEmbed response:

```html
<iframe class="scribd_iframe_embed" src="https://www.scribd.com/embeds/110799637/content" scrolling="no" id="110799637" width="100%" height="500" frameborder="0"></iframe>
<script type="text/javascript">
          (function() { var scribd = document.createElement("script"); scribd.type = "text/javascript"; scribd.async = true; scribd.src = "https://www.scribd.com/javascripts/embed_code/inject.js"; var s = document.getElementsByTagName("script")[0]; s.parentNode.insertBefore(scribd, s); })()
 </script>
```

In spite of this, the block renders fine on the frontend, _except_ the full screen button does not work. An error appears in the console:

> Blocked opening 'https://www.scribd.com/document/110799637/Synthesis-of-Knowledge-Effects-of-Fire-and-Thinning-Treatments-on-Understory-Vegetation-in-Dry-U-S-Forests#fullscreen&from_embed' in a new window because the request was made in a sandboxed frame whose 'allow-popups' permission is not set.

## Expected Behaviour

The block renders on the frontend without any validation errors. Full screen functionality should be available.

## Steps to reproduce

1. Add a Scribd embed, for example https://www.scribd.com/document/110799637/Synthesis-of-Knowledge-Effects-of-Fire-and-Thinning-Treatments-on-Understory-Vegetation-in-Dry-U-S-Forests
2. View the frontend.
3. Notice AMP validation errors appearing in the admin bar.
4. Clicking the Full screen button has no effect.
5. The height of the iframe is not the same as the non-AMP version.

## Screenshots

Screen Shot 2019-10-02 at 14 20 35

image

## Additional context

 - WordPress version: 5.3-beta
 - Plugin version: 1.3.0
 - Gutenberg plugin version (if applicable): 6.6.0
 - AMP plugin template mode: Transitional
 - PHP version: 7.3
 - OS: Mac
 - Browser: Chrome
 - Device: Mac

---------------

## Acceptance criteria

* Embedding Scribd results in no AMP validation errors.
* An embedded Scribd can be opened in full screen from an AMP page.
* [Optional] An embedded Scribd can resize itself.

## Implementation brief

* Following the model in the `AMP_SoundCloud_Embed_Handler`, add a `embed_oembed_html` filter which looks for embeds for Scribd.
* Use filter to strip out the `script`.
* Use filter to inject `sandbox="allow-scripts allow-popups"` attribute into the `<iframe>`
* Try to also add the `resizable` attribute, to [allow the iframe to resize](https://amp.dev/documentation/components/amp-iframe/?format=websites#iframe-resizing). This may not have any effect if Scribd does not send the required messages.

## QA testing instructions

1. Set template mode to Transitional.
1. Create a new Embed block or Scribd block provide the  URL: https://www.scribd.com/document/110799637/Synthesis-of-Knowledge-Effects-of-Fire-and-Thinning-Treatments-on-Understory-Vegetation-in-Dry-U-S-Forests
1. Click save draft and preview.
1. The AMP version should not show any validation errors in the Admin Bar and the embed should be fully functional.

## Demo

* 

## Changelog entry

* Add direct support for Scribd embeds.

pierlon:
## Summary

Removes the script element from oEmbed response and adds a sandbox attribute to allow the 'Full screen' functionality to work.

At the moment Scribd does not perform a [iframe resize request](https://amp.dev/documentation/components/amp-iframe/?format=websites#iframe-resizing), so the height of the iframe is not the same as the non-AMP version.

Fixes #3424.

## Checklist

### DIFF
--- a/includes/embeds/class-amp-scribd-embed-handler.php
+++ b/includes/embeds/class-amp-scribd-embed-handler.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Class AMP_SoundCloud_Embed_Handler
+ * Class AMP_Scribd_Embed_Handler
  *
  * @package AMP
  * @since 1.3.1
@@ -22,7 +22,7 @@ class AMP_Scribd_Embed_Handler extends AMP_Base_Embed_Handler {
 	 * Unregisters embed.
 	 */
 	public function unregister_embed() {
-		remove_filter( 'embed_oembed_html', [ $this, 'filter_embed_oembed_html' ], 10 );
+		remove_filter( 'embed_oembed_html', [ $this, 'filter_embed_oembed_html' ] );
 	}
 
 	/**
@@ -50,7 +50,7 @@ class AMP_Scribd_Embed_Handler extends AMP_Base_Embed_Handler {
 	 * @return string
 	 */
 	private function remove_script( $html ) {
-		$html_without_script = preg_replace( '#<script[^>].+?</script>#s', '', $html );
+		$html_without_script = preg_replace( '#<script.+?</script>#s', '', $html );
 
 		if ( null !== $html_without_script ) {
 			return $html_without_script;
@@ -68,8 +68,7 @@ class AMP_Scribd_Embed_Handler extends AMP_Base_Embed_Handler {
 	 */
 	private function inject_sandbox_attribute( $html ) {
 		if ( false === strpos( $html, 'sandbox="allow-scripts allow-popups"' ) ) {
-			preg_match( '#<iframe(?P<attributes>.+?)>#s', $html, $matches );
-			return "<iframe sandbox='allow-scripts allow-popups' ${matches['attributes']}></iframe>";
+			return str_replace( '<iframe ', '<iframe sandbox="allow-scripts allow-popups" ', $html );
 		}
 
 		return $html;
--- a/includes/embeds/class-amp-scribd-embed-handler.php
+++ b/includes/embeds/class-amp-scribd-embed-handler.php
@@ -1,0 +1,78 @@
+<?php
+/**
+ * Class AMP_SoundCloud_Embed_Handler
+ *
+ * @package AMP
+ * @since 1.3.1
+ */
+
+/**
+ * Class AMP_Scribd_Embed_Handler
+ */
+class AMP_Scribd_Embed_Handler extends AMP_Base_Embed_Handler {
+
+	/**
+	 * Registers embed.
+	 */
+	public function register_embed() {
+		add_filter( 'embed_oembed_html', [ $this, 'filter_embed_oembed_html' ], 10, 2 );
+	}
+
+	/**
+	 * Unregisters embed.
+	 */
+	public function unregister_embed() {
+		remove_filter( 'embed_oembed_html', [ $this, 'filter_embed_oembed_html' ], 10 );
+	}
+
+	/**
+	 * Filter oEmbed HTML for Scribd to to be AMP compatible.
+	 *
+	 * @param string $cache Cache for oEmbed.
+	 * @param string $url   Embed URL.
+	 * @return string Embed.
+	 */
+	public function filter_embed_oembed_html( $cache, $url ) {
+		if ( false === strpos( wp_parse_url( $url, PHP_URL_HOST ), 'scribd.com' ) ) {
+			return $cache;
+		}
+
+		$embed = $this->remove_script( $cache );
+		$embed = $this->inject_sandbox_attribute( $embed );
+
+		return $embed;
+	}
+
+	/**
+	 * Remove the 'script' element from the provided HTML if there is any.
+	 *
+	 * @param string $html HTML string.
+	 * @return string
+	 */
+	private function remove_script( $html ) {
+		$html_without_script = preg_replace( '#<script[^>].+?</script>#s', '', $html );
+
+		if ( null !== $html_without_script ) {
+			return $html_without_script;
+		}
+
+		return $html;
+	}
+
+	/**
+	 * Add the sandbox attribute with 'allow-popups' & 'allow-scripts' permissions set so that the
+	 * full screen button works.
+	 *
+	 * @param string $html HTML string.
+	 * @return string
+	 */
+	private function inject_sandbox_attribute( $html ) {
+		if ( false === strpos( $html, 'sandbox="allow-scripts allow-popups"' ) ) {
+			preg_match( '#<iframe(?P<attributes>.+?)>#s', $html, $matches );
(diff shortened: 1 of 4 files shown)


includes/embeds/class-amp-scribd-embed-handler.php
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Filter oEmbed HTML for Scribd to to be AMP compatible.
+	 *
+	 * @param string $cache Cache for oEmbed.
+	 * @param string $url   Embed URL.
+	 * @return string Embed.
+	 */
+	public function filter_embed_oembed_html( $cache, $url ) {
+		if ( false === strpos( wp_parse_url( $url, PHP_URL_HOST ), 'scribd.com' ) ) {
```
schlessera: This is not 100% foolproof, as it only checks for a partial domain match. An embed for something like `https://unscribd.com` would match as well.

It would be better to check for exact domains to make be safe.
```suggestion
		if ( ! in_array( wp_parse_url( $url, PHP_URL_HOST ), [ 'scribd.com', 'www.scribd.com' ], true ) ) {
```

includes/embeds/class-amp-scribd-embed-handler.php
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	/**
+	 * Remove the 'script' element from the provided HTML if there is any.
+	 *
+	 * @param string $html HTML string.
+	 * @return string
+	 */
+	private function remove_script( $html ) {
+		$html_without_script = preg_replace( '#<script[^>].+?</script>#s', '', $html );
```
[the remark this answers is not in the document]
schlessera: Yes, the `[^>]` is not needed here indeed, and would only be used to start matching against attributes of the script tag.
However, with the rise of web components, we cannot rely too much on a given set of tag names anymore. So, the above would completely strip prepending elements whose tag name happens to start with `script`. So it would turn this:
```html
<script-highlighter>
   <script>alert();</script>
</script-highlighter>
```
into this, breaking the HTML:
```html
</script-highlighter>
```
Therefore, it would be safer to only match a strict `script` tag, with optional attributes that are separated by whitespace:
```suggestion
		$html_without_script = preg_replace( '#<script(?:\s.*?)?>.+?</script>#s', '', $html );
```

includes/embeds/class-amp-scribd-embed-handler.php
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Add the sandbox attribute with 'allow-popups' & 'allow-scripts' permissions set so that the
+	 * full screen button works.
+	 *
+	 * @param string $html HTML string.
+	 * @return string
+	 */
+	private function inject_sandbox_attribute( $html ) {
+		if ( false === strpos( $html, 'sandbox="allow-scripts allow-popups"' ) ) {
```
schlessera: This can lead to broken HTML because of a duplicate `sandbox` attribute if the iframe already contains a sandbox attribute, but it has a different set or a different order of values, i.e. `<iframe sandbox="allow-popups allow-scripts">`.

The code should check for an existing sandbox attribute. If none is found, just add the one you need. If one is found, fetch the attribute values and add the two we need without introducing duplicates.

So, the code should be able to accept `<iframe sandbox="allow-forms allow-scripts">` and turn it into `<iframe sandbox="allow-forms allow-scripts allow-popups">`.

schlessera:
@pierlon Please add a test to the above PR to make sure that:
* the code does what it is supposed to do
* we won't break this code in the future through a regression

pierlon:
Thanks for the detailed review @kienstra. I've added a test to for the embed.

schlessera:
Good work on the test! 👍

includes/embeds/class-amp-scribd-embed-handler.php
```diff
@@ -0,0 +1,93 @@
...
+			return preg_replace(
+				'#(<iframe.+?sandbox=")(.+?)(".+</iframe>)#s',
+				"$1${matches['sandbox_attr']}$3",
+				$html,
+				1
+			);
+		} else {
+			return str_replace( '<iframe ', '<iframe sandbox="allow-scripts allow-popups" ', $html );
+		}
+	}
```
westonruter: I think this can be simplified a bit which will also eliminate the need for `remove_script`. 

Like this:

```php
private function sanitize_iframe( $html ) {
	return preg_replace_callback(
		'#^.*<iframe(?P<iframe_attributes>[^>]+?)></iframe>.*$#s',
		function ( $matches ) {
			$attr = $matches['iframe_attributes'];

			// Amend the required keywords to the iframe's sandbox.
			$sandbox  = 'allow-popups allow-scripts';
			$replaced = 0;
			$attr = preg_replace(
				'/(?<=\ssandbox=["\'])/',
				$sandbox,
				$attr,
				1,
				$replaced
			);

			// If no sandbox attribute was found, then add the attribute.
			if ( 0 === $replaced ) {
				$attr .= sprintf( ' sandbox="%s"', $sandbox );
			}

			// The iframe sanitizer will convert this into an amp-iframe.
			return "<iframe{$attr}></iframe>";
		},
		$html
	);
}
```

It's fine for `sandbox` keywords to be repeated.

includes/embeds/class-amp-scribd-embed-handler.php
```diff
@@ -0,0 +1,93 @@
...
+	 * @param string $url   Embed URL.
+	 * @return string Embed.
+	 */
+	public function filter_embed_oembed_html( $cache, $url ) {
+		if ( ! in_array( wp_parse_url( $url, PHP_URL_HOST ), [ 'scribd.com', 'www.scribd.com' ], true ) ) {
+			return $cache;
+		}
+
+		$embed = $this->remove_script( $cache );
+		$embed = $this->inject_sandbox_permissions( $embed );
```
westonruter: Per below, this could be replaced with:

```php
return $this->sanitize_iframe( $embed );
```

includes/embeds/class-amp-scribd-embed-handler.php
pierlon: So simple, yet I never thought of something like that. I have much to learn from this project. Thanks @westonruter!

pierlon:
I must thank everyone for giving me such detailed reviews and suggestions. I have much to learn from this project!

swissspidy:
@pierlon Thank _you_ for showing up and contributing to the plugin! :-)

Issue #3458, opened 2019-10-07
Lag when dragging elements


swissspidy:
## Bug Description

In #3060 it was mentioned:

* It seems to start lagging after adding a few blocks, and generally behaves so that the usability is not the best, it became not possible to drag.
* I have seen that for a while actually. It's in develop too and have been for about a week
* But when there's multiple elements on the page, you can't select-and-drag in one click (not always at least). You have to click to select first, then drag.

The last part might be covered by #3455, not sure though.

## Expected Behaviour

Consistently smooth interactions without lagging. No issues with selecting and dragging blocks.

## Screenshots

![dragging](https://user-images.githubusercontent.com/3294597/66332235-8de9ed00-e934-11e9-9396-018eee3bd854.gif)

## Additional context

 - WordPress version: 5.3 Beta
 - Plugin version: current `develop`
 - Gutenberg plugin version (if applicable): 6.6
 - OS: macOS
 - Browser: Chrome

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* AC1: When dragging a block, the block should move, along with the cursor, in real-time, so no lag is experienced on the screen.

## Implementation brief

Two optimizations are recommended for getting smoother snap line display:

* [_Debounce_ `setSnapLines` by 10 ms](https://github.com/ampproject/amp-wp/pull/new/fix/3458-smooth-snaplines).
* We currently make way more snap lines than "normal" tools (using Google Slides for comparison here). We currently allow _edge-to-edge_, _center-to-center_, _edge-to-center_ and _center-to-edge_ snap lines. We should reduce this by only allowing _edge-to-edge_ and _center-to-center_ snap lines.
* * The easiest way to do this is to split snap targets in two for each direction, e.g. separate `horizontalSnaps` in `horizontalEdgeSnaps` and `horizontalCenterSnaps` in `with-snap-target.js`. And consequently use these separately for calculations when finding optimal snap lines in `getBestSnapLines`.
* * Do note however, that snapping _edge-to-center-of-page_ is normal. It's only snapping _edge-to-center-of-other-object_, that's weird. Snapping _center-to-edge-of-page_ is still weird though, so centers should only ever snap to centers of other things.

## QA testing instructions

Verify smooth drag without lagging:
* Add several elements to a page.
* Drag any one around and see no delay while snap lines are calculated.

Verify proper snap lines between element and page:
* Add an image to a page.
[...]

PR #3484, opened 2019-10-11
Fix console warning for SnapLines

### LINKED ISSUE #3462: snapLines prop accidentally passed to DOM elements
## Bug Description

After adding the snapping, it looks like the `snapLines` prop is accidentally passed to the underlying DOM elements. This causes a console warning:

## Expected Behaviour

No console warning.

## Steps to reproduce

1. Insert video block
2. See console warning

## Screenshots

Screenshot 2019-10-08 at 10 15 11

## Additional context

 - WordPress version: 5.3 Beta
 - Plugin version: current `develop`
 - Gutenberg plugin version (if applicable): 6.6
 - Browser: Chrome

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

spacedmonkey:
## Summary
Fixes #3462

## Checklist

### DIFF
--- a/assets/src/stories-editor/components/contexts/snapping/index.js
+++ b/assets/src/stories-editor/components/contexts/snapping/index.js
@@ -25,7 +25,6 @@ const Snapping = ( { children } ) => {
 
 	const context = {
 		setSnapLines,
-		snapLines,
 		clearSnapLines,
 	};
 


miina:
Based on a comment from @barklund (https://github.com/ampproject/amp-wp/issues/3458#issuecomment-540759386) then we might not need the `snapLines` in the context either if it's not used by any of the components (now when it's removed from `resizable-box`)

spacedmonkey:
Thanks for the feedback @miina 

I have updated the PR so that snapLines are passed. It is a valid change and does fix the console error.

miina:
Thanks for the update.

Hmm, it looks like removing the `snapLines` from `resizable-box` was reverted. Was that intentional, to add it back? Since `snapLines` is not part of the context anymore then it looks like it's always `undefined` in `resizable-box` and could also be removed, as you had it before. Thoughts?

spacedmonkey:
> Hmm, it looks like removing the `snapLines` from `resizable-box` was reverted. Was that intentional, to add it back? Since `snapLines` is not part of the context anymore then it looks like it's always `undefined` in `resizable-box` and could also be removed, as you had it before. Thoughts?

 The first commit was removed and replaced with a second commit that fixes the issue as suggested. The current PR fixes the console warning. So can rereview as current commited change.

miina:
> The first commit was removed and replaced with a second commit that fixes the issue as suggested. The current PR fixes the console warning. So can rereview as current commited change.

I understand, what I'm saying is that now when the `snapLines` is removed from the `context` then it can be removed from `resizable-box` as well. Since it's not passed down anymore then it's always `undefined` in the `resizable-box` anyway, it has become kind of a "left-over" there. Thoughts?

Or what was the reasoning for removing it from `context` but leaving it to `resizable-box`?

spacedmonkey:
> I understand, what I'm saying is that now when the `snapLines` is removed from the `context` then it can be removed from `resizable-box` as well. Since it's not passed down anymore then it's always `undefined` in the `resizable-box` anyway, it has become kind of a "left-over" there. Thoughts?

Remove the snapLines from context, means that that prop is not passed down. When I console logged the props of `EnhancedResizableBox` see below.

```
ampFitText: undefined
angle: 0
blockName: "core/image"
children: {$$typeof: Symbol(react.element), key: null, ref: null, props: {…}, type: ƒ, …}
clearSnapLines: ƒ clearSnapLines()
clientId: "3742269e-c61a-4375-b5fd-095791049f28"
hasTextContent: false
height: 30
horizontalSnaps: ƒ (targetMin, targetMax)
minHeight: 30
minWidth: 40
onResizeStart: ƒ onResizeStart()
onResizeStop: ƒ onResizeStop(value)
parentBlockElement: div
setSnapLines: ƒ ()
snapGap: 8
verticalSnaps: ƒ (targetMin, targetMax)
width: 328
```

I can't see snapLines in there, so there is no need to remove it. Unless I am missing something.

miina:
> Remove the snapLines from context, means that that prop is not passed down. When I console logged the props of EnhancedResizableBox see below.

Exactly!

> I can't see snapLines in there, so there is no need to remove it. Unless I am missing something.

But why leave it in to the code if it's always undefined (since it's not passed down) and it's not used at all in the component?

Why do this:
```
const {
			...
			snapLines,
                         ...
		} = otherProps;
```

for a variable that's always `undefined`? It's an extra line of code which doesn't seem to have any purpose at all. It would be the same as if adding a random undefined variable there -- it doesn't break the code but it's unnecessary.

Could you explain what would be the purpose of leaving it in, what is the benefit?

spacedmonkey:
> Could you explain what would be the purpose of leaving it in, what is the benefit?

I haven't left it in? That line is not in the changelog. 

Screenshot from 2019-10-11 12-52-30

miina:
Why am I seeing this locally in the PR when the local branch claims to be up to date 😕

Uh! If there's something wrong with the branch locally then I apologize for the confusing comments.

miina:
Hmm, yep, there's something wrong with the local branch and it's not properly updating. I'm sorry for all the confusing comments, completely my bad!

PR #3529, opened 2019-10-15
Fix jump with font change

### LINKED ISSUE #3444: Blocks "jumping" when resizing to minimum limits fast

### LINKED ISSUE #3461: Text block jumps when changing normal to huge font
## Bug Description

Text block jumps when changing normal to huge font size.

## Expected Behaviour

Text block should not change its position 

## Steps to reproduce

1. Add a Text block and turn off Automatically fitting the text (amp-fit-text)
2. Add some text inside.
3. Rotate the block.
4. Change the font to "huge", without resizing the block larger before.
5. Observe results.

## Screenshots

jumpinh

## Additional context

 N/A

---------------

## Acceptance criteria

* AC1: When a block is placed on a page, regardless of size or rotational orientation, it should have a centered focal point, so when the block is rotated, or resized, it does not "jump" visually on the page.

Related to #3444 

### Cause of the bug and possible fix
The issue here is that when the text size is changed to "huge" then the block is resized, it's resized not considering the rotation.

We should adjust the logic so that if block size is increased, the rotation angle should be considered as well and the `positionTop` and `positionLeft` values adjusted accordingly.

The logic needs updating within the `maybeUpdateBlockDimensions` function.

The `resizable-box` component is already doing similar calculations, the relevant helpers are in the `helpers.js` file of that component. We can reuse the helpers and move the shared ones to the helper's folder instead, where applicable.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

miina:
## Summary

Fixes #3461 
Fixes #3444 

This PR creates a helper for getting an updated position after resizing. It also removes some seemingly redundant logic from resizing which seems to fix #3444 

## Checklist


miina:
Hmm, just saw that some tests are failing, will look into that tomorrow.

assets/src/stories-editor/components/resizable-box/helpers.js
```diff
@@ -190,3 +191,46 @@ export const getBlockPositioning = ( width, height, radian, direction ) => {
...
+ * @return {Object}  Top and left positions adjusted based on the rotation angle.
+ */
+export const getPositionAfterResizing = ( attributes ) => {
+	const {
+		direction,
+		angle,
+		isText,
+		width,
+		height,
+		appliedWidth,
```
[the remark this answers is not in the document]
spacedmonkey: I think it might be a bigger task to change the name of this variable as it used in the other component like this. Could we just add jsdocs explaining the attributes.

assets/src/stories-editor/helpers/maybeUpdateBlockDimensions.js
```diff
@@ -54,13 +56,8 @@ const maybeUpdateBlockDimensions = ( block ) => {
...
 				metaBlockElement.classList.toggle( 'is-measuring' );
 
-				if ( metaBlockElement.offsetHeight > height ) {
-					updateBlockAttributes( clientId, { height: metaBlockElement.offsetHeight } );
-				}
-
-				if ( metaBlockElement.offsetWidth > width ) {
-					updateBlockAttributes( clientId, { width: metaBlockElement.offsetWidth } );
-				}
+				newHeight = metaBlockElement.offsetHeight > height ? metaBlockElement.offsetHeight : null;
```
spacedmonkey: See above.

assets/src/stories-editor/components/resizable-box/test/getPositionAfterResizing.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Internal dependencies
+ */
+import { getPositionAfterResizing } from '../helpers';
+
+describe( 'getPositionAfterResizing', () => {
+	it( 'should return correct left and top position for text block', () => {
```
spacedmonkey: As the issues with #3444 . Can we also add tests to resize from top and right would prove that #3444 was fixed.

spacedmonkey:
One thing I noticed that I was looking into #3444 . 

The re-resizable component supports minHeight and minWidth. See the docs [here](https://github.com/bokuweb/re-resizable#minwidth-number--string). Min height and width are passed  `EnhancedResizableBox`, however they are [never passed](https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/components/resizable-box/index.js#L89)  `ResizableBox` as these fields are [exacted](https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/components/resizable-box/index.js#L57-L58) from props so are not part of `childProps`.  Not sure if it effects anything, but it just a strange.

spacedmonkey:
The `reachedMinLimit` is only used for min reached for text blocks is that correct. Why is this the case? [Other types of blocks](https://github.com/ampproject/amp-wp/blob/develop/assets/src/stories-editor/constants.js#L11-L17) have min heights as well.

miina:
> The reachedMinLimit is only used for min reached for text blocks is that correct. Why is this the case? Other types of blocks have min heights as well.

`reachedMinLimit` is meant for tracking specifically if the resizing has reached the limit to be able to _fit the text content_ inside:
https://github.com/ampproject/amp-wp/blob/5799032d4cc9882ceac94966f67c318ebb60eb68/assets/src/stories-editor/components/resizable-box/index.js#L187-L188

Ensuring that the minimum height and width for all the blocks would be respected happens here though:
https://github.com/ampproject/amp-wp/blob/5799032d4cc9882ceac94966f67c318ebb60eb68/assets/src/stories-editor/components/resizable-box/index.js#L181-L183

assets/src/stories-editor/components/resizable-box/test/getPositionAfterResizing.js
miina: There are some resizing tests which are resizing from top which are [failing right now](https://travis-ci.org/ampproject/amp-wp/jobs/598763242#L920-L970), I'm looking into those but it's possible that these are actually failing now due to #3444 being fixed, perhaps these were incorrect before.

spacedmonkey:
I found the variable name `reachedMinLimit` a little confusing, as I thought it would be related to any block reaching it's min and not just text blocks. Might be worth changing the variable name for readability. 

First of all, it would make the below lines 
```
let appliedWidth = minWidth <= width + deltaW ? width + deltaW : minWidth;
```
more readable if they looked like this 
```
let appliedWidth = Math.max ( minWidth, ( width + deltaW ) );
```

Why both with our own logic for min Height and width here?

miina:
> Why both with our own logic for min Height and width here?

We need to determine the `appliedWidth` and `appliedHeight` for being able to do the correct position calculations, it's not only for stopping resizing to respect the limits.

assets/src/stories-editor/components/resizable-box/helpers.js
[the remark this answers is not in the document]
miina: Both seemed like good ideas, so did both: 
641908c

assets/src/stories-editor/helpers/maybeUpdateBlockDimensions.js
```diff
@@ -24,24 +28,22 @@ const { updateBlockAttributes } = dispatch( 'core/block-editor' );
...
 
 			if ( element && content.length ) {
-				if ( element.scrollHeight > height ) {
-					updateBlockAttributes( clientId, { height: element.scrollHeight } );
-				}
-
-				if ( element.scrollWidth > width ) {
-					updateBlockAttributes( clientId, { width: element.scrollWidth } );
-				}
+				newHeight = element.scrollHeight > height ? element.scrollHeight : null;
```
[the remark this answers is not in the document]
miina: Added in 54749e6

assets/src/stories-editor/helpers/maybeUpdateBlockDimensions.js
miina: Added in 54749e6

assets/src/stories-editor/helpers/maybeUpdateBlockDimensions.js
```diff
@@ -70,6 +67,33 @@ const maybeUpdateBlockDimensions = ( block ) => {
 		default:
 			break;
 	}
+
+	if ( rotationAngle && ( newWidth || newHeight ) ) {
```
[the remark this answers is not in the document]
miina: Added in 54749e6

assets/src/stories-editor/components/resizable-box/test/getPositionAfterResizing.js
miina: After some investigation came to the conclusion that this was indeed the case: the tests had incorrect results since it jumped before when resizing. Adjusted the values within fe5cba0.

assets/src/stories-editor/helpers/maybeUpdateBlockDimensions.js
```diff
@@ -24,24 +28,23 @@ const { updateBlockAttributes } = dispatch( 'core/block-editor' );
...
 const maybeUpdateBlockDimensions = ( block ) => {
 	const { name, clientId, attributes } = block;
-	const { width, height, ampFitText, content } = attributes;
+	const { width, height, ampFitText, content, rotationAngle, positionLeft, positionTop } = attributes;
 
 	if ( ampFitText ) {
 		return;
 	}
 
+	let newHeight,
```
spacedmonkey: Alignment?

spacedmonkey:
Other than an alignment, the code seems good. Once that change was made, I will approve.

assets/src/stories-editor/helpers/maybeUpdateBlockDimensions.js
miina: It looks correct in the IDE at least. The first line as it is and the second line with one tab. What would you suggest?

spacedmonkey: I checked it out locally and you are right. That is a github preview issue. Very strange, it happens [here](https://github.com/ampproject/amp-wp/blob/fe5cba0211be36728b65a643d6ab43c5ff20edf2/assets/src/stories-editor/components/resizable-box/index.js#L27-L36) too.

Not a code issue. Resolving.

Issue #3279, opened 2019-09-17
Simple way to remove AMP story pages


spacedmonkey:
## Feature description

As a user I want there to be an easier way to remove pages from within the editor.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* AC1: It is currently possible to remove AMP story pages by pressing delete or using the remove block button on the in the top bar. However, there should be a simple way to remove pages within the UI.
* AC2: Add "Remove page" to the right-click context menu, upon selection, page will be removed.
* AC3: Include a "remove" link in the page panel, when a page is selected. Selecting the "remove" link will remove the page (see visual in comment below).

## Implementation brief
Implementation for this issue is broken into two piece of work. 

1. Add another a Remove page back on the `InspectorControls` for the edit method the page block. 
Mock up of which looks like this. 

![Screenshot from 2019-10-15 16-17-30](https://user-images.githubusercontent.com/237508/66845059-4f24ea00-ef67-11e9-959e-75bbca08c51d.png)

2. Add new options to the context menu that appears when a user right clicks in the stories editor. 
Currently the context menu is limiting to only appearing in one of two cases. First if a block other the the page block is selected, it shows options. Case two, if a page block is selected and there a block in the clipboard, show a 'Paste block' option. 

To implement this feature, some of the limitations and checks for if page is selected. These will have to be removed. The following limitations will need to be respected. 

- When a page block is selected, a user will not be able to copy or cut or page a page. 
- However, they is no limitation on duplication, so if simple, this should also be implemented. 
- Elements in popover menu, should say, remove page instead of remove block. 
- If only one page is in the story, remove page should not be displayed. 

A mockup of what this will look like. 
![Screenshot from 2019-10-15 15-40-31](https://user-images.githubusercontent.com/237508/66846117-34537500-ef69-11e9-81ed-62aa77137b4a.png)

### Test 1
- Create new story post. 
- Create a second page. 
- Select second page. 
- Click page to show right context menu (background settings). 
- Click remove page. 
- Confirm page is removed.   

### Test 2
- Create new story post. 
- Create a second page. 
- Select second page. 
- Right Click page to show context menu
- Click remove page. 
- Confirm page is removed.  

## Demo

* https://youtu.be/IyWXV3H9ysc
* https://youtu.be/ncaN-N2dLfA

## Changelog entry

* Added a new button to the sidebar to easily remove pages.

swissspidy:
[...]

PR #3536, opened 2019-10-15
Adjust issue reference in pull request template

barklund:
A small change to the PR template to invoke Github magic by using the phrase `Fixes #<id>` instead of `Addressses issue #<id>`.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,7 @@
 ## Summary
 
 <!-- Please reference the issue this PR addresses. -->
-Addresses issue #
+Fixes #
 
 ## Checklist
 


swissspidy:
It‘s like that on purpose as an issue is only considered closed when it has moved through the board(s) successfully. But perhaps it doesn‘t make too much of a difference.

Curious to hear how you‘d see this change the workflow.

barklund:
> It‘s like that on purpose as an issue is only considered closed when it has moved through the board(s) successfully. But perhaps it doesn‘t make too much of a difference.

Merging a PR that "fixes #x" does automatically close X yes, that's one of the features of this magic.

> Curious to hear how you‘d see this change the workflow.

Another feature of this magic is the link from issue to PR:

Screenshot 2019-10-15 at 15 57 50

The PR's listed up there in the header as _"May be fixed by..."_ are directly generated by them having "fixes"-references to the issue in question.

It's just a lot quicker to navigate from an issue to the corresponding PR without having to scroll through the issue to find a comment or similar reference to the PR.

Compare navigating to the PRs from #3279 **with** links and from #3447 **without** link (made with the old template without `fixes`).

swissspidy:
An alternative would be to add "Addressed By" to the issue template, though that would be a manual process. Plus not mutually exclusive.

PR #3544, opened 2019-10-16
Remove invalid processing instructions

### LINKED ISSUE #3409: Erroneous XML processing instructions (e.g. <?php ?>) are not sanitized
As reported on the [support forums](https://wordpress.org/support/topic/amp-1-3-0-validation-errors/), upon upgrading to AMP 1.3 they they started getting the AMP validator complaining about validation errors for this markup:

```html
<?php
$schema = get_post_meta(get_the_ID(), 'schema', true);
if(!empty($schema)) {
echo $schema;
}
?>
```

Naturally this is a bug since the PHP code is getting incorrectly included on the page. But still this `<?php ... ?>` processing instruction and others like it should be getting sanitized. It was actually getting removed in the previous version of the plugin, with an error code of `unknown`:

image

We should restore this behavior, or rather add a more specific error code like `invalid_processing_instruction`. The validator doesn't apparently allow _any_ processing instructions, including:

* 🚫 `<?xml version="1.0" encoding="UTF-8" ?>`
* 🚫 `<?xml-stylesheet type="text/xsl" href="style.xsl"?>`
* 🚫 `<?php /*...*/ ?>`

This would be a regression likely introduced by #3243.

pierlon:
Co-Authored-By: Weston Ruter <westonruter@google.com>

## Summary

Remove all invalid processing instructions and also display the `invalid_processing_instruction` error code for such occurrences.

## Screenshot:

image

Fixes #3409.

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -291,6 +291,8 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 						$result
 					);
 				}
+			} elseif ( $this_child instanceof DOMProcessingInstruction ) {
+				$this->remove_invalid_child( $this_child, [ 'code' => 'invalid_processing_instruction' ] );
 			}
 			$this_child = $next_child;
 		}
--- a/tests/php/test-tag-and-attribute-sanitizer.php
+++ b/tests/php/test-tag-and-attribute-sanitizer.php
@@ -1689,6 +1689,13 @@ class AMP_Tag_And_Attribute_Sanitizer_Test extends WP_UnitTestCase {
 				[],
 				[ 'invalid_processing_instruction' ],
 			],
+
+			'invalid_xml_pi'                               => [
+				'<?xml version="1.0" encoding="utf-8"?>',
+				'',
+				[],
+				[ 'invalid_processing_instruction' ],
+			],
 		];
 	}
 
--- a/tests/php/test-tag-and-attribute-sanitizer.php
+++ b/tests/php/test-tag-and-attribute-sanitizer.php
@@ -1682,6 +1682,13 @@ class AMP_Tag_And_Attribute_Sanitizer_Test extends WP_UnitTestCase {
 				[],
 				[ 'invalid_element' ],
 			],
+
+			'invalid_php_pi'                               => [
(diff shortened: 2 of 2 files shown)


tests/php/test-tag-and-attribute-sanitizer.php
```diff
@@ -1682,6 +1682,13 @@ static function() {
 				[],
 				[ 'invalid_element' ],
 			],
+
+			'invalid_php_pi'                               => [
```
westonruter: Please also add a test for (formerly) very-common `<?xml ... ?>` processing instruction.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -291,6 +291,8 @@ private function sanitize_element( DOMElement $element ) {
 						$result
 					);
 				}
+			} elseif ( $this_child instanceof DOMProcessingInstruction ) {
+				$this->remove_invalid_child( $this_child, [ 'code' => 'invalid_processing_instruction' ] );
```
westonruter: Please also check the other DOM node types that are not allowed. For example, a `DOMCdataSection` is not allowed in AMP:

image

Try adding checks for all of the various `dom` classes that can be generated by the HTML DOM parser: https://www.php.net/manual/en/book.dom.php

Perhaps `DOMNotation` (e.g. `<!NOTATION …>`), `DOMEntity` (e.g. `<!ENTITY …>`), and `DOMEntityReference` (e.g. `&foo;`)? Not every DOM class (or [node type](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType)) is possible to be instantiated from parsing, but please check them for good measure.

pierlon: According to the [DOM standard](https://dom.spec.whatwg.org/#node-trees) the possible nodes are `Document`, `DocumentType`, `DocumentFragment`, `Element`, `Text`, `ProcessingInstruction`, and `Comment`, which I think is all we need to consider.

The only invalid node then that should not be allowed in AMP would be `ProcessingInstruction`.

westonruter: OK, I wasn't able to cause the addition of a CDATA to result in an AMP validation error; the HTML parser actually just drops them from the document it seems. So yes, it appears we don't need to worry about them.

pierlon: > the HTML parser actually just drops them from the document 

Yes, that's what I found out as well. Reduces the work for us :smile: .

Issue #2487, opened 2019-05-31
Provide user opt-in for automatic updates


westonruter:
WordPress by default only auto-updates core patch versions. Plugins can also opt-in to automatic updates via:

```php
add_filter(
    'auto_update_plugin',
    function ( $update, $item ) {
        if ( in_array( $item->slug, [ 'amp' ], true ) ) {
            $update = true;
        }
        return $update;
    },
    10,
    2
);
```

However, doing this by default could be a violation of user expectations, since plugins normally only auto-update if the site owner has configured them to do so.

We could however provide a way for the user to opt-in to automatic updates. For example:

![image](https://user-images.githubusercontent.com/134745/58726290-3809a500-8396-11e9-8473-534a65b3f7dd.png)

Upon activation there could be an admin notice encouraging auto-updates to be enabled.
amending the default “Plugin activated.” message:

![image](https://user-images.githubusercontent.com/134745/58726310-43f56700-8396-11e9-983f-6aa3ad9f32eb.png)

Nevertheless, this could be a short-lived feature since auto-updating plugins is one of the 9 projects for 2019: https://make.wordpress.org/core/2018/12/08/9-priorities-for-2019/

igrigorik:
I would like to suggest the opposite: enable automatic updates by default and provide an opt-out. We can provide an explicit notification when this is first enabled, that site owners can dismiss or go to disable this feature if they want to. 

There is no reason why enabling this policy should block on wider WP rollout of automatic updates. To the contrary, that rollout will be gated on informed experience which the community lacks today, and we can start that journey by taking these baby steps. In the spirit of AMP... baking in best practices and guardrails that lead site owners to the right outcomes.

westonruter:
In contrast to core, the biggest problem with plugin auto-updates at the moment is that users are much more likely to have hacked plugin files. There is the "Plugin Editor" in WordPress that allows for this, even though it warns the user against it. (And there is no warning when users have modified and updated files via Git/FTP.) For them, auto-updates can mean losing changes, especially if they are counting updates always being manual.

Given that automatic updates for themes and plugins is already on the radar this year (as part of Matt's [9 Projects for 2019](https://make.wordpress.org/core/2018/12/08/9-priorities-for-2019/)), I think we should support that effort instead. If the whole ecosystem is transitioned to automatic updates for themes/plugins, then there will be a much wider impact, and we won't risk potential blowback for rolling it out for the AMP plugin prematurely.

amedina:
[...]

PR #3560, opened 2019-10-19
Opt in to receiving non-stable updates

### LINKED ISSUE #2084: Add ability to opt-in to a beta release channel

pierlon:
## Summary

This PR adds a plugin called "AMP Beta Tester" to opt-in to receiving updates released on GitHub.

Addresses issue #2084.

## To-do:

- [x] Hide plugin updates if working from source
- [x] Force WP to do a plugin update check on plugin activation (this allows for the plugin update cache to be modified)
- [x] Remove any plugin data and restore plugin update cache on deactivation
- [x] Update plugin details to reflect that of GitHub release
- [x] Move plugin to the correct folder when updated
- [x] Add an option to control auto updating of the plugin
- [x] Update tests for relevant sections of code

## Checklist:


beta-tester/amp-beta-tester.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace AMP\Beta_Tester;
+
+define( 'AMP__BETA_TESTER__DIR__', dirname( __FILE__ ) );
+
+require_once AMP__BETA_TESTER__DIR__ . '/includes/class-autoloader.php';
+Autoloader::register( __NAMESPACE__ );
+
+$amp_beta_tester_plugin = new AMP_Beta_Tester_Plugin();
+add_action( 'plugins_loaded', [ $amp_beta_tester_plugin, 'init' ] );
```
westonruter: This entire file can actually be located in the root of the plugin directory. So `amp-beta-tester.php` and `amp.php` both in the same directory. When this is done, then WordPress will automatically see both of the plugins. It may make sense to put all of the code in this single file and not use any classes for the logic in it. It could all be handled with namespaced functions.

pierlon: I forgot to mention that the reason for putting the plugin in its own folder was because the WordPress Plugin Update API uses the folder of the main PHP file (`amp` in this case) as an identifier for updates. So any other plugin file that is put in the `amp` folder will have the same version number as the AMP plugin.

One workaround for that would be to utilize the `all_plugins` filter hook to correct the flaw when the plugins list table is being shown.

pierlon: Here's an example of the workaround: https://github.com/ampproject/amp-wp/pull/3560/commits/060463ec2928435d8c6b7fd86ab0d24aeaa64990#diff-4805c837d3390517773926fb26726bcdR165-R178

amp-beta-tester.php
```diff
@@ -0,0 +1,52 @@
...
+ * @return void
+ */
+function show_missing_amp_notice() {
+	$error = sprintf(
+		// translators: %s: AMP plugin URL.
+		esc_html__(
+			'AMP Beta Tester requires AMP to be installed and active. You can download AMP %s.',
+			'amp-beta-tester'
+		),
+		'<a href="https://amp-wp.org" target="_blank">here</a>'
```
westonruter: Since the AMP Beta Tester plugin is distributed _with_ the AMP plugin pre-release builds, there would not be a possibility for the AMP plugin to not be installed. So I suggest this message can be simplified to just advising that the AMP plugin needs to be active, and the link can be removed.

beta-tester/amp-beta-tester.php
pierlon: The workaround doesn't work if AMP is deactivated, however:

image

westonruter: I suggest including the workaround code `amp.php` as well then.

amp-beta-tester.php
swissspidy: Worth noting that any notice in this file will not be translatable by users as the file is not part of the stable build and thus not available on the WordPress.org translation platform.

To solve this:

1. Add any translatable strings needed in this file also to `amp.php`.
2. Use the `amp` text domain in all cases

For example, we'd add this to the bottom of `amp.php`:

```php
__( 'AMP Beta Tester requires AMP to be installed and active. You can download AMP %s.', 'amp );
```

Then it will be translatable by users and part of the language packs downloaded by WordPress.

In `amp-beta-tester.php` you could then use the same string as usual.

pierlon:
At the moment, the plugin _works_ as intended; that is, upgrading the AMP plugin to the latest non-stable release on GitHub. Things to note:

- Upgrading will remove the plugin directory

- Upgrading to [1.4.0-beta1](https://github.com/ampproject/amp-wp/releases/tag/1.4.0-beta1) is not recommended, since it does not include this plugin

Regarding the failing tests, the main blocker is that files are not being `require`d correctly due to a problem with having multiple plugin files in one directory (see PR at https://github.com/xwp/wp-dev-lib/pull/307). I'm currently waiting on @kasparsd's approval so that it can get merged and released as a minor update.

tests/php/bootstrap.php
```diff
@@ -0,0 +1,20 @@
...
+ * `$_plugin_file`. One solution (hack) to fix this is to change the current directory to one where
+ * no plugin files can be found, then provide the path to the actual plugin file to require through
+ * the aforementioned global variable :D .
+ */
+
+global $_plugin_file;
+$current_dir = __DIR__;
+
+// Path to plugin to require.
+$_plugin_file = $current_dir . '/../../amp.php';
```
kasparsd: Could we extend the `WP_TEST_ACTIVATED_PLUGINS` environment variable with the additional plugin file instead?

https://github.com/ampproject/amp-wp/blob/e9aa2f3ebc2080106c8d54a07fe4a75f1df84184/phpunit.xml#L10

swissspidy: My understanding is that this is a workaround to ensure `amp.php` is loaded for the test framework, whereas `WP_TEST_ACTIVATED_PLUGINS` is for loading additional plugins that are required by the tests.

Without this workaround, the test framework tries to bootstrap by wrongly loading `amp-beta-tester.php` (as it's the first plugin file it finds)

swissspidy:
> Upgrading to `1.4.0-beta1` is not recommended, since it does not include this plugin

When I am running the current beta / `develop` branch, does it even suggest upgrading to `1.4.0-beta1`? If so, that sounds wrong.

phpcs.xml
```diff
@@ -11,6 +11,7 @@
 		<properties>
 			<property name="text_domain" type="array">
 				<element value="amp"/>
+				<element value="amp-beta-tester"/>
```
swissspidy: As mentioned before, only the `amp` text domain should be used.

```suggestion
```

amp-beta-tester.php
```diff
@@ -0,0 +1,204 @@
...
+}
+
+/**
+ * Determine if the supplied version code is a prerelease.
+ *
+ * @param string $plugin_version Plugin version code.
+ * @return bool
+ */
+function is_pre_release( $plugin_version ) {
+	return (bool) preg_match( '/^\d+\.\d+\.\d+-(beta|alpha)\d?$/', $plugin_version );
```
swissspidy: What about RC?

amp-beta-tester.php
```diff
@@ -0,0 +1,204 @@
...
+		ob_start();
+		?>
+		<script>
+			document.addEventListener('DOMContentLoaded', function() {
+				const link = document.querySelectorAll("[data-slug='amp'] a.thickbox.open-plugin-details-modal");
+
+				link.forEach( (link) => {
+					link.className = 'overridden'; // Override class so that onclick listeners are disabled.
+					link.target = '_blank';
+					link.href = '<?php echo $plugin_data['url']; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>';
```
swissspidy: Use `esc_js` or `wp_json_encode`

amp-beta-tester.php
```diff
@@ -0,0 +1,204 @@
...
+function get_amp_github_releases() {
+	$raw_response = wp_remote_get( 'https://api.github.com/repos/ampproject/amp-wp/releases' );
+	if ( is_wp_error( $raw_response ) ) {
+		return null;
+	}
+	return json_decode( $raw_response['body'] );
+}
+
+/**
+ * Replace the 'View version details' link with the link to the release on GitHub.
```
swissspidy: Might wanna look at how [Site Kit](https://github.com/google/site-kit-wp) is doing this.

amp-beta-tester.php
```diff
@@ -0,0 +1,204 @@
...
+	if ( ! isset( $updates->no_update ) ) {
+		return $updates;
+	}
+
+	$amp_zip_file = 'amp.zip';
+	$amp_manifest = isset( $updates->response[ AMP_PLUGIN_FILE ] )
+		? $updates->response[ AMP_PLUGIN_FILE ]
+		: $updates->no_update[ AMP_PLUGIN_FILE ];
+
+	$github_releases = get_amp_github_releases();
```
swissspidy: Where is this cached?

amp-beta-tester.php
```diff
@@ -0,0 +1,204 @@
...
+				if ( array_key_exists( 'amp/amp-beta-tester.php', $updates->response ) ) {
+					unset( $updates->response['amp/amp-beta-tester.php'] );
+				}
+			}
+
+			return $updates;
+		}
+	);
+}
+register_activation_hook( __FILE__, __NAMESPACE__ . '\force_plugin_update_check' );
```
swissspidy: The plugin activation hook could `activate_plugin( 'amp/amp.php' )` ([docs](https://developer.wordpress.org/reference/functions/activate_plugin/))

tests/php/bootstrap.php
kasparsd: Do we want to include `amp/amp-beta-tester.php` in `WP_TEST_ACTIVATED_PLUGINS` if the main plugin file is set to `amp.php`?

kasparsd:
@pierlon [Version 1.3.0 of wp-dev-lib](https://github.com/xwp/wp-dev-lib/releases/tag/1.3.0) is out now.

westonruter:
> @pierlon [Version 1.3.0 of wp-dev-lib](https://github.com/xwp/wp-dev-lib/releases/tag/1.3.0) is out now.

Updated `develop`: https://github.com/ampproject/amp-wp/pull/3589

amp-beta-tester.php
pierlon: Missed that :man_facepalming: .

amp-beta-tester.php
pierlon: I don't think it should be. The filter hook that calls this function (`pre_set_site_transient_update_plugins`) is only executed when the `update_plugins` transient is being set (which is usually not often).

amp-beta-tester.php
pierlon: Should it be activated programatically? There might be a scenario where the user wishes to keep the AMP plugin updated but not activated (for whatever reason). I think the notice telling the user that the AMP plugin must be activated is good enough.

swissspidy: > There might be a scenario where the user wishes to keep the AMP plugin updated but not activated (for whatever reason).

Makes sense. But how is this doable when the beta tester plugin aborts early when the AMP plugin is not active?

Following that logic, my understanding is that in this case the beta tester plugin should work even when the AMP plugin is inactive, and should not show any warning.

Just my two cents while reading through the code :-)

pierlon: As far as I can tell the `register_activation_hook` hook would only be called upon plugin activation, while the `init` hook would be called upon every request, so it should show the notice regardless.

swissspidy: Right. What I meant is that due to the early `return` in `init()`, the beta tester plugin does not work when the AMP plugin is not active.

However, above you mentioned that might be a valid use case:

> There might be a scenario where the user wishes to keep the AMP plugin updated but not activated

Correct me if I'm wrong :-)

amp-beta-tester.php
```diff
@@ -0,0 +1,204 @@
...
+				link.forEach( (link) => {
+					link.className = 'overridden'; // Override class so that onclick listeners are disabled.
+					link.target = '_blank';
+					link.href = '<?php echo $plugin_data['url']; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>';
+				} );
+			}, false);
+		</script>
+		<?php
+
+		echo ob_get_clean(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
swissspidy: Output buffering doesn't really make sense here. I'd just do:

```php
function replace_view_version_details_link( $file, $plugin_data ) {
	$plugin_version = $plugin_data['Version'];
	if ( ! is_pre_release( $plugin_version ) ) {
		return;
	}
	?>
		<script>...</script>
	<?php
}
```

amp-beta-tester.php
pierlon: Oh, misunderstood your concern there. The status of the AMP plugin will not interfere with this one, as WP detects them as two separate plugins :smile:.

pierlon: Oh, misunderstood your concern. I'll modify `init` to accommodate that scenario :smile:.

pierlon: Here's the new logic:

https://github.com/ampproject/amp-wp/blob/ac056ebec93825bdf2725d30b0d2d0e852d8358d/amp-beta-tester.php#L55-L63

amp-beta-tester.php
pierlon: > > Upgrading to `1.4.0-beta1` is not recommended, since it does not include this plugin
> 
> When I am running the current beta / `develop` branch, does it even suggest upgrading to `1.4.0-beta1`? If so, that sounds wrong.

@swissspidy, fixed.

pierlon:
> > Upgrading to `1.4.0-beta1` is not recommended, since it does not include this plugin
> 
> When I am running the current beta / `develop` branch, does it even suggest upgrading to `1.4.0-beta1`? If so, that sounds wrong.

@swissspidy, fixed.

pierlon:
Just to note, an update for the beta plugin will show if both the AMP and beta tester plugins are not active:

image

See https://github.com/ampproject/amp-wp/pull/3560#discussion_r336763645 as to why.

swissspidy:
Lots of great progress here, really awesome to see! 🎉 

> Just to note, an update for the beta plugin will show if both the AMP and beta tester plugins are not active:

I think the AMP Beta Tester plugin should just have the same version number as the plugin then to solve this.

amp-beta-tester.php
```diff
@@ -0,0 +1,202 @@
...
+
+register_activation_hook( __FILE__, __NAMESPACE__ . '\force_plugin_update_check' );
+add_action( 'plugins_loaded', __NAMESPACE__ . '\init' );
+
+/**
+ * Force a plugin update check. This will allows us to modify the plugin update cache so we
+ * can set a custom update.
+ */
+function force_plugin_update_check() {
+	if ( defined( 'DOING_CRON' ) ) {
```
swissspidy:
```suggestion
	if ( wp_doing_cron() ) {
```

amp-beta-tester.php
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * @param string $plugin_version Plugin version code.
+ * @return bool
+ */
+function is_pre_release( $plugin_version ) {
+	return (bool) preg_match( '/^\d+\.\d+\.\d+-(beta|alpha|RC)\d?$/', $plugin_version );
+}
+
+/**
+ * Displays the version code in the admin bar to act as a reminder that an unstable version
```
swissspidy: My main concern with this reminder is that it's shown to **every** user, even those who cannot `manage_options` and thus have no idea that AMP is even installed.

If I _can_ `manage_options`, I am likely aware that I am running the beta plugin of AMP. Thus, I don't think any such notice is really useful.

I'm curious to learn how other plugins like Jetpack Beta Tester or WooCommerce Beta Tester do this.

.dev-lib
```diff
@@ -6,6 +6,7 @@ PROJECT_SLUG=amp
 README_MD_TITLE="AMP Plugin for WordPress"
 DEV_LIB_SKIP="$DEV_LIB_SKIP,jshint"
 CHECK_SCOPE=all
+PLUGIN_FILE="$(pwd)/amp.php"
```
westonruter: To make this a bit more robust, it would probably be best if `wp-dev-lib` supplied the directory name and the environment variable here was only set as `amp.php`. Just thinking that as it stands right now, if the user is not currently in the plugin's root directory, then this could fail to work as expected. Granted this is unlikely. Nevertheless, it seems unnecessary to always require adding `$(pwd)/` every time.

amp-beta-tester.php
```diff
@@ -0,0 +1,202 @@
...
+		<script>
+			document.addEventListener('DOMContentLoaded', function() {
+				const links = document.querySelectorAll("[data-slug='amp'] a.thickbox.open-plugin-details-modal");
+
+				links.forEach( (link) => {
+					link.className = 'overridden'; // Override class so that onclick listeners are disabled.
+					link.target = '_blank';
+					<?php
+					if ( isset( $plugin_data['url'] ) ) {
+						echo "link.href = '" . esc_url( $plugin_data['url'] ) . "';";
```
westonruter: The `esc_url()` function intends an HTML target, rather than a JS one. This should be better:

```suggestion
						printf( 'link.href = %s;', wp_json_encode( esc_url_raw( $plugin_data['url'] ) ) );
```

amp-beta-tester.php
```diff
@@ -0,0 +1,202 @@
...
+ * @param string $file Plugin file.
+ * @param array  $plugin_data Plugin data.
+ */
+function replace_view_version_details_link( $file, $plugin_data ) {
+	$plugin_version = $plugin_data['Version'];
+
+	if ( is_pre_release( $plugin_version ) ) {
+		?>
+		<script>
+			document.addEventListener('DOMContentLoaded', function() {
```
westonruter: For good measure:

```suggestion
			document.addEventListener('DOMContentLoaded', () => {
```

amp-beta-tester.php
```diff
@@ -0,0 +1,202 @@
...
+}
+
+/**
+ * Determine if the supplied version code is a prerelease.
+ *
+ * @param string $plugin_version Plugin version code.
+ * @return bool
+ */
+function is_pre_release( $plugin_version ) {
+	return (bool) preg_match( '/^\d+\.\d+\.\d+-(beta|alpha|RC)\d?$/', $plugin_version );
```
westonruter: The major pre-releases have historically not always included the patch version.

```suggestion
	return (bool) preg_match( '/^\d+\.\d+(\.\d+)?-(beta|alpha|RC)\d?$/', $plugin_version );
```

amp-beta-tester.php
```diff
@@ -0,0 +1,202 @@
...
+ * of AMP is being used.
+ */
+function show_unstable_reminder() {
+	global $wp_admin_bar;
+
+	if ( is_pre_release( AMP__VERSION ) ) {
+		$args = [
+			'id'     => 'amp-beta-tester-admin-bar',
+			'title'  => sprintf( 'AMP v%s', AMP__VERSION ),
+			'parent' => 'top-secondary',
```
westonruter: Rather than have this as a top-level menu item, I suggest adding it as a submenu item under the top-level AMP admin bar menu item. Then it won't be persistently taking up real estate on the top-level.

amp-beta-tester.php
```diff
@@ -0,0 +1,202 @@
...
+			'href'   => admin_url( 'admin.php?page=amp-options' ),
```
westonruter: It needn't have a link even, or if it does, it could link to the plugins screen with "amp" being the search term to filter the list down to just the AMP plugin and the AMP Beta Tester plugin (among others), for example:

```suggestion
			'href'   => admin_url( 'plugins.php?s=amp&plugin_status=active' ),
```

amp-beta-tester.php
```diff
@@ -0,0 +1,202 @@
...
+ * Displays the version code in the admin bar to act as a reminder that an unstable version
+ * of AMP is being used.
+ */
+function show_unstable_reminder() {
+	global $wp_admin_bar;
+
+	if ( is_pre_release( AMP__VERSION ) ) {
+		$args = [
+			'id'     => 'amp-beta-tester-admin-bar',
+			'title'  => sprintf( 'AMP v%s', AMP__VERSION ),
```
westonruter:
```suggestion
			'title'  => sprintf( 'v%s', AMP__VERSION ),
```

amp-beta-tester.php
```diff
@@ -0,0 +1,202 @@
...
+			'parent' => 'top-secondary',
+			'href'   => admin_url( 'admin.php?page=amp-options' ),
+		];
+		$wp_admin_bar->add_node( $args );
+
+		// Highlight the menu.
+		echo '<style>#wpadminbar #wp-admin-bar-amp-beta-tester-admin-bar { background: #0075C2; }</style>';
```
westonruter: This may then be unnecessary, but if it is, then make sure to either add `data-amp-dev-mode` to the `style` element or even better do (which should do this automatically):

```suggestion
		wp_add_inline_style( 'admin-bar', '#wpadminbar #wp-admin-bar-amp-beta-tester-admin-bar { background: #0075C2; }' );
```

westonruter:
When the user has the AMP Beta Tester plugin active, what about automatically updating them to the latest pre-release version? See https://github.com/ampproject/amp-wp/issues/2487. Perhaps this should be an opt-in setting? Probably better to be opt-in because sometimes you'll make a plugin build from a branch and you'll want to test that specifically, and it would be undesirable to have the plugin update itself unexpectedly. 

(Alternatively, we could change the logic for generating the plugin version number to include the current branch name if not `develop` and only auto-update if a branch build is currently active.)

Perhaps the auto-update functionality should actually be a setting in the main plugin itself, so if the AMP Beta Tester plugin is active then the user will keep on the latest pre-release version. And if the AMP Beta Tester plugin is _not_ active then it could keep them up-to-date to the latest stable version.

amp-beta-tester.php
pierlon: I actually followed [Jetpack Beta Tester's approach](https://github.com/Automattic/jetpack-beta/blob/929878af22965aacb6db1544a5d69c50807e2caa/jetpack-beta.php#L306-L357).

amp-beta-tester.php
pierlon: The plugin now hooks into the WP plugins API to show the correct version and also updates the download URL.

pierlon:
Plugin now updates the WP plugins API to reflect the pre-release details; the version code and download URL is updated:

image

pierlon:
Things to note:

- Activating the plugin right after installation will only activate AMP
- Deleting either plugin deletes the folder, so both plugins are deleted

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+
+/**
+ * Fetch AMP releases from GitHub.
+ *
+ * @return array|false
+ */
+function get_amp_github_releases() {
+	$raw_response = wp_remote_get( 'https://api.github.com/repos/ampproject/amp-wp/releases' );
+	if ( is_wp_error( $raw_response ) ) {
+		return false;
```
schlessera: The way this is combined with the transient can cause major issues if the connection is faulty or GitHub has issues.

As this returns `false` on error, the consuming code will store `false` in a transient, and in a subsequent request check if `empty( false )`, thus making a remote GitHub request on every single page request until the situation is resolved. That can easily get the server blacklisted.

Returning an empty array here, and checking for `false` in the consuming code would store the faulty response and only retry much later instead.
```suggestion
		return [];
```

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+ * Get the AMP plugin update manifest for the specified version from GitHub.
+ *
+ * @param string $version Version to get manifest for. Defaults to getting the latest pre-release.
+ * @return object|false Latest release, or false on failure.
+ */
+function get_amp_update_manifest($version = 'pre-release' ) {
+	$amp_manifest = null;
+	$releases     = get_site_transient( AMP__BETA__TESTER__RELEASES__TRANSIENT );
+
+	if ( empty( $releases ) ) {
```
schlessera: See comment above regarding storing the transient.
```suggestion
	if ( false === $releases ) {
```

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+	return json_decode( $raw_response['body'] );
+}
+
+/**
+ * Retrieves the download url for amp.zip, if it exists.
+ *
+ * @param object $release GitHub release JSON object.
+ * @return string|false Download URL if it exists, false if not.
+ */
+function get_download_url_from_amp_release($release ) {
```
schlessera:
```suggestion
function get_download_url_from_amp_release( $release ) {
```

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+}
+
+/**
+ * Retrieves the current AMP update manifest, and updates it to include the analogous information
+ * from its GitHub release.
+ *
+ * @param object $release GitHub release JSON object.
+ * @return array|false Updated manifest, or false if it fails to retrieve the current update manifest.
+ */
+function generate_amp_update_manifest($release ) {
```
schlessera:
```suggestion
function generate_amp_update_manifest( $release ) {
```

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+function get_current_amp_update_manifest() {
+	$updates = get_site_transient( 'update_plugins' );
+
+	if ( ! isset( $updates->response, $updates->no_update ) ) {
+		return false;
+	}
+
+	if ( isset( $updates->response[ AMP__PLUGIN__BASENAME ] ) ) {
+		$manifest = $updates->response[ AMP__PLUGIN__BASENAME ];
+	} elseif ( isset( $updates->no_update[ AMP__PLUGIN__BASENAME ] ) ) {
```
schlessera: As you've already checked that both cannot be unset at the same time, `isset( $updates->response[ AMP__PLUGIN__BASENAME ] )` being `false` implies that `isset( $updates->response[ AMP__PLUGIN__BASENAME ] )` is `true`.

So you can turn this into a regular `else` and drop the third condition entirely.

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+}
+
+/**
+ * Determine if the supplied version code is a prerelease.
+ *
+ * @param string $plugin_version Plugin version code.
+ * @return bool
+ */
+function is_pre_release( $plugin_version ) {
+	return (bool) preg_match( '/^\d+\.\d+(\.\d+)?-/', $plugin_version );
```
schlessera: Maybe add an explanatory comment here about what the regex is checking.
```suggestion
	// A plugin version is a pre-release if it has a `-<suffix>`, so we're
	// checking for a hyphen after the version numbers.
	return (bool) preg_match( '/^\d+\.\d+(\.\d+)?-/', $plugin_version );
```

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+}
+
+/**
+ * Get the current AMP version.
+ *
+ * @param bool $strip_build_info Whether to strip build information or not.
```
schlessera:
```suggestion
 * @param bool $strip_build_info Optional. Whether to strip build information or not. Defaults to true.
```

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+function get_amp_version( $strip_build_info = true ) {
+	$amp_version = defined( 'AMP__VERSION' )
+		? AMP__VERSION
+		: get_plugin_data( WP_PLUGIN_DIR . '/' . AMP__PLUGIN__BASENAME )['Version'];
+
+	if ( $strip_build_info ) {
+		// Strip the timestamp and commit hash from the plugin version if it exists.
+		preg_match( '/[^-]*-[^-]*/', $amp_version, $amp_version );
+		$amp_version = $amp_version[0];
+	}
```
schlessera: Is this actually being used somewhere? It seems that it is dead code.

schlessera:
Awesome work so far, @pierlon !

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+
+define( 'AMP__BETA_TESTER__DIR__', dirname( __FILE__ ) );
+define( 'AMP__BETA__TESTER__RELEASES__TRANSIENT', 'amp_releases' );
+define( 'AMP__PLUGIN__BASENAME', 'amp/amp.php' );
+
+// DEV_CODE. This block of code is removed during the build process.
+if ( file_exists( AMP__BETA_TESTER__DIR__ . '/amp.php' ) ) {
+	add_filter(
+		'site_transient_update_plugins',
+		function ( $updates ) {
```
schlessera: No `$this` context, so closure can be static for performance reasons.
```suggestion
		static function ( $updates ) {
```

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+	return array_merge( (array) $current_manifest, $manifest );
+}
+
+/**
+ * Get the AMP plugin update manifest for the specified version from GitHub.
+ *
+ * @param string $version Version to get manifest for. Defaults to getting the latest pre-release.
+ * @return object|false Latest release, or false on failure.
+ */
+function get_amp_update_manifest($version = 'pre-release' ) {
```
schlessera:
```suggestion
function get_amp_update_manifest( $version = 'pre-release' ) {
```

amp.php
```diff
@@ -161,6 +161,22 @@
 	);
 }
 
+// DEV_CODE. This block of code is removed during the build process.
+if ( file_exists( AMP__DIR__ . '/amp-beta-tester.php' ) ) {
+	add_filter(
+		'site_transient_update_plugins',
+		function ( $updates ) {
```
schlessera: No `$this` context, so closure can be static for performance reasons.
```suggestion
		static function ( $updates ) {
```

composer.json
```diff
@@ -10,6 +10,7 @@
     "ext-date": "*",
     "ext-dom": "*",
     "ext-iconv": "*",
+    "ext-json": "*",
```
schlessera: If you add the extension here as required, it needs to be removed from the `"suggest"` section below.

tests/php/test-amp-beta-tester.php
```diff
@@ -0,0 +1,528 @@
...
+		];
+	}
+
+	/**
+	 * Test \AMP_Beta_Tester\update_amp_manifest().
+	 *
+	 * @dataProvider get_plugin_updates_data
+	 * @covers \AMP_Beta_Tester\update_amp_manifest()
+	 *
+	 * @param stdClass $source   Source.
```
schlessera:
```suggestion
	 * @param stdClass $github_response GitHub response.
	 * @param stdClass $source          Source.
```

tests/php/test-amp-beta-tester.php
```diff
@@ -0,0 +1,528 @@
...
+	 *
+	 * @dataProvider get_view_version_details_data
+	 * @covers \AMP_Beta_Tester\replace_view_version_details_link()
+	 *
+	 * @param array $source   Source.
+	 * @param string $expected Expected.
+	 */
+	public function test__view_version_details( $source, $expected ) {
+		ob_start();
+		AMP_Beta_Tester\replace_view_version_details_link( null, $source );
```
schlessera: This function doesn't seem to exist.

tests/php/test-amp-beta-tester.php
```diff
@@ -0,0 +1,528 @@
...
+			$this->markTestSkipped( 'Unstable reminder is only shown on non-stable versions.' );
+		}
+
+		global $show_admin_bar, $wp_admin_bar;
+
+		$show_admin_bar = true;
+
+		_wp_admin_bar_init();
+
+		AMP_Beta_Tester\show_unstable_reminder( $wp_admin_bar );
```
schlessera: This function doesn't seem to exist.

tests/php/test-amp-beta-tester.php
```diff
@@ -0,0 +1,528 @@
...
+	 */
+	private function generate_github_api_response( $version = '1.0.0-beta', $is_pre_released = true, $with_zip = true ) {
+		$response = [
+			'url'              => 'https://api.github.com/repos/ampproject/amp-wp/releases/20819569',
+			'assets_url'       => 'https://api.github.com/repos/ampproject/amp-wp/releases/20819569/assets',
+			'upload_url'       => 'https://uploads.github.com/repos/ampproject/amp-wp/releases/20819569/assets{?name,label}',
+			'html_url'         => "https://github.com/ampproject/amp-wp/releases/tag/${version}",
+			'id'               => 20819569,
+			'node_id'          => 'MDc6UmVsZWFzZTIwODE5NTY5',
+			'tag_name'         => "${version}",
```
schlessera: An actual cast will better communicate the intention here.
```suggestion
			'tag_name'         => (string) $version,
```

tests/php/test-amp-beta-tester.php
```diff
@@ -0,0 +1,528 @@
...
+			'target_commitish' => 'develop',
+			'name'             => "${version}",
```
schlessera:
```suggestion
			'name'             => (string) $version,
```

pierlon:
All that's left here is to update the tests, and a good code review of course :smile:.

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
+<?php
+/**
+ * Plugin Name: AMP Beta Tester
+ * Description: Opt-in to receive non-stable release builds for the AMP plugin.
+ * Plugin URI: https://amp-wp.org
+ * Author: AMP Project Contributors
+ * Author URI: https://github.com/ampproject/amp-wp/graphs/contributors
+ * Version: 1.4.0-beta1
+ * Text Domain: amp
```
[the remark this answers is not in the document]
pierlon: Since the least version of WordPress being tested against is 4.9 on Travis, I think that should be used.

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+ * Domain Path: /languages/
+ * License: GPLv2 or later
+ *
+ * @package AMP Beta Tester
+ */
+
+namespace AMP_Beta_Tester;
+
+define( 'AMP__BETA_TESTER__DIR__', dirname( __FILE__ ) );
+define( 'AMP__BETA__TESTER__RELEASES__TRANSIENT', 'amp_releases' );
```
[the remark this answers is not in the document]
pierlon: No actually, I was only following the same convention in [amp.php](https://github.com/ampproject/amp-wp/blob/enhancement/2084-add-opt-in-to-beta-feature/amp.php#L16-L18).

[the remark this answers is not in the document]
pierlon: I've normalized them in https://github.com/ampproject/amp-wp/pull/3560/commits/c392dba8bebee093979d6e126bc44c35e304983b.

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+
+	$latest_manifest = get_amp_update_manifest();
+
+	if ( ! $latest_manifest ) {
+		$current_manifest = get_amp_update_manifest( get_amp_version() );
+
+		if ( $current_manifest ) {
+			unset( $updates->response[ AMP__PLUGIN__BASENAME ] );
+			$updates->no_update[ AMP__PLUGIN__BASENAME ] = $current_manifest;
+		}
```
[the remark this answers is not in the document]
pierlon: If I remember correctly, `get_amp_update_manifest()` here would return `false` if it failed to retrieve the updates from GitHub or failed to return an update manifest. Nevertheless, the logic for this has been updated and can be found [here](https://github.com/ampproject/amp-wp/blob/enhancement/2084-add-opt-in-to-beta-feature/amp-beta-tester.php#L184-L205).

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+ * Text Domain: amp
+ * Domain Path: /languages/
+ * License: GPLv2 or later
+ *
+ * @package AMP Beta Tester
+ */
+
+namespace AMP_Beta_Tester;
+
+define( 'AMP__BETA_TESTER__DIR__', dirname( __FILE__ ) );
```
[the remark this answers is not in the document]
pierlon: Done in https://github.com/ampproject/amp-wp/pull/3560/commits/c392dba8bebee093979d6e126bc44c35e304983b.

amp-beta-tester.php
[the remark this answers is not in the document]
pierlon: Added requirements in https://github.com/ampproject/amp-wp/pull/3560/commits/e13f56ccf63213d6940269a385a3f69c7aab83e5.

amp-beta-tester.php
```diff
@@ -0,0 +1,292 @@
...
+/**
+ * Update the AMP plugin information data to reflect that of the GitHub release.
+ *
+ * @param false|object|array $value  The result object or array. Default false.
+ * @param string             $action The type of information being requested from the Plugin Installation API.
+ * @param object             $args   Plugin API arguments.
+ * @return false|object|array Updated $value, or passed-through $value on failure.
+ */
+function update_amp_plugin_information( $value, $action, $args ) {
+	if ( 'plugin_information' !== $action || 'amp' !== $args->slug ) {
```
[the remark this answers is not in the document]
pierlon: Done in https://github.com/ampproject/amp-wp/pull/3560/commits/35caa1390b7d529e51fbb08bb633523a0c6a65c5.

amp-beta-tester.php
pierlon: No longer the case anymore as of e2c009787c95cb7cff7e4f73bb13e32ac64f184a.

amp-beta-tester.php
pierlon: Yes, that makes sense. Done in https://github.com/ampproject/amp-wp/pull/3560/commits/1dafb7fecc143e920fce7c494f8c09931fcaff44.

amp-beta-tester.php
pierlon: No longer necessary as function has been removed.

amp-beta-tester.php
pierlon: No longer necessary as parameter has been removed.

amp-beta-tester.php
pierlon: Code block here has been removed, no longer dead code :smile: .

amp-beta-tester.php
pierlon: Thanks, done in https://github.com/ampproject/amp-wp/pull/3560/commits/7c19a8e351362ea70e6c1c6ffbe0adfb097fbd67.

composer.json
pierlon: Thanks, done in 4165a25a86f917eb2ab8668b00a244c925691528.

PR #3582, opened 2019-10-21
CTA responsiveness fixes

### LINKED ISSUE #3343: CTA button is only initially centered, needing re-centering after adding text

miina:
## Summary

Fixes #3343 

- [x] Positions the CTA button 5% from left as other blocks.
- [x] Fix width and height by considering 12px 24px padding assigned in the FE. Currently that creates inconsistencies between the FE and editor.

## Checklist

### DIFF
--- a/assets/css/src/amp-stories.css
+++ b/assets/css/src/amp-stories.css
@@ -34,7 +34,7 @@ amp-story-grid-layer[template="fill"] .wp-block-image {
 	font-size: 18px;
 	margin: 0;
 	max-height: calc(100% - 24px);
-	padding: 12px 24px;
+	padding: 7px 24px;
 	text-align: center;
 	text-decoration: none;
 	white-space: normal;
--- a/assets/src/stories-editor/blocks/amp-story-cta/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-cta/edit.js
@@ -22,9 +22,9 @@ import { getBackgroundColorWithOpacity } from '../../../common/helpers';
 import { DraggableText } from '../../components';
 import { STORY_PAGE_INNER_HEIGHT_FOR_CTA } from '../../constants';
 
-// Total padding of top + bottom / left + right.
-const CTA_BUTTON_PADDING_TOP_BOTTOM = 7;
-const CTA_BUTTON_PADDING_LEFT_RIGHT = 38;
+// Total padding of top + bottom (vertical) / left + right (horizontal).
+const CTA_BUTTON_PADDING_VERTICAL = 14;
+const CTA_BUTTON_PADDING_HORIZONTAL = 48;
 
 const CallToActionEdit = ( {
 	attributes,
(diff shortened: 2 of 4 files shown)


miina:
Found additional issue when testing: padding in the FE is added in addition to the assigned width/height and this causes differences between FE and editor. Will look into fixing this so that it wouldn't cause block validation errors.

assets/src/stories-editor/blocks/amp-story-cta/edit.js
```diff
@@ -22,6 +22,10 @@ import { getBackgroundColorWithOpacity } from '../../../common/helpers';
 import { DraggableText } from '../../components';
 import { STORY_PAGE_INNER_HEIGHT_FOR_CTA } from '../../constants';
 
+// Total padding of top + bottom / left + right.
+const CTA_BUTTON_PADDING_TOP_BOTTOM = 7;
+const CTA_BUTTON_PADDING_LEFT_RIGHT = 38;
```
swissspidy: In `amp-stories.css`, `.amp-block-story-cta__link` has `padding: 7px 24px;` - where does `38` come from?

assets/css/src/amp-stories.css
```diff
@@ -34,7 +34,7 @@ amp-story-grid-layer[template="fill"] .wp-block-image {
 	font-size: 18px;
 	margin: 0;
 	max-height: calc(100% - 24px);
-	padding: 12px 24px;
+	padding: 7px 24px;
```
barklund: This reflects the button in FE, or?

miina: Yep, this change should remove the difference between the editor and FE. Are you thinking that perhaps it would be better to change it in the editor instead? The 12px seemed too much when testing out with the added amp-fit-text.

barklund: I was just curious. Seems legit to me :)

assets/src/stories-editor/blocks/amp-story-cta/edit.js
miina: This padding is based on padding the values in the editor since the width and height are taken from the editor. The 7px is the top padding that the CTA button has in the editor (there is no bottom padding) and 38 is the left + right padding together (19 + 19).
Additionally, in the editor, there is a wrapper element with extra padding of 5px on the left and right which in total makes 24px as in the FE, however, since the width and height are not taken from the outer wrapper then we only have to deduct 19px.

I'll check if we can take the width and height from the outer wrapper instead to make it more understandable.

barklund: I think a comment (and maybe variable renaming, maybe vertical and horizontal?) is fine here to explain the origins.

miina: Updated within 0eb23d6, let me know if it's more understandable now.

miina: Note that the selector is updated as well now, it is actually easier this way, fewer calculations with nested paddings :)

swissspidy: Much clearer now indeed, thanks!

PR #3586, opened 2019-10-22
Convert .focus class selectors to :focus-within pseudo-class selectors

### LINKED ISSUE #3585: Class selectors for .focus should be automatically converted into :focus-within pseudo class selectors
## Feature description

Many themes include JS logic to toggle a `focus` class on an element when focus is anywhere inside. For example:

* Twenty Seventeen: https://github.com/WordPress/wordpress-develop/blob/67a937a2fd93cc2227aa5472239365b4c7eaf803/src/wp-content/themes/twentyseventeen/assets/js/navigation.js#L68-L107
* Twenty Sixteen: https://github.com/WordPress/wordpress-develop/blob/67a937a2fd93cc2227aa5472239365b4c7eaf803/src/wp-content/themes/twentysixteen/js/functions.js#L73-L109
* Underscores: https://github.com/Automattic/_s/blob/28f0a5b4e56a5dc3cba94749a208eb478fe224d4/js/navigation.js#L48-L104

They then include CSS like Twenty Seventeen has here:

```css
.main-navigation ul ul li:hover > ul,
.main-navigation ul ul li.focus > ul {
	left: 100%;
	right: auto;
}
```

This causes a submenu to appear when focus is anywhere inside of the subtree.

The JS logic to toggle the `focus` class on elements is not feasible in AMP, so the AMP plugin uses a drop-in replacement by patching patching the core themes to use the `:focus-within` pseudo class in selectors that originally used `.focus`, for example:

https://github.com/ampproject/amp-wp/blob/884b805aa1b754383c44a7a909e859dd39427327/includes/sanitizers/class-amp-core-theme-sanitizer.php#L1065-L1073

This is not great because it means there are duplicated CSS rules, resulting in unoptimized CSS such as:

```css
.main-navigation ul ul li:hover > ul {
	left: 100%;
	right: auto;
	/* ... */
}
.main-navigation ul ul li:focus-within > ul {
	left: 100%;
	right: auto;
	/* ... */
}
```

Ideally these should be combined into one style rule because the properties are the same:

```css
.main-navigation ul ul li:hover > ul,
.main-navigation ul ul li:focus-within > ul {
	left: 100%;
	right: auto;
	/* ... */
}
```

Moreover, this is also not good because the process of making any theme AMP compatible would require copying `.focus`-selector style rules to `:focus-within`-selector style rules. (Note that `:focus-within` is supported by [80%+ of browsers](https://caniuse.com/#feat=css-focus-within).)

We can eliminate these painpoints by simply enhancing the CSS sanitizer to automatically replace `.focus` with `:focus-within` when processing a pared stylesheet. There is a natural place for this as well, inside of `\AMP_Style_Sanitizer::ampify_ruleset_selectors()`.

The class name(s) used should be configurable in an arg to the style sanitizer. For example:

```php
	$sanitizers['AMP_Style_Sanitizer']['focus_within_classes'] = [ 'enfocado' ];
	return $sanitizers;
} );
```

This would allow a theme to prevent this replacement behavior entirely by doing:

```php
	$sanitizers['AMP_Style_Sanitizer']['focus_within_classes'] = [];
	return $sanitizers;
} );
```

Alternatively, if a page initially has element(s) with the `focus` class, the replacement would ideally be skipped. But this would seem somewhat rare.

---------------

## Acceptance criteria

* A theme that uses a CSS rule with a selector containing `.focus` should automatically get converted to `:focus-within`.

## Implementation brief

* See above and https://github.com/ampproject/amp-wp/pull/3586

## QA testing instructions

* Ensure that nav submenu items are accessible when just navigating by tabbing with the keyboard.

## Demo

* 

## Changelog entry

* Convert `.focus` class selectors to `:focus-within` pseudo-class selectors

westonruter:
## Summary

* Applies `.focus` => `:focus-within` conversion for all themes, including all themes based on _s.
* Reduces the amount of CSS required by preventing rule duplication for AMP compatibility (reduces Twenty Fourteen's CSS by 119 bytes). This also reduces the maintenance burden.

Fixes #3585.

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -185,7 +185,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	private $used_class_names;
 
 	/**
-	 * Regular expression pattern to match focus_class_names in selectors.
+	 * Regular expression pattern to match focus class names in selectors.
 	 *
 	 * The computed pattern is cached to prevent re-constructing for each processed selector.
 	 *
@@ -2877,7 +2877,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			$before_type_selector_pattern = '(?<=^|\(|\s|>|\+|~|,|})';
 			$after_type_selector_pattern  = '(?=$|[^a-zA-Z0-9_-])';
 
-			// Replace .focus selector with :focus-within.
+			// Replace focus selectors with :focus-within.
 			if ( $this->focus_class_name_selector_pattern ) {
 				$count    = 0;
 				$selector = preg_replace(
@@ -2945,12 +2945,12 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	}
 
 	/**
-	 * Given list of class names, create a regular expression pattern to match them in a selector.
+	 * Given a list of class names, create a regular expression pattern to match them in a selector.
 	 *
 	 * @since 1.4
 	 *
 	 * @param string[] $class_names Class names.
-	 * @return string Pattern,
+	 * @return string Regular expression pattern.
 	 */
 	private static function get_class_name_selector_pattern( $class_names ) {
 		$pattern  = '/\.(';
@@ -2960,7 +2960,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 				function ( $class_name ) {
 					return preg_quote( $class_name, '/' );
 				},
-				$class_names
+				(array) $class_names
 			)
 		);
 		$pattern .= ')(?=$|[^a-zA-Z0-9_-])';
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -88,6 +88,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 *      @type bool     $should_locate_sources      Whether to locate the sources when reporting validation errors.
 	 *      @type string   $parsed_cache_variant       Additional value by which to vary parsed cache.
 	 *      @type string   $include_manifest_comment   Whether to show the manifest HTML comment in the response before the style[amp-custom] element. Can be 'always', 'never', or 'when_excessive'.
+	 *      @type string[] $focus_within_classes       Class names in selectors that should be replaced with :focus-within pseudo classes.
 	 * }
 	 */
 	protected $args;
@@ -108,6 +109,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 		'should_locate_sources'     => false,
 		'parsed_cache_variant'      => null,
 		'include_manifest_comment'  => 'always',
+		'focus_within_classes'      => [ 'focus' ],
 	];
 
 	/**
@@ -181,6 +183,15 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 * @var array
 	 */
 	private $used_class_names;
+
+	/**
+	 * Regular expression pattern to match focus_class_names in selectors.
+	 *
+	 * The computed pattern is cached to prevent re-constructing for each processed selector.
+	 *
+	 * @var string|null
+	 */
+	private $focus_class_name_selector_pattern;
 
 	/**
 	 * Attributes used in the document.
@@ -773,6 +784,12 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 		if ( ! empty( $this->args['allow_dirty_styles'] ) ) {
 			return;
 		}
+
+		$this->focus_class_name_selector_pattern = (
+			! empty( $this->args['focus_within_classes'] ) ?
+				self::get_class_name_selector_pattern( $this->args['focus_within_classes'] ) :
+				null
+		);
 
 		$this->head = $this->dom->getElementsByTagName( 'head' )->item( 0 );
 		if ( ! $this->head ) {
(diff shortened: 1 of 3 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -182,6 +184,15 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 */
 	private $used_class_names;
 
+	/**
+	 * Regular expression pattern to match focus_class_names in selectors.
```
schlessera:
```suggestion
	 * Regular expression pattern to match focus class names in selectors.
```

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -2860,6 +2877,21 @@ static function ( $selector_language ) {
 			$before_type_selector_pattern = '(?<=^|\(|\s|>|\+|~|,|})';
 			$after_type_selector_pattern  = '(?=$|[^a-zA-Z0-9_-])';
 
+			// Replace .focus selector with :focus-within.
```
schlessera:
```suggestion
			// Replace focus selectors with :focus-within.
```

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -2912,6 +2944,30 @@ static function ( $selector_language ) {
 		}
 	}
 
+	/**
+	 * Given list of class names, create a regular expression pattern to match them in a selector.
```
schlessera:
```suggestion
	 * Given a list of class names, create a regular expression pattern to match them in a selector.
```

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -2912,6 +2944,30 @@ static function ( $selector_language ) {
...
+	 *
+	 * @since 1.4
+	 *
+	 * @param string[] $class_names Class names.
+	 * @return string Pattern,
```
schlessera:
```suggestion
	 * @return string Regular expression pattern.
```

tests/php/test-amp-style-sanitizer.php
```diff
@@ -700,12 +700,19 @@ public function get_amp_selector_data() {
 				'.video{color:blue;} audio.audio{color:purple;} .iframe{color:black;} .img{color:purple;} .form:not(form){color:green;}',
 				'.video{color:blue}amp-audio.audio{color:purple}.iframe{color:black}.img{color:purple}.form:not(form){color:green}',
 			],
+			'focus_within_classes' => [
```
schlessera: Test case has both positive (replace `focus`) and negative (don't replace `focused`) cases. 👍

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -182,6 +184,15 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
...
 	private $used_class_names;
 
+	/**
+	 * Regular expression pattern to match focus_class_names in selectors.
+	 *
+	 * The computed pattern is cached to prevent re-constructing for each processed selector.
+	 *
+	 * @var string|null
+	 */
+	private $focus_class_name_selector_pattern;
```
schlessera: I would generally avoid properties for caching purposes only and rather use a static variable within the method that caches. Turning this into a property seems to give it more importance than it actually has, as it seems to be something "actionable".

But that's mostly a matter of preference, this is fine as-is as well.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -2912,6 +2944,30 @@ static function ( $selector_language ) {
...
+	 */
+	private static function get_class_name_selector_pattern( $class_names ) {
+		$pattern  = '/\.(';
+		$pattern .= implode(
+			'|',
+			array_map(
+				function ( $class_name ) {
+					return preg_quote( $class_name, '/' );
+				},
+				$class_names
```
schlessera: We have no guarantee here we'll receive an array, so we should be defensive:
```suggestion
				(array) $class_names
```

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -2912,6 +2944,30 @@ static function ( $selector_language ) {
...
+	/**
+	 * Given list of class names, create a regular expression pattern to match them in a selector.
+	 *
+	 * @since 1.4
+	 *
+	 * @param string[] $class_names Class names.
+	 * @return string Pattern,
+	 */
+	private static function get_class_name_selector_pattern( $class_names ) {
+		$pattern  = '/\.(';
```
schlessera: Although it is pretty irrelevant at the scale here, building a pattern in this way is very wasteful, with multiple assignments and memory concatenations.

I think the following structure would make it both more efficient and more readable:
```php
$class_pattern = implode(
	'|',
	array_map(
		static function ( $class_name ) {
			return preg_quote( $class_name, '/' );
		},
		$class_names
	)
);
return "/\.({$class_pattern})(?=$|[^a-zA-Z0-9_-])/";

schlessera:
I went through the stylesheets of the core themes, and afaict, the replacement should correctly target all needed changes with the above code.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: It's a good point, and it's something I debated, but the reason why I went with this approach is that it is _possible_ that the `focus_within_classes` could be changed for the instance of `AMP_Style_Sanitizer` after the initial `sanitize` call. By using a class member variable we can reset it each time the `sanitize` method is called.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: Done in 8cc8e81e. If you like, you can also add an additional CLA consent 😄

PR #3593, opened 2019-10-22
Fix page crashing when duplicating immovable blocks

### LINKED ISSUE #3467: Duplicating CTA / Page Attachment block crashes browser
## Bug Description

In the stories editor, there when a block is a selected, a context menu icon appears in the toolbar that looks like this. 

Screenshot from 2019-10-08 16-42-39

If a CTA block is selected and the duplicate block is selected, the browser becomes unresponsive and sometimes crashes. 

There might be some infinite loop, perhaps caused by some faulty React hook somewhere.

## Expected Behaviour

Block should fail to duplicate or at least a warning about duplicate CTA blocks is shown.

## Steps to reproduce
1. Create new story in editor
1. Create a second page
1. Insert CTA block.
1. Select CTA block. 
1. Click three dots in top toolbar. 
1. Select duplicate block. 

## Screenshots
Screenshot from 2019-10-08 16-56-13

If the browser doesn't crash, then two CTA blocks are shown. 

## Additional context

 - WordPress version: 5.3
 - Plugin version: 1.3
 - AMP plugin template mode:
 - PHP version:
 - OS: Ubuntu 18.04
 - Browser: Chrome 77.0.3865.75 (Official Build) (64-bit)
 - Device: Dell XPS 13

---------------

## Acceptance criteria

* AC1: Duplicating a CTA block should not cause the browser to lag / crash
* AC2: Duplicating a Page Attachment block should not cause the browser to lag / crash

### Reason for crashing ###
When either a CTA or a Page Attachment block is present on a Page, enforcing the block to be the last block on the Page is triggered. If there are more than one CTA / Page Attachment blocks then this causes the logic to be triggered infinitely since only one block can be the last block:
https://github.com/ampproject/amp-wp/blob/744b640f069c43bca209ad2bf988b105e345a8d4/assets/src/stories-editor/blocks/amp-story-page/edit.js#L145-L166

### Suggested solution ###
If more than one CTA/Page Attachment blocks are detected then before enforcing the block to be the last, two things should happen:
- The extra block(s) should be removed.
- A snackbar notice is displayed to let know about the block being removed.

We could add an additional condition to the same `useEffect` hook that enforces the CTA/Attachment block to be the last, to only run if there is only one CTA or one Page Attachment block. Additionally, we can add another `useEffect` which removes the extra block(s) if detected which can also display a snackbar notice saying that the block was removed since each Page can have only 1 CTA block or 1 Attachment block.

## QA testing instructions

1. Create a new story
1. Add a new page
1. Add a CTA block
1. In the block toolbar, click on the three dots and then on "Duplicate"
    -> Verify that block is not duplicated, but a warning message appears in the bottom left.
1. Add a new page
1. Add a Page Attachment block
1. In the block toolbar, click on the three dots and then on "Duplicate"
    -> Verify that block is not duplicated, but a warning message appears in the bottom left.

## Demo

https://www.youtube.com/watch?v=iItTJkcF7Qg&feature=youtu.be

## Changelog entry

* Prevent duplicating CTA or Page Attachment blocks.

miina:
## Summary

Fixes #3467 
Fixes Page Crashing when duplicating immovable blocks.

## Checklist

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -153,9 +153,10 @@ const PageEdit = ( {
 			const blocksToRemove = immovableBlocks.slice( 0, -1 ).map( ( { clientId: blockId } ) => blockId );
 			removeBlocks( blocksToRemove );
 			const removeMessage = sprintf(
+				/* translators: %d: number of removed blocks. */
 				_n(
-					'%d block removed. Only one CTA/Attachment block allowed per Page.',
-					'%d blocks removed. Only one CTA/Attachment block allowed per Page.',
+					'%d block removed. Only one block of this type is allowed per page.',
+					'%d blocks removed. Only one block of this type is allowed per page.',
 					blocksToRemove.length,
 					'amp'
 				),
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -181,11 +181,9 @@ const PageEdit = ( {
 		}
 		const blockToMove = immovableBlocks[ 0 ];
 
-		if ( blockToMove ) {
-			// If the either CTA or Attachment is not the last block, move it there.
-			if ( childrenOrder[ childrenOrder.length - 1 ] !== blockToMove.clientId ) {
-				moveBlockToPosition( blockToMove.clientId, clientId, clientId, childrenOrder.length - 1 );
-			}
+		// If the either CTA or Attachment is not the last block, move it there.
+		if ( childrenOrder[ childrenOrder.length - 1 ] !== blockToMove.clientId ) {
+			moveBlockToPosition( blockToMove.clientId, clientId, clientId, childrenOrder.length - 1 );
 		}
 	}, [ childrenOrder, clientId, moveBlockToPosition, immovableBlocks ] );
 
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -176,19 +176,10 @@ const PageEdit = ( {
 			return;
 		}
 		// If there is an illegal number of immovable blocks (more than 1).
-		if ( immovableBlocks.length > 1 ) {
+		if ( 1 !== immovableBlocks.length ) {
 			return;
 		}
-		const ctaBlock = getCallToActionBlock( clientId );
-		const attachmentBlock = getPageAttachmentBlock( clientId );
-
-		let blockToMove = null;
-
-		if ( ctaBlock ) {
-			blockToMove = ctaBlock;
-		} else if ( attachmentBlock ) {
-			blockToMove = attachmentBlock;
-		}
+		const blockToMove = immovableBlocks[ 0 ];
 
 		if ( blockToMove ) {
 			// If the either CTA or Attachment is not the last block, move it there.
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -150,8 +150,7 @@ const PageEdit = ( {
 	// If there is more than one immovable block, leave only the last and remove the others.
 	useEffect( () => {
 		if ( immovableBlocks.length > 1 ) {
-			immovableBlocks.pop();
-			const blocksToRemove = immovableBlocks.slice( 0, -1 ).map( ( { clientId } ) => clientId );
+			const blocksToRemove = immovableBlocks.slice( 0, -1 ).map( ( { clientId: blockId } ) => blockId );
 			removeBlocks( blocksToRemove );
 			const removeMessage = sprintf(
 				_n(
--- a/assets/src/stories-editor/blocks/amp-story-page/edit.js
+++ b/assets/src/stories-editor/blocks/amp-story-page/edit.js
@@ -86,14 +86,15 @@ const PageEdit = ( {
 		const { getEditedPostAttribute } = select( 'core/editor' );
 		const postMeta = getEditedPostAttribute( 'meta' ) || {};
 		const { storySettings } = getSettings();
-		const innerBlocks = getBlocksByClientId( getBlockOrder( clientId ) );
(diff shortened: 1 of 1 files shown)


assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -142,10 +149,30 @@ const PageEdit = ( {
...
 
+	// If there is more than one immovable block, leave only the last and remove the others.
+	useEffect( () => {
+		const immovableBlocks = getImmovableBlocks( clientId );
+		if ( immovableBlocks.length > 1 ) {
+			immovableBlocks.pop();
+			const blocksToRemove = immovableBlocks.map( ( { clientId: blockId } ) => blockId );
+			removeBlocks( blocksToRemove );
+			createErrorNotice(
+				__( 'Block removed. Only one CTA/Attachment block allowed per Page.', 'amp' ),
```
barklund: Add count here - technically there could have been more than 1 block removed. Not probable under current circumstances, but to be safe.

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -88,6 +91,10 @@ const PageEdit = ( {
 			media: mediaObject,
 			videoFeaturedImage: videoThumbnail,
 			getBlockOrder,
+			getImmovableBlocks: ( pageClientId ) => {
```
barklund: You only ever invoke this with `clientId` - why not actually extract the list here and store it in a `immovableBlocks` variable? Or is there a need for a function here? It also makes the dependencies of the subsequent `useEffect`'s more clear.

However, do remember to add `clientId` to the dependencies of this `useSelect` (it's already used here, so it's a mistake it hasn't been done already).

@swissspidy, do you know if we can add `useSelect` to the automatic eslint-based dependency-check? There's probably an option to add custom hooks to the validator? This hook is probably missing more than just `clientId`.

assets/src/stories-editor/blocks/amp-story-page/edit.js
spacedmonkey: We know which block was removed here. Just say CTA or attachment. Much clearer

assets/src/stories-editor/blocks/amp-story-page/edit.js
miina: Changed in 0197f85.

assets/src/stories-editor/blocks/amp-story-page/edit.js
miina: Count added in 58b22d6.

assets/src/stories-editor/blocks/amp-story-page/edit.js
swissspidy: @barklund Good question! I always thought it's either a limitation in the lint rule or a bug in the `useSelect` implementation. Looking at [the code](https://github.com/facebook/react/blob/9e64bf18e11828d6b4c0363bff5ed2eca1ccd838/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L28-L35), it seems like we can just pass a regex/list of additional hooks to check against. Will whip up a quick PR to test this!

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -142,10 +146,39 @@ const PageEdit = ( {
 		}
 	}, [ mediaType, mediaUrl ] );
 
+	// If there is more than one immovable block, leave only the last and remove the others.
+	useEffect( () => {
+		if ( immovableBlocks.length > 1 ) {
+			immovableBlocks.pop();
```
barklund: I'm not sure mutating it is such a great idea given that it's used as a dependency for this and other hooks. I think all hook dependencies ought to be kept immutable.

Rather, just `slice` it in the next line:

```js
// keep first
const blocksToRemove = immovableBlocks.slice( 1 ).map( ( { clientId: blockId } ) => blockId );
// keep last
const blocksToRemove = immovableBlocks.slice( 0, -1 ).map( ( { clientId: blockId } ) => blockId );
```

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -142,10 +146,39 @@ const PageEdit = ( {
...
+			const blocksToRemove = immovableBlocks.map( ( { clientId: blockId } ) => blockId );
```
barklund: There's no need to rename the variable here at all:

```suggestion
			const blocksToRemove = immovableBlocks.slice( 0, -1 ).map( ( { clientId } ) => clientId );
```

assets/src/stories-editor/blocks/amp-story-page/edit.js
```diff
@@ -142,10 +146,39 @@ const PageEdit = ( {
...
+
 	useEffect( () => {
 		if ( childrenOrder.length <= 1 ) {
 			return;
 		}
+		// If there is an illegal number of immovable blocks (more than 1).
+		if ( immovableBlocks.length > 1 ) {
+			return;
+		}
 		const ctaBlock = getCallToActionBlock( clientId );
```
barklund: Couldn't all of this logic ([lines 178-193](https://github.com/ampproject/amp-wp/pull/3593/files#diff-3ebaa9cdbb7e5502ab37969c8f31350eR182-R191)) be reduced to:

```js
if (immovableBlocks.length === 1) {
  const blockToMove = immovableBlocks[0];
  // continue check for order and do move
}
```

assets/src/stories-editor/blocks/amp-story-page/edit.js
miina: Fixing in a minute.

assets/src/stories-editor/blocks/amp-story-page/edit.js
miina: Good point! Should be. _Testing_

miina: We could actually use 
```
if ( immovableBlocks.length !== 1 ) {
			return;
		}
```
above, then we could just do `const blockToMove = immovableBlocks[0];`.

miina: Changed in 63850c9.

PR #3597, opened 2019-10-22
Ensure useSelect is tested by react-hooks/exhaustive-deps lint rule

swissspidy:
## Summary

Mentioned while working on #3467.

## Checklist

### DIFF
--- a/assets/src/stories-editor/components/block-preview-label.js
+++ b/assets/src/stories-editor/components/block-preview-label.js
@@ -9,17 +9,20 @@ import PropTypes from 'prop-types';
 import { getBlockType } from '@wordpress/blocks';
 import { BlockIcon } from '@wordpress/block-editor';
 import { useSelect } from '@wordpress/data';
+import { useRef } from '@wordpress/element';
 import { __experimentalGetSettings as getDateSettings, dateI18n } from '@wordpress/date';
 import { __ } from '@wordpress/i18n';
 
 const BlockPreviewLabel = ( { block, label, displayIcon = true, alignIcon = 'left', accessibilityText = false } ) => {
+	const blockPreviewLabel = useRef( label );
+
 	const {
 		content,
 		icon,
 	} = useSelect( ( select ) => {
 		if ( ! block ) {
 			return {
-				content: label,
+				content: blockPreviewLabel.current,
 				icon: null,
 			};
 		}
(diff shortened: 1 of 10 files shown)


assets/src/stories-editor/components/block-preview-label.js
```diff
@@ -76,19 +79,19 @@ const BlockPreviewLabel = ( { block, label, displayIcon = true, alignIcon = 'lef
...
 			default:
 				break;
 		}
 
 		return {
-			content: label,
+			content: blockPreviewLabel.current,
 			icon: blockType.icon,
 		};
 	}, [ block ] );
```
barklund: When you use a `ref`, you don't add it as a dependency?

barklund:
There weren't any other issues than those two files? Good discipline! 👍

assets/src/stories-editor/components/block-preview-label.js
swissspidy: I was wondering myself. I changed it to `useRef` because the lint rule complained about modifying `label` in the hook, recommending to use `useRef` instead.

https://github.com/facebook/react/issues/14387 seems to have an answer, especially https://github.com/facebook/react/issues/14387#issuecomment-455415008 and https://github.com/facebook/react/issues/14387#issuecomment-493677168

barklund: Ah, looking at the code in more detail, is just a local variable in the function. The first reference (the early abort) should just be `label` whereas the others should be to a local `let newLabel = label` that might change during the function. And then still add it to the dependencies.

swissspidy:
I was a bit surprised tbh, but yeah those were the only reported issues. 😅 

In many places we use `useSelect` in trivial ways, e.g. `const currentPage = useSelect( ( select ) => select( 'amp/story' ).getCurrentPage() );`

In this particular example, the lint rule does not complain about a missing `[]`. But I don't think that would make a difference.

-----

While writing this:

I noticed that the lint rule does not complain when no dependencies are passed. See b635d33be4fff780d6d206a820669d20d2f2d8ab

barklund:
> I noticed that the lint rule does not complain when no dependencies are passed.

For `useEffect`, passing no dependencies is the same as passing all props as dependencies, as it means the effect will re-run on every render.

I don't know if `useSelect` works in the same way though.

assets/src/stories-editor/components/block-preview-label.js
swissspidy: Makes sense 👍

PR #3627, opened 2019-10-26
Add AMP-to-AMP navigation

### LINKED ISSUE #1389: Automatically append ?amp to URLs in Paired mode (AMP-to-AMP navigation)
Hey again,

Wanted to say thanks for the awesome work you put into this project and that the v1 is looking better and better every day :clap: 

Now for my issue :
I want the entirety of my Wordpress to be available as AMP.
I have activated the plugin in Paired mode and started transitioning my templates with is_amp_endpoint()
The **problem I am having** is that any links that get rendered on an AMP page will point to the non-amp version of the page.

I have beta 2 and activated automatic stripping of non valid markup.
I have tested accessing said pages (the first one and the one that is linked) directly with `?amp` and all went well.

I was planning on maybe checking the query parameters inside a hook and append ?amp to every link that gets rendered for that call.
I am pretty sure however that there might be a better way to do this.

I might also have a problem on my end or a bad understanding of how things work as I'd expect that this would be a feature that would be enabled by default for an AMP plugin. I tried searing in the Wiki and the Issues but could not find anything relating to this apart from trying to manually enable paired mode and setting a template folder in my `functions.php` but that did not help.

If there is a more appropriate place to ask question like those so that I don't end up spamming the issues I'd be glad to hear about it ! 
Cheers

schlessera:
## Summary

This PR adds a new `AMP_Link_Sanitizer` that redirects internal to their AMP variants in Transitional mode.

The code is adapted from https://gist.github.com/westonruter/f9ee9ea717d52471bae092879e3d52b0 and has added tests to complete the integration.

Fixes #1389

## Checklist


includes/amp-helper-functions.php
```diff
@@ -917,6 +917,10 @@ function amp_get_content_sanitizers( $post = null ) {
 		'AMP_Style_Sanitizer'             => [
 			'include_manifest_comment' => ( defined( 'WP_DEBUG' ) && WP_DEBUG ) ? 'always' : 'when_excessive',
 		],
+		'AMP_Link_Sanitizer'              => [
+			'add_query_vars'    => ! amp_is_canonical(),
```
westonruter: We'll want to only enable AMP-to-AMP by default in Transitional mode, since in Reader mode the AMP version is not full parity with the non-AMP version. So by default users should navigate to non-AMP pages when leaving Reader mode templates.

```suggestion
			'add_query_vars'    => AMP_Theme_Support::TRANSITIONAL_MODE_SLUG === AMP_Theme_Support::get_support_mode(),
```

includes/sanitizers/class-amp-link-sanitizer.php
```diff
@@ -0,0 +1,193 @@
...
+	const DEFAULT_A2A_META_CONTENT = 'AMP-Redirect-To; AMP.navigateTo';
+
+	/**
+	 * Placeholder for default args, to be set in child classes.
+	 *
+	 * @var array
+	 */
+	protected $DEFAULT_ARGS = [ // phpcs:ignore WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase
+		'add_amphtml_rel'   => true,
+		'add_query_vars'    => true, // Set to false when in native mode. Overridden in \AMP_To_AMP\filter_content_sanitizers().
```
westonruter:
```suggestion
		'add_query_vars'    => false, // Only set to true when in Transitional mode.
```

schlessera:
@westonruter All that is missing is to test the filter I added in https://github.com/ampproject/amp-wp/pull/3627/commits/49e56068beb2a603358e64e8120bdca67ec6ecab. However, I'm unsure how best to test this meaningfully. Any pointers to existing tests that already have such a case?

schlessera:
Image 2019-10-28 at 8 26 44 PM

This is what I have right now, but it doesn't do anything, I don't trigger the filter. Probably missing something obvious here...

westonruter:
@schlessera Since the filter is added in the `amp_get_content_sanitizers()` function, don't you just have to call the function to make sure that `add_query_vars` flag has been set to `true`?

Ultimately, this flag could also be set via the `amp_content_sanitizers` filter as well.

schlessera:
> Since the filter is added in the amp_get_content_sanitizers() function, don't you just have to call the function to make sure that add_query_vars flag has been set to true?

Yes, I'll probably fall back to that. I wanted to test the effects in the HTML, but the other tests already do that, so I guess it would be fine.

includes/sanitizers/class-amp-link-sanitizer.php
```diff
@@ -0,0 +1,202 @@
...
+	 */
+	public function process_links() {
+		/**
+		 * Element.
+		 *
+		 * @var DOMElement $element
+		 */
+		$xpath = new DOMXPath( $this->dom );
+
+		// Remove admin bar from DOM to prevent mutating it.
```
schlessera: ~Making two mutations here seems wasteful. Wouldn't it be preferable to adapt the XPath queries to skip ancestors of the admin bar instead.~
Some quick googling lets me assume that query would be too expensive.

PR #3630, opened 2019-10-26
Rebrand validation errors

### LINKED ISSUE #2023: Enhance Compatibility debugging workflow narrative

westonruter:
## Summary

Fixes #2023.

* Rebrand “accepting” the sanitization the invalid markup causing a validation error to rather be “removing” the invalid markup. Similarly, rebrand “reject” as “keep”. As such, “Status” generally now becomes “Markup Status”
* Use different colors of AMP logo as icons for the various states, as opposed to the many different icons previously used (e.g. ✅ and ❌).
* Improve descriptions of stylesheet error codes to indicate they relate to CSS.

### Validated URLs List

* Include whether AMP is enabled or disabled in the list.
* Combine the counts of new errors into the remove/keep counts.
* Rename “Invalid” column to “Invalid Markup”
* Remove red color from Invalid Markup column.
* Improve sources summary (see section below).
* Update filter dropdowns to refer to actions taken on invalid markup rather than validation errors sanitization statuses.

Before | After
-------|------
amp validated urls before | amp validated urls after

### Validation Error Index

* Per above, use “Removed” instead of “Accepted” and “Kept” instead of “Rejected”.
* Per above, use only red and green AMP logos to indicate whether the invalid markup is kept or removed, respectively.

Before | After
-------|------
validation error index before | validation error index after

### Validation Error Detail

* Use green/red buttons to correspond to the removed/kept actions.
* Include “confirm” in the action buttons/links for remove/keep when in the new state, as removing already-removed invalid markup just clears the new state.

Before | After
-------|------
validation error before | validation error after

### Validated URL with Kept Markup (Rejected Error)

* Re-work the page heading to just say “AMP Validated URL”, with subheading for the actual title of the queried object (if available).
* Use same summary from validated URL list in the status metabox.

Before | After
-------|------
validated url kept error before | validated url kept error after

### Validated URL with Errors in all 4 states

* As above, re-work the page heading to just say “AMP Validated URL”, with subheading for the actual title of the queried object (if available).
* Eliminate separate dropdown options for the “New” states, keeping the choices to just: Removed and Kept.
* As above, merge new error counts into the total counts for invalid markup removed/kept.

Before | After
-------|------
validated url all error states before | validated url all error states after

### Improvement to sources summary

Previously when there was a validation error occurring in the content, the summary would show:

Screen Shot 2019-10-26 at 09 31 18

Now, however, a translatable “Content” label is used:

Screen Shot 2019-10-26 at 09 34 14

And, more importantly, when there is more specific information available, like the block name, it is used instead:

Screen Shot 2019-10-26 at 09 31 36

Additionally, if Gutenberg is one of the sources, it will be suppressed since it should be considered the same as core.

## Checklist


includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -1647,28 +1650,34 @@ public static function filter_tag_row_actions( $actions, WP_Term $tag ) {
...
+
+				if ( self::VALIDATION_ERROR_ACK_ACCEPTED_STATUS !== $sanitization['status'] ) {
+					$actions[ self::VALIDATION_ERROR_ACCEPT_ACTION ] = sprintf(
+						'<a href="%s">%s</a>',
+						wp_nonce_url(
+							add_query_arg( array_merge( [ 'action' => self::VALIDATION_ERROR_ACCEPT_ACTION ], compact( 'term_id' ) ) ),
+							self::VALIDATION_ERROR_ACCEPT_ACTION
+						),
+						esc_html(
+							self::VALIDATION_ERROR_NEW_ACCEPTED_STATUS === $sanitization['term_status'] ? __( 'Confirm remove', 'amp' ) : __( 'Remove', 'amp' )
```
swissspidy: Should this be `Confirm Removal`? Feels more 'English' to me. But not a native speaker.

westonruter:
### Updates on Validated URL status metabox

An explanation is added when there are AMP is disabled:

image

And when AMP is enabled while there is invalid markup being removed, another message is displayed:

image

When there is no invalid markup on the URL, then no message is required:

image

### AMP Validated URLs list

The “AMP enabled” and “AMP disabled” lines are removed from the validated URLs list, and “Status” column is changed to “Markup Status” to match the other screens:

image

The icons are reduced in size.

includes/validation/class-amp-validation-error-taxonomy.php
westonruter: That is my gut sense as well. Nevertheless, I can't think of a similar word for “keep”, so that's why I left it as a bare verb. “Confirm keepal” doesn't work.

westonruter: Maybe “Confirm removed” and “Confirm kept” would be better.

amedina:
These changes are superb as they make the functionality of the plugin much more accessible, making it easier for developers to follow an AMP debugging workflow fully integrated with WordPress content generation mechanisms.

includes/validation/class-amp-validation-error-taxonomy.php
westonruter: Yes, I think so, because “confirm removed” has more or a sense that an action has been performed, whereas “confirm removal” has more of a timeless sense. And using the past participle means we can use “confirm kept”, and be consistent with other places we use the past participles in the markup status dropdowns.

westonruter: Done in 803d00f0

PR #3633, opened 2019-10-27
Fix context menu for RTL users.

### LINKED ISSUE #3607: Context menu not accessible in RTL mode
## Bug Description

I'm splitting this out from #3012 as it seems to be a big enough task on its own.

It looks like our custom context menu (accessible via right click or keyboard shortcut) is not working when using the editor in RTL mode.

## Expected Behaviour

The context menu should be accessible just like as in LTR mode.

## Steps to reproduce

1. Install [RTL Tester](https://wordpress.org/plugins/rtl-tester/) plugin and click on "Switch to RTL" in the admin bar. _Or_, alternatively, switch user language to an RTL language like Arabic.
2. Open stories editor
3. Try to access context menu

## Additional context

 - WordPress version: 5.4
 - Plugin version: `develop`
 - Gutenberg plugin version (if applicable): 6.7
 - AMP plugin template mode:
 - PHP version:
 - OS:
 - Browser: Chrome on macOS

---------------

## Acceptance criteria

* The context menu should be accessible in RTL mode just like as in LTR mode.

## Implementation brief

* Check if RTL mode is set and use right offset over left offset to workout position of position of popover. 
* Also make sure RTL css works correctly.

## QA testing instructions
1. Set mode to RTL 
1. Create a new story.
1. Create new text block. 
1. Right click on text block. 
1. Verify context menu appeareance 

## Demo

https://youtu.be/4HEK_gFI1Zs

## Changelog entry

* Improved context menu RTL support

spacedmonkey:
## Summary

Use right offset over left offset if RTL. 

Fixes #3607

## Checklist

### DIFF
--- a/assets/src/stories-editor/components/context-menu/index.js
+++ b/assets/src/stories-editor/components/context-menu/index.js
@@ -201,6 +201,10 @@ const ContextMenu = ( props ) => {
 		left: clientX,
 	};
 
+	if ( isRTL ) {
+		POPOVER_PROPS.position = 'bottom right';
+	}
+
 	let blockActions = [];
 	const pageList = getBlockOrder();
 	const numPages = pageList.length;
--- a/assets/src/stories-editor/components/context-menu/index.js
+++ b/assets/src/stories-editor/components/context-menu/index.js
@@ -200,7 +200,7 @@ const ContextMenu = ( props ) => {
 		left: clientX,
 	};
 
-	const POPOVER_POSITION = ( isRTL ) ? 'bottom right' : 'bottom left';
+	const popoverPosition = ( isRTL ) ? 'bottom right' : 'bottom left';
 
 	let blockActions = [];
 	const pageList = getBlockOrder();
@@ -329,7 +329,7 @@ const ContextMenu = ( props ) => {
 			{ isOpen && (
 				<Popover
 					className={ POPOVER_PROPS.className }
-					position={ POPOVER_POSITION }
+					position={ popoverPosition }
 					onClose={ onClose }
 					focusOnMount="firstElement"
 				>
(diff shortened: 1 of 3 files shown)


assets/src/stories-editor/components/context-menu/index.js
```diff
@@ -201,6 +201,10 @@ const ContextMenu = ( props ) => {
 		left: clientX,
 	};
 
+	if ( isRTL ) {
+		POPOVER_PROPS.position = 'bottom right';
```
barklund: This is directly mutating the "constant", which is definitely not recommended, as it will change the value of something that looks to be constant. Sure, RTL won't suddenly change in the middle of the app, but still, it's bad practice.

Either, create two constants (re-using the common prop) and select which to use based on `isRTL` as:

```js
const popoverProps = isRTL ? POPOVER_PROPS_RTL : POPOVER_PROPS_LTR;
```

Or keep only the `className` as a constant:

```js
const popoverProps = {
	position: isRTL ? 'bottom right' : 'bottom left',
	className = POPOVER_CLASSNAME,
};
```

Maybe the latter is the purer option, but any like this works for me.

spacedmonkey: @barklund What do you think of https://github.com/ampproject/amp-wp/pull/3633/commits/e960c73138f6abd823a0736e62abdb224e79efa9 ?

swissspidy:
Glad to see that it was merely a position problem, and not an issue with right click not working or anything.

spacedmonkey:
Honestly this PR took me ages to work out. So happy it was something simple.

assets/src/stories-editor/components/context-menu/index.js
```diff
@@ -201,6 +200,8 @@ const ContextMenu = ( props ) => {
 		left: clientX,
 	};
 
+	const POPOVER_POSITION = ( isRTL ) ? 'bottom right' : 'bottom left';
```
barklund: It really shouldn't be `SCREAMING` if it's not a constant. And it's just a local variable.

```suggestion
	const popoverPosition = ( isRTL ) ? 'bottom right' : 'bottom left';
```

spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3633/commits/4e5eadff03f9486e5ad8c91f54c679dbe7687ba2

PR #3645, opened 2019-10-28
Remove Alignment button.

### LINKED ISSUE #3504: Image alignment not functional
## Bug Description

Image alignment is once again possible in Story editor (probably due to changes upstream) and is once again broken.

If you try to align the image using any of the options, nothing useful happens.

This has been fixed before in #2115.

## Expected Behaviour

No image alignment option should be displayed.

## Steps to reproduce

1. Create a new story
2. Add and select an image.
3. Click the image alignment dropdown in top bar.
4. See useless options.

## Screenshots

Screenshot 2019-10-14 at 18 03 56

## Additional context

 - WordPress version:
 - Plugin version: **develop**
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* No alignment controls should be visible for image block
* No alignment controls should be visible for video block
* No alignment controls should be visible for pullquote block
* No alignment controls should be visible for embed block

## Implementation brief

* Update the CSS selectors/add new selectors to ensure that the alignment toolbar is hidden.
* The current selectors are here:
https://github.com/ampproject/amp-wp/blob/468b3492fe8611a13fcbebee8cbd62966ac71376/assets/src/stories-editor/blocks/amp-story-page/edit.css#L217-L226

## QA testing instructions

* Create a new story
* Create a new page
* Create Image / table / pullquote / embed / video block. 
* Verify block options do not show. 

## Demo

Screenshot from 2019-10-29 13-02-54
Screenshot from 2019-10-29 13-02-27

## Changelog entry

* Fixed an issue with disabled block alignment support

spacedmonkey:
## Summary

Fixes #3504

## Checklist


assets/src/stories-editor/blocks/amp-story-page/edit.css
```diff
@@ -218,10 +218,7 @@ div[data-type="amp/amp-story-page"] .wp-block-image {
  * Hackily hide the alignment options from the image block since they are irrelevant for an image block in a story.
  * In the future the underlying alignment buttons should be removed entirely. See <https://github.com/ampproject/amp-wp/issues/2115>.
  */
-.edit-post-layout[data-block-name="core/image"] .edit-post-header-toolbar__block-toolbar .block-editor-block-toolbar > .components-toolbar:nth-child(2),
-.edit-post-layout[data-block-name="core/pullquote"] .edit-post-header-toolbar__block-toolbar  .block-editor-block-toolbar > .components-toolbar:nth-child(2),
-.edit-post-layout[data-block-name="core/embed"] .edit-post-header-toolbar__block-toolbar .block-editor-block-toolbar > .components-toolbar:nth-child(2),
-.edit-post-layout[data-block-name="core/video"] .edit-post-header-toolbar__block-toolbar .block-editor-block-toolbar > .components-toolbar:nth-child(2) {
+.edit-post-layout[data-block-name="core/image"]  .edit-post-header-toolbar__block-toolbar .block-editor-block-toolbar .components-dropdown-menu.components-toolbar {
```
miina: ^ Doesn't the `align: false` work for image block?

assets/src/stories-editor/helpers/addAMPAttributes.js
```diff
@@ -194,6 +194,7 @@ const addAMPAttributes = ( settings, name ) => {
 		supports: {
 			...settings.supports,
 			anchor: false,
+			align: false,
```
miina: Would this set `align:false` to all the blocks, including `table`, `verse`, etc.?

spacedmonkey: Don't we want to remove alignment for all blocks? If not, I can easily just remove support conditionally.

assets/src/stories-editor/blocks/amp-story-page/edit.css
spacedmonkey: Sadly not. This is an issue with this upstream. https://github.com/WordPress/gutenberg/issues/13780

assets/src/stories-editor/helpers/addAMPAttributes.js
spacedmonkey: Just checked and the table block does support alignments but they don't work. 
If we are going to support alignments then that is a bigger issue that we should look at.

miina: Hmm, based on what I can see then #3504 mentions specific blocks only (Image, pullquote, embed, video). Probably some of the blocks, e.g. Verse, List, possibly Table, should still have the option for alignment. Although now when testing, it looks like the `align:false` doesn't influence those blocks either, it looks like "Verse" still has alignment despite setting it to `false`. I guess that's also the Gutenberg bug? Probably should still make it conditional since when the bug gets fixed this would remove the support for the blocks that need alignment, too.

If the Gutenberg bug would get fixed, would this remove the alignment from the Text block, too with this?

It feels "more correct" to specifically remove the alignment from the blocks that don't need it instead of all. Thoughts?

(Note: I can see the alignment for Table working):
Screenshot 2019-10-28 at 18 38 49

spacedmonkey: @swissspidy Can you confirm what the action should be here? I believe we should not support alignment for any block. In my tests, only right, center and left kind of worked for table, but wide and full did not. We either support property or not at all.

miina: Note that there is also `alignWide: false` option.

Also, in case of Table, the full width does work, e.g. compare left alignment with full width:
Screenshot 2019-10-28 at 19 42 18
Screenshot 2019-10-28 at 19 42 11

miina: Table block has alignment options for the Table generally and separately for the Text within. Probably alignment for the text inside makes sense, for the Table not (since resizing replaces that).

swissspidy: > Table block has alignment options for the Table generally and separately for the Text within. Probably alignment for the text inside makes sense, for the Table not (since resizing replaces that).

That's my thinking as well.

We have draggable and resizable blocks; any alignment options that target the  _block_ and not just its _text_ are not useful and should be disabled.

swissspidy: To answer the question:

> Don't we want to remove alignment for all blocks?

Yes, we do. For all blocks.

To give an example:

Remove block aligments:

Screenshot 2019-10-29 at 11 49 29

Keep text or column alignments:

Screenshot 2019-10-29 at 11 49 24

spacedmonkey: So the PR as it stands is correct. `align: false` remove the block alignment and not the column / text alignment. With alignment disabled `alignWide: true` does nothing.

spacedmonkey:
We can added `alignWide: false` if that helps.

PR #3323, opened 2019-09-23
Add an 'AMP Preview' button to the block editor

kienstra:
### Update: this has changed, please see [more recent comments](https://github.com/ampproject/amp-wp/pull/3323#issuecomment-546745329) below

Like [Pascal mentioned](https://github.com/ampproject/amp-wp/issues/2934#issuecomment-518236586), this is mainly forked from Gutenberg's [PostPreviewButton](https://github.com/WordPress/gutenberg/blob/95e769df1f82f6b0ef587d81af65dd2f48cd1c38/packages/editor/src/components/post-preview-button/index.js#L1).

amp-preview-g

Maybe the preview button, or its location, could use improvement:
amp-preview-button

[Option 2](https://github.com/ampproject/amp-wp/issues/2934#issuecomment-528993707) looks really good, but I didn't see a way to implement that.

For the preview tab, this uses a fork of an AMP icon from amp.dev:

svg-loading-amp

Closes #2934

Issue #2934, opened 2019-08-02
How can I preview AMP pages in Gutenberg?


TwisterMc:
In the classic editor, next to the preview button was an AMP icon which allowed a user to preview the AMP page. In Gutenberg this button doesn't show and there doesn't seem to be a way to preview AMP pages in Gutenberg.

Is there a button someplace that I just don't see? Or is it not available in Gutenberg?

I do know I can just add `&amp=1` to the preview URL, but that's not client friendly. They like buttons. :)

swissspidy:
I don't think we currently add an AMP Preview button in Gutenberg. But I suppose we should :-)

westonruter:
Yeah, the classic editor has a preview button for AMP in Transitional and Reader modes (IIRC). We didn't add it to Gutenberg because... we probably didn't know how.

TwisterMc:
Love the honesty. What about adding a text link below the _Enable AMP_ toggle that says _Preview AMP Post_? It doesn't have to be next to the preview button.

swissspidy:
For comparison, here's the preview button in the classic editor:

![Screenshot 2019-08-05 at 13 00 06](https://user-images.githubusercontent.com/841956/62467881-131a1f80-b795-11e9-9e1f-206b4c63603c.png)

And here is the same section in the block editor:

![Screenshot 2019-08-05 at 13 02 20](https://user-images.githubusercontent.com/841956/62467896-19a89700-b795-11e9-9e54-751ab2c85057.png)

Since the preview button in the block editor could only be modified using direct DOM manipulation or perhaps by overriding `wp.editor.PostPreviewButton`. Both options would be prone to breakage when new versions.

So using the same `PluginPostStatusInfo` slot as we already do for the AMP toggle makes most sense to me.

I think that is a reasonably small enhancement to include in the next release. It might mean duplicating larges parts of the [`PostPreviewButton`](https://github.com/WordPress/gutenberg/blob/1df83a3998d1acbf224f96bbe4320b607747cb45/packages/editor/src/components/post-preview-button/index.js) component though.

swissspidy:
Right, that's my suggested location as well. Needs lots of the logic from `PostPreviewButton` either way though in order to get the correct preview link, saveable status, to open the preview window etc. Otherwise the preview might not be accurate.

This is currently in the backlog, so we will get to it at some point.

westonruter:
[...]

PR #3656, opened 2019-10-28
Add paired browsing interface

### LINKED ISSUE #3365: Add paired browsing interface for Transitional mode to aid with checking parity
Currently when checking a site for AMP compatibility, we have guidance to use Transitional mode and then check the AMP vs non-AMP version to see if there are discrepancies. One way to do that is to have the non-AMP version open in one window and the AMP version in another and browse around the site in both windows to check for parity. There are a couple problems with this:

1. When in Transitional mode, links on an AMP page do not link to the AMP version for URLs of the site. In other words, there is not yet any AMP-to-AMP linking (#1389). This means the user would have to constantly click the admin bar item to go to the AMP version after navigating. That's annoying and the user is prone to forget they aren't looking at the AMP version.
2. Manually navigating in two windows is tedious, as you have to perform every navigation twice, while also keep scrolling both windows to ensure visual consistency all the way down the page.

What if administrators had an interface in Transitional mode which facilitates a side-by-side comparisons? Now that the Admin Bar can be used in AMP pages along with custom scripts (#3187), we can use this to include an AMP admin bar menu item called “Compare with AMP version”. Clicking this link can start a paired browsing session in another window:

Imagine being on the non-AMP version and clicking that admin menu option to open a new window with the AMP version. As you navigate between around the site in the initial non-AMP window, the corresponding slave AMP window could be synchronized, including the current URL and scroll position. That would make it really easy to quickly navigate around the site and compare the AMP and non-AMP version for parity.

Note that `window.resizeTo()` does not work unless you've explicitly opened a window via `window.open()`. Therefore, it wouldn't be possible to programmatically resize the current window. Therefore, achieving this may require using two iframes placed side-by-side on a page. The two windows would each report to the parent window the current scroll position and URL via `postMessage` and the parent window would relay the position and current URL to the other window.

In order to achieve this, we'd need to make sure that when in this mode, the frontend of the site gets `X-Frame-Options: SAMEORIGIN` and `Content-Security-Policy: frame-ancestors 'self'` sent, as the Customizer preview does in `\WP_Customize_Manager::filter_iframe_security_headers()` (and in `send_frame_options_header()`). Perhaps we could just send those headers always when the Admin Bar is shown. We should also consider just going ahead and enabling A2A linking (#1389), at least when the user is logged-in. There could be a filter to override whether A2A is enabled, where it is enabled by default on Transitional mode and disabled by default when in Reader mode.

pierlon:
## Summary

Fixes #3365.

## Todo

- [x] Extract the template for paired browsing into its own file
- [ ] ~Add support for paired browsing while in Standard mode.~
- [x] Add paired browsing button to AMP Validated URL status metabox.
- [x] Provide means of exiting paired browsing mode, perhaps meaning an additional paired browsing bar appearing at top of page.
- [x] Allow paired browsing while admin bar is not showing. This will likely mean persisting certain paired-mode query vars on links and forms so that the required JS in the window will be loaded. Or else, the parent window can instead be polling for access to the iframe window contents.
- [ ] If AMP window initializes without AMP query var, then show AMP not available message, and make iframe hidden. Make sure that that if AMP is disabled for a given post or template that the failure to show AMP version is handled.
- [x] Add a label to each iframe for clear identification

## Checklist


pierlon:
For pages with validation errors, once the URL includes the `amp_validation_errors` query var it prevents any further redirection and alerts the user with the reason why the AMP version can't be viewed:

deepin-screen-recorder_brave-browser_20191028165319

Any suggestions on how this could be improved?

schlessera:
I suggest changing the background colors to something less aggressive and having big centered text in the middle of each iframe saying "non-AMP" and "AMP":

paired-browsing-iframe-backgrounds

assets/src/admin/paired-browsing/app.js
```diff
@@ -0,0 +1,189 @@
+/* global ampSlug, ampPairedBrowsingQueryVar */
```
swissspidy: If deconstructing these variables from `window`, neither such a `global` definition nor the change to the ESLint config should not be needed:

```js
const { ampSlug, ampPairedBrowsingQueryVar, alert, history, parent } = window;
```

.eslintrc
```diff
@@ -1,4 +1,9 @@
 {
+  "globals": {
```
swissspidy: Should not be needed as per my other comment

westonruter:
I've fleshed out the todos based on additional requirements that I hadn't yet added to the corresponding issue.

westonruter:
> * Add support for paired browsing while in Standard mode.

This is worth having a conversation about. The thinking here is that if you are debugging an issue on an AMP-first site (Standard mode), you may want to compare the AMP page with a non-AMP page to see what is going on. Think of, for example, adding a new block that tries to enqueue some JS (like a [Countdown](https://ultimateblocks.com/countdown-block/)). Nevertheless, allowing paired browsing may not have the desired effect. It could be that the theme is not even designed to work without AMP, so trying to load the theme without AMP could result in a broken template. The alternative would be to expose the debug flags (#1294, being worked on in #3448), specifically the debug option to disable AMP. Come to think of it, this would essentially be a corollary to the addition of the “Preview AMP” button (#2934 via #3323): on a Standard-mode site, the Validated URL screen should provide a “View non-AMP” link which would provide that query var to disable AMP for the request. This may have a better result than trying to force paired browsing to work in Standard mode, when there's no guarantee that non-AMP version would even work. /cc @kienstra @amedina

assets/src/admin/paired-browsing/app.js
pierlon: Yep that's a better solution, thanks :100:

pierlon:
> I suggest changing the background colors to something less aggressive and having big centered text in the middle of each iframe saying "non-AMP" and "AMP":
> 
> paired-browsing-iframe-backgrounds

Thinking about this some more: are the background colors necessary? The labels should be sufficient to identify the iframes.

westonruter:
The background colors are not necessary. They were used as an initial way to clearly differentiate the two.

pierlon:
In 964f981, I've added a Paired Browsing navigation bar:

image

pierlon:
> - Provide means of exiting paired browsing mode

I'm not sure what this means. Would it be closing the tab that was opened, or upon exiting it redirects the parent window to the URL from one of the iframes?

I think closing the tab would be the better option here, but there is one major caveat with that approach. It is possible to close the tab once its opened from the "Start paired browsing" link, but not so if the page is accessed directly from that same link

pierlon:
> -  Allow paired browsing while admin bar is not showing.

Could a reason be given for this? I can't think of any scenarios where something like this would occur.

assets/src/admin/paired-browsing/client.js
```diff
@@ -0,0 +1,33 @@
+const { parent } = window;
+
+if ( parent.pairedBrowsingApp ) {
+	window.ampPairedBrowsingClient = true;
+	const app = parent.pairedBrowsingApp;
+
+	app.registerClientWindow( window );
+
+	document.addEventListener( 'DOMContentLoaded', () => {
```
swissspidy: Could use `@wordpress/dom-ready` instead as we do already everywhere else. Example:

```js
/**
 * WordPress dependencies
 */
import domReady from '@wordpress/dom-ready';

domReady( () => {
	// ...
} );
```

It's basically the same, but also checks `document.readyState`

assets/src/admin/paired-browsing/app.js
```diff
@@ -0,0 +1,194 @@
...
+		);
+	}
+
+	/**
+	 * Removes AMP related query variables from the supplied URL.
+	 *
+	 * @param {Location} url URL string.
+	 * @return {string} Modified URL without any AMP related query variables.
+	 */
+	removeAmpQueryVars( url ) {
```
swissspidy: The `@wordpress/url` package contains a bunch of such URL helpers, which could help make things DRY here.

Example:

```js
/**
 * WordPress dependencies
 */
import { removeQueryArgs, addQueryArgs, hasQueryArg } from '@wordpress/url';

// ...

removeAmpQueryVars( url ) {
	return removeQueryArgs( url, ampSlug, 'amp_validation_errors' );
}

addAmpQueryVar( url ) {
	return addQueryArgs(
		url,
		{
			ampSlug: '',
		},
	);
}

addPairedBrowsingQueryVar( url ) {
	return addQueryArgs(
		url,
		{
			ampPairedBrowsingQueryVar: '1',
		},
	);
}

urlHasValidationErrorQueryVar( url ) {
	return hasQueryArg( url, 'amp_validation_errors' );
}
```

webpack.config.js
```diff
@@ -173,9 +173,28 @@ const admin = {
...
+					MiniCssExtractPlugin.loader,
+					'css-loader',
+					'postcss-loader',
+				],
+			},
+		],
 	},
 	plugins: [
 		...defaultConfig.plugins,
+		new MiniCssExtractPlugin( {
```
swissspidy: Let's also use `RtlCssPlugin` for RTL support

templates/amp-paired-browsing.php
```diff
@@ -0,0 +1,54 @@
...
+	</head>
+	<body>
+		<section>
+			<nav id="header">
+				<ul>
+					<li>
+						<img src="<?php echo esc_url( amp_get_asset_url( 'images/amp-white-icon.svg' ) ); ?>" alt="">
+					</li>
+					<li>
+						<span>Paired Browsing</span>
```
swissspidy:
```suggestion
						<span><?php esc_html_e( 'Paired Browsing', 'amp' ); ?></span>
```

includes/class-amp-theme-support.php
```diff
@@ -2463,6 +2470,59 @@ public static function enqueue_assets() {
 		wp_styles()->add_data( 'amp-default', 'rtl', 'replace' );
 	}
 
+	/**
+	 * Enqueue AMP paired browsing client script.
+	 *
+	 * @return void
+	 */
+	public static function enqueue_paired_browsing_client() {
+		wp_enqueue_script(
```
swissspidy: For the version and the list of dependencies we can use something like this:

```php
$asset_file   = AMP__DIR__ . '/assets/js/amp-paired-browsing-client.asset.php';
$asset        = require $asset_file;
$dependencies = $asset['dependencies'];
$version      = $asset['version'];

wp_enqueue_script(
	'amp-paired-browsing-client',
	amp_get_asset_url( '/js/amp-paired-browsing-client.js' ),
	$dependencies,
	$version,
	true
);
```

This way we get the list of dependencies automatically for free, and the script version changes every time there is an actual change to the script (no need to clear caches).

includes/class-amp-theme-support.php
```diff
@@ -2463,6 +2470,59 @@ public static function enqueue_assets() {
...
+		}
+
+		wp_enqueue_style(
+			'amp-paired-browsing-app',
+			amp_get_asset_url( '/css/amp-paired-browsing-app-compiled.css' ),
+			[ 'dashicons' ],
+			AMP__VERSION
+		);
+
+		wp_enqueue_script(
```
swissspidy: Same here with dependencies list & version

swissspidy:
> I'm not sure what this means. Would it be closing the tab that was opened, or upon exiting it redirects the parent window to the URL from one of the iframes?

My understanding would be the latter. If my website is the only tab open and I enter paired browsing, how do I get back to my website? Closing the tab would not be ideal.

> Could a reason be given for this? I can't think of any scenarios where something like this would occur.

If I disable the admin bar from being shown on the frontend (can be done in the user profile), how can I still use paired browsing?

includes/class-amp-theme-support.php
pierlon: Thanks for the review, @swissspidy!

pierlon:
Paired browsing is now accessible from the Validated URL screen:

image

The button only appears if both AMP is enabled for that post and transitional mode is available.

pierlon:
Exiting paired browsing mode is now possible, and will redirect to the current non-AMP page:

deepin-screen-recorder_Select area_20191030171049

templates/amp-paired-browsing.php
```diff
@@ -0,0 +1,57 @@
...
+			<nav id="header">
+				<ul>
+					<li>
+						<img src="<?php echo esc_url( amp_get_asset_url( 'images/amp-white-icon.svg' ) ); ?>" alt="">
+					</li>
+					<li>
+						<span><?php esc_html_e( 'Paired Browsing', 'amp' ); ?></span>
+					</li>
+					<li>
+						<span id="exit-link" class="dashicons-before dashicons-migrate"><?php esc_html_e( 'Exit', 'amp' ); ?></span>
```
swissspidy: This should be a proper `<a>` link. By default it can use `$url` as the `href`.

assets/src/admin/paired-browsing/app.js
```diff
@@ -0,0 +1,207 @@
...
+		this.nonAmpIframe = document.getElementById( 'non-amp' ).firstElementChild;
+		this.ampIframe = document.getElementById( 'amp' ).firstElementChild;
+
+		Promise.all( this.getIframeLoadedPromises() ).then( () => {
+			// Check every second to see if any iframe has disconnected.
+			setInterval( () => this.checkConnectedIframes(), 1000 );
+
+			const exitLink = document.getElementById( 'exit-link' );
+
+			exitLink.addEventListener( 'click', () => {
```
swissspidy: As mentioned below, we should make this accessible by using a proper anchor. Instead of using `window.location.replace` and a click handler we can update the link's `href` value with `this.nonAmpIframe.contentWindow.location.href` if they don't match (e.g. when navigating)

pierlon:
> If I disable the admin bar from being shown on the frontend (can be done in the user profile), how can I still use paired browsing?

Is this what you had in mind for the todo: "Allow paired browsing while admin bar is not showing", @westonruter?

assets/src/admin/paired-browsing/app.js
pierlon: Thanks for the guidance @swissspidy.

westonruter:
> Is this what you had in mind for the todo: "Allow paired browsing while admin bar is not showing"

Yes, but also @schlessera noted that he wanted to be able to do paired browsing while not being logged-in to WordPress at all. So ideally paired browsing would be possible not only when the admin bar is hidden, but also when the user is not logged-in at all.

pierlon:
Ah, OK. That's already been implemented then.

pierlon:
Each iframe now has a label to clear identification:

deepin-screen-recorder_Select area_20191106105214

templates/amp-paired-browsing.php
```diff
@@ -0,0 +1,58 @@
...
+							<?php esc_html_e( 'Exit', 'amp' ); ?>
+						</a>
+					</li>
+				</ul>
+			</nav>
+		</section>
+
+		<div class="container">
+			<div id="non-amp">
+				<iframe src="<?php echo esc_url( $url ); ?>" sandbox="allow-forms allow-scripts allow-same-origin allow-popups"></iframe>
```
swissspidy: Let's add a `title` attribute here to these iframes for accessibility reasons.

e.g. `AMP Version` and `Regular Version`

assets/src/admin/paired-browsing/app.css
```diff
@@ -0,0 +1,90 @@
...
+	text-decoration: none;
+}
+
+#header img {
+	display: block;
+	width: 24px;
+}
+
+#amp::before {
+	content: "AMP";
```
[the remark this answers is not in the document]
schlessera: In case we do want to have the text visible (which I wouldn't recommend), we should at the very least add `pointer-events: none` to make sure they don't hinder interaction with the page.

assets/src/admin/paired-browsing/app.css
```diff
@@ -0,0 +1,90 @@
...
+#header li {
+	margin: 0 5px;
+}
+
+#header li:last-child {
+	margin-left: auto;
+}
+
+#exit-link {
+	text-decoration: none;
```
schlessera: The exit link currently does not show any sign of being clickable, even when hovering over it.
Maybe add the underline back on hover?

schlessera:
I would suggest also looking into adding the two iframes as skip links to improve keyboard traversal.

assets/src/admin/paired-browsing/app.css
pierlon: I thought the cursor becoming a pointer was enough, but yes an underline on hover would make it more understandable.

assets/src/admin/paired-browsing/app.css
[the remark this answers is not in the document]
pierlon: The initial thought when implementing this was that it may not be obvious to the user what each iframe is showing, so adding a distinct label for each should solve that. After reading your review and @swissspidy's however, I realize that this creates more problems than that of the one I intended to fix, so I will revert it for now.

With that said, do you think that the AMP admin bar item in each iframe acts as an adequate label? Maybe I'm overthinking this, so no worries either way :see_no_evil:  .

tests/php/validation/test-class-amp-validation-manager.php
```diff
@@ -314,6 +314,25 @@ public function test_add_admin_bar_menu_items() {
...
 
+		// Admin bar item available in paired mode.
+		add_theme_support( AMP_Theme_Support::SLUG, [ AMP_Theme_Support::PAIRED_FLAG => true ] );
+		$admin_bar = new WP_Admin_Bar();
+		AMP_Validation_Manager::add_admin_bar_menu_items( $admin_bar );
+		$node = $admin_bar->get_node( 'amp' );
+		$this->assertInternalType( 'object', $node );
+		$this->assertInternalType( 'object', $admin_bar->get_node( 'amp-paired-browsing' ) );
+
+		/**
```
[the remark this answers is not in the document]
pierlon: Thanks, missed that one :+1:

pierlon:
@schlessera, @swissspidy could you please consent :relaxed:

pierlon:
@googlebot I consent.

pierlon:
> If AMP window initializes without AMP query var, then show AMP not available message, and make iframe hidden.

Hey @westonruter, could you clarify what you meant by "make iframe hidden"? Currently if the AMP window fails to load the AMP page, it shows an alert with a message as to why and the iframe becomes "disconnected":

deepin-screen-recorder_Select area_20191111211239

Also, regarding the second to-do (adding support for paired browsing while in Standard mode), I agree with the alternative you suggested as outlined in https://github.com/ampproject/amp-wp/pull/3656#issuecomment-547200551, but of course that depends on https://github.com/ampproject/amp-wp/pull/3448 being accepted and merged, so I guess this PR will be on pause until that happens.

westonruter:
> could you clarify what you meant by "make iframe hidden"? Currently if the AMP window fails to load the AMP page, it shows an alert with a message as to why and the iframe becomes "disconnected":

The scenario I'm thinking of is what happens if you have disabled AMP for a given post or AMP is disabled because there are kept validation errors. In such cases, we need to make sure that the user knows the AMP iframe is not showing an AMP page. So making it clear that the AMP iframe is disconnected seems like it would work well. 

There seem to be two kinds of disconnections: one due to an AMP version not being available, and two due to the user clicking a link off to some other domain (or going to the admin). Should the iframe be shown as disconnected in the same way for both scenarios? Perhaps we should not be facilitating browsing in the disconnected state at all. Initially I was thinking this would be nice, for example to allow the user to enter the admin in one of the iframes to fix some content and then to go back to the frontend to then see the change. But this overly complicates things I think. Therefore, I think the disconnected state should actually be a full page overlay that takes over the entire page saying something like, “The navigated URL is not available for paired browsing. Would you like to go back or exit to continue to that URL?” 

However, I do not think an `alert()` should be used, as it is overall poor UX and blocks the main thread. How about this: the disconnected state could cause the blue bar on the top could extend down to cover the entire page and have the message overlaying it in the middle. The bar could even extend down with an animation to make that less jarring. If the initially-loaded URL is not available for paired browsing, then only “Exit” should be offered. Otherwise, a Back link could be provided which just does `history.back()`.

/cc @cathibosco 

> Also, regarding the second to-do (adding support for paired browsing while in Standard mode), I agree with the alternative you suggested as outlined in [#3656 (comment)](https://github.com/ampproject/amp-wp/pull/3656#issuecomment-547200551), but of course that depends on #3448 being accepted and merged, so I guess this PR will be on pause until that happens.

In that case I'd say just proceed with only letting Paired Browsing be available in Transitional mode. We can revisit later if we think it is warranted.

assets/src/admin/paired-browsing/app.js
```diff
@@ -0,0 +1,215 @@
...
+	 * @param {Window} win Document window.
+	 */
+	registerClientWindow( win ) {
+		let oppositeWindow;
+
+		if ( win === this.ampIframe.contentWindow ) {
+			if ( ! this.documentIsAmp( win.document ) ) {
+				if ( this.urlHasValidationErrorQueryVar( win.location.href ) ) {
+					// eslint-disable-next-line no-alert
+					alert( 'The AMP version of this page could not be rendered due to validation errors.' );
```
westonruter: As noted in comment, let's find an alternative to using `alert()`. 

Aside: this needs to be translatable (in `__()`).

assets/src/admin/paired-browsing/client.js
```diff
@@ -0,0 +1,38 @@
...
+				ampViewBrowsingItem.remove();
+			}
+		} else {
+			/**
+			 * Override the entire AMP menu item with just "Non-AMP". There should be no link to
+			 * the AMP version since it is already being shown.
+			 */
+			const ampMenuItem = document.getElementById( 'wp-admin-bar-amp' );
+			if ( ampMenuItem ) {
+				ampMenuItem.innerHTML = 'Non-AMP';
```
westonruter: Will need to be translatable.

templates/amp-paired-browsing.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * AMP Paired Browsing experience template.
+ *
+ * 🚫🚫🚫
+ * DO NOT EDIT THIS FILE WHILE INSIDE THE PLUGIN! Changes You make will be lost when a new version
+ * of the AMP plugin is released.
+ * 🚫🚫🚫
```
westonruter: This comment isn't relevant here because it should not be overridden by any theme.

templates/amp-paired-browsing.php
```diff
@@ -0,0 +1,58 @@
...
+			</div>
+
+			<div id="amp">
+				<iframe src="<?php echo esc_url( $amp_url ); ?>" sandbox="allow-forms allow-scripts allow-same-origin allow-popups" title="<?php esc_attr__( 'AMP version', 'amp' ); ?>"></iframe>
+			</div>
+		</div>
+
+		<?php print_footer_scripts(); ?>
+	</body>
+</html>
```
westonruter: This file should rather be placed in `includes/templates/` as it is not a Reader mode template, and it should not be able to be overridden.

includes/validation/class-amp-validation-manager.php
```diff
@@ -469,6 +469,17 @@ public static function add_admin_bar_menu_items( $wp_admin_bar ) {
...
 		];
 
+		// Construct admin bar item to link to AMP version or non-AMP version.
+		$paired_browsing_item = [
+			'parent' => 'amp',
+			'id'     => 'amp-paired-browsing',
+			'title'  => esc_html__( 'Start paired browsing', 'amp' ),
+			'href'   => remove_query_arg( amp_get_slug(), add_query_arg( AMP_Theme_Support::PAIRED_BROWSING_QUERY_VAR, '1' ) ),
+			'meta'   => [
+				'target' => 'amp-paired-browsing',
```
westonruter: We can remove the `target` now. Originally I added a `target` to open paired browsing in a new window because there was no link to exit. But now there is. So I don't think we need it.

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -1833,6 +1833,21 @@ public static function print_status_meta_box( $post ) {
 						<?php elseif ( ! $is_amp_enabled ) : ?>
 							<?php esc_html_e( 'AMP is disabled because there is invalid markup kept. To unblock AMP from being served, either mark the invalid markup as removed or fix the code that adds the invalid markup.', 'amp' ); ?>
 						<?php endif; ?>
+
+						<?php
+						if ( $is_amp_enabled && AMP_Theme_Support::is_paired_available() ) {
+							?>
+							<div id="paired-browsing-action">
+								<a class="button button-secondary"
+									target="amp-paired-browsing"
```
westonruter: I think the `target` here can be removed, per below.

pierlon:
Does anyone have issues with scrolling in Firefox while in paired browsing mode?

westonruter:
It's working fine for me. What theme? What issues?

pierlon:
The theme doesn't seem to be a factor. As you can see scrolling using the scrollbar is not an issue, but with the mouse or track-pad its a bit choppy:

choppy scroll

westonruter:
A bit of choppiness I think is fine. I don't know what could be done to make it smoother while also ensure the scroll position is kept in sync.

pierlon:
Some noticeable changes have been made since my last commit, which include:

- [Skip links are added to quickly navigate to an iframe](https://github.com/ampproject/amp-wp/commit/1aad73e57f1698ea496d0be75157dd6339720f83):

skip links

- [The exit link is underlined upon a hover or focus](https://github.com/ampproject/amp-wp/commit/8bf8dc420064455c9b9007a0b55bd3c534f510be):

exit-link-hover

- [Labels for the iframes are now on the nav bar](https://github.com/ampproject/amp-wp/commit/2eb69af5eae44235298187082d5754515686d734):

image

Perhaps it could stand out more, but I'm not sure what can be done to accomplish that.

- [An overlay now appears over a disconnected iframe](https://github.com/ampproject/amp-wp/commit/d86ba3f0b31488bc6900e61a4f367e4439bab3ba):

If the navigated URL is some other domain (or going to the admin), the 'Go Back' button returns to the previous page, while the 'Continue' button hides the overlay to allow interaction with the page:

disconnected-overlay-general

However, an AMP version of the page is not available, the 'Continue' button will be hidden and the only options will be to exit paired browsing or to go back to the previous page:

amp-page-error

Note that the 'Go back' button is only shown if the parent window has history.

westonruter:
> Some noticeable changes have been made since my last commit

Thanks for the GIFs. Looking good!

assets/src/admin/paired-browsing/app.js
```diff
@@ -0,0 +1,304 @@
...
+
+	/**
+	 * Validates whether or not the window document is AMP compatible.
+	 *
+	 * @param {Document} doc Window document.
+	 * @return {boolean} True if AMP compatible, false if not.
+	 */
+	documentIsAmp( doc ) {
+		return doc.documentElement.hasAttribute( 'amp' ) ||
+			doc.documentElement.hasAttribute( '⚡️' );
```
westonruter: I think this check needs to also look for whether the `doc.querySelector( 'head > script[src="https://cdn.ampproject.org/v0.js]' )` is true. This will be more robust to handle the case of where we intentionally serve the AMP framework in a non-AMP page. Granted, this is only done in Standard mode at present, but it would be good to include here for completeness.

assets/src/admin/paired-browsing/client.js
westonruter: Still needs to be translatable. At the same time, now that we have the AMP/non-AMP message in the header, do we even need this label in the admin bar anymore? Should we just remove the AMP nav menu item altogether here?

```suggestion
				ampMenuItem.parentNode.removeChild( ampMenuItem );
```

This would prevent the user from trying to access the AMP version from the non-AMP version, which doesn't make sense since they are in paired browsing. There's no need for that admin bar menu item in the non-AMP page, now that there is the blue paired browsing bar.

includes/class-amp-theme-support.php
```diff
@@ -2495,6 +2503,128 @@ public static function enqueue_assets() {
...
+			true
+		);
+
+		// Force dev mode to be enabled. This ensures that the enqueued script and its dependencies
+		// will be present when the admin bar is not showing.
+		add_filter(
+			'amp_dev_mode_enabled',
+			static function () {
+				return true;
+			}
```
westonruter: This could be replaced with `'__return_true'`.

includes/class-amp-theme-support.php
```diff
@@ -2495,6 +2503,128 @@ public static function enqueue_assets() {
...
+			$tag = preg_replace( '/(?<=<script)(?=\s|>)/i', ' ' . AMP_Rule_Spec::DEV_MODE_ATTRIBUTE, $tag );
+		}
+
+		return $tag;
+	}
+
+	/**
+	 * Remove any unnecessary query vars that could hamper the paired browsing experience.
+	 */
+	public static function sanitize_url_for_paired_browsing() {
```
westonruter: Oh good. You've accounted for the most recent changes to https://gist.github.com/westonruter/f9ee9ea717d52471bae092879e3d52b0/revisions?

Namely, paired browsing should be possible to enter when you're previewing invalid markup being kept.

includes/validation/class-amp-validation-manager.php
```diff
@@ -469,6 +469,14 @@ public static function add_admin_bar_menu_items( $wp_admin_bar ) {
...
 		];
 
+		// Construct admin bar item to link to AMP version or non-AMP version.
+		$paired_browsing_item = [
+			'parent' => 'amp',
+			'id'     => 'amp-paired-browsing',
+			'title'  => esc_html__( 'Start paired browsing', 'amp' ),
+			'href'   => remove_query_arg( amp_get_slug(), add_query_arg( AMP_Theme_Support::PAIRED_BROWSING_QUERY_VAR, '1' ) ),
+		];
+
```
westonruter: Should this be moved into the condition below? No need to create this variable if it's not going to be used.

schlessera:
@googlebot I consent.

assets/src/admin/paired-browsing/client.js
schlessera: @pierlon This is still missing its "translatability". You should add the translated string on the PHP side to the obejct you're sending over to the client side via `wp_localize_script()`.

assets/src/admin/paired-browsing/app.css
```diff
@@ -0,0 +1,207 @@
...
+	right: 20px;
+	height: 32px;
+}
+
+#header .iframe-label {
+	margin: 0 auto;
+	font-weight: 600;
+	width: 50%;
+	justify-content: center;
+	text-decoration: underline;
```
westonruter: The underline here is making me think it is a link when it is not. Is this needed?

assets/src/admin/paired-browsing/app.js
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	/**
+	 * Add event listeners for buttons on disconnect overlay.
+	 */
+	addDisconnectButtonListeners() {
+		// The 'Continue' button simply hides the 'disconnected' overlay.
+		this.disconnectButtons.continue.addEventListener( 'click', () => {
+			this.disconnectOverlay.classList.remove( 'disconnected' );
+		} );
```
westonruter: I think perhaps the Continue button should result in actually exiting paired browsing. Thoughts?

pierlon:
> https://github.com/ampproject/amp-wp/pull/3656#discussion_r351621250: The underline here is making me think it is a link when it is not. Is this needed?

No, its not. I was only experimenting on how to make those labels more noticeable.

> https://github.com/ampproject/amp-wp/pull/3656#discussion_r351622253: I think perhaps the Continue button should result in actually exiting paired browsing. Thoughts?

That's a valid suggestion. You could go on browsing within the iframe, but upon the load of each page you will be nagged with the same 'disconnected' overlay until the paired browsing client is available again.

assets/src/admin/paired-browsing/app.js
```diff
@@ -0,0 +1,294 @@
...
+			oppositeWindow = this.nonAmpIframe.contentWindow;
+		} else {
+			// Force the non-AMP iframe to always have a non-AMP URL.
+			if ( this.documentIsAmp( win.document ) ) {
+				win.location.replace( this.removeAmpQueryVars( win.location.href ) );
+				return;
+			}
+
+			// Update the link used for exiting paired browsing.
+			this.exitLink.href = this.nonAmpIframe.contentWindow.location.href;
```
westonruter: As opposed to being the non-AMP window's URL, should this not rather persist whether the user was on an AMP or non-AMP page when they started paired browsing in the first place? If I was on an AMP page on the frontend, and I click “Start paired browsing” then when I exit paired browsing, then I should be put back onto an AMP page on the frontend. Right?

westonruter:
> That's a valid suggestion. You could go on browsing within the iframe, but upon the load of each page you will be nagged with the same 'disconnected' overlay until the paired browsing client is available again.

You know what paired browsing will eventually be able to make excellent use of? Instead of iframes it could use [portals](https://web.dev/hands-on-portals/), when they become available. This would allow seamlessly exiting the paired browsing context without having to reload the page. I think this will also work when _entering_ paired browsing as well:

1. Clicking the “Start paired browsing” link on in the admin bar could create a portal that contains the paired browsing interface.
2. The paired browsing interface can be activated.
3. The previous window could then be adopted as a portal _inside_ the paired browsing window and displayed as either the AMP or non-AMP frame, depending on if the user started from an AMP or non-AMP page.
4. And a second portal can then be created to contain the corresponding window (non-AMP for AMP, and vice versa), and this portal could likewise be used for exiting paired browsing.

This brings something to light. A user could actually exit to _either_ the AMP or non-AMP window when they are done with paired browsing. At the moment it forces the user to go to the non-AMP version (see https://github.com/ampproject/amp-wp/pull/3656/files#r352265423) but eventually they could be directed to either. There's no need to implement this now, but once portals become an option, this could be a nice touch.

westonruter:
I noticed a problem with the centered AMP/non-AMP frame titles. When resizing the window to show a mobile viewport, the “Paired Browsing” text overlaps the non-AMP label:

Screen Shot 2019-11-29 at 20 45 59

westonruter:
I have an idea for fixing the overlap problem and also exit link returning the user to where they came from: we can just turn both “Non-AMP” and “AMP” into links and keep them updated to be the same as the iframe `src`. We can then eliminate the separate Exit link altogether, and I think we could also remove the “Paired Browsing” text as well. I'm going to try it.

westonruter:
There, take a look a look at these:

Screen Shot 2019-11-29 at 21 23 50

Screen Shot 2019-11-29 at 21 24 06

pierlon:
That looks much neater! I love the use of the background colors in the nav bar to help differentiating the iframe labels. The use of the labels as links to exit paired browsing is also a great idea.

> You know what paired browsing will eventually be able to make excellent use of? Instead of iframes it could use portals, when they become available.

Yes! The use of that API would fit perfectly here and would greatly improve the UX. I guess the current implementation of using `iframes` will still be needed as a fallback, no?

westonruter:
That's right. If portals aren't available then we'd fall back to iframes.

includes/class-amp-theme-support.php
westonruter: OK, I see not. Well, this isn't so important. The thinking here is that if you had a validated URL which has removed markup, and you then go to mark one invalid markup as kept and click preview, at that point on the frontend it could be possible to click the “Start paired browsing” link to then see the original and the _previewed_ side-by-side. But this could get confusing. So let's defer that to think about later.

includes/class-amp-theme-support.php
```diff
@@ -2495,6 +2503,137 @@ public static function enqueue_assets() {
...
+	 * @param string $handle The script's registered handle.
+	 * @return string Tag.
+	 */
+	public static function filter_paired_browsing_client_script_loader_tag( $tag, $handle ) {
+		if ( self::has_dependency( wp_scripts(), 'amp-paired-browsing-client', $handle ) ) {
+			$tag = preg_replace( '/(?<=<script)(?=\s|>)/i', ' ' . AMP_Rule_Spec::DEV_MODE_ATTRIBUTE, $tag );
+		}
+
+		return $tag;
+	}
```
westonruter: FYI: This should be able to be largely eliminated if we implement this: https://github.com/google/site-kit-wp/pull/505#discussion_r348683617

includes/class-amp-theme-support.php
```diff
@@ -2495,6 +2503,140 @@ public static function enqueue_assets() {
...
+			'amp-paired-browsing-client',
+			amp_get_asset_url( '/js/amp-paired-browsing-client.js' ),
+			$dependencies,
+			$version,
+			true
+		);
+
+		// Force dev mode to be enabled. This ensures that the enqueued script and its dependencies
+		// will be present when the admin bar is not showing.
+		add_filter( 'amp_dev_mode_enabled', '__return_true' );
```
westonruter: @pierlon I just realized that this makes every page invalid AMP, because every page is forced to be in dev mode and the custom script is enqueued. Two options I see:

1. Short-circuit this function if an user is not logged-in (or rather, if `amp_is_dev_mode()` returns `false`). Since search crawlers will never be logged-in, they will never then see an invalid AMP page. This would also require entire paired browsing app should `wp_die()` if the user is not logged in.
2. Go back to only including this JS if the admin bar is showing. Accessing paired browsing here would still need to include the `wp_die()` if `! amp_is_dev_mode()`.

westonruter: See #3863.

westonruter:
I started experimenting with portals in #3864.

PR #3285, opened 2019-09-17
Add caption support to the Gallery block in <amp-carousel> and fix displaying Gallery block as carousel in WP 5.3

kienstra:
### Update: this has changed a lot, please see [more recent comments](https://github.com/ampproject/amp-wp/pull/3285#issuecomment-546171299)

For images in a gallery that have a caption, this adds caption support when 'Display as carousel' is selected:
captions-in-carousel

# Before

before-caption

# After

caption-appears-now

Closes #2855

(Please do `npm run build:css` for the CSS here)

PR #3659, opened 2019-10-29
Enable captions on Gallery shortcodes

### LINKED ISSUE #3658: Add support for captions to the Gallery shortcode
## Feature description

The `[gallery]` shortcode does not display captions. It'd be good to add support for them, as the Classic editor uses this for galleries.

A follow-up to #2855

---------------

## Acceptance criteria

1. If a `[gallery]` shortcode has images with captions, the captions should display in the `<amp-carousel>`.
2. Images without captions should still look good.

## Implementation brief

* PR #3285 [added logic](https://github.com/ampproject/amp-wp/pull/3285/files#diff-3f3a86f42d679f81a48116543bcb3b38R166) for creating an `<amp-carousel>` with slides that can have captions.

*  It'd be good to reuse as much of this as possible, though getting the actual caption text will be different for Gallery shortcodes.

## QA testing instructions

1. In `/wp-admin`, click 'Media' 
2. Click any image:
media-here-image

3. Add a caption, and make note of the ID:
add-caption

4. Repeat steps 2-3 
5. You should now have at least 2 IDs for images
6. Create a new post in the block editor
7. Add a 'Shortcode' block
8. Add a Gallery shortcode, with IDs of the images from step 2. For example, if the IDs were `4572` and `3619`, the shortcode would be:

```html
[gallery ids="4572, 3619"]
```
9. Click 'Preview,' and go to the AMP URL
10. Expected: The captions should be visible in the gallery
11. Repeat steps 1-9, with galleries that have no images with captions, very long captions, only a few images with captions, etc...
12. Please re-test the [Gallery block testing steps](https://github.com/ampproject/amp-wp/pull/3285#issuecomment-546171299). PR #3659 refactors that code, but there's no change intended to the Gallery block.

## Demo

* 

## Changelog entry

*

### LINKED ISSUE #3810: Namespace structure
## Feature description

In #3659 we started discussing adding namespaces for new objects to start reasoning about the main architecture of v2.0 of the plugin.

We need to decide what the overall namespace structure should look like, and what conventions we want to use for new names, subnamespaces and file/older names.

We've already agreed on PSR-4 so far, so the files and folders should match the namespaces and class/interface/trait names.

What is still unclear is:

1.) **Package structure in terms of overall PHP space**

Instead of just going with `AMP` as the namespace for the plugin, we should think about the bigger picture in terms of what the PHP space might need in total for AMP support.

In that context, `AMP` should be the "vendor" prefix for anything related to the AMP open source project. With the usual convention of `<vendor>/<package>` structure as package names and root namespaces, we would then have `AMP\<package>` for the plugin package name and root namespace.

The `<package>` could be something like "AMP-WP" (in whatever notation form).

What we currently have: `Amp\AmpWp`.

2.) **Notation or capitalization form**

With PSR-4, the biggest point of discussion is usually capitalization of acronyms and trademarks. There are two basic approaches: capitalize acronyms and trademarks as they would be capitalized in regular written form, or capitalize purely on the word boundary.

Capitalizing as in written form often seems to be more correct, but can lead to difficult-to-read combinations, as there is no word spacing to keep word boundaries intact. In our case, the term "AMP WP" could for example be written as `AMPWP`, but that is hardly legible.

Sometimes, this is overcome by adding underscores, like `AMP_WP`. However this is not directly part of PSR-4 and more of a remnant of PHP < 5.3 where underscores were used to overcome the lack of namespaces. So `AMP_WP` would have been used in stead of `AMP\WP`.

Capitalizing om the word boundary only is the more popular choice in the overall PHP space, as it remove the contentious topics of acronyms and trademarks from the equation. Trademarks in particular are quite troublesome, as they might also enforce lowercase starting letters, completely breaking the PascalCase flow. Therefore, using only the word boundaries as reference makes everything more consistent and predictable and avoids discussing trademark rules. This would produce `AmpWp` for us.

Of course we can go with neither as well, and look only on the aesthetic part of the namespace. `AmpWP` was suggested as being the best-looking one. It seems to combine the word boundary approach (`Amp` instead of `AMP`) with the trademarked approach ("capital P dangit").

3.) **First level of subnamespaces**

The general approach in PHP is to have the most critical parts of the code that all subsystems need to use in the root namespace. This is often mostly interfaces, as that is all you should need for cross-component communication.

Then, the first level of subnamespaces should represent the subsystems of the package.

That could be something like this (random example, nothing definitive yet):

```
- Amp
	- AmpWp
		- Administration
		- Component
		- EmbedHandler
		- Sanitizer
		- Validation
		- Widget
```

As we want to plan to extract some of these subsystems into separate packages, we might already shape everything in that way to make it easy to separate later:

```
- Amp
	- AmpWp
	    - Administration
	    - Component
	    - EmbedHandler
	    - Widget
	- Sanitizer
	- Validation
```

It would regroup the WP-specific stuff, and keep the reusable, generic-PHP code separate.

We could then later easily extract these out into separate packages `amp/sanitizer` and `amp/validation`.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

kienstra:
## Summary
This allows displaying captions in Gallery shortcodes. It adds a class `AMP_Carousel` that [uses logic](https://github.com/ampproject/amp-wp/pull/3285/files#diff-3f3a86f42d679f81a48116543bcb3b38R166) in the Gallery block sanitizer from PR #3285. 

Fixes #3658
Closes #3810

## Checklist


includes/sanitizers/class-amp-gallery-block-sanitizer.php
```diff
@@ -149,105 +125,17 @@ public function sanitize() {
...
-					);
-					$caption_span    = AMP_DOM_Utils::create_node( $this->dom, 'span', [] );
-					$text_node       = $this->dom->createTextNode( $possible_caption_text );
-
-					$caption_span->appendChild( $text_node );
-					$caption_wrapper->appendChild( $caption_span );
-					$slide->appendChild( $caption_wrapper );
-				}
-
-				$amp_carousel->appendChild( $slide );
```
kienstra: This 'red' section in the diff is moved to `AMP_Carousel::create_and_get()`, with few changes.

includes/sanitizers/class-amp-gallery-block-sanitizer.php
```diff
@@ -149,105 +125,17 @@ public function sanitize() {
...
-	 *
-	 * @param DOMElement $element The UL element.
-	 * @return array {
-	 *     Dimensions.
-	 *
-	 *     @type int $width  Width.
-	 *     @type int $height Height.
-	 * }
-	 */
-	protected function get_carousel_dimensions( $element ) {
```
kienstra: This is mainly moved to [AMP_Carousel](https://github.com/ampproject/amp-wp/pull/3659/files#diff-1892bee5b65cec2225d075857448d046R124), with few changes.

includes/utils/class-amp-carousel.php
```diff
@@ -0,0 +1,160 @@
...
+	 *
+	 * @param array[] {
+	 *     The images and their respective captions, if any.
+	 *
+	 *     @type DOMElement  $image A representation of an image, in index 0.
+	 *     @type string|null $caption The caption for that image, if any, in index 1.
+	 * }
+	 * @return DOMElement A representation of the <amp-carousel>.
+	 */
+	public function create_and_get( $images_and_captions ) {
```
kienstra: It'd be better if `$images_and_captions` were an associative array of images to their respective captions.

But a `DOMElement` (the image) isn't allowed as key of an associative array.

includes/utils/class-amp-carousel.php
```diff
@@ -0,0 +1,160 @@
...
+	 * @param array[] $images_and_captions {
+	 *     The images and their respective captions, if any.
+	 *
+	 *     @type DOMElement  $image A representation of an image, in index 0.
+	 *     @type string|null $caption The caption for that image, if any, in index 1.
+	 * }
+	 * @return DOMElement A representation of the <amp-carousel>.
+	 */
+	public function create_and_get( $images_and_captions ) {
+		$images = [];
```
kienstra: This function is mainly moved from `AMP_Gallery_Block_Sanitizer`, with few changes.

includes/utils/class-amp-carousel.php
```diff
@@ -0,0 +1,160 @@
...
+	 *
+	 * @param array[] $images The images and captions to get the dimensions from.
+	 * @return array {
+	 *     Dimensions.
+	 *
+	 *     @type int $width  Width.
+	 *     @type int $height Height.
+	 * }
+	 */
+	public function get_dimensions( $images ) {
```
kienstra: This function is mainly moved from `AMP_Gallery_Block_Sanitizer::get_carousel_dimensions()`, with few changes.

includes/embeds/class-amp-gallery-embed.php
```diff
@@ -129,25 +129,14 @@ public function shortcode( $attr ) {
...
-				'amp-image-lightbox',
-				[
-					'id'                           => AMP_Base_Sanitizer::AMP_IMAGE_LIGHTBOX_ID,
-					'layout'                       => 'nodisplay',
-					'data-close-button-aria-label' => __( 'Close', 'amp' ),
-				]
-			);
-			/* We need to add lightbox tag, too. @todo Could there be a better alternative for this? */
-			return $this->render( $args ) . $lightbox_tag;
-		}
```
kienstra: Now that this is using [amp-lightbox-gallery](https://amp.dev/documentation/components/amp-lightbox-gallery/) as [recommended by amphtml](https://github.com/ampproject/amphtml/issues/25021#issuecomment-541846264), all that's needed is a `lightbox` attribute on the `<amp-img>` or `<amp-carousel>`

includes/utils/class-amp-carousel.php
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * @package AMP
+ */
+
+/**
+ * Class AMP_Carousel
+ *
+ * Gets the markup for an <amp-carousel>.
+ *
+ * @since 1.4.1
```
kienstra: Maybe this will be different.

includes/sanitizers/class-amp-gallery-block-sanitizer.php
```diff
@@ -12,24 +12,6 @@
  */
 class AMP_Gallery_Block_Sanitizer extends AMP_Base_Sanitizer {
 
-	/**
```
kienstra: There's no intended change to the Gallery block, only a refactoring of this logic into the class `AMP_Carousel`.

kienstra:
**Request For Review, But No Hurry**

Hi @westonruter,
Good to see you earlier.

No hurry reviewing this, with the stable release and WCUS coming up. Just wanted to let you know that's it's ready when you have a chance.

This extracts much of the logic from the Gallery block caption PR (#3285) into a class `AMP_Carousel` that is also used for the Gallery shortcode.

captions-supported

There's no intended change to Gallery blocks, and I didn't see any in my testing. But I asked for regression testing for them in the [QA testing instructions](https://github.com/ampproject/amp-wp/issues/3658#issue-513686753).

Like [in the Gallery block caption PR (#3285)](https://github.com/ampproject/amp-wp/pull/3285#issuecomment-546168357), the Gallery shortcode now looks different if there's a mix of landscape and square images.

I think it's an improvement, but some might disagree.

Thanks, Weston!

## Before 

before-pr-car

## After

fill-width-amp-amp-carousel

includes/utils/class-amp-carousel.php
westonruter: Yes, let's have the changes here by targeting 1.5.0

includes/utils/class-amp-carousel.php
westonruter: That's pretty slick. I suppose my only hesitation is overly engineering an API for this which we don't necessarily want others to use externally, until we've given it it enough thought into how it relates to the other pieces. If we just use a simple array as the input, then it is more flexible in the case we decide to go a different direction. But then again, if we don't want people to be using this class yet anyway, then I don't suppose it matters. Just make sure it is marked with `@internal`.

If we do go this route, then it should definitely fall under the `AMP` namespace and the autoloader should automatically locate the filename purely based off of the name of the class/interface/trait. If we start creating a ton of little PHP files, we don't want to be having to embiggen the `\AMP_Autoloader::$classmap` endlessly.

includes/utils/class-amp-carousel.php
kienstra: Sounds good

includes/utils/class-amp-carousel.php
```diff
@@ -0,0 +1,163 @@
...
+			return [ self::FALLBACK_WIDTH, self::FALLBACK_HEIGHT ];
+		}
+
+		foreach ( $images as $image ) {
+			// Account for an <amp-img> that's wrapped in an <a>.
+			if ( 'amp-img' !== $image->tagName && isset( $image->firstChild->tagName ) && 'amp-img' === $image->firstChild->tagName ) {
+				$image = $image->firstChild;
+			}
+
+			if ( ! is_numeric( $image->getAttribute( 'width' ) ) || ! is_numeric( $image->getAttribute( 'height' ) ) ) {
```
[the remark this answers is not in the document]
kienstra: Good point, it's possible that there'd be a `'0'` width or height.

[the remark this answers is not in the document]
kienstra: https://github.com/ampproject/amp-wp/pull/3659/commits/93aa3ea9d34f1b9b65fb1dba012be3fb1d39aaab continues if there's a `0` or `0.0` width or height.

kienstra:
Hi @schlessera,
>As an example, here's how an object-based representation would look like and behave (comments mostly skipped for brevity):

Thanks a lot for your image collection [code](https://github.com/ampproject/amp-wp/pull/3659#discussion_r344376957) above. https://github.com/ampproject/amp-wp/pull/3659/commits/d3ba063aa808c894bc6aae0486e7ae948898d84b commits the code mainly as you wrote it.

includes/utils/class-amp-carousel.php
```diff
@@ -0,0 +1,163 @@
...
+		$this->dom = $dom;
+	}
+
+	/**
+	 * Creates and gets an <amp-carousel> with the given images and captions.
+	 *
+	 * @param array[] $images_and_captions {
+	 *     The images and their respective captions, if any.
+	 *
+	 *     @type DOMElement  $image A representation of an image, in index 0.
```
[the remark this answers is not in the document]
kienstra: This is a good suggestion, but luckily the entire `@param` tag was simplified, thanks to your code below.

includes/utils/class-amp-image-list.php
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * @package AMP
+ */
+
+/**
+ * Interface Has_Caption
+ *
+ * @since 1.5.0
+ */
+interface Has_Caption {
```
westonruter: This should be in separate file.

westonruter: This either needs prefixing or (better) the `AMP` namespace.

includes/utils/class-amp-image-list.php
```diff
@@ -0,0 +1,138 @@
...
+		return $this->image_node;
+	}
+}
+
+/**
+ * Class AMP_Captioned_Image
+ *
+ * @since 1.5.0
+ */
+final class AMP_Captioned_Image extends AMP_Image implements Has_Caption {
```
westonruter: Should be in a separate file.

includes/utils/class-amp-image-list.php
```diff
@@ -0,0 +1,138 @@
...
+		return $this->caption;
+	}
+}
+
+/**
+ * Class AMP_Image_List
+ *
+ * @since 1.5.0
+ */
+final class AMP_Image_List implements IteratorAggregate, Countable {
```
westonruter: Should be in a separate file.

includes/class-amp-autoloader.php
```diff
@@ -96,8 +96,10 @@ class AMP_Autoloader {
 		'AMP_Content'                        => 'includes/templates/class-amp-content',
 		'AMP_Content_Sanitizer'              => 'includes/templates/class-amp-content-sanitizer',
 		'AMP_Post_Template'                  => 'includes/templates/class-amp-post-template',
+		'AMP_Carousel'                       => 'includes/utils/class-amp-carousel',
 		'AMP_DOM_Utils'                      => 'includes/utils/class-amp-dom-utils',
 		'AMP_HTML_Utils'                     => 'includes/utils/class-amp-html-utils',
+		'AMP_Image_List'                     => 'includes/utils/class-amp-image-list',
```
westonruter: As opposed to adding new AMP classes in the global namespace needing to add them to this autoloader map, we should take this as an opportunity to introduce the `AMP` namespace and then to have a PSR autoloader be able to automatically locate the class file purely based on the class name.

includes/utils/class-amp-image-list.php
kienstra: Sounds good, I'll move it.

schlessera:
Yes, absolutely, the code I provided was just meant to be an example. If we go that route (meaning we add more classes at a smaller granularity), we need to get away from the manually managed class map.

We can still generate one at build time for performance reasons, but we should need to maintain it manually.

includes/class-amp-autoloader.php
kienstra: What do you think about https://github.com/ampproject/amp-wp/pull/3659/commits/5794193bfc992da28eddbee4aff2fd28c4dbc322#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R48? 

This adds a directory `includes/amp/` that can have files with the `AMP` namespace.

includes/utils/class-amp-image-list.php
kienstra: Sure, this now [has the AMP namespace](https://github.com/ampproject/amp-wp/pull/3659/files#diff-97f7f9bbd8f82d4b8782110a007991cbR8).

kienstra: This is [now in its own file](https://github.com/ampproject/amp-wp/pull/3659/files#diff-97f7f9bbd8f82d4b8782110a007991cbR1).

includes/utils/class-amp-image-list.php
kienstra: Sure, this is now [in its own file](https://github.com/ampproject/amp-wp/pull/3659/files#diff-672fa2445cf206c6ba3b93fca58976c0R1).

includes/utils/class-amp-image-list.php
kienstra: Sure, this now has [its own file](https://github.com/ampproject/amp-wp/pull/3659/files#diff-5666c973c3a9ea12bb05da93645c37d2R16).

composer.json
```diff
@@ -33,6 +33,7 @@
     "ext-zip": "Enables the use of ZipArchive to export AMP Stories."
   },
   "config": {
+    "optimize-autoloader": true,
```
kienstra: This ensures that the classmap for `includes/amp/` is regenerated with `composer install`. Otherwise, I think it's necessary to run `composer install -o`.

includes/class-amp-autoloader.php
```diff
@@ -96,6 +96,7 @@ class AMP_Autoloader {
 		'AMP_Content'                        => 'includes/templates/class-amp-content',
 		'AMP_Content_Sanitizer'              => 'includes/templates/class-amp-content-sanitizer',
 		'AMP_Post_Template'                  => 'includes/templates/class-amp-post-template',
+		'AMP_Carousel'                       => 'includes/utils/class-amp-carousel',
```
kienstra: I could also move this into the `AMP` namespace.

includes/sanitizers/class-amp-gallery-block-sanitizer.php
kienstra: (Now it's `AMP\Carousel`)

includes/amp/class-captioned-image.php
```diff
@@ -0,0 +1,44 @@
...
+	 */
+	private $caption;
+
+	/**
+	 * Constructs the class.
+	 *
+	 * @param \DOMElement $image_node The image node.
+	 * @param string      $caption    The caption text.
+	 */
+	public function __construct( \DOMElement $image_node, $caption ) {
```
schlessera: I suggest always importing all classes, even the non-namespaced ones. This will not only get rid of the `\` in front of them, make it more readable, but also will add a clear list of external dependencies at the top of each class:
```php
use DOMElement;
```

This list of dependencies is very valuable later down the road for debugging and refactoring.

```suggestion
	public function __construct( DOMElement $image_node, $caption ) {
```

includes/amp/class-image-list.php
```diff
@@ -0,0 +1,57 @@
+<?php
+/**
+ * Class Image_List
+ *
+ * @package AMP
+ */
+
+namespace AMP;
+
+/**
```
schlessera: Again, I suggest importing all external dependencies, as that makes it very clear what to watch out for later when refactoring or diagnosing bugs.
```suggestion
use ArrayIterator;
use Countable;
use DOMElement;
use IteratorAggregate;

/**
```

includes/embeds/class-amp-gallery-embed.php
```diff
@@ -241,47 +230,38 @@ public function render( $args ) {
...
-			'amp-carousel',
-			[
-				'width'  => $carousel_width,
-				'height' => $carousel_height,
-				'type'   => 'slides',
-				'layout' => 'responsive',
-			],
-			implode( PHP_EOL, $images )
-		);
+		$amp_carousel  = new Carousel( $dom );
```
schlessera: The `Carousel` does not behave like an object, but rather like a collection of procedural functions:
- create $carousel
- use $carousel on $image_list to create node

Note that the carousel on itself is not functional, and it can be used with multiple image lists. It is a procedural function collection instead of an object.

A `Carousel` does not really make sense without an image list (even if only an empty one), so a better OOP structure would be to accept the image list in the constructor.

Then, the "API" would rather look like this:

```php
$amp_carousel = new Carousel( $dom, $image_list );
$dom->appendChild( $amp_carousel->as_dom_element() );
```

If you want to get fancy, you could even let `AMP\Carousel` extend `DOMElement`, so you could just simply use `$dom->appendChild( $amp_carousel );`.

includes/utils/class-amp-carousel.php
schlessera: Yes, I agree with marking as `@internal` for now. We should slowly build up our understanding and infrastructure with regards to an OOP approach of the plugin, and only really advertise an explicit set of interfaces with v2.0.

schlessera:
Another note regarding the autoloader:

The overall PHP convention is to use `src/` as the root folder for the autoloader class/interface/trait definitions, and `includes/` for the procedural code (generally, files that need to be directly included and cause side-effects).

I don't want to overcomplicate things, so I'm not sure whether it would be a good or a bad idea to build the new `AMP` namespace in a separate `src` folder.

Pros:
- sticks to PHP convention
- allows us to keep separate code styles in terms of file/folder structure for `src/` and `includes/`

Cons:
- splits our current selection of classes into two separate locations until the conversion from the old system to the new is complete

As the con here is not minor, I'd tend to suggest keeping everything within `includes/` for v1.5.0 but plan on moving the `AMP` namespace over to `src/` with v2.0.

@kienstra, @westonruter Thoughts?

westonruter:
@schlessera I'm fine with `src` as it kicks off the new codebase for the strangler pattern you've mentioned. I'd be especially keen on it if it means we can avoid having an `amp` subdirectory like `includes/amp` which seems somewhat redundant. Couldn't the base directory for the `AMP` namespace be designated as `src` and so if there was an `AMP\Carousel\Element` class it could be found at `src/Carousel/Element.php`, for example?

schlessera:
Yes, that's how it is usually done. The `src/` maps to the namespace root. Usually, the namespace root would be `<vendor>/<package>`. But as I don't think we want to have a vendor here, having `AMP` as namespace root is fine. So the `src/Component/Carousel.php` file would be the `AMP\Component\Carousel` class.

What we might want to discuss is how the `AMP` namespace relates to the overall PHP ecosystem.

As we plan to extract a reusable `AMP` library out of the sanitizer system, it would logically make sense to distinguish between `AMP` and `AMP_WP` for example, as the `AMP` package might be used in Drupal as well and will not contain the WP integration code.

In terms of packages in Composer, we would then end up with something like this, for example:

- package `amp/amp` - reusable library under the `AMP` namespace
- package `amp/amp-wp` - WP plugin under the `AMP_WP` namespace.

That would be problematic however in a single package, so when we switch to v2, we should discuss packaging as well and come back to the namespace.

Aöternatively, for now, we could have all WP-specific code under a sub-namespace within `AMP`. This could be something like `AMP\WordPress` or `AMP\Integration`. However, that might just make everything a bit too spread out at this point.

westonruter:
@schlessera This brings up a great point. Perhaps the root namespace used by the plugin should actually be `AMP\WP` (essentially vendor/package). Then when we extract parts for agnostic packages, we can use `AMP` but could keep to using packages as well, like `AMP\Validator`.

schlessera:
Yes, something like that would be preferable if we want to play nice with the bigger PHP ecosystem.

However, `WP` is not a good package name, I would suggest `AMP\AMP_WP` or `Amp\AmpWP` instead.

includes/amp/class-image-list.php
kienstra: Ah, good idea. https://github.com/ampproject/amp-wp/pull/3659/commits/3f0ee523069fcda4e8588f71d781256dad3346ca imports the classes. 

Your commit above looks good, but I moved the files to `src/component/` earlier. I should have committed that suggestion first 😄

includes/amp/class-captioned-image.php
kienstra: Thanks, https://github.com/ampproject/amp-wp/pull/3659/commits/3f0ee523069fcda4e8588f71d781256dad3346ca imports the classes.

includes/amp/class-carousel.php
```diff
@@ -0,0 +1,164 @@
+<?php
+/**
+ * Class Carousel.
+ *
+ * @package AMP
+ */
+
+namespace AMP;
```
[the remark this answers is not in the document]
kienstra: Good idea, no need to fill the new namespace with these classes.

How about the [namespace of Amp\AmpWP\Component](https://github.com/ampproject/amp-wp/pull/3659/files#diff-e72cc13437c94de96b8c154d6353b40eR8), per the [discussion below](https://github.com/ampproject/amp-wp/pull/3659#issuecomment-555141508)?

includes/amp/class-captioned-image.php
```diff
@@ -0,0 +1,44 @@
+<?php
+/**
+ * Class Captioned_Image
+ *
+ * @package AMP
+ */
+
+namespace AMP;
```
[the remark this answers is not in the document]
kienstra: Good idea, the current [directory to namespace mapping](https://github.com/ampproject/amp-wp/pull/3659/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R51) is:

`src/components` : `Amp\AmpWP\Components`

How does that look?

includes/embeds/class-amp-gallery-embed.php
kienstra: Thanks, https://github.com/ampproject/amp-wp/pull/3659/commits/2b4a98fd6ae549430da57883aea4d60c0bf66945 changes the constructor to accept the `$dom` and `$image_list`. Then, it can get the `DOMElement` in a separate function.

kienstra: Is that what you had in mind?

kienstra:
Thanks for discussing the namespaces, sorry for not taking part.

How does the [current directory to namespace mapping](https://github.com/ampproject/amp-wp/pull/3659/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R51) look?

`src/components/` : `Amp\AmpWP\Components`

For example, the new class `src/components/class-carousel.php` has a namespace of `Amp\AmpWP\Components\`.

includes/sanitizers/class-amp-gallery-block-sanitizer.php
kienstra: Now, it's `Amp\AmpWp\Component\Carousel`

schlessera:
@kienstra As I highly doubt this plugin will ever be merged into Core, I strongly recommend going with PSR-4 instead.

`src/Component/Carousel.php` => `Amp\AmpWp\Component\Carousel`.

schlessera:
Hmm, I'm having a hard time making up my mind about whether we should make the `Image(_List)` objects immutable or not. In general, immutable objects are easier to handle as you can't have bugs where one piece of code breaks a completely unrelated piece of code somewhere else. However, it would mean that the way the `Carousel` is set up, we wouldn't be able to add images after its construction. This is generally a good thing, but developers hooking into this later on might expect this to happen.

Would love to discuss the use of immutable objects more genrally, @kienstra & @westonruter.

westonruter:
> This is generally a good thing, but developers hooking into this later on might expect this to happen.

Since we're not intending this to be used by other developers (for now at least, hence `@internal`), keeping it immutable is fine.

kienstra:
>@kienstra As I highly doubt this plugin will ever be merged into Core, I strongly recommend going with PSR-4 instead.

>src/Component/Carousel.php => Amp\AmpWp\Component\Carousel.

Sounds good, https://github.com/ampproject/amp-wp/pull/3659/commits/d7fdbd240677228dd1987208b991d13b396c6e23 makes that change.

phpcs.xml
```diff
@@ -22,6 +22,7 @@
 
 	<rule ref="WordPress.Files.FileName.InvalidClassFileName">
 		<exclude-pattern>tests/*</exclude-pattern>
+		<exclude-pattern>src/*</exclude-pattern>
```
westonruter: In addition to excluding the WordPress sniff for the `src` directory, what about including another sniff that applies _only_ for the `src` directory to enforce PSR-4 filenames?

includes/sanitizers/class-amp-gallery-block-sanitizer.php
westonruter: Actually, it's `Amp\AmpWP` (capital_P_dangit? 😄) 

For what it's worth, I do like `AmpWP` more than `AmpWp`.

src/Component/Carousel.php
```diff
@@ -0,0 +1,173 @@
+<?php
+/**
+ * Class Carousel.
+ *
+ * @package AMP
```
westonruter: Question: The root namespace is `Amp` but the package is `AMP`. Does this matter?

Also: Should the root namespace actually be `AMP`? Now that “AMP” is no longer an acronym of “Accelerated Mobile Pages” it's possible it may eventually be spelled “Amp” but this is not the case today. 

/cc @schlessera

composer.json
```diff
@@ -44,5 +45,10 @@
         "PHP-CSS-Parser: Fix parsing CSS selectors which contain commas <https://github.com/sabberworm/PHP-CSS-Parser/pull/138>": "https://github.com/sabberworm/PHP-CSS-Parser/commit/fa139f65c5b098ae652c970b25e6eb03fc495eb4.diff"
       }
     }
+  },
+  "autoload": {
+    "psr-4": {
+      "Amp\\AmpWP\\": "src/"
```
westonruter: See questions below about whether root namespace should be `AMP` or `Amp`. That being said, should the package be `AmpWP` or `AMP_WP`? I see WP-CLI uses `WP_CLI`. /cc @schlessera

src/Component/Image.php
```diff
@@ -0,0 +1,44 @@
...
+
+use DOMElement;
+
+/**
+ * Class Image
+ *
+ * @internal
+ * @since 1.5.0
+ */
+class Image {
```
westonruter: This class seems somewhat useless. It's just a wrapper for a `DOMElement`. Should `DOMElement` just be used instead? I suppose it is needed for extending to then have a caption.

src/Component/Image.php
```diff
@@ -0,0 +1,44 @@
...
+	 */
+	protected $image_node;
+
+	/**
+	 * Constructs the class.
+	 *
+	 * @param DOMElement $image_node The image node.
+	 */
+	public function __construct( DOMElement $image_node ) {
+		$this->image_node = $image_node;
```
westonruter: Should this throw an Exception if `$image_node` is not an `amp-img`, `img`, or `amp-anim`?

phpcs.xml
kienstra: That's a good idea.

src/Component/Image.php
kienstra: That's a good idea. I'll make a commit for that.

src/Component/Image.php
kienstra: Hm, that's a good point. I'll look at this more.

src/Component/Image.php
westonruter: Also, it seems somewhat needless to have this class, as it is just a wrapper around `DOMElement`. Since non-image elements can also be used in a carousel, it would be more flexible to allow any DOM element actually.

So instead of having an `Image` class, could we just use `DOMElement`. And then for elements with captions, could this then be a subclass of `DOMElement` like `CaptionedElement` which takes a `$caption` arg in its constructor?

westonruter: Or rather, `CaptionedSlide`. I suppose it couldn't be a subclass because you're working with existing DOM elements. But it could be a container object that has the existing DOM element as a member?

kienstra: Sure, what do you think about https://github.com/ampproject/amp-wp/pull/3659/commits/12c6d2694589adce9e9293550f62a0a3eeb4e18f ?

`class Image` is deleted, but much of it is [moved to CaptionedSlide](https://github.com/ampproject/amp-wp/pull/3659/commits/12c6d2694589adce9e9293550f62a0a3eeb4e18f#diff-7d849810f36483c7db89977ff845c00fR25).

Maybe it's a slight improvement.

phpcs.xml
kienstra: So far, I couldn't find a sniff for this. But I'll look more.

A sniff that [this issue](https://github.com/squizlabs/PHP_CodeSniffer/issues/2163#issuecomment-422986914) recomments, Slevomat, [requires PHP ^7.2](https://github.com/slevomat/coding-standard/blob/master/composer.json#L12). 

Likewise, [Suin requires >=7.1](https://github.com/suin/phpcs-psr4-sniff/blob/master/composer.json#L30).

westonruter: What about [`Squiz.Sniffs.ClassFileName`](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Squiz/Sniffs/Classes/ClassFileNameSniff.php)?

kienstra: Ah, good idea. I'll try that out.

kienstra: Great idea, https://github.com/ampproject/amp-wp/pull/3659/commits/951cd3f76b1f67538363a702c0714ed464f47797 adds that rule only for `src/*`

src/Component/Image.php
schlessera: The point of the `Image` class was to ensure that the `ImageList` only ever can accept images, not random elements.

This is one of the cornerstones of strong typing, that you actually have types to check against.

If our `ImageList` accepts any element, it is actually an `ElementList`, and it defeats the purpose.

schlessera: Ah, I just saw the comment about the carousel also accepting other elements.

But we should rename the `ImageList` then if it accepts non-images as well.

There's already a `DOMNodeList`, so we could indeed add `DOMElementList` just for typing purposes, it would improve typing and remove the need for these silly large docblocks stating a node is an element, just to get its attributes.

schlessera:
Regarding the namespace and package structure, I opened another issue to unblock this PR: https://github.com/ampproject/amp-wp/issues/3810

schlessera:
> Since we're not intending this to be used by other developers (for now at least, hence @internal), keeping it immutable is fine.

That makes sense. However, right now the objects are _not_ immutable, so that would require an additional change.

@kienstra Can you make the `ImageList` immutable`? If I remember correctly, you'd only need to clone on `add()` to achieve that, and always re-assign the return value.

src/Component/CaptionedSlide.php
```diff
@@ -0,0 +1,62 @@
+<?php
+/**
+ * Class CaptionedSlide
+ *
+ * @package AMP
```
westonruter:
```suggestion
 * @package Amp\AmpWP
```

Correct?

src/Component/DOMElementList.php
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * Adds an image to the list.
+	 *
+	 * @param DOMElement $element The element to add, possibly an image.
+	 * @param string     $caption The caption to add.
+	 * @return self
+	 */
+	public function add( DOMElement $element, $caption = '' ) {
+		$this->elements[] = empty( $caption ) ? $element->cloneNode() : new CaptionedSlide( $element->cloneNode(), $caption );
```
westonruter: Shouldn't this be a deep clone? Otherwise the `noscript > img` children won't be included.

```suggestion
		$this->elements[] = empty( $caption ) ? $element->cloneNode( true ) : new CaptionedSlide( $element->cloneNode( true ), $caption );
```

kienstra: Yes, good idea.

kienstra: That helped, but the tests still failed locally. I'm working on it.

kienstra:
Hi @schlessera,
>@kienstra Can you make the ImageList immutable? If I remember correctly, you'd only need to clone on add() to achieve that, and always re-assign the return value.

So far, I couldn't find a way to make `ImageList` (now `ElementList`) immutable, without causing an error.

This looked to cause an error [in this unit test](https://travis-ci.org/ampproject/amp-wp/jobs/615776421#L1240) on [this line](https://github.com/ampproject/amp-wp/blob/7b603eaeda0d87c71815f8014c9a5ae09fd3eac4/src/Component/Carousel.php#L152)

getAttribute(): Couldn't fetch DOMElement
It looks like there was no longer a reference
to the DOMElement after cloning it.

kienstra:
Though maybe I'm not clear on what immutable means in this context.

Here's a (now reverted) [unit test](https://github.com/ampproject/amp-wp/pull/3659/commits/8de234083b123b0c52e03cfb9473d4893a447f4c#diff-eef4617d0d784367468f2afe7614165bL110) to check if it was possible to change an attribute when [iterating over](https://github.com/ampproject/amp-wp/pull/3659/files#diff-accd9bb2db2b5a5ef55b03be951fe83eR50) the `DOMElementList`.

schlessera:
I'm sorry, @kienstra, my comments were probably misleading.

You cannot really make the `DOMNodes` reliably immutable, as they are not designed that way.

What I meant was make the "container" `ImageList` immutable in the sense that when you have an image list, it will stick to the images it has, even if some other codes adds an image later on.

So, just cloning the list itself before adding and returning the clone, and always assigning the return value of the `add()` operation to accept the clone as the new image list.

It should be pretty straight-forward. Anything more is overkill and will still never work reliably.

kienstra:
Hi @schlessera,
Thanks, that's a good idea 😄 

https://github.com/ampproject/amp-wp/pull/3659/commits/3cdd86f52be726ba15e2174e08d075842cf43d7a [clones the list](https://github.com/ampproject/amp-wp/pull/3659/commits/3cdd86f52be726ba15e2174e08d075842cf43d7a#diff-accd9bb2db2b5a5ef55b03be951fe83eR38) in `add()` and returns that clone.

composer.json
schlessera: This should be removed, and a corresponding change should be made to the build process instead.

If an optimized autoloader is forced, then you'll have to constantly rebuild the plugin on filesystem changes even during development.

composer.json
schlessera: Discussed in https://github.com/ampproject/amp-wp/issues/3810

src/Component/Carousel.php
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Class Carousel
+ *
+ * Gets the markup for an <amp-carousel>.
+ *
+ * @internal
+ * @since 1.5.0
+ */
+class Carousel {
```
schlessera: As we marked the `Carousel` as `@internal`, we should mark the `Carousel` as `final` and use `private` accessibility for the properties.

In general, final+private should be the default, unless we explicitly want something to be extendable, not the other way around.

We can always turn a final class into an extendable class. The other way it not so easy because third-party code might already be relying on it being extendable.

```suggestion
final class Carousel {
```

src/Component/Carousel.php
```diff
@@ -0,0 +1,173 @@
...
+	 * @var int
+	 */
+	const FALLBACK_HEIGHT = 480;
+
+	/**
+	 * An object representation of the DOM.
+	 *
+	 * @var DOMDocument
+	 */
+	protected $dom;
```
schlessera:
```suggestion
	private $dom;
```

src/Component/Carousel.php
```diff
@@ -0,0 +1,173 @@
...
+
+	/**
+	 * The images to add to the carousel.
+	 *
+	 * @var DOMElementList
+	 */
+	protected $images;
```
schlessera:
```suggestion
	private $images;
```

src/Component/Carousel.php
```diff
@@ -0,0 +1,173 @@
...
+	 * not necessarily the image with the biggest absolute width.
+	 *
+	 * @return array {
+	 *     The carousel dimensions.
+	 *
+	 *     @type int $width  The width of the carousel, at index 0.
+	 *     @type int $height The height of the carousel, at index 1.
+	 * }
+	 */
+	public function get_dimensions() {
```
schlessera: Is this supposed to be publicly accessible, i.e. do we want to give access to this to third-party developers as our public interface?

src/Component/Carousel.php
```diff
@@ -0,0 +1,173 @@
...
+
+	/**
+	 * Instantiates the class.
+	 *
+	 * @param DOMDocument    $dom    The dom to use to create a carousel.
+	 * @param DOMElementList $images The images from which to create a carousel.
+	 */
+	public function __construct( $dom, DOMElementList $images ) {
+		$this->dom    = $dom;
+		$this->images = $images;
```
schlessera: As the carousel can contain other elements than just images, should we rename this property and the corresponding variables into `$slides` instead to match the `CaptionedSlide` adjustment?

src/Component/DOMElementList.php
```diff
@@ -25,7 +25,7 @@ final class DOMElementList implements IteratorAggregate, Countable {
 	 *
 	 * @var DOMElement[]
 	 */
-	private $elements = [];
+	public $elements = [];
```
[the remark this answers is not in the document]
kienstra: My mistake, https://github.com/ampproject/amp-wp/pull/3659/commits/38df209f9b66a3e0ce9eefff468d527c5cd8ce1d changes it back to `private`.

I originally changed it because of failing unit tests, but they pass now.

src/Component/Carousel.php
kienstra: Ah, good point. It's now changed in https://github.com/ampproject/amp-wp/pull/3659/commits/450f9cdf106074db9a941ebcf0af8b91bcbd53d8

I should have [brought this up before](https://github.com/ampproject/amp-wp/pull/3659#discussion_r349240797): the `Carousel` class probably [makes the most sense for images](https://github.com/ampproject/amp-wp/pull/3659/files#diff-5d38610362507a06dc80a89b9151dff7R90-R98), and it's now only used for images.

But as you mentioned, the `DOMElementList` can now have any element.

composer.json
kienstra: Sure, it's removed in https://github.com/ampproject/amp-wp/pull/3659/commits/4ac1d2528732709d8cf3daf9152908ab91e7f20d. Next, I'll apply in in the build processes.

src/Component/Carousel.php
kienstra: Uf, there's probably not a need for it to be public.

kienstra: https://github.com/ampproject/amp-wp/pull/3659/commits/478f71e28c4c333c31f1d4abc1c057af0379c0eb makes it `private`

composer.json
kienstra: https://github.com/ampproject/amp-wp/pull/3659/commits/c7d5d6e69f94289144d67b9a6adae85f8c7fa080 adds the `-o` flag to `composer install` only for production build processes and when creating a pre-release.

Maybe it should should also be added to `.travis.yml`, but I thought the optimization [wouldn't be worth](https://getcomposer.org/doc/articles/autoloader-optimization.md#autoloader-optimization) the trouble in debugging if the build is cached:

>The performance gains are not worth the trouble in a development setting.

PR #3661, opened 2019-10-29
Check if element is not in top toolbar.

### LINKED ISSUE #3647: Context menu opening when right-clicking on Toolbar buttons
## Bug Description

When doing right-click on the Toolbar (e.g. alignment button), the custom Context menu opens

## Expected Behaviour

Context menu should only open when right-clicking on the block.

## Steps to reproduce

1. Add a block (e.g. Text)
2. Right-click on any of the Toolbar buttons (e.g. alignment menu)
3. See the custom context menu opening.

## Screenshots
Screenshot 2019-10-28 at 18 10 04

## Additional context

 - WordPress version:
 - Plugin version: `develop`
 - Gutenberg plugin version (if applicable): `develop`
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* After right-clicking on a block, the context menu should open 
* After right-clicking on something other than a block, the context menu should not open
   * e.g. Toolbar

## Implementation brief

* Use [compareDocumentPosition](https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition) to check if element is user right clicks on top toolbar. See the change [here](https://github.com/ampproject/amp-wp/compare/fix/right-click-toolbar?expand=1).

## QA testing instructions

1. Add a block (e.g. Text)
2. Right-click on any of the Toolbar buttons (e.g. alignment menu)
3. Confirm, browser context menu shows and not story editor context menu.

## Demo

Screenshot from 2019-10-29 18-24-55

## Changelog entry

*

spacedmonkey:
## Summary

Fixes #3647

## Checklist


spacedmonkey:
The reason behind 20 check is explained [here](https://blog.absolutedisaster.co.uk/js-compare-document-position/).

assets/src/stories-editor/components/higher-order/with-context-menu.js
```diff
@@ -47,9 +46,14 @@ const applyWithSelect = withSelect( ( select ) => {
 			editLayout.appendChild( menuWrapper );
 		}
 
+		// Make sure that user did not right click topbar.
+		const toolBar = document.querySelector( '.edit-post-header' );
+		if ( toolBar.compareDocumentPosition( event.target ) === 20 ) {
```
barklund: I get why the value 20 is returned, but checking directly for that is irrelevant. Of course any node contained in another node follows it.

I still think the proper check is to do:

```suggestion
		if ( toolBar.compareDocumentPosition( event.target ) & Node.DOCUMENT_POSITION_CONTAINED_BY ) {
```

If you really want to validate, that it's both contained and followed by (not that the fact that it's followed by matters here) you should still use the bit mask pattern:

```suggestion
		const mask = Node.DOCUMENT_POSITION_CONTAINED_BY | Node. DOCUMENT_POSITION_FOLLOWING;
		if ( ( toolBar.compareDocumentPosition( event.target ) & mask ) === mask ) {
```

And not just plain integers. 

However the only important fact here is the `DOCUMENT_POSITION_CONTAINED_BY`, so no need to check for other than that in my opinion.

You could move it to a variable and wrap it in `Boolean()` for increased clarity:

```suggestion
		const isEventInsideToolbar = Boolean( toolBar.compareDocumentPosition( event.target ) & Node.DOCUMENT_POSITION_CONTAINED_BY );
		if ( isEventInsideToolbar ) {
```

spacedmonkey:
```

const isEventInsideToolbar = Boolean( toolBar.compareDocumentPosition( event.target ) & Node.DOCUMENT_POSITION_CONTAINED_BY );
--
  | if ( isEventInsideToolbar ) {

```
I get these errors


51:41  error  Unexpected use of '&'  no-bitwise
  51:91  error  'Node' is not defined  no-undef

barklund: That rule should clearly be disabled for a use-case like this.

Create [a helper function akin to this one](https://github.com/StephenGrider/MeteorCasts/blob/master/markbin/node_modules/fbjs/lib/containsNode.js.flow), it contains the eslint disable as well. [It's pretty common for this use-case - even used in React core](https://www.codota.com/code/javascript/functions/builtins/containsNode).

I can live with 16 being directly used in the code, though I feel it's a bit of [a magic number](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad). `window.Node.DOCUMENT_POSITION_CONTAINED_BY` would probably be the correct notation.

barklund: I can live with that too! I don't know if there might be issues with text nodes being event targets (the linked helper function has specific workarounds for those), but if it works, it works.

spacedmonkey: Changed in https://github.com/ampproject/amp-wp/pull/3661/commits/d85df416c880a8b63b20ad84e459c1f23a9fa1f5

swissspidy:
Curious to learn why this happened. Is this an issue for the sidebar as well?

If so, instead of checking `isEventInsideToolbar` couldn't we just reverse the logic and check `isEventInsideCurrentPage` / `isEventInsideCurrentBlock` or something?

barklund:
> Curious to learn why this happened. Is this an issue for the sidebar as well?

I wondered the same thing too. It seems that the alignment dropdown in the top toolbar is somehow a direct virtual dom child node of the block, but the sidebar components or anything else aren't.

Screen Shot 2019-10-31 at 2 23 03 PM

I think checking that the click isn't in the top toolbar covers all necessary checks (for now at least).

swissspidy:
I see. Interesting!

PR #3678, opened 2019-10-31
Remove handling of Jetpack shortcodes

### LINKED ISSUE #3309: Migrate Jetpack shortcode handling out of AMP plugin and into Jetpack

kienstra:
## Summary
This plugin's [handling](https://github.com/ampproject/amp-wp/blob/ea9e6fb9d262e699ea64978794840ba9868715f6/includes/embeds/class-amp-vimeo-embed.php#L71) of Jetpack shortcodes, like `[vimeo]` and `[tweet]`, is possibly being [moved to Jetpack](https://github.com/ampproject/amp-wp/issues/3309). 

So this PR looks for the presence of a new filter that Jetpack should have. If that's present, this plugin doesn't run the logic for that shortcode.

Related to #3309

## Checklist

- [x] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests). (It's not really tested, but it's almost entirely deletions. So I don't think more tests are needed.)


includes/embeds/class-amp-soundcloud-embed.php
```diff
@@ -23,9 +30,8 @@ class AMP_SoundCloud_Embed_Handler extends AMP_Base_Embed_Handler {
 	 * Register embed.
 	 */
 	public function register_embed() {
-		if ( function_exists( 'soundcloud_shortcode' ) ) {
-			// @todo Move this to Jetpack.
-			add_shortcode( 'soundcloud', [ $this, 'shortcode' ] );
+		if ( ! $this->is_amp_shortcode_available_in_jetpack( self::SHORTCODE_TAG ) ) {
```
westonruter: Instead of this method, wouldn't it be the same to just do `! class_exists( 'Jetpack_AMP_Soundcloud_Shortcode' )`?  This code will quickly become dead code that should be removed entirely from the AMP plugin, once a new version of Jetpack has been released, so we don't need to worry too much about keeping this code around too long.

includes/embeds/class-amp-soundcloud-embed.php
```diff
@@ -23,9 +30,8 @@ class AMP_SoundCloud_Embed_Handler extends AMP_Base_Embed_Handler {
...
+			add_shortcode( self::SHORTCODE_TAG, [ $this, 'shortcode' ] );
```
westonruter: Please mark the `shortcode` methods as `@deprecated`.

includes/embeds/class-amp-twitter-embed.php
```diff
@@ -46,7 +53,9 @@ class AMP_Twitter_Embed_Handler extends AMP_Base_Embed_Handler {
 	 * Registers embed.
 	 */
 	public function register_embed() {
-		add_shortcode( 'tweet', [ $this, 'shortcode' ] ); // Note: This is a Jetpack shortcode.
+		if ( ! $this->is_amp_shortcode_available_in_jetpack( self::SHORTCODE_TAG ) ) {
```
westonruter: See notes above for Soundcloud.

includes/embeds/class-amp-vimeo-embed.php
```diff
@@ -50,7 +67,9 @@ public function __construct( $args = [] ) {
 	 */
 	public function register_embed() {
 		wp_embed_register_handler( 'amp-vimeo', self::URL_PATTERN, [ $this, 'oembed' ], -1 );
-		add_shortcode( 'vimeo', [ $this, 'shortcode' ] );
+		if ( ! $this->is_amp_shortcode_available_in_jetpack( self::SHORTCODE_TAG ) ) {
```
westonruter: See notes above for Soundcloud.

tests/php/test-amp-vimeo-embed.php
```diff
@@ -12,6 +12,15 @@
  */
 class AMP_Vimeo_Embed_Test extends WP_UnitTestCase {
 
+	/**
+	 * Tears down the environment after each test.
+	 *
+	 * @inheritDoc
+	 */
+	public function tearDown() {
+		remove_all_filters( 'do_shortcode_tag' );
```
westonruter: I believe this isn't necessary because filters get reset after each test anyway.

tests/php/test-amp-vimeo-embed.php
```diff
@@ -109,4 +118,23 @@ public function test__get_scripts( $source, $expected ) {
 
 		$this->assertEquals( $expected, $scripts );
 	}
+
+	/**
+	 * Test is_amp_shortcode_available_in_jetpack.
+	 *
+	 * @covers AMP_Base_Embed_Handler::is_amp_shortcode_available_in_jetpack()
+	 */
+	public function test_is_amp_shortcode_available_in_jetpack() {
```
westonruter: This test may not be necessary. I think we can rather just check if the `Jetpack_AMP_Vimeo_Shortcode` class exists.

tests/php/test-amp-vimeo-embed.php
kienstra: OK, sounds good.

includes/embeds/class-amp-soundcloud-embed.php
kienstra: Sure, https://github.com/ampproject/amp-wp/pull/3678/commits/3e0e649706533cbe6fbf96cb5c63710f32887d93 does that.

tests/php/test-amp-vimeo-embed.php
kienstra: https://github.com/ampproject/amp-wp/pull/3678/commits/3e0e649706533cbe6fbf96cb5c63710f32887d93#diff-58db4548e89d6f4290514ec825c906ecL15 removes this `tearDown()` method.

includes/embeds/class-amp-vimeo-embed.php
kienstra: Sure, https://github.com/ampproject/amp-wp/pull/3678/commits/b64392bf469255d7573e074a8bc0d0d1f1dc14c8 changes this

includes/embeds/class-amp-twitter-embed.php
kienstra: Sure, applied in https://github.com/ampproject/amp-wp/pull/3678/commits/b64392bf469255d7573e074a8bc0d0d1f1dc14c8

includes/embeds/class-amp-soundcloud-embed.php
kienstra: Sure, https://github.com/ampproject/amp-wp/pull/3678/commits/b64392bf469255d7573e074a8bc0d0d1f1dc14c8 checks whether the class exists.

includes/embeds/class-amp-soundcloud-embed.php
```diff
@@ -116,7 +115,7 @@ private function parse_amp_component_from_iframe( $html, $url ) {
 	/**
 	 * Render shortcode.
 	 *
-	 * @todo Move this to Jetpack.
+	 * @deprecated 1.4.1 Moved to Jetpack in soundcloud_shortcode().
```
kienstra: With the deletion of `add_shortcode( ... )` above, this function is dead code. Maybe it'd be good to delete it entirely.

The [PR](https://github.com/Automattic/jetpack/pull/14028) to migrate Soundcloud shortcode to Jetpack doesn't add a function that this can detect, like:

```
if ( ! function_exists( 'jetpack_soundcloud_amp_shortcode' ) ) {
        add_shortcode( 'soundcloud', [ $this, 'shortcode' ] );
}

File: includes/embeds/class-amp-twitter-embed.php [hunk 8]
```diff
@@ -28,6 +28,13 @@ class AMP_Twitter_Embed_Handler extends AMP_Base_Embed_Handler {
 	 */
 	const URL_PATTERN_TIMELINE = '#https?:\/\/twitter\.com(?:\/\#\!\/|\/)(?P<username>[a-zA-Z0-9_]{1,20})(?:$|\/(?P<type>likes|lists)(\/(?P<id>[a-zA-Z0-9_-]+))?)#i';
 
+	/**
+	 * The tag (name) of the shortcode.
+	 *
+	 * @var string
+	 */
+	const SHORTCODE_TAG = 'vimeo';
```
jeherve: This should probably be `tweet`.

kienstra: Ah, good point.

kienstra: Addressed in https://github.com/ampproject/amp-wp/pull/3678/commits/ec5257ee50907b44281e8405918aa2fe1548aaf6. It turns out that the constant wasn't needed at all 😄

kienstra:
The [Jetpack PRs](https://github.com/ampproject/amp-wp/issues/3309#issuecomment-553989965) with the migrated shortcode callbacks should be released in [8.0.0](https://github.com/Automattic/jetpack/milestone/138). So this PR probably shouldn't be released until that is.

westonruter:
Excellent. Let's snooze this issue until `8.0.0` is released.

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -51,7 +51,6 @@ public function __construct( $args = [] ) {
 	 */
 	public function register_embed() {
 		add_filter( 'embed_oembed_html', [ $this, 'filter_embed_oembed_html' ], 10, 2 );
-		add_shortcode( 'youtube', [ $this, 'shortcode' ] ); // @todo Deprecated. See <https://github.com/ampproject/amp-wp/issues/3309>.
 		add_filter( 'wp_video_shortcode_override', [ $this, 'video_override' ], 10, 2 );
```
westonruter: This `video_override` method needs to be restored.

kienstra: Ah, good point. I'll do that.

kienstra:
It looks like I didn't resolve the conflicts very well.

includes/embeds/class-amp-youtube-embed-handler.php
kienstra: It's restored in https://github.com/ampproject/amp-wp/pull/3678/commits/509704014bca92220b4c7240faf213981357878c

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -257,10 +257,14 @@ public function video_override( $html, $attr ) {
 		}
 		$src             = $attr['src'];
 		$youtube_pattern = '#^https?://(?:www\.)?(?:youtube\.com/watch|youtu\.be/)#';
-		if ( 1 === preg_match( $youtube_pattern, $src ) ) {
-			return $this->shortcode( [ $src ] );
```
kienstra: I moved the small amount of logic from `shortcode()` that this used into this function, so `shortcode()` can be `@deprecated`.

kienstra:
>However, there is one left which needs some more work from you, namely AMP_Vimeo_Embed_Handler::video_override()

Good catch. That method's call of `shortcode()` [is now replaced with](https://github.com/ampproject/amp-wp/pull/3678/commits/fdb463e531a2a47417d922df83542b99ddc9bb99) the small amount of logic from `shortcode()` that it used.

kienstra:
The `[video]` shortcodes for YouTube and Vimeo work again.

For example:

```html
[video src="https://www.youtube.com/watch?v=NeXQEJNWO5w"]
[video src="https://vimeo.com/45879568"]
```

includes/embeds/class-amp-twitter-embed-handler.php
```diff
@@ -64,11 +64,31 @@ public function unregister_embed() {
...
+		$this->did_convert_elements = true;
+		return AMP_HTML_Utils::build_tag(
+			$this->amp_tag,
+			[
+				'data-tweetid' => $id,
+				'layout'       => 'responsive',
+				'width'        => $this->args['width'],
+				'height'       => $this->args['height'],
+			]
+		);
```
westonruter: I don't think we need this method anymore, do we? Can't it just be emptied out, as I did in: e168813e521d0138383499a84632bc8637ae6727

```php
	/**
	 * Render oEmbed.
	 *
	 * @deprecated Since 1.1 as now the sanitize_raw_embeds() is used exclusively, allowing the
	 *             original oEmbed response to be wrapped with `amp-twitter`.
	 *
	 * @see \WP_Embed::shortcode()
	 *
	 * @return string Rendered oEmbed.
	 */
	public function oembed() {
		_deprecated_function( __METHOD__, '1.1' );
		return '';
	}
```

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -124,17 +124,49 @@ private function parse_props( $html, $url, $video_id ) {
...
 	 * @return string Rendered oEmbed.
 	 */
-	public function oembed() {
+	public function oembed( $matches, $attr, $url ) {
+		unset( $matches, $attr );
 		_deprecated_function( __METHOD__, '1.5.0' );
-		return '';
+
+		return $this->shortcode( [ $url ] );
 	}
```
westonruter: Since these methods aren't used anymore, can't they just be made to return an empty string while having the `_deprecated_function()` calls? 

We might as well just remove them altogether because they aren't being used.

includes/embeds/class-amp-twitter-embed-handler.php
westonruter: Actually, we might as well just remove it entirely, and it is not being used. It has been deprecated since 1.1.

includes/embeds/class-amp-youtube-embed-handler.php
westonruter: Can't we just remove it altogether? If methods aren't being referenced anymore, then we should try to remove them altogether.

includes/embeds/class-amp-twitter-embed-handler.php
kienstra: Good point, it's now removed in https://github.com/ampproject/amp-wp/pull/3678/commits/6ae3419ae2d2c7aae7f93fee751d4f368441b115.

includes/embeds/class-amp-youtube-embed-handler.php
kienstra: Good point, https://github.com/ampproject/amp-wp/pull/3678/commits/4726922a1788e0d034e6bf5c955051c653d0231b removes `AMP_YouTube_Embed_Handler::shortcode()` and `oembed()`, as they're not used anywhere.

PR #3697, opened 2019-11-07
Add AMP on protected posts

### LINKED ISSUE #3428: Add support for forcing AMP on protected posts
## Feature

Presently password-protected posts are excluded from AMP processing. However as `post_password_required` is used to determine if a post should be processed, this behaviour is only temporary. Should the `wp-postpass_*` be present & valid for the current post/page then processing occurs without issue.

I would like to propose making available a filter which would allow theme authors to have more control over whether something should be processed as AMP.

The rationale for this is to be able to better support protected posts when running a site as a **PWA** with **AMP Shadow Doc**.

Consider the code below. If one, or both where available it would provide a greater degree of control & negate the need for registering a filter for `post_password_required` to workaround this.

**amp-helper-functions.php**

```php

    $errors = AMP_Post_Type_Support::get_support_errors( $post );
    $errors = apply_filters( 'amp_post_supports_amp', $errors );

	return 0 === count( $errors );
}
```

**AMP_Post_Type_Support:get_support_errors**

```php
function get_support_errors( $post ) {

    ......

    return apply_filters( 'amp_get_support_errors', $errors );
}
```

Should the team see value in such a proposal I'd be more than happy to implement.

Alternatively if there is something far more suitable & presently available it would be great to know.

Thanks.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

- Password protected post no longer disables AMP
- Password form is now shown for Reader mode

Fixes #3428.

## Checklist

### DIFF
--- a/includes/class-amp-post-type-support.php
+++ b/includes/class-amp-post-type-support.php
@@ -88,8 +88,9 @@ class AMP_Post_Type_Support {
 			$errors[] = 'post-type-support';
 		}
 
-		if ( post_password_required( $post ) ) {
-			$errors[] = 'password-protected';
+		// Show password form instead of the post content.
+		if ( ! current_theme_supports( 'amp' ) && post_password_required( $post ) ) {
+			add_filter( 'the_content', [ __CLASS__, 'show_password_form' ] );
 		}
 
 		/**
@@ -152,5 +153,16 @@ class AMP_Post_Type_Support {
 			}
 		}
 		return $errors;
+	}
+
+	/**
+	 * Show the password form if the post password is required.
+	 *
+	 * @return string
+	 */
+	public static function show_password_form() {
+		$post = get_post();
+
+		return get_the_password_form( $post );
 	}
 }
--- a/includes/class-amp-post-type-support.php
+++ b/includes/class-amp-post-type-support.php
@@ -88,11 +88,6 @@ class AMP_Post_Type_Support {
 			$errors[] = 'post-type-support';
 		}
 
-		// Show password form instead of the post content.
-		if ( ! current_theme_supports( 'amp' ) && post_password_required( $post ) ) {
-			add_filter( 'the_content', [ __CLASS__, 'show_password_form' ] );
-		}
-
 		/**
 		 * Filters whether to skip the post from AMP.
 		 *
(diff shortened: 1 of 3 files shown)


includes/class-amp-post-type-support.php
```diff
@@ -88,8 +88,9 @@ public static function get_support_errors( $post ) {
 			$errors[] = 'post-type-support';
 		}
 
-		if ( post_password_required( $post ) ) {
-			$errors[] = 'password-protected';
+		// Show password form instead of the post content.
+		if ( ! current_theme_supports( 'amp' ) && post_password_required( $post ) ) {
+			add_filter( 'the_content', [ __CLASS__, 'show_password_form' ] );
```
westonruter: I don't think this isn't really the right place for this filter to be added. Adding a filter is introducing global side effects which don't seem necessarily expected.

westonruter: I think this can be simplified quite a bit: 8519e4b

includes/templates/class-amp-post-template.php
```diff
@@ -322,7 +322,7 @@ private function build_post_comments_data() {
 	 */
 	private function build_post_content() {
 		$amp_content = new AMP_Content(
-			$this->post->post_content,
+			post_password_required( $this->post ) ? get_the_password_form( $this->post ) : $this->post->post_content,
```
westonruter: This is now somewhat like what [`get_the_content()`](https://developer.wordpress.org/reference/functions/get_the_content/) is doing:

```php
	// If post password required and it doesn't match the cookie.
	if ( post_password_required( $_post ) ) {
		return get_the_password_form( $_post );
	}
```

includes/class-amp-post-type-support.php
pierlon: Ah, that is indeed a much better solution. Thanks :ok_hand:

includes/templates/class-amp-post-template.php
pierlon: Thanks for simplifying the password form logic, @westonruter.

pierlon:
Thanks for simplifying the password form logic, @westonruter.

PR #3701, opened 2019-11-08
Remove redundant data-ampdevmode attributes from admin bar descendants

westonruter:
## Summary

I [found out](https://github.com/ampproject/amphtml/issues/24987#issuecomment-551350675) from @Gregable that `data-ampdevmode` applies to the entire descendant DOM tree when it is added to an element:

> The way it works in the validator is that you add `data-ampdevmode` to the root `<html>` tag first. Then, any other tag that includes `data-ampdevmode` will have any errors suppressed on it **or it's descendants**.
> 
> If you want to do:
> 
> ```
> <html amp data-ampdevmode>
> <head>...</head>
> <body data-ampdevmode>
>   ... Anything goes in here ...
> </body>
> </html>
> ```
> 
> Then that would be consistent with the validator mode. The idea of requiring it twice is that you can decide which parts of the document you want errors for and which parts you don't. You can add it to the body and/or head (*) if you want to just disable all errors.
> 
> (*) This doesn't get rid of the mandatory document-level requirement errors, just the tag-level errors. Missing boilerplate will still be reported for example.

This means we can eliminate adding `data-ampdevmode` to every single element under `#wpadminbar` in addition to adding it to the `#wpadminbar` container element as well.

This is a follow-up on #3187.

## Checklist

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -973,7 +973,6 @@ function amp_get_content_sanitizers( $post = null ) {
 		$dev_mode_xpaths = (array) apply_filters( 'amp_dev_mode_element_xpaths', [] );
 		if ( is_admin_bar_showing() ) {
 			$dev_mode_xpaths[] = '//*[ @id = "wpadminbar" ]';
-			$dev_mode_xpaths[] = '//*[ @id = "wpadminbar" ]//*';
 			$dev_mode_xpaths[] = '//style[ @id = "admin-bar-inline-css" ]';
 		}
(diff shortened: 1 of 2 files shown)


westonruter:
I just realized that this may not have the fully intended effect, as the function to check if an element is in dev mode does not look at its ancestors.

westonruter:
While this is redundant, it does allow the sanitizers to not have to keep track of whether or not they are inside of a subtree of an element that has the `data-ampdevmode` attribute. 

Maybe later.

Issue #2769, opened 2019-07-08
Revisit PHP representation of AMP validator spec


westonruter:
The AMP validator spec is converted from protoascii into PHP via the [`bin/amphtml-update.py`](https://github.com/ampproject/amp-wp/blob/4d47a064dc7e0e833682318a4afe3428a07fc934/bin/amphtml-update.py) Python script. This file tries to only extract the information that is needed for the tag-and-attribute sanitizer, but still, it is large: 451KB—so large in fact that it crashes phpcs: https://github.com/ampproject/amp-wp/pull/2767#issuecomment-509386369. Just requiring this file incurs memory usage of +4MB. We could look at splitting up file by tag so that we only load the data for tags that are actually encountered.

Stepping back a bit further, we should also consider whether the Python script is the best way to extract the validator spec into PHP. It turns out that @fstanis has worked on making available the validator spec in JSON format: https://github.com/ampproject/amphtml/pull/22528. This would mean at the very least that we could rewrite the spec extraction logic in PHP (or even JS) instead of Python (which has the required protobuf library, though there is probably a PHP protobuf library that could have been used instead). In any case, it is much more comfortable to work with JSON than protoascii, as long as there is no loss of fidelity in the [conversion](https://developers.google.com/protocol-buffers/docs/reference/python/google.protobuf.json_format-module), which  is done in [`validator/validator_gen_js.py`](https://github.com/ampproject/amphtml/blob/b66f2fa8dde074f0adadc030316e69c92c47da8d/validator/validator_gen_js.py#L899-L919):

```py
  rules = validator_pb2.ValidatorRules()
  text_format.Merge(open(specfile).read(), rules)
  out.append(json_format.MessageToJson(rules))
```

Another benefit here is this would avoid us having to download the entire amphtml repo, since the entire spec in JSON format is always available at https://cdn.ampproject.org/v0/validator.json 

This JSON file is only ~250KB as opposed to an archive export of the amphtml repo which is 100MB+.

fstanis:
If you go for a JS solution, ampproject/amp-toolbox#377 might be of interest.

It's meant to be a general purpose querying library for the validator rules. It'll be minimal at first, but you're free to extend it to include the functionality you need.

schlessera:
Related #3730

westonruter:
Implemented in https://github.com/ampproject/amp-toolbox-php/pull/100

PR #3705, opened 2019-11-10
Remove scripts for components that were not detected in output buffer

### LINKED ISSUE #3489: Remove duplicate and unused AMP component scripts from the response
## Feature description

If you include more than one copy of an AMP component script on a page, e.g. `amp-form`, then the validator gives a warning:

image

In the same way, if there are scripts on the page which are not used, a validation error is also raised:

image

These issues come up often in the Reader mode templates which don't send the entire templates through the post-processor (which will be fixed in #2202), but it can also happen in Standard/Transitional mode when duplicate scripts are manually output in `wp_head` or `wp_footer` actions (which normally should not be done since the plugin adds them already).

Cf. https://github.com/ampproject/amp-wp/pull/3253#discussion_r326385709

---------------

## Acceptance criteria

* 

## Implementation brief

The [`AMP_Theme_Support::ensure_required_markup()`](https://github.com/ampproject/amp-wp/blob/394b2023aaf7bb83146c3968d5b43e8be194cac7/includes/class-amp-theme-support.php#L1526-L1894) method is passed a list of `$script_handles` for the AMP scripts that are required to be in the page. If those scripts are not already present in the DOM, then they are added:

https://github.com/ampproject/amp-wp/blob/394b2023aaf7bb83146c3968d5b43e8be194cac7/includes/class-amp-theme-support.php#L1730-L1746

There are two things missing here:

1. If there is more than one copy of a given script on the page, the duplicates need to be removed.
2. If there are AMP scripts present on the page which are _not_ present in `$script_handles` then they should be removed, as otherwise there can be an AMP validation error (as they are unused).

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
## Summary

If there are AMP scripts present on the page which are not present in `$script_handles` then they are removed, as otherwise there can be an AMP validation error (as they are unused).

Fixes #3489

## Checklist


schlessera:
@westonruter I had to change a few existing tests to get them to pass, as this now removes any AMP component script that is not detected in the output buffer.

This might have side effects:
- If we do check for a component that a script is associated with.
- If a script is not directly tied to a component (I added special handling for `amp-dynamic-css-class` for now, are there others?).
- If a component is outside of the markup we check. This might happen for reader mode, maybe?

westonruter:
> * If a script is not directly tied to a component (I added special handling for `amp-dynamic-css-class` for now, are there others?).

That's a great question. I'll inquire.

> * If a component is outside of the markup we check. This might happen for reader mode, maybe?

Yes, the featured image for example gets sanitized separately from the content. However, the scripts gleaned from it will get merged with the scripts gleaned from `the_content` processing, so I think it OK.

westonruter:
> * If we do check for a component that a script is associated with.

@schlessera Can you elaborate on this? I'm not sure I understand.

westonruter:
I've [inquired in #validator-discuss](https://amphtml.slack.com/archives/C4J8SLKMH/p1573448835014300) on AMP Slack:

> We're looking to improve the AMP plugin to remove AMP component scripts on the page if the AMP components aren't used on the page. For example, if the `amp-facebook-like` script is on the page, but there is no `<amp-facebook-like>` on the page, there is a validation error. I have two questions:
>
> 1) Why does this apply only to some component scripts but not others? For example I can include `amp-form` script on the page or `amp-video` on the page without there being any `<amp-form>` or `<amp-video>` just fine. But for `amp-facebook-like` it is different.
> 2) Is there an enumeration of the component scripts that do _not_ have associated components (or attributes) expected to be in the document? One example we're aware of is `amp-dynamic-css-classes`. What would the others be? Perhaps we should rather be using an enumerated list of the component scripts that specifically cause validation errors if present in the document without the components/attributes present. How can that list be derived?

westonruter:
FYI: If you rebase against `develop` the build should pass.

includes/class-amp-theme-support.php
```diff
@@ -1745,6 +1745,30 @@ public static function ensure_required_markup( DOMDocument $dom, $script_handles
 			$amp_scripts[ $missing_script_handle ] = AMP_DOM_Utils::create_node( $dom, 'script', $attrs );
 		}
 
+		// Remove scripts that had already been added but couldn't be detected from output buffering.
+		foreach ( array_diff( array_keys( $amp_scripts ), $script_handles ) as $superfluous_script_handle ) {
+			$src = $amp_scripts[ $superfluous_script_handle ]->getAttribute( 'src' );
+
+			// Skip scripts unrelated to AMP.
+			if ( ! $src || 0 !== strpos( $src, 'https://cdn.ampproject.org/' ) ) {
```
westonruter: Nit: This could be made marginally faster via:

```suggestion
			if ( ! $src || 'https://cdn.ampproject.org/' !== substr( $src, 0, 27 ) ) {
```

(No need to search the _entire_ large string when we're just wanting to know if it `startsWith` another string.)

includes/class-amp-theme-support.php
```diff
@@ -1745,6 +1745,30 @@ public static function ensure_required_markup( DOMDocument $dom, $script_handles
...
+			if ( 'https://cdn.ampproject.org/v0.js' === $src ) {
+				continue;
+			}
+
+			$custom_element = $amp_scripts[ $superfluous_script_handle ]->getAttribute( 'custom-element' );
+
+			// Skip dynamic CSS classes script, it has no associated element.
+			if ( 'amp-dynamic-css-classes' === $custom_element ) {
+				continue;
+			}
```
westonruter: Instead of doing this single check here, and depending on what the AMP validator team comes back with regarding any list of components, perhaps `[ 'amp-dynamic-css-classes' ]` should be merged with `$script_handles` when it is passed into `array_diff()`. Then there would be no need for `$custom_element`, as the script handles are the component names.

includes/class-amp-theme-support.php
```diff
@@ -1745,6 +1745,30 @@ public static function ensure_required_markup( DOMDocument $dom, $script_handles
 			$amp_scripts[ $missing_script_handle ] = AMP_DOM_Utils::create_node( $dom, 'script', $attrs );
 		}
 
+		// Remove scripts that had already been added but couldn't be detected from output buffering.
+		foreach ( array_diff( array_keys( $amp_scripts ), $script_handles ) as $superfluous_script_handle ) {
```
westonruter: `array_diff( array_keys( $amp_scripts ), $script_handles )` may need to be `array_intersect()`'ed with a list of components that are actually prone to cause validation errors. Scripts like `amp-video` and `amp-form` don't cause any issue when being included. Perhaps this is due to some having built-in status. We need to find out.

In that case, there would be no need for the `amp-dynamic-css-classes` check below either.

schlessera:
> > * If we do check for a component that a script is associated with.
> 
> @schlessera Can you elaborate on this? I'm not sure I understand.

Sorry, I updated the comment, it was missing a word:

> If we **don't** do check for a component that a script is associated with.

What I mean by that is that the scripts are only added to our list of detected scripts if we actually do check for the component that is associated for that script. If a component is part of the markup that we don't check for (new component, maybe?), then we will also not detect the need for that script, and will consequently strip it if it was already included.

westonruter:
Replies [courtesy](https://amphtml.slack.com/archives/C4J8SLKMH/p1573485482019200) of @honeybadgerdontcare.

----

> Why does this apply only to some component scripts but not others? For example I can include `amp-form` script on the page or `amp-video` on the page without there being any `<amp-form>` or `<amp-video>` just fine. But for `amp-facebook-like` it is different.

The feature to enforce this was released after AMP was released and to enforce it on all previous scripts at that time would have broken many AMP documents. Instead those previous scripts were grandfathered in. More can be read at https://github.com/ampproject/amphtml/blob/master/validator/validator.proto#L430

> Is there an enumeration of the component scripts that do _not_ have associated components (or attributes) expected to be in the document? One example we're aware of is `amp-dynamic-css-classes`. What would the others be? Perhaps we should rather be using an enumerated list of the component scripts that specifically cause validation errors if present in the document without the components/attributes present. How can that list be derived?

These would be declared as requires_usage: NONE.

The list could be derived from parsing the protoascii and looking at the requires_usage portion of an extension_spec. https://github.com/ampproject/amphtml/blob/master/validator/validator.proto#L454

PR #22528 introduced a JSON version of the validator spec that might be useful here. https://github.com/ampproject/amphtml/pull/22528

westonruter:
> PR #22528 introduced a JSON version of the validator spec that might be useful here. [ampproject/amphtml#22528](https://github.com/ampproject/amphtml/pull/22528)

For this see #2769.

> These would be declared as requires_usage: NONE.
> The list could be derived from parsing the protoascii and looking at the requires_usage portion of an extension_spec. https://github.com/ampproject/amphtml/blob/master/validator/validator.proto#L454

This is great. What we can do in the immediate term is update the spec parser Python script to extract the `requires_usage` information alongside where we extract the `extension_spec` info. Then this information would be available for the post-processor to query for the extensions that require a tag to be present and vice-versa.

includes/class-amp-theme-support.php
schlessera: In benchmarks with a 4800 character string, the end result is that these are pretty much exactly the same in terms of performance: http://www.spudsdesign.com/benchmark/index.php?t=strpos4&history

However, the `substr()` variant has a maintainability issue in that it requires you to make changes in two places if the string to check for changes: the string itself, and its character count need to be adapted separately - and you can even get it wrong by miscounting.

Therefore, I think the `strpos()` is the superior one.

westonruter: Yeah, it's fine. My wink was intended to acknowledge the exceedingly marginal performance impact.

includes/amp-helper-functions.php
```diff
@@ -464,38 +464,43 @@ function amp_register_default_scripts( $wp_scripts ) {
...
 
 		$wp_scripts->add(
-			$extension,
+			$extension_spec['name'],
 			$src,
 			[ 'amp-runtime' ],
 			null
 		);
+
+		$wp_scripts->add_data( $extension_spec['name'], 'amp_requires_usage', ! empty( $extension_spec['requires_usage'] ) );
```
westonruter: I'm not super happy with what I did here. The post-processor shouldn't be looking at the `wp_scripts()` object to determine later if this requires usage. It would be better if the `extension_specs` were obtained again during post-processing and the `required_usage` was obtained at that point.

includes/amp-helper-functions.php
```diff
@@ -464,38 +464,43 @@ function amp_register_default_scripts( $wp_scripts ) {
...
 
-	// Get all AMP components as defined in the spec.
-	$extensions = [];
+	// Register all AMP components as defined in the spec.
 	foreach ( AMP_Allowed_Tags_Generated::get_allowed_tag( 'script' ) as $script_spec ) {
-		if ( isset( $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec']['name'], $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec']['version'] ) ) {
-			$versions = $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec']['version'];
-			array_pop( $versions );
-			$extensions[ $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec']['name'] ] = array_pop( $versions );
+		if ( ! isset( $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec'] ) ) {
```
westonruter: The Python parser should probably be smarter about extracting all of the extension specs up front to then provide them in a single call, like `AMP_Allowed_Tags_Generated::get_extension_specs()`.

includes/amp-helper-functions.php
```diff
@@ -464,38 +464,43 @@ function amp_register_default_scripts( $wp_scripts ) {
...
-		 * @todo Revisit once amp-base-carousel is no longer experimental. Add support for obtaining a list of extensions that depend on other extensions to include in the script dependencies when registering below.
-		 */
-		$extensions['amp-carousel'] = '0.1';
-	}
+		$extension_spec = $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec'];
+		$versions       = array_filter(
+			$extension_spec['version'],
+			function ( $version ) {
+				// @todo Why are we including the latest version in the generated spec in the first place? We can just include the highest number version.
+				return 'latest' !== $version;
```
westonruter: And we can omit the `latest` version, and any version other than the largest-versioned number.

includes/sanitizers/class-amp-allowed-tags-generated.php
```diff
@@ -12026,7 +12014,7 @@ class AMP_Allowed_Tags_Generated {
 					'extension_spec' => array(
 						'deprecated_allow_duplicates' => true,
```
westonruter: This seems directly related to the need in #3489 to de-duplicate. In reality, we can just omit this field entirely from the generated spec data, as we can just _always_ remove duplicates.

includes/amp-helper-functions.php
schlessera: I think we should at one point rethink the way the spec is parsed. I've created an issue for that: #3730

includes/amp-helper-functions.php
schlessera: How about even moving the entire registration to a point where we know what is needed and what is not? That would avoid registering scripts we don't need, and you could have all checks against the spec in one place...

westonruter: These scripts are needing to be registered in `WP_Scripts` to ensure that they will be available to enqueue for AMP components used outside of AMP pages. For example, Newspack is using `amp-carousel` in its Carousel block and the block will `wp_enqueue_script( 'amp-carousel' )` which will ensure it works in AMP pages and non-AMP pages alike. (Granted, this is not guaranteed to work prior to Bento AMP, but still, it works most of the time and is better than a blank element.)

includes/amp-helper-functions.php
westonruter: I _love_ that.

westonruter: While we plan out improving how the spec is parsed, I've implemented a `AMP_Allowed_Tags_Generated::get_extension_specs()` helper to improve things in the meantime. See 9956c90e2d27f2183be41471cf42275f0b51585f.

bin/amphtml-update.py
```diff
@@ -211,6 +211,23 @@ def GenerateFooterPHP(out):
...
+	 */
+	public static function get_extension_specs() {
+		$extension_specs = [];
+		foreach ( self::get_allowed_tag( 'script' ) as $script_spec ) {
+			if ( isset( $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec'] ) ) {
+				$extension_specs[ $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec']['name'] ] = $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec'];
+			}
+		}
+		return $extension_specs;
+	}
```
schlessera: No need to process this every time it is requested (which is at least _twice_ per request).
```php
public static function get_extension_specs() {
	static $extension_specs = [];

	if ( ! empty( $extension_specs ) ) {
		return $extension_specs;
	}

	foreach ( self::get_allowed_tag( 'script' ) as $script_spec ) {
		if ( isset( $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec'] ) ) {
			$extension_specs[ $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec']['name'] ] = $script_spec[ AMP_Rule_Spec::TAG_SPEC ]['extension_spec'];
		}
	}

	return $extension_specs;
}
```

westonruter: Good point. Done in 16f35b4.

PR #3717, opened 2019-11-12
Include text content of style element in validation error

westonruter:
## Summary

In https://github.com/ampproject/amp-wp/issues/2326#issuecomment-552726579 it was discovered that an excessive CSS validation error for one `style` element will get conflated with validation error for another `style` element, if the `style` elements don't vary in their attributes. This is because the text content of the `style` element was not being included in the validation error attributes; this was unlike what is being done for inline `script` elements. So this PR ensures that the validation errors are distinct for distinct `style` elements.

This PR also improves the display of the text content by putting the contents in `code` element, and putting it inside of a `details` element that has the byte count.

### Before

image

image

### After

image

Screen Shot 2019-11-11 at 22 53 23

image

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -551,8 +551,8 @@ abstract class AMP_Base_Sanitizer {
 				}
 			}
 
-			// Capture script contents.
-			if ( 'script' === $node->nodeName && ! $node->hasAttribute( 'src' ) ) {
+			// Capture element contents.
+			if ( ( 'script' === $node->nodeName && ! $node->hasAttribute( 'src' ) ) || 'style' === $node->nodeName ) {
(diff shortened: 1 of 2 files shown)


westonruter:
Let me re-clarify. 

In Standard mode if I add a Custom HTML block that has 50KB+ of CSS:

image

I will then get, as expected, the invalid markup kept by default:

image

If I go ahead and mark that as removed, and then add two other 50KB+ `style` elements:

image

Then the other invalid markup is automatically marked as removed, even though in Standard mode it should have defaulted to kept:

image

This doesn't happen because the status of the validation error is being applied to the other validation errors since they all have the same attributes (since the `text` is not included).

What should actually have been the result, and what this PR does, is make the other two `style` elements also validation errors that have markup default to kept:

image

csossi:
Verified in QA:
image

PR #3728, opened 2019-11-14
Validate height, width & layout attributes

### LINKED ISSUE #2146: Percentage units and bogus values are erroneously permitted for width/height attributes

pierlon:
## Summary

As [noted by @westonruter](https://github.com/ampproject/amp-wp/issues/2146#issue-434545734), there are no constraints on the width and height attributes as defined in the [Validator protoascii](https://github.com/ampproject/amphtml/blob/master/validator/validator-main.protoascii#L5331-L5345), as the conditional logic is not feasible for the protoascii format.

This PR extends the method `AMP_Tag_And_Attribute_Sanitizer::sanitize_disallowed_attribute_values_in_node` so that the logic for the supported layouts of a given element are considered, and constraints for the width and height attributes are applied not only according to the layout, but also if a layout is not given.

The logic for validating these attributes is adapted from [validator.js](https://github.com/ampproject/amphtml/blob/ee0ad23d78800959f361a97af1bd0c99d91597f6/validator/engine/validator.js#L3938) of the [amproject/amphtml](/amproject/amphtml) repository.

### Video Cover Block

This PR also fixes the Cover Block when a video is selected as the background; the video in a Cover Block is filtered to get the `fill` layout and to have a `cover` object fit.

### Todo:

- [x] Remove the element if the width or height is invalid
- [x] Remove the element if the width or height does not conform to the specified layout
- [x] Set the width of the element to 'auto' if the width is not already set to 'auto' & the layout is 'fixed_height'

Fixes #2146.

## Checklist

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -916,6 +916,7 @@ function amp_get_content_sanitizers( $post = null ) {
 			'include_manifest_comment' => ( defined( 'WP_DEBUG' ) && WP_DEBUG ) ? 'always' : 'when_excessive',
 		],
 		'AMP_Meta_Sanitizer'              => [],
+		'AMP_Layout_Sanitizer'            => [],
 		'AMP_Tag_And_Attribute_Sanitizer' => [], // Note: This whitelist sanitizer must come at the end to clean up any remaining issues the other sanitizers didn't catch.
 	];
 
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -1348,7 +1348,10 @@ function amp_generate_script_hash( $script ) {
 }
 
 /**
- * Check if an HTML attribute has a value.
+ * Validate that an HTML attribute is empty.
+ *
+ * The function `empty()` can't be used instead as it returns true for the values
+ * `0` (int), `"0"` (string), and `0.0` (float), all of which are valid attribute values.
  *
  * @param string $attribute Attribute value.
  * @return bool
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -1347,6 +1347,16 @@ function amp_generate_script_hash( $script ) {
 	return 'sha384-' . $hash;
 }
 
+/**
+ * Check if an HTML attribute has a value.
+ *
+ * @param string $attribute Attribute value.
+ * @return bool
+ */
+function amp_is_attribute_empty( $attribute ) {
+	return ! isset( $attribute ) || '' === $attribute;
+}
+
 /*
  * The function below is copied from the ramsey/array_column package.
  *
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -1348,19 +1348,6 @@ function amp_generate_script_hash( $script ) {
 	return 'sha384-' . $hash;
 }
 
-/**
- * Validate that an HTML attribute is empty.
- *
- * The function `empty()` can't be used instead as it returns true for the values
- * `0` (int), `"0"` (string), and `0.0` (float), all of which are valid attribute values.
- *
- * @param string $attribute Attribute value.
- * @return bool
- */
-function amp_is_attribute_empty( $attribute ) {
-	return ! isset( $attribute ) || '' === $attribute;
-}
-
 /*
  * The function below is copied from the ramsey/array_column package.
  *
--- a/includes/embeds/class-amp-facebook-embed.php
+++ b/includes/embeds/class-amp-facebook-embed.php
@@ -192,6 +192,7 @@ class AMP_Facebook_Embed_Handler extends AMP_Base_Embed_Handler {
 
 		$attributes = [
 			// The layout sanitizer will convert this to `layout` when being sanitized.
+			// The data attribute needs to be used so that the layout sanitizer will process it.
 			'data-amp-layout' => 'responsive',
 			'width'           => $node->hasAttribute( 'data-width' ) ? $node->getAttribute( 'data-width' ) : $this->DEFAULT_WIDTH,
 			'height'          => $node->hasAttribute( 'data-height' ) ? $node->getAttribute( 'data-height' ) : $this->DEFAULT_HEIGHT,
--- a/includes/embeds/class-amp-facebook-embed.php
+++ b/includes/embeds/class-amp-facebook-embed.php
@@ -195,10 +195,6 @@ class AMP_Facebook_Embed_Handler extends AMP_Base_Embed_Handler {
 			'width'  => $node->hasAttribute( 'data-width' ) ? $node->getAttribute( 'data-width' ) : $this->DEFAULT_WIDTH,
 			'height' => $node->hasAttribute( 'data-height' ) ? $node->getAttribute( 'data-height' ) : $this->DEFAULT_HEIGHT,
 		];
-		if ( '100%' === $attributes['width'] ) {
-			$attributes['layout'] = 'fixed-height';
-			$attributes['width']  = 'auto';
-		}
 
 		$node->removeAttribute( 'data-width' );
 		$node->removeAttribute( 'data-height' );
--- a/includes/embeds/class-amp-facebook-embed.php
+++ b/includes/embeds/class-amp-facebook-embed.php
@@ -191,9 +191,10 @@ class AMP_Facebook_Embed_Handler extends AMP_Base_Embed_Handler {
 	private function create_amp_facebook_and_replace_node( Document $dom, DOMElement $node, $embed_type ) {
 
 		$attributes = [
-			'layout' => 'responsive',
-			'width'  => $node->hasAttribute( 'data-width' ) ? $node->getAttribute( 'data-width' ) : $this->DEFAULT_WIDTH,
-			'height' => $node->hasAttribute( 'data-height' ) ? $node->getAttribute( 'data-height' ) : $this->DEFAULT_HEIGHT,
+			// The layout sanitizer will convert this to `layout` when being sanitized.
+			'data-amp-layout' => 'responsive',
+			'width'           => $node->hasAttribute( 'data-width' ) ? $node->getAttribute( 'data-width' ) : $this->DEFAULT_WIDTH,
+			'height'          => $node->hasAttribute( 'data-height' ) ? $node->getAttribute( 'data-height' ) : $this->DEFAULT_HEIGHT,
 		];
 
 		$node->removeAttribute( 'data-width' );
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -335,14 +335,18 @@ abstract class AMP_Base_Sanitizer {
 			}
 		}
 
-		if ( empty( $attributes['height'] ) ) {
-			unset( $attributes['width'] );
-			$attributes['height'] = self::FALLBACK_HEIGHT;
-		}
-
-		if ( empty( $attributes['width'] ) || '100%' === $attributes['width'] ) {
-			$attributes['layout'] = 'fixed-height';
-			$attributes['width']  = 'auto';
+		if ( isset( $attributes['width'], $attributes['height'] ) && '100%' === $attributes['width'] && '100%' === $attributes['height'] ) {
+			unset( $attributes['width'], $attributes['height'] );
+			$attributes['layout'] = 'fill';
+		} else {
+			if ( empty( $attributes['height'] ) ) {
+				unset( $attributes['width'] );
+				$attributes['height'] = self::FALLBACK_HEIGHT;
+			}
+			if ( empty( $attributes['width'] ) || '100%' === $attributes['width'] ) {
+				$attributes['layout'] = 'fixed-height';
+				$attributes['width']  = 'auto';
+			}
 		}
 
 		return $attributes;
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -249,6 +249,16 @@ abstract class AMP_Base_Sanitizer {
 	}
 
 	/**
+	 * Determine if an attribute is empty.
+	 *
+	 * @param string $attribute Attribute value.
+	 * @return bool True if empty, false if not.
+	 */
+	public function attr_empty( $attribute ) {
+		return ! isset( $attribute ) || '' === $attribute;
+	}
+
+	/**
 	 * Sets the layout, and possibly the 'height' and 'width' attributes.
 	 *
 	 * @param array $attributes {
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -326,9 +326,11 @@ abstract class AMP_Base_Sanitizer {
 				&& '100%' === $attributes['width']
 				&& '100%' === $attributes['height']
 			) {
-				unset( $attributes['style'], $attributes['width'], $attributes['height'] );
+				unset( $attributes['style'], $styles['position'], $attributes['width'], $attributes['height'] );
+				if ( ! empty( $styles ) ) {
+					$attributes['style'] = $this->reassemble_style_string( $styles );
+				}
 				$attributes['layout'] = 'fill';
-				unset( $attributes['height'], $attributes['width'] );
 				return $attributes;
 			}
 		}
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -254,7 +254,7 @@ abstract class AMP_Base_Sanitizer {
 	 * @param string $attribute Attribute value.
 	 * @return bool True if empty, false if not.
 	 */
-	public function attr_empty( $attribute ) {
+	public function attribute_empty( $attribute ) {
 		return ! isset( $attribute ) || '' === $attribute;
 	}
 
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -249,13 +249,13 @@ abstract class AMP_Base_Sanitizer {
 	}
 
 	/**
-	 * Determine if an attribute is empty.
+	 * Determine if an attribute value is empty.
 	 *
-	 * @param string $attribute Attribute value.
+	 * @param string|null $value Attribute value.
 	 * @return bool True if empty, false if not.
 	 */
-	public function attribute_empty( $attribute ) {
-		return ! isset( $attribute ) || '' === $attribute;
+	public function is_empty_attribute_value( $value ) {
+		return ! isset( $value ) || '' === $value;
 	}
 
 	/**
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -254,11 +254,15 @@ abstract class AMP_Base_Sanitizer {
 	 * @param array $attributes {
 	 *      Attributes.
 	 *
+	 *      @type string     $bottom
 	 *      @type int|string $height
-	 *      @type int|string $width
-	 *      @type string     $sizes
-	 *      @type string     $class
 	 *      @type string     $layout
+	 *      @type string     $left
+	 *      @type string     $position
+	 *      @type string     $right
+	 *      @type string     $style
+	 *      @type string     $top
+	 *      @type int|string $width
 	 * }
 	 * @return array Attributes.
 	 */
--- a/includes/sanitizers/class-amp-layout-sanitizer.php
+++ b/includes/sanitizers/class-amp-layout-sanitizer.php
@@ -18,7 +18,7 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 	 */
 	public function sanitize() {
 		$xpath = new DOMXPath( $this->dom );
-		$nodes = $xpath->query( '//*[ @layout or @data-amp-layout ]' );
+		$nodes = $xpath->query( '//*[ ( @data-amp-layout ) and ( @width or @height ) ]' );
 
 		foreach ( $nodes as $node ) {
 			$width  = $node->getAttribute( 'width' );
--- a/includes/sanitizers/class-amp-layout-sanitizer.php
+++ b/includes/sanitizers/class-amp-layout-sanitizer.php
@@ -33,7 +33,7 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 				$node->removeAttribute( 'data-amp-layout' );
 			}
 
-			if ( ! $this->attr_empty( $style ) ) {
+			if ( ! $this->attribute_empty( $style ) ) {
 				$styles = $this->parse_style_string( $style );
 
 				// If both height & width descriptors are 100%, apply fill layout.
--- a/includes/sanitizers/class-amp-layout-sanitizer.php
+++ b/includes/sanitizers/class-amp-layout-sanitizer.php
@@ -36,12 +36,29 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 			if ( ! $this->attribute_empty( $style ) ) {
 				$styles = $this->parse_style_string( $style );
 
-				// If both height & width descriptors are 100%, apply fill layout.
+				/*
+				 * If both height & width descriptors are 100%, or
+				 *    width attribute is 100% and height style descriptor is 100%, or
+				 *    height attribute is 100% and width style descriptor is 100%
+				 * then apply fill layout.
+				 */
 				if (
-					isset( $styles['width'], $styles['height'] ) &&
-					( '100%' === $styles['width'] && '100%' === $styles['height'] )
+					(
+						isset( $styles['width'], $styles['height'] ) &&
+						( '100%' === $styles['width'] && '100%' === $styles['height'] )
+					) ||
+					(
+						( ! $this->attribute_empty( $width ) && '100%' === $width ) &&
+						( isset( $styles['height'] ) && '100%' === $styles['height'] )
+					) ||
+					(
+						( ! $this->attribute_empty( $height ) && '100%' === $height ) &&
+						( isset( $styles['width'] ) && '100%' === $styles['width'] )
+					)
 				) {
 					unset( $styles['width'], $styles['height'] );
+					$node->removeAttribute( 'width' );
+					$node->removeAttribute( 'height' );
 
 					if ( empty( $styles ) ) {
 						$node->removeAttribute( 'style' );
--- a/includes/sanitizers/class-amp-layout-sanitizer.php
+++ b/includes/sanitizers/class-amp-layout-sanitizer.php
@@ -18,7 +18,8 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 	 */
 	public function sanitize() {
 		$xpath = new DOMXPath( $this->dom );
-		$nodes = $xpath->query( '//*[ ( @data-amp-layout ) and ( @width or @height ) ]' );
+		// Elements with the `layout` attribute will be validated by `AMP_Tag_And_Attribute_Sanitizer`.
+		$nodes = $xpath->query( '//*[ not( @layout ) and ( @data-amp-layout or @width or @height ) ]' );
 
 		foreach ( $nodes as $node ) {
 			$width  = $node->getAttribute( 'width' );
--- a/includes/sanitizers/class-amp-layout-sanitizer.php
+++ b/includes/sanitizers/class-amp-layout-sanitizer.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Class AMP_Layout_Sanitizer
+ *
+ * @since 1.5.0
+ * @package AMP
+ */
+
+/**
+ * Class AMP_Layout_Sanitizer
+ *
+ * @since 1.5.0
+ */
+class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
+
+	/**
+	 * Sanitize any element that has the `layout` or `data-amp-layout` attribute.
+	 */
+	public function sanitize() {
+		$xpath = new DOMXPath( $this->dom );
+		$nodes = $xpath->query( '//*[ @layout or @data-amp-layout ]' );
+
+		foreach ( $nodes as $node ) {
+			$width  = $node->getAttribute( 'width' );
+			$height = $node->getAttribute( 'height' );
+
+			// The `layout` attribute can also be defined through the `data-amp-layout` attribute.
+			if ( $node->hasAttribute( 'data-amp-layout' ) ) {
+				$layout = $node->getAttribute( 'data-amp-layout' );
+				$node->setAttribute( 'layout', $layout );
+				$node->removeAttribute( 'data-amp-layout' );
+			}
+
+			// If the width & height are `100%` the layout must be `fill`.
+			if ( '100%' === $width && '100%' === $height ) {
+				$node->removeAttribute( 'width' );
+				$node->removeAttribute( 'height' );
+				$node->setAttribute( 'layout', AMP_Rule_Spec::LAYOUT_FILL );
+				return;
+			}
+
+			// If the width is `100%`, convert the layout to `fixed-height` and width to `auto`.
+			if ( '100%' === $width ) {
+				$node->setAttribute( 'width', 'auto' );
+				$node->setAttribute( 'layout', AMP_Rule_Spec::LAYOUT_FIXED_HEIGHT );
+			}
+		}
+
+		$this->did_convert_elements = true;
+	}
+}
--- a/includes/sanitizers/class-amp-layout-sanitizer.php
+++ b/includes/sanitizers/class-amp-layout-sanitizer.php
@@ -19,6 +19,7 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 	 */
 	public function sanitize() {
 		$xpath = new DOMXPath( $this->dom );
+
 		// Elements with the `layout` attribute will be validated by `AMP_Tag_And_Attribute_Sanitizer`.
 		$nodes = $xpath->query( '//*[ not( @layout ) and ( @data-amp-layout or @width or @height or @style ) ]' );
 
@@ -28,6 +29,8 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 			 *
 			 * @var DOMElement $node
 			 */
+
+			// Layout does not apply inside of noscript.
 			if ( $this->is_inside_amp_noscript( $node ) ) {
 				continue;
 			}
@@ -43,7 +46,7 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 				$node->removeAttribute( 'data-amp-layout' );
 			}
 
-			if ( ! $this->attribute_empty( $style ) ) {
+			if ( ! $this->is_empty_attribute_value( $style ) ) {
 				$styles = $this->parse_style_string( $style );
 
 				/*
@@ -58,11 +61,11 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 						( '100%' === $styles['width'] && '100%' === $styles['height'] )
 					) ||
 					(
-						( ! $this->attribute_empty( $width ) && '100%' === $width ) &&
+						( ! $this->is_empty_attribute_value( $width ) && '100%' === $width ) &&
 						( isset( $styles['height'] ) && '100%' === $styles['height'] )
 					) ||
 					(
-						( ! $this->attribute_empty( $height ) && '100%' === $height ) &&
+						( ! $this->is_empty_attribute_value( $height ) && '100%' === $height ) &&
 						( isset( $styles['width'] ) && '100%' === $styles['width'] )
 					)
 				) {
@@ -102,7 +105,7 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 	 *
 	 * @param DOMElement $node Node.
 	 */
-	private function set_layout_fill( $node ) {
+	private function set_layout_fill( DOMElement $node ) {
 		if ( $node->hasAttribute( 'width' ) && $node->hasAttribute( 'height' ) ) {
 			$node->removeAttribute( 'width' );
 			$node->removeAttribute( 'height' );
--- a/includes/sanitizers/class-amp-layout-sanitizer.php
+++ b/includes/sanitizers/class-amp-layout-sanitizer.php
@@ -19,11 +19,12 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 	public function sanitize() {
 		$xpath = new DOMXPath( $this->dom );
 		// Elements with the `layout` attribute will be validated by `AMP_Tag_And_Attribute_Sanitizer`.
-		$nodes = $xpath->query( '//*[ not( @layout ) and ( @data-amp-layout or @width or @height ) ]' );
+		$nodes = $xpath->query( '//*[ not( @layout ) and ( @data-amp-layout or @width or @height or @style ) ]' );
 
 		foreach ( $nodes as $node ) {
 			$width  = $node->getAttribute( 'width' );
 			$height = $node->getAttribute( 'height' );
+			$style  = $node->getAttribute( 'style' );
 
 			// The `layout` attribute can also be defined through the `data-amp-layout` attribute.
 			if ( $node->hasAttribute( 'data-amp-layout' ) ) {
@@ -32,12 +33,31 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 				$node->removeAttribute( 'data-amp-layout' );
 			}
 
-			// If the width & height are `100%` the layout must be `fill`.
+			if ( ! $this->attr_empty( $style ) ) {
+				$styles = $this->parse_style_string( $style );
+
+				// If both height & width descriptors are 100%, apply fill layout.
+				if (
+					isset( $styles['width'], $styles['height'] ) &&
+					( '100%' === $styles['width'] && '100%' === $styles['height'] )
+				) {
+					unset( $styles['width'], $styles['height'] );
+
+					if ( empty( $styles ) ) {
+						$node->removeAttribute( 'style' );
+					} else {
+						$node->setAttribute( 'style', $this->reassemble_style_string( $styles ) );
+					}
+
+					$this->set_layout_fill( $node );
+					continue;
+				}
+			}
+
+			// If the width & height are `100%` then apply fill layout.
 			if ( '100%' === $width && '100%' === $height ) {
-				$node->removeAttribute( 'width' );
-				$node->removeAttribute( 'height' );
-				$node->setAttribute( 'layout', AMP_Rule_Spec::LAYOUT_FILL );
-				return;
+				$this->set_layout_fill( $node );
+				continue;
 			}
 
 			// If the width is `100%`, convert the layout to `fixed-height` and width to `auto`.
@@ -48,5 +68,21 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 		}
 
 		$this->did_convert_elements = true;
+	}
+
+	/**
+	 * Apply the `fill` layout.
+	 *
+	 * @param DOMElement $node Node.
+	 */
+	private function set_layout_fill( $node ) {
+		if ( $node->hasAttribute( 'width' ) && $node->hasAttribute( 'height' ) ) {
+			$node->removeAttribute( 'width' );
+			$node->removeAttribute( 'height' );
+		}
+
+		if ( AMP_Rule_Spec::LAYOUT_FILL !== $node->getAttribute( 'layout' ) ) {
+			$node->setAttribute( 'layout', AMP_Rule_Spec::LAYOUT_FILL );
+		}
 	}
 }
--- a/includes/sanitizers/class-amp-layout-sanitizer.php
+++ b/includes/sanitizers/class-amp-layout-sanitizer.php
@@ -12,6 +12,7 @@
  * @since 1.5.0
  */
 class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
+	use AMP_Noscript_Fallback;
 
 	/**
 	 * Sanitize any element that has the `layout` or `data-amp-layout` attribute.
@@ -22,6 +23,15 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 		$nodes = $xpath->query( '//*[ not( @layout ) and ( @data-amp-layout or @width or @height or @style ) ]' );
 
 		foreach ( $nodes as $node ) {
+			/**
+			 * Element.
+			 *
+			 * @var DOMElement $node
+			 */
+			if ( $this->is_inside_amp_noscript( $node ) ) {
+				continue;
+			}
+
 			$width  = $node->getAttribute( 'width' );
 			$height = $node->getAttribute( 'height' );
 			$style  = $node->getAttribute( 'style' );
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -646,9 +646,6 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 			}
 		}
 
-		// Fix any values for width & height that may be AMP incompatible.
-		$this->adapt_dimensions_to_layout( $node );
-
 		// Identify attribute values that don't conform to the attr_spec.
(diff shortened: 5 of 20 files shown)


includes/validation/class-amp-css-length.php
```diff
@@ -0,0 +1,136 @@
...
+			return;
+		} elseif ( 'fluid' === $attr_value ) {
+			$this->is_fluid = true;
+			$this->is_valid = $allow_fluid;
+		}
+
+		$pattern = '/^(?P<numeral>\d+(?:\.\d+)?)(?P<unit>px|em|rem|vh|vw|vmin|vmax)?$/';
+		if ( preg_match( $pattern, $attr_value, $match ) ) {
+			$this->is_valid = true;
+			$this->numeral  = isset( $match['numeral'] ) ? floatval( $match['numeral'] ) : $this->numeral;
```
schlessera: `floatval()` is PHP-4-style casting. In PHP 5+, the actual casting operator should be preferred over a function.
```suggestion
			$this->numeral  = isset( $match['numeral'] ) ? (float) $match['numeral'] : $this->numeral;
```

tests/php/test-tag-and-attribute-sanitizer.php
```diff
@@ -1714,6 +1714,26 @@ static function() {
...
+
+			'illegal_width_attribute_with_layout'          => [
+				'<amp-img src="/img1.png" width="50" height="50%" layout="responsive"></amp-img>',
+				'',
+			],
+
+			'illegal_height_attribute_with_layout'         => [
+				'<amp-img src="/img1.png" width="50" height="50%" layout="responsive"></amp-img>',
+				'',
+			],
```
schlessera: I suggest adding the following tests, to cover edge cases:
- a value of `0` => `width="0"`
- an empty value => `width=""`
- multiple contradicting values => `width="10" width="20"` (not sure what the result should be... last one used, or invalidated?)

includes/validation/class-amp-css-length.php
```diff
@@ -0,0 +1,136 @@
+<?php
+/**
+ * Class AMP_CSS_Length
+ *
+ * @package AMP
+ */
+
+/**
+ * Class AMP_CSS_Length
+ *
```
schlessera: Add a reference and link to the JS class that this was adapted from, and at which version.
As we copy behavior from the JS validator, we'll need to make sure we keep it in sync over time.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1098,6 +1117,213 @@ private function sanitize_disallowed_attribute_values_in_node( DOMElement $node,
 		return $attributes_pending_removal;
 	}
 
+	/**
+	 * Ensure the width and height for the element are valid.
+	 *
```
schlessera: For all of the following methods, add a reference and link to the JS code that this was adapted from, and at which version.
As we copy behavior from the JS validator, we'll need to make sure we keep it in sync over time.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1098,6 +1117,213 @@ private function sanitize_disallowed_attribute_values_in_node( DOMElement $node,
...
+			$node->getAttribute( 'height' ),
+			true,
+			AMP_Rule_Spec::LAYOUT_FLUID === $layout_attr
+		);
+
+		if ( ! $input_height->is_valid() ) {
+			return false;
+		}
+
+		if ( isset( $tag_spec['amp_layout'] ) ) {
```
schlessera: The JS validates starts off its `validateLayout()` function with this check as an assert: https://github.com/ampproject/amphtml/blob/b203bbac123337a83752083ac6018fab92393f9f/validator/engine/validator.js#L3940

That means it wouldn't even calculate the `$input_width` and `$input_height` when the `'amp_layout'` spec is not set, so this deviates from the JS validator behavior.

Was there a specific reason why you moved this check to a later point in the code?

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1098,6 +1117,213 @@ private function sanitize_disallowed_attribute_values_in_node( DOMElement $node,
...
+				(
+					AMP_Rule_Spec::LAYOUT_INTRINSIC === $layout ||
+					AMP_Rule_Spec::LAYOUT_RESPONSIVE === $layout
+				) &&
+				$width->get_unit() !== $height->get_unit()
+			) {
+				return false;
+			}
+
+			// RESPONSIVE only allows heights attribute.
```
schlessera: This comment seems to be wrong in the JS code already, as the code doesn't do what the comment claims.
```suggestion
			// Height attribute is only allowed for RESPONSIVE layout.
```

I made an upstream PR to that regard: https://github.com/ampproject/amphtml/pull/25581

includes/validation/class-amp-css-length.php
```diff
@@ -0,0 +1,136 @@
...
+
+	/**
+	 * AMP_CSS_Length constructor.
+	 *
+	 * @param string $attr_value  Attribute value to be parsed.
+	 * @param bool   $allow_auto  Whether or not to allow the 'auto' value as a value.
+	 * @param bool   $allow_fluid Whether or not to allow the 'fluid' value as a value.
+	 */
+	public function __construct( $attr_value, $allow_auto, $allow_fluid ) {
+		if ( empty( $attr_value ) ) {
```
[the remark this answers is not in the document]
pierlon: Yes, thanks for pointing that out.

The `$attr_value` here would be retrieved from the node's attribute through `$node->getAttribute()`, which would either return the value of the attribute, or an empty string.

I think having something like `! isset( $attr_value ) || '' === $attr_value` would be sufficient here.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
pierlon: Ah, I see what you mean now. That assertion is not meant to be placed at a later point in the `is_valid_layout` method, but instead when checking if [the element has the layout attribute set](https://github.com/ampproject/amp-wp/pull/3728/files#diff-ee41b32e64c90445a71328ca93cfa799R1026).

includes/validation/class-amp-css-length.php
```diff
@@ -0,0 +1,142 @@
...
+	protected $unit = 'px';
+
+	/**
+	 * AMP_CSS_Length constructor.
+	 *
+	 * @param string $attr_value  Attribute value to be parsed.
+	 * @param bool   $allow_auto  Whether or not to allow the 'auto' value as a value.
+	 * @param bool   $allow_fluid Whether or not to allow the 'fluid' value as a value.
+	 */
+	public function __construct( $attr_value, $allow_auto, $allow_fluid ) {
```
westonruter: I don't think that passing `$allow_auto` and `$allow_fluid` should be done here. For one thing, it makes the instantiations hard to read, for example:

```php
		$input_height = new AMP_CSS_Length(
			$node->getAttribute( 'height' ),
			true,
			true
		);
```

Secondly, it feels like it's somewhat overloading the constructor by doing both capturing the value while also performing validation.

There is validation is being done in the `AMP_Tag_And_Attribute_Sanitizer::has_valid_dimensions()` method, which I think should take more of the responsibility than what is being done in the constructor here.

tests/php/test-amp-facebook-embed.php
```diff
@@ -184,13 +184,6 @@ public function get_raw_embed_dataset() {
 				'<amp-facebook-comments layout="responsive" width="600" height="400" data-href="https://developers.facebook.com/docs/plugins/comments#configurator" data-numposts="5"></amp-facebook-comments>',
 			],
 
-			'comments_full_width'   => [
-				'
-					<div class="fb-comments" data-href="https://developers.facebook.com/docs/plugins/comments#configurator" data-width="100%" data-numposts="5"></div>
-				',
-				'<amp-facebook-comments layout="fixed-height" width="auto" height="400" data-href="https://developers.facebook.com/docs/plugins/comments#configurator" data-numposts="5"></amp-facebook-comments>',
-			],
-
```
westonruter: This should not have been removed. See https://github.com/ampproject/amp-wp/pull/2776

includes/embeds/class-amp-facebook-embed.php
```diff
@@ -199,10 +199,6 @@ private function create_amp_facebook_and_replace_node( $dom, $node, $embed_type
 			'width'  => $node->hasAttribute( 'data-width' ) ? $node->getAttribute( 'data-width' ) : $this->DEFAULT_WIDTH,
 			'height' => $node->hasAttribute( 'data-height' ) ? $node->getAttribute( 'data-height' ) : $this->DEFAULT_HEIGHT,
 		];
-		if ( '100%' === $attributes['width'] ) {
-			$attributes['layout'] = 'fixed-height';
-			$attributes['width']  = 'auto';
-		}
```
westonruter: Where is this conversion of `100%` width to fixed-height `auto` being done now? This was added specifically to address an issue in https://github.com/ampproject/amp-wp/pull/2776.

The code here actually should be generalized to apply to all sanitizers. Anywhere that `width=100%` is found it should be converted into `layout=fixed-height` and `width=auto`.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1005,13 +1005,33 @@ private function get_disallowed_attributes_in_node( DOMElement $node, $attr_spec
...
 	 */
-	private function sanitize_disallowed_attribute_values_in_node( DOMElement $node, $attr_spec_list, $attributes_pending_removal ) {
+	private function sanitize_disallowed_attribute_values_in_node( DOMElement $node, $tag_spec, $attr_spec_list, $attributes_pending_removal ) {
 		$attrs_to_remove = [];
 
+		if (
+			! $node->hasAttribute( 'layout' ) &&
+			( $node->hasAttribute( 'width' ) || $node->hasAttribute( 'height' ) ) &&
+			isset( $tag_spec['amp_layout'] ) &&
+			! $this->has_valid_dimensions( $node )
```
westonruter: Perhaps the `has_valid_dimensions` method should take the `$node` and `$tag_spec` and perform all this logic.

includes/embeds/class-amp-facebook-embed.php
pierlon: I was under the impression from #2146 that this was a temporary fix, but the sanitizers should remove such cases. Will re-include it.

includes/validation/class-amp-css-length.php
```diff
@@ -0,0 +1,142 @@
...
+		if ( 'auto' === $attr_value ) {
+			$this->is_auto  = true;
+			$this->is_valid = $allow_auto;
+			return;
+		} elseif ( 'fluid' === $attr_value ) {
+			$this->is_fluid = true;
+			$this->is_valid = $allow_fluid;
+		}
+
+		$pattern = '/^(?<numeral>\d+(?:\.\d+)?)(?<unit>px|em|rem|vh|vw|vmin|vmax)?$/';
```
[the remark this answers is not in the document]
pierlon: I think letting it match any string of characters would defeat the purpose of the validation being done here. Also, the `%` wasn't included since its invalid according to the AMP validator:

https://github.com/ampproject/amphtml/blob/1911070201440/validator/engine/validator.js#L3407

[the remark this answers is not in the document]
westonruter: OK, you're right. Limiting to the recognized units is probably right. Nevertheless, we do need to account for `%` units in conversions for `fixed-height` and `fill` layouts. I would think that `AMP_CSS_Length` would be involved for parsing that out.

includes/embeds/class-amp-facebook-embed.php
pierlon: This is now taken care of by `AMP_Tag_And_Attribute_Sanitizer::adapt_dimensions_to_layout()`.

includes/validation/class-amp-css-length.php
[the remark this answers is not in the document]
pierlon: Doesn't need to be. `AMP_Tag_And_Attribute_Sanitizer::adapt_dimensions_to_layout()` takes care of `%` units now. Anything else out of that norm should be removed.

includes/validation/class-amp-css-length.php
pierlon: This was a verbatim of [AMP's validator](https://github.com/ampproject/amphtml/blob/1911070201440/validator/engine/validator.js#L3351), but you do make valid points. Will refactor this.

includes/amp-helper-functions.php
```diff
@@ -1358,6 +1358,16 @@ function amp_generate_script_hash( $script ) {
 	return 'sha384-' . $hash;
 }
 
+/**
+ * Check if an HTML attribute has a value.
+ *
+ * @param string $attribute Attribute value.
+ * @return bool
+ */
+function amp_is_attribute_empty( $attribute ) {
```
westonruter: Is this necessary? Can't we just use `empty()` instead?

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -999,19 +1002,54 @@ private function get_disallowed_attributes_in_node( DOMElement $node, $attr_spec
...
 
+	/**
+	 * Adapt width and height to conform to the AMP layout.
+	 *
+	 * @param DOMElement $node Node.
+	 */
+	private function adapt_dimensions_to_layout( DOMElement $node ) {
+		$layout = $node->getAttribute( 'layout' );
+
+		if ( ! amp_is_attribute_empty( $layout ) ) {
```
westonruter: This could rather use `$node->hasAttribute('layout')`

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -559,11 +559,14 @@ private function process_node( DOMElement $node ) {
 			}
 		}
 
+		// Fix any values for width & height that may be AMP incompatible.
+		$this->adapt_dimensions_to_layout( $node );
```
westonruter: Rather than calling this method on every single node, what about adding an `AMP_Layout_Sanitizer` which is dedicated for the purpose of setting dimensions and layout? This sanitizer could do an XPath query for `//*[ @layout ]` to only get the elements that have the `layout` attribute to begin with. (This would also apply to querying for `data-amp-layout` attribute.)

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
westonruter: Note also that we support a `data-amp-layout` attribute which should get converted into `layout` on AMP pages. This is intended to allow authors to define the attribute on HTML pages without adding unrecognized HTML5 attributes.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -999,19 +1002,54 @@ private function get_disallowed_attributes_in_node( DOMElement $node, $attr_spec
...
 	 * @param DOMAttr[]  $attributes_pending_removal Attributes pending removal.
 	 * @return DOMAttr[]|false Attributes to remove, or false if the element itself should be removed.
 	 */
-	private function sanitize_disallowed_attribute_values_in_node( DOMElement $node, $attr_spec_list, $attributes_pending_removal ) {
+	private function sanitize_disallowed_attribute_values_in_node( DOMElement $node, $tag_spec, $attr_spec_list, $attributes_pending_removal ) {
 		$attrs_to_remove = [];
 
+		if ( ! $this->is_valid_layout( $tag_spec, $node ) ) {
+			// If the layout is invalid, the node should be removed.
+			$this->remove_node( $node );
```
westonruter: Should the node be removed or just the _attribute_ be removed (adding `layout` attribute to `$attrs_to_remove`)? It should only be removed if `layout` is mandatory, per below.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -999,19 +1002,54 @@ private function get_disallowed_attributes_in_node( DOMElement $node, $attr_spec
...
+	 */
+	private function adapt_dimensions_to_layout( DOMElement $node ) {
+		$layout = $node->getAttribute( 'layout' );
+
+		if ( ! amp_is_attribute_empty( $layout ) ) {
+			$width  = $node->getAttribute( 'width' );
+			$height = $node->getAttribute( 'height' );
+
+			// If the width & height are `100%` the layout must be `fill`.
+			if ( '100%' === $width && '100%' === $height ) {
```
westonruter: This condition and the condition below need to run even when `layout` is not set, as they imply the required `layout`.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1098,6 +1136,227 @@ private function sanitize_disallowed_attribute_values_in_node( DOMElement $node,
...
+
+		$input_height = new AMP_CSS_Length( $node->getAttribute( 'height' ) );
+		$input_height->validate( $allow_auto, $allow_fluid );
+
+		if ( ! $input_height->is_valid() ) {
+			return false;
+		}
+
+		// If no layout is supplied, there is nothing else to test.
+		if ( amp_is_attribute_empty( $layout_attr ) ) {
```
westonruter: Couldn't this do `$node->hasAttribute( 'layout' )`? 

Should the check not be done at the beginning of this method?

includes/amp-helper-functions.php
pierlon: As noted in https://github.com/ampproject/amp-wp/pull/3728#discussion_r346164920, `empty()` would return true for the values `0` (int), `"0"` (string) and `0.0` (float), which can result in unforeseen bugs in the future.

westonruter: This is fine for the purposes of `layout`, however. If there is `layout=0` then this seems fine to consider as if it is empty; it will get removed by the tag-and-attribute sanitizer.

I'm just wary off adding new functions to the global namespace unless absolutely needed.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
westonruter: I recall also discussing moving this logic to `\AMP_Base_Sanitizer::set_layout()`.

In general, however, the `AMP_Tag_And_Attribute_Sanitizer` should be considered purely used for validation and sanitization. It is the last check to ensure valid AMP. So it can _remove_ elements and attributes, but in general it shouldn't be using to _add_ things to the DOM. This would be better suited for other sanitizers dedicated for that purpose.

includes/amp-helper-functions.php
pierlon: Ah I see, I'll refactor this.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
pierlon: I'd say it should be removed. For example, an `amp-img` with `width=""` and `layout="responsive"` has a valid width and layout on its own, but because the layout is not mandatory, it will be considered valid markup.

westonruter: I don't understand. Given markup like this:

```
<amp-img src="https://example.com/foo.jpg" width="100" height="200" layout="bogus"></amp-img>
```

The `layout` attribute is not valid, so only the _attribute_ should be removed. If this ends up being a mandatory attribute, then the entire element can be removed as well, but it shouldn't be done here.

westonruter:
With the introduction of layout validation, this redundant/obsolete logic should get removed:

https://github.com/ampproject/amp-wp/blob/be6781844c0ed6acd66c9a7ad7367be458a11666/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L536-L545

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
pierlon: Sorry my previous comment was a bit mangled. What I meant to say:

Given markup like this:

```html
<amp-img src="https://example.com/foo.jpg" width="" height="200" layout="responsive"></amp-img>
```

The layout is indeed invalid as there is no specified width, but as an attribute, the width itself is valid. Because there is no spec to validate the width, it will not be sanitized and the markup will be outputted.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
pierlon: Will be done at the beginning.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
westonruter: OK, well then the change to `sanitize_disallowed_attribute_values_in_node()` here should be reverted in favor of something like:

```diff
diff --git a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
index 14e5cc00..a0c6b3bd 100644
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -562,6 +562,11 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 		// Fix any values for width & height that may be AMP incompatible.
 		$this->adapt_dimensions_to_layout( $node );
 
+		// If the layout is invalid, the node should be removed.
+		if ( ! $this->is_valid_layout( $tag_spec, $node ) && $this->remove_invalid_child( $node, [ 'code' => 'invalid_layout' ] ) ) {
+			return null;
+		}
+
 		// Identify any remaining disallowed attributes.
 		$disallowed_attributes = $this->get_disallowed_attributes_in_node( $node, $merged_attr_spec_list );
 
@@ -1044,12 +1049,6 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 	private function sanitize_disallowed_attribute_values_in_node( DOMElement $node, $tag_spec, $attr_spec_list, $attributes_pending_removal ) {
 		$attrs_to_remove = [];
 
-		if ( ! $this->is_valid_layout( $tag_spec, $node ) ) {
-			// If the layout is invalid, the node should be removed.
-			$this->remove_node( $node );
-			return false;
-		}
-
 		foreach ( $attr_spec_list as $attr_name => $attr_val ) {
 			if ( isset( $attr_spec_list[ $attr_name ][ AMP_Rule_Spec::ALTERNATIVE_NAMES ] ) ) {
 				foreach ( $attr_spec_list[ $attr_name ][ AMP_Rule_Spec::ALTERNATIVE_NAMES ] as $attr_alt_name ) {
```

pierlon:
That would still be needed when `AMP_Tag_And_Attribute_Sanitizer::sanitize_disallowed_attribute_values_in_node` does its sanitization.

This is needed to make sure [the layout attributes](https://github.com/ampproject/amp-wp/blob/a782d7e4be11a20e98b68494508d7a7d1e73c50d/includes/sanitizers/class-amp-allowed-tags-generated.php#L17075-L17083) are allowed:
https://github.com/ampproject/amp-wp/blob/a782d7e4be11a20e98b68494508d7a7d1e73c50d/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L537

And this is to ensure that the defined layout is supported by the element:

https://github.com/ampproject/amp-wp/blob/a782d7e4be11a20e98b68494508d7a7d1e73c50d/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L540-L542

includes/sanitizers/class-amp-layout-sanitizer.php
```diff
@@ -0,0 +1,51 @@
...
+				$node->removeAttribute( 'height' );
+				$node->setAttribute( 'layout', AMP_Rule_Spec::LAYOUT_FILL );
+				return;
+			}
+
+			// If the width is `100%`, convert the layout to `fixed-height` and width to `auto`.
+			if ( '100%' === $width ) {
+				$node->setAttribute( 'width', 'auto' );
+				$node->setAttribute( 'layout', AMP_Rule_Spec::LAYOUT_FIXED_HEIGHT );
+			}
```
westonruter: Inferring the `layout=fill` and `layout=fixed-height` need to be done on elements that _don't_ already have the `layout` defined.

In particular, these are needed for converting non-AMP elements (which don't have `layout`) into AMP elements that _do_ have `layout`.

westonruter: In other words, if someone has provided an invalid `layout` attribute for a given width/height, then then this should freely result in an `invalid_layout` validation error. If someone is writing AMP then they need to be told they are doing it wrong.

pierlon: Done in https://github.com/ampproject/amp-wp/pull/3728/commits/18c0833f63a9eebf054ba067e67bef3a585a6252.

includes/sanitizers/class-amp-layout-sanitizer.php
```diff
@@ -0,0 +1,53 @@
...
+			$height = $node->getAttribute( 'height' );
+
+			// The `layout` attribute can also be defined through the `data-amp-layout` attribute.
+			if ( $node->hasAttribute( 'data-amp-layout' ) ) {
+				$layout = $node->getAttribute( 'data-amp-layout' );
+				$node->setAttribute( 'layout', $layout );
+				$node->removeAttribute( 'data-amp-layout' );
+			}
+
+			if ( ! $node->hasAttribute( 'layout' ) ) {
```
westonruter: Does this make sense here? The `$node` at this point is guaranteed to _always_ have a `layout` attribute, so this code will never run.

pierlon: I see two ways to make this happen:

- Put `AMP_Layout_Sanitizer` to be the last sanitizer to run. This ensures that `AMP_Tag_And_Attribute_Sanitizer` would remove elements with invalid layouts before sanitizing the layout itself.

- Keep the current order of the sanitizers and move the logic for validating invalid layouts to `AMP_Layout_Sanitizer`.

Which would you prefer?

pierlon: Implement the first point in https://github.com/ampproject/amp-wp/pull/3728/commits/8a2b94d3ab8b58ef972db3bd1661d3cb9bc67637.

westonruter: I think the validation of the `layout` attributes should remain in the `AMP_Tag_And_Attribute_Sanitizer`, where it removes any elements with invalid `layout`. The `AMP_Layout_Sanitizer` can then be purely focused on fixing up non-AMP HTML to infer the right `layout` based on the HTML.

So I think `AMP_Tag_And_Attribute_Sanitizer` should continue to be the last sanitizer that runs.

includes/sanitizers/class-amp-layout-sanitizer.php
```diff
@@ -0,0 +1,51 @@
...
+ * @since 1.5.0
+ */
+class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
+
+	/**
+	 * Sanitize any element that has the `layout` or `data-amp-layout` attribute.
+	 */
+	public function sanitize() {
+		$xpath = new DOMXPath( $this->dom );
+		$nodes = $xpath->query( '//*[ @layout or @data-amp-layout ]' );
```
westonruter: Given that `layout` is validated in `AMP_Tag_And_Attribute_Sanitizer`, this query needs only to be for `data-amp-layout` to copy it over to `layout`.

includes/sanitizers/class-amp-layout-sanitizer.php
```diff
@@ -0,0 +1,51 @@
...
+
+		foreach ( $nodes as $node ) {
```
westonruter: The query here needs to actually be for any nodes that contain `width` or `height` attributes.

includes/sanitizers/class-amp-layout-sanitizer.php
pierlon: So invalid layouts that can be fixed should be ignored?

includes/sanitizers/class-amp-layout-sanitizer.php
pierlon: This would then mean that the fix for #2146 is to remove the embed, as the layout would be invalid.

includes/sanitizers/class-amp-layout-sanitizer.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/3728/commits/27ac54f7e43e6df22686c28cca9bff7e5269bda1.

includes/sanitizers/class-amp-layout-sanitizer.php
westonruter: For any element that that has `width=100%` we should always convert this into a `fixed-height` layout. I see that the Facebook comment is a special case because it has a `data-width` attribute, not just a `width`. So yes, that conversion would have to go into the Facebook sanitizer. 

But otherwise, if encountering any element with `width=100%` it should be converted into `width=auto layout=fixed-height`. 

But, if the element already has a `layout` attribute specified, then this indicates the author says they are (trying) to write AMP, and in that case, we should go ahead and raise an `INVALID_LAYOUT` error and remove the element. This is because we don't know how to recover from the error (as we could override the `layout` the user supplied), and the author wouldn't be informed of their error. Since the site is attempting to write AMP markup, then we should apply the validation rules and tell them they are doing it wrong.

pierlon: Thanks for the detailed explanation. A simple fix for that Facebook comment could be to simply use `data-layout` when setting the node's attributes, then let the layout sanitizer take care of it:

```diff
diff --git a/includes/embeds/class-amp-facebook-embed.php b/includes/embeds/class-amp-facebook-embed.php
index d10ec8dc..8f5ff75a 100644
--- a/includes/embeds/class-amp-facebook-embed.php
+++ b/includes/embeds/class-amp-facebook-embed.php
@@ -195,7 +195,7 @@ class AMP_Facebook_Embed_Handler extends AMP_Base_Embed_Handler {
        private function create_amp_facebook_and_replace_node( $dom, $node, $embed_type ) {
 
                $attributes = [
-                       'layout' => 'responsive',
+                       'data-layout' => 'responsive',
                        'width'  => $node->hasAttribute( 'data-width' ) ? $node->getAttribute( 'data-width' ) : $this->DEFAULT_WIDTH,
                        'height' => $node->hasAttribute( 'data-height' ) ? $node->getAttribute( 'data-height' ) : $this->DEFAULT_HEIGHT,
                ];
```

westonruter:
@pierlon Please make sure that this scenario is accounted for as reported at https://wordpress.org/support/topic/invalid-layout-specified-by-its-attributes-height/

Namely, code like this:

```html
<div class="embed-vidweb"><iframe id="NATCOM_iframe" src="//vidweb.com/embedcfg/izmqP6DO9n/&amp;idioma=ES&amp;producto=FYI&amp;p=40932" width="100%" height="100%" frameborder="0" scrolling="no" allowfullscreen="allowfullscreen"></iframe></div>
```

When an `iframe` or any element has `width`/`height` of `100%`, either specified via HTML attributes or `style`, that it gets converted to `layout=fill` and the `width`/`height` are removed. You may be accounting for this already in the PR, but I didn't have a chance to check:

```diff
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -304,8 +304,7 @@ abstract class AMP_Base_Sanitizer {
 			}
 
 			// Apply fill layout if width & height are 100%.
-			if ( isset( $styles['position'], $attributes['width'], $attributes['height'] )
-				&& 'absolute' === $styles['position']
+			if ( isset( $attributes['width'], $attributes['height'] )
 				&& '100%' === $attributes['width']
 				&& '100%' === $attributes['height']
 			) {
@@ -321,7 +320,10 @@ abstract class AMP_Base_Sanitizer {
 			$attributes['height'] = self::FALLBACK_HEIGHT;
 		}
 
-		if ( empty( $attributes['width'] ) || '100%' === $attributes['width'] ) {
+		if ( isset( $attributes['width'], $attributes['height'] ) && '100%' === $attributes['width'] && '100%' === $attributes['height'] ) {
+			$attributes['layout'] = 'fill';
+			unset( $attributes['width'], $attributes['height'] );
+		} elseif ( empty( $attributes['width'] ) || '100%' === $attributes['width'] ) {
 			$attributes['layout'] = 'fixed-height';
 			$attributes['width']  = 'auto';
 		}
```

pierlon:
The layout sanitizer wasn't accounting for the dimensions being in `style`, will add that in.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -318,9 +332,11 @@ public function set_layout( $attributes ) {
 				&& '100%' === $attributes['width']
 				&& '100%' === $attributes['height']
 			) {
-				unset( $attributes['style'], $attributes['width'], $attributes['height'] );
+				unset( $attributes['style'], $styles['position'], $attributes['width'], $attributes['height'] );
+				if ( ! empty( $styles ) ) {
+					$attributes['style'] = $this->reassemble_style_string( $styles );
+				}
```
pierlon: Unrelated to this PR, but came across it while perusing this function.

pierlon:
> Please make sure that this scenario is accounted for as reported at https://wordpress.org/support/topic/invalid-layout-specified-by-its-attributes-height/

That's already accounted for :smile:. I also accounted for the dimensions being defined within the inline style in https://github.com/ampproject/amp-wp/pull/3728/commits/a85af7208269dd35d44ba6ded8d4485e88c2631f.

includes/sanitizers/class-amp-base-sanitizer.php
schlessera: This might warrant a test, as the bug went unnoticed.

includes/sanitizers/class-amp-layout-sanitizer.php
```diff
@@ -0,0 +1,88 @@
...
+		// Elements with the `layout` attribute will be validated by `AMP_Tag_And_Attribute_Sanitizer`.
+		$nodes = $xpath->query( '//*[ not( @layout ) and ( @data-amp-layout or @width or @height or @style ) ]' );
+
+		foreach ( $nodes as $node ) {
+			$width  = $node->getAttribute( 'width' );
+			$height = $node->getAttribute( 'height' );
+			$style  = $node->getAttribute( 'style' );
+
+			// The `layout` attribute can also be defined through the `data-amp-layout` attribute.
+			if ( $node->hasAttribute( 'data-amp-layout' ) ) {
```
schlessera: What if there is already a `layout` attribute? Is the `data-amp-layout` attribute always meant to override an existing `layout`?

pierlon: The [XPath query](https://github.com/ampproject/amp-wp/blob/fix/2146-erroneous-attributes/includes/sanitizers/class-amp-layout-sanitizer.php#L22) above ignores elements already with a layout attribute so that won't happen.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1098,6 +1107,216 @@ private function sanitize_disallowed_attribute_values_in_node( DOMElement $node,
...
 	}
 
+	/**
+	 * Validates the layout attributes for the given tag. This involves checking the
+	 * layout, width, height, sizes attributes with AMP specific logic.
+	 *
+	 * @version 1911070201440
+	 * @link https://github.com/ampproject/amphtml/blob/1911070201440/validator/engine/validator.js#L3937
+	 *
+	 * Adapted from the <code>validateLayout</code> method found in validator.js of the
```
schlessera: I don't think we have HTML markup in the docblocks anywhere else, have we? I'd suggest sticking with backticks instead of `<code></code>` and remove the active link, as there's a `@link` annotation directly underneath.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -254,17 +254,31 @@ public function sanitize_dimension( $value, $dimension ) {
 		return '';
 	}
 
+	/**
+	 * Determine if an attribute is empty.
+	 *
+	 * @param string $attribute Attribute value.
+	 * @return bool True if empty, false if not.
+	 */
+	public function attr_empty( $attribute ) {
```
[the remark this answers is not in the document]
pierlon: Yes that's a valid reason, I've amended it in https://github.com/ampproject/amp-wp/pull/3728/commits/6ddf3b8df6903b13dea7ad7ce7627350ca3e9f12.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1098,6 +1107,216 @@ private function sanitize_disallowed_attribute_values_in_node( DOMElement $node,
 		return $attributes_pending_removal;
 	}
 
+	/**
+	 * Validates the layout attributes for the given tag. This involves checking the
+	 * layout, width, height, sizes attributes with AMP specific logic.
```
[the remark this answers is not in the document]
pierlon: Done in https://github.com/ampproject/amp-wp/pull/3728/commits/5253a44989fbd49f2c83539a1e370340bf671af6.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
pierlon: A quick search of the codebase yielded no such cases, so that's good. Removed the markup in  https://github.com/ampproject/amp-wp/pull/3728/commits/08857d9f5bfc314a172a9be98c2f29846d644c96.

includes/sanitizers/class-amp-base-sanitizer.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/3728/commits/bb9c7cdca5992b772630044006c1dc2cf259565f.

includes/sanitizers/class-amp-layout-sanitizer.php
```diff
@@ -0,0 +1,88 @@
...
+					continue;
+				}
+			}
+
+			// If the width & height are `100%` then apply fill layout.
+			if ( '100%' === $width && '100%' === $height ) {
+				$this->set_layout_fill( $node );
+				continue;
+			}
+
```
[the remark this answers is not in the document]
pierlon: My guess is that it would also warrant a layout of `fill`.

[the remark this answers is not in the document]
pierlon: Done in https://github.com/ampproject/amp-wp/pull/3728/commits/e3efc98ac1e18b6916038d98c6f987953cf85248.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -559,6 +559,15 @@ private function process_node( DOMElement $node ) {
...
 
+		// Remove the element if it has a invalid layout attributes.
+		if (
+			! $this->is_valid_layout( $tag_spec, $node ) &&
+			$this->remove_invalid_child( $node, [ 'code' => 'invalid_layout' ] )
+		) {
+			$this->remove_node( $node );
+			return null;
+		}
+
```
westonruter: There is a merge conflict now with this code:

image

Now instead of using `invalid_layout` string, a new error code constant should be introduced. For example:

```diff
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -55,6 +55,7 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 	const INVALID_BLACKLISTED_VALUE_REGEX      = 'INVALID_BLACKLISTED_VALUE_REGEX';
 	const DISALLOWED_PROPERTY_IN_ATTR_VALUE    = 'DISALLOWED_PROPERTY_IN_ATTR_VALUE';
 	const ATTR_REQUIRED_BUT_MISSING            = 'ATTR_REQUIRED_BUT_MISSING';
+	const INVALID_LAYOUT                       = 'INVALID_LAYOUT';
 
 	/**
 	 * Allowed tags.
```

I think it can be resolved as:

```php

		// Remove the element if it has a invalid layout attributes.
		if (
			! $this->is_valid_layout( $tag_spec, $node ) &&
			$this->remove_invalid_child( $node, [ 'code' => self::INVALID_LAYOUT ] )
		) {
			$this->remove_node( $node );
			return null;
		}

		// Identify attribute values that don't conform to the attr_spec.
		$disallowed_attributes = $this->sanitize_disallowed_attribute_values_in_node( $node, $merged_attr_spec_list );
```

westonruter: Actually instead of there being one `INVALID_LAYOUT` error, the `is_valid_layout` method could now return not just a boolean but rather `true` when success and an error array when a failure happens, which will indicate the specific error that is happening. So for example:

```diff
diff --git a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
index dd608d58..7deb7935 100644
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -1142,8 +1142,11 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 		$input_width->validate( $allow_auto, $allow_fluid );
 		$input_height->validate( $allow_auto, $allow_fluid );
 
-		if ( ! $input_width->is_valid() || ! $input_height->is_valid() ) {
-			return false;
+		if ( ! $input_width->is_valid() ) {
+			return [ 'code' => self::INVALID_LAYOUT_WIDTH ];
+		}
+		if ( ! $input_height->is_valid() ) {
+			return [ 'code' => self::INVALID_LAYOUT_HEIGHT ];
 		}
 
 		// No need to go further if there is no layout attribute.
@@ -1161,7 +1164,7 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 
 		// Only FLEX_ITEM allows for height to be set to auto.
 		if ( $height->is_auto() && AMP_Rule_Spec::LAYOUT_FLEX_ITEM !== $layout ) {
-			return false;
+			return [ 'code' => self::INVALID_LAYOUT_AUTO_HEIGHT ];
 		}
 
 		// FIXED, FIXED_HEIGHT, INTRINSIC, RESPONSIVE must have height set.
```

And so on. 

The conditional to remove the element can then be changed to:

```php
// Remove the element if it has a invalid layout attributes.
$layout_validity = $this->is_valid_layout( $tag_spec, $node )
if ( true !== $layout_validity && $this->remove_invalid_child( $node, $layout_validity ) ) {
	return null;
}
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -559,6 +559,15 @@ private function process_node( DOMElement $node ) {
 			}
 		}
 
+		// Remove the element if it has a invalid layout attributes.
+		if (
+			! $this->is_valid_layout( $tag_spec, $node ) &&
+			$this->remove_invalid_child( $node, [ 'code' => 'invalid_layout' ] )
+		) {
+			$this->remove_node( $node );
```
westonruter: Note that `remove_invalid_child` will have already removed the node, so this `$this->remove_node( $node );` should not be here.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
pierlon: That was a fun merge conflict :sweat_smile:. I've added error codes for each invalid layout in https://github.com/ampproject/amp-wp/pull/3728/commits/6a6ab8ce3d509d37d08d6ee8ecb2beb208ac8866,

westonruter:
Note: Test failure should be fixed if pulling latest changes from `develop`.

pierlon:
:warning: Rebasing with latest changes from `develop`.

pierlon:
:warning: Rebasing with latest changes from `develop`.

westonruter:
One tiny merge conflict needs to be resolved. I also need to re-review this.

pierlon:
:warning: Rebased onto `develop`.

westonruter:
@pierlon Now that I've pushed to this branch, please refrain from rebasing as it will cause headaches.

westonruter:
> That's already accounted for 😄. I also accounted for the dimensions being defined within the inline style in [a85af72](https://github.com/ampproject/amp-wp/commit/a85af7208269dd35d44ba6ded8d4485e88c2631f).

Note that this didn't seem to be the case. I've addressed in 99c7b4a.

westonruter:
This PR has dragged on long enough. Great work here. Let's get it into testing to shake out any bugs or things we didn't account for.

westonruter:
@schlessera Please advise whether there are outstanding issues you wanted addressed with this PR prior to merging.

rgllm:
@westonruter any timeline when this will be released?

westonruter:
@rgllm The 1.5 release is due out by end of March to coincide with WordPress 5.4. 

In the meantime, you can use a pre-release version from `develop`: [amp.zip](https://github.com/ampproject/amp-wp/files/4182053/amp.zip) (v1.5.0-alpha-20200210T165753Z-094272913)

PR #3732, opened 2019-11-14
Remove duplicate script entries

schlessera:
## Summary

This adds a test for removing duplicate scripts for now, as the current logic seems to already take care of duplicate scripts in the output-buffered code path.

See #3489

## Checklist

### DIFF
--- a/tests/php/test-class-amp-theme-support.php
+++ b/tests/php/test-class-amp-theme-support.php
@@ -1439,7 +1439,9 @@ class Test_AMP_Theme_Support extends WP_UnitTestCase {
 		ob_start();
 		?>
 		<html>
-			<head></head>
+			<head>
+				<?php wp_print_scripts( [ 'amp-runtime', 'amp-mustache', 'amp-list', 'amp-runtime', 'amp-mustache', 'amp-list' ] ); ?>
+			</head>
 			<body>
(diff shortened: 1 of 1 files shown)


schlessera:
@westonruter Due to the way the scripts are managed, the script handle is the array index, so this is automatically deduplicated.

As you wrote in https://github.com/ampproject/amp-wp/issues/3489#issuecomment-549215627 that we should focus on the output-buffered mode for now and ignore the Reader mode (which is not covered here), this would mean that there's no change to do for fixing this part of the #3489 ticket, only adding the above test for regression testing.

Did I miss something?

tests/php/test-class-amp-theme-support.php
```diff
@@ -1424,6 +1424,49 @@ public function test_scripts_get_moved_to_head() {
...
+		wp();
+		add_theme_support( AMP_Theme_Support::SLUG );
+		AMP_Theme_Support::init();
+		AMP_Theme_Support::finish_init();
+
+		ob_start();
+		?>
+		<html>
+			<head>
+				<?php wp_print_scripts( [ 'amp-runtime', 'amp-mustache', 'amp-list', 'amp-runtime', 'amp-mustache', 'amp-list' ] ); ?>
```
westonruter: WordPress's dependencies system automatically skip printing dependencies that have already been printed, so duplicating handles won't have any effect here. Nor will printing the same thing in the footer have any effect.

PR #3758, opened 2019-11-15
Add AMP_DOM_Document & meta tag sanitizer

### LINKED ISSUE #855: Support WordPress installs with non-UTF-8 charsets
The AMP spec requires `<meta charset="utf-8">` and this is naturally problematic for sites that still use a Latin1 charset. We need to do a few things:

* For existing paired mode theme system, we'll need to use `mb_convert_encoding()` from the `get_bloginfo('charset')` to `utf-8` when reading from the DB. We'll also need to forcibly add `header( 'Content-Type: text/html; charset=utf-8' )` to override what WordPress is sending by default.
* With canonical AMP, we'll be output buffering the response and so we could convert the encoding at the time of output. We'll still need to be wary of `\AMP_DOM_Utils::get_dom_from_content()` in how it is currently assuming UTF-8 content for the input.

https://github.com/Automattic/amp-wp/blob/e60cb152a50e2ffbf5504d41aee859ad1d0e0baa/includes/class-amp-theme-support.php#L192

### LINKED ISSUE #3469: The HTML4 meta[http-equiv=Content-Type] causes validation error to leak and should convert automatically to HTML5 meta[charset]
## Bug Description

As reported in a [support forum topic](https://wordpress.org/support/topic/http-equivcontent-type-attribute-in-meta-tag/), if a theme like [Koji](https://wordpress.org/themes/koji/) declares the `Content-Type` using the HTML4 method like this:

```php
<meta http-equiv="content-type" content="<?php bloginfo( 'html_type' ); ?>" charset="<?php bloginfo( 'charset' ); ?>" />
```

Then the result the AMP plugin generating:

```html
<meta content="text/html; charset=UTF-8" charset="UTF-8" http-equiv="Content-Type">
```

Somehow the validating sanitizer is leaking the `http-equiv` and `content` attributes here, and this causes an error in the validator:

> image

This is even though the sanitizer is reporting the former as being invalid:

> image

Even though it is accepted, it is not getting sanitized. So that's confusing.

The AMP plugin should transparently convert any such `meta` Content-Type tag into the HTML meta charset that AMP requires:

```html
<meta charset=utf-8>
```

See also #2970. This would have avoided an error in adding support for Twenty Twenty (https://github.com/ampproject/amp-wp/issues/3342#issuecomment-536197548) which was fixed in Twenty Twenty via https://github.com/WordPress/twentytwenty/pull/640.

Here is a workaround plugin for the issue: https://gist.github.com/westonruter/d18fd37fb9a58a26ae375a0cb9570831

There should not be any need to show a validation error as this should automatically get converted. If the site is using an encoding other than UTF-8, then technically the encoding should be getting converted to UTF-8 before being parsed. See:

https://github.com/ampproject/amp-wp/blob/603328895e81305718bbe2afd1b913b7f31ac9be/includes/class-amp-theme-support.php#L2290-L2294

This would fix #855.

## Expected Behaviour

An AMP-valid HTML5 meta charset should only be output.

## Steps to reproduce

1. Install and activate Koji theme.
2. Activate Standard or Transitional mode.
3. See the the AMP Validator report a validation error for `http-equiv` attribute.
4. See the admin bar report a validation error for the same.

---------------

## Acceptance criteria

* The `http-equiv` attribute should not be leaked to the frontend.
* Convert HTML4 `meta[http-equiv]` to HTML5 `meta[charset]`.
* (Nice to have) Convert encoding of input documents to UTF-8.

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
## Summary

This PR adds an abstraction for the DOM document and a new sanitizer `AMP_Meta_Sanitizer` that sanitizes meta tags in general, but more specifically for now the charset tag.

- [x] Provide `Dom\Document` to abstract away charset and document structure requirements.
- [x] Provide `Dom\Document::from_html()` to construct from markup (and deprecate `AMP_DOM_Utils::get_dom()`).
- [x] Provide `Dom\Document::from_node()` to construct from a node (and deprecate `AMP_DOM_Utils::get_dom_from_content_node()`).
- [x] Enforce basic HTML markup structure, including `<head>` and `<body>` elements.
- [x] Detect existing charset across possible HTML 4 / HTML 5 combinations.
- [x] Normalize all charsets to HTML 5 `<meta charset="<charset>">` format.
- [x] Ensure a charset is present and add a default one as needed.
- [x] Detect when the AMP requirement for `utf-8` is not met.
- [x] Convert non-UTF-8 encoding to UTF-8.
- [x] Move all relevant implementation specifics and compat code from `AMP_DOM_Utils` to internal `Dom\Document` methods and deprecate accordingly.

Fixes #3469
Fixes #855 

## Checklist


schlessera:
Note to self: don't push into a PR you've initiated but haven't created yet, it will force-create an empty one.

schlessera:
We already have most of the meta tag logic in `AMP_Theme_Support`, but in a rather haphazard way.

Right now, the above code seems to fix the main bug we were chasing, but it introduces duplicate code.

I'll make some more changes now to extract the meta tag handling out of the `AMP_Theme_Support` class and put them into the new sanitizer instead. This will improve the quality of the sanitizers for stand-alone use and simplify the procedural mess that happens in `AMP_Theme_Support`.

schlessera:
Note: the code in the `AMP_Theme_Support` class already states that it would be better off in a sanitizer instead:
https://github.com/ampproject/amp-wp/blob/3b70f6d8e20593920b7b711455e491960b1d0418/includes/class-amp-theme-support.php#L1536

schlessera:
What was supposed to be a small refactor here turned out to be more of a mess than expected.

I started with a `AMP_Meta_Sanitizer` as the gist plugin that solved the initial bug used that, but I'm not happy with the current shape of that sanitizer.

I don't want to spend much more time on refactoring here, as I've already collected higher-level thoughts about this in #3763. I think one principle we should adhere to in future refactorings is to split sanitization (changing from incorrect into correct) from optimization (reordering and similar). The reordering is what not only causes the entire `AMP_Theme_Support::ensure_required_markup()` to be strictly procedural and hard to untangle, but also causes many other headaches when refactoring into more encapsulated code.

schlessera:
There's two open issues here:

**What do we do if we encounter a charset other than `utf-8`?**

Do we try to convert the entire content encoding? Do we work on a known subset only that we can test? Do we just throw an error and force the user to deal with this before trying to use AMP?

**What are the requirements for the viewport?**

As already stated in Slack, the existing tests currently expect `maximum-scale=1.0` to be sanitized away, while I can't find a reference in the documentation that states this as a requirement.

So, given the input `<meta name="viewport" content="maximum-scale=1.0">`, should the output be:
a.) `<meta name="viewport" content="width=device-width">` (what the test expects)
b.) `<meta name="viewport" content="width=device-width,maximum-scale=1.0">` (what I'm currently producing based on how I interpreted the docs.

westonruter:
> **What are the requirements for the viewport?**
> 
> As already stated in Slack, the existing tests currently expect `maximum-scale=1.0` to be sanitized away, while I can't find a reference in the documentation that states this as a requirement.
> 
> So, given the input `<meta name="viewport" content="maximum-scale=1.0">`, should the output be:
> a.) `<meta name="viewport" content="width=device-width">` (what the test expects)
> b.) `<meta name="viewport" content="width=device-width,maximum-scale=1.0">` (what I'm currently producing based on how I interpreted the docs.

Here are the allowed properties as well as the required `device-width=width`:

https://github.com/ampproject/amphtml/blob/e135c18c9460d4834578d6809ae5f77a1e3a231a/validator/validator-main.protoascii#L436-L445

You're right that it `<meta name="viewport" content="maximum-scale=1.0">` should ideally be sanitized as `<meta name="viewport" content="width=device-width,maximum-scale=1.0">`. What is happening now is a bit more clear cut since it detects an invalid `meta` tag and then removes it (while triggering a validation error) and then later the proper one is put in its place. I suppose this could be done equally well with conversion merging… if we encounter a meta `viewport` tag we parse the properties, and if it doesn't include `width=device-width` then we can still trigger the validation error, and if the validation error is sanitized then instead of _removing_ the node we can instead merge the properties.

schlessera:
The PR already does the merging, I just wanted to verify with you whether I got the requirements right because of the broken test.

westonruter:
> **What do we do if we encounter a charset other than `utf-8`?**
> 
> Do we try to convert the entire content encoding? Do we work on a known subset only that we can test? Do we just throw an error and force the user to deal with this before trying to use AMP?

Throwing a warning other than what we're currently doing?

https://github.com/ampproject/amp-wp/blob/f15760234e6b2b6bae81769f2e80077fcf6d657d/includes/class-amp-theme-support.php#L2361-L2365

I think we should try to convert the entire encoding if possible. I was thinking that something like this should be possible (untested code):

```diff
diff --git a/includes/class-amp-theme-support.php b/includes/class-amp-theme-support.php
index 11b7d9a3..fb518e4c 100644
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -2249,6 +2249,16 @@ class AMP_Theme_Support {
 			);
 		}
 
+		// AMP requires UTF-8, so convert encoding.
+		if ( strtolower( get_bloginfo( 'charset' ) ) !== 'utf-8' ) {
+			if ( function_exists( 'mb_convert_encoding' ) ) {
+				$response = mb_convert_encoding( $response, 'utf-8', get_bloginfo( 'charset' ) );
+			} else {
+				/* translators: %s: the charset of the current site. */
+				trigger_error( esc_html( sprintf( __( 'The database has the %s encoding when it needs to be utf-8 to work with AMP.', 'amp' ), get_bloginfo( 'charset' ) ) ), E_USER_WARNING ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+			}
+		}
+
 		$dom   = AMP_DOM_Utils::get_dom( $response );
 		$xpath = new DOMXPath( $dom );
 		$head  = $dom->getElementsByTagName( 'head' )->item( 0 );
@@ -2358,12 +2368,6 @@ class AMP_Theme_Support {
 			}
 		}
 
-		// @todo If 'utf-8' is not the blog charset, then we'll need to do some character encoding conversation or "entityification".
-		if ( 'utf-8' !== strtolower( get_bloginfo( 'charset' ) ) ) {
-			/* translators: %s: the charset of the current site. */
-			trigger_error( esc_html( sprintf( __( 'The database has the %s encoding when it needs to be utf-8 to work with AMP.', 'amp' ), get_bloginfo( 'charset' ) ) ), E_USER_WARNING ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
-		}
-
 		AMP_Validation_Manager::finalize_validation(
 			$dom,
 			[
```

However, last I tried it I don't recall having success. It has been awhile however.

Even so, if UTF-8 is not the encoding, we should still at least issue a warning in Site Health.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -0,0 +1,333 @@
...
+			} elseif ( 'amp-script-src' === $element->getAttribute( 'name' ) ) {
+				$this->meta_tags[ self::TAG_AMP_SCRIPT_SRC ][] = $element;
+			} else {
+				$this->meta_tags[ self::TAG_OTHER ][] = $element;
+			}
+		}
+
+		$this->ensure_charset_is_present( $charset );
+
+		if ( ! $this->is_correct_charset() ) { // phpcs:ignore Generic.CodeAnalysis.EmptyStatement
```
westonruter: I believe this will need to be done prior to the `DOMDocument` being constructed, but it didn't occur to me that perhaps `DOMDocument` allows for encoding to be changed after parsing. I don't think it facilitates this, however.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -0,0 +1,333 @@
...
+		$this->head  = $this->ensure_head_is_present();
+		$charset     = $this->detect_charset();
+		$elements    = $this->dom->getElementsByTagName( static::$tag );
+
+		// Remove all nodes for easy reordering later on.
+		$elements = array_map(
+			static function ( $element ) {
+				return $element->parentNode->removeChild( $element );
+			},
+			iterator_to_array( $elements, false )
```
westonruter: TIL! We could be using this lots of places where we're currently iterating over a `DOMNodeList` to push onto an array which we then iterate over to potentially remove elements. Removing elements from the DOM while iterating over a `DOMNodeList` is problematic since it is a live node list.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -0,0 +1,333 @@
...
+
+		$this->re_add_meta_tags_in_optimized_order();
+	}
+
+	/**
+	 * Ensure that the <head> element is present in the document.
+	 *
+	 * @return DOMElement The document's <head> element.
+	 */
+	protected function ensure_head_is_present() {
```
westonruter: I wonder if this is, in fact, needed, because now we prevent processing responses when no `<head>` is present:

https://github.com/ampproject/amp-wp/blob/f15760234e6b2b6bae81769f2e80077fcf6d657d/includes/class-amp-theme-support.php#L2052-L2058

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -0,0 +1,333 @@
...
+	/**
+	 * Detect the charset of the document.
+	 *
+	 * @return string|false Detected charset of the document, or false if none.
+	 */
+	protected function detect_charset() {
+		$charset = false;
+
+		// Check for HTML 4 http-equiv meta tags.
+		$http_equiv_tag = $this->xpath->query( '//meta[ @http-equiv and @content ]' )->item( 0 );
```
westonruter:
```suggestion
		$http_equiv_tag = $this->xpath->query( '//meta[ translate(@http-equiv, 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz') = 'content-type' and @content ]' )->item( 0 );
```

Alas, `lower-case()` is only in XPath 2.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -0,0 +1,333 @@
...
+
+			// Check for the existence of a proper charset attribute first.
+			$charset = $http_equiv_tag->getAttribute( 'charset' );
+			if ( ! $charset ) {
+				// If not, check whether the charset is included with the content type, and use that.
+				$content = $http_equiv_tag->getAttribute( 'content' );
+
+				$matches = [];
+				if ( preg_match( '/;\s*charset=(?<charset>[^;]+)/', $content, $matches ) && ! empty( $matches['charset'] ) ) {
+					$charset = $matches['charset'];
```
westonruter:
```suggestion
					$charset = trim( $matches['charset'] );
```

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -0,0 +1,333 @@
...
+	 * The charset defaults to utf-8, which is also what AMP requires.
+	 *
+	 * @param string|false $charset Optional. Charset that was already detected. False if none. Defaults to false.
+	 */
+	protected function ensure_charset_is_present( $charset = false ) {
+		if ( ! empty( $this->meta_tags[ self::TAG_CHARSET ] ) ) {
+			return;
+		}
+
+		$this->meta_tags[ self::TAG_CHARSET ][] = $this->create_charset_element( $charset ?: static::AMP_CHARSET );
```
westonruter: I think this can just always use `static::AMP_CHARSET` because this is the only thing that AMP allows (`utf-8`).

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -0,0 +1,333 @@
...
+				break;
+			}
+		}
+
+		if ( ! $width_found ) {
+			array_unshift( $viewport_settings, 'width=device-width' );
+		}
+
+		$viewport_tag->setAttribute( 'content', implode( ',', $viewport_settings ) );
+	}
```
westonruter: This will need to obtain the tag spec for this meta tag essentially incorporate this logic:

https://github.com/ampproject/amp-wp/blob/f15760234e6b2b6bae81769f2e80077fcf6d657d/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L1593-L1652

Otherwise, if someone creates a meta tag like:

```html
<meta name=viewport content="width=content-width,BOGUS=SUPER_BOGUS">
```

Then the tag-and-attribute sanitizer will end up _removing_ this `meta` tag that you created here.

In order to facilitate obtaining the tag spec for `meta name=viewport`, instead of iterating over all `meta` tags it and finding the one that has that `spec_name`, it would be useful to perhaps to refactor the spec generation logic to use the `spec_name` as the array key, like so:

```diff
--- a/includes/sanitizers/class-amp-allowed-tags-generated.php
+++ b/includes/sanitizers/class-amp-allowed-tags-generated.php
@@ -10106,7 +10106,7 @@ class AMP_Allowed_Tags_Generated {
 					'unique' => true,
 				),
 			),
-			array(
+			'meta name=viewport' => array(
 				'attr_spec_list' => array(
 					'content' => array(
 						'mandatory' => true,
@@ -10135,7 +10135,6 @@ class AMP_Allowed_Tags_Generated {
 				'tag_spec' => array(
 					'mandatory' => true,
 					'mandatory_parent' => 'head',
-					'spec_name' => 'meta name=viewport',
 					'spec_url' => 'https://amp.dev/documentation/guides-and-tutorials/learn/spec/amphtml#required-markup',
 					'unique' => true,
 				),
``` 

Then we could extend the `\AMP_Allowed_Tags_Generated::get_allowed_tag()` with a new `$spec_name` arg:

```diff
--- a/includes/sanitizers/class-amp-allowed-tags-generated.php
+++ b/includes/sanitizers/class-amp-allowed-tags-generated.php
@@ -17625,10 +17625,13 @@ class AMP_Allowed_Tags_Generated {
 	 *
 	 * @since 0.7
 	 * @param string $node_name Tag name.
+	 * @param string $spec_name Spec name, to reduce the tag specs to a single one.
 	 * @return array|null Allowed tag, or null if the tag does not exist.
 	 */
-	public static function get_allowed_tag( $node_name ) {
-		if ( isset( self::$allowed_tags[ $node_name ] ) ) {
+	public static function get_allowed_tag( $node_name, $spec_name = null ) {
+		if ( isset( $spec_name, self::$allowed_tags[ $node_name ][ $spec_name ] ) ) {
+			return self::$allowed_tags[ $node_name ][ $spec_name ];
+		} elseif ( isset( self::$allowed_tags[ $node_name ] ) ) {
 			return self::$allowed_tags[ $node_name ];
 		}
 		return null;
```

And then here in the `ensure_viewport_is_present` it could easily locate the tag spec via:

```php
$tag_spec = \AMP_Allowed_Tags_Generated::get_allowed_tag( 'meta', 'meta name=viewport' );
```

This would be useful elsewhere that we refer to tag specs, including in the style sanitizer.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -0,0 +1,333 @@
...
+
+		$charset_element = $this->meta_tags[ self::TAG_CHARSET ][0];
+
+		return static::AMP_CHARSET === strtolower( $charset_element->getAttribute( 'charset' ) );
+	}
+
+	/**
+	 * Re-add the meta tags to the <head> node in the optimized order.
+	 *
+	 * The order is defined by the array entries in $this->meta_tags.
```
westonruter: Note the optimized order of DOM nodes is specified in https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/optimize_amp/

The comments in `\AMP_Theme_Support::ensure_required_markup()` are citing this document.

includes/sanitizers/class-amp-meta-sanitizer.php
schlessera: DOMDocument works with UTF-8 internally, so I think we need to have it converted before sending it to DOMDocument, or otherwise we might already have messed it up via `loadHTML()`.

includes/sanitizers/class-amp-meta-sanitizer.php
schlessera: Yes, the `DOMNodeList` does not seem to have a built-in way to turn itself into an array, but any iterator can be turned into one via `iterator_to_array()`.

Array/Traversable/Iterable handling in PHP is just a big mess, unfortunately.

includes/sanitizers/class-amp-meta-sanitizer.php
schlessera: I think we should treat the sanitizers independently from the `AMP_Theme_Support` processing here, as we want to produce a stand-alone sanitizer library in the long term.

While it is perfectly fine to do early bails in `AMP_Theme_Support` for various reasons as part of the "WP integration" of the plugin, I think that "asserting" the requirements in the sanitizer should still be done nevertheless.

includes/sanitizers/class-amp-meta-sanitizer.php
schlessera: Yes, but only if we do indeed transform the document first. Otherwise, we're not only in the wrong charset, but we've also lost the information about what the actual charset is => broken^2. :)

includes/sanitizers/class-amp-meta-sanitizer.php
schlessera: More generally, would it be possible to refactor the tag & meta sanitizer to strip the offending properties only, and then check if the entire is still valid or not? This way, we wouldn't have to duplicate this logic all over the place...

> it would be useful to perhaps to refactor the spec generation logic to use the spec_name as the array key

Is this guaranteed not to have collisions?

includes/sanitizers/class-amp-meta-sanitizer.php
schlessera: Yes, I'll add an additional reference here. The comment above is meant to hint at the mechanism of how you can change the order that is being produced here.

Btw, the comments in the `AMP_Theme_Support::ensure_required_markup()` are the only good thing about procedural code... ;)

includes/sanitizers/class-amp-meta-sanitizer.php
westonruter: > More generally, would it be possible to refactor the tag & meta sanitizer to strip the offending properties only, and then check if the entire thing is still valid or not? This way, we wouldn't have to duplicate this logic all over the place...

Yes, this could be done. There could be validation errors with type `invalid_meta_property`. I like that.

> Is this guaranteed not to have collisions?

It turns out, yes! Sometimes the `spec_name` is omitted when it is just a regular HTML element, I think. And for `script` we just need to derive a `spec_name` from the `extension_spec`. When this is done, all tag specs are unique: https://gist.github.com/westonruter/8e6a8ca427c69b23cdd8e26dcccbfc3d

includes/amp-helper-functions.php
```diff
@@ -931,6 +931,7 @@ function amp_get_content_sanitizers( $post = null ) {
 		'AMP_Style_Sanitizer'             => [
 			'include_manifest_comment' => ( defined( 'WP_DEBUG' ) && WP_DEBUG ) ? 'always' : 'when_excessive',
 		],
+		'AMP_Meta_Sanitizer'              => [],
```
westonruter: This needs to be added below as well, under `Force style sanitizer and whitelist sanitizer to be at end.`

schlessera: Resolved in https://github.com/ampproject/amp-wp/pull/3758/commits/af541751fc9706fceb9de77c0ef339b40cd4b08c

includes/sanitizers/class-amp-meta-sanitizer.php
schlessera: As discussed during the plugin sync, I'll tackle the enhanced validation-stripping-granularity in a separate PR to keep this one moderately sane.

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,313 @@
...
+		$this->original_encoding = $this->detect_and_strip_encoding( $source );
+
+		if ( self::AMP_ENCODING !== strtolower( $this->original_encoding ) ) {
+			$source = $this->adapt_encoding( $source );
+		}
+
+		// Force-add http-equiv charset to make DOMDocument behave as it should.
+		// See: http://php.net/manual/en/domdocument.loadhtml.php#78243.
+		$source = str_replace(
+			'<head>',
```
westonruter: What if `$source` has a `<head>` that hsa attributes like `<head profile="http://www.acme.com/profiles/core">`

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,313 @@
...
+			if ( false === strpos( $substring, '<head>' ) ) {
+				// Create the required HTML structure if none exists yet.
+				$content = "<html><head></head><body>{$content}</body></html>";
+			} else {
+				// <head> seems to be present without <body>.
+				$content = preg_replace( '#</head>(.*)</html>#', '</head><body>$1</body>', $content );
+			}
+		} elseif ( false === strpos( $substring, '<head>' ) ) {
+			// Create a <head> element if none exists yet.
+			$content = str_replace( '<body', '<head></head><body', $content );
```
westonruter: What if `<head>` is present but it contains attributes, like:

```html
<head profile='http://www.acme.com/profiles/core'>
```

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,313 @@
...
+			$this,
+			'meta',
+			[
+				'http-equiv' => 'content-type',
+				'content'    => 'text/html; charset=' . self::AMP_ENCODING,
+			]
+		);
+		$head->insertBefore( $charset, $head->firstChild );
+
+		return str_replace( '<meta http-equiv="content-type" content="text/html; charset=' . self::AMP_ENCODING . '">', '', parent::saveHTML( $node ) );
```
westonruter: This replacement key makes me a bit nervous, as it means we have have to rely on libxml always serializing attributes the same way. As long as we have tests for it, then I suppose nothing to worry about.

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,313 @@
...
+			'<head><meta http-equiv="content-type" content="text/html; charset=' . self::AMP_ENCODING . '">',
+			$source
+		);
+
+		$success = parent::loadHTML( $source, $options );
+
+		if ( $success ) {
+			$this->encoding = self::AMP_ENCODING;
+			$head           = $this->getElementsByTagName( 'head' )->item( 0 );
+			$head->removeChild( $head->firstChild );
```
westonruter: What if the above `str_replace()` didn't actually do any replacement. It could remove the wrong node here.

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,313 @@
...
+		$source = str_replace(
+			'<head>',
+			'<head><meta http-equiv="content-type" content="text/html; charset=' . self::AMP_ENCODING . '">',
+			$source
+		);
+
+		$success = parent::loadHTML( $source, $options );
+
+		if ( $success ) {
+			$this->encoding = self::AMP_ENCODING;
```
westonruter: Wouldn't `DOMDocument` populate this? Why is this needed?

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,313 @@
...
+		}
+
+		// Guessing the encoding seems to have failed, so we assume UTF-8 instead.
+		if ( empty( $this->original_encoding ) ) {
+			$this->original_encoding = self::AMP_ENCODING;
+		}
+
+		$this->original_encoding = $this->sanitize_encoding( $this->original_encoding );
+
+		$target = mb_convert_encoding( $source, self::AMP_ENCODING, $this->original_encoding );
```
westonruter: This will need to short-circuit if the `mbstring` extension is not loaded, since we do not currently include it among the `$_amp_required_extensions`.

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,313 @@
...
+		// Check for HTML 5 charset meta tag. This overrides the HTML 4 charset.
+		$charset_tag = $this->find_tag( $content, 'meta', 'charset' );
+		if ( $charset_tag ) {
+			$encoding = $this->extract_value( $charset_tag, 'charset' );
+		}
+
+		// Strip charset tags if they don't fit the AMP UTF-8 requirement.
+		if ( self::AMP_ENCODING !== strtolower( $encoding ) ) {
+			$http_equiv_tag && str_replace( $http_equiv_tag, '', $content );
+			$charset_tag && str_replace( $charset_tag, '', $content );
```
westonruter: These seem to be unused expressions. Shouldn't they be assigned to something?

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,313 @@
...
+	 * Magic getter to implement lazily-created, cached properties for the document.
+	 *
+	 * @param string $name Name of the property to get.
+	 * @return mixed Value of the property, or null if unknown property was requested.
+	 */
+	public function __get( $name ) {
+		switch ( $name ) {
+			case 'xpath':
+				$this->xpath = new DOMXPath( $this );
+				return $this->xpath;
```
westonruter: Very nice.

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,313 @@
...
+ *
+ * @package AMP
+ */
+
+/**
+ * Class AMP_DOM_Document.
+ *
+ * @since 1.5
+ *
+ * @property DOMXpath $xpath XPath query object for this document.
```
westonruter:
```suggestion
 * @property DOMXPath $xpath XPath query object for this document.
```

includes/utils/class-amp-dom-document.php
westonruter: Actually, I suggest that if a document is not `utf-8` and we cannot convert into `utf-8` that we throw an exception. AMP would just not be available.

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,313 @@
...
+	 * logic try to figure it out itself.
+	 *
+	 * @param string $encoding Encoding to sanitize.
+	 * @return string Sanitized encoding. Falls back to 'auto' on failure.
+	 */
+	private function sanitize_encoding( $encoding ) {
+		static $known_encodings = null;
+
+		if ( null === $known_encodings ) {
+			$known_encodings = array_map( 'strtolower', mb_list_encodings() );
```
westonruter: Need to make sure that the `mb_list_encodings()` function exists.

includes/sanitizers/class-amp-meta-sanitizer.php
westonruter: Related: #3780. May be best to wait to work on this until that PR is merged.

westonruter: Actually, I can include the property sanitization as part of #3780.

schlessera:
Yes, some more hardening is needed here, this was the first attempt that finally passed all the tests.

Also, I need write a lot of additional tests for the new code, move some tests over, and think about how to best test the actual encoding conversion.

includes/utils/class-amp-dom-document.php
schlessera: These are string replacements that only happen when the strings evaluate to `true`. Too cryptic?

includes/utils/class-amp-dom-document.php
schlessera: It's one of the many bits I found online that might or might not improve DOMDocument behavior with UTF-8.

However, I don't think this does much, so I'll remove it.

includes/utils/class-amp-dom-document.php
schlessera: I'll add a safeguard around it.

includes/utils/class-amp-dom-document.php
schlessera: It's a tag with two string-based attributes. I'm not sure how far off this could get.

However, I'll replace it with a `preg_replace()` to make it case-insensitive and "quote-insensitive".

includes/utils/class-amp-dom-document.php
schlessera: Yes, will change to a `preg_replace()`.

includes/utils/class-amp-dom-document.php
westonruter: The string replacements aren't done in place though, as `$content` is not passed by reference. I would expect something like this:

```php
if ( $charset_tag ) {
    $content = str_replace( $charset_tag, '', $content );
}
```

schlessera: You're right, this is nonsense. I'll change it.

includes/sanitizers/class-amp-meta-sanitizer.php
schlessera: Yes, I've put this into `AMP_DOM_Document` now.

includes/utils/class-amp-dom-document.php
schlessera: I modified this to use a regex to match any head tag. However, tests revealed now that `DOMDocument` seems to strip all attributes from the head tag...

schlessera:
There's still a few minor kinks to figure out, but I think I've got general automated character set conversion working.

Also, the new `AMP_DOM_Document` provides a way of centralizing some operations that are reused again and again, saving some processing time. I've done a preliminary change for fetching `$xpath`, `$head` and `$body` for some classes now.

includes/class-amp-autoloader.php
```diff
@@ -96,6 +97,7 @@ class AMP_Autoloader {
 		'AMP_Content'                        => 'includes/templates/class-amp-content',
 		'AMP_Content_Sanitizer'              => 'includes/templates/class-amp-content-sanitizer',
 		'AMP_Post_Template'                  => 'includes/templates/class-amp-post-template',
+		'AMP_DOM_Document'                   => 'includes/utils/class-amp-dom-document',
```
westonruter: Should this go into the `Amp\AmpWP` namespace now that we have it? Or is this premature?

includes/class-amp-theme-support.php
```diff
@@ -1583,7 +1574,7 @@ public static function ensure_required_markup( DOMDocument $dom, $script_handles
 				),
 			],
 		];
-		$link_elements = $head->getElementsByTagName( 'link' );
+		$link_elements = $dom->head->getElementsByTagName( 'link' );
```
westonruter: Love this.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -0,0 +1,246 @@
...
+	public static $tag = 'meta';
+
+	/**
+	 * Placeholder for default arguments, to be set in child classes.
+	 *
+	 * @var array
+	 */
+	protected $DEFAULT_ARGS = [ // phpcs:ignore WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase
+		'use_document_element' => true, // We want to work on the header, so we need the entire document.
+	];
```
westonruter: Not sure if this will have any practical effect. Since we're not referencing `$this-args['use_document_element']` in the sanitizer, this is just setting a member which is going to be either unused or just overridden in the first place.

includes/sanitizers/class-amp-meta-sanitizer.php
westonruter: Or rather a subsequent PR built off of that PR.

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,424 @@
...
+	public function __get( $name ) {
+		switch ( $name ) {
+			case 'xpath':
+				$this->xpath = new DOMXPath( $this );
+				return $this->xpath;
+			case 'head':
+				$this->head = $this->getElementsByTagName( 'head' )->item( 0 );
+				return $this->head;
+			case 'body':
+				$this->body = $this->getElementsByTagName( 'body' )->item( 0 );
```
westonruter: If `$this->head` or `$this->body` are null, should this create those elements just in time?

includes/utils/class-amp-dom-document.php
```diff
@@ -0,0 +1,424 @@
...
+				return $this->body;
+		}
+
+		// Mimic regular PHP behavior for missing notices.
+		trigger_error( "Undefined property: AMP_DOM_Document::${$name}", E_NOTICE ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions,WordPress.Security.EscapeOutput
```
westonruter:
```suggestion
		trigger_error( "Undefined property: " . __CLASS__ . "::${$name}", E_NOTICE ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions,WordPress.Security.EscapeOutput
```

includes/utils/class-amp-dom-utils.php
```diff
@@ -175,26 +159,7 @@ static function( $noscript_matches ) {
...
-		$head = $dom->getElementsByTagName( 'head' )->item( 0 );
-		if ( ! $head ) {
-			$head = $dom->createElement( 'head' );
-			$dom->documentElement->insertBefore( $head, $dom->documentElement->firstChild );
-		}
-		$body = $dom->getElementsByTagName( 'body' )->item( 0 );
-		if ( ! $body ) {
-			$body = $dom->createElement( 'body' );
-			$dom->documentElement->appendChild( $body );
-		}
```
westonruter: Is the ensuring of the `body` and `head` being done now?

includes/class-amp-autoloader.php
schlessera: No, I think it makes sense, if we can agree that we'll use this new abstraction everywhere going forward.

I assume it should be `Amp\AmpWP\DOMDocument`, due to its importance?

includes/sanitizers/class-amp-meta-sanitizer.php
schlessera: Yes, I think this can be removed now.

includes/utils/class-amp-dom-document.php
schlessera: It does. We're not defining the internal properties `$head` and `$body`, so the first time around, they don't exist and a call to `$this->head` will trigger the magic `__get()`. Within the magic getter, we dynamically set the property `$head`, so the next call will immediately retrieve it and not hit the magic getter anymore.

includes/utils/class-amp-dom-utils.php
schlessera: The `AMP_DOM_Document` now enforces a base structure:
https://github.com/ampproject/amp-wp/blob/7513e4145c393d699cede7ec32fd4c037a38835b/includes/utils/class-amp-dom-document.php#L170-L188

includes/class-amp-autoloader.php
westonruter: Sure, or what about `Amp\AmpWP\DOM\Document`?

schlessera: If we use a subnamespace like that, we would also add `DOMElementList` in there as well.

It would allow us to important the namespace only and then use the relative `DOM` subnamespace in the code, if we want:

```php
use Amp\AmpWP\DOM;

$dom = new DOM\Document();
$list = ( new DOM\ElementList() )
    ->add( $image1 )
    ->add( $image2 );
```

However, we might want to go with `Amp\AmpWP\Dom\Document` to match `Amp`. It would make the result less "shouty".

westonruter:
Sounds good to me.

schlessera:
⚠️ Rebasing now to get access to the new namespace code.

schlessera:
⚠️ Rebasing now to get resolve conflicts.

pierlon:
Hey @schlessera, reverting 4b391698a36b6a773e7b93a54ce8628a4aec1756 does resolve the failing Twitter embed tests, but I'm not sure as to why they are failing in the first place as yet.

schlessera:
⚠️ Rebasing now to get resolve conflicts.

schlessera:
@westonruter This is ready for a proper review now. The `Dom\Document` is feature-complete for a first iteration, and all tests are passing.

A few design decisions that are worth noting:
- I deprecated some `AMP_DOM_Utils` methods in favor of methods on `Dom\Document`. It seems more consistent and reliable that way. I left the two methods that deal with "partial" documents, but the ones that deal with the document as a whole should use `Dom\Document` instead.
- The `Dom\Document` enforces the following structure, and it will always be present:
	```html
	<!doctype html>
	<html>
	  <head>
	    <meta charset="utf-8">
	  </head>
	  <body>
	  </body>
	</html>
	```
	The `<meta charset="utf-8">` is not strictly necessary according to the AMP spec, but I think it is a good practice to just always have it around, especially as we now might have a different charset for the AMP page than for the non-AMP page.
- As we "cache" `body`, `head` and `xpath` now, there might be edge cases where these should be rest. I added one of these for now for the case where you use `loadHTML()` again on the same document. There might be others, so this is something to be aware of for now.
Cloning the `Dom\Document` can lead to such edge cases, so this is also something to use with caution until we were able to play around more with it.

schlessera:
Note: The new `AMP_Meta_Sanitizer` is in an intermediary state here, and it should be revisited when working on building sanitizers that mirror the optimizing transformers defined in https://docs.google.com/document/d/1lZl7UJ5eUrhwTmfbOZ7g-HFIgMnTeqGwfUxE98kALzU/edit#heading=h.fzwfkw32haw7

westonruter:
> * The `<meta charset="utf-8">` is not strictly necessary according to the AMP spec, but I think it is a good practice to just always have it around, especially as we now might have a different charset for the AMP page than for the non-AMP page.

Actually, it is required according to the spec: 

* https://amp.dev/documentation/guides-and-tutorials/learn/spec/amphtml/#chrs
* https://github.com/ampproject/amphtml/blob/4917e7096594df556404c757d449eb295acc3fd3/validator/validator-main.protoascii#L404-L423

src/Dom/Document.php
```diff
@@ -0,0 +1,968 @@
...
+			$known_encodings = array_map( 'strtolower', mb_list_encodings() );
+		}
+
+		if ( array_key_exists( strtolower( $encoding ), self::$encoding_map ) ) {
+			$encoding = self::$encoding_map[ strtolower( $encoding ) ];
+		}
+
+		if ( ! in_array( strtolower( $encoding ), $known_encodings, true ) ) {
+			return self::UNKNOWN_ENCODING;
+		}
```
westonruter: Suggested:

```php
		$encoding = strtolower( $encoding );
		if ( isset( self::$encoding_map[ $encoding ] ) ) {
			$encoding = self::$encoding_map[ self::$encoding_map[ $encoding ] ];
		}

		if ( ! in_array( $encoding, $known_encodings, true ) ) {
			return self::UNKNOWN_ENCODING;
		}
```

Reduce calls to `strtolower()` and `isset()` is marginally faster than `array_key_exists()`, last I checked.

tests/php/test-class-amp-dom-document.php
```diff
@@ -0,0 +1,328 @@
...
+	 *
+	 * @param string $charset  Charset to use.
+	 * @param string $source   Source content.
+	 * @param string $expected Expected target content.
+	 *
+	 * @dataProvider data_dom_document
+	 * @covers Document::loadHTML()
+	 * @covers Document::saveHTML()
+	 */
+	public function test_dom_document( $charset, $source, $expected ) {
```
westonruter: The `$charset` param is unused. Is that intentional?

tests/php/test-class-amp-dom-document.php
```diff
@@ -0,0 +1,328 @@
...
+		$this->assertEquals( 1, $dom->head->childNodes->length );
+		$this->assertEquals( 'body', $dom->documentElement->lastChild->nodeName );
+		$this->assertEquals( 'Hello world', $dom->body->lastChild->textContent );
+	}
+
+
+	/**
+	 * Test that invalid head nodes are moved to body.
+	 *
+	 * @covers \Amp\AmpWP\Dom\Document::move_invalid_head_nodes_to_body()
```
westonruter: This is fully-qualified but `Document::loadHTML()` above is not. Intentional?

src/Dom/Document.php
```diff
@@ -0,0 +1,968 @@
...
+				')>#s',
+			]
+		);
+		$converted = preg_replace_callback(
+			$pattern,
+			$replace_callback,
+			$html
+		);
+
+		/**
```
westonruter:
```suggestion
		/*
```

tests/php/test-class-amp-dom-document.php
```diff
@@ -0,0 +1,328 @@
...
+
+		$html = 'Hello world';
+		$dom  = Document::from_html( $html );
+		$this->assertEquals( 'head', $dom->documentElement->firstChild->nodeName );
+		$this->assertEquals( 1, $dom->head->childNodes->length );
+		$this->assertEquals( 'body', $dom->documentElement->lastChild->nodeName );
+		$this->assertEquals( 'Hello world', $dom->body->lastChild->textContent );
+	}
+
+
```
westonruter:
```suggestion
```

src/Dom/Document.php
```diff
@@ -0,0 +1,968 @@
...
+	const HTML_GET_HTTP_EQUIV_TAG_PATTERN       = '#<meta http-equiv=([\'"])content-type\1 content=([\'"])text/html; charset=utf-8\2>#i';
+	const HTML_HTTP_EQUIV_VALUE                 = 'content-type';
+	const HTML_HTTP_EQUIV_CONTENT_VALUE         = 'text/html; charset=utf-8';
+
+	// Regex patterns used for finding tags or extracting attribute values in an HTML string.
+	const HTML_FIND_TAG_WITHOUT_ATTRIBUTE_PATTERN = '/<%1$s[^>]*?>[^<]*(?:<\/%1$s>)?/i';
+	const HTML_FIND_TAG_WITH_ATTRIBUTE_PATTERN    = '/<%1$s [^>]*?\s*%2$s=[^>]*?>[^<]*(?:<\/%1$s>)?/i';
+	const HTML_EXTRACT_ATTRIBUTE_VALUE_PATTERN    = '/%s=(?:([\'"])(?<full>.*)?\1|(?<partial>[^ \'";]+))/';
+
+
```
westonruter:
```suggestion
```

westonruter:
Alpha build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3980938/amp.zip) (v1.5.0-alpha-20191218T231919Z-8cbe6d60)

westonruter:
> * Cloning the `Dom\Document` can lead to such edge cases, so this is also something to use with caution until we were able to play around more with it.

Wouldn't this be addressed by implementing [`__clone()`](https://www.php.net/manual/en/language.oop5.cloning.php#object.clone)? For example:

```php
function __clone() {
    unset( $this->xpath, $this->head, $this->body );
}
```

This will reset those properties so that the next time the getter is invoked, they will get re-populated.

includes/class-amp-theme-support.php
```diff
@@ -1665,97 +1658,18 @@ public static function ensure_required_markup( DOMDocument $dom, $script_handles
...
-		// Insert all the the meta elements next in the head.
-		$previous_node = $meta_viewport;
-		foreach ( $meta_elements as $meta_element ) {
-			$meta_element->parentNode->removeChild( $meta_element );
-			$head->insertBefore( $meta_element, $previous_node->nextSibling );
-			$previous_node = $meta_element;
-		}
+		// Store the last meta tag as the previous node to append to.
+		$meta_tags     = $dom->head->getElementsByTagName( 'meta' );
+		$previous_node = $meta_tags->length > 0 ? $meta_tags->item( $meta_tags->length - 1 ) : $dom->head->firstChild;
```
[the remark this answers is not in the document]
schlessera: The head would never be empty, because we're enforcing the `<meta charset="utf-8">` to be present. But I can nevertheless harden the calls to `nextSbiling`to make sure we won't randomly break this in the future.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -396,10 +382,10 @@ public static function has_required_php_css_parser() {
 	 *
 	 * @since 0.7
 	 *
-	 * @param DOMDocument $dom  Represents the HTML document to sanitize.
-	 * @param array       $args Args.
+	 * @param Document $dom  Represents the HTML document to sanitize.
+	 * @param array    $args Args.
 	 */
-	public function __construct( DOMDocument $dom, array $args = [] ) {
+	public function __construct( $dom, array $args = [] ) {
```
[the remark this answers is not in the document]
schlessera: Yes, to stick with the base signature. I meant to typehint againt `Amp\AmpWP\Dom\Document` in all sanitizers at first, but that would be a breaking change for people that already extend the base sanitizer in their own plugins/themes. We should do that with v2.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1595,10 +1573,10 @@ private function parse_import_stylesheet( Import $item, CSSList $css_list, $opti
 	 * @return array {
 	 *    Parsed stylesheet.
 	 *
-	 *    @type Document $css_document       CSS Document.
-	 *    @type array    $validation_results Validation results, array containing arrays with error and sanitized keys.
-	 *    @type string   $stylesheet_url     Stylesheet URL, if available.
-	 *    @type string   $spec_name          Spec name.
+	 *    @type CSSDocument $css_document       CSS Document.
```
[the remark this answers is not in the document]
schlessera: That was already correct, but I ended up with two `Document` classes: `Css\Document` and `Dom\Document`. So I aliased `Css\Document` to `CSSDocument`.

src/Dom/Document.php
```diff
@@ -0,0 +1,968 @@
...
+	/**
+	 * Move invalid head nodes back to the body.
+	 */
+	private function move_invalid_head_nodes_to_body() {
+		// Walking backwards makes it easier to move elements in the expected order.
+		$node = $this->head->lastChild;
+		while ( $node ) {
+			$next_sibling = $node->previousSibling;
+			if ( ! AMP_DOM_Utils::is_valid_head_node( $node ) ) {
```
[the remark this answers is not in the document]
schlessera: It is in the Document class...?

tests/php/test-class-amp-dom-document.php
schlessera: No, it's not, good catch!

schlessera:
> > * The `<meta charset="utf-8">` is not strictly necessary according to the AMP spec, but I think it is a good practice to just always have it around, especially as we now might have a different charset for the AMP page than for the non-AMP page.
> 
> Actually, it is required according to the spec:

Ah, yes, I mixed stuff up in my head. In terms of parsing the DOMDocument, it is not required (as it doesn't work properly, actually, and DOMDocument needs an HTML4 http-equiv) and adding it would normally happen in sanitizers.

But, still, I think it is good to have here, and when we agree it is worth changing the existing tests for that, then its settled.

schlessera:
> > * Cloning the `Dom\Document` can lead to such edge cases, so this is also something to use with caution until we were able to play around more with it.
> 
> Wouldn't this be addressed by implementing [`__clone()`](https://www.php.net/manual/en/language.oop5.cloning.php#object.clone)? For example:
> 
> ```
> function __clone() {
>     unset( $this->xpath, $this->head, $this->body );
> }
> ```
> 
> This will reset those properties so that the next time the getter is invoked, they will get re-populated.

I thought about that but was wondering how far that rabbit-hole might take me if the DOMDocument does by itself already cause issues. But I can try adding this and reverting the test change to see what we'll get.

tests/php/test-class-amp-dom-document.php
schlessera: Shouldn't matter, but I'll make them consistent.

When using namespace, I prefer to stick with omitting the root namespace, so I'll go with that.

src/Dom/Document.php
schlessera: Yes, that's better (apart from the double look-up). `isset()` is not technically the same, but still good to use here.

src/Dom/Document.php
```diff
@@ -0,0 +1,968 @@
...
+		return false !== $target ? $target : $source;
+	}
+
+	/**
+	 * Detect the encoding of the document.
+	 *
+	 * @param string $content Content of which to detect the encoding.
+	 * @return string|false Detected encoding of the document, or false if none.
+	 */
+	private function detect_and_strip_encoding( &$content ) {
```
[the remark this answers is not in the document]
schlessera: Yes, I can do that. I don't neither of the two approaches, but I guess returning an array offers less surprises.

includes/class-amp-theme-support.php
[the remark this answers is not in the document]
schlessera: Hmm, actually, I don't think it's worth hardening at the moment. Right now, the charset meta tag ensures the head always has at least 1 node, and I'm starting work on the optimized AMP stuff after this, where I'll rip most of the rest of this stuff out of `AMP_Theme_Support` anyway.

src/Dom/Document.php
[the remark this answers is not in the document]
schlessera: Oh, nevermind, looked at the wrong method in the file view.

The `AMP_DOM_Utils::is_valid_head_node()` is publicly accessible, so moving it would mean deprecating the old version, not removing that one. But I can do that as well.

tests/php/test-class-amp-dom-document.php
schlessera: Oh, our tests don't have namespaces, so I guess I can only still with the FQCN anyway.

src/Dom/Document.php
schlessera: Oh, but `$encoding` itself needs to stay in its original form. I'll adapt.

tests/php/test-class-amp-dom-document.php
schlessera: Ah, we're importing Document, so just using `Document` should be fine, and causes less visual clutter.

src/Dom/Document.php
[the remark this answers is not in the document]
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/3758/commits/41094b488fa7977effd9e25e1473f7197169ea6b

src/Dom/Document.php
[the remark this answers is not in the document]
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/3758/commits/6ff286ca11dbec56aab7323d8383ae81b7f3fad3

src/Dom/Document.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/3758/commits/e94e3caa06d321a56b0e394ba7e7effc4e2471e8

tests/php/test-class-amp-dom-document.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/3758/commits/c5415d2bbf9ae25f120c5de36d30a5c1068e0b80

tests/php/test-class-amp-dom-document.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/3758/commits/7c037a94f72b8e3ea07b2a60d3ad8179f9740da7

schlessera:
@westonruter All of your feedback from the review should be addressed now. Also, I got rid of the dependency on `AMP_DOM_Utils` in `Dom\Document`, as I think that doesn't make sense and creates a circular dependency.

src/Dom/Document.php
```diff
@@ -258,6 +278,22 @@ public static function from_node( DOMNode &$node ) {
...
 
+	/**
+	 * Reset the internal optimizations of the Document object.
+	 *
+	 * This might be needed if you are doing an operation that causes the cached
+	 * nodes and XPath objects to point to the wrong document.
+	 *
+	 * @return self Reset version of the Document object.
+	 */
+	public function reset() {
```
westonruter: Does this need to be public?

tests/php/test-class-amp-dom-document.php
```diff
@@ -325,4 +324,127 @@ public function test_invalid_head_nodes() {
 		$this->assertNull( $dom->head->firstChild->nextSibling );
 		$this->assertEquals( 6, $dom->body->childNodes->length );
 	}
+
+
```
westonruter: Extra line break?

src/Dom/Document.php
schlessera: No, it's actually better to default to `private` for now, and only make it public if we actually now it is needed.

schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/3758/commits/e006454b725fd4c1a0589af73211c0a82c5a97ed

tests/php/test-class-amp-dom-document.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/3758/commits/b70ea783efddf747e5e6001854e59d350a0d511e

schlessera:
Note: I force-pushed because I messed up a commit.

westonruter:
Great job on this large effort.

includes/sanitizers/class-amp-meta-sanitizer.php
westonruter: Opened a new issue to track this: #4070

src/Dom/Document.php
```diff
@@ -0,0 +1,1033 @@
...
+	 * Regular expressions to fetch the individual structural tags.
+	 * These patterns were optimized to avoid extreme backtracking on large documents.
+	 */
+	const HTML_STRUCTURE_DOCTYPE_PATTERN = '/^[^<]*<!doctype(?:\s+[^>]+)?>/i';
+	const HTML_STRUCTURE_HTML_START_TAG  = '/^[^<]*(?<html_start><html(?:\s+[^>]*)?>)/i';
+	const HTML_STRUCTURE_HTML_END_TAG    = '/(?:<\/html(?:\s+[^>]*)?>)[^<>]*$/i';
+	const HTML_STRUCTURE_HEAD_START_TAG  = '/^[^<]*(?:<head(?:\s+[^>]*)?>)/i';
+	const HTML_STRUCTURE_BODY_START_TAG  = '/^[^<]*(?:<body(?:\s+[^>]*)?>)/i';
+	const HTML_STRUCTURE_BODY_END_TAG    = '/(?:<\/body(?:\s+[^>]*)?>)[^<>]*$/i';
+	const HTML_STRUCTURE_HEAD_TAG        = '/^(?:[^<]*(?:<head(?:\s+[^>]*)?>).*?<\/head(?:\s+[^>]*)?>)/is';
```
westonruter: These patterns aren't all accounting for the inclusion of HTML comments which are added when doing validation requests. This results in HTML documents being corrupted when validating. See https://github.com/ampproject/amp-wp/issues/4104.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -0,0 +1,243 @@
...
+	 *
+	 * @var string
+	 */
+	const AMP_VIEWPORT = 'width=device-width';
+
+	/**
+	 * Sanitize.
+	 */
+	public function sanitize() {
+		$elements = $this->dom->getElementsByTagName( static::$tag );
```
westonruter, line 71: This caused a bug. See #4502.

Issue #3662, opened 2019-10-29
Refactor content transformation engine into a middleware stack


schlessera:
## Feature description

We currently have multiple code paths that deal with transforming the content for the different modes. In the future, we plan to add REST API requests, optimized AMP and more that will also need to be accommodated. With the current design, this would probably mean copy-pasting and adapting large parts of the code for each use case, making extension and long-term maintenance difficult and costly.

I suggest refactoring the content transformation conceptually into a middleware stack. This would allow easy reuse of code paths while making them extensible to address each use case as needed.

The model can be similar to what most modern frameworks use right now for their HTTP stack, only that it would start slightly later in the process. So where a traditional HTTP stack starts with an HTTP request and has as an end-result an HTTP response, we would also target the HTTP response as the end result but start after the initial WordPress processing of the HTTP request.

The middleware concept allows all the different behaviors we need to account for to be cleanly separated from each other and the main use case to be a matter of assembling the correct stack. It is a flexible pipeline that takes us from input to output in a structured way.

Here's a (very improvised and probably incorrect) example of how such a middleware stack would be constructed:

```php
$amp_stack = new AMP_Stack(
	new AMP_Error_Logger(),
	new AMP_Dev_Mode(),
	new AMP_Reader_Mode_Theme_Switcher(),
	new AMP_Content_Sanitizer(),
	new AMP_Optimizer(),
	new AMP_Internal_Cache(),
);

$output = $amp_stack->process( $input );
```

If we need different modes, we encode them as one or more steps to add to the middleware (no duplication required).

If we need conditional behavior, we encode them as one or more steps to add to the middleware (no wading through lots of places in the code to find the correct injection point).

If we need cross-cutting concerns like caching or logging, we encode them as one or more steps to add to the middleware (no obfuscation of the actual logic by these cross-cutting concerns).

Everything adheres to the same structure, and it becomes obvious where to look for which part of the logic.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
As discussed in today's meeting, I'll work on a design document together with @westonruter to flesh out the details of how this would work.

schlessera:
[...]

PR #3781, opened 2019-11-16
Pass entire Reader mode templates through post-processor

### LINKED ISSUE #688: CSS in Reader mode is not minified

### LINKED ISSUE #2202: Send entire templates in Reader/Classic mode through the sanitizers

westonruter:
## Summary

- [x] Eliminate majority of Reader mode template's unique rendering path, rendering them with the same processing logic in `AMP_Theme_Support`. This ensures the entire documents are sanitized and all CSS is tree-shaken.
- [x] Use `template_include` filter to load a template which initializes the `AMP_Post_Template` class while output buffering is being done. This ensures themes which extend the Reader mode templates and make use of `$this` may continue to do so.
- [x] Deprecate functions and `AMP_Content` class which are no longer used.
- [x] Ensure validation error reporting (compatibility tool) remains disabled in Reader mode (for now).
- [x] Move deprecated functions to `deprecated.php`.
- [x] Fix viewing Reader template in Customizer.

Fixes #2202.
Fixes #688.

## Checklist


includes/templates/class-amp-post-template.php
```diff
@@ -164,10 +148,41 @@ public function __construct( $post ) {
...
+			case 'post_modified_timestamp':
+				return get_post_modified_time( 'U', false, $this->post );
+			case 'post':
+				return $this->post;
+			case 'post_id':
+				return $this->post->ID;
+			case 'site_icon_url':
+				/** This filter is documented in includes/amp-helper-functions.php */
+				return apply_filters( 'amp_site_icon_url', function_exists( 'get_site_icon_url' ) ? get_site_icon_url( self::SITE_ICON_SIZE ) : '' );
+		}
```
westonruter: This makes me nervous. By making the fields here lazy-loaded when invoking the getter, it bypasses the `amp_post_template_data` filter. Perhaps this should be reverted, and the introduction of regular WordPress template tags be replaced with `AMP_Post_Template::get()` to ensure no breakage of backwards-compatibility. See search for `amp_post_template_data` filter across plugins:  https://wpdirectory.net/search/01DSTQ6DK2GCDKF9QDEN0E8X7R

Nevertheless, going forward sites would be free to use WordPress template tags when writing Reader mode templates.

westonruter: Undone in aed6c0e6550ac9404fa96b838b6ab1ddd541a933

includes/class-amp-theme-support.php
```diff
@@ -417,15 +419,26 @@ public static function finish_init() {
...
 
-		$theme_support = self::get_theme_support_args();
+		$is_reader_mode = self::READER_MODE_SLUG === self::get_support_mode();
+		$theme_support  = self::get_theme_support_args();
 		if ( ! empty( $theme_support['template_dir'] ) ) {
 			self::add_amp_template_filters();
+		} elseif ( $is_reader_mode ) {
+			add_action(
+				'template_include',
+				function() {
```
westonruter: Should either be `static` or to add a new global function.

amp.php
```diff
@@ -449,6 +446,7 @@ function amp_force_query_var_value( $query_vars ) {
  * @return void
  */
 function amp_maybe_add_actions() {
+	_deprecated_function( __FUNCTION__, '1.5' );
```
westonruter: The deprecated functions here should be moved to `includes/deprecated.php` (which interestingly is not being included anywhere!)

Also, it would be good to move all of the non-bootstrapping functions in `amp.php` into `includes/amp-helper-functions.php`. This will allow us to use PHP>5.2 in these functions.

westonruter:
The `AMP_Content` class can be hard-deprecated once we have the new API established in #3662.

includes/class-amp-theme-support.php
```diff
@@ -417,15 +419,26 @@ public static function finish_init() {
 
 		self::ensure_proper_amp_location();
 
-		$theme_support = self::get_theme_support_args();
+		$is_reader_mode = self::READER_MODE_SLUG === self::get_support_mode();
+		$theme_support  = self::get_theme_support_args();
 		if ( ! empty( $theme_support['template_dir'] ) ) {
 			self::add_amp_template_filters();
+		} elseif ( $is_reader_mode ) {
+			add_action(
```
westonruter:
```suggestion
			add_filter(
```

includes/class-amp-theme-support.php
```diff
@@ -417,15 +419,26 @@ public static function finish_init() {
 
 		self::ensure_proper_amp_location();
 
-		$theme_support = self::get_theme_support_args();
+		$is_reader_mode = self::READER_MODE_SLUG === self::get_support_mode();
+		$theme_support  = self::get_theme_support_args();
 		if ( ! empty( $theme_support['template_dir'] ) ) {
 			self::add_amp_template_filters();
+		} elseif ( $is_reader_mode ) {
```
westonruter: Instead of having a separate condition for reader mode, it may be good to combine with `add_amp_template_filters`. The thinking is we may want to go ahead and try to load a template as if it is in Transitional mode, but then if we are about to load a `single.php` or `page.php` that contains `$this`, then we can at that point load it via `AMP_Post_Template`. This will avoid necessarily doing the current hard-coding of `single.php` and `page.php` as seen in:

https://github.com/ampproject/amp-wp/blob/451dd22cb6f80a631c95bec7fef1b0e02f8e9c48/includes/templates/class-amp-post-template.php#L198-L199

This would allow Reader templates to make use of the full template hierarchy, and ease transition to _Transitional_ templates.

In other words, if someone does:

```php
add_theme_support( 'amp', [ 'template_dir' => 'amp' ] );
```

This should either allow the templates in the `amp` directory to be classic `AMP_Post_Template`-based templates, or regular WordPress theme templates. We can start phasing out the use of `AMP_Post_Template`, only loading it when we detect it is needed.

In fact, there shouldn't be any real difference between Reader mode and Transitional mode when a `template_dir` is present.

includes/templates/class-amp-post-template.php
```diff
@@ -111,23 +111,24 @@ public function __construct( $post ) {
 		$this->data = [
```
westonruter: Instead of assembling all of this data in the constructor, I think it would be better to lazily construct it upon invoking the getter the first time. This would allow us to construct the `AMP_Post_Template` preemptively without potentially it ever being used (and thus avoiding wasted queries), such as when normal WordPress templates are being used in the `amp` theme directory.

We could then still pass the `AMP_Post_Template` object in the existing hooks like `amp_post_template_head`, and as long as no hooks call the getter, then the data is not fetched.

amp.php
```diff
@@ -667,6 +668,7 @@ function amp_render() {
  * @global WP_Query $wp_query
  */
 function amp_render_post( $post ) {
+	_deprecated_function( __FUNCTION__, '1.5' );
```
westonruter: We need to check if these functions are being used in themes/plugins, and if so, potentially defer this hard deprecation until we have the AMP middleware rendering stack available to use instead (same for `AMP_Content`).

includes/class-amp-theme-support.php
schlessera: Instead of detecting `$this` inside of a template (which requires parsing the entire template file), we could also look into wrapping _all_ templates in a compatibility wrapper object by default, as regular WordPress templates don't use `$this`. However, I'm not 100% sure whether that cleanly works across global state (which is expected in regular WordPress templates).

amp.php
schlessera: > which interestingly is not being included anywhere!

Hehe, does that mean that the current contents of that file can be hard-deprecated ?

templates/html-end.php
```diff
@@ -19,7 +19,17 @@
  */
 ?>
 
-<?php do_action( 'amp_post_template_footer', $this ); ?>
+<?php
+/**
+ * Fires when just before after printing the </body>.
```
schlessera: Huh? 🤔

templates/html-start.php
```diff
@@ -23,12 +23,44 @@
...
+		do_action( 'amp_post_template_css', $this );
+		?>
 	</style>
 </head>
 
 <body class="<?php echo esc_attr( $this->get( 'body_class' ) ); ?>">
-<?php do_action( 'amp_post_template_body_open', $this ); ?>
+<?php
+/**
+ * Fires when immediately after printing the <body>.
```
schlessera:
```suggestion
 * Fires immediately after printing the <body>.
```

includes/class-amp-theme-support.php
schlessera: If you turn the action into a filter, you'll need to actually use the return value as well.

includes/class-amp-theme-support.php
```diff
@@ -458,6 +471,18 @@ public static function ensure_proper_amp_location( $exit = true ) {
...
+			$path_args = [];
+			wp_parse_str( $wp->matched_query, $path_args );
+			if ( isset( $path_args[ amp_get_slug() ] ) && '' !== $path_args[ amp_get_slug() ] ) {
+				wp_safe_redirect( amp_get_permalink( get_queried_object_id() ), 301 );
+				if ( $exit ) {
+					exit;
+				}
+				return true;
+			}
 		} else {
```
schlessera: There's a bogus if below here that can be included in the else.
```suggestion
		} elseif ( $has_query_var && ! $has_url_param ) {
```

includes/templates/class-amp-content.php
```diff
@@ -7,6 +7,8 @@
 
 /**
  * Class AMP_Content
+ *
+ * @deprecated
```
schlessera:
```suggestion
 * @deprecated 1.5 Reader mode now sanitizes its entire template through the standard post-processor.
```

includes/templates/class-amp-post-template.php
schlessera: A "fun" way of doing that (which doesn't require conditionals all over the place) is to not define the `$data` property at all in the class and then use `__get()` to add it the first time it is being requested (through whatever means). After that, it is just present, so the `__get()` won't be hit a second time.

amp.php
westonruter: Yes, probably. The file was introduced in 9750909638bc0a9864d76e1a57ef9bd91e35f794 by yours truly but I forgot to include it. So much backwards-incompatibility breakage that didn't happen to anyone!

includes/class-amp-theme-support.php
westonruter: I don't understand. The filter callback is returning a value:

```php
return __DIR__ . '/templates/reader-template-loader.php';
```

It's not applying filters here.

schlessera: Nevermind, I'm dumb.

includes/templates/class-amp-post-template.php
westonruter: Yes, this is what I intended to say. It's just in this class that logic would be in `\AMP_Post_Template::get()`.

templates/html-end.php
westonruter: Extraneous “after“.

schlessera: More than that... probably extraneous "when" as well.

includes/class-amp-theme-support.php
westonruter: Fixed in 93cede44d26e0bdadbc3fa99db0ed6cedbfad36d

includes/templates/class-amp-post-template.php
westonruter: Done in c74ba974

includes/class-amp-theme-support.php
westonruter: As opposed to parsing the entire template file, I was intending to just use a simple check to see if `'$this->'` occurs inside of the file. If so, it could then load it via the `AMP_Post_Template`. Otherwise, it could do normal `include`.

amp.php
westonruter: No relevant usages: https://wpdirectory.net/search/01DT14RRMMNGT84GHZGZC900F1

amp.php
```diff
@@ -639,6 +637,7 @@ function amp_add_post_template_actions() {
  * @deprecated This function is not used when 'amp' theme support is added.
  */
 function amp_prepare_render() {
+	_deprecated_function( __FUNCTION__, '1.5' );
```
westonruter: No relevant usages: https://wpdirectory.net/search/01DT14VJW5JX1X1F65C7R9VY5D

amp.php
```diff
@@ -649,6 +648,8 @@ function amp_prepare_render() {
  * @deprecated This function is not used when 'amp' theme support is added.
  */
 function amp_render() {
+	_deprecated_function( __FUNCTION__, '1.5' );
```
westonruter: One relevant usage: https://wpdirectory.net/search/01DT14Z9DAB3X0VQWZGJPG7XA3

It's in Facebook Instant Articles & Google AMP Pages by PageFrog. However, this plugin has been closed: https://wordpress.org/plugins/pagefrog/

amp.php
westonruter: No relevant usages: https://wpdirectory.net/search/01DT152D0C97YRHGGJHYQ19X5X

includes/class-amp-theme-support.php
schlessera: You'll still have to read the entire file to find a potential `$this`. And this reading is separate from `include`ing the file, which you'll need to do as well, and which would normally happen via opcache and not hit the filesystem, so this adds more filesystem reads.

The compatibility wrapper I suggested would be read once and wrapped around each include. So most of the time, everything would still stay within the opcache. However, it adds complexity.

westonruter: Doesn't `file_get_contents()` use caching as well? Per [docs](https://www.php.net/manual/en/function.file-get-contents.php):

> `file_get_contents()` is the preferred way to read the contents of a file into a string. It will use memory mapping techniques if supported by your OS to enhance performance.

I'm not sure what these “memory mapping techniques” entail.

In any case, we're doing `file_get_contents()` for other files with each page load, for example:

https://github.com/ampproject/amp-wp/blob/c900d8484de1751bf56fd8b1c392154fa922eebe/templates/style.php#L55

Why would this be inherently different?

In core I see `get_post_embed_html()` uses `file_get_contents()` if `SCRIPT_DEBUG` is enabled. Also, `file_get_contents()` is used by `load_script_translations()` in core.

westonruter: In any case, migrating Reader mode templates to use a normal theme templates is beyond the scope of this PR.

kienstra:
**Looks good!**

Hi @westonruter,
This looks good. It's great to have the post-processor working for the entire Reader mode templates.

For example, with:

```

add_action(
	'amp_post_template_head',
	static function() {
		?>
		<script>disallowed();</script>
		<?php
	}
);
```

This removed the script:
script-removed-as-expected

...though before this PR, there was a validation error on the front-end (in the Chrome extension).

Also, the `style[amp-custom]` looks good:

tree-shaking-done

A basic testing post looked good. It was the same with this PR, and on the `develop` branch:

testing-here

kienstra:
The [e2e failures](https://travis-ci.org/ampproject/amp-wp/jobs/614285931#L923) don't look to be related to this PR at all. They're all for Stories.

includes/templates/class-amp-post-template.php
```diff
@@ -321,18 +342,10 @@ private function build_post_comments_data() {
...
-		$amp_content = new AMP_Content(
-			post_password_required( $this->post ) ? get_the_password_form( $this->post ) : $this->post->post_content,
-			amp_get_content_embed_handlers( $this->post ),
-			amp_get_content_sanitizers( $this->post ),
-			[
-				'content_max_width' => $this->get( 'content_max_width' ),
-			]
-		);
+		/** This filter is documented in wp-includes/post-template.php */
+		$content = apply_filters( 'the_content', get_the_content( null, false, $this->post ) );
```
westonruter, line 346: This introduced a regression. Prior to this change, posts having content with page breaks (`<!--nextpage-->`) would have their pages all displayed together. This needs to be restored, with a big reason being that there has been no multipage pagination in the Reader mode template. It's also not consistent with the previous behavior.

westonruter: Workaround plugin to restore the previous behavior: https://gist.github.com/westonruter/559c7a9cbb9460d1099c3d5d9bfa5ad1

westonruter: Fix: https://github.com/ampproject/amp-wp/pull/4547

PR #3790, opened 2019-11-18
Fix expanded menu when screen width is <1000px for Twenty Twenty theme

pierlon:
## Summary

With the Twenty Twenty theme, when the expanded menu is active on a screen width <1000px, it is incorrectly shown:

### Before
image

The root cause of this is due to the container that the menu is placed in by `amp-lightbox` upon initialization is not made to fill the content because [the `amp-lightbox` has the `scrollable` attribute](https://github.com/ampproject/amphtml/blob/master/extensions/amp-lightbox/0.1/amp-lightbox.js#L224-L226).

This PR fixes this issue by unsetting the `display` CSS property of the `amp-lightbox`, reverting it from `flex` to `block`.

### After

image

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-core-theme-sanitizer.php
+++ b/includes/sanitizers/class-amp-core-theme-sanitizer.php
@@ -1031,8 +1031,8 @@ class AMP_Core_Theme_Sanitizer extends AMP_Base_Sanitizer {
 					}
 
 					@media (max-width: 999px) {
-						.show-modal {
-							display: unset !important;
+						amp-lightbox.cover-modal.show-modal {
+							display: unset;
 						}
 					}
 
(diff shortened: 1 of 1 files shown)


includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -1030,6 +1030,12 @@ static function() use ( $args ) {
 						}
 					}
 
+					@media (max-width: 999px) {
+						.show-modal {
+							display: unset !important;
```
schlessera: I'd like to avoid the use of `!important` if at all possible.

Can you try with the following, more specific selector: `amp-lightbox.cover-modal.show-modal` ? I think that should fix the problem as well without using `!important`.

pierlon: Sure that works and would prevent future CSS rule conflicts.

csossi:
Verified in QA

PR #3797, opened 2019-11-20
Ensure Stories preview in e2e tests meets minimum viewport requirements.

kopepasah:
When running the e2e tests interactively, there was a consistent failure of the `amp-story` preview page which seems to be causing an issue with the testing suite. 

In short, when AMP detects that the screen is of a specific larger size, it uses the desktop version of Stories, which requires a minimum width of 1024px and height of 550px. When this viewport requirement is not met, AMP displays an notice to the user to adjust the size of the viewport, which may have been causing those errors.

This pull request sets the Stories preview to the minimum AMP requirements (for desktop).

### DIFF
--- a/tests/e2e/utils/open-preview-page.js
+++ b/tests/e2e/utils/open-preview-page.js
@@ -22,6 +22,15 @@ export async function openPreviewPage( editorPage, selector ) {
 	}
 
 	const previewPage = last( openTabs );
+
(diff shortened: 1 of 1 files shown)


westonruter:
I've added E2E tests to `allow_failures` in d10ab1817b87f296e5447c97b808ae1c97c14cb1

Issue #3783, opened 2019-11-18
Moveable Library for Resizing Rotating and Dragging


jauyong:
## Feature description

This ticket was taken from rows in this [spreadsheet](https://docs.google.com/spreadsheets/d/1nqJ-iOypTziWFjWBIBoJUc0LHW3gTeUnLkbjVH_eW4g/edit#gid=0). This ticket can be used to track research and work related to using the Moveable library. 

Resizing:
Handles resizing a block from each side and corner.

Rotation:
Handles rotating a block

Block Mover/Dragging:
Handles dragging a block

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Research https://daybrush.com/moveable/ to see if it can handle the resizing of blocks.
* Research https://daybrush.com/moveable/ to see if it can handle the rotation of blocks.
* Research https://daybrush.com/moveable/ to see if it can handle the moving of blocks.

AC: There are documented findings with a recommendation if to use Moveable library or something else.
AC: The chosen library has been implemented to provide basic resizing, rotation and dragging.

## Implementation brief

The conclusion from looking into [Moveable](https://www.npmjs.com/package/react-moveable) is that it can be used for Resizing, Dragging, and Rotation.

* See [this commit](https://github.com/ampproject/amp-wp/pull/3798/commits/ae18f64999f5fbdce670d49d87c67011ab6e8488) for a partially working demo of Moveable within our code and attaching it to an element. Note that this code does not update the properties of the moved/resized/rotated block yet, however, it shows that we can use Moveable.

**Overview**
We will use Moveable and implement basic dragging/rotating/resizing. For the initial version, we will follow a similar logic as Google Slides has: an outline of the element is being dragged/resized/rotated instead of the element directly. Once the action has stopped, the properties of the element are updated, too. Using just the outline has the benefit of simplicity -- we won't have to worry about updating the element while it's being resized (e.g. font sizes, image ratio, etc.) which was causing a lot of issues in the previous iteration of the Story editor. This can be improved in a future iteration.
## QA testing instructions

* 

## Demo

* 
![demo-moveable](https://user-images.githubusercontent.com/3294597/69666806-2840f400-108d-11ea-8084-0c7a04f6afd1.gif)

## Changelog entry

*

miina:
[...]

PR #3798, opened 2019-11-20
Basic implementation of react-moveable npm package for dragging, resizing, rotating

miina:
## Summary

See #3783 

## Checklist

### DIFF
--- a/assets/src/edit-story/components/canvas/page.js
+++ b/assets/src/edit-story/components/canvas/page.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import styled from 'styled-components';
+import Moveable from 'react-moveable';
 
 /**
  * WordPress dependencies
@@ -58,19 +59,41 @@ function Page() {
 	useEffect( () => {
 		setBackgroundClickHandler( () => clearSelection() );
 	}, [ setBackgroundClickHandler, clearSelection ] );
+	let selector;
 	return (
 		<Background>
 			{ currentPage && currentPage.elements.map( ( { type, id, ...rest } ) => {
 				const comp = getComponentForType( type );
 				const Comp = comp; // why u do dis, eslint?
+				selector = id; // @todo We should actually assign the selector only if the specific block is selected, too.
 				return (
-					<Element key={ id } onClick={ handleSelectElement( id ) }>
+					<Element id={ 'element-' + id } key={ id } onClick={ handleSelectElement( id ) }>
 						<Comp { ...rest } />
 					</Element>
 				);
 			} ) }
 			{ hasSelection && (
 				<Selection { ...selectionProps } />
+			) }
+			{ hasSelection && selector && (
+				<Moveable
+					target={ document.querySelector( '#element-' + selector + ' > :first-child' ) }
+					pinchThreshold={ 20 }
+					draggable={ true }
+					resizable={ true }
+					rotatable={ true }
+					onDrag={ ( { target, top, left } ) => {
+						target.style.left = `${ left }px`;
+						target.style.top = `${ top }px`;
+					} }
+					onResize={ ( { target, width, height } ) => {
+						target.style.width = `${ width }px`;
+						target.style.height = `${ height }px`;
+					} }
+					onRotate={ ( { target, transform } ) => {
+						target.style.transform = transform;
+					} }
+				/>
 			) }
 		</Background>
 	);
--- a/assets/src/edit-story/components/canvas/page.js
+++ b/assets/src/edit-story/components/canvas/page.js
@@ -66,6 +66,10 @@ function Page() {
 		rotate: selectionProps.rotationAngle,
 	};
 
+	const setStyle = ( target ) => {
+		target.style.transform = `translate(${ frame.translate[ 0 ] }px, ${ frame.translate[ 1 ] }px) rotate(${ frame.rotate }deg)`;
+	};
+
 	return (
 		<Background>
 			{ currentPage && currentPage.elements.map( ( { type, id, ...rest } ) => {
@@ -87,13 +91,12 @@ function Page() {
 			{ displayMoveable && (
 				<Moveable
 					target={ document.querySelector( '.selection' ) }
-					pinchThreshold={ 20 }
 					draggable={ true }
 					resizable={ true }
 					rotatable={ true }
 					onDrag={ ( { target, beforeTranslate } ) => {
 						frame.translate = beforeTranslate;
-						target.style.transform = `translate(${ beforeTranslate[ 0 ] }px, ${ beforeTranslate[ 1 ] }px)`;
+						setStyle( target );
 					} }
 					onDragStart={ ( { set } ) => {
 						set( frame.translate );
@@ -101,13 +104,19 @@ function Page() {
 					onDragEnd={ () => {
 						setPropertiesOnSelectedElements( { x: selectionProps.x + frame.translate[ 0 ], y: selectionProps.y + frame.translate[ 1 ] } );
 					} }
-					onResizeStart={ ( { target, set, setOrigin, dragStart } ) => {
-						// @todo
+					onResizeStart={ ( { setOrigin, dragStart } ) => {
+						setOrigin( [ '%', '%' ] );
+						if ( dragStart ) {
+							dragStart.set( frame.translate );
+						}
 					} }
 					onResize={ ( { target, width, height, drag } ) => {
+						// @todo This is not really working, it's sliding away.
 						target.style.width = `${ width }px`;
 						target.style.height = `${ height }px`;
-						// @todo properly by setting frame, too.
+						console.log( drag.beforeTranslate );
+						//frame.translate = drag.beforeTranslate;
+						setStyle( target );
 					} }
 					onResizeEnd={ () => {
 						// @todo Set the correct width/height.
@@ -117,13 +126,12 @@ function Page() {
 					} }
 					onRotate={ ( { target, beforeRotate } ) => {
 						frame.rotate = beforeRotate;
-						target.style.transform = `rotate(${ beforeRotate }deg)`;
+						setStyle( target );
 					} }
 					onRotateEnd={ () => {
 						setPropertiesOnSelectedElements( { rotationAngle: frame.rotate } );
 					} }
 					origin={ false }
-					pinchable={ true }
 					keepRatio={ 'image' === selectedElements[ 0 ].type }
 					renderDirections={ 'image' === selectedElements[ 0 ].type ? [ 'nw', 'ne', 'sw', 'se' ] : null }
 				/>
--- a/assets/src/edit-story/components/canvas/page.js
+++ b/assets/src/edit-story/components/canvas/page.js
@@ -7,7 +7,7 @@ import Moveable from 'react-moveable';
 /**
  * WordPress dependencies
  */
-import { useCallback, useEffect } from '@wordpress/element';
+import { useCallback, useEffect, createRef, useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -40,6 +40,13 @@ const Element = styled.div`
 `;
 
 function Page() {
+	const selectionRef = createRef();
+	const [ targetEl, setTargetEl ] = useState( null );
+
+	useEffect( () => {
+		setTargetEl( selectionRef.current );
+	}, [ selectionRef ] );
+
 	const {
 		state: { currentPage, hasSelection, selectedElements },
 		actions: { clearSelection, selectElementById, setPropertiesOnSelectedElements, toggleElementIdInSelection },
@@ -86,11 +93,11 @@ function Page() {
 				);
 			} ) }
 			{ hasSelection && (
-				<Selection { ...selectionProps } />
+				<Selection { ...selectionProps } ref={ selectionRef } />
 			) }
-			{ displayMoveable && (
+			{ displayMoveable && targetEl && (
 				<Moveable
-					target={ document.querySelector( '.selection' ) }
+					target={ targetEl }
 					draggable={ true }
 					resizable={ true }
 					rotatable={ true }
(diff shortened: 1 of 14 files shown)


assets/src/edit-story/components/canvas/page.js
```diff
@@ -72,16 +85,67 @@ function Page() {
...
+			y: y - padding,
+			width: width + ( 2 * padding ),
+			height: height + ( 2 * padding ),
+			rotationAngle,
+		};
+	}
 	// return x,y,width,height that will encompass all elements in list
-	const { x1, y1, x2, y2 } = list
-		// first convert x1,y1 as upper left and x2,y2 as lower right
+	const { x1, y1, x2, y2, rotationAngle } = list
```
dvoytenko: IMHO we should just skip this entirely for now. The multi-select should be able to take care of multiple selection for us, we just need to render "stub" elements. I think I even noticed that Moveable draws outlines themselves so we may not even need a border.

assets/src/edit-story/components/canvas/page.js
```diff
@@ -72,16 +85,67 @@ function Page() {
 			{ hasSelection && (
 				<Selection { ...selectionProps } />
 			) }
+			{ displayMoveable && (
+				<Moveable
+					target={ document.querySelector( '.selection' ) }
```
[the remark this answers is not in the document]
miina: Correct, this assumes a separate selection frame, the idea being simplifying the logic for now to avoid worrying about updating the original element and its "insides" while an event is happening.

Do you mean using `ref` for the target instead of `document.querySelector`? Tried using refs initially but this didn't seem to work right away so moved on to avoid debugging too much. Will circle back to it a bit later though. I'll look a bit more into the Moveable documentation as well.

assets/src/edit-story/components/canvas/page.js
```diff
@@ -72,16 +85,67 @@ function Page() {
 			{ hasSelection && (
 				<Selection { ...selectionProps } />
 			) }
+			{ displayMoveable && (
```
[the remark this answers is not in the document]
miina: Yes, definitely makes sense to move it separately 👍

assets/src/edit-story/components/canvas/page.js
```diff
@@ -72,16 +85,67 @@ function Page() {
...
+						set( frame.translate );
+					} }
+					onDragEnd={ () => {
+						setPropertiesOnSelectedElements( { x: selectionProps.x + frame.translate[ 0 ], y: selectionProps.y + frame.translate[ 1 ] } );
+					} }
+					onResizeStart={ ( { target, set, setOrigin, dragStart } ) => {
+						// @todo
+					} }
+					onResize={ ( { target, width, height, drag } ) => {
+						target.style.width = `${ width }px`;
```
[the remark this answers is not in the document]
miina: As far as I've seen it's using `width` and `height` in guidelines and has a separate scaling feature instead in addition to resizing. In case of some elements (such as image) which should always keep their ratio and only have resizing handles in the corners (based on Figma) then using scalable instead might be an option.

Will check the options.

assets/src/edit-story/components/canvas/page.js
```diff
@@ -72,16 +85,67 @@ function Page() {
...
+					} }
+					onRotateStart={ ( { set } ) => {
+						set( frame.rotate );
+					} }
+					onRotate={ ( { target, beforeRotate } ) => {
+						frame.rotate = beforeRotate;
+						target.style.transform = `rotate(${ beforeRotate }deg)`;
+					} }
+					onRotateEnd={ () => {
+						setPropertiesOnSelectedElements( { rotationAngle: frame.rotate } );
```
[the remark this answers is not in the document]
miina: Good point. Currently, the implementation just ignores the multi-selection completely. Haven't look into the [grouping + Moveable](https://github.com/daybrush/moveable/blob/master/handbook/handbook.md#toc-group) yet but hopefully they do consider this.

miina:
Note: Currently looking into getting resizing working properly. Due to absolute positioning, the position also needs an update when resizing, always when the element is rotated and otherwise also for every resizing handle except for the bottom, right, and bottom-right. In the previous version of AMP Stories, we wrote [all the logic of adjusting the position while resizing ourselves](https://github.com/ampproject/amp-wp/blob/d26786376615c4310f85dd60fb0e71e2d5df5b10/assets/src/stories-editor/components/resizable-box/index.js#L165-L282) but it was quite often buggy and influenced by other changes in the codebase.

In theory, Moveable should work with absolute positioning as well, however, not currently working here. Will continue looking into this, perhaps I'm missing something obvious.

Currently "sliding" around due to missing position compensating:
resizing-sliding

miina:
Looks like the obvious thing might have been missing `pinchable` which allows using drag events while resizing. However, there's still some odd sliding happening -- see the upper left corner continuously sliding:
odd-behaviour

Might create a support ticket to the repo unless I figure it out tomorrow quickly enough, the developer seems to be very responsive and active.

assets/src/edit-story/components/canvas/page.js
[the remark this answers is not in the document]
dvoytenko: Agree. Let's circle back with ref->state.

dvoytenko:
> Note: Currently looking into getting resizing working properly. Due to absolute positioning, the position also needs an update when resizing, always when the element is rotated and otherwise also for every resizing handle except for the bottom, right, and bottom-right. 

I believe `drag.beforeTranslate` is used for this. It'd look something like this:

```
target.style.top = `${oldTop + drag.beforeTranslate[0]}px`;
target.style.left = `${oldLeft + drag.beforeTranslate[1]}px`;
```

assets/src/edit-story/components/canvas/page.js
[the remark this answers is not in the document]
miina: That's replaced now, using ref insteaf of `document.querySelector`.

miina:
> I believe `drag.beforeTranslate` is used for this. It'd look something like this:

Seems so, that's currently [used in the PR as well](https://github.com/ampproject/amp-wp/pull/3798/files#diff-3688303d740704ce521b8f44dbb09923R124), also tried a few variations, still sliding a little bit. Will look more into this.

assets/src/edit-story/components/canvas/page.js
```diff
@@ -39,9 +40,16 @@ const Element = styled.div`
 `;
 
 function Page() {
+	const selectionRef = createRef();
+	const [ targetEl, setTargetEl ] = useState( null );
+
+	useEffect( () => {
+		setTargetEl( selectionRef.current );
```
dvoytenko: I noticed this in couple of places: is there something bad from just doing call ref-callback? E.g.

```
<X ref={current => setTargetEl(current)}
```

assets/src/edit-story/components/canvas/page.js
```diff
@@ -39,9 +40,16 @@ const Element = styled.div`
...
+	}, [ selectionRef ] );
```
dvoytenko: `selectionRef` is guaranteed to never change. Should probably be `[selectionRef.current]` instead? Or ref-callback I mentioned above?

assets/src/edit-story/components/canvas/page.js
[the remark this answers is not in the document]
dvoytenko: I looked around Moveable - I think width/height for now are good enough - they do a lot of mutations themselves that would bust layout. We can circle back to this later if we find a better way.

assets/src/edit-story/components/canvas/page.js
miina: Since `selectionRef` is an object then it would probably always be considered as not equal and always run. Wouldn't `selectionRef.current` always be `null` at this point instead?

assets/src/edit-story/components/canvas/page.js
miina: Hmm, I'm not sure if there's a difference/downside doing that. ^  @barklund Any thoughts here?

miina:
Here are some of the initial follow-up tasks for this issue:
- Move Moveable into a separate component
- Hide handles while dragging / resizing, only display outline
- Minimum limits for resizing
- Limit dragging out of the Page (at least to some extent for now)
- Allow group dragging
- Allow group rotation
- Allow group resizing

Perhaps creating two separate issue for these would be good, one for general/UX improvements:
- Move Moveable into a separate component
- Hide handles while dragging / resizing, only display outline
- Minimum limits for resizing (not allowing resizing to non-existing)
- Limit dragging out of the Page (at least to some extent for now)

The other for enabling group actions:
- Allow group dragging
- Allow group rotation
- Allow group resizing

Does this sound reasonable?

assets/src/edit-story/components/canvas/page.js
dvoytenko: The result of `useRef()`  is guaranteed to always be the same object. E.g. `Object.is(old, new) === true`. If we use `selectionRef.current` it will first start as `null` and then will re-update itself with the actual element value. You'd have one extra render in that case, which is not ideal. A callback-ref could thus be better...

assets/src/edit-story/components/canvas/page.js
dvoytenko: Usually the main downside to using a ref-callback is that you don't have `ref.current` handy. But since we're writing the value into an easily-available state, we will have it anyway. So I think there's no real downside that I can see at this point. You'd remove an extra re-render this way vs `useEffect -> setState` approach. Not a huge deal, but still. Also the code itself would actually be simpler: `<X ref={setTargetEl}>` - so it looks pretty natural, imho.

assets/src/edit-story/components/canvas/page.js
miina: Doesn't `Object.is( old, new )` always result in `false` since the references are compared here not the actual values of the two objects? E.g. if you do `a = createRef(); b = createRef()`, then `Object.is( a, b )` is always `false`.

(Note: just tested out with `selectionRef.current` and it doesn't seem to update -- the `targetEl` always stays `null` and `useEffect` is run just once. Perhaps I'm missing something & you tried it out and it worked for you?)

dvoytenko:
@miina this all sounds very reasonable. I'd only add: "display of guidelines" item.

assets/src/edit-story/components/canvas/page.js
miina: 👍 Will update this and then merge this PR.

assets/src/edit-story/components/canvas/page.js
dvoytenko: With `[selectionRef]` the `useEffect` will definitely run only once since it never changes. In `a = useRef(); b = useRef()`, the `Object.is(a, b)` would definitely be `false`. But the `a = useRef()` in the same position will always return the same `a` object for any number of render calls. I'll try it now and see what's happening there. Possibly I'm missing something in the code. Does it work for you for the `selectionRef`?

miina: Yes, it does work with `selectionRef`.

dvoytenko: I just realized that the code uses `createRef`. I thought it was `useRef` hook instead? `createRef` indeed would create the new reference for each render. But that's not what you want here, is it? With hooks a `createRef` is no different that just a simple `let/const` var?

miina: Okay, just realized that the code is using `createRef` instead of `useRef` which seems to be creating a new object on each render.

dvoytenko: I tried with `useRef` just now and it all seems to work pretty well. Also I tried to just using `<Selection ref={setTargetEl}>` - that also works well. I guess either way works well.

miina:
>  I'd only add: "display of guidelines" item.

By guidelines, do you mean the snapping lines? We have a separate issue for that currently: https://github.com/ampproject/amp-wp/issues/3773, which seems to be P1 as well, others being P0, perhaps good to keep it separately? (Let me know if you meant something else by Guidelines.)

assets/src/edit-story/components/canvas/page.js
miina: The last commit changed the code to use `<Selection ref={ setTargetEl }>` instead of `useEffect`. Thank you for going into such detail in reviewing btw!

Will merge this as a basis for future iterations.

dvoytenko: Thanks! I'm just trying to learn codebase this way :)

dvoytenko:
> Currently "sliding" around due to missing position compensating:

Definitely seeing some odd offsetting. But I think this is good to merge - we'll have to figure it out later.

miina:
> Definitely seeing some odd offsetting. But I think this is good to merge - we'll have to figure it out later.

What's odd is that it seems to be sliding with the image element only. Also created a sandbox for testing it out and asking the dev of the library to check what's wrong but it works as expected here: https://codesandbox.io/s/react-moveable-sliding-74s5q

dvoytenko:
> > I'd only add: "display of guidelines" item.
> 
> By guidelines, do you mean the snapping lines? We have a separate issue for that currently: #3773, which seems to be P1 as well, others being P0, perhaps good to keep it separately? (Let me know if you meant something else by Guidelines.)

Great. Didn't know it was already there. Let's keep separately.

dvoytenko:
> > Definitely seeing some odd offsetting. But I think this is good to merge - we'll have to figure it out later.
> 
> What's odd is that it seems to be sliding with the image element only. Also created a sandbox for testing it out and asking the dev of the library to check what's wrong but it works as expected here: https://codesandbox.io/s/react-moveable-sliding-74s5q

Really strange, but I was just going to check it out, so I pulled the merged version and I'm no longer seeing any drifting effect from the animations above. Not sure if it's the code state or maybe something environmental interfering? Sometimes I see some weirdness when I collapse the outline to 0, but the same can be seen in the code snippet as well and probably not a real issue.

miina:
Discovered the issue in my case: Chrome was zoomed out to 75% and this made it slide -- I've also let the [developer know](https://github.com/daybrush/moveable/issues/105#issuecomment-558990727). Works as expected when zoomed in and 100%.

In your case the "fix" might possibly be due to updating the version of `react-moveable` package -- apparently it was sliding a little bit in earlier versions but got fixed last Friday (version `14.5`) (unless you already had that version before).

It's still odd that only in an image's case it slides. Does it slide for you when zoomed out and resizing image when rotated?

dvoytenko:
@miina yes, definitely scales when the page is zoomed-in/zoomed-out. Odd, but I guess not a big issue for us since it's not entirely an expected mode.

miina:
@dvoytenko There is a new version of the package installed now which should fix the issue, it did at least for me locally and for the dev of the package. It got merged within #3841, looks like it's still not 100% fixed then.

PR #3804, opened 2019-11-21
Adding in-place editing with RTE

### LINKED ISSUE #3800: [Research] In-place editing
## Feature description

A conflict of needs occur, when you both want the ability to drag a text element as well as select text for editing. The two interactions are indistinguishable on the surface.

The previous editor used double click to enable object editing, and then you had to click either the border or deselect/reselect to be able to drag the element around on the canvas.

Another approach could be the one used by Google Slides and similar tools. In order to drag an object, you have to click in certain places - if you just click/drag in the center of a text element, you select the text / try to insert text. We can also even have a physical anchor-icon placed next to the element that you need to use to drag the element around on the page.

In the most recent version of the Figma design document, a similar double-click-to-edit approach is suggested for editing images - in this case editing is changing the focal point and maybe even the zoom inside the available viewbox. The Figma sketch suggests a view change when double clicking (graying out the background), so it's clear that you're in “edit mode”. This same visual style can be used for other elements, text elements in particular.

For either case it’s necessary to define some sort of link from an element to the element’s editable version. Who handles this switching? Who checks if edit mode is even supported by the given element (as not all elements will have that)?

These ideas must be explored and in particular merged with whatever transformation library we go with in the above. This is however also a UX discussion.

---------------

## Acceptance criteria

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

barklund:
## Summary

*Fixes #3800 and #3808.*

This PR reorganizes elements and allows for elements to have an editable mode (`hasEditMode`). If so, the main canvas allows users to double-click an element to enter edit mode for said element.

Edit mode does 4 things:

* Selects only the element in question (and thus keeps the editing panels for this element open in the inspector). If you had multiple elements selected when double-clicking one of them, selection is thus changed.
* Disables all other canvas interaction (you can't select other elements, can't change page, add a page, delete/duplicate page, etc) and shows this by CSS ignoring pointer events, fading out all the elements and changes the background color to a brighter color.
* Adds a click event to the background (and thus clicking anywhere on the canvas including the rest of the page or any of the previously mentioned disabled components) that will exit edit mode.
* Swaps the select element component from its `<Element.Display />` to its `<Element.Edit />` with the exact same attributes.

The edit element component can update its own properties by invoking `setPropertiesOnSelectedElements` from the `useStory` hook and otherwise present whatever is needed for the user to interact.

It is important for the edit element component to capture and not propagate any click events, that shouldn't exit the edit mode (e.g. clicking inside the RTE editor when editing text).

It would make sense to add keyboard functionality here where ENTER can be used to enter and exit edit mode and persists the change when existing whereas ESCAPE while in edit mode would exit the edit mode and cancel any changes made. This work has not been made nor prepared, and is left as a future refactoring if UX and Product agrees.

## Checklist

- [x] ~My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).~


assets/src/edit-story/components/canvas/page.js
```diff
@@ -44,32 +45,59 @@ function Page() {
...
+				// Are we editing this element, display this as Edit component.
+				if ( editingElement === id ) {
+					return (
+						<Element key={ id }>
+							<Edit { ...rest } />
+						</Element>
+					);
+				}
+
+				// Are we editing some other element, display this as passive Display.
```
dvoytenko: Could you clarify what's this mode does exactly?

assets/src/edit-story/elements/image/display.js
```diff
@@ -0,0 +1,54 @@
...
+	overflow: hidden;
+`;
+
+const Img = styled.img`
+	position: relative;
+	${ ImageWithScale }
+`;
+
+function ImageDisplay( { src, origRatio, width, height, x, y, scale, offsetX, offsetY } ) {
+	const actualRatio = width / height * PAGE_WIDTH / PAGE_HEIGHT;
```
dvoytenko: How does this formula work? Why is page's ratio necessary for an image?

assets/src/edit-story/utils/useDoubleClick.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * WordPress dependencies
+ */
+import { useEffect, useState } from '@wordpress/element';
+
+const useDoubleClick = ( onClick, onDoubleClick ) => {
```
dvoytenko: There's a discussion in #3800 re: the editing condition triggering. If we consider the triggering to be based on a single click (or Enter key) on a selected element, we may not need a special double-click handling?

barklund: Yes, I removed the double-click usage from the codebase, but (for now) kept the util hook. If image editing is also to be made without double-click, I'll remove it completely.

assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -30,25 +30,31 @@ const Background = styled.div`
...
+		<Background isPassive={ isEditing } onClick={ onClick || null }>
 			<Area area="page">
 				<Page />
 			</Area>
-			<Area area="meta">
+			<Area area="meta" isPassive={ isEditing }>
 				<Meta />
 			</Area>
-			<Area area="carrousel">
+			<Area area="carrousel" isPassive={ isEditing }>
```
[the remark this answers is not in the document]
barklund: It means that the buttons and icons inside those sections are in-active and will allow clicks to pass through them. And any click on the background will exist edit mode. So yes, you'd have to exit edit mode explicitly before being able to add a page, delete a page, change page, etc. However, a click anywhere outside the edited element will exit edit mode.

It is still not clear whether we are going to need edit mode in the sense I have create it here at all. Awaiting UX on that part.

assets/src/edit-story/components/canvas/page.js
barklund: `isPassive` here on the individual element just refers to [the element being faded a bit](https://github.com/ampproject/amp-wp/pull/3804/files#diff-3688303d740704ce521b8f44dbb09923R37).

assets/src/edit-story/elements/image/display.js
barklund: `width` and `height` are in percent if the actual page size. Thus, a size of 20x20 doesn't mean a square viewport, rather a viewport that is 20% of the page width and 20% of the page height. To figure out the actual viewport ratio we have to multiply it with the actual page ratio. With a static page ratio of 775/434 a 20x20 viewport image will have an actual ratio of about `1.79`.

assets/src/edit-story/components/canvas/canvasLayout.js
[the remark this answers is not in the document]
dvoytenko: Would clicking on another element in this case automatically "exit" edit mode for the previously selected element, unselect it, and update selection to the new element? Or is it really everything disabled? That'd feel a bit too drastic...

spacedmonkey:
@barklund this PR need rebasing.

spacedmonkey:
I am not sure about the the background change when an element is being edited. 

Screenshot from 2019-11-29 11-27-19

I found it is a little jarring. What do you think @samitron7 ?

assets/src/edit-story/utils/getCaretCharacterOffsetWithin.js
```diff
@@ -0,0 +1,28 @@
+function getCaretCharacterOffsetWithin( element ) {
```
spacedmonkey: Some comments would be useful here.

assets/src/edit-story/elements/text/edit.js
```diff
@@ -0,0 +1,147 @@
...
+			EditorState.createWithContent( stateFromHTML( content ) )
+	);
+	const [ editorState, setEditorState ] = useState( initialState );
+	const mustAddOffset = useRef( offset ? 2 : 0 );
+
+	// This is to allow the finalizing useEffect to *not* depend on editorState,
+	// as would otherwise be a lint error.
+	const lastKnownState = useRef( null );
+
+	// This filters out illegal content on paste and updates state accordingly.
```
spacedmonkey: What does illegal content mean in this context?

assets/src/edit-story/elements/text/display.js
```diff
@@ -0,0 +1,129 @@
...
+		setEditingElementWithState( id, { offset: getCaretCharacterOffsetWithin( element.current ) } );
+		evt.stopPropagation();
+	};
+	const handleKeyDown = ( evt ) => {
+		if ( evt.metaKey || evt.altKey || evt.ctrlKey ) {
+			// Some modifier (except shift) was pressed. Ignore and bubble
+			return;
+		}
+
+		if ( evt.key === 'Enter' ) {
```
spacedmonkey: Isn't better to do a key code check here like [this](https://github.com/ampproject/amp-wp/blob/8ae3c05f0f525ed85305527777d5ef07d40a3ae0/assets/src/stories-editor/components/draggable-text.js#L68)

assets/src/edit-story/elements/text/display.js
```diff
@@ -0,0 +1,129 @@
...
+			// Enter editing without writing or selecting anything
+			setEditingElement( id );
+			evt.stopPropagation();
+			// Make sure no actual Enter is pressed
+			evt.preventDefault();
+		} else if ( /^\w$/.test( evt.key ) ) {
```
spacedmonkey: What about special characters here? All punctuation, none roman characters and spaces need to be considered here.

assets/src/edit-story/elements/image/display.js
spacedmonkey: A comment above explaining this might be useful here.

assets/src/edit-story/elements/text/display.js
barklund: [`keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode).

assets/src/edit-story/elements/text/display.js
barklund: As mentioned in the `TODO` in the very next line, yes.

assets/src/edit-story/elements/image/display.js
barklund: [Added comment](https://github.com/ampproject/amp-wp/pull/3804/files#diff-3351695c7709cca397b8b3229f4a4a94R27-R28).

assets/src/edit-story/elements/text/edit.js
barklund: [Added some context](https://github.com/ampproject/amp-wp/pull/3804/files#diff-5f4f624867fd96363e2603afb736f301R85-R86).

barklund: [Added some context](https://github.com/ampproject/amp-wp/pull/3804/files#diff-5f4f624867fd96363e2603afb736f301R85-R86).

assets/src/edit-story/elements/text/display.js
spacedmonkey: Sorry, I didn't get that from your comment.

spacedmonkey:
Comparing this PR to google slides. There is a mega difference. To edit text content, requires two click, one click to select the text element and second to focus into the text editting mode. In google slides, this is not how it works. One click both select and put user into editting mode. It just depends where they click. Text elements are only draggable at the edge. A click cursor appears when moving over the corner to indicate that you are in a draggable appear. 

There need to be an IB for this feature so we can discuss this.

assets/src/edit-story/components/canvas/canvasLayout.js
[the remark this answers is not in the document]
barklund: I've updated this part to match the requirements in UX. All other buttons are now still enabled and any click outside the editing element will both exit editing (with `useCapture=true`) as well as do whatever the click would normally do.

assets/src/edit-story/utils/useDoubleClick.js
barklund: Image editing is still done with double-click in the UX and thus stays. I have added documentation.

barklund:
> I am not sure about the the background change when an element is being edited.
 I have now updated as per UX.

barklund:
> @barklund this PR need rebasing.

This has been merged with `develop-stories` and overhauled.

barklund:
This PR is now ready for a full review.

Do note that image editing is not correct and does not follow UX. Entering and exiting image editing does though, but the actual editing is not proper. This is left as a future task.

However, do ignore the complexity of `page.js` - it does way too much and needs to be split up (even further), but as @miina is working on the `Moveable` part of it, I didn't want to touch that part needlessly too much.

assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -30,15 +30,16 @@ const Background = styled.div`
...
-  width: 100%;
+	grid-area: ${ ( { area } ) => area };
+	height: 100%;
+	width: 100%;
 `;
 
 function CanvasLayout() {
-	const { state: { backgroundClickHandler } } = useCanvas();
+	const { state: { isEditing, backgroundClickHandler } } = useCanvas();
+	const onClick = isEditing ? () => {} : backgroundClickHandler;
```
barklund: Hm, this might not be relevant anymore actually. Will check later.

assets/src/edit-story/components/canvas/page.js
```diff
@@ -33,22 +33,29 @@ const Selection = styled.div`
...
+	const moveable = useRef();
+
+	const setClickHandler = useCallback(
+		( id, handler ) =>
+			setClickHandlers( ( handlers ) => ( { ...handlers, [ id ]: handler } ) ),
+		[ setClickHandlers ] );
+
+	const getClickHandler = useCallback(
+		( id ) => clickHandlers[ id ] || ( () => {} ),
+		[ clickHandlers ] );
```
barklund: This part about click handlers in particular is a bit messy. We might be able to do it in a better way.

And maybe @miina's efforts will make this completely superfluous. It is needed in the version as Moveable is working on a selection object next to the actual objects, but as I understand it, Moveable will in the future work on the actual objects and thus we hopefully don't need this propagation of clicks from one element to another.

assets/src/edit-story/components/canvas/page.js
```diff
@@ -71,26 +76,50 @@ function Page() {
...
+		const clickHandler = getClickHandler( selectedElement.id );
+		if ( clickHandler ) {
+			// Make selection box "transparent", click element and make selection opaque again,
+			evt.persist();
+			targetEl.style.pointerEvents = 'none';
+			clickHandler( evt );
+			targetEl.style.pointerEvents = '';
+		}
+		evt.stopPropagation();
+	}, [ getClickHandler, selectedElement, targetEl ] );
```
barklund: This is the second part of the messy bit, that we can hopefully skip in the future setup.

assets/src/edit-story/components/canvas/page.js
```diff
@@ -166,13 +195,13 @@ function getUnionSelection( list, padding = 0 ) {
 		.map( ( { x, y, width, height } ) => ( { x1: x, y1: y, x2: x + width, y2: y + height } ) )
 		// then reduce to a single object by finding lowest {x,y}1 and highest {x,y}2
 		.reduce(
-			( el, sum ) => ( {
+			( sum, el ) => ( {
```
barklund: Luckily it didn't matter this time around 😜.

assets/src/edit-story/elements/text/display.js
```diff
@@ -0,0 +1,139 @@
+/**
```
barklund: This file is a bit long, but I don't see how it can be easily compressed. It's fairly well-contained IMHO, but maybe some of the keyboard click handling can be moved to a helper hook.

assets/src/edit-story/utils/getCaretCharacterOffsetWithin.js
```diff
@@ -0,0 +1,60 @@
...
+	if ( typeof win.getSelection !== 'undefined' ) {
+		sel = win.getSelection();
+		if ( sel.rangeCount > 0 ) {
+			let range = win.getSelection().getRangeAt( 0 );
+			if ( clientX && clientY ) {
+				if ( doc.caretPositionFromPoint ) {
+					range = document.caretPositionFromPoint( clientX, clientY );
+				} else if ( doc.caretRangeFromPoint ) {
+					range = document.caretRangeFromPoint( clientX, clientY );
+				}
```
barklund: So, both of these methods are non-standard, but cover most browsers currently.

[The former, `caretPositionFromPoint`, is drafted](https://drafts.csswg.org/cssom-view/#dom-document-caretpositionfrompoint), but [not implemented in Chrome or Edge](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/caretPositionFromPoint#Browser_compatibility) so far.

assets/src/edit-story/utils/useDoubleClick.js
```diff
@@ -0,0 +1,86 @@
+/**
+ * WordPress dependencies
+ */
+import { useCallback, useEffect, useState } from '@wordpress/element';
+
+// Disable reason: Should not check indentation in code samples inside
+// markdown, but it does.
+
+/* eslint-disable jsdoc/check-indentation */
```
barklund: Why isn't this working properly? Does WP eslint have weird config for this rule?

assets/src/edit-story/components/canvas/canvasLayout.js
barklund: Solved it!

assets/src/edit-story/elements/text/display.js
barklund: Actually some of the keyboard stuff will be removed when `<Focusable />` is included here. Optimizations will wait for then.

assets/src/edit-story/components/canvas/element.js
```diff
@@ -0,0 +1,67 @@
...
+	addNodeForElement,
+	setClickHandler,
+	handleSelectElement,
+	element: {
+		id,
+		type,
+		...rest
+	},
+} ) {
+	const { Display, Edit } = getDefinitionForType( type );
```
dvoytenko: Can there be an element that's not editable (or deep-editable as it might be)?

assets/src/edit-story/components/canvas/element.js
```diff
@@ -0,0 +1,67 @@
...
+			</Wrapper>
+		);
+	}
+
+	return (
+		<Wrapper
+			onClick={ ( evt ) => handleSelectElement( id, evt ) }
+			ref={ element }
+		>
+			<Display { ...props } setClickHandler={ setClickHandler } />
```
dvoytenko: Is there an explanation in one sentence what's the role of the `setClickHandler`?

barklund: It's a temporary work-around to forward click-events from the selection object to the actual on-page element. In separate work, that @miina's doing, the actual elements will be child elements of the selection and this work-around can be removed then.

assets/src/edit-story/components/canvas/element.js
barklund: Yes and in that case `Edit` will just be `undefined`. Elements only enter edit-mode, when they themselves request it based on user interaction.

assets/src/edit-story/components/canvas/element.js
barklund: > Hmm. I'm a bit lost. By "selection object" do you mean a selection DOM element?

I mean the React component `<Selection />`. But this is temporary anyway, this whole mess will be fixed once merged with @miina's efforts elsewere. Better review it then.

assets/src/edit-story/components/canvas/element.js
```diff
@@ -0,0 +1,67 @@
...
+		type,
+		...rest
+	},
+} ) {
+	const { Display, Edit } = getDefinitionForType( type );
+	const element = useRef();
+	const props = { ...rest, id };
+
+	useLayoutEffect( () => {
+		addNodeForElement( id, element.current );
```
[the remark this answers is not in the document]
barklund: It's weird and pretty bad structure. It needs an overhaul when someone thinks of a better way to do it. Yes, it should probably be `setNodeForElement`. I'll rename it here, but this will probably also be completely changed when merged with @miina's stuff.

assets/src/edit-story/components/canvas/useEditingElement.js
```diff
@@ -0,0 +1,74 @@
+/**
+ * WordPress dependencies
+ */
+import { useState, useCallback, useEffect } from '@wordpress/element';
+
+function useEditingElement() {
+	const [ editingElement, setEditingElement ] = useState( null );
```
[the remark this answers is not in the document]
barklund: That's a UX/Product question. Personally, I don't think so.

Often edits "inside" an object are committed as a single history state, e.g. if changing text in a text layer in Photoshop or in a cell in Sheets. You can go back to before the edit, but you can't back to a partial edit and thus no need to go back to the cell/element being in an editable state.

Should we loop in @jasti for this? Or just @samitron7?

assets/src/edit-story/components/canvas/page.js
```diff
@@ -71,26 +76,50 @@ function Page() {
...
+	// TODO: proper multi-select
+	// TODO: don't invoke if *any* transformation has happened.
+	const handleSelectionClick = useCallback( ( evt ) => {
+		const clickHandler = getClickHandler( selectedElement.id );
+		if ( clickHandler ) {
+			// Make selection box "transparent", click element and make selection opaque again,
+			evt.persist();
+			targetEl.style.pointerEvents = 'none';
+			clickHandler( evt );
+			targetEl.style.pointerEvents = '';
```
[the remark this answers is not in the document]
barklund: If disabled, Moveable doesn't work. But again, this is temporary.

assets/src/edit-story/components/library/mediaLibrary.js
```diff
@@ -45,14 +45,15 @@ function MediaLibrary( { onInsert } ) {
...
-			{ media.map( ( { src } ) => (
+			{ media.map( ( { src, origWidth, origHeight } ) => (
 				<Image
 					key={ src }
 					src={ src }
 					width={ 150 }
 					height={ 150 }
 					onClick={ () => onInsert( 'image', {
 						src,
+						origRatio: origWidth / origHeight,
```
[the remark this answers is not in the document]
barklund: I believe @spacedmonkey has already done this in his work with media library.

assets/src/edit-story/elements/text/display.js
```diff
@@ -0,0 +1,139 @@
...
+
+	const handleKeyDown = ( evt ) => {
+		if ( evt.metaKey || evt.altKey || evt.ctrlKey ) {
+			// Some modifier (except shift) was pressed. Ignore and bubble
+			return;
+		}
+
+		if ( evt.key === 'Enter' ) {
+			// Enter editing without writing or selecting anything
+			setEditingElement( id );
```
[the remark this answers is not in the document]
barklund: Besides the need to pass additional state to the editable state, there's how edit mode is even entered.

Text elements can enter edit mode:

- If clicked a single time
- If a character key is pressed
- If enter is pressed

Image elements can enter edit mode:

- If double-clicked

For now, no other element can enter edit-mode (as we don't really have others).

These interactions are so different, that to me it made sense to have the elements themselves dictate when and if they should switch to edit mode and not the page manager.

When we get further and have more elements that can enter edit mode, we might be able to generalize and refactor, but for now I feel this makes the most sense.

assets/src/edit-story/utils/useDoubleClick.js
```diff
@@ -0,0 +1,86 @@
...
+ * @param {number}   ms             Timeout in ms to wait - defaults to 200.
+ * @return {Function} Handler retrieval function to get an onClick listener (invoke with unique value).
+ */
+const useDoubleClick = ( onSingleClick, onDoubleClick, ms = null ) => {
+	const [ target, setTarget ] = useState( null );
+	const [ lastEvent, setLastEvent ] = useState( null );
+	const getHandler = useCallback( ( newTarget ) => ( evt ) => {
+		evt.stopPropagation();
+
+		if ( target !== newTarget ) {
```
miina: I guess it's intended that in case of an image, it first has to be selected and only then double click is required to go to edit mode?

barklund: It's actually because I wanted this handler to be able to be used in a component will multiple children, like [in the example given in the description of the hook](https://github.com/ampproject/amp-wp/pull/3804/files/575fccf668fbff8c6a004ea45180ac9f9ee9fb4e#diff-55be7466fec1b29684bc0dcb019ca439R23-R39).

If it just counted clicks, it would count a click on one item and then another fast click on another as a double click.

As you can't use a variable number of hooks in a component (it always has to be the exact same amount), this was the best way to create it.

assets/src/edit-story/components/canvas/useEditingElement.js
```diff
@@ -0,0 +1,76 @@
...
+
+	const setEditingElementWithoutState = useCallback(
+		( id ) => {
+			setEditingElement( id );
+			setEditingElementState( {} );
+		},
+		[],
+	);
+
+	const setEditingElementWithState = useCallback(
```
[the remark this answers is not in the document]
barklund: This is used in two instances, all when editing a text field:

- When you click to enter a text field, the initial pointer inside the WYSIWYG has to be placed where the mouse pointer is currently located - this information is passed as state to the editable component.
- When you type a character while having a text field focused, the text field must enter edit mode and replace the entire contents with the character pressed. This character and the clearing of the text field is passed as state.

Essentially, it's the only way for the Display component to tell the Edit component how to initialize itself, if there are some special circumstances.

assets/src/edit-story/components/canvas/canvasProvider.js
```diff
@@ -3,43 +3,43 @@
...
+		setEditingElementWithState,
+		setEditingElementWithoutState,
+		clearEditing,
+		setNodeForElement,
+	} = useEditingElement();
+
+	const [
+		backgroundClickHandler,
+		setBackgroundClickHandler,
+		clearBackgroundClickHandler,
```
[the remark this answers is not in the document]
barklund: I originally thought it would be used for multiple purposes, (like exiting edit mode), but it turns out to not be the case. I suggest keeping it for now, so we have the option of changing what happens when you click on the background under special circumstances, but it's a good idea to mark it as potentially obsolete.

miina:
@barklund From testing: currently, it looks like an image has to be first selected and only then the double-click works for entering the edit mode. Shouldn't the image go into edit mode without having to be selected first?

barklund:
> @barklund From testing: currently, it looks like an image has to be first selected and only then the double-click works for entering the edit mode. Shouldn't the image go into edit mode without having to be selected first?

I think the UX says you have to click to select it and *then* double-click to enter edit-mode. Also, it's easier so if UX doesn't require it to be changed, I'd rather not change it.

miina:
> I think the UX says you have to click to select it and then double-click to enter edit-mode. Also, it's easier so if UX doesn't require it to be changed, I'd rather not change it.

Hmm, okay. I didn't read out UX specifically stating that. But okay, let's keep it like this for now, it doesn't specifically state otherwise either.

barklund:
> > I think the UX says you have to click to select it and then double-click to enter edit-mode. Also, it's easier so if UX doesn't require it to be changed, I'd rather not change it.
> 
> Hmm, okay. I didn't read out UX specifically stating that. But okay, let's keep it like this for now, it doesn't specifically state otherwise either.

@miina and now it works as you commented after the merge. Not intentionally, but that's how it is now. I can easily change it though. Let's see what people think :)

@dima, @spacedmonkey, @swissspidy, please review - it's much neater now after the merge.

assets/src/edit-story/components/canvas/element.js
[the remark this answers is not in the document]
dvoytenko: I actually just commented on one of @miina's pull requests about the same. If we can set a DOM node on an element, things could get a lot simpler. Otherwise, we'll have to maintain some pretty long forwardRef chains to constantly try to find these mappings.

assets/src/edit-story/components/canvas/useEditingElement.js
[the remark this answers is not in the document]
dvoytenko: I think you're right. But I'll follow up with @samitron7.

assets/src/edit-story/elements/text/display.js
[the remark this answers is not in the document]
dvoytenko: Ok. This makes sense.

assets/src/edit-story/components/canvas/useEditingElement.js
```diff
@@ -0,0 +1,76 @@
...
+
+		const handleClick = ( evt ) => {
+			if ( ! root.contains( evt.target ) ) {
+				clearEditing();
+			}
+		};
+
+		// as soon as something is clicked/touched, check if we should exit edit mode.
+		const doc = root.ownerDocument || root.document;
+		doc.addEventListener( 'click', handleClick, true );
```
dvoytenko: Should we handle Esc right away as well?

assets/src/edit-story/elements/image/display.js
```diff
@@ -0,0 +1,85 @@
...
+
+const Img = styled.img`
+	position: relative;
+	${ ImageWithScale }
+`;
+
+function ImageDisplay( { id, src, origRatio, width, height, x, y, scale, offsetX, offsetY, rotationAngle, forwardedRef, onPointerDown } ) {
+	// Width and height are percent of the actual page dimensions,
+	// Thus 20-by-20 doesn't mean square, but "same as page ratio".
+	const actualRatio = width / height * PAGE_WIDTH / PAGE_HEIGHT;
```
miina: Just realized that currently, any element can be resized to `0`, this would break the logic, I assume?

barklund: Division by zero might cause a black hole or similar singularity, yes. But let's fix that elsewhere.

assets/src/edit-story/components/canvas/useEditingElement.js
barklund: We'll leave that for later pending UX decision on this.

PR #3815, opened 2019-11-24
Optimize is_amp_allowed_attribute checking for reference points

westonruter:
Following up on #3243 we can now optimize `\AMP_Tag_And_Attribute_Sanitizer::is_amp_allowed_attribute()` by looking at the currently-open elements rather than traversing all the ancestor nodes to see whether an element is open. This PR implements that optimization.

### DIFF
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -1693,18 +1693,16 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 		 * https://github.com/ampproject/amphtml/blob/1526498116488/extensions/amp-selector/validator-amp-selector.protoascii#L81-L91
 		 */
 		$descendant_reference_points = [
-			'amp-selector'         => AMP_Allowed_Tags_Generated::get_reference_point_spec( 'AMP-SELECTOR option' ),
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1693,18 +1693,16 @@ private function is_amp_allowed_attribute( DOMAttr $attr_node, $attr_spec_list )
...
 		 */
 		$descendant_reference_points = [
-			'amp-selector'         => AMP_Allowed_Tags_Generated::get_reference_point_spec( 'AMP-SELECTOR option' ),
-			'amp-story-grid-layer' => AMP_Allowed_Tags_Generated::get_reference_point_spec( 'AMP-STORY-GRID-LAYER default' ), // @todo Consider the more restrictive 'AMP-STORY-GRID-LAYER animate-in'.
+			'amp-selector'         => 'AMP-SELECTOR option',
+			'amp-story-grid-layer' => 'AMP-STORY-GRID-LAYER default', // @todo Consider the more restrictive 'AMP-STORY-GRID-LAYER animate-in'.
 		];
-		foreach ( $descendant_reference_points as $ancestor_name => $reference_point_spec ) {
+		foreach ( $descendant_reference_points as $ancestor_name => $reference_point_spec_name ) {
+			if ( empty( $this->open_elements[ $ancestor_name ] ) ) {
```
schlessera: This seems to do what the previous code did, so it's the same thing in faster.

However, I'd note that it is not technically correct, as the relationship between "some ancestor" and the descendant is not clear and could be invalid, depending on which nodes are in-between.

westonruter: I'm trying to think of an example that would be invalid. Do you have one in mind?

Issue #3831, opened 2019-11-27
Allow deleting the selected element(s) via keyboard


miina:
## Feature description

Deleting selected elements (one or more) should be possible via keyboard interaction by pressing "Delete".
Note that this depends on #3826 as it's first necessary to understand the general approach for keyboard events handling.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The user should be able to remove selected elements by clicking the "Delete" key on the keyboard (Backspace on Mac).

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

Issue #3826, opened 2019-11-27
Keyboard handling


barklund:
## Feature description

We need either to have:

1. Global-only keyboard handling, where elements can listen for desired keyboard presses (if focused)
2. Global keyboard handling for global-only interactions (copy, cut, paste, undo, redo, select all) and local keyboard handlers in individual focusable elements (delete, enter, exit, type).

Weigh pros and cons between the two solutions, get consensus and implement.

This is not about creating the actual [keyboard shortcuts](https://github.com/ampproject/amp-wp/issues/3785), but about creating the framework for adding them later.

Ping me @barklund, if you need more details on this.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Select between global-only keyboard handling or global+local keyboard handling
* Implement the selection

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

jauyong:
We need to be careful about making sure the text edit mode does not collide with keyboard shortcuts.

spacedmonkey:
I think there are a different set of handlerings depending on if an element is selected or not.

When a element is selected I would expect the following. 
- Arrow keys would move the element in that direction. 
- ESC key would deselect element 
- DELETE / backspace would remove the selected element. 
- ENTER would make a text element editable. 
- CTRL + C / X  - Would cut and copy the current selected element. 

When an element is selected in the media library
- ESC key would deselect element 
- ENTER would insert the element into the editor and select this new element. 

When a form element is selected in the sidebar. 
- ESC refocuses to the selected element. 

Globally 
I would expect all elements to be able to be tabbed to, with hidden skip jump that take you to different sections quicker, such as the canvas. 

Should DETELE / backspace remove the current page? Maybe after confirming with a dialog? Should all deleting of elements via DELETE / backspace require confirmation.

dvoytenko:
To clarify, copy/paste are probably not the keyboard actions, but the direct `copy`/`paste` events, right?

Initially, it'd seem, for the canvas specifically, it'd be more convenient to have a single set of keyboard/clipboard handlers. But in the current model it might be hard to dispatch all types of actions. `DELETE` and such global actions should be easy enough to handle universally. But there might be some element-specific actions as well? How do you see it? Similar to your `panels` approach? E.g. something like this?

```
Image.actions = {
  'ACTION1': (properties, setProperties, dispatcher) => {...},
};
```

[...]

Issue #3824, opened 2019-11-27
Allow deleting a Page via "Delete" icon


miina:
## Feature description

Currently, there is no way to remove an element (other than Undo after creating it). It should be possible to delete a Page.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The user should be able to remove a Page by clicking the Trash icon above the Page. 

## Implementation brief

* Add a new action for deleting the currently selected Page.
* When the Delete icon is clicked, the currently selected Page will be removed.

## QA testing instructions

* 

## Demo

![deleting](https://user-images.githubusercontent.com/3294597/70433466-20685300-1a82-11ea-82ec-bef36c46174a.gif)

## Changelog entry

*

PR #3837, opened 2019-11-28
Allow deleting Pages/elements.

miina:
## Summary

See #3824 

Makes the "Delete" icon above a Page functional.
Also adds a temporary panel for being able to delete selected elements. This will be addressed properly within #3831 once the approach in #3826 has been confirmed.

## Checklist


assets/src/edit-story/app/story/actions/useAddBlankPage.js
```diff
@@ -15,7 +15,7 @@ function useAddBlankPage( {
 	clearSelection,
 } ) {
 	const addBlankPage = useCallback( () => {
-		const newPage = createPage( { index: pages.length } );
+		const newPage = createPage();
```
miina: Removed adding `index` as a param since otherwise we'd have to update all the index params once the order changes or a Page is deleted. Getting the actual index from the actual `pages` seems to make more sense at this moment.

assets/src/edit-story/panels/actions.js
```diff
@@ -0,0 +1,44 @@
...
+	&:hover {
+		color: ${ ( { theme } ) => theme.colors.danger };
+	}
+`;
+
+function ActionsPanel( { deleteCurrentSelection } ) {
+	return (
+		<Panel>
+			<Title>
+				{ 'Actions' }
```
spacedmonkey: Shouldn't this string be translable?

assets/src/edit-story/panels/actions.js
```diff
@@ -0,0 +1,44 @@
...
+			</Title>
+			<Delete onClick={ deleteCurrentSelection } >
+				{ 'Remove element' }
```
spacedmonkey: Shouldn't this string be translable?

assets/src/edit-story/app/story/actions/useDeletePage.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * WordPress dependencies
+ */
+import { useCallback } from '@wordpress/element';
+
+// @todo this might need more appropriate naming depending on how the deleting Page logic will be.
+// Delete the currently selected Page.
+function useDeletePage( {
```
spacedmonkey: Should this name be changed to `deleteSelectedPage`? Make sure it clear what this function does when you call it.

assets/src/edit-story/panels/actions.js
miina: Yes, it should. Currently, however, no strings are translatable with `amp` domain in the app so I was just consciously ignoring this for now until a PR to make everything translatable. Perhaps there's an issue missing for making everything translatable?

assets/src/edit-story/app/story/actions/useDeletePage.js
miina: Yep, fare point. There's a Todo note as well in the comment 2 lines above to make the naming more appropriate -- it will be possible to delete a Page selecting it from the thumbnails in the carousel and clicking "Delete" as well so wasn't sure which name would be the best considering the future actions.

Can def make it `useDeleteCurrentPage` or something like that instead for now.

assets/src/edit-story/app/story/actions/useDeleteCurrentSelection.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * WordPress dependencies
+ */
+import { useCallback } from '@wordpress/element';
+
+// Delete the elements currently selected.
+function useDeleteCurrentSelection( {
```
spacedmonkey: `useDeleteCurrentSelectedElement` ?

assets/src/edit-story/app/story/actions/useDeletePage.js
spacedmonkey: `useDeleteCurrentSelectedPage`?

assets/src/edit-story/app/story/actions/useDeleteCurrentSelection.js
miina: That's actually for deleting the full selection, even if there's multiple elements selected. That's why the general name of `deleteCurrentSelection`

assets/src/edit-story/app/story/actions/useDeletePage.js
miina: There's the state value for `currentPage`, however, a Page isn't actually added to a selection -- the comment above saying "currently selected Page" is a bit misleading. `useDeleteCurrentPage` would be the most consistent with the `useState` values, I suppose.

Will update this, thanks for noting that.

spacedmonkey: Some jsdocs would be good here too.

assets/src/edit-story/app/story/actions/useDeleteCurrentSelection.js
spacedmonkey: Some jsdocs would be good here too.

spacedmonkey: It is not clear from the name what is being deleted.

assets/src/edit-story/panels/actions.js
spacedmonkey: I have created a follow up ticket regarding translations - https://github.com/ampproject/amp-wp/issues/3842

spacedmonkey:
Shouldn't there be some sort of unit test that goes with this PR?

miina:
> Shouldn't there be some sort of unit test that goes with this PR?

Not until the testing infrastructure gets set up (see #3249), also, note that the functionality in this PR is not necessarily final, it's mainly adding a missing part of PoC

assets/src/edit-story/app/story/actions/useDeleteCurrentSelection.js
miina: Changed to `useDeleteSelectedElements`, this matches with the `selectedElements` state, too.

assets/src/edit-story/app/story/actions/useAddBlankPage.js
miina: @barklund Do you see any concerns with removing the `index` param from here in favor of using index from the actual updated list of pages?

assets/src/edit-story/app/story/storyProvider.js
```diff
@@ -71,6 +75,8 @@ function StoryProvider( { storyId, children } ) {
 			setCurrentPageByIndex,
 			addBlankPage,
 			clearSelection,
+			deleteSelectedElements,
```
dvoytenko: There's a similar question with history work, but maybe it's worth doing reducer/dispatch for all datamodel updates? Totally ok deferring this question until further pull requests. Just brining attention to whether the APIs are starting to grow pretty wide.

miina: Probably better to discuss this under a more general issue. If that's OK for you then I'd prefer merging the PR and later changing together with the other actions if needed. Thoughts?

spacedmonkey: I think that using `reducer/dispatch` is a larger architectural discussion as other part of this implementation would have to be changed. @dvoytenko can you create another ticket for this so we can move the discussion there?

miina:
@dvoytenko @pbakaus Let me know if you have any concerns about merging this PR.

assets/src/edit-story/app/story/storyProvider.js
dvoytenko: I created a new "Definition" story in https://github.com/ampproject/amp-wp/issues/3894. LGTM to this PR.

PR #3863, opened 2019-12-01
Restrict paired browsing to when dev mode is enabled

westonruter:
## Summary

See https://github.com/ampproject/amp-wp/pull/3656/files#r352310999:

> I just realized that this makes every page invalid AMP, because every page is forced to be in dev mode and the custom script is enqueued. Two options I see:
> 
> 1. Short-circuit this function if an user is not logged-in (or rather, if `amp_is_dev_mode()` returns `false`). Since search crawlers will never be logged-in, they will never then see an invalid AMP page. This would also require entire paired browsing app should `wp_die()` if the user is not logged in.
> 2. Go back to only including this JS if the admin bar is showing. Accessing paired browsing here would still need to include the `wp_die()` if `! amp_is_dev_mode()`.

This PR implements the former. This allows a site to continue to allow paired browsing for unauthenticated users by adding plugin code that does `add_filter( 'amp_dev_mode_enabled', '__return_true' );`, though naturally they'd want to use some other condition (e.g. setting a cookie).

See #3365. Amends #3656.

## Checklist

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -2516,6 +2516,11 @@ class AMP_Theme_Support {
 			return;
 		}
 
+		// Paired browsing requires a custom script which in turn requires dev mode.
+		if ( ! amp_is_dev_mode() ) {
+			return;
+		}
+
 		$asset_file   = AMP__DIR__ . '/assets/js/amp-paired-browsing-client.asset.php';
 		$asset        = require $asset_file;
 		$dependencies = $asset['dependencies'];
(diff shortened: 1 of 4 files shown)


pierlon:
While reviewing this I realized that the dialog text does not match the options that are shown if the user can't exit the page:

image

I see two options:

1. Remove the second sentence
2. Only show the second sentence if the 'Exit' button is also shown

westonruter:
> * Remove the second sentence

Good point. I've done this in c10276e. I think it makes sense to remove because the text is redundant as the presence or absence of the buttons say the same thing.

PR #3864, opened 2019-12-01
Try: Use portals instead of iframes in paired browsing, when available

westonruter:
## Summary

As noted in https://github.com/ampproject/amp-wp/pull/3656#issuecomment-559910535:

> You know what paired browsing will eventually be able to make excellent use of? Instead of iframes it could use [portals](https://web.dev/hands-on-portals/), when they become available. This would allow seamlessly exiting the paired browsing context without having to reload the page. I think this will also work when _entering_ paired browsing as well:
> 
> 1. Clicking the “Start paired browsing” link on in the admin bar could create a portal that contains the paired browsing interface.
> 2. The paired browsing interface can be activated.
> 3. The previous window could then be adopted as a portal _inside_ the paired browsing window and displayed as either the AMP or non-AMP frame, depending on if the user started from an AMP or non-AMP page.
> 4. And a second portal can then be created to contain the corresponding window (non-AMP for AMP, and vice versa), and this portal could likewise be used for exiting paired browsing.
> 
> This brings something to light. A user could actually exit to _either_ the AMP or non-AMP window when they are done with paired browsing. At the moment it forces the user to go to the non-AMP version (see https://github.com/ampproject/amp-wp/pull/3656/files#r352265423) but eventually they could be directed to either. There's no need to implement this now, but once portals become an option, this could be a nice touch.

See #3365.

### Outstanding Issues

- [ ] When a window is loaded in a `portal`, it cannot access its `parent` like an `iframe` can; nor can the portal host access `contentWindow` on the portals in the page. Is this by design? If so, we may need to refactor the app to make use of message passing exclusively rather than direct cross-window calls.
- [ ] Windows embedded in portals are not scrollable; in fact, they are completely non-interactive, at least in Chrome Canary 80.0.3981.0. See [tweet question](https://twitter.com/westonruter/status/1201083296362323968).

## Checklist

### DIFF
--- a/assets/src/admin/paired-browsing/client.js
+++ b/assets/src/admin/paired-browsing/client.js
@@ -1,35 +1,91 @@
 /**
  * WordPress dependencies
  */
-import domReady from '@wordpress/dom-ready';
+// import domReady from '@wordpress/dom-ready';
+
+function domReady( callback ) {
+	if (
+		document.readyState === 'complete' || // DOMContentLoaded + Images/Styles/etc loaded, so we call directly.
+		document.readyState === 'interactive' // DOMContentLoaded fires at this point, so we call directly.
+	) {
+		return callback();
+	} // DOMContentLoaded has not fired yet, delay callback until then.
+
+	document.addEventListener( 'DOMContentLoaded', callback );
+	return null;
+}
 
 const { parent } = window;
 
-if ( parent.pairedBrowsingApp ) {
-	window.ampPairedBrowsingClient = true;
-	const app = parent.pairedBrowsingApp;
-
-	app.registerClientWindow( window );
-
-	domReady( () => {
-		if ( app.documentIsAmp( document ) ) {
-			// Hide the paired browsing menu item.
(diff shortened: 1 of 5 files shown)


assets/src/admin/paired-browsing/client.js
```diff
@@ -1,35 +1,91 @@
...
+	if (
+		document.readyState === 'complete' || // DOMContentLoaded + Images/Styles/etc loaded, so we call directly.
+		document.readyState === 'interactive' // DOMContentLoaded fires at this point, so we call directly.
+	) {
+		return callback();
+	} // DOMContentLoaded has not fired yet, delay callback until then.
+
+	document.addEventListener( 'DOMContentLoaded', callback );
+	return null;
+}
```
westonruter: I did this because I was getting mysterious sporadic failures with the imported `domReady`:

image

westonruter:
It turns out portals cannot be used in this way currently. See https://github.com/WICG/portals/issues/113#issuecomment-560229183.

Issue #3838, opened 2019-11-28
Storing Story content in the DB (AMP HTML)


miina:
## Feature description

The AMP HTML for rendering in the FE should be kept in `post_content` based on the JSON saved to `post_meta` (see #3794)

This would be updated only when JSON is updated. 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* AMP Story HTML is saved to `post_content` based on JSON from editor. 
* The HTML renders in the FE as a valid AMP Story.

## Implementation brief

Every time when post_meta is updated in the DB to save the data from the editor then we should also update AMP HTML in the `post_content`.

For portability reasons, putting together the AMP HTML will be done on the **client-side** within the React app. The content will be saved to the DB as-is and it will leverage the AMP Plugin's sanitization logic for always rendering valid AMP. (To clarify: this effectively means that we could potentially save invalid AMP to the `post_content` even though the rendered output will always be valid AMP. The validation is purposefully not making any changes to the DB directly, different behavior could be discussed within a separate Validation issue)

The following will be considered:
- The `amp-story` tag will be part of the `post_content` to make it easier to port, everything outside of the `amp-story` tag will be part of the WordPress template and not saved to `post_content`.
- Each Page (`amp-story-page`) will have one `amp-story-grid-layer` child element with `vertical` layout which will include all the common elements with absolute positioning, including shapes, images, Text, etc.
- Additionally, `amp-story-cta-layer` and `amp-story-page-attachment` will be handled separately and always added as the last element after `amp-story-grid-layer`. It has to be ensured that there is only one CTA layer or one Attachment on each Page. Also, there can be no CTA layer on the first Page. This should already be done before starting putting together the AMP HTML output.
- Depending on the approach chosen within #3802 for full-bleed components, then those might potentially have their own `amp-story-grid-layer` with template `fill`.

We will need mapping for the attributes from JSON to attributes added to the tags, for example:
`x`  in JSON will be converted to `left` in `style` attribute (likely with some kind of logic for converting from pixels, TBD later)
`animationType` => `animate-in`, etc.

We will then loop through the Pages, create the content and save the result AMP HTML to `post_content` via API request.

[...]

PR #3871, opened 2019-12-02
Save page elements 

### LINKED ISSUE #3794: Storing Story content in the DB (JSON)
## Description

Currently, the idea based on the new PoC is to save the data for the editor in JSON format and separately the content for rendering in the FE, in AMP HTML.

We should look into the different ideas and options to determine which is the preferred and viable way for saving the content and if there are better alternatives for the currently suggested idea.

Note that Gutenberg stores data in the `post_content` field, including both the blocks' configuration and the HTML to render.

Also consider CPT suggestion from #3743 

## Acceptance Criteria

AC: There is a document comparing at least 2 different options for storing content in the DB for editor + the FE.
AC: A suggestion is made based on the findings.

## Implementation brief

*  Modify the REST API controller, so it accepts `story_data` and saves this to the `post_content_filtered` field on the `WP_Post` object.
* `content_filtered` will be registered as an array, so care must be taken to use `wp_json_encode` and `json_decode` where possible.
- Implementation will tracking the following fields in WP
   - Title
   - slug ( post name ). 
   - Status ( Post status ) i.e Published, auto-draft or draft. 
   - Author
- `useLoadStory` will be changed to load these fields into state. 
- Add a `useSavePost`, to change the call to the API hook  `useAPI` and setting the `isSaving` global state.
- An `isSaving` state will be tracked, so a spinner can be displayed to user and save button disabled while saving. 

PoC can be found [here](https://github.com/ampproject/amp-wp/tree/try/save-post).

 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

spacedmonkey:
## Summary

Fixes #3794

## Checklist


includes/class-amp-story-post-type.php
```diff
@@ -105,6 +105,39 @@ public static function register() {
...
+							'type'    => 'integer',
+							'default' => 0,
+						],
+					],
+				],
+			],
+		];
+		// Hide the pages data from none logged in users.
+		if ( ! is_user_logged_in() ) {
+			$meta_args['show_in_rest'] = false;
```
swissspidy: Couldn't this always be `true` but just use an `auth_callback`? ([docs](https://developer.wordpress.org/reference/functions/register_meta/#parameters))

spacedmonkey: `auth_callback` is just for writes and not reads. See the docs 
> (string) Optional. A function or method to call when performing edit_post_meta, add_post_meta, and delete_post_meta capability checks.`

includes/class-amp-story-post-type.php
```diff
@@ -105,6 +105,39 @@ public static function register() {
...
+		}
+		register_meta( 'post', 'amp_pages', $meta_args );
```
miina: I wonder if we should register a more generic `post_meta` instead of `amp_pages` (e.g. `story_content`) and make `amp_pages` a part of it?

Reason: there might be additional information that might not necessarily be related to a specific page (general settings), such as `Advance to next Page` setting in the Legacy Stories. There might potentially be more of those in the future. Or do you think it would make more sense to register a separate meta for each new setting? Or register a separate meta for all generic settings? Thoughts?

barklund: I agree with this. For future compatibility we shouldn't (through naming) restrict what to save in the field. `story_content` (or `amp_story_content`) is a candidate.

assets/src/edit-story/app/api/apiProvider.js
```diff
@@ -23,6 +23,25 @@ function APIProvider( { children } ) {
 		[ stories ],
 	);
 
+	const saveStoryById = useCallback(
+		( storyId, title, status, pages, author, slug ) => {
```
dvoytenko: To clarify, is the `pages` here an array of separated story pages? And all the story-wide properties would go as separate props here? Is there a reason not to put them all together into a single `story` JSON structure?

assets/src/edit-story/app/story/storyProvider.js
```diff
@@ -26,12 +26,19 @@ import useSelectElementById from './actions/useSelectElementById';
...
 import useSetCurrentPageByIndex from './actions/useSetCurrentPageByIndex';
 import useSetPropertiesOnSelectedElements from './actions/useSetPropertiesOnSelectedElements';
+import useSavePost from './actions/useSavePost';
 
 function StoryProvider( { storyId, children } ) {
 	// Story state is stored in these three immutable variables only!
 	// Don't update 1 of these in an effect based off another base variable.
 	// Only update these directly as a response to user or api interactions.
 	const [ pages, setPages ] = useState( [] );
+	const [ title, setTitle ] = useState( '' );
```
dvoytenko: Would it be better to just treat these all as story's props (just like page's props, etc) and reduce number of state read/writers? Then it'd all be part of the data model, including history undo/redo. Obviously, this would not include things like `setIsSaving`, etc.

includes/class-amp-story-post-type.php
spacedmonkey: I think we should still prefix our key with something, just in case someone else tries to use `story_content`. I can work with `amp_story_content`. 

This meta key is only for the pages JSON object. Other meta data like deep linking etc, would be stored in another object / meta key.

assets/src/edit-story/app/api/apiProvider.js
spacedmonkey: `Pages` here is a array of object that represents the pages in the story. The other fields are just WordPress fields.

assets/src/edit-story/app/story/storyProvider.js
spacedmonkey: I separated these out, as means you can be more selective about what you write to. Different component will write to state. If all the properties are one object, it may mean a redraw of components that do not need it. @barklund Thoughts ^

assets/src/edit-story/app/api/apiProvider.js
dvoytenko: I'd think that we should probably make `pages` a field in a separate `story` object. There will definitely be some story wide properties.

assets/src/edit-story/app/story/storyProvider.js
dvoytenko: Good for now. But IMHO it's a good time to huddle again w.r.t. data model and how it's updated. In some cases the state is convenient, but I think a simpler reducer could have some benefits for us. Can we have that discussion with @barklund ?

assets/src/edit-story/app/api/apiProvider.js
swissspidy: Agreed. A `story` object makes sense to me 👍

includes/rest-api/class-wp-rest-stories-controller.php
```diff
@@ -0,0 +1,131 @@
...
+				$response->add_link( $rel, $link['href'], $link['attributes'] );
+			}
+		}
+
+		/**
+		 * Filters the post data for a response.
+		 *
+		 * The dynamic portion of the hook name, `$this->post_type`, refers to the post type slug.
+		 *
+		 *
```
westonruter: Remove line.

assets/src/edit-story/app/api/apiProvider.js
spacedmonkey: A story is a group of separate fields. I am still not sure what the benefit of a story object here.

includes/rest-api/class-wp-rest-stories-controller.php
```diff
@@ -0,0 +1,131 @@
...
+		$schema['properties']['content_filtered'] = array(
+			'description' => __( 'The post content filtered for the object.', 'amp' ),
+			'type'        => 'object',
+			'context'     => array( 'view', 'edit', 'embed' ),
+			'arg_options' => array(
+				'sanitize_callback' => null, // Note: sanitization implemented in self::prepare_item_for_database()
+				'validate_callback' => null, // Note: validation implemented in self::prepare_item_for_database()
+			),
+			'properties'  => array(
+				'raw' => array(
```
westonruter: Do we need `raw`? Since there is only one property, couldn't this level be removed?

includes/rest-api/class-wp-rest-stories-controller.php
```diff
@@ -0,0 +1,131 @@
...
+	 *
+	 * @return array Item schema as an array.
+	 */
+	public function get_item_schema() {
+		if ( $this->schema ) {
+			return $this->add_additional_fields_schema( $this->schema );
+		}
+		$schema = parent::get_item_schema();
+
+		$schema['properties']['content_filtered'] = array(
```
westonruter: As opposed to `content_filtered` what about something that avoids exposing some of the internals, that we're using `post_content_filtered`? In other words, perhaps this should rather be something like `story_data`.

It would be kinda nice if the field were added alongside `content.raw` and `content.rendered`, but I understand there is a limitation in the REST API which doesn't allow for this. See https://core.trac.wordpress.org/ticket/47779

includes/rest-api/class-wp-rest-stories-controller.php
```diff
@@ -0,0 +1,131 @@
...
+	 * @return stdClass|WP_Error Post object or WP_Error.
+	 */
+	protected function prepare_item_for_database( $request ) {
+		$prepared_story = parent::prepare_item_for_database( $request );
+
+		if ( is_wp_error( $prepared_story ) ) {
+			return $prepared_story;
+		}
+
+		if ( isset( $request['content_filtered'] ) ) {
```
westonruter: Shouldn't this field be marked as `required`? It shouldn't be allowed to update the `content` without also updating the underlying corresponding story data. I don't think this `if` condition should be here then.

includes/rest-api/class-wp-rest-stories-controller.php
```diff
@@ -0,0 +1,131 @@
...
+			$prepared_story->post_content_filtered = $request['content_filtered'];
+			if ( isset( $request['content_filtered']['raw'] ) ) {
+				$prepared_story->post_content_filtered = $request['content_filtered']['raw'];
+			}
+			$prepared_story->post_content_filtered = maybe_serialize( $prepared_story->post_content_filtered );
```
westonruter: Is `maybe_serialize` right? Shouldn't it be `wp_json_encode`?

includes/rest-api/class-wp-rest-stories-controller.php
```diff
@@ -0,0 +1,131 @@
...
+		}
+		$schema = parent::get_item_schema();
+
+		$schema['properties']['content_filtered'] = array(
+			'description' => __( 'The post content filtered for the object.', 'amp' ),
+			'type'        => 'object',
+			'context'     => array( 'view', 'edit', 'embed' ),
+			'arg_options' => array(
+				'sanitize_callback' => null, // Note: sanitization implemented in self::prepare_item_for_database()
+				'validate_callback' => null, // Note: validation implemented in self::prepare_item_for_database()
```
westonruter: Sanitization and validation are currently not implemented in that method.

Couldn't `rest_sanitize_request_arg` and  `rest_validate_request_arg` be used at least? (The defaults as supplied by `\WP_REST_Request::sanitize_params()`.)

In other words, if these lines were removed would that not allow basic validation to be performed?

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3871/commits/00846a7433a5909b8efe6d671f302965efa4484f

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: I think that is confusing. If we wanted it named `story_data` we should have gone with register meta. The field should reflect what it maps to in core.

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: To save an array into a string field in the database the array has converted to a string for saving. So this array has be `serialized`.

includes/rest-api/class-wp-rest-stories-controller.php
westonruter: Why? The REST API intentionally diverged from what core uses internally in `wp_insert_post()` because the legacy naming was bad. For example, the REST API uses `content` instead of `post_content` and `slug` instead of `post_name`. It seems totally fine to use a more specific field name than `content_filtered`. The REST API client shouldn't care at all where the data is being stored.

includes/rest-api/class-wp-rest-stories-controller.php
westonruter: `wp_json_encode()` also converts the array into a string. The data is known to be JSON-serializable since it came via the REST API. 

In general, PHP `serialize()` should be avoided when possible since there are security impacts, hence why they are flagged in PHPCS: https://github.com/WordPress/WordPress-Coding-Standards/blob/fab1ed525e057c19d723bf7c93feaf0a48512abe/WordPress/Sniffs/PHP/DiscouragedPHPFunctionsSniff.php#L39-L47

Granted, this is likely not exploitable here. But still, JSON seems preferable to serialized PHP. JSON is also not susceptible to corruption when doing search/replace when changing string lengths.

Also, I understand that `json_decode()` is faster than `unserialize()` in PHP.

barklund:
Regarding the data management, remember that we'll change those things once we merge in #3910.

I can prepare the data storage for these attributes though.

My recommendation would be:

* Add a new state variable `meta` to the story reducer state.
* Add a new API to update `meta` variables.
* Also store the `meta` variable array in history.
* When loading a story, initialising `meta` from a set of fixed properties on the loaded data object.
* When saving a story, copying a fixed list of properties from the meta object to the request object.

Thoughts?

assets/src/edit-story/app/story/storyProvider.js
barklund: See https://github.com/ampproject/amp-wp/pull/3871#issuecomment-565461091.

assets/src/edit-story/app/api/apiProvider.js
barklund: See https://github.com/ampproject/amp-wp/pull/3871#issuecomment-565461091

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: Works for me. https://github.com/ampproject/amp-wp/pull/3871/commits/b43d2df908fe7afeccd5562cc1bee045dda9a949

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3871/commits/b09e7cc93b31e102212976ce451252731d46d395

includes/rest-api/class-amp-rest-stories-controller.php
```diff
@@ -0,0 +1,134 @@
...
+	 * @return WP_REST_Response Response object.
+	 */
+	public function prepare_item_for_response( $post, $request ) {
+		$response = parent::prepare_item_for_response( $post, $request );
+		$fields   = $this->get_fields_for_response( $request );
+		$data     = $response->get_data();
+		$schema   = $this->get_item_schema();
+
+		if ( in_array( 'content_filtered', $fields, true ) ) {
+			$post_content_filtered    = json_decode( $post->post_content_filtered );
```
westonruter: This needs to ensure an array is returned rather than an object, right?

```suggestion
			$post_content_filtered    = json_decode( $post->post_content_filtered, true );
```

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: This keeps the endpoint inline with other core endpoints. Both `title` and `content` both follow this pattern.

includes/class-amp-story-post-type.php
spacedmonkey: Now we are using `post_content_filtered`, resolving.

includes/rest-api/class-wp-rest-stories-controller.php
westonruter: Only because they get `rendered` output that needs to be sent. There will never be a `rendered` field for this `content_filtered` because it is essentially the `content.rendered`.

I do not think it make sense to include `raw` for this field.

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: Why does it need to be required?

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: It is confusing for new developer, as they would have to look through the code to see what field that maps too. I don't see the benefit of hiding the field name here.

includes/rest-api/class-wp-rest-stories-controller.php
westonruter: Well, it should at least be `required` if the `content` is also supplied. We would want to force the client to ensure when updating a story that they provide both the rendered AMP story HTML as well as the underlying data that was used to render it. This is my understanding of how the editor will work with the data.

The same should be true vice-versa: if the story data is supplied then the `content` must should also be supplied.

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: It is the pattern the REST API uses, why diverge from that?

includes/rest-api/class-wp-rest-stories-controller.php
westonruter: It's not hiding the field. It's using a more semantic field name for the REST API client. If a developer wants to find how the data is stored in PHP, they'd have to look at the controller anyway. When doing so, they'd see how it maps.

includes/rest-api/class-wp-rest-stories-controller.php
westonruter: The REST API only uses `raw`/`rendered` for fields that have PHP filters applying on the underlying value. There is no `raw`/`rendered` for fields that don't have PHP filters applying (more or less), e.g. `status`, `date`, `link`, `featured_media`. 

Because there is no filters applying on this field, then no `raw` should be included.

assets/src/edit-story/app/api/apiProvider.js
spacedmonkey: Marking as resolved, as all questions of state are well out of scope of this PR. Let's move this conversion to #3910

assets/src/edit-story/app/story/storyProvider.js
spacedmonkey: See https://github.com/ampproject/amp-wp/pull/3871#discussion_r357696957

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: I mean if we are going to start renaming fields, should we rename all the fields in the REST API to what they are in our new editor? Either we go with WP naming or we don't here. Some fields is really confusing.

swissspidy: What does this have to do with _naming_? 🤔 I thought this is just about making sure required data is supplied.

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: Field name changed - https://github.com/ampproject/amp-wp/pull/3871/commits/25c14233ebe86581b6ed385871fe85ef7784f669

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: @swissspidy This comment was designed to go with https://github.com/ampproject/amp-wp/pull/3871#discussion_r357643691 . Something must have gone wrong and it was posted here.

includes/rest-api/class-wp-rest-stories-controller.php
spacedmonkey: I was trying to give us an option for us to have a rendered version in the future. But we can always change it later if required. 

Changed in https://github.com/ampproject/amp-wp/pull/3871/commits/25c14233ebe86581b6ed385871fe85ef7784f669

assets/src/edit-story/app/story/actions/useSavePost.js
```diff
@@ -0,0 +1,73 @@
...
+	const savePost = useCallback( () => {
+		if ( ! isSaving ) {
+			setIsSaving( true );
+			saveStoryById( storyId, title, status, pages, postAuthor, slug ).then( ( post ) => {
+				const { status: thisPostStatus, link } = post;
+				setPostStatus( thisPostStatus );
+				setLink( link );
+				refreshPostEditURL( storyId );
+			} ).catch( () => {
+				// TODO Display error message to user as save as failed.
```
swissspidy: Sounds like something we first need UX for. Don't see anything in Figma about displaying such notifications.

spacedmonkey: It makes sense to show an error here. A lot of that work has been done yet for the helpers. If no error message is to be shown, I will remove this TODO later.

swissspidy: What I meant is UX for _how_ it should be displayed. It makes sense to have error messages for sure.

includes/rest-api/class-wp-rest-stories-controller.php
miina: @westonruter I'll be addressing it being required if `content` is supplied within #3838. Will mark this resolved for the purposes of this PR.

PR #1753, opened 2018-12-19
Provide a placeholder element for Twitter embeds

westonruter:
Refer to prior work on SoundCloud embeds (https://github.com/ampproject/amp-wp/pull/2722) and YouTube embeds (#3358).

Todo:

- [x] Use Twitter oEmbed response as placeholder content for `amp-twitter`.
- [x] Update tests to mock HTTP requests to prevent external requests during testing.
- [x] Fix tests.
- [x] ~Use YouTube poster image linked to YouTube URL as the `fallback` content for `amp-youtube`.~ See https://github.com/ampproject/amp-wp/pull/1753#issuecomment-511618954. Done in #3358.

# Twitter

Before:

```html
<amp-twitter data-tweetid="1065026506068504577" layout="responsive" width="600" height="480"></amp-twitter>
```

After:

```html
<amp-twitter width="600" height="480" layout="responsive" data-tweetid="1065026506068504577">
    <blockquote class="twitter-tweet" data-width="525" data-dnt="true" placeholder>
        <p lang="en" dir="ltr">🦖 HTM: JSX without the transpiler.<br><br>700 bytes to run in the browser,<br>or
            compile it away using babel-plugin-htm.<br><br>(works with any VDOM library!)<a href="https://t.co/KrlyiYP9Aj">https://t.co/KrlyiYP9Aj</a>
            <a href="https://t.co/qbEFfMT1Dy">pic.twitter.com/qbEFfMT1Dy</a></p>— Jason Miller 🦊⚛ (@_developit) <a
            href="https://twitter.com/_developit/status/1065026506068504577?ref_src=twsrc%5Etfw">November 20, 2018</a>
    </blockquote>
</amp-twitter>
```

# YouTube (Done in #3358)

Before:

```html
<figure class="wp-block-embed-youtube wp-block-embed is-type-video is-provider-youtube  wp-has-aspect-ratio">
    <div class="wp-block-embed__wrapper">
        <amp-youtube data-videoid="8EtmqU844hU" layout="responsive" width="640" height="360"></amp-youtube>
    </div>
</figure>
```

After:

 ```html
<figure class="wp-block-embed-youtube wp-block-embed is-type-video is-provider-youtube  wp-has-aspect-ratio">
    <div class="wp-block-embed__wrapper">
        <amp-youtube data-videoid="8EtmqU844hU" layout="responsive" width="640" height="360">
            <a fallback href="https://www.youtube.com/watch?v=8EtmqU844hU">Finding Portland</a>
[...]

PR #3874, opened 2019-12-03
Prevent `wpautop()` from modifying Twitter embed

### LINKED ISSUE #3865: Twitter embed in Classic Block is duplicated
## Bug Description

Given `post_content` (from theme unit test data) that consists of:

```
https://twitter.com/nacin/status/319508408669708289

This post tests WordPress' <a title="Twitter Embeds" href="https://en.support.wordpress.com/twitter/twitter-embeds/" target="_blank" rel="noopener noreferrer">Twitter Embeds</a> feature.
```

This is incorrectly resulting in the Tweet being duplicated:

image

The generated HTML even includes the duplication:

```html
<p><amp-twitter width="600" height="480" layout="responsive" data-tweetid="319508408669708289"></amp-twitter></p>
<amp-twitter width="600" height="480" layout="responsive" data-tweetid="319508408669708289" class="twitter-tweet" data-width="550" data-dnt="true" placeholder=""><blockquote class="twitter-tweet" data-width="550" data-dnt="true" placeholder="">
<p lang="en" dir="ltr">Really cool to read through and find so much awesomeness added to WordPress 3.6 while I was gone. I should take three weeks off more often.</p>
<p>— Andrew Nacin (@nacin) <a href="https://twitter.com/nacin/status/319508408669708289?ref_src=twsrc%5Etfw">April 3, 2013</a></p></blockquote></amp-twitter>
<p></p>
<p>This post tests WordPress’ <a title="Twitter Embeds" href="https://en.support.wordpress.com/twitter/twitter-embeds/" target="_blank" rel="noopener noreferrer">Twitter Embeds</a> feature.</p>
```

## Expected Behaviour

The non-AMP version does not have a duplicated embed, and this is what it should look like in AMP:

image

## Steps to reproduce

1. Create a post and enter code editor mode, pasting the above `post_content` into it.
2. Publish and view the AMP version/
3. See duplicated tweet.

## Additional context

The issue doesn't seem to occur in 1.4.1, so it seems to be some regression introduced in `develop` since the 1.4 branch.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

The cause of the duplication was due to `wpautop()` mucking with the content, wrapping tags it deems as "block-level" with a `<p>` which causes the `blockquote` tag to be placed outside the AMP component. Once outside, the `blockquote` is santitized as a raw embed and another Twitter AMP component would be rendered.

Removing the HTML oEmbed filter in `AMP_Twitter_Embed_Handler` will allow for `wpautop` to process the original HTML, and allows the raw embed sanitizer within the class to convert it into an AMP component, without any interference.

Fixes #3865.

## Checklist

### DIFF
--- a/tests/php/test-amp-twitter-embed.php
+++ b/tests/php/test-amp-twitter-embed.php
@@ -91,34 +91,28 @@ class AMP_Twitter_Embed_Test extends WP_UnitTestCase {
 			],
 			'url_simple'                           => [
 				'https://twitter.com/wordpress/status/987437752164737025' . PHP_EOL,
-				'<p><amp-twitter width="600" height="480" layout="responsive" data-tweetid="987437752164737025">',
-				"</amp-twitter></p>\n",
+				"<amp-twitter width=\"600\" height=\"480\" layout=\"responsive\" data-tweetid=\"987437752164737025\" class=\"twitter-tweet\" data-width=\"500\" data-dnt=\"true\"><blockquote class=\"twitter-tweet\" data-width=\"500\" data-dnt=\"true\" placeholder=\"\">\n<p lang=\"en\" dir=\"ltr\">Celebrate the WordPress 15th Anniversary on May 27 <a href=\"https://t.co/jv62WkI9lr\">https://t.co/jv62WkI9lr</a> <a href=\"https://t.co/4ZECodSK78\">pic.twitter.com/4ZECodSK78</a></p>\n<p>— WordPress (@WordPress) <a href=\"https://twitter.com/WordPress/status/987437752164737025?ref_src=twsrc%5Etfw\">April 20, 2018</a></p></blockquote></amp-twitter>\n\n",
 			],
 			'url_with_big_tweet_id'                => [
 				'https://twitter.com/wordpress/status/705219971425574912' . PHP_EOL,
-				'<p><amp-twitter width="600" height="480" layout="responsive" data-tweetid="705219971425574912">',
-				"</amp-twitter></p>\n",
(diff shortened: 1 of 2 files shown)


westonruter:
Oh, I see. This now brings to mind an issue we had with placeholder content in `amp-iframe`, namely that we had to use a `span` to prevent this from happening.

This being the case, I am wary of modifying the behavior of `wpautop` as this seems like it could cause headaches later. Instead, I suggest populating the Twitter placeholder content with a... placeholder. For example, the placeholder could be some unique string of characters which then get replaced with the actual `blockquote` after `wpautop` has run.

pierlon:
Just wanted to point out that setting `wpautop` to run last is how [dynamic blocks are rendered](https://github.com/WordPress/wordpress-develop/blob/46581088d5c3609f8ca9c13aedb501ecf68b8106/src/wp-includes/blocks.php#L317-L321). I agree with using a placeholder for the placeholder (:laughing:) though, it would seem to be more future-proof.

pierlon:
Rebased to remove the commit messing with `wpautop`.

westonruter:
@pierlon I thought this PR was going to be needed for 1.4.2, hence the squash merge. But then I found it wouldn't cherry-pick cleanly onto the 1.4 branch. This is due to #1753, which is only in `develop` (which this PR reverts part of). So I believe this isn't needed in a patch release prior to 1.5.

pierlon:
Yes that makes sense.

tests/php/test-amp-twitter-embed.php
```diff
@@ -162,18 +152,19 @@ public function get_conversion_data() {
...
+		$dom              = AMP_DOM_Utils::get_dom_from_content( $filtered_content );
+		$embed->sanitize_raw_embeds( $dom );
 
-		$this->assertStringStartsWith( $expected_start, $filtered_content );
-		if ( $expected_end ) {
-			$this->assertStringEndsWith( $expected_end, $filtered_content );
-		}
+		$content = AMP_DOM_Utils::get_content_from_dom( $dom );
+
+		$this->assertEquals( $content, $expected );
```
schlessera: @pierlon For PHPUnit assertion, the expected value always comes first, followed by the actual value.

This is important because PHPUnit will tell you what it expected, and what it got instead. If you put the values in the wrong order (like here), this is misleading and causes you to look in the wrong places.

I'm rectifying this in #3758, as I'm touching those tests anyway.

pierlon: Ah thanks for pointing that out, noted :+1: .

Issue #3788, opened 2019-11-18
Remove `the_content` Filter


jauyong:
## Feature description

This issue was inspire by this [slack comment. ](https://xwp.slack.com/archives/CCCAUAH6F/p1573920050484300)

> I think we should consider removing all the_content filters when rendering the story post type other than the ones we know we are using. This will prevent things like this from happening: https://wordpress.org/support/topic/amp-stories-blank-page/

This came up before as well for someone who had an estimated reading time plugin that prepended a paragraph to the content.
https://github.com/ampproject/amp-wp/issues/3321

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

PR #3876, opened 2019-12-03
Saving AMP HTML to post_content

miina:
## Summary

See #3838

## Checklist


includes/class-amp-story-legacy-post-type.php
```diff
@@ -44,9 +44,10 @@ class AMP_Story_Legacy_Post_Type {
 	/**
 	 * The rewrite slug for this post type.
 	 *
+	 * @todo Figure out a suitable rewrite slug for this.
 	 * @var string
 	 */
-	const REWRITE_SLUG = 'stories';
+	const REWRITE_SLUG = 'amp-stories';
```
miina: Changed this since both the new and the old Story CPT-s were using the same at this moment.

spacedmonkey: Good spot :+1:

includes/templates/single-amp-story.php
```diff
@@ -0,0 +1,57 @@
...
+} elseif ( isset( $metadata['publisher']['logo'] ) && is_string( $metadata['publisher']['logo'] ) ) {
+	$publisher_logo_src = $metadata['publisher']['logo'];
+} else {
+	$publisher_logo_src = admin_url( 'images/wordpress-logo.png' );
+}
+$publisher = isset( $metadata['publisher']['name'] ) ? $metadata['publisher']['name'] : get_option( 'blogname' );
+
+$meta_images = AMP_Story_Media::get_story_meta_images();
+?>
+<amp-story
```
miina: `amp-story` would ideally also be added to the `post_content` for portability, we could do that on the PHP side initially for simplicity, e.g. via filtering while saving.

includes/class-amp-story-post-type.php
```diff
@@ -142,6 +142,98 @@ public static function register() {
...
+					$sanitizers['AMP_Img_Sanitizer']['add_noscript_fallback']    = false;
+					$sanitizers['AMP_Audio_Sanitizer']['add_noscript_fallback']  = false;
+					$sanitizers['AMP_Video_Sanitizer']['add_noscript_fallback']  = false;
+					$sanitizers['AMP_Iframe_Sanitizer']['add_noscript_fallback'] = false; // Note that iframe is not yet allowed in an AMP Story.
+				}
+				return $sanitizers;
+			}
+		);
+
+		add_filter(
```
miina: Note: this should be handled by the validator, IIRC, needs looking into why the `p` is added/not removed, perhaps there's something missing still.

assets/src/edit-story/app/story/actions/useSavePost.js
```diff
@@ -1,14 +1,44 @@
...
+		return renderToString(
+			<amp-story-page id={ id }>
+				<amp-story-grid-layer template="vertical">
+					{ page.elements.map( ( { type, src, width, height, content } ) => {
+						const style = {
+							position: 'absolute',
+						};
+						// @todo this should be redone by using dynamic tag.
+						if ( 'image' === type ) {
+							return <amp-img key={ 'el' + id } style={ { ...style } } src={ src } layout="fixed" width={ width } height={ height } />;
```
spacedmonkey: These different elements should be different component files and use a switch statement.

assets/src/edit-story/app/story/actions/useSavePost.js
```diff
@@ -36,7 +66,8 @@ function useSavePost( {
 	const savePost = useCallback( () => {
 		if ( ! isSaving ) {
 			setIsSaving( true );
-			saveStoryById( storyId, title, status, pages, postAuthor, slug ).then( ( post ) => {
+			const content = getStoryMarkupFromPages( pages );
```
spacedmonkey: Love how simple this is! :+1:

spacedmonkey: I have a spinner for saving, so this takes some time then the spinner will show. Should `getStoryMarkupFromPages` return a promise?

includes/validation/class-amp-validation-manager.php
```diff
@@ -262,7 +262,7 @@ public static function post_supports_validation( $post ) {
 		}
 
 		// Story post type always supports validation.
-		if ( AMP_Story_Legacy_Post_Type::POST_TYPE_SLUG === $post->post_type ) {
+		if ( AMP_Story_Legacy_Post_Type::POST_TYPE_SLUG === $post->post_type || AMP_Story_Post_Type::POST_TYPE_SLUG === $post->post_type ) {
```
spacedmonkey: I had this in my original PR for post type.

spacedmonkey:
```suggestion
		if ( in_array( [ AMP_Story_Legacy_Post_Type::POST_TYPE_SLUG, AMP_Story_Post_Type::POST_TYPE_SLUG ], $post->post_type, true ) ) {
```

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -615,7 +615,7 @@ public static function get_url_from_post( $post ) {
 		$url = $post->post_title;
 
 		$queried_object = get_post_meta( $post->ID, '_amp_queried_object', true );
-		$is_amp_story   = isset( $queried_object['id'], $queried_object['type'] ) && 'post' === $queried_object['type'] && AMP_Story_Legacy_Post_Type::POST_TYPE_SLUG === get_post_type( $queried_object['id'] );
+		$is_amp_story   = isset( $queried_object['id'], $queried_object['type'] ) && 'post' === $queried_object['type'] && ( AMP_Story_Legacy_Post_Type::POST_TYPE_SLUG === get_post_type( $queried_object['id'] || AMP_Story_Post_Type::POST_TYPE_SLUG === get_post_type( $queried_object['id'] ) ) );
```
spacedmonkey:
```suggestion
		$is_amp_story   = isset( $queried_object['id'], $queried_object['type'] ) && 'post' === $queried_object['type'] && ( in_array( [ AMP_Story_Legacy_Post_Type::POST_TYPE_SLUG, AMP_Story_Post_Type::POST_TYPE_SLUG ], get_post_type( $queried_object['id'] ), true ) );
```

spacedmonkey: This code is cleaner and fixes an issue with `get_post_type( $queried_object['id']` that doesn't have a closing bracket.

includes/class-amp-story-post-type.php
```diff
@@ -305,4 +397,54 @@ public static function admin_body_class( $class ) {
...
 		return $class;
 	}
+
+	/**
+	 * Filter the allowed tags for KSES to allow for amp-story children.
+	 *
+	 * @param array $allowed_tags Allowed tags.
+	 * @return array Allowed tags.
+	 */
+	public static function filter_kses_allowed_html( $allowed_tags ) {
```
spacedmonkey: Like looks like repeated code. Instead of repeating code, we should use a [Abstraction Class](https://www.php.net/manual/en/language.oop5.abstract.php) or [traits](https://www.php.net/manual/en/language.oop5.traits.php) to stop repeated methods.

assets/src/edit-story/app/story/actions/useSavePost.js
```diff
@@ -1,14 +1,44 @@
...
 
+const getStoryMarkupFromPages = ( pages ) => {
+	// @todo look into: We should rather use renderToStaticMarkup here, however, it doesn't seem to be exposed via @wordpress/element.
+	const markup = pages.map( ( page ) => {
+		const { id } = page;
+		return renderToString(
+			<amp-story-page id={ id }>
+				<amp-story-grid-layer template="vertical">
+					{ page.elements.map( ( { type, src, width, height, content } ) => {
+						const style = {
```
dvoytenko: Does this updates the element's style or also uses some form of JS->CSS stylesheet?

miina: That's saved as inline-style currently, we could also use a stylesheet for the FE, however, it might be more portable with inline-styles. We might need a stylesheet probably anyway.

(Note that this PR is currently PoC and waiting for the Implementation Brief to get Approved before moving to In Progress.)

dvoytenko: Thanks! I actually wanted to see if we can get away with using inline styles for most of things here.

miina:
@barklund @dvoytenko Would you mind giving the PR another round of review on the React/JS side? Should be quite final now for the initial implementation.

Will address the PHP side tomorrow (cc @spacedmonkey).

assets/src/edit-story/app/story/actions/useSavePost.js
```diff
@@ -1,13 +1,34 @@
...
+			<amp-story-page id={ id }>
+				<amp-story-grid-layer template="vertical">
+					{ page.elements.map( ( { type, ...rest } ) => {
+						const { id: elId } = rest;
+						// eslint-disable-next-line @wordpress/no-unused-vars-before-return
+						const { Save } = getDefinitionForType( type );
+						return <Save key={ 'element-' + elId } { ...rest } />;
+					} ) }
+				</amp-story-grid-layer>
+			</amp-story-page>,
```
dvoytenko: Is this comma doing anything?

assets/src/edit-story/elements/image/save.js
```diff
@@ -0,0 +1,48 @@
...
+	const wrapperProps = {
+		id: 'el-' + id,
+	};
+	const style = {
+		position: 'absolute',
+		left: getPercentageFromPixels( x, 'x' ) + '%',
+		top: getPercentageFromPixels( y, 'y' ) + '%',
+		width: getPercentageFromPixels( width, 'x' ) + '%',
+		height: getPercentageFromPixels( height, 'y' ) + '%',
+		transform: `rotate(${ rotationAngle }deg)`,
```
dvoytenko: Not the focus of this PR, but just quick nuance: `transform: rotate(0deg)` has a lot of performance and other nuances. Generally it'd be safer to not set `transform` at all if no transformation is needed.

assets/src/edit-story/elements/square/save.js
```diff
@@ -0,0 +1,39 @@
...
+ * Internal dependencies
+ */
+import getPercentageFromPixels from '../../utils/getPercentageFromPixels';
+
+/**
+ * Returns AMP HTML for saving into post content for displaying in the FE.
+ */
+function SquareSave( { id, backgroundColor, width, height, x, y, rotationAngle } ) {
+	const style = {
+		position: 'absolute',
```
dvoytenko: Same, not the focus here, but still this is similar to our convo on wrappers and the repeated code that supports the element placement/sizing/rotation basics.

assets/src/edit-story/app/story/actions/useSavePost.js
miina: It's the mandatory trailing comma.

assets/src/edit-story/elements/square/save.js
miina: Improved in 5469f09.

assets/src/edit-story/elements/image/save.js
miina: Improved in 5469f09.

assets/src/edit-story/app/story/actions/useSavePost.js
miina: Currently it's just rendering so shouldn't take any time.

includes/class-amp-story-post-type.php
miina: Actually both the story components and allowed tags are different. It's not final yet either and there's a possibility that we can remove this from the legacy post if editing (thus saving) will be disabled anyway. Will keep it like this for now until things are more clear.

includes/validation/class-amp-validated-url-post-type.php
miina: Updated within 178f07b

includes/validation/class-amp-validation-manager.php
miina: Updated within 178f07b

includes/class-amp-story-post-type.php
miina: @westonruter For some reason, I remembered as if the Story sanitizer had removing the `p` tag added by `wpautop` as part of it, however, that's not the case. Any reason why it would get removed in the Legacy Story Post (currently in the released plugin) and not in the new Story Post?

Note: added `remove_filter( 'the_content', 'wpautop' );` for now but since it's not necessary for the Legacy Story Post then I suppose there's something missing from the new.

Thoughts?

westonruter: @miina The removal of `wpautop` may be due to the issue where it incorrectly tries to add `p` tags around AMP components. We've had this issue elsewhere when trying to use `<div>` or some other block-level element as a `placeholder` inside of `<amp-iframe>`, which causes it to break. I think it is similar thing here.

In any case, I don't think _any_ of `the_content` filters should apply in the new Story editor. They will just introduce areas for validation errors to occur (see #3788). If any filters are provided, they should be explicitly added one-by-one.

assets/src/edit-story/app/story/actions/useSavePost.js
```diff
@@ -1,13 +1,40 @@
...
+import { getDefinitionForType } from '../../../elements';
+
+/**
+ * Creates AMP HTML markup for saving to DB for rendering in the FE.
+ *
+ * @param {Object} pages Object of pages.
+ * @return {Element} Markup of pages.
+ */
+const getStoryMarkupFromPages = ( pages ) => {
+	// @todo look into: We should rather use renderToStaticMarkup here, however, it doesn't seem to be exposed via @wordpress/element.
```
swissspidy: `wp.element.renderToString` seems to be a custom implementation, and not just an alias for `ReactDOMServer.renderToString`:

https://github.com/WordPress/gutenberg/blob/67cb1cc7d85e5134c580fba0ed4b362a5d95e07f/packages/element/src/serialize.js#L348-L419

miina: Oh! Good catch, thanks for checking that. Perhaps that's what we need then, will check.

includes/class-amp-story-post-type.php
miina: > In any case, I don't think any of the_content filters should apply in the new Story editor. They will just introduce areas for validation errors to occur (see #3788). If any filters are provided, they should be explicitly added one-by-one.

Hmm, okay, I guess it's OK then to leave the filter here for the purposes of this PR. Without this filter, the `p` is added and the story displayed blank in the FE due to sanitization.

Generally, the `the_content` filter will probably be addressed/discussed here: #3788

assets/src/edit-story/app/story/actions/useSavePost.js
miina: Looks like that works!

miina:
@dvoytenko The issues from the review should be addressed now.

The blocker PR (#3871) got merged now so this PR is ready for merging as well, as soon as it gets approved.

assets/src/edit-story/elements/image/save.js
dvoytenko: Just a tiny comment on this w.r.t. `null` vs `''` style values. Not 100% this is still a case today, but it used to be pretty riddled with browser issues.

miina: Hmm, I was certain that React would omit that style then, however, will check with @barklund

dvoytenko: I think you're right. At least in this case, since we are writing only once, I don't think it actually matters. Let's proceed.

miina: Sounds good!

includes/class-amp-story-post-type.php
```diff
@@ -110,6 +109,98 @@ public static function register() {
...
+		add_action( 'amp_story_head', 'rel_canonical' );
+		add_action( 'amp_story_head', 'wp_shortlink_wp_head', 10, 0 );
+		add_action( 'amp_story_head', 'wp_site_icon', 99 );
+		add_action( 'amp_story_head', 'wp_oembed_add_discovery_links' );
+
+		// Limit the styles that are printed in a story.
+		add_filter( 'print_styles_array', [ __CLASS__, 'filter_frontend_print_styles_array' ] );
+	}
+
+	/**
```
spacedmonkey: What does this function do right now? If is doesn't do anything, maybe it should be removed?

includes/class-amp-story-post-type.php
spacedmonkey: What is the benefit of a anonymous function here? Does this make this harder to test?

spacedmonkey:
I think we should rethink the use of anonymous functions here. The only benefit of them seems to be that they can't be removed. But it makes the code messy and hard to test. 

There also seem to be some PHP warnings. 
```
FILE: includes/validation/class-amp-validation-manager.php
-------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
-------------------------------------------------------------------------------------------------------------------------------------------
 265 | WARNING | Not using strict comparison for in_array; supply true for third argument.
     |         | (WordPress.PHP.StrictInArray.MissingTrueStrict)
-------------------------------------------------------------------------------------------------------------------------------------------
FILE: includes/class-amp-story-post-type.php
--------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
--------------------------------------------------------------------------------------------------------------------
 199 | WARNING | Unused function parameter $handle.
     |         | (VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable)
--------------------------------------------------------------------------------------------------------------------
```

miina:
On anonymous functions:
Note that this is not a finalized feature, the goal of this PoC PR was to create base functionality for creating AMP HTML from JSON, the focus being on the JS side. As discussed over the sync last week, currently the idea is to get basic features working and then build on that.

The filters copied over from the legacy post mostly unchanged were with the purpose of enabling displaying and saving the Stories, we definitely need to revise all of it in future iterations once we start looking into everything in more detail.

assets/src/edit-story/app/story/actions/useSavePost.js
swissspidy: @miina This comma now seems to be saved as part of post_content.

miina: Hmph, okay, removed and added eslint ignore: https://github.com/ampproject/amp-wp/pull/3999

Issue #3869, opened 2019-12-02
Theme stylesheet gets pushed out even though it is priority 1


schlessera:
## Bug Description

Whenever the CSS hits the 50KB limit, the theme stylesheet, being one of the most important ones, is being purged first, even though it is prioritized at priority 1.

## Expected Behaviour

The theme's stylesheet should be prioritized to only be dropped last, as it has the most impact on the visual similitude between AMP and non-AMP.

## Steps to reproduce

1. Activate the TwentyTwenty theme and activate it.
2. Install the Query Monitor plugin and activate it.
3. Go to a page with lots of diverse content.
4. **Verify:** The page looks visually broken as the theme stylesheet is not applied.
5. Open the "View Page Source..." page.
6. **Verify:** The CSS manifest comment at the top of the HTML source shows that the big Query Monitor stylesheet is kept while the theme stylesheet is dropped.

## Screenshots

![Image 2019-12-02 at 9 49 53 AM](https://user-images.githubusercontent.com/83631/69945278-86d0fc80-14e9-11ea-9081-577e7e1e1313.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
It seems the above is a different issue than what we discussed in Slack. In Slack we talked about the theme's inline styles not getting any prioritization.

The above issue is the regular stylesheet, though, which gets a priority of `1` but is still pushed out over multiple other ones with less priority.

The reason might be that the stylesheet is just so large it will never really fit. I'll dive in deeper.

schlessera:
I'm closing this issue, as it seems to only be caused because the theme stylesheet is already too big on its own to fit within 50kB after the AMP bare requirements are included.

PR #3877, opened 2019-12-03
Prioritize twentytwenty inline CSS

schlessera:
## Summary

This PR adds the following:
- prioritization of Twentytwenty inline CSS (at priority 2)
- unit test for `AMP_Style_Sanitizer::get_stylesheet_priority()`
- bugfix for plugin CSS bug detected through the new tests
- deprioritization of dev plugins like Query Monitor
- `Amp\AmpWP\Tests\PrivateAccess` trait to access private/protected methods and properties
- `Amp\AmpWP\Tests` development autoloader
- add priorities to CSS manifest
- improve CSS manifest alignment

Related #3869 

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -2607,9 +2607,11 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 				if ( 'custom' !== $pending_stylesheet['group'] || ! ( $pending_stylesheet['node'] instanceof DOMElement ) || ! empty( $pending_stylesheet['duplicate'] ) ) {
 					continue;
 				}
-				$message = sprintf( '% 6d B', $pending_stylesheet['size'] );
+				$message = sprintf( '[%3d] % 6d B', $pending_stylesheet['priority'], $pending_stylesheet['size'] );
 				if ( $pending_stylesheet['size'] && $pending_stylesheet['size'] !== $pending_stylesheet['original_size'] ) {
-					$message .= sprintf( ' (%d%%)', $pending_stylesheet['size'] / $pending_stylesheet['original_size'] * 100 );
+					$message .= sprintf( ' (%2d%%)', $pending_stylesheet['size'] / $pending_stylesheet['original_size'] * 100 );
+				} else {
+					$message .= '      ';
 				}
 				$message .= ': ';
 				$message .= $pending_stylesheet['node']->nodeName;
@@ -3132,45 +3134,45 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 
 		unset( $pending_stylesheet );
 
-		// Determine which stylesheets are included based on their priorities.
-		$pending_stylesheet_indices = array_keys( $this->pending_stylesheets );
+		// Sort stylesheets by their priority.
 		usort(
-			$pending_stylesheet_indices,
-			function( $a, $b ) {
-				return $this->pending_stylesheets[ $a ]['priority'] - $this->pending_stylesheets[ $b ]['priority'];
+			$this->pending_stylesheets,
+			static function ( $a, $b ) {
+				return $a['priority'] - $b['priority'];
 			}
 		);
+
 		$current_concatenated_size = 0;
-		foreach ( $pending_stylesheet_indices as $i ) {
-			if ( $group !== $this->pending_stylesheets[ $i ]['group'] ) {
+		foreach ( $this->pending_stylesheets as $index => $pending_stylesheet ) {
+			if ( $group !== $pending_stylesheet['group'] ) {
 				continue;
 			}
 
 			// Skip duplicates.
-			if ( false === $this->pending_stylesheets[ $i ]['included'] ) {
+			if ( false === $pending_stylesheet['included'] ) {
 				continue;
 			}
 
 			// Report validation error if size is now too big.
-			if ( $current_concatenated_size + $this->pending_stylesheets[ $i ]['size'] > $max_bytes ) {
+			if ( $current_concatenated_size + $pending_stylesheet['size'] > $max_bytes ) {
 				$validation_error = [
 					'code' => 'excessive_css',
 					'type' => AMP_Validation_Error_Taxonomy::CSS_ERROR_TYPE,
 				];
-				if ( isset( $this->pending_stylesheets[ $i ]['sources'] ) ) {
-					$validation_error['sources'] = $this->pending_stylesheets[ $i ]['sources'];
+				if ( isset( $pending_stylesheet['sources'] ) ) {
+					$validation_error['sources'] = $pending_stylesheet['sources'];
 				}
 
-				if ( $this->should_sanitize_validation_error( $validation_error, wp_array_slice_assoc( $this->pending_stylesheets[ $i ], [ 'node' ] ) ) ) {
-					$this->pending_stylesheets[ $i ]['included'] = false;
+				if ( $this->should_sanitize_validation_error( $validation_error, wp_array_slice_assoc( $pending_stylesheet, [ 'node' ] ) ) ) {
+					$this->pending_stylesheets[ $index ]['included'] = false;
 					continue; // Skip to the next stylesheet.
 				}
 			}
 
-			if ( ! isset( $this->pending_stylesheets[ $i ]['included'] ) ) {
-				$this->pending_stylesheets[ $i ]['included'] = true;
+			if ( ! isset( $pending_stylesheet['included'] ) ) {
+				$this->pending_stylesheets[ $index ]['included'] = true;
 				$included_count++;
-				$current_concatenated_size += $this->pending_stylesheets[ $i ]['size'];
+				$current_concatenated_size += $pending_stylesheet['size'];
 			}
 		}
 
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -89,6 +89,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 *      @type string   $parsed_cache_variant       Additional value by which to vary parsed cache.
 	 *      @type string   $include_manifest_comment   Whether to show the manifest HTML comment in the response before the style[amp-custom] element. Can be 'always', 'never', or 'when_excessive'.
 	 *      @type string[] $focus_within_classes       Class names in selectors that should be replaced with :focus-within pseudo classes.
+	 *      @type string[] $low_priority_plugins       Plugin slugs of the plugins to deprioritize when hitting the CSS limit.
 	 * }
 	 */
 	protected $args;
@@ -110,6 +111,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 		'parsed_cache_variant'      => null,
 		'include_manifest_comment'  => 'always',
 		'focus_within_classes'      => [ 'focus' ],
+		'low_priority_plugins'      => [ 'query-monitor' ],
 	];
 
 	/**
@@ -931,9 +933,6 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 				'wp-mediaelement',
 				'thickbox',
 			];
-			$low_priority_plugins  = [
-				'query-monitor',
-			];
 
 			if ( in_array( $style_handle, $non_amp_handles, true ) ) {
 				// Styles are for non-AMP JS only so not be used in AMP at all.
@@ -955,7 +954,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 				$priority = 20;
 			} elseif ( $plugin ) {
 				// Styles from plugins are next-highest priority, unless they are in the list of low-priority plugins.
-				$priority = in_array( $plugin, $low_priority_plugins, true ) ? 150 : 30;
+				$priority = in_array( $plugin, $this->args['low_priority_plugins'], true ) ? 150 : 30;
 			} elseif ( 0 === strpos( $schemeless_href, $remove_url_scheme( includes_url() ) ) ) {
 				// Other styles from wp-includes come next.
 				$priority = 40;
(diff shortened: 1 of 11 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -904,9 +904,9 @@ private function get_stylesheet_priority( DOMNode $node ) {
 			$element_id      = (string) $node->getAttribute( 'id' );
 			$schemeless_href = $remove_url_scheme( $node->getAttribute( 'href' ) );
 			$is_plugin_asset = (
-				0 === strpos( $schemeless_href, $remove_url_scheme( trailingslashit( plugins_url( WP_PLUGIN_DIR ) ) ) )
```
schlessera: The plugin prioritization did never work correctly, I think. Not only does the non-MU version use the `_DIR` instead of the `_URL`, they are also wrapped in `plugins_url()`.

`plugins_url()` assumes its argument is a relative path to be appended to the plugin folder URL. So the result was something like `//example.org/wp-content/plugins/Users/alain/wp-content/plugins/`

tests/php/src/PrivateAccess.php
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Trait PrivateAccess.
+ *
+ * Allows accessing private methods and properties for testing.
+ *
+ * @internal
+ * @since 1.5.0
+ */
+trait PrivateAccess {
```
schlessera: As this is something we need multiple times across multiple tests, I thought it made sense to abstract it away into a trait.

I started a new namespace for testing using Composer's `autoload-dev` directive, so it is always accessible without additional includes during tests.

tests/php/test-amp-style-sanitizer.php
```diff
@@ -2409,4 +2413,91 @@ public function test_prioritized_stylesheets( $html_generator, $assert ) {
...
+	 * Test retrieval of stylesheet priorities.
+	 *
+	 * @covers       \AMP_Style_Sanitizer::get_stylesheet_priority()
+	 *
+	 * @dataProvider get_get_stylesheet_priority_data
+	 *
+	 * @param string $node     Node to check the priority of.
+	 * @param int    $expected Expected priority.
+	 */
+	public function test_get_stylesheet_priority( $node, $expected ) {
```
schlessera: The existing tests were not very thorough, so I added a more specific unit test for this method. This uncovered the bug with plugin CSS mentioned above.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -922,6 +931,9 @@ private function get_stylesheet_priority( DOMNode $node ) {
 				'wp-mediaelement',
 				'thickbox',
 			];
+			$low_priority_plugins  = [
+				'query-monitor',
```
schlessera: Any suggestions of what else to add here?

tests/php/src/PrivateAccess.php
kienstra: Great idea, no sense repeating this logic.

includes/sanitizers/class-amp-style-sanitizer.php
pierlon: None that I can think of at the moment.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -956,15 +968,22 @@ private function get_stylesheet_priority( DOMNode $node ) {
...
-			if ( 'admin-bar-inline-css' === $element_id ) {
-				$priority = $admin_bar_priority;
-			} elseif ( 'wp-custom-css' === $element_id ) {
-				// Additional CSS from Customizer.
-				$priority = 60;
-			} else {
-				// Other style elements, including from Recent Comments widget.
-				$priority = 70;
+			switch ( (string) $node->getAttribute( 'id' ) ) {
+				case 'twentytwenty-style-inline-css':
```
westonruter: This should be expanded to account any theme, not just Twenty Twenty. A quick way to do that would be:

```suggestion
				case get_template() . '-style-inline-css':
				case get_stylesheet() . '-style-inline-css':
```

Core themes starting with Twenty Twelve enqueue stylesheets like so:

```php
wp_enqueue_style( 'twentytwelve-style', get_stylesheet_uri(), array(), '20190507' );
```

A more robust way of determining the priority would be derive the underlying style handle from the ID and then look up the registered stylesheet to then determine if it is coming from a stylesheet.

```php
if ( $node->getAttribute( 'id' ) && 
     preg_match( '/^(?P<handle>.+)-inline-css$/', $node->getAttribute( 'id' ), $matches ) &&
     wp_style_is( $matches['handle'], 'registered' ) &&
     (
         0 === strpos( wp_styles()->registered[ $matches['handle'] ]->src, get_template_directory_uri() ) ||
         0 === strpos( wp_styles()->registered[ $matches['handle'] ]->src, get_stylesheet_directory_uri() )
     )
) {
    $priority = 2;
}
```

So I think it's pretty easily implementable, and it would resolve the below todo comment.

tests/php/src/PrivateAccess.php
westonruter: Very nice.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: What about making this list of plugins a sanitizer arg, with `[ 'query-monitor' ]` being the initial value? So instead of `$low_priority_plugins` here it could use `$this->args['low_priority_plugins']`. This would allow the plugins to be configurable via the `amp_content_sanitizers` filter.

Also: Query Monitor should ideally be short-lived here since we want to add dev mode support directly to Query Monitor.

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: Good catch.

schlessera:
> These tests are failing locally for me due to the non-default site URL Lando sets. I think using `content_url()` to form the `href` in the `link` tags is a possible workaround.

What is the URL that is being generated in Lando? Why doesn't it use the default one for tests?

schlessera:
I'm adding the calculated priority to the CSS manifest for easier debugging:

Image 2019-12-04 at 8 40 55 AM

pierlon:
> What is the URL that is being generated in Lando? Why doesn't it use the default one for tests?

In my case the WordPress address during tests is `http://example.org/core-dev/src`, and the site address is `http://example.org`.

It seems `core-dev/src` is appended to the Wordpress address to determine the path to the WordPress codebase being tested against.

westonruter:
> It seems `core-dev/src` is appended to the Wordpress address to determine the path to the WordPress codebase being tested against.

See [`_wordpressdev_set_directory_constants()`](https://github.com/GoogleChromeLabs/wordpressdev/blob/11ae31d62b09ac31f974229b7fb3b49473cadc40/public/functions.php#L88-L130).

I'm also getting failures. Specifically:

```4) AMP_Style_Sanitizer_Test::test_get_stylesheet_priority with data set #3 (DOMElement Object (), 1)
Failed asserting that 50 matches expected 1.

/app/public/content/plugins/amp/tests/php/test-amp-style-sanitizer.php:2647

5) AMP_Style_Sanitizer_Test::test_get_stylesheet_priority with data set #4 (DOMElement Object (), 10)
Failed asserting that 50 matches expected 10.

/app/public/content/plugins/amp/tests/php/test-amp-style-sanitizer.php:2647

6) AMP_Style_Sanitizer_Test::test_get_stylesheet_priority with data set #6 (DOMElement Object (), 30)
Failed asserting that 50 matches expected 30.

/app/public/content/plugins/amp/tests/php/test-amp-style-sanitizer.php:2647

7) AMP_Style_Sanitizer_Test::test_get_stylesheet_priority with data set #7 (DOMElement Object (), 150)
Failed asserting that 50 matches expected 150.

/app/public/content/plugins/amp/tests/php/test-amp-style-sanitizer.php:2647

8) AMP_Style_Sanitizer_Test::test_get_stylesheet_priority with data set #8 (DOMElement Object (), 40)
Failed asserting that 50 matches expected 40.

/app/public/content/plugins/amp/tests/php/test-amp-style-sanitizer.php:2647

9) AMP_Style_Sanitizer_Test::test_get_stylesheet_priority with data set #17 (DOMElement Object (), 101)
Failed asserting that 150 matches expected 101.

/app/public/content/plugins/amp/tests/php/test-amp-style-sanitizer.php:2647

10) AMP_Style_Sanitizer_Test::test_get_stylesheet_priority with data set #18 (DOMElement Object (), 110)
Failed asserting that 150 matches expected 110.

/app/public/content/plugins/amp/tests/php/test-amp-style-sanitizer.php:2647

11) AMP_Style_Sanitizer_Test::test_get_stylesheet_priority with data set #20 (DOMElement Object (), 130)
Failed asserting that 150 matches expected 130.

/app/public/content/plugins/amp/tests/php/test-amp-style-sanitizer.php:2647

12) AMP_Style_Sanitizer_Test::test_get_stylesheet_priority with data set #21 (DOMElement Object (), 250)
Failed asserting that 150 matches expected 250.

/app/public/content/plugins/amp/tests/php/test-amp-style-sanitizer.php:2647

13) AMP_Style_Sanitizer_Test::test_get_stylesheet_priority with data set #22 (DOMElement Object (), 140)
Failed asserting that 150 matches expected 140.

/app/public/content/plugins/amp/tests/php/test-amp-style-sanitizer.php:2647
```

schlessera:
While I was at it, I also changed the sorting algorithm and tweaked the formatting:
Image 2019-12-04 at 10 06 11 AM

Makes more sense to show them in the prioritized order, I think.

schlessera:
> In my case the WordPress address during tests is http://example.org/core-dev/src, and the site address is http://example.org.

Wouldn't it be preferable to fix this on the Lando side, then? The code we test and the actual tests should make be based on the same environment assumptions, otherwise we might have tests that pass locally but code that fails on production instead.

includes/sanitizers/class-amp-style-sanitizer.php
schlessera: Via https://github.com/ampproject/amp-wp/pull/3877/commits/518930d9adbeab05546921120f91ce4cadae51d2

includes/sanitizers/class-amp-style-sanitizer.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/3877/commits/e0ba933ed168ad2bd87d2d5e90bd99f7041b1f5c

westonruter:
> Makes more sense to show them in the prioritized order, I think.

I don't think so. It's important to be in document order because CSS is _cascading_ and styles later override styles before. The entries should match the order of the stylesheets concatenated in `style[amp-custom]`.

westonruter:
> Wouldn't it be preferable to fix this on the Lando side, then? The code we test and the actual tests should be based on the same environment assumptions, otherwise we might have tests that pass locally but code that fails on production instead.

It's not specific to the Lando configuration, though. The test failures would happen on any site that uses non-standard installation configuration, such as a custom `WP_CONTENT_DIR`. I've fixed this in a5018251b2bf3720bef9bb82977eb8fee69dc159 by replacing hard-coded URL paths with the use of `get_theme_root_uri()`, `plugins_url()`, and `includes_url()`.

schlessera:
> I don't think so. It's important to be in document order because CSS is cascading and styles later override styles before. The entries should match the order of the stylesheets concatenated in style[amp-custom].

Oh, indeed, I didn't consider that. I'll revert the change.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -955,11 +966,25 @@ private function get_stylesheet_priority( DOMNode $node ) {
...
+			$id = $node->getAttribute( 'id' );
+			if (
+				preg_match( '/^(?<handle>.+)-inline-css$/', $id, $matches ) &&
+				wp_style_is( $matches['handle'], 'registered' ) &&
+				0 === strpos( wp_styles()->registered[ $matches['handle'] ]->src, get_template_directory_uri() )
+			) {
+				// Parent theme inline style.
+				$priority = 2;
+			} elseif (
+				preg_match( '/^(?<handle>.+)-inline-css$/', $id, $matches ) &&
```
westonruter: Minor points: 

1. The `preg_match()` call with its `$matches` variable, and the `wp_style_is( $matches['handle'], 'registered' )` call as well could be factored out of the conditions and then used in both
2. The second condition should include a check for `get_stylesheet() !== get_template()` for strict correctness.

schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/3877/commits/93c2760e4176bf1b159cafd04d6d42ae520996a6

tests/php/test-amp-style-sanitizer.php
```diff
@@ -2409,4 +2413,255 @@ public function test_prioritized_stylesheets( $html_generator, $assert ) {
...
+
+	/**
+	 * Provide data to test_get_stylesheet_priority().
+	 *
+	 * @return array[] Array of test data.
+	 */
+	public function get_get_stylesheet_priority_data() {
+		return [
+			// Non-AMP handle.
+			[ [ 'link', [ 'id' => 'mediaelement-css' ] ], 1000 ],
```
westonruter: Note: It would be better if the comment "Non-AMP handle" were made into the array key, as this would then make it easier to identify the dataset when failures occur.

For example:

```suggestion
			'Non-AMP handle' => [ [ 'link', [ 'id' => 'mediaelement-css' ] ], 1000 ],
```

schlessera:
*Testing instructions:*
1. Use TwentyTwenty as active theme.
2. Go to an AMP page on the frontend.
3. View the page source via the context menu's "View source..." entry.
4. *Verify:* The CSS manifest comment near the start of the page source contains priority numbers in square brackets before each entry (`[  x]` at the start with `x` being the priority number).
5. *Verify:* The entry `style#twentytwenty-style-inline-css` has the priority `2`.

csossi:
Hi @schlessera - went to view-source:https://story-test-wordpress-amp.pantheonsite.io/2019/12/12/gallery-block-test/?amp but not seeing the above step 4 and 5 notes

schlessera:
@csossi I think this is only displayed when `WP_DEBUG` is set to true, so that's probably the reason why it is not showing.

@westonruter Is `WP_DEBUG` off on the Pantheon testing site?

westonruter:
I've enabled a plugin which forces the manifest comment to appear.

Plugin code:

```php
        $sanitizers['AMP_Style_Sanitizer']['include_manifest_comment'] = 'always';
        return $sanitizers;
} );
```

csossi:
verified in QA

Issue #3819, opened 2019-11-26
Moveable: Visual/general Improvements for Dragging, Resizing, Rotating


miina:
## Feature description

This is a follow-up issue for #3798.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Moveable component is a separate component outside of `Page`.
* Resizing/rotation handles are hidden while dragging (NEEDS UX CONFIRMATION)
* Resizing has minimum limits and doesn't allow decreasing an element to non-existing  (NEEDS UX CONFIRMATION).
* Move the element itself instead of the outline.
* Dragging is possible without selecting the element first. The element will be selected after the dragging event (see Figma for details).

## Implementation brief

General overview:

- The implementation of `Moveable` will be moved to a separate file under `components` and imported to the `Page` instead of being part of it.

- Instead of dragging/resizing/rotating just the outline and then moving the element after that, the element is manipulated directly.

For doing that, we will remove the separate `Selection` element and set the element itself as the target for `Moveable` directly. The `Selection` element was originally to mark an outline for the selected area, however, the element (outline) created by `Moveable` can replace that. We might want to add a wrapper around the original element for target for Moveable, with the actual element being inside it.

For now, we will ignore the content of the element being resizing, meaning that if there's text inside the element, the font size will stay as it is. This can later be addressed once Text editing and automatic resizing has been implemented.

- Dragging will be possible without selecting the element first.
This is related to the previous point as well. Previously, the dragging was applied to `Selection` which only existed if the element was selected which makes it possible to drag it even if not selected previously. This means that we will need to add a `Moveable` component for each element by default for not having to repaint everything. Currently, the elements are re-rendered after selecting and there is only one `Moveable` for the selected element.

- Additionally, the dragged element will be selected within the `onDragEnd` event. The selection should ideally persist without having to set it after each action, currently the selection "gets lost" due to `backgroundClickHandler` being triggered after dragging, which clears the current selection -- needs debugging.

## QA testing instructions

* 

## Demo

* 
![dragging](https://user-images.githubusercontent.com/3294597/70712117-5b1cf600-1ce3-11ea-8b83-b556fba3be2c.gif)

## Changelog entry

*

PR #3884, opened 2019-12-05
Improvements for Moveable

miina:
## Summary

See #3819 

- Allows dragging an element without selecting first.
- Manipulates the element directly instead of the outline only

## Checklist


assets/src/edit-story/app/story/actions/useSetPropertiesById.js
```diff
@@ -0,0 +1,46 @@
...
+	pages,
+	setPages,
+} ) {
+	const setPropertiesById = useCallback( ( elementId, properties ) => {
+		const currentPage = pages[ currentPageIndex ];
+		const newPages = [
+			...pages.slice( 0, currentPageIndex ),
+			{
+				...currentPage,
+				elements: [
```
miina: This new action as well brings up the question if we should combine these.

assets/src/edit-story/components/movable-element/index.js
```diff
@@ -0,0 +1,156 @@
...
+		<>
+			<Element
+				ref={ setTargetEl }
+				key={ id }
+				onClick={ ( evt ) => handleSelectElement( id, evt ) }
+			>
+				<Comp { ...rest } />
+			</Element>
+			{ targetEl && <Moveable
+				target={ targetEl.firstChild }
```
miina: This needs looking into to address the correct element directly.

assets/src/edit-story/components/movable-element/index.js
```diff
@@ -0,0 +1,156 @@
...
+const MovableElement = ( props ) => {
+	const {
+		rotationAngle,
+		x,
+		y,
+		type,
+		selected,
+		id,
+		rest,
+	} = props;
```
barklund: I think the new convention is to do `function` expressions. Also, destruct directly in definition as:

```js
function MovableElement( {
	rotationAngle,
	x,
	y,
	type,
	selected,
	id,
	rest,
} ) {
```

assets/src/edit-story/app/story/actions/useSetPropertiesById.js
```diff
@@ -0,0 +1,46 @@
...
+						return {
+							...element,
+							...properties,
+						};
+					} ),
+				],
+			},
+			...pages.slice( currentPageIndex + 1 ),
+		];
+		setPages( newPages );
```
barklund: This probably needs to use the functional setter to allow multiple elements being updated in the same render loop:

```js
setPages( ( pages ) => {
	// create newPages
	return newPages;
} );
```

That'll also completely remove the need for passing `pages` in at all and just use it inline in the setter.

Actually, all the methods ought to do this now that I think about it.

assets/src/edit-story/elements/image.js
```diff
@@ -16,17 +16,18 @@ const Element = styled.img`
...
 		height,
 		x,
 		y,
 		rotationAngle,
 		src,
+		ref: forwardedRef,
 	};
 	return (
-		<Element { ...props } />
+		<Element ref={ forwardedRef } { ...props } />
```
barklund: Not needed both here and in `props`.

assets/src/edit-story/app/story/actions/useSetPropertiesById.js
miina: Should we perhaps generally have smaller exposed actions that are using non-exposed internal logic within the `useStory`?

E.g. we could have `setPropertiesById` and `setPropertiesByIds` both exposed via `useStory`, however, internally this would always use `setPropertiesByIds` anyway. Perhaps even `setPropertiesBy` could be a general internal function which would accept `by` (e.g. selection or ids)?

Basically just to have the base logic of setting the properties via slicing the `pages` in one place. Thoughts?

assets/src/edit-story/components/movable-element/index.js
miina: Good to know, thanks, noted.

assets/src/edit-story/app/story/actions/useSetPropertiesById.js
barklund: > Basically just to have the base logic of setting the properties via slicing the `pages` in one place. Thoughts?

We could easily do this where relevant. Perhaps it's even better to do it with a reducer and simple actions (but expose it as functions) to move the logic to a more centralized place, as suggested by @dvoytenko elsewhere.

We could probably keep it to the following actions in the reducer and re-use those for the different exposed functions:

```
PAGE_DELETE
PAGE_ADD
PAGE_UPDATE
ELEMENT_DELETE
ELEMENT_ADD
ELEMENT_UPDATE
```

All of the exposed functions would then be a series of reductions of the same state. But can we actually easily do a series of reductions?

Or we would create reducers for each atomic action - we'd need 4 extra actions for duplicating both pages and elements, updating multiple elements and deleting multiple elements. Seems doable too.

And we could get all of the state variables such as currentPage, currentPageId, currentPageNumber, etc directly from the state.

It's probably the right way to do it.

I'll create a (high-priority) ticket about it.

dvoytenko: I added this ticket a bit ago: https://github.com/ampproject/amp-wp/issues/3894

assets/src/edit-story/components/movable-element/index.js
```diff
@@ -0,0 +1,156 @@
...
+
+	return (
+		<>
+			<Element
+				key={ id }
+				onClick={ ( evt ) => handleSelectElement( id, evt ) }
+			>
+				<Comp { ...rest } forwardedRef={ setTargetEl } />
+			</Element>
+			{ targetEl && <Moveable
```
dvoytenko: Are we doing a `Moveable` per element? How would we then do grouped transformations?

assets/src/edit-story/components/movable-element/index.js
```diff
@@ -0,0 +1,156 @@
...
+				ref={ moveable }
+				target={ targetEl }
+				draggable={ true }
+				resizable={ selected }
```
dvoytenko: Would be happy to be convinced otherwise or dig into this deeper. But I think your previous model was better: single moveable can support a single element, or a group of elements, or any additional customizations. Perhaps it's not an issue with moveable as much as we need to switch to do selection with `onPointerDown`? I looked around in other products (such as Google Slides) and it looks like they all do `onPointerDown` selection. That would solve our initial issue, right?

assets/src/edit-story/components/movable-element/index.js
```diff
@@ -0,0 +1,156 @@
...
+ */
+import { useStory } from '../../app';
+import { getComponentForType } from '../../elements';
+
+const Element = styled.div`
+	cursor: pointer;
+	user-select: none;
+`;
+
+const MovableElement = ( props ) => {
```
dvoytenko: This is unrelated to the question of a single moveable vs many moveables below. But I just wanted to note that if we have a concept of "movable/draggable/panable/resizable/rotatable" - they are dependent on the component's states. E.g. an image is "movable", but when it's in the "background/fullbleed" state, it stops being "movable". Not sure ref forwarding would work well for that.

assets/src/edit-story/components/movable-element/index.js
```diff
@@ -0,0 +1,156 @@
...
+		if ( evt.metaKey ) {
+			toggleElementIdInSelection( elId );
+		} else {
+			selectElementById( elId );
+		}
+		evt.stopPropagation();
+	}, [ toggleElementIdInSelection, selectElementById ] );
+
+	const comp = getComponentForType( type );
+	const Comp = comp; // why u do dis, eslint?
```
spacedmonkey: The comment :rofl:

assets/src/edit-story/components/movable-element/index.js
```diff
@@ -0,0 +1,156 @@
...
+					frame.rotate = beforeRotate;
+					setStyle( target );
+				} }
+				onRotateEnd={ () => {
+					setPropertiesOnSelectedElements( { rotationAngle: frame.rotate } );
+				} }
+				origin={ false }
+				pinchable={ true }
+				keepRatio={ 'image' === type }
+				renderDirections={ 'image' === type ? [ 'nw', 'ne', 'sw', 'se' ] : [ 'n', 's', 'e', 'w', 'nw', 'ne', 'sw', 'se' ] }
```
spacedmonkey: This may get more complex as we add new elements. Maybe this is better as a function?

assets/src/edit-story/elements/image.js
spacedmonkey: The name of the variable is different here. Is that why this is done?

spacedmonkey:
```suggestion
		<Element { ...props } />
```
Not needed as not par of any other element.

assets/src/edit-story/components/movable-element/index.js
miina: It would solve the issue for draggable, however, should it be possible to resize/rotate the element before it is selected first? E.g. displaying the handles when hovering close to the element, without selecting (while keeping the same for the selected element, too). Perhaps only dragging should be possible without selecting first, that's how it works in Google Slides and Keynote, too. Will confirm just in case with the UX but I guess you're right and only dragging can be done without selecting first, which could be addressed with using `onPointerDown` for selecting.

assets/src/edit-story/components/movable-element/index.js
miina: The honor of this comment belongs to @barklund 😄

assets/src/edit-story/components/movable-element/index.js
miina: Grouped transformation would have a separate `Moveable`, however, if we can use just `onPointerDown` and won't have to display multiple `Moveable`-s at the same time, then we can move back to using just one `Moveable` for now,

assets/src/edit-story/components/movable-element/index.js
miina: Hmm, actually, I wonder if `onPointerDown` would work since if `Moveable` would be added only once selected, then a new drag event should be initiated for the newly added `Moveable` to take effect.

miina: @dvoytenko Perhaps I'm missing something but it's not working currently for me -- setting the moveable target with `onPointerDown` (selecting the element which triggers the target setting) without having anything selected previously, it needs a new dragevent. I'll continue looking into this tomorrow morning, let me know if you had a working test somewhere, checked it briefly only at this moment.

assets/src/edit-story/components/movable-element/index.js
dvoytenko: I think this is mainly a property such as `preserveAspectRatio=true|false`.

assets/src/edit-story/app/story/actions/useSetPropertiesById.js
miina: @barklund On updating multiple elements in a loop: decided to go with [updateElementsByIds action](https://github.com/ampproject/amp-wp/pull/3909/files#diff-368f816669cc6567ef5ec19001aa7210) instead which would work for both single and multiple element updates. It's currently in the multi-selection PR only, however, will replace the one within this PR with that one. Let me know if you have any concerns.

miina:
@dvoytenko The PR is updated to use only one Moveable now.

Todo and discuss (perhaps in a follow-up issue) is to consider creating a wrapper element around the "actual" element that would get assigned the properties and would also be a target for Moveable instead. Points in favor:
- Easier keypress events handling.
- Easier to add sibling elements in the future if needed, e.g. if the sibling needs to move together with the element, e.g. action buttons etc.
- Having the control over the selection in one place, e.g. the pointer events etc. which right now have to be set for each element separately.

cc @barklund

assets/src/edit-story/app/story/actions/useSetPropertiesById.js
barklund: All this will be irrelevant when #3910 is merged in.

barklund:
> Todo and discuss (perhaps in a follow-up issue) is to consider creating a wrapper element around the "actual" element that would get assigned the properties and would also be a target for Moveable instead.

I think that's a great idea, but let's follow that up when both this and in-place editing is merged in to see where we are then.

assets/src/edit-story/components/movable/index.js
```diff
@@ -0,0 +1,141 @@
...
+		actions: { setPropertiesOnSelectedElements },
+	} = useStory();
+
+	useEffect( () => {
+		if ( moveable.current ) {
+			// If we have persistent event then let's use that.
+			if ( pushEvent ) {
+				moveable.current.moveable.dragStart( pushEvent );
+				// Remove the event right after using, not to influence other elements.
+				setPushEvent( null );
```
dvoytenko: We can avoid passing `setPushEvent` down and forcing re-rendering by either:

1. We don't have to reset it. We can simply check above that `pushEvent.currentTarget == targetEl` and it's a recent event.
2. We could instead pass `pushEventRef` as a ref and use `pushEventRef.current` and clear it out w/o triggering re-rendering.
3. We can keep the re-rendering. It's probably not very expensive :) But might be worth looking into it.

assets/src/edit-story/components/movable/index.js
```diff
@@ -0,0 +1,141 @@
...
+			}
+			moveable.current.updateRect();
+		}
+		// Disable reason: we should not run this when pushEvent changes!
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [ targetEl, moveable ] );
+
+	const frame = {
```
dvoytenko: We can reconsider this with grouping, but just realized - is it ok that this element is reset on each re-render? Should it be a `ref`?

assets/src/edit-story/components/movable/index.js
```diff
@@ -0,0 +1,141 @@
...
+		translate: [ 0, 0 ],
+		rotate: rotationAngle,
+	};
+
+	const setStyle = ( target ) => {
```
dvoytenko: Realistically, is the `target` here ever not `targetEl` from above?

assets/src/edit-story/components/movable/index.js
```diff
@@ -0,0 +1,141 @@
...
+			onRotate={ ( { target, beforeRotate } ) => {
+				frame.rotate = beforeRotate;
+				setStyle( target );
+			} }
+			onRotateEnd={ () => {
+				setPropertiesOnSelectedElements( { rotationAngle: frame.rotate } );
+			} }
+			origin={ false }
+			pinchable={ true }
+			keepRatio={ 'image' === type }
```
dvoytenko: The more I think about this, the more I'm convinced we should just have `keepRatio` property in our element's model. Because even images do not have to keep ratio when the "lock" checkmark unset in the edit panel.

assets/src/edit-story/components/canvas/page.js
```diff
@@ -23,158 +23,83 @@ const Background = styled.div.attrs( { className: 'container' } )`
...
 				return (
-					<Element key={ id } onClick={ handleSelectElement( id ) }>
-						<Comp { ...rest } />
+					<Element
+						key={ id }
+						onClick={ ( evt ) => handleSelectElement( id, evt ) }
+					>
+						<Comp
+							{ ...rest }
+							onPointerDown={ ( evt ) => {
```
dvoytenko: Actually, nothing prevents us from moving up `onPointerDown` to the `<Element>` above, right?

assets/src/edit-story/components/movable/index.js
swissspidy: Makes sense to me 👍

assets/src/edit-story/components/canvas/page.js
```diff
@@ -23,158 +23,83 @@ const Background = styled.div.attrs( { className: 'container' } )`
...
-	useEffect( () => {
-		setBackgroundClickHandler( () => clearSelection() );
-	}, [ setBackgroundClickHandler, clearSelection ] );
 
-	let displayMoveable = false;
-	const frame = {
-		translate: [ 0, 0 ],
-		rotate: selectionProps.rotationAngle,
-	};
+		// @todo That's not necessary for multi-selection.
```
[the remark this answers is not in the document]
miina: Hmm, to think of it, it's actually relevant for multi-selection anyway as well since even if there's multi-selection previously, then it should be able to start dragging a non-selected element.

assets/src/edit-story/components/movable/index.js
miina: With group elements, it would. We can probably keep it for now, will remove it within #3909 if it's not needed.

assets/src/edit-story/components/movable/index.js
miina: Perhaps for now (within this PR), we can keep it as it is? Since currently we don't have the lock mark and without that, it makes more sense for the images always to keep ratio.

assets/src/edit-story/components/canvas/page.js
miina: Just that then the `targetEl` and the `target` of the `pushEvent` would be different, probably better to keep it in sync. If we'd do that, it would already make sense to add the wrapper element and change the `targetEl` target as well.

assets/src/edit-story/components/movable/index.js
miina: Probably should be reset when re-rendering, or which case do you have in mind?

assets/src/edit-story/components/movable/index.js
miina: We should probably do the check for the target elements matching even if we do reset the `pushEvent` afterward :)

On resetting the `pushEvent` -- it feels cleaner not to leave an old event persisted after it's purpose has been fulfilled. However, I guess it shouldn't cause any side effects if we'll always check that the target is correct.

Will go with the option 1. and remove resetting the event for now.

assets/src/edit-story/components/canvas/page.js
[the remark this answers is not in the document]
dvoytenko: Do you mean how should onPointerDown -> drag behave with multi-selection? I think that case is quite obscure and either way is ok in that case.

assets/src/edit-story/components/movable/index.js
dvoytenko: It seemed like re-rendering could happen somewhat randomly while in the middle of dragging. Might be a too obscure case for now, but there are some many dependencies via props, state, and context, that re-rendering is probably pretty chatty.

assets/src/edit-story/components/movable/index.js
dvoytenko: 👍

assets/src/edit-story/components/canvas/page.js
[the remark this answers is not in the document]
pbakaus: @miina that's correct. If you have elements 1-5 selected in multi-selection, and start dragging on unselected element 6, the multi-selection should go away, element 6 should be selected and dragged.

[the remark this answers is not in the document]
barklund: @pbakaus, what if the user holds down the meta key for multi-selecting elements (be that <kbd>shift</kbd>, <kbd>ctrl</kbd> or <kbd>cmd</kbd>) when click-dragging the 6th element?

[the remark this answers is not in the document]
miina: Note: In #3909, the idea is to ignore `onPointerDown` if a meta key is held down and add the element to the selection. Perhaps that's a good discussion for that PR.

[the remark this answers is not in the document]
pbakaus: yes, I think ignoring the click then is likely the most logical UX. Let's go with that for now, we can revisit if it feels strange.

PR #3793, opened 2019-11-19
Improve error messages when validation requests fail

westonruter:
## Summary

This PR addresses something I noticed when troubleshooting a [support topic](https://wordpress.org/support/topic/links-and-buttons-are-disabled/#post-12146493), namely that when a validation request fails the error message is too generic and not provide enough context for why the problem is happening and how it can be resolved. This includes direction to check Site Health and how to find the support forum and submit topics.

For testing, there is a [plugin](https://gist.github.com/westonruter/d34f49b6288ed52604f7e9b6926810a9) which allows you to (re-)validate URLs with particular query vars to trigger various error scenarios, for example:

* `?amp_simulate_validate_request_error=response_comment_absent`
* `?amp_simulate_validate_request_error=wsod`
* `?amp_simulate_validate_request_error=bad_host`
* `?amp_simulate_validate_request_error=timeout`

This PR also addresses an issue discovered where query params added to the original URLs are normalized-away in the resulting URL that is used for the `amp_validated_url` post type. It also fixes encoding query params in URLs that are added to links in the admin bar.

### Before

Screen Shot 2019-11-18 at 21 51 25

### After

Screen Shot 2019-11-18 at 21 50 33

Relates to #3789 and #1371 and #2199.

## Checklist

- [ ] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [ ] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [ ] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

PR #3898, opened 2019-12-07
Serve back JSON exclusively in amp_validate requests

### LINKED ISSUE #3887: Problem: "URL validation failed to due to the absence of the expected JSON-containing AMP_VALIDATION comment after the body"
When validate url this message show:
"URL validation failed to due to the absence of the expected JSON-containing AMP_VALIDATION comment after the body."

## Expected Behaviour

Show the errors of validation

- WordPress version: 5.3
 - Plugin version: 1.4.1
 - AMP plugin template mode: transitional
 - PHP version: 7.3
 - OS: Centos 7
 - Browser: [e.g. chrome, safari] Firefox

In the log files I have messages related to the use of is_amp_function related to a plugin, but when this plugin is disabled, the messages go away and the issue with validat url continue.

In the acces log file I get this:
`/wp-admin/edit.php?post_type=amp_validated_url&amp_validate_error=response_comment_absent&amp_urls_tested=0 `

I have tried adding ?amp&amp_validate (as stated in #1733) to the url, get the same mesage when a try to view the validation errors.

iconv is installed as estated in #3207 

With all plugins disabled and a default theme, the issue is here.
I have a list of plugins that when enabled the amp validator icon in admin side have issue, but cannot view cause this:

Autoptimize
jQuery Updater
Sassy Social Share
Lazy Loader

But urls get amp validated in google and in amp validator.

Captura

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

westonruter:
## Summary

When performing an AMP validation request, the AMP plugin currently gathers up that information and injects it into the HTML response as an HTML comment after the `</body>`. This HTML comment contains the validation data represented as JSON. The HTML comment, however, is prone to be corrupted or (in the case of PageSpeed) removed entirely.

In thinking about how validation requests are performed, I realized that there is no need to inject the validation JSON data into an HTML document: we can simply return the JSON as the response entirely. This simplifies the process of generating the response, while also improving performance of validation requests by avoiding needlessly to remove HTML source stack comments and to serialize the DOM into HTML.

Additionally, in this PR:

* If nonce is not supplied, the validation request fails early rather than failing silently. An error message is communicated if the nonce check fails.
* A specific error message is displayed when there are too many redirects.
* Fixed PHP≤5.5 DOM compatibility where `textContent` is not writable, in updating the AMP admin bar item.

Fixes #3887. Improvements to error messages follow up on #3793.

## Todo

- [x] Add tests for `\AMP_Validation_Manager::finalize_validation()`.
- [x] Send back error when doing validate request without authorization.

## Checklist


westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3935397/amp.zip) (v1.5.0-alpha-20191207T181223Z-c3e9208b)

westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3936209/amp.zip) (v1.5.0-alpha-20191208T122237Z-31dc013c)

tests/php/validation/test-class-amp-validation-manager.php
```diff
@@ -1536,50 +1529,71 @@ public function test_get_amp_validate_nonce() {
...
 
-		// Ensure that should_validate_response() is true, so finalize_validation() will append the AMP_VALIDATION comment.
-		$post = self::factory()->post->create();
-		$_GET[ AMP_Validation_Manager::VALIDATE_QUERY_VAR ] = 1;
-		$this->set_capability();
 		AMP_Validation_Manager::finalize_validation( $dom );
-		$this->assertTrue( (bool) preg_match( '#AMP_VALIDATION:({.+})#s', $dom->documentElement->lastChild->nodeValue, $matches ) );
-		$returned_valudation = json_decode( $matches[1], true );
-		$this->assertEquals( $validation_results, $returned_valudation['results'] );
+		$this->assertEquals( '⚠️', $status_icon_element->textContent );
```
westonruter: This is failing for some reason on WordPress 5.0 and 4.9.

westonruter: Fixed in ef657b5ba3320c7e6a6d6169b4ab613fc8ca2fc3

includes/validation/class-amp-validation-manager.php
```diff
@@ -1884,14 +1866,13 @@ public static function validate_url( $url ) {
 			$validation_url
 		);
 
-		$response = wp_remote_retrieve_body( $r );
-		if ( trim( $response ) === '' ) {
+		if ( '' === $response ) {
 			return new WP_Error( 'white_screen_of_death' );
```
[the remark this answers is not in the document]
westonruter: Yes, the white screen of death can still happen. For example, if an exception is thrown in one of the sanitizers, I believe.

includes/class-amp-theme-support.php
```diff
@@ -2314,6 +2314,12 @@ public static function prepare_response( $response, $args = [] ) {
 
 		$assets = AMP_Content_Sanitizer::sanitize_document( $dom, self::$sanitizer_classes, $args );
 
+		// Respond early with validation results if performing a validation request.
+		if ( AMP_Validation_Manager::$is_validate_request ) {
+			header( 'Content-Type: application/json; charset=utf-8' );
+			return wp_json_encode( AMP_Validation_Manager::get_validation_response_data(), JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES );
```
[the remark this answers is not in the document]
westonruter: No, probable not. It was originally pretty-printed because it appeared in the HTML comment and so it could be easily read. But now that this is no longer done, `JSON_PRETTY_PRINT` can be removed.

includes/validation/class-amp-validation-manager.php
```diff
@@ -1975,32 +1956,72 @@ public static function get_validate_url_error_message( $error_code, $error_messa
...
+						esc_html( $error_message ),
+						$check_error_log,
+						$support_forum_message,
+					]
+				);
 			case '404':
-				return __( 'The fetched URL was not found. It may have been deleted. If so, you can trash this.', 'amp' ) . $error_message . $support_forum_message;
+				return $implode_non_empty_strings_with_spaces(
+					[
+						esc_html__( 'The fetched URL was not found. It may have been deleted. If so, you can trash this.', 'amp' ),
```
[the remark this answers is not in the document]
westonruter: You're right. Trash isn't right. Actually the terminology used is “forget”:

https://github.com/ampproject/amp-wp/blob/e6f9dc3587954e40c0c21bc66ccfa88f99aaa1b3/includes/validation/class-amp-validated-url-post-type.php#L2281-L2292

So how about:

```suggestion
						esc_html__( 'The fetched URL was not found. It may have been deleted. If so, you can forget this validated URL.', 'amp' ),
```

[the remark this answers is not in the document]
pierlon: `forget` is what I had originally in my previous comment, but it sounded confusing to me. If `forget` is going to be used though, I would suggest making it a link and adding a perhaps adding a `title` attribute to provide context.

includes/class-amp-theme-support.php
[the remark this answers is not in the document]
pierlon: Right, thanks for the clarification.

includes/validation/class-amp-validation-manager.php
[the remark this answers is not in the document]
westonruter: In reality, this error message is very rarely shown. It can only happen if someone has a validated URL post open in one tab, then someone forgets the URL in another tab, and then tries to re-check the URL that had been forgotten. As such, I don't think we need to worry too much about tailoring the error message.

includes/validation/class-amp-validation-manager.php
```diff
@@ -1975,32 +1956,72 @@ public static function get_validate_url_error_message( $error_code, $error_messa
...
+						esc_html__( 'An internal server error occurred when fetching the URL for validation.', 'amp' ),
+						esc_html( $error_message ),
+						$check_error_log,
+						$support_forum_message,
+					]
+				);
+			case 'response_not_json':
+				return $implode_non_empty_strings_with_spaces(
+					[
+						esc_html__( 'URL validation failed to due to the AMP validation request not returning JSON data. This is may be due to a PHP fatal error occurring.', 'amp' ),
```
schlessera:
```suggestion
						esc_html__( 'URL validation failed due to the AMP validation request not returning JSON data. This may be due to a PHP fatal error occurring.', 'amp' ),
```

includes/validation/class-amp-validation-manager.php
```diff
@@ -1975,32 +1956,72 @@ public static function get_validate_url_error_message( $error_code, $error_messa
...
 			case 'malformed_json_validation_errors':
-				return sprintf(
-					/* translators: %1$s: AMP_VALIDATION, %2$s: </body> */
-					__( 'URL validation failed to due to unexpected JSON in the %1$s HTML comment after %2$s.', 'amp' ),
-					'<code>AMP_VALIDATION</code>',
-					'<code>&lt;/body&gt;</code>'
-				) . $error_message . $support_forum_message;
+				return $implode_non_empty_strings_with_spaces(
+					[
+						esc_html__( 'URL validation failed to due to unexpected JSON in AMP validation response.', 'amp' ),
```
schlessera:
```suggestion
						esc_html__( 'URL validation failed due to unexpected JSON in AMP validation response.', 'amp' ),
```

includes/validation/class-amp-validation-manager.php
```diff
@@ -1946,15 +1927,15 @@ public static function unserialize_validation_error_messages( $serialized ) {
 	public static function get_validate_url_error_message( $error_code, $error_message = '' ) {
 		$check_error_log = sprintf(
 			/* translators: %s is link to Debugging in WordPress */
-			' ' . __( 'Please check your server PHP error logs; to do this you may need to <a href="%s" target="_blank">enable</a> <code>WP_DEBUG_LOG</code>.', 'amp' ),
+			__( 'Please check your server PHP error logs; to do this you may need to <a href="%s" target="_blank">enable</a> <code>WP_DEBUG_LOG</code>.', 'amp' ),
```
schlessera: Translations should be escaped, as the actual HTML that will be sent to the browser is crowd-sourced user-input.

includes/validation/class-amp-validation-manager.php
```diff
@@ -1946,15 +1927,15 @@ public static function unserialize_validation_error_messages( $serialized ) {
...
 
 		if ( $error_message ) {
-			$error_message = ' ' . rtrim( $error_message, '.' ) . '.';
+			$error_message = rtrim( $error_message, '.' ) . '.';
 		}
 
-		$support_forum_message = ' ' . sprintf(
+		$support_forum_message = sprintf(
 			/* translators: %1$s: Link to support forum. %2$s: Link to new topic form in support forum. */
 			__( 'If you are stuck, please search the <a href="%1$s">support forum</a> for possible related topics, or otherwise start a <a href="%2$s">new support topic</a> including the error message, the URL to your site, and your active theme/plugins.', 'amp' ),
```
schlessera: Translations should be escaped, as the actual HTML that will be sent to the browser is crowd-sourced user-input.

includes/validation/class-amp-validation-manager.php
```diff
@@ -1963,7 +1944,7 @@ public static function get_validate_url_error_message( $error_code, $error_messa
 
 		$site_health_message = '';
 		if ( version_compare( get_bloginfo( 'version' ), '5.2', '>=' ) ) {
-			$site_health_message .= ' ' . sprintf(
+			$site_health_message .= sprintf(
 				/* translators: %s is link to Site Health */
 				__( 'Please check your <a href="%s">Site Health</a> to verify it can perform loopback requests.', 'amp' ),
```
schlessera: Translations should be escaped, as the actual HTML that will be sent to the browser is crowd-sourced user-input.

schlessera:
Love the simplification!

includes/validation/class-amp-validation-manager.php
```diff
@@ -168,24 +168,25 @@ class AMP_Validation_Manager {
...
 
-		self::$should_locate_sources = $args['should_locate_sources'];
+		// Short-circuit validation requests that are unauthorized.
+		if ( $should_validate_response instanceof WP_Error ) {
+			wp_send_json(
+				[
+					'code'    => $should_validate_response->get_error_code(),
+					'message' => $should_validate_response->get_error_message(),
+				],
+				400
```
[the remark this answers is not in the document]
westonruter: I did have it a 401 initially, but I switched it to be a generic 400 later since I had more conditions in this in `should_validate_response`, but now I can flip it back to 401.

includes/validation/class-amp-validation-manager.php
```diff
@@ -182,12 +182,12 @@ public static function init() {
 					'code'    => $should_validate_response->get_error_code(),
 					'message' => $should_validate_response->get_error_message(),
 				],
-				400
+				401
```
schlessera: The `401` code requires a `WWW-Authenticate` to be sent along as well: https://httpstatuses.com/401

I suggest using something like the following:
```
WWW-Authenticate: WordPress realm="wp-login.php", title="Log in via /wp-login.php"
```

The above is a guess based on reading up on the syntax (https://tools.ietf.org/html/rfc7235#section-4.1). Feel free to adapt if you know more about this header tag.

includes/class-amp-theme-support.php
```diff
@@ -2104,7 +2104,7 @@ public static function prepare_response( $response, $args = [] ) {
 		$enable_response_caching = (
 			$enable_response_caching
 			&&
-			! AMP_Validation_Manager::should_validate_response()
+			! AMP_Validation_Manager::$is_validate_request
```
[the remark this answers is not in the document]
westonruter: I went with `validate` because query param is `amp_validate` and there are methods like: 
* `\AMP_Validation_Manager::get_amp_validate_nonce()`
* `\AMP_Validation_Manager::get_validate_url_error_message()`
* `\AMP_Validation_Manager::should_validate_response()`

The outlier here is `\AMP_Validation_Manager::get_validation_response_data()`. So to be consistent, I think actually it should be renamed to `\AMP_Validation_Manager::get_validate_response_data()`.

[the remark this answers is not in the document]
schlessera: You don't really care about grammar, do you...? 😜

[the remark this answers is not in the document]
westonruter: See 792f6bb223bab3f99a54bdd837a020b8f0a67d51. 

I don't really see it as being ungrammatical. It's a “request to validate” aka a “validate request”, like a “request to pull” is also a “pull request”.

This is somewhat of an attempt to distinguish validation which happens with generation of every AMP page, with the “deep” validation that obtains the source information for the underlying validation errors. In other words, `validation` happens all the time as a side effect of requesting an AMP page, but `validate` only happens if you are explicitly requesting the validation results.

Where I know you will agree is that `AMP_Validation_Manager` is doing too much and needs to be split up. So we'll be sure to revisit the terminology when we do that. 😄

includes/validation/class-amp-validation-manager.php
westonruter: The authentication is not performed via logging-in to WordPress however. With the changes in this PR, authentication is always performed via nonce query param. So instead of `WordPress` would it be `nonce`?

Isn't “requires” a bit strong here? WordPress isn't sending back this header. For example:

```
$ curl -i -X PUT -d 'title=hello' 'https://wordpressdev.lndo.site/wp-json/wp/v2/posts/1'
HTTP/2 401
access-control-allow-headers: Authorization, Content-Type
access-control-expose-headers: X-WP-Total, X-WP-TotalPages
allow: GET
content-type: application/json; charset=UTF-8
date: Thu, 12 Dec 2019 07:59:25 GMT
link: <https://wordpressdev.lndo.site/wp-json/>; rel="https://api.w.org/"
server: nginx/1.14.2
x-content-type-options: nosniff
x-powered-by: PHP/7.3.12
x-robots-tag: noindex
content-length: 107

{"code":"rest_cannot_edit","message":"Sorry, you are not allowed to edit this post.","data":{"status":401}}
```

schlessera: Well, WordPress is hardly a good reference.

From the RFC 7235 (Hypertext Transfer Protocol (HTTP/1.1): Authentication):
```
The server generating a 401 response MUST send a WWW-Authenticate header field
(Section 4.1) containing at least one challenge applicable to the target resource.
```
See https://tools.ietf.org/html/rfc7235#section-3.1

schlessera: I don't think anything breaks if we don't send it, but the response would technically be invalid.

schlessera: If I understand the RFCs correctly, `nonce` would be a property of the authentication type `Digest`. And when you send `nonce`, you're supposed to send a new one as a value so the client can retry with a valid nonce:
```
WWW-Authenticate: Digest nonce=12345
```
See https://tools.ietf.org/html/rfc2617#section-3.2.1

schlessera: I don't feel strongly about this, though. Seems like this is all so old and weird and badly documented that no-one gets it right anyway.

includes/validation/class-amp-validation-manager.php
```diff
@@ -168,24 +168,25 @@ class AMP_Validation_Manager {
...
 	 */
-	public static function init( $args = [] ) {
-		$args = array_merge(
-			[
-				'should_locate_sources' => self::should_validate_response(),
-			],
-			$args
-		);
+	public static function init() {
+		$should_validate_response = self::should_validate_response();
```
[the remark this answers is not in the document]
westonruter: True, but a reason against this is that `\AMP_Validation_Manager::$is_validate_request` is defined as a `bool`. It would seem a bit weird to save a `WP_Error` instance into it.

includes/validation/class-amp-validation-manager.php
schlessera: Just tested with a Google API:
Image 2019-12-12 at 9 20 12 AM

includes/class-amp-theme-support.php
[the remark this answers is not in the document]
schlessera: Ha, dangling refactoring carrots in front of me now...

Alright, works for me! 😄

includes/validation/class-amp-validation-manager.php
```diff
@@ -1617,19 +1594,56 @@ public static function get_amp_validate_nonce() {
...
+	/**
+	 * Get validation response data.
+	 *
+	 * @return array Validation response data.
+	 */
+	public static function get_validation_response_data() {
+		$data = [
+			'results' => self::$validation_results,
+		];
+		if ( get_queried_object() ) {
```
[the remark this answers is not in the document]
westonruter: By pulling in global data you mean populating properties of the global `$wp_query`? Note that `\WP_Query::get_queried_object()` will short-circuit when `\WP_Query::$queried_object` has been set.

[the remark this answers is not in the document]
westonruter: Updated in 2a1dc540.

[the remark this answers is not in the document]
schlessera: Yes, but you're going through `get_queried_object()` first, and that fetches the global `$wp_query` object first. Although, unlikely, this is technically brittle code, as an operation in-between could change the object that variable points to, and you'll end up with a different queried object.

You're assuming that you'll get the same object every time, instead of ensuring you're only working with a single one yourself. (see https://ocramius.github.io/extremely-defensive-php/#/71)

includes/validation/class-amp-validation-manager.php
westonruter: True. I was not doing it here because the function notes that it returns “Error message with HTML markup.” When the return value is used, it is getting passed through `wp_kses()` when the error is actually being printed:

https://github.com/ampproject/amp-wp/blob/2a1dc540e6ecc2621a48cf4324308b2c23f7b6ce/includes/validation/class-amp-validated-url-post-type.php#L1281-L1291

But I suppose it would be a good idea to add more hardening.

westonruter: See d48ceae6f50623a6e0abb63057f7dd294e98d54c

includes/validation/class-amp-validation-manager.php
westonruter: See d48ceae6f50623a6e0abb63057f7dd294e98d54c

includes/validation/class-amp-validation-manager.php
westonruter: See d48ceae6f50623a6e0abb63057f7dd294e98d54c

includes/validation/class-amp-validation-manager.php
westonruter: Google is hardly an authority here! 

😉

westonruter: Well, we might as well go full on board here and actually send the nonce via the `Authorization` request header as well…

westonruter: Humm. Except this would cause problems for sites that are already behind some HTTP authentication and they'd need to provide double authentication.

westonruter: In that case I suppose multiple headers could be sent: https://stackoverflow.com/a/38515091

westonruter: I started going down this road, but I think it's overkill for now.

Stashed patch:

```diff
diff --git a/includes/validation/class-amp-validation-manager.php b/includes/validation/class-amp-validation-manager.php
index b724aae9..feb496c6 100644
--- a/includes/validation/class-amp-validation-manager.php
+++ b/includes/validation/class-amp-validation-manager.php
@@ -19,6 +19,13 @@ class AMP_Validation_Manager {
 	 */
 	const VALIDATE_QUERY_VAR = 'amp_validate';
 
+	/**
+	 * Custom HTTP authorization type that is sent in Authorization request header and WWW-Authenticate response header.
+	 *
+	 * @var string
+	 */
+	const VALIDATE_AUTHORIZATION_TYPE = 'AMP-WP-Validate-Nonce-v1';
+
 	/**
 	 * Query var for containing the number of validation errors on the frontend after redirection when invalid.
 	 *
@@ -1614,7 +1621,7 @@ class AMP_Validation_Manager {
 	 * @return string Nonce.
 	 */
 	public static function get_amp_validate_nonce() {
-		return substr( wp_hash( self::VALIDATE_QUERY_VAR . wp_nonce_tick(), 'nonce' ), -12, 10 );
+		return wp_hash( self::VALIDATE_QUERY_VAR . wp_nonce_tick(), 'nonce' );
 	}
 
 	/**
@@ -1632,11 +1639,44 @@ class AMP_Validation_Manager {
 			return false;
 		}
 
-		$validate_key = wp_unslash( $_GET[ self::VALIDATE_QUERY_VAR ] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		if ( self::get_amp_validate_nonce() !== $validate_key ) {
+		$validate_nonce = self::get_amp_validate_nonce();
+		$validate_key   = wp_unslash( $_GET[ self::VALIDATE_QUERY_VAR ] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		if ( $validate_nonce === $validate_key ) {
+			return true;
+		}
+
+		if ( empty( $_SERVER['HTTP_AUTHORIZATION'] ) ) {
+			return new WP_Error(
+				'http_request_failed',
+				esc_html__( 'Missing authorization header.', 'amp' )
+			);
+		}
+
+		$request_nonce  = null;
+		$authorizations = array_map( 'trim', explode( ',', wp_unslash( $_SERVER['HTTP_AUTHORIZATION'] ) ) );
+		$pattern        = sprintf( '/^%s\s+(?P<nonce>\w+)$/', self::VALIDATE_AUTHORIZATION_TYPE );
+		foreach ( $authorizations as $authorization ) {
+			if ( preg_match( $pattern, $authorization, $match ) ) {
+				$request_nonce = $match['nonce'];
+				break;
+			}
+		}
+
+		if ( empty( $request_nonce ) ) {
+			return new WP_Error(
+				'http_request_failed',
+				esc_html__( 'Missing authorization header.', 'amp' )
+			);
+		}
+
+		if ( $validate_nonce !== $request_nonce ) {
 			return new WP_Error(
 				'http_request_failed',
-				__( 'Nonce authentication failed.', 'amp' )
+				sprintf(
+					/* translators: %s is the WP-CLI command to generate the nonce */
+					esc_html__( 'Nonce authentication failed. The required nonce is returned by %s.', 'amp' ),
+					'<code>wp amp validation generate-nonce</code>'
+				)
 			);
 		}
 
@@ -1817,7 +1857,7 @@ class AMP_Validation_Manager {
 	public static function validate_url( $url ) {
 
 		$added_query_vars = [
-			self::VALIDATE_QUERY_VAR   => self::get_amp_validate_nonce(),
+			self::VALIDATE_QUERY_VAR   => '',
 			self::CACHE_BUST_QUERY_VAR => wp_rand(),
 		];
 		$validation_url   = add_query_arg( $added_query_vars, $url );
@@ -1836,6 +1876,7 @@ class AMP_Validation_Manager {
 					'redirection' => 0, // Because we're in a loop for redirection.
 					'headers'     => [
 						'Cache-Control' => 'no-cache',
+						'Authorization' => sprintf( '%s %s', self::VALIDATE_AUTHORIZATION_TYPE, self::get_amp_validate_nonce() ),
 					],
 				]
 			);
```

includes/cli/class-amp-cli-validation-command.php
```diff
@@ -314,6 +314,70 @@ public function reset( $args, $assoc_args ) {
...
+
+		$result = AMP_Validation_Manager::validate_url( $url );
+		if ( $result instanceof WP_Error ) {
+			WP_CLI::error( $result );
+		}
+
+		print wp_json_encode( AMP_Validation_Manager::validate_url( $url ), JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES );
+	}
+
 	/**
```
westonruter: @schlessera Thoughts on these new commands?

includes/cli/class-amp-cli-validation-command.php
```diff
@@ -314,6 +314,69 @@ public function reset( $args, $assoc_args ) {
...
+			}
+			$url = $origin . '/' . ltrim( $url, '/' );
+		}
+
+		$result = AMP_Validation_Manager::validate_url( $url );
+		if ( $result instanceof WP_Error ) {
+			WP_CLI::error( $result );
+		}
+
+		print wp_json_encode( AMP_Validation_Manager::validate_url( $url ), JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES );
```
westonruter: Would it be better to use `WP_CLI::line()` here?

schlessera: `WP_CLI::log()` should be used here. `WP_CLI::line()` is a legacy compat method.

schlessera: No, forget my rambling. You are correct, `WP_CLI::line()` should be used here, as this output should not be suppressed when using the `--quiet` flag.

includes/cli/class-amp-cli-validation-command.php
```diff
@@ -314,6 +314,69 @@ public function reset( $args, $assoc_args ) {
...
+		list( $url ) = $args;
+
+		$host            = wp_parse_url( $url, PHP_URL_HOST );
+		$parsed_home_url = wp_parse_url( home_url( '/' ) );
+
+		if ( $host && $host !== $parsed_home_url['host'] ) {
+			WP_CLI::error(
+				sprintf(
+					/* translators: %1$s is the expected host, %2$s is the actual host */
+					__( 'Supplied URL must be for this WordPress install. Expected host "%1$s" but provided is "%2$s".', 'amp' ),
```
schlessera: WP-CLI output is generally not translated, to keep it easily parseable by scripts.

includes/cli/class-amp-cli-validation-command.php
```diff
@@ -314,6 +314,69 @@ public function reset( $args, $assoc_args ) {
...
+	 *
+	 * @param array $args Args.
+	 */
+	public function check_url( $args ) {
+		list( $url ) = $args;
+
+		$host            = wp_parse_url( $url, PHP_URL_HOST );
+		$parsed_home_url = wp_parse_url( home_url( '/' ) );
+
+		if ( $host && $host !== $parsed_home_url['host'] ) {
```
schlessera: The `'host'` index could be missing here.

includes/cli/class-amp-cli-validation-command.php
```diff
@@ -314,6 +314,69 @@ public function reset( $args, $assoc_args ) {
...
+					/* translators: %1$s is the expected host, %2$s is the actual host */
+					__( 'Supplied URL must be for this WordPress install. Expected host "%1$s" but provided is "%2$s".', 'amp' ),
+					$parsed_home_url['host'],
+					$host
+				)
+			);
+		}
+
+		if ( ! $host ) {
+			$origin = $parsed_home_url['scheme'] . '://' . $parsed_home_url['host'];
```
schlessera: Same here, some of these array indexes might be missing.

includes/cli/class-amp-cli-validation-command.php
schlessera: Actually, given the input, the command should support a `--format` flag and the output should be processed via `WP_CLI\Utils\format_items()` instead.

This way it would default to the normal WP-CLI tabular output, but you could request JSON if you wanted to.

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -1885,7 +1901,7 @@ public static function print_status_meta_box( $post ) {
 		<script>
 		jQuery( function( $ ) {
 			var validateUrl, postId;
-			validateUrl = <?php echo wp_json_encode( add_query_arg( AMP_Validation_Manager::VALIDATE_QUERY_VAR, AMP_Validation_Manager::get_amp_validate_nonce(), self::get_url_from_post( $post ) ) ); ?>;
+			validateUrl = <?php echo wp_json_encode( self::get_markup_status_preview_url( self::get_url_from_post( $post ) ) ); ?>;
```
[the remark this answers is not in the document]
westonruter: True, but we don't allow it yet in the PHPCS config, so we're not using it anywhere else in the codebase yet.

includes/cli/class-amp-cli-validation-command.php
westonruter: Could it? If `home_url()` returns a URL without a `host` then that would be bad!

westonruter: Hardened in b6adafa.

includes/cli/class-amp-cli-validation-command.php
westonruter: I don't think the data would work well in a tabular format. Would it? It looks like this:

```json
{
    "results": [
        {
            "error": {
                "node_name": "script",
                "parent_name": "div",
                "code": "DISALLOWED_TAG",
                "type": "js_error",
                "node_attributes": [],
                "text": "document.write(\"BAD!\");",
                "sources": [
                    {
                        "hook": "the_content",
                        "filter": true,
                        "post_id": 2531,
                        "post_type": "post",
                        "sources": [
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "class-wp-embed.php",
                                "line": 59,
                                "function": "WP_Embed::run_shortcode"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "class-wp-embed.php",
                                "line": 404,
                                "function": "WP_Embed::autoembed"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "blocks.php",
                                "line": 532,
                                "function": "do_blocks"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 51,
                                "function": "wptexturize"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 455,
                                "function": "wpautop"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 838,
                                "function": "shortcode_unautop"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "post-template.php",
                                "line": 1649,
                                "function": "prepend_attachment"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "media.php",
                                "line": 1432,
                                "function": "wp_make_content_images_responsive"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 5726,
                                "function": "wp_staticize_emoji"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 5359,
                                "function": "capital_P_dangit"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "shortcodes.php",
                                "line": 177,
                                "function": "do_shortcode"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 3335,
                                "function": "convert_smilies"
                            }
                        ]
                    },
                    {
                        "block_name": "core/shortcode",
                        "post_id": 2531,
                        "block_content_index": 0,
                        "type": "plugin",
                        "name": "gutenberg",
                        "file": "build/block-library/blocks/shortcode.php",
                        "line": 16,
                        "function": "gutenberg_render_block_core_shortcode"
                    },
                    {
                        "type": "plugin",
                        "name": "bad-shortcode.php",
                        "file": "bad-shortcode.php",
                        "line": 7,
                        "function": "Bad\\shortcode",
                        "shortcode": "bad"
                    }
                ]
            },
            "sanitized": true
        },
        {
            "error": {
                "node_name": "script",
                "parent_name": "div",
                "code": "DISALLOWED_TAG",
                "type": "js_error",
                "node_attributes": [],
                "text": "badness",
                "sources": [
                    {
                        "hook": "the_content",
                        "filter": true,
                        "post_id": 2531,
                        "post_type": "post",
                        "sources": [
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "class-wp-embed.php",
                                "line": 59,
                                "function": "WP_Embed::run_shortcode"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "class-wp-embed.php",
                                "line": 404,
                                "function": "WP_Embed::autoembed"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "blocks.php",
                                "line": 532,
                                "function": "do_blocks"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 51,
                                "function": "wptexturize"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 455,
                                "function": "wpautop"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 838,
                                "function": "shortcode_unautop"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "post-template.php",
                                "line": 1649,
                                "function": "prepend_attachment"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "media.php",
                                "line": 1432,
                                "function": "wp_make_content_images_responsive"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 5726,
                                "function": "wp_staticize_emoji"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 5359,
                                "function": "capital_P_dangit"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "shortcodes.php",
                                "line": 177,
                                "function": "do_shortcode"
                            },
                            {
                                "type": "core",
                                "name": "wp-includes",
                                "file": "formatting.php",
                                "line": 3335,
                                "function": "convert_smilies"
                            }
                        ]
                    },
                    {
                        "block_name": "core/html",
                        "post_id": 2531,
                        "block_content_index": 1
                    }
                ]
            },
            "sanitized": true
        },
        {
            "error": {
                "node_name": "script",
                "parent_name": "body",
                "code": "DISALLOWED_TAG",
                "type": "js_error",
                "node_attributes": [],
                "text": "document.write( 'I am invalid!!' );",
                "sources": [
                    {
                        "type": "plugin",
                        "name": "print-invaliud-footer.php",
                        "file": "print-invaliud-footer.php",
                        "line": 8,
                        "function": "{closure}",
                        "hook": "wp_footer",
                        "priority": 10
                    }
                ]
            },
            "sanitized": true
        }
    ],
    "queried_object": {
        "id": 2531,
        "type": "post"
    },
    "url": "https://wordpressdev.lndo.site/2019/12/05/bad-shortcode/?amp"
}
```

includes/validation/class-amp-validation-manager.php
```diff
@@ -1638,96 +1679,54 @@ public static function should_validate_response() {
...
+						'amp'
+					),
+					number_format_i18n( $error_count )
+				);
 			}
 		}
+
+		$admin_bar_icon = $dom->getElementById( 'amp-admin-bar-item-status-icon' );
+		if ( $admin_bar_icon ) {
+			$admin_bar_icon->firstChild->nodeValue = "\xE2\x9A\xA0\xEF\xB8\x8F"; // WARNING SIGN: U+26A0, U+FE0F.
```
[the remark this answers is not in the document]
westonruter: Perhaps. Corresponding Unicode character are referenced here, but in HTML entity form:

https://github.com/ampproject/amp-wp/blob/bf0a0a0d8394facd57193f8e0ec08bf3483b7f07/includes/validation/class-amp-validation-manager.php#L425-L435

includes/cli/class-amp-cli-validation-command.php
schlessera: I've improved the command to make use of the standard WP-CLI data offerings. This now includes JSON, YAML and table formatting, counting, exit codes. But also filtering on a given field and selecting which fields. For table output, I improved the formatting so it is useful as the human-readable format.

Here's an example of table output:

Image 2019-12-14 at 2 25 21 AM

To filter, you can do something like this:
```
wp amp validation check-url example.org --type=css_error
```

What do you think about that format, it seems more useful for human consumption than JSON, doesn't it?

Should I commit this into this current PR, or create a separate one?

westonruter: A problem with that is the fields vary by error type. Maybe you accounted for this. Doing this in a separate PR makes sense to me.

The table does look nice. But at the same time, the admin screen has this information already presented in an HTML table. I think that perhaps this command should in addition to obtaining the validation results, also store them and then provide the URL to the validated URL screen via a message to `STDERR`.

includes/cli/class-amp-cli-validation-command.php
```diff
@@ -314,6 +314,77 @@ public function reset( $args, $assoc_args ) {
...
+			}
+			$url = $origin . '/' . ltrim( $url, '/' );
+		}
+
+		$result = AMP_Validation_Manager::validate_url( $url );
+		if ( $result instanceof WP_Error ) {
+			WP_CLI::error( $result );
+		}
+
+		WP_CLI::line( wp_json_encode( AMP_Validation_Manager::validate_url( $url ), JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES ) );
```
schlessera: The validation is run twice.
```suggestion
		WP_CLI::line( wp_json_encode( $result, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES ) );
```

schlessera:
If I'm not mistaken, this modifies the response that the loopback request returns (server => server). Therefore, this cannot easily be verified from the staging site during QA, as the Network tab in the developer tools will only show requests initiated on the client.

@westonruter Thoughts on whether this needs QA and how we can do it? Otherwise I'd say the automated tests in place are enough to assert that the response has changed.

westonruter:
Agreed.

PR #3909, opened 2019-12-10
Add group dragging and rotating.

miina:
## Summary

See #3820 

## Checklist

- [x] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [ ] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [ ] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

PR #3910, opened 2019-12-10
New story state management

### LINKED ISSUE #3893: Rewrite story provider to internally use a reducer over states
## Feature description

Rewrite the story provider to internally use a reducer over states.

This is motivated by a number of things, but most importantly because it's best practice.

Motivated partially by these comment threads: https://github.com/ampproject/amp-wp/pull/3884/files/6fcc6941dfed5121438e3295c5f9db5128ec5a5e#r354935318, https://github.com/ampproject/amp-wp/pull/3871#discussion_r352911357, https://github.com/ampproject/amp-wp/pull/3837#discussion_r352952939, https://github.com/ampproject/amp-wp/issues/3827#issuecomment-560932488.

An additional question is whether it's good to keep history and data model reducers separate.

## Acceptance criteria

* Story provider uses a reducer internally.

## Implementation brief

* PoC: #3910

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

barklund:
## Summary

This PR creates a new reducer based story state management with all possible state interactions necessary based on the UX and priority list for P1.

This is in full use across the application.

Thorough unit testing is included.

Explicitly excluded from this:
* Multi-page or multi-element reordering (as it's not P1).

Fixes #3893 

## Checklist


assets/src/edit-story/app/story/useStoryReducer/actions.js
```diff
@@ -0,0 +1,119 @@
...
+const updateElementsById = ( dispatch ) => ( { elementIds, properties } ) =>
+	dispatch( { type: types.UPDATE_ELEMENTS, payload: { elementIds, properties } } );
+
+const updateElementById = ( dispatch ) => ( { elementId, properties } ) =>
+	dispatch( { type: types.UPDATE_ELEMENTS, payload: { elementIds: [ elementId ], properties } } );
+
+const updateSelectedElements = ( dispatch ) => ( { properties } ) =>
+	dispatch( { type: types.UPDATE_ELEMENTS, payload: { elementIds: null, properties } } );
+
+const moveElement = ( dispatch ) => ( { elementId, position } ) =>
```
dvoytenko: What does "moving" of element mean? Reordering in layer tree? Doesn't seem to be related to moving between pages here.

assets/src/edit-story/app/story/useStoryReducer/actions.js
```diff
@@ -0,0 +1,119 @@
...
+	dispatch( { type: types.MOVE_ELEMENT, payload: { pageIndex: null, elementId, position } } );
+
+const moveSelectedElement = ( dispatch ) => ( { position } ) =>
```
dvoytenko: Should this be plural?

assets/src/edit-story/app/story/useStoryReducer/actions.js
```diff
@@ -0,0 +1,119 @@
...
+	dispatch( { type: types.MOVE_ELEMENT, payload: { pageIndex: null, elementId: null, position } } );
+
+const moveElementOnPage = ( dispatch ) => ( { pageIndex, elementId, position } ) =>
```
dvoytenko: Move _to_ page?

assets/src/edit-story/app/story/useStoryReducer/reducers/addElements.js
```diff
@@ -0,0 +1,62 @@
...
+ * Current page is unchanged.
+ *
+ * @param {Object} state Current state
+ * @param {Object} payload Action payload
+ * @param {number} payload.pageIndex Page to insert elements on or null if current page should be used
+ * @param {Array.<Object>} payload.elements Elements to insert on the given page.
+ * @return {Object} New state
+ */
+function addElements( state, { pageIndex, elements } ) {
+	const indexToAddTo = pageIndex === null ? state.current : pageIndex;
```
dvoytenko: Nit, but maybe `state.currentPageIndex`?

assets/src/edit-story/app/story/useStoryReducer/reducers/addElements.js
```diff
@@ -0,0 +1,62 @@
...
+	const isInsertingOnCurrentPage = indexToAddTo === state.current;
+
+	const isWithinBounds = isInsideRange( indexToAddTo, 0, state.pages.length - 1 );
+	if ( isWithinBounds || ! Array.isArray( elements ) ) {
```
dvoytenko: Might be missing something, but shouldn't it be `!isWithinBounds`?

assets/src/edit-story/app/story/useStoryReducer/reducers/addPage.js
```diff
@@ -0,0 +1,26 @@
...
+ * Current page will be updated to point to the newly inserted page.
+ *
+ * Selection is cleared.
+ *
+ * @param {Object} state Current state
+ * @param {Object} payload Action payload
+ * @param {Object} payload.page Object with properties of new page
+ * @return {Object} New state
+ */
+function addPage( state, { page } ) {
```
dvoytenko: Nit, but we will likely have `addPage` at `position` (or after `position`/`page`) in short order. Ok to leave until later.

assets/src/edit-story/app/story/useStoryReducer/reducers/duplicateElements.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Duplicate the given list of elements.
+ *
+ * TODO: Describe and implement if we even need this. Do we?
```
dvoytenko: IMHO this could just be an `addElements` with manual copy.

assets/src/edit-story/app/story/useStoryReducer/actions.js
```diff
@@ -0,0 +1,119 @@
+/**
+ * Internal dependencies
+ */
+import * as types from './types';
+
+const addPage = ( dispatch ) => ( { properties } ) =>
+	dispatch( { type: types.ADD_PAGE, payload: { properties } } );
+
+const deletePage = ( dispatch ) => ( { pageIndex } ) =>
```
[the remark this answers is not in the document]
barklund: I considered if I should switch to ID's half-way through (because pages have ID's too), but was too lazy. Damn, you caught me! Will update to ID's for everything related to page pointers.

assets/src/edit-story/app/story/useStoryReducer/actions.js
barklund: Moving an element in this context is re-layering the element only. The name isn't great, I can see that (@miina asked about it too elsewhere). Is `reorderElement` better? Or maybe `arrangeElement`? (and similar for pages)

assets/src/edit-story/app/story/useStoryReducer/actions.js
barklund: As stated i[n the todo for the specific underlying operation](https://github.com/ampproject/amp-wp/pull/3910/files#diff-a2eff1a69cc4279b3c8fcd9c88338b6bR28), moving/arranging elements currently only handles single element reordering.

@samitron7 is checking whether [multi-element (and multi-page) reordering is P1](https://xwp.slack.com/archives/CCCAUAH6F/p1576039887124500) (internal Slack link, sorry guests).

assets/src/edit-story/app/story/useStoryReducer/actions.js
barklund: Nope, it's if you want to reorder an element on a specific page which is *not* the current page. I don't think it'll ever be used, but is included for consistency (as the underlying operation allows it).

There's a bunch of element operations that allow manipulations to elements on other pages than the current page. Maybe all of that won't ever be needed and we can delete all of that, but I wasn't sure. I'd rather have the underlying API support it, but the exposed API limiting it if necessary.

assets/src/edit-story/app/story/useStoryReducer/reducers/addElements.js
barklund: Well spotted! You win a donut of your own choice! 😄

assets/src/edit-story/app/story/useStoryReducer/reducers/addElements.js
barklund: I considered it, but it's an internal variable only, so I went with only the three variables [`pages`, `current` and `selection`](https://github.com/ampproject/amp-wp/pull/3910/files#diff-018c60aaf786b41efe49a8ba88150c6fR12-R16).

They will never be used outside the scope of the reducer, so I went for brevity over specificity here.

I can revert though to more explicit variables, `currentPageId` and `selectedElementIds`.

assets/src/edit-story/app/story/useStoryReducer/reducers/addPage.js
barklund: Let's push back to @samitron7. I'll ask in the Figma.

"Does adding a page add the new page after the current page or only to the end of the list of pages?"

barklund: @samitron7 confirms, that pressing "add page" adds a blank page *directly after the current page*. So maybe this underlying function should do exactly that? Or should it be `addPageAtPosition`?

assets/src/edit-story/app/story/useStoryReducer/actions.js
```diff
@@ -0,0 +1,119 @@
...
+const addElementsToPage = ( dispatch ) => ( { pageIndex, elements } ) =>
+	dispatch( { type: types.ADD_ELEMENTS, payload: { pageIndex, elements } } );
+
+const addElementToPage = ( dispatch ) => ( { pageIndex, element } ) =>
+	dispatch( { type: types.ADD_ELEMENTS, payload: { pageIndex, elements: [ element ] } } );
+
+const deleteElementsById = ( dispatch ) => ( { elementIds } ) =>
+	dispatch( { type: types.DELETE_ELEMENTS, payload: { elementIds } } );
+
+const deleteSelectedElements = ( dispatch ) => () =>
```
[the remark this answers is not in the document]
barklund: I always wanted selection to be part of this, but not "editing element" (that's left for the canvas only). Selection is directly influenced by actions in the story state, e.g. adding a new page clears selection, adding a new element makes that element the selection, etc.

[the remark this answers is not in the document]
barklund: Also, `selection` here refers specifically to the elements that have focus right now on the current page and is a list of their ids. Not the selected page (current page) nor the focused page, element in edit mode, etc.

barklund:
Several things have come up, that I will address before moving forward with this:

- [ ] Pages have ids and should be referred to be ids, not by index.
- [ ] Adding a page by pressing the "add page" button adds a page right after the current page, not at the end of the page list. Add page should thus take a position for where to add the page (as other situations will require adding pages not necessarily after the current page), just default to it.
- [ ] Duplicating a page isn't an internal function of the API, as it requires the new page to have a new id. Thus duplicating a page will be a combination of element cloning logic (which must generate a new id) and simply adding a page at a specified position.
- [ ] Element manipulation only ever happens on the current page. All functions manipulating elements can safely assume to only be operating on the current page, never any other page nor all pages (as they do now).
- [ ] A page can have zero or one background element. A background element will always be the rear-most in the element stack (at position 0). If another element is made background on a page with an existing background element, the old background element will become element 1 in the stack. These invariants will be kept by the system.

I will address these items in subsequent commits.

barklund:
Also:

- [ ] Rename `movePage` and `moveElement` to `arrangePage` and `arrangeElement` to make it clear we're talking about changing ordering of pages/elements, not moving things around on page or similar.

assets/src/edit-story/app/story/useStoryReducer/actions.js
dvoytenko: I like "reorder".

barklund: I just renamed them to `arrange` because it sounds nicer.

Looking at Keynote and Google Slides, `arrange` is used, but normally includes more than just reordering – also alignment, rotation, grouping, locking, etc.

Keynote | Google Slides
--- | ---
Screen Shot 2019-12-11 at 6 16 30 PM | Screen Shot 2019-12-11 at 6 17 05 PM

What does the people say? @spacedmonkey, @miina, @swissspidy, @pbakaus?

Should we name a function that moves a page's position in the page order as (the more correct but less elegant) `reorderPage` or (the cleaner but less specific) `arrangePage`?

barklund:
> A page can have zero or one background element. A background element will always be the rear-most in the element stack (at position 0). If another element is made background on a page with an existing background element, the old background element will become element 1 in the stack. These invariants will be kept by the system.

I have several ideas about how to do this, but not sure which is best:

1. Expose an API as `setBackgroundElement( elementId )` and use it to set a "private" variable on the page, `backgroundElementId`, and then keep this element as element 0 until set to `null` or another element. Don't allow pages to update this property directly (just like page updates should disallow updating `id` and `elements` actually).
1. Check if a page update includes a `backgroundElementId` attribute and if so, make sure to reorder elements and keep at 0 until set to `null` or another element.
1. Check if an element is updated with `isBackground: true` attribute and if so, reorder elements and make sure it remains bottom in all future reorderings. Any other element will be marked with `isBackground: false`. If multiple elements are updated with the same attribute, only respect the first one. If the current element is later updated with `isBackground: false`, background element status is disabled.
1. Don't do anything at all, but require the API consumer to user reorder functions correctly.

All of them involve using magic attributes on either elements or pages, but so are the attributes `id` and `elements` already.

Thinking more about it, I personally prefer method 1.

assets/src/edit-story/app/story/useStoryReducer/actions.js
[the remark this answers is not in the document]
dvoytenko: Sounds good. My only (smaller) concerns are:

1. There are two selection operations: toggle and set.
2. There could be some "addNew" or "duplicate" cases where the selection stays on the previous element.

I don't think we'll run into these right away, so it's probably ok for now. But I'm slightly concern that we will have a bunch of complicated "selectionOperation" flags, such as "add", "toggle", "set", "auto", "none".

assets/src/edit-story/app/story/useStoryReducer/actions.js
dvoytenko: I think the action should be for plural and the current UI implementation only happens to be executing it as singular. Otherwise, we don't know whether the selection is currently single or multi.

assets/src/edit-story/app/story/useStoryReducer/actions.js
dvoytenko: I see. So you are basically not going to do anything if the `pageIndex` page does not have `elementId`. Right?

assets/src/edit-story/app/story/useStoryReducer/reducers/addElements.js
dvoytenko: sg

dvoytenko: sg to the short names, i mean :)

assets/src/edit-story/app/story/useStoryReducer/reducers/addElements.js
dvoytenko: I'll take kale ones :)

assets/src/edit-story/app/story/useStoryReducer/reducers/addPage.js
dvoytenko: imho, just having `{page, position}` as payload would be good enough?

dvoytenko:
> Thinking more about it, I personally prefer method 1.

I think 1 or 2 are mostly similar. The reason I'd go for 1 is because, I believe, when a new background element is set, the old one has to be removed completely.

assets/src/edit-story/app/story/useStoryReducer/actions.js
swissspidy: No strong opinion here. I think `arrange` is OK

assets/src/edit-story/app/story/useStoryReducer/actions.js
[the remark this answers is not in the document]
barklund: > 1. There are two selection operations: toggle and set.

If you click an element (on page or in layer panel), it and only it should be selected. If you lasso-select a bunch of elements on page, only those exact elements should be selected. If you shift-click from one element to another in the layer panel, those two elements and all elements in between should be selected. All these ways of selecting will use `setSelectedElements`.

If you shift-click an element, it will be toggled in the current selection and doesn't need to know which other elements are currently selected. This will use `toggleElementInSelection`.

Maybe `add` and `remove` will never be used, now that I think of it. If not, let's definitely remove them. They're very simple operations however.

> 2. There could be some "addNew" or "duplicate" cases where the selection stays on the previous element.

I've tried to think hard about what the selection state should be after all operations as specified in the jsdoc within each reducer - see for example [`addPage`](https://github.com/ampproject/amp-wp/pull/3910/files#diff-2140a3febf9327cf423699f615121608R13) or [`deleteElements`](https://github.com/ampproject/amp-wp/pull/3910/files#diff-8952b07de8f790551a5a26bed1b97b4fR16-R17). If you have specific concerns, please mention where appropriate.

assets/src/edit-story/app/story/useStoryReducer/actions.js
barklund: Makes sense, I'll note it and get to it tomorrow.

assets/src/edit-story/app/story/useStoryReducer/actions.js
barklund: [Yes](https://github.com/ampproject/amp-wp/pull/3910/files#diff-08b4230faff7f35682a0f9f8b71366e7R12). Well, it's change to `arrangeElement` and takes a `pageId` now, but [ignores an unknown element id](https://github.com/ampproject/amp-wp/pull/3910/files#diff-08b4230faff7f35682a0f9f8b71366e7R12).

barklund:
> when a new background element is set, the old one has to be removed completely.

Oh? I'll check with @samitron7 in the morning.

assets/src/edit-story/app/story/useStoryReducer/reducers/addPage.js
barklund: Yes, there's now [`addPage`](https://github.com/ampproject/amp-wp/pull/3910/files#diff-30af242e1170b1c57b72a85d7d353543R6-R7) assuming _"add right after current page"_ and [`addPageAt`](https://github.com/ampproject/amp-wp/pull/3910/files#diff-30af242e1170b1c57b72a85d7d353543R9-R10) taking a position.

Also see [the underlying implementation](https://github.com/ampproject/amp-wp/pull/3910/files#diff-2140a3febf9327cf423699f615121608R21-R37).

assets/src/edit-story/app/story/useStoryReducer/actions.js
miina: What about something like `movePageToPosition`? If `arrange` seems clear to most then that's OK with me, too.

assets/src/edit-story/app/story/useStoryReducer/reducers/addElements.js
```diff
@@ -0,0 +1,43 @@
...
+ * @return {Object} New state
+ */
+function addElements( state, { elements } ) {
+	const pageIndex = state.pages.find( ( { id } ) => id === state.current );
+	const oldPage = state.pages[ pageIndex ];
+	const newPage = {
+		...oldPage,
+		elements: [
+			...oldPage.elements,
+			...elements,
```
miina: Would elements always be added to the end, e.g. in case of duplication as well? Perhaps that's a UX question, just wondering if there should be an optional index. Probably not important until we actually need it.

barklund: I considered it but didn't see a clear use-case for it. Maybe if the user cut-pastes in the layer panel, but we might want to just use `arrange` for that.

assets/src/edit-story/app/story/useStoryReducer/actions.js
barklund: It would make sense to have the element reorder function named the same at the page one, and `moveElementToPosition` is pretty ambiguous.

assets/src/edit-story/app/story/useStoryReducer/actions.js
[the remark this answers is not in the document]
dvoytenko: Ok. I think we won't figure it out fully in this PR. Let's go ahead and complete/merge and we'll be following up while implementing all of these features.

barklund:
Two topics have come up (in https://github.com/ampproject/amp-wp/pull/3871#discussion_r352910490 and https://github.com/ampproject/amp-wp/pull/3911#discussion_r357451424 respectively) for possible inclusion here.

### Story meta data
Do we need a story state variable for story meta information such as title, post status, author, tags, categories, etc?

As I wrote [elsewhere](https://github.com/ampproject/amp-wp/pull/3871#issuecomment-565461091):

> My recommendation would be:
>
> - Add a new state variable meta to the story reducer state.
> - Add a new API to update meta variables.
> - Also store the meta variable array in history.
> - When loading a story, initialising meta from a set of fixed properties on the loaded data object
> - When saving a story, copying a fixed list of properties from the meta object to the request object.

I think this is a necessary extension of the story state management and will go ahead with this unless I hear loud objections.

### Multi-selecting background elements
Should we completely disallow background elements from being part of multi-selections?

As I wrote [elsewhere](https://github.com/ampproject/amp-wp/pull/3911#discussion_r357679239):

> We can block this on the underlying data management level, e.g. as the following invariant:
> 
> - The selection object can be either a list of *only* the background element id or a list of element ids of which none are equal to the background element id.
>
> It will be enforced as:
>
> - If you are setting the selection list, the background element must either be the only element or will be filtered out of the list.
> - If you are adding the background element to the selection list, the previous selection must be empty or nothing happens.
> - If you are adding a non-background element to the selection list, and the previous selection was the background element, the new element will be the only selected element.

On this one I'm a bit more hesitant. It's not extremely complicated, but does muddy the API a bit. Feel free to chime in.

pbakaus:
> Should we completely disallow background elements from being part of multi-selections?

Yes. Left the same comment on Dima's recent PR. Background can only be selected on its own.

spacedmonkey:
I think it makes senses to have a `meta` object to track all of the WP properties of a story. I would recommend through, keeping this as just WP fields on the `WP_Post` object. Fields that are important to us in WP_Post. 
- ID
- Title
- Post status
- Post content ( rendered AMP tags )
- Post content filtered ( JSON object of data )
- Published date 
- Updated date. 
- Author (int of WP_user id ).
- Post name ( slug ). 

Taxonomies like categories and tags are handled a little differently and not part `WP_Post` object. So we would end up with 4 state objects. 

- Post ( WP_Post object fields ) 
- Taxonomies ( Object of linked tags / categories )
- Pages - array of pages and elements, as currently
- HIstory

I also don't like the naming of meta. WordPress has meta data and naming it our code is confusing as it is not meta data, is the primary data on a `WP_Post` object. We may later need to have a meta object if we have extra fields that do not map to core fields in `WP_Post`.

miina:
@barklund @pbakaus 
> Should we completely disallow background elements from being part of multi-selections?

Would this be needed because of the Moveable functionality (dragging, resizing, etc.) or for other reasons as well? For example, shouldn't the user be able to multi-select some elements + the background and delete these? Or copy or cut?

pbakaus:
> Would this be needed because of the Moveable functionality (dragging, resizing, etc.) or for other reasons as well? For example, shouldn't the user be able to multi-select some elements + the background and delete these? Or copy or cut?

@miina I could be convinced that this is needed, but the behaviour would be super inconsistent: It would mean that you cannot drag a multi-selection with a lasso that includes the background, but that you can shift+click yourself into a multi-selection that includes it. When in doubt, I usually aim for consistency.

barklund:
> I think it makes senses to have a `meta` object to track all of the WP properties of a story. I would recommend through, keeping this as just WP fields on the `WP_Post` object. Fields that are important to us in WP_Post.

Come to think of it, it's about more than just WP post data. Any story setting that's not related to a specific page would have to be saved on a more general "story" level.

One thing is the featured image, which is an required property for any AMP story. We need that pretty soon (how to edit it in the editor is another question).

Another thing, background audio can be set globally for the entire story. This could go in the data for the first page, but that wouldn't be proper,

We will most likely have other story-global settings, that need to be stored here as well.

> Taxonomies like categories and tags are handled a little differently and not part `WP_Post` object. So we would end up with 4 state objects.

This only matters when saving, not when manipulating the data internally. I don't think there's a need for storing it differently - I'd just keep it as a list of id's on the meta data object named `categories` and `tags` respectively. When loading and saving, we'd need to unpack this correctly.

> I also don't like the naming of meta. WordPress has meta data and naming it our code is confusing as it is not meta data, is the primary data on a `WP_Post` object. We may later need to have a meta object if we have extra fields that do not map to core fields in `WP_Post`.

Great point about `meta`. However, `post` is also a weird name, as this is a story, not a post (seen without WP glasses). Maybe `info` or simply `story` could be a reasonable name?

I suggest we create a `story` variable for all meta data. And when saving, we'd then need to unpack some of the values in here and name them correctly for saving (and similar when loading), and the rest would be sent to the API as the `story` variable (to go alongside `pages`).

We'd then have things like:

```
story.post_id = 'string'
story.title = 'string'
story.slug = 'string'
story.categories = [1,2,3]
story.featuredImage = { src: '...', .... }
story.backgroundAudio = { src: '...', volume: 60, .... }
```

When saving, we'd unpack it like:

```js
const { story, pages } = state;
const {
  id,
  title,
  slug,
  authorId,
  categoryIds,
  tagIds,
  ...storyData,
} = story;
const wpPostData = {
  post_id: id,
  title,
  slug,
  author: authorId,
  post_content: generateAMPHTML( story, pages ),
  post_content_filtered: { pages, story: storyData },
};
const wpTaxonomies = { categories: categoryIds, tags: tagIds };
// send request!
fetch( api, {
  body: {
    post: wpPostData,
    taxonomies: wpTaxonomies,
  } );
```

And in reverse when loading. It does create a bit of extra work when getting things in and out. However when working with the data inside the editor, all global story meta data goes in the same place, regardless of where it's coming from or going to.

Actual WP meta variables would go to the same place, just packed and unpacked differently in a third way.

We can of course create dictionaries of which variables to move where:

```js
const POST_DATA = {
  post_id: 'id',
  ...
  post_content: ( { story, pages } ) => generateAMPHTML( story, pages ),
  post_content_filtered: ( { story, pages } ) => ( { pages, story: storyData } ),
};
const TAXONOMY_DATA = { categories: 'categoryIds', ...  };
const META_DATA = { ...  };
```

And then use this to unpack properly. It's still fairly neat IMHO. However, this is also slightly outside the scope of this ticket which is just about the state management.

miina:
Having one variable for the general story data makes sense to me, had that thought by default -- one state variable that worries about the settings per pages and the other for the whole story general settings.

I don't think that the App should always be aware of/depend on how the data is saved to the DB / handled in WP -- a layer mapping between the App and WP would take care of that. This would make it more friendly for platforms other than WP and also seems to be a more logical way to store data (when not thinking about WP).

assets/src/edit-story/app/story/useStoryReducer/reducers/deletePage.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * Delete page by id or delete current page if no id given.
+ *
+ * If another page than current page is deleted, it will remain current page.
+ *
+ * If the current page is deleted, the next page will become current.
+ * If no next page, previous page will become current.
+ *
+ * If state only has one or zero pages, nothing happens.
```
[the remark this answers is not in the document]
barklund: It's a UX/Product decision, and the decision is currently to not allow to delete a page, if it's the only one.

assets/src/edit-story/app/story/useStoryReducer/reducers/utils.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * Internal dependencies
+ */
+import { LAYER_DIRECTIONS } from '../../../../constants';
+
+export function intersect( first, ...rest ) {
```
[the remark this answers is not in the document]
barklund: It's arguably one of the longer lodash functions to reimplement with native JS functionality, but I'll add some unit tests to validate its correctness.

assets/src/edit-story/app/story/useStoryReducer/reducers/utils.js
```diff
@@ -0,0 +1,76 @@
...
+			return Math.min( maxPosition, currentPosition + 1 );
+		case LAYER_DIRECTIONS.BACKWARD:
+			return Math.max( minPosition, currentPosition - 1 );
+		default:
+			return currentPosition;
+	}
+}
+
+export function objectWithout( obj, propertiesToRemove ) {
+	return Object.keys( obj )
```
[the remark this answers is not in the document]
barklund: It's pretty trivial to see how it works with the implementation directly in the codebase.

assets/src/edit-story/app/story/useStoryReducer/reducers/updateElements.js
```diff
@@ -0,0 +1,70 @@
...
+	const idsToUpdate = elementIds === null ? state.selection : elementIds;
+
+	if ( idsToUpdate.length === 0 ) {
+		return state;
+	}
+
+	const pageIndex = state.pages.findIndex( ( { id } ) => id === state.current );
+
+	const oldPage = state.pages[ pageIndex ];
+	const pageElementIds = oldPage.elements.map( ( { id } ) => id );
```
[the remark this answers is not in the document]
barklund: This is more or less exactly what `Array.prototype.map` was made for.

assets/src/edit-story/app/story/useStoryReducer/reducers/selectElement.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * If no id is given or id is already in the current selection, nothing happens.
+ *
+ * @param {Object} state Current state
+ * @param {Object} payload Action payload
+ * @param {string} payload.elementId Element id to add to the current selection.
+ * @return {Object} New state
+ */
+function selectElement( state, { elementId } ) {
+	if ( ! elementId || state.selection.includes( elementId ) ) {
```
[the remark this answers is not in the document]
barklund: It's an array. Checking if it contains something using a native array method. I don't see any way to make it clearer than that using a helper method? If `selection` somehow changed to a different data type, a lot of the API would be impacted.

spacedmonkey:
> I don't think that the App should always be aware of/depend on how the data is saved to the DB / handled in WP -- a layer mapping between the App and WP would take care of that.

I agree with this, I think I am thinking to much about WP here. 

I do however, believe that taxonomies should treated differently. We should make these taxonomies optional, as other platforms we integrate into may not have this functionality. Unlike publish date, title, content and author, that are pretty generic. 

I think we should go with the naming of `story` for the variable, `post`, `info` and `meta` all these seem confusing. We can also add properties on these variables later. Some good examples of global data. 
- SEO fields, like title, description. 
- Animation settings. 
- Background audio for all pages. 
- Popular elements or fonts.

miina:
> I do however, believe that taxonomies should treated differently. We should make these taxonomies optional, as other platforms we integrate into may not have this functionality.

It's possible that a platform doesn't have categories implemented, in this case, the field could be just unused with an empty default value (`[]`). Definitely agreed that it should work without categories being available, too.

@barklund Looks like there's a consensus on going with the `story` variable at least, perhaps we could go with that for now for the data we're already using.

Q: How much of the rest of the discussion is a blocker for this state management PR right now?

barklund:
I will implement both of the above improvements, a `story` variable for setting arbitrary story-global settings (with no validation of keys or values) and a restriction on multi-select to not include background element.

- [ ] New story-global state variables.
- [ ] Restrict multi-select to not include background element.

I will amend tests as well.

After that, I will actually implement the usage of this new state throughout the application. I don't think this will be done before X-mas, but I'll work a bit during the holidays to get it ready for final review just after.

barklund:
@miina, could I get you to see if you have a simple fix for the weird moveable behaviour in this one? Maybe you have experienced the same thing.

@swissspidy, @pbakaus, @dvoytenko and @spacedmonkey, PTAL and review the source code now it has been rewritten and taken into use throughout the application.

barklund:
> After that, I will actually implement the usage of this new state throughout the application. I don't think this will be done before X-mas, but I'll work a bit during the holidays to get it ready for final review just after.

And done. It was a lot easier than expected.

assets/src/edit-story/app/story/useStoryReducer/reducers/setBackgroundElement.js
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * If page had a background element before, that element is deleted!
+ * And if that element was selected, selection is cleared.
+ *
+ * @param {Object} state Current state
+ * @param {Object} payload Action payload
+ * @param {number} payload.elementId Element id to set as background on the current page.
+ * @return {Object} New state
+ */
+function setBackgroundElement( state, { elementId } ) {
```
dvoytenko: The typical pattern so far was to pass `pageId` that could be `null` to default to `state.current`. Should we do this here as well?

miina:
> @miina, could I get you to see if you have a simple fix for the weird moveable behaviour in this one? Maybe you have experienced the same thing.

@barklund Yes, there's even a PR for it and you are tagged there as well for feedback ;)
https://github.com/ampproject/amp-wp/pull/3959

miina:
@barklund Testing the branch and seeing an additional (related to the previous) issue with Moveble right now. Would be good to get #3959 in first to see if the issue here comes from this PR.

For example, in this PR the issue happens with Text element as well (not only the Image), also, when one element is selected, then clicking another element causes multi-selecting instead of selecting the new element. These things would be good to fix before merging.

assets/src/edit-story/app/story/storyProvider.js
```diff
@@ -6,98 +6,111 @@ import PropTypes from 'prop-types';
...
+	useHistoryEntry( { pages, current, selection, story } );
+	useHistoryReplay( { restore } );
 
-	useLoadStory( { storyId, pages, setPages, setTitle, setPostStatus, setPostAuthor, setSlug, setLink, setCurrentPageIndex, clearSelection } );
-	useCurrentPage( { currentPageIndex, pages, setCurrentPage, setCurrentPageNumber } );
-	useHistoryEntry( { currentPageIndex, pages, selectedElementIds } );
-	useHistoryReplay( { setCurrentPageIndex, setPages, setSelectedElementIds } );
-	useSelectedElements( { currentPageIndex, pages, selectedElementIds, setSelectedElements } );
+	// This action allows the user to save the story
+	// (and it will have side-effects, thus the need for `updateStory`)
```
miina: Q: What do you mean by side-effects here?

assets/src/edit-story/app/story/storyProvider.js
```diff
@@ -6,98 +6,111 @@ import PropTypes from 'prop-types';
...
+	const { updateStory } = api;
+	const saveStory = useSaveStory( { storyId, pages, story, updateStory } );
 
 	const state = {
 		state: {
 			pages,
+			currentPageId,
 			currentPageIndex,
 			currentPageNumber,
 			currentPage,
```
miina: ^ Any specific reason to not have just one state variable for `currentPage` with params for `id`, `index`, etc.? (Not thinking it would be better otherwise, just curious)

miina: I suppose these are needed separately most of the time instead of using/updating all the `currentPage` information?

assets/src/edit-story/app/story/useStoryReducer/reducers/arrangePage.js
```diff
@@ -0,0 +1,40 @@
...
+ * TODO: Handle multi-page re-order when UX and priority is finalized.
+ *
+ * @param {Object} state Current state
+ * @param {Object} payload Action payload
+ * @param {number} payload.pageId Id of page to move to a new position.
+ * @param {number} payload.position Index of where page should be moved to.
+ * @return {Object} New state
+ */
+function arrangePage( state, { pageId, position } ) {
+	const pageIndex = state.pages.findIndex( ( { id } ) => id === pageId );
```
miina: Is there's only one Page, should we just return state?

assets/src/edit-story/app/story/useStoryReducer/reducers/setBackgroundElement.js
miina: Could a background element ever be set to a non-current Page? In this case, would it ever be possible to `addElements` to the non-current page, too?

assets/src/edit-story/app/story/useStoryReducer/reducers/setSelectedElements.js
```diff
@@ -0,0 +1,54 @@
...
+		const commonElements = intersect( state.selection, uniqueElementIds );
+		if ( commonElements.length === state.selection.length ) {
+			return state;
+		}
+	}
+
+	// If it's a multi-selection, filter out the background element
+	const currentPage = state.pages.find( ( { id } ) => id === state.current );
+	const isNotBackgroundElement = ( id ) => currentPage.backgroundElementId !== id;
+	const newSelection = uniqueElementIds.length > 1 ?
```
miina: What if the background element was selected first and then a second element is clicked? Should then only the background element stay?

assets/src/edit-story/app/story/useStoryReducer/reducers/utils.js
```diff
@@ -0,0 +1,74 @@
...
+
+	return rest.reduce(
+		( intersection, list ) =>
+			intersection.filter( ( value ) => list.includes( value ) ),
+		first,
+	);
+}
+
+export function isInsideRange( index, start, end ) {
+	return index >= start && index <= end;
```
miina: Should this ensure that `index` is numeric/a number?

assets/src/edit-story/app/story/useStoryReducer/reducers/test/utils.js
```diff
@@ -0,0 +1,180 @@
...
+		expect( isSingleDigit( 10 ) ).toStrictEqual( false );
+
+		expect( isSingleDigit( Number.POSITIVE_INFINITY ) ).toStrictEqual( false );
+		expect( isSingleDigit( Number.NEGATIVE_INFINITY ) ).toStrictEqual( false );
+		expect( isSingleDigit( Number.NaN ) ).toStrictEqual( false );
+	} );
+} );
+
+describe( 'moveArrayElement', () => {
+	it( 'should move element forwards', () => {
```
miina: Could we add a test for trying moving the element outside of limits?

assets/src/edit-story/app/story/useStoryReducer/test/addElement.js
```diff
@@ -0,0 +1,43 @@
...
+		const result = addElement( { element: { id: '123' } } );
+
+		expect( result.pages[ 0 ] ).toStrictEqual( {
+			id: '111',
+			elements: [ { id: '123' } ],
+		} );
+		expect( result.selection ).toStrictEqual( [ '123' ] );
+	} );
+
+	it( 'should add an element to the end of the list on the current page and replace selection', () => {
```
miina: Could we add a test for adding 2 elements with the same ID?

assets/src/edit-story/app/story/useStoryReducer/test/addElements.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Internal dependencies
+ */
+import { setupReducer } from './_utils';
+
+describe( 'addElements', () => {
+	it( 'should add all element to the current page and select them', () => {
```
miina: Could we again add a test for adding two elements with the same ID?

assets/src/edit-story/app/story/useStoryReducer/test/arrangeElement.js
```diff
@@ -0,0 +1,171 @@
+/**
+ * Internal dependencies
+ */
+import { LAYER_DIRECTIONS } from '../../../../constants';
+import { setupReducer } from './_utils';
+
+describe( 'arrangeElement', () => {
+	it( 'should move element to specified position', () => {
```
miina: Could we add a test with an illegal position?

assets/src/edit-story/app/story/storyProvider.js
barklund: Saving a post in WP can [change its status and link](https://github.com/ampproject/amp-wp/pull/3910/files#diff-5157e5359482c52319bc0f4b51134219R76-R79) and that'll update the state.

assets/src/edit-story/app/story/storyProvider.js
barklund: I wanted `currentPage` to be the exact same object also stored in the `pages` array and at the same time I did not want to update index and number on all pages all the time. The reducer could internally manage these, but I didn't find it important. We could however add that quite easily if we wanted to.

assets/src/edit-story/app/story/useStoryReducer/reducers/arrangePage.js
barklund: That seems like a fair optimization - done in c07622e.

assets/src/edit-story/app/story/useStoryReducer/reducers/setSelectedElements.js
barklund: Yeah, that's the big question. @pbakaus, what do you think?

Currently I've implemented it as follows:

* If the background element is selected and another element is added to the selection (e.g. by meta-clicking it), only the new element is selected, the background element is not selected anymore.
* If a non-background element is selected and the background element is added to the selection (e.g. by meta-clicking it), it is ignored and only the non-background element is in the selection.

assets/src/edit-story/app/story/useStoryReducer/reducers/utils.js
barklund: It'll return `NaN` for most thing, which is the only reasonable return value anyway.

assets/src/edit-story/app/story/useStoryReducer/reducers/test/utils.js
barklund: Good idea - done in d6c410c9.

assets/src/edit-story/app/story/useStoryReducer/test/addElements.js
barklund: So, the thing is. No check is actually made for this and it's not one of the invariants maintained by the system (no duplicate ids allowed).

Should it validate this?

assets/src/edit-story/app/story/useStoryReducer/test/arrangeElement.js
barklund: Done in 757a64a5.

assets/src/edit-story/app/story/useStoryReducer/test/addElements.js
barklund: (I resolved the other similar comment, just FYI)

assets/src/edit-story/app/story/useStoryReducer/reducers/setBackgroundElement.js
barklund: I actually asked @samitron7 about editing elements on non-current pages, and she said it wasn't possible in any way. I thus removed all references to pageId on all element editing functions. I can add them, but I don't think we'll ever need them. You'll see the exact same thing for `deleteElements`, `addElements`, `updateElements` and `arrangeElement`.

assets/src/edit-story/app/story/useStoryReducer/reducers/setSelectedElements.js
pbakaus: sorry for the late reply. The way @barklund has implemented it makes sense to me intuitively. Let's roll with that and see how it feels for now.

assets/src/edit-story/app/story/useStoryReducer/actions.js
```diff
@@ -0,0 +1,120 @@
...
+const deleteElementById = ( dispatch ) => ( { elementId } ) =>
+	dispatch( { type: types.DELETE_ELEMENTS, payload: { elementIds: [ elementId ] } } );
+
+const updateElementsById = ( dispatch ) => ( { elementIds, properties } ) =>
+	dispatch( { type: types.UPDATE_ELEMENTS, payload: { elementIds, properties } } );
+
+const updateElementById = ( dispatch ) => ( { elementId, properties } ) =>
+	dispatch( { type: types.UPDATE_ELEMENTS, payload: { elementIds: [ elementId ], properties } } );
+
+const updateSelectedElements = ( dispatch ) => ( { properties } ) =>
```
dvoytenko: One comment here, which is ok to delay until later. But `updateSelectedElements({properties})` only works for side panel. The transforming movables have more complicated rules for updating the properties of elements post transformation, but they are definitely _not_ shared values. For instance:

* Drag: `elements[i].x += deltaX` and `elements[i].y += deltaY`. I.e. the same `deltaX` and `deltaY` offsets for all elements, but different base `x`/`y`.
* Rotate: `elements[i].rotationAngle = angle` plus `elements[i].x += deltaX[i]` and `elements[i].y += deltaY[i]`. I.e. the same angle value, but different offsets `deltaX` and `deltaY`.
* Resize: likely all `width[i]`, `height[i]`, `deltaX[i]` and `deltaY[i]` are different per each element.

It might be just easier to update the selected elements by ID in such cases. Or we could try an updater function that takes existing properties and recalculates the new ones.

dvoytenko: /cc @miina

assets/src/edit-story/app/story/useStoryReducer/reducers/setBackgroundElement.js
miina: Makes sense to me!

assets/src/edit-story/app/story/storyProvider.js
miina: Could we specify that in the comment about the side-effects?
Also, is this something that should be mentioned in `useSaveStory` as well?

assets/src/edit-story/app/story/useStoryReducer/test/addElements.js
miina: Hmm, okay, sure. Saw that `addElements` is removing duplicates just in case and thought that perhaps testing that could make sense. Would then adding an element twice as a test make more sense?

If you think that's not necessary, that's fine, feel free to ignore :)

assets/src/edit-story/app/story/useStoryReducer/actions.js
miina: Maybe `updateElements` would be appropriate for that case. E.g. something like [here](https://github.com/ampproject/amp-wp/pull/3909/files#diff-ee03834cf840c0325129d4839b2e65f0R117-R127).

assets/src/edit-story/components/header/buttons.js
```diff
@@ -67,14 +67,14 @@ function PreviewButton() {
...
+		state: { isSaving, story: { status } },
+		actions: { saveStory },
 	} = useStory();
 
-	const text = ( postStatus !== 'publish' ) ? __( 'Publish' ) : __( 'Update' );
+	const text = ( status !== 'publish' ) ? __( 'Publish' ) : __( 'Update' );
 
 	return (
-		<Primary onClick={ savePost } isDisabled={ isSaving }>
+		<Primary onClick={ saveStory } isDisabled={ isSaving }>
```
swissspidy: This PR removes `isSaving` state. Can we perhaps keep that to retain functionality (disable publish button while saving)?

assets/src/edit-story/app/story/actions/useSaveStory.js
```diff
@@ -0,0 +1,89 @@
...
+	}, [] );
+
+	const saveStory = useCallback( () => {
+		const { title, status: postStatus, author, slug } = story;
+		const status = ( postStatus !== 'publish' ) ? 'publish' : postStatus;
+
+		const content = getStoryMarkupFromPages( pages );
+		saveStoryById( storyId, title, status, pages, author, slug, content ).then( ( post ) => {
+			const { status: newStatus, link } = post;
+			updateStory( {
```
swissspidy: Looks like this needs to be changed to:

```js
updateStory( {
	properties: {
		status: newStatus,
		link,
	},
} );
```

swissspidy: Fixed in 450dfd4199a95cb82480617145bad7212d98f952

miina:
> >@miina, could I get you to see if you have a simple fix for the weird moveable behaviour in this one? Maybe you have experienced the same thing.

>@barklund Yes, there's even a PR for it and you are tagged there as well for feedback ;)
#3959

@barklund FYI: Looks like #4018 fixes this issue of stickiness, likely it will get merged soon, we'll see then if it fixes the issues in this PR, too, the issues being:
1. Elements "sticking" to mouse even when not dragging
2. When one element is selected, then clicking another element causes multi-selecting instead of selecting the new element.

assets/src/edit-story/app/story/useStoryReducer/test/addElements.js
swissspidy: The tests for `addElementToSelection` reflect that selection should always be unique, and since `addElements` also keeps selection unique, it makes sense to have a test for it IMO.

assets/src/edit-story/app/story/useStoryReducer/reducers/addPage.js
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * If position is outside bounds or no position given, new page will be inserted after current page.
+ *
+ * Current page will be updated to point to the newly inserted page.
+ *
+ * Selection is cleared.
+ *
+ * @param {Object} state Current state
+ * @param {Object} payload Action payload
+ * @param {Object} payload.properties Object with properties of new page
```
swissspidy: Unused property

spacedmonkey: Missing page docs

assets/src/edit-story/app/story/useStoryReducer/actions.js
barklund: That's a really good point, @dvoytenko. It could definitely make sense to add the option of supplying properties as an object or as a function receiving the old props and returning new ones (that'll be merged in) as a result. Do you want to do this now or can we wait until we have multiple uses for this to see what the best interface is?

assets/src/edit-story/app/story/useStoryReducer/test/addElements.js
barklund: I've changed this since the original comment and added an invariant for no duplicates in selection. I'll add a test as suggested.

assets/src/edit-story/components/header/buttons.js
barklund: But where to put it? It doesn't belong inside the story reducer (as it's not part of the story, it's just UX meta). All the way back in the API would probably make the most sense, but then this publish button needs to know, which underlying API method (or even methods) are used for saving a story? It might go together with another variable remembering whether the story is dirty (changed since last save) in some state object on the story, e.g.:
```js
	const {
		state: {
			meta: { isSaving, isDirty },
			story: { status },
		},
		actions: { saveStory },
	} = useStory();
```

And then create a separate reducer for updating these two states when relevant?

assets/src/edit-story/app/story/useStoryReducer/test/addElements.js
barklund: Oh, I hadn't actually added that invariant, but am doing that now - as it does make a hell of a lot of sense to do 😛.

assets/src/edit-story/app/story/useStoryReducer/reducers/addPage.js
barklund: Fixed in 81692ed.

barklund: Fixed in 81692ed.

assets/src/edit-story/app/story/useStoryReducer/test/addElements.js
barklund: Fixed in 81692ed.

assets/src/edit-story/app/story/storyProvider.js
barklund: It's pretty obvious inside the function, as the side-effect is right there in the code, but it'll update the description.

barklund: Fixed in 3d70ba8.

assets/src/edit-story/components/header/buttons.js
swissspidy: Good question. Your suggestion makes sense to me.

Perhaps there's gonna be more meta variables like this in the future too. #4020 comes to mind here.

assets/src/edit-story/app/story/useStoryReducer/actions.js
dvoytenko: Definitely ok to wait. It just means there might be more `updateById` vs `updateSelected` used for now.

assets/src/edit-story/components/header/buttons.js
barklund: Done in bf5d6fb.

barklund: (only the `isSaving` part though - `isDirty` is better dealt with through the history provider, I think)

PR #3911, opened 2019-12-11
PoC: Insertion and editing of full-bleed components

### LINKED ISSUE #3802: Full-bleed Components
## Feature description

How does full-bleed components work? This will be used a lot, so it has to be considered carefully. Should width-height be disabled in the inspector panel? What if you want to zoom in even more on a full-bleed image?

Also, we need to consider how to select the page itself when a full-bleed element is covering the entire canvas. You can click outside the page on the background, but is that sufficiently clear to users?
It's not complex, but it's important to consider this, as it's a day-1 feature.

---------------

## Acceptance criteria

## Implementation brief

* PoC: #3911
* Fullbleed is related to background, but not the same thing. The background feature is in TODO.
* There are three major issues with fullbleed components: "Layout fullbleed component", "Toggle fullbleed mode in the edit panel", "selectable/transformable changes". The PoC address "layout" and "edit panel" cases.
* PoC opts for a separate `isFullbleed` property. The reasons: it's now a truly logical property and we may want to be able to toggle back to the previously available width/height/x/y.
* At least x/y should be hidden/disabled in the edit panel. The PoC currently disables x/y/w/h/r. Will confirm this with the UX.
* PoC introduces `units` context to translate between "story units" and "editor pixels".
* PoC pushes `isFullbleed` editor into a single `FullbleedLayerPanel`.
* This story has strong dependencies on #3816, Display/Edit mode, and decision on whether the wrapper can be used for x/y/w/h/r.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

dvoytenko:
## Summary

Fixes #3802

## Checklist

### DIFF
--- a/assets/src/edit-story/components/canvas/page.js
+++ b/assets/src/edit-story/components/canvas/page.js
@@ -6,7 +6,7 @@ import styled from 'styled-components';
 /**
  * WordPress dependencies
  */
-import { useCallback, useEffect, useState, useRef } from '@wordpress/element';
+import { useCallback, useEffect, useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -78,12 +78,9 @@ function Page() {
 
 			{ selectedElement && targetEl && (
 				<Movable
-					rotationAngle={ selectedElement.rotationAngle }
+					selectedElement={ selectedElement }
 					targetEl={ targetEl }
 					pushEvent={ pushEvent }
-					type={ selectedElement.type }
-					x={ selectedElement.x }
-					y={ selectedElement.y }
 				/>
 			) }
 		</Background>
--- a/assets/src/edit-story/components/movable/index.js
+++ b/assets/src/edit-story/components/movable/index.js
@@ -69,6 +69,8 @@ function Movable( {
 	const resetMoveable = ( target ) => {
 		frame.translate = [ 0, 0 ];
 		setStyle( target );
+		target.style.width = '';
+		target.style.height = '';
 		if ( moveable.current ) {
 			moveable.current.updateRect();
 		}
--- a/assets/src/edit-story/components/movable/index.js
+++ b/assets/src/edit-story/components/movable/index.js
@@ -18,10 +18,7 @@ const CORNER_HANDLES = [ 'nw', 'ne', 'sw', 'se' ];
 const ALL_HANDLES = [ 'n', 's', 'e', 'w', 'nw', 'ne', 'sw', 'se' ];
 
 function Movable( {
-	rotationAngle,
-	x,
-	y,
-	type,
+	selectedElement,
 	targetEl,
 	pushEvent,
 } ) {
@@ -47,9 +44,17 @@ function Movable( {
 		}
 	}, [ targetEl, moveable ] );
 
+	// Update moveable with whatever properties could be updated outside moveable
+	// itself.
+	useEffect( () => {
+		if ( moveable.current ) {
+			moveable.current.updateRect();
+		}
+	} );
+
 	const frame = {
 		translate: [ 0, 0 ],
-		rotate: rotationAngle,
+		rotate: selectedElement.rotationAngle,
 	};
 
 	const setStyle = ( target ) => {
@@ -73,9 +78,9 @@ function Movable( {
 		<Moveable
 			ref={ moveable }
 			target={ targetEl }
-			draggable={ true }
-			resizable={ true }
-			rotatable={ true }
+			draggable={ ! selectedElement.isFullbleed }
+			resizable={ ! selectedElement.isFullbleed }
+			rotatable={ ! selectedElement.isFullbleed }
 			onDrag={ ( { target, beforeTranslate } ) => {
 				frame.translate = beforeTranslate;
 				setStyle( target );
@@ -86,7 +91,7 @@ function Movable( {
 			} }
 			onDragEnd={ ( { target } ) => {
 				// When dragging finishes, set the new properties based on the original + what moved meanwhile.
-				const newProps = { x: x + frame.translate[ 0 ], y: y + frame.translate[ 1 ] };
+				const newProps = { x: selectedElement.x + frame.translate[ 0 ], y: selectedElement.y + frame.translate[ 1 ] };
 				setPropertiesOnSelectedElements( newProps );
 				resetMoveable( target );
 			} }
@@ -106,8 +111,8 @@ function Movable( {
 				setPropertiesOnSelectedElements( {
 					width: parseInt( target.style.width ),
 					height: parseInt( target.style.height ),
-					x: x + frame.translate[ 0 ],
-					y: y + frame.translate[ 1 ],
+					x: selectedElement.x + frame.translate[ 0 ],
+					y: selectedElement.y + frame.translate[ 1 ],
 				} );
 				resetMoveable( target );
 			} }
@@ -123,19 +128,16 @@ function Movable( {
 			} }
 			origin={ false }
 			pinchable={ true }
-			keepRatio={ 'image' === type } // @†odo Even image doesn't always keep ratio, consider moving to element's model.
-			renderDirections={ 'image' === type ? CORNER_HANDLES : ALL_HANDLES }
+			keepRatio={ 'image' === selectedElement.type } // @†odo Even image doesn't always keep ratio, consider moving to element's model.
+			renderDirections={ 'image' === selectedElement.type ? CORNER_HANDLES : ALL_HANDLES }
 		/>
 	);
 }
 
 Movable.propTypes = {
-	rotationAngle: PropTypes.number.isRequired,
+	selectedElement: PropTypes.object,
 	targetEl: PropTypes.object.isRequired,
 	pushEvent: PropTypes.object,
-	type: PropTypes.string.isRequired,
-	x: PropTypes.number.isRequired,
-	y: PropTypes.number.isRequired,
 };
 
 export default Movable;
--- a/assets/src/edit-story/elements/image/display.js
+++ b/assets/src/edit-story/elements/image/display.js
@@ -12,10 +12,9 @@ import { useCallback } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import { PAGE_WIDTH, PAGE_HEIGHT } from '../../constants';
 import { useCanvas } from '../../components/canvas';
 import useDoubleClick from '../../utils/useDoubleClick';
-import { ElementWithPosition, ElementWithSize, ElementWithRotation } from '../shared';
+import { ElementWithPosition, ElementWithSize, ElementWithRotation, getBox } from '../shared';
 import { getImgProps, ImageWithScale } from './util';
 
 const Element = styled.div`
@@ -26,24 +25,17 @@ const Element = styled.div`
 `;
 
 const Img = styled.img`
-	position: relative;
+	position: absolute;
 	${ ImageWithScale }
 `;
 
-function ImageDisplay( { id, src, origRatio, width, height, x, y, scale, offsetX, offsetY, rotationAngle, forwardedRef, onPointerDown } ) {
-	// Width and height are percent of the actual page dimensions,
-	// Thus 20-by-20 doesn't mean square, but "same as page ratio".
-	const actualRatio = width / height * PAGE_WIDTH / PAGE_HEIGHT;
-	const imgProps = getImgProps( scale, offsetX, offsetY, origRatio, actualRatio );
+function ImageDisplay( { id, src, origRatio, width, height, x, y, scale, focalX, focalY, rotationAngle, isFullbleed, forwardedRef, onPointerDown } ) {
 	const elementProps = {
-		width,
-		height,
-		x,
-		y,
-		rotationAngle,
+		...getBox( { x, y, width, height, rotationAngle, isFullbleed } ),
 		ref: forwardedRef,
 		onPointerDown,
 	};
+	const imgProps = getImgProps( elementProps.width, elementProps.height, scale, focalX, focalY, origRatio );
 	const {
 		actions: { setEditingElement },
 	} = useCanvas();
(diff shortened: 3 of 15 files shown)


assets/src/edit-story/components/canvas/page.js
```diff
@@ -40,6 +41,24 @@ const Element = styled.div`
 
 function Page() {
 	const [ targetEl, setTargetEl ] = useState( null );
+	const [ editorUnits, setEditorUnits ] = useState( { width: 0, height: 0 } );
+	const backgroundRef = useRef( null );
+
+	// Measure the editor's size and handle resize events.
```
barklund: Does the page background resize? Isn't it fixed?

assets/src/edit-story/elements/image.js
```diff
@@ -16,13 +17,19 @@ const Element = styled.img`
...
+	} = useUnits();
 	const props = {
-		width,
-		height,
-		x,
-		y,
-		rotationAngle,
+		// TODO: what to do with 100% width/height?
+		width: isBackground ? percentToEditorPx( 100, 'w' ) : width,
+		height: isBackground ? percentToEditorPx( 100, 'v' ) : height,
```
barklund: `'v'`? I know that any non-`'w'` value works, but isn't `h` the logical choice? 😃 

```suggestion
		height: isBackground ? percentToEditorPx( 100, 'h' ) : height,
```

assets/src/edit-story/units/unitsProvider.js
```diff
@@ -0,0 +1,48 @@
...
+import { useCallback } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import Context from './context';
+
+function UnitsProvider( { width, height, children } ) {
+	const percentToEditorPx = useCallback(
+		( percent, dir ) => dirValue( dir, width, height ) * percent * 0.01,
```
miina: I wonder if background images should keep the ratio as well, e.g. use `cover` in the FE and in the editor make the (relatively) shorter side 100%, the other 100%+. UX question?

assets/src/edit-story/components/canvas/page.js
```diff
@@ -72,76 +91,78 @@ function Page() {
...
-				/>
-			) }
+		<Background ref={ backgroundRef }>
+			<UnitsProvider width={ editorUnits.width } height={ editorUnits.height }>
+				{ currentPage && currentPage.elements.map( ( { type, id, ...rest } ) => {
+					const comp = getComponentForType( type );
+					const Comp = comp; // why u do dis, eslint?
+					// Ignore multi-selection for now.
+					if ( 1 === selectedElements.length && selectedElements[ 0 ].id === id ) {
+						displayMoveable = true;
```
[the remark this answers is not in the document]
miina: Or at least not the one which allows resizing and rotating and is taking width and height and `x` and `y` from the original element.

assets/src/edit-story/components/library/mediaLibrary.js
```diff
@@ -16,14 +16,58 @@ import { useEffect } from '@wordpress/element';
...
+
+	opacity: 0;
+  ${ Thumb }:hover & {
+    opacity: 1;
+  }
+`;
+
+const ThumbButton = styled.button`
+	background: rgba(0, 0, 0, 0.7);
+	color: #fff;
```
miina: Should these come from the theme colors? Or is this temporary, waiting for the final UX anyway?

assets/src/edit-story/components/canvas/page.js
dvoytenko: I'm confirming that part right now. But for safety I added resizer. Was there a specific discussion w.r.t. to the editor's page size? Perhaps for the older editor? I wasn't sure if 434x775 was based on anything specific.

assets/src/edit-story/elements/image.js
dvoytenko: Right. Do we use any asserts? I couldn't find any yet.

assets/src/edit-story/components/canvas/page.js
dvoytenko: But assuming we just deal with the fixed 434x775, are we ok to use `PAGE_WIDTH` and `PAGE_HEIGHT` throughout our code for this? I can delay `units` part for a bit then.

barklund: For P1 I think this is a safe assumption. @pbakaus?

barklund: This is actually an open question regarding the overall responsive design strategy (created in #3920). For this ticket, I think it's safe to assume that the page never changes size, and if it does, we'll handle it in #3920.

assets/src/edit-story/components/library/mediaLibrary.js
```diff
@@ -56,15 +100,19 @@ function MediaLibrary( { onInsert } ) {
...
+	const onSelect = ( { url, src, width, height, x, y, rotationAngle, isBackground }, event ) => {
 		onInsert( 'image', {
-			src: url,
-			width: 100,
-			height: 100,
-			x: 5,
-			y: 5,
-			rotationAngle: 0,
+			src: src || url,
+			isBackground: isBackground || false,
```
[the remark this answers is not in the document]
barklund: Hm, we need to think about how to handle this in the story state management actually. Let's keep it out of this ticket for now.

assets/src/edit-story/elements/image.js
barklund: I think its unusual to use `assert` in React development. Also here we could use a bool rather than magic strings - `isHorizontal` perhaps. If not, then an enum would be better IMO.

assets/src/edit-story/units/unitsProvider.js
barklund: Images should always keep ratio AFAIK.

assets/src/edit-story/components/canvas/page.js
pbakaus: Page canvas shouldn't change in the editor, for now.

dvoytenko: Great! Rolling back this code for now then.

assets/src/edit-story/components/library/mediaLibrary.js
dvoytenko: Rolling back this file, but asked on the Slack how to use theme colors. I wasn't sure if I understood it right...

assets/src/edit-story/elements/image.js
dvoytenko: Ok. Switching to constants for now.

dvoytenko: But for the future: let's do enums.

assets/src/edit-story/components/canvas/page.js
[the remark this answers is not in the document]
dvoytenko: I kept the moveable/selection for now. I'm just setting `draggable`, `resizable` and `rotatable` to `false`. I'm still following up with UX on whether full-bleed elements allow rotation.

dvoytenko:
@miina @barklund PTAL. I removed "isBackground" work and only left "isFullbleed". This code is a lot more targeted now. Also updated image ratio and moveable parameters for full-bleed component.

assets/src/edit-story/panels/fullbleed.js
```diff
@@ -0,0 +1,47 @@
...
+	const isFullbleed = getCommonValue( selectedElements, 'isFullbleed' );
+	const [ state, setState ] = useState( { isFullbleed } );
+	useEffect( () => {
+		setState( { isFullbleed } );
+	}, [ isFullbleed ] );
+	const handleClick = ( evt ) => {
+		const newState = { isFullbleed: ! state.isFullbleed };
+		setState( newState );
+		onSetProperties( newState );
+		evt.preventDefault();
```
pbakaus: why is preventDefault needed here?

assets/src/edit-story/panels/shared.js
```diff
@@ -36,16 +36,23 @@ const Group = styled.label`
 	display: flex;
 	align-items: center;
 	margin-bottom: 5px;
+	opacity: ${ ( { disabled } ) => disabled ? 0.7 : 1 };
 `;
 
-function InputGroup( { label, value, isMultiple, onChange, postfix } ) {
+export const ActionButton = styled.button`
+	color: ${ ( { theme } ) => theme.colors.mg.v1 };
+	font-size: 11px;
```
pbakaus: I'd rather not define font colors on individual elements – we should handle these further up the stack somewhere, to enforce consistency.

assets/src/edit-story/components/canvas/page.js
```diff
@@ -91,10 +101,11 @@ function Page() {
 			) }
 			{ displayMoveable && targetEl && (
 				<Moveable
+					ref={ moveable }
 					target={ targetEl }
-					draggable={ true }
-					resizable={ true }
-					rotatable={ true }
+					draggable={ ! selectedElements[ 0 ].isFullbleed }
```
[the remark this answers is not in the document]
barklund: We can block this on the underlying data management level, e.g. as the following invariant:

* The selection object can be either a list of *only* the background element id or a list of element ids of which none are equal to the background element id.

It will be enforced as:

* If you are setting the selection list, the background element must either be the only element or will be filtered out of the list.
* If you are adding the background element to the selection list, the previous selection must be empty or nothing happens.
* If you are adding a non-background element to the selection list, and the previous selection was the background element, the new element will be the only selected element.

Would that make sense?

assets/src/edit-story/panels/shared.js
barklund: So, the thing is... with styled-components we aren't actually doing the C of CSS - there's no real cascading going on. The stylesheet is generated as a _one-class-per-element_ style of CSS.

Cascading can be done in code (but won't be reflected in the generated CSS) by "extending" existing styled objects (conceptually similar to @mixin rather than @extend in sass), and that makes sense for conceptual inheritance/cascading. But for something as simple as colors, just using the theme is best practice.

For reference, Material UI and it's React implementation, uses a [similar approach](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Badge/Badge.js#L43-L57) of a common color scheme defined in one place and then using these colors on individual components directly (albeit in a much more complex scenario). Generated CSS is in this case also not actually cascading.

assets/src/edit-story/components/canvas/page.js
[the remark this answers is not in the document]
pbakaus: SGTM.

assets/src/edit-story/panels/shared.js
pbakaus: Yep, understood. I still greatly prefer the cascade personally, but willing to work with this for now. Let's extend the objects then, in this case, or do the same thing as colors and use a theme variable.

assets/src/edit-story/components/canvas/page.js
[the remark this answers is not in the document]
dvoytenko: @barklund @pbakaus to clarify, this PR is about "fullbleed" elements, not "background" elements. Background implies fullbleed, but also has more properties, such as being the only allowed "background" element on the page, and always being the bottom layer.

Selection rules make sense to me here, but

> If you are adding the background element to the selection list, the previous selection must be empty or nothing happens.

I think the last user action should win. I.e. when background element is added to a non-empty selection, the previously selected elements should be unselected and the new selection should only contain the background element.

assets/src/edit-story/panels/fullbleed.js
dvoytenko: Removed. Not needed. Just copy/pasted it from another panel that used `submit` event making the `preventDefault` necessary. But no need in this case.

miina:
Needs resolving the conflicts with `develop-stories`.

dvoytenko:
@miina @barklund This is the rebase to catch up the latest in develop-stories branch. Many things are unchanged, but an important diff: I changed the image's data model by switching from offsetX/Y to focalX/Y which is similar but measured from center. IMHO, focalX/Y have a more clear semantical meaning and thus they simplify a lot of operations fairly naturally. More importantly, focal points usually do not change as often as offsets do, e.g. when the scale changes. They are a bit more inline with `object-fit: cover` that we most often aim for. Consequently, the meaning of `scale` is also different here. Inline with `object-fit:cover; transform: scale(2)`, the `scale` defines the scale over the image's shell. This simplifies things as well.

assets/src/edit-story/elements/image/display.js
```diff
@@ -26,24 +25,17 @@ const Element = styled.div`
...
-	const actualRatio = width / height * PAGE_WIDTH / PAGE_HEIGHT;
-	const imgProps = getImgProps( scale, offsetX, offsetY, origRatio, actualRatio );
+function ImageDisplay( { id, src, origRatio, width, height, x, y, scale, focalX, focalY, rotationAngle, isFullbleed, forwardedRef, onPointerDown } ) {
 	const elementProps = {
-		width,
-		height,
-		x,
-		y,
-		rotationAngle,
+		...getBox( { x, y, width, height, rotationAngle, isFullbleed } ),
```
miina: Currently, after resizing an Image, the full-bleed doesn't work as expected anymore. Although the attributes for the `div` (`Element`) are updated, the actual style is not updated.
To reproduce: resize an image and then try to set as full-bleed.

assets/src/edit-story/components/movable/index.js
```diff
@@ -73,9 +78,9 @@ function Movable( {
 		<Moveable
 			ref={ moveable }
 			target={ targetEl }
-			draggable={ true }
-			resizable={ true }
-			rotatable={ true }
+			draggable={ ! selectedElement.isFullbleed }
```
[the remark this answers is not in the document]
dvoytenko: Yes. I'm pretty sure we still want to show the selection via the same Moveable. But not yet 100% we will want to use the same  Moveable from crop lines. I was going to try to reuse and see how hard that looks.

assets/src/edit-story/elements/image/display.js
dvoytenko: Great catch! The issue is the conflict of styles between `Moveable` and `SetProperties`. `Moveable` works via inline styles and `SetProperties` via stylesheet (via styled elements). Thus the `Moveable` with inline styles always takes precedence. Fixed by resetting inline styles after `Moveable` is done.

dvoytenko:
> There's a general issue with the Image element (displaying stretched out / scaled) but doesn't look like it's related to this PR.

Confirmed as unrelated. But I will look into this separately.

PR #3912, opened 2019-12-11
Eliminate JS scroll fix for Twenty Twenty theme

### LINKED ISSUE #3900: Eliminate add_smooth_scrolling from Twenty Twenty core theme sanitization
In WordPress 5.3.1, Twenty Twenty will remove its JS-based smooth scrolling in favor of `scroll-behavior` CSS. See [Make/Core post](https://make.wordpress.org/core/2019/12/08/twenty-twenty-animated-scroll-changes-in-wordpress-5-3-1/). This being the case, we should be able to eliminate this code from the core theme sanitizer:

https://github.com/ampproject/amp-wp/blob/bf8a1740e3561a6852b36155f6b3c9998deeaa7e/includes/sanitizers/class-amp-core-theme-sanitizer.php#L67-L70

Trac ticket: https://core.trac.wordpress.org/ticket/48763

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

This PR removes the now defunct fix for adding smooth scrolling for Twenty Twenty theme on WP > 5.3.0, and keeps it for WP < 5.3.1.

It also fixes an issue with synchronized scrolling in paired browsing due to the `scroll-behavior:smooth` rule being applied to the clients.

Fixes #3900.

## Checklist

### DIFF
--- a/assets/src/admin/paired-browsing/app.js
+++ b/assets/src/admin/paired-browsing/app.js
@@ -299,6 +299,9 @@ class PairedBrowsingApp {
 			{ passive: true },
 		);
 
+		// Scrolling is not synchronized if `scroll-behavior` is set to `scroll`.
+		win.document.querySelector( 'html' ).style.setProperty( 'scroll-behavior', 'auto', 'important' );
+
 		// Make sure the opposite iframe is set to match.
 		if (
 			oppositeWindow &&
--- a/assets/src/admin/paired-browsing/app.js
+++ b/assets/src/admin/paired-browsing/app.js
@@ -299,8 +299,8 @@ class PairedBrowsingApp {
 			{ passive: true },
 		);
 
-		// Scrolling is not synchronized if `scroll-behavior` is set to `scroll`.
-		win.document.querySelector( 'html' ).style.setProperty( 'scroll-behavior', 'auto', 'important' );
+		// Scrolling is not synchronized if `scroll-behavior` is set to `smooth`.
+		win.document.documentElement.style.setProperty( 'scroll-behavior', 'auto', 'important' );
 
 		// Make sure the opposite iframe is set to match.
 		if (
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -932,11 +932,14 @@ function amp_get_content_sanitizers( $post = null ) {
 		];
 	}
 
-	if ( 'twentytwenty' === get_template() && version_compare( get_bloginfo( 'version' ), '5.3.1', '<' ) ) {
-		$sanitizers['AMP_Core_Theme_Sanitizer']['theme_features']['add_smooth_scrolling'] = [
-			// @todo Only replaces twentytwenty.smoothscroll.scrollToAnchor, but not twentytwenty.smoothscroll.scrollToElement
-			'//a[ starts-with( @href, "#" ) and not( @href = "#" )and not( @href = "#0" ) and not( contains( @class, "do-not-scroll" ) ) and not( contains( @class, "skip-link" ) ) ]',
-		];
+	if ( 'twentytwenty' === get_stylesheet() ) {
+		$theme = wp_get_theme( 'twentytwenty' );
+
+		if ( $theme->exists() && version_compare( $theme->get( 'Version' ), '1.0.0', '<=' ) ) {
+			$sanitizers['AMP_Core_Theme_Sanitizer']['theme_features']['add_smooth_scrolling'] = [
+				'//a[ starts-with( @href, "#" ) and not( @href = "#" )and not( @href = "#0" ) and not( contains( @class, "do-not-scroll" ) ) and not( contains( @class, "skip-link" ) ) ]',
+			];
+		}
 	}
 
 	/**
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -932,16 +932,6 @@ function amp_get_content_sanitizers( $post = null ) {
 		];
 	}
 
-	if ( 'twentytwenty' === get_stylesheet() ) {
-		$theme = wp_get_theme( 'twentytwenty' );
-
-		if ( $theme->exists() && version_compare( $theme->get( 'Version' ), '1.0.0', '<=' ) ) {
-			$sanitizers['AMP_Core_Theme_Sanitizer']['theme_features']['add_smooth_scrolling'] = [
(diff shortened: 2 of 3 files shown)


pierlon:
@schlessera I tried to implement the `todo` you added for replacing `twentytwenty.smoothscroll.scrollToElement`, but I'm not sure if the JS function itself even works as intended.

It [retrieves the element to scroll to](https://github.com/WordPress/twentytwenty/blob/6d0a5240af108d02f58ec797fd49b32458a4c698/assets/js/index.js#L392):

```js
target = event.target.dataset.twentytwentyScrollTo;
```

but I think the element should be retrieved before [calculating the scroll offset](https://github.com/WordPress/twentytwenty/blob/master/assets/js/index.js#L402-L404):

```js
// Determine offset
originalOffset = target.getBoundingClientRect().top + window.pageYOffset;
scrollOffset = additionalOffset ? originalOffset + additionalOffset : originalOffset;
```

----

If my assumption is correct the `todo` should be removed, and possibly alert the maintainers for Core Themes about the issue.

assets/src/admin/paired-browsing/app.js
```diff
@@ -299,6 +299,9 @@ class PairedBrowsingApp {
 			{ passive: true },
 		);
 
+		// Scrolling is not synchronized if `scroll-behavior` is set to `scroll`.
```
westonruter:
```suggestion
		// Scrolling is not synchronized if `scroll-behavior` is set to `smooth`.
```

assets/src/admin/paired-browsing/app.js
```diff
@@ -299,6 +299,9 @@ class PairedBrowsingApp {
...
+		win.document.querySelector( 'html' ).style.setProperty( 'scroll-behavior', 'auto', 'important' );
```
westonruter: Great idea to set this with JS. And TIL `setProperty`!

```suggestion
		win.document.documentElement.style.setProperty( 'scroll-behavior', 'auto', 'important' );
```

westonruter: Please also check what happens if someone sets `scroll-behavior` on the `body` instead.

includes/amp-helper-functions.php
```diff
@@ -932,6 +932,13 @@ function amp_get_content_sanitizers( $post = null ) {
 		];
 	}
 
+	if ( 'twentytwenty' === get_template() && version_compare( get_bloginfo( 'version' ), '5.3.1', '<' ) ) {
```
westonruter: This isn't technically correct.  It should be checking the version of the theme not core. It's possible for someone to have an older version of Twenty Twenty but have the current version of core.

pierlon: That's what I initially had in mind as well, but [the version code](https://github.com/WordPress/wordpress-develop/blob/388524c939f5efa66010231ba60ff2a8756c75e8/src/wp-content/themes/twentytwenty/style.css#L4) for Twenty Twenty hasn't been bumped since it was integrated into Core.

pierlon: > It's possible for someone to have an older version of Twenty Twenty but have the current version of core.

Hmm. The current version of the theme is `1.0.0`. Would checking for that also suffice?

westonruter: Yeah, if the version is greater than 1.0.0 then the config should be omitted.

includes/sanitizers/class-amp-core-theme-sanitizer.php
```diff
@@ -64,10 +64,6 @@ class AMP_Core_Theme_Sanitizer extends AMP_Base_Sanitizer {
 					'twentytwenty_skip_link_focus_fix', // See <https://github.com/WordPress/twentynineteen/pull/47>.
 				],
 			],
-			'add_smooth_scrolling'                    => [
```
schlessera: It's annoying that we need to move this piece of logic just because it needs to be done conditionally now.

In the past I've used closures for something like this. So, for a generic configuration system like we have here, the property getter would check if the property is callable first, and execute it (with some basic context) to fetch the actual dynamic property if that's the case.

So, this would stay in place but turn into something like this instead:
```php
'add_smooth_scrolling' => function ( $some_context ) {
	return ( 'twentytwenty' === get_template() && version_compare( get_bloginfo( 'version' ), '5.3.1', '<' ) )
		? [ '//a[ starts-with( @href, "#" ) and not( @href = "#" )and not( @href = "#0" ) and not( contains( @class, "do-not-scroll" ) ) and not( contains( @class, "skip-link" ) ) ]' ]
		: []
	}
},
```
Not sure we want something like there but I wanted to mention that possibility nevertheless.

includes/amp-helper-functions.php
pierlon: Version number for Twenty Twenty should be bumped via Trac [ticket #48944](https://core.trac.wordpress.org/ticket/489440), so only checking the theme's version code should be fine, as stated earlier.

assets/src/admin/paired-browsing/app.js
pierlon: Setting `scroll-behavior`set to `smooth on the `body` element doesn't have the same effect.

includes/amp-helper-functions.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/3912/commits/a1528d53e0f95c1274855a2ad9897dfdb3fcc7dd.

includes/sanitizers/class-amp-core-theme-sanitizer.php
pierlon: Agreed. Using a convention such as that outlined above would be one way of preventing further logic from being moved elsewhere to then be applied conditionally.

Suggestions @westonruter?

westonruter: Yes, that sounds good. It could be simply a matter of refactoring `\AMP_Core_Theme_Sanitizer::$theme_features` to instead be a method like `\AMP_Core_Theme_Sanitizer::get_theme_features()`

pierlon: Done in https://github.com/ampproject/amp-wp/pull/3912/commits/961cd23af1b7673abddaecf8b740acf5ec4a96eb.

westonruter:
Cherry-picked commits onto 1.4 branch, excluding 853c6021fa5ddccaccc731a1f392b75abd8866cf and a2e4b792f6b124404b2f733e075be4106dd68591 since specific to paired browsing.

PR #3928, opened 2019-12-13
Include admin-bar script deps in dev mode (e.g. hoverintent-js)

### LINKED ISSUE #3927: Validation error for hoverintent-js appearing in WP 5.3.1
Hi
I have upgraded WP to version 5.3.1 and now in validation have a new issue with a js file loaded in the admin side,
`wp-includes/js/hoverintent-js.min.js`
Is loaded by the WP core file 
script-loader.php:2647 
cript-loader.php:2637

westonruter:
## Summary

Fixes #3927.

Before:

image

After:

image

## Checklist

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1498,6 +1498,38 @@ class AMP_Theme_Support {
 	}
 
 	/**
+	 * Check if a handle is exclusively a dependency of another handle.
+	 *
+	 * For example, check if dashicons is being added exclusively because it is a dependency of admin-bar, as opposed
+	 * to being added because it was directly enqueued by a theme or a dependency of some other style.
+	 *
+	 * @since 1.4.2
+	 *
+	 * @param WP_Dependencies $dependencies         Dependencies.
+	 * @param string          $handle               Handle.
+	 * @param string          $exclusive_dependency Exclusive dependency handle.
+	 * @return bool Whether the $handle is exclusively a handle of the $exclusive_dependency handle.
+	 */
+	protected static function is_exclusive_dependency( WP_Dependencies $dependencies, $handle, $exclusive_dependency ) {
+		foreach ( $dependencies->queue as $queued_handle ) {
+			if (
+				// If a theme or plugin directly enqueued handle, then it is not added via $exclusive_dependency and it is not part of dev mode.
+				$handle === $queued_handle
+				||
+				// If a stylesheet has $handle as a dependency without also having $exclusive_dependency as a dependency, then no dev mode.
+				(
+					self::has_dependency( $dependencies, $queued_handle, $handle )
+					&&
+					! self::has_dependency( $dependencies, $queued_handle, $exclusive_dependency )
+				)
+			) {
+				return false;
+			}
+		}
+		return true;
+	}
+
+	/**
 	 * Add data-ampdevmode attribute to any enqueued style that depends on the admin-bar.
 	 *
 	 * @since 1.3
@@ -1507,30 +1539,11 @@ class AMP_Theme_Support {
 	 * @return string Tag.
 	 */
 	public static function filter_admin_bar_style_loader_tag( $tag, $handle ) {
-		if ( 'dashicons' === $handle ) {
-			// Conditionally include Dashicons in dev mode only if was included because it is a dependency of admin-bar.
-			$needs_dev_mode = true;
-			foreach ( wp_styles()->queue as $queued_handle ) {
-				if (
-					// If a theme or plugin directly enqueued dashicons, then it is not added via admin-bar dependency and it is not part of dev mode.
-					'dashicons' === $queued_handle
-					||
-					// If a stylesheet has dashicons as a dependency without also having admin-bar as a dependency, then no dev mode.
-					(
-						self::has_dependency( wp_styles(), $queued_handle, 'dashicons' )
-						&&
-						! self::has_dependency( wp_styles(), $queued_handle, 'admin-bar' )
-					)
-				) {
-					$needs_dev_mode = false;
-					break;
-				}
-			}
-		} else {
-			$needs_dev_mode = self::has_dependency( wp_styles(), $handle, 'admin-bar' );
-		}
-
-		if ( $needs_dev_mode ) {
+		if (
+			in_array( $handle, wp_styles()->registered['admin-bar']->deps, true ) ?
+				self::is_exclusive_dependency( wp_styles(), $handle, 'admin-bar' ) :
+				self::has_dependency( wp_styles(), $handle, 'admin-bar' )
+		) {
 			$tag = preg_replace( '/(?<=<link)(?=\s|>)/i', ' ' . AMP_Rule_Spec::DEV_MODE_ATTRIBUTE, $tag );
 		}
 		return $tag;
@@ -1547,9 +1560,9 @@ class AMP_Theme_Support {
 	 */
 	public static function filter_admin_bar_script_loader_tag( $tag, $handle ) {
 		if (
-			self::has_dependency( wp_scripts(), $handle, 'admin-bar' )
-			||
-			in_array( $handle, wp_scripts()->registered['admin-bar']->deps, true )
+			in_array( $handle, wp_scripts()->registered['admin-bar']->deps, true ) ?
+				self::is_exclusive_dependency( wp_scripts(), $handle, 'admin-bar' ) :
+				self::has_dependency( wp_scripts(), $handle, 'admin-bar' )
 		) {
 			$tag = preg_replace( '/(?<=<script)(?=\s|>)/i', ' ' . AMP_Rule_Spec::DEV_MODE_ATTRIBUTE, $tag );
 		}
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1546,7 +1546,11 @@ class AMP_Theme_Support {
 	 * @return string Tag.
 	 */
 	public static function filter_admin_bar_script_loader_tag( $tag, $handle ) {
-		if ( self::has_dependency( wp_scripts(), $handle, 'admin-bar' ) ) {
+		if (
+			self::has_dependency( wp_scripts(), $handle, 'admin-bar' )
+			||
+			in_array( $handle, wp_scripts()->registered['admin-bar']->deps, true )
+		) {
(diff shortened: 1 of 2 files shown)


includes/class-amp-theme-support.php
```diff
@@ -1497,6 +1497,38 @@ protected static function has_dependency( WP_Dependencies $dependencies, $curren
...
+	 *
+	 * @param WP_Dependencies $dependencies         Dependencies.
+	 * @param string          $handle               Handle.
+	 * @param string          $exclusive_dependency Exclusive dependency handle.
+	 * @return bool Whether the $handle is exclusively a handle of the $exclusive_dependency handle.
+	 */
+	protected static function is_exclusive_dependency( WP_Dependencies $dependencies, $handle, $exclusive_dependency ) {
+		foreach ( $dependencies->queue as $queued_handle ) {
+			if (
+				// If a theme or plugin directly enqueued handle, then it is not added via $exclusive_dependency and it is not part of dev mode.
```
schlessera: The comments here should not refer to the dev-mode, as the code that was now extracted is not directly concerned with dev-mode anymore.

includes/class-amp-theme-support.php
```diff
@@ -1497,6 +1497,38 @@ protected static function has_dependency( WP_Dependencies $dependencies, $curren
...
+	 * Check if a handle is exclusively a dependency of another handle.
+	 *
+	 * For example, check if dashicons is being added exclusively because it is a dependency of admin-bar, as opposed
+	 * to being added because it was directly enqueued by a theme or a dependency of some other style.
+	 *
+	 * @since 1.4.2
+	 *
+	 * @param WP_Dependencies $dependencies         Dependencies.
+	 * @param string          $handle               Handle.
+	 * @param string          $exclusive_dependency Exclusive dependency handle.
```
schlessera: If I understand the logic correctly, then `$exclusive_dependency` is the wrong name for the third argument here. The dependency that is checked on its exclusivity is `$handle`.

includes/class-amp-theme-support.php
```diff
@@ -1497,6 +1497,38 @@ protected static function has_dependency( WP_Dependencies $dependencies, $curren
...
+	 * @return bool Whether the $handle is exclusively a handle of the $exclusive_dependency handle.
+	 */
+	protected static function is_exclusive_dependency( WP_Dependencies $dependencies, $handle, $exclusive_dependency ) {
+		foreach ( $dependencies->queue as $queued_handle ) {
+			if (
+				// If a theme or plugin directly enqueued handle, then it is not added via $exclusive_dependency and it is not part of dev mode.
+				$handle === $queued_handle
+				||
+				// If a stylesheet has $handle as a dependency without also having $exclusive_dependency as a dependency, then no dev mode.
```
schlessera: Same issue with the mention of no-dev in this comment.

westonruter:
@schlessera I've revisited the logic for the method used to determine if one dependency is exclusively the dependency of another dependency. The logic is making my brain hurt and my jet lag is not helping. I've added unit tests to help bring some assurance that the logic is correct, but it may need further tweaking.

Once this PR is merged we'll cut the 1.4.2-RC1 release.

westonruter:
Note for merger: squash merging cannot be used because 4f0983ab970e8924ddaf22eccd06a72ee0d9594e will conflict with 1.4 branch. So each other commit should be cherry-picked onto the 1.4 branch individually, with the reference to this PR added to the commit message.

includes/class-amp-theme-support.php
```diff
@@ -1497,6 +1497,57 @@ protected static function has_dependency( WP_Dependencies $dependencies, $curren
 		return false;
 	}
 
+	/**
+	 * Check if a handle is exclusively a dependency of another handle.
```
schlessera: Before reviewing the actual logic, let me try to rephrase here to make sure I understand the intent and we're on the same page here.

The way I understand it is that this method checks whether a dependency is *only* being pulled in as a child dependency of a given parent dependency and nowhere else (neither as a child dependency of another parent dependency, nor as a main dependency itself).

This is meant to be used to detect dependencies that are either safe to remove when their parent was removed or safe to add to devmode if their parent was added to devmode.

Is the above correct?

tests/php/test-class-amp-theme-support.php
```diff
@@ -1289,6 +1289,213 @@ function() {
...
+				},
+			],
+
+			'dashicons_in_dev_mode_because_all_dependents_depend_on_admin_bar' => [
+				static function () {
+					wp_enqueue_style( 'admin-bar' );
+					wp_enqueue_style( 'special-icons', 'https://example.com/special-icons.css', [ 'dashicons', 'admin-bar' ], '0.1' );
+				},
+				function ( DOMXPath $xpath ) {
+					$element = $xpath->query( '//link[ @id = "dashicons-css" ]' )->item( 0 );
```
schlessera: Should probably use `$this->assert_dev_mode_is_on_queried_element` as well here.

tests/php/test-class-amp-theme-support.php
```diff
@@ -1289,6 +1289,213 @@ function() {
...
+				static function () {
+					wp_enqueue_style( 'admin-bar' );
+					wp_enqueue_style( 'special-icons', 'https://example.com/special-icons.css', [ 'dashicons' ], '0.1' );
+				},
+				function ( DOMXPath $xpath ) {
+					$this->assert_dev_mode_is_not_on_queried_element( $xpath, '//link[ @id = "dashicons-css" ]' );
+				},
+			],
+
+			'dashicons_in_dev_mode_because_all_dependents_depend_on_admin_bar' => [
```
schlessera: This is the case that I think is debatable. I don't think that having both `'dashicons'` and `'admin-bar'` as dependencies warrants the conclusion that `dashicons` is not used outside the admin bar.

westonruter: I'm not sure. The `special-icons` style here depends on `admin-bar` so if you enqueue it then you'd also get the `admin-bar`. In other words, it's true that perhaps some of the icons are being used outside the admin bar area, but perhaps this is because of some inline editing features. This is good because those special icons are needed by that authenticated user in dev mode. So if the user is logged-out, and the admin bar is not being included, then these should not be getting enqueued.

includes/class-amp-theme-support.php
```diff
@@ -1497,6 +1497,57 @@ protected static function has_dependency( WP_Dependencies $dependencies, $curren
...
+	 * @since 1.4.2
+	 *
+	 * @param WP_Dependencies $dependencies      Dependencies.
+	 * @param string          $dependency_handle Dependency handle.
+	 * @param string          $dependent_handle  Dependent handle.
+	 * @return bool Whether the $handle is exclusively a handle of the $exclusive_dependency handle.
+	 */
+	protected static function is_exclusively_dependent( WP_Dependencies $dependencies, $dependency_handle, $dependent_handle ) {
+
+		// If a dependency handle is the same as the dependent handle, then this self-referential relationship is exclusive.
```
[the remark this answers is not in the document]
westonruter: I included this condition as somewhat of an edge case test since if someone has `jquery` as a dependency of `jquery` this actually causes WordPress to self-destruct:

> `PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 20480 bytes) in /wp-includes/class.wp-dependencies.php on line 151`

This is due to `WP_Dependencies::all_deps()` not catching infinite loops: https://github.com/WordPress/wordpress-develop/blob/3b25bc43628a3b0c033248aef260581277880a32/src/wp-includes/class.wp-dependencies.php#L132-L133

The code path won't ever be traversed in how this code is used because `admin-bar` will never a dependency of `admin-bar`. 

So we might as well remove this condition.

includes/class-amp-theme-support.php
westonruter: > (neither as a child dependency of another parent dependency

I don't think this part is quite right. It should be allowed to a (recursive) dependency of something else, if that thing is also ultimately a dependency of the dependency in question.

> This is meant to be used to detect dependencies that are either safe to remove when their parent was removed or safe to add to devmode if their parent was added to devmode.

The latter, yes, whether we can add dev mode to the dependency. We only want to add dev mode to dependences which are ultimately and exclusively _connected_ with the admin bar. So this is what this method is trying to check for.

includes/class-amp-theme-support.php
```diff
@@ -1497,6 +1497,57 @@ protected static function has_dependency( WP_Dependencies $dependencies, $curren
...
+				return false;
+			}
+
+			// Otherwise, if the dependency handle is depended on by the queued handle while at the same time the queued
+			// handle _does_ have a dependency on the supplied dependent handle, then the dependency handle is not
+			// exclusively dependent on the dependent handle.
+			if (
+				self::has_dependency( $dependencies, $queued_handle, $dependency_handle )
+				&&
+				! self::has_dependency( $dependencies, $queued_handle, $dependent_handle )
```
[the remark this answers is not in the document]
westonruter: If this check is not here then the `admin_bar_scripts_have_dev_mode` dataset test fails with:

> Expected dev mode to be enabled on element for query: `//script[ contains( @src, "/hoverintent-js" ) ]`

The `hoverintent-js` asset should _NOT_ get dev mode if:

1. There is a queued handle that depends on it, _and_
2. This queued handle does _not_ ultimately depend on admin bar.

The second point corresponds to this second condition above.

westonruter:
@pierlon There is a test failure in WordPress 4.9 due to the amp-paired-browsing-client not being printed: https://travis-ci.org/ampproject/amp-wp/jobs/625330248#L1062-L1068

I believe this is due to the fact that `amp-paired-browsing-client` depends on both `wp-dom-ready` and `wp-polyfill` but only the former is being polyfilled for WordPress 4.9:

https://github.com/ampproject/amp-wp/blob/80e70e3a48a55e80597c698d748180e94716c471/webpack.config.js#L261-L263

In other words, there is no polyfill for the polyfills.

Three options I see here:

1. Increase minimum WP version to 5.0 and eliminate these pre-Gutenberg polyfills altogether.
2. Polyfill the polyfills.
3. Prevent paired browsing from being available in WordPress 4.9.

I've gone ahead and removed the test case in de90b0d5, but this will need to be fixed for v1.5 in a separate PR.

westonruter:
⚠️ Rebasing with latest changes from `develop`.

westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/3965614/amp.zip) (1.5.0-alpha-20191216T000003Z-64a80518)

Issue #3895, opened 2019-12-06
Image: Cropped/masked editing


dvoytenko:
## Feature description

* How does the image editing work?
* How does it interop with the selection transformers?
* How are the editing tools overlayed?
* How do we show the background for image cropping?

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A suggestion is made based on the findings

## Implementation brief

* A very preliminary implementation demo is [here](https://gist.github.com/dvoytenko/5b75b685595a1e4da0a89f6e61fbba4e). Here's a short [video](https://drive.google.com/file/d/1sh31zM7o5rw0uIYtu0QBkK56ioC2TUqr/view?usp=sharing).
* There are two transformable elements involved: the image's shell (the element itself) and the image panning within the cropped shell. Ideally we will use the Moveable library for them as well.
* These transformables are mostly independent from each other: the elements are not grouped and not operating as a group. But they define boundaries and guidelines for each other. Thus this will require two `Moveable` components.
* I can reuse the existing "selection" `Moveable` implemented in #3798 for the element's shell or we can add a new one. The differences: we have to restyle the resizing grips to crop style and we need to disable dragging and rotation. The demo borrows the `Moveable` from #3798 but changes it quite a bit. But, as in #3798 resizing is done by outline and the actual element is resized once it's released. I think it's better since it shows a clear indication of before/after state. The resize operation itself is expensive: it updates pan position, scale factor, etc.
* The second `Moveable` only enables dragging. It's for panning of the full image in the cropped shell. Unlike resize - the demo implements panning immediately to show exactly how focal point changes. Note that panning is very cheap comparing to resize.
* It's very important that panning `Moveable` is on top of the resizing one. This appears a limitation of the library.
* The demo uses state for element properties. This will, of course, change.
* The panning/cropping implemented using two images on top of each other that are 100% synchronized. The bottom image has opacity of 0.5 and is used to show the "ghost" image behind the actual image. The top image has opacity of 1 and is clipped to the element's outline.
* Demo adds scaler per design mocks. However, this scaler "feels" more like an edit panel, with the only exception that it's not docked in the side-panel on the right. It looks like design mocks add more such "floating" panels for other elements. Perhaps we can implement them the same way as side-panels, except that they will be floating. E.g. `IMAGE_SCALE_PANEL: {float: true, anchor: 'bottom'}` or such.
[...]

PR #3953, opened 2019-12-17
Keep ratio property: edit panels and moveable

dvoytenko:
## Summary

Partial for #3895

`keepRatio: boolean` property has been introduced. Size edit panel and moveable transformers have been modified to use it.

## Checklist

### DIFF
--- a/assets/src/edit-story/components/movable/index.js
+++ b/assets/src/edit-story/components/movable/index.js
@@ -7,14 +7,13 @@ import PropTypes from 'prop-types';
 /**
  * WordPress dependencies
  */
-import { useRef, useEffect } from '@wordpress/element';
+import { useRef, useEffect, useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
  */
 import { useStory } from '../../app';
 
-const CORNER_HANDLES = [ 'nw', 'ne', 'sw', 'se' ];
 const ALL_HANDLES = [ 'n', 's', 'e', 'w', 'nw', 'ne', 'sw', 'se' ];
 
 function Movable( {
@@ -23,6 +22,7 @@ function Movable( {
 	pushEvent,
 } ) {
 	const moveable = useRef();
+	const [ keepRatioMode, setKeepRatioMode ] = useState( true );
 
 	const {
 		actions: { setPropertiesOnSelectedElements },
@@ -71,6 +71,7 @@ function Movable( {
 		setStyle( target );
 		target.style.width = '';
 		target.style.height = '';
+		setKeepRatioMode( true );
 		if ( moveable.current ) {
 			moveable.current.updateRect();
 		}
@@ -97,10 +98,14 @@ function Movable( {
 				setPropertiesOnSelectedElements( newProps );
 				resetMoveable( target );
 			} }
-			onResizeStart={ ( { setOrigin, dragStart } ) => {
+			onResizeStart={ ( { setOrigin, dragStart, direction } ) => {
 				setOrigin( [ '%', '%' ] );
 				if ( dragStart ) {
 					dragStart.set( frame.translate );
+				}
+				const newKeepRatioMode = direction[ 0 ] !== 0 && direction[ 1 ] !== 0;
+				if ( keepRatioMode !== newKeepRatioMode ) {
+					setKeepRatioMode( newKeepRatioMode );
 				}
 			} }
 			onResize={ ( { target, width, height, drag } ) => {
@@ -130,8 +135,8 @@ function Movable( {
 			} }
 			origin={ false }
 			pinchable={ true }
-			keepRatio={ selectedElement.keepRatio || false }
-			renderDirections={ selectedElement.keepRatio ? CORNER_HANDLES : ALL_HANDLES }
+			keepRatio={ 'image' === selectedElement.type && keepRatioMode }
+			renderDirections={ ALL_HANDLES }
 		/>
 	);
 }
--- a/assets/src/edit-story/components/movable/index.js
+++ b/assets/src/edit-story/components/movable/index.js
@@ -130,8 +130,8 @@ function Movable( {
 			} }
 			origin={ false }
 			pinchable={ true }
-			keepRatio={ 'image' === selectedElement.type } // @†odo Even image doesn't always keep ratio, consider moving to element's model.
-			renderDirections={ 'image' === selectedElement.type ? CORNER_HANDLES : ALL_HANDLES }
+			keepRatio={ selectedElement.keepRatio || false }
+			renderDirections={ selectedElement.keepRatio ? CORNER_HANDLES : ALL_HANDLES }
 		/>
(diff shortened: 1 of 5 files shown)


assets/src/edit-story/panels/shared.js
```diff
@@ -44,26 +44,30 @@ export const ActionButton = styled.button`
...
+				value={ isCheckbox ? null : value }
+				checked={ isCheckbox ? value : null }
 			/>
 			{ postfix }
 		</Group>
 	);
 }
 
 InputGroup.propTypes = {
+	type: PropTypes.string,
```
barklund: Doesn't this need a default value or React will start throwing warnings when missing? A default of `number` seems appropriate.

assets/src/edit-story/panels/size.js
```diff
@@ -31,21 +33,61 @@ function SizePanel( { selectedElements, onSetProperties } ) {
...
 				label="Width"
 				value={ state.width }
 				isMultiple={ width === '' }
-				onChange={ ( value ) => setState( { ...state, width: isNaN( value ) || value === '' ? '' : parseFloat( value ) } ) }
+				onChange={ ( value ) => {
+					const newWidth = isNaN( value ) || value === '' ? '' : parseFloat( value );
+					setState( {
+						...state,
+						width: newWidth,
+						// @todo: Move to the reducer once available. This is especially critical
```
barklund: Which reducer in particular? It's a pretty specific invariant.

dvoytenko: This is worth discussing. The business rules here are pretty specific and sensitive to what's actually specifically being updated. E.g. the width-only update may need to be reflected in the height for `keepRatio=true` elements. The `keepRatio` update may need to be updated in both width and height. And so on. This is all complicated in case if we update a width for multiple elements with different `keepRatio`/`origRatio` values. I think for us this means that:

1. We need to be able to run an update separately for each element.
2. We'd want to contain the update rules centrally.

Hence I'd like to see if we can move these rules to the reducer for element update dispatches that you're working on in the #3910.

assets/src/edit-story/panels/shared.js
dvoytenko: Done

dvoytenko:
@googlebot I consent.

assets/src/edit-story/components/movable/index.js
```diff
@@ -130,8 +130,8 @@ function Movable( {
 			} }
 			origin={ false }
 			pinchable={ true }
-			keepRatio={ 'image' === selectedElement.type } // @†odo Even image doesn't always keep ratio, consider moving to element's model.
-			renderDirections={ 'image' === selectedElement.type ? CORNER_HANDLES : ALL_HANDLES }
+			keepRatio={ selectedElement.keepRatio || false }
```
miina: Any reason for not using just `selectedElement.keepRatio` here, or its boolean value? Readability?

assets/src/edit-story/panels/size.js
```diff
@@ -34,18 +36,56 @@ function SizePanel( { selectedElements, onSetProperties } ) {
...
 			/>
+			<InputGroup
+				type="checkbox"
+				label="Keep ratio"
+				value={ state.keepRatio }
+				isMultiple={ keepRatio === '' }
+				onChange={ ( value ) => {
+					if ( value && typeof origRatio === 'number' ) {
+						onSetProperties( {
+							keepRatio: true,
```
miina: Hmm, should the checkbox influence the behavior of Moveable? Perhaps this should only lock the numbers when manually changing?

assets/src/edit-story/components/movable/index.js
dvoytenko: For now `keepRatio` is optional since we weren't sure if we wanted it in other elements. Do you think we should just make it required?

assets/src/edit-story/panels/size.js
dvoytenko: It updates both: the numbers and moveable's properties. I think once we have reducers in place we will move the width/height update there. But we definitely want to update `Moveable` props, such as which resizing corners are allowed, etc.

miina: Yep, saw that it's changing both, I'm thinking that maybe it should not change the behavior of Moveable? Maybe the lock icon in the Figma files is for manual updating only? For some reason that's what I'd assume from the UX -- that Moveable works as it does and the lock icon would be for the manual fields only. Anyway, perhaps you know better what's intended for UX.

dvoytenko: I'll confirm with the @samitron7.

assets/src/edit-story/components/movable/index.js
dvoytenko: This part is reimplemented.

assets/src/edit-story/panels/size.js
dvoytenko: This requirement is removed per clarification with @samitron7. See this thread: https://github.com/ampproject/amp-wp/pull/3953/files#r359940148 for details.

assets/src/edit-story/panels/size.js
dvoytenko: @miina I talked to @samitron7 and you were exactly right. These changes shouldn't cause Moveable changes. Further, here are the changes in the latest commit:

1. Ratio "lock" is no longer in the model. It's only part of edit panel.
2. Changing "lock" value does not update the width/height, but simply locks the ratio between them.
3. Moveable is no longer affected by this value either. Per latest discussion with @samitron7, all resize handles are now enabled for the image: corners preserve aspect ratio and the mid-points do not.

dvoytenko:
@miina @barklund PTAL again. This PR changed a lot based on your and @samitron7's feedback. See the changes described in the https://github.com/ampproject/amp-wp/pull/3953#discussion_r360195952

assets/src/edit-story/components/movable/index.js
```diff
@@ -97,11 +98,15 @@ function Movable( {
 				setPropertiesOnSelectedElements( newProps );
 				resetMoveable( target );
 			} }
-			onResizeStart={ ( { setOrigin, dragStart } ) => {
+			onResizeStart={ ( { setOrigin, dragStart, direction } ) => {
 				setOrigin( [ '%', '%' ] );
 				if ( dragStart ) {
 					dragStart.set( frame.translate );
 				}
+				const newKeepRatioMode = direction[ 0 ] !== 0 && direction[ 1 ] !== 0;
```
miina: Could we add a comment here explaining what this means -- `direction[ 0 ]` and `direction[ 1 ]`?

dvoytenko: Done

dvoytenko:
> Also seeing this in the console:
> Screenshot 2019-12-20 at 15 26 52
> 
> Otherwise looks good.

Done

PR #3957, opened 2019-12-18
Create a separate layer for all Moveable instances

dvoytenko:
## Summary

Partial for #3895

Based on the prototype in #3895, it appears that:

1. It will be fairly hard to reuse the selection moveable for cropping. So an additional 2 moveable instances are needed.
2. It's much easier to keep all moveables in the same place in DOM: for coordinate system, but also because in some cases their `z-index` order is important.

The proposal thus creates a portal inside the `Background` and all moveables are hoisted into it. This seems to work pretty well and would allow us to keep moveable instances inside the `Edit` components. We could even further try to make this a generic approach to host other overlaying tools.

## Checklist

### DIFF
--- a/assets/src/edit-story/components/canvas/page.js
+++ b/assets/src/edit-story/components/canvas/page.js
@@ -12,9 +12,10 @@ import { useCallback, useEffect, useState } from '@wordpress/element';
  * Internal dependencies
  */
 import { useStory } from '../../app';
+import SelectionMovable from '../selection';
+import MovableLayer from '../movable/movableLayer';
 import useCanvas from './useCanvas';
 import Element from './element';
-import Movable from './../movable';
 
 const Background = styled.div.attrs( { className: 'container' } )`
 	background-color: ${ ( { theme } ) => theme.colors.fg.v1 };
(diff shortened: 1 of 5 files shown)


miina:
This might be good for handling the moveable keyboard events as well, to have a wrapping layer.

> This seems to work pretty well and would allow us to keep moveable instances inside the `Edit` components.

Do you mean here the Moveable instances which don't exist yet? The ones for cropped image editing?

Q: Should there perhaps still be just one `Movable` component, maybe with a `type` attribute where `Movable` then would take care of loading the correct React Component instead of using a separate `SelectionMovable`? Wondering if keeping the Movable implementation under one component and folder would be better. Thoughts?

dvoytenko:
@miina 

> This might be good for handling the moveable keyboard events as well, to have a wrapping layer.

It might be useful for other things, yeah. But I'm not sure yet if we'll handle movable keyboard events on the element's wrapper or on the Moveable wrapper. We'll have to see which one is more natural.

> Do you mean here the Moveable instances that don't exist yet? The ones for cropped image editing?
> Q: Should there perhaps still be just one `Movable` component, maybe with a `type` attribute where `Movable` then would take care of loading the correct React Component instead of using a separate `SelectionMovable`? Wondering if keeping the Movable implementation under one component and folder would be better. Thoughts?

Yes, I added more moveables in https://github.com/ampproject/amp-wp/pull/3964. I tried a single moveable, but it simply didn't work. I needed two moveables with completely different targets: one for cropping and one for panning the bigger image within the cropped borders. Among other things the order of moveables (`zIndex`)  is also important here. But I don't think it's also too far of a departure for us: we would still want to minimize the number of moveables - but we will have one per type at a time: one moveable for selection, one for cropping, etc.

miina:
> But I'm not sure yet if we'll handle movable keyboard events on the element's wrapper or on the Moveable wrapper. We'll have to see which one is more natural.

I was mainly thinking of the group keyboard events here actually.

> I tried a single moveable, but it simply didn't work. I needed two moveables with completely different targets: one for cropping and one for panning the bigger image within the cropped borders. 

Already discussed in the other PR but I should have clarified that meant here having separate components per each Moveable implementation behind the scenes.

dvoytenko:
@miina 

> > But I'm not sure yet if we'll handle movable keyboard events on the element's wrapper or on the Moveable wrapper. We'll have to see which one is more natural.
> 
> I was mainly thinking of the group keyboard events here actually.

Oh yeah. That's an interesting case. Especially since none of the Moveable control boxes are focusable currently, right? So I don't think we can make the wrappers focusable either. A bit related, I also filed https://github.com/ampproject/amp-wp/issues/3994. I think we may need a more advanced global canvas keyboard processing.

> Already discussed in the other PR but I should have clarified that meant here having separate components per each Moveable implementation behind the scenes.

Let's finish that discussion on https://github.com/ampproject/amp-wp/pull/3964#discussion_r359977687

dvoytenko:
Closing in favor of #4010.

PR #3959, opened 2019-12-18
[WIP] Try: fix some event handler issues with Moveable

miina:
## Summary

Fixes #

## Checklist

### DIFF
--- a/assets/src/edit-story/components/canvas/element.js
+++ b/assets/src/edit-story/components/canvas/element.js
@@ -21,6 +21,7 @@ function Element( {
 	isSelected,
 	setNodeForElement,
 	handleSelectElement,
+	setPushEvent,
 	forwardedRef,
 	element: {
 		id,
@@ -49,7 +50,9 @@ function Element( {
 
 	return (
 		<Wrapper
-			onClick={ ( evt ) => handleSelectElement( id, evt ) }
+			onClick={ ( evt ) => {
+				handleSelectElement( id, evt );
+			} }
 			ref={ element }
 		>
 			<Display
@@ -59,6 +62,7 @@ function Element( {
 						handleSelectElement( id, evt );
 					}
 				} }
+				onPointerUp={ () => setPushEvent( null ) }
 				forwardedRef={ forwardedRef }
 			/>
 		</Wrapper>
@@ -70,6 +74,7 @@ Element.propTypes = {
 	isSelected: PropTypes.bool.isRequired,
 	setNodeForElement: PropTypes.func.isRequired,
 	handleSelectElement: PropTypes.func.isRequired,
+	setPushEvent: PropTypes.func.isRequired,
 	element: PropTypes.object.isRequired,
 	forwardedRef: PropTypes.oneOfType( [
 		PropTypes.object,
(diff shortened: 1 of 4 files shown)


miina:
There's an issue currently with the Image element and the persisted event when dragging without selecting the image first.
To reproduce:
- Add an image.
- Deselect image.
- Click the image (without holding) and move the cursor somewhere else.
- See the Image being dragged with just moving the mouse.

See here just after a click (no holding pointer down):
dragging


This PR "fixes it" by adding an `onPointerUp` even to cancel reset the persisted event, however, there's likely a better solution.

There's an issue currently with `onPointerDown` being triggered by the inner `img` element while `targetEl` is its parent element:
https://github.com/ampproject/amp-wp/blob/295131bc00ffdb5fb31d47af2480507cd1c4a726/assets/src/edit-story/elements/image/display.js#L46-L48

This results in the persistent event saved with `img` as the target of the event and triggering dragging with incorrect target since the check is for containing the element:
https://github.com/ampproject/amp-wp/blob/295131bc00ffdb5fb31d47af2480507cd1c4a726/assets/src/edit-story/components/movable/index.js#L39-L42

@barklund @dvoytenko Perhaps you have ideas on how to resolve this issue in a more elegant way for the Image block instead of using `onPointerUp`? I'm likely missing something obvious here...

dvoytenko:
@miina strangely enough i can't seem to repro. I've been following your steps, but the image stays put.

miina:
@dvoytenko That's odd, I tested the issue on `develop-stories` on different browsers just in case, always able to reproduce.

@spacedmonkey @barklund Would you mind testing if you can see the issue?

To reproduce (on `develop-stories`):
- Add an image
- Deselect the image first by clicking somewhere on the Page.
- Click the image (without dragging)
- Move the cursor and see the image moving with the cursor without actually dragging.

dvoytenko:
@miina I still can't repro no matter how I try. But even if using `onPointerUp` makes sense here, can we just set it up on the `Background` in the `canvas/page.js`?

barklund:
@miina, I also don't experience this problem in `develop-stories` at all. I do in #3910, but not elsewhere.

miina:
@barklund @dvoytenko I'm still seeing the same issue on `develop-stories`, updated everything again, no luck. No idea why it happens for me and not for you.

Could you test again but this way:
- Add 2-3 image elements and save the post.
- Refresh the page
- Now click on an unselected image (without dragging)
- Move the cursor

Can you see the issue now? I just realized that _sometimes_ it works as expected for me when I start on a clean story, however, so far, when trying on a previously existing story, the image always "sticks" to the cursor.

Let me know if you can see the issue now, if still not, then I'll change the base issue to #3910 instead.

dvoytenko:
@miina I can't repro using these steps and trying many different other ways. Though I'd still take this issue seriously. If one of us can repro then there has to be something there. But I think there could be an easier way for us to do this? Maybe we can simply catch `onPointerUp` to reset the last stored event? I'm also curious if you still would be able to repro this issue on #4010. I did bunch of variations on the events there. And even deeper changes are done in #4018.

miina:
@dvoytenko It's still happening in #4010 but not in #4018 -- not being able to reproduce it there. I'm happy to omit this PR in favor of #4018 if we go that way.

swissspidy:
@miina Is this obsolete now?

miina:
Yes, that's correct, forgot about it, thanks!

PR #3964, opened 2019-12-18
Cropped image editor

### LINKED ISSUE #3895: Image: Cropped/masked editing

dvoytenko:
## Summary

Fixes #3895

## Checklist

### DIFF
--- a/assets/src/edit-story/app/index.js
+++ b/assets/src/edit-story/app/index.js
@@ -17,6 +17,7 @@ import {
  * Internal dependencies
  */
 import theme, { GlobalStyle } from '../theme';
+import {GlobalStyle as CropMoveableGlobalStyle} from '../components/movable/cropStyle';
 import { useHistory, HistoryProvider } from './history';
 import { useAPI, APIProvider } from './api';
 import { useConfig, ConfigProvider } from './config';
@@ -33,6 +34,7 @@ function App( { config } ) {
 						<HistoryProvider size={ 50 }>
 							<StoryProvider storyId={ storyId }>
 								<GlobalStyle />
+								<CropMoveableGlobalStyle />
 								<Layout />
 								<Popover.Slot />
 							</StoryProvider>
--- a/assets/src/edit-story/app/index.js
+++ b/assets/src/edit-story/app/index.js
@@ -17,7 +17,7 @@ import {
  * Internal dependencies
  */
 import theme, { GlobalStyle } from '../theme';
-import {GlobalStyle as CropMoveableGlobalStyle} from '../components/movable/cropStyle';
+import { GlobalStyle as CropMoveableGlobalStyle } from '../components/movable/cropStyle';
 import { useHistory, HistoryProvider } from './history';
 import { useAPI, APIProvider } from './api';
 import { useConfig, ConfigProvider } from './config';
--- a/assets/src/edit-story/components/movable/cropStyle.js
+++ b/assets/src/edit-story/components/movable/cropStyle.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import {createGlobalStyle} from 'styled-components';
+import { createGlobalStyle } from 'styled-components';
 
 export const GlobalStyle = createGlobalStyle`
   .crop-movable .moveable-control {
--- a/assets/src/edit-story/components/movable/cropStyle.js
+++ b/assets/src/edit-story/components/movable/cropStyle.js
@@ -75,6 +75,7 @@ export const GlobalStyle = createGlobalStyle`
   }
 
   .crop-movable .moveable-control.moveable-nw {
+    transform-origin: 2px 2px !important;
     clip-path: polygon(0px 0px, 16px 0px, 16px 4px, 4px 4px, 4px 16px, 0px 16px) !important;
   }
   .crop-movable .moveable-control.moveable-nw:after {
@@ -82,6 +83,7 @@ export const GlobalStyle = createGlobalStyle`
   }
 
   .crop-movable .moveable-control.moveable-ne {
+    transform-origin: 14px 2px !important;
     clip-path: polygon(0px 0px, 16px 0px, 16px 16px, 12px 16px, 12px 4px, 0px 4px) !important;
   }
   .crop-movable .moveable-control.moveable-ne:after {
@@ -89,6 +91,7 @@ export const GlobalStyle = createGlobalStyle`
   }
 
   .crop-movable .moveable-control.moveable-sw {
+    transform-origin: 2px 14px !important;
     clip-path: polygon(0px 0px, 0px 16px, 16px 16px, 16px 12px, 4px 12px, 4px 0px) !important;
   }
   .crop-movable .moveable-control.moveable-sw:after {
@@ -96,6 +99,7 @@ export const GlobalStyle = createGlobalStyle`
   }
 
   .crop-movable .moveable-control.moveable-se {
+    transform-origin: 14px 14px !important;
     clip-path: polygon(16px 0px, 16px 16px, 0px 16px, 0px 12px, 12px 12px, 12px 0px) !important;
   }
   .crop-movable .moveable-control.moveable-se:after {
--- a/assets/src/edit-story/components/movable/cropStyle.js
+++ b/assets/src/edit-story/components/movable/cropStyle.js
@@ -1,0 +1,104 @@
+/**
+ * External dependencies
+ */
+import {createGlobalStyle} from 'styled-components';
+
+export const GlobalStyle = createGlobalStyle`
+  .crop-movable .moveable-control {
+    background: #222 !important;
+    border-radius: 0px !important;
+    border: none !important;
+    box-sizing: border-box !important;
+  }
+
+  .crop-movable .moveable-control.moveable-n,
+  .crop-movable .moveable-control.moveable-s,
+  .crop-movable .moveable-control.moveable-e,
+  .crop-movable .moveable-control.moveable-w {
+    border: 1px solid #fff !important;
+  }
+
+  .crop-movable .moveable-control.moveable-n,
+  .crop-movable .moveable-control.moveable-s {
+    width: 16px !important;
+    height: 4px !important;
+    margin-left: -8px !important;
+    margin-top: -2px !important;
+  }
+
+  .crop-movable .moveable-control.moveable-e,
+  .crop-movable .moveable-control.moveable-w {
+    width: 4px !important;
+    height: 16px !important;
+    margin-left: -2px !important;
+    margin-top: -8px !important;
+  }
+
+  .crop-movable .moveable-control.moveable-nw,
+  .crop-movable .moveable-control.moveable-ne,
+  .crop-movable .moveable-control.moveable-sw,
+  .crop-movable .moveable-control.moveable-se {
+    width: 16px !important;
+    height: 16px !important;
+    background: #fff !important;
+  }
+
+  .crop-movable .moveable-control.moveable-nw:after,
+  .crop-movable .moveable-control.moveable-ne:after,
+  .crop-movable .moveable-control.moveable-sw:after,
+  .crop-movable .moveable-control.moveable-se:after {
+    content: "" !important;
+    display: block !important;
+    position: absolute !important;
+    inset: 1px !important;
+    background: #222 !important;
+  }
+
+  .crop-movable .moveable-control.moveable-nw,
+  .crop-movable .moveable-control.moveable-ne {
+    margin-top: -2px !important;
+  }
+
+  .crop-movable .moveable-control.moveable-sw,
+  .crop-movable .moveable-control.moveable-se {
+    margin-top: -14px !important;
+  }
+
+  .crop-movable .moveable-control.moveable-nw,
+  .crop-movable .moveable-control.moveable-sw {
+    margin-left: -2px !important;
+  }
+
+  .crop-movable .moveable-control.moveable-ne,
+  .crop-movable .moveable-control.moveable-se {
+    margin-left: -14px !important;
+  }
+
+  .crop-movable .moveable-control.moveable-nw {
+    clip-path: polygon(0px 0px, 16px 0px, 16px 4px, 4px 4px, 4px 16px, 0px 16px) !important;
+  }
+  .crop-movable .moveable-control.moveable-nw:after {
+    clip-path: polygon(0px 0px, 14px 0px, 14px 2px, 2px 2px, 2px 14px, 0px 14px) !important;
+  }
+
+  .crop-movable .moveable-control.moveable-ne {
+    clip-path: polygon(0px 0px, 16px 0px, 16px 16px, 12px 16px, 12px 4px, 0px 4px) !important;
+  }
+  .crop-movable .moveable-control.moveable-ne:after {
+    clip-path: polygon(0px 0px, 14px 0px, 14px 14px, 12px 14px, 12px 2px, 0px 2px) !important;
+  }
+
+  .crop-movable .moveable-control.moveable-sw {
+    clip-path: polygon(0px 0px, 0px 16px, 16px 16px, 16px 12px, 4px 12px, 4px 0px) !important;
+  }
+  .crop-movable .moveable-control.moveable-sw:after {
+    clip-path: polygon(0px 0px, 0px 14px, 14px 14px, 14px 12px, 2px 12px, 2px 0px) !important;
+  }
+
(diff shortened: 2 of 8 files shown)


assets/src/edit-story/app/index.js
```diff
@@ -33,6 +34,7 @@ function App( { config } ) {
 						<HistoryProvider size={ 50 }>
 							<StoryProvider storyId={ storyId }>
 								<GlobalStyle />
+								<CropMoveableGlobalStyle />
```
miina: Should the existing `GlobalStyle` perhaps include all the global style? Not sure what's the best practice here. cc @barklund

assets/src/edit-story/components/movable/cropStyle.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  .crop-movable .moveable-control.moveable-nw:after,
+  .crop-movable .moveable-control.moveable-ne:after,
+  .crop-movable .moveable-control.moveable-sw:after,
+  .crop-movable .moveable-control.moveable-se:after {
+    content: "" !important;
+    display: block !important;
+    position: absolute !important;
+    inset: 1px !important;
```
miina: Is this supported in most of the browsers? Perhaps it's not that important :)

assets/src/edit-story/elements/image/edit.js
```diff
@@ -41,25 +48,172 @@ const FadedImg = styled.img`
...
-			<ActualBox>
-				<ActualImg src={ src } { ...imgProps } />
-			</ActualBox>
+			<FadedImg ref={ setFullImage } draggable={ false } src={ src } { ...imgProps } />
+			<CropBox ref={ setCropBox }>
+				<CropImg ref={ setCroppedImage } draggable={ false } src={ src } { ...imgProps } />
+			</CropBox>
+
+			{ /* Resizable moveable for cropping */ }
+			{ ! isFullbleed && (
```
miina: Shouldn't the logic here consider `rotationAngle`, too?

assets/src/edit-story/elements/image/edit.js
```diff
@@ -41,25 +48,172 @@ const FadedImg = styled.img`
...
+				<Movable
```
miina: Commented elsewhere as well but perhaps we could have all the Movable implementations under the one `Movable` component (a separate file per each implementation)? Would make these reusable (where applicable), and also keep the components smaller for testing/maintenance purposes.

At least the `Edit` for image is growing quite big like this.

Thoughts?

assets/src/edit-story/app/index.js
dvoytenko: Yeah. Just let me know the best practice here.

assets/src/edit-story/components/movable/cropStyle.js
dvoytenko: I assumed the CSS-in-JS library we use would expand this for non-supporting browsers. But if not, it'd be pretty easy to find them all and replace at any time.

assets/src/edit-story/elements/image/edit.js
dvoytenko: `rotationAngle` seems to work correctly now. Do you have some particular feature in mind?

miina: Does the cropping work as expected while the image is rotated? For some reason, I expected that this would need some extra handling.

_Testing_

assets/src/edit-story/elements/image/edit.js
dvoytenko: It could be possible as long as we use separate `Moveable` components behind the scenes. But one argument here: the implementation seems fairly closely connected with the very specific structure of the `Edit` component: the moveables manipulate separate targets in the subtree. We could also move these movables into separate modules, e.g. `EditResizeMovable` and `EditPanMovable`?

assets/src/edit-story/elements/image/edit.js
miina: I'm seeing this currently:

croppibg

Is it working as expected for you?

assets/src/edit-story/elements/image/edit.js
miina: > It could be possible as long as we use separate Moveable components behind the scenes.

Yes, I was thinking exactly that -- having separate Moveable components behind the scene.

>  But one argument here: the implementation seems fairly closely connected with the very specific structure of the Edit component: the moveables manipulate separate targets in the subtree

Could the Video element or some of the shapes potentially have a similar editing experience?

> We could also move these movables into separate modules, e.g. EditResizeMovable and EditPanMovable?

Could also be an option. These could potentially then just live together in one folder instead of using one `Moveable` with separate components behind the scenes.

assets/src/edit-story/elements/image/edit.js
dvoytenko: Oh. I see. It's the snap lines that mess with panning. Let me see what can be done here.

dvoytenko: @miina I had to disable most of bounds and guidelines for the rotated shapes. The horizontal/vertical implementation that Moveable provides is just not good enough UX. I kept the central guides since they look clear enough. The offset calculations can already readjust if the shape is moved out of bounds. I think this looks ok. PTAL.

assets/src/edit-story/elements/image/edit.js
```diff
@@ -41,25 +48,176 @@ const FadedImg = styled.img`
...
+					targets={ cropBox }
+
+					origin={ false }
+					edge={ false }
+
+					draggable={ false }
+					scalable={ false }
+					rotatable={ false }
+					warpable={ false }
+					pinchable={ false }
```
miina: ^ These should be `false` by default mostly, I think.

assets/src/edit-story/elements/image/edit.js
miina: Definitely better and now it works, however, the experience while rotated feels very different:
cropping

Especially having the boundaries would be great. Would it be very complicated to make custom calculations for that? E.g. to check that the top left corner of the cropbox would not have higher value `x` and `y` coordinates than the visible element, etc.?

One alternative could be removing the rotation (setting it to 0) for the cropping time only, that would potentially make the element move quite a lot though and be annoying for the user.

Or we could leave it as-is for now and deal with the rotation boundaries later, as part of an improvement.

Thoughts?

dvoytenko: There's no good option here for now. The only good way to express it would be using the rotated boundary box like the red one here:

Rotated boundaries

That could only be done using a polygon (or a set of bounding lines). Unfortunately, `Moveable` only allows us to specify bounding region using vertical/horizontal lines. We have to then expand the box to fully fit our rotated rectangle (the blue lines). This ends up working like [this demo](https://drive.google.com/file/d/1NPgl0QjtKW-j2BuhVApxIUhVtoYof2Er/view?usp=sharing). I played around with it, and the UX is just disastrous. As a user it's really confusing why some of these boundaries pop up and when.

Perhaps we should if we could contribute the polygon boundaries to the `Moveable` project.

assets/src/edit-story/elements/image/edit.js
dvoytenko: done

assets/src/edit-story/elements/image/edit.js
dvoytenko: @miina I moved the movables into separate modules. PTAL.

assets/src/edit-story/elements/image/edit.js
miina: Hmm, maybe that's partly a product question in terms of what to do -- if we should support the rotated cropping "partially" since currently the polygon boundaries aren't supported.

To me it feels like from the product perspective we should (eventually) either have it working correctly and as expected or not support it at all, as opposed to having poor/confusing UX.

Maybe for now, we could leave it as-is and connect with the dev of Moveable on contributing to the library to support polygon boundaries.

There's also the (potentially annoying but simpler) alternative of setting the rotation angle to 0 for the time of cropping only. That might also be poor UX, however...

Thoughts?

cc @pbakaus

assets/src/edit-story/elements/image/editCropMovable.js
```diff
@@ -0,0 +1,102 @@
...
+				cropRef.current = [ tx, ty, resizeWidth, resizeHeight ];
+			} }
+			onResizeEnd={ () => {
+				cropBox.style.transform = '';
+				croppedImage.style.transform = '';
+				cropBox.style.width = '';
+				cropBox.style.height = '';
+				const [ tx, ty, resizeWidth, resizeHeight ] = cropRef.current;
+				cropRef.current = [ 0, 0, 0, 0 ];
+				if ( resizeWidth === 0 || resizeHeight === 0 ) {
```
miina: Could we instead limit the resizing within `onResize`, to ensure it doesn't get to 0? I suppose this should be done within the display mode `Moveable`, too. 

We can follow up on this within a separate PR though.

assets/src/edit-story/elements/image/edit.js
miina: LGTM!

assets/src/edit-story/app/index.js
miina: Pinging @barklund again in case that was missed.

miina:
LGTM, just waiting for feedback on the best practices for global style and thoughts on if it's OK to merge the rotation + missing boundaries as-is.

assets/src/edit-story/elements/image/edit.js
pbakaus: I'd say either:

1. disable rotation while cropping, or
2. implement perfectly (guides, snap points rotate as well)

I was already in touch with the Moveable guy on Twitter, shouldn't be hard to contribute.

assets/src/edit-story/elements/image/editCropMovable.js
dvoytenko: This serves a bit simpler purpose. I noticed that if I just click and release on the anchor, I get a sequence of events: `onResizeStart` and `onResizeEnd`. There's no `onResize` in between. The meaning of this simply that "no resize happened". Thus I simply ignore this sequence here. We had the same problem with the selection transformer and I fixed it elsewhere.

dvoytenko:
/to @barklund, could you PTAL the global styling question? Is this PR ok in that regard, or would you prefer some other way?

assets/src/edit-story/elements/image/editCropMovable.js
miina: Ah, now I understand, makes sense.

assets/src/edit-story/elements/image/edit.js
miina: Okay, in this case, we could merge it as it is and create a follow-up issue to contribute to Moveable to support polygon boundaries for making this complete.

miina:
@dvoytenko Just realized there's actually one more part missing: reflecting the changes in the front-end, too (adjusting the style / saving the necessary attributes in AMP HTML).

dvoytenko:
> @dvoytenko Just realized there's actually one more part missing: reflecting the changes in the front-end, too (adjusting the style / saving the necessary attributes in AMP HTML).

Didn't realize you've already submitted that pull request :) Mind if I follow up on this in a separate PR?

miina:
> Didn't realize you've already submitted that pull request :) Mind if I follow up on this in a separate PR?

Sure. Generally, it would be ideal to keep the editor and FE sync but at this stage of development it's still OK, will keep the PR more maintainable as well.

Created two follow-up issues for this ticket:
- https://github.com/ampproject/amp-wp/issues/4038 (FE implementation)
- https://github.com/ampproject/amp-wp/issues/4039 (Polygon boundaries)

Will approve this PR, we can still wait to address the `GlobalStyle` question, or address it post-merge.

assets/src/edit-story/app/index.js
swissspidy: I think I prefer the separation here, but if that's not considered a good practice we can change it in a follow-up PR.

swissspidy:
So cropping is only possible when the image isn't set as full-bleed?

assets/src/edit-story/app/index.js
barklund: This works for me. It's a bit of a special case as we're dealing with styling an external component, so this makes sense to me.

assets/src/edit-story/elements/image/editCropMovable.js
```diff
@@ -0,0 +1,102 @@
...
+} ) {
+	const moveableRef = useRef();
+	const cropRef = useRef( [ 0, 0, 0, 0 ] );
+
+	// Refresh moveables to ensure that the selection rect is always correct.
+	useEffect( () => {
+		if ( moveableRef.current ) {
+			moveableRef.current.updateRect();
+		}
+	} );
```
barklund: This could be rewritten to using `useLayoutEffect` - then you can remove the `if` as it's always run after the component has been rendered. But I'd like to see some dependencies - should it really be done on every render or only if something specific changes?

dvoytenko:
> So cropping is only possible when the image isn't set as full-bleed?

@swissspidy Correct. Cropping is not enabled in the fullbleed mode. However, panning and scaling are allowed.

assets/src/edit-story/elements/image/editCropMovable.js
dvoytenko: Removed `if`s. The `useEffect` works fine with this as well. It's actually a leftover from the time when these were combined together in a single module and were on/off selectively. Now that on/off control is done by the main component, so the refs can no longer be null.

Unfortunately, we've found that `updateRect()` is needed very often. On the plus side, however, it's very cheap if nothing really changed.

dvoytenko:
> Will approve this PR, we can still wait to address the `GlobalStyle` question, or address it post-merge.

@miina, @barklund has approved the `GlobalStyle` usage for this case.

PR #3998, opened 2019-12-20
Font picker

### LINKED ISSUE #3759: Font Picker
## Feature description

This ticket was taken from a row in this [spreadsheet](https://docs.google.com/spreadsheets/d/1nqJ-iOypTziWFjWBIBoJUc0LHW3gTeUnLkbjVH_eW4g/edit#gid=0). This ticket can be used to track research and work related to porting this existing feature from the "old" editor to the "new" editor. 

Font Picker:
- An autocomplete field used to pick font from list of defined google fonts. 
- The font picker simple implements an autocomplete library. The autocomplete library is overwritten to add a clear autocoplete field. The list of google fonts will need be hydrated to new component.

UX Required: Yes

---------------

## Acceptance criteria

* The existing font picker is ported to the new stories editor
* The font picker is integrated into the new [design](https://www.figma.com/file/Cd9iacVknHK62RO4pRwb68/Experimental?node-id=0%3A1)

1. A simple dropdown is displayed to select fonts (font family). (Additional functionality will be added later in new issues)
2. Ensure the dropdown is performant (previously we were loading 1000s of fonts)

### PHP 
* Abstract off current font code `AMP_Fonts`, so not linked to old implemenation found in `class-amp-story-legacy-post-type.php`
* Create a basic REST API to all font data into the stories editor. 
  * Only load new API if new editor is enabled. 
  * Make the API limited to only those who can editor a post can access. 
  * Set a limit to API of  1000. ( In the future we may wish to paginate results )
* Load all fonts into state when editor is loaded. It is likely that font data will be used. ( In future will may wish to look at lazy loading this data in for performance reasons ). 
* Load fonts from google fonts in Admin and front end, by looping through list of elements and exacting font family. 
### JS 
* Create a basic select ( dropdown ) menu UI element for panels. 
* Load all fonts into state when editor is init. 
* Port `maybeEnqueueFontStyle` function from old stories editor.
* Load font when text element is displayed. 
* Create a dropdown listing all fonts. 
* Create a new `FontProvider` to store state of fonts. 
* Change font weights list of avalible when font is selected. 

[PoC](https://github.com/ampproject/amp-wp/pull/3998)

## QA testing instructions

* 

## Demo

https://youtu.be/lglp7C7y8U8

## Changelog entry

*

spacedmonkey:
## Summary

Fixes #3759

## Checklist


assets/src/edit-story/app/story/storyProvider.js
```diff
@@ -66,6 +68,7 @@ function StoryProvider( { storyId, children } ) {
 	const setPropertiesById = useSetPropertiesById( { currentPageIndex, pages, setPages } );
 
 	useLoadStory( { storyId, pages, setPages, setTitle, setPostStatus, setPostAuthor, setSlug, setLink, setCurrentPageIndex, clearSelection } );
+	useLoadFonts( { fonts, setFonts } );
```
spacedmonkey: In the future, I think we should either lazy load this request or just load fonts that are in current use. 
Thoughts @barklund

assets/src/edit-story/elements/text/edit.js
```diff
@@ -123,6 +123,10 @@ function TextEdit( { content, color, backgroundColor, width, height, x, y, fontF
 		editor.current.focus();
 	}, [] );
 
+	useEffect( () => {
+		maybeEnqueueFontStyle( fontFamily, fonts );
```
spacedmonkey: Not sure if we need this here.

assets/src/edit-story/panels/font.js
```diff
@@ -26,27 +28,40 @@ function FontPanel( { selectedElements, onSetProperties } ) {
 		onSetProperties( state );
 		evt.preventDefault();
 	};
+
+	const fontWeights = [
```
spacedmonkey: Later, we should be move this settings somewhere else.

assets/src/edit-story/app/story/storyProvider.js
```diff
@@ -36,6 +37,7 @@ function StoryProvider( { storyId, children } ) {
 	// Don't update 1 of these in an effect based off another base variable.
 	// Only update these directly as a response to user or api interactions.
 	const [ pages, setPages ] = useState( [] );
+	const [ fonts, setFonts ] = useState( [] );
```
barklund: I don't think this belongs in the story provider. I think a new `FontProvider` would be a better choice for his logic. It would need to be a parent of the story provider in the layering in `<App>`.

Yes, the story provider needs to know which fonts (_font presets_) are used in the story, but it doesn't need to concern itself with which fonts are available nor how and when these fonts are loaded.

assets/src/edit-story/app/story/effects/useLoadFonts.js
```diff
@@ -0,0 +1,35 @@
...
+ * Internal dependencies
+ */
+import { useAPI } from '../../';
+
+function useLoadFonts( { fonts, setFonts } ) {
+	const { actions: { getAllFonts } } = useAPI();
+
+	useEffect( () => {
+		if ( fonts.length === 0 ) {
+			getAllFonts( {} ).then( ( loadedFont ) => {
```
barklund: These three lines could simply be:
```suggestion
			getAllFonts( {} ).then( setFonts );
```

assets/src/edit-story/panels/font.js
barklund: Wouldn't these (ideally) be on a per-font basis and come from the fonts array? Similar for styles?

assets/src/edit-story/panels/shared.js
```diff
@@ -76,6 +76,43 @@ InputGroup.defaultProps = {
...
+	label: PropTypes.string.isRequired,
+	value: PropTypes.any.isRequired,
+	options: PropTypes.array.isRequired,
+	onChange: PropTypes.func.isRequired,
+	postfix: PropTypes.string,
+	disabled: PropTypes.bool,
+};
+
+SelectMenu.defaultProps = {
+	postfix: '',
```
barklund: `disabled` should probably default to `false`.

spacedmonkey: Fixed https://github.com/ampproject/amp-wp/pull/3998/commits/614d307b959ee55c9dab20966e0300ed41afb789

assets/src/edit-story/elements/text/index.js
```diff
@@ -8,7 +8,7 @@ export { default as Save } from './save';
 
 export const defaultAttributes = {
 	fontFamily: 'Arial',
-	fontWeight: 'normal',
+	fontWeight: '400',
```
spacedmonkey: Font weights are now handled as numbers.

assets/src/edit-story/panels/font.js
```diff
@@ -11,49 +11,66 @@ import { useEffect, useState } from '@wordpress/element';
...
-			<InputGroup
-				type="text"
+			<SelectMenu
 				label="Font family"
+				options={ fonts }
 				value={ state.fontFamily }
-				isMultiple={ fontFamily === '' }
-				onChange={ ( value ) => setState( { ...state, fontFamily: value } ) }
+				onChange={ ( value ) => {
+					setState( { ...state, fontFamily: value, fontWeight: 400 } );
```
spacedmonkey: Reset back to 400, just in case.

miina:
From testing:
Currently, the when changing the font, the value in the display first seems to change to a default value (or something else) for a moment, and only then to the selected font. See flickering of the font here:
font-flicker

As can be seen, three different fonts are displayed in total: the first one, a second (perhaps the default), and only then the selected font.

assets/src/edit-story/panels/font.js
spacedmonkey: This is now fixed.

assets/src/edit-story/app/story/storyProvider.js
spacedmonkey: Font provider now added.

spacedmonkey:
> From testing:
> Currently, the when changing the font, the value in the display first seems to change to a default value (or something else) for a moment, and only then to the selected font. See flickering of the font here:
> font-flicker
> 
> As can be seen, three different fonts are displayed in total: the first one, a second (perhaps the default), and only then the selected font.

Thanks for the feedback @miina . I am still actively working on this PR, so you may want to pull as lots of changes. 

Any ideas on how we can improve how this looks. The issue is the browser is waiting for  font to download, so it change from font 1, to font 2 (broswer default) and to font 3 when font is downloaded. It is going to hard to know when this happens.

miina:
> Any ideas on how we can improve how this looks. The issue is the browser is waiting for font to download, so it change from font 1, to font 2 (broswer default) and to font 3 when font is downloaded. It is going to hard to know when this happens.

Hmm, probably the list itself will also display the preview of the actual font later, correct? I guess that will fix the issue since the chosen font would have already been pulled for displaying the font preview in the selection list.

swissspidy:
I noticed that too the last time I checked out this PR. Could we perhaps turn `maybeEnqueueFontStyle` into a hook that tracks whether the style is loaded? Similar to what this [`useScript`](https://usehooks.com/useScript/) hook is doing.

Then, `useRef` could be used to track the previous font, like this [`usePrevious`](https://usehooks.com/usePrevious/) example is doing.

dvoytenko:
> Any ideas on how we can improve how this looks. The issue is the browser is waiting for font to download, so it change from font 1, to font 2 (broswer default) and to font 3 when font is downloaded. It is going to hard to know when this happens.

We can use `document.fonts.load()` which is an async API? I think we can even do the text sampling to ensure that the right part of unicode table has been loaded.

spacedmonkey:
The issue here, is the object we are loading is CSS. We could track when this request for CSS has completed, but it not as easy to track all the font files have loaded correctly as this is dynamic, unless I am missing something here. 

Document.fonts.load seem very promising ( mind the pun there, as this API returns promises ). This [example](https://googlechrome.github.io/samples/font-face-set/) looks like the start of something.  I can't find details on browser support ( see [Can I use](https://caniuse.com/#feat=mdn-api_fontface_load) and [mozilla](https://developer.mozilla.org/en-US/docs/Web/API/FontFace/load) ) . 

Doing a little research, I would like to explore [webfontloader](https://github.com/typekit/webfontloader) ( [npm](https://www.npmjs.com/package/webfontloader) ), developed by Google and Typekit. This support font loading from Google fonts, Typekit, Fonts.com, Adobe Edge Web Fonts, Fontdeck and custom loading. Supporting custom and typekit was on the original roadmap, so this could be useful here. 

Personally, I think this is massively out of scope this ticket which is to make a simple dropdown, which this down. The font does load, on both the FE and BE. How fonts are loaded and preview is for the later implementation of search functionality IMO. 

 I worked on this PR, all of yesterday ( 6th of Jan ). Can everyone how is interested take a look again.

swissspidy:
Sure, we can tackle that in a separate issue, but it's valuable to brainstorm about it as it pops up 🙂 

Will have another look at the current state 👍

assets/src/edit-story/panels/font.js
swissspidy: Hmm I'd rather not reset if it can be avoided. Can the effect above just check whether the new font supports the existing font weight and only fall back to 400 if it doesn't?

includes/rest-api/class-amp-rest-fonts-controller.php
```diff
@@ -0,0 +1,256 @@
...
+		// Wrap the data in a response object.
+		$response = rest_ensure_response( $data );
+
+		/**
+		 * Filters a font returned from the API.
+		 *
+		 * Allows modification of the font right before it is returned.
+		 *
+		 * @param WP_REST_Response $response The response object.
+		 * @param Object $font The original comment object.
```
swissspidy:
```suggestion
		 * @param Object $font The original font object.
```

includes/rest-api/class-amp-rest-fonts-controller.php
```diff
@@ -0,0 +1,256 @@
...
+			$response->link_header( 'next', $next_link );
+		}
+
+		return $response;
+	}
+
+	/**
+	 * Prepares a single font output for response.
+	 *
+	 * @param Object          $font Comment object.
```
swissspidy:
```suggestion
	 * @param Object          $font Font object.
```

includes/rest-api/class-amp-rest-fonts-controller.php
```diff
@@ -0,0 +1,256 @@
...
+				],
+				'weights'   => [
+					'description' => __( 'Array of fallback fonts', 'amp' ),
+					'type'        => 'array',
+					'context'     => [ 'embed', 'view', 'edit' ],
+					'readonly'    => true,
+					'default'     => [ 'normal', 'bold', 'bolder', 'lighter' ],
+				],
+				'src'       => [
+					'description' => __( 'URL for the object font.', 'amp' ),
```
swissspidy:
```suggestion
					'description' => __( 'URL to enqueue the font', 'amp' ),
```

includes/rest-api/class-amp-rest-fonts-controller.php
spacedmonkey: 5 guesses where I copied and pasted this from...

includes/rest-api/class-amp-rest-fonts-controller.php
```diff
@@ -0,0 +1,256 @@
...
+	public function get_items_permissions_check( $request ) {
+		return current_user_can( 'edit_posts' );
+	}
+
+	/**
+	 * Checks if the font can be read.
+	 *
+	 * @param Object $font Font object.
+	 *
+	 * @return bool Whether the comment can be read.
```
swissspidy:
```suggestion
	 * @return bool Whether the font can be read.
```

Although we can probably just remove this method, and only add it back when we actually need per-font permission checks.

spacedmonkey: Yeah, I thought it might overkill. Not sure how per font permissions would work anyway, what if you edit a post by someone..

spacedmonkey: Removed in https://github.com/ampproject/amp-wp/pull/3998/commits/060b83b2257028b4e06c37bec91b905513ecbb6f

includes/rest-api/class-amp-rest-fonts-controller.php
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3998/commits/060b83b2257028b4e06c37bec91b905513ecbb6f

includes/rest-api/class-amp-rest-fonts-controller.php
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3998/commits/060b83b2257028b4e06c37bec91b905513ecbb6f

assets/src/edit-story/elements/text/util.js
```diff
@@ -44,3 +44,43 @@ export const getHandleKeyCommand = ( setEditorState ) => ( command, currentEdito
 	}
 	return 'not-handled';
 };
+
+/**
+ * Adds a <link> element to the <head> for a given font in case there is none yet.
```
spacedmonkey: Should this be moved the font provider?

swissspidy: Seems fitting 👍

includes/class-amp-story-post-type.php
```diff
@@ -323,6 +347,55 @@ public static function admin_enqueue_scripts( $hook ) {
 
 	}
 
+	/**
+	 * Load font from story data.
+	 *
+	 * @param WP_Post $post Post Object.
+	 */
+	public static function load_fonts( $post ) {
```
swissspidy: The downside of the batching here is that `maybeEnqueueFontStyle` can potentially unnecessarily  enqueue a font twice because of the non-matching style handles.

I think it's OK for now though as we will eventually move to a more sophisticated method for loading fonts anyway. Just wanted to write it down :-)

spacedmonkey: I could do something different in admin, so link tags are formatted with the correct id, so javascript doesn't load it. But honestly thought a couple small css calls was not a problem here.

assets/src/edit-story/elements/text/util.js
spacedmonkey: Done https://github.com/ampproject/amp-wp/pull/3998/commits/3eba1c0d98481e5ade47de8c69534961f918986b

assets/src/edit-story/panels/font.js
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3998/commits/1d13d7e12912b808fdf217ae55bdff3470d95547

assets/src/edit-story/app/font/fontProvider.js
```diff
@@ -0,0 +1,102 @@
...
+
+import useLoadFonts from './effects/useLoadFonts';
+import useLoadFontFiles from './actions/useLoadFontFiles';
+
+function FontProvider( { children } ) {
+	const [ fonts, setFonts ] = useState( [] );
+
+	useLoadFonts( { fonts, setFonts } );
+
+	const getFontByName = useCallback(
```
miina: Would a single function something like `getFontBy` make sense here instead of two separate?
E.g. `fonts.find( ( thisFont ) => thisFont[ key ] === value );`?

(Not a strong preference if it makes code much more readable as it is, feel free to ignore).

assets/src/edit-story/panels/shared.js
```diff
@@ -79,6 +79,45 @@ InputGroup.propTypes = {
...
+	disabled: false,
+};
+
+function SelectMenu( { label, options, value, onChange, postfix, disabled } ) {
+	return (
+		<Group disabled={ disabled }>
+			<Label>
+				{ label }
+			</Label>
+			<select
```
miina: Should we use styled-components here instead for consistency?

assets/src/edit-story/app/font/fontProvider.js
spacedmonkey: Not against this as removed repeated code. `getFontBy` could then just become an alias of `getFontByName` and getFontBySlug`.

assets/src/edit-story/panels/shared.js
spacedmonkey: This select is a temp element for now that soon will be replaced. I am not sure I see the point of styling these select nicely nice now.

includes/class-amp-fonts.php
```diff
@@ -0,0 +1,295 @@
...
+	public static function get_subsets() {
+		$subsets = [ 'latin', 'latin-ext' ];
+
+		/*
+		 * Translators: To add an additional character subset specific to your language,
+		 * translate this to 'greek', 'cyrillic', 'devanagari' or 'vietnamese'. Do not translate into your own language.
+		 */
+		$subset = _x( 'no-subset', 'Add new subset (greek, cyrillic, devanagari, vietnamese)', 'amp' );
+
+		if ( 'cyrillic' === $subset ) {
```
miina: Perhaps a `switch` here instead to make it more readable?

spacedmonkey: This is just a copy and paste from code found elsewhere  - https://github.com/ampproject/amp-wp/pull/3998/files/f9befc5462a684639b420aa7021963d24a1db5a0#diff-4c59a5c0c9c106525335ea3d5aa2a9c6L1224-L1234

includes/class-amp-story-legacy-post-type.php
```diff
@@ -1115,244 +1115,6 @@ public static function add_custom_stories_styles() {
 		self::enqueue_general_styles();
 	}
 
-	/**
```
miina: Just to confirm: the functionality of the legacy Stories editor didn't change with this change, correct? Looks like not, just ensuring that we're keeping the legacy editor as it was before.

includes/class-amp-story-post-type.php
swissspidy: Let's revisit this later if needed 👍

includes/class-amp-story-post-type.php
```diff
@@ -323,6 +347,55 @@ public static function admin_enqueue_scripts( $hook ) {
...
+		if ( $post_story_data ) {
+			foreach ( $post_story_data as $page ) {
+				foreach ( $page['elements'] as $element ) {
+					$font = AMP_Fonts::get_font( $element['fontFamily'] );
+
+					if ( $font && isset( $font['gfont'] ) && $font['gfont'] ) {
+						if ( isset( $g_fonts[ $font['name'] ] ) && ! in_array( $element['fontWeight'], $g_fonts[ $font['name'] ], true ) ) {
+							$g_fonts[ $font['name'] ][] = $element['fontWeight'];
+						} else {
+							$g_fonts[ $font['name'] ][] = $element['fontWeight'];
```
miina: The `if` and `else` contents seem to be identical (both `$g_fonts[ $font['name'] ][] = $element['fontWeight'];`).

assets/src/edit-story/panels/shared.js
miina: Agreed that no point stying nicely, was thinking rather using something just like this (since it's used elsewhere everywhere as well, even if there's no styling at all):

`const Select = styled.select``;` or 
```
const Select = styled.select`
width: 100%;
`;
```

includes/class-amp-fonts.php
miina: OK, is this part still WIP and going to change? Let's keep as is if that's the case.

assets/src/edit-story/panels/shared.js
spacedmonkey: Changed in https://github.com/ampproject/amp-wp/pull/3998/commits/6f37350ad93201af7003de73c8a31cfb068684ef

includes/class-amp-fonts.php
spacedmonkey: This is still a PoC so it can be changed if you want.

includes/class-amp-story-legacy-post-type.php
spacedmonkey: Just moved some logic around to make it share about. Should be no breaking changes to legacy CPT.

includes/rest-api/class-amp-rest-fonts-controller.php
```diff
@@ -0,0 +1,242 @@
...
+
+	/**
+	 * Checks if a given request has access to get fonts.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 *
+	 * @return true|WP_Error True if the request has read access, WP_Error object otherwise.
+	 */
+	public function get_items_permissions_check( $request ) {
+		return current_user_can( 'edit_posts' );
```
miina: Should this perhaps check the read access instead of edit?
The comment above mentions `read` access, too.

spacedmonkey: The function returns if a user / request can read the fonts. The check if they can edit posts.

swissspidy: This way the REST API endpoint is only accessible if you can edit posts, which makes sense to me.

includes/rest-api/class-amp-rest-fonts-controller.php
```diff
@@ -0,0 +1,242 @@
...
+	/**
+	 * Gets a collection of fonts.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 *
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	public function get_items( $request ) {
+		$fonts       = AMP_Fonts::get_fonts();
+		$total_posts = count( $fonts );
```
miina: I suppose this should be `$total_fonts`?

Is the check for that necessary for fonts? Could you explain the logic behind allowing requesting the fonts using `per_page` and `page` params, in which cases are these being used in the editor currently? Is this for the future, for loading the fonts in batches in the dropdown?

includes/rest-api/class-amp-rest-fonts-controller.php
```diff
@@ -0,0 +1,242 @@
...
+
+		if ( $page > 1 ) {
+			$prev_page = $page - 1;
+
+			if ( $prev_page > $max_pages ) {
+				$prev_page = $max_pages;
+			}
+
+			$prev_link = add_query_arg( 'page', $prev_page, $base );
+			$response->link_header( 'prev', $prev_link );
```
miina: Curious to know what is the plan to use the prev and next for pages of fonts.

includes/rest-api/class-amp-rest-fonts-controller.php
spacedmonkey: Correct, this variable should be changed. I have written the API should it can be changed to be paginated easily later, if we want to lazy load in fonts.

includes/rest-api/class-amp-rest-fonts-controller.php
```diff
@@ -0,0 +1,242 @@
...
+		if ( $this->schema ) {
+			return $this->add_additional_fields_schema( $this->schema );
+		}
+		$schema       = [
+			'$schema'    => 'http://json-schema.org/draft-04/schema#',
+			'title'      => 'font',
+			'type'       => 'object',
+			'properties' => [
+				'name'      => [
+					'description' => __( 'The title for the status.', 'amp' ),
```
miina: Should this be for `font` and not `status`?

includes/rest-api/class-amp-rest-fonts-controller.php
```diff
@@ -0,0 +1,242 @@
...
+					'type'        => 'string',
+					'context'     => [ 'embed', 'view', 'edit' ],
+					'readonly'    => true,
+				],
+				'slug'      => [
+					'description' => __( 'An alphanumeric identifier for the status.', 'amp' ),
```
miina: Should this be for `font` and not `status`?

includes/rest-api/class-amp-rest-fonts-controller.php
spacedmonkey: Just following the format of other apis here. We may never use this functionality.

includes/rest-api/class-amp-rest-fonts-controller.php
miina: Yes, to think of it then it's only used in the admin anyway, the endpoint.

includes/rest-api/class-amp-rest-fonts-controller.php
miina: The other apis where pagination is not used with prev-next links don't necessarily include this, e.g. autosaves, post types, taxonomies, etc. Since it's not foreseen in the near future that we'd need it, perhaps we can remove it for now and add back if it should become necessary for some reason?

dvoytenko:
Certainly ok to defer this to further PRs, but few replies:

> Document.fonts.load seem very promising ( mind the pun there, as this API returns promises ). This [example](https://googlechrome.github.io/samples/font-face-set/) looks like the start of something. I can't find details on browser support ( see [Can I use](https://caniuse.com/#feat=mdn-api_fontface_load) and [mozilla](https://developer.mozilla.org/en-US/docs/Web/API/FontFace/load) ) .

The support is pretty good. Definitely covers Chrome/Safari/Firefox. And I think it'd be reasonable to simply fallback to immediately applying font and taking a hit on it jumping in UI a bit. E.g. something like this:

```
function waitForFontFaceToLoad(fontFace, fontWeight, ..., sample) {
  if (!document.fonts) {
   return Promise.resolve();
  }
  const font = `${fontFace} ...`;
  return document.fonts.load(font, sample).catch(() => {/*ignore*/});
}
```

> Doing a little research, I would like to explore [webfontloader](https://github.com/typekit/webfontloader) ( [npm](https://www.npmjs.com/package/webfontloader) ), developed by Google and Typekit. This support font loading from Google fonts, Typekit, Fonts.com, Adobe Edge Web Fonts, Fontdeck and custom loading. Supporting custom and typekit was on the original roadmap, so this could be useful here.

I think even using these APIs, you'd still need to use `document.fonts` to be certain the font is ready to be used b/c fonts are loaded lazily when corresponding styles are resolved in the active connected DOM.

includes/rest-api/class-amp-rest-fonts-controller.php
spacedmonkey: Removed https://github.com/ampproject/amp-wp/pull/3998/commits/7764e31e9501a0756f280850ee8602703b54f949

includes/rest-api/class-amp-rest-fonts-controller.php
spacedmonkey: Changed in https://github.com/ampproject/amp-wp/pull/3998/commits/7a80cd18b87d6fab2c775a0f037d366996fc49cb

includes/rest-api/class-amp-rest-fonts-controller.php
spacedmonkey: Changed in https://github.com/ampproject/amp-wp/pull/3998/commits/7a80cd18b87d6fab2c775a0f037d366996fc49cb

includes/rest-api/class-amp-rest-fonts-controller.php
spacedmonkey: Changed in https://github.com/ampproject/amp-wp/pull/3998/commits/b6bffec9e1b0972aff684ead1b8ed9062cd40897

assets/src/edit-story/app/font/fontProvider.js
spacedmonkey: Changed in https://github.com/ampproject/amp-wp/pull/3998/commits/eac50b04f7e3d1829eaefe62772008b5f56ad8d2

includes/class-amp-fonts.php
spacedmonkey: Changed in https://github.com/ampproject/amp-wp/pull/3998/commits/f9ed841c56fb16fa5b351d3edda1d6cf66668837

includes/class-amp-story-post-type.php
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3998/commits/8bbb28b930ae5f0a2b85bf7f8f1acb7d5c13100d

assets/src/edit-story/app/index.js
```diff
@@ -33,10 +34,12 @@ function App( { config } ) {
...
-								<GlobalStyle />
-								<CropMoveableGlobalStyle />
-								<Layout />
-								<Popover.Slot />
+								<FontProvider>
+								  <GlobalStyle />
+								  <CropMoveableGlobalStyle />
+								  <Layout />
+								  <Popover.Slot />
+                </FontProvider>
```
swissspidy: The indentation in this area is a bit off (spaces vs tabs)

assets/src/edit-story/elements/text/util.js
```diff
@@ -44,3 +44,11 @@ export const getHandleKeyCommand = ( setEditorState ) => ( command, currentEdito
 	}
 	return 'not-handled';
 };
+
+export const generateFontFamily = ( fontFamily, fontFallback ) => {
+	let fontFamilyDisplay = fontFamily ? `${ fontFamily }` : null;
+	if ( fontFallback && fontFallback.length ) {
+		fontFamilyDisplay += `,${ fontFallback.join( `,` ) }`;
```
swissspidy: Looks like a good utility function to add 1-2 tests for.

If `fontFamily` is `null` and `fontFallback` is `[ 'Arial', 'Helvetica' ]`, this function will return `null,Arial,Helvetica`.

spacedmonkey: When would this happen? Font family and font fallback are set the same time.

swissspidy: When someone accidentally passes a falsy value for `fontFamily`? Might not happen in our controlled environment, but I was just looking at this function in isolation.

assets/src/edit-story/app/index.js
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/3998/commits/3ee619201154af1ef3c75c4dd1f55fcc1c3b2cde

assets/src/edit-story/elements/text/util.js
spacedmonkey: Any recommendations on what the logic should look like?

swissspidy: Not having that leading comma when `fontFamilyDisplay` is `null`.

spacedmonkey: Does this work for you @swissspidy - https://github.com/ampproject/amp-wp/pull/3998/commits/c326c9ae1bc59423e53f3a9fb3c6a56fc12e7570

swissspidy: Yeah it does 👍

PR #4003, opened 2019-12-20
Fix conversion of forms with relative action URLs

westonruter:
## Summary

The subscription `form` in Jetpack [uses](https://github.com/Automattic/jetpack/blob/e67bea661ea10a2396bf6efd3a484e9c274a1e2a/modules/subscriptions/views.php#L329) an `action` of `#`:

```html
<form action="#" method="post" accept-charset="utf-8" id="subscribe-blog-blog_subscription-2">
```

When the form is being displayed on a URL like `https://example.org/about/` it is getting converted by the form sanitizer as:

```html
<form action-xhr="//example.org?_wp_amp_action_xhr_converted=1#" method="post" accept-charset="utf-8" id="subscribe-blog-blog_subscription-2" target="_top">
```

This is wrong. It should instead be:

```html
<form action-xhr="//example.org/about/?_wp_amp_action_xhr_converted=1#" method="post" accept-charset="utf-8" id="subscribe-blog-blog_subscription-2" target="_top">
```

This PR fixes the sanitization of `action` attribute containing values such as:

* `../`
* `?foo=bar`
* `#foo`

Issue discovered in a [WordPress.org support topic](https://wordpress.org/support/topic/jetpack-subscribe-to-blog-widget-not-working/#post-12259691).

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-form-sanitizer.php
+++ b/includes/sanitizers/class-amp-form-sanitizer.php
@@ -67,9 +67,21 @@ class AMP_Form_Sanitizer extends AMP_Base_Sanitizer {
 				$action_url = esc_url_raw( '//' . $_SERVER['HTTP_HOST'] . wp_unslash( $_SERVER['REQUEST_URI'] ) );
 			} else {
 				$action_url = $node->getAttribute( 'action' );
-				// Check if action_url is a relative path and add the host to it.
+
+				// Handle relative URLs.
 				if ( ! preg_match( '#^(https?:)?//#', $action_url ) ) {
-					$action_url = esc_url_raw( '//' . $_SERVER['HTTP_HOST'] . $action_url );
+					$schemeless_host = '//' . $_SERVER['HTTP_HOST'];
+					if ( '?' === $action_url[0] || '#' === $action_url[0] ) {
+						// For actions consisting of only a query or URL fragment, include the schemeless-host and the REQUEST URI of the current page.
+						$action_url = '//' . $_SERVER['HTTP_HOST'] . wp_unslash( $_SERVER['REQUEST_URI'] ) . $action_url;
+					} elseif ( '.' === $action_url[0] ) {
+						// For actions consisting of relative paths (e.g. '../'), prepend the schemeless-host and a trailing-slashed REQUEST URI.
+						$action_url = '//' . $_SERVER['HTTP_HOST'] . trailingslashit( wp_unslash( $_SERVER['REQUEST_URI'] ) ) . $action_url;
+					} else {
+						// Otherwise, when the action URL includes an absolute path, just append it to the schemeless-host.
+						$action_url = $schemeless_host . $action_url;
+					}
+					$action_url = esc_url_raw( $action_url );
 				}
 			}
 			$xhr_action = $node->getAttribute( 'action-xhr' );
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -67,9 +67,21 @@ public function sanitize() {
...
+				// Handle relative URLs.
 				if ( ! preg_match( '#^(https?:)?//#', $action_url ) ) {
-					$action_url = esc_url_raw( '//' . $_SERVER['HTTP_HOST'] . $action_url );
+					$schemeless_host = '//' . $_SERVER['HTTP_HOST'];
+					if ( '?' === $action_url[0] || '#' === $action_url[0] ) {
+						// For actions consisting of only a query or URL fragment, include the schemeless-host and the REQUEST URI of the current page.
+						$action_url = '//' . $_SERVER['HTTP_HOST'] . wp_unslash( $_SERVER['REQUEST_URI'] ) . $action_url;
+					} elseif ( '.' === $action_url[0] ) {
+						// For actions consisting of relative paths (e.g. '../'), prepend the schemeless-host and a trailing-slashed REQUEST URI.
+						$action_url = '//' . $_SERVER['HTTP_HOST'] . trailingslashit( wp_unslash( $_SERVER['REQUEST_URI'] ) ) . $action_url;
```
pierlon: I don't think the trailing slash should always be added. Taking the action `.foo` for example, depending on the server configuration `.foo` could be a file while `.foo/` could be a directory.

westonruter: Correct, but the trailing slash is being added after the path in order to append the relative segment. For example, if the user is currently on `/about/team` and there is a form that has an action of `../` the resulting `action` URL path needs to be `/about/team/../` not `/about/team../`. Does that make sense?

pierlon: Yes, thanks for the clarification.

pierlon:
Testing instructions:

1. Create a post with a Custom HTML block filled with the following HTML:

    ```html
    <form method="post" action="#"></form>
    ``` 

2. On the AMP page, verify that the `action` attribute has been converted to `action-xhr`, and the form element now looks like:

    ```html
    <form method="post" action-xhr="//example.com/about/?amp&_wp_amp_action_xhr_converted=1#" ...>...</form>
    ```
     _(where `example.com/about/?amp` is the post URL)_

csossi:
Verified in QA

image

includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -67,9 +67,21 @@ public function sanitize() {
...
-				// Check if action_url is a relative path and add the host to it.
+
+				// Handle relative URLs.
 				if ( ! preg_match( '#^(https?:)?//#', $action_url ) ) {
-					$action_url = esc_url_raw( '//' . $_SERVER['HTTP_HOST'] . $action_url );
+					$schemeless_host = '//' . $_SERVER['HTTP_HOST'];
+					if ( '?' === $action_url[0] || '#' === $action_url[0] ) {
+						// For actions consisting of only a query or URL fragment, include the schemeless-host and the REQUEST URI of the current page.
+						$action_url = $schemeless_host . wp_unslash( $_SERVER['REQUEST_URI'] ) . $action_url;
+					} elseif ( '.' === $action_url[0] ) {
```
westonruter: Note: the `'.'` check was not correct here. See https://github.com/ampproject/amp-wp/pull/4250#pullrequestreview-355559313

PR #3954, opened 2019-12-17
Handle sanitization of security, marginwidth & marginheight iframe attributes

pierlon:
## Summary

When sanitizing iframes:

- Remove the obsolete `security` attribute
- Remove the obsolete `marginwidth` and `marginheight` attributes if their values are `0`.

Fixes #3950.

## Checklist

- [x] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [x] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [x] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

PR #4004, opened 2019-12-21
Don't mock HTTP requests for the external-http testsuite

pierlon:
## Summary

While working on #3954, [an error](https://travis-ci.org/ampproject/amp-wp/jobs/626452964#L1069) with the Crowdsignal embed test surfaced (not pertaining to the aforementioned PR). That error should not have occurred as it would receive a mocked HTTP response body to sanitize, but due to a mishandled case in WP < 5.1 the HTTP request was not mocked (fixed in #3956). Because of this, it was discovered that Crowdsignal's embed response had been changed, causing the assertion of the expected HTML to fail.

This has brought into light the possibility of the other supported embeds returning HTML being unaccounted for and could lead to broken reports of them, while we are happily contented with the passing tests.

This PR seeks to change the way how embeds are tested by adding a new PHPUnit testsuite called `external-http`, which will allow the HTTP requests of these embeds to pass through unfiltered and allow for a true comparison of the expected HTML.

## Checklist


tests/php/test-amp-image-dimension-extractor.php
```diff
@@ -5,14 +5,6 @@
  * @package AMP
  */
 
-define( 'AMP_IMG_DIMENSION_TEST_INVALID_FILE', dirname( __FILE__ ) . '/assets/not-exists.png' );
```
pierlon: I've refactored the `external-http` group of tests to their own class called `AMP_Image_Dimension_Extract_Download_Test`, which makes it easier to add and remove from test suites.

tests/php/test-class-amp-imgur-embed-handler.php
```diff
@@ -69,18 +78,18 @@ public function get_conversion_data() {
 			],
 
 			'url_simple'      => [
-				'https://imgur.com/f462IUj' . PHP_EOL,
```
pierlon: This endpoint throws a `404` so I've updated it.

tests/php/test-amp-crowdsignal-embed-handler.php
```diff
@@ -82,6 +82,10 @@ public function test_conversion( $url, $expected, $oembed_response ) {
 		add_filter(
 			'pre_http_request',
 			static function ( $pre, $r, $request_url ) use ( $oembed_response ) {
+				if ( in_array( 'external-http', $_SERVER['argv'], true ) ) {
```
pierlon: This would check if the `external-http` testsuite is being run, and to act accordingly. Cleaner solutions for this are welcomed.

pierlon:
:warning: Rebasing with latest changes from `develop`.

tests/php/test-amp-crowdsignal-embed-handler.php
westonruter: The ideal seems to be to obtain the name of the test suite bring run via some phpunit API. I'm not seeing this, however. Perhaps @schlessera is familiar with one.

pierlon: Due to how PHPUnit is designed there is no way of dynamically adding or removing tests at runtime, As far as I am concerned.

schlessera: Maybe you could use the `@require` annotation to do this.
Not tested, but it might work to use something like `@require function do_external_http_request` and then conditionally define the `do_external_http_request()` function during bootstrap or in a listener...

See https://phpunit.readthedocs.io/en/8.4/incomplete-and-skipped-tests.html#incomplete-and-skipped-tests-requires-tables-api

pierlon: Using a listener wouldn't be possible as the signature of  the interface to implement changes between each major PHPUnit release, so I'll opt for the defining the function when bootstrapping.

pierlon: Using the `@requires` annotation wouldn't allow for the flexibility of mocking a HTTP request during the test. The aim is not having to duplicate the logic just to not mock the request and maintain both instances.

pierlon:
:warning: Rebasing with latest changes from `develop` to resolve merge conflict.

PR #4013, opened 2019-12-27
A floating scaling panel

### LINKED ISSUE #3905: Floating edit panels
## Feature description

Currently the edit panels are used in the sidebar on the right. The additional proposal is to extend this tool to also implement edit panels that float next to the elements on the canvas. This has first been raised for the cropped image scaling editor in #3895. However, there are several similar panels planned in the near future for other elements as well, such as text editor.

Such panels would only be shown in the edit mode and would be docked either under the element or to the side.

---------------

## Acceptance criteria

* The extension to the architecture of edit panels is proposed.

## Implementation brief

* See POC in #4013.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

dvoytenko:
## Summary

Fixes #3905

A lot to discuss in this pull request:

1. It builds heavily on #3957. And it further proposes to use a separate overlaying layer for all such tools: moveables, floating panels, etc.
2. Without overlay, the moveables or floating panels could get obscured by other elements, which would not work for edit mode, selection, transformers, etc.
3. It's still convenient to keep "edit mode" code close to the element's `edit.js`. But we could externalize it similar to the edit panels, but we'd need a wrapper element for position/size, which we discussed elsewhere already.
4. There's a nuance with ongoing cropping. If the element's bottom boundary is used to crop, the floating panel is unaligned with the bottom until cropping is complete. In practice this doesn't seem [too bad](https://user-images.githubusercontent.com/726049/71532545-b684e180-28a8-11ea-83e4-3bca1c40f6a9.gif). But if we want to synchronize a Moveable and the positioning of the floating panel, they may need to stay close to each other to exchange the transforms.

## Checklist


dvoytenko:
This is now ready for review.

assets/src/edit-story/elements/image/scalePanel.js
```diff
@@ -0,0 +1,86 @@
...
+`;
+
+const Range = styled.input`
+  flex: 1 1;
+  margin: 4px;
+  min-width: 100px;
+  cursor: pointer;
+`;
+
+const ResetButton = styled.button`
```
barklund: Is there no existing "standard" button for this? Is it a new type of standard button to be used elsewhere?

assets/src/edit-story/elements/image/scalePanel.js
```diff
@@ -0,0 +1,86 @@
...
+  border-radius: 4px;
+
+  display: flex;
+  flex-direction: row;
+  flex-wrap: nowrap;
+  align-items: center;
+  padding: 0 4px;
+`;
+
+const Range = styled.input`
```
barklund: Nitpick: Maybe move non-dynamic attributes to `.attrs()` here (and remove them from the JSX):

```suggestion
const Range = styled.input.attrs( {
	type: "range",
	min: 100,
	max: MAX_SCALE,
	step: 10,
} )`
```

assets/src/edit-story/elements/image/scalePanel.js
```diff
@@ -0,0 +1,86 @@
...
+  font-size: 10px;
+  color: rgba(255, 255, 255, 0.3);
+  background: rgba(255, 255, 255, 0.2);
+  border-radius: 4px;
+  border: none;
+`;
+
+function ScalePanel( { setProperties, width, height, x, y, scale } ) {
+	return (
+		<InOverlay zIndex={ 11 }>
```
barklund: That's a weirdly specific magic value. Can `InOverlay` maybe export some suggested `zIndex` values?

assets/src/edit-story/theme.js
```diff
@@ -17,6 +17,7 @@ const theme = {
 			v3: '#242A3B',
 			v4: '#2F3449',
 			v5: '#575D65',
+			v6: '#07080C',
```
barklund: I feel like at least one other PR has a `bg.v6` value (probably different from this) clearly illustrating the fragility of this scheme.

assets/src/edit-story/elements/image/scalePanel.js
dvoytenko: Not sure if there's much more here than `<button>Reset</button>` at this time. The main "commons" candidate is `text-transform: uppercase`. But we might want to eventually apply to most of the editor buttons. All other styles are, for now, very tightly applied to the the "scale" panel and I've not yet seen any panel similar to it in designs.

assets/src/edit-story/elements/image/scalePanel.js
dvoytenko: Makes sense. I applied this suggestion. There might be a time when we may calculate the max scale, but for now this is better.

assets/src/edit-story/theme.js
dvoytenko: Let's see who gets to merge first :)

assets/src/edit-story/elements/image/scalePanel.js
dvoytenko: Yeah. We typically do `zIndex` constants. So I redid this similarly here. LMK if it works.

Issue #3994, opened 2019-12-19
Discovery: cancelation stack for Esc and click-outside


dvoytenko:
## Feature description

The "canceling" interactions seem to be fairly complex and need to be clarified in the UX. As observed in the Google Slides:

* _Both_ <kbd>esc</kbd> and click-outside in the edit mode exit the edit mode, but keep the selection unchanged on the edited element. In either case the edited state is "committed".
* _Both_ <kbd>esc</kbd> and click-outside in the display mode unselect the selected element.
* The <kbd>esc</kbd> cancels the transforming operation (drag, rotation, resize). The edited state is "reset" back to pre-transformation.

Technically, we could potentially get away with scoping <kbd>esc</kbd> to the edit elements and `event.stopPropagation()`. With "click-outside" it might be more challenging and we may have to find a way to use a single canvas-global click listener for this.

/cc @barklund @miina @samitron7 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Clarify UX
* Prototype technical implementation

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

PR #4018, opened 2019-12-31
Selection canvas with lasso

### LINKED ISSUE #3906: Lasso selection
## Feature description

This is a classical laso selection mechanism where a user can start dragging with the pointer down and on pointer-up select 1-N elements that overlap with the laso's area. This issue assumes the multi-selection is already (at least) in the execution phase.

Some nuances:

1. Any part of unobscured canvas area can be used to start laso.
2. We may or may not want to allow Laso to be cancelable with ESC.
3. (Controversial part) The "background" elements do not participate in laso, but can be used to start it.
4. We may or may not be able (or want) to use `Moveable` for this. I'm not sure `Moveable` by default provides drag-to-resize ability.

---------------

## Acceptance criteria

* The laso implementation architecture is proposed.

## Implementation brief

* See #4018
* Pointer events are handled on the canvas' background
* The special overlay layer is used to draw the laso rectangle
* The laso rectange coordinates are recalculated to the page's coordinates
* A simple element lookup finds all intersecting elements and selects them

TODO:

- [ ] Background elements cannot participate in the laso selection, otherwise nothing can ever be selected this way.
- [ ] Possibly similar behavior should apply to fullscreen elements

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

dvoytenko:
## Summary

Fixes #3906

If we go ahead with this pull request, it will supersede #3957 and #4010.

This is a big change, but it also addresses numerous questions:

1. It implements laso selection.
2. It consolidates more selection logic in fewer places. Though there are still several levels split between `CanvasContext`, `SelectionCanvas`, and `Element`.
3. It reduces number of bugs related to selection sequence of mouse events.
4. It aligns editing and selection, as well as "editing cancel" with "selection cancel". This should help a lot with #3994.
5. It moves positioning, sizing, rotation, selection from `Display` and `Edit` components to the single wrapper element. Note that not all elements are migrated yet. But this significantly reduces amount of code duplication for these features.

TODO:

- [x] Finish element migration for position/size/rotation/selection.

## Checklist


miina:
@dvoytenko Haven't reviewed this yet but I suppose this would also make #3909 obsolete?

miina:
> Haven't reviewed this yet but I suppose this would also make #3909 obsolete?

Ok, looks like this PR is for the selection part only, and not the moving itself.

dvoytenko:
> > Haven't reviewed this yet but I suppose this would also make #3909 obsolete?
> 
> Ok, looks like this PR is for the selection part only, and not the moving itself.

@miina, yes, it builds on similar approach, but takes it a lot deeper.

dvoytenko:
> > > Haven't reviewed this yet but I suppose this would also make #3909 obsolete?
> > 
> > 
> > Ok, looks like this PR is for the selection part only, and not the moving itself.
> 
> @miina, yes, it builds on similar approach, but takes it a lot deeper.

It doesn't change moving logic, but does place moveables all together in an overlay.

assets/src/edit-story/components/canvas/canvasProvider.js
```diff
@@ -20,26 +33,67 @@ function CanvasProvider( { children } ) {
...
+					y <= ly + lh &&
+					ly <= y + height
+				);
+			} ).map( ( { id } ) => id );
+		setSelectedElementIds( newSelectedElementIds );
+	}, [ currentPage, setSelectedElementIds ] );
+
+	// Reset editing mode when selection changes.
+	useEffect( () => {
+		if ( editingElement &&
```
miina: In which case would it be true? Meaning, in which case would an element be in edit mode and multiple elements being selected at the same time?

assets/src/edit-story/components/canvas/selectionCanvas.js
```diff
@@ -0,0 +1,149 @@
...
+ */
+import { useStory } from '../../app';
+import useCanvas from '../canvas/useCanvas';
+import withOverlay from '../overlay/withOverlay';
+import InOverlay from '../overlay';
+
+const LasoMode = {
+	OFF: 0,
+	ON: 1,
+	MAYBE: 2,
```
miina: What's the "maybe" for?

assets/src/edit-story/components/canvas/selectionCanvas.js
```diff
@@ -0,0 +1,149 @@
...
+		laso.style.left = `${ x }px`;
+		laso.style.top = `${ y }px`;
+		laso.style.width = `${ w }px`;
+		laso.style.height = `${ h }px`;
+		laso.style.display = lasoMode === LasoMode.ON ? 'block' : 'none';
+	};
+
+	const onMouseDown = ( evt ) => {
+		clearSelection();
+		clearEditing();
```
miina: In case of an image being edited, I wonder if this should clear the selection only once the mouse is already moving? Considering that just clicking shouldn't deselect an image. Anyway -- that's probably something to tackle in a separate issue.

assets/src/edit-story/components/canvas/selectionCanvas.js
```diff
@@ -0,0 +1,149 @@
...
+		const overlay = overlayRef.current;
+		let offsetX = 0,
+			offsetY = 0;
+		for ( let offsetNode = overlay; offsetNode; offsetNode = offsetNode.offsetParent ) {
+			offsetX += offsetNode.offsetLeft;
+			offsetY += offsetNode.offsetTop;
+		}
+		const x = evt.pageX - offsetX;
+		const y = evt.pageY - offsetY;
+		offsetRef.current = [ offsetX, offsetY ];
```
miina: I'm likely missing something but couldn't we use just the start and the end of the mouse event here instead of the overlay element offset?

assets/src/edit-story/elements/image/display.js
```diff
@@ -29,20 +27,16 @@ const Img = styled.img`
 	${ ImageWithScale }
 `;
 
-function ImageDisplay( { id, src, origRatio, width, height, x, y, scale, focalX, focalY, rotationAngle, isFullbleed, forwardedRef } ) {
-	const elementProps = {
```
miina: Probably similar changes should be for other elements as well then, at least the Text? Not to have duplicate attributes on the element itself.

assets/src/edit-story/components/canvas/selectionCanvas.js
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { useStory } from '../../app';
+import useCanvas from '../canvas/useCanvas';
+import withOverlay from '../overlay/withOverlay';
+import InOverlay from '../overlay';
+
+const LasoMode = {
```
miina: Should probably by "lasso" instead of "laso"?

assets/src/edit-story/elements/image/display.js
miina: Just noticed that you mentioned that not all the elements have been migrated yet.

miina:
A lot of great changes -- definitely in favor of having the attributes in the wrapper and the element just filling in the available space.

Left some questions/notes.

assets/src/edit-story/components/canvas/canvasProvider.js
dvoytenko: The selection and editing are managed on different levels: the selection is managed by the reducer model in the `StoryProvider` and the edit mode is managed in the `Canvas` below it. I thought it was very likely that a selection could change via story model action (`setSelectedElementIds` method), while edit element remains unchanged. In reality, however, edit element must always also be selected one and it makes no sense to have an edit element for multiple selections. So this effect is a basically catch-all way to exit edit mode when the selection changes.

assets/src/edit-story/components/canvas/selectionCanvas.js
dvoytenko: The sequence we are looking for is pointer down -> move -> up. So the current algo is:

1. Ignore all "move" if there pointer is not down.
2. Once "down" arrives, we are still not sure if we are in the laso mode until we see some "move" events. Switch to MAYBE mode.
3. Once "move" arrives and we're in MAYBE mode, switch to ON mode.
4. Once "up" arrives, switch to OFF mode. If selection is valid, select the intersecting elements.

Further with possible drift we may, at some point, want to introduce a "move" threshold to control for false positives, e.g. how `Moveable` does. But for now it didn't seem necessary to me.

assets/src/edit-story/components/canvas/selectionCanvas.js
dvoytenko: Do you mean #3994? If so - yes, this would be trivially solvable here. But I opted to not do it yet since there might be some additional questions with <kbd>esc</kbd> processing and we already have a special issue to look into that.

assets/src/edit-story/components/canvas/selectionCanvas.js
dvoytenko: Hmm. Not sure how. The `pageX/Y` are relative to the document. And I need to calculate coordinates relative to the canvas. Otherwise I'd not know how to position the laso element.

assets/src/edit-story/elements/image/display.js
dvoytenko: Yep. Wanted to first see if there were any objections to this approach. Will do migration before merging of course.

assets/src/edit-story/components/canvas/selectionCanvas.js
dvoytenko: :)

dvoytenko: Done

assets/src/edit-story/components/canvas/selectionCanvas.js
dvoytenko: Do you mean #3994? If so - yes, this would be trivially solvable here. But I opted to not do it yet since there might be some additional questions with esc processing and we already have a special issue to look into that.

assets/src/edit-story/components/canvas/selectionCanvas.js
miina: Yep, makes sense, perhaps it would be good to leave a comment here that "maybe" is for when the mouse is down but no move events have happened yet.

miina:
Generally looks good to me, works well, and fixes some annoying issues (such as elements sticking to the mouse which happens only to me).

Requested additional review from @swissspidy since it's quite a large PR, otherwise, just the implementation for the other elements is missing.

assets/src/edit-story/components/canvas/selectionCanvas.js
```diff
@@ -0,0 +1,149 @@
...
+	};
+
+	const onMouseUp = ( ) => {
+		if ( lassoModeRef.current === LassoMode.ON ) {
+			const [ ox, oy, width, height ] = getLassoBox();
+			const x = ox - pageContainer.offsetLeft;
+			const y = oy - pageContainer.offsetTop;
+			clearSelection();
+			clearEditing();
+			selectIntersection( { x, y, width, height } );
```
swissspidy: Should the selection perhaps continuously update while drawing the lasso? That's what Keynote does, for example. Makes it easier to see which elements one is going to end up selecting.

dvoytenko: I'd say "no". Because somewhere, I believe, we wanted <kbd>esc</kbd> to be able to cancel lasso once we support keyboard events. Partially related to #3994. But /cc @samitron7 to clarify.

dvoytenko:
@miina @swissspidy ptal: other elements are done.

assets/src/edit-story/elements/square/display.js
```diff
@@ -8,40 +8,30 @@ import styled from 'styled-components';
...
-	${ ElementWithRotation }
+	${ ElementFillContent }
 	${ ElementWithBackgroundColor }
 `;
 
-function SquareDisplay( { backgroundColor, width, height, x, y, rotationAngle } ) {
+function SquareDisplay( { backgroundColor, width, height } ) {
 	const props = {
 		backgroundColor,
 		width,
```
miina: Should width and height be removed here as well? (It's mainly a placeholder element anyway, so probably not that important)

assets/src/edit-story/elements/video/display.js
```diff
@@ -29,16 +27,13 @@ function VideoDisplay( props ) {
 }
```
miina: ^ Probably the `x` and `y` etc. properties should be removed above in the `VideoDisplay` as well? Currently, it's adding all properties to `Element` ( `...props` ), this still includes these, even if the `propTypes` were updated, correct?

assets/src/edit-story/elements/square/display.js
dvoytenko: Good point. Removed.

assets/src/edit-story/elements/video/display.js
dvoytenko: Yeah. Generally these attributes are just element attributes. The template may or may not use them. But template should declare what it needs. So, yeah, I removed these from the declaration.

Issue #3801, opened 2019-11-21
Animations


jauyong:
## Feature description

Elements on the page can be animated. The old editor has a basic working implementation, but we had even more future ideas for this section in the works including allowing multiple animations per object. The main ideas can be reused. Animations are simply kept as state on supported elements and is essentially trivial to implement.

The tricky part here is animation replay and in particular animation replay of all elements on the page. The main idea here is probably to disable main UI while animations are replaying (graying out everything but the canvas, add a big stop button somewhere - similar in style to in-place editing) and let the animations play without any interaction available.

Also, should selected-object-animation-replay be supported if multiple elements are selected? It makes sense to be able to assign the same animation to multiple elements, but can you play them all simultaneously as well? What if they are set to play after one another? What if set to play with an unselected element in between? Some very complex scenarios can occur here.

Some UX thoughts have to be applied first, but due to the complex scenarios, we have to consider this well. It will probably not be hard to implement, but it has to be well designed first.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A suggestion is made based on the findings

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

Issue #3836, opened 2019-11-27
Add page grid view


barklund:
## Feature description

Add a page grid view as a new global state for the entire editor as well as a button in the page carousel that enables it.

In grid view, show page outlines of all pages in story and allow them to be both dragged and cut-pasted as designed.

Page reordering/duplication takes immediate effect.

Page outlines can be resized using the slider in the footer.

The close button in the footer will close grid view (changes have been persisted instantly when made).

This ticket does not include showing actual previews of the pages in the grid view.

The keyboard interactions are dependent on the outcome of #3826.

This is still pending UX review.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Users can enter grid view when pressing the button in the footer.
* Users can reorder pages in grid view by dragging.
* Users can focus pages and spaces in grid view by clicking.
* Users can move pages in grid view by using keyboard shortcuts as designed.
* Users can resize pages in grid view.
* Users can exist grid view.

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

jauyong:
Hey @barklund what exactly is this AC? Are you talking about the zoom slider?
> Users can resize pages in grid view.

barklund:
> Hey @barklund what exactly is this AC? Are you talking about the zoom slider?
> 
> > Users can resize pages in grid view.

Yes, it was referring to the preview thumbnail sizing. I have updated and clarified the text (and added _deleting pages_).

Note that the AC does not include multi-select as we're still exploring if this is a P1 or later.

adamsilverstein:
I love this issue, having a grid mode will make layout control so much easier for users! 

One idea from a recent demo: how about enabling the grid mode automatically (or a "grid automatic mode") for specific interactions such as drag-and-drop or image resizing. In this mode, the grid isn't visible until until interaction - when a user starts resizing or dragging image for example, then the grid appears and interactions "snap" to it. Once the stop interacting, the grid disappears.

swissspidy:
@adamsilverstein I think this issue is more about a grid view as described in https://github.com/ampproject/amp-wp/issues/3974#issuecomment-572136169. Snapping is handled in ampproject/amp-wp#3773.

adamsilverstein:
@swissspidy - thanks for the link, that issue does cover my idea - I had missed it earlier when reviewing the issue backlog.

swissspidy:
**Two open UX questions:**

[...]

PR #4020, opened 2020-01-03
DRAFT/PoC: animations and replay mode

dvoytenko:
## Summary

Partial for #3801

Notes:

1. I introduced `DesignMode={design, replay}`. The `replay` mode servers to main functions: disabling parts of the editor that obscure or complicated animations replay. For instance, selection really gets in a way of the replay, so it's completely disabled. Second reason: intermediary animated `divs` are inserted for animation that are otherwise not needed (and possibly harmful) for editing itself.
2. Design mode is currently canceled when clicking on the background. We will certainly also add <kbd>esc</kbd> canceling, and cancelling if any part of the model has been updated.
3. I think we might be able to adopt AMP animation presets more or less exactly with a small transformation of the code. But for now I just forked couple of presets for PoC/demo purposes only.
4. The element wrapper ideally would be already defining positioning, size, and rotation (see #4018). For now the status-quo works, but it feels rather fragile.

## Checklist

### DIFF
--- a/assets/src/edit-story/app/story/storyProvider.js
+++ b/assets/src/edit-story/app/story/storyProvider.js
@@ -11,6 +11,7 @@ import { useState } from '@wordpress/element';
 /**
  * Internal dependencies
  */
+import DesignMode from '../designMode';
 import Context from './context';
 
 import useLoadStory from './effects/useLoadStory';
@@ -35,6 +36,7 @@ function StoryProvider( { storyId, children } ) {
 	// Story state is stored in these three immutable variables only!
 	// Don't update 1 of these in an effect based off another base variable.
 	// Only update these directly as a response to user or api interactions.
+	const [ designMode, setDesignMode ] = useState( DesignMode.DESIGN );
 	const [ pages, setPages ] = useState( [] );
 	const [ title, setTitle ] = useState( '' );
 	const [ link, setLink ] = useState( '' );
@@ -73,6 +75,7 @@ function StoryProvider( { storyId, children } ) {
 
 	const state = {
 		state: {
+			designMode,
 			pages,
 			currentPageIndex,
 			currentPageNumber,
@@ -86,6 +89,7 @@ function StoryProvider( { storyId, children } ) {
 			link,
 		},
 		actions: {
+			setDesignMode,
 			setCurrentPageByIndex,
 			addBlankPage,
 			clearSelection,
--- a/assets/src/edit-story/app/designMode.js
+++ b/assets/src/edit-story/app/designMode.js
@@ -1,0 +1,7 @@
+
+const DesignMode = {
+	DESIGN: 'design',
+	REPLAY: 'replay',
+};
+
+export default DesignMode;
--- a/assets/src/edit-story/app/index.js
+++ b/assets/src/edit-story/app/index.js
@@ -17,6 +17,7 @@ import {
  * Internal dependencies
  */
 import theme, { GlobalStyle } from '../theme';
+import DesignMode from './designMode';
 import { useHistory, HistoryProvider } from './history';
 import { useAPI, APIProvider } from './api';
 import { useConfig, ConfigProvider } from './config';
@@ -51,6 +52,7 @@ App.propTypes = {
 export default App;
 
 export {
+	DesignMode,
 	useHistory,
 	useAPI,
 	useStory,
--- a/assets/src/edit-story/components/animations/animationPlayer.js
+++ b/assets/src/edit-story/components/animations/animationPlayer.js
@@ -1,0 +1,84 @@
+
+/**
+ * Rouhgly follows `Animation` API.
+ * See https://developer.mozilla.org/en-US/docs/Web/API/Animation.
+ */
+class AnimationPlayer {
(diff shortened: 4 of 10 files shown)


assets/src/edit-story/app/story/storyProvider.js
```diff
@@ -35,6 +36,7 @@ function StoryProvider( { storyId, children } ) {
 	// Story state is stored in these three immutable variables only!
 	// Don't update 1 of these in an effect based off another base variable.
 	// Only update these directly as a response to user or api interactions.
+	const [ designMode, setDesignMode ] = useState( DesignMode.DESIGN );
```
miina: Wondering if the `designMode` should be part of the `storyProvider` as it is now or if it would make sense in the canvas level? Is the design state relevant for a Page only?

miina: It's needed in the sidebar as well of course, just realized.

miina:
Q: Do we already have a description from the UX in terms of how the animations / combined animations are going to be handled? E.g. will we have only pre-set combined animations (such as Ken Burns) which user would add just as "Ken Burns"?
Couldn't find from Figma at this moment, was curious if there's anything set already.

swissspidy:
I don't think we have an epic or UX for this part yet.

assets/src/edit-story/app/story/storyProvider.js
dvoytenko: @miina I think your intuition is more correct here and maybe we should have the canvas state to sit above the sidebar? I don't think they are ever out of sync. Or maybe it's a higher-level state altogether - editor-wide state.

dvoytenko:
@swissspidy No epic/UX yet. And this PR is not focused on the editing nuances either. The main focus is on what's needed for replaying the animations from the editor's state point of view. As such as thought this is an important part of fundamentals that would be easier to capture now rather than later. I think the animation editor itself is going to be a multi-iteration part of the product spanning MVP and beyond.

dvoytenko:
@miina Ditto on this. The main goal is how to bring the editor state and how to render elements so that animations are possible at all, not how animations are edited itself. Specifically, composite animations story is not yet clear. But "Ken Burns" is likely not going to be a composite animation. With composites we will mostly be concerned with multiple elements and coordination over multiple elements. In this case, imho Ken Burns is relatively easily handled with the model in this PR. Though we should probably make a PoC for it to be certain.

assets/src/edit-story/app/story/storyProvider.js
swissspidy: If we could use it for the grid view as well (#3836) then it seems more like an editor-wide state.

barklund: As we do need some kind of editor-wide flag to indicate _grid-view-mode_, this seems like a great candidate for it.

However I suggest that we implement it with a strategy pattern and create different "editor view strategies". Then all relevant components would merely check the active editor view strategy. Thus e.g. the inspector panel doesn't need to know whether or not a certain design mode includes the inspector or not - it'll just refer to the strategy and update accordingly.

In implementation, it would be something like:

```js
const EDITOR_GRID = {
  hasInspector: false,
  hasLibrary: false,
  hasMoveable: false,
  hasPage: false,
  ...
};
const EDITOR_DESIGN = {
  hasInspector: true,
  hasLibrary: true,
  hasMoveable: true,
  hasPage: true,
  ...
};
const DESIGN_REPLAY = {
  hasInspector: true,
  hasLibrary: true,
  hasMoveable: false,
  hasPage: true,
  ...
};
```

It can be done with spreading a default strategy first of course and only overwriting where relevant.

I think this is a cleaner solution than sprinkling a lot of `const inspectorVisible = editorMode !== GRID;` all over the code. In stead we would just have `const inspectorVisible = editorMode.hasInspector;`.

dvoytenko: @barklund, one thing to clarify: the main nuance in "replay" mode right now is that we have to use a somewhat special markup and hence we also need separate set of templates - similar to `Display` vs `Edit`. I guess we could simply have `elementTemplate` property/enum that would indicate which template we use in which mode?

Beyond that, I'm wondering if we mode with page grid view is really a special mode or an overlay above the editor itself (e.g. kind of like a dialog). Do you have reasoning one way or another for this?

barklund: @dvoytenko, I don't have strong feelings on way or the other about either. They could probably be either two different small local states or both be "big" editor-wide view modes. I just wanted to avoid have a lot of constants and a lot of if's showing and hiding things throughout the application.

Thinking more closely about it, and can't really come up with any other global state like this. Maybe if we had something akin to "switch template", that would allow the user to "go back" to template selection and apply a different one, but that's only in the far future if ever.

So any implementation here is fine with me. When @swissspidy gets to implement grid view, he can figure out the best way to implement the state for that.

swissspidy: Dialog seems reasonable. Just saw that the spreadsheet refers to it as a modal too FWIW.

PR #4034, opened 2020-01-07
Attempt to fix test__multiple_valid_image_files

kienstra:
## Summary

This attempts to fix test__multiple_valid_image_files. That test had [failed](https://travis-ci.org/ampproject/amp-wp/jobs/630544394?utm_medium=notification&utm_source=github_status) intermittently.

Following the example of Weston's work on a FasterImage test, this hosts an image in a Gist instead of amp-wp.org
https://github.com/willwashburn/fasterimage/blob/42d125a15dc124520aff2157bbed9a4b2d4f310a/tests/FasterImageTest.php#L117

The actual file could be hosted in a Gist in Weston's account, if this approach is valid.

## Checklist


westonruter:
The intention behind hosting the file on `amp-wp.org` was to avoid it being tied to anyone's GitHub account. Since GitHub isn't a CDN it should seem that loading a file from a Gist would be _less_ reliable than loading it from a production website, not _more_ reliable.

kienstra:
OK, good point.

kienstra:
The [external-http](https://travis-ci.org/ampproject/amp-wp/jobs/634013328?utm_medium=notification&utm_source=github_status) build that had failed before just passed. But it had also passed sometimes before. I'll run the build a few more times just to test this, and then revert the previous commit https://github.com/ampproject/amp-wp/pull/4034/commits/aadfaa14b81ba47550faaf519dc5ab36d7b5aea3

kienstra:
Without https://github.com/ampproject/amp-wp/pull/4034/commits/aadfaa14b81ba47550faaf519dc5ab36d7b5aea3, `test__multiple_valid_image_files` [failed](https://travis-ci.org/ampproject/amp-wp/jobs/634024564#L1154) 1 of 2 times.

With https://github.com/ampproject/amp-wp/pull/4034/commits/aadfaa14b81ba47550faaf519dc5ab36d7b5aea3, it passed 5 of 5 time.

Understanding that using a GitHub gist isn't possible, I'll close this PR soon.

westonruter:
Using a GitHub Gist is possible, but it doesn't seem correct. There must be some other reason for why this is happening. Perhaps the issue is there is some Pantheon protection in place against a bots trying to DDoS? I don't see why that would be the case and only for one file (and a static one at that), but it's a possibility.

kienstra:
Yeah, I agree that it doesn't seem right to use a Gist. 

It does seem strange that there's only one file failing in the test, I'm still not sure what the cause is.

westonruter:
What happens if the order of items in the array is changed, so this SVG is fetched first?

kienstra:
I'll try that.

kienstra:
Hm, the same `.svg` file [failed](https://travis-ci.org/ampproject/amp-wp/jobs/634039782#L1155), even with it moved to the beginning of the array.

westonruter:
That somewhat rules out the timeout theory. So strange.

westonruter:
Let's try clearing the Travis caches...

kienstra:
Hm, it passed.

westonruter:
If it passes 5 times in a row, then I think we found the problem.

kienstra:
The [external-http](https://travis-ci.org/ampproject/amp-wp/jobs/634039782?utm_medium=notification&utm_source=github_status) test passed, and I restarted it.

kienstra:
Unfortunately, I don't know why `test__multiple_valid_image_files` [had failed](https://travis-ci.org/ampproject/amp-wp/jobs/634024564#L1158).

https://github.com/ampproject/amp-wp/pull/4034/commits/4d7218c1920c6c7e1b8f7019630dc05bef01b838#diff-9d88cb48c8afba5f63546855275bb633R241 tested whether basic curl requests for https://amp-wp.org/wp-content/plugin-test-files/google.svg succeeded. And they did. 

But FasterImage's curl requests [use curl_multi_*](https://github.com/willwashburn/fasterimage/blob/42d125a15dc124520aff2157bbed9a4b2d4f310a/src/FasterImage/FasterImage.php#L69).

When reproducing this locally by stepping through Xdebug breakpoints very slowly, there was never an exception when the curl request failed. 

And it looks like FasterImage didn't [throw exceptions](https://github.com/willwashburn/fasterimage/blob/42d125a15dc124520aff2157bbed9a4b2d4f310a/src/FasterImage/FasterImage.php#L113) in the failed Travis builds.

kienstra:
Hi @westonruter, 
What do you think about the approach here, understanding it's not ideal?

The test is reliably passing. But like you mentioned, it'd be much better to know why it failed.

Also, if you'd like, maybe the Gist could be from your GitHub account.

Thanks, and have a great weekend.

PR #4029, opened 2020-01-07
Update wp-dev-lib package

pierlon:
See: xwp/wp-dev-lib/pull/310

## Checklist

- [ ] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [ ] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [ ] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

PR #4043, opened 2020-01-08
Ignore Story editor tests

pierlon:
## Summary

This PR ignores the Stories Editor E2E tests due to their flakiness.

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,8 +67,6 @@ jobs:
     # PHP unit tests (7.4, WordPress trunk)
     - env: WP_VERSION=trunk  DEV_LIB_ONLY=phpunit                               INSTALL_PWA_PLUGIN=1
       php: 7.4snapshot
-    # E2E tests, since flaky.
-    - env: WP_VERSION=latest DEV_LIB_SKIP=phpcs,eslint,xmllint,phpsyntax,phpunit PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=
   include:
     - stage: lint
       name: Lint (PHP, JavaScript, and configuration files)
--- a/tests/e2e/jest.config.js
+++ b/tests/e2e/jest.config.js
@@ -14,6 +14,8 @@ module.exports = {
 	testPathIgnorePatterns: [
 		'.git',
 		'node_modules',
+		'tests/e2e/specs/block-editor',
+		'tests/e2e/specs/stories-editor',
 	],
 	reporters: [ [ 'jest-silent-reporter', { useDots: true } ] ],
 };
--- a/tests/e2e/jest.config.js
+++ b/tests/e2e/jest.config.js
@@ -14,7 +14,6 @@ module.exports = {
 	testPathIgnorePatterns: [
 		'.git',
 		'node_modules',
-		'tests/e2e/specs/block-editor',
 		'tests/e2e/specs/stories-editor',
 	],
 	reporters: [ [ 'jest-silent-reporter', { useDots: true } ] ],
--- a/bin/local-env/run-e2e-tests.js
+++ b/bin/local-env/run-e2e-tests.js
@@ -24,7 +24,7 @@ prc.stdout.on( 'data', ( data ) => {
 
 	const testsToIgnore = [];
 
-	if ( semver.gte( '5.3.0', semver.coerce( wpVersion ) ) ) {
+	if ( semver.gte( semver.clean( wpVersion ), '5.3.0' ) ) {
 		// Ignore tests that are not to be run in WP >= 5.3.0.
 		testsToIgnore.push( 'AMP Settings Screen should not allow AMP Stories to be enabled when Gutenberg is not active' );
 	}
--- a/tests/e2e/specs/block-editor/featured-image-notice.js
+++ b/tests/e2e/specs/block-editor/featured-image-notice.js
@@ -22,7 +22,7 @@ describe( 'Featured Image Notice', () => {
 	beforeEach( async () => {
 		await createNewPost( { postType: 'post' } );
 		await clickButton( 'Document' );
-		await clickButton( 'Featured Image' );
+		await clickButton( 'Featured image' );
 		await clickButton( 'Set featured image' );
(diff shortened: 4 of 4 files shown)


.travis.yml
```diff
@@ -94,20 +87,6 @@ jobs:
...
-        - nvm install
-        - composer install
-        - npm install
-      script:
-        - npm run build:js
-        - npm run build:css
-        - npm run env:start
-        - npm run test:e2e:ci
-        - npm run env:stop
-
```
westonruter: I don't think the tests have to be removed as a while. Rather, perhaps the `tests/e2e/specs/stories-editor` directory should be removed.

.travis.yml
```diff
@@ -60,15 +60,8 @@ jobs:
   fast_finish: true
   # These need to be exact matches, including whitespace!
   allow_failures:
-    # PHP unit tests (7.3, WordPress trunk)
-    - env: WP_VERSION=trunk  DEV_LIB_ONLY=phpunit                               INSTALL_PWA_PLUGIN=1
     # PHP and JavaScript unit tests (7.3, WordPress trunk, with code coverage)
     - env: WP_VERSION=latest DEV_LIB_ONLY=phpunit                               INSTALL_PWA_PLUGIN=1 RUN_PHPUNIT_COVERAGE=1
```
westonruter: @swissspidy  Should this still be allowed to fail? Why was added initially?

.travis.yml
pierlon: Ah OK.

pierlon:
PHP 7.4.0 on Travis does not include the `gd` extension, causing [errors](https://travis-ci.org/ampproject/amp-wp/jobs/634544781#L1065-L1069) to be thrown. I've [reported](https://travis-ci.community/t/gd-library-extension-not-available-with-php-7-4/6324/4?u=pierlon) it on their forum.

.travis.yml
swissspidy: I would not delete the directory.

Instead, I recommend adjusting `testPathIgnorePatterns` in `tests/e2e/jest.config.js` to ignore the `stories-editor` directory.

.travis.yml
swissspidy: Because trunk is not guaranteed to be stable. #4029 just showed it this week :-)

See #2408 for our discussion about adding this test & marking it as allowed to fail.

.travis.yml
pierlon: Some tests in `tests/e2e/specs/block-editor` are also failing. Should those be ignored as well?

pierlon:
:warning: Rebasing to remove reverted commits.

tests/e2e/jest.config.js
```diff
@@ -14,6 +14,8 @@ module.exports = {
 	testPathIgnorePatterns: [
 		'.git',
 		'node_modules',
+		'tests/e2e/specs/block-editor',
```
westonruter: Why exclude the `block-editor`?

westonruter: Oh, because you said some tests are failing.

westonruter: If this is included, then no tests would be run at all, except for `tests/e2e/specs/amp-options.js`, right?

Can the `block-editor` test be fixed? What is the failure?

pierlon: Fixed the failing `block-editor` tests in https://github.com/ampproject/amp-wp/pull/4043/commits/1f842d3deff649af9494772cbb29d5db34f226fa.

westonruter:
Should this not also remove:

https://github.com/ampproject/amp-wp/blob/1f842d3deff649af9494772cbb29d5db34f226fa/.travis.yml#L71

pierlon:
Yes, nice catch.

westonruter:
I cherry-picked the changes here onto the 1.4 branch, but the E2E are failing for some reason: https://travis-ci.org/ampproject/amp-wp/jobs/635359942

Any ideas?

pierlon:
The Gutenberg plugin [failed to activate](https://travis-ci.org/ampproject/amp-wp/jobs/635359942#L870), which led to tests failing due to them being unable to find the case-sensitive words "Featured image" (see https://github.com/ampproject/amp-wp/pull/4043/commits/1f842d3deff649af9494772cbb29d5db34f226fa#diff-792685570ec094fec9f40f84e3af8d3bR25). Apparently WP CLI failed to install and activate Gutenberg due to clash in a function declaration (`WP_CLI\UpgraderSkin::feedback($string)` vs `WP_Upgrader_Skin::feedback($string, ...$args)`.

On the `1.4` branch, WP CLI v2.3.0 is installed via Composer while within the Docker container v2.4.0 is used. Further investigation shows v2.3.0 uses `WP_CLI\UpgraderSkin::feedback($string)` while v2.4.0 on PHP > 5.6 (which is the case here) uses `WP_Upgrader_Skin::feedback($string, ...$args)`, which leads me to speculate whether both versions are being loaded at the same time. Thoughts @schlessera?

The `develop` branch uses WP CLI v2.4.0 in both instances, however, which I think is why the error does not occur there.

schlessera:
This is due to a breaking change in WordPress Core 5.3. While making the move to PHP 5.6 code, the signature for the upgrader base class `feedback()` method changed from `WP_Upgrader_Skin::feedback($string)` to `WP_Upgrader_Skin::feedback($string, ...$args)`.

WP-CLI 2.3.0 is compatible with older versions of WP (< 5.3) only, while WordPress Core 5.3+ requires WP-CLI 2.4.0+.

This is unfortunate, but that's WordPress' way of backward compatibility I'm afraid...

schlessera:
In case that wasn't clear, the solution is to use WP-CLI v2.4.0+ in the Docker container as well, as it seems to be using WP Core 5.3+.

schlessera:
No, the other way around, install WP-CLI 2.4.0+ on the v1.4 branch if you're using WP Core 5.3+ there. 🤪

westonruter:
Moving this to a separate issue: #4174.

Issue #3779, opened 2019-11-15
Port Existing Copy and Paste Keyboard Shortcut


jauyong:
## Feature description

This ticket was taken from a row in this [spreadsheet](https://docs.google.com/spreadsheets/d/1nqJ-iOypTziWFjWBIBoJUc0LHW3gTeUnLkbjVH_eW4g/edit#gid=0). This ticket can be used to track research and work related to porting this existing feature from the "old" editor to the "new" editor. 

UX Required: Yes
- Would like to tie these action to keyboard shortcuts

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Ensure that Copy&Paste works with keyboard shortcuts outside of the context menu.
* The Page block should handle the pasting event.

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

PR #4044, opened 2020-01-09
Copy/paste for selected elements inside the canvas

dvoytenko:
## Summary

Partial for #3779.

Blocked by #4018.

This doesn't deal with all aspects of shortcuts and popup menus. Just focus on the selected element copy/paste using platform events. 

Some nuances:

* The focus state of elements is currently ignored. Even if we decide to support it, we can't rely on this because the focus in this kind of editor is relatively fragile and also because we need to support multi-select.
* The implementation intercepts document-level copy/cut/paste events and sets the payload.
* The document-level events part is important. Copy/paste target outside of inputs and selection ranges is most likely will be `document.body` element.
* This PR uses a pre-existing `Save()` template, but adds a new `TextContent` template to support both: in-editor and out-of-editor copy/paste.
* The HTML payloads that are not packaged by our editor could also be processed. E.g. `` and other could be imported this way. However, that requires a deeper media library integration to reupload resources. Thus I'm skipping it in this PR.

## Checklist

### DIFF
--- a/assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
+++ b/assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
@@ -1,0 +1,126 @@
+/**
+ * External dependencies
+ */
+import uuid from 'uuid/v4';
+
+/**
+ * WordPress dependencies
+ */
+import { useCallback, renderToString } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { useStory } from '../../app';
+import useClipboardHandlers from '../../utils/useClipboardHandlers';
+import { getDefinitionForType } from '../../elements';
+
+const DOUBLE_DASH_ESCAPE = '_DOUBLEDASH_';
+
+/**
+ * @param {?Element} container
+ */
+function useCanvasSelectionCopyPaste( container ) {
+	const {
+		state: { currentPage, selectedElements },
+		actions: { appendElementToCurrentPage, deleteSelectedElements },
+	} = useStory();
+
+	const copyCutHandler = useCallback(
+		( evt ) => {
+			const { type: eventType, clipboardData } = evt;
+
+			if ( selectedElements.length === 0 ) {
+				return;
+			}
+
+			const payload = {
+				sentinel: 'story-elements',
+				// @todo: Ensure that there's no unserializable data here. The easiest
+				// would be to keep all serializable data together and all non-serializable
+				// in a separate property.
+				items: selectedElements.map( ( element ) => ( {
+					...element,
+					basedOn: element.id,
+					id: undefined,
+				} ) ),
+			};
+			const serializedPayload = JSON.stringify( payload ).replace( /\-\-/g, DOUBLE_DASH_ESCAPE );
+
+			const textContent = selectedElements
+				.map( ( { type, ...rest } ) => {
+					const { TextContent } = getDefinitionForType( type );
+					if ( TextContent ) {
+						return TextContent( { ...rest } );
+					}
+					return type;
+				} )
+				.join( '\n' );
+
+			const htmlContent = selectedElements
+				.map( ( { type, ...rest } ) => {
+					// eslint-disable-next-line @wordpress/no-unused-vars-before-return
+					const { Save } = getDefinitionForType( type );
+					return renderToString( <Save { ...rest } /> );
+				} )
+				.join( '\n' );
+
+			clipboardData.setData( 'text/plain', textContent );
+			clipboardData.setData( 'text/html', `<!-- ${ serializedPayload } -->${ htmlContent }` );
+
+			if ( eventType === 'cut' ) {
+				deleteSelectedElements();
+			}
+
+			evt.preventDefault();
+		},
+		[ deleteSelectedElements, selectedElements ],
+	);
+
+	const pasteHandler = useCallback(
+		( evt ) => {
+			const { clipboardData } = evt;
+
+			try {
+				const html = clipboardData.getData( 'text/html' );
+				if ( html ) {
+					const template = document.createElement( 'template' );
+					template.innerHTML = html;
+					for ( let n = template.content.firstChild; n; n = n.nextSibling ) {
+						if ( n.nodeType === /* COMMENT */ 8 ) {
+							const payload = JSON.parse( n.nodeValue.replace( new RegExp( DOUBLE_DASH_ESCAPE, 'g' ), '--' ) );
+							if ( payload.sentinel === 'story-elements' ) {
+								payload.items.forEach( ( { x, y, basedOn, ...rest } ) => {
+									currentPage.elements.forEach( ( element ) => {
+										if ( element.id === basedOn || element.basedOn === basedOn ) {
+											x = Math.max( x, element.x + 20 );
+											y = Math.max( y, element.y + 20 );
+										}
+									} );
+									const element = {
+										...rest,
+										basedOn,
+										id: uuid(),
+										x,
+										y,
+									};
+									appendElementToCurrentPage( element );
+								} );
+								evt.preventDefault();
+							}
+						}
+					}
+				}
+			} catch ( e ) {
+				// Ignore.
+			}
+		},
+		[ appendElementToCurrentPage, currentPage ],
+	);
+
+	useClipboardHandlers( container, copyCutHandler, pasteHandler );
+
+	// @todo: return copy/cut/pasteAction that can be used in the context menus.
+}
+
+export default useCanvasSelectionCopyPaste;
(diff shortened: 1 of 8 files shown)


assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
```diff
@@ -0,0 +1,126 @@
...
+				if ( html ) {
+					const template = document.createElement( 'template' );
+					template.innerHTML = html;
+					for ( let n = template.content.firstChild; n; n = n.nextSibling ) {
+						if ( n.nodeType === /* COMMENT */ 8 ) {
+							const payload = JSON.parse( n.nodeValue.replace( new RegExp( DOUBLE_DASH_ESCAPE, 'g' ), '--' ) );
+							if ( payload.sentinel === 'story-elements' ) {
+								payload.items.forEach( ( { x, y, basedOn, ...rest } ) => {
+									currentPage.elements.forEach( ( element ) => {
+										if ( element.id === basedOn || element.basedOn === basedOn ) {
```
swissspidy: Interesting idea with the positioning here!

assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
```diff
@@ -0,0 +1,126 @@
...
+
+			const htmlContent = selectedElements
+				.map( ( { type, ...rest } ) => {
+					// eslint-disable-next-line @wordpress/no-unused-vars-before-return
+					const { Save } = getDefinitionForType( type );
+					return renderToString( <Save { ...rest } /> );
+				} )
+				.join( '\n' );
+
+			clipboardData.setData( 'text/plain', textContent );
```
miina: Does this need separate handling for IE due to IE accepting only `Text` (e.g. `clipboardData.setData ( 'Text', input.value )`?

assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
```diff
@@ -0,0 +1,126 @@
...
+		},
+		[ deleteSelectedElements, selectedElements ],
+	);
+
+	const pasteHandler = useCallback(
+		( evt ) => {
+			const { clipboardData } = evt;
+
+			try {
+				const html = clipboardData.getData( 'text/html' );
```
miina: Same question here for IE.

assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
swissspidy: For reference, the code from the legacy editor: https://github.com/ampproject/amp-wp/blob/eba2f3b1cd42f81b7e2b7d4687d74a5e606603ba/assets/src/stories-editor/blocks/amp-story-page/copy-paste-handler.js#L48-L65

This begs the question of which browsers we want to support. Perhaps we don't want to support IE 11 at all? cc @pbakaus

assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
```diff
@@ -0,0 +1,126 @@
...
+				if ( html ) {
+					const template = document.createElement( 'template' );
+					template.innerHTML = html;
+					for ( let n = template.content.firstChild; n; n = n.nextSibling ) {
+						if ( n.nodeType === /* COMMENT */ 8 ) {
```
miina: Could we perhaps reduce nesting here a bit by using `continue` were possible to make it easier to read?
E.g. `if ( n.nodeType !== 8 ) continue` and below as well for `story-elements`?

assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
```diff
@@ -0,0 +1,126 @@
...
+								payload.items.forEach( ( { x, y, basedOn, ...rest } ) => {
+									currentPage.elements.forEach( ( element ) => {
+										if ( element.id === basedOn || element.basedOn === basedOn ) {
+											x = Math.max( x, element.x + 20 );
+											y = Math.max( y, element.y + 20 );
+										}
+									} );
+									const element = {
+										...rest,
+										basedOn,
```
miina: Will this be used anywhere?

assets/src/edit-story/utils/useClipboardHandlers.js
```diff
@@ -0,0 +1,73 @@
...
+			}
+
+			// A target can be anywhere in the container's full subtree, but not
+			// in its siblings.
+			if ( ! container.contains( target ) && ! target.contains( container ) ) {
+				return;
+			}
+
+			// Someone has already put something in the clipboard. Do not override.
+			if ( clipboardData.types.length !== 0 ) {
```
miina: Curious, in which case shouldn't it override what's already in the clipboard?

assets/src/edit-story/utils/useClipboardHandlers.js
```diff
@@ -0,0 +1,73 @@
...
+
+			copyCutHandler( evt );
+		};
+
+		const pasteHandlerWrapper = ( evt ) => {
+			const { target } = evt;
+
+			// Elements that either handle their own clipboard or use platform.
+			if ( ! target ||
+          BLACKLIST_CLIPBOARD_ELEMENTS.includes( target.tagName ) ||
```
miina: Indentation adjustment needed?

spacedmonkey:
To confirm this PR is only designed to handle text / html content. 

What if the is a video / image / audio file in the clipboard? 
Can another element be copied? 
Can another element be cut? 
If broken HTML is copied into the clipboard. How is this html validated? 
Is the user informed if paste fails? 
What happens if the user copies and pastes via the context menu ( right click)?

assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
pbakaus: Let's drop IE11. We will only support last two versions of Chrome, Safari, Edge and Firefox (we will try to cover 90%+ market that way)

assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
pbakaus: no IE.

assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
dvoytenko: 👍

assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
dvoytenko: Good idea! Done.

assets/src/edit-story/components/canvas/useCanvasSelectionCopyPaste.js
dvoytenko: Yeah. It's used on the repeated paste. To try it, do one copy on an element and then do couple of paste one after another. They will all stack up to avoid covering each other. The same algo also ensures that if it's a cut/paste, then the new element comes back to the exactly same spot.

assets/src/edit-story/utils/useClipboardHandlers.js
dvoytenko: TBH this might be overprotection. Just something I've done before, but don't remember any special problematic case.

dvoytenko:
@spacedmonkey 

> To confirm this PR is only designed to handle text / html content.

This PR exclusively deals with in-editor copy/paste (updated the description). Extending it to out-of-editor copy/paste is more akin dealing with drag-n-drop. So I think it should be done using separate handler and probably better wait until we have the first draft of d-n-d.

> What if the is a video / image / audio file in the clipboard?

This has to go via media import and more similar to d-n-d per above.

> Can another element be copied?
> Can another element be cut?

Another HTML element? Definitely, but not sure the exact behavior. If this is an img/video HTML element, then it should be the same as image/video copy/paste, but we will have a bit more meta-data (likely width/height), so we should use that. But media-import has to come first.

> If broken HTML is copied into the clipboard. How is this html validated?

Right now it's not an issue. I don't process HTML here, instead I only parse the comment with a predefined JSON structure. And only when it's valid and parses. For a wider HTML support - this shouldn't be a much of an issue either - we can either do something about payload or not.

> Is the user informed if paste fails?

Failure is very much expected. The handler simply fails and the next handler in the chain gets a turn to try to process this event. If all fail - the platform will get to process the event. I typically don't see paste errors out there, but I guess we could show one. Or, alternatively, as a catch-all we could extract text and paste it as a text block.

> What happens if the user copies and pastes via the context menu ( right click)?

Do you mean a system context menu or our own? With a system context menu the copy/paste events are propagated starting with the target. This would typically bypass the handler in this PR. If it's our own context menu, we just need to do `executeCommand('copy'/'paste')` and that would activate these handlers the same way as a keyboard shortcut. This is the only way we get a chance to read/write the clipboard w/o special permissions.

swissspidy:
> Or, alternatively, as a catch-all we could extract text and paste it as a text block.

Sounds like a reasonable future enhancement

PR #4048, opened 2020-01-09
Page reordering from carousel

### LINKED ISSUE #3835: Allow focus and reordering in page carousel
## Feature description

Allow pages in carousel to receive focus if clicked (as well as switch current page on the canvas above) and reordering if dragged.

Pages need not be focused first in order to be dragged.

There are some open issues around keyboard navigation (changing focus only or changing page as well) and similarly for keyboard-only reordering (maybe shift-arrow to move a page to the right or left in the order?).

This does not include page grid view or page grid view button.

---------------

## Acceptance criteria

1. Page outlines in carousel can be focused by clicking on them
1a. This will of course still also change current page to the clicked one
1b. Clicking anywhere else on the page removes focus from pages
2. Page not in focus, 25% opacity
3. Page currently in focus, 100% opacity
4. Drag page thumbnails to reorder pages
5. Focus can be manipulated with arrow keys or (<kbd>shift</kbd> +) <kbd>tab</kbd>.
6. The focused page can be opened (if not already open) with <kbd>enter</kbd> and/or deleted by pressing <kbd>del</kbd> or <kbd>backspace</kbd>.

## Implementation brief

* **NB:** use `tabindex` and `:focus` for focus for accessibility and ease of implementation.

- Create a new `DropZone` component which will be used to wrap the element that needs dropzone.
- The `DropZone` implementation will have a placeholder for files drop as well but only implement general onDrop functionality
- Each Page icon below the Pages (<button>) will be wrapped into DropZone which will then decide where should the Page be dropped.

- Determine approach and UX for reordering with keyboard (dependent on #3897).

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

miina:
## Summary

Fixes #3835 

- Adds basic DropZone.
- Adds DropZoneProvider which keeps track of the DropZones and detects which is hovered on.
- Adds functionality for reordering Pages via Thumbnails.
- Assumes that focusing and keyboard interaction will be part of #3826 (PR: https://github.com/ampproject/amp-wp/pull/3897

## Checklist


assets/src/edit-story/components/dropzone/index.js
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * WordPress dependencies
+ */
+import { useRef } from '@wordpress/element';
+
+const DropZoneComponent = styled.div`
+	display: inherit;
+`;
+
+function DropZone( { children, onDrop } ) {
```
swissspidy: I just stumbled across https://github.com/WordPress/gutenberg/pull/19514 and was wondering if we could turn this into a hook too.

assets/src/edit-story/components/canvas/carrousel/index.js
```diff
@@ -30,12 +31,45 @@ const Page = styled.a`
...
+				// @todo Create a Draggable component for setting data and setting "draggable".
+				const onDragStart = ( evt ) => {
+					const pageData = {
+						type: 'page',
+						index,
+					};
+					evt.dataTransfer.setData( 'text', JSON.stringify( pageData ) );
+				};
+				return (
+					<DropZone key={ index } onDrop={ onDrop } >
```
dvoytenko: So, a `DropZone` always wraps a page thumbnail? Is there no way to drop between the two thumbnail elements? I mean is it physically possible, if the thumbnails have margins?

assets/src/edit-story/components/canvas/carrousel/index.js
```diff
@@ -30,12 +31,45 @@ const Page = styled.a`
...
+						<Page draggable="true" onDragStart={ onDragStart } onClick={ () => setCurrentPageByIndex( index ) } isActive={ index === currentPageIndex } />
```
dvoytenko: I'm also wondering if the `draggable` a right tool here? It works great from out-of-app drag-n-drop, but it depends on the kind of UX we want to apply. E.g. we might want to highlight a place where the page will drop in the new order, shift elements a bit to demo how this element will fit, provide constraints to dragging, stuff like this. It's definitely possible to do most of it with `draggable=true`, but at that point it might seem like we doing most of it manually anyway. Maybe it'd be better to do this `Moveable`-style? We have a couple of components that do something like this: layers, page carousel, maybe animations too. So there might be savings there as well. LMK if you've already considered that approach and discarded it.

assets/src/edit-story/components/canvas/carrousel/index.js
miina: Yes, it's possible, currently, too. The thumbnails have margin currently as well, meaning there is no space between the two dropzones and dropping between the thumbnails will still result in dropping into a dropzone. In theory, it would be possible to add a separate dropzone element between the thumbnails instead but it seems to be easier to just wrap the related element and adjust it via styling as necessary.

assets/src/edit-story/components/canvas/carrousel/index.js
miina: I haven't considered using Moveable, however, I did think through the scenarios mentioned above for highlighting the area where a Page is dropped (e.g. using border for the dropzone), shifting the elements (as well doable via styling or adding an extra space element), etc. I did not find any UX related to that currently from Figma and purposefully left out handling those cases.

Would be good to confirm with @samitron7 to understand the UX for these cases.

It feels like at this moment, we can do everything we'd like to do using draggable as well, and unless there's a specific reason to create a "custom draggable" using Moveable then we could just stick to the existing option. Thoughts?

assets/src/edit-story/components/dropzone/index.js
miina: Maybe. I wonder if there could be cases where we'd benefit from the dropzone being a separate element though, e.g. if we'd want to leave space between the border of the dropzone / the actual element and add a border just to the dropzone to highlight the dropzone. Thoughts?

swissspidy: > if we'd want to leave space between the border of the dropzone / the actual element and add a border just to the dropzone to highlight the dropzone

Looks like that is the case for the grid view. So I guess it makes sense the way it is now.

assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -37,22 +37,22 @@ const Area = styled.div`
 
 function CanvasLayout() {
 	return (
-		<SelectionCanvas>
-			<Background>
-				<Area area="page">
+		<Background>
+			<Area area="page">
+				<SelectionCanvas>
```
miina: @dvoytenko Looks like currently the lasso selection was happening outside of Page, too, including the carousel under the page, creating a selection instead of dragging these. Moved the `SelectionCanvas` around the `Page` only, meaning that now the lasso selection can be drawn within the `Page` only and dragging outside of it won't be handled by lasso selection. Perhaps that's not the best option either though and multi-selecting should be allowed outside of the Page, too? Thoughts?

assets/src/edit-story/components/canvas/carrousel/index.js
miina: Actually, I just saw that Figma has been updated meanwhile and there should be a blue highlight displayed indicating the place where the thumbnail will be dropped.

LMK if keeping the current approach of "draggable" seems reasonable to you as well, then I+ll continue working on the PR and implement the highlighting, too.

dvoytenko: Hard to say if we will run into issues with native draggable here: we might be able to solve all we need just fine. But couple of thoughts:

1. I didn't mean to use `Moveable` as is, but more to follow a similar approach: e.g. using a component with mouse/touch/pointer events + transforms. Maybe there's even an existing component we could use? Ironically I've been running into react-movable by mistake all the time, but no idea how good it is.
2. If we do use native drop events: would it be easier for us to have a single drop zone - the whole `nav` for page thumbnails and use drop coordinates to decide where the reordering will occur?
3. We also will be supporting keyboard events for this. So I'm also wondering how much we would need to do for this anyway that we'd otherwise sharable with a would-be non-native mouse-based implementation. If not a lot, then it doesn't really matter if we use pointer or drag events at this time.
4. Another thought: we have at this time three places where something like this is done: (a) page carousel, (b) page grid, and (c) layers panel. Given dnd and keyboard support, does that change at all how you'd approach this? Not 100% if all these are entirely similar, but they are at least as far as keyboard support is concerned.

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: I don't think this is a right approach. I'm pretty certain lasso should work in the canvas. I think it's the other way around instead: carousel should not be part of the canvas. I think logically it'd make more sense: canvas includes the fully drawable area including margin, frame, etc. All tools should be off the canvas. Generally there's very little reason why an inspector is out of canvas but page carousel is on the canvas. We just need to find a way to provide a similar layout with a slightly different layering.

We can, of course, simply rename things and nest them a little different instead of overhauling completely. But I think it's important that the selection area (just like moveable area) geometrically span the whole current canvas area under the current "page", "carousel", "inspector", and "media panel" - we just need to make sure they are not structured in a parent/child relationships so that the event propagation is logical.

In the very short-term, if you don't want to deal with this now, you can just put `onMouseDown={ evt => evt.stopPropagation()}` on the page carousel container. And I can do a pass on the nesting/layering in the meantime so we can discuss it separately?

miina: Yep, I basically added this for now just to make it work but didn't think either that this is the right approach. As far as I remember then actually the user should be able to drag elements out of the Page as well, meaning that the selection should work anywhere on the Canvas. Will confirm with Sam just in case but I agree that the carousel should be separated instead of isolating the Page.

I don't mind dealing with this now at all, just wasn't certain of the UX, will take a look tomorrow.

assets/src/edit-story/components/canvas/carrousel/index.js
miina: 1. Hmm, I'd guess that the original element would still have to stay in place, at least in case of reordering, and we'd only need to drag the clone?
2. I did consider this, however, in this specific case, it seemed to me to be more straightforward to have a separate dropzone for each thumbnail and just checking on which side of it the element was dropped, instead of figuring out the general coordinates in the dropzone. It seems to be sufficient for any of the use cases that are currently in Figma (checking if dropped left/right/top/bottom). I suppose the issue you're thinking might be with areas which for some reason between the elements would be left without the dropzone coverage?
3. That's a good point, I'd like to understand better the UX of keyboard events for this specific case as well, e.g. if would still just the clone move and the original element stay in place. I'll look into keyboard dnd in more detail tomorrow to understand this better -- if the default draggable still makes sense.
4. I did think about the other occasions as well where we'd need dnd but not necessarily from the keyboard point of view -- I don't have experience with keyboard dnd really, TBH. Again, will look into this tomorrow to know better.

The current Dropzone/draggable implementation is definitely not final at all -- I did consider that we might need a general Draggable component (be that based on the default "draggable" or something custom then), however, I purposefully did not look into this within this PR so far to improve it as we need in the future, thinking of YAGNI while keeping in mind the use cases we have planned so far. Even if we decide to just keep the current "draggable" then it'll be very easy to replace if needed, the current implementation is not a very complex masterpiece :)

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: There are actually numerous other related bugs with this layout. I reprioritized this a bit on my side. This is still work in progress, but here's what it might involve: https://github.com/ampproject/amp-wp/pull/4110. This PR is pretty close, but couple of issues are still outstanding...

assets/src/edit-story/components/canvas/carrousel/index.js
miina: Looked into dnd + keyboard and although I might very well not be anticipating something that you are, then it still seems to me that for now we should stick to the default "draggable".

Thinking through how the keyboard dnd would work then, in that case, we would only need to deal with keyboard events, "draggable" already solves the mouse events for us, so it does seem that there's not that much shareable between the mouse events and keyboard events logic (other than "dropping", which would already be handled in dropzones).

I will create a separate issue for keyboard page reordering once #3826 is in place and work on that separately. As for dropzones, it might still be OK to use separate dropzones for keyboard events as well, in my mind (I might be wrong) it seems to be more straightforward this way for applying for other use cases as well. WDYT?

miina:
Follow-up issue for keyboard reordering only: https://github.com/ampproject/amp-wp/issues/4112

assets/src/edit-story/components/canvas/carrousel/index.js
miina: Okay, thought about it a bit more.

I'm thinking of reworking this to implement the originally planned DropZoneProvider which would handle `onDragOver` event to determine which dropzone within it is being hovered on and where the element is being dropped. This means that DropZoneProvider would keep track of DropZone elements within it.

This way, when doing the keyboard implementation, it could use the same logic of `onDragOver` to determine the hovered dropzone, only that this logic would be triggered by keyboard events while in dragging state (e.g. clicking left/right arrow).

Thoughts?

miina: Started reworking in b01dd59, still WIP.

dvoytenko: @miina, this sounds good. I agree - we can definitely start with draggable interface and then change if we don't like something. I did ask @pbakaus to look around if any library superficially look good to him and he found this: https://github.com/clauderic/react-sortable-hoc. It states:

> "However, most of those libraries rely on the HTML5 Drag & Drop API, which has some severe limitations. For instance, things rapidly become tricky if you need to support touch devices, if you need to lock dragging to an axis, or want to animate the nodes as they're being sorted. React Sortable HOC aims to provide a simple set of higher-order components to fill those gaps. If you're looking for a dead-simple, mobile-friendly way to add sortable functionality to your lists, then you're in the right place."

But let's keep to your plan and maybe see if any of those statement relevant to us, in case we will run into any problems in the future.

assets/src/edit-story/components/canvas/canvasLayout.js
miina: @dvoytenko Would you prefer first getting the layers change done and then this PR? Alternatively, we could merge this PR with "incorrect" `SelectionCanvas` and fix it when reorganizing the layers in your PR. LMK.

Note that it looks like lasso selection might not make sense fully outside of the Page since the [elements are deleted](https://www.figma.com/file/Cd9iacVknHK62RO4pRwb68/Stories-Editor?node-id=2832%3A208) when dragged off the Page.

spacedmonkey:
What is the expected behaviour when a page is reordered and the presses undo / redo?

miina:
> What is the expected behaviour when a page is reordered and the presses undo / redo?

Good question. It should undo it :)

Note: When testing, it looks like the history is generally not working properly, e.g. when adding a Page and clicking Undo, it removes the Page, however, clicking again adds the Page back again.

Same with adding an element -- the first Undo removes it, the second adds it back again, instead of undoing a previous action. Try adding a few elements and then clicking Undo:
history-issue

It looks like the Undo/Redo should be looked at separately, I was (incorrectly) assuming that it already worked as expected with the actions in the reducer but looks like not.

I think it would be good to create a separate issue for fixing History first. Thoughts? cc @barklund

barklund:
> I think it would be good to create a separate issue for fixing History first. Thoughts? cc @barklund

Yes, I think it actually broke all the way back in #3910 at some stage in the revisions of the PR. I have an idea about what the problem is (essentially that restore doesn't actually set the same objects, but clone at least one of them, which makes the history think that state hasn't be restored to an old state, it's just a new state).

I think it's a fairly easy fix once looked at separately.

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: @miina I think we'll have to merge this PR first. The layering is a bit more work then we first anticipated. 

> Note that it looks like lasso selection might not make sense fully outside of the Page since the elements are deleted when dragged off the Page.

Yeah, the latest version asks the elements to be deleted when off the screen. But I think the lasso behavior is unchanged. We just want to be able to start the selection anywhere on the workspace (outside the headers, footers, etc). I'll double-check though to be certain.

assets/src/edit-story/components/dropzone/index.js
```diff
@@ -0,0 +1,102 @@
...
+ */
+import { useRef, useLayoutEffect } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import useDropZone from './useDropZone';
+
+const DropZoneComponent = styled.div`
+	display: inherit;
```
dvoytenko: `display: inherit` seems very uncommon. Why is it needed?

assets/src/edit-story/components/dropzone/index.js
```diff
@@ -0,0 +1,102 @@
...
+`;
+
+function DropZone( { children, onDrop } ) {
+	const dropZoneElement = useRef( null );
+	const { actions: { addDropZone, resetHoverState }, state: { hoveredDropZone, dropZones } } = useDropZone();
+
+	useLayoutEffect( () => {
+		if ( ! dropZones.some( ( { ref } ) => ref === dropZoneElement.current ) ) {
+			addDropZone( {
+				ref: dropZoneElement.current,
```
dvoytenko: As in the other similar case, this is a `node`, right? Not the `ref<node>`?

assets/src/edit-story/components/dropzone/index.js
```diff
@@ -0,0 +1,102 @@
...
+			} );
+		}
+	}, [ addDropZone, dropZones ] );
```
dvoytenko: This is a bit recursive. The side-effect modifies its own dependency. Maybe it'd be easier to just rename `addDropZone` to `registerDropZone` and leave it to its internal implementation? As a benefit, you would also be able to do do something like this:

```
useLayoutAffect( () =>
   registerDropZone(dropZoneElement.current), [registerDropZone] )
```

And the `registerDropZone` could return the unregister function which would allow us to unregister drop zones when a component is unmounted automatically. E.g.

```
function registerDropZone(node) {
  ...
  return () => {
    // unregister here.
  };
}
```

barklund: Then you should make it a "real" custom hook, call it `useRegisterDropZone` and use it like that. The above is a bit fragile if someone doesn't see that the implicit return of the arrow function is actually important and changes it to:

```js
useLayoutAffect( () => {
  // something here
  registerDropZone(dropZoneElement.current);
}, [registerDropZone] )
```

Now the unregistering doesn't work, because the return value of `registerDropZone` is returned forward.

With a custom hook, it's a lot more clear how it works.

assets/src/edit-story/components/canvas/carrousel/index.js
```diff
@@ -30,13 +31,47 @@ const Page = styled.a`
...
+		return dstIndex + 1 + indexAdjustment;
+	};
+	const handleClickPage = ( page ) => setCurrentPage( { pageId: page.id } );
 	return (
 		<List>
-			{ pages.map( ( page, index ) => (
-				<Page key={ index } onClick={ handleClickPage( page ) } isActive={ index === currentPageIndex } />
-			) ) }
+			{ pages.map( ( page, index ) => {
+				const onDrop = ( evt, position ) => {
```
spacedmonkey: Why is the function defined in the map. Wouldn't that mean it was redefined on every loop. Can't this function be moved?

assets/src/edit-story/components/canvas/carrousel/index.js
```diff
@@ -30,13 +31,47 @@ const Page = styled.a`
...
+					const arrangedIndex = getArrangeIndex( droppedEl.index, index, position );
+					// Do nothing if the index didn't change.
+					if ( droppedEl.index !== arrangedIndex ) {
+						const pageId = pages[ droppedEl.index ].id;
+						arrangePage( { pageId, position: arrangedIndex } );
+						setCurrentPage( { pageId } );
+					}
+				};
+				// @todo Create a Draggable component for setting data and setting "draggable".
+				const onDragStart = ( evt ) => {
```
spacedmonkey: Again, can't this be defined in a high scope.

spacedmonkey:
```
const onDragStart =  useCallback( ( index ) => ( evt ) => {
        const pageData = { type: 'page', 	index, };
	evt.dataTransfer.setData( 'text', JSON.stringify( pageData ) );
} ); 
```

```
<DropZone key={ index } onDrop={ onDrop( index ) } >
```

assets/src/edit-story/components/dropzone/index.js
miina: In page navigation context, the dropzone's display should be `flex` (as its parent), in some other context probably not, thus the `inherit`. Thoughts?

assets/src/edit-story/components/dropzone/index.js
miina: Correct, changed.

assets/src/edit-story/components/dropzone/index.js
miina: Added `unregisterDropZone` and moved the checks to the callbacks directly.

assets/src/edit-story/components/canvas/carrousel/index.js
miina: Moved this by adding `pageIndex` attribute to dropzone. Might be useful elsewhere anyway for determining which page a dropzone belongs to.

assets/src/edit-story/components/canvas/carrousel/index.js
miina: Thanks! Done.

swissspidy:
There are now some merge conflicts after the page carousel revamp.

miina:
Just pushed the change, the conflicts should be resolved now.

swissspidy:
Seems like the editor is crashing when adding new pages. Getting `Uncaught Invariant Violation: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate.` within `DropZone`.

miina:
Thanks, @swissspidy! Should be fixed now.

assets/src/edit-story/components/dropzone/index.js
dvoytenko: I'm not sure that would work like this. You can inherit `display:flex`, but what about `flex-direction` and very many other flex/grid/etc CSS properties? I this for this reason there was `display:contents` introduced sometime ago. But not sure if that ever became a reality and in what form.

miina: Good point. I'll remove this, we can handle any specific display needs per case basis later if they should come.

assets/src/edit-story/components/canvas/carousel/index.js
```diff
@@ -105,6 +106,38 @@ function Carousel() {
...
+		const droppedEl = JSON.parse( evt.dataTransfer.getData( 'text' ) );
+		if ( ! droppedEl || 'page' !== droppedEl.type ) {
+			return;
+		}
+		const arrangedIndex = getArrangeIndex( droppedEl.index, pageIndex, position );
+		// Do nothing if the index didn't change.
+		if ( droppedEl.index !== arrangedIndex ) {
+			const pageId = pages[ droppedEl.index ].id;
+			arrangePage( { pageId, position: arrangedIndex } );
+			setCurrentPage( { pageId } );
```
dvoytenko: I know I'm a bit late with this question, but just to confirm: does reordering of pages ever change the currently shown page?

miina: Yes, currently it sets the page that was reordered as current. Perhaps it shouldn't be like this? That was an assumption but maybe the current page should not change? It made more sense without the page preview that we have now.

swissspidy: Opened an issue to confirm this: https://github.com/google/web-stories-wp/issues/141

dvoytenko: Yeah. I tried both ways and either seem pretty logical to me. I think this is low priority and let's assign to Sam.

PR #3931, opened 2019-12-14
Add/spa amp shadow dom updated with latest in develop

mehigh:
## Summary

Updated the spa-amp-shadow-dom branch with the latest in develop (8 files conflicted handled) and issued it against the branch part of this PR: https://github.com/ampproject/amp-wp/pull/1519 

I've fetched the latest from upstream and force pushed to match it:
https://github.com/xwp/amp-wp/

I've checked out locally the add/spa-amp-shadow-dom branch 
https://github.com/ampproject/amp-wp/pull/1519/files

I've pulled the latest in the develop branch and merged develop into this branch add/spa-amp-shadow-dom-updated using 3-way comparison to make sure nothing is missed.

## Checklist

- [x] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [ ] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [x] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

PR #4055, opened 2020-01-10
SPA with amp shadow DOM extras

mehigh:
## Summary

Issued it against the branch part of this PR: #1519

This contains some improvements to amp-wp-app-shell.js contributed by Jonathan Barnett ( @indieisaconcept ). Consent was [already provided](https://github.com/ampproject/amp-wp/pull/3931#issuecomment-572302342). 

Ensure .site-content-contain element exists

Rationale
A prior change introduced an early return for isHeaderVisible should the
site-branding element not be found in the document. However this early
return caused code to trigger elsewhere related to scrolling.

Changes

Ensure .site-content-contain document lookup is also guarded to prevent
calling scrollIntoView on a non-existing dom element.
Rationale
Due to caching restrictions with WPengine the use of a query-string parameter to
denote an inner shell request is problematic as it prevents the request from
being cached independently. The request needs to be distinctly different from
the outer shell in-order for it to be treated as a seperate cache item.

Changes

Query string parameter is now appended to the path instead of
Fixed bug where code was always expecting a .site-branding element to exist

## Checklist

### DIFF
--- a/assets/js/amp-wp-app-shell.js
+++ b/assets/js/amp-wp-app-shell.js
@@ -112,6 +112,9 @@
 	 */
 	function isHeaderVisible() {
 		const element = document.querySelector( '.site-branding' );
+		if ( ! element ) { 
+			return;
+		} 
 		const clientRect = element.getBoundingClientRect();
 		return clientRect.height + clientRect.top >= 0;
 	}
@@ -124,7 +127,8 @@
 	 */
 	function fetchShadowDocResponse( url ) {
 		const ampUrl = new URL( url );
-		ampUrl.searchParams.set( ampAppShell.componentQueryVar, 'inner' );
+		const pathSuffix = '_' + ampAppShell.componentQueryVar + '_inner';
+		ampUrl.pathname = ampUrl.pathname + pathSuffix;
 
 		return fetch( ampUrl.toString(), {
 			method: 'GET',
--- a/assets/js/amp-wp-app-shell.js
+++ b/assets/js/amp-wp-app-shell.js
@@ -240,8 +240,14 @@
 					}
 
 					if ( scrollIntoView && ! isHeaderVisible() ) {
+						const siteContent = document.querySelector( '.site-content-contain' );
+
+						if ( ! siteContent ) {
+							return;
+						}
+
 						// @todo The scroll position is not correct when admin bar is used. Consider scrolling to Y coordinate smoothly instead.
-						document.querySelector( '.site-content-contain' ).scrollIntoView( {
+						siteContent.scrollIntoView( {
 							block: 'start',
 							inline: 'start',
 							behavior: 'smooth'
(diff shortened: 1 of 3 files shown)


westonruter:
You can now rebase against `add/spa-amp-shadow-dom`. Note I independently also fixed  c5b6ee9 so you can omit that commit from being included.

mehigh:
Thank you for the feedback, glad to hear one of the items were already fixed. I'll only include the others.

mehigh:
This was rebased, only the remaining 4 commits are present now.

mehigh:
I consent to this being merged, and Jonathan Barnett gave it in a different PR where I originally added this in, so we should be good with the consent needs.

westonruter:
@indieisaconcept sorry, I need you to do another `@googlebot I consent.`

includes/class-amp-theme-support.php
```diff
@@ -2033,7 +2033,7 @@ public static function is_output_buffering() {
 	 */
 	public static function finish_output_buffering( $response ) {
 		self::$is_output_buffering = false;
-		return self::prepare_response( $response );
+		return apply_filters( 'amp_document_output', self::prepare_response( $response ) );
```
westonruter: This is a very powerful filter that has the possibility of overriding AMP validity. Any such mutations to the document should rather be done through sanitizers, which the AMP plugin will then guarantee to be valid AMP since the tag-and-attribute sanitizer runs last. So I think bb371b1c0115009d82289c49a72dd6009bababba should be reverted.

assets/js/amp-wp-app-shell.js
```diff
@@ -124,7 +127,8 @@
 	 */
 	function fetchShadowDocResponse( url ) {
 		const ampUrl = new URL( url );
-		ampUrl.searchParams.set( ampAppShell.componentQueryVar, 'inner' );
+		const pathSuffix = '_' + ampAppShell.componentQueryVar + '_inner';
+		ampUrl.pathname = ampUrl.pathname + pathSuffix;
```
westonruter, line 131: Does this not require the addition of new rewrite rules?

includes/class-amp-theme-support.php
indieisaconcept: @westonruter we found this feature to be necessary in-order to integrate non AMP compatible code into our documents. This was a conscious decision on our part & we were well aware of the implications of such a change in terms of AMP validation. As a publisher this level of flexibility is really important as unfortunately AMP compatibility is not a priority for some third-party services.

As an aside I think AMP validation in general needs to be overhauled to provide a broader degree of opt-in/opt-out flexibility but I see that as a larger discussion outside of the scope of this specific PR & better suited to [ampproject/amphtml](https://github.com/ampproject/amphtml).

indieisaconcept:
@googlebot I consent.

includes/class-amp-theme-support.php
westonruter: @indieisaconcept What specific changes did you need to make? There is now a way to do this officially via AMP dev mode, depending on your use case.

westonruter: I've reverted this for now in f59a9c9.

PR #4074, opened 2020-01-13
Improve compatibility of legacy stories editor with Gutenberg plugin

swissspidy:
## Summary

Addresses https://wordpress.org/support/topic/new-version-of-gutenberg-breaks-stories-layout/

* Updates obsolete class names in CSS
* Fixes selectors used for DOM manipulation
* Replace usage of deprecated `IconButton` with `Button`
* Hides some newly added elements from the UI

This way it works better with the Gutenberg plugin active, but doesn't fully work with WordPress 5.3 at the moment. Don't think it's worth trying to make it work for both.

## Checklist

### DIFF
--- a/assets/src/stories-editor/blocks/amp-story-cta/edit.css
+++ b/assets/src/stories-editor/blocks/amp-story-cta/edit.css
@@ -1,4 +1,4 @@
-.editor-block-list__layout div[data-type="amp/amp-story-cta"] {
+.block-editor-block-list__layout div[data-type="amp/amp-story-cta"] {
 	top: 80%;
 	left: 0;
 	width: 100%;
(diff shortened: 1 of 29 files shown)


westonruter:
Mileatoning for 1.4.3 assuming we can cherry-pick onto the 1.4 branch.

westonruter:
@swissspidy Do I understand correctly that this PR is not going forward? If so, is another PR needed which disables the AMP Stories experience if either the user is on WordPress version under 5.3 or they have a Gutenberg version that is too new?

In addition to these, we need to hide the Stories experience entirely (the experience checkboxes) if the user has no `amp_story` posts created. The thinking being that we can continue to allow access to Stories for users who are using them, but we discontinue offering the Stories feature for new installs or people who don't have stories created. In the place of the experience checkboxes, perhaps there should be a link to the Web Stories repo to get a peek of what's coming?

swissspidy:
Yeah that seems accurate for now.

We should write down all these things in #4172 and then create multiple PRs for the individual parts.

PR #4107, opened 2020-01-14
Preserve leading comments to main HTML elements

### LINKED ISSUE #4104: Validation broken in Genesis themes by Document::normalize_document_structure()
## Bug Description

I was testing out a Genesis theme and I saw in the admin bar that it was ✅ valid AMP. When I clicked to validate the page, however, I unexpectedly saw two validation errors:

image

This made no sense to me and I was racking my brain trying to figure it out. It turns out to be that the `link[rel=canonical]` element was located in the `body` during validation, when for non-validation requests it is in the `head` as expected. 

The issue turns out to be `\Amp\AmpWP\Dom\Document::normalize_document_structure()` introduced in #3758, namely that it is not currently written to account for HTML comments occuring before the doctype.

Normally, WordPress themes hard-code the `<!DOCTYPE html>`, `<html>`, and `<head>` tags. Genesis, however, uses `genesis_doctype()` to generate them. When a validation request is being performed, source stack comments are added before and after `genesis_doctype()` like so:

```html
<!--amp-source-stack {"type":"theme","name":"genesis","file":"lib\/structure\/header.php","line":28,"function":"genesis_do_doctype","hook":"genesis_doctype","priority":10}--><!DOCTYPE html>
<html lang="en-US">
<head >
<meta charset="UTF-8" />
<!--/amp-source-stack {"type":"theme","name":"genesis","file":"lib\/structure\/header.php","line":28,"function":"genesis_do_doctype","hook":"genesis_doctype","priority":10}-->
```

This has the effect of `Document::normalize_document_structure()` normalizing the document to start as:

```html
<!DOCTYPE html><html><head></head><body><!--amp-source-stack {"type":"theme","name":"genesis","file":"lib\/structure\/header.php","line":28,"function":"genesis_do_doctype","hook":"genesis_doctype","priority":10}--><!DOCTYPE html>
<html lang="en-US">
<head >
```

## Expected Behaviour

The DOM used for validation should be normalized to be the same as the DOM used for rendering.

## Steps to reproduce

1. Activate Genesis.
2. Enable Transitional or Standard mode in the AMP plugin.
3. Validate a URL.
4. Notice unexpected validation errors for `link[rel=canonical]` or other tags. 

---------------

## Acceptance criteria

* Validating a page generated by Genesis should not result in validation errors when there are none on the frontend.
* HTML comments that occur before and around the `<!DOCTYPE>`, `<html>`, and `<head>` tags must be preserved so that validation errors can be properly sourced.

## Implementation brief

* Perhaps capture the whitespace before/after the `<!DOCTYPE>`, `<html>`, and `<head>` and ensure that they are retained after normalization.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
## Summary

As we use HTML comments to track sources of each of the pieces that are written into the resulting HTML document, we need to ensure that all of these comments are properly preserved, especially around the main HTML elements that we are normalizing.

This PR changes the way the `<!doctype>` is handled so that it maintains and preceding HTML comments.

Fixes #4104

## Checklist


src/Dom/Document.php
```diff
@@ -219,6 +217,15 @@ final class Document extends DOMDocument {
...
 	private $mustache_tags_replaced = false;
 
+	/**
+	 * Whether we had secured a doctype that needs restoring or not.
+	 *
+	 * This is an int as it receives the $count from the preg_replace().
+	 *
+	 * @var int
+	 */
+	private $secured_doctype = 0;
```
westonruter: Does this need to be reset in the `reset` method?

westonruter:
Another test case to check here is `header.php` templates that start with:

```php
<!DOCTYPE html>
<!--[if IE 8]> <html <?php language_attributes(); ?> class="ie8"> <![endif]-->
<!--[if !IE]><!--> <html <?php language_attributes(); ?>> <!--<![endif]-->

<head>
	<meta charset="<?php bloginfo( 'charset' ); ?>">
	<title><?php wp_title( '|', true, 'right' ); ?></title>
```

Example taken from the [vision](https://wordpress.org/themes/vision/)  theme, which is also exhibiting this issue.

westonruter:
Some more examples from themes on WordPress.org.

## [ferry](https://wordpress.org/themes/ferry/)

The response starts with:

```html
<!-- =========================
     Page Breadcrumb
============================== -->
<!DOCTYPE html>
<html lang="en-US">
<head>
<meta charset="UTF-8"></meta>
<meta name="viewport" content="width=device-width, initial-scale=1"></meta>
...
```

## [catch-adaptive](https://wordpress.org/themes/catch-adaptive/)

The `header.php` starts with:

```php
<?php
/**
 * The default template for displaying header
 *
 * @package Catch Themes
 * @subpackage Catch Adaptive
 * @since Catch Adaptive 0.1
 */

	/** 
	 * catchadaptive_doctype hook
	 *
	 * @hooked catchadaptive_doctype -  10
	 *
	 */
	do_action( 'catchadaptive_doctype' );?>

<head>
<?php	
	/** 
	 * catchadaptive_before_wp_head hook
	 *
	 * @hooked catchadaptive_head -  10
	 * 
	 */
	do_action( 'catchadaptive_before_wp_head' );

	wp_head(); ?>
</head>

<body <?php body_class(); ?>>
```

## [edulab](https://wordpress.org/themes/edulab/)

The `header.php` starts with:

```php
<!doctype html>
<html <?php language_attributes(); ?>>

<!-- Head starts -->
<head itemscope itemtype="http://schema.org/WebSite">
	<meta charset="<?php bloginfo( 'charset' ); ?>">
	<meta name="viewport" content="width=device-width, initial-scale=1">
	<link rel="profile" href="https://gmpg.org/xfn/11">
	<?php wp_head(); ?>
</head><!-- Head ends -->
```

schlessera:
Regarding the conditional IE tags you mentioned above, we should discuss how to handle these. AMP doesn't support IE < 11, so we might just strip tags like the above (`if IE8`). Thoughts?

src/Dom/Document.php
schlessera: No, it mustn't be reset. It's a scalar value, so cloning is not an issue and it needs to stay in the state it was set.

schlessera:
> Somehow an additional meta[charset] is showing up when doing a validation request

I think that is a different issue, namely that the existing charset is not properly detected because it has a closing slash.

schlessera:
> Regarding the conditional IE tags you mentioned above, we should discuss how to handle these. AMP doesn't support IE < 11, so we might just strip tags like the above (if IE8). Thoughts?

Conditional comments seem to be exclusively supported by IE 5 to 9, none of which AMP supports. So I suggest we just strip these completely.

westonruter:
> Conditional comments seem to be exclusively supported by IE 5 to 9, none of which AMP supports. So I suggest we just strip these completely.

I think that may be fine, but the main issue is needing to make sure that comments that appear between the initial tags are preserved. They could be IE conditional comments or they could be HTML source stack comments.

src/Dom/Document.php
```diff
@@ -530,10 +547,10 @@ private function replace_self_closing_tags( $html ) {
 		static $regex_pattern = null;
 
 		if ( null === $regex_pattern ) {
-			$regex_pattern = '#<(' . implode( '|', self::$self_closing_tags ) . ')[^>]*>(?!</\1>)#';
+			$regex_pattern = '#<(' . implode( '|', self::$self_closing_tags ) . ')([^>]*?)(?:\s*\/)?>(?!</\1>)#';
```
westonruter: Good catch!

schlessera:
> the main issue is needing to make sure that comments that appear between the initial tags are preserved

The comments are already preserved now with the current state of the PR.

I'm currently adding a few more edge cases to the tests and adapting as needed to make it more robust.

westonruter:
Two test failures:

```
1) AMP_Tag_And_Attribute_Sanitizer_Test::test_replace_node_with_children_validation_errors with data set "nested_valid_and_invalud" ('\n					<div class="parent">\n...\n				', '\n<div class="parent">\n...\n				', array(array('invalid_p', 'div', 'DISALLOWED_TAG', array('invalid'), 'html_element_error'), array('bazfoo', 'div', 'DISALLOWED_TAG', array(), 'html_element_error')))
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     1 => '<div class="parent">'
     3 => '<p>'
     5 => 'Nesting valid and invalid elements.'
     7 => '</p>'
-    9 => ' Is an invalid "invalid" tagIs an invalid "foo" tag '
+    9 => ' Is an invalid "invalid" tag Is an invalid "foo" tag '

/app/public/content/plugins/amp/tests/php/test-tag-and-attribute-sanitizer.php:2726
/app/public/content/plugins/amp/tests/php/test-tag-and-attribute-sanitizer.php:2709

2) AMP_Tag_And_Attribute_Sanitizer_Test::test_replace_node_with_children_validation_errors with data set "bad_lili" ('\n					<ul>\n						<li>hello<...\n			', '<ul><li>hello</li> world</ul>', array(array('lili', 'ul', 'DISALLOWED_TAG', array(), 'html_element_error')))
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     1 => '<ul>'
     3 => '<li>'
     5 => 'hello'
     7 => '</li>'
-    9 => ' world'
+    9 => ' world '

/app/public/content/plugins/amp/tests/php/test-tag-and-attribute-sanitizer.php:2726
/app/public/content/plugins/amp/tests/php/test-tag-and-attribute-sanitizer.php:2709
```

westonruter:
It actually seems those broken tests actually indicate that you _fixed_ something in the plugin. Namely, whitespace was getting stripped when it should have been retained. Strangely, it seems that adding either `LIBXML_COMPACT` or `LIBXML_HTML_NODEFDTD` options result in this fix. 

Test updated in 9c9c74f.

schlessera:
I'm still curious why this would then have been correct on my local setup all the time while it was broken on Travis...?
Oh well, libxml is like a box of chocolates ...

PR #4108, opened 2020-01-14
Group dragging and rotating.

### LINKED ISSUE #3820: Moveable: dragging and rotating multi-selected elements
## Feature description

This is a follow-up issue for #3798.

Also to consider: keypress events for resizing, rotating, dragging. That's on hold though until we have information about the best practices using keyboard for doing that.

---------------

## Acceptance criteria

* It's possible to resize, drag, rotate multiple selected elements at once.

## Implementation brief

 

When more than one element is selected, one additional Moveable will be displayed which has all the selected elements as targets in an array (`target={ arrayOfTargets }`). Moveable displays an outline around the selected elements by default:
Screenshot 2019-12-06 at 18 02 55

For dragging, resizing, and rotating, we will need to set the initial values to the `frames` as an array of objects, e.g.
```
const frames = [
	{
		translate: [ 0, 0 ],
		rotate: rotationAngle1,
	},
	{
		translate: [ 0, 0 ],
		rotate: rotationAngle2,
	},
];
```
where rotation angle should be the respective initial rotation angle of each element.
### Dragging

Based on the documentation, this should work for updating the style for all the elements while dragging in `onDragGroup` (tested and it does work):
```
events.forEach( ( { target, beforeTranslate }, i ) => {
	const frame = frames[ i ];
	frame.translate = beforeTranslate;
	target.style.transform = `translate(${ beforeTranslate[ 0 ] }px, ${ beforeTranslate[ 1 ] }px)`;
} );
```
Then, in `onDragGroupEnd` we can update the `x` and `y` params of all the elements by adding the `translate` values to the original `x` and `y`.

### Rotating
Rotation works in a similar way with a difference that once we finish rotating, we can just set the rotation angle directly from the `frames` values as the updated element property since it already considers the initial rotation angle as well.

### Resizing
Resizing works in a similar way as resizing for a single element with the difference of having to loop through the targets to assign the values to each target separately.

---
Within this issue, in all three cases, the element values in the state are updated only when ending the activity, within `onResizeGroupEnd`, `onDragGroupEnd` and `onRotateGroupEnd`.

The selected elements should remain selected.

For simplicity, this implementation of this issue will not "care" about the following:
- Boundaries
- Resizing limits
- Keyboard events (common practices TBD by UX)
- Content within the elements

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

miina:
## Summary

Fixes #3820

## Checklist

### DIFF
--- a/assets/src/edit-story/components/canvas/multiSelectionMovable.js
+++ b/assets/src/edit-story/components/canvas/multiSelectionMovable.js
@@ -28,9 +28,6 @@ function MultiSelectionMovable( { selectedElements, nodesById } ) {
 	const { actions: { updateElementsById } } = useStory();
 
 	const targetList = selectedElements.map( ( element ) => {
-		if ( undefined === nodesById[ element.id ] ) {
-			return undefined;
-		}
 		return {
 			ref: nodesById[ element.id ],
 			id: element.id,
@@ -40,7 +37,7 @@ function MultiSelectionMovable( { selectedElements, nodesById } ) {
 		};
 	} );
 	// Not all targets have been defined yet.
-	if ( targetList.includes( undefined ) ) {
+	if ( targetList.some( ( { ref } ) => ref === undefined ) ) {
 		return null;
 	}
 
--- a/assets/src/edit-story/components/canvas/multiSelectionMovable.js
+++ b/assets/src/edit-story/components/canvas/multiSelectionMovable.js
@@ -54,7 +54,6 @@ function MultiSelectionMovable( { selectedElements, nodesById } ) {
 		target.style.transform = `translate(${ frame.translate[ 0 ] }px, ${ frame.translate[ 1 ] }px) rotate(${ frame.rotate }deg)`;
 	};
 
-	// @todo Is there any time when translate would differ among the elements? If not, we caould use just one translate for these.
 	const frames = targetList ? targetList.map( ( target ) => ( {
 		translate: [ 0, 0 ],
 		rotate: target.rotationAngle,
@@ -81,7 +80,7 @@ function MultiSelectionMovable( { selectedElements, nodesById } ) {
 			zIndex={ 0 }
 			target={ targetList.map( ( { ref } ) => ref ) }
 
-			// todo@: implement group transformations.
+			// @todo: implement group resizing.
 			draggable={ true }
 			resizable={ false }
 			rotatable={ true }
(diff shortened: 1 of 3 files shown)


assets/src/edit-story/components/canvas/multiSelectionMovable.js
```diff
@@ -22,23 +23,113 @@ function MultiSelectionMovable( { selectedElements, nodesById } ) {
...
+
+			onDragGroup={ ( { events } ) => {
+				events.forEach( ( { target, beforeTranslate }, i ) => {
+					const sFrame = frames[ i ];
+					sFrame.translate = beforeTranslate;
+					setTransformStyle( target, sFrame );
+				} );
+			} }
+			onDragGroupStart={ ( { events } ) => {
+				events.forEach( ( ev, i ) => {
```
spacedmonkey: This and logic `onRotateGroupStart` are very similar. Can these be extracted to a function?

spacedmonkey:
@googlebot I consent.

spacedmonkey:
I pushed a small change to allow for multiple value in a select menu https://github.com/ampproject/amp-wp/pull/4108/commits/7b51c279c4d3e3fdc1c082f4f21a4bf3be23a89a

spacedmonkey:
Other than a small bit of logic to change and something I fixed. This works great. Great work @miina

assets/src/edit-story/components/canvas/multiSelectionMovable.js
miina: Thanks for noting that, created a function for this and also for ending the event, see 0c1d528.

assets/src/edit-story/components/canvas/multiSelectionMovable.js
```diff
@@ -22,23 +23,120 @@ function MultiSelectionMovable( { selectedElements, nodesById } ) {
...
-	} );
+	}, [ selectedElements, moveable, nodesById ] );
+
+	const { actions: { updateElementsById } } = useStory();
 
-	const targets = selectedElements.map( ( element ) => nodesById[ element.id ] );
+	// Create targets list including nodes and also necessary attributes.
+	const targetList = selectedElements.map( ( element ) => {
+		return {
+			ref: nodesById[ element.id ],
```
dvoytenko: The `ref` naming here is slightly confusing since it usually indicates `Ref` type. Instead we have a `node`. Maybe rename to `node`?

miina: Good point, changed.

PR #4110, opened 2020-01-15
Work in progress: Rearrange layout of the editor

dvoytenko:
## Summary

This PR adds some more nesting in the editor's layout and unnests canvas. There are still couple of things that are not quite right here and I'm working on, but I wanted to get an early feedback due to some dependencies with other PRs. But what I was trying to do here is:

1. Cleanly separate the editing canvas from all the overlaying tools in terms of event propagation.
2. Ensure that overlaying tools never collide with selection, lasso, moveables, floating panels, etc.
3. Structurally ensure that canvas is about positioning of elements as on the page and outside. All other tools are not parented to it. E.g. an "inspector" is treated mostly the same as "page carousel" regardless of the layout we apply.

## Checklist


miina:
Makes sense to me, especially since I specifically ran into the issue of canvas being the parent of everything. Some questions:
1. I suppose this means that dragging elements to the non-canvas area should be restricted as well?
2. Should the meta and the carousel area (or the area they are in) be full-width and thus the canvas be "cut off" from the top and the bottom instead of still having canvas area around (or on the left and right) of the carousel/meta? I guess that's a UX question.

dvoytenko:
@miina, I think the answer to these both questions is the same: this approach allows us to have the right tools to do either in both cases. The layering separates the containers and gives each container more independence, but they are still related on some level. And thus we can decide on which layer we host a particular feature. I'll forward these questions to @samitron7, but I'd think the answers would be:

> 1. I suppose this means that dragging elements to the non-canvas area should be restricted as well?

A conditioned yes:
 - Dragging from media library should only be allowed on canvas
 - Dragging from (e.g.) desktop should be allowed on canvas and possibly library as well. If allowed on the library, however, the handler would be somewhat different, so the separation still helps.

> 2. Should the meta and the carousel area (or the area they are in) be full-width and thus the canvas be "cut off" from the top and the bottom instead of still having canvas area around (or on the left and right) of the carousel/meta? I guess that's a UX question.

I don't think they need to be full width. I'd think of them more as a "floating" overlays. They should have a clear padding though and rounded borders. Even though they might not be visible normally, they come in play when there's overlap. That being said, making them full width should be simple as well.

dvoytenko:
I'm gonna start breaking this PR down a bit to address independent issues.

dvoytenko:
Add-page changes are forked into https://github.com/ampproject/amp-wp/pull/4124

dvoytenko:
Closing this PR to rearrange according to the new workspace specs.

PR #4111, opened 2020-01-15
Fix invalid layout on non-AMP elements

### LINKED ISSUE #4036: Non-AMP element with height=100% results in invalid layout=fill added
## Bug Description

Given this markup generated by the Co-Blocks plugin'  `coblocks/shape-divider` block:

```html
<div class="wp-block-coblocks-shape-divider alignfull coblocks-shape-divider-11595047829 mb-0 mt-0" style="color:#111" aria-hidden="true">
	<div class="wp-block-coblocks-shape-divider__svg-wrapper" style="height:100px">
		<svg class="divider--wavy" height="100%" viewbox="0 0 100 10" width="100%" xmlns="http://www.w3.org/2000/svg" preserveaspectratio="none">
			<path d="m42.19.65c2.26-.25 5.15.04 7.55.53 2.36.49 7.09 2.35 10.05 3.57 7.58 3.22 13.37 4.45 19.26 4.97 2.36.21 4.87.35 10.34-.25s10.62-2.56 10.62-2.56v-6.91h-100.01v3.03s7.2 3.26 15.84 3.05c3.92-.07 9.28-.67 13.4-2.24 2.12-.81 5.22-1.82 7.97-2.42 2.72-.63 3.95-.67 4.98-.77z" fill-rule="evenodd" transform="matrix(1 0 0 -1 0 10)"></path>
		</svg>
	</div>
	<div class="wp-block-coblocks-shape-divider__alt-wrapper" style="height:50px"></div>
</div>
```

This is resulting in a validation error:

image

I believe this is a regression introduced in #3728.

## Expected Behaviour

The `layout` attribute should only be added to AMP components. Handling a `height=100%` attribute on a non-AMP should get converted to a `style="height:100%"`.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

This PR addresses the regression introduced in #3728 where in some cases non-AMP components were being sanitized and the `layout` attribute was appended to the node.

Also, this handles the conversion of percentage-based height and width attribute values to inline style properties. Do note, however, that if the node already has such a style property in place along with the corresponding dimension attribute value, no action will be taken place. For example, given:

```html
<svg width="50%" style="width:100%; height:80%">
```

The width value `50%` will not replace the width style property `100%`.

---

Fixes #4036.

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-layout-sanitizer.php
+++ b/includes/sanitizers/class-amp-layout-sanitizer.php
@@ -20,6 +20,52 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 	public function sanitize() {
 		$xpath = new DOMXPath( $this->dom );
 
+		/*
+		 * Convert all percentage-based width or height values into style properties.
+		 *
+		 * The following query could be made simpler by using the `ends-with` function, but it is not a valid function in
+		 * XPath 1.0, which the `DOMXPath` class uses.
+		 */
+		$nodes = $xpath->query( '//*[ "%" = substring( @width, string-length( @width ) ) or "%" = substring( @height, string-length( @height ) ) ]' );
+
+		foreach ( $nodes as $node ) {
+			$width  = $node->getAttribute( 'width' );
+			$height = $node->getAttribute( 'height' );
+			$style  = $node->getAttribute( 'style' );
+
+			$styles         = $this->is_empty_attribute_value( $style ) ? [] : $this->parse_style_string( $style );
+			$attr_converted = false;
+
+			// Convert the percentage-based width attribute to a style property.
+			if ( ! isset( $styles['width'] ) && '%' === substr( $width, -1 ) ) {
+				// Ignore if its an AMP component and the width is `100%`.
+				if ( '100%' === $width && strpos( $node->tagName, 'amp-' ) === 0 ) {
+					continue;
+				}
+
+				$styles['width'] = $width;
+				$attr_converted  = true;
+				$node->removeAttribute( 'width' );
+			}
+
+			// Convert the percentage-based height attribute to a style property.
+			if ( ! isset( $styles['height'] ) && '%' === substr( $height, -1 ) ) {
+				// Ignore if its an AMP component and the height is `100%`.
+				if ( '100%' === $height && strpos( $node->tagName, 'amp-' ) === 0 ) {
+					continue;
+				}
+
+				$styles['height'] = $height;
+				$attr_converted   = true;
+				$node->removeAttribute( 'height' );
+			}
+
+			// If either dimension was converted, update the style property with it.
+			if ( $attr_converted ) {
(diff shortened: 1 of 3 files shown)


schlessera:
Isn't that something the tree-shaker should take into account? I mean, the produced outcome from the tree-shaker seems incorrect, so it is missing a rule, I guess?

westonruter:
> I mean, the produced outcome from the tree-shaker seems incorrect, so it is missing a rule, I guess?

Are you referring to this rule?

image

It's an internal browser rule derived from the `width` and `height` HTML attributes, which aren't available to us. Once we lift those from the HTML attribute into CSS, then we override rules with less specificity.

pierlon:
I suppose the conversion of the `width`/`height` attributes to style properties may have to be reverted then.

> Something else to consider is whether or not the actual amp-layout element should be considered?

I'm not sure how this would be applied. Will it only be for certain elements? Will the aspect ratio always be the same?

schlessera:
> It's an internal browser rule derived from the width and height HTML attributes, which aren't available to us. Once we lift those from the HTML attribute into CSS, then we override rules with less specificity.

Wouldn't those be consistent across all browsers? We could think about encoding them for the tree-shaker, so that deviating from them will create added styling as needed to keep the end-result for the browser the same.

westonruter:
> Wouldn't those be consistent across all (relevant) browsers? We could think about encoding them for the tree-shaker, so that deviating from them will create added styling as needed to keep the end-result for the browser the same.

How would they be encoded? I can't envision what you are proposing or how it would work.

westonruter:
At issue is a `width="100%"` effectively creates a browser style rule. Any style rule defined in the website CSS will always override such rules, as I just was reaffirmed of today when seeing this: https://wattenberger.com/blog/css-cascade#origin

> image

pierlon:
Thanks for the reaffirmation. I'll be reverting https://github.com/ampproject/amp-wp/pull/4111/commits/bbc87ad47a51860aeff8db817f7effb532be818e then.

includes/sanitizers/class-amp-layout-sanitizer.php
```diff
@@ -20,8 +20,11 @@ class AMP_Layout_Sanitizer extends AMP_Base_Sanitizer {
 	public function sanitize() {
 		$xpath = new DOMXPath( $this->dom );
 
-		// Elements with the `layout` attribute will be validated by `AMP_Tag_And_Attribute_Sanitizer`.
-		$nodes = $xpath->query( '//*[ not( @layout ) and ( @data-amp-layout or @width or @height or @style ) ]' );
+		/**
```
westonruter: Just noticed this, that it should have been `/*` instead of `/**` since it is not a phpDoc comment. Not a big deal but we can fix it as part of another ongoing PR.

PR #4122, opened 2020-01-16
First pass at document sidebar

### LINKED ISSUE #3784: Sidebar / Document Inspector functionality.
## Feature description

This ticket was made out of this [conversation in slack](https://xwp.slack.com/archives/CCCAUAH6F/p1574088174001600). 

"Old" Editor:
image

---------------

## Acceptance criteria

* The post visibility: Private functionality is ported into the new editor
* The post visibility: Public functionality is ported into the new editor
* The post visibility: Password Protected functionality is ported into the new editor
* The post status: Published functionality is ported into the new editor
* The post status: Draft functionality is ported into the new editor
* The post status: Scheduled functionality ( selection of publish date ) is ported into the new editor
* The post author functionality is ported into the new editor
* The move post to trash functionality is ported into the new editor
* The post permalink functionality is ported into the new editor
* The post excerpt functionality is ported into the new editor
* The featured image functionality for setting meta tags as primary image is ported into the new editor
* Author assignment and publish fields should only show to user that have this capability. 

## Implementation brief

* Load a list of users into state from REST API
* Load a list of post into state from REST API
* Do capability checks for publish / select author. If user does not have cap, do not show option. 
* Refactor uploadMedia to be more reusable. 
* Add field to REST API to get feature image. 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

spacedmonkey:
## Summary

Fixes #3784

## Checklist


assets/src/edit-story/components/header/buttons.js
```diff
@@ -71,7 +71,20 @@ function Publish() {
...
-	const text = ( status !== 'publish' ) ? __( 'Publish' ) : __( 'Update' );
+	let text;
+
+	switch ( status ) {
+		case 'publish':
+		case 'private':
+			text = __( 'Update' );
+			break;
+		case 'future':
+			text = __( 'Sceduled' );
```
swissspidy:
```suggestion
			text = __( 'Scheduled' );
```

assets/src/edit-story/app/story/useStoryReducer/reducers/restore.js
```diff
@@ -12,12 +12,13 @@
...
  * @return {Object} New state
  */
-function restore( state, { pages, current, selection, story } ) {
+function restore( state, { pages, current, selection, story, capabilities } ) {
 	if ( ! Array.isArray( pages ) || pages.length === 0 ) {
 		return state;
 	}
 
 	const newStory = typeof story === 'object' ? story : {};
+	const newCapabilities = typeof story === 'object' ? capabilities : {};
```
swissspidy: Should this be like this?

```suggestion
	const newCapabilities = typeof capabilities === 'object' ? capabilities : {};
```

assets/src/edit-story/app/story/effects/useLoadStory.js
```diff
@@ -25,31 +35,50 @@ function useLoadStory( {
 					status,
 					author,
 					slug,
+					date,
+					modified,
+					excerpt: { raw: excerpt },
```
miina: I wonder if excerpt is going to be used for Stories? It wasn't in use before at least, IIRC. Perhaps would be good to confirm? (Unless you did already)

assets/src/edit-story/app/story/effects/useLoadStory.js
```diff
@@ -9,6 +14,10 @@ import { useEffect } from '@wordpress/element';
 import { useAPI, useHistory } from '../../';
 import { createPage } from '../../../elements';
 
+const getPerm = ( post, field ) => {
```
miina: Could we write the name of this func out for clarity or add a comment?

assets/src/edit-story/components/inspector/documentInspector.js
```diff
@@ -1,9 +1,187 @@
...
+		state: { meta: { isSaving }, story: { author, status, slug, date, excerpt, featuredMediaUrl, password }, capabilities },
+		actions: { updateStory, deleteStory },
+	} = useStory();
+
+	const { postThumbnails } = useConfig();
+
+	useEffect( () => {
+		loadStatuses();
+		loadUsers();
+	} );
```
miina: Should this have dependencies? Or is it supposed to always run?

assets/src/edit-story/components/inspector/documentInspector.js
```diff
@@ -1,9 +1,187 @@
...
+
+	useEffect( () => {
+		const states = ( status === 'future' ) ? [ 'pending' ] : [ 'future', 'pending' ];
+		setDisabledStatuses( states );
+	}, [ status, setDisabledStatuses ] );
+
+	const handleChangeStatus = useCallback(
+		( evt ) => 	updateStory( { properties: { status: evt.target.value } } ),
```
miina: ^ Indent issue

assets/src/edit-story/components/inspector/documentInspector.js
```diff
@@ -1,9 +1,187 @@
...
+
 	return (
-		<h2>
-			{ 'Document' }
-		</h2>
+		<>
+			<h2>
+				{ 'Document' }
+			</h2>
+			{ capabilities && capabilities.hasPublishAction && statuses && <SelectMenu
```
miina: Would be great to make the naming a bit more clear, e.g. use `canPublish` instead of `hasPublishAction` etc. Thoughts?

assets/src/edit-story/app/config/context.js
```diff
@@ -3,4 +3,4 @@
  */
 import { createContext } from '@wordpress/element';
 
-export default createContext( { api: {} } );
+export default createContext( { api: {}, statuses: {} } );
```
miina: Isn't `statuses` part of `api`?

assets/src/edit-story/components/inspector/documentInspector.js
```diff
@@ -1,9 +1,187 @@
...
+				label="Status"
+				name="status"
+				options={ statuses }
+				disabled={ isSaving }
+				disabledOptions={ disabledStatuses }
```
miina: Would it be more clear to filter the statuses right away instead of passing a separate `disabledOptions`?

assets/src/edit-story/components/inspector/documentInspector.js
```diff
@@ -1,9 +1,187 @@
...
+				value={ status }
+				onChange={ handleChangeStatus }
+			/> }
+			{ capabilities && capabilities.hasPublishAction && status !== 'private' && <InputGroup
```
miina: When testing, I currently can't see the option for setting the visibility to Password Protected (just the password field). Not sure if I'm missing something?

includes/class-amp-story-post-type.php
```diff
@@ -407,10 +412,12 @@ public static function load_admin_fonts( $post ) {
 		if ( $post_story_data ) {
 			foreach ( $post_story_data as $page ) {
 				foreach ( $page['elements'] as $element ) {
-					$font = AMP_Fonts::get_font( $element['fontFamily'] );
-					if ( $font && ! in_array( $font['slug'], $font_slugs, true ) ) {
-						$fonts[]      = $font;
-						$font_slugs[] = $font['slug'];
+					if ( isset( $element['fontFamily'] ) ) {
```
miina: Could we do `continue` if `! isset( $element['fontFamily'] )` instead? To reduce nesting and improve readability.

assets/src/edit-story/components/inspector/documentInspector.js
```diff
@@ -1,9 +1,187 @@
...
+
+			<RemoveButton onClick={ handleRemoveStory } dangerouslySetInnerHTML={ { __html: 'Move to trash' } } />
+			<InputGroup
+				label={ 'Published date' }
+				type={ 'datetime-local' }
+				value={ date }
+				disabled={ isSaving }
+				onChange={ handleChangeDate }
+			/>
+			{ capabilities && capabilities.hasAssignAuthorAction && users && <SelectMenu
```
swissspidy: It seems like something weird is going on with the capability checks.

1. Edit story.
1. Assign story to user with subscriber role
1. Update story
1. Click Undo button.
1. See that the author field and others disappear unless the page is reloaded.

This might be an issue with Undo and not the capability checks added here, but I wanted to flag it either way.

swissspidy: Ah, looks like @miina flagged the same thing too :-)

Resolving this as a duplicate.

assets/src/edit-story/app/story/useStoryReducer/reducers/restore.js
```diff
@@ -26,6 +27,7 @@ function restore( state, { pages, current, selection, story } ) {
 		current: newCurrent,
 		selection: newSelection,
 		story: newStory,
+		capabilities: newCapabilities,
```
barklund: Please remember to add it to [the initial state](https://github.com/ampproject/amp-wp/blob/feature/document-sidebar1/assets/src/edit-story/app/story/useStoryReducer/useStoryReducer.js#L12-L17) of the reducer.

Also, please remember to update all tests (ensure 100% test coverage in [`reducers`](https://github.com/ampproject/amp-wp/tree/feature/document-sidebar1/assets/src/edit-story/app/story/useStoryReducer/reducers) folder).

assets/src/edit-story/components/header/title.js
```diff
@@ -25,26 +27,42 @@ const Input = styled.input`
...
+
 	const handleChange = useCallback(
 		( evt ) => updateStory( { properties: { title: evt.target.value } } ),
 		[ updateStory ],
 	);
 
+	const handleBlur = useCallback(
+		() => {
+			if ( ! slug || slug === storyId ) {
+				const clearSlug = encodeURIComponent( cleanForSlug( titleFormatted( title ) ) ) || storyId;
```
barklund: I don't understand the name `clearSlug` here.

assets/src/edit-story/components/header/title.js
```diff
@@ -25,26 +27,42 @@ const Input = styled.input`
...
 		return null;
 	}
 
-	// TODO Make sure that Auto Draft checks translations.
-	const titleFormatted = ( status === 'auto-draft' && title === 'Auto Draft' ) ? '' : title;
-
 	return (
 		<Input
-			value={ titleFormatted }
+			value={ titleFormatted( title ) }
```
barklund: Doesn't this give warnings about switching from controlled to uncontrolled or is that not an issue?

assets/src/edit-story/components/inspector/documentInspector.js
```diff
@@ -1,9 +1,187 @@
...
+
+	const handleChangeExcerpt = useCallback(
+		( value, evt ) => updateStory( { properties: { excerpt: evt.target.value } } ),
+		[ updateStory ],
+	);
+
+	const handleChangeSlug = useCallback(
+		( value, evt ) => updateStory( { properties: { slug: evt.target.value } } ),
+		[ updateStory ],
+	);
```
barklund: These can all be generalised to:

```js
	const handleChangeValue = useCallback(
		( prop ) => ( value, evt ) => updateStory( { properties: { [ prop ]: evt.target.value } } ),
		[ updateStory ],
	);
```

And then further down for the individual input:

```js
		onChange={ handleChangeValue( 'slug' ) }
```

assets/src/edit-story/components/inspector/documentInspector.js
```diff
@@ -1,9 +1,187 @@
...
+	display: block;
+	align-items: center;
+	margin: 15px 0px;
+`;
+
+const RemoveButton = styled.button`
+	${ ButtonCSS }
+`;
 
 function DocumentInspector() {
```
barklund: This component is way too big and should be split up somehow. I don't know what would be a natural split off the top of my head, but it should have concerns split in several smaller components.

assets/src/edit-story/components/inspector/inspectorProvider.js
```diff
@@ -6,26 +6,78 @@ import PropTypes from 'prop-types';
...
 const PREPUBLISH = 'prepublish';
 
 function InspectorProvider( { children } ) {
+	const { actions: { getAllStatuses, getAllUsers } } = useAPI();
 	const [ tab, setTab ] = useState( DESIGN );
+	const [ users, setUsers ] = useState( [] );
+	const [ statuses, setStatuses ] = useState( [] );
+
+	const [ isUsersLoading, setIsUsersLoading ] = useState( false );
+	const [ isStatusesLoading, setIsStatusesLoading ] = useState( false );
```
barklund: This ought to be changed to two instances of the same reducer-based hook, that can update a list and remember loading state (and probably handle error too).

assets/src/edit-story/components/inspector/inspectorProvider.js
```diff
@@ -6,26 +6,78 @@ import PropTypes from 'prop-types';
...
+	const [ disabledStatuses, setDisabledStatuses ] = useState( [] );
+
+	const loadStatuses = useCallback( () => {
+		if ( ! isStatusesLoading && statuses.length === 0 ) {
+			setIsStatusesLoading( true );
+			getAllStatuses().then( ( data ) => {
+				data = Object.values( data );
+				// eslint-disable-next-line camelcase
+				data = data.filter( ( { show_in_list } ) => show_in_list );
```
barklund: You could also just rename it inline to avoid the disable such as:

```js
				data = data.filter( ( { show_in_list:isShown } ) => isShown );
```

But yes, it's a bit silly.

assets/src/edit-story/components/inspector/inspectorProvider.js
```diff
@@ -6,26 +6,78 @@ import PropTypes from 'prop-types';
...
 	const [ tab, setTab ] = useState( DESIGN );
+	const [ users, setUsers ] = useState( [] );
+	const [ statuses, setStatuses ] = useState( [] );
+
+	const [ isUsersLoading, setIsUsersLoading ] = useState( false );
+	const [ isStatusesLoading, setIsStatusesLoading ] = useState( false );
+	const [ disabledStatuses, setDisabledStatuses ] = useState( [] );
+
+	const loadStatuses = useCallback( () => {
+		if ( ! isStatusesLoading && statuses.length === 0 ) {
```
barklund: I'd much rather not call this callback if there already are statuses rather than check every time.

assets/src/edit-story/components/inspector/inspectorProvider.js
```diff
@@ -6,26 +6,78 @@ import PropTypes from 'prop-types';
...
+				} ) => ( {
+					value: id,
+					name,
+				} ) );
+
+				setUsers( saveData );
+				setIsUsersLoading( false );
+			} );
+		}
+	}, [ isUsersLoading, users.length, getAllUsers ] );
```
barklund: These two callbacks are also very similar now and could be baked into the above-mentioned reducer-based hooks, that will manage their loading. Will make this much cleaner.

assets/src/edit-story/components/inspector/documentInspector.js
```diff
@@ -1,9 +1,187 @@
...
+
+	useEffect( () => {
+		loadStatuses();
+		loadUsers();
+	} );
+
+	useEffect( () => {
+		const states = ( status === 'future' ) ? [ 'pending' ] : [ 'future', 'pending' ];
+		setDisabledStatuses( states );
+	}, [ status, setDisabledStatuses ] );
```
barklund: Disabled statuses are only ever used in this component, are they not? Can this just be converted to:

```js
const disabledStatuses = useMemo( () => ( status === 'future' ) ? [ 'pending' ] : [ 'future', 'pending' ], [ status ] );
```

And then everything about disabled statues in the context and provider can be completely deleted? Unless it's required elsewhere of course.

Memoizing isn't even that important, but it looks nice and saves on some re-renders.

barklund:
Really great effort here, adding a lot of new valuable functionality to the editor! Don't let all the comments confuse you, excellent job! 👍

assets/src/edit-story/app/config/context.js
spacedmonkey: Original post statuses came in via config. I didnt remove this. Removed in https://github.com/ampproject/amp-wp/pull/4122/commits/5f9eb6f5b5dc81a15cc5e571b1f3e6412cd9f2a0

assets/src/edit-story/app/story/effects/useLoadStory.js
spacedmonkey: Added comment here https://github.com/ampproject/amp-wp/pull/4122/commits/6e1e71e9397d71efad6685a01ef9111648c0d876

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: Good catch. Fixed in https://github.com/ampproject/amp-wp/pull/4122/commits/9c4c650652a0962d34eb04c0b0f682773f412aa8

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: This one was copied from gutenberg - https://github.com/WordPress/gutenberg/blob/94f54892c214ebd014ec126a9575b7e902198ad9/packages/editor/src/components/post-pending-status/check.js#L24 
Not against changing it...

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: Can you explain this, not sure what you mean here.

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: You are not. Current there is not check to make password protected. Currently, a post becomes password protected by adding a password. I will change this when UX comes in.

assets/src/edit-story/components/inspector/inspectorProvider.js
spacedmonkey: Works for me. Fixed in https://github.com/ampproject/amp-wp/pull/4122/commits/2ca09099dd8f3274206534faa0fec332ec7890c9

includes/class-amp-story-post-type.php
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/4122/commits/b7aa6d26c7839c7134782ecd0113fd9a66333697

assets/src/edit-story/components/inspector/inspectorProvider.js
spacedmonkey: Did you have something in mind here?

assets/src/edit-story/app/story/effects/useLoadStory.js
spacedmonkey: @pbakaus Thoughts on excerpt?

assets/src/edit-story/app/story/useStoryReducer/reducers/restore.js
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/4122/commits/bae3059721739ba3ffa181c0fda173da440ddff3

assets/src/edit-story/components/header/title.js
spacedmonkey: Changed in https://github.com/ampproject/amp-wp/pull/4122/commits/27a750e4e53b9418c7ea0f6a7adbfdd1806c60d7

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: Good shout. https://github.com/ampproject/amp-wp/pull/4122/commits/f6fd82f3eae64b9f1e3f01f29358ab967618091a

assets/src/edit-story/components/inspector/inspectorProvider.js
spacedmonkey: This two functions are similar but not the same. The output of the two different api calls are different so require some different logic inside them, making it not easy to refactor into one generate function with params. If I am missing something, please go ahead @barklund give me a code snippet.

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: These should always run.

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: I think you are correct here. There is a lot and is makes it a little hard to follow. But until we know the final UX, it seems unwise to split this now. This PR is just so we can get the functionality in, we can style it up and move it around later.

assets/src/edit-story/components/header/title.js
spacedmonkey: Can you explain this a little. I haven't seen any issue as of yet.

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: This callback has been removed.

spacedmonkey:
> * Undo and Redo don't seem to work for at least some of the fields (didn't test all), e.g. Status and some others.
> * Undo and Redo _do something_ for some of the fields but incorrectly. E.g. changing the slug and then clicking Undo a few times will result in not displaying some of the fields at all:

Good spot @miina and @swissspidy . I have fixed the issue with undo / redo. It was simply a matter of passing the capabilities to the restored state. I had missed a couple of places. See this commit https://github.com/ampproject/amp-wp/pull/4122/commits/62ac276fac3a40c6f2c60b7af27030ccb849d60c

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/commit/62ac276fac3a40c6f2c60b7af27030ccb849d60c

assets/src/edit-story/components/inspector/documentInspector.js
miina: What I meant is that perhaps instead of having a separate `disabledOptions` attribute, we could directly just filter `statuses` and pass the correct `options` right away, instead of the `SelectMenu` having to be aware of irrelevant options.

E.g. `options={ statuses.filter( status ) => ! disabledStatuses.includes( status ) }` or sth similar.

 Thoughts?

assets/src/edit-story/components/inspector/documentInspector.js
miina: Hmm, perhaps it shouldn't try to load these if there are already statuses/users loaded? Otherwise, it will try to do this with every render.

assets/src/edit-story/app/story/effects/useHistoryEntry.js
```diff
@@ -22,8 +23,9 @@ function useHistoryEntry( {
 			current,
 			pages,
 			selection,
+			capabilities,
 		} );
-	}, [ appendToHistory, story, current, pages, selection ] );
+	}, [ appendToHistory, story, current, pages, selection, capabilities ] );
```
miina: Actually, (I might be missing something but) why are capabilities relevant to history entry? Aren't capabilities static and coming from DB, and "Undo" and "Redo" shouldn't influence the capabilities? Thoughts?

spacedmonkey: ATM capabilities are stored in the story state. If capabilities are not copied when state is changed, then the capabilities lost and the error you saw happens. 

That is an argument to be said that they shouldn't be stored there in the first place. That is another discussion.

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: No it will not. In those callbacks it checks to see if they have been loaded. See https://github.com/ampproject/amp-wp/pull/4122/files#diff-89184968733e534046330610da4d7328R31

assets/src/edit-story/components/inspector/documentInspector.js
spacedmonkey: Works for me.

spacedmonkey: Fixed in https://github.com/ampproject/amp-wp/pull/4122/commits/c47fcf1e39e549cd797892635c1986946bedc0db

assets/src/edit-story/app/story/effects/useHistoryEntry.js
swissspidy: Editor-wide state was a topic the other day already. Thoughts, @barklund?

barklund: I did look at this one, and didn't really know how to feel about it. On one hand, it doesn't feel like something that belongs inside the story reducer, as it's not in any way part of the story nor is it anything you can edit. But one the other hand, where else would it go?

I would probably prefer this being taken back out if the story reducer, but not completely sure about where to put it instead. Maybe it's an API level thing and belongs in the API state? It's a tricky one.

assets/src/edit-story/app/story/effects/useLoadStory.js
pbakaus: I think this might be useful to support as the excerpt is often used as description of something. But I don't think it's P0.

assets/src/edit-story/app/story/effects/useHistoryEntry.js
miina: Sounds good to me to merge as-is for the purposes of this issue, and open a separate ticket for discussing how to store it "properly".

pbakaus: sgtm.

assets/src/edit-story/app/story/useStoryReducer/reducers/restore.js
```diff
@@ -12,12 +12,13 @@
...
  * @return {Object} New state
  */
-function restore( state, { pages, current, selection, story } ) {
+function restore( state, { pages, current, selection, story, capabilities } ) {
 	if ( ! Array.isArray( pages ) || pages.length === 0 ) {
 		return state;
 	}
 
 	const newStory = typeof story === 'object' ? story : {};
+	const newCapabilities = typeof capabilities === 'object' ? capabilities : {};
```
dvoytenko: Just accidentally noticed this. Is this still ok given that `typeof null === 'object'`?

assets/src/edit-story/components/header/title.js
dvoytenko: Looks like title is never `null`? So there's no controlled/uncontrolled issue?

barklund: Seems to work fine, my bad.

PR #4106, opened 2020-01-14
Rework page carousel

swissspidy:
## Summary

Reworks page carousel to match updated design. Supersedes #3958.

Fixes #3834.

Screenshots:

Screenshot 2020-01-14 at 16 12 11
Screenshot 2020-01-14 at 16 11 58

## Checklist

- [x] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [x] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [x] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

Issue #3920, opened 2019-12-11
Responsive design


barklund:
## Feature description

We need responsive design making the editor look good in every size down to iPad (768x1024) and up to WQHD (2560 x 1440) – and of course everything in-between.

How, when and what is as yet undefined.

Product priority first, then UX, then engineering.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

PR #4129, opened 2020-01-18
Preview: reimplement workspace layering

dvoytenko:
## Summary

Supersedes #4110. This is a major redo of the workspace layering to support new "Anatomy of the workspace" in designs (/cc @samitron7). The current state is very preliminary to get the feedback ASAP. Many features are still broken.

The following layering has been implemented:
WP Story editor

This might seem complicated, but this is the most straightforward way I see at this moment that supports the needed features, while being less fragile overall. Key notes here:

* Layers are layered on top of each other using `position:absolute; inset:0`.
* All layers use the same `grid-template` to ensure that the positions of the areas always match.
* Lasso layer is at the very bottom to allow lasso to start on the page's area and outside of it.
* Display area is right on top of lasso layer.
* Critically, the display area is now clipped to borders (`overflow:hidden`). And element overflowing off the side of the page is no longer displayed.
* Tools layer (`NavLayer` in code) overlays the headers, footers, menus on top of the display area. The event propagation is blocked to avoid lasso.
* Frames & selection are on top of both display and tools. Per spec, frames and selection has to be shown on top of everything else in the workplace (i.e. never clipped). The frames and selection are also selectable/interactable outside the page area, even if on top of other workspace elements.
* Frames also show outline on hover. The gray outline is only for debugging. Note that frames are not obscured by other elements either.
* `Display` templates are split into two categories: `Display` and `Frame`. `Display` is only to render the element: all event handlers are removed. `Frame` would normally only render the outline, but it handles all events (focus, mouse, keyboard).
* Edit layer now moves the `Edit` mode into a separate layer. This is done so that the editing element is always on top and we can, when needed, show graypane over the workspace, safe/unsafe area, etc - as requested by the UX.

You can see the actual layers by changing `DEBUG_ANGLE` constant to a non-zero value, e.g. `const DEBUG_ANGLE = 60;`.

## Checklist

### DIFF
--- a/assets/src/edit-story/components/canvas/canvasLayout.js
+++ b/assets/src/edit-story/components/canvas/canvasLayout.js
@@ -6,60 +6,29 @@ import styled from 'styled-components';
 /**
  * Internal dependencies
  */
-import { PAGE_NAV_WIDTH, PAGE_WIDTH, PAGE_HEIGHT } from '../../constants';
-import Page from './page';
-import PageMenu from './pagemenu';
-import PageNav from './pagenav';
-import Carousel from './carousel';
+import EditLayer from './editLayer';
+import DisplayLayer from './page';
+import FramesLayer from './framesLayer';
+import NavLayer from './navLayer';
 import SelectionCanvas from './selectionCanvas';
-
-const PAGE_PADDING = 30;
 
 const Background = styled.div`
 	background-color: ${ ( { theme } ) => theme.colors.bg.v1 };
-	display: flex;
-	flex-direction: column;
-	align-items: center;
-	justify-content: center;
-	height: 100%;
-
-	display: grid;
-	grid:
-    ".         .         .         .         .       " minmax(${ PAGE_PADDING }px, 1fr)
-    ".         prev      page      next      .       " ${ PAGE_HEIGHT }px
-    ".         .         menu      .         .       " 48px
-    ".         .         .         .         .       " minmax(${ PAGE_PADDING }px, 1fr)
-    "carousel  carousel  carousel  carousel  carousel" auto
-    / 1fr ${ PAGE_NAV_WIDTH }px ${ PAGE_WIDTH }px ${ PAGE_NAV_WIDTH }px 1fr;
-`;
-
-const Area = styled.div`
-	grid-area: ${ ( { area } ) => area };
-	height: 100%;
 	width: 100%;
+	height: 100%;
+	position: relative;
+	user-select: none;
 `;
 
 function CanvasLayout() {
-	// @todo SelectionCanvas should not be there, will be addressed separately.
 	return (
 		<Background>
-			<Area area="page">
-				<SelectionCanvas>
-					<Page />
-				</SelectionCanvas>
-			</Area>
-			<Area area="menu">
-				<PageMenu />
-			</Area>
-			<Area area="prev">
-				<PageNav isNext={ false } />
-			</Area>
-			<Area area="next">
-				<PageNav />
-			</Area>
-			<Area area="carousel">
-				<Carousel />
-			</Area>
+			<SelectionCanvas>
+				<DisplayLayer />
+				<NavLayer />
+				<FramesLayer />
+			</SelectionCanvas>
+			<EditLayer />
 		</Background>
 	);
 }
--- a/assets/src/edit-story/components/canvas/canvasLayout.js
+++ b/assets/src/edit-story/components/canvas/canvasLayout.js
@@ -7,7 +7,7 @@ import styled from 'styled-components';
  * Internal dependencies
  */
 import EditLayer from './editLayer';
-import DisplayLayer from './page';
+import DisplayLayer from './displayLayer';
 import FramesLayer from './framesLayer';
 import NavLayer from './navLayer';
 import SelectionCanvas from './selectionCanvas';
--- a/assets/src/edit-story/components/canvas/elementEdit.js
+++ b/assets/src/edit-story/components/canvas/elementEdit.js
@@ -1,0 +1,57 @@
+/**
+ * External dependencies
+ */
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+/**
+ * Internal dependencies
+ */
+import { getDefinitionForType } from '../../elements';
+import { ElementWithPosition, ElementWithSize, ElementWithRotation, getBox } from '../../elements/shared';
+
+// Background color is used to make the edited element more prominent and
+// easier to see.
+const Wrapper = styled.div`
+  ${ ElementWithPosition }
+  ${ ElementWithSize }
+  ${ ElementWithRotation }
+  pointer-events: initial;
+	background-color: ${ ( { theme } ) => theme.colors.whiteout };
+`;
+
+function ElementEdit( {
+	element: {
+		id,
+		type,
+		x,
+		y,
+		width,
+		height,
+		rotationAngle,
+		isFullbleed,
+		...rest
+	},
+} ) {
+	// eslint-disable-next-line @wordpress/no-unused-vars-before-return
+	const { Edit } = getDefinitionForType( type );
+
+	const box = getBox( { x, y, width, height, rotationAngle, isFullbleed } );
+	// eslint-disable-next-line @wordpress/no-unused-vars-before-return
+	const props = { ...box, ...rest, id };
+
+	return (
+		<Wrapper
+			{ ...box }
+			onMouseDown={ ( evt ) => evt.stopPropagation() }
+		>
+			<Edit { ...props } />
+		</Wrapper>
+	);
+}
+
+ElementEdit.propTypes = {
+	element: PropTypes.object.isRequired,
+};
+
+export default ElementEdit;
--- a/assets/src/edit-story/components/canvas/selectionCanvas.js
+++ b/assets/src/edit-story/components/canvas/selectionCanvas.js
@@ -24,23 +24,27 @@ const LassoMode = {
 };
 
 const Container = withOverlay( styled.div`
-  width: 100%;
-  height: 100%;
-  user-select: none;
+	position: absolute;
+	top: 0;
+	left: 0;
+	right: 0;
+	bottom: 0;
+	width: 100%;
+	height: 100%;
+	user-select: none;
 ` );
 
 const Lasso = styled.div`
-  display: none;
-  position: absolute;
-  border: 1px dotted ${ ( { theme } ) => theme.colors.selection };
-  z-index: 1;
+	display: none;
+	position: absolute;
+	border: 1px dotted ${ ( { theme } ) => theme.colors.selection };
+	z-index: 1;
 `;
 
 function SelectionCanvas( { children } ) {
 	const {
 		actions: { clearSelection },
 	} = useStory();
-
 	const {
 		state: { pageContainer },
 		actions: { clearEditing, selectIntersection },
--- a/assets/src/edit-story/app/layout/index.js
+++ b/assets/src/edit-story/app/layout/index.js
@@ -37,6 +37,7 @@ const Editor = styled.div`
 const Area = styled.div`
   grid-area: ${ ( { area } ) => area };
   position: relative;
+  overflow: hidden;
   z-index: ${ ( { area } ) => area === 'canv' ? 1 : 2 };
 `;
 
--- a/assets/src/edit-story/app/layout/index.js
+++ b/assets/src/edit-story/app/layout/index.js
@@ -34,10 +34,10 @@ const Editor = styled.div`
 		/ minmax(${ LIBRARY_MIN_WIDTH }px, ${ LIBRARY_MAX_WIDTH }px) 1fr minmax(${ INSPECTOR_MIN_WIDTH }px, ${ INSPECTOR_MAX_WIDTH }px);
 `;
 
+// @todo: set `overflow: hidden;` once page size is responsive.
 const Area = styled.div`
   grid-area: ${ ( { area } ) => area };
   position: relative;
-  overflow: hidden;
   z-index: ${ ( { area } ) => area === 'canv' ? 1 : 2 };
 `;
 
--- a/assets/src/edit-story/app/layout/index.js
+++ b/assets/src/edit-story/app/layout/index.js
@@ -37,6 +37,7 @@ const Editor = styled.div`
 const Area = styled.div`
   grid-area: ${ ( { area } ) => area };
   position: relative;
+  z-index: ${ ( { area } ) => area === 'canv' ? 1 : 2 };
 `;
 
 function Layout() {
--- a/assets/src/edit-story/components/canvas/canvasProvider.js
+++ b/assets/src/edit-story/components/canvas/canvasProvider.js
@@ -6,7 +6,7 @@ import PropTypes from 'prop-types';
 /**
  * WordPress dependencies
  */
-import { useCallback, useEffect, useState } from '@wordpress/element';
+import { useCallback, useEffect, useRef, useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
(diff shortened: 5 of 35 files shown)


assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -1,62 +1,223 @@
...
+	${ DEBUG_ANGLE && ( `
+		background-color: darkgray;
+		backface-visibility: hidden;
+		transform-style: preserve-3d;
+		transform-origin: 50% 50%;
+		transform: perspective(2500px) scale(0.4) rotateY(${ DEBUG_ANGLE }deg);
+	` ) }
+`;
+
+const Layer = css`
```
pbakaus: "Position" instead of Layer, maybe?

pbakaus: also would prefer to make variable names more explicit plus lower case as they're not components (styled components docs also use lowercase), e.g. `positionCss` or `positionCSS`.

assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -1,62 +1,223 @@
...
-    ".    .      .          . .        ." 1fr
-    ".    prev   page       . next     ." ${ PAGE_HEIGHT }px
-    ".    .      menu       . .        ." 48px
-    ".    .      carrousel  . .        ." 1fr
-    / 1fr ${ LEFT_NAV_WIDTH }px ${ PAGE_WIDTH }px ${ PAGE_NAV_PADDING }px 1fr ${ CENTRAL_RIGHT_PADDING }px;
+    ".    .      . .          . .        ." 1fr
+    ".    prev   . page       . next     ." ${ PAGE_HEIGHT }px
+    ".    .      . menu       . .        ." 48px
+    ".    .      . .          . .        ." 1fr
+    ".    carrousel      carrousel carrousel  carrousel carrousel        ." 60px
```
pbakaus: feel free to ignore this as this isn't really part of this PR, but my OCD demands that we fix `carousel` misspelling everywhere ;)

assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -1,62 +1,223 @@
...
+    / 1fr ${ LEFT_NAV_WIDTH }px ${ PAGE_NAV_BUTTON_WIDTH }px ${ PAGE_WIDTH }px ${ PAGE_NAV_BUTTON_WIDTH }px ${ LEFT_NAV_WIDTH }px 1fr;
+`;
+
+const Canvas = css`
```
pbakaus: only used once, consider inlining below?

assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -1,62 +1,223 @@
...
+	` ) }
 `;
 
-const Area = styled.div`
+const NavArea = styled.div`
+	background-color: ${ ( { theme } ) => theme.colors.bg.v1 };
 	grid-area: ${ ( { area } ) => area };
 	height: 100%;
 	width: 100%;
+	pointer-events: initial;
```
pbakaus: why is this needed? or just for readability?

assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -1,62 +1,223 @@
...
+	` ) }
+`;
+
+// QQQQ: only show graypane for few element types.
+const EditLayer = styled.div`
+	${ Layer }
+	${ Layout }
+	pointer-events: none;
+	background-color: rgba(0, 0, 0, 0.5);
+	z-index: 1;
```
pbakaus: consider moving z-index out to constants, unless we can get away with just this one

assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -1,62 +1,223 @@
...
+		background-color: rgba(255, 255, 0, 0.2);
+	` ) }
+`;
+
+// QQQQ: only show graypane for few element types.
+const EditLayer = styled.div`
+	${ Layer }
+	${ Layout }
+	pointer-events: none;
+	background-color: rgba(0, 0, 0, 0.5);
```
pbakaus: put into theme

assets/src/edit-story/components/canvas/elementEdit.js
```diff
@@ -0,0 +1,57 @@
...
+import { ElementWithPosition, ElementWithSize, ElementWithRotation, getBox } from '../../elements/shared';
+
+// Background color is used to make the edited element more prominent and
+// easier to see.
+const Wrapper = styled.div`
+  ${ ElementWithPosition }
+  ${ ElementWithSize }
+  ${ ElementWithRotation }
+  pointer-events: initial;
+	background-color: rgba(255, 255, 255, 0.5);
```
pbakaus: move to theme

assets/src/edit-story/components/canvas/selectionCanvas.js
```diff
@@ -24,6 +24,11 @@ const LassoMode = {
 };
 
 const Container = withOverlay( styled.div`
+	position: absolute;
```
pbakaus: nit: fix indentation

pbakaus:
`inset:0`?? I just thought 'wait a sec, why inset, that's pretty obscure and new', then double checked and yep, [it's only supported in Firefox](https://caniuse.com/#feat=mdn-css_properties_inset). 🤣

It's also nowhere in the code...mysterious!

assets/src/edit-story/components/canvas/canvasLayout.js
swissspidy: Also covered by #4106 FWIW

assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -1,62 +1,223 @@
...
 `;
 
 function CanvasLayout() {
+	const {
+		state: { currentPage },
+	} = useStory();
+	const {
+		state: { editingElement: editingElementId },
+	} = useCanvas();
+	// eslint-disable-next-line @wordpress/no-unused-vars-before-return
```
swissspidy: This `eslint-disable-next-line` seems to be unneeded. Also: indentation.

assets/src/edit-story/components/canvas/canvasLayout.js
```diff
@@ -1,62 +1,223 @@
...
+				<DisplayLayer>
+					<DisplayCanvas>
+						<PageDisplay />
+					</DisplayCanvas>
+				</DisplayLayer>
+				<NavLayer onMouseDown={ ( evt ) => evt.stopPropagation() }>
+					<NavArea area="menu">
+						<PageMenu />
+					</NavArea>
+					{ /* QQQ: buttons are too big. especially given the pointerEvents
```
swissspidy: #4106 is gonna improve the layout there.

In the meantime, this could be a quick fix:


```patch
 	display: grid;
 	grid:
-    ".    .      . .          . .        ." 1fr
-    ".    prev   . page       . next     ." ${ PAGE_HEIGHT }px
-    ".    .      . menu       . .        ." 48px
-    ".    .      . .          . .        ." 1fr
-    ".    carrousel      carrousel carrousel  carrousel carrousel        ." 60px
-    / 1fr ${ LEFT_NAV_WIDTH }px ${ PAGE_NAV_BUTTON_WIDTH }px ${ PAGE_WIDTH }px ${ PAGE_NAV_BUTTON_WIDTH }px ${ LEFT_NAV_WIDTH }px 1fr;
+    ".         .         .         .         .         " 1fr
+    ".         prev      page      next      .         " ${ PAGE_HEIGHT }px
+    ".         .         menu      .         .         " 48px
+    ".         .         .         .         .         " 1fr
+    "carousel  carousel  carousel  carousel  carousel  " 60px
+    / 1fr ${ LEFT_NAV_WIDTH }px ${ PAGE_WIDTH }px ${ LEFT_NAV_WIDTH }px 1fr;
```

miina:
Great work separating the layers. A few notes from testing, just in case, although I know that it's still very WIP and things are not supposed to work yet:
- Should the header (the Story Title) have lower z-index than the PageFrames (or Selection) Layer? Currently, rotation handle and multi-selection outline get hidden under it.
- As Pascal mentioned as well then multi-selection actions (rotation & dragging) aren't working, doesn't seem to be handling the drag events, unselects instead:
multiselection
- Lasso selection isn't always working, selecting just one element, e.g.
lasso
- Sometimes clicking on workspace selects an element unexpectedly:
clicking

Perhaps these are all expected issues but bringing these up in case they hint to some underlying issue.

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: Still prefer "Layer" or "CanvasLayer" since they do not reflect positions of any kind - only `z-index` layering. This will soon stop being "css" and become an actual component - still deciding what would look better here.

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: Will sync it.

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: Layers are normally `pointer-events: none` - they are only used for `z-index` in terms of display, clipping, and event propagation - but otherwise do not directly respond to events. The children of layers actually handle events, so the `pointer-events` is restored.

dvoytenko: To add: I've been trying to keep this `pointer-events` off -> on logic in this one module. Not yet sure it'd work like this for everything.

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: Also still evaluating. IMHO if all `z-index` values can be localized here (e.g. not shared across multiple modules), this would be better like this. For now I hope we can fully localize `z-index`.

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: Do we put shading into themes? There's no actual color here, just `0.5` value that matters. This is similar to `opacity: 0.5`, but applied directly to a color. I was under an impression that those cases do not belong in a theme.

/cc @barklund @miina

assets/src/edit-story/components/canvas/elementEdit.js
dvoytenko: Ditto, theme vs shading.

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: Removed.

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: I'll just sync.

dvoytenko:
@miina 

> Should the header (the Story Title) have lower z-index than the PageFrames (or Selection) Layer? Currently, rotation handle and multi-selection outline get hidden under it.

Correct. The header will move to "workspace" area very soon. Either in this PR or the next. Just avoiding two bits of work on header that are being done elsewhere. So depending on their merge state...

> As Pascal mentioned as well then multi-selection actions (rotation & dragging) aren't working, doesn't seem to be handling the drag events, unselects instead
> Lasso selection isn't always working, selecting just one element

Yep. Still working on this. Fairly small fixes, so should be available soon.

> Sometimes clicking on workspace selects an element unexpectedly

That's very unexpected. Looking...

assets/src/edit-story/components/canvas/canvasLayout.js
pbakaus: sure, if it becomes a component then works for me.

assets/src/edit-story/components/canvas/canvasLayout.js
pbakaus: oh I missed that these are children. All good then.

assets/src/edit-story/components/canvas/canvasLayout.js
pbakaus: if the goal is to isolate them to this module, then that works. Would still move them into constants to the top of the file then though.

assets/src/edit-story/components/canvas/canvasLayout.js
pbakaus: I guess "color" is a matter of definition :) I'd count black as one for our purposes?

Regardless, I think it might be worthwhile to put even `opacity` values into the theme, as they're stylistic choices (similar to fonts, font style, ...).

assets/src/edit-story/components/canvas/elementEdit.js
pbakaus: Ditto from above – it's used for shading, but effectively you are shading by applying white at 50% opacity on top of something else. If we don't use these in other modules though, then I'm fine with defining them as constants locally – that way we can still easily change the values if need to be.

assets/src/edit-story/components/canvas/canvasLayout.js
dvoytenko: Ok. Will do then.

assets/src/edit-story/components/canvas/elementEdit.js
dvoytenko: Ok. Agree.

assets/src/edit-story/components/canvas/selectionCanvas.js
dvoytenko: hmm. the `--fix` does some rather strange things...

dvoytenko:
@pbakaus @miina @swissspidy PTAL. The outstanding bugs should be fixed now. I'm still going to do a few renamings and syncing with the base, but structurally this work should be complete.

assets/src/edit-story/components/canvas/elementEdit.js
dvoytenko: To clarify though: one is grayout and the other one is whiteout. There's really not much in-between. The only thing that could be a bit different here is the transparency. I'll start with these two as named theme colors and we'll see from there.

pbakaus: yeah, you're right. I guess the color isn't something we would want to change here, realistically. The opacity value more so.

assets/src/edit-story/components/canvas/selectionCanvas.js
swissspidy: Might be improved a bit with the new repo, but linting and autofixing styled components is tricky

miina:
@dvoytenko I'm currently seeing this, is this still expected at this point:
Screenshot 2020-01-22 at 13 40 27

Note that also the Page isn't centered, in addition to the other oddities in the bottom of the Page.

miina:
^ Note that the situation looks better when zoomed out.

dvoytenko:
> ^ Note that the situation looks better when zoomed out.

@miina, yes, this is still outstanding. Two issues (roughly covered by #3920):

1. We need to support a responsive size of the canvas. That's a biggish change since we'll need to do a two-way conversion between data and display pixels. I think I'll get to it next.
2. The side panels need to have their own scrollers. We keep reintroducing the bugs where we forget the scrollers. But I think there's a simple way to make it a bit harder to forget. So I'll send a follow up PR for that.

dvoytenko:
@miina @pbakaus @swissspidy I finished the fixes and cleanup and sync'd with the HEAD. The conflicts here are pretty hard, so I'd like to merge it the sooner the better. I'm ok to do some emergency fixes on my end if I miss anything. LMK.

miina:
@dvoytenko Currently it makes the editor a bit unusable on my screen:
The whole meta bar with actions is hidden behind gray area:
Screenshot 2020-01-23 at 11 46 54

The navigation shows up when zooming out quite a lot, which, however, makes the items too tiny for normally being able to work on the editor.

How much more time do you think it would take to fix the responsiveness issue?
I'd rather hold off with other merges (to avoid further conflicts) until this gets fixed and merged instead of merging as-is.

However, if all the others think it's OK to merge like this then I can just continue working on the pre-merge branch until it gets fixed, perhaps for wider screens it works fine for now.

Thoughts?

swissspidy:
Hmm I don't see these drastic issues. Just the carousel is a bit off:

Screenshot 2020-01-23 at 15 43 44
Screenshot 2020-01-23 at 15 43 22
Screenshot 2020-01-23 at 15 43 16

That seems easily fixable though with tweaking the CSS grid again.

Apart from that there are just a bunch of React warnings in the console, which might be pre-existing though.

dvoytenko:
@swissspidy @miina, let me see if there's an easy fix for the layout. But at this point we are a bit in the corner. The only way to make it work is by resizing the "page" card.

@miina What's your screen resolution?

dvoytenko:
@miina I removed overflow from the editor area until we can make page size responsive. PTAL. The gray area is still there, but at least the controls are not clipped. I will work on making the page size itself responsive next so that the editor fits within the predefined sizes.

swissspidy:
Oh I see now, the carousel is so small because you deliberately made them so in `canvas/carousel/index.js`, citing lack of responsiveness as the reason. Makes sense. Then I don't really have objections on merging.

Still cannot reproduce the thing with the grey area.

dvoytenko:
@swissspidy the gray area comes up if you continue to making the window smaller. There's currently a conflict between a fill-space grid and the unresponsive page size. I.e. sum(height) > 100%.

miina:
@dvoytenko Looks better now, still seeing the gray area, but as you said, at least the controls are there now. Ok to merge from my side as well then. 👍

swissspidy:
💥

PR #4135, opened 2020-01-19
Add source information for script/style dependencies and inline scripts/styles

### LINKED ISSUE #4133: Scripts and styles enqueued before wp_enqueue_scripts fail to get sources identified
## Bug Description

Scripts and stylesheets enqueued at `init` do not get sourcing information. The same goes for assets registered at `wp_default_styles` and `wp_scripts`.

See [demo plugin](https://gist.github.com/westonruter/7ce1e6270aa5b1bd9e66a3f0adf93b07) which demonstrates the issue.

(Additionally, inline styles added via `wp_add_inline_style()` do not get their source attributed.)

The `wp-includes` source is erroneous for 2 scripts and 3 stylesheets shown on the page.

Related: #4134

## Expected Behaviour

The source column should indicate the theme/plugin that is actually responsible for the style or script being added to the page.

The case of the inline stylesheet is a tricky one, since it is one stylesheet with multiple sources. At the very least it should have the same source information as the stylesheet it is being added inline for. Best case is for it to include all themes/plugins that added inline styles to it.

## Steps to reproduce

0. Enable Standard or Transitional mode in the AMP plugin.
1. Activate the [above plugin](https://gist.github.com/westonruter/7ce1e6270aa5b1bd9e66a3f0adf93b07)
2. Validate a page.
3. Notice the `wp-includes` mentioned unexpectedly.

## Screenshots

When active (and #4026 checked out) the validated URL screen looks like the following, with the red entries erroneously saying they are from `wp-includes` when they should actually be from “Add Assets before wp_enqueue_scripts”.

72671921-d66e7780-3a06-11ea-8135-d2f12e3dc6b5

## Additional context

 - WordPress version:
 - Plugin version:
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* 

## Implementation brief

* 

* This may be resolved simply by adding the validation hooks earlier. Origination is also a solution, perhaps long-term.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

### LINKED ISSUE #4134: Inline scripts/styles should have sources identified and core script/style dependencies should get enqueueing theme/plugin as source
## Feature description

Related: #4133

If a theme enqueues a `navigation.js` that has jQuery as a dependency, currently this appears in the validated URL screen as “wp-includes” as the source. This is not very helpful. It would be helpful that if a theme or plugin enqueues a script or a style, that the original theme/plugin that enqueued it would be attributed as a source in some way. As an example, consider this [demo plugin](https://gist.github.com/westonruter/b8f559338bc9ecf1c9f9967c48676219).

All of the scripts here should get “Enqueue assets with core dependencies” as the source, at least indirectly:

image

And for the stylesheets, the same goes for the circled stylesheets:

image

I believe that [Origination](https://github.com/GoogleChromeLabs/wp-origination) already provides support here for this, so this should perhaps be deferred to the incorporation of that library.

Also: For each script/stylesheet, the asset which caused the dependency to be printed should be listed.

Additionally, inline styles added via `wp_add_inline_style()` do not get their source attributed.

### Inline Scripts

Scripts added via `wp_localize_script()` and `wp_add_inline_script()` should all get properly identified. For this, see [another example plugin](https://gist.github.com/westonruter/2b1b64249befa3b39adccc9f139f9aba).

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

westonruter:
## Summary

Fixes #4134
Fixes #4133

* Scripts and styles that are printed due to being dependencies of enqueued scripts and styles will now show the plugin/theme responsible for enqueueing as the source. This eliminates most of the unhelpful `wp-includes` entries from appearing as the validation error source. The handle of the dependent script/style is also listed among the sources.
* Inline scripts now indicate among their sources _all_ of the plugins and theme(s) that caused them to be added, as opposed to just one. Also indicated is whether the script was added via `wp_localize_script()` or `wp_add_inline_script()`. The source stack also includes the specific inline script that was added.
* Inline styles added via `wp_add_inline_style()` are now attributed to theme/plugin sources for the first time. As with scripts, all plugins and theme(s) that contribute to a given inline script will now be listed among the sources.
* Scripts and styles that are enqueued before `wp_enqueue_scripts` will now be properly attributed. This includes styles registered at `wp_default_styles` and scripts at `wp_default_scripts`. 

## Testing

0. Activate the twentytwenty theme.
1. Copy each file from [this gist](https://gist.github.com/westonruter/2b1b64249befa3b39adccc9f139f9aba) into the `wp-content/plugins` directory. These are separate plugins so they must be added as separate files in the `plugins` directory. Then do `wp plugin activate add-inline-scripts add-more-inline-scripts`.
2. Clone [another gist](https://gist.github.com/westonruter/b8f559338bc9ecf1c9f9967c48676219) into the plugins directory with the name `enqueue-assets-with-core-dependencies` and activate via `wp plugin activate enqueue-assets-with-core-dependencies`.
3. Clone [yet another gist](https://gist.github.com/westonruter/7ce1e6270aa5b1bd9e66a3f0adf93b07) into the `wp-content/plugins` with the directory name of `add-assets-before-enqueue-scripts`. Then do `wp plugin activate add-assets-before-enqueue-scripts`.

## Screenshots

Compare before and after:

Before | After
-------|------
Screen Shot 2020-01-23 at 22 32 36 | Screen Shot 2020-01-23 at 22 33 19

The stylesheets have the same sourcing:

Before | After
-------|------
image | image

The source stack now has information about how a given inline script was added, either via `wp_add_inline_script()` or `wp_localize_script()`. For example:

Before | After
-------|------
image | image

If I activate a theme (like Hestia) that also has scripts that depend on jQuery, then I see both the plugin and the theme as sources:

image

# Todo

- [x] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests). See https://github.com/ampproject/amp-wp/issues/4175.
  - [x] Add tests for `AMP_Validation_Callback_Wrapper`.
  - [x] Add tests for `AMP_Validation_Manager::locate_sources()`
- [x] Add stylesheet URL and enqueued handle as top level `dl` items? At least linkify the URL in the `<link>` original markup.


westonruter:
The `wp-block-library-theme` stylesheet currently has `wp-includes` as the source:

image

This is fine because it comes from core, but technically it could have the current theme attributed as a source because `wp_common_block_scripts_and_styles()` conditionally enqueues it based on whether or not the theme has `wp-block-styles` theme support. It's not important, however. And it could be argued that having `wp-includes` is just fine.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -2885,13 +2920,13 @@ public static function get_error_title_from_code( $validation_error ) {
...
-					$title = esc_html__( 'Invalid script', 'amp' );
-
 					if ( isset( $validation_error['node_attributes']['src'] ) ) {
-						$title .= sprintf(
+						$title = esc_html__( 'Invalid script', 'amp' ) . sprintf(
 							': <code>%s</code>',
 							esc_html( basename( wp_parse_url( $validation_error['node_attributes']['src'], PHP_URL_PATH ) ) )
 						);
+					} else {
+						$title = esc_html__( 'Invalid inline script', 'amp' );
```
westonruter: Would be nice to add the handle that the inline script is for.

westonruter:
The `stylesheet-enqueued-at-init` style enqueued and registered at `init` is not getting sourced for some reason:

image

amp.php
```diff
@@ -307,6 +307,9 @@ function amp_deactivate() {
 	flush_rewrite_rules( false );
 }
 
+// The plugins_loaded action is the earliest we can run this since that is when pluggable.php has been required and wp_hash() is available.
```
schlessera: Why not replace `wp_hash` with something else, then? It doesn't need to be cryptographically secure, so any type of hash will work, really.

schlessera: I think it would even make sense to include a hash function in the `amp/common` library, so that we have consistent hashing across all libraries.

westonruter: Good question. It's not just purely a hash that we need. It's hashing using one of the WordPress secrets, which entails using `wp_salt()` which is in turn another pluggable function. Then the input for hashing is `wp_nonce_tick()` which is yet another pluggable function. So these three functions all require this to be run at `plugins_loaded` and I believe we need to depend on them specifically.

https://github.com/ampproject/amp-wp/blob/1c3c8539aeb59b564da3bc6db16584dfc25dd88f/includes/validation/class-amp-validation-manager.php#L1773-L1775

westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/4110365/amp.zip) (1.5.0-alpha-20200124T194146Z-186d71920)

westonruter:
Todo: 

* Shorten columns by dropping “Size” from “Original Size” and “Final Size”. Use “Percent” instead of “Percentage“.
* Add hyperlink to URL in `<link>` original markup.

westonruter:
> * Shorten columns by dropping “Size” from “Original Size” and “Final Size”. Use “Percent” instead of “Percentage“.

Done in e54bb6118.

Before:
Screen Shot 2020-01-24 at 17 13 49

After:
Screen Shot 2020-01-24 at 17 14 18

pierlon:
With the `add-inline-scripts` & `add-more-inline-scripts` plugiins activated, there seems to be an issue with how scripts added via `wp_localize_script` are shown:

image

The `Hello_Underscores` variable was added by `add-inline-scripts`, not `add-more-inline-scripts`. Scripts added via `wp_add_inline_script` don't seem to be affected.

westonruter:
@pierlon I think that is expected because `\WP_Scripts::localize()` concatenates the strings. So when the first time `wp_localize_script()` is called, only the first text is present. I think this can be improved by splitting the text by newlines. I'll try it.

westonruter:
@pierlon I believe 6059817 resolves this.

includes/validation/class-amp-validation-manager.php
```diff
@@ -1314,10 +1461,17 @@ public static function wrap_hook_callbacks( $hook ) {
 					)
 				);
 
-				$callback['function'] = function() use ( &$callback, $wrapped_callback, $original_function ) {
-					$callback['function'] = $original_function; // Restore original.
-					return call_user_func_array( $wrapped_callback, func_get_args() );
-				};
+				if ( 1 === $passed_by_ref ) {
+					$callback['function'] = function( &$first, ...$other_args ) use ( &$callback, $wrapped_callback, $original_function ) {
```
pierlon:
```suggestion
					$callback['function'] = static function( &$first, ...$other_args ) use ( &$callback, $wrapped_callback, $original_function ) {
```

includes/validation/class-amp-validation-manager.php
```diff
@@ -1314,10 +1461,17 @@ public static function wrap_hook_callbacks( $hook ) {
...
+						$callback['function'] = $original_function; // Restore original.
+						return $wrapped_callback->invoke_with_first_ref_arg( $first, ...$other_args );
+					};
+				} else {
+					$callback['function'] = function( ...$args ) use ( &$callback, $wrapped_callback, $original_function ) {
```
pierlon:
```suggestion
					$callback['function'] = static function( ...$args ) use ( &$callback, $wrapped_callback, $original_function ) {
```

pierlon:
The patch for the `mustache/mustache` library (requirement of `wp-cli/wp-cli`) is [failing to be patched](https://travis-ci.org/ampproject/amp-wp/jobs/651793279#L569-L571) since being upgraded to v2.13.0 via e35a144d93a4435b16d4f43f229a86b809b0bde6. Downgrading it to v2.12.0 should resolve the issue.

westonruter:
> The patch for the `mustache/mustache` library (requirement of `wp-cli/wp-cli`) is [failing to be patched](https://travis-ci.org/ampproject/amp-wp/jobs/651793279#L569-L571) since being upgraded to v2.13.0 via [e35a144](https://github.com/ampproject/amp-wp/commit/e35a144d93a4435b16d4f43f229a86b809b0bde6). Downgrading it to v2.12.0 should resolve the issue.

How do you recommend doing this? We don't directly have `mustache/mustache` as a dependency. Would we add it as a dev dependency and pin it at `2.12.0`? We'd have to then tell Renovate to not try to update it, or else just reject updates.

pierlon:
The patch being applied by `wp-cli/wp-cli` is to make `mustache/mustache` PHP 7.4 compatible, which has been [fixed in v2.13.0](https://github.com/bobthecow/mustache.php/releases/tag/v2.13.0). The patch can be ignored by instructing `cweagans/composer-patches` to [ignore it](https://github.com/cweagans/composer-patches#ignoring-patches). Something like:

```diff
Index: composer.json
===================================================================
--- composer.json	(revision 6059817247734c2e7c1f15c39450e152100b0e6a)
+++ composer.json	(date 1581998688688)
@@ -44,6 +44,13 @@
         "Fix parsing CSS selectors which contain commas <https://github.com/sabberworm/PHP-CSS-Parser/pull/138>": "https://github.com/sabberworm/PHP-CSS-Parser/commit/fa139f65c5b098ae652c970b25e6eb03fc495eb4.diff",
         "Validate name-start code points for identifier <https://github.com/sabberworm/PHP-CSS-Parser/pull/185>": "patches/php-css-parser-pull-185.patch"
       }
+    },
+    "patches-ignore": {
+      "wp-cli/wp-cli": {
+        "mustache/mustache": {
+          "Patch no longer relevant as of mustache/mustache v2.13.0": "https://patch-diff.githubusercontent.com/raw/bobthecow/mustache.php/pull/349.patch"
+        }
+      }
     }
   },
   "autoload": {
```

PR #4137, opened 2020-01-19
Raise validation errors for properties in attributes

### LINKED ISSUE #4070: Add validation of individual properties in meta content attributes
## Feature description

Split out from #3758 via the conversation with @schlessera in https://github.com/ampproject/amp-wp/pull/3758#discussion_r348074703:

> More generally, would it be possible to refactor the tag & meta sanitizer to strip the offending properties only, and then check if the entire thing is still valid or not? This way, we wouldn't have to duplicate this logic all over the place...

This is in relation to this code:

https://github.com/ampproject/amp-wp/blob/a7e0497b72b624f6e3032f857f6dcbbf85039f6f/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L1155-L1159

---------------

## Acceptance criteria

* Adding invalid properties to a `meta[viewport]` should not result in the entire attribute being removed, if there are other valid properties.
* A separate validation error should be raised for each invalid property in a `meta[content]` attribute.

## Implementation brief

Instead of raising one single `DISALLOWED_PROPERTY_IN_ATTR_VALUE` for the entire `meta[content]` attribute and then causing the entire `content` attribute to be removed, we instead need to:

1. Raise a `DISALLOWED_PROPERTY_IN_ATTR_VALUE` validation error for each individual property in the `content` attribute.
2. Remove the invalid property from the `content` attribute if the validation error if `\AMP_Base_Sanitizer::should_sanitize_validation_error()` returns `true`.

What this means is that the `\AMP_Tag_And_Attribute_Sanitizer::sanitize_disallowed_attribute_values_in_node()` method will need to be augmented to not simply return a list of `[ $attr_node, $error_code ]` tuples. I the case of invalid properties, it will also need to return `[ $attr_node, self::DISALLOWED_PROPERTY_IN_ATTR_VALUE, $property_name, $property_value ]`. 

This resulting list of ` $disallowed_attributes` will then need to be iterated over in the `foreach ( $disallowed_attributes as $disallowed_attribute ) {…}` here:

https://github.com/ampproject/amp-wp/blob/a7e0497b72b624f6e3032f857f6dcbbf85039f6f/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L667-L695

Where instead of calling the `\AMP_Base_Sanitizer::remove_invalid_attribute()`, there would need to be a direct call to `AMP_Base_Sanitizer::should_sanitize_validation_error()` and if it returns true, update the value of the attribute to omit that invalid property.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

This PR allows for separate validation errors to be raised when an attribute has invalid properties, missing mandatory properties and missing required property values. Also, suitable titles for their error codes have been added, along with the `AMP_Tag_And_Attribute_Sanitizer::ATTR_REQUIRED_BUT_MISSING` error code.

Fixes #4070.

### Invalid properties

If there are invalid properties within an attribute, each occurrence will be sanitized. For example, given the meta tag:

```html
<meta name="viewport" content="foo=, widh=device-width, initial-scale=1, user-scalabe=no,">
```

The following validation errors will be raised:

image

### Missing mandatory properties

If a mandatory property within an attribute is missing, the attribute will be removed. For example, given the meta tag:

```html
<meta name="viewport" content="initial-scale=1, user-scalable=no,">
```

A validation error will be raised for the missing `width` property, as well as for the `content` attribute being removed:

image

_Note: This scenario can only be replicated by disabling `AMP_Meta_Sanitizer` so that it will not set the correct `width` property value. Currently, only the `width` property [requires for a specific value to be set](https://github.com/ampproject/amphtml/blob/e7a8e7f687e5e63a2ecb63566828ec25296a7f96/validator/validator-main.protoascii#L437)._

### Missing required property values

When a property does not have the required value,  the attribute will be removed. For example, given the meta tag:

```html
<meta http-equiv="X-UA-Compatible" content="IE=9,chrome=1">
```

A validation error will be raised as the `IE` property does not have the required value `edge`, and the `content` attribute will be removed:

image

## Checklist

- [ ] Add unit tests for validation error properties.
- [ ] Ensure tests are accounting for the various scenarios.
- [ ] Ensure validation error messages are fully populated for the various error scenarios.


includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -685,12 +687,44 @@ static function ( $validation_error ) {
...
-				if ( $this->remove_invalid_attribute( $node, $attr_node, $validation_error, $attr_spec ) ) {
-					$removed_attributes[] = $attr_node;
+				if ( self::DISALLOWED_PROPERTY_IN_ATTR_VALUE === $error_code ) {
+					$properties         = $error_data['original_properties'];
+					$invalid_properties = $error_data['invalid_properties'];
+
+					foreach ( $invalid_properties as $invalid_property ) {
+						$validation_error['attr_property_name'] = $invalid_property;
+						if ( $this->should_sanitize_validation_error( $validation_error, [ 'node' => $attr_node ] ) ) {
+							unset( $properties[ $invalid_property ] );
```
westonruter: While this complicates the logic here, removal of the invalid properties needs to be deferred later outside of this method. The reason is that if the removal is done too early, then if you opt to remove or keep the invalid property then this will impact the validation error objects for subsequent checks for invalid attributes. This can result in the latter validation errors needing to be re-removed after changeling the status of former validation errors.

You can see the logic later for this in relation to removal of invalid attributes. In particular, you can see it gathers up all the attributes and then passes them into the sanitize method to check if it should be removed. This ensures there is consistency across the validation errors as one is removed.

Alternatively perhaps the array of original attribute values could be obtained _before_ even this method is called, and then it wouldn't matter.

westonruter: @pierlon I'm sorry, I believe I am mistaken here. I first thought this diff was part of the `sanitize_disallowed_attribute_values_in_node` method but I see it is not. It looks like you are doing the right thing. I can see that `$validation_error['element_attributes']` has already been populated before this, so my concern was incorrect.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1920,12 +1976,21 @@ private function check_attr_spec_rule_value_properties( DOMElement $node, $attr_
...
 				}
 				if ( isset( $required_value ) && $prop_value !== $required_value ) {
-					return AMP_Rule_Spec::FAIL;
+					return [
+						AMP_Rule_Spec::FAIL,
+						[
+							self::MISSING_REQUIRED_PROPERTY_VALUE,
+							[
+								'property' => $prop_name,
+								'value'    => $required_value,
```
westonruter: What about `property_name` and `property_value`? Or just `name` and `value`?

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -685,12 +687,44 @@ static function ( $validation_error ) {
...
+						}
+					);
+
+					$node->setAttribute( $attr_node->nodeName, $valid_properties );
+				} else {
+					if ( self::MISSING_MANDATORY_PROPERTY === $error_code ) {
+						$validation_error['attr_property_name'] = $error_data['property'];
+					} elseif ( self::MISSING_REQUIRED_PROPERTY_VALUE === $error_code ) {
+						$validation_error['attr_property_name']  = $error_data['property'];
+						$validation_error['attr_property_value'] = $error_data['value'];
```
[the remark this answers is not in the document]
pierlon: I chose to prefix `attr_` because `property_name` and `property_value` were already in use in the style sanitizer. I suppose renaming those instances to `css_property_*` would resolve that to resolve any ambiguity.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
pierlon: `name` and `value` it is!

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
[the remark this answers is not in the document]
pierlon: Renamed `property_*` to `css_property_*` and `attr_property_*` to `property_*` in https://github.com/ampproject/amp-wp/pull/4137/commits/73d04ff259a2ee23f13cbc749a3a49829255fe59.

pierlon:
While reviewing the `postmag` theme, the following validation error can be seen in the AMP validator:

> The property 'width' in attribute 'content' in tag 'meta name=viewport' is set to ' device-width', which is invalid.

Note the space prepended to the value `device-width`.

The plugin does not show this validation error because the [property value is trimmed](https://github.com/ampproject/amp-wp/blob/73d04ff259a2ee23f13cbc749a3a49829255fe59/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L1942) before [comparison to the required value](https://github.com/ampproject/amp-wp/blob/73d04ff259a2ee23f13cbc749a3a49829255fe59/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L1987). This can be solved in two ways:

- Compare the untrimmed value, which would cause the `MISSING_REQUIRED_PROPERTY_VALUE`  validation error to occur

- Save the trimmed value to the attribute to resolve the error

Thoughts @westonruter?

westonruter:
The meta viewport in question [in postmag](https://themes.trac.wordpress.org/browser/postmag/1.1.7/header.php#L15) is:

```html
<meta name = "viewport" content = "width = device-width, initial-scale = 1.0" >
```

I can see this is indeed a problem for the AMP Validator:

image

Nevertheless, both Chrome and Firefox seem to read the `viewport` properties just fine with the extra whitespace. So this appears to perhaps be a bug with the AMP Validator.

Therefore, perhaps the right solution now is to “save the trimmed value to the attribute to resolve the error”. I'll open an issue on the AMP project to see if this is a bug. If so, the saved-trimming could then be removed.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -2964,14 +3006,32 @@ public static function get_source_key_label( $key, $validation_error ) {
...
-			case 'property_name':
+			case 'css_property_name':
 				return __( 'CSS property', 'amp' );
 			case 'text':
 				return __( 'Text content', 'amp' );
 			case 'type':
 				return __( 'Type', 'amp' );
 			case 'sources':
 				return __( 'Sources', 'amp' );
+			case 'property_name':
```
westonruter: Should this be renamed to `meta_property_name` then?

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -137,10 +137,9 @@ protected function ensure_viewport_is_present() {
 		 */
 		$viewport_tag      = $this->meta_tags[ self::TAG_VIEWPORT ][0];
 		$viewport_content  = $viewport_tag->getAttribute( 'content' );
-		$viewport_settings = array_map( 'trim', explode( ',', $viewport_content ) );
+		$viewport_settings = array_filter( array_map( 'trim', explode( ',', $viewport_content ) ) );
 		$width_found       = false;
 
-		// @todo The invalid/missing properties should raise validation errors. See <https://github.com/ampproject/amp-wp/pull/3758/files#r348074703>.
 		foreach ( $viewport_settings as $index => $viewport_setting ) {
 			list( $property, $value ) = array_map( 'trim', explode( '=', $viewport_setting ) );
```
westonruter: A missing change is to put the trimmed `$property` and `$value` into the `$viewport_settings`, correct? This would then address https://github.com/ampproject/amp-wp/pull/4137#issuecomment-578467681

includes/validation/class-amp-validation-error-taxonomy.php
westonruter: Alternatively, the `$validation_error['code']` can be used to differentiate which property type is in view here. And that could avoid the need to use `css_property_name`. Thoughts?

pierlon: Personally, I prefer renaming it to `meta_property_name` to be more descriptive.

includes/sanitizers/class-amp-meta-sanitizer.php
pierlon: I made the change to `AMP_Tag_And_Attribute_Sanitizer` in https://github.com/ampproject/amp-wp/pull/4137/commits/c4a5581701e21d9ed02d40ff25533f3e8d728a0c before seeing this. It definitely would be suited for `AMP_Meta_Sanitizer`.

pierlon: Done via https://github.com/ampproject/amp-wp/pull/4137/commits/b25781997dac38d6337bfd89a4752f285981de2d.

includes/validation/class-amp-validation-error-taxonomy.php
pierlon: Done via https://github.com/ampproject/amp-wp/pull/4137/commits/fa1299e503eeb853d8bb185d0b660e8eeec724bf.

westonruter:
Ready for re-review?

pierlon:
Yep!

pierlon:
### Invalid properties

> There is no validation error for width=BAD

In that case `AMP_Meta_Sanitizer` ensures the proper `width` property value is set before being validated.

### Missing mandatory properties

> I see _no_ validation errors reported. Why?

I forgot to mentioned I had disabled `AMP_Meta_Sanitizer` to produce that scenario. There isn't any property value in the spec that has a required mandatory value, with `width` being the exception.

westonruter:
There's one more thing that isn't expected, and that is if an invalid property value is provided for viewport `width`. For example, if I add:

```html
<meta name="viewport" content="width=600, initial-scale=1.0" >
```

Then this gets sanitized as:

```html
<meta name="viewport" content="width=device-width,initial-scale=1.0">
```

And no validation error is raised. This is is an issue with the `AMP_Meta_Sanitizer` where it is silently replacing the value with something that is allowed.

westonruter:
I'm not sure what the best resolution is for this. If the `AMP_Meta_Sanitizer` raises a validation error due to `width=600`, the behavior of the sanitizer should logically replace this with `width=device-width` when the status of the error is “removed”. However, if the user decides to make the status “kept” then the `AMP_Tag_And_Attribute_Sanitizer` will come along later and remove invalid `width=600` property _without_ replacing it with the required `width=device-width`. And the result would be a `meta` viewport that lacks the required property. It's a similar situation we have for the style sanitizer for when in Standard mode and the style sanitizer allows for the total CSS to go over 50KB. In that case we special-case the condition in the `AMP_Tag_And_Attribute_Sanitizer`:

https://github.com/ampproject/amp-wp/blob/19d54bdf0ac4c41a37927e11241b723cdb754de8/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L807-L817

Which isn't great, but I can't think of a better solution.

pierlon:
Would there be a case where a user would want to use a specific width? The page can't be AMP compatible unless it being set to `device-width`.

westonruter:
> Would there be a case where a user would want to use a specific width? The page can't be AMP compatible unless it being set to `device-width`.

In practice I've only ever seen `width` supplied with `device-width` as the value. However, it can have [different values](https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag). When someone explicitly marks a validation error as “kept” then this means they _don't_ want a valid AMP page. In Standard mode, such a page would be served without the `amp` attribute on the HTML element. Granted, this should be an exceedingly rare edge case, so I don't think we should do anything too involved to support this.

pierlon:
In https://github.com/ampproject/amp-wp/pull/4137/commits/cb7b96aa88cd961dabee3bfb3d2b1d9d12c69b69, the `width` property will no longer be overridden, and a validation error will be raised if the value is not `device-width`:

image

If the `width` property does not exist, one will be added with the value being set to `device-width`.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -135,28 +135,18 @@ protected function ensure_viewport_is_present() {
...
-				$width_found = true;
-				break;
-			}
-		}
+		$viewport_tag     = $this->meta_tags[ self::TAG_VIEWPORT ][0];
+		$viewport_content = $viewport_tag->getAttribute( 'content' );
+
+		// Remove any whitespace and random delimiters (commas).
+		$viewport_settings = preg_replace( '/^,|,(?=[^A-Za-z0-9 ])|,$|\s+/', '', $viewport_content );
+		$width_found       = false !== strpos( $viewport_settings, 'width=' );
```
westonruter: I think the previous method for finding whether the `$width` was found is more robust. What if someone has a property like `body-width=foo`? This will incorrectly match here.

pierlon: Yep agreed. Fixed in https://github.com/ampproject/amp-wp/pull/4137/commits/4cde75a7766e388968506bfc4724abd924cdf8d0.

pierlon:
In https://github.com/ampproject/amp-wp/pull/4137/commits/13a0f13a2ef743a57c08e133033bc28c58314b50, I've added the option to set the correct value for a mandatory property if the error is "removed".

schlessera:
I noticed that the source is not filled in my case. I edited the `header.php` file in the twentytwenty theme to change the viewport.

Image 2020-02-14 at 6 18 11 PM

westonruter:
> I noticed that the source is not filled in my case. I edited the `header.php` file in the twentytwenty theme to change the viewport.

@schlessera This would be expected. If markup is not generated by a hook callback then no source information would be available. In the past the source stack info would show the theme as a fallback, followed by "(?)". That wasn't very helpful either, and it wasn't reliable as I recall.

pierlon:
> If markup is not generated by a hook callback then no source information would be available.

That was my understanding as well, thanks for clarifying.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -126,37 +135,7 @@ protected function ensure_charset_is_present() {
...
-				$width_found = true;
-				break;
-			}
 		}
-
-		if ( ! $width_found ) {
-			array_unshift( $viewport_settings, 'width=device-width' );
-		}
-
-		$viewport_tag->setAttribute( 'content', implode( ',', $viewport_settings ) );
```
westonruter: @schlessera I found that this code was no longer necessary now that required property values for _all_ `meta` tags (not just `meta[content=viewport]`) in `\AMP_Tag_And_Attribute_Sanitizer::process_node()`.

westonruter:
Now given a `head` that contains:

```html
<meta name="viewport" content="initial-scale=1.0" >
<meta http-equiv="X-UA-Compatible" content="IE=6,chrome=1,netscape=4">
```

Three validation errors are raised:

wordpressdev lndo site_core-dev_src_wp-admin_post php_post=1951 action=edit

And with the errors being sanitized, the result is valid AMP:

```html
<meta name="viewport" content="initial-scale=1.0,width=device-width">
<meta http-equiv="X-UA-Compatible" content="chrome=1">
```

westonruter:
@schlessera @pierlon Please verify my additional changes and we'll merge.

pierlon:
Added a more specific label for the "required property value" validation error:

Before:
image

After:
image

Otherwise from that, the prior changes made by @westonruter work as described, so :heavy_check_mark: for me.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -2141,26 +2147,43 @@ public static function render_single_url_error_details( $validation_error, $term
...
 								'<code>%s: %s</code>',
 								esc_html( $value ),
-								esc_html( $validation_error['property_value'] )
+								esc_html( $validation_error['css_property_value'] )
 							);
 						} else {
 							printf( '<code>%s</code>', esc_html( $value ) );
 						}
 						?>
+					<?php elseif ( 'meta_property_name' === $key ) : ?>
```
schlessera: As we're using these array keys in multiple places, I would suggest using constants for them.

However, given that the long-term plan is to use objects for validation errors, this might just be a waste of effort at this point.

tests/php/test-class-amp-meta-sanitizer.php
```diff
@@ -41,6 +41,12 @@ public function get_data_for_sanitize() {
 				'<!DOCTYPE html><html><head><meta charset="utf-8"><meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1"></head><body></body></html>',
 			],
 
+			// Trim whitespace between the properties and their values.
```
schlessera: This file should be changed so that these single-line comments for each test case are converted into array keys. This way, PHPUnit would display the exact test case that is failing.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -2099,6 +2102,9 @@ public static function render_single_url_error_details( $validation_error, $term
 				</p>
 			</dd>
 
+			<dt><?php esc_html_e( 'Error code', 'amp' ); ?></dt>
+			<dd><code><?php echo esc_html( $validation_error['code'] ); ?></code></dd>
```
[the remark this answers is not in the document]
westonruter: > My favorite would be `<?= esc_html() ?>`,

Don't you mean `<?= esc_html__() ?>`?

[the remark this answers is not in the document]
westonruter: Actually, the latter is `esc_html()` because it is not using a translated string. I think I misunderstood your consistency request. So eventually, you'd like:

`<?= esc_html__() ?>` with `<?= esc_html() ?>`

rather than

`<?php esc_html_e() ?>` with `<?php echo esc_html() ?>`

PR #4146, opened 2020-01-21
Allow opting-out of AMP-to-AMP links, ensure 'Exit Reader Mode' link is non-AMP

### LINKED ISSUE #3689: Problem with AMP to AMP navigation
Hello,

AMP to AMP is what we were waiting for, but there is a bug in it.
We are using Transitional mode for single posts only.

## Bug Description

1. On AMP site version we have this link to go to non-AMP version of the post and we would like to keep it:
```<a href="' . get_permalink() . '">View non-AMP version</a>```
2. As mentioned before we only support single posts but AMP 1.4 is also changing all the tags/categories/search etc links to AMP version and it will create a lot of unnecessary redirects.

## Expected Behaviour

1. Somehow force keeping non-AMP link maybe by adding some attribute like `data-force-non-amp` or `data-keep-non-amp` or something like that.

Simple solution would be to change this:
https://github.com/ampproject/amp-wp/blob/develop/includes/sanitizers/class-amp-link-sanitizer.php#L130
```
if ( ! $element->hasAttribute( 'href' ) ) {
```
into:
```
if ( ! $element->hasAttribute( 'href' ) || $element->hasAttribute('data-force-non-amp')) {
```
And then we could just use something like this:
```<a href="' . get_permalink() . '" data-force-non-amp>View non-AMP version</a>```

2. Account for AMP settings:
obraz

`Homepage` and `Search` should be simple just matching home_url() and get_search_link() should be easy.
`Not Found (404)` I believe can be skipped.

I'm afraid it can be a little trickier for Archives, but in our case it can be solved by `data-force-non-amp` attribute, so when we are showing tags/categories/etc we could just simply add this attribute and problem solved.

## Additional context

 - WordPress version: 5.2.4
 - Plugin version: 1.4
 - AMP plugin template mode: Transitional
 - PHP version: 7.3
 - Browser: all
 - Device: all

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

### PR DESCRIPTION
**Pull request by kienstra**
## Summary
This allows opting out of AMP-to-AMP links via the filter `'excluded_links_from_amp_to_amp'` or adding `rel="not-amphtml"` to the `<a>`.

#### Rel attribute

```html
<a href="https://am.test/?p=195" rel="not-amphtml">To non-AMP</a>
```

...will produce a link to a non-AMP URL, with the `rel` stripped:

```html
<a href="https://am.test/?p=195">To non-AMP</a>
```

#### Filter
```php
// Excludes URLs from having AMP-to-AMP links.
add_filter(
	'amp_to_amp_excluded_links',
	static function() {
		return [
			'https://am.test/?p=195',
			'https://am.test/?p=199',
		 ];
	}
);
```

#### 'Exit Reader Mode' Link

Even with `add_filter( 'amp_to_amp_linking_enabled', '__return_true' );`, the 'Exit Reader Mode' link will now go to non-AMP:

filter-mode

Fixes #3689 

## Checklist

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -928,7 +928,7 @@ function amp_get_content_sanitizers( $post = null ) {
 		$sanitizers['AMP_Nav_Menu_Dropdown_Sanitizer'] = $theme_support_args['nav_menu_dropdown'];
 	}
 
-	if ( $amp_to_amp_linking_enabled ) {
+	if ( $amp_to_amp_linking_enabled && AMP_Theme_Support::STANDARD_MODE_SLUG !== AMP_Theme_Support::get_support_mode() ) {
 
 		/**
 		 * Filters the list of URLs which are excluded from being included in AMP-to-AMP linking.
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -929,9 +929,20 @@ function amp_get_content_sanitizers( $post = null ) {
 	}
 
 	if ( $amp_to_amp_linking_enabled ) {
-		$sanitizers['AMP_Link_Sanitizer'] = [
-			'paired' => ! amp_is_canonical(),
-		];
+
+		/**
+		 * Normally, when in a paired mode, links to the same origin will be to AMP.
+		 *
+		 * This allows passing URLs to exclude from having AMP-to-AMP links.
+		 *
+		 * @param string[] The URLs to exclude from having AMP-to-AMP links.
+		 */
+		$excluded_amp_links = apply_filters( 'excluded_links_from_amp_to_amp', [] );
+
+		$sanitizers['AMP_Link_Sanitizer'] = array_merge(
+			[ 'paired' => ! amp_is_canonical() ],
+			compact( 'excluded_amp_links' )
+		);
 	}
 
 	/**
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -938,6 +938,8 @@ function amp_get_content_sanitizers( $post = null ) {
 		 * when in Reader mode. This does not apply in Standard mode.
 		 * Only frontend URLs on the frontend need be excluded, as all other URLs are never made into AMP links.
 		 *
+		 * @since 1.5.0
+		 *
 		 * @param string[] The URLs to exclude from having AMP-to-AMP links.
 		 */
 		$excluded_urls = apply_filters( 'amp_to_amp_excluded_urls', [] );
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -937,7 +937,7 @@ function amp_get_content_sanitizers( $post = null ) {
 		 *
 		 * @param string[] The URLs to exclude from having AMP-to-AMP links.
 		 */
-		$excluded_amp_links = apply_filters( 'excluded_links_from_amp_to_amp', [] );
+		$excluded_amp_links = apply_filters( 'amp_to_amp_excluded_links', [] );
 
 		$sanitizers['AMP_Link_Sanitizer'] = array_merge(
 			[ 'paired' => ! amp_is_canonical() ],
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -931,9 +931,12 @@ function amp_get_content_sanitizers( $post = null ) {
 	if ( $amp_to_amp_linking_enabled ) {
 
 		/**
-		 * Normally, when in a paired mode, links to the same origin will be to AMP.
+		 * Filters the list of URLs which are excluded from being included in AMP-to-AMP linking.
 		 *
-		 * This allows passing URLs to exclude from having AMP-to-AMP links.
+		 * This only applies when the amp_to_amp_linking_enabled filter returns true,
+		 * which it does by default in Transitional mode. This filter can be used to opt-in
+		 * when in Reader mode. This does not apply in Standard mode.
+		 * Only frontend URLs on the frontend need be excluded, as all other URLs are never made into AMP links.
 		 *
 		 * @param string[] The URLs to exclude from having AMP-to-AMP links.
 		 */
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -940,11 +940,11 @@ function amp_get_content_sanitizers( $post = null ) {
 		 *
 		 * @param string[] The URLs to exclude from having AMP-to-AMP links.
 		 */
-		$excluded_amp_links = apply_filters( 'amp_to_amp_excluded_urls', [] );
+		$excluded_urls = apply_filters( 'amp_to_amp_excluded_urls', [] );
 
 		$sanitizers['AMP_Link_Sanitizer'] = array_merge(
 			[ 'paired' => ! amp_is_canonical() ],
-			compact( 'excluded_amp_links' )
+			compact( 'excluded_urls' )
 		);
 	}
 
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -940,7 +940,7 @@ function amp_get_content_sanitizers( $post = null ) {
 		 *
 		 * @param string[] The URLs to exclude from having AMP-to-AMP links.
 		 */
-		$excluded_amp_links = apply_filters( 'amp_to_amp_excluded_links', [] );
+		$excluded_amp_links = apply_filters( 'amp_to_amp_excluded_urls', [] );
 
 		$sanitizers['AMP_Link_Sanitizer'] = array_merge(
 			[ 'paired' => ! amp_is_canonical() ],
--- a/includes/sanitizers/class-amp-link-sanitizer.php
+++ b/includes/sanitizers/class-amp-link-sanitizer.php
@@ -149,11 +149,10 @@ class AMP_Link_Sanitizer extends AMP_Base_Sanitizer {
 				continue;
 			}
 
-			$href           = $element->getAttribute( 'href' );
-			$rel            = $element->hasAttribute( 'rel' ) ? trim( $element->getAttribute( 'rel' ) ) : '';
-			$is_rel_non_amp = self::REL_VALUE_NON_AMP_TO_AMP === $rel;
+			$href = $element->getAttribute( 'href' );
+			$rel  = $element->hasAttribute( 'rel' ) ? trim( $element->getAttribute( 'rel' ) ) : '';
 
-			if ( $is_rel_non_amp ) {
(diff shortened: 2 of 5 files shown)


includes/sanitizers/class-amp-link-sanitizer.php
```diff
@@ -132,11 +150,20 @@ public function process_links() {
 			}
 
 			$href = $element->getAttribute( 'href' );
-
-			if ( $this->is_frontend_url( $href ) && '#' !== substr( $href, 0, 1 ) ) {
+			$rel  = $element->hasAttribute( 'rel' ) ? trim( $element->getAttribute( 'rel' ) ) : '';
```
kienstra: @todo: handle a case where there is more than one rel, like `rel="not-amphtml nofollow"`. It looks like there can be multiple [space-separated values](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-rel).

westonruter: That's correct. It should rather do something like:

```suggestion
			$rel  = $element->hasAttribute( 'rel' ) ? array_filter( preg_split( '/\s+/' $element->getAttribute( 'rel' ) ) ) : [];
```

And then check:

```php
$pos = array_search( self::REL_VALUE_NON_AMP_TO_AMP, $rel );
if ( false !== $pos ) {
    unset( $rel[ $pos ] );
    if ( empty( $rel ) ) {
        $element->removeAttribute('rel');
    } else {
        $element->setAttribute('rel', implode( ' ', $rel ));
    }
}
```

includes/sanitizers/class-amp-link-sanitizer.php
```diff
@@ -31,14 +31,32 @@ class AMP_Link_Sanitizer extends AMP_Base_Sanitizer {
...
+
+	/**
+	 * The rel attribute value that will force non-AMP links.
+	 *
+	 * Normally, in a paired mode, links to the same origin will be for AMP.
+	 * But by adding this rel value, the link will be to non-AMP.
+	 *
+	 * @var string
+	 */
+	const REL_VALUE_NON_AMP_TO_AMP = 'not-amphtml';
```
westonruter: I don't necessarily have a better suggestion than `not-amphtml` but I don't feel entirely satisfied with this.

Maybe instead of `not-amphtml` it should be just `noamphtml`. This could make it more inline with `nofollow`, noopener`, and `noreferrer`. Note there is a [`noopener`](https://html.spec.whatwg.org/multipage/links.html#link-type-noopener) _and_ an [`opener`](https://html.spec.whatwg.org/multipage/links.html#link-type-opener) link relation, as seen in the [Link Relations](https://www.iana.org/assignments/link-relations/link-relations.xhtml) registry:

> image


So having an `amphtml` and a `noamphtml` would seem like it better follows this precedent.

```suggestion
	const REL_VALUE_NON_AMP_TO_AMP = 'noamphtml';
```

This would also help avoid confusion between `not-amphtml` and `non-amphtml`.

includes/sanitizers/class-amp-link-sanitizer.php
kienstra: Thanks, https://github.com/ampproject/amp-wp/pull/4146/commits/8a2e3dadf769d403b9b1e52866d1c7aab8d84a7d uses these snippets and adds some test cases.

includes/sanitizers/class-amp-link-sanitizer.php
kienstra: Good point, from that [Link Relations link](https://www.iana.org/assignments/link-relations/link-relations.xhtml) you shared, `no*` is the convention.

kienstra: Committed with https://github.com/ampproject/amp-wp/pull/4146/commits/640834ff7046455cbe91138e388c8e9f6e8147be

templates/header-bar.php
```diff
@@ -34,7 +34,7 @@
 		<?php $canonical_link_url = $this->get( 'post_canonical_link_url' ); ?>
 		<?php if ( $canonical_link_url ) : ?>
 			<?php $canonical_link_text = $this->get( 'post_canonical_link_text' ); ?>
-			<a class="amp-wp-canonical-link" href="<?php echo esc_url( $canonical_link_url ); ?>">
+			<a class="amp-wp-canonical-link" rel="not-amphtml" href="<?php echo esc_url( $canonical_link_url ); ?>">
```
westonruter:
```suggestion
			<a class="amp-wp-canonical-link" rel="noamphtml" href="<?php echo esc_url( $canonical_link_url ); ?>">
```

includes/amp-helper-functions.php
```diff
@@ -929,9 +929,20 @@ function amp_get_content_sanitizers( $post = null ) {
...
 	if ( $amp_to_amp_linking_enabled ) {
-		$sanitizers['AMP_Link_Sanitizer'] = [
-			'paired' => ! amp_is_canonical(),
-		];
+
+		/**
+		 * Normally, when in a paired mode, links to the same origin will be to AMP.
+		 *
+		 * This allows passing URLs to exclude from having AMP-to-AMP links.
+		 *
```
westonruter: Please add a `@since` version here.

includes/amp-helper-functions.php
```diff
@@ -929,9 +929,20 @@ function amp_get_content_sanitizers( $post = null ) {
 	}
 
 	if ( $amp_to_amp_linking_enabled ) {
```
westonruter: Let's add a check here to make sure this conditional never is entered when in Standard mode.

```suggestion
	if ( $amp_to_amp_linking_enabled && AMP_Theme_Support::STANDARD_MODE_SLUG !== AMP_Theme_Support::get_support_mode() ) {
```

AMP-to-AMP linking is only relevant in a paired AMP mode, that is Transitional and Reader.

includes/amp-helper-functions.php
```diff
@@ -929,9 +929,20 @@ function amp_get_content_sanitizers( $post = null ) {
...
-		$sanitizers['AMP_Link_Sanitizer'] = [
-			'paired' => ! amp_is_canonical(),
-		];
+
+		/**
+		 * Normally, when in a paired mode, links to the same origin will be to AMP.
+		 *
+		 * This allows passing URLs to exclude from having AMP-to-AMP links.
```
westonruter: Suggested docblock:

> Filters the list of URLs which are excluded from being included in AMP-to-AMP linking.
> 
> This only applies when the `amp_to_amp_linking_enabled` filter returns true,
> which it does by default in Transitional mode. This filter can be used to opt-in
> when in Reader mode. This does not apply in Standard mode. Only frontend
> URLs on the frontend need be excluded, as all other URLs are never made into AMP links.

includes/amp-helper-functions.php
```diff
@@ -929,9 +929,20 @@ function amp_get_content_sanitizers( $post = null ) {
...
+		 *
+		 * @param string[] The URLs to exclude from having AMP-to-AMP links.
+		 */
+		$excluded_amp_links = apply_filters( 'amp_to_amp_excluded_links', [] );
```
westonruter: A subtle difference, but it is the list of URLs that is being filtered not the list of links.

```suggestion
		$excluded_amp_links = apply_filters( 'amp_to_amp_excluded_urls', [] );
```

includes/sanitizers/class-amp-link-sanitizer.php
```diff
@@ -31,14 +31,32 @@ class AMP_Link_Sanitizer extends AMP_Base_Sanitizer {
...
 	 * Placeholder for default arguments, to be set in child classes.
 	 *
 	 * @var array
 	 */
 	protected $DEFAULT_ARGS = [ // phpcs:ignore WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase
-		'paired'       => false, // Only set to true when in a paired mode (will be false when amp_is_canonical()). Controls whether query var is added.
-		'meta_content' => self::DEFAULT_META_CONTENT,
+		'paired'             => false, // Only set to true when in a paired mode (will be false when amp_is_canonical()). Controls whether query var is added.
+		'meta_content'       => self::DEFAULT_META_CONTENT,
+		'excluded_amp_links' => [], // URLs in this won't have AMP-to-AMP links in a paired mode.
```
westonruter:
```suggestion
		'excluded_urls' => [], // URLs in this won't have AMP-to-AMP links in a paired mode.
```

and change the other instances of this string.

includes/sanitizers/class-amp-link-sanitizer.php
```diff
@@ -132,12 +150,26 @@ public function process_links() {
...
+					$element->removeAttribute( 'rel' );
+				} else {
+					$element->setAttribute( 'rel', implode( ' ', $rel ) );
+				}
+			} elseif (
+				$this->is_frontend_url( $href )
+				&&
+				'#' !== substr( $href, 0, 1 )
+				&&
+				! in_array( $href, $this->args['excluded_amp_links'], true )
```
westonruter: Any fragment identifier should be omitted from consideration here.

```suggestion
				! in_array( strtok( $href, '#' ), $this->args['excluded_amp_links'], true )
```

includes/amp-helper-functions.php
kienstra: Sure, I committed this verbatim in https://github.com/ampproject/amp-wp/pull/4146/commits/77118d73956f7551eb4e108eb8150b1dc4585ab6, though the line breaks are slightly different.

includes/amp-helper-functions.php
kienstra: Sure, good point. That's changed in https://github.com/ampproject/amp-wp/pull/4146/commits/d953a27f88346787df5be725425dd45b1fee4915

includes/sanitizers/class-amp-link-sanitizer.php
kienstra: Sure, that parameter is renamed in https://github.com/ampproject/amp-wp/pull/4146/commits/b6dda38531c77838a944ef1d119b635115b1bce7

includes/sanitizers/class-amp-link-sanitizer.php
kienstra: Sure, https://github.com/ampproject/amp-wp/pull/4146/commits/f85e8be23179001e348fd2a2dd8d0639a73a411f applies that change, and adds a test case for it.

includes/amp-helper-functions.php
kienstra: Sure, it's added in https://github.com/ampproject/amp-wp/pull/4146/commits/160eb877bb8b838ddc61a60b5c60e9c2e93ee0ff

PR #4155, opened 2020-01-21
Add group resizing.

### LINKED ISSUE #3904: Moveable: Resizing multiple elements
## Feature description

Split out from #3820 to address resizing multiple elements.

---------------

## Acceptance criteria

* It should be possible to resize multiple elements at the same time
* The elements always need to keep ratio while resizing.
* Text element's font size has to adjust while resizing.
* Only corner handles should be visible.

## Implementation brief

When more than one element is selected, one additional Moveable will be displayed which has all the selected elements as targets in an array (`target={ arrayOfTargets }`). Moveable displays an outline around the selected elements by default:
Screenshot 2019-12-06 at 18 02 55

We will need to set the initial values to the `frames` as an array of objects, e.g.
```
const frames = [
	{
		translate: [ 0, 0 ],
		rotate: rotationAngle1,
	},
	{
		translate: [ 0, 0 ],
		rotate: rotationAngle2,
	},
];
```
where rotation angle should be the respective initial rotation angle of each element.

### Resizing
Resizing works in a similar way as resizing for a single element with the difference of having to loop through the targets to assign the values to each target separately.

Within this issue, the element values in the state are updated only when ending the activity, within `onResizeGroupEnd`.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

miina:
## Summary

Fixes #3904 

## Checklist

### DIFF
--- a/assets/src/edit-story/components/canvas/multiSelectionMovable.js
+++ b/assets/src/edit-story/components/canvas/multiSelectionMovable.js
@@ -13,6 +13,9 @@ import { useRef, useEffect } from '@wordpress/element';
  */
 import Movable from '../movable';
 import { useStory } from '../../app/story';
+import calculateFitTextFontSize from '../../utils/calculateFitTextFontSize';
+
+const CORNER_HANDLES = [ 'nw', 'ne', 'sw', 'se' ];
 
 function MultiSelectionMovable( { selectedElements, nodesById } ) {
 	const moveable = useRef();
@@ -35,6 +38,8 @@ function MultiSelectionMovable( { selectedElements, nodesById } ) {
 			x: element.x,
 			y: element.y,
 			rotationAngle: element.rotationAngle,
+			type: element.type,
+			content: element.content,
 		};
 	} );
 	// Not all targets have been defined yet.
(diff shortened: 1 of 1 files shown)


assets/src/edit-story/components/canvas/multiSelectionMovable.js
```diff
@@ -137,6 +145,35 @@ function MultiSelectionMovable( { selectedElements, nodesById } ) {
...
+			} }
+			onResizeGroup={ ( { events } ) => {
+				events.forEach( ( { target, width, height, drag }, i ) => {
+					const sFrame = frames[ i ];
+					const isText = 'text' === targetList[ i ].type;
+					target.style.width = `${ width }px`;
+					target.style.height = `${ height }px`;
+					if ( isText ) {
+						// For text: update font size, too.
+						target.style.fontSize = calculateFitTextFontSize( target.firstChild, height, width );
```
swissspidy: Shouldn't this also be called in `onGroupEventEnd` to properly update the element? Just like `SingleSelectionMovable` does in `onResizeEnd`

miina: Thanks for catching that, fixed in 982719e.

PR #4163, opened 2020-01-22
Address notice and console error in Paired Browsing from async

### LINKED ISSUE #4127: Paired browsing sporadically fails to detect loaded AMP page
## Bug Description

When I have DevTools open and I have “Disable Cache” checked, about 10% of the time that I load paired browsing the AMP version shows the modal:

> The navigated URL is not available for paired browsing.

image

If I reload to try again, then the issue normally goes away, but this is because the issue happens sporadically. 

I can reproduce it both with DevTools open and with doing a hard-reload (Cmd+Shift+R).

There appears to be a race condition that is causing the detection in the AMP window to fail or vice-versa, or rather, detection that the page was able to load properly.

In the console I see an error that may be a clue for what is going on:

image

This appears to be due to the fact that the paired browsing scripts are being added with `async` attributes:

image

This is resulting in inconsistent loading order, and thus sporadic failures.

So we need to prevent adding the `async` attribute on these scripts, which probably just means removing it from this line:

https://github.com/ampproject/amp-wp/blob/0b91f99566b7254cc5688523291864d3ff49488e/includes/class-amp-theme-support.php#L2485

## Steps to reproduce

1. Enable Transitional mode
2. Open DevTools and Disable Cache
3. Open paired browsing
4. Try reloading 20 times and notice that a couple times the above error is shown.

## Additional context

 - WordPress version:
 - Plugin version:
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

kienstra:
## Summary
In Paired Browsing, this prevents the occasional notice and console error that Weston found, using his suggested fix.

Fixes #4127

### Before 
Sometimes, this would appear on a refresh with a cleared cache:

>TypeError: Cannot read property 'domReady' of undefined.

navigated-url-cache

### After
That issue didn't occur after refreshing 20 times.

## Checklist

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -2476,14 +2476,13 @@ class AMP_Theme_Support {
 			true
 		);
 
-		// Whitelist enqueued script for AMP dev mdoe so that it is not removed.
+		// Whitelist enqueued script for AMP dev mode so that it is not removed.
 		// @todo Revisit with <https://github.com/google/site-kit-wp/pull/505#discussion_r348683617>.
 		add_filter(
 			'script_loader_tag',
 			static function( $tag, $handle ) {
 				if ( is_amp_endpoint() && self::has_dependency( wp_scripts(), 'amp-paired-browsing-client', $handle ) ) {
-					$attrs = [ AMP_Rule_Spec::DEV_MODE_ATTRIBUTE, 'async' ];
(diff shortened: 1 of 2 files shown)


kienstra:
Hi Weston, it looks like you saw this already, but this is ready for review whenever you have a chance.

You're fast with seeing these 😄

PR #4153, opened 2020-01-21
Grid View Modal

swissspidy:
## Summary

Implements the reorder grid view using `react-modal`.

It is still a bit rough and needs some polishing, but generally works.

Fixes #3836.

## Checklist

- [x] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [x] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [x] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

PR #4167, opened 2020-01-23
Page thumb previews

### LINKED ISSUE #3777: Port Existing Page (Block) Preview
## Feature description

This ticket was taken from a row in this [spreadsheet](https://docs.google.com/spreadsheets/d/1nqJ-iOypTziWFjWBIBoJUc0LHW3gTeUnLkbjVH_eW4g/edit#gid=0). This ticket can be used to track research and work related to porting this existing feature from the "old" editor to the "new" editor. 

Page (Block) Preview:
- Component that allows displaying a block (or a set of blocks aka template) in a disabled mode just for previewing, users can't interact with it but can see how it looks like. That was used for templates and also for reordering Pages.

UX Required: Yes

---------------

## Acceptance criteria

* The existing Page (Block) Preview is ported to the new stories editor
* The Page (Block) Preview is integrated into the new [design](https://www.figma.com/file/Cd9iacVknHK62RO4pRwb68/Experimental?node-id=0%3A1)

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

miina:
## Summary

Currently based on #4153 

Fixes #3777 (https://github.com/swissspidy/web-stories-wp/issues/91)

- Uses the same element for both FE output and output for previews.
See the comments above: this might not be a good idea, we might prefer a separate `Preview` component, especially when things will get more complicated and the FE will use more AMP HTML attributes and components.

- Also adds fullbleed support for FE, that was missing.

## Checklist


miina:
Note: The logic is reusing the `Output` of each element (previously `Save`) which also brings out issues where output hasn't been implemented. E.g. full-bleed images currently do not work in the FE.

Will check if I can include this within this PR as well.

swissspidy:
I wonder whether it makes sense for `Output` to handle both the save and the preview case, or whether those should be separate. Some thoughts on that would be good.

Also, `previewSizeMultiplier` indicates issues with responsive page size. In https://github.com/ampproject/amp-wp/pull/4129#issuecomment-577334689 @dvoytenko indicated to look into responsive size of the canvas next. Sounds like some overlap?

miina:
> I wonder whether it makes sense for Output to handle both the save and the preview case, or whether those should be separate. Some thoughts on that would be good.

Thought about it and at this point at least, it seemed to be a good case for matching `Output` for both preview and FE. The main difference is AMP vs "normal" HTML for images and videos and font size. However, it might get more complicated in the future with more AMP-related params being added. Not sure. 

IDK what's better. WDYT?

> Also, `previewSizeMultiplier` indicates issues with responsive page size. In #4129 (comment) @dvoytenko indicated to look into responsive size of the canvas next. Sounds like some overlap?

`previewSizeMultiplier` is used for font size adjustment only based, the rest is already responsive due to using `%`. Not sure what would be a better solution here for fonts, an alternative would be calculating and assigning the font size dynamically post-rendering based on the difference between the full Page size and the preview size. That could also be added later though if needed.

Thoughts?

miina:
I suppose that for clarity and not making things confusing, it would be better to have a separate `Preview` instead.

miina:
Restored `save.js` and added separate `preview.js`. It makes sense especially with #4169 that @dvoytenko is working on where the shared parts of `save.js` and `preview.js` are separated into `saveElement`.

miina:
@swissspidy This should be ready for review now.

PR #4169, opened 2020-01-24
Editor page resize and units conversion

dvoytenko:
## Summary

Partial for #27.

TODO: 

- [ ] Migrate all element types
- [ ] Migrate multi-selection movable
- [ ] Migrate carousel to use `useResizeEffect` hook
- [ ] Remove old APIs

This PR does the following:

* Implements breakpoints for a couple of different allowed sizes of the "page" area.
* Uses `units` system to convert between "data" and "editor" pixels.
* Migrates element templates to use `{element, box, ...}` shape to allow use of both data and physical pixels as necessary.
* Cleans up typing a little.

## Checklist

### DIFF
--- a/assets/src/edit-story/components/canvas/carousel/index.js
+++ b/assets/src/edit-story/components/canvas/carousel/index.js
@@ -66,6 +66,7 @@ function Carousel() {
 	const openModal = useCallback( () => setIsGridViewOpen( true ), [ setIsGridViewOpen ] );
 	const closeModal = useCallback( () => setIsGridViewOpen( false ), [ setIsGridViewOpen ] );
 
+	// QQQQ: useResizeObserver
 	useLayoutEffect( () => {
 		const observer = new ResizeObserver( ( entries ) => {
 			for ( const entry of entries ) {
--- a/assets/src/edit-story/units/dimensions.js
+++ b/assets/src/edit-story/units/dimensions.js
@@ -1,0 +1,39 @@
+
+/**
+ * Internal dependencies
+ */
+import { PAGE_WIDTH, PAGE_HEIGHT } from '../constants';
+
+export function dataUnit( v ) {
+	return Number( v.toFixed( 0 ) );
+}
+
+export function editorUnit( v ) {
+	return Number( v.toFixed( 5 ) );
+}
+
+export function dataToEditorX( x, pageWidth ) {
+	return editorUnit( x * pageWidth / PAGE_WIDTH );
+}
+
+export function dataToEditorY( y, pageHeight ) {
+	return editorUnit( y * pageHeight / PAGE_HEIGHT );
+}
+
+export function editorToDataX( x, pageWidth ) {
+	return dataUnit( x * PAGE_WIDTH / pageWidth );
+}
+
+export function editorToDataY( y, pageHeight ) {
+	return dataUnit( y * PAGE_HEIGHT / pageHeight );
+}
+
+export function getBox( { x, y, width, height, rotationAngle, isFullbleed }, pageWidth, pageHeight ) {
+	return {
+		x: dataToEditorX( isFullbleed ? 0 : x, pageWidth ),
+		y: dataToEditorY( isFullbleed ? 0 : y, pageHeight ),
+		width: dataToEditorX( isFullbleed ? PAGE_WIDTH : width, pageWidth ),
+		height: dataToEditorY( isFullbleed ? PAGE_HEIGHT : height, pageHeight ),
+		rotationAngle: isFullbleed ? 0 : rotationAngle,
+	};
+}
--- a/assets/src/edit-story/units/dimensions.js
+++ b/assets/src/edit-story/units/dimensions.js
@@ -4,30 +4,85 @@
  */
 import { PAGE_WIDTH, PAGE_HEIGHT } from '../constants';
 
-export function dataUnit( v ) {
+/**
+ * Rounds the pixel value to the max allowed precision in the "data" space.
+ *
+ * @param {number} v The value to be rounded.
+ * @return {number} The value rounded to the "data" space precision.
+ */
+export function dataPixels( v ) {
 	return Number( v.toFixed( 0 ) );
 }
 
-export function editorUnit( v ) {
+/**
+ * Rounds the pixel value to the max allowed precision in the "editor" space.
+ *
+ * @param {number} v The value to be rounded.
+ * @return {number} The value rounded to the "editor" space precision.
+ */
+export function editorPixels( v ) {
 	return Number( v.toFixed( 5 ) );
 }
 
+/**
+ * Converts a "data" pixel value to the value in the "editor" space along
+ * the horizontal (X) dimension.
+ *
+ * @param {number} x The value to be converted.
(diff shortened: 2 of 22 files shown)


assets/src/edit-story/components/canvas/carousel/index.js
```diff
@@ -74,6 +74,7 @@ function Carousel() {
 	const listRef = useRef();
 	const pageRefs = useRef( [] );
 
+	// QQQQ: useResizeObserver
```
barklund: 👍

assets/src/edit-story/app/story/save/saveElement.js
```diff
@@ -0,0 +1,37 @@
+/**
```
miina: I wonder if `saveElement` would be better placed under `elements` folder? Currently, half of the element output is separated. Perhaps it's fine.

Also, I'm wondering if `Save` should be changed to `Output` instead. Two reasons:
- In another PR (#4167) the save and preview are sharing quite a lot -- like all the wrapper params for example, only the inner element (`p` or `amp-img` vs `img`) is different. `Output` would allow using `saveElement` for both, just create a separate `preview.js` for the inner element.
- `Save` isn't actually doing/allowing the `Save` action, `Output` might be more accurate to describe the component anyway.

Not a crucial thing here though, could do that in #4167, too. Thoughts?

assets/src/edit-story/units/dimensions.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export function editorToDataY( y, pageHeight ) {
+	return dataUnit( y * PAGE_HEIGHT / pageHeight );
+}
+
+export function getBox( { x, y, width, height, rotationAngle, isFullbleed }, pageWidth, pageHeight ) {
+	return {
+		x: dataToEditorX( isFullbleed ? 0 : x, pageWidth ),
+		y: dataToEditorY( isFullbleed ? 0 : y, pageHeight ),
```
miina: I wonder if it would make sense to use the percentage values directly (storing data in % instead of px as now) and instead do the conversion in places before saving and displaying the data instead where necessary (e.g. in the Size panel and Moveable). Thoughts?

assets/src/edit-story/units/dimensions.js
```diff
@@ -0,0 +1,39 @@
...
+
+export function dataUnit( v ) {
+	return Number( v.toFixed( 0 ) );
+}
+
+export function editorUnit( v ) {
+	return Number( v.toFixed( 5 ) );
+}
+
+export function dataToEditorX( x, pageWidth ) {
```
miina: It would be good to have some short function comments here.

miina:
Looks good!

General question: what's the goal with responsiveness, should the Page + nav menu always be visible without scrolling, while displaying the Page as large as possible (with correct ratio)?

Also, just FI that some relevant PRs got merged meanwhile, there's also now [Grid View Modal](#4153) and new `preview.js` file for all elements for [preview in Thumbnails](#4167) which is similar to `save.js` and using `%` (except for the `fontSize` for Text element).

assets/src/edit-story/app/story/save/saveElement.js
dvoytenko: Yeah, I like `Output` better too. And, yeah, we could move `saveElement` as well. I do think they'll be the same for the most part. Not sure about preview right now, but the most drastic issue I discovered with `saveElement` and `renderToString` is that I wasn't able to use hooks inside of `Save`/`Output` templates. That partially makes sense and hopefully won't be an issue for `Preview` templates either. But basically, these render functions are not quite like other render functions (e.g. `Display` or `Edit`). Have you run into this as well?

assets/src/edit-story/units/dimensions.js
dvoytenko: I think we've did some analysis for this before and it seemed like we may want to arrive at an integer value at some point and hence we decided to pick a predefined data pixel. I still think it's easier. For instance, `Moveable` then uses the same coordinate system as we. While we could try to define our position in `%`, `Moveable` still measures and uses pixel values that we then compare, add, etc. If you'd like to dig deeper, should we reopen https://github.com/google/web-stories-wp/issues/71?

miina: Pixels definitely made sense with the idea to keep the width/height for Page the same all the time. Yes, `Moveable` would be one place where we'd need to convert between `%` and `px` which I suppose, would still require responsive converting anyway.

Perhaps it's safer to use `px` and if you think it's still easier then let's stick to that. Was just wondering for a moment if now, when the Page will be responsive, `%` might make sense in the editor, too. However, to prevent struggle and potential future surprises with `Moveable` and possible other libs then `px` can still be a better option.

dvoytenko:
> General question: what's the goal with responsiveness, should the Page + nav menu always be visible without scrolling, while displaying the Page as large as possible (with correct ratio)?

Yeah, pretty much. The goals are:
 1. All elements have to be always be visible within a reasonable workspace size.
 2. Header and menu are fixed height. Page and carousel are responsive.
 3. Page is responsive with breakpoints: only a few options available because we don't want to deal with numerous resolutions. So maybe 3 or so in the long term.
 4. Eventually carousel will have two modes: compact mode (only bars are visible), and responsive up to X px.

This PR handles everything, except carousel.

> Also, just FI that some relevant PRs got merged meanwhile, there's also now Grid View Modal and new preview.js file for all elements for preview in Thumbnails which is similar to save.js and using % (except for the fontSize for Text element).

Yep, I saw the `preview.js`. I think the best course of action now would be to move `save` stuff from this pull request into a separate one and combine/rearrange save/preview/output there.

assets/src/edit-story/units/dimensions.js
dvoytenko: Let's continue this thread in that issue. This PR makes this a bit more systematic. So it should help in migration if we decide to go down that path.

assets/src/edit-story/app/story/save/saveElement.js
dvoytenko: Moved this to https://github.com/ampproject/amp-wp/pull/4173

dvoytenko:
@miina I moved the "save" stuff to https://github.com/ampproject/amp-wp/pull/4173 for now. I'll rework it to `Output` and try to combine it with "preview" as well. But it definitely better to disentangle it from responsiveness questions for now.

assets/src/edit-story/units/dimensions.js
dvoytenko: Done.

assets/src/edit-story/app/story/save/saveElement.js
miina: > Have you run into this as well?

Nope, haven't noticed this issue so far, which hooks did you need to use within saving the output?

Note: we can continue the discussion in #4173

dvoytenko:
Moved to https://github.com/google/web-stories-wp/pull/136

assets/src/edit-story/app/story/save/saveElement.js
dvoytenko: Yep. I'll bring this back up in #4173

Issue #3284, opened 2019-09-17
Reduce scenarios where DB writes are performed during frontend requests


westonruter:
There are a few places where DB writes are being performed during plugin initialization or when performing normal frontend requests to the site:

https://github.com/ampproject/amp-wp/blob/f1af0aba711f25cd93ad9efb903764adeab54d4f/amp.php#L404-L420

(Removed `class-amp-story-templates.php` instance since code removed in #4315.)

(Removed `AMP_Theme_Support::check_for_cache_misses()` example since removed in #4178.)

These can potentially cause race conditions where concurrent requests both cause an update to be done at the same time. It's also not great for performance to have DB writes being performed in response to frontend requests.

I suggest that each of these cases be refactored with `wp_schedule_single_event()` which can then be used to perform the writes during WP Cron, and thus avoid the race condition.

It's true that scheduling a WP Cron event is itself a DB write, but at least this a more proper way to do it. Otherwise, there could be a repeating task every X minutes to see if the above tasks need to be performed.

archon810:
The primary reason we critically need something to be done here is in a master/slave WP env (using https://github.com/stuttter/ludicrousdb), after a master reboot, AMP plugin settings got completely reverted to their defaults.

I was able to replicate this twice after rebooting the master db server, and while it's unclear exactly why it happens or which race condition does it, it's very likely the cause is the code above that runs on every page load and initializes AMP plugin options if some data it expects isn't there (like AMP plugin version from db).

westonruter:
@archon810 Please comment out those lines above in the plugin and monitor if you ever see the issue occur.

archon810:
We're not planning on restarting the master for a while (it was a planned OS upgrade), and the changes would be wiped on the next plugin update when I run composer unless I clone the project into our local repo.

I wish I had a more reliable way to reproduce this :(

westonruter:
Good point by @schlessera in that some of these should continue to be synchronous, but gated behind requests to the admin.

archon810:
What's the status of this issue please? I haven't observed the bug lately but we haven't been restarting the master much, and it's not a guarantee that the bug is fixed.

westonruter:
There is no status update. It hasn't been prioritized since we haven't had any other reports of this being an issue, and you haven't been able to reliably reproduce it.

[...]

PR #4177, opened 2020-01-25
Add ability to prevent storing parsed CSS in transients

westonruter:
_Note: This PR's true author is @schlessera!_

## Summary

- Adds a new `amp_parsed_css_transient_caching_allowed` filter which allows a site to prevent storing parsed CSS in transients, by filtering the value to be false. This is important on sites which have highly variable CSS, resulting in a site filling up its `wp_options` table with transients.

	Note that when transient storage is disabled, the parsed CSS is still cached in the object cache even when an external object cache is not available. This ensures that the same CSS appearing multiple times on a page will not have to be re-parsed.

- Adds detection for a huge number of `SELECT COUNT(*) FROM $wpdb->options WHERE option_name LIKE '_transient_amp-parsed-stylesheet%'` and when a threshold is met, persists the fact that transient caching of stylesheets is disabled. This happens during a daily cron.

- Adds Site Health warning when transient caching is disabled (and external object cache is not present).

- Adds Site Health debugging information.

- Adds a warning to the Network Plugins screen on every large sites, to inform the administrator that orphaned scheduled events will be left behind.

### Transient caching of stylesheets enabled

Screen Shot 2020-03-21 at 16 21 34

### Transient caching of stylesheets disabled

Screen Shot 2020-03-21 at 16 22 54

### Transient caching of stylesheets not applicable

Screen Shot 2020-03-21 at 16 22 11

### Debug information

Image 2020-03-20 at 8 45 02 AM

### Network plugins screen warning

image

See #2092.

## Checklist


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1439,6 +1441,7 @@ private function get_parsed_stylesheet( $stylesheet, $options = [] ) {
 		$cache_key   = null;
 		$cached      = true;
 		$cache_group = 'amp-parsed-stylesheet-v24'; // This should be bumped whenever the PHP-CSS-Parser is updated or parsed format is updated.
+		$use_cache   = wp_using_ext_object_cache() || ! $this->args['allow_transient_caching'];
```
schlessera: The name is misleading, both object cache and transient cache are caches in some way.
`$use_cache` implies that caching in general is disabled if this is `false`.
```suggestion
		$use_object_cache = wp_using_ext_object_cache() || ! $this->args['allow_transient_caching'];
```

schlessera:
Force-pushing a rebase.

schlessera:
@googlebot I consent.

schlessera:
@westonruter To follow up with the discussion we just had in the sync call, we should talk about how best to handle this for now: https://github.com/ampproject/amp-wp/pull/4177/files#diff-4805c837d3390517773926fb26726bcdR306-R323

schlessera:
Also, I'm open for any suggestions if you know a better way of testing the background task.

tests/php/test-class-monitor-css-transient-caching.php
```diff
@@ -0,0 +1,128 @@
...
+
+		$monitor = new MonitorCssTransientCaching();
+		$monitor->register();
+
+		$timestamp = wp_next_scheduled( MonitorCssTransientCaching::EVENT_NAME );
+
+		$this->assertNotFalse( $timestamp );
+		$this->assertInternalType( 'int', $timestamp );
+		$this->assertGreaterThan( 0, $timestamp );
+	}
```
westonruter: This test is failing in PHP 5.6 for some reason:

```
1) Test_Monitor_CSS_Transient_Caching::test_event_gets_scheduled
AmpProject\AmpWP\Exception\FailedToRegisterBackgroundTask: Failed to schedule the event 'amp_monitor_css_transient_caching' with WordPress.
/tmp/wordpress/src/wp-content/plugins/amp/src/Exception/FailedToRegisterBackgroundTask.php:30
/tmp/wordpress/src/wp-content/plugins/amp/src/BackgroundTask/CronBasedBackgroundTask.php:49
/tmp/wordpress/src/wp-content/plugins/amp/tests/php/test-class-monitor-css-transient-caching.php:41
```

Or maybe its a WordPress≤5.0 issue:

image

src/BackgroundTask/MonitorCssTransientCaching.php
```diff
@@ -0,0 +1,216 @@
...
+
+	/**
+	 * Get the threshold to check the moving average against.
+	 *
+	 * This can be filtered via the 'amp_css_transient_monitoring_threshold' filter.
+	 *
+	 * @return float Threshold to use.
+	 */
+	private function get_threshold() {
+		$threshold = (float) apply_filters( 'amp_css_transient_monitoring_threshold', self::DEFAULT_THRESHOLD );
```
westonruter: Should add phpdoc comment for this.

src/BackgroundTask/MonitorCssTransientCaching.php
```diff
@@ -0,0 +1,216 @@
...
+
+	/**
+	 * Get the sampling range to limit the time series to for calculating the moving average.
+	 *
+	 * This can be filtered via the 'amp_css_transient_monitoring_sampling_range' filter.
+	 *
+	 * @return int Sampling range to use.
+	 */
+	private function get_sampling_range() {
+		$sampling_range = (int) apply_filters( 'amp_css_transient_monitoring_sampling_range', self::DEFAULT_SAMPLING_RANGE );
```
westonruter: Should add phpdoc comment for this.

amp.php
```diff
@@ -303,6 +303,24 @@ function amp_deactivate() {
...
+		$exception_class = get_class( $exception );
+		$error_message   = "Exception '{$exception_class}' thrown in background task '{$background_task}': {$exception->getMessage()}";
+		if ( class_exists( 'WP_CLI' ) ) {
+			WP_CLI::warning( $error_message );
+		} else {
+			error_log( $error_message ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+		}
+	}
+}
+
```
westonruter: > To follow up with the discussion we just had in the sync call, we should talk about how best to handle this for now: https://github.com/ampproject/amp-wp/pull/4177/files#diff-4805c837d3390517773926fb26726bcdR306-R323

@schlessera I don't really have any better suggestion for the immediate term on where to put this, other than to plop it in `amp.php` like you've done. We'll need to look in the future about how better to organize the `amp.php` bootstrap as part of modularization effort.

The one thing that I think should be done, however, is to avoid calling this right here immediately when the bootstrap is being invoked. This will cause a DB write to occur and potentially race condition on a heavily-trafficked site. This would be adding one more place in addition to the frontend writes being called out in https://github.com/ampproject/amp-wp/issues/3284. 

Since `register_activation_hook()` doesn't work with multisite, what about using `admin_init` instead? This will ensure it only fires once a user has logged in. See:  https://developer.wordpress.org/reference/functions/register_activation_hook/#comment-2100

westonruter: Another option would be to put it inside of an `amp_plugin_update` action, but this still runs during frontend requests. As noted in https://github.com/ampproject/amp-wp/issues/3284#issuecomment-601461083 it could also be moved to `admin_init`. In that way it would be similar to the trigger for the DB upgrade.

schlessera: > This will cause a DB write to occur and potentially race condition on a heavily-trafficked site.

AFAICT, right now, the DB writes are:
- once per day in the `register()` task to schedule a new cron job if none is scheduled;
- once per day in the `process()` that is run via cron.

Am I missing something?

westonruter: Yes, there is the additional write to schedule the interval in the first place. That is, when `register` is called for the very first time upon upgrading, then `wp_next_scheduled()` will return `false` which will then cause `wp_schedule_event()` to be called:

https://github.com/ampproject/amp-wp/blob/d28a5ef1f6e8f930230c357c75c9c30492c60026/src/BackgroundTask/CronBasedBackgroundTask.php#L100-L106

This will then cause `_set_cron_array()` to be called, which will do a DB write. If `wp_schedule_event()` is limited to only happen at `admin_init`, then no cron update would happen on the frontend. When an event transpires, the next event would get scheduled during the cron request, not during a frontend request.

By putting this logic into an `admin_init` hook, this DB write to schedule the event won't happen on the frontend. Granted, this would mean splitting the `register` method into two: one to schedule and one to add the hooks. The schedule part could be done at `admin_init` whereas the hook addition would need to be done with every request (or at least when doing cron).

Maybe I'm being overly cautious about this, and there invariably a lot of other DB writes that happen during frontend requests (especially when transients are used without a persistent object cache). But I'm just trying to think of any ways to make this more robust.

schlessera: I just realised that the schedule call should actually be done on plugin activation instead. That makes it more robust and ensures it is only run once, which is enough, as we are scheduling a recurring event.

schlessera: I've rewritten some parts now to both streamline the way everything is executed and heavily simplify the cron logic.

schlessera:
So I've added a very, _very_ small "Services" subsystem now to get around the hacky instantiation that we had in the previous iteration of this PR.

This subsystem:
- is very lightweight and straight-forward
- doesn't use any magic of heavy engineering concepts
- can be used gradually and does not need to be enforced

More importantly, though, it gives structure to the way we register, activate or deactivate code.

For the next piece of logic we want to add like this:
1. create a class that does its thing
2. add needed hooks into the `register()` method
3. if it needs to act on plugin activation, implement the `HasActivation` interface
4. if it needs to act on plugin deactivation, implement the `HasDeactivation` interface

No change to `amp.php` or other parts of the code needed anywhere, the change will be completely self-contained (which should be the goal of maintainable code).

Please let me know what you think!

amp.php
schlessera: The situation now should be:
- one DB write on plugin activation
- one DB write on plugin deactivation
- one DB write per day during a wp_cron execution

schlessera: Note: The wp_cron one can be filtered out of the frontend processing and only be run as an actual cron background task via filtering and server setup. This should solve the problem completely for high-traffic sites.

schlessera: The plugin activation hook does work on multisite. However, it will only schedule the cron event for the primary site. I can add code to cycle through the sites and schedule an event for each of them.

schlessera: The above is the case if you network-activate the plugin. We could also disallow network-activation, so you'll have to activate on a subsite by subsite basis instead...

westonruter: Here's some code that apparently handles the network activation case and the multisite blog creation case: https://wordpress.stackexchange.com/a/181150

I am not an authority on this!

westonruter: cc @felixarntz!

westonruter: > We could also disallow network-activation, so you'll have to activate on a subsite by subsite basis instead...

No, we should still allow it to be network-activated.

felixarntz: @westonruter @schlessera In order to make a site-wide action like this "reliable" for multisite it needs to happen on a hook like `admin_init` (or if earlier is preferable something like `init` with `is_admin()` check).

It's a pain, but the activation hooks don't cascade to the per-site level in multisite during network activation (scale is an issue either way), and they also don't fire when a new site is created if the plugin is already network-active. So using an as-inexpensive-as-possible DB call to check whether the action still needs to be done (only in admin!) would be my recommendation for now.

schlessera: @felixarntz But that will inevitably run into race conditions then, and we'll end up running the cron job multiple times per day, potentially disabling the transient caching much too soon.

schlessera: Note: In this case, not running the cron job at all is highly preferable to running it multiple times because of a race condition.

schlessera: Why would it not be reliable to cycle over all blogs (once!) on plugin activation (something that is under the control of the administrator, and listening to site creation?

felixarntz: @schlessera Potentially yes, but we can reduce the chance for this to a minimum, e.g. we could only schedule the cron job if an administrator is logged in. I would say it's highly unlikely that a site has so many administrators that more than one loads an admin page in the same given second right in the moment where this action is executed.

felixarntz: @schlessera 
> Why would it not be reliable to cycle over all blogs (once!) on plugin activation (something that is under the control of the administrator, and listening to site creation?

Yes we could do that but then you'll run into scaling issues with larger multisites.

src/Services.php
```diff
@@ -0,0 +1,100 @@
...
+final class Services {
+
+	/**
+	 * List of known services.
+	 *
+	 * @var string[]
+	 */
+	const ALL = [
+		MonitorCssTransientCaching::class,
+	];
```
westonruter: Eventually would this be specified externally? It seems a bit weird to put them in here.

src/Services.php
```diff
@@ -0,0 +1,100 @@
...
+	 */
+	private static function maybe_instantiate() {
+		if ( ! empty( self::$instances ) ) {
+			return;
+		}
+
+		foreach ( self::ALL as $service_class ) {
+			self::$instances[ $service_class ] = new $service_class();
+		}
+	}
```
westonruter: Could this be turned into an `instances` method instead? Then instead of directly calling `maybe_instantiate` and then directly iterating over `$instances`, other methods can just call the `instances` method.

```suggestion
	/**
	 * Get the services.
	 *
	 * @return Service[] Services.
	 */
	private static function instances() {
		if ( empty( self::$instances ) ) {
			foreach ( self::ALL as $service_class ) {
				self::$instances[ $service_class ] = new $service_class();
			}
		}
		return self::$instances;
	}
```

src/Services.php
```diff
@@ -0,0 +1,100 @@
...
+	 * Run deactivation logic.
+	 *
+	 * This should be hooked up to the WordPress deactivation hook.
+	 *
+	 * @return void
+	 */
+	public static function deactivate() {
+		self::maybe_instantiate();
+
+		foreach ( self::$instances as $service ) {
```
westonruter:
```suggestion
		foreach ( self::instances() as $service ) {
```

src/Services.php
```diff
@@ -0,0 +1,100 @@
...
+	 * Run activation logic.
+	 *
+	 * This should be hooked up to the WordPress deactivation hook.
+	 *
+	 * @return void
+	 */
+	public static function activate() {
+		self::maybe_instantiate();
+
+		foreach ( self::$instances as $service ) {
```
westonruter:
```suggestion
		foreach ( self::instances() as $service ) {
```

src/Services.php
```diff
@@ -0,0 +1,100 @@
...
+
+	/**
+	 * Register the services with the system.
+	 *
+	 * @return void
+	 */
+	public static function register() {
+		self::maybe_instantiate();
+
+		foreach ( self::$instances as $service ) {
```
westonruter:
```suggestion
		foreach ( self::instances() as $service ) {
```

src/BackgroundTask/MonitorCssTransientCaching.php
```diff
@@ -0,0 +1,228 @@
...
+	 * @todo This has arbitrary arguments to allow for testing, as we don't have dependency injection for services.
+	 *       With dependency injection, we could for example inject a Clock object and mock it for testing.
+	 *
+	 * @param DateTimeInterface $date            Optional. Date to use for timestamping the processing (for testing).
+	 * @param int               $transient_count Optional. Count of transients to use for the processing (for testing).
+	 * @return void
+	 * @throws Exception If a date could not be instantiated.
+	 */
+	public function process( DateTimeInterface $date = null, $transient_count = null ) {
+		if ( $this->is_css_transient_caching_disabled() ) {
```
westonruter: The monitor should short-circuit if an external object cache is used, no? The query in `query_css_transient_count` will always be 0:

```suggestion
		if ( wp_using_ext_object_cache() || $this->is_css_transient_caching_disabled() ) {
```

src/Admin/SiteHealth.php
```diff
@@ -267,6 +291,26 @@ public function add_debug_information( $debugging_information ) {
...
+						'amp_css_transient_caching_sampling_range' => [
+							'label'   => esc_html__( 'Sampling range for monitoring stylesheet caching', 'amp' ),
+							'value'   => $this->get_css_transient_caching_sampling_range(),
+							'private' => false,
+						],
+						'amp_css_transient_caching_transient_count' => [
+							'label'   => esc_html__( 'Number of stylesheet transient cache entries', 'amp' ),
+							'value'   => MonitorCssTransientCaching::query_css_transient_count(),
+							'private' => false,
+						],
```
westonruter: Should not some of these be omitted if `wp_using_ext_object_cache()` and thus transients are not being used?

src/Admin/SiteHealth.php
```diff
@@ -207,33 +208,56 @@ public function icu_version() {
...
+				'color' => $is_disabled ? 'orange' : 'green',
+			],
+			'description' => esc_html__( 'On sites which have highly variable CSS and are not using a persistent object cache, the transient caching of stylesheets may be automatically disabled in order to prevent a site from filling up its wp_options table with too many transients.', 'amp' ),
+			'test'        => 'amp_css_transient_caching',
+		];
+
+		$status = $is_disabled ? 'recommended' : 'good';
+		$label  = $is_disabled
+			? __( 'Transient caching of stylesheets is disabled', 'amp' )
+			: __( 'Transient caching of stylesheets is not disabled or not being used', 'amp' );
```
westonruter: For consistency:

```suggestion
			? esc_html__( 'Transient caching of stylesheets is disabled', 'amp' )
			: esc_html__( 'Transient caching of stylesheets is not disabled or not being used', 'amp' );
```

schlessera: The escaping is down a few lines below.

src/Admin/SiteHealth.php
schlessera: I'd keep them around. If people have trouble with transient caching and switch to the object cache just to diagnose, we'd still want to have the values to help with debugging.

In general, you'd want to make as few assumptions as possible when collecting debugging information, as you'll make the same assumptions you've already done when coding up the bug in the first place.

src/BackgroundTask/CronBasedBackgroundTask.php
```diff
@@ -0,0 +1,114 @@
...
+	 *
+	 * This does nothing if the event is already scheduled.
+	 *
+	 * @return void
+	 */
+	public function schedule_event() {
+		if ( ! current_user_can( 'administrator' ) ) {
+			return;
+		}
+
```
westonruter: Do you think this is required? I think the event should be scheduled if any user logs in.

Also, this is using a role name in the capability check. It could be that a site doesn't have a traditional administrator role, and instead has fine-grained roles that have the capabilities (e.g. `manage_options`)

```suggestion
```

The important thing is `is_admin()` not is _administrator_.

src/BackgroundTask/CronBasedBackgroundTask.php
```diff
@@ -0,0 +1,114 @@
...
+	 * This should be hooked up to the WordPress deactivation hook.
+	 *
+	 * @param bool $network_wide Whether the deactivation was done network-wide.
+	 * @return void
+	 */
+	public function deactivate( $network_wide ) {
+		if ( $network_wide && is_multisite() && ! wp_is_large_network( 'sites' ) ) {
+			global $wpdb;
+
+			foreach ( $wpdb->get_col( "SELECT blog_id FROM {$wpdb->blogs}" ) as $blog_id ) {
```
westonruter: Would it be better to use `get_sites()` here?

```suggestion
			foreach ( get_sites( [ 'fields' => 'ids' ] ) as $blog_id ) {
```

src/BackgroundTask/CronBasedBackgroundTask.php
felixarntz: That was based on something @schlessera and I discussed. By only running this for administrators and not _any_ users we can reduce the chance for race conditions (there may be massive sites with so many logged in users that that could happen, but there's probably no site in the world with enough administrator users).

I personally have never used this approach, I'd argue without it in the majority of cases there  still won't be a race condition, and for sites where it would, they'd have much bigger problems with WordPress overall.

I'll defer to you guys for which approach to take, but I agree that _if_ this is kept it should use a capability, not a role name (`manage_options` makes sense).

src/BackgroundTask/CronBasedBackgroundTask.php
schlessera: We can use that, but we have to bump the `'number'` value as well then, it defaults to `100`.

schlessera: Looks like a `'number'` of `0` disables the limit. As we already check for `wp_is_large_network()`, I think that can be used.

schlessera:
I added a screenshot with the warning to the description up above.

westonruter:
Hummm. I'm seeing test failures/errors:

```
There was 1 error:

1) Test_Monitor_CSS_Transient_Caching::test_event_gets_scheduled_and_unscheduled
Error: Call to undefined method AmpProject\AmpWP\BackgroundTask\MonitorCssTransientCaching::activate()

/tmp/wordpress/src/wp-content/plugins/amp/tests/php/test-class-monitor-css-transient-caching.php:42

--

There were 2 failures:

1) Test_Monitor_CSS_Transient_Caching::test_event_can_be_processed
Failed asserting that false is not false.

/tmp/wordpress/tests/phpunit/includes/phpunit7/testcase.php:30
/tmp/wordpress/src/wp-content/plugins/amp/tests/php/test-class-monitor-css-transient-caching.php:66

2) Test_Monitor_CSS_Transient_Caching::test_transient_caching_is_disabled
false does not match expected type "array".
````

File: src/HasActivation.php [hunk 16]
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace AmpProject\AmpWP;
+
+/**
+ * An service that needs to run logic during plugin activation.
+ *
+ * @package AmpProject\AmpWP
+ */
+interface HasActivation {
```
westonruter: Note this interface is not currently used.

src/BackgroundTask/CronBasedBackgroundTask.php
```diff
@@ -0,0 +1,191 @@
...
+	const DEFAULT_INTERVAL_HOURLY      = 'hourly';
+	const DEFAULT_INTERVAL_TWICE_DAILY = 'twicedaily';
+	const DEFAULT_INTERVAL_DAILY       = 'daily';
+
+	/**
+	 * List of default interval names that are shipped with WordPress.
+	 *
+	 * @var string[]
+	 */
+	const DEFAULT_INTERVAL_NAMES = [
```
pierlon: This is currently unused.

src/BackgroundTask/CronBasedBackgroundTask.php
```diff
@@ -0,0 +1,191 @@
...
+	public function add_warning_sign_to_network_deactivate_action( $actions ) {
+		if ( ! wp_is_large_network() ) {
+			return $actions;
+		}
+
+		if ( ! array_key_exists( 'deactivate', $actions ) ) {
+			return $actions;
+		}
+
+		$actions['deactivate'] = preg_replace( '#(?=</a>)#i', ' ' . self::get_warning_icon(), $actions['deactivate'] );
```
pierlon: `get_warning_icon()` is being called statically although the function is not.

src/Admin/SiteHealth.php
```diff
@@ -334,6 +378,51 @@ public function get_serve_all_templates() {
...
+	 * @return string Threshold for the transient caching of stylesheets.
+	 */
+	private function get_css_transient_caching_threshold() {
+		/** This filter is documented in src/BackgroundTask/MonitorCssTransientCaching.php */
+		$threshold = (float) apply_filters(
+			'amp_css_transient_monitoring_threshold',
+			MonitorCssTransientCaching::DEFAULT_THRESHOLD
+		);
+
+		return "{$threshold} transients per day";
```
pierlon: Should this be translated?

src/Admin/SiteHealth.php
```diff
@@ -334,6 +378,51 @@ public function get_serve_all_templates() {
...
+	 * @return string Sampling range for the transient caching of stylesheets.
+	 */
+	private function get_css_transient_caching_sampling_range() {
+		/** This filter is documented in src/BackgroundTask/MonitorCssTransientCaching.php */
+		$sampling_range = (float) apply_filters(
+			'amp_css_transient_monitoring_sampling_range',
+			MonitorCssTransientCaching::DEFAULT_SAMPLING_RANGE
+		);
+
+		return "{$sampling_range} days";
```
pierlon: This as well, should it be translated?

src/HasActivation.php
pierlon: It's not being implemented, but it's used in `\AmpProject\AmpWP\Services::class`:

https://github.com/ampproject/amp-wp/blob/08506c2bc6ff0e8f67d158315e4a8e4767e7e3a0/src/Services.php#L58

src/BackgroundTask/CronBasedBackgroundTask.php
westonruter: Fixed in 9ad9b10f6/

src/Admin/SiteHealth.php
westonruter: I don't think so. It's part of the Site Health info export.

src/Admin/SiteHealth.php
westonruter: I don't think so. It's part of the Site Health info export.

src/BackgroundTask/CronBasedBackgroundTask.php
```diff
@@ -0,0 +1,191 @@
...
+	public function add_warning_to_plugin_meta( $plugin_meta, $plugin_file ) {
+		if ( ! is_multisite() || ! wp_is_large_network() ) {
+			return $plugin_meta;
+		}
+
+		if ( $plugin_file !== $this->plugin_file ) {
+			return $plugin_meta;
+		}
+
+		$plugin_meta[] = self::get_warning_icon() . ' ' . esc_html__( 'Large site detected. Deactivation will leave orphaned scheduled events behind.', 'amp' ) . ' ' . self::get_warning_icon();
```
pierlon:
```suggestion
		$plugin_meta[] = $this->get_warning_icon() . ' ' . esc_html__( 'Large site detected. Deactivation will leave orphaned scheduled events behind.', 'amp' ) . ' ' . $this->get_warning_icon();
```

src/Admin/SiteHealth.php
pierlon: Ah OK.

amp.php
```diff
@@ -14,13 +14,13 @@
  */
 
 define( 'AMP__FILE__', __FILE__ );
-define( 'AMP__DIR__', __DIR__ );
+define( 'AMP__DIR__', __DIR__ ); // phpcs:ignore PHPCompatibility.Keywords.NewKeywords.t_dirFound
```
westonruter: Maybe this should rather go back to being:

```suggestion
define( 'AMP__DIR__', dirname( __FILE__ ) );
```

This would prevent a PHP warning in PHP 5.2 from being raised. But that would be the only benefit to doing so. The plugin would be short-circuting before this constant is actually used to include a file.

schlessera: I'd probably avoid throwing a notice on PHP 5.2 in this particular case, but I don't feel strongly about it.

src/BackgroundTask/CronBasedBackgroundTask.php
schlessera: Ah, good catch. When the cron-handling was still a bit more complex, this was used for validation. However, I ripped all of that out, and just assume someone using the class is using it correctly for now.

So this one can go.

PR #4184, opened 2020-01-27
Redirect to non-AMP version when making non-singular AMP request; prevent using Reader mode template for Stories

westonruter:
## Summary

This PR fixes three issues with how Reader mode is served:

1. Reader mode only supports rendering AMP pages for the singular template. When attempting to access the AMP version for a non-singular template, the AMP plugin needs to redirect to the non-AMP URL. This was not the case in v1.4, where the `?amp` query parameter was just ignored. This was not good because if a site switched from Transitional mode to Reader mode, then the AMP URLs for non-singular templates would start serving back non-AMP pages. The AMP plugin should redirect to the non-AMP version to make it clear that an AMP page is not being served, either to inform search crawlers who may have indexed an AMP version or for users who are manually trying to access the AMP URL.
2. In 1.5-alpha after #3781, the above issue is more problematic because the lack of a redirect is resulting in a non-singular template being routed into the `single.php` Reader template. For example, if doing a search for “Sticky” with the URL`/?s=sticky` and you add the `amp` query var to go to `/?s=sticky&amp`, currently in `develop` the AMP plugin will try to render the `WP_Query` onto the `single.php` template, resulting in just the first post in the query being rendered. This is wrong! In this case we need to redirect to the non-AMP version in Reader mode when AMP is not available.
3. Lastly, when Reader mode was active, accessing a story permalink was resulting in the Reader template being used as opposed to the proper story template.

## Testing Instructions

Non-singular templates:

1. Enable Reader mode.
1. Try viewing the AMP version of a non-singular template, e.g. `/category/uncategorized/?amp`.
1. PASS: User is redirected to `/category/uncategorized/`; FAIL: Reader mode template renders the most recent post in the category.

Story template:

1. Enable the Stories experience and create a story.
1. View a single story.
1. PASS: Story renders as expected. FAIL: Reader mode template is unexpectedly used to render a story.

## Checklist

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -410,7 +410,12 @@ class AMP_Theme_Support {
 			add_filter( 'template_include', [ __CLASS__, 'serve_paired_browsing_experience' ] );
 		}
 
-		if ( ! is_amp_endpoint() ) {
+		if ( self::READER_MODE_SLUG === self::get_support_mode() && ! is_singular() && false !== get_query_var( amp_get_slug(), false ) ) {
+			// Reader mode only supports the singular template (for now) so redirect non-singular queries in reader mode to non-AMP version.
+			// A temporary redirect is used for admin users to allow them to see changes between reader mode and transitional modes.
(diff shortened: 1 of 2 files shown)


kienstra:
Hi @westonruter, sorry for the delay. I'll review this today.

csossi:
Verified in QA

PR #4193, opened 2020-01-29
Convert `head[profile]` attribute to `link[rel=profile]`

### LINKED ISSUE #4131: Automatically convert head[profile] to head > link[rel=profile]
## Feature description

When doing an analysis of themes on WordPress, I got a `DISALLOWED_ATTR` validation error for the `profile` attribute on the `head` element, such as code in :

```html
<head profile="http://gmpg.org/xfn/11">
```

The `profile` attribute on the `head` element is part of HTML4 but not HTML5. 

The W3 validator provides this advice:

image

According to [this SO answer](https://stackoverflow.com/a/14002644), the right way to represent this in HTML5 is:

```html
<head>
    <!-- … -->
    <link rel="profile" href="http://gmpg.org/xfn/11">
```

This should perhaps be done as part of the normalization done in the `Document` class (appending the `link` element to the end of the `head`). It could also be done in a separate sanitizer, though that may be overkill. Since the `Document` class is normalizing to HTML5, I think it makes sense to let the class handle the `profile` transformation.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

kienstra:
## Summary
Like Weston mentioned, `head[profile]` is invalid, so this converts it to a `link[rel=profile]`.

To test this, I made this change in Twenty Twenty:

```diff
diff --git a/header.php b/header.php
index d66d08f..99b02e0 100644
--- a/header.php
+++ b/header.php
@@ -13,13 +13,11 @@
 
 <html class="no-js" <?php language_attributes(); ?>>
 
-       <head>
+       <head profile="https://example-profile.com">
 
                <meta charset="<?php bloginfo( 'charset' ); ?>">
                <meta name="viewport" content="width=device-width, initial-scale=1.0" >
 
-               <link rel="profile" href="https://gmpg.org/xfn/11">
-
                <?php wp_head(); ?>
 
        </head>
```

...and with this PR, the `head[profile]` is stripped and moved to a new `head > link[rel=profile]`:

rel-profile

In the unlikely case that there's already a `head > link[rel=profile]`, this doesn't create a new one. It simply strips the `head[profile]` if it exists.

Fixes #4131

## Checklist

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -541,17 +541,22 @@ final class Document extends DOMDocument {
 	 * Converts a possible head[profile] attribute to link[rel=profile].
 	 *
 	 * The head[profile] attribute is only valid in HTML4, not HTML5.
-	 * So if it exists, add it to the <head> as a link[rel=profile] and strip the attribute.a -
+	 * So if it exists and isn't empty, add it to the <head> as a link[rel=profile] and strip the attribute.
 	 */
 	private function convert_head_profile_to_link() {
+		if ( ! $this->head->hasAttribute( 'profile' ) ) {
+			return;
+		}
+
 		$profile = $this->head->getAttribute( 'profile' );
 		if ( $profile ) {
 			$link = $this->createElement( 'link' );
 			$link->setAttribute( 'rel', 'profile' );
 			$link->setAttribute( 'href', $profile );
 			$this->head->appendChild( $link );
-			$this->head->removeAttribute( 'profile' );
 		}
+
+		$this->head->removeAttribute( 'profile' );
 	}
 
 	/**
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -359,6 +359,7 @@ final class Document extends DOMDocument {
 			$this->deduplicate_tag( self::TAG_HEAD );
 			$this->deduplicate_tag( self::TAG_BODY );
 			$this->move_invalid_head_nodes_to_body();
+			$this->convert_head_profile_to_link();
 		}
 
 		return $success;
@@ -533,6 +534,23 @@ final class Document extends DOMDocument {
 				$this->body->insertBefore( $this->head->removeChild( $node ), $this->body->firstChild );
 			}
 			$node = $next_sibling;
+		}
+	}
+
+	/**
+	 * Converts a possible head[profile] attribute to link[rel=profile].
+	 *
+	 * The head[profile] attribute is only valid in HTML4, not HTML5.
+	 * So if it exists, add it to the <head> as a link[rel=profile] and strip the attribute.
+	 */
+	private function convert_head_profile_to_link() {
+		$profile = $this->head->getAttribute( 'profile' );
+		if ( $profile ) {
+			$link = $this->createElement( 'link' );
+			$link->setAttribute( 'rel', 'profile' );
+			$link->setAttribute( 'href', $profile );
+			$this->head->appendChild( $link );
+			$this->head->removeAttribute( 'profile' );
 		}
 	}
 
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -541,17 +541,29 @@ final class Document extends DOMDocument {
 	 * Converts a possible head[profile] attribute to link[rel=profile].
 	 *
 	 * The head[profile] attribute is only valid in HTML4, not HTML5.
-	 * So if it exists, add it to the <head> as a link[rel=profile] and strip the attribute.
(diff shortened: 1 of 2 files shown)


kienstra:
@todo: consider a case where there is both a `head[profile]` and a `link[profile]`. For example, [Twenty Twenty has](https://github.com/WordPress/twentytwenty/blob/6d0a5240af108d02f58ec797fd49b32458a4c698/header.php#L21) a `link[profile]`, though it doesn't have a `head[profile]`.

src/Dom/Document.php
```diff
@@ -536,6 +537,35 @@ private function move_invalid_head_nodes_to_body() {
...
 
+	/**
+	 * Converts a possible head[profile] attribute to link[rel=profile].
+	 *
+	 * The head[profile] attribute is only valid in HTML4, not HTML5.
+	 * So if it exists and there's no link[rel=profile], add it to the <head> as a link[rel=profile].
+	 */
+	private function convert_head_profile_to_link() {
+		$profile = $this->head->getAttribute( 'profile' );
+		if ( ! $profile ) {
```
westonruter: This should handle the case of someone putting `profile=""` perhaps, so:

```suggestion
		if ( ! $this->head->hasAttribute( 'profile' ) ) {
```

src/Dom/Document.php
```diff
@@ -536,6 +537,35 @@ private function move_invalid_head_nodes_to_body() {
...
+		}
+
+		// If there's already a head > link[rel=profile], don't create another.
+		$links = $this->head->getElementsByTagName( 'link' );
+		foreach ( $links as $link ) {
+			if ( 'profile' === $link->getAttribute( 'rel' ) ) {
+				$this->head->removeAttribute( 'profile' );
+				return;
+			}
+		}
```
westonruter: I don't think this is necessary. It's not a problem to have another, is it?

kienstra: Yeah, good point. According to the [w3.org validator](https://validator.w3.org/), it's still valid HTML if there are 2 `link[rel=profile]` tags:

valid-html

kienstra: Reverted in https://github.com/ampproject/amp-wp/pull/4193/commits/2294c6e9412683aba413ca9e1286776add32b9e6

src/Dom/Document.php
```diff
@@ -536,6 +537,23 @@ private function move_invalid_head_nodes_to_body() {
...
+	 */
+	private function convert_head_profile_to_link() {
+		$profile = $this->head->getAttribute( 'profile' );
+		if ( $profile ) {
+			$link = $this->createElement( 'link' );
+			$link->setAttribute( 'rel', 'profile' );
+			$link->setAttribute( 'href', $profile );
+			$this->head->appendChild( $link );
+			$this->head->removeAttribute( 'profile' );
+		}
```
westonruter: Only thing left is to make sure that a falsy `profile` attribute is removed, so like:

```php
if ( ! $this->head->hasAttribute( 'profile' ) ) {
	return;
}

$profile = $this->head->getAttribute( 'profile' );
if ( $profile ) {
	$link = $this->createElement( 'link' );
	$link->setAttribute( 'rel', 'profile' );
	$link->setAttribute( 'href', $profile );
	$this->head->appendChild( $link );
}

$this->head->removeAttribute( 'profile' );
```

src/Dom/Document.php
kienstra: If `<head profile="">`, do you think we should strip the `profile` but not create a `<link re="profile">` ?

kienstra: Sorry, you just answered that in https://github.com/ampproject/amp-wp/pull/4193#discussion_r372636143

src/Dom/Document.php
kienstra: Sounds good, that's applied in https://github.com/ampproject/amp-wp/pull/4193/commits/6ebae07a5a2f804b73eb465120e41f06259ec2bd

src/Dom/Document.php
kienstra: Applied in https://github.com/ampproject/amp-wp/pull/4193/commits/6ebae07a5a2f804b73eb465120e41f06259ec2bd

westonruter:
> consider a case where there is both a `head[profile]` and a `link[profile]`. For example, [Twenty Twenty has](https://github.com/WordPress/twentytwenty/blob/6d0a5240af108d02f58ec797fd49b32458a4c698/header.php#L21) a `link[profile]`, though it doesn't have a `head[profile]`.

Nice. I didn't notice that Twenty Twenty had this. Added validation that the normalization here is the right approach.

PR #4194, opened 2020-01-29
Conditionally remove html[xmlns], convert html[xml:lang] to html[lang]

### LINKED ISSUE #4132: Normalize html[xmlns] and html[xml:lang] attributes
## Feature description

When doing an analysis of WordPress themes, I got validation errors being reported for these attributes:

* `html@xml:lang`
* `html@xmlns`

These are included in many WordPress themes because WordPress to strive to be valid XHTML. WordPress has since moved to HTML5, and HTML5 [specifies](https://html.spec.whatwg.org/multipage/infrastructure.html#xml) that the `http://www.w3.org/1999/xhtml` namespace is implied.

Since the `Document` class normalizes HTML documents as HTML5 with the HTML5 doctype, I think it makes sense to also have `Document::loadHTML()` automatically normalize these attributes.

We should consider normalizing `xml:lang` to just `lang`, or removing `xml:lang` if the `lang` attribute is already there. The `xmlns` attribute here is redundant and unnecessary; it can simply be removed.

So given:

```html
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
```

Should be normalized to just:

```html
<html lang="en">
```

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

kienstra:
## Summary
* If http://www.w3.org/1999/xhtml === `html[xmlns]`, this strips the attribute
* If theres an `html[xml:lang]` but not `html[lang]`, it moves the `html[xml:lang]` value to `html[lang]`
* This strips the `html[xml:lang]` attribute if it exists, no matter its value

For example, with this change to Twenty Twenty:
```diff
diff --git a/header.php b/header.php
index d66d08f..f65eb8d 100644
--- a/header.php
+++ b/header.php
@@ -11,7 +11,7 @@
 
 ?><!DOCTYPE html>
 
-<html class="no-js" <?php language_attributes(); ?>>
+<html class="no-js" xmlns="http://www.w3.org/1999/xhtml" xml:lang="es">
```

...the `html[xmlns]` is stripped, and the `html[xml:lang]` is moved to `html[lang]`:

xml-lang-converted

Fixes #4132 

## Checklist

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -1029,7 +1029,7 @@ final class Document extends DOMDocument {
 	 * Conditionally removes html[xmlns], and converts html[xml:lang] to html[lang].
 	 */
 	private function normalize_html_attributes() {
-		$html = $this->getElementsByTagName( 'html' )->item( 0 );
+		$html = $this->documentElement;
 		if ( ! $html->hasAttributes()  ) {
 			return;
 		}
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -1030,7 +1030,7 @@ final class Document extends DOMDocument {
 	 */
 	private function normalize_html_attributes() {
 		$html = $this->documentElement;
-		if ( ! $html->hasAttributes()  ) {
+		if ( ! $html->hasAttributes() ) {
 			return;
 		}
 
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -1034,9 +1034,8 @@ final class Document extends DOMDocument {
 			return;
 		}
 
-		$xmlns                = $html->attributes->getNamedItem( 'xmlns' );
-		$xmlns_value_to_strip = 'http://www.w3.org/1999/xhtml';
-		if ( $xmlns && $xmlns_value_to_strip === $xmlns->nodeValue ) {
+		$xmlns = $html->attributes->getNamedItem( 'xmlns' );
+		if ( $xmlns && 'http://www.w3.org/1999/xhtml' === $xmlns->nodeValue ) {
 			$html->removeAttributeNode( $xmlns );
 		}
 
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -340,6 +340,8 @@ final class Document extends DOMDocument {
 		libxml_use_internal_errors( $libxml_previous_state );
 
 		if ( $success ) {
+			$this->normalize_html_attributes();
+
 			// Remove http-equiv charset again.
 			$meta = $this->head->firstChild;
 			if (
@@ -1019,6 +1021,33 @@ final class Document extends DOMDocument {
 		}
 
 		return preg_replace( self::HTML_RESTORE_DOCTYPE_PATTERN, '\1!\3\4>', $html, 1 );
+	}
+
+	/**
+	 * Normalizes HTML attributes to be HTML5 compatible.
+	 *
+	 * Conditionally removes html[xmlns], and converts html[xml:lang] to html[lang].
+	 */
+	private function normalize_html_attributes() {
+		$html = $this->getElementsByTagName( 'html' )->item( 0 );
+		if ( ! $html->attributes ) {
+			return;
+		}
+
+		$xmlns                = $html->attributes->getNamedItem( 'xmlns' );
+		$xmlns_value_to_strip = 'http://www.w3.org/1999/xhtml';
+		if ( $xmlns && $xmlns_value_to_strip === $xmlns->nodeValue ) {
+			$html->removeAttributeNode( $xmlns );
+		}
+
+		$xml_lang = $html->attributes->getNamedItem( 'xml:lang' );
+		if ( $xml_lang ) {
+			if ( ! $html->attributes->getNamedItem( 'lang' ) && $xml_lang->nodeValue ) {
+				// Move the html[xml:lang] value to html[lang].
+				$html->setAttribute( 'lang', $xml_lang->nodeValue );
+			}
+			$html->removeAttributeNode( $xml_lang );
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/Dom/Document.php
```diff
@@ -1021,6 +1023,33 @@ private function restore_doctype_node( $html ) {
...
+	 *
+	 * Conditionally removes html[xmlns], and converts html[xml:lang] to html[lang].
+	 */
+	private function normalize_html_attributes() {
+		$html = $this->getElementsByTagName( 'html' )->item( 0 );
+		if ( ! $html->attributes ) {
+			return;
+		}
+
+		$xmlns                = $html->attributes->getNamedItem( 'xmlns' );
```
kienstra: It'd probably be better to do something like `$xmlns = $html->getAttribute( 'xmlns' )`.

But for some reason, `$html->getAttribute( 'xmlns' )` returned `""`, even when `$html->attributes->getNamedItem( 'xmlns' )` returned a `DOMAttr`.

src/Dom/Document.php
```diff
@@ -1021,6 +1023,33 @@ private function restore_doctype_node( $html ) {
...
+		$xmlns_value_to_strip = 'http://www.w3.org/1999/xhtml';
+		if ( $xmlns && $xmlns_value_to_strip === $xmlns->nodeValue ) {
+			$html->removeAttributeNode( $xmlns );
+		}
+
+		$xml_lang = $html->attributes->getNamedItem( 'xml:lang' );
+		if ( $xml_lang ) {
+			if ( ! $html->attributes->getNamedItem( 'lang' ) && $xml_lang->nodeValue ) {
+				// Move the html[xml:lang] value to html[lang].
+				$html->setAttribute( 'lang', $xml_lang->nodeValue );
```
kienstra: @todo: handle case of `lang=""`.

If `<head xml:lang="es" lang="">`, this will produce `<head lang="">`. 

Currently, it won't copy the `xml:lang` value to `lang` if `lang=""`.

kienstra: Applied in https://github.com/ampproject/amp-wp/pull/4194/commits/93d96f4a9e92e021597ecae524bf9d476f84f552

src/Dom/Document.php
```diff
@@ -1021,6 +1023,34 @@ private function restore_doctype_node( $html ) {
 		return preg_replace( self::HTML_RESTORE_DOCTYPE_PATTERN, '\1!\3\4>', $html, 1 );
 	}
 
+	/**
+	 * Normalizes HTML attributes to be HTML5 compatible.
+	 *
+	 * Conditionally removes html[xmlns], and converts html[xml:lang] to html[lang].
+	 */
+	private function normalize_html_attributes() {
+		$html = $this->getElementsByTagName( 'html' )->item( 0 );
```
westonruter:
```suggestion
		$html = $this->documentElement;
```

src/Dom/Document.php
```diff
@@ -1021,6 +1023,34 @@ private function restore_doctype_node( $html ) {
...
+		if ( ! $html->hasAttributes()  ) {
+			return;
+		}
+
+		$xmlns                = $html->attributes->getNamedItem( 'xmlns' );
+		$xmlns_value_to_strip = 'http://www.w3.org/1999/xhtml';
+		if ( $xmlns && $xmlns_value_to_strip === $xmlns->nodeValue ) {
+			$html->removeAttributeNode( $xmlns );
+		}
```
westonruter: Since the variable is only used once, might as well remove `$xmlns_value_to_strip` and just use the string literal:

```php
		if ( $xmlns && 'http://www.w3.org/1999/xhtml' === $xmlns->nodeValue ) {
```

src/Dom/Document.php
```diff
@@ -1021,6 +1023,34 @@ private function restore_doctype_node( $html ) {
...
+
+		$xml_lang = $html->attributes->getNamedItem( 'xml:lang' );
```
westonruter: Have you considered [`hasAttributeNS`](https://www.php.net/manual/en/domelement.hasattributens.php) or [`getAttributeNS`](https://www.php.net/manual/en/domelement.getattributens.php)? 

It seems these don't work either. I would have thought at least that this would work: 

```php
$dom->documentElement->getAttributeNodeNS( 'http://www.w3.org/XML/1998/namespace', 'lang' );
```

But it doesn't. Perhaps it's due to `loadHTML` and that the document is not XML in the first place.

Using `$dom->documentElement->attributes->getNamedItem( 'xml:lang' )` works in all PHP versions 5+. So seems like a good one to use.

src/Dom/Document.php
westonruter: Yeah, I'm not sure. I can see that `$html->attributes->getNamedItem( 'xmlns' )` works across all PHP versions. I don't have a better option.

src/Dom/Document.php
kienstra: Nice, thanks for looking at this.

src/Dom/Document.php
kienstra: Yeah, that probably makes sense to remove the variable declaration.

src/Dom/Document.php
kienstra: Ah, good point. I didn't think about that.

kienstra:
I'l make the remaining changes soon, maybe in 30 minutes to an hour.

src/Dom/Document.php
kienstra: Applied in https://github.com/ampproject/amp-wp/pull/4194/commits/096cf7b880794ac352e9b0235a1eeb46a802f144

src/Dom/Document.php
kienstra: Committed in https://github.com/ampproject/amp-wp/pull/4194/commits/7aaf8fd016967e130c9564bdeed9f0c9aee944f7

kienstra:
The outstanding feedback should be addressed now, unless I misunderstood something.

westonruter:
Changes look good. Note that it is difficult to test this because without #4190 the validation errors do not show up in the Validated URL screen.

kienstra:
Ah, good point. Thanks for reviewing this.

Issue #2169, opened 2019-04-22
Suggestion: display CSS sizes in Compatibility Tool (move manifest)


kienstra:
@billerickson had a good suggestion, that it would be useful to display the CSS sizes in the Compatibility Tool. 

It would help developers see how close they are to the limit.

This could include the total CSS size for all files, the size of each file or `<style>` element, and the source of each.

It might be similar to the comment printed above the `<style amp-custom>`. For example, in https://make.xwp.co/?amp:

```html
<!--
The style[amp-custom] element is populated with:
     0 B: style[amp-custom=]
   198 B: link#twentyfifteen-jetpack-css[rel=stylesheet][id=twentyfifteen-jetpack-css][href=https://make.xwp.co/wp-content/plugins/jetpack/modules/theme-tools/compat/twentyfifteen.css?ver=6.5][type=text/css][media=all]
    86 B: link#jetpack_related-posts-css[rel=stylesheet][id=jetpack_related-posts-css][href=https://make.xwp.co/wp-content/plugins/jetpack/modules/related-posts/related-posts.css?ver=20150408][type=text/css][media=all]
     0 B: style#jetpack_related-posts-inline-css[id=jetpack_related-posts-inline-css][type=text/css]
   403 B: link#dashicons-css[rel=stylesheet][id=dashicons-css][href=https://make.xwp.co/wp-includes/css/dashicons.min.css?ver=4.9.8][type=text/css][media=all]
    90 B: link#jetpack_likes-css[rel=stylesheet][id=jetpack_likes-css][href=https://make.xwp.co/wp-content/plugins/jetpack/modules/likes/style.css?ver=6.5][type=text/css][media=all]
   942 B: link#genericons-css[rel=stylesheet][id=genericons-css][href=https://make.xwp.co/wp-content/plugins/jetpack/_inc/genericons/genericons/genericons.css?ver=3.1][type=text/css][media=all]
 26293 B: link#twentyfifteen-style-css[rel=stylesheet][id=twentyfifteen-style-css][href=https://make.xwp.co/wp-content/themes/twentyfifteen/style.css?ver=1.9][type=text/css][media=all]
   109 B: style#twentyfifteen-style-inline-css[id=twentyfifteen-style-inline-css][type=text/css]
[...]

Issue #2044, opened 2019-04-01
Extend classic/reader templates with commonly-requested features


westonruter:
Now that we plan to rebrand the Classic mode and maintain it as a “Reader” mode (#2034) we should address common requests by users for the classic post templates.

The most common request I believe is to add a nav menu (#1772, #792, #615, #220, #435). Note that if this involves adding a nav menu location, this could be re-used for navigation in AMP Stories. As with AMP Stories, the nav menu should be presented in an `amp-sidebar`.

What are the other most-commonly requested template features that make sense for us to incorporate into the plugin? One request is a [search bar](https://wordpress.org/support/topic/nothing-except-converting-to-amp/#post-11375989).

swissspidy:
Some users also requested filtering all internal links so that they stay in reader mode while navigating.

westonruter:
For that, there is also https://github.com/ampproject/amp-wp/issues/1389

westonruter:
Here is a basic plugin which can be used to add a nav menu sidebar on Reader mode templates: https://gist.github.com/westonruter/b22cfae13f0862d5b9716c776c60f882

mrjuri:
Is a good idea make two location menu?
1. location for AMP page;
2. location for AMP stories.

So in AMP stories, the menu will be active or not.

swissspidy:
@mrjuri Those would be separate things either way eventually, see #2639.

jeannyhaliman:
Hi Weston,

I heard similar feedbacks of publishers requesting sidebar & search bar. Beside that common asks I've heard:

1. Feature to upload publisher logo on top of the page. (upon click - redirect to home)
2. Flexibility of changing the colour of the top bar (currently only available in blue)

Pubs also ask for share to social media, but I usually recommend them to check supported plugins in amp-wp.org/ecosystem

Hope this helps!

Sincerely,
Jeanny

westonruter:
@jeannyhaliman 

> Feature to upload publisher logo on top of the page. (upon click - redirect to home)

Currently the site icon is shown in the upper-right corner of the header.

![image](https://user-images.githubusercontent.com/134745/61751528-4b216b80-ad6e-11e9-8fd2-3d888dd5004d.png)

But instead of this they want to have a bigger logo appear instead of the site title in the top-left (here “WordPress Develop”)? Should the site icon then be hidden when that is done?

> Flexibility of changing the colour of the top bar (currently only available in blue)

Actually, you can change the color of the header currently:

[...]

PR #4205, opened 2020-01-30
Split Reader mode style template part stylesheet from action-supplied styles

westonruter:
### Splitting up stylesheets in Reader mode

The Reader mode templates have for a long time output all styles in a single `<style amp-custom>` tag. This was needed because the `AMP_Style_Sanitizer` was originally not able to handle the parsing of stylesheets but only `style` attributes. Now that the Reader mode templates are passed through the post-processor (via #2202) all `style` elements will be combined into a single `style[amp-custom]` element automatically. It is advantageous to split styles into multiple stylesheets because when a stylesheet goes above the limit of 50KB, it will get excluded. If a plugin adds a lot of styles via the `amp_post_template_head` action, the result can be that all of the styles in the `style` template part would also get excluded. 

By splitting the two sets of styles into separate stylesheets, only the styles in `amp_post_template_head` will be excluded if they are excessive. This is better because those styles are less important than those being included in the `style` template part, which control the overall layout and design of the template. Otherwise, there is no difference in how the styles are included in the page.

The inclusion of the class names on the `style` elements is purely for the purpose of debugging, when/if Reader mode templates get shown as Validated URLs in the admin, with the new stylesheet analysis (see #2169).

Here's some example code that causes too much CSS to be added in `amp_post_template_css`:

```php
add_action( 'amp_post_template_css', function() {
	printf( 'body:after{ content: "%s"; }', str_repeat( 'a', 50000 ) );
} );
```

See also #2044.

### Preventing invalid `style[amp-custom]` from passing through

I also noticed that in the `develop` branch this code actually does not get sanitized at all. In fact, in Reader mode more than 50KB of CSS are getting added to the page:

image

The reason is that when there is only one single single `style` on the page and this stylesheet is greater than 50KB, then this condition is not entered since there are no included stylesheets:

https://github.com/ampproject/amp-wp/blob/0d3ce780414af6306c59dfd132967a8907ae0412/includes/sanitizers/class-amp-style-sanitizer.php#L2625-L2626

When that condition is not entered, then the previously-captured `amp_custom_style_element` element is not subsequently emptied out to populate with the processed CSS (since there is none):

https://github.com/ampproject/amp-wp/blob/0d3ce780414af6306c59dfd132967a8907ae0412/includes/sanitizers/class-amp-style-sanitizer.php#L2644-L2653

So the fix is simply to discontinue capturing any original `style[amp-custom]` in favor of just creating a fresh empty new one to add to the document only if there are valid stylesheets to include.

## Checklist

### DIFF
--- a/templates/html-start.php
+++ b/templates/html-start.php
@@ -24,6 +24,10 @@
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width,initial-scale=1,minimum-scale=1">
 	<?php
+	// Note: The following two <style> tags are combined into <style amp-custom> via AMP_Style_Sanitizer.
+	// Splitting up the styles into two stylesheets allows for plugin-supplied styles via the amp_post_template_css action
+	// to be excluded from the styles in the style template part, which are more important given they style the overall page.
+
 	/**
 	 * Fires when rendering <head> in Reader mode templates.
 	 *
@@ -33,8 +37,10 @@
 	 */
 	do_action( 'amp_post_template_head', $this );
 	?>
-	<style amp-custom>
+	<style class="style-template-part">
 		<?php $this->load_parts( [ 'style' ] ); ?>
+	</style>
+	<style class="amp-post-template-css-action">
 		<?php
 		/**
 		 * Fires when printing CSS styles in Reader mode templates.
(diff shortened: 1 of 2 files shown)


templates/html-start.php
```diff
@@ -33,8 +37,10 @@
 	 */
 	do_action( 'amp_post_template_head', $this );
 	?>
-	<style amp-custom>
+	<style class="style-template-part">
```
kienstra: This makes sense, given that the document is passed through the sanitizers.

PR #4217, opened 2020-02-03
Limit Stories experience to WP 5.3 & Gutenberg 7.1.0

### LINKED ISSUE #4200: Disable AMP Stories experience when recent Gutenberg is active
## Feature description

As discussed in #4074 and [reported](https://wordpress.org/support/topic/new-version-of-gutenberg-breaks-stories-layout/) in the support forum, recent versions of Gutenberg break the AMP Stories editor. 

We need to disable the AMP Stories experience if either the user is on WordPress version under 5.3 and/or they have a Gutenberg version that is too new.

Note that if the user does _not_ have any stories created, then the experience would be hidden entirely. See #4201. So this only is needed for users who have existing `amp_story` posts.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

Disables Stories experience if Gutenberg > v7.1.0 or WordPress > v5.3. The Stories editor is broken in Gutenberg v7.2.0, and Wordpress 5.4 ( which will include an incompatible version of the block editor).

Fixes #4200.

## Checklist

### DIFF
--- a/.dev-lib
+++ b/.dev-lib
@@ -10,7 +10,7 @@ CHECK_SCOPE=all
 function after_wp_install {
 	if [[ "$WP_VERSION" != "4.9" ]]; then
 		echo -n "Installing Gutenberg 7.1.0..."
-		gutenberg_plugin_svn_url=http://plugins.svn.wordpress.org/gutenberg/tags/7.1.0/
+		gutenberg_plugin_svn_url=https://plugins.svn.wordpress.org/gutenberg/tags/7.1.0/
 		svn export -q "$gutenberg_plugin_svn_url" "$WP_CORE_DIR/src/wp-content/plugins/gutenberg"
 		echo "done"
 	fi
--- a/.dev-lib
+++ b/.dev-lib
@@ -9,8 +9,8 @@ CHECK_SCOPE=all
 
 function after_wp_install {
 	if [[ "$WP_VERSION" != "4.9" ]]; then
-		echo -n "Installing Gutenberg..."
-		gutenberg_plugin_svn_url=https://plugins.svn.wordpress.org/gutenberg/trunk/
+		echo -n "Installing Gutenberg 7.1.0..."
+		gutenberg_plugin_svn_url=http://plugins.svn.wordpress.org/gutenberg/tags/7.1.0/
 		svn export -q "$gutenberg_plugin_svn_url" "$WP_CORE_DIR/src/wp-content/plugins/gutenberg"
 		echo "done"
 	fi
--- a/includes/class-amp-story-post-type.php
+++ b/includes/class-amp-story-post-type.php
@@ -104,7 +104,7 @@ class AMP_Story_Post_Type {
 	 */
 	public static function has_required_block_capabilities() {
 		if ( defined( 'GUTENBERG_DEVELOPMENT_MODE' ) ) {
-			return (bool) GUTENBERG_DEVELOPMENT_MODE;
+			return false;
 		}
 
 		// If Gutenberg is installed only versions 6.6.0 - 7.1.0 are supported.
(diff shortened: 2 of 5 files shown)


includes/class-amp-story-post-type.php
```diff
@@ -32,7 +32,14 @@ class AMP_Story_Post_Type {
 	 *
 	 * @var string
 	 */
-	const REQUIRED_GUTENBERG_VERSION = '6.6';
+	const MIN_REQUIRED_GUTENBERG_VERSION = '6.6.0';
+
+	/**
+	 * Minimum required version of Gutenberg required.
```
swissspidy:
```suggestion
	 * Maximum allowed version of Gutenberg.
```

includes/class-amp-story-post-type.php
```diff
@@ -93,20 +100,22 @@ class AMP_Story_Post_Type {
...
 	public static function has_required_block_capabilities() {
-		return (
-			( defined( 'GUTENBERG_DEVELOPMENT_MODE' ) && GUTENBERG_DEVELOPMENT_MODE )
-			||
-			( defined( 'GUTENBERG_VERSION' ) && version_compare( GUTENBERG_VERSION, self::REQUIRED_GUTENBERG_VERSION, '>=' ) )
-			||
-			version_compare( get_bloginfo( 'version' ), '5.3-RC2', '>=' )
-		);
+		if ( defined( 'GUTENBERG_DEVELOPMENT_MODE' ) ) {
+			return (bool) GUTENBERG_DEVELOPMENT_MODE;
```
swissspidy: Wouldn't that indicate that the editor works with current Gutenberg master?

I think if `GUTENBERG_DEVELOPMENT_MODE` is defined we should return `false` as it's most likely the current master version.

pierlon: Yes good call.

Would it be better to get the version via `get_plugin_data()`? The likelihood of someone having an outdated Gutenberg with `GUTENBERG_DEVELOPMENT_MODE` set would be low I assume, but I'm not entirely on board with returning `false` if that constant is set.

westonruter: Let's go ahead and return `false` if `GUTENBERG_DEVELOPMENT_MODE` is on. The reality is that this is not a widely-used feature yet and we'll be removing it in the next version of the AMP plugin. The liklihood that someone is using an old version of Gutenberg _with_  `GUTENBERG_DEVELOPMENT_MODE` on _and_ having Stories active seems very rare. If `GUTENBERG_DEVELOPMENT_MODE` is `true`, then a site just has to deactivate Gutenberg to access Stories.

pierlon: Done in https://github.com/ampproject/amp-wp/pull/4217/commits/92e7c3481444a40801b439e94c874886f145afa1.

.dev-lib
```diff
@@ -9,8 +9,8 @@ CHECK_SCOPE=all
 
 function after_wp_install {
 	if [[ "$WP_VERSION" != "4.9" ]]; then
-		echo -n "Installing Gutenberg..."
-		gutenberg_plugin_svn_url=https://plugins.svn.wordpress.org/gutenberg/trunk/
+		echo -n "Installing Gutenberg 7.1.0..."
+		gutenberg_plugin_svn_url=http://plugins.svn.wordpress.org/gutenberg/tags/7.1.0/
```
westonruter: Oh, why the change from `https` to `http`?

westonruter: Revised in 983efd4a064fb54acb3a46a8ed5b2f7e5356b6f8

westonruter:
Once the build passes I'll squash merge this and cherry-pick into the `1.4` branch.

PR #4212, opened 2020-02-03
Improve handling of form submissions that redirect and implement form submission proxy

westonruter:
## Summary

* Add support for automatically redirecting users when WordPress uses `header('Location: …')` in the same way as before `wp_redirect(…)` was able to automatically supply the `AMP-Redirect-To` response header. This should eliminate the need for [this shim code](https://gist.github.com/westonruter/b2554073461707a88057bb86b6b603d6#file-amp-gravity-form-submission-shim-php-L12-L29) in Gravity Forms. The corresponding [playbook information](https://amp-wp.org/documentation/playbooks/handling-form-submissions/) will also be obsolete with this PR.
* Add support for submitting forms to external domains by means of a new REST API form submission proxy. 

Fixes #4191.

Iterates on #2425. Revisits #911. See also #923.

## Testing

1. First of all, fire up [this Glitch to be able to test remote form submissions](https://glitch.com/~brook-paranthodon-k6ttvyw9a).
2. Then install the [Test POST Form](https://gist.github.com/westonruter/6c0eb253972ccdfc31f841629e2f4344) plugin.
3. Add two shortcodes to a post: `[test_post_form]` and `[test_post_form action="https://brook-paranthodon-k6ttvyw9a.glitch.me/"]`. For example:
```html
<!-- wp:paragraph -->
<p><strong>Internal action:</strong></p>
<!-- /wp:paragraph -->

<!-- wp:shortcode -->
[test_post_form]
<!-- /wp:shortcode -->

<!-- wp:paragraph -->
<p><strong>External action:</strong></p>
<!-- /wp:paragraph -->

<!-- wp:shortcode -->
[test_post_form action="https://brook-paranthodon-k6ttvyw9a.glitch.me/?foo=bar"]
<!-- /wp:shortcode -->
```
4. Check the behavior of submitting the internal action and external action forms when on an AMP page.

Any submission request type that has a JSON response (including the use of `wp_die()`) should have a response looking like:

image

or 

image

If the form handler does _not_ respond with something that can be represented as JSON (e.g. returning with an HTML page) then the expected output should look like:

image

or 

image

Note that the underlying HTTP status code and status text should be displayed in this message.

Additionally, if a form submission results in a redirect, either via `wp_redirect()` or using an underlying `header('Location: …')` call, then the user should be redirected after seeing a “Redirecting…” message momentarily:

image

Similar behavior should work both for when the form submission is handled internally by the present WordPress or if the form submission is pointing to another domain.

## Checklist

[...]

PR #4250, opened 2020-02-06
Fix malformed conversion of relative action URLs for forms

### LINKED ISSUE #4208: Malformed conversion of forms with relative action URLs
## Bug Description

When a `form` element has a relative URL (such as `../`) or a query (such as `?foo=bar`) for the `action` attribute, the value is incorrectly sanitized and a malformed URL is produced.

For example, given that the request URL is `https://example.org/about/?amp` with Transitional mode active, the following can be observed:

`action` | Sanitized `action` | Expected `action`
---|---|---
`../`|`//example.org/about/?amp/__/&_wp_amp_action_xhr_converted=1&amp`|`//example.org/?amp&_wp_amp_action_xhr_converted=1`
`?foo=bar`|`//example.org/about?amp?foo=bar&_wp_amp_action_xhr_converted=1&amp`|`//example.org/about/?amp&foo=bar&_wp_amp_action_xhr_converted=1`

This is also reproducible in Standard mode.

Do note that actions with relative anchor tags (eg. `#foo`) are correctly converted, which was fixed by #4003.

## Steps to reproduce

1. Activate Transitional mode

2. Create a post and add a Custom HTML block with the following content:

```html
<form method="post" action="../"></form> 
<form method="post" action="?foo=bar"></form> 
```

3. View the AMP version of the post

4. Given that the current URL is `https://example.org/about/?amp`, the following malformed `action-xhr` URL values for each `form` element can be observed:

```html
<form method="post" action-xhr="//example.com/about/?amp/__/&_wp_amp_action_xhr_converted=1&amp">...</form>

<form method="post" action-xhr="//example.com/about/?amp?foo=bar&_wp_amp_action_xhr_converted=1&amp">...</form>
```

## Additional context

 - WordPress version:
 - Plugin version:
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

Fixes #4208.

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-form-sanitizer.php
+++ b/includes/sanitizers/class-amp-form-sanitizer.php
@@ -58,60 +58,8 @@ class AMP_Form_Sanitizer extends AMP_Base_Sanitizer {
 				$node->setAttribute( 'method', $method );
 			}
 
-			/*
-			 * In HTML, the default action is just the current URL that the page is served from.
-			 * The action "specifies a server endpoint to handle the form input. The value must be an
-			 * https URL and must not be a link to a CDN".
-			 */
-			if ( ! $node->getAttribute( 'action' ) ) {
-				$action_url = esc_url_raw( '//' . $_SERVER['HTTP_HOST'] . wp_unslash( $_SERVER['REQUEST_URI'] ) );
-			} else {
-				$action_url = $node->getAttribute( 'action' );
-				$parsed_url = wp_parse_url( $action_url );
+			$action_url = $this->get_action_url( $node->getAttribute( 'action' ) );
 
-				// Ignore a malformed URL - it will be later sanitized.
-				if ( false !== $parsed_url ) {
-					// Handle relative URLs.
-					if ( ! isset( $parsed_url['scheme'] ) || 'https' !== $parsed_url['scheme'] ) {
-						$parsed_url['scheme'] = '//';
-
-						if ( ! isset( $parsed_url['host'] ) ) {
-							$parsed_url['host'] = $_SERVER['HTTP_HOST'];
-						}
-
-						if ( ! isset( $parsed_url['path'] ) ) {
-							// If there is action URL path, use the one from the request.
-							$parsed_url['path'] = trailingslashit( wp_unslash( $_SERVER['REQUEST_URI'] ) );
-						} elseif ( '' !== $parsed_url['path'] && '/' !== $parsed_url['path'][0] ) {
-							// If the path is relative, append it to the current request path.
-							$parsed_url['path'] = trailingslashit( wp_unslash( $_SERVER['REQUEST_URI'] ) ) . trailingslashit( $parsed_url['path'] );
-						}
-
-						// Rebuild the URL.
-						$action_url = $parsed_url['scheme'];
-						if ( isset( $parsed_url['user'] ) ) {
-							$action_url .= $parsed_url['user'];
-							if ( isset( $parsed_url['pass'] ) ) {
-								$action_url .= ':' . $parsed_url['pass'];
-							}
-							$action_url .= '@';
-						}
-						$action_url .= $parsed_url['host'];
-						if ( isset( $parsed_url['port'] ) ) {
-							$action_url .= ':' . $parsed_url['port'];
-						}
-						$action_url .= $parsed_url['path'];
-						if ( isset( $parsed_url['query'] ) ) {
-							$action_url .= '?' . $parsed_url['query'];
-						}
-						if ( isset( $parsed_url['fragment'] ) ) {
-							$action_url .= '#' . $parsed_url['fragment'];
-						}
-					}
-				}
-			}
-
-			$action_url = esc_url_raw( $action_url );
 			$xhr_action = $node->getAttribute( 'action-xhr' );
 
 			// Make HTTP URLs protocol-less, since HTTPS is required for forms.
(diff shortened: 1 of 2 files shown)


includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -67,23 +67,41 @@ public function sanitize() {
...
-					if ( '?' === $action_url[0] || '#' === $action_url[0] ) {
-						// For actions consisting of only a query or URL fragment, include the schemeless-host and the REQUEST URI of the current page.
-						$action_url = $schemeless_host . wp_unslash( $_SERVER['REQUEST_URI'] ) . $action_url;
-					} elseif ( '.' === $action_url[0] ) {
-						// For actions consisting of relative paths (e.g. '../'), prepend the schemeless-host and a trailing-slashed REQUEST URI.
-						$action_url = $schemeless_host . trailingslashit( wp_unslash( $_SERVER['REQUEST_URI'] ) ) . $action_url;
-					} else {
-						// Otherwise, when the action URL includes an absolute path, just append it to the schemeless-host.
-						$action_url = $schemeless_host . $action_url;
+				$parsed_url = wp_parse_url( $action_url );
```
westonruter: This conditional is getting big. Please move it into a helper method as I've done in #4212. This will make the condition needing less nesting, as you can just `return` at more points.

includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -67,23 +67,41 @@ public function sanitize() {
...
+
+				// Ignore a malformed URL - it will be later sanitized.
+				if ( false !== $parsed_url ) {
+					// Handle relative URLs.
+					if ( ! isset( $parsed_url['scheme'] ) || 'https' !== $parsed_url['scheme'] ) {
```
westonruter: Why the `'https' !== $parsed_url['scheme']` condition here?

includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -67,23 +67,51 @@ public function sanitize() {
...
+							$parsed_url['host'] = $_SERVER['HTTP_HOST'];
+						}
+
+						if ( ! isset( $parsed_url['path'] ) ) {
+							// If there is action URL path, use the one from the request.
+							$parsed_url['path'] = trailingslashit( wp_unslash( $_SERVER['REQUEST_URI'] ) );
+						} elseif ( '' !== $parsed_url['path'] && '/' !== $parsed_url['path'][0] ) {
+							// If the path is relative, append it to the current request path.
+							$parsed_url['path'] = trailingslashit( wp_unslash( $_SERVER['REQUEST_URI'] ) ) . trailingslashit( $parsed_url['path'] );
+						}
```
westonruter: If no path is supplied and yet the host is present, then the path should assume to be empty. The current `REQUEST_URI` should not be supplied. I added a failing test case to show the proper behavior: f27f5a264251f4183ebfc35e94a79197d093aa07.

pierlon: Fixed in https://github.com/ampproject/amp-wp/pull/4250/commits/16b7fa8a7ee9b5ea7ebc47111f4495d8f6a54537.

includes/sanitizers/class-amp-form-sanitizer.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/4250/commits/1c9ab5326755b4f5374c4fda580aea66fb15c15f.

includes/sanitizers/class-amp-form-sanitizer.php
pierlon: To ignore URLs with the `https` scheme, otherwise it becomes schemeless which is unnecessary.

I've updated the comment there to reflect this:
https://github.com/ampproject/amp-wp/blob/16b7fa8a7ee9b5ea7ebc47111f4495d8f6a54537/includes/sanitizers/class-amp-form-sanitizer.php#L132

westonruter: @pierlon I think I simplified this further, and also added short-circuiting when it is a scheme-less URL (in which case nothing is left to do either). Please see e99eade20214510dd9fbd34c3930116fa26c5fc7.

includes/sanitizers/class-amp-form-sanitizer.php
```diff
@@ -58,32 +58,8 @@ public function sanitize() {
...
-			} else {
-				$action_url = $node->getAttribute( 'action' );
-
-				// Handle relative URLs.
-				if ( ! preg_match( '#^(https?:)?//#', $action_url ) ) {
-					$schemeless_host = '//' . $_SERVER['HTTP_HOST'];
-					if ( '?' === $action_url[0] || '#' === $action_url[0] ) {
-						// For actions consisting of only a query or URL fragment, include the schemeless-host and the REQUEST URI of the current page.
-						$action_url = $schemeless_host . wp_unslash( $_SERVER['REQUEST_URI'] ) . $action_url;
-					} elseif ( '.' === $action_url[0] ) {
```
westonruter: In your changes, you correctly changed the `'.' === $action_url[0]` to `'/' !== $action_url[0]`. This I believe fixes the case where there is an `action` URL like `foo/` as opposed to `../`. I added a test to confirm this: 585be89e054a1009a561e36904bfa51d1fe18361.

includes/sanitizers/class-amp-form-sanitizer.php
pierlon: Thanks! And its more readable :+1: .

PR #4183, opened 2020-01-26
Fix validation of links with mailto: protocol

westonruter:
## Summary

Fixes #4182.

- [ ] Prevent `INVALID_URL_PROTOCOL` validation error from being raised for links to `mailto:?&subject=Example&body=https://example.com/` 

## Checklist

- [ ] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [ ] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [ ] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

PR #4251, opened 2020-02-06
Prevent validation errors on valid mailto: links; restrict AMP-to-AMP linking to HTTP(S) protocol links

### LINKED ISSUE #4182: Invalid validation error and AMP-to-AMP linking failure for mailto: links
## Bug Description

Add a Custom HTML block that contains the following markup:

```html
<a href="mailto:?&#038;subject=Example&#038;body=https://example.com/">Email this</a>
```

This results in the AMP plugin detecting an `INVALID_URL_PROTOCOL` validation error:

image

This is in spite of the markup being entirely valid in AMP:

image

The issue also happens when an email address is provided:

```html
<a href="mailto:foo@example.com?&#038;subject=Example&#038;body=https://example.com/">Email this</a>
```

Please note: the issue _only occurs in Standard mode_. In Transitional mode, it turns out that the `AMP_Link_Sanitizer` is mutating the `mailto:` link in a way that causes it to no longer be invalid. This turns out to also be an error, as `\AMP_Link_Sanitizer::is_frontend_url()` needs to only apply to URLs with the HTTP and HTTPS scheme. (See #4183 which fixes this so you can test in Transitional mode if checking out that branch.)

The reason for why `AMP_Link_Sanitizer` prevents the links from being invalid is that it mutates the URL via `add_query_arg()` to append `?amp`. So it is making this change:

```diff
- mailto:?&subject=Example&body=https://example.com/
+ mailto:?subject=Example&body=https%3A%2F%2Fexample.com%2F&amp
```

The tag-and-attribute sanitizer likes the latter but not the former. But the AMP validator is OK with both, so we need to update our behavior to match. Our validator appears to be failing when the URL contains an unencoded colon (`:`), even though that is perfectly fine.

## Expected Behaviour

Links with `mailto:` protocol that contain a query parameter that have a URL value get flagged as a validation error.

## Steps to reproduce

1. Enable Standard mode
2. Create a Custom HTML block with the above `mailto:` link.
3. Notice validation error.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

kienstra:
## Summary

* Prevents validation errors on `mailto:` links, like Weston's example of `<a href="mailto:foo@example.com?&#038;subject=Example&#038;body=https://example.com/">Email this</a>`
* Uses lazy instead of greedy matching, so the lookahead of `(?=:)` will only match the first `:`

Fixes #4182 

## Testing
1. Activate Standard mode
2. Create a new post
3. Add a Custom HTML block with this markup that Weston shared in #4182 
```html
<a href="mailto:?&#038;subject=Example&#038;body=https://example.com/">Email this</a>
<a href="mailto:foo@example.com?&#038;subject=Example&#038;body=https://example.com/">Email this</a>
```
4. Go to the AMP URL
5. Expected: there's no validation error

## Regex
Your suggestion worked pretty well overall:

```diff
diff --git a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
index 9227c6a62..73944f731 100644
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -1666,7 +1666,7 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 	 * @return string|null Protocol without colon if matched. Otherwise null.
 	 */
 	private function parse_protocol( $url ) {
-		if ( preg_match( '#^[^/]+?(?=:)#', $url, $matches ) ) {
+		if ( preg_match( '#^[a-z][a-z0-9+.-]*(?=:)#', $url, $matches ) ) {
 			return $matches[0];
 		}
 		return null;
```

...but there were 2 unit test failures:

https://github.com/ampproject/amp-wp/blob/19d54bdf0ac4c41a37927e11241b723cdb754de8/tests/php/test-tag-and-attribute-sanitizer.php#L2018
and 
https://github.com/ampproject/amp-wp/blob/19d54bdf0ac4c41a37927e11241b723cdb754de8/tests/php/test-amp-tag-and-attribute-sanitizer-private-methods.php#L1439

For the `'malformed_attribute_syntax_curly_quotes'` failure, I think that's because `$this->parse_protocol()` returned `null` when passed `%E2%80%9Chttps://example.com/path/to/post/%E2%80%9D` here:
https://github.com/ampproject/amp-wp/blob/19d54bdf0ac4c41a37927e11241b723cdb754de8/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L1702-L1705

...so the conditional on the next line was false, and it didn't return an invalid protocol error as it should have.

It doesn't look like I could edit this to return `AMP_Rule_Spec::FAIL` if `$this->parse_protocol()` returns `null`. But maybe there is a fix to this that I missed 😄 

Thanks, Weston!

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-link-sanitizer.php
+++ b/includes/sanitizers/class-amp-link-sanitizer.php
@@ -204,6 +204,10 @@ class AMP_Link_Sanitizer extends AMP_Base_Sanitizer {
 	public function is_frontend_url( $url ) {
 		$parsed_url = wp_parse_url( $url );
 
+		if ( ! empty( $parsed_url['scheme'] ) && ! in_array( strtolower( $parsed_url['scheme'] ), [ 'http', 'https' ], true ) ) {
+			return false;
+		}
+
 		// Skip adding query var to links on other URLs.
 		if ( ! empty( $parsed_url['host'] ) && $this->home_host !== $parsed_url['host'] ) {
 			return false;
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -1666,7 +1666,7 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 	 * @return string|null Protocol without colon if matched. Otherwise null.
 	 */
 	private function parse_protocol( $url ) {
-		if ( preg_match( '#^[^/]+(?=:)#', $url, $matches ) ) {
+		if ( preg_match( '#^[^/]+?(?=:)#', $url, $matches ) ) {
 			return $matches[0];
 		}
 		return null;
--- a/tests/php/test-amp-tag-and-attribute-sanitizer-private-methods.php
+++ b/tests/php/test-amp-tag-and-attribute-sanitizer-private-methods.php
@@ -1616,6 +1616,22 @@ class AMP_Tag_And_Attribute_Sanitizer_Attr_Spec_Rules_Test extends WP_UnitTestCa
 				],
 				AMP_Rule_Spec::FAIL,
 			],
+			'protocol_mailto_pass'        => [
+				[
+					'source'         => '<a href="mailto:foo@example.com?&#038;subject=Example&#038;body=https://example.com/"></a>',
+					'node_tag_name'  => 'a',
+					'attr_name'      => 'href',
+					'attr_spec_rule' => [
+						'value_url' => [
+							'protocol' => [
+								'mailto',
+								'https',
+							],
+						],
+					],
+				],
+				AMP_Rule_Spec::PASS,
+			],
 		];
 	}
 
@@ -1646,9 +1662,9 @@ class AMP_Tag_And_Attribute_Sanitizer_Attr_Spec_Rules_Test extends WP_UnitTestCa
 				'//image.png ',
 				false,
 			],
-			'two_colons'        => [
-				'foo:baz://image.png ',
-				'foo:baz',
+			'mailto'            => [
+				'mailto:?&#038;subject=Foo&#038;body=https://example.com/',
+				'mailto',
 			],
 		];
 	}
--- a/tests/php/test-class-amp-link-sanitizer.php
+++ b/tests/php/test-class-amp-link-sanitizer.php
@@ -142,6 +142,11 @@ class AMP_Link_Sanitizer_Test extends WP_UnitTestCase {
 				'expected_amp' => false,
 				'expected_rel' => null,
 			],
+			'mailto-link'         => [
+				'href'         => 'mailto:nobody@example.com',
+				'expected_amp' => false,
+				'expected_rel' => null,
+			],
 		];
 
 		$admin_bar_link_href = home_url( '/?do_something' );
--- a/tests/php/test-tag-and-attribute-sanitizer.php
+++ b/tests/php/test-tag-and-attribute-sanitizer.php
@@ -1272,11 +1272,9 @@ class AMP_Tag_And_Attribute_Sanitizer_Test extends WP_UnitTestCase {
 				[],
 				[ AMP_Tag_And_Attribute_Sanitizer::INVALID_URL ],
 			],
-			'a_with_mail_host'                             => [
-				'<a class="foo" href="mail to:foo@bar.com">value</a>',
-				'<a class="foo">value</a>',
-				[],
-				[ AMP_Tag_And_Attribute_Sanitizer::INVALID_URL_PROTOCOL ],
+			'a_with_valid_mail_host'                       => [
+				'<a class="foo" href="mailto:foo@bar.com&amp;body=https://example.org/">value</a>',
+				null,
 			],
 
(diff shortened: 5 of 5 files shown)


kienstra:
Hi Weston,
Could you please review this when you have a chance? Maybe I missed something in the unit tests [mentioned above](https://github.com/ampproject/amp-wp/pull/4251#issue-371730520).

Thanks, Weston! See you soon.

westonruter:
> For the `'malformed_attribute_syntax_curly_quotes'` failure

Humm. It is indeed something that should be invalid:

image

It looks like my suggested regex does not align properly with the AMP validator. Please refer to this code to see if it can inform how the regex is constructed, or if non-regex is needed:

https://github.com/ampproject/amphtml/blob/a89d625c0538f40644ece37572cd6bef9f0a09c3/validator/engine/parse-url.js#L235-L282

kienstra:
Hi @westonruter,
Good idea to look at the AmpHtml sanitizer.

It looks like it:
1. [Sets the protocol](https://github.com/ampproject/amphtml/blob/a89d625c0538f40644ece37572cd6bef9f0a09c3/validator/engine/parse-url.js#L242-L252) as `http` or `https` if the URL starts with those and it's followed by `:`
2. [Looks for](https://github.com/ampproject/amphtml/blob/a89d625c0538f40644ece37572cd6bef9f0a09c3/validator/engine/parse-url.js#L254-L259) the first `:`
3. [Ensures that](https://github.com/ampproject/amphtml/blob/a89d625c0538f40644ece37572cd6bef9f0a09c3/validator/engine/parse-url.js#L261-L267) there's no invalid character before the `:`, very similar to your regex `[a-z][a-z0-9+.-]`

I think we could do something similar by reusing this:

https://github.com/ampproject/amp-wp/blob/19d54bdf0ac4c41a37927e11241b723cdb754de8/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L1637-L1642

...and adding a similar check here:

```diff
diff --git a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
index 9227c6a62..f1961e8ff 100644
--- a/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
+++ b/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
@@ -1700,7 +1700,14 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
                        if ( $node->hasAttribute( $attr_name ) ) {
                                foreach ( $this->extract_attribute_urls( $node->getAttributeNode( $attr_name ) ) as $url ) {
                                        $url_scheme = $this->parse_protocol( $this->normalize_url_from_attribute_value( $url ) );
-                                       if ( isset( $url_scheme ) && ! in_array( strtolower( $url_scheme ), $attr_spec_rule[ AMP_Rule_Spec::VALUE_URL ][ AMP_Rule_Spec::ALLOWED_PROTOCOL ], true ) ) {
+                                       if (
+                                               isset( $url_scheme ) &&
+                                               (
+                                                       ! in_array( strtolower( $url_scheme ), $attr_spec_rule[ AMP_Rule_Spec::VALUE_URL ][ AMP_Rule_Spec::ALLOWED_PROTOCOL ], true )
+                                                       ||
+                                                       ! preg_match( '/^[a-zA-Z0-9+-]+/i', $url_scheme )
+                                               )
+                                       ) {
                                                return AMP_Rule_Spec::FAIL;
                                        }
                                }
```

Of course, it'd probably be in a private method to be more DRY, if we did go with this approach.

kienstra:
Though maybe that diff above wouldn't help. 

There's probably not much need to evaluate:
```php
! preg_match( '/^[a-zA-Z0-9+-]+/i', $url_scheme )
```
...as this will very likely be `true` if that expression above is `true`:

```php
! in_array( strtolower( $url_scheme ), $attr_spec_rule[ AMP_Rule_Spec::VALUE_URL ][ AMP_Rule_Spec::ALLOWED_PROTOCOL ], true )
```

westonruter:
This PR is missing this part from the issue:

> Please note: the issue _only occurs in Standard mode_. In Transitional mode, it turns out that the `AMP_Link_Sanitizer` is mutating the `mailto:` link in a way that causes it to no longer be invalid. This turns out to also be an error, as `\AMP_Link_Sanitizer::is_frontend_url()` needs to only apply to URLs with the HTTP and HTTPS scheme. (See #4183 which fixes this so you can test in Transitional mode if checking out that branch.)

So this PR could have been finishing up #4183 or else just cherry-picking 1a213cd72278ce8638877719a4c65bb4e30d1bd9. I'll pull it in here.

kienstra:
>So this PR could have been finishing up #4183 or else just cherry-picking 1a213cd. I'll pull it in here.

Thanks, for cherry-picking that in. Sorry, I must have misread that.

PR #4257, opened 2020-02-09
Improve handling of fetching external stylesheets

westonruter:
## Summary

I discovered that an external stylesheet with a schemeless URL fails to be fetched because `wp_remote_get()` does not support such URLs. 

For example, given this plugin:

```php
<?php
/**
 * Plugin Name: Try External Stylesheet Schemeless URL
 */
add_action( 'wp_head', function() {
	?>
	<link rel="stylesheet" href="//make.wordpress.org/core/wp-content/plugins/o2/modules/follow/css/style.css?ver=<?php echo (int) time(); ?>">
	<?php
} );
```

The result is an error like so:

image

Not only should the URL be fetchable but the error message displayed here is not helpful.

So this PR ensures that a URL scheme is supplied when fetching an external stylesheet to prevent `wp_remote_get()` from failing. It uses the same URL scheme as the scheme for the `home_url()`, as this is the behavior of a schemeless URL.

Additionally, when the URL is actually invalid the error messages are now much more helpful (resolving some code todos), as the WP HTTP error message is passed through. For example:

```html
<link rel="stylesheet" href="bad://example.com/foo.css">
```

image

```html
<link rel="stylesheet" href="http://not-existing.example.com/foo.css">
```

image

```html
<link rel="stylesheet" href="http://example.com/notfound.css">
```

image

Relates to #1420.

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -1366,6 +1366,12 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 * @return string|WP_Error Stylesheet contents or WP_Error.
 	 */
 	private function fetch_external_stylesheet( $url ) {
+
+		// Prepend schemeless stylesheet URL with the same URL scheme as the current site.
+		if ( '//' === substr( $url, 0, 2 ) ) {
+			$url = wp_parse_url( home_url(), PHP_URL_SCHEME ) . ':' . $url;
+		}
+
 		$cache_key = md5( $url );
 		$contents  = get_transient( $cache_key );
 		if ( false === $contents ) {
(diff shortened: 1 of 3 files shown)


pierlon:
Testing instructions:

Create a new post and a Custom HTML block with the following content:

```html
<link rel="stylesheet" href="bad://example.com/foo.css">
<link rel="stylesheet" href="http://not-existing.example.com/foo.css">
<link rel="stylesheet" href="https://make.wordpress.org/notfound.css">
<link rel="stylesheet" href="//make.wordpress.org/core/wp-content/plugins/o2/modules/follow/css/style.css">
```

The following set of validation errors should be raised:

image

With the:
- First error being caused by `bad://example.com/foo.css`

- Second by `http://not-existing.example.com/foo.css`

- Third by `https://make.wordpress.org/notfound.css`

And no issue should be reported for `//make.wordpress.org/core/wp-content/plugins/o2/modules/follow/css/style.css`.

csossi:
Verified in QA:
image

PR #3619, opened 2019-10-24
Update allowed tags/attributes from spec in amphtml 1910161528000

westonruter:
Previously #3239.

- [x] Run `./bin/amphtml-update.sh`
- [x] Examine diff for changelog
- [x] Update spec generator as needed based on spec format changes.
- [x] Modify validating sanitizer based on changes to spec, if needed.
- [x] Add tests for key changes

# Changelog

* Support [inline `script` for `amp-script`](https://amp.dev/documentation/components/amp-script/#load-javascript-from-a-local-element), eliminating requirement that external scripts be used. Take note of the required `meta[name=amp-script-src]` tag, but the plugin will automatically merge multiple instances of the `meta` tag in the `head` into the one required. A new `amp_generate_script_hash()` function is provided to aid with generating the sha384 script hashes. See [demonstration plugin](https://gist.github.com/westonruter/d0bd2f39a45e664368d6e4ffc82da041) for usage. ✨
image 
* Add to list of bindable attributes for `amp-audio`.
* Add `smoothing` attribute to `amp-orientation-observer`.
* Add `max-age` attribute to `amp-script`.
* Add `amp-experiment story extension .json script`. Allow `script` in `amp-story-grid-layer`.

# Details

```bash
(
    PREV_VERSION=1910071803120;
    THIS_VERSION=1910161528000; 
    git checkout $THIS_VERSION;
    git diff $PREV_VERSION...$THIS_VERSION -w -- $( git ls-files | grep '.protoascii' );
    git checkout - > /dev/null
)
```

<details>
<summary>Compare 1910071803120..1910161528000</summary>

```diff
diff --git a/extensions/amp-accordion/validator-amp-accordion.protoascii b/extensions/amp-accordion/validator-amp-accordion.protoascii
index 66eea7306..c646ff1a8 100644
--- a/extensions/amp-accordion/validator-amp-accordion.protoascii
+++ b/extensions/amp-accordion/validator-amp-accordion.protoascii
@@ -16,7 +16,6 @@
 tags: {  # amp-accordion
   html_format: AMP
   html_format: AMP4ADS
-  html_format: AMP4EMAIL
   html_format: ACTIONS
   tag_name: "SCRIPT"
   extension_spec: {
@@ -28,6 +27,19 @@ tags: {  # amp-accordion
   }
   attr_lists: "common-extension-attrs"
 }
+tags: {  # amp-accordion
+  html_format: AMP4EMAIL
+  tag_name: "SCRIPT"
[...]

PR #4285, opened 2020-02-12
Support `mandatory_oneof` and `mandatory_anyof` constraints

### LINKED ISSUE #938: Add support for mandatory_oneof and mandatory_anyof attribute constraints
## Feature description

Someone had a `<iframe src="about:blank">` in a post on https://wordpress.org/support/topic/google-console-error/#post-9946202

This is resulting in an `<amp-iframe>` being rendered but without a `src` attribute. The result is in valid AMP because `src` is mandatory.

The whitelist sanitizer should be removing the `<amp-iframe>` in this case I should think. But otherwise, the iframe sanitizer itself can preemptively remove the iframe to ensure validity. Perhaps a placeholder could be put there in its place.

---------------

## Acceptance criteria

1. The iframe sanitizer should preemptively remove the iframe to ensure validity
2. When `<iframe src="about:blank">`is  in a post the whitelist sanitizer should be removing the `<amp-iframe>`

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

kienstra:
## Summary

Adds `mandatory_anyof` and `mandatory_oneof` to `amphtml-update.py`, and verifies against it in the sanitizer

## Testing
1. As the description of #938 mentions, adding this to a Custom HTML block:

```html
<iframe src="about:blank">
```

...now results in the `<amp-iframe>` being stripped, and the validation errors now include:

mandatory-oneof

2. And adding this to a Custom HTML block:
```html
<amp-list width="400" height="400"></amp-list>
```
...now results in this validation error:

yan-spec

Fixes #938

## Checklist

### DIFF
--- a/bin/amphtml-update.py
+++ b/bin/amphtml-update.py
@@ -655,6 +655,14 @@ def GetTagRules(tag_spec):
 				amp_layout[ field[0].name ] = field[1]
 		tag_rules['amp_layout'] = amp_layout
 
+	mandatory_anyof = GetMandatoryOf(tag_spec.attrs, 'mandatory_anyof')
+	if mandatory_anyof:
+		tag_rules.update( mandatory_anyof )
+
+	mandatory_oneof = GetMandatoryOf(tag_spec.attrs, 'mandatory_oneof')
+	if mandatory_oneof:
+		tag_rules.update( mandatory_oneof )
+
 	logging.info('... done')
 	return tag_rules
 
@@ -708,9 +716,6 @@ def GetValues(attr_spec):
 	# mandatory is a boolean
 	if attr_spec.HasField('mandatory'):
 		value_dict['mandatory'] = attr_spec.mandatory
-
-	value_dict = AddMandatoryOf( value_dict, attr_spec, 'mandatory_anyof' )
-	value_dict = AddMandatoryOf( value_dict, attr_spec, 'mandatory_oneof' )
 
 	# Add allowed value
 	if attr_spec.value:
@@ -776,21 +781,21 @@ def UnicodeEscape(string):
 	"""
 	return ('' + string).encode('unicode-escape')
 
-def AddMandatoryOf( spec_dict, attribute_spec, constraint ):
+def GetMandatoryOf( attr, constraint ):
 	"""Gets the spec dictionary, with mandatory_*of constraints possibly added.
 
 	Args:
-		spec_dict: A dictionary of the attr_spec values to possibly add the constraint to.
-		attribute_spec: The attribute spec in which to look for the mandatory_*of constraint.
+		attr: The attributes in which to look for the mandatory_*of constraint.
 		constraint: A string of the mandatory_*of constraint, like 'mandatory_anyof'.
 	Returns:
-		The spec_dict, with the constraint possibly added.
+		A spec_dict with the constraint added, or null.
 	"""
-	if attribute_spec.HasField(constraint):
-		mandatory_of = getattr(attribute_spec, constraint).lstrip('[').rstrip(']').split(',')
-		spec_dict[constraint] = [GetMandatoryOfAttr( oneof ) for oneof in mandatory_of]
-
-	return spec_dict
+	for attr_spec in attr:
+		if attr_spec.HasField(constraint):
+			spec_dict = {}
+			mandatory_of = getattr(attr_spec, constraint).lstrip('[').rstrip(']').split(',')
+			spec_dict[constraint] = [GetMandatoryOfAttr(oneof) for oneof in mandatory_of]
+			return spec_dict
 
 def GetMandatoryOfAttr( attribute ):
 	"""Gets mandatory_*of attribute value to add to the generated file.
--- a/bin/amphtml-update.py
+++ b/bin/amphtml-update.py
@@ -28,6 +28,7 @@ import json
 import google
 from collections import defaultdict
 import imp
+import re
 
 seen_spec_names = set()
 
@@ -787,9 +788,26 @@ def AddMandatoryOf( spec_dict, attribute_spec, constraint ):
 	"""
 	if attribute_spec.HasField(constraint):
 		mandatory_of = getattr(attribute_spec, constraint).lstrip('[').rstrip(']').split(',')
-		spec_dict[constraint] = [oneof.strip(' ').strip("'").replace('[', 'data-amp-bind-').rstrip(']') for oneof in mandatory_of]
+		spec_dict[constraint] = [GetMandatoryOfAttr( oneof ) for oneof in mandatory_of]
 
 	return spec_dict
+
+def GetMandatoryOfAttr( attribute ):
+	"""Gets mandatory_*of attribute value to add to the generated file.
+
+	Args:
+		attribute: A string mandatory_*of attribute value, as parsed from the spec file.
+	Returns:
+		The attribute to pass to the generated PHP file.
+	"""
+	parsed_attribute = attribute.strip(' ').strip("'")
+
+	# Convert something like [src] to data-amp-bind-src.
+	return re.sub(
+		"^\[(\S+)\]$",
+		r"data-amp-bind-\1",
+		parsed_attribute
+	)
 
 def Phpize(data, indent=0):
 	"""Helper function to convert JSON-serializable data into PHP literals.
--- a/bin/amphtml-update.py
+++ b/bin/amphtml-update.py
@@ -655,13 +655,10 @@ def GetTagRules(tag_spec):
 				amp_layout[ field[0].name ] = field[1]
 		tag_rules['amp_layout'] = amp_layout
 
-	mandatory_anyof = GetMandatoryOf(tag_spec.attrs, 'mandatory_anyof')
-	if mandatory_anyof:
-		tag_rules.update(mandatory_anyof)
-
-	mandatory_oneof = GetMandatoryOf(tag_spec.attrs, 'mandatory_oneof')
-	if mandatory_oneof:
-		tag_rules.update(mandatory_oneof)
+	for mandatory_of_constraint in ['mandatory_anyof', 'mandatory_oneof']:
+		mandatory_of_spec = GetMandatoryOf(tag_spec.attrs, mandatory_of_constraint)
+		if mandatory_of_spec:
+			tag_rules[ mandatory_of_constraint ] = mandatory_of_spec
 
 	logging.info('... done')
 	return tag_rules
@@ -782,37 +779,27 @@ def UnicodeEscape(string):
 	return ('' + string).encode('unicode-escape')
 
 def GetMandatoryOf( attr, constraint ):
-	"""Gets the mandatory_*of value, if it exists.
+	"""Gets the attributes with the passed mandatory_*of constraint, if there are any.
 
 	Args:
 		attr: The attributes in which to look for the mandatory_*of constraint.
 		constraint: A string of the mandatory_*of constraint, like 'mandatory_anyof'.
 	Returns:
-		A dictionary with the constraint name mapped to its value.
+		A list of attributes that have that constraint name.
 	"""
+	attributes = []
 	for attr_spec in attr:
 		if attr_spec.HasField(constraint):
-			spec_dict = {}
-			mandatory_of = getattr(attr_spec, constraint).strip('[]').split(',')
-			spec_dict[constraint] = [GetMandatoryOfAttr(oneof) for oneof in mandatory_of]
-			return spec_dict
+			attributes.append(
+				# Convert something like [src] to data-amp-bind-src.
+				re.sub(
+					"^\[(\S+)\]$",
+					r"data-amp-bind-\1",
+					attr_spec.name
+				)
+			)
 
-def GetMandatoryOfAttr( attribute ):
-	"""Gets mandatory_*of attribute value to add to the generated file.
-
-	Args:
-		attribute: A string mandatory_*of attribute value, as parsed from the spec file.
-	Returns:
-		The attribute to pass to the generated PHP file.
-	"""
-	parsed_attribute = attribute.strip(' ').strip("'")
-
-	# Convert something like [src] to data-amp-bind-src.
-	return re.sub(
-		"^\[(\S+)\]$",
-		r"data-amp-bind-\1",
-		parsed_attribute
-	)
+	return attributes
 
 def Phpize(data, indent=0):
 	"""Helper function to convert JSON-serializable data into PHP literals.
--- a/bin/amphtml-update.py
+++ b/bin/amphtml-update.py
@@ -657,11 +657,11 @@ def GetTagRules(tag_spec):
 
 	mandatory_anyof = GetMandatoryOf(tag_spec.attrs, 'mandatory_anyof')
 	if mandatory_anyof:
-		tag_rules.update( mandatory_anyof )
+		tag_rules.update(mandatory_anyof)
 
 	mandatory_oneof = GetMandatoryOf(tag_spec.attrs, 'mandatory_oneof')
 	if mandatory_oneof:
-		tag_rules.update( mandatory_oneof )
+		tag_rules.update(mandatory_oneof)
 
 	logging.info('... done')
 	return tag_rules
--- a/bin/amphtml-update.py
+++ b/bin/amphtml-update.py
@@ -782,18 +782,18 @@ def UnicodeEscape(string):
 	return ('' + string).encode('unicode-escape')
 
 def GetMandatoryOf( attr, constraint ):
-	"""Gets the spec dictionary, with mandatory_*of constraints possibly added.
+	"""Gets the mandatory_*of value, if it exists.
 
 	Args:
 		attr: The attributes in which to look for the mandatory_*of constraint.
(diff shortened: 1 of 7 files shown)


includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1423,6 +1448,38 @@ private function check_attr_spec_rule_mandatory( DOMElement $node, $attr_name, $
...
+	 * @param array[]    $attr_spec The full attribute spec.
+	 *
+	 * @return string:
+	 *      - AMP_Rule_Spec::PASS - $attr_name is mandatory and it exists
+	 *      - AMP_Rule_Spec::FAIL - $attr_name is mandatory, but doesn't exist
+	 *      - AMP_Rule_Spec::NOT_APPLICABLE - $attr_name is not mandatory
+	 */
+	private function check_attr_spec_rule_mandatory_oneof( DOMElement $node, $attr_spec ) {
+		foreach ( $attr_spec as $attr_name => $attr_spec_rule_value ) {
+			if ( ! empty( $attr_spec_rule_value[ AMP_Rule_Spec::MANDATORY_ONEOF ] ) ) {
```
kienstra: This needs more thought, as it only checks one `mandatory_oneof` per node.

westonruter:
For updating the spec, please update at the same version that it was last updated from so that other spec changes are not included. That would be specifically 1910161528000 as done in #3619.

bin/amphtml-update.py
```diff
@@ -772,6 +775,22 @@ def UnicodeEscape(string):
...
+
+	Args:
+		spec_dict: A dictionary of the attr_spec values to possibly add the constraint to.
+		attribute_spec: The attribute spec in which to look for the mandatory_*of constraint.
+		constraint: A string of the mandatory_*of constraint, like 'mandatory_anyof'.
+	Returns:
+		The spec_dict, with the constraint possibly added.
+	"""
+	if attribute_spec.HasField(constraint):
+		mandatory_of = getattr(attribute_spec, constraint).lstrip('[').rstrip(']').split(',')
```
westonruter: Why are the square brackets being stripped? I believe `[` should rather be replaced with `data-amp-bind-` if I understand this right.

kienstra: Yeah, this stripping of square brackets isn't very elegant. 


Those calls are there because the `.protoascii` files have [this format](https://github.com/ampproject/amphtml/blob/b765e57b8748e31de55fc8249ed0540500107997/extensions/amp-addthis/validator-amp-addthis.protoascii#L34) for `mandatory_oneof` and `mandatory_anyof`:

```
mandatory_oneof: "['data-product-code', 'data-widget-id']"
```

Without the `lstrip()` and `rstrip()` calls above, here's what it looks like:

```diff
diff --git a/includes/sanitizers/class-amp-allowed-tags-generated.php b/includes/sanitizers/class-amp-allowed-tags-generated.php
index cd8679638..db20c21b2 100644
--- a/includes/sanitizers/class-amp-allowed-tags-generated.php
+++ b/includes/sanitizers/class-amp-allowed-tags-generated.php
@@ -948,8 +948,8 @@ class AMP_Allowed_Tags_Generated {
                                'attr_spec_list' => array(
                                        'data-product-code' => array(
                                                'mandatory_oneof' => array(
-                                                       'data-product-code',
-                                                       'data-widget-id',
+                                                       '[\'data-product-code',
+                                                       'data-widget-id\']',
                                                ),
                                        ),

```

Still, there could be a better way to handle that.

westonruter: I see. Strange that the value is a string and not already parsed.

kienstra: Yeah, it's a little weird. Maybe a regex would be more elegant, but it might be a little more code.

bin/amphtml-update.py
```diff
@@ -772,6 +775,22 @@ def UnicodeEscape(string):
...
+		spec_dict[constraint] = [oneof.strip(' ').strip("'") for oneof in mandatory_of]
```
westonruter: This will need to ensure that `'[src]'` becomes `'data-amp-bind-src'`, so replacing the first `[` with `data-amp-bind-` and stripping the last ']'.

includes/sanitizers/class-amp-allowed-tags-generated.php
```diff
@@ -3751,7 +3840,12 @@ class AMP_Allowed_Tags_Generated {
 					),
 					'credentials' => array(),
 					'data-amp-bind-is-layout-container' => array(),
-					'data-amp-bind-src' => array(),
+					'data-amp-bind-src' => array(
+						'mandatory_anyof' => array(
+							'src',
+							'[src]',
```
westonruter: This should get generated as `'data-amp-bind-src'`. See https://github.com/ampproject/amp-wp/pull/4285/files#r378581454.

bin/amphtml-update.py
kienstra: Ah, good point.

includes/sanitizers/class-amp-allowed-tags-generated.php
kienstra: Good idea, it's applied in https://github.com/ampproject/amp-wp/compare/39926b690632afdfe86458844d5e34614e85fb76...bcec1fa7e40d487b656f5dde699c650ef72a06cc

bin/amphtml-update.py
kienstra: Applied in https://github.com/ampproject/amp-wp/compare/39926b690632afdfe86458844d5e34614e85fb76...bcec1fa7e40d487b656f5dde699c650ef72a06cc

kienstra:
Hi @westonruter,
Thanks for looking at this. It should be ready for another round of review when you have a chance.

Thanks, and looking forward to talking tomorrow.

includes/sanitizers/class-amp-allowed-tags-generated.php
```diff
@@ -3266,7 +3342,12 @@ class AMP_Allowed_Tags_Generated {
 							),
 						),
 					),
-					'srcdoc' => array(),
+					'srcdoc' => array(
+						'mandatory_oneof' => array(
+							'src',
+							'srcdoc',
+						),
+					),
```
westonruter: I know this isn't really something you can answer, but something seems off about the data format in [the validator spec](https://github.com/ampproject/amphtml/blob/5fb4e7bb72012c2e088ffda951c3b815b6e1caba/extensions/amp-iframe/validator-amp-iframe.protoascii#L67-L80):

```
  attrs: {
    name: "src"
    mandatory_oneof: "['src', 'srcdoc']"
    value_url: {
      protocol: "data"
      protocol: "https"
      allow_relative: true  # Will be set to false at a future date.
    }
    blacklisted_value_regex: "__amp_source_origin"
  }
  attrs: {
    name: "srcdoc"
    mandatory_oneof: "['src', 'srcdoc']"
  }
```

It's also somewhat strange that `"['src', 'srcdoc']"` is serialized as pseudo-JSON, as opposed to being represented as an actual array in the protoascii.

Why is `mandatory_oneof` duplicated in both `srcdoc` and `src`?

In other words, it would seem more logical for the `mandatory_oneof` property to be placed under the `tag_spec` as opposed to duplicated inside of each attribute. For example:

```diff
diff --git a/includes/sanitizers/class-amp-allowed-tags-generated.php b/includes/sanitizers/class-amp-allowed-tags-generated.php
index b60135580..1aae2b432 100644
--- a/includes/sanitizers/class-amp-allowed-tags-generated.php
+++ b/includes/sanitizers/class-amp-allowed-tags-generated.php
@@ -3290,6 +3290,10 @@ class AMP_Allowed_Tags_Generated {
 					),
 				),
 				'tag_spec' => array(
+					'mandatory_oneof' => array(
+						'src',
+						'srcdoc',
+					),
 					'amp_layout' => array(
 						'supported_layouts' => array(
 							6,
```

Would it make sense to gather up the `mandatory_oneof` list across all attributes and then store them once in the `tag_spec`?

westonruter: It turns out the value of the `mandatory_oneof` is just an opaque string that is used for the error message. The validator doesn't actually parse it. See the logic in the JS validator: https://github.com/ampproject/amphtml/blob/5fb4e7bb72012c2e088ffda951c3b815b6e1caba/validator/engine/validator.js#L4732-L4761

That is why the string is duplicated across the attributes, since it is just a key.

So in theory, instead of:

```
  attrs: {
    name: "src"
    mandatory_oneof: "['src', 'srcdoc']"
    value_url: {
      protocol: "data"
      protocol: "https"
      allow_relative: true  # Will be set to false at a future date.
    }
    blacklisted_value_regex: "__amp_source_origin"
  }
  attrs: {
    name: "srcdoc"
    mandatory_oneof: "['src', 'srcdoc']"
  }
```

The protoascii could have factored out the `mandatory_oneof` like so:

```
  attrs: {
    name: "src"
    value_url: {
      protocol: "data"
      protocol: "https"
      allow_relative: true  # Will be set to false at a future date.
    }
    blacklisted_value_regex: "__amp_source_origin"
  }
  attrs: {
    name: "srcdoc"
  }
  mandatory_oneof: ['src', 'srcdoc']
```

If we represented the data in this way, this would eliminate the current duplication of checking the same constraints across all attributes that have the same `mandatory_oneof` value in a given tag.

westonruter: Otherwise, the only way to prevent the duplication now would seem to be to replicate what the JS validator is doing and instead of checking `check_attr_spec_rule_mandatory_number_of` and `check_attr_spec_rule_mandatory_number_of` _inside_ the loop that iterates over the attributes, to instead check it once after they have been iterated over.

Nevertheless, it seems better to improve the data model by factoring out the duplicated data from the attributes and into the `tag_spec`.

kienstra: Yeah, that's a good idea to remove the duplication in the generated PHP file. I'll work on applying it.

kienstra: https://github.com/ampproject/amp-wp/pull/4285/commits/969def47eb93e9e993e04e9cde8221b353c09c88 updates the generated file, but of course I need to update the PHP logic.

includes/sanitizers/class-amp-allowed-tags-generated.php
```diff
@@ -3834,6 +3870,12 @@ class AMP_Allowed_Tags_Generated {
 					),
 				),
 				'tag_spec' => array(
+					'mandatory_oneof' => array(
```
kienstra: The `mandatory_*of` rules are now in the `tag_spec`

kienstra:
Hi @westonruter,
Thanks for waiting. this should be ready for another look whenever you have a chance.

bin/amphtml-update.py
```diff
@@ -772,6 +781,39 @@ def UnicodeEscape(string):
...
+		constraint: A string of the mandatory_*of constraint, like 'mandatory_anyof'.
+	Returns:
+		A dictionary with the constraint name mapped to its value.
+	"""
+	for attr_spec in attr:
+		if attr_spec.HasField(constraint):
+			spec_dict = {}
+			mandatory_of = getattr(attr_spec, constraint).strip('[]').split(',')
+			spec_dict[constraint] = [GetMandatoryOfAttr(oneof) for oneof in mandatory_of]
+			return spec_dict
```
westonruter: I have a suggestion to improve the logic here which would eliminate the need for `GetMandatoryOfAttr`. Given that the value is not actually parsed, what can be done is merely to check if the key exists for a given attribute, and of so, add the attribute name to the list for the constraint. 

So this could be replaced with:

```python
attributes = []
for attr_spec in attr:
	if attr_spec.HasField(constraint):
		attributes.append( attr_spec.name )
if len( attributes ) == 0:
	return None
return attributes
```

And then `GetMandatoryOfAttr` could be removed. 

Is that right?

kienstra: Thanks, that's a great idea. 

https://github.com/ampproject/amp-wp/pull/4285/commits/3b1dbf8a830e75d9ac55ffecaecad25124ab40d8 applies it.

westonruter:
@kienstra I found that the `MANDATORY_ONEOF_ATTR_MISSING` error was conflating two different error scenarios: (1) missing any attributes and (2) duplicate exclusive attributes.

In f18f6c2 I've addressed that by introducing a new error code: `DUPLICATE_ONEOF_ATTRS`.

Given a post with this content:

```html
<!-- wp:html -->
<amp-iframe src="about:blank"></amp-iframe>
<!-- /wp:html -->

<!-- wp:html -->
<amp-list width="400" height="400"></amp-list>
<!-- /wp:html -->

<!-- wp:html -->
<amp-iframe src="https://example.com" srcdoc="https://foo.com" width="200" height="100"></amp-iframe><!-- Bad because both present! -->
<!-- /wp:html -->
```

The errors in the validated URL screen are now:

image

westonruter:
The message for the first validation error (`INVALID_URL_PROTOCOL`) needs to be improved in the context of #1420.

@kienstra Please review my enhancements and let me know if they check out.

includes/sanitizers/class-amp-allowed-tags-generated.php
```diff
@@ -3296,6 +3296,10 @@ class AMP_Allowed_Tags_Generated {
...
+						'src',
+						'srcdoc',
+					),
```
kienstra: Hm, this looks to be the same as the `mandatory_oneof` spec below:

```
'mandatory_oneof' => array(
	'src',
	'srcdoc',
),

File: includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php [hunk 9]
```diff
@@ -57,6 +57,7 @@ class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 	const ATTR_REQUIRED_BUT_MISSING            = 'ATTR_REQUIRED_BUT_MISSING';
 	const MANDATORY_ANYOF_ATTR_MISSING         = 'MANDATORY_ANYOF_ATTR_MISSING';
 	const MANDATORY_ONEOF_ATTR_MISSING         = 'MANDATORY_ONEOF_ATTR_MISSING';
+	const DUPLICATE_ONEOF_ATTRS                = 'DUPLICATE_ONEOF_ATTRS';
```
kienstra: Good idea, having `MISSING` in the error code when there are too many is misleading.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -1450,50 +1465,20 @@ private function check_attr_spec_rule_mandatory( DOMElement $node, $attr_name, $
...
-					$tag_spec[ $constraint_type ],
-					static function( $attribute ) use ( $node ) {
-						return $node->hasAttribute( $attribute );
-					}
-				)
-			);
-
-			if ( ! $this->is_matched_attribute_count_acceptable( $matched_attribute_count, $constraint_type ) ) {
-				return $tag_spec[ $constraint_type ];
+	private function get_element_attribute_intersection( DOMElement $element, $attribute_names ) {
```
kienstra: Good idea to have this return the intersection.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -2951,6 +2962,48 @@ public static function get_error_title_from_code( $validation_error ) {
...
+				}
+				break;
+			case AMP_Tag_And_Attribute_Sanitizer::MANDATORY_ONEOF_ATTR_MISSING:
+			case AMP_Tag_And_Attribute_Sanitizer::MANDATORY_ANYOF_ATTR_MISSING:
+				$attributes_key = null;
+				if ( AMP_Tag_And_Attribute_Sanitizer::MANDATORY_ONEOF_ATTR_MISSING === $validation_error['code'] ) {
+					$title          = __( 'Missing exclusive mandatory attribute', 'amp' );
+					$attributes_key = 'mandatory_oneof_attrs';
+				} else {
+					$title          = __( 'Missing at least one mandatory attribute', 'amp' );
```
kienstra: This looks good, it's much more descriptive than before:

amp-list-here

includes/sanitizers/class-amp-allowed-tags-generated.php
westonruter: Oh, that's bizzare. How is Python generating an array with duplicate keys?

```php
				'tag_spec' => array(
					'mandatory_oneof' => array(
						'src',
						'srcdoc',
					),
					'amp_layout' => array(
						'supported_layouts' => array(
							6,
							2,
							3,
							7,
							9,
							1,
							4,
						),
					),
					'mandatory_oneof' => array(
						'src',
						'srcdoc',
					),
					'requires_extension' => array(
						'amp-iframe',
					),
				),
```

westonruter: I just re-ran `./bin/amphtml-update.sh` and it removed the duplicate: 4fb49daa0.

kienstra: Yeah, that's a little strange.

kienstra:
🎉

PR #4296, opened 2020-02-15
Enfore UTF-8 content-type header

### LINKED ISSUE #855: Support WordPress installs with non-UTF-8 charsets

schlessera:
## Summary

We're already convert the encoding from non-UTF-8 to UTF-8 within the DOMDocument extension, but WordPress still outputs a content-type header with the old charset.

This PR fixes this, and also removes a `trigger_error` that is not needed anymore.

Fixes #855

## Checklist

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -2054,7 +2054,9 @@ class AMP_Theme_Support {
 		}
 
 		// Enforce UTF-8 encoding as it is a requirement for AMP.
-		header( 'Content-Type: text/html; charset=utf-8' );
+		if ( ! headers_sent() ) {
+			header( 'Content-Type: text/html; charset=utf-8' );
+		}
 
 		/**
 		 * Filters whether response (post-processor) caching is enabled.
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -2053,6 +2053,9 @@ class AMP_Theme_Support {
 			return $response;
 		}
 
+		// Enforce UTF-8 encoding as it is a requirement for AMP.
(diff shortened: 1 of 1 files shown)


westonruter:
I confirmed this worked by forcing a site to be created in latin1 and then creating a post with the table in https://cs.stanford.edu/people/miles/iso8859.html

The Latin1 characters rendered as expected in AMP both in Transitional and Reader modes, with the page itself being actually UTF-8 because of the AMP plugin.

kienstra:
**Looks Good!**

Hi @schlessera,
Nice, the AMP and non-AMP URLs now look the same, even when the non-AMP URL uses `latin1`.

Testing this PR with these [instructions](https://github.com/ampproject/amp-wp/issues/855#issuecomment-586044812), the look the same:

# Non-AMP

non-amp-version

# AMP

amp-version

schlessera:
Phew! The endless fiddling with DOMDocument bugs and encoding weirdnesses was successful after all!

PR #4297, opened 2020-02-15
Support content after HTML end tag

### LINKED ISSUE #4282: Failure to normalize documents with HTML comments after </body>
## Bug Description

When markup appears after `</body>` or `</html>`, the document's original `<body>` is lost. 
There is a bug in `\Amp\AmpWP\Dom\Document::normalize_document_structure()` in particular with the pattern in `\Amp\AmpWP\Dom\Document::HTML_STRUCTURE_BODY_END_TAG` as used here:

https://github.com/ampproject/amp-wp/blob/2026afc6fa017ad6f3ed0305d95138115526c41b/src/Dom/Document.php#L492-L497

This is causing new `<body>` to be added to the page which is blowing away the existing one.

## Expected Behaviour

The original `<body>` with all of its valid attributes should not be removed during normalization when there is markup or HTML comments appearing after `</body>` or `</html>`. Any markup appearing after `</body>` should get moved inside during normalization, although beware of moving HTML comment nodes as this may cause problems with validation. This came up before previously in #4104.

## Steps to reproduce

Modify the `footer.php` of Twenty Twenty to end with:

```html
	</body>
</html>
<!-- a comment! -->
```

or

```html
	</body>
<!-- a comment! -->
</html>
```

When then look at an AMP page in Standard/Transitional mode. Notice that the `body` element of the page is now just:

```html
<body id="body">
```

Whereas it should be:

```html
<body class="home blog logged-in admin-bar enable-search-modal has-no-pagination showing-comments show-avatars footer-top-visible customize-support" id="body">
```

## Additional context

 - WordPress version:
 - Plugin version:
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
## Summary

The `normalize_document_structure()` method in `Dom\Document` didn't support content after the HTML end tag. In that case, it failed to remove the end tag, and the subsequent manipulations to normalize further resulted in an overwritten <body> tag that lost all attributes.

This PR changes the regular expressions and the reassembly behavior so that comments are properly kept intact and in the correct ordering across the HTML structure.

It also changes most regular expressions to use atomic groups where needed to avoid any backtracking and improve regex performance.

Fixes #4282

## Checklist


src/Dom/Document.php
```diff
@@ -477,7 +478,10 @@ private function normalize_document_structure( $content ) {
 		if ( preg_match( self::HTML_STRUCTURE_HTML_START_TAG, $content, $matches ) ) {
 			$html_start = $matches['html_start'];
 			$content    = preg_replace( self::HTML_STRUCTURE_HTML_START_TAG, '', $content, 1 );
-			$content    = preg_replace( self::HTML_STRUCTURE_HTML_END_TAG, '', $content, 1 );
+
+			preg_match( self::HTML_STRUCTURE_HTML_END_TAG, $content, $matches );
+			$html_end = $matches['html_end'];
```
pierlon:
```suggestion
			$html_end = isset( $matches['html_end'] ) ? $matches['html_end'] : $html_end;
```

westonruter:
> However, it is _not_ fixing the problem when there is a comment after the `</body>` but before the `</body>`.

In Twenty Twenty's `footer.php`, if I change the end to be:

```html
	</body>
<!-- a comment! -->
</html>
```

Then the result is a `body` tag that looks like `<body id="body">`, where the `class` attribute is lost.

westonruter:
Accidentally closed when merging #4299.

schlessera:
> Also, the comments should maintain to their original relative location, which is important for source stack tracing.

I think that particular comment is not being moved by the normalization, but by PHP's `DOMDocument`. I'll look into it.

schlessera:
Oh, no, I'm mistaken, it is indeed moved by the normalization. I'll fix it.

schlessera:
I think I'll hav to recode the normalization to tokenize first and then iterate over tokens.

While trying to make more robust tests for comment positioning I used this HTML:
```html
<!-- before <!doctype> --><!DOCTYPE html><!-- before <html> --><html><!-- before <head> --><head><!-- first within <head> --><meta charset="utf-8"><!-- last within <head> --></head><!-- before <body> --><body class="something" data-something="something"><!-- within <body> --></body><!-- after </body> --></html><!-- after </html> -->
```

And lo and behold, I get this test result:
```
-    1 => '<!-- before <!doctype>'
-    3 => ' --'
-    4 => '>'
-    5 => '<!DOCTYPE html>'
-    7 => '<!-- before <html>'
-    9 => ' --'
-    10 => '>'
-    11 => '<html>'
-    13 => '<!-- before <head>'
+    1 => '<!DOCTYPE html>'
+    3 => '<html>'
+    5 => '<head>'
+    7 => '<meta charset="utf-8">'
+    9 => '</head>'
+    11 => '<body>'
+    13 => '<!-- before <!doctype>'
     15 => ' --'
     16 => '>'
-    17 => '<head>'
-    19 => '<!-- first within <head>'
-    21 => ' --'
-    22 => '>'
-    23 => '<meta charset="utf-8">'
-    25 => '<!-- last within <head>'
+    17 => '<!-- before <html>'
+    19 => ' --'
+    21 => '<!-- before <head>'
+    23 => ' --'
+    25 => '<!-- first within <head>'
     27 => ' --'
     28 => '>'
-    29 => '</head>'
-    31 => '<!-- before <body>'
-    33 => ' --'
-    34 => '>'
-    35 => '<body class="something" data-something="something">'
+    29 => '<!-- last within <head>'
+    31 => ' --'
+    33 => '<!-- before <body>'
+    35 => ' --'
     37 => '<!-- within <body>'
     39 => ' --'
     40 => '>'
-    41 => '</body>'
-    43 => '<!-- after </body>'
-    45 => ' --'
-    46 => '>'
-    47 => '</html>'
-    49 => '<!-- after </html>'
-    51 => ' --'
-    52 => '>'
+    41 => '<!-- after </body>'
+    43 => ' --'
+    45 => '<!-- after </html>'
+    47 => ' --'
+    49 => '</body>'
+    51 => '</html>'
+    20 => '>'
+    24 => '>'
+    32 => '>'
+    36 => '>'
+    44 => '>'
+    48 => '>'
```

This means that, of course, my regular expressions are also triggered by tags that are found within comments. I don't think there's a clean and performant way of actually ignoring these comments with a singular regular expression. So I'll rewrite this to tokenize everything first and then rearrange. I don't need to tokenize every single tag, only the big pieces like first-level comments, body, head, etc...
That should make it robust enough to deal with the comments properly.

schlessera:
Maybe not all hope is lost. A large part of the problem is the `assertEqualMarkup` which splits up the comments. I'll first try to salvage the existing code therefore.

schlessera:
@westonruter This is what the code currently supports and keeps intext in terms of comment ordering:
```html
<!DOCTYPE html><!-- before <html> --><html><!-- before <head> --><head><meta charset="utf-8"><!-- within <head> --></head><!-- before <body> --><body class="something" data-something="something"><!-- within <body> --></body><!-- after </body> --></html><!-- after </html> -->
```

What doesn't work yet is:
1. a comment **before the <!doctype>**, as the `DOMDocument` moves it to be the very first tag automatically;
2. a comment **as first node of <head>**, as we currently enforce our `<meta charset="utf-8">` to be the very first node within <head>.

For 1., it would be cumbersome to fix, as we'd have to store additional markup in some form and then re-establish order after parsing.
For 2., it would mean making the code to deal with the charset a bit more complex to keep the charset intact but adapt the actual encoding stored in it, as opposed to removed whatever charset and re-adding a normalized one as we do now.

For these two, do you want me to invest time in fixing them, or as these not worth it?

westonruter:
@schlessera in the case of the doctype, it probably doesn't matter because we force it to be the HTML5 version anyway, thus there is no possibility for a validation error. So if a source stack open comment gets moved to be after the doctype and then is immediately followed by the source stack closing comment, then that is fine.

I suppose the same could be said for the meta charset.

As long as the source stack comments retain their relative position to each other (so a closing comment doesn't come before an opening one) then if the markup inside being annotated gone because we force it to be something, then I think it's okay.

Aside: this brings up something important for the Optimizer work. We need to make sure that the Optimizer does not run during validation requests, since the reordering of elements in the `head` will probably cause elements to lose the source stack context.

westonruter:
Reader for conflict resolution. Take note of https://github.com/ampproject/amp-wp/pull/4333#issuecomment-596052096

westonruter:
@schlessera I tried modifying Twenty Twenty's `footer.php` to:

```html
	</body>
</html>
<!--after html!-->
```

But when accessing the AMP page, I see:

```html
	<!--after html!--></body>
</html>
```

That doesn't seem right?

schlessera:
Maybe I messed it up during the conflict resolution. I'll have a look at it.

schlessera:
So the test case for `Dom\Document` covers this case and passes. This means it would be caused somewhere else.

Can you confirm that you're testing the exact PR in its latest state, and that your `vendor` folders point to the right files?

schlessera:
Nevermind, I can confirm it happens on my site too.

schlessera:
It's caused by this piece of code: https://github.com/ampproject/amp-wp/blob/1.4.4/includes/class-amp-theme-support.php#L2281-L2295

westonruter:
In that case I think what you have may be correct. The important thing is maintaining relative position to other markup. So if there is some action that is done after the closing HTML tag, then the markup including comments should get moved inside the body with their order preserved.

includes/class-amp-theme-support.php
```diff
@@ -2183,17 +2183,22 @@ public static function prepare_response( $response, $args = [] ) {
...
+					$dom->body->appendChild( $next_sibling->firstChild );
 				}
-				$dom->removeChild( $dom->documentElement->nextSibling );
-			} else {
-				$dom->body->appendChild( $dom->documentElement->nextSibling );
+				$dom->removeChild( $next_sibling );
+			} elseif ( ! $next_sibling instanceof DOMComment ) {
+				$dom->body->appendChild( $next_sibling );
 			}
+			$next_sibling = $next_sibling->nextSibling;
```
westonruter: This turns out to not be quite right. If element are moved but not comment nodes, then the source stacks can get confused.

Consider a `footer.php` ending with:

```php
		<?php wp_footer(); ?>

	</body>
	<?php do_action( 'after_closing_body' ); ?>
</html>
<?php do_action( 'after_closing_html' ); ?>
```

And a plugin like this:

```php
<?php
/**
 * Plugin Name: Try After Actions
 */

add_action( 'after_closing_body', function () {
	echo '<script>after_closing_body</script>';
} );

add_action( 'after_closing_html', function () {
	echo '<script>after_closing_html</script>';
} );
```

At the moment this results in validation errors looking like this:

image

Notice the lack of source identified for the first.

So what I think needs to be done is _every_ node appearing after `</body>` and `</html>` should be moved to be appended to the `body`, regardless of the node type. This will ensure that the source stack comments maintain their relative position with the elements they annotate.

Ultimately it should rather go into `\AmpProject\Dom\Document::loadHTML()`

westonruter: Patch incoming…

westonruter: See e4f10d1 and 62fd552

lib/common/src/Dom/Document.php
```diff
@@ -573,19 +586,54 @@ private function normalizeDocumentStructure($content)
      */
     public function normalizeDomStructure()
     {
-        $head = $this->getElementsByTagName(Tag::HEAD)->item(0);
-        if (! $head) {
-            $this->head = $this->createElement(Tag::HEAD);
-            $this->insertBefore($this->head, $this->firstChild);
```
westonruter: This, like the `body` handling below, appears to have been a bug where it was appending the `head` to the document and not to the `documentElement`.

PR #4299, opened 2020-02-15
Load block styles via amp_post_template_head action

schlessera:
Load block styles in Reader mode

## Summary

This PR loads block styles in Reader mode via the `'amp_post_template_head'` action:

It also adds the theme support for `'wp-block-styles'`.

The styles are directly printed. Scripts are ignored, as they wouldn't be used in AMP anyway.

Fixes #4297

## Checklist

### DIFF
--- a/includes/amp-post-template-functions.php
+++ b/includes/amp-post-template-functions.php
@@ -57,10 +57,8 @@ function amp_post_template_add_fonts( $amp_template ) {
 
 /**
  * Add block styles for core blocks and third-party blocks.
- *
- * @param AMP_Post_Template $amp_template Template.
  */
-function amp_post_template_add_block_styles( $amp_template ) {
+function amp_post_template_add_block_styles() {
 
 	add_theme_support( 'wp-block-styles' );
 
--- a/includes/amp-post-template-functions.php
+++ b/includes/amp-post-template-functions.php
@@ -59,9 +59,7 @@ function amp_post_template_add_fonts( $amp_template ) {
  * Add block styles for core blocks and third-party blocks.
  */
 function amp_post_template_add_block_styles() {
-
 	add_theme_support( 'wp-block-styles' );
-
 	wp_common_block_scripts_and_styles();
 	wp_styles()->do_items();
 }
--- a/includes/amp-post-template-functions.php
+++ b/includes/amp-post-template-functions.php
@@ -57,6 +57,8 @@ function amp_post_template_add_fonts( $amp_template ) {
 
 /**
  * Add block styles for core blocks and third-party blocks.
+ *
+ * @since 1.5.0
  */
 function amp_post_template_add_block_styles() {
 	add_theme_support( 'wp-block-styles' );
--- a/includes/amp-post-template-functions.php
+++ b/includes/amp-post-template-functions.php
@@ -62,7 +62,7 @@ function amp_post_template_add_fonts( $amp_template ) {
  */
 function amp_post_template_add_block_styles( $amp_template ) {
 
-	current_theme_supports( 'wp-block-styles' );
+	add_theme_supports( 'wp-block-styles' );
 
 	wp_common_block_scripts_and_styles();
 	wp_styles()->do_items();
--- a/includes/amp-post-template-functions.php
+++ b/includes/amp-post-template-functions.php
@@ -62,7 +62,7 @@ function amp_post_template_add_fonts( $amp_template ) {
  */
 function amp_post_template_add_block_styles( $amp_template ) {
 
-	add_theme_supports( 'wp-block-styles' );
+	add_theme_support( 'wp-block-styles' );
 
 	wp_common_block_scripts_and_styles();
 	wp_styles()->do_items();
--- a/includes/amp-post-template-functions.php
+++ b/includes/amp-post-template-functions.php
@@ -62,7 +62,9 @@ function amp_post_template_add_fonts( $amp_template ) {
  */
 function amp_post_template_add_block_styles() {
 	add_theme_support( 'wp-block-styles' );
-	wp_common_block_scripts_and_styles();
+	if ( function_exists( 'wp_common_block_scripts_and_styles' ) ) {
+		wp_common_block_scripts_and_styles();
+	}
 	wp_styles()->do_items();
 }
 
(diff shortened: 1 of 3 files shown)


schlessera:
@westonruter What do you think of this approach? I think it would catch most if not all styles for core and third-party blocks. It should ignore scripts.

Also, I didn't find any preexisting tests for Reader mode and its templating engine when I went over the test files. Maybe I missed something?

schlessera:
I'll integrate it into the `amp_post_template_init_hooks`, that should be cleaner.

includes/amp-post-template-functions.php
```diff
@@ -15,6 +15,7 @@ function amp_post_template_init_hooks() {
 	add_action( 'amp_post_template_head', 'amp_print_schemaorg_metadata' );
 	add_action( 'amp_post_template_head', 'amp_add_generator_metadata' );
 	add_action( 'amp_post_template_head', 'wp_generator' );
+	add_action( 'amp_post_template_head', 'amp_post_template_add_block_styles' );
```
westonruter: Good idea to do this at this action so that it is added to sites that have customized the `html-start.php` template.

includes/amp-post-template-functions.php
```diff
@@ -54,6 +55,20 @@ function amp_post_template_add_fonts( $amp_template ) {
...
 
+/**
+ * Add block styles for core blocks and third-party blocks.
+ *
+ * @param AMP_Post_Template $amp_template Template.
+ */
+function amp_post_template_add_block_styles( $amp_template ) {
+	global $wp_styles;
+
+	current_theme_supports( 'wp-block-styles' );
```
westonruter:
```suggestion
	add_theme_supports( 'wp-block-styles' );
```

includes/amp-post-template-functions.php
```diff
@@ -54,6 +55,20 @@ function amp_post_template_add_fonts( $amp_template ) {
 	}
 }
 
+/**
+ * Add block styles for core blocks and third-party blocks.
+ *
+ * @param AMP_Post_Template $amp_template Template.
+ */
+function amp_post_template_add_block_styles( $amp_template ) {
+	global $wp_styles;
```
westonruter:
```suggestion
```

includes/amp-post-template-functions.php
```diff
@@ -54,6 +55,20 @@ function amp_post_template_add_fonts( $amp_template ) {
...
+
+	current_theme_supports( 'wp-block-styles' );
+
+	wp_common_block_scripts_and_styles();
+	$wp_styles->do_items();
```
westonruter:
```suggestion
	wp_styles()->do_items();
```

includes/amp-post-template-functions.php
```diff
@@ -54,6 +55,20 @@ function amp_post_template_add_fonts( $amp_template ) {
 	}
 }
 
+/**
+ * Add block styles for core blocks and third-party blocks.
+ *
+ * @param AMP_Post_Template $amp_template Template.
+ */
+function amp_post_template_add_block_styles( $amp_template ) {
```
westonruter: The `$amp_template` arg can be removed, no?

includes/amp-post-template-functions.php
schlessera: 🤦‍♂ 

Yesterday was definitely not my brightest day...

includes/amp-post-template-functions.php
schlessera: It can be removed, yes, although it is indeed being passed-in.

I'll remove it.

schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4299/commits/0baf7b90c13de89b585fd91fd274b260def57b6e

includes/amp-post-template-functions.php
```diff
@@ -54,6 +55,15 @@ function amp_post_template_add_fonts( $amp_template ) {
 	}
 }
 
+/**
+ * Add block styles for core blocks and third-party blocks.
```
westonruter:
```suggestion
 * Add block styles for core blocks and third-party blocks.
 *
 * @since 1.5.0
```

westonruter:
The block styles greatly improve the presentation of the theme unit test data for posts with blocks. 

I looked at each and the one that seems not quite right is the Cover Block post. Notice the lack of a margin:

image

It should rather look like:

image

Also, what if the theme adds `add_theme_support( 'align-wide' )` for [alignwide and alighfull](https://developer.wordpress.org/block-editor/developers/themes/theme-support/#wide-alignment).  Should this theme support get removed?

schlessera:
I can't see a difference between your two screenshots, you've probably uploaded the same one twice (or am just blind).

schlessera:
Do you suggest we add specific styling to the Reader mode to make the core blocks look nice?

We could add basic margins and padding to the block-level wrapper elements.

westonruter:
> I can't see a difference between your two screenshots, you've probably uploaded the same one twice (or am just blind).

My apologies. Fixed.

westonruter:
> Do you suggest we add specific styling to the Reader mode to make the core blocks look nice?
> 
> We could add basic margins and padding to the block-level wrapper elements.

The other blocks have margins as expected. There just seems to be something about the Cover block. I'm not an expert on theme block styles. I'm not familiar what what themes are expected to include in terms of styles on their own and which styles should be considered to make block styling “just work” by enqueueing the core styles.

schlessera:
I'll have a look to see if there's an easy fix, or whether it might be more involved.

westonruter:
Aside: The block stylesheet added in the Syntax Highlighting Code Block plugin will be successfully added now in Reader mode after https://github.com/westonruter/syntax-highlighting-code-block/pull/35.

Themes that register/enqueue stylesheets at `enqueue_block_assets` will just work. Or rather, if the stylesheet is registered at `init` as shown at https://developer.wordpress.org/block-editor/tutorials/block-tutorial/applying-styles-with-stylesheets/ then this will work as well.

schlessera:
@westonruter I added a few basic styles to keep margins intact with cover blocks. The reason why these weren't working as expected was that they don't use an `amp-img` (which is targeted by the styling we need), but rather use a `background` property on a `div`.

> Also, what if the theme adds add_theme_support( 'align-wide' ) for alignwide and alighfull. Should this theme support get removed?

I also added basic support for `alignwide` and `alignfull`. If you look at the test page you were referring to, you can see that `alignfull` always covers the full width of the screen, while `alignwide` grows beyong the article width while respecting the page border, up to a certain maximum.

Let me know what you think.

westonruter:
Scribd is currently returning with `503 Service Unavailable` so that is why the external-http requests are failing.

Issue #2062, opened 2019-04-04
Consider automatically redirecting /amp/ to ?amp in case of 404


westonruter:
As reported in a [support forum topic](https://wordpress.org/support/topic/how-to-change-url-structure-2/):

> I tried to find solution for my problem but unable to.
> I want to change the URL format from
> https://hinditips.com/best-ayurvedic-tips-for-hair-growth/?amp
> to
> https://hinditips.com/best-ayurvedic-tips-for-hair-growth/amp/
>
> How can it be done.
> 
> Because the previous plugin that I have used has the same structure. Now i switched to this plugin. Search engines may find those old URLs as 404.

The alternative to this I suggested is actually there should just be a redirect from `/amp/` to `?amp` in the case of 404, such as when migrating to this AMP plugin. 

The `amp` query var works always works for serving AMP regardless of the mode and it always avoids returning a 404 when the plugin gets deactivated. This also ensures that AMP requests don't fail when the rewrite rules have not been flushed. It is overall better to use `?amp` instead of `/amp/`.

Proof of concept plugin: https://gist.github.com/westonruter/6b7d3b168711ee03c4ee29b68c3719b2

----

When using another plugin that tries to use `/amp/` for all URLs, switching to this AMP plugin can result in non-singular URLs 404ing. To deal with that, I created a helper migration plugin: https://gist.github.com/westonruter/6b7d3b168711ee03c4ee29b68c3719b2

Nevertheless, this seems like something that would make sense to be part of the the AMP plugin directly. In particular, it would make sense to add after #2204 is figured out.

----

A related issue came up in https://github.com/ampproject/amp-wp/issues/474#issuecomment-609764881.

Namely, if someone originally had `/amp/` for pages, upon switching to the AMP plugin these URLs won't work anymore and they may end up getting routed to WordPress's own canonical redirect system, so in that case `/some-page/amp/` will redirect to `/some-page/` and not `/some-page/?amp` as expected. While we don't want to register an `amp` endpoint for pages, we should perhaps hook into WordPress's canonical redirects to check for `/amp/` being present, and make sure that WordPress will redirect to `?amp` (or `?amp=1` per #2204) instead of removing it from the URL altogether.

westonruter:
**QA Passed**

I set the Template Mode to Transitional and the Paired URL Structure to “Query parameter”. 

I then tried going to `/sample-page/amp/` and I was redirected to `/sample-page/?amp=1`.

[...]

PR #4312, opened 2020-02-18
Default to ?amp=1 query parameter

### LINKED ISSUE #2204: Default to ?amp=1 query param instead of /amp/ endpoint slug in all scenarios (but persist legacy config and allow customization)
## Feature description

We should consider eliminating the `/amp/` in favor of `?amp` everywhere. The endpoint slug is unnecessary for SEO, and it improve consistency when switching between the paired modes (transitional and reader). Additionally, there will no longer be a need to flush rewrite rules.

See also #2062 to automatically redirect from `/amp/` to `?amp` when `is_404()`.

See also #1390 where `?amp=1` should be considered instead of `?amp`.

See also #1389 where the AMP plugin should automatically convert all URLs in AMP pages to point to other AMP pages in Transitional mode.

In order to prevent breaking analytics, SEO, and other things… these changes would need to not happen for existing sites. They should only apply to new plugin installs. Furthermore, the specific AMP URL format should be configurable (yes, including the allowing of a site to use `/amp/`).

Also relates to #2817 where there could be a filter added to `is_amp_endpoint()` to determine if a request is AMP or not. This would be a necessary corollary to allowing the URL for AMP pages to be arbitrary filterable: there would need to be a way to reverse the process.

Example AMP URLs that should be supported:

Type | Example
-----|---------
Endpoint | https://www.example.com/…/foo/amp/
Root directory | https://www.example.com/amp/…/foo/
File extension | https://www.example.com/…/foo.amp
Bare query var | https://www.example.com/…/foo/?amp
Query var with value | https://www.example.com/…/foo/?amp=1
Subdomain | https://amp.example.com/…/foo/

In each case, the `amp` corresponds to whatever is returned by `amp_get_slug()`.

---------------

## Acceptance criteria

1.  For new plugin installs, replace all instances of `/amp/` in favor of `?amp` 
2.  The specific AMP URL format should be configurable (yes, including the allowing of a site to use `/amp/`).

## Implementation brief

* ~Let `amp_add_paired_endpoint()` allow the `$url` to be omitted, and when this is done, pass either the current `WP_Query` (if available) or the `WP_Post` (if available) for the URL in question. Or else allow for the `WP_Query` or `WP_Post` to be added for every occurance of the aforementioned functions?~
* Introduce a new `PairedAmpRouting` service.
  * Introduce a new AMP option for the AMP `permalink_structure`.
  * Move the rewrite endpoint registration from `amp_init()` to this service.
* Add a new settings drawer for “Paired URL Structure” which is shown only when Reader mode or Transitional mode is selected. This is similar to the Permalinks admin screen in WordPress core. Four options are available:
  * Query var: `?amp=1` everywhere (the **default** for new installs).
  * Endpoint: `/amp/` whenever possible (or else `?amp=1`).
  * Legacy Reader: `/amp/` for posts and `?amp` for everything else (default for sites upgraded when in legacy Reader mode).
  * Legacy Transitional: `?amp` for everything (default for sites upgraded when in Transitional mode or non-legacy Reader mode).
  * Custom: This will not be an option _per se_, but if the site has a given filter in place to modify the AMP URL structure, then the other options will not be available and this will be the forced option.
* Update these functions which manipulate AMP URLs to invoke corresponding methods in the service:
  * `amp_add_paired_endpoint()`
  * `amp_remove_paired_endpoint()`
  * `amp_has_paired_endpoint()`

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

### LINKED ISSUE #4442: AMP Helper Function Chooses Not To Append Query Value - ($use_query_var) Evaluates To False
## Bug Description

On `amp-helper-functions.php` in the `amp_get_permalink()` function the `$use_query_var` conditional evaluates to false where I would expect it should evaluate to true. I have a custom permalink structure using the permalink settings in WordPress - `/%category%/%postname%/`. I don't understand why the `$use_query_var` conditional is structured in this way, as using `/amp` instead of `/?amp` can cause WordPress redirection issues - it thinks `amp` is part of the post slug and redirects it to a post (the wrong post - and not AMP). Oddly, this bug only occurred for me on a multisite, where the multisite  was appended to the url as a directory, so like this:

`/my-multisite-identifier/%category%/%postname%/amp` - causes redirect issue

This broke the amp link and caused redirection issues as described. To be clear, however - the `/amp` was occurring on my main site too - it just only caused a redirect issue on the site that has its multisite identifier (url string specific to that multisite) as part of the url. I manually removed the `$use_query_var` conditional to make it always use a query var and this was always reliable and works for me across all my sites in the network - but this modification will be lost on plugin update. So my question is why don't we always use the query var or can we do it in a way that still fires for my use case, as it seems appropriate (and working) there.

## Expected Behaviour

I expected the plugin to use a query var not append `/amp` without the query string character `?`.

## Steps to reproduce

Issue can't be seen anymore because I fixed it as it was breaking my site and causing search console issues. But if you run a multisite (not a subdomain) with my url scheme it should reproduce for you on the multisite.

## Additional context

Thank you very much for all that you do and for your consideration.

 - WordPress version: Version 5.3.2
 - Plugin version: Version 1.4.4
 - Gutenberg plugin version (if applicable): N/A
 - AMP plugin template mode: Reader
 - PHP version: PHP 7.2.11
 - OS: Windows and Amazon AMI 
 - Browser: All (Server-side Issue)
 - Device: All (Server-side Issue)

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

Closes #4442
Relates to #2204

This PR enforces the `?amp=1` query parameter throughout the plugin.


includes/class-amp-theme-support.php
```diff
@@ -494,30 +490,18 @@ public static function ensure_proper_amp_location( $exit = true ) {
...
-			}
-		} elseif ( $has_query_var && ! $has_url_param ) {
+		} elseif ( ! $is_amp_url ) {
 			/*
-			 * When in AMP transitional mode *with* theme support, then the proper AMP URL has the 'amp' URL param
-			 * and not the /amp/ endpoint. The URL param is now the exclusive way to mark AMP in transitional mode
-			 * when amp theme support present. This is important for plugins to be able to reliably call
+			 * When in AMP reader or transitional mode *with* theme support, then the proper AMP URL has the 'amp' URL
+			 * param and not the /amp/ endpoint. The URL param is now the exclusive way to mark AMP in reader mode &
+			 * transitional mode when amp theme support present. This is important for plugins to be able to reliably call
```
schlessera:
```suggestion
			 * transitional mode when amp theme support is present. This is important for plugins to be able to reliably call
```

includes/class-amp-theme-support.php
```diff
@@ -483,8 +480,7 @@ static function() {
 	 * @return bool Whether redirection was done. Naturally this is irrelevant if $exit is true.
 	 */
 	public static function ensure_proper_amp_location( $exit = true ) {
-		$has_query_var = false !== get_query_var( amp_get_slug(), false ); // May come from URL param or endpoint slug.
-		$has_url_param = isset( $_GET[ amp_get_slug() ] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$is_amp_url = is_amp_url();
```
schlessera: No need really to store this as a variable here.
As both calls of the function are within mutually exclusive conditionals, you can just call the function when needed.
```suggestion
```

includes/class-amp-theme-support.php
```diff
@@ -494,30 +490,18 @@ public static function ensure_proper_amp_location( $exit = true ) {
 			 * should happen infrequently. For admin users, this is kept temporary to allow them
 			 * to not be hampered by browser remembering permanent redirects and preventing test.
 			 */
-			if ( $has_query_var || $has_url_param ) {
+			if ( $is_amp_url ) {
```
schlessera:
```suggestion
			if ( is_amp_url() ) {
```

includes/class-amp-theme-support.php
```diff
@@ -494,30 +490,18 @@ public static function ensure_proper_amp_location( $exit = true ) {
...
-			wp_parse_str( $wp->matched_query, $path_args );
-			if ( isset( $path_args[ amp_get_slug() ] ) && '' !== $path_args[ amp_get_slug() ] ) {
-				wp_safe_redirect( amp_get_permalink( get_queried_object_id() ), 301 );
-				if ( $exit ) {
-					exit;
-				}
-				return true;
-			}
-		} elseif ( $has_query_var && ! $has_url_param ) {
+		} elseif ( ! $is_amp_url ) {
```
schlessera:
```suggestion
		} elseif ( ! is_amp_url() ) {
```

includes/amp-helper-functions.php
```diff
@@ -1360,6 +1350,73 @@ function amp_generate_script_hash( $script ) {
...
+		 * @since 1.5.0
+		 *
+		 * @param false  $is_amp_url Whether current URL is AMP or not.
+		 * @param string $url        Current URL for the request.
+		 * @param string $amp_slug   AMP slug.
+		 */
+		return true === apply_filters( 'is_amp_url', false, amp_get_current_url(), amp_get_slug() );
+	}
+
+	return isset( $_GET[ amp_get_slug() ] ) && 1 === intval( $_GET[ amp_get_slug() ] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
schlessera: I usually prefer the PHP filtering functions for dealing with `$_GET` and `$_POST`. I think this woud also remove the need for the PHPCS ignore rule.

```suggestion
	return 1 === (int) filter_input( INPUT_GET, amp_get_slug(), FILTER_SANITIZE_NUMBER_INT );
```

includes/amp-helper-functions.php
```diff
@@ -1360,6 +1350,73 @@ function amp_generate_script_hash( $script ) {
...
+
+	return add_query_arg( amp_get_slug(), 1, $url );
+}
+
+/**
+ * Set the AMP endpoint.
+ */
+function amp_set_rewrite_rule() {
+	if ( is_amp_url_overridden() ) {
+		do_action( 'amp_set_rewrite_rule', amp_get_slug() );
```
schlessera: The action is missing documentation here.

includes/amp-helper-functions.php
pierlon: Ah, TIL about `filter_input`!

westonruter:
@schlessera in regards to:

> I usually prefer the PHP filtering functions for dealing with $_GET and $_POST. I think this woud also remove the need for the PHPCS ignore rule.

How do you unit test these? AFAIK you can't manipulate the underlying data source that `filter_input()` reads from.

pierlon:
Starting this PR anew; this will focus on enforcing the `?amp=1` query parameter, and will then be followed by a separate PR focusing on the customization of AMP URLs.

includes/amp-helper-functions.php
```diff
@@ -1943,3 +1905,26 @@ function amp_generate_script_hash( $script ) {
...
+/**
+ * Get the AMP version of a URL.
+ *
+ * @since 2.1
+ *
+ * @param string $url URL.
+ * @return string AMP URL.
+ */
+function amp_get_url( $url ) {
+	return add_query_arg( amp_get_slug(), 1, $url );
```
westonruter: Minor thing:

```suggestion
	return add_query_arg( amp_get_slug(), '1', $url );
```

includes/amp-helper-functions.php
```diff
@@ -1943,3 +1905,26 @@ function amp_generate_script_hash( $script ) {
...
+
+/**
+ * Determine whether the current request has the AMP query parameter set with the required value '1'.
+ *
+ * @since 2.1
+ *
+ * @return bool True if the AMP query parameter is set with the required value, false if not.
+ */
+function amp_has_query_var() {
+	return isset( $_GET[ amp_get_slug() ] ) && 1 === filter_var( $_GET[ amp_get_slug() ], FILTER_VALIDATE_INT ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
westonruter: I don't think the value should need to be _required_ to be `1`. It is just `1` as a generic non-empty value.

```suggestion
	return isset( $_GET[ amp_get_slug() ] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```

includes/amp-helper-functions.php
```diff
@@ -1943,3 +1905,26 @@ function amp_generate_script_hash( $script ) {
...
+ *
+ * @param string $url URL.
+ * @return string AMP URL.
+ */
+function amp_get_url( $url ) {
+	return add_query_arg( amp_get_slug(), 1, $url );
+}
+
+/**
+ * Determine whether the current request has the AMP query parameter set with the required value '1'.
```
```suggestion
 * Determine whether the current request has the AMP query parameter set.
```

includes/class-amp-theme-support.php
```diff
@@ -411,23 +411,23 @@ public static function ensure_proper_amp_location() {
 			global $wp;
 			$path_args = [];
 			wp_parse_str( $wp->matched_query, $path_args );
-			if ( isset( $path_args[ amp_get_slug() ] ) && '' !== $path_args[ amp_get_slug() ] ) {
+			if ( isset( $path_args[ amp_get_slug() ] ) && '1' !== $path_args[ amp_get_slug() ] ) {
```
westonruter: I don't think this change is quite right. This code here is asking if the URL looks like this `/foo/amp/bar/`, where there is _something_ after the `/amp/` (like for feeds there is `/feed/rss/` and `/feed/atom/`). The `$path_args` is _not_ the query vars from `$_GET` but rather wat was matched from the rewrite rule. So for a request `/foo/amp/bar/`, here `$path_args[ amp_get_slug() ]` would be set to `bar`. If it is any non-empty string value, then this indicates a redirect is needed to avoid serving an infinite URL space.

westonruter (changes requested):

includes/class-amp-theme-support.php
westonruter: So this change should be reverted.

```suggestion
			if ( isset( $path_args[ amp_get_slug() ] ) && '' !== $path_args[ amp_get_slug() ] ) {
```

pierlon: Yes that makes sense. I've made the changes in https://github.com/ampproject/amp-wp/pull/4312/commits/83ec4a869442e89fda638f7bd408f6db7b3458c2.

includes/amp-helper-functions.php
```diff
@@ -1983,3 +1947,39 @@ function amp_generate_script_hash( $script ) {
...
+/**
+ * Get the AMP version of a URL.
+ *
+ * @since 2.1
+ * @todo Rename this to clear up confusion with amp_get_current_url().
+ *
+ * @param string $url URL.
+ * @return string AMP URL.
+ */
+function amp_get_url( $url ) {
```
westonruter: We have `amp_get_current_url()` and `amp_get_asset_url()` and `amp_get_customizer_url()`. These being here, I think `amp_get_url()` is too ambiguous. 

How about we call it `amp_get_paired_url()` since its purpose is to get the paired AMP URL for the current page. 

We also have a `\AMP_Theme_Support::get_current_canonical_url()` which could be extracted into a separate global function called `amp_get_canonical_url()`.

includes/amp-helper-functions.php
```diff
@@ -1983,3 +1947,39 @@ function amp_generate_script_hash( $script ) {
...
+/**
+ * Determine whether the current request has the AMP query parameter set.
+ *
+ * @since 2.1
+ * @todo Rename this to be more agnostic to what the URL contains.
+ *
+ * @return bool True if the AMP query parameter is set with the required value, false if not.
+ * @global WP_Query $wp_query
+ */
+function amp_has_query_var() {
```
westonruter: Given that in the future a paired AMP URL will be able to be indicated arbitrarily, such as via a subdomain, I think we can anticipate that by making `amp_has_query_var()` more abstract. 

We have `amp_is_request()` to indicate that the request is being made for an AMP page. What we need is a method that says that the request was made for a paired AMP URL. What about `amp_is_paired_request()`? I'm not entirely satisfied with that since it implies that it entails `amp_is_request()`, being more specific. It's actually just giving an indication that the current URL is for paired AMP.

Maybe `amp_is_paired_url()`? We could pass it a `$url` argument as well, and it could tell you if _that_ URL is paired rather than the current request to WP. This would be useful in other contexts, like in the AMP-to-AMP Linking sanitizer perhaps.

includes/amp-helper-functions.php
westonruter: Done in b66c233fa

includes/amp-helper-functions.php
westonruter: Thinking about this some more… what about `amp_is_paired_endpoint`? Using “endpoint” terminology would align with what we already have in `amp_remove_endpoint()` (which we could technically rename to `amp_remove_paired_endpoint()`).

While in a WordPress context, an “endpoint” refers to the _end_ of the URL, in reality an API endpoint doesn't necessarily refer to any physical part of the URL. Normally the endpoint will still be at the end (`?amp=1`) but sites can customize the paired AMP endpoint URLs to look however they want.

westonruter: I suppose then we'd rename `amp_get_paired_url( $url )` to `amp_get_paired_endpoint( $url )`.

westonruter: We basically need CRUD functions for manipulating paired AMP URLs:

* Read: `has`/`is`
* Create/Update: `get`
* Delete: `remove`

westonruter: So:

* Read: `amp_is_paired_endpoint( $url )`
* Create/Update: `amp_get_paired_endpoint( $url )`
* Delete: `amp_remove_paired_endpoint( $url )` <em>(formerly `amp_remove_endpoint()`)</em>

Thoughts?

pierlon: > Using “endpoint” terminology would align with what we already have in amp_remove_endpoint() (which we could technically rename to amp_remove_paired_endpoint()).

That reasoning makes sense. I'm also in support of renaming the functions to represent the CRUD methods of manipulating the paired URL :+1:.

Sidenote: That means `amp_remove_endpoint()` would have to be deprecated and annotated accordingly since its documented.

westonruter: Here, check this out: d5610f2

tests/php/test-amp-helper-functions.php
```diff
@@ -494,13 +495,16 @@ public function test_amp_get_permalink_without_pretty_permalinks() {
...
+			remove_filter( 'amp_get_permalink', [ $this, 'return_example_url' ] );
+			$this->assertStringEndsWith( '&amp=1', amp_get_permalink( $published_post ) );
+			$this->assertStringEndsWith( '&amp=1', amp_get_permalink( $drafted_post ) );
+			$this->assertStringEndsWith( '&amp=1', amp_get_permalink( $published_page ) );
 			add_filter( 'amp_get_permalink', [ $this, 'return_example_url' ], 10, 2 );
-			$this->assertStringNotContains( 'current_filter=amp_get_permalink', amp_get_permalink( $published_post ) ); // Filter does not apply.
+			$this->assertStringEndsWith( 'current_filter=amp_get_permalink', amp_get_permalink( $published_post ) );
 			add_filter( 'amp_pre_get_permalink', [ $this, 'return_example_url' ], 10, 2 );
-			$this->assertStringNotContains( 'current_filter=amp_pre_get_permalink', amp_get_permalink( $published_post ) ); // Filter does not apply.
+			$this->assertStringEndsWith( 'current_filter=amp_pre_get_permalink', amp_get_permalink( $published_post ) );
```
westonruter, line 507: Nice. Looks like you found a bug in the test here.

westonruter (approved):
> Fixes #2062

It appears that this does not actually fix that issue. 

With Legacy AMP Reader mode active, I tried going to `/category/uncategorized/amp/` and I was taken to a 404 not being redirected to `/category/uncategorized/?amp=1` as I expected.

@pierlon That being the case, should we just remove that `Fixes` to address in another PR?

pierlon:
> @pierlon That being the case, should we just remove that Fixes to address in another PR?

Yes let's do that. I'll update the PR description.

Also, reviewed the recent changes and it's a :+1: from me.

schlessera:
I find the naming to be confusing, for the reason that `get` usually represents the READ of CRUD, not the CREATE/UPDATE.

I don't think it's a bad idea to use `is`, `get` & `remove`, necessarily, but it confuses when referred to as CRUD operations.

westonruter:
> I don't think it's a bad idea to use `is`, `get` & `remove`, necessarily, but it confuses when referred to as CRUD operations.

OK, that makes sense. In any case, there is no reference to CRUD in the codebase.

schlessera:
Alright. CRUD really is about manipulating entities. That is not the case here.

That being said, I think the names are good and make sense to me.

schlessera:
I wonder whether it is clear for users, though...

I was just now thinking about the following constellation:

```
amp_is_paired_url()
amp_get_paired_url()
amp_get_source_url()
```

westonruter:
That would work, though `amp_get_source_url()` I'm not sure about, in regards to the “source”. In particular, it would make more sense to use `amp_get_canonical_url()` since the URL returned would correspond with the `link[rel=canonical]` on the AMP page. (But then this may not actually correspond with the `link[rel=canonical]` on the canonical page.)

schlessera:
Yes, the following would work:
```
amp_is_paired_url()
amp_get_paired_url()
amp_get_canonical_url()
```

I guess I have an issue with the `remove_endpoint` bit, as it sounds like disabling something or deleting something. By having two `get_*()` variations, it makes it clear that you're getting a URL in both cases, it's just that we happen to have two variations of that URL.

westonruter:
My only concern with `amp_get_canonical_url()` is users may expect that its return value will be the same as what is returned by `wp_get_canonical_url()` but this won't necessarily be the case. Really it's about getting the “_unpaired_” URL.

westonruter:
Here's an alternative suggestion: 

* `amp_has_paired_endpoint( $url )` (instead of `is`)
* `amp_add_paired_endpoint( $url )` (instead of `get`)
* `amp_remove_paired_endpoint( $url )`

westonruter:
Note we currently have `amp_remove_endpoint()`.

westonruter:
cf. `add_query_arg()` and `remove_querg_arg()` in core.

westonruter:
I'm happy with the naming but we can adjust further in a subsequent PR.

PR #4300, opened 2020-02-16
Require unreleased version of PHP-CSS-Parser

pierlon:
## Summary

This PR updates the version of [PHP-CSS-Parser](/sabberworm/PHP-CSS-Parser) to the commit https://github.com/sabberworm/PHP-CSS-Parser/commit/134f4e62fe8ab9f316425f3c0f480b3f4f52d804, instead of the latest release version [8.3.0](https://github.com/sabberworm/PHP-CSS-Parser/releases/tag/8.3.0). The commit aforementioned includes a fix that would resolve #4197.

A patch is also included for PHP-CSS-Parser to fix the validation of selectors, namely adding `,`, `\` and `@` as valid characters.

Also, I've used the [undocumented flag `no-api`](https://github.com/composer/composer/issues/5433) to instruct Composer to retrieve the repository via Git instead of using the GitHub API, due to a rate limit on their API that usually occurs when `composer update` is run. The alternative to not using `no-api` is giving Composer your GitHub credentials or creating a GitHub OAuth token, as described [here](https://getcomposer.org/doc/articles/troubleshooting.md#api-rate-limit-and-oauth-tokens).
  

Fixes #4197.

## Checklist

- [x] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [x] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [x] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

PR #4324, opened 2020-02-24
Prevent any vendor .git directories from being added to build

westonruter:
I noticed when doing `npm run build` that the `build` directory was unexpectedly getting `vendor/sabberworm/php-css-parser/.git` added to it. This PR prevents that from happening.

This may be related to #4300.


westonruter:
@pierlon Good point. Done in f104458fc.

PR #4339, opened 2020-02-27
Don't tree-shake allowed child classes of `amp-date-picker` if it's in the DOM

### LINKED ISSUE #4334: Tree shaker removes CSS style rules for amp-date-picker classes
## Bug Description

As reported in the [support forum](https://wordpress.org/support/topic/amp-datepicker-unable-to-appy-custom-style-css/), styling the `amp-date-picker` component is difficult because style rules that target classes which the component adds via JS are removed by the tree-shaker. For example:

```css
amp-date-picker .CalendarMonth_caption {  
  background-image: linear-gradient(#b8b8b8, #5e5e5e);
  padding: 12px 0px;
}
amp-date-picker .amp-date-picker-calendar-container .CalendarMonth_caption {
  background-image: linear-gradient(#b8b8b8, #5e5e5e);
  padding: 12px 0px;
}
.amp-date-picker-calendar-container .CalendarMonth_caption {
  background-image: linear-gradient(#b8b8b8, #5e5e5e);
  padding: 12px 0px;	
}
```

A workaround is to ensure that all CSS selectors have an `amp-date-picker` ancestor prefix, and then to add this plugin code:

```php
// Temporary workaround for amp-date-picker class names having CSS selectors being tree-shaken.
add_filter( 'amp_content_sanitizers',
	function ( $sanitizers ) {
			! empty( $sanitizers['AMP_Style_Sanitizer']['dynamic_element_selectors'] ) ? $sanitizers['AMP_Style_Sanitizer']['dynamic_element_selectors'] : [],
			[
				// Modified from from protected AMP_Style_Sanitizer::$DEFAULT_ARGS.
				'amp-list',
				'amp-live-list',
				'[submit-error]',
				'[submit-success]',
				'amp-script',

				// New.
				'amp-date-picker',
			]
		);

		return $sanitizers;
	}
);
```

## Expected Behaviour

If an `amp-date-picker` is on the page, all of the class names that can be used in an `amp-date-picker` should be considered exempt from tree-shaking.

## Steps to reproduce

Create a Custom HTML block with the following contents:

```html
<style>
amp-date-picker .CalendarMonth_caption {  
  background-image: linear-gradient(#b8b8b8, #5e5e5e);
  padding: 12px 0px;
}
amp-date-picker .amp-date-picker-calendar-container .CalendarMonth_caption {
  background-image: linear-gradient(#b8b8b8, #5e5e5e);
  padding: 12px 0px;
}
.amp-date-picker-calendar-container .CalendarMonth_caption {
  background-image: linear-gradient(#b8b8b8, #5e5e5e);
  padding: 12px 0px;	
}
</style>
<amp-date-picker layout="fixed-height" height="360"></amp-date-picker>
```

View the AMP version of the page and notice that the date picker header does not have the expected gradient.

## Screenshots

Expected:

image

Actual:

image

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

kienstra:
## Summary
Prevents tree-shaking of valid child classes of `amp-date-picker`, if that tag is in the document.

When adding the markup and styling that Weston [shared](https://github.com/ampproject/amp-wp/issues/4334#issue-570785294) to a Custom HTML block:

```html
<style>
amp-date-picker .CalendarMonth_caption {  
  background-image: linear-gradient(#b8b8b8, #5e5e5e);
  padding: 12px 0px;
}
amp-date-picker .amp-date-picker-calendar-container .CalendarMonth_caption {
  background-image: linear-gradient(#b8b8b8, #5e5e5e);
  padding: 12px 0px;
}
.amp-date-picker-calendar-container .CalendarMonth_caption {
  background-image: linear-gradient(#b8b8b8, #5e5e5e);
  padding: 12px 0px;	
}
</style>
<amp-date-picker layout="fixed-height" height="360"></amp-date-picker>
```

...it's not tree-shaken:

date-picker-not-tree-shaken

Fixes #4334

## Checklist


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -755,6 +761,41 @@ private function has_used_attributes( $attribute_names ) {
...
+	 * @since 1.5.0
+	 * @link https://github.com/airbnb/react-dates/tree/05356/src/components
+	 *
+	 * @param string $class The name of the class to evaluate.
+	 * @return bool Whether the class is allowed as a child of <amp-date-picker>.
+	 */
+	private function is_class_allowed_in_amp_date_picker( $class ) {
+		// Some prefixes, some full class names.
+		$allowed_class_beginnings = [
+			'CalendarDay',
```
kienstra: This might be missing some classes, though I tried to be thorough.

I looked at the [components](https://github.com/airbnb/react-dates/tree/05356400baa8e05a9d7ea17e74fbb04bb3b23fa8/src/components) that have styles. 

This doesn't include components that are just an `.svg`, like [CalendarIcon](https://github.com/airbnb/react-dates/blob/05356400baa8e05a9d7ea17e74fbb04bb3b23fa8/src/components/CalendarIcon.jsx).

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -755,6 +761,41 @@ private function has_used_attributes( $attribute_names ) {
...
+			'CalendarMonth',
+			'DateInput',
+			'DateRangePicker',
+			'DayPicker',
```
kienstra: These classes could be more specific, though that would mean a longer array.

For example, instead of just `DayPicker`, this array could have `DayPicker_`, `DayPickerKeyboardShortcuts_`, `DayPickerNavigation_`, etc...

This will of course still accept those, though it'll also accept something like `DayPickerRandomIncorrect`.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -755,6 +761,41 @@ private function has_used_attributes( $attribute_names ) {
...
+			'KeyboardShortcutRow',
+			'SingleDatePicker',
+			'amp-date-picker-calendar-container',
```
kienstra: If needed, this check for class names could be more specific. For example, this class isn't a prefix, it should be an exact match of the class.

Most other classes in this array are prefixes. For example, `DayPicker` is a prefix of `DayPicker_focusRegion`.

kienstra: I'm going to make a commit for this.

kienstra: Applied in https://github.com/ampproject/amp-wp/pull/4339/commits/28d037e1d92c5f1404e53adbaa951f237622ac75

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -755,6 +761,49 @@ private function has_used_attributes( $attribute_names ) {
...
+			'CalendarMonth',
+			'DateInput',
+			'DateRangePicker',
+			'DayPicker',
+			'KeyboardShortcutRow',
+			'SingleDatePicker',
+		];
+
+		foreach ( $class_prefixes as $prefix ) {
+			if ( 0 === strpos( $class, $prefix ) ) {
```
kienstra: I'm working on making this into a `preg_match()`.

kienstra: Applied in https://github.com/ampproject/amp-wp/pull/4339/commits/bd9313d462aca486423006824ff6725beac8df0d

kienstra:
Hi @westonruter,
No hurry...could you please review this when you have a chance? Thanks!

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -755,6 +761,36 @@ private function has_used_attributes( $attribute_names ) {
...
+			'CalendarDay',
+			'CalendarMonth',
+			'DateInput',
+			'DateRangePicker',
+			'DayPicker',
+			'KeyboardShortcutRow',
+			'SingleDatePicker',
+		];
+
+		if ( preg_match( '#^(' . implode( '|', $class_prefixes ) . ')[a-zA-Z_]+#', $class ) ) {
```
westonruter: This can be made simpler/faster like so:

```suggestion
		if ( in_array( $class_prefixes, strtok( $class, '_' ), true ) ) ) {
```

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -755,6 +761,36 @@ private function has_used_attributes( $attribute_names ) {
...
+			return true;
+		}
+
+		return in_array( $class, [ 'amp-date-picker-calendar-container', 'amp-date-picker-resize-bug' ], true );
```
westonruter:
```suggestion
		return 'amp-date-picker-' === substr( $class, 0, 16 );
```

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: Do any of these component names get carried over to be class names in the DOM? If so, they should be included in this list.

includes/sanitizers/class-amp-style-sanitizer.php
kienstra: Committed in https://github.com/ampproject/amp-wp/pull/4339/commits/36cba7d67a34f61ef97f0e93322ec00567251529

includes/sanitizers/class-amp-style-sanitizer.php
kienstra: Yes, all of these appear as names in the DOM except for `DateInput`, `DateRangePicker`, `DateRangePicker`:

amp-date-picker

For those 3 that didn't appear, I couldn't find them, when using most of the snippets in the [amp-date-picker](https://amp.dev/documentation/examples/components/amp-date-picker/) documentation. I should probably remove those, though maybe I just missed a case where they are used.

includes/sanitizers/class-amp-style-sanitizer.php
kienstra: Reverted, then re-committed in https://github.com/ampproject/amp-wp/pull/4339/commits/5bffd65ecba76846919095ea883c93d6df20954a

includes/sanitizers/class-amp-style-sanitizer.php
kienstra: The 3 that I couldn't find in the DOM as class names are removed in https://github.com/ampproject/amp-wp/pull/4339/commits/5bffd65ecba76846919095ea883c93d6df20954a

includes/sanitizers/class-amp-style-sanitizer.php
kienstra: They're now all specific with https://github.com/ampproject/amp-wp/pull/4339/commits/5bffd65ecba76846919095ea883c93d6df20954a, like `DayPickerNavigation` instead of only `DayPicker`

includes/sanitizers/class-amp-style-sanitizer.php
kienstra: Committed in https://github.com/ampproject/amp-wp/pull/4339/commits/eea565f5a78d441428e049073354b40ae94d6399

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -755,6 +761,36 @@ private function has_used_attributes( $attribute_names ) {
...
+	 */
+	private function is_class_allowed_in_amp_date_picker( $class ) {
+		$class_prefixes = [
+			'CalendarDay',
+			'CalendarMonth',
+			'CalendarMonthGrid',
+			'DayPicker',
+			'DayPickerKeyboardShortcuts',
+			'DayPickerNavigation',
+			'KeyboardShortcutRow',
```
westonruter: I'm using this function in the console:

```js
function gatherClassNames( elements ) {
	const classNames = new Set();
	const add = ( className ) => {
		if ( !/^i-amphtml-/.test( className ) ) {
			classNames.add( className.replace( /_.+/, '' ) );
		}
	};
	for ( const element of elements ) {
		element.classList.forEach( add );
		gatherClassNames( element.children ).forEach( add );
	}
	return classNames;
}
```

And I'm invoking it via:

```js
gatherClassNames(document.querySelectorAll('amp-date-picker'))
```

I'm not able to find any other class names in use when looking at the `amp-date-picker` examples.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -755,6 +761,36 @@ private function has_used_attributes( $attribute_names ) {
...
+		];
+
+		if ( in_array( strtok( $class, '_' ), $class_prefixes, true ) ) {
+			return true;
+		}
+
+		return 'amp-date-picker-' === substr( $class, 0, 16 );
```
westonruter: Actually, is this not redundant with:

https://github.com/ampproject/amp-wp/blob/eea565f5a78d441428e049073354b40ae94d6399/includes/sanitizers/class-amp-style-sanitizer.php#L650-L655

includes/sanitizers/class-amp-style-sanitizer.php
kienstra: Thanks, great function to get the class names.

kienstra: The `KeyboardShortcutRow` class only appears on clicking the `?` icon in the bottom right. But so far, I couldn't find more classes that appear via user interaction.

includes/sanitizers/class-amp-style-sanitizer.php
kienstra: Good point, I think it is redundant.

What do you think of deleting that snippet you [shared](https://github.com/ampproject/amp-wp/pull/4339/files#r387407995)?


Otherwise, it might be confusing if that's there, and there's a call below of `->is_class_allowed_in_amp_date_picker()`

westonruter: I'd defer to @schlessera for his recommendation, since he highly optimized this method.

I'd probably rather change this method to:

```php
return in_array( strtok( $class, '_' ), $class_prefixes, true );
```

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -662,6 +662,12 @@ private function has_used_class_name( $class_names ) {
 				}
 			}
 
+			// If the document has an amp-date-picker tag, check if this class is an allowed child of it.
+			// That component's child classes won't be present yet in the document, so prevent tree-shaking valid classes.
+			if ( $this->has_used_tag_names( [ 'amp-date-picker' ] ) && $this->is_class_allowed_in_amp_date_picker( $class_name ) ) {
```
schlessera: This needs to be integrated into the switch logic for the datepicker above that starts at line 651. Otherwise, we will _always_ run the optimized loop, and then _always_ run another check for the datepicker, suddenly making the entire loop work twice as much (and thus much slower).

includes/sanitizers/class-amp-style-sanitizer.php
schlessera: Indeed, as explained in my comment above, the logic needs to be integrated into the switch statement that already has the datepicker included, otherwise this is getting a big performance hit again.

This would then take care of the `'amp-date-picker-' === substr( $class, 0, 16 )` bit, which would be a given once you enter here.

Keep in mind that this logic runs 20k to 30k times on a single request for a regular page. It is important to keep this highly optimized when making changes.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -755,6 +761,36 @@ private function has_used_attributes( $attribute_names ) {
...
+	 * So get whether a class is an allowed child.
+	 *
+	 * @since 1.5.0
+	 * @link https://github.com/airbnb/react-dates/tree/05356/src/components
+	 *
+	 * @param string $class The name of the class to evaluate.
+	 * @return bool Whether the class is allowed as a child of <amp-date-picker>.
+	 */
+	private function is_class_allowed_in_amp_date_picker( $class ) {
+		$class_prefixes = [
```
schlessera: This should be either a const array, or a static variable. Otherwise, you're creating and filling up a fresh new array every single time this method runs.

kienstra: Sure, this is applied in https://github.com/ampproject/amp-wp/pull/4339/commits/01a8e8fc265543052b4ab1858236a12855c4bedd

includes/sanitizers/class-amp-style-sanitizer.php
kienstra: Thanks, good point that this would be a big performance cost always running this other check.

Maybe I'm not understanding your suggestion. 

But on moving this into the switch for the datepicker at line 651:`

```diff
diff --git a/includes/sanitizers/class-amp-style-sanitizer.php b/includes/sanitizers/class-amp-style-sanitizer.php
index e9e2b4be6..5f807791a 100644
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -649,7 +649,7 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
                                switch ( substr( $class_name, 0, 16 ) ) {
                                        // Class names for amp-date-picker, see <https://www.ampproject.org/docs/reference/components/amp-date-picker>.
                                        case 'amp-date-picker-':
-                                               if ( ! $this->has_used_tag_names( [ 'amp-date-picker' ] ) ) {
+                                               if ( ! $this->has_used_tag_names( [ 'amp-date-picker' ] ) || ! $this->is_class_allowed_in_amp_date_picker( $class_name ) ) {  // Though this conditional will probably never be false.
                                                        return false;
                                                }
                                                continue 2;
@@ -662,12 +662,6 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
                                }
                        }
 
-                       // If the document has an amp-date-picker tag, check if this class is an allowed child of it.
-                       // That component's child classes won't be present yet in the document, so prevent tree-shaking valid classes.
-                       if ( $this->has_used_tag_names( [ 'amp-date-picker' ] ) && $this->is_class_allowed_in_amp_date_picker( $class_name ) ) {
-                               continue;
-                       }
-
                        if ( ! isset( $this->used_class_names[ $class_name ] ) ) {
                                return false;
                        }
```

...this won't allow the classes in `is_class_allowed_in_amp_date_picker()` 

```php		
static $class_prefixes = [
	'CalendarDay',
	'CalendarMonth',
	'CalendarMonthGrid',
	'DayPicker',
	'DayPickerKeyboardShortcuts',
	'DayPickerNavigation',
	'KeyboardShortcutRow',
];
```
...as they don't begin with `amp-date-picker-`. 

But maybe I'm not understanding your point. And your point about the performance cost is well-taken.

kienstra: Here's an example `<amp-date-picker>` 

amp-date-picker

includes/sanitizers/class-amp-style-sanitizer.php
westonruter: @schlessera what do you think of this approach: b6f1a2360c84a82882f12c1f0f1126e4ee2a3450?

schlessera: @westonruter See https://github.com/ampproject/amp-wp/commit/b6f1a2360c84a82882f12c1f0f1126e4ee2a3450#r37853546

kienstra:
Nice, thanks!

PR #4340, opened 2020-02-28
In scripts expected to have JSON, ensure the JSON is valid

### LINKED ISSUE #4318: Emoji in JSON causes syntax error and AMP validation error
## Bug Description

When JSON data in post content contains emoji characters, they are getting converted into Twemoji images and break the JSON.

Paste this into a Custom HTML block:

```html
<script type="application/ld+json">
	{"broken":" 🚫 🚫 🚫"}
</script>
```

Something else I noticed with the above `<script type="application/json">` example that WordPress is erroneously converting the emoji into images, so the JSON is even worse:

```html
<script type="application/ld+json">
	{"broken":" <img src="https://s.w.org/images/core/emoji/12.0.0-1/72x72/1f6ab.png" alt="🚫" class="wp-smiley" style="height: 1em; max-height: 1em;"> <img src="https://s.w.org/images/core/emoji/12.0.0-1/72x72/1f6ab.png" alt="🚫" class="wp-smiley" style="height: 1em; max-height: 1em;"> <img src="https://s.w.org/images/core/emoji/12.0.0-1/72x72/1f6ab.png" alt="🚫" class="wp-smiley" style="height: 1em; max-height: 1em;">"}
</script>
```

This is due to this logic:

https://github.com/ampproject/amp-wp/blob/b059e63c1de983a94d598142fded3c6758b0d0b9/includes/class-amp-theme-support.php#L1082-L1092

We may need to replace this logic with a sanitizer that is more discerning about doing the replacements. A sanitizer would also ensure that emoji are converted when used outside of title and content, for example the admin bar or the site title.

## Expected Behaviour

* Emoji in JSON should not be mutated into images.
* Emoji used directly in templates should be converted to Twemoji if `has_action( 'wp_head', 'print_emoji_detection_script' )` is true.

## Additional context

 - WordPress version:
 - Plugin version:
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

### LINKED ISSUE #4320: Add sanitization of JSON to prevent AMP validation errors
## Bug Description

I found out that the AMP validator validates JSON for parse errors, but we are not.

Given this markup:

```html
<script type="application/ld+json">
	{BAD  🚫 🚫 🚫
</script>
<amp-analytics type="nielsen">
		{
			"vars": {
				"apid": "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX",
				"apv": "1.0",
				"ap BAD  🚫 🚫 🚫
	</script>
</amp-analytics>
```

No validation errors are reported by the AMP plugin, and yet the AMP validator does report them:

image

Such invalid JSON should be sanitized while also reporting a validation error to the user to prevent GSC from alerting users to an error when it crawls next that we can proactively alert them of immediately.

Note that the AMP validator does not yet validate `application/ld+json` data, but it should after https://github.com/ampproject/amphtml/issues/25610.

One way that this JSON parse error can occur is if JSON contains emoji. See https://github.com/ampproject/amp-wp/issues/4318.

The code needing to be patched here is [`AMP_Tag_And_Attribute_Sanitizer::validate_cdata_for_node()`](https://github.com/ampproject/amp-wp/blob/b059e63c1de983a94d598142fded3c6758b0d0b9/includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php#L881-L926). Note this code will also need to be modified to fix a separate bug with how `blacklisted_cdata_regex` is handled, so beware of merge conflicts; see https://github.com/ampproject/amp-wp/issues/4319.

The logic can be adopted from the `amphtml-validator` here: https://github.com/ampproject/amphtml/blob/e1fd0e15b85e6f80f5d6382600b0765875d37b23/validator/engine/validator.js#L6226-L6237

Note the `INVALID_JSON_CDATA` error code. The error message should be: “The script tag contains invalid JSON that cannot be parsed.”

Related support topics:

* https://wordpress.org/support/topic/the-script-tag-contains-invalid-json-that-cannot-be-parsed-4/
* https://wordpress.org/support/topic/the-script-tag-contains-invalid-json-that-cannot-be-parsed-3/
* https://wordpress.org/support/topic/the-script-tag-contains-invalid-json-that-cannot-be-parsed-2/

## Expected Behaviour

* 

## Additional context

 - WordPress version:
 - Plugin version:
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

kienstra:
## Summary

* For `<script>` tags with `type="application/json"` or
`type="application/ld+json"`, this ensures that the `textContent` is valid JSON, and produces a `INVALID_JSON_CDATA` error if it's not.
* This doesn't help with #4318. It doesn't prevent `wp_staticize_emoji()` from making the emojis invalid. But it now reports emojis that have become invalid due to that issue.
* For example, adding a Custom HTML block with this will now cause an `INVALID_JSON_CDATA` error in the plugin:

```html
<amp-analytics><script type="application/ld+json">{"anEmoji": "🍫"}</script></amp-analytics>
```

## Testing
1. Add this to a plugin:
```php
add_action(
	'wp_head',
	static function() {
		?>
		<script type="application/ld+json">{"invalid": ⛷️}</script>
		<script type="application/ld+json">{"valid": "something"}</script>
		<script type="application/ld+json">{"anotherValid":" 🚫 🚫 🚫"}</script>
		<?php
	}
);

```
2. Create a new post, and add a Custom HTML block with this:

```html
<amp-analytics><script type="application/json">{"malformed": 🍫}</script>
```
3. Go to the front-end, and validate the URL
4. Expected: there are 2 validation errors:

expected-actual

One should be from the first `<script>` in the `add_action( 'wp_head' ...` call above, and one should be from the Custom HTML block.

The bottom two scripts in the `add_action( 'wp_head' ...` call should not cause a validation error:

```html
<script type="application/ld+json">{"valid": "something"}</script>
<script type="application/ld+json">{"anotherValid":" 🚫 🚫 🚫"}</script>
```

Fixes #4320 

## Checklist


composer.json
```diff
@@ -10,6 +10,7 @@
     "ext-date": "*",
     "ext-dom": "*",
     "ext-iconv": "*",
+    "ext-json": "*",
```
kienstra: PhpStorm reported the lack of this in `composer.json`.

It looks like [json_decode()](https://www.php.net/manual/en/function.json-decode.php), with `AMP_HTML_Utils::is_valid_json()` calls, is from the `json` extension.

And [Jetpack also uses this](https://github.com/Automattic/jetpack/blob/3fca2c56fc064d0c1b1e180fa223feaf47e72c8c/composer.json#L12) in `composer.json`.

kienstra:
https://github.com/WordPress/gutenberg/pull/20547 is pretty similar to the commit above: https://github.com/ampproject/amp-wp/pull/4340/commits/26d97a5323205f5ef46749167598fd957b4215de

kienstra:
Hi @westonruter,
No hurry, it being Friday afternoon. But could you please review this when you have a chance?

Thanks, Weston!

.travis.yml
```diff
@@ -94,7 +90,7 @@ jobs:
 
     - name: E2E tests
       php: "7.3"
-      env: WP_VERSION=latest DEV_LIB_SKIP=phpcs,eslint,xmllint,phpsyntax,phpunit PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=
+      env: WP_VERSION=latest DEV_LIB_SKIP=phpcs,eslint,xmllint,phpsyntax,phpunit
```
kienstra: This is like https://github.com/WordPress/gutenberg/pull/20547. It's because of a failed `E2E` Travis job.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -922,6 +923,15 @@ private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
 				return [ 'code' => self::MANDATORY_CDATA_MISSING_OR_INCORRECT ];
 			}
 		}
+
+		// When the CDATA is expected to be JSON, ensure it's valid JSON.
+		if ( 'script' === $element->tagName &&
+			in_array( $element->getAttribute( 'type' ), [ 'application/json', 'application/ld+json' ], true ) &&
+			null === json_decode( $element->textContent )
```
westonruter: This isn't quite right. What if the JSON being decoded is itself `null`? This needs to make use of `json_last_error()` to detect the error scenario.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -922,6 +923,15 @@ private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
...
+		) {
+			return [ 'code' => self::INVALID_JSON_CDATA ];
```
westonruter: This could include the JSON error code as part of the error data.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
kienstra: OK, sure

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
kienstra: Here's the new error message added in https://github.com/ampproject/amp-wp/pull/4340/commits/ea26570a49ca6f3cbfc8a32c7e64aab3a4a91803. How does that look?

new-error-message

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
kienstra: https://github.com/ampproject/amp-wp/pull/4340/commits/ea26570a49ca6f3cbfc8a32c7e64aab3a4a91803 uses `json_last_error()` instead.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
kienstra: The most common error would probably be probably 'Syntax error'.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -922,6 +923,25 @@ private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
 				return [ 'code' => self::MANDATORY_CDATA_MISSING_OR_INCORRECT ];
 			}
 		}
+
+		// When the CDATA is expected to be JSON, ensure it's valid JSON.
+		if ( 'script' === $element->tagName && in_array( $element->getAttribute( 'type' ), [ 'application/json', 'application/ld+json' ], true ) ) {
+			if ( empty( $element->textContent ) ) {
+				return [
```
kienstra: It looks like in PHP 5.6 and earlier, `json_last_error()` returns `JSON_ERROR_NONE` if passed an empty string, for example.

In the [Changelog](https://www.php.net/manual/en/function.json-decode.php#refsect1-function.json-decode-changelog):
> 7.0.0 | An empty PHP string or value that after casting to string is an empty string (NULL, FALSE) results in JSON syntax error.

These [failed unit tests](https://travis-ci.org/ampproject/amp-wp/jobs/656510177#L1278) are from that different behavior, I think.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -922,6 +923,25 @@ private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
...
+					'code'    => self::INVALID_JSON_CDATA,
+					'message' => esc_html__( 'Expected JSON, got an empty value', 'amp' ),
```
kienstra: Here's what this looks when the expected JSON is `empty()`:

ms-new-error

tests/php/test-tag-and-attribute-sanitizer.php
```diff
@@ -2463,8 +2465,32 @@ public function get_html_data() {
...
 				[ AMP_Tag_And_Attribute_Sanitizer::INVALID_CDATA_HTML_COMMENTS ],
 			],
+			'cdata_malformed_json'                    => [
+				'<html><head><meta charset="utf-8"><script type="application/ld+json">{"example": </script></head><body></body></html>',
+				'<html><head><meta charset="utf-8"></head><body></body></html>',
+				[],
+				[ AMP_Tag_And_Attribute_Sanitizer::INVALID_JSON_CDATA ],
+			],
+			'cdata_malformed_json_with_emojis'        => [
+				'<html><head><meta charset="utf-8"><script type="application/ld+json">{"wrong": 🚧 🚧 }</script></head><body></body></html>',
```
westonruter: This test isn't quite capturing the original issue observed. This should do it:

```suggestion
				'<html><head><meta charset="utf-8"><script type="application/ld+json">{"wrong": "<?php echo wp_staticize_emoji( "🚧 🚧" ); ?>"}</script></head><body></body></html>',
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
westonruter:
```suggestion
					'message' => __( 'Expected JSON, got an empty value', 'amp' ),
```

Escaping should be done at printing not storage. The validation error may be sent back via JSON where HTML entities should not be present.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -922,6 +923,25 @@ private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
...
+				];
+			}
+
+			json_decode( $element->textContent );
+			if ( JSON_ERROR_NONE !== json_last_error() ) {
+				return [
+					'code'    => self::INVALID_JSON_CDATA,
```
westonruter: Please also include the JSON error code (perhaps `json_error`). Ultimately, this may be preferable than using the `message` as it would mean the strings could be translated, e.g. using https://www.php.net/manual/en/function.json-last-error-msg.php#117393

westonruter: Or rather, each of the error codes could have their own dedicated validation error codes. So instead of just `INVALID_JSON_DATA` there could be one for each of the constants in https://www.php.net/manual/en/function.json-last-error.php

Then the specific error message could be displayed at runtime rather than being stored in the actual error object, allowing for translations tailored to the user.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
westonruter: See note below about how having a more specific error code for this case could be preferable, as the error message could be translated for the user at display time.

tests/php/test-tag-and-attribute-sanitizer.php
kienstra: OK, committed in https://github.com/ampproject/amp-wp/pull/4340/commits/5b187d36165ee6361d494039b78a7772204876b4

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -922,6 +923,25 @@ private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
 				return [ 'code' => self::MANDATORY_CDATA_MISSING_OR_INCORRECT ];
 			}
 		}
+
+		// When the CDATA is expected to be JSON, ensure it's valid JSON.
+		if ( 'script' === $element->tagName && in_array( $element->getAttribute( 'type' ), [ 'application/json', 'application/ld+json' ], true ) ) {
+			if ( empty( $element->textContent ) ) {
```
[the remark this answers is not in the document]
kienstra: At least in PHP 7.0+, `json_decode( trim( ' ' ) )`  results in `json_last_error()` returning `JSON_ERROR_SYNTAX`. 

But your suggestion would probably give a more useful message to the user.

If `$element->textContent` is `' '`, having a message that it's empty is probably more useful than a message about invalid syntax.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -922,6 +923,25 @@ private function validate_cdata_for_node( DOMElement $element, $cdata_spec ) {
...
+					'code'    => self::INVALID_JSON_CDATA,
+					'message' => esc_html__( 'Expected JSON, got an empty value', 'amp' ),
+				];
+			}
+
+			json_decode( $element->textContent );
+			if ( JSON_ERROR_NONE !== json_last_error() ) {
+				return [
+					'code'    => self::INVALID_JSON_CDATA,
+					'message' => esc_html( json_last_error_msg() ),
```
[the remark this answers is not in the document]
kienstra: Sure, applied in https://github.com/ampproject/amp-wp/pull/4340/commits/81ea67dd590cb41f515a26b0225b9eceabfccb56

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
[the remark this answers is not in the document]
kienstra: Applied in https://github.com/ampproject/amp-wp/pull/4340/commits/81ea67dd590cb41f515a26b0225b9eceabfccb56

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
kienstra: Good point, these error codes could be from the constants that `json_last_error()` returns. I'll apply that later tonight if that's alright.

kienstra: >Then the specific error message could be displayed at runtime rather than being stored in the actual error object, allowing for translations tailored to the user.

Sure, it'd be good to have actual translations of these error messages.

But do you have in mind passing each of the [10 error messages](https://www.php.net/manual/en/function.json-last-error.php#refsect1-function.json-last-error-returnvalues) through `__()`, like:

```php
private static function get_error_message() {
	...
	case AMP_Tag_And_Attribute_Sanitizer::JSON_ERROR_DEPTH:
		return __( 'The maximum stack depth has been exceeded', 'amp' );
	case AMP_Tag_And_Attribute_Sanitizer::JSON_ERROR_STATE_MISMATCH:
		return __( 'Invalid or malformed JSON', 'amp' );
	etc...
```

This could get pretty verbose, but maybe I'm not understanding this.

Thanks, Weston!

westonruter: Yes, something like that. Maybe let the error title be normalized to just “Invalid JSON” for each error code, but then have the message have more elaboration, for example:

```diff
diff --git a/includes/validation/class-amp-validation-error-taxonomy.php b/includes/validation/class-amp-validation-error-taxonomy.php
index 6a6158c61..ed6c49e40 100644
--- a/includes/validation/class-amp-validation-error-taxonomy.php
+++ b/includes/validation/class-amp-validation-error-taxonomy.php
@@ -1833,8 +1833,22 @@ class AMP_Validation_Error_Taxonomy {
 					$content .= '</p>';
 				}
 
-				if ( isset( $validation_error['message'] ) ) {
-					$content .= sprintf( '<p>%s</p>', esc_html( $validation_error['message'] ) );
+				$message = null;
+				switch ( $validation_error['code'] ) {
+					case AMP_Tag_And_Attribute_Sanitizer::JSON_ERROR_DEPTH:
+						$message = __( 'The maximum stack depth has been exceeded', 'amp' );
+						break;
+					case AMP_Tag_And_Attribute_Sanitizer::JSON_ERROR_STATE_MISMATCH:
+						$message = __( 'Invalid or malformed JSON', 'amp' );
+						break;
+					// ...
+					default:
+						if ( isset( $validation_error['message'] ) ) {
+							$message = $validation_error['message'];
+						}
+				}
+				if ( $message ) {
+					$content .= sprintf( '<p>%s</p>', esc_html( $message ) );
 				}
 
 				break;
@@ -3099,6 +3113,11 @@ class AMP_Validation_Error_Taxonomy {
 				}
 				return $title;
 
+			case AMP_Tag_And_Attribute_Sanitizer::JSON_ERROR_DEPTH:
+			case AMP_Tag_And_Attribute_Sanitizer::JSON_ERROR_STATE_MISMATCH:
+			// ...
+				return __( 'Invalid JSON', 'amp' );
+
 			default:
 				/* translators: %s error code */
 				return sprintf( __( 'Unknown error (%s)', 'amp' ), $validation_error['code'] );

```

kienstra: Thanks, how does https://github.com/ampproject/amp-wp/pull/4340/commits/47565f19398e500141b1ea839804c1de1a530268 look?

`JSON_ERROR_EMPTY` isn't one of the [json_last_error()](https://www.php.net/manual/en/function.json-last-error.php#refsect1-function.json-last-error-returnvalues) values, but I made it up for `<script>` textContent of `''` or `' '`.

If it's alright, later today I'll add a test for `AMP_Validation_Error_Taxonomy::filter_manage_custom_columns()` that covers that new `switch` statement for the message.

kienstra: https://github.com/ampproject/amp-wp/pull/4340/commits/47b8e8706e56a634bbe745701765f1bb4762ad24 adds a test for the error message.

tests/php/validation/test-class-amp-validation-error-taxonomy.php
```diff
@@ -1130,6 +1130,75 @@ public function test_filter_manage_custom_columns() {
...
+				'taxonomy'    => AMP_Validation_Error_Taxonomy::TAXONOMY_SLUG,
+				'description' => wp_json_encode( $validation_error ),
+			]
+		);
+
+		$GLOBALS['pagenow'] = 'post.php';
+		$filtered_content   = AMP_Validation_Error_Taxonomy::filter_manage_custom_columns( $initial_content, 'error_code', $term_id );
+
+		$this->assertStringStartsWith( $initial_content . '<button type="button" aria-label="Toggle error details"', $filtered_content );
+		$this->assertStringContainsString( $expected_error_message, $filtered_content );
```
westonruter: This method is causing test failures:

> `PHP Fatal error:  Call to undefined method Test_AMP_Validation_Error_Taxonomy::assertStringContainsString() in /tmp/wordpress/src/wp-content/plugins/amp/tests/php/validation/test-class-amp-validation-error-taxonomy.php on line 1199`

westonruter:
> The test failure for `Test_AMP_Validation_Manager::test_has_parameters_passed_by_reference` is not caused by the code here, but by a change in core: [WordPress/wordpress-develop@f40c6c9#diff-1f96a4b7aed0d895bbbcfc265fcb4fc6](https://github.com/WordPress/wordpress-develop/commit/f40c6c9ebe6490eebd4d716d0796f29b59fa963d#diff-1f96a4b7aed0d895bbbcfc265fcb4fc6)

This is being fixed in #4357.

tests/php/validation/test-class-amp-validation-error-taxonomy.php
kienstra: Sorry for missing that. https://github.com/ampproject/amp-wp/pull/4340/commits/ae5df635383dc7d2304b0564574683d3ad42871a fixes it.

PR #4357, opened 2020-03-03
Add Tiktok embed support

### LINKED ISSUE #4326: Tiktok embed support - Something for the plugin, or custom scripting?
## Feature description

[Tiktok has an embed code and oEmbed endpoint](https://developers.tiktok.com/doc/Embed) that works with core WP, but so far it does nothing in AMP and I can't find anyone on the internet talking about the issue. 

What's the right approach for me to take if I want to support it? Wait for an `amp-tiktok` script like the Twitter/Instagram ones? 

Should I be trying to support their iframe directly somehow?

Thanks for any feedback. Sorry if the question doesn't belong here. I realize there's a never-ending list of potential embeds for AMP to support, just trying to figure out if this one is a reasonable goal for one person to tackle, or if I should just forbid my users from embedding Tiktok until AMP supports it (if they ever do).

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

Fixes #4326.

## Todo

- [x] Fix failing tests

## Checklist


includes/embeds/class-amp-tiktok-embed-handler.php
```diff
@@ -0,0 +1,94 @@
...
+				'amp-iframe',
+				[
+					'width'   => $attr['width'],
+					'height'  => $attr['height'],
+
+					/*
+					 * A `lang` query parameter is added to the URL via JS. This can't be determined here so it is not
+					 * added. Whether it alters the embed in any way or not has not been determined.
+					 */
+					'src'     => 'https://www.tiktok.com/embed/v2/' . $video_id,
```
pierlon: Want to point out a potential issue here; the `lang` query parameter is added by the accompanied JS and so cannot be determined here. I'm not sure if this affects the embed in any way.

includes/embeds/class-amp-tiktok-embed-handler.php
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Class AMP_Tiktok_Embed_Handler
+ */
+class AMP_Tiktok_Embed_Handler extends AMP_Base_Embed_Handler {
+
+	/**
+	 * Registers embed.
+	 */
+	public function register_embed() {
+		add_filter( 'embed_oembed_html', [ $this, 'filter_embed_oembed_html' ], 10, 3 );
```
westonruter: Note that the raw embeds also need to be handled, that is the `blockquote` that someone pastes into a Custom HTML block for example. See `\AMP_Twitter_Embed_Handler::sanitize_raw_embeds()`. The Twitter embed handler doesn't filter `embed_oembed_html` at all. I recall discussing this with you in December, that we can just rely on `sanitize_raw_embeds` alone.

includes/embeds/class-amp-tiktok-embed-handler.php
westonruter: One possibility would be to make use of a URL substitution. 

```suggestion
					/*
					 * A `lang` query parameter is added to the URL via JS. This can't be determined here so it is not
					 * added. Whether it alters the embed in any way or not has not been determined.
					 */
					'src'     => 'https://www.tiktok.com/embed/v2/' . $video_id . '?lang=BROWSER_LANGUAGE',
```

See https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md

_But_ it seems AMP doesn't support this yet, as this PR was closed: https://github.com/ampproject/amphtml/pull/10884

westonruter:
Aside: if it turns out the `amp-iframe` can't be used to faithfully embed TikTok, please do open an amphtml issue to flag it. Since TikTok is becoming very popular, it's something AMP needs to support.

includes/embeds/class-amp-tiktok-embed-handler.php
pierlon: Right, I haven't found any definite proof of URL substitution being done on the `src` attribute of an `amp-iframe`.

includes/embeds/class-amp-tiktok-embed-handler.php
pierlon: Embeds are now being handled via `sanitize_raw_embeds`; see https://github.com/ampproject/amp-wp/pull/4357/commits/e8fa07baddd451272c0a9117c47e31ca3518d857.

pierlon:
With the exception of tests being run against WP trunk, the others will be failing until WP 5.4 is released due to the Tiktok embed is not being available as yet.

westonruter:
> With the exception of tests being run against WP trunk, the others will be failing until WP 5.4 is released due to the Tiktok embed is not being available as yet.

I suggest `markTestSkipped` if < WP 5.4-beta.

includes/embeds/class-amp-tiktok-embed-handler.php
```diff
@@ -0,0 +1,161 @@
...
+		// Create fallback node.
+		$fallback_node = AMP_DOM_Utils::create_node(
+			$dom,
+			'a',
+			[
+				'href'  => esc_url_raw( $embed_url ),
+				'class' => 'amp-wp-embed-fallback',
+			]
+		);
+		$fallback_node->appendChild( new DOMText( esc_html( $embed_url ) ) );
```
westonruter: No need to `esc_html()` here. The DOM will automatically escape.

includes/embeds/class-amp-tiktok-embed-handler.php
```diff
@@ -0,0 +1,161 @@
...
+			if ( 'section' === $child->nodeName ) {
+				$placeholder_node = $node->removeChild( $child );
+				break;
+			}
+		}
+
+		// If the placeholder was not found, use the fallback instead.
+		if ( ! $placeholder_node ) {
+			$placeholder_node = $fallback_node;
+		}
```
westonruter: Is `$fallback` even needed? By very nature of exclusively sanitizing raw embeds, the `blockquote` will always be present and thus there will always be a `placeholder`.

includes/embeds/class-amp-tiktok-embed-handler.php
```diff
@@ -0,0 +1,161 @@
...
+		// Not implemented.
+	}
+
+	/**
+	 * Sanitize Tiktok embeds to be AMP compatible.
+	 *
+	 * @param Document $dom DOM.
+	 */
+	public function sanitize_raw_embeds( Document $dom ) {
+		$nodes = $dom->xpath->query( '//blockquote[ @class="tiktok-embed" ]' );
```
westonruter: May be more robust to check if `contains( @class, 'tiktok-embed' )`.

includes/embeds/class-amp-tiktok-embed-handler.php
```diff
@@ -0,0 +1,161 @@
...
+			[
+				'href'  => esc_url_raw( $embed_url ),
+				'class' => 'amp-wp-embed-fallback',
+			]
+		);
+		$fallback_node->appendChild( new DOMText( esc_html( $embed_url ) ) );
+
+		// If there is no video ID, replace the blockquote with the fallback and return.
+		if ( empty( $video_id ) ) {
+			$node->parentNode->replaceChild( $fallback_node, $node );
```
westonruter: I think rather than replace with a fallback, this method should short-circuit without doing anything.

includes/embeds/class-amp-tiktok-embed-handler.php
```diff
@@ -0,0 +1,161 @@
...
+
+		$placeholder_node->setAttribute( 'placeholder', '' );
+
+		$amp_iframe_node = AMP_DOM_Utils::create_node(
+			$dom,
+			'amp-iframe',
+			[
+				'layout'  => 'responsive',
+				'width'   => $this->DEFAULT_WIDTH,
+				'height'  => $this->DEFAULT_HEIGHT,
```
westonruter: Should these note be derived from the `style="max-width: 605px;min-width: 325px;"` on the `blockquote`? Those properties should then be removed from the `blockquote`. 

Otherwise, the `blockquote` should get a `position:relative` and the `amp-iframe` should perhaps get `layout=fill`, no?

pierlon: The `blockquote` and `iframe` are resized during page load, so the width and height can't be derived from the `blockquote` style.

> Otherwise, the `blockquote` should get a `position:relative` and the `amp-iframe` should perhaps get `layout=fill`, no?

With that approach the `blockquote` nor `amp-iframe` is resized, leading to an almost non-visible embed:

image

westonruter:
I think this is about as good as it's going to get until AMP has specific support for TikTok (or better TikTok directly supports `amp-iframe`). See https://github.com/ampproject/amphtml/issues/27080.

westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/4303814/amp.zip) (v1.5.0-alpha-20200308T201719Z-28a05d38a)

PR #4371, opened 2020-03-09
Add DevMode helper and keep devmode nodes during optimization

### LINKED ISSUE #4369: Optimizer strips admin bar style[data-ampdevmode]
## Bug Description

When the admin bar is shown and dev mode is enabled, when the Optimizer is turned off an AMP page will include:

```html
<style id="admin-bar-inline-css" data-ampdevmode="">
html { margin-top: 32px !important; }
	@media screen and ( max-width: 782px ) {
		html { margin-top: 46px !important; }
	}html:not(#_) > body { position:unset !important; }
</style>
```

However, when the Optimizer is enabled, this `style` element is getting removed. The effect is the admin bar is not properly positioned:

image

## Expected Behaviour

Elements with `data-ampdevmode` should be exempted from the Optimizer.

## Steps to reproduce

1. Make sure the Optimizer is enabled, e.g. via `add_filter( 'amp_enable_optimizer', '__return_true' ); add_filter( 'amp_enable_ssr', '__return_true' );`
2. Log-in to the WordPress admin so the admin bar us displayed.
3. View the frontend with Twenty Twenty active.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
## Summary

This PR makes the following changes:

- Adds a new `AmpProject\DevMode` helper class (with tests);
- Deprecates the devmode-related methods in `AMP_Base_Sanitizer`;
- Moves usage of the deprecated `AMP_Base_Sanitizer` methods over to the new `AmpProject\DevMode` helper class;
- Ensures that the `AmpBoilerplate` transformer does not remove nodes that are marked as being in devmode.

The `AmpProject\DevMode` helper class was introduced (and the `AMP_BaseSanitizer` code deprecated) because the `ampproject/optimizer` cannot have a dependency pointing to the plugin package. Dependencies can only ever be allowed the other way around.

Fixes #4369

## Checklist

### DIFF
--- a/lib/optimizer/src/Transformer/AmpBoilerplate.php
+++ b/lib/optimizer/src/Transformer/AmpBoilerplate.php
@@ -4,6 +4,7 @@ namespace AmpProject\Optimizer\Transformer;
 
 use AmpProject\Amp;
 use AmpProject\Attribute;
+use AmpProject\DevMode;
 use AmpProject\Dom\Document;
 use AmpProject\Optimizer\ErrorCollection;
 use AmpProject\Optimizer\Transformer;
@@ -74,9 +75,10 @@ final class AmpBoilerplate implements Transformer
     private function removeStyleAndNoscriptTags(Document $document)
     {
         $headNode = $document->head->firstChild;
+        $devMode = DevMode::isActiveForDocument($document);
         while ($headNode) {
             $nextSibling = $headNode->nextSibling;
-            if ($headNode instanceof DOMElement) {
+            if ($headNode instanceof DOMElement && (! $devMode || ! DevMode::hasExemptionForNode($headNode))) {
                 switch ($headNode->tagName) {
                     case Tag::STYLE:
                         if (! $headNode->hasAttribute(Attribute::AMP_CUSTOM)) {
--- a/includes/sanitizers/class-amp-audio-sanitizer.php
+++ b/includes/sanitizers/class-amp-audio-sanitizer.php
@@ -5,6 +5,8 @@
  * @package AMP
  */
 
+use AmpProject\DevMode;
+
 /**
  * Class AMP_Audio_Sanitizer
  *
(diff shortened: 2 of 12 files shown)


lib/optimizer/src/Transformer/AmpBoilerplate.php
```diff
@@ -74,9 +75,10 @@ public function transform(Document $document, ErrorCollection $errors)
     private function removeStyleAndNoscriptTags(Document $document)
     {
         $headNode = $document->head->firstChild;
+        $devMode = DevMode::isActiveForDocument($document);
         while ($headNode) {
             $nextSibling = $headNode->nextSibling;
-            if ($headNode instanceof DOMElement) {
+            if ($headNode instanceof DOMElement && (! $devMode || ! DevMode::hasExemptionForNode($headNode))) {
```
westonruter: Why not use `isExemptFromValidation`?

```suggestion
            if ($headNode instanceof DOMElement && ! DevMode::isExemptFromValidation($headNode)) {
```

Then `$devMode` variable could be removed.

lib/optimizer/src/Transformer/AmpBoilerplate.php
```diff
@@ -74,9 +75,10 @@ public function transform(Document $document, ErrorCollection $errors)
     private function removeStyleAndNoscriptTags(Document $document)
     {
         $headNode = $document->head->firstChild;
+        $devMode = DevMode::isActiveForDocument($document);
```
westonruter:
```suggestion
```

lib/optimizer/src/Transformer/AmpBoilerplate.php
schlessera: Because it's a loop, and that would do the check against the document multiple times although it's only needed once.

westonruter: OK. But that's almost always true for other instances of where `isExemptFromValidation` is used currently. 

I don't feel strongly about this.

Issue #4276, opened 2020-02-11
CSS Parser does not preserve styles used inside templates script[type=text/plain]


spasibych:
## Bug Description

The plugin implements the removal of unused CSS rules on the page. This works well for all page content, including `<template type="amp-mustache">...</template>`, but does not work with `<script type="text/plain" template="amp-mustache">...</script>`.

## Expected Behaviour

Preserve CSS rules for elements inside `<script type="text/plain" template="amp-mustache">...</script>`.

## Steps to reproduce

1. Add code to AMP page:
```html
<style>
    h2.one { color: green }
    h2.two { color: red }
</style>
<template type="amp-mustache">
    <h2 class="one">One</h2>
</template>
<script type="text/plain" template="amp-mustache">
    <h2 class="two">Two</h2>
</script>
```
2. View the resulting source code for an AMP page.
3. See content `<style amp-custom>...</style>` in page head where there is a rule ` h2.one{color:green}` but there is no rule `h2.two{color:red}`.

## Additional context

 - WordPress version: 5.3.2
 - Plugin version: 1.4.2
 - AMP plugin template mode: Website; Standard; Serve all templates as AMP regardless of what is being queried.
 - PHP version: 7.3.11

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

westonruter:
I believe this may be resolved by fixing #4254.

spasibych:
@westonruter thanks!

This is especially important if full support of `<!--{{#foo}}-->` is not implemented (issue [#26656](https://github.com/ampproject/amphtml/issues/26656)).

schlessera:
> I believe this may be resolved by fixing #4254.

Just verified that this is indeed fixed via the PR that @kienstra did for #4254 :

<img width="370" alt="Image 2020-03-02 at 6 42 36 AM" src="https://user-images.githubusercontent.com/83631/75648694-1f922800-5c51-11ea-841f-a855f089184b.png">

I'll add a "Fixes #4276" to that PR.

csossi:
Verified in QA:

</header>
--
  |  
  | <div class="post-inner thin ">
  |  
  | <div class="entry-content">
  |  
  |  
  | <style>
  | h2.one { color: green }
  | h2.two { color: red }
  | </style>
  | <template type="amp-mustache">
  | <h2 class="one">One</h2>
  | </template>
[...]

PR #4374, opened 2020-03-10
Exclude processing style attributes inside of Mustache templates

westonruter:
## Summary

As reported in a [support topic](https://wordpress.org/support/topic/amp-autocomplete-inline-style-not-rendering/) by @artneo7, there is an issue with the style sanitizer in how it handles inline `style` attributes that appear inside of Mustache `template` elements. Consider this markup:

```html
<form id="myform" role="search" class="search-form" method="get" action="https://example.com/" target="_top">
	<amp-autocomplete filter="substring" items="." filter-value="title" max-entries="6" min-characters="2" submit-on-enter="" src="https://example.com/autocomplete/">
		<template type="amp-mustache" id="amp-template-custom">
			<div class="city-item" data-value="{{city}}, {{state}}">
				<div style="color: {{regionColor}}">{{city}}, {{state}}</div>
				<div class="custom-population">Population: {{population}}</div>
			</div>
		</template>
	</amp-autocomplete>
</form>
```

Note specifically `<div style="color: {{regionColor}}">{{city}}, {{state}}</div>`.

At the moment, the AMP plugin is trying to extract this inline `style` to be added to the `amp-custom` stylesheet, leaving behind a CSS class:

```html
<form id="myform" role="search" class="search-form" method="get" action="https://example.com/" target="_top">
	<amp-autocomplete filter="substring" items="." filter-value="title" max-entries="6" min-characters="2" submit-on-enter="" src="https://example.com/autocomplete/">
		<template type="amp-mustache" id="amp-template-custom">
			<div class="city-item" data-value="{{city}}, {{state}}">
				<div class="amp-wp-88b51c5">{{city}}, {{state}}</div>
				<div class="custom-population">Population: {{population}}</div>
			</div>
		</template>
	</amp-autocomplete>
</form>
```

But no CSS rule for `.amp-wp-88b51c5 { color:: {{regionColor}} }` appears in `style[amp-custom]`, which of course it shouldn't be cause: (1) it's a syntax error and (2) it doesn't make sense outside of a mustache `template` anyway.

In this special case, the style sanitizer needs to prevent converting inline `style` attributes to be part of `style[amp-custom]`. The reality is that AMP now allows for inline `style` attributes as long as they aren't `!important`. (See also #1313.) Since we can't move the style to `style[amp-custom]` it's fine to leave it in place. The reason for always moving inline `style` elements even when not `!important` is to preserve the relative CSS cascade (see https://github.com/ampproject/amp-wp/issues/1313#issuecomment-535731938), but in the case of styles appearing in Mustache templates the concern is minimized because the author already indicates they are writing AMP markup, and so it will be up to them to ensure the inline `style` is fine to have the final word.

With the changes in this PR, inline `style` attributes that contain Mustache variable tags are left unchanged.

Relates to #4276 and #3137.

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -2527,11 +2527,23 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 	 */
 	private function collect_inline_styles( $element ) {
 		$style_attribute = $element->getAttributeNode( 'style' );
-		if ( ! $style_attribute || ! trim( $style_attribute->nodeValue ) ) {
+		if ( empty( $style_attribute ) ) {
 			return;
 		}
 
-		$value = $style_attribute->nodeValue;
+		$value = trim( $style_attribute->nodeValue );
+		if ( empty( $value ) ) {
+			return;
+		}
+
+		// Skip processing stylesheets that contain mustache template variables if the element is inside of a mustache template.
+		if (
+			preg_match( '/{{[^}]+?}}/', $value ) &&
+			0 !== $this->dom->xpath->query( '//template[ @type="amp-mustache" ]//.', $element )->length
+		) {
+			return;
+		}
+
 		$class = 'amp-wp-' . substr( md5( $value ), 0, 7 );
 		$root  = ':root' . str_repeat( ':not(#_)', self::INLINE_SPECIFICITY_MULTIPLIER );
 		$rule  = sprintf( '%s .%s { %s }', $root, $class, $style_attribute->nodeValue );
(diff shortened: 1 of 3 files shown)


westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/4314105/amp.zip) (v1.5.0-alpha-20200310T182805Z-effc46f80)

westonruter:
Fix verified by @artneo7: https://wordpress.org/support/topic/amp-autocomplete-inline-style-not-rendering/#post-12529258

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -2527,11 +2527,23 @@ static function( Selector $old_selector ) {
 	 */
 	private function collect_inline_styles( $element ) {
 		$style_attribute = $element->getAttributeNode( 'style' );
-		if ( ! $style_attribute || ! trim( $style_attribute->nodeValue ) ) {
+		if ( empty( $style_attribute ) ) {
+			return;
+		}
```
westonruter: I suppose this is redundant.

westonruter: Nah, better safe than sorry.

PR #4375, opened 2020-03-11
Move eligible viewport rules to `meta[name="viewport"]`

### LINKED ISSUE #4118: Automatically extract @viewport CSS at-rules to meta viewport tag
## Feature description

A very common validation error encountered when checking themes is:

> Illegal CSS at-rule: @-ms-viewport

We even suppress them in core themes. For example:

https://github.com/ampproject/amp-wp/blob/1da327d32753d811af023416b75c67f1d0b1dbb1/includes/sanitizers/class-amp-core-theme-sanitizer.php#L251-L265

(There are other vendor prefixes which are possible which we are not accounting for in the core themes.)

We should consider extracting the properties from the the `@viewport` at-rules to populate the `meta[name=viewport]` element, and then just remove the rules from the CSS and skip reporting any validation error.

Something needing to be checked whether the `meta` tag or the at-rule has precedence, and then prepend or append the properties accordingly. Validation would also need to be performed to only copy valid properties.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

kienstra:
## Summary
- [x] Prevent validation errors for `@viewport` rules and other vendor prefixes
- [x] When those rules would be valid in `meta[name="viewport"]`, move them there
- [x] When there is an existing `meta[name="viewport"]`, consider the rules that already has, and merge the style `@viewport` rules with those

Fixes #4118

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-meta-sanitizer.php
+++ b/includes/sanitizers/class-amp-meta-sanitizer.php
@@ -135,10 +135,8 @@ class AMP_Meta_Sanitizer extends AMP_Base_Sanitizer {
 	 * The viewport defaults to 'width=device-width', which is the bare minimum that AMP requires.
 	 */
 	protected function ensure_viewport_is_present() {
-		$viewport_rules = $this->get_valid_viewport_rules( AMP_Style_Sanitizer::get_extracted_viewport_rules() );
-
 		if ( empty( $this->meta_tags[ self::TAG_VIEWPORT ] ) ) {
-			$this->meta_tags[ self::TAG_VIEWPORT ][] = $this->create_viewport_element( $viewport_rules );
+			$this->meta_tags[ self::TAG_VIEWPORT ][] = $this->create_viewport_element( static::AMP_VIEWPORT );
 		}
 		// @todo: In an else block, add $viewport rules to the existing meta[name="viewport"], prioritizing the existing ones.
 	}
--- a/includes/sanitizers/class-amp-meta-sanitizer.php
+++ b/includes/sanitizers/class-amp-meta-sanitizer.php
@@ -139,7 +139,9 @@ class AMP_Meta_Sanitizer extends AMP_Base_Sanitizer {
 			$this->meta_tags[ self::TAG_VIEWPORT ][] = $this->create_viewport_element( static::AMP_VIEWPORT );
 		} else {
 			$parsed_rules = [];
-			foreach ( $this->meta_tags[ self::TAG_VIEWPORT ] as $meta_viewport ) {
+
+			// Reverse the meta[name="viewport"] tags, so original meta[name="viewport"] tags have precedence @viewport style rules.
+			foreach ( array_reverse( $this->meta_tags[ self::TAG_VIEWPORT ] ) as $meta_viewport ) {
 				$viewport_content = explode( ',', $meta_viewport->getAttribute( 'content' ) );
 				foreach ( $viewport_content as $rule ) {
 					list( $name, $value )          = explode( '=', $rule, 2 );
@@ -155,7 +157,7 @@ class AMP_Meta_Sanitizer extends AMP_Base_Sanitizer {
 	/**
 	 * Gets the viewport rules that would be valid in meta[name="viewport"].
 	 *
-	 * @link https://github.com/ampproject/amphtml/blob/5f75efe35b734a4fcf0884d373315d
+	 * @link https://github.com/ampproject/amphtml/blob/6b439aec75c1d9b52ba19435f6730b27a457bf42/validator/validator-main.protoascii#L436-L445
 	 *
 	 * @param array $rules The rules to evaluate, an associative array of $rule_name => $rule_value.
 	 * @return string The rules of those that are valid.
--- a/includes/sanitizers/class-amp-meta-sanitizer.php
+++ b/includes/sanitizers/class-amp-meta-sanitizer.php
@@ -133,81 +133,45 @@ class AMP_Meta_Sanitizer extends AMP_Base_Sanitizer {
 	 * Always ensure we have a viewport tag.
 	 *
 	 * The viewport defaults to 'width=device-width', which is the bare minimum that AMP requires.
-	 * If there are @viewport style rules, these will have been moved into the content attribute of their own meta[name="viewport"] tags.
-	 * So this iterates over all of those tags, gets the content value,
-	 * and merges all of the valid values into a single meta tag.
+	 * If there are `@viewport` style rules, these will have been moved into the content attribute of their own meta[name=viewport]
+	 * tags by the style sanitizer. When there are multiple such meta tags, this method extracts the viewport properties of each
+	 * and then merges them into a single meta[name=viewport] tag. Any invalid properties will get removed by the
+	 * tag-and-attribute sanitizer.
 	 */
 	protected function ensure_viewport_is_present() {
 		if ( empty( $this->meta_tags[ self::TAG_VIEWPORT ] ) ) {
 			$this->meta_tags[ self::TAG_VIEWPORT ][] = $this->create_viewport_element( static::AMP_VIEWPORT );
 		} else {
+			// Merge one or more meta[name=viewport] tags into one.
 			$parsed_rules = [];
 
-			// Reverse the meta[name="viewport"] tags, so original meta[name="viewport"] tags have precedence over @viewport style rules.
-			foreach ( array_reverse( $this->meta_tags[ self::TAG_VIEWPORT ] ) as $meta_viewport ) {
-				$viewport_content = explode( ',', $meta_viewport->getAttribute( 'content' ) );
-				foreach ( $viewport_content as $rule ) {
-					$exploded_rule = explode( '=', $rule, 2 );
-					if ( ! isset( $exploded_rule[1] ) ) {
-						continue;
+			/**
+			 * Meta viewport element.
+			 *
+			 * @var DOMElement $meta_viewport
+			 */
+			foreach ( $this->meta_tags[ self::TAG_VIEWPORT ] as $meta_viewport ) {
+				$property_pairs = explode( ',', $meta_viewport->getAttribute( 'content' ) );
+				foreach ( $property_pairs as $property_pair ) {
+					$exploded_pair = explode( '=', $property_pair, 2 );
+					if ( isset( $exploded_pair[1] ) ) {
+						$parsed_rules[ trim( $exploded_pair[0] ) ] = trim( $exploded_pair[1] );
 					}
-
-					list( $name, $value )          = $exploded_rule;
-					$parsed_rules[ trim( $name ) ] = trim( $value );
 				}
 			}
 
-			$valid_rules                           = $this->get_valid_viewport_rules( $parsed_rules );
-			$this->meta_tags[ self::TAG_VIEWPORT ] = [ $this->create_viewport_element( $valid_rules ) ];
+			$viewport_value = implode(
+				',',
+				array_map(
+					static function ( $rule_name ) use ( $parsed_rules ) {
+						return $rule_name . '=' . $parsed_rules[ $rule_name ];
+					},
+					array_keys( $parsed_rules )
+				)
+			);
+
+			$this->meta_tags[ self::TAG_VIEWPORT ] = [ $this->create_viewport_element( $viewport_value ) ];
 		}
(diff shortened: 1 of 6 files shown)


includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1987,7 +2005,18 @@ private function process_css_list( CSSList $css_list, $options ) {
...
-				if ( ! in_array( $css_item->atRuleName(), $options['allowed_at_rules'], true ) ) {
+				if ( in_array( $css_item->atRuleName(), $this->allowed_viewport_rules, true ) ) {
+					$output_format = new OutputFormat();
+					foreach ( $css_item->getRules() as $rule ) {
+						$rule_value = $rule->getValue();
+						if ( $rule_value instanceof Value ) {
+							$rule_value = $rule_value->render( $output_format );
+						}
+
+						self::$extracted_viewport_rules[] = [ $rule->getRule() => $rule_value ];
```
westonruter: Shouldn't we just store the properties in one array?

```suggestion
						self::$extracted_viewport_rules[ $rule->getRule() ] = $rule_value;
```

(Note the other discussion about changing `self::$extracted_viewport_rules` to be part of an array sent in the return value.)

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -3241,4 +3270,13 @@ function ( $a, $b ) {
...
 	}
+
+	/**
+	 * Gets the @viewport CSS rules that were extracted from stylesheet, and vendor-prefixed rules like @-ms-viewport.
+	 *
+	 * @return array The rules that were extracted.
+	 */
+	public static function get_extracted_viewport_rules() {
+		return self::$extracted_viewport_rules;
+	}
```
westonruter: I don't think this should be added. Instead of exposing the extracted viewport rules, let's instead inside of `AMP_Style_Sanitizer::process_style_element()` and `AMP_Style_Sanitizer::process_link_element()` obtain the previously-obtained 'viewport_rules` inject them as a new `meta` tag into the document:

```diff
diff --git a/includes/sanitizers/class-amp-style-sanitizer.php b/includes/sanitizers/class-amp-style-sanitizer.php
index 81bfc64f0..5afe1e5a9 100644
--- a/includes/sanitizers/class-amp-style-sanitizer.php
+++ b/includes/sanitizers/class-amp-style-sanitizer.php
@@ -1315,6 +1315,26 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
 			]
 		);
 
+		if ( ! empty( $parsed['viewport_rules'] ) ) {
+			$viewport_meta = $this->dom->createElement( 'meta' );
+			$viewport_meta->setAttribute( 'name', 'viewport' );
+			$viewport_meta->setAttribute(
+				'content',
+				implode(
+					',',
+					array_map(
+						static function ( $property_name ) use ( $parsed ) {
+							return $property_name . '=' . $parsed['viewport_rules'][ $property_name ];
+						},
+						array_keys( $parsed['viewport_rules'] )
+					)
+				)
+			);
+
+			// Inject aa potential duplicate meta viewport element, to later be merged in AMP_Meta_Sanitizer.
+			$element->parentNode->insertBefore( $viewport_meta, $element );
+		}
+
 		$this->pending_stylesheets[] = [
 			'group'              => self::STYLE_AMP_CUSTOM_GROUP_INDEX,
 			'original_size'      => strlen( $stylesheet ),
```

With this in place, note that `AMP_Meta_Sanitizer` already is forced to run _after_ `AMP_Style_Sanitizer`, so what that means is that the `AMP_Style_Sanitizer` can then be updated to iterate over all `meta[name=viewport]` elements and _merge_ them together into one.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -303,6 +305,22 @@ class AMP_Style_Sanitizer extends AMP_Base_Sanitizer {
...
+	private $allowed_viewport_rules = [ 'viewport', '-ms-viewport', '-o-viewport' ];
+
+	/**
+	 * The viewport rules that were extracted from stylesheets.
+	 *
+	 * These aren't valid in CSS, but were extracted so they can be moved to the meta[name="viewport"],
+	 *
+	 * @var array
+	 */
+	private static $extracted_viewport_rules = [];
```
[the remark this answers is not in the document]
westonruter: Update: instead of passing around an object, the methods should be updated to return an array that contains both the `validation_errors` _and_ the `viewport_rules` (along with `imported_stylesheet_urls`, etc). These can then be merged with the return values of any recursive calls.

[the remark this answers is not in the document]
kienstra: Ah, good idea. I started applying this, and will continue.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -133,11 +133,81 @@ protected function ensure_charset_is_present() {
...
+	 * and merges all of the valid values into a single meta tag.
 	 */
 	protected function ensure_viewport_is_present() {
 		if ( empty( $this->meta_tags[ self::TAG_VIEWPORT ] ) ) {
 			$this->meta_tags[ self::TAG_VIEWPORT ][] = $this->create_viewport_element( static::AMP_VIEWPORT );
+		} else {
+			$parsed_rules = [];
+
+			// Reverse the meta[name="viewport"] tags, so original meta[name="viewport"] tags have precedence over @viewport style rules.
+			foreach ( array_reverse( $this->meta_tags[ self::TAG_VIEWPORT ] ) as $meta_viewport ) {
```
kienstra: This prioritizes original `<meta name="viewport">` tags, not `@viewport` style rules that were added to a meta tag. Maybe this isn't what you had in mind, though.

kienstra:
Hi @westonruter,
Thanks a lot for your snippets here.

How does this approach look?

It's [now working](https://github.com/ampproject/amp-wp/issues/4118#issuecomment-598144376).

It merges all of the `meta[name="viewport"]` content values, prioritizing the original meta tag(s), not tags created from `@viewport` rules. Maybe that's not the best approach, though.

Thanks, Weston!

tests/php/test-amp-style-sanitizer.php
```diff
@@ -2628,4 +2628,92 @@ public function assertEqualMarkup( $expected, $actual ) {
...
+			'ms_vendor_prefix_recognized' => [
+				'<style amp-custom>@-ms-viewport{ initial-scale: .8 }</style>',
+				'<meta name="viewport" content="initial-scale=.8,width=device-width">',
+			],
+			'o_vendor_prefix_recognized' => [
+				'<style amp-custom>@-o-viewport{ height: 80% }</style>',
+				'<meta name="viewport" content="height=80%,width=device-width">',
+			],
+			'invalid_vendor_prefix_not_recognized' => [
+				'<style amp-custom>@-baz-viewport{ height: 400px }</style>',
```
westonruter: Custom vendor prefixes should be recognized.

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -133,11 +133,81 @@ protected function ensure_charset_is_present() {
 	 * Always ensure we have a viewport tag.
 	 *
 	 * The viewport defaults to 'width=device-width', which is the bare minimum that AMP requires.
+	 * If there are @viewport style rules, these will have been moved into the content attribute of their own meta[name="viewport"] tags.
+	 * So this iterates over all of those tags, gets the content value,
+	 * and merges all of the valid values into a single meta tag.
 	 */
 	protected function ensure_viewport_is_present() {
```
westonruter: It appears this code is not tested?

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1952,73 +1969,96 @@ static function( $matches ) {
...
 					);
 				}
 			} elseif ( $css_item instanceof Import ) {
-				$results = array_merge(
-					$results,
-					$this->splice_imported_stylesheet( $css_item, $css_list, $options )
-				);
+				$imported_stylesheet = $this->splice_imported_stylesheet( $css_item, $css_list, $options );
+				$imported_font_urls  = array_merge( $imported_font_urls, $imported_stylesheet['imported_font_urls'] );
+				$validation_errors   = array_merge( $validation_errors, $imported_stylesheet['validation_errors'] );
```
westonruter: This is missing the extracted viewport rules.

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -1952,73 +1969,96 @@ static function( $matches ) {
...
-					$error     = [
+				if ( in_array( $css_item->atRuleName(), $this->allowed_viewport_rules, true ) ) {
+					$output_format = new OutputFormat();
+					foreach ( $css_item->getRules() as $rule ) {
+						$rule_value = $rule->getValue();
+						if ( $rule_value instanceof Value ) {
+							$rule_value = $rule_value->render( $output_format );
+						}
+
+						$viewport_rules[ $rule->getRule() ] = $rule_value;
```
westonruter: What else would it be? If it is not a `Value`, should it be added to `$viewport_rules`?

includes/sanitizers/class-amp-meta-sanitizer.php
```diff
@@ -133,11 +133,81 @@ protected function ensure_charset_is_present() {
...
+					if ( ! isset( $exploded_rule[1] ) ) {
+						continue;
+					}
+
+					list( $name, $value )          = $exploded_rule;
+					$parsed_rules[ trim( $name ) ] = trim( $value );
+				}
+			}
+
+			$valid_rules                           = $this->get_valid_viewport_rules( $parsed_rules );
```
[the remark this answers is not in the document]
kienstra: Yeah, they are removed 😄 

Though it might be confusing how the validation error references '<meta>`.

For example, on adding this to a Custom HTML block:

```html
<style>@viewport { disallowed: 100px }</style>
```

...here's the validation error:

validation-error-here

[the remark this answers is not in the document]
westonruter: Yeah, but better than silently removing them. Later we can think about adding attribution.

includes/sanitizers/class-amp-style-sanitizer.php
kienstra: In several unit tests, `$rule_value` is a string, like `device-width`

The logic above is [taken from](https://github.com/sabberworm/PHP-CSS-Parser/blob/a80a97d3fdb2164d72cbf0fb3cfc40b1d630a877/lib/Sabberworm/CSS/Rule/Rule.php#L200-L204

In [AMP_Style_Sanitizer_Test](https://github.com/ampproject/amp-wp/blob/88a136b484ff8a9c0d30986ce95a739a3e70fafc/tests/php/test-amp-style-sanitizer.php), there were 3 times when `$rule_value instanceof Value` was `true`, and 8 times where it was `false`.

kienstra: Yes, I think it should be added to `$vewport_rules` if it's not a `Value`.

westonruter: Thanks. Good explanation. Makes sense.

westonruter:
@kienstra Let me know if any of my changes seem incorrect. Please verify that the core themes that have `viewport` at-rules are not causing any validation errors to occur now. Namely:

* twentysixteen
* twentyeleven
* twentythirteen
* twentyfifteen
* twentyfourteen
* twentytwelve

kienstra:
**Thanks, looks good!**

Hi @westonruter,
Thanks for your commits above, they look good.

Yes, all of the themes you listed are free of validation errors from `@viewport` or `@-ms-viewport`, despite all of them having those rules [in their style.css](https://github.com/WordPress/wordpress-develop/blob/c8d918e47634bd423e0c7ccfc0e8f7a34515c3f5/src/wp-content/themes/twentyeleven/style.css#L2393-L2398).

* twentysixteen
* twentyeleven
* twentythirteen
* twentyfifteen
* twentyfourteen
* twentytwelve

For example, there was no validation error with Twenty Sixteen:

twentysixteen-no-validation-error

...even though its `style.css` file had a `@viewport` style rule:

viewport-rule-non-amp

Issue #3726, opened 2019-11-13
Validation error status icons


schlessera:
The validation error status icons have been changed to be differently colored versions of the AMP logo:

![Image 2019-11-13 at 6 16 04 PM](https://user-images.githubusercontent.com/83631/68787419-c9698c80-0641-11ea-8f57-370024b5154d.png)

However, one of the guiding principles of WCAG when it comes to color is that it should never serve as the only indicator. More specifically, we would talk about WCAG G14 here: https://www.w3.org/TR/WCAG20-TECHS/G14.html

Following are a few examples of how this is perceived with different forms of color blindness.

**Protanopia (red-blind):**

![image](https://user-images.githubusercontent.com/83631/68787809-83f98f00-0642-11ea-9315-bf7c0da21cd0.png)

**Deuteranopia (green-blind):**

![image](https://user-images.githubusercontent.com/83631/68787980-c91dc100-0642-11ea-97f0-33fbc94bbf99.png)

**Achromatopsia (no colours):**

![image](https://user-images.githubusercontent.com/83631/68787995-d2a72900-0642-11ea-8f42-cdb678c6e1e9.png)

First of all, we do have a text representation of the same information here, so it is not necessarily broken right now.

However, there is no direct visual indication that the icon is supposed to relate to the text, and no semantic coding to clearly associate them. It's just an image in a div that is located close to a `<select>`:

![Image 2019-11-13 at 6 26 15 PM](https://user-images.githubusercontent.com/83631/68788137-274aa400-0643-11ea-9b8a-1d205f3f0792.png)

I have two suggestions to improve on this:

**Put the shape into the select element**

It would be helpful to add the icon inside of the select, instead of outside. That would also clearly show that this is supposed to be the same information, which is not currently the case, especially if you open the dropdown. It just still shows the currently selected one, while showing the text of both options:

![Image 2019-11-13 at 6 27 43 PM](https://user-images.githubusercontent.com/83631/68788250-53662500-0643-11ea-857c-d398a77ecf87.png)

**Use two distinct shapes that convey meaning**

Another aspect, though, is that the actual shape of the icon is meaningless and useless in this context. It is truly only used as a container for the color. Therefore, it might be preferable to use the text itself or its background for the color instead. And if that is not wanted, we should look for two distinct shapes to use as the container for the colors, that also serve to communicate the intent properly.

westonruter:
> **ut the shape into the select element**
> 
[...]

PR #4381, opened 2020-03-13
Fix warning icon no longer appears as emoji in macOS Catalina

westonruter:
## Summary

Upon upgrading to macOS Catalina, I noticed the ⚠️ emoji was no longer appearing as color but rather as:

image

This PR ensures that color emoji fonts are used when emoji are added to the page. Originally reported at https://github.com/ampproject/amp-wp/issues/3726#issuecomment-594341592.

Before | After
-------|-------
Screen Shot 2020-03-13 at 09 30 49 | Screen Shot 2020-03-13 at 09 37 43
image | image
image | image

## Checklist

### DIFF
--- a/lib/common/src/Fonts.php
+++ b/lib/common/src/Fonts.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace AmpProject;
+
+/**
+ * Class for fonts.
+ *
+ * @package ampproject/common
+ */
+final class Fonts
+{
+
+    const EMOJI_FONT_STACK = [
+        'Apple Color Emoji',
+        'Android Emoji',
+        'Segoe UI Emoji',
+        'Noto Color Emoji',
+        'EmojiSymbols',
+        'Symbola',
+        'Segoe UI Symbol',
+        'emoji',
+    ];
+
+    /**
+     * Get emoji font family property value.
+     *
+     * @return string Font-family property value.
+     */
+    public static function getEmojiFontFamilyValue()
+    {
+        return implode(
+            ', ',
+            array_map(
+                static function ($font) {
+                    return '"' . $font . '"';
+                },
+                self::EMOJI_FONT_STACK
+            )
+        );
+    }
+}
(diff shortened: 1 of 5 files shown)


westonruter:
> I'm not able to verify that this fixes the issue on Catalina (🐧), but I can confirm the emojis are still being shown correctly for me.

Excellent. I thought you were on Linux so that's partly why I wanted your confirmation that the color emojis were still working.

lib/common/src/Fonts.php
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Class for fonts.
+ *
+ * @package ampproject/common
+ */
+final class Fonts
+{
+
+    const EMOJI_FONT_STACK = [
```
schlessera:
```suggestion
    /**
     * Font stack to cycle through for displaying emojis reliably.
     *
     * @var string[]
     */
    const EMOJI_FONT_STACK = [
```

lib/common/src/Fonts.php
```diff
@@ -0,0 +1,41 @@
...
+    ];
+
+    /**
+     * Get emoji font family property value.
+     *
+     * @return string Font-family property value.
+     */
+    public static function getEmojiFontFamilyValue()
+    {
+        return implode(
```
schlessera: This should be stored in a static variable, otherwise it is computed every single time the value is requested, even though it is purely static and logic is only for easy maintainability.

The usual construct for this is:
```php
static $fontFamilyValue = null;

if ($fontFamilyValue === null) {
	$fontFamilyValue = implode...;
}

return $fontFamilyValue;
```

lib/common/tests/FontsTest.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @covers Amp
+ * @package ampproject/common
+ */
+class FontsTest extends TestCase
+{
+    use AssertContainsCompatibility;
+
+    /**
+     * Test the check for an AMP runtime method.
```
schlessera:
```suggestion
     * Test retrieval of the emoji-specific font-family stack.
```

lib/common/src/Fonts.php
westonruter: Thanks. Fixed in a4ed98330.

PR #4382, opened 2020-03-13
Prevent ACK state of non-modified validation errors from being lost; update validation error icons

### LINKED ISSUE #3725: The ACK state of all validation errors is lost when changing a single one
## Bug Description

When changing a single validation error's state (from "Removed" to "Kept", for example, the ACK state of all validation errors is lost. As new ones are created as well, this makes for a very consuming and random experience.

I assume this is due to the fact that the dropdown with the values does not distinguish between ACK or not anymore, so WP will just store it as a user value either way.

## Expected Behaviour

The ACK state should only change if I specifically act on something. This might need a new UI element, though.

## Steps to reproduce

1. Go to the validation errors screen for a page with multiple new AMP validation errors.
2. **Verify:** The screen should have multiple errors with a reddish border at the left.
3. Change a single validation error from "Removed" to "Kept".
4. Hit the "Update" button.
5. **Verify:** Not only the changed validation error has lost its reddish border, but multiple other ones as well.

## Screenshots

https://youtu.be/yV82f6isY7U

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

This PR prevents the ACK state of other validation errors from being affected when modifying a single one. Also, it updates the validation error icons.

Addresses issue #3725
Addresses issue #3726

## Todo

- [x] Remove validation error count summary from "Validation Error" post metabox
- [x] Update validation error icons with ones from Dashicons
    - [x] Update icons in admin bar
    - [x] Update icons on error index page
    - [x] Update icons validated URLs page
    - [x] Update icons on single validation error page
    - [x] Update icons on single error index page
    - [x] Update icons used in validation error stylesheet summary
    - [x] Remove red and green AMP SVG icons
- [x] Toggle 'new' state for validation error row upon approval/disapproval
- [x] Change validation error status `<select>` color when toggling between error states
- [x] Show a warning message in post metabox if there are invalid or removed validation errors that are not approved
- [x] Add a new row action for Approve/Unapprove on error index page
- [x] Rebrand “New errors” to “Unapproved errors” on error index page

## Screenshots

- Icons in admin bar:

- Updated error index page:

- Updated single error page:

- Updated 'Validated URLs' page:

- Updated single 'Validation URL' page:

- Toggling markup and approval status gives visual feedback:

## Checklist


includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -735,51 +735,48 @@ public static function store_validation_errors( $validation_errors, $url, $args
...
-							$term_id,
-							AMP_Validation_Error_Taxonomy::TAXONOMY_SLUG,
-							[
-								'term_group' => $term_data['term_group'],
-							]
-						);
-					}
-
-					$term = get_term( $term_id );
+			// Not using WP_Term_Query since more likely individual terms are cached and wp_insert_term() will itself look at this cache anyway.
```
pierlon: I've only removed the unneeded if statement here, since `$terms` will always be an empty array.

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -1673,7 +1670,10 @@ public static function handle_validation_error_status_update() {
 				continue;
 			}
 			$term_group = AMP_Validation_Error_Taxonomy::sanitize_term_status( $status );
-			if ( null !== $term_group && $term_group !== $term->term_group ) {
+
+			$status_changed = ( $term->term_group | AMP_Validation_Error_Taxonomy::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK ) !== $term_group;
```
pierlon: Doing a bit-wise OR here gives the acknowledged status code, and would then allow for comparison against the submitted status code (which is always acknowledged).

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -744,51 +755,48 @@ public static function store_validation_errors( $validation_errors, $url, $args
 			$term_data = AMP_Validation_Error_Taxonomy::prepare_validation_error_taxonomy_term( $data );
 			$term_slug = $term_data['slug'];
 
-			if ( ! isset( $terms[ $term_slug ] ) ) {
```
pierlon, line 747: The only change here was removing the if condition wrapping this block of code. `$terms` will always be an empty array, and so there is need to check if `$terms[ $term_slug ]` is set.

src/Icon.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+namespace AmpProject\AmpWP;
+
+/**
+ * Icons used to visually represent validity of validation errors.
+ *
+ * @package AmpProject\AmpWP
+ */
+interface Icon {
```
schlessera: How about turning this simple interface into an actual helper class?

The syntax could then look something like the following:
```php
// Right now:
echo sprint(
	'<span id="amp-admin-bar-item-status-icon" class="ab-icon amp-icon %s"></span>',
	Icon::WARNING
);

// As a helper class:
echo Icon::warning()->with_id( 'amp-admin-bar-item-status-icon' );
```

The main benefit is that it centralizes the markup generation for icons, so that when we make changes at a later point, we don't need to go over all the code and adapt it in multiple places.

As an example, what if we want to add ARIA attributes to improve the accessibility for the icons?

src/Icon.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Interface Icon.
+ *
+ * @package AmpProject\AmpWP
+ */
+
+namespace AmpProject\AmpWP;
+
+/**
+ * Icons used to visually represent validity of validation errors.
```
schlessera: The `validity of validation errors` bit is probably not a very clear phrasing here. I don't think there are invalid validation errors.

Isn't this more about the "state" and/or "effects" of a validation error?

assets/css/src/amp-icons.css
```diff
@@ -0,0 +1,28 @@
...
+
+.amp-icon.amp-warning::before {
+	content: "\f534";
+	color: #d98500 !important;
+}
+
+.amp-icon.amp-link::before {
+	content: "\f103";
+	color: #4183f2 !important;
+}
```
schlessera: Where are these color values coming from?

We should probably stick to the color values from the official admin backend color palette: https://make.wordpress.org/design/handbook/design-guide/foundations/colors/

assets/src/amp-validation/amp-validated-url-post-edit-screen.js
```diff
@@ -283,32 +283,36 @@ const handleSearching = () => {
...
-		if ( event.target.matches( 'select' ) ) {
-			updateSelectIcon( event.target );
-		}
-	};
+const handleRowEvents = () => {
 
 	document.querySelectorAll( 'tr[id^="tag-"]' ).forEach( ( row ) => {
-		const select = row.querySelector( '.amp-validation-error-status' );
+		const statusSelect   = row.querySelector( '.amp-validation-error-status' );
+		const ackCheckbox    = row.querySelector( '.amp-validation-error-status-ack' );
```
schlessera:
```suggestion
		const statusSelect = row.querySelector( '.amp-validation-error-status' );
		const ackCheckbox  = row.querySelector( '.amp-validation-error-status-ack' );
```

assets/src/amp-validation/amp-validated-url-post-edit-screen.js
```diff
@@ -283,32 +283,36 @@ const handleSearching = () => {
...
 
-		if ( select ) {
-			select.addEventListener( 'change', ( event ) => {
-				onChange( { event, row, select } );
+		if ( statusSelect ) {
+			/*
+			 * Handle a change in the error status, like from 'Removed' to 'Kept'. It gets the data-status-icon value
+			 * from the newly-selected <option>. And sets this as the src of the status icon <img>.
```
schlessera:
```suggestion
			 * from the newly-selected <option> and sets this as the src of the status icon <img>.
```

includes/sanitizers/class-amp-style-sanitizer.php
```diff
@@ -2962,15 +2963,15 @@ public function add_css_budget_to_admin_bar() {
...
 		}
-		if ( isset( $emoji ) ) {
-			$stylesheets_a_element->appendChild( $this->dom->createTextNode( ' ' ) );
+		if ( isset( $icon ) ) {
 			$span = $this->dom->createElement( 'span' );
-			$span->setAttribute( 'style', sprintf( 'font-family: %s;', Fonts::getEmojiFontFamilyValue() ) );
-			$span->appendChild( $this->dom->createTextNode( $emoji ) );
+			$span->setAttribute( 'class', 'amp-icon ' . $icon );
+			// Setting the style here prevents it from being removed during tree-shaking.
+			$span->setAttribute( 'style', 'position: absolute; margin-left: 2px;' );
```
schlessera: Sounds like this might be a bug to file against the tree-shaker.

includes/validation/class-amp-validated-url-post-type.php
schlessera: The way I read the logic is that this loops over all validation errors and processes the `$term_slug` for each validation error.

So, `$terms` is only an empty array during the very first iteration of that loop, and on subsequent iterations, the `! isset( $terms[ $term_slug ] )` condition ensures we don't process the same `$term_slug` more than once.

Am I missing something?

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -991,6 +984,17 @@ public static function add_single_post_columns() {
...
 				)
 			),
+			'approved'                    => sprintf(
+				'%s<span class="dashicons dashicons-editor-help tooltip-button" tabindex="0"></span><div class="tooltip" hidden data-content="%s"></div>',
+				esc_html__( 'Approved', 'amp' ),
+				esc_attr(
+					sprintf(
+						'<h3>%s</h3><p>%s</p>',
+						esc_html__( 'Approved', 'amp' ),
+						__( 'Allows confirmation of action taken to resolve the validation error.', 'amp' )
```
schlessera:
```suggestion
						esc_html__( 'Allows confirmation of action taken to resolve the validation error.', 'amp' )
```

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -1653,7 +1657,7 @@ public static function recheck_post( $post ) {
 	public static function handle_validation_error_status_update() {
 		check_admin_referer( self::UPDATE_POST_TERM_STATUS_ACTION, self::UPDATE_POST_TERM_STATUS_ACTION . '_nonce' );
 
-		if ( empty( $_POST[ AMP_Validation_Manager::VALIDATION_ERROR_TERM_STATUS_QUERY_VAR ] ) || ! is_array( $_POST[ AMP_Validation_Manager::VALIDATION_ERROR_TERM_STATUS_QUERY_VAR ] ) ) {
+		if ( empty( $_POST['val_errors'] ) || ! is_array( $_POST['val_errors'] ) ) {
```
schlessera: Maybe introduce a new local constant for `'val_errors'` ?

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -1682,7 +1696,15 @@ public static function handle_validation_error_status_update() {
...
 			}
 			$term_group = AMP_Validation_Error_Taxonomy::sanitize_term_status( $status );
-			if ( null !== $term_group && $term_group !== $term->term_group ) {
+			if ( null === $term_group ) {
+				continue;
+			}
+
+			$term_group = $status_acknowledged
+				? $term_group | AMP_Validation_Error_Taxonomy::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK
+				: $term_group - AMP_Validation_Error_Taxonomy::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK;
```
schlessera: I assume this is meant to _disable_  the `ACK` flag?

This is done with bitmasks by an "and not" operation:
```suggestion
				: $term_group &  ~AMP_Validation_Error_Taxonomy::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK;
```

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -1653,29 +1663,36 @@ public static function filter_tag_row_actions( $actions, WP_Term $tag ) {
...
 						self::VALIDATION_ERROR_ACCEPT_ACTION
 					),
-					esc_html(
-						self::VALIDATION_ERROR_NEW_ACCEPTED_STATUS === $sanitization['term_status'] ? __( 'Confirm removed', 'amp' ) : __( 'Remove', 'amp' )
-					)
+					esc_html__( 'Remove', 'amp' )
 				);
-			}
-			if ( self::VALIDATION_ERROR_ACK_REJECTED_STATUS !== $sanitization['status'] ) {
+			} elseif ( $is_accepted ) {
```
schlessera: This condition is already implied by not matching the above `! $is_accepted`.
```suggestion
			} else {
```

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -1653,29 +1663,36 @@ public static function filter_tag_row_actions( $actions, WP_Term $tag ) {
...
+					add_query_arg( array_merge( [ 'action' => self::VALIDATION_ERROR_ACKNOWLEDGE_ACTION ], compact( 'term_id' ) ) ),
+					self::VALIDATION_ERROR_ACKNOWLEDGE_ACTION
+				),
+				esc_html(
+					! ( $sanitization['term_status'] & self::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK ) ? __( 'Approve', 'amp' ) : __( 'Disapprove', 'amp' )
+				)
+			);
 
-			if ( self::VALIDATION_ERROR_ACK_ACCEPTED_STATUS !== $sanitization['status'] ) {
+			if ( ! $is_accepted ) {
```
schlessera: Small nitpick, but it might be preferable to invert the cases here to use the positive one for the initial code and the negative one for the `else` clause.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -1868,26 +1885,23 @@ public static function filter_manage_custom_columns( $content, $column_name, $te
 				$is_removed = (bool) ( (int) $term->term_group & self::ACCEPTED_VALIDATION_ERROR_BIT_MASK );
 
 				if ( 'post.php' === $pagenow ) {
-					$status_select_name = sprintf( '%s[%s]', AMP_Validation_Manager::VALIDATION_ERROR_TERM_STATUS_QUERY_VAR, $term->slug );
+					$valid_color   = '#0a9f1c';
+					$invalid_color = '#e92c1a';
```
schlessera: We could consider adding the color codes as constants into the `Icon` helper and add their CSS generation as a method on that class.

This would make the color values DRY.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -2029,6 +2043,12 @@ static function( $error ) use ( $term ) {
 					$content .= esc_html__( 'Misc', 'amp' );
 				}
 				break;
+			case 'approved':
+				if ( 'post.php' === $pagenow ) {
+					$checked    = ! ( (int) $term->term_group & self::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK ) ? '' : 'checked="checked"';
```
schlessera:
```suggestion
					$checked    = (int) $term->term_group & self::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK ? 'checked="checked"' : '';
```

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -2888,13 +2907,17 @@ public static function handle_validation_error_update( $redirect_to, $action, $t
...
-			$term_group = $term->term_group;
+			$term_group = (int) $term->term_group;
 
-			// The action of marking an error as removed/kept (aka accepted/rejected) also results in it being marked as not-new.
 			if ( self::VALIDATION_ERROR_ACCEPT_ACTION === $action ) {
-				$term_group = self::VALIDATION_ERROR_ACK_ACCEPTED_STATUS;
+				$term_group |= self::ACCEPTED_VALIDATION_ERROR_BIT_MASK;
 			} elseif ( self::VALIDATION_ERROR_REJECT_ACTION === $action ) {
-				$term_group = self::VALIDATION_ERROR_ACK_REJECTED_STATUS;
+				$term_group -= self::ACCEPTED_VALIDATION_ERROR_BIT_MASK;
```
schlessera:
```suggestion
				$term_group &= ~self::ACCEPTED_VALIDATION_ERROR_BIT_MASK;
```

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -2888,13 +2907,17 @@ public static function handle_validation_error_update( $redirect_to, $action, $t
...
+			} elseif ( self::VALIDATION_ERROR_ACKNOWLEDGE_ACTION === $action ) {
+				$acknowledged = $term_group & self::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK;
+				$term_group   = $acknowledged
+					? $term_group - self::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK
```
schlessera:
```suggestion
					? $term_group & ~self::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK
```

includes/validation/class-amp-validation-manager.php
```diff
@@ -423,23 +438,38 @@ public static function add_admin_bar_menu_items( $wp_admin_bar ) {
...
+				 * may be added during post-processing, such as overriding the admin bar icon to represent a new validation
+				 * state or when new menu items are being added (eg. "CSS Usage") and a icon needs to displayed.
+				 */
+					'<span style="display: none" class="amp-icon %s %s %s %s"></span>',
+					Icon::LINK,
+					Icon::VALID,
+					Icon::WARNING,
+					Icon::INVALID
+				),
+			],
```
schlessera: Would it make sense to add this to the style sanitizer's `'dynamic_element_selectors'` configuration instead?

See https://github.com/ampproject/amp-wp/blob/5718373e576d15b05328b9cf687adf86d0aa9050/includes/sanitizers/class-amp-style-sanitizer.php#L138-L144

tests/php/validation/test-class-amp-validation-error-taxonomy.php
```diff
@@ -995,12 +995,23 @@ public function test_filter_tag_row_actions() {
...
 		$reject_action      = $filtered_actions[ AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_REJECT_ACTION ];
-		$accept_action      = $filtered_actions[ AMP_Validation_Error_Taxonomy::VALIDATION_ERROR_ACCEPT_ACTION ];
+		$this->assertStringContains( strval( $term_this_taxonomy->term_id ), $approve_action );
 		$this->assertStringContains( strval( $term_this_taxonomy->term_id ), $reject_action );
-		$this->assertStringContains( strval( $term_this_taxonomy->term_id ), $accept_action );
+		$this->assertStringContains( 'Approve', $approve_action );
 		$this->assertStringContains( 'Keep', $reject_action );
-		$this->assertStringContains( 'Confirm removed', $accept_action );
+
+		// If its a kept error only 'Remove' action should be shown, and not 'Keep' action.
```
schlessera:
```suggestion
		// If it's a kept error, only 'Remove' action should be shown, and not 'Keep' action.
```

pierlon: Resolved via https://github.com/ampproject/amp-wp/pull/4382/commits/64e47f343d3cf289de6ef3e0e0a9724a46bfc0be.

includes/validation/class-amp-validation-error-taxonomy.php
pierlon: Resolved via https://github.com/ampproject/amp-wp/pull/4382/commits/64e47f343d3cf289de6ef3e0e0a9724a46bfc0be.

includes/validation/class-amp-validated-url-post-type.php
pierlon: I've replaced these with the constant `AMP_Validated_URL_Post_Type::VALIDATION_ERRORS_INPUT_KEY`.

includes/validation/class-amp-validated-url-post-type.php
pierlon: Ah yes you're right. Reverted the change in 64e47f343d3cf289de6ef3e0e0a9724a46bfc0be.

includes/validation/class-amp-validation-manager.php
pierlon: That would mean the HTML markup has to be included prior to sanitizing styles, no?

assets/css/src/amp-icons.css
pierlon: Just colors I selected from various parts of the admin screen. Using colors from the official color scheme makes much more sense here; I've updated them where appropriate:

https://github.com/ampproject/amp-wp/blob/64e47f343d3cf289de6ef3e0e0a9724a46bfc0be/assets/css/src/amp-icons.css#L10-L40

includes/sanitizers/class-amp-style-sanitizer.php
pierlon: I meant to say here that the style would have already been removed during sanitization, since there would be no HTML markup for it.

includes/validation/class-amp-validation-manager.php
schlessera: Yes, indeed. Let's leave it like this then, although it feels very hacky to me. Might be one of the scenarios we should consider when doing "the big refactor" for the sanitizer pipeline.

westonruter:
> We need to confirm that all authors are ok with their commits being contributed to this project. Please have them confirm that by leaving a comment that contains only `@googlebot I consent.` in this pull request.

This request is for @schlessera.

schlessera:
@googlebot I consent.

westonruter:
I don't think we need the link icon here to be colorized:

image

westonruter:
> Do note that the validated URL status will not be acknowledged if the updated status is the same. There has been discussion of updating the UI to allow explicit acknowledgement in such cases (see [#2023 (comment)](https://github.com/ampproject/amp-wp/issues/2023#issuecomment-546476699)).

This description is outdated, correct? I can successfully update the removed/kept status independent new/ack state.

pierlon:
Yes, updating description now.

westonruter:
The icons have inconsistent margin/spacing with the text on the right:

> image

`2px` vs `6px`. Could you fix that?

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -1936,11 +1961,24 @@ public static function print_status_meta_box( $post ) {
...
+						$counts = self::count_invalid_url_validation_errors( self::get_invalid_url_validation_errors( $post ) );
+
+						if ( 0 < ( $counts['new_rejected'] + $counts['new_accepted'] ) ) {
+							?>
+							<strong id="amp-invalid-markup" class="status-text">
+								<span class="amp-icon amp-warning"></span>
+								<?php esc_html_e( 'Invalid markup not approved', 'amp' ); ?>
+							</strong>
+							<?php
+							esc_html_e( 'Please approve all the invalid markup before updating.', 'amp' );
```
westonruter: This should be iterated on. Something like this:

```suggestion
							esc_html_e( 'Review each instance of invalid markup and determine whether the action of removing (or keeping) is approved. You can change the status and click preview to see what impact the invalid markup has on the page. Invalid markup that is not approved is markup which you may not have seen yet or which you want to fix directly in underlying code.', 'amp' );
```

src/Icon.php
pierlon: Done via https://github.com/ampproject/amp-wp/pull/4382/commits/c615be8243bceff183c0ba88e96ea9d5d207c24b.

src/Icon.php
pierlon: Updated comment, see https://github.com/ampproject/amp-wp/pull/4382/commits/c615be8243bceff183c0ba88e96ea9d5d207c24b.

includes/validation/class-amp-validation-error-taxonomy.php
pierlon: I've moved them to a helper function in the `Icon` class, see https://github.com/ampproject/amp-wp/pull/4382/commits/c615be8243bceff183c0ba88e96ea9d5d207c24b.

src/Icon.php
```diff
@@ -0,0 +1,136 @@
...
+	 * Indicates there are validation errors which have not been explicitly accepted.
+	 */
+	const WARNING = 'amp-warning';
+
+	/**
+	 * Current icon.
+	 *
+	 * @var string
+	 */
+	private static $icon;
```
schlessera: No need for this to be static. You have named constructors below, so you are always instantiating an actual object instance.

src/Icon.php
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}
+
+	/**
+	 * Render icon as HTML.
+	 *
+	 * @param array $attributes List of attributes to add to HTML output.
+	 * @return string Rendered HTML.
+	 */
+	public static function to_html( $attributes = [] ) {
```
schlessera: This method shouldn't be static. I don't think it makes sense to render an "undefined icon" as HTML.

src/Services.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Interface BackgroundTask.
+ * Class Services.
```
schlessera, line 3: Good catch! I'm the king of copypasta, it seems! :)

src/Icon.php
```diff
@@ -0,0 +1,136 @@
...
+				if ( ! $attribute ) {
+					return '';
+				}
+
+				return $attributes_string . sprintf( '%s="%s" ', $attribute, $attributes[ $attribute ] );
+			},
+			''
+		);
+
+		self::$icon = null;
```
schlessera: This is dangerous code because of the use of static state. If someone calls `to_html()` on the same Icon twice, it will produce different results...

src/Icon.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @var string
+	 */
+	private static $icon;
+
+	/**
+	 * Invalid icon.
+	 *
+	 * @return Icon
+	 */
+	public static function invalid() {
```
schlessera, line 58: The named constructors should not set static state, this can get messy easily. They should just call the regular constructor with the correct argument.

If you want to avoid an open constructor that requires additional validation, you can make the constructor `private`, which will force consumers to pass through one of these named constructors.

src/Icon.php
pierlon: Done via https://github.com/ampproject/amp-wp/pull/4382/commits/04eacacab8ff40899eeb41c79f3801137882afb5.

src/Icon.php
```diff
@@ -0,0 +1,142 @@
...
+ */
+
+namespace AmpProject\AmpWP;
+
+/**
+ * Icons used to visually represent the state of a validation error.
+ *
+ * @package AmpProject\AmpWP
+ */
+class Icon {
```
schlessera: Class should be `final` by default:
```suggestion
final class Icon {
```

Otherwise, you'd need to change the calls to `self::` to `static::` to use late static binding so an extending class would not unexpectedly break.

This would mean we advertise this class for extension, though, and we should only do that if the actual benefit outweighs the additional maintenance cost.

src/Icon.php
```diff
@@ -0,0 +1,142 @@
...
+	 *
+	 * @param array $attributes List of attributes to add to HTML output.
+	 * @return string Rendered HTML.
+	 */
+	public function to_html( $attributes = [] ) {
+		if ( ! $this->icon ) {
+			// Silently fail.
+			return '';
+		}
+
```
schlessera: Not needed, as we ensured we can only ever have a valid instance.
```suggestion
```

src/Icon.php
```diff
@@ -0,0 +1,142 @@
...
+			array_keys( $attributes ),
+			static function ( $attributes_string, $attribute ) use ( $attributes ) {
+				if ( ! $attribute ) {
+					return '';
+				}
+
+				return $attributes_string . sprintf( '%s="%s" ', $attribute, $attributes[ $attribute ] );
+			},
+			''
+		);
```
schlessera: This will leave an unneeded `' '` at the end.

How about something like this:
```suggestion
		$attributes_string = implode(
			' ',
			array_map(
				static function ( $key, $value ) {
					return sprint(
						'%s="%s"',
						htmlspecialchars( $key ),
						htmlspecialchars( $value )
					);
				},
				array_keys( $attributes ),
				$attributes
			);
		);
```

amedina:
It seems the semantics of the icons changed. When I add an invalid markup, the plugin removes it (it indicates "Invalid markup removed: 1") but still the icon is a reddish warning sign. Need to discuss this. 

Screen Shot 2020-04-25 at 9 14 02 AM

amedina:
The color of the red circle with the exclamation point should be `#ff0000` (right now it is orangy). And the exclamation mark should be `#ffffff`

Screen Shot 2020-04-25 at 9 12 02 AM

amedina:
The warning semantics should be conveyed with both color and shape. The triangular yellow warning sign achieved that. We should maintain that semantic message.

This is also achieved by making the circle yellow:

- red circle + x ==> Invalid
- yellow circle + ! ==> Valid but warning, there may be problems
- green circle + checkmark ==> Valid

amedina:
At the same time I see the orange warning circle in the Validated URL screen, I see this in the admin bar. There seems to be an issue with the icon, and also a semantic mismatch between the admin bar and the Validated URL.

Update: Clearing the browser caches does not fix it.

Screen Shot 2020-04-25 at 9 18 26 AM

assets/css/src/amp-icons.css
```diff
@@ -0,0 +1,48 @@
...
+
+	/* Accent Red */
+	color: #dc3232 !important;
+}
+
+.amp-icon.amp-warning::before {
+	content: "\f534";
+
+	/* Accent Orange */
+	color: #f56e28 !important;
```
amedina: Let's keep the color semantics we had before. Accent orange does not convey the desired semantics. This should be around `#ffff00`.

assets/css/src/amp-icons.css
```diff
@@ -0,0 +1,48 @@
...
+
+	/* Accent Green */
+	color: #46b450 !important;
+}
+
+.amp-icon.amp-invalid::before {
+	content: "\f153";
+
+	/* Accent Red */
+	color: #dc3232 !important;
```
amedina, line 31: Let's keep the color semantics we had before. This should be `#ff0000`.

assets/css/src/amp-icons.css
```diff
@@ -0,0 +1,48 @@
...
+
+/*
+ * Colors are taken from the official WordPress color palette.
+ * See https://make.wordpress.org/design/handbook/design-guide/foundations/colors/#auxiliary-hues
+ */
+.amp-icon.amp-valid::before {
+	content: "\f12a";
+
+	/* Accent Green */
+	color: #46b450 !important;
```
amedina, line 24: Let's keep the color semantics we had before. This should be `#00ff00`.

src/Icon.php
```diff
@@ -0,0 +1,142 @@
...
+	public static function warning() {
+		return new self( self::WARNING );
+	}
+
+	/**
+	 * Get color for current icon.
+	 *
+	 * @return string Hex color for icon.
+	 */
+	public function get_color() {
```
amedina, line 94: Let's discuss/review the color assignments.

amedina:
I find the circles in all icons a bit boring. The use of different shapes together with colors make the interface more appealing. We may go with this and follow up with improvements as part of the upcoming UXD work.

assets/css/src/amp-icons.css
pierlon: That color might be a bit pale when in use, for example:

image

The current colors for the icons are taken from the official [WordPress color palette](https://make.wordpress.org/design/handbook/design-guide/foundations/colors/). I do agree  with the warning icon looking a bit reddish, would the Accent Yellow with 20% tint (`#ffeab3`) suffice?

image

pierlon:
> I find the circles in all icons a bit boring. The use of different shapes together with colors make the interface more appealing.

Yes, that's a limitation with the current set of Dashicons, unfortunately. If such customization is desired, I suppose the best way to go forward would be to replace these icons with SVGs. That would give us much more flexibility with the shapes. Another benefit would be not having to rely upon the Dashicons font being loaded.

Aside: The Dashicons project is [no longer accepting icon requests](https://make.wordpress.org/design/2020/04/20/next-steps-for-dashicons/) and efforts are now being focused on the Gutenberg [Icons package](https://github.com/WordPress/gutenberg/tree/master/packages/icons).

assets/css/src/amp-icons.css
amedina: That would work.

amedina:
The semantics of 'Approve' on error entries in the Error Index page is confusing.  

`Approve` should be associated with approving the action taken by the plugin. That semantics got overloaded with toggling the "New" status of an error and it does not work well.

westonruter:
Copying some of my messages from chat…

One thing we talked about is changing “Approved” terminology to “Reviewed”

Another thing we talked about is, on the Error Index screen, to iterate on the Approve/Unapprove and Remove/Keep row action links. In 1.5.3, these links can be either Confirm Removed, Confirm Kept, Remove, or Keep. Can this sense be retained somehow? Still not sure entirely about this.

It's also important that we make sure the Validated URL screen semantics works the best, as that is the screen users will most often be interacting with. The error index screen is more included for completeness as a way to find the validated URLs that an error occurs on. Manipulating the status of errors on this screen should be rarely done because (1) the source stack information is not available for the errors, and (2) there is no way to preview the effect of removing/keeping from that context, as this can be only done in the context of an individual URL.

This leads me to think that perhaps we should actually eliminate row actions from the error index entirely. A user should only be able to confirm/approve removal in the context of a Validated URL. 

We're probably doing users a disservice even to provide them with a way to toggle the status of validation errors outside the context of the validated URL screen.

So the row action for a validated URL should be to take the user to Details of the error, which is to say to take the user to the URLs that have the error.

If no URLs have the error, then there could be a new row action to Forget an individual validation error, which would do do the same as clicking the Clear Empty button.

Doing this would remove the complexity of the validation screens, and it would consistently draw the user into the workflow on the validated URL screen.

includes/validation/class-amp-validation-manager.php
```diff
@@ -245,6 +246,20 @@ function( $query_vars ) {
...
 
 			add_action( 'admin_bar_menu', [ __CLASS__, 'add_admin_bar_menu_items' ], 101 );
+
+			add_action(
+				'admin_bar_init',
+				static function () {
+					wp_enqueue_style(
+						'amp-icons',
+						amp_get_asset_url( 'css/amp-icons.css' ),
+						[ 'dashicons' ],
```
westonruter:
```suggestion
						[ 'admin-bar', 'dashicons' ],
```

pierlon: Done in https://github.com/ampproject/amp-wp/pull/4382/commits/335e863ea78dbdeb69abae8cfa26bee7b2844ede.

pierlon:
Once https://github.com/ampproject/amp-wp/pull/4642 is merged I'll merge those changes here to get these tests passing.

westonruter:
As discussed over chat, two more things:

- [x] Restore the markup status on the validation error details screen. Instead of removing it, move it to the `dl` below that has the details of the error.
- [x] While at it, unwrap the `details` from the details as there's no need to be able to collapse the section. No need for the `summary` element there either.

westonruter:
To clarify, the validation error status should be moved as follows:

image

westonruter:
- [x] The 2px margin between the warning icon and AMP icon and the text is a bit too small. Please increase to 4px.

image

pierlon:
Markup status being moved below to the error details section, and also unwrapping `<details>`:

image

assets/src/amp-validation/amp-validated-url-post-edit-screen.js
```diff
@@ -283,32 +285,35 @@ const handleSearching = () => {
 const updateSelectIcon = ( select ) => {
```
westonruter: This function should be renamed, correct? Like `updateSelectBorderColor`?

assets/src/amp-validation/amp-validation-error-edit-screen.js
```diff
@@ -0,0 +1,34 @@
...
+ * WordPress dependencies
+ */
+import domReady from '@wordpress/dom-ready';
+import { __ } from '@wordpress/i18n';
+
+domReady( () => {
+	const errorDetailList = document.querySelector( 'details > dl.detailed' );
+
+	addAdditionalErrorDetails( errorDetailList );
+	unwrapErrorDetails( errorDetailList );
```
westonruter: As opposed to doing this with JS can not the `details` just be omitted when PHP generates it from the start? As it stands right now, there is a FODC (flash of `detail`'ed content)

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -77,20 +77,15 @@ class AMP_Validated_URL_Post_Type {
...
-	const VALIDATION_ERRORS_META_BOX = 'amp_validation_errors';
-
 	/**
 	 * The transient key to use for caching the number of URLs with new validation errors.
 	 *
 	 * @var string
 	 */
 	const NEW_VALIDATION_ERROR_URLS_COUNT_TRANSIENT = 'amp_new_validation_error_urls_count';
 
+	const VALIDATION_ERRORS_INPUT_KEY = 'validation_errors';
```
westonruter: Add phpDoc for this?

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -995,7 +994,18 @@ public static function add_single_post_columns() {
...
+				)
+			),
+			'reviewed'                    => sprintf(
+				'%s<span class="dashicons dashicons-editor-help tooltip-button" tabindex="0"></span><div class="tooltip" hidden data-content="%s"></div>',
+				esc_html__( 'Reviewed', 'amp' ),
+				esc_attr(
+					sprintf(
+						'<h3>%s</h3><p>%s</p>',
+						esc_html__( 'Reviewed', 'amp' ),
+						esc_html__( 'Confirm that the status of the validation error has been seen and approved of.', 'amp' )
```
westonruter:
```suggestion
						esc_html__( 'Confirm that the action being taken on the invalid markup (causing a validation error) has been seen and approved.', 'amp' )
```

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -1936,11 +1955,31 @@ public static function print_status_meta_box( $post ) {
...
+								<span class="amp-icon amp-warning"></span>
+								<?php esc_html_e( 'Invalid markup not reviewed', 'amp' ); ?>
+							</strong>
+							<?php
+							esc_html_e(
+								'Review each instance of invalid markup and determine whether the action of removing
+								(or keeping) is approved. You can change the status and click preview to see what
+								impact the invalid markup has on the page. Invalid markup that is not approved is
+								markup which you may not have seen yet or which you want to fix directly in
+								underlying code.',
```
westonruter: I think the linebreaks perhaps should be removed? I don't recall translation strings having these. The extra tabs may confuse translators. I think it should all be on one line.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -2035,6 +2025,17 @@ static function( $error ) use ( $term ) {
 					$content .= esc_html__( 'Misc', 'amp' );
 				}
 				break;
+			case 'reviewed':
+				if ( 'post.php' === $pagenow ) {
+					$checked    = ( (int) $term->term_group & self::ACKNOWLEDGED_VALIDATION_ERROR_BIT_MASK ) ? 'checked="checked"' : '';
```
westonruter: This could use the `checked()` helper function if desired.

assets/css/src/amp-validation-single-error-url.css
```diff
@@ -152,4 +152,11 @@ table.striped > tbody > tr.even {
 	margin-left: 5px;
 	width: auto;
 	float: none;
+	border-width: 2px;
+}
+
+#amp-invalid-markup > .amp-icon::before {
+	margin-right: 2px;
```
westonruter:
```suggestion
	margin-right: 3px;
```

assets/src/amp-validation/amp-validation-error-edit-screen.js
pierlon: Done in https://github.com/ampproject/amp-wp/pull/4382/commits/34e3e806433000e0758ff0af68dcde6dc1b94818.

assets/src/amp-validation/amp-validated-url-post-edit-screen.js
pierlon: Right, missed that one :sweat_smile:.

assets/src/amp-validation/amp-validation-error-edit-screen.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * WordPress dependencies
+ */
+import domReady from '@wordpress/dom-ready';
+import { __ } from '@wordpress/i18n';
+
+domReady( () => {
+	const errorDetailList = document.querySelector( 'dl.detailed' );
+	addAdditionalErrorDetails( errorDetailList );
```
westonruter: As opposed to using JS for this, couldn't the new `dt`/`dd` be added in PHP by injecting them before the first `<dt>`?

pierlon: Done in https://github.com/ampproject/amp-wp/pull/4382/commits/f571c17a827cc209c63bbf55c562667568b209ea.

includes/validation/class-amp-validation-manager.php
```diff
@@ -245,6 +246,20 @@ function( $query_vars ) {
...
+					wp_enqueue_style(
+						'amp-icons',
+						amp_get_asset_url( 'css/amp-icons.css' ),
+						[ 'dashicons', 'admin-bar' ],
+						AMP__VERSION
+					);
+
+					wp_styles()->add_data( 'amp-icons', 'rtl', 'replace' );
+				}
+			);
```
westonruter: @pierlon I found an issue. This is not currently getting run in Reader mode. When Reader mode is selected, then no Link icon appears in the header.

I tried moving this code to `amp_init` and that fixed the issue, but then it revealed two others:

1. The `amp-icons` stylesheet here is not only used for the admin bar but also for other parts of the screens.
2. The stylesheet was being added to every screen in the admin (and the frontend) regardless of whether it was being used.

I've implemented a fix in c4cb9ab47f639070a8410c2fa4cf64da1c20c377.

assets/src/amp-validation/amp-validated-url-post-edit-screen.js
```diff
@@ -56,12 +56,14 @@ const addBeforeUnloadPrompt = () => {
...
- * unless the user is pressing the trash link or update button.
+ * Add an beforeunload warning when there are unsaved changes for the markup or review status.
  */
 const watchForUnsavedChanges = () => {
 	const onChange = ( event ) => {
-		if ( event.target.matches( 'select' ) && event.target.getAttribute( 'id' ) !== 'amp_validation_error_type' ) {
+		if (
+			event.target.matches( '.amp-validation-error-status' ) ||
+			event.target.matches( '.amp-validation-error-status-review' )
+		) {
```
schlessera, line 66: This is overly simplistic and leads to cases where the warning appears even though, for the user, no change has happened as whatever the user did he also undid again.

Screen Recording 2020-05-04 at 11 18 AM

Fixing this might be overly complicated for now, so I'm fine with keeping as-is as long as we're aware of that issue.

westonruter: That's a good point, but it's the same behavior as the block editor. You can make a change and then either manually undo the change or click the Undo button, and in both cases clicking reload will show the same AYS dialog. So it's not any worse than what users expect elsewhere in WP.

PR #4384, opened 2020-03-14
Remove registration of Facebook embed handler

### LINKED ISSUE #4358: amp-facebook is wrapped in a <p> tag when it should be in post root like amp-twitter
## Bug Description
Screen Shot 2020-03-02 at 6 49 51 PM

As you can see, the `amp-facebook` element is inside a `<p>`. To me this is not very elegant and it's also different from the `amp-twitter` output, which is in the post root:

Screen Shot 2020-03-02 at 6 51 43 PM

The specific thing I want to do is style the `blockquote.translation` that comes after the embed, which should be possible with `amp-facebook+blockquote.translation`, but becomes essentially impossible due to the wrapping `<p>`. 

## Expected Behaviour

Seems to me that both `amp-facebook` and `amp-twitter` should render without the containing `<p>`. 

Maybe I'm wrong? Is it maybe something wrong with my setup?

#### Example embed code for twitter:

```
<blockquote class="twitter-tweet" data-lang="en">
<p dir="ltr" lang="en">Facebook Instant Articles doesn't allow horizontal rules because lines are dangerous and need to be kept under control.</p>
— jeremyclarke (@jeremyclarke) <a href="https://twitter.com/jeremyclarke/status/826895852350345217">February 1, 2017</a></blockquote>
<blockquote class="translation">Les Articles Instants de Facebook ne permet pas de règles horizontales car les lignes sont dangereuses et doivent être gardées sous contrôle.</blockquote>
```

#### Example embed code for Facebook

```
[embed]https://www.facebook.com/globalvoicesonline/videos/10154528534218876/[/embed]
<blockquote class="translation">This is a translation of the facebook thing above.</blockquote>
```

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

This PR removes the custom WordPress embed handler for Facebook and instead relies on `\AMP_Facebook_Embed_Handler::sanitize_raw_embeds` to convert such embeds to their appropriate AMP components.

Fixes #4358.

## Checklist

### DIFF
--- a/includes/embeds/class-amp-facebook-embed.php
+++ b/includes/embeds/class-amp-facebook-embed.php
@@ -11,15 +11,6 @@ use AmpProject\Dom\Document;
  * Class AMP_Facebook_Embed_Handler
  */
 class AMP_Facebook_Embed_Handler extends AMP_Base_Embed_Handler {
-	const URL_PATTERN = '#https?://(www\.)?facebook\.com/.*#i';
-
-	/**
-	 * Default width.
-	 *
-	 * @var int
-	 */
-	protected $DEFAULT_WIDTH = 600;
-
 	/**
 	 * Default height.
 	 *
@@ -45,57 +36,14 @@ class AMP_Facebook_Embed_Handler extends AMP_Base_Embed_Handler {
 	 * Registers embed.
 	 */
 	public function register_embed() {
-		wp_embed_register_handler( $this->amp_tag, self::URL_PATTERN, [ $this, 'oembed' ], -1 );
+		// Not implemented.
 	}
 
 	/**
 	 * Unregisters embed.
 	 */
 	public function unregister_embed() {
-		wp_embed_unregister_handler( $this->amp_tag, -1 );
-	}
-
-	/**
-	 * WordPress OEmbed rendering callback.
-	 *
-	 * @param array  $matches URL pattern matches.
-	 * @param array  $attr    Matched attributes.
-	 * @param string $url     Matched URL.
-	 * @return string HTML markup for rendered embed.
-	 */
-	public function oembed( $matches, $attr, $url ) {
-		return $this->render( [ 'url' => $url ] );
-	}
-
-	/**
-	 * Gets the rendered embed markup.
-	 *
-	 * @param array $args Embed rendering arguments.
-	 * @return string HTML markup for rendered embed.
-	 */
-	public function render( $args ) {
-		$args = wp_parse_args(
-			$args,
-			[
-				'url' => false,
-			]
-		);
-
-		if ( empty( $args['url'] ) ) {
-			return '';
-		}
-
-		$this->did_convert_elements = true;
-
-		return AMP_HTML_Utils::build_tag(
-			$this->amp_tag,
-			[
-				'data-href' => $args['url'],
-				'layout'    => 'responsive',
-				'width'     => $this->args['width'],
-				'height'    => $this->args['height'],
-			]
-		);
+		// Not implemented.
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/embeds/class-amp-facebook-embed.php
```diff
@@ -136,7 +84,13 @@ public function sanitize_raw_embeds( Document $dom ) {
 				$scripts[] = $script;
 			}
 			foreach ( $scripts as $script ) {
-				$script->parentNode->removeChild( $script );
+				$parent_node = $script->parentNode;
+				$parent_node->removeChild( $script );
+
+				// Remove parent node if it is an empty <p> tag.
+				if ( 'p' === $parent_node->tagName && null === $parent_node->firstChild ) {
```
pierlon: The `script` tag is wrapped with a `p` tag (from `wpautop()`), so I've removed the lingering empty tag if it's there.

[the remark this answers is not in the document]
pierlon: Its also an issue for [TikTok](https://github.com/ampproject/amp-wp/blob/c6314bfcb853fb70b6232dd42ac9f9789301f856/includes/embeds/class-amp-tiktok-embed-handler.php#L130-L137), [Twitter](https://github.com/ampproject/amp-wp/blob/c6314bfcb853fb70b6232dd42ac9f9789301f856/includes/embeds/class-amp-twitter-embed-handler.php#L239-L246) and [Instagram](https://github.com/ampproject/amp-wp/blob/c6314bfcb853fb70b6232dd42ac9f9789301f856/includes/embeds/class-amp-instagram-embed.php#L206-L213).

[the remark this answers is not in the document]
pierlon: After looking at some of the other embeds, they are for the most part also wrapped with a `<p>` tag. That isn't an issue per se, but if this PR is merged we should probably standardize the other embeds to not be wrapped with a paragraph element.

westonruter:
Humm, there are build errors.

includes/embeds/class-amp-facebook-embed.php
```diff
@@ -131,12 +79,20 @@ public function sanitize_raw_embeds( Document $dom ) {
...
+			 */
+			$script_query = $dom->xpath->query( '//script[ starts-with( @src, "https://connect.facebook.net" ) and contains( @src, "sdk.js" ) ]' );
+			foreach ( $script_query as $script ) {
+				$parent_node = $script->parentNode;
+				$parent_node->removeChild( $script );
+
+				// Remove parent node if it is an empty <p> tag.
+				if ( 'p' === $parent_node->nodeName && null === $parent_node->firstChild ) {
+					$parent_node->parentNode->removeChild( $parent_node );
+				}
```
westonruter: I don't see a test for this.

includes/embeds/class-amp-facebook-embed.php
```diff
@@ -131,12 +79,20 @@ public function sanitize_raw_embeds( Document $dom ) {
...
 			}
 			$fb_root->parentNode->removeChild( $fb_root );
```
westonruter: It turns out that when you have multiple Facebook embeds on a page, each instance will include a `div#fb-root`. So I suppose this should be changed to:

```suggestion
			while ( $fb_root ) {
				$fb_root->parentNode->removeChild( $fb_root );
				$fb_root = $dom->getElementById( 'fb-root' );
			}
```

pierlon: In theory this should work, but when it retrieves the `#fb_root` `div` again it's returning the same `div` as before. I'll adapt this to use an XPath query instead.

includes/embeds/class-amp-facebook-embed.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/4384/commits/c6869a9380c7842ed0f176bff2a0579012fff3a0.

includes/embeds/class-amp-facebook-embed.php
westonruter: Why would it return the same `div` as before if it was removed from the document?

westonruter:
Note: It appears Facebook tests don't have mocks for the non-`external-http` test suite.

PR #4385, opened 2020-03-14
Improve security of external href (make it harder for malicious people).

netchild:
Initially I reported this via the Google Seucrity program (in
https://issuetracker.google.com/issues/151134576), but it was
rejected with this explanation:
https://sites.google.com/site/bughunteruniversity/nonvuln/phishing-with-window-opener

While the change may not fix all possible attack vectors, it at least
takes care about one (window.opener). It adds one more layer to the
security onion, and makes it one step harder for malicious people.

The probability that this issues can be exploitet is low, as amp-wp.org
needs to be compromised, and an admin of an wordpress installation needs
to click on the getting started link.

The security impact is medium, as a breach can theoretically affect a
lot of websites (currently 400.000+ installations counted on the
wordpress plugin page for this plugin) and there are other ways to
do malicious things which are not prevented by this change.

There are also static security scanners (commercial and open source)
which flag this as a blocking security issue. Adding this change will
make that this is not showing up as an issue in such scans, which may
make a difference for companies where such scans are mandatory to not
show up security issues.

All in all this is a security related improvement, but not a big security
hole.

## Summary

Fixes #

## Checklist

### DIFF
--- a/includes/options/class-amp-options-manager.php
+++ b/includes/options/class-amp-options-manager.php
@@ -413,7 +413,7 @@ class AMP_Options_Manager {
 			<h2><?php esc_html_e( 'Welcome to AMP for WordPress', 'amp' ); ?></h2>
 			<h3><?php esc_html_e( 'Bring the speed and features of the open source AMP project to your site, complete with the tools to support content authoring and website development.', 'amp' ); ?></h3>
 			<h3><?php esc_html_e( 'From granular controls that help you create AMP content, to Core Gutenberg support, to a sanitizer that only shows visitors error-free pages, to a full error workflow for developers, this release enables rich, performant experiences for your WordPress site.', 'amp' ); ?></h3>
-			<a href="https://amp-wp.org/getting-started/" target="_blank" class="button button-primary"><?php esc_html_e( 'Learn More', 'amp' ); ?></a>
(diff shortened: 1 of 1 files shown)


netchild:
@googlebot I signed it!

westonruter:
Thanks for the PR.

There are other links to https://amp-wp.org than just this one, however. For example, in `class-amp-options-menu.php` and `class-amp-form-sanitizer.php`. 

There are also links to https://amp.dev in various places. 

Both amp.dev and amp-wp.org are domains that are managed by the AMP team (and my team), thus the possibly to exploit is exceedingly low, so low that I don't think it is worth it to try to maintain.

WordPress core itself does not seem to consider this worthy of doing, as the WordPress link in the footer of each admin page links off to https://wordpress.org without any such `noreferrer` or `noopener`.

So I believe we'll close this PR for the same reason outlined in https://sites.google.com/site/bughunteruniversity/nonvuln/phishing-with-window-opener

netchild:
FYI: There also seems to be a performance benefit in various browsers when using rel=noopener. See https://developers.google.com/web/tools/lighthouse/audits/noopener and https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/ for more on that.

PR #4388, opened 2020-03-15
Prevent failed attempts to override the reply-title heading in comment form

### LINKED ISSUE #2489: Comment form “Leave a Reply” heading misbehaves in Twenty Nineteen
The `#reply-title` element is not getting the correctly populated in Twenty Nineteen. It is being output as:

```html
<span [text]='commentform_post_2013.replyToName ? "Leave a Reply to " + commentform_post_2013.replyToName + "" : "Leave a Reply"'></span>
```

Notice the text content of the `span` is empty, but no value of `[text]` is empty.

In `#development=1` mode, this is generating a warning:

> `Default value () does not match first result (Leave a Reply) for <SPAN [text]="commentform_post_2013.replyToName ? "Leave a Reply to " + commentform_post_2013.replyToName + "" : "Leave a Reply"">. We recommend writing expressions with matching default values, but this can be safely ignored if intentional.`

It also exhibits a strange behavior when adding comment text to the form. 

Upon entering text:

> image

Then hit tab to go to the Submit button, and the “Leave a Reply” heading all of a sudden appears:

> Screen Shot 2019-06-01 at 23 42 35

The problem here is that Twenty Nineteen renders its comment form via `twentynineteen_comment_form()` which does:

```php
comment_form(
	array(
		'logged_in_as' => null,
		'title_reply'  => null,
	)
);
```

In particular, the `title_reply` here is supplied as `null`. This is causing problems in the `\AMP_Theme_Support::filter_comment_form_defaults()` method which is doing the following:

https://github.com/ampproject/amp-wp/blob/3985eaef6f5c412f52b0e6eed992d19d9729b89b/includes/class-amp-theme-support.php#L1162-L1191

The problem is that `comment_form()` applies filters on `comment_form_defaults`, not on the _actual_ args passed to `comment_form()`. Therefore, we don't actually know what value for `title_reply` was passed. 

This hasn't been a problem before Twenty Nineteen because themes normally are calling `comment_form()` without any args at all.

westonruter:
## Summary

This is a short-term fix to prevent attempting to integrate amp-bind with the  “Leave a Reply” heading in the comment form. Because WordPress only provides a `comment_form_defaults` filter for the `comment_form()` function, and not a filter like `comment_form_args` to modify the actual args passed in, we cannot use the `comment_form_defaults` filter to inject amp-bind logic into the form.

Longer-term solutions would be to introduce the `comment_form_args` filter to core, or even better to not use amp-bind and instead use amp-script for this.

The effect of this PR is that the “Leave a Reply” heading will remain unchanged in Twenty Nineteen and Twenty Twenty when you enter a comment or click a “Reply” link in the comments list. In Twenty Seventeen, which does not pass any arguments to `comment_form()`, the heading will change from “Leave a Reply” to “Leave a Reply to John Smith” when clicking the reply link. In all themes, the “Cancel Reply” link should still appear regardless of the heading changing.

Additionally, this PR prevents the amp-bind warning message from appearing:

> `Default value () does not match first result (Leave a Reply) for <SPAN [text]="commentform_post_2013.replyToName ? "Leave a Reply to " + commentform_post_2013.replyToName + "" : "Leave a Reply"">. We recommend writing expressions with matching default values, but this can be safely ignored if intentional.`

Fixes #2489.

## Checklist

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -1130,7 +1130,7 @@ class AMP_Theme_Support {
 		add_action( 'template_redirect', [ __CLASS__, 'start_output_buffering' ], $priority );
 
 		// Commenting hooks.
-		add_filter( 'comment_form_defaults', [ __CLASS__, 'filter_comment_form_defaults' ] );
+		add_filter( 'comment_form_defaults', [ __CLASS__, 'filter_comment_form_defaults' ], PHP_INT_MAX );
 		add_filter( 'comment_reply_link', [ __CLASS__, 'filter_comment_reply_link' ], 10, 4 );
 		add_filter( 'cancel_comment_reply_link', [ __CLASS__, 'filter_cancel_comment_reply_link' ], 10, 3 );
 		add_action( 'comment_form', [ __CLASS__, 'amend_comment_form' ], 100 );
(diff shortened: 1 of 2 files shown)


includes/class-amp-theme-support.php
```diff
@@ -1338,29 +1338,48 @@ public static function get_comment_form_state_id( $post_id ) {
...
+	 * @param array $default_args Comment form arg defaults.
 	 * @return array Filtered comment form args.
 	 */
-	public static function filter_comment_form_defaults( $args ) {
-		$state_id = self::get_comment_form_state_id( get_the_ID() );
+	public static function filter_comment_form_defaults( $default_args ) {
 
+		// Obtain the actual args provided to the comment_form() function since it is not available in the filter.
+		$args      = [];
+		$backtrace = debug_backtrace(); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_debug_backtrace -- Due to limitation in WordPress core.
```
pierlon: Ignoring the `args` index for each backtrace should save memory.

```suggestion
		$backtrace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_debug_backtrace -- Due to limitation in WordPress core.
```

westonruter: The args are the specific thing that we need to capture.

includes/class-amp-theme-support.php
```diff
@@ -1338,29 +1338,48 @@ public static function get_comment_form_state_id( $post_id ) {
...
+		foreach ( $backtrace as $call ) {
+			if ( 'comment_form' === $call['function'] ) {
+				$args = isset( $call['args'][0] ) ? $call['args'][0] : [];
```
westonruter: Note args

pierlon: Ah right, I did not see that :sweat: .

laurelfulford:
Thanks Weston! 

I've tested this PR with Twenty Nineteen, and with the Newspack theme and [our plugin that renames comment labels](https://github.com/Automattic/newspack-rename-comments). In both cases, it worked like a charm -- the duplicate title no longer happened, and the custom text you can add with the Rename Comments plugin was respected 👍

PR #4404, opened 2020-03-18
Use Cache-Control directive for external stylesheet caching

### LINKED ISSUE #4293: Improve caching of externally-fetched stylesheets
When `AMP_Style_Sanitizer::fetch_external_stylesheet()` fails to fetch an external stylesheet, it caches the a `WP_Error` in the transient as opposed to the stylesheet string. In both cases, the transient expires after a month. This should be more dynamic, namely:

* Fetches that result in failures should be cached for a `DAY_IN_SECONDS` rather than `MONTH_IN_SECONDS`.
* Fetches that are successful should actually use a transient expiration that reflects the `Cache-Control` response header, as long as the `max-age` is greater than some minimum amount (e.g. 1 day).

See also this todo:

https://github.com/ampproject/amp-wp/blob/a3aabc7c8115734a9c15a4f85a19d5117c892595/includes/sanitizers/class-amp-style-sanitizer.php#L1363

Originally referenced in in https://github.com/ampproject/amp-wp/issues/2449#issuecomment-496733384.

schlessera:
## Summary

This PR does the following:
- Moves `CachedRemoteGetRequest` and `CachedData` into a subnamespace `AmpProject\AmpWP\RemoteRequest`
- Adds a `WpHttpRemoteGetRequest` implementation using the WordPress `WP_Http` abstraction to implement `RemoteGetRequest`
- Makes the `CachedRemoteGetRequest` implementation more robust
- Uses `WpHttpRemoteGetRequest` for remote requests in the optimizer
- Uses `WpHttpRemoteGetRequest` to fetch and cache external stylesheets

The `WpHttpRemoteGetRequest` implementation will respect the `max-age` directive of the `Cache-Control` header and will automatically retry failed remote requests for the following status codes:
- `WP_Http::REQUEST_TIMEOUT` (408)
- `WP_Http::LOCKED` (423)
- `WP_Http::TOO_MANY_REQUESTS` (429)
- `WP_Http::INTERNAL_SERVER_ERROR` (500)
- `WP_Http::SERVICE_UNAVAILABLE` (503)
- `WP_Http::GATEWAY_TIMEOUT` (504)

Fixes #4293

## Checklist


westonruter:
@schlessera I'm getting warnings still, but different than before:

```
[18-Mar-2020 16:55:25 UTC] PHP Warning:  array_keys() expects parameter 1 to be array, object given in /app/public/content/plugins/amp/lib/common/src/RemoteRequest/RemoteGetRequestResponse.php on line 178
[18-Mar-2020 16:55:25 UTC] PHP Warning:  array_combine() expects parameter 1 to be array, null given in /app/public/content/plugins/amp/lib/common/src/RemoteRequest/RemoteGetRequestResponse.php on line 178
[18-Mar-2020 16:55:25 UTC] PHP Warning:  array_change_key_case() expects parameter 1 to be array, null given in /app/public/content/plugins/amp/lib/common/src/RemoteRequest/RemoteGetRequestResponse.php on line 179
[18-Mar-2020 16:55:25 UTC] PHP Warning:  array_key_exists() expects parameter 2 to be array, null given in /app/public/content/plugins/amp/lib/common/src/RemoteRequest/RemoteGetRequestResponse.php on line 98
```

I can reproduce this each time after I `wp transient delete --all` and then access an AMP page.

westonruter:
I've determined the cause to be attempted passing `headers` as  `Requests_Utility_CaseInsensitiveDictionary` instead of as an array. Fixed in 2f0652a by converting to an array. You may have a better way to do that however.

schlessera:
I'll add a regression test and fix it.

westonruter:
> Maybe check first whether it actually is an iterator object, as it could also still be an array instead.

Good point. Since `is_iterable()` is only available in PHP 7, what about:

```php
$headers = is_object( $response['headers'] ) ? iterator_to_array( $response['headers'] ) : (array) $response['headers'];
```

Anyway, I'll leave you to it 😄

schlessera:
The correct interface to check would be https://www.php.net/manual/en/class.traversable.php

westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/4351156/amp.zip) (v1.5.0-alpha-20200318T200637Z-f7034c1c7)

src/RemoteRequest/CachedRemoteGetRequest.php
```diff
@@ -100,10 +102,16 @@ public function get( $url ) {
...
 			$expiry      = $this->get_expiry_time( $response );
 			$cached_data = new CachedData( $response->getBody(), $expiry );
-			set_transient( $cache_key, serialize( $cached_data ) ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
+			$headers     = $response->getHeaders();
+
+			// Only store response if it was successful.
+			$status = $response->getStatusCode();
+			if ( $status >= 200 && $status < 300 ) {
+				set_transient( $cache_key, serialize( $cached_data ) ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
+			}
```
westonruter: What about if the request fails with every request? Doesn't this mean an HTTP request will be issued for every page load? Before this PR would cache failures for a month the same as successes. But #4293 notes:

> * Fetches that result in failures should be cached for a `DAY_IN_SECONDS` rather than `MONTH_IN_SECONDS`.

Is that not accounted for?

src/RemoteRequest/WpHttpRemoteGetRequest.php
```diff
@@ -0,0 +1,164 @@
...
+
+		if ( ! is_int( $retries ) || $retries < 0 ) {
+			$retries = self::DEFAULT_RETRIES;
+		}
+
+		$this->ssl_verify = $ssl_verify;
+		$this->timeout    = $timeout;
+		$this->retries    = $retries;
+
+		$this->wp_http = _wp_http_get_object();
```
westonruter: Question: `_wp_http_get_object()` is marked as `@private`. While WordPress themes and plugins rarely respect such things, there's a chance in the future it could go away, no? Why not use `wp_remote_get()` directly?

src/RemoteRequest/CachedRemoteGetRequest.php
westonruter: What if `$expiry` were split into two separate arguments, one for `$expiry_success` and another for `$expiry_failure`?

One more thing: what if the `max-age` in the `Cache-Control` is set to zero or something very short? We should not honor such max-ages. In that case, what about making `$expiry` (or `$expiry_success`) into a `$min_expiry`? This could then be paired with the `$use_cache_control` to be whatever the `max` value is.

I don't think we should allow for an expiry of less than a certain amount of time. This would go hand-in-hand with #2092, where if a `max-age` is `0` or very short, and the stylesheet is responding with random CSS, then it will cause transients to fill up.

schlessera: Right now it caches `DAY_IN_SECONDS` by default, _unless_ a non-zero `max-age` is provided via `cache-control`.

However, I can go with a default and a min expiry instead, to have a few basic guarantees.

src/RemoteRequest/WpHttpRemoteGetRequest.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4404/commits/163228630ecaf26bfd587807f2d1e1e8917a0f0c

src/RemoteRequest/CachedRemoteGetRequest.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4404/commits/b8e10da7b4adc2e8a03b458fa546c14318e1ebe9

src/RemoteRequest/CachedRemoteGetRequest.php
```diff
@@ -117,21 +147,41 @@ public function get( $url ) {
...
 	private function get_expiry_time( Response $response ) {
 		if ( $this->use_cache_control && $response->hasHeader( self::CACHE_CONTROL ) ) {
-			$max_age              = 0;
-			$cache_control_string = $response->getHeader( self::CACHE_CONTROL );
-			$cache_control_parts  = array_map( 'trim', explode( ',', $cache_control_string ) );
+			$max_age = $this->get_max_age( $response->getHeader( self::CACHE_CONTROL ) );
+
+			if ( $max_age >= $this->min_expiry ) {
+				return new DateTimeImmutable( "+ {$max_age} seconds" );
+			}
```
westonruter: Minor thing. Should this not rather be:

```suggestion
			$expiry = max( $this->min_expiry, $this->get_max_age( $response->getHeader( self::CACHE_CONTROL ) ) );
			return new DateTimeImmutable( "+ {$expiry} seconds" );
```

This will prevent a `max-age` of 10 minutes from automatically extending to the default expiry of 1 month, rather than the minimum expiry of 1 day.

schlessera: Ah, I see the problem. However, I don't think this is the solution, as you've eliminated the default expiry of 1 month altogether now. I'll update the PR to avoid the case you're trying to fix.

schlessera: Hmm, no, I' got confused with the alignment here in the PR. Let me take a look at my IDE instead...

westonruter: Is the default expiry needed then in this case? The default expiry is only needed when `max-age` is not used.

schlessera: Yes, indeed.

schlessera: However, the above code runs if `cache-control` is used. Is it guaranteed that it always contains a `max-age` directive?

westonruter: Right, it wouldn't. But then `get_max_age` would return `0` and then automatically the `min_expiry` would end up getting used.

kienstra:
**Testing looks good**

This looks good! 

1. An external URL that does exist (200 response) is cached as expected. It stores an expiration, and successive calls to [fetch_external_stylesheet()](https://github.com/ampproject/amp-wp/pull/4404/files#diff-924f049482f90f79f49e7346b3240a31L1404) return the cached value.

expiry-stored

2. An external URL that doesn't exist (404 response) is cached as expected, and this doesn't  [retry the request](https://github.com/ampproject/amp-wp/pull/4404/files#diff-b8ec7c1a036293a6b75635c308f5b6f6R44) again.

Here's a snippet I used to test this:

```php
add_action(
	'wp_enqueue_scripts',
	static function() {
		wp_enqueue_style( 'example-valid-external-style', 'https://cdnjs.cloudflare.com/ajax/libs/normalize/8.0.1/normalize.min.css' );
		wp_enqueue_style( 'non-existent-external-style', 'https://cdnjs.cloudflare.com/ajax/libs/modern-normalize/0.6.0/modern-normalize.min.css' );
	}
);
```

Sorry for the delay on this.

src/RemoteRequest/CachedRemoteGetRequest.php
```diff
@@ -97,13 +119,24 @@ public function get( $url ) {
...
+				$response = $this->remote_request->get( $url );
+				$status   = $response->getStatusCode();
+				$expiry   = $this->get_expiry_time( $response );
+				$headers  = $response->getHeaders();
+				$body     = $response->getBody();
+			} catch ( FailedToGetFromRemoteUrl $exception ) {
+				$status = $exception->getStatusCode();
+				$expiry = new DateTimeImmutable( "+ {$this->min_expiry} seconds" );
+				$body   = $exception->getMessage();
+			}
```
westonruter: @schlessera FYI, I added this in 00036be when testing whether caching worked as expected in the failure case, and I found it didn't. Please advise if anything looks off.

schlessera: I would hardcode `$status` in that case to `500`. Otherwise, you might have an exception status between 200 and 299 and the subsequent code would then consider it a successful request.

westonruter: Will it? Won't it only get thrown if the status code is not in the 2xx range?

https://github.com/ampproject/amp-wp/blob/182634c9363b897fe154c0f9db1b0f2b18fe88f5/src/RemoteRequest/WpHttpRemoteGetRequest.php#L130-L136

schlessera: Ah, yes, indeed. Alright then, looks good to me otherwise.

PR #3817, opened 2019-11-26
Refactor generated spec data to facilitate looking up by spec name

westonruter:
## Summary

Fixes #4597.

### Todo

- [ ] There should be an efficient way to look up a spec by a spec name regardless of the tag it occurs in.
- [ ] Instead of looping over the tag specs to find the one with the matching spec name, it would be better if the spec name was the array key.

## Checklist

- [x] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [ ] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [ ] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

PR #4405, opened 2020-03-19
Add missing error titles for error codes

### LINKED ISSUE #1420: Expose reason for why a validation error is raised

pierlon:
## Summary

This PR adds the missing error titles (and appropriate labels for error sources) for the error codes which have none.

Fixes #1420

## Todo

- [x] Add error titles for custom error codes

## Checklist


westonruter:
@pierlon Thanks for opening this draft PR. However, I am not sure this is the right approach for the immediate term (v1.5). Namely, the error codes that we are using are not 1:1 mappings with the error codes used by the AMP validator. There are many error codes that we don't use at all (e.g. `DOCUMENT_TOO_COMPLEX`) whereas others that the validator doesn't use at all (e.g. `MISSING_REQUIRED_PROPERTY_VALUE`). So I don't see a lot of value in automating the extraction of the strings from the AMP validator spec right now. What's more is that if we make all of these strings translatable, then translators may end up translating a bunch of strings that are not ever used. 

I think the initial target to solve #1420 should be less ambitious. And that is to extend what we have today in `\AMP_Validation_Error_Taxonomy::get_error_title_from_code()` to add additional `case` statements for the error codes we aren't currently accounting for, and incorporating the error messages from the validator where it makes sense.

So I think the patch should largely be limited to making changes to this method:

https://github.com/ampproject/amp-wp/blob/15fb873f72500c1c593ec769b364bc3937b65740/includes/validation/class-amp-validation-error-taxonomy.php#L2978-L3136

westonruter:
I say this in part because I already tried going down that road in 8ae5c3b5959af631e8404111315b128f1ade610a and 0b30e4a593793843a64c10f261812f2e58331e74, but I decided ultimately to abandon it in a6d0fc9315f917ebbc756b2d98c84d20b7fe46e2. This was done in the context of #3780.

pierlon:
Ah OK, I'm changing the direction of this PR from automating the extraction of the error codes to identifying the missing error titles for error codes and adding them.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -993,14 +993,16 @@ private function validate_tag_spec_for_node( DOMElement $node, $tag_spec ) {
 
 		if ( ! empty( $tag_spec[ AMP_Rule_Spec::MANDATORY_PARENT ] ) && ! $this->has_parent( $node, $tag_spec[ AMP_Rule_Spec::MANDATORY_PARENT ] ) ) {
 			return [
-				'code' => self::WRONG_PARENT_TAG,
+				'code'                 => self::WRONG_PARENT_TAG,
+				'required_parent_name' => $tag_spec[ AMP_Rule_Spec::MANDATORY_PARENT ],
```
westonruter: Note that at some point we will not need to capture the `required_parent_name` in the error. We could instead capture the `spec_name` alone. Then the `spec_name` could be looked up to obtain the `required_parent_name` (or other properties depending on the error). This will be made possible by https://github.com/ampproject/amp-wp/pull/3817. However, that is not yet available, so what you are doing here is the perfect approach in the time being.

pierlon: Yep I initially was retrieving it by looking up the tag spec in `AMP_Allowed_Tags_Generated::class`, but I realized it was available in the sanitizer so I took it from there.

> This will be made possible by #3817.

Awesome, made a note of that.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -3129,6 +3129,101 @@ static function ( $attribute_name ) {
...
+				return sprintf(
+					__( "The attribute '%1\$s' in tag '%2\$s' is set to the invalid value '%3\$s'", 'amp' ),
+					esc_html( $validation_error['node_name'] ),
+					esc_html( $validation_error['parent_name'] ),
+					esc_html( $validation_error['element_attributes'][ $validation_error['node_name'] ] )
+				);
+
+			case AMP_Tag_And_Attribute_Sanitizer::INVALID_URL_PROTOCOL:
+				$parsed_url = wp_parse_url( $validation_error['element_attributes'][ $validation_error['node_name'] ] );
+				$invalid_protocol = isset( $parsed_url['scheme'] ) ? $parsed_url['scheme'] : '';
```
pierlon: Not sure what to do here if the scheme could not be parsed, defaulting to an empty string for now.

westonruter: Maybe `null` would be more appropriate?

westonruter: Oh I see, it's used in a string. Yeah, an empty string seems fine for now.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -3129,6 +3129,101 @@ static function ( $attribute_name ) {
 				}
 				return $title;
 
+			case AMP_Tag_And_Attribute_Sanitizer::DISALLOWED_CHILD_TAG:
+				/* translators: %1$s is the child tag, %2$s is node name */
+				return sprintf(
+					__( "Tag '%1\$s' is disallowed as child of tag '%2\$s'", 'amp' ),
```
westonruter: Two alternative suggestions:

1. Use curly quotes instead of straight ones, as this is WordPress core standard.
2. Alternatively, remove the quotes altogether and instead wrap the `esc_html()` calls below with `code` tags. 

Using `code` throughout these error messages I think will lead to better results.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -27,7 +27,6 @@
 class AMP_Tag_And_Attribute_Sanitizer extends AMP_Base_Sanitizer {
 
 	const DISALLOWED_TAG                       = 'DISALLOWED_TAG';
-	const DISALLOWED_TAG_MULTIPLE_CHOICES      = 'DISALLOWED_TAG_MULTIPLE_CHOICES';
```
pierlon: I've went ahead and removed the `DISALLOWED_TAG_MULTIPLE_CHOICES` error code as I don't see a meaningful way of displaying the multiple validation errors on the page. Instead, each error will be sanitized separately which allows for each to be viewed.

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -505,7 +504,15 @@ private function process_node( DOMElement $node ) {
 					array_unique(
 						array_map(
 							static function ( $validation_error ) {
-								unset( $validation_error['spec_name'] );
+								unset(
+									$validation_error['spec_name'],
+									// Remove other keys that may not not make the error unique.
```
pierlon: I'm removing any other keys that could potentially make the validation error unique (but it should not). Take for example this HTML element:

```html
<stop offset="5%" stop-color="gold" />
```

This would raise two separate `MANDATORY_TAG_ANCESTOR` validation errors, but each would have its own `required_ancestor_name` (`lineargradient` and `radialgradient`), making each validation error unique.

includes/validation/class-amp-validation-error-taxonomy.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/4405/commits/29eea3c31ad668d83671d2159c931caf86a0a0b3.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -3129,6 +3130,136 @@ static function ( $attribute_name ) {
 				}
 				return $title;
 
+			case AMP_Tag_And_Attribute_Sanitizer::DISALLOWED_CHILD_TAG:
+				return sprintf(
+					/* translators: %1$s is the child tag, %2$s is node name */
+					__( 'Tag <code>%1$s</code> is disallowed as child of tag <code>%2$s</code>', 'amp' ),
+					esc_html( $validation_error['child_tag'] ),
+					esc_html( $validation_error['node_name'] )
```
westonruter: The markup doesn't need to be in the translation string, so:

```suggestion
					esc_html__( 'Tag %1$s is disallowed as child of tag %2$s', 'amp' ),
					'<code>' . esc_html( $validation_error['child_tag'] ) . '</code>',
					'<code>' . esc_html( $validation_error['node_name'] ) . '</code>'
```

Note also that this allows `esc_html__()` to be used.

westonruter: Same goes for instances below.

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -3129,6 +3130,136 @@ static function ( $attribute_name ) {
...
+
+			case AMP_Tag_And_Attribute_Sanitizer::INVALID_URL_PROTOCOL:
+				$parsed_url       = wp_parse_url( $validation_error['element_attributes'][ $validation_error['node_name'] ] );
+				$invalid_protocol = isset( $parsed_url['scheme'] ) ? $parsed_url['scheme'] : '';
+
+				return sprintf(
+					/* translators: %1$s is the invalid protocol, %2$s is attribute name */
+					__( 'Invalid URL protocol <code>%1$s:</code> for attribute <code>%2$s</code>', 'amp' ),
+					$invalid_protocol,
+					esc_html( $validation_error['node_name'] )
```
westonruter: In addition to moving the `code` markup, the `:` should be moved below:

```suggestion
					__( 'Invalid URL protocol %1$s for attribute %2$s</code>', 'amp' ),
					'<code>' . esc_html( $invalid_protocol ) . ':</code>',
					'<code>' . esc_html( $validation_error['node_name'] ) . '</code>'
```

Or rather the `:` should be made part of the `$invalid_protocol` above.

includes/validation/class-amp-validation-error-taxonomy.php
westonruter: Actually, since `code` is now being used as opposed to quotes, then I think something else is needed:

```suggestion
				$invalid_protocol = isset( $parsed_url['scheme'] ) ? $parsed_url['scheme'] . ':' : '(failure)';
```

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -3129,6 +3130,136 @@ static function ( $attribute_name ) {
...
+					__( 'The relative URL <code>%1$s</code> for attribute <code>%2$s</code> is disallowed', 'amp' ),
+					esc_html( $validation_error['element_attributes'][ $validation_error['node_name'] ] ),
+					esc_html( $validation_error['node_name'] )
+				);
+
+			case AMP_Tag_And_Attribute_Sanitizer::MISSING_URL:
+				return sprintf(
+					/* translators: %1$s is attribute name */
+					__( "Missing URL for attribute '%s'", 'amp' ),
+					esc_html( $validation_error['node_name'] )
```
westonruter: Missing `code`:

```suggestion
					esc_html__( "Missing URL for attribute %s", 'amp' ),
					'<code>' . esc_html( $validation_error['node_name'] ) . '</code>'
```

includes/validation/class-amp-validation-error-taxonomy.php
```diff
@@ -3129,6 +3130,136 @@ static function ( $attribute_name ) {
...
+			case AMP_Tag_And_Attribute_Sanitizer::INVALID_LAYOUT_HEIGHTS:
+				return sprintf(
+					/* translators: %1$s is the invalid attribute value, %2$s is the attribute name */
+					__( 'Invalid value <code>%1$s</code> for attribute <code>%2$s</code>', 'amp' ),
+					esc_html( $validation_error['node_attributes'][ $validation_error['attribute'] ] ),
+					esc_html( $validation_error['attribute'] )
+				);
+
+			case AMP_Tag_And_Attribute_Sanitizer::INVALID_LAYOUT_UNIT_DIMENSIONS:
+				return __( 'Inconsistent units for width and height', 'amp' );
```
westonruter:
```suggestion
				return esc_html__( 'Inconsistent units for width and height', 'amp' );
```

includes/sanitizers/class-amp-tag-and-attribute-sanitizer.php
```diff
@@ -527,13 +534,11 @@ static function ( $validation_error ) {
...
-						[
-							'code'   => self::DISALLOWED_TAG_MULTIPLE_CHOICES,
-							'errors' => $validation_errors,
-						]
-					);
+					foreach ( $validation_errors as $validation_error ) {
+						if ( true === $this->remove_invalid_child( $node, $validation_error ) ) {
+							break; // Once removed, ignore remaining errors.
+						}
+					}
```
westonruter: Given a Custom HTML block with the following markup:

```html
<amp-accordion id="my-accordion" disable-session-states="">
  <section>
    <h2>Section 1</h2>
    <p>Content in section 12.</p>
  </section>
  <section>
    <h2>Section 2</h2>
    <div>Content in section 2.</div>
  </section>
  <section expanded="">
    <!--<h2>Section 3</h2>-->
    <amp-img src="/static/inline-examples/images/squirrel.jpg" width="320" height="256"></amp-img>
  </section>
</amp-accordion>
```

This is invalid because of the `h2` being commented out. The `amp-accordion` component requires an `h2` followed by a `div` or a `p`. The validation errors for this are not expected, however:

image

Notice the `amp-img` lacks the expected Custom HTML source.

I've fixed this here in feababc by breaking the loop of validation error reporting once a markup causing validation error is removed.

This will probably make this code no longer necessary:

https://github.com/ampproject/amp-wp/blob/3c29860c139d0706f9bb1029df0752803e625681/includes/sanitizers/class-amp-base-sanitizer.php#L477-L481

Though it doesn't hurt to have it in there.

pierlon: > Notice the `amp-img` lacks the expected Custom HTML source.

Good catch!

> This will probably make this code no longer necessary:

In this context yes, but it could potentially prevent an error being thrown when attempting to remove a node multiple times.

PR #4440, opened 2020-03-24
Validate <html> element if 'data-ampdevmode' attribute was not initially set

### LINKED ISSUE #4387: Validation errors not reported for invalid attributes on the HTML element
## Bug Description

Split out from #4190.

In #2500 the admin bar was suppressed from being shown when doing validation requests. Then in #3187 this was reverted because the addition of dev mode meant that the presence of the admin bar would no longer make pages invalid. Nevertheless, this had an unexpected consequence. When the `<html>` element has invalid attributes, these are never detected during validation because a validation request is made as an authenticated user, and thus the admin bar is present, and the admin bar entails dev mode being enabled. Since the `data-ampdevmode` attribute is on the `html` element, any invalid attributes on the element are ignored for validation purposes. This is not expected.

## Steps to reproduce

Activate Twenty Twenty and add this plugin code:

```php
add_filter(
	'language_attributes',
	function ( $attr ) {
		if ( ! is_admin() ) {
			$attr .= ' amp-version="1902151859190" onclick="alert(\'hello\')" ';
		}
		return $attr;
	}
);
```

This adds two invalid AMP attributes, and when I look at an AMP page while logged-in, I see them:

```html
<html class="no-js" lang="en-US" amp-version="1902151859190" onclick="alert('hello')" amp="" data-ampdevmode="">
```

Nevertheless, the admin bar shows ✅ for validity.

However, if I access the same AMP page while being logged-out, I see:

```html
<html class="no-js" lang="en-US" amp="">
```

Similarly, if I go to validate the URL I see no validation errors:

image

But the validation errors _are_ there. It's just they are not being reported.

---------------

## Acceptance criteria

* 

## Implementation brief

* We should capture whether the `html` element _originally_ had `data-ampdevmode` and if so, ignore validation errors. Otherwise if it was _added_ by the plugin to enable dev mode, then go ahead and report validation errors on the element even though it has the attribute. This would be a special case unique for the `html` element.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

If the `data-ampdevmode` attribute was set on the `<html>` element by the plugin, the element will be validated for errors. If it was set externally, validation errors will be ignored.

Fixes #4387.

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -440,14 +440,13 @@ abstract class AMP_Base_Sanitizer {
 	/**
 	 * Check whether a certain node should be exempt from validation.
 	 *
-	 * @deprecated Use AmpProject\DevMode::isExemptFromValidation( $node ) instead.
+	 * @since 1.5 Exempts DOMNode if the 'data-ampdevmode' attribute has the value '_amp_exempt'.
 	 *
 	 * @param DOMNode $node Node to check.
 	 * @return bool Whether the node should be exempt from validation.
 	 */
 	protected function is_exempt_from_validation( DOMNode $node ) {
-		_deprecated_function( 'AMP_Base_Sanitizer::is_exempt_from_validation', '1.5', 'AmpProject\DevMode::isExemptFromValidation' );
-		return DevMode::isExemptFromValidation( $node );
+		return DevMode::isExemptFromValidation( $node ) && '_amp_exempt' !== $node->getAttribute( AMP_Rule_Spec::DEV_MODE_ATTRIBUTE );
 	}
 
 	/**
@@ -463,7 +462,7 @@ abstract class AMP_Base_Sanitizer {
 	 * @return bool Whether the node should have been removed, that is, that the node was sanitized for validity.
 	 */
 	public function remove_invalid_child( $node, $validation_error = [] ) {
-		if ( DevMode::isExemptFromValidation( $node ) ) {
+		if ( $this->is_exempt_from_validation( $node ) ) {
 			return false;
 		}
 
@@ -501,7 +500,7 @@ abstract class AMP_Base_Sanitizer {
 	 * @return bool Whether the node should have been removed, that is, that the node was sanitized for validity.
 	 */
 	public function remove_invalid_attribute( $element, $attribute, $validation_error = [], $attr_spec = [] ) {
-		if ( DevMode::isExemptFromValidation( $element ) ) {
+		if ( $this->is_exempt_from_validation( $element ) ) {
 			return false;
 		}
 
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -446,7 +446,9 @@ abstract class AMP_Base_Sanitizer {
 	 * @return bool Whether the node should be exempt from validation.
 	 */
 	protected function is_exempt_from_validation( DOMNode $node ) {
-		return DevMode::isExemptFromValidation( $node ) && '_amp_exempt' !== $node->getAttribute( AMP_Rule_Spec::DEV_MODE_ATTRIBUTE );
+		return DevMode::isExemptFromValidation( $node )
+			&& $node instanceof DOMElement
+			&& '_amp_exempt' !== $node->getAttribute( AMP_Rule_Spec::DEV_MODE_ATTRIBUTE );
 	}
 
 	/**
--- a/includes/sanitizers/class-amp-base-sanitizer.php
+++ b/includes/sanitizers/class-amp-base-sanitizer.php
@@ -440,15 +440,14 @@ abstract class AMP_Base_Sanitizer {
 	/**
 	 * Check whether a certain node should be exempt from validation.
 	 *
-	 * @since 1.5 Exempts DOMNode if the 'data-ampdevmode' attribute has the value '_amp_exempt'.
+	 * @deprecated Use AmpProject\DevMode::isExemptFromValidation( $node ) instead.
 	 *
 	 * @param DOMNode $node Node to check.
 	 * @return bool Whether the node should be exempt from validation.
 	 */
 	protected function is_exempt_from_validation( DOMNode $node ) {
-		return DevMode::isExemptFromValidation( $node )
-			&& $node instanceof DOMElement
-			&& '_amp_exempt' !== $node->getAttribute( AMP_Rule_Spec::DEV_MODE_ATTRIBUTE );
+		_deprecated_function( 'AMP_Base_Sanitizer::is_exempt_from_validation', '1.5', 'AmpProject\DevMode::isExemptFromValidation' );
+		return DevMode::isExemptFromValidation( $node );
 	}
 
 	/**
@@ -464,7 +463,7 @@ abstract class AMP_Base_Sanitizer {
 	 * @return bool Whether the node should have been removed, that is, that the node was sanitized for validity.
 	 */
 	public function remove_invalid_child( $node, $validation_error = [] ) {
-		if ( $this->is_exempt_from_validation( $node ) ) {
+		if ( DevMode::isExemptFromValidation( $node ) ) {
 			return false;
 		}
 
@@ -502,7 +501,7 @@ abstract class AMP_Base_Sanitizer {
 	 * @return bool Whether the node should have been removed, that is, that the node was sanitized for validity.
 	 */
 	public function remove_invalid_attribute( $element, $attribute, $validation_error = [], $attr_spec = [] ) {
-		if ( $this->is_exempt_from_validation( $element ) ) {
+		if ( DevMode::isExemptFromValidation( $element ) ) {
 			return false;
 		}
 
--- a/includes/sanitizers/class-amp-dev-mode-sanitizer.php
+++ b/includes/sanitizers/class-amp-dev-mode-sanitizer.php
@@ -8,6 +8,8 @@
  * @package AMP
  */
 
+use AmpProject\DevMode;
+
 /**
  * Class AMP_Dev_Mode_Sanitizer
  *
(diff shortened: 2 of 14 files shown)


includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -440,14 +440,13 @@ protected function has_dev_mode_exemption( DOMNode $node ) {
 	/**
 	 * Check whether a certain node should be exempt from validation.
 	 *
-	 * @deprecated Use AmpProject\DevMode::isExemptFromValidation( $node ) instead.
+	 * @since 1.5 Exempts DOMNode if the 'data-ampdevmode' attribute has the value '_amp_exempt'.
```
pierlon: Can I un-deprecate this method, or should it stay and create a new function instead?

includes/sanitizers/class-amp-dev-mode-sanitizer.php
```diff
@@ -30,13 +32,17 @@ final class AMP_Dev_Mode_Sanitizer extends AMP_Base_Sanitizer {
 	 * @since 1.3
 	 */
 	public function sanitize() {
-		$this->dom->documentElement->setAttribute( AMP_Rule_Spec::DEV_MODE_ATTRIBUTE, '' );
+		$document_element = $this->dom->documentElement;
+
+		if ( ! DevMode::hasExemptionForNode( $document_element ) ) {
+			$document_element->setAttribute( AMP_Rule_Spec::DEV_MODE_ATTRIBUTE, '_amp_exempt' );
```
pierlon: Any objections to setting `data-ampdevmode` to `_amp_exempt` to mark it as set by the plugin?

westonruter: Humm. Would it be better to rather introduce a new property directly on the `Document` object to indicate whether the `documentElement` has always had the `data-ampdevmode` attribute.

In `\AmpProject\Dom\Document::loadHTMLFragment()` perhaps, it could do something like:

```php
$this->intialAmpDevMode = $this->hasAttribute( DevMode::DEV_MODE_ATTRIBUTE );
```

Then instead of checking for some special value, this flag could be checked. This would prevent a sanitizer from setting the value of the attribute to be this special`_amp_exempt` value.

So then the `\AmpProject\DevMode::hasExemptionForNode` method could be changed to be something like this:

```diff
diff --git a/lib/common/src/DevMode.php b/lib/common/src/DevMode.php
index 11ceeeb28..207e89a15 100644
--- a/lib/common/src/DevMode.php
+++ b/lib/common/src/DevMode.php
@@ -51,6 +51,12 @@ final class DevMode
             return false;
         }
 
+        // If the node is the documentElement and it did not originally have dev-mode, then count it as not being exempt.
+        $document = $node->ownerDocument;
+        if ( $document instanceof Document && $node === $document->documentElement && ! $document->intialAmpDevMode ) {
+            return false;
+        }
+
         return $node->hasAttribute(self::DEV_MODE_ATTRIBUTE);
     }
```

And then `DevMode::isExemptFromValidation()` could be continued to be used as before, rather than resurrecting the `is_exempt_from_validation` method.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -501,7 +502,7 @@ public function remove_invalid_child( $node, $validation_error = [] ) {
 	 * @return bool Whether the node should have been removed, that is, that the node was sanitized for validity.
 	 */
 	public function remove_invalid_attribute( $element, $attribute, $validation_error = [], $attr_spec = [] ) {
-		if ( DevMode::isExemptFromValidation( $element ) ) {
+		if ( $this->is_exempt_from_validation( $element ) ) {
```
westonruter: I think this can be reverted based on my comment on the `sanitize` method below.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -463,7 +464,7 @@ protected function is_exempt_from_validation( DOMNode $node ) {
 	 * @return bool Whether the node should have been removed, that is, that the node was sanitized for validity.
 	 */
 	public function remove_invalid_child( $node, $validation_error = [] ) {
-		if ( DevMode::isExemptFromValidation( $node ) ) {
+		if ( $this->is_exempt_from_validation( $node ) ) {
```
westonruter: I think this can be reverted based on my comment on the `sanitize` method below.

includes/sanitizers/class-amp-base-sanitizer.php
```diff
@@ -440,14 +440,15 @@ protected function has_dev_mode_exemption( DOMNode $node ) {
...
 	 *
 	 * @param DOMNode $node Node to check.
 	 * @return bool Whether the node should be exempt from validation.
 	 */
 	protected function is_exempt_from_validation( DOMNode $node ) {
-		_deprecated_function( 'AMP_Base_Sanitizer::is_exempt_from_validation', '1.5', 'AmpProject\DevMode::isExemptFromValidation' );
-		return DevMode::isExemptFromValidation( $node );
+		return DevMode::isExemptFromValidation( $node )
+			&& $node instanceof DOMElement
+			&& '_amp_exempt' !== $node->getAttribute( AMP_Rule_Spec::DEV_MODE_ATTRIBUTE );
```
westonruter: I think this can be reverted based on my comment on the `sanitize` method below.

includes/sanitizers/class-amp-base-sanitizer.php
westonruter: I think this can be reverted based on my comment on the `sanitize` method below.

includes/sanitizers/class-amp-dev-mode-sanitizer.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/4440/commits/f9296cab3c7f30426639fe14a836c2921607cba4.

tests/php/test-amp-audio-converter.php
```diff
@@ -235,6 +235,7 @@ public function test_converter( $source, $expected = null, $args = [] ) {
 		}
 		$dom = AMP_DOM_Utils::get_dom_from_content( $source );
 		if ( ! empty( $args['add_dev_mode'] ) ) {
+			$dom->ampDevModeActive = true; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
westonruter: I don't think this is right. This property should only be relevant when `loadHTML` first parses the document. The `ampDevModeActive`  property on the document should be read-only. It's kinda indicating the flavor of the HTML document.

pierlon: Applied suggestion in https://github.com/ampproject/amp-wp/pull/4440/commits/fa005500c92f503c9dc0b81ebccd598f8f2d1ba5.

lib/common/src/DevMode.php
```diff
@@ -62,9 +62,15 @@ public static function hasExemptionForNode(DOMNode $node)
...
+        $document = $node->ownerDocument;
+        if (! $node->ownerDocument instanceof Document) {
+            $document = Document::fromNode($node);
+        }
 
-        if (! $document) {
+        /**
+         * @var Document $document
+         */
+        if ($node === $document->documentElement && ! $document->hasInitialAmpDevMode()) {
```
schlessera: There are a few places in the code that forgo `DevMode::isExemptFromValidation()` and use `DevMode:: hasExemptionForNode()` instead (to duplicate calls to `DevMode::isActiveForDocument()`).

You should check these and make sure they behave correctly still, or adapt them as needed.

pierlon: The only change here is to determine whether the `html` element should be validated or not; the behavior of the class is the same otherwise.

I haven't seen any adverse effects of determining the dev mode for nodes with this change.

lib/common/src/Dom/Document.php
```diff
@@ -1446,6 +1458,16 @@ public function isValidHeadNode(DOMNode $node)
         );
     }
 
+    /**
+     * Determine whether `data-ampdevmode` was initially set on the document element.
+     *
+     * @return bool
+     */
+    public function hasInitialAmpDevMode()
```
westonruter: This method should be tested in `DocumentTest`, verifying the method returns:

1. `false` when the `data-ampdevmode` attribute is absent on `html`.
1. `false` when the `data-ampdevmode` attribute is added on `html` after parsing.
1. `true`  when the `data-ampdevmode` attribute is initially on `html`.

lib/common/tests/DevModeTest.php
```diff
@@ -34,7 +34,7 @@ public function dataHasExemptionForNode()
     {
         return [
             [ '<html><body id="node_to_test"><div data-ampdevmode></div></body></html>', false ],
-            [ '<html><body><div id="node_to_test" data-ampdevmode></div></body></html>', true ],
+            [ '<html data-ampdevmode><body><div id="node_to_test" data-ampdevmode></div></body></html>', true ],
```
westonruter: Two more test cases are needed, namely:

1. On `<html id="node_to_test">` when `data-ampdevmode` is initially present, `DevMode::hasExemptionForNode($node)` should be `true`.
1. On `<html id="node_to_test">` when `data-ampdevmode` is not initially present (added after document created), `DevMode::hasExemptionForNode($node)` should be `false`.

lib/common/src/Dom/Document.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/4440/commits/3081a0a5de6e45c3d01ea0c9059f9914dc37128b.

lib/common/tests/DevModeTest.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/4440/commits/3081a0a5de6e45c3d01ea0c9059f9914dc37128b.

lib/common/src/DevMode.php
pierlon: I rescind my comment above; please see changes in https://github.com/ampproject/amp-wp/pull/4440/commits/3081a0a5de6e45c3d01ea0c9059f9914dc37128b.

lib/common/src/DevMode.php
```diff
@@ -62,12 +71,23 @@ public static function hasExemptionForNode(DOMNode $node)
...
+        return self::isActiveForDocument($document) && self::hasExemptionForNode($node);
+    }
 
-        if (! $document) {
-            return false;
+    /**
+     * Get the document from the specified node.
+     *
+     * @param DOMNode $node
+     * @return Document|DOMDocument|null
```
westonruter: Won't this only ever return a `Document` instance? The `Document::fromNode()` method only returns a `Document`.

```suggestion
     * @return Document
```

lib/common/src/DevMode.php
```diff
@@ -51,6 +51,15 @@ public static function hasExemptionForNode(DOMNode $node)
             return false;
         }
 
+        /**
+         * @var Document $document
+         */
```
westonruter: I don't think this is required because it will only be a `Document` per below:

```suggestion
```

westonruter:
Note: I've filed an upstream issue (https://github.com/ampproject/amphtml/issues/27467) as well for what appears to be a validator issue where validation errors are being raised for invalid attributes on the root element even when it has `data-ampdevmode`.

lib/common/src/DevMode.php
pierlon: This is just to make `phpstan` happy. It incorrectly detects `$document` here as a `DOMDocument`,which would not have the `hasInitialAmpDevMode` method.

westonruter: But it shouldn't be needed now that you did 6314af6.

westonruter: I found what was missing. Fixed in 472c6fc1c.

PR #4482, opened 2020-03-31
Adapt heights, sizes and media attributes for SSR

### LINKED ISSUE #4439: Implement transformer in optimizer for media, sizes, and heights
## Feature description

Currently SSR is blocked when an AMP page contains AMP components that make use of the attributes `sizes`, `heights`, or `media`. While the AMP Optimizer does not currently implement a transformer for this, there is a transformation which can be made as noted in Future Work of [Server-side rendering for AMP caches](https://docs.google.com/document/d/1gViU1hxtGXwMSTNnum2zY_p9ZWvFwMKoachNdIRUgh0/edit#heading=h.opq6yorpt6ly):

> Handle rewriting of media, sizes and heights attributes into CSS.
>
> Rewriting `sizes` attribute
> 
>
> ```html
> <amp-img id="img1" sizes="(min-width: 320px) 320px, 100vw"></amp-img>
> ```
>
> Server-side inserted CSS
>
> ```html
> <style i-amp-layout-style>
> #img1 {
>   width: 100vw;
> }
> @media (min-width: 320px) {
>   #img1 {
>     width: 320px;
>   }
> }
> </style>
> ```
> 
> Rewriting `heights` attribute
> 
> ```html
> <amp-img id="img1" heights="(min-width: 320px) 50%, 200px"></amp-img>
> ```
> 
> Server-side inserted CSS
> 
> ```html
> <style i-amp-layout-style>
> #img1 > i-amphtml-sizer {
>   height: 200px;
> }
> @media (min-width: 320px) {
>   #img1 > i-amphtml-sizer {
>     height: 50%;
>   }
> }
> </style>
> ```

Note that `media` can be transformed just by moving the attribute into a CSS rule. This involved ensuring that the element has an ID to target in a selector.

Related #4213

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
## Summary

This PR adds logic to the `ServerSideRendering` transformer in `ampproject/optimizer` to adapt the `heights`, `sizes` and `media` attributes on AMP elements and turn them into CSS. This is done so that they don't block server-side layout application.

The PR also moves the `AMP_DOM_Utils::get_element_id()` to `AmpProject\Dom\Document::getElementId()` and deprecates the previous method.

Addresses issue #4439

## Checklist


westonruter:
Here's my plan for getting this merged:

- [x] 1. Update the AMP spec to the latest via #4548.
- [x] 2. Create a new PR which eliminates the removal of the `sizes` attribute in favor of adding the new `disable-inline-width` attribute. See #4606.
- [x] 3. Merge the `develop` branch into this branch to test those changes with the SSR changes. At the moment the `sizes` attribute is stripped from all images generated by WordPress, but this is no longer done then the changes in _this_ PR to adapt `sizes` for SSR will be hugely more important. Also, I suspect that `amp-img[sizes][disable-inline-width]` will actually be exempt from requiring the boilerplate, so that will be a new condition to check for in the Optimizer. cc @sebastianbenz

sebastianbenz:
Good find! Really interesting! I'm wondering whether it'd make sense to make this the default behavior for sizes (for transformed AMP) or whether there are cases where the behavior differs.

westonruter:
@schlessera This is ready for picking up again. Now that #4622 is merged, images converted to `amp-img` will default to having the `sizes` attribute retained, so this will make testing this PR here much easier and dramatic in terms of impact.

schlessera:
This needs to be adapted for a special condition I wasn't aware of until now. From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-sizes:

> If the srcset attribute is absent, or contains no values with a width descriptor, then the sizes attribute has no effect.

As we're unconditionally transforming the `sizes` attribute into straight media queries right now, we're turning inactive `sizes` attributes into active media queries that shouldn't exist.

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -46,6 +56,72 @@ final class ServerSideRendering implements Transformer
...
+     * The <style amp-custom> element that custom CSS styles need to be added to.
+     *
+     * @var DOMElement
+     */
+    private $ampCustomStyleElement;
+
+    /**
+     * Count of bytes to calculate against the AMP size limit for the custom CSS styling.
+     *
+     * AMP only allows for 75.000 bytes of CSS across <style amp-custom> and inline style attributes.
```
westonruter:
```suggestion
     * AMP only allows for 75000 bytes of CSS across <style amp-custom> and inline style attributes.
```

schlessera: As that is a regional thing and both versions read like 75 bytes with three zeros of decimal precision in one half of the world, shouldn't we go with a more neutral `75000` instead?

schlessera:
> This still needs to incorporate the logic to skip extracting sizes when the element has disable-inline-width?

Yes, I will adapt the code today.

lib/optimizer/src/Transformer/ServerSideRendering.php
westonruter: Sounds good 👍

schlessera:
Two broken tests in Optimizer are being skipped, they should be dealt with in a separate PR: https://github.com/ampproject/amp-wp/issues/4654

schlessera:
@westonruter From my testing, I think this works as expected in terms of markup. However, the styling is currently still broken because of https://github.com/ampproject/amphtml/pull/28020

As the above is already merged and expected to go into prod soon, are we good to merge this as-is? Or should I add the above as a blocker and put this one on hold?

westonruter:
> As the above is already merged and expected to go into prod soon, are we good to merge this as-is? Or should I add the above as a blocker and put this one on hold?

Yes, it will be in prod soon. By next week at the latest, I believe.

lib/common/src/Dom/CssByteCountCalculator.php
```diff
@@ -0,0 +1,79 @@
...
+     *
+     * @param DOMNodeList $nodeList Node list to calculate the byte count of CSS styles for.
+     * @return int Byte count of CSS styles for the given node list.
+     */
+    private function calculateForNodeList(DOMNodeList $nodeList)
+    {
+        $byteCount = 0;
+
+        foreach ($nodeList as $node) {
+            $byteCount += strlen($node->textContent);
```
westonruter, line 74: TIL that `textContent` works for `DOMAttr` as well. It's not [called out](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) in MDN even.

lib/common/src/Dom/Document.php
```diff
@@ -314,7 +323,7 @@ public static function fromNode(DOMNode &$node)
         // We replace the $node by reference, to make sure the next lines of code will
         // work as expected with the new document.
         // Otherwise $dom and $node would refer to two different DOMDocuments.
-        $node = $dom->importNode($root->documentElement ?: $root, true);
+        $node = $dom->importNode($node, true);
```
westonruter, line 326: Bugfix?

lib/common/src/Dom/Document.php
```diff
@@ -1474,6 +1483,39 @@ public function isValidHeadNode(DOMNode $node)
...
+            $id = "{$prefix}-{$this->indexCounter[ $prefix ]}";
+        } else {
+            $id                          = $prefix;
+            $this->indexCounter[$prefix] = 1;
+        }
+
+        while ($this->getElementById($id) instanceof DOMElement) {
+            ++$this->indexCounter[$prefix];
+            $id = "{$prefix}-{$this->indexCounter[ $prefix ]}";
+        }
```
westonruter, line 1512: This is a fallback scenario in which case it just so happens a pre-existing ID exists on the document which collides?

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -64,41 +140,40 @@ public function transform(Document $document, ErrorCollection $errors)
...
             // Skip tags inside a template tag.
-            if ($this->hasAncestorWithTag($amp_element, Tag::TEMPLATE)) {
+            if ($this->hasAncestorWithTag($ampElement, Tag::TEMPLATE)) {
                 continue;
             }
 
             /*
-             * If these attributes are used on any AMP custom element tags within the document, we can't remove the
-             * boilerplate - they require the boilerplate.
+             * amp-experiment is a render delaying extension iff the tag is used in the doc. We check for that here
```
westonruter, line 150:
```suggestion
             * amp-experiment is a render delaying extension if the tag is used in the doc. We check for that here
```

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,260 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+            $normalizedAttributeName = strtolower($attribute->name);
+            if (array_key_exists($normalizedAttributeName, self::EXTRACTION_METHOD_MAP)) {
+                if (
+                    $normalizedAttributeName === Attribute::SIZES
+                    && $ampElement->hasAttribute(Attribute::DISABLE_INLINE_WIDTH)
+                ) {
+                    // Don't remove sizes when disable-inline-width is set.
+                    // @see https://github.com/ampproject/amphtml/pull/27083
+                    continue;
+                }
```
westonruter: @sebastianbenz FYI

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,260 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+
+                $method = self::EXTRACTION_METHOD_MAP[$normalizedAttributeName];
+
+                try {
+                    $customCss            .= $this->$method($document, $ampElement, $attribute->value);
+                    $attributesToRemove[] = $attribute->name;
```
westonruter: The equals signs aren't lined up. Is that intended?

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,260 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+                    $ampElement->removeAttribute($attribute->name);
```
westonruter: Should this be getting removed here? I believe it should only be getting removed below after the `style[amp-custom]` is successfully amended.

```suggestion
```

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,260 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+     */
+    private function addCustomCss(Document $document, $customCss)
+    {
+        $additionalBytes = strlen($customCss);
+
+        if (($this->ampCustomCssByteCount + $additionalBytes) > self::MAX_CSS_BYTE_COUNT) {
+            return false;
+        }
+
+        if (empty($this->ampCustomStyleElement)) {
```
westonruter, line 795: Should this check to see if the style element is still in the document (e.g. if it doesn't have a `parentNode`)? If not, then shouldn't `ampCustomCssByteCount` be zeroed out? I realize this is a condition that could never happen given the current transformers, but it's just something that came to mind.

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,260 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+     * @return string Extract custom CSS styling.
+     */
+    private function extractSizesAttributeCss(Document $document, DOMElement $element, $attributeValue)
+    {
+        if (!$element->hasAttribute(Attribute::SRCSET) || empty($element->getAttribute(Attribute::SRCSET))) {
+            // According to the Mozilla docs, a sizes attribute without a valid srcset attribute should have no effect.
+            // Therefore, it should simply be stripped, without producing media queries.
+            // @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-sizes
+            return '';
+        }
```
westonruter, line 848: @sebastianbenz FYI

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,260 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+
+        return $this->extractAttributeCss(
+            $document,
+            $element,
+            $attributeValue,
+            '#__ID__{width:%s};',
+            '@media %s{#__ID__{width:%s;}}'
```
westonruter: We can save a byte:

```suggestion
            '@media %s{#__ID__{width:%s}}'
```

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,260 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+            // Therefore, it should simply be stripped, without producing media queries.
+            // @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-sizes
+            return '';
+        }
+
+        return $this->extractAttributeCss(
+            $document,
+            $element,
+            $attributeValue,
+            '#__ID__{width:%s};',
```
westonruter: Semicolon should be removed here, no?

```suggestion
            '#__ID__{width:%s}',
```

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,260 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+     * @param string     $attributeValue Value of the attribute to extract the CSS styling from.
+     * @return string Extract custom CSS styling.
+     */
+    private function extractHeightsAttributeCss(Document $document, DOMElement $element, $attributeValue)
+    {
+        return $this->extractAttributeCss(
+            $document,
+            $element,
+            $attributeValue,
+            '#__ID__>i-amphtml-sizer{height:%s};',
```
westonruter:
```suggestion
            '#__ID__>i-amphtml-sizer{height:%s}',
```

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,260 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+            '@media %s{#__ID__>i-amphtml-sizer{height:%s;}}'
```
westonruter:
```suggestion
            '@media %s{#__ID__>i-amphtml-sizer{height:%s}}'
```

lib/optimizer/src/Transformer/ServerSideRendering.php
westonruter: This selector is going to be problematic as it breaks AMP validation (see [spec](https://github.com/ampproject/amphtml/blob/cb57fe236c0e904b2030efcf611332d4c83fcea3/validator/validator-css.protoascii#L542-L547)):

image

We may have to omit extraction of `heights` for the time being.

cc @sebastianbenz

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,260 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+        if (empty($attributeValue)) {
+            return '';
+        }
+
+        if ($attributeValue[0] === '(') {
+            $attributeValue = 'all and ' . $attributeValue;
+        }
+
+        return sprintf(
+            '@media not %s{#%s{display:none;}}',
```
westonruter:
```suggestion
            '@media not %s{#%s{display:none}}',
```

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: `iff` is commonly used for "if and only if" and is also what is being used in the Node.js code from where this is taken from.

lib/common/src/Dom/Document.php
schlessera: Yes, as we cannot make any assumptions about what will be passed into the Optimizer. A sanitizer might already have created an ID, for example.

lib/common/src/Dom/CssByteCountCalculator.php
schlessera: Image 2020-05-07 at 8 54 24 AM

lib/common/src/Dom/Document.php
schlessera: Yes. I also can't think of a reason why I imported `$root` instead of `$node` in the first place here. Doesn't seem to make sense.

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: Hmm, that seems to be a left-over, as that is also added to the `$attributesToRemove[]` in the line just above.

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: I'm not happy with the way the `ampCustomStyleElement` is currently being handled, as it creates a dependency between transformers.

I was thinking of adding an abstraction into `DOM\Document` so that you'll have:
```php
$document->addCustomCss( string $customCss, bool $ignoreByteCount = false );
```

This would work consistently across transformers and add or omit the style element as needed. It would also mean we could count bytes on addition, or even do stuff as automatic minification behind the scenes.

schlessera: Oh, I reread your comment and I think you misread the code here.

The `empty($this->ampCustomStyleElement)` is because we're storing the element as an internal property, however it is not initialized right away.

So the empty check here is just to provide the initialization when we hit `addCustomCss()` for the first time.

Then it tries to retrieve the style element from the actual document to see if one already existed before we ran this transformer, and count the bytes accordingly.

If not (meaning we didn't have a style element AND the document didn't contain one either), we start with a byte count of zero.

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4482/commits/facc01606bab638a58ff32a991becc6686ceb94a

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: Is it worth it to experiment with different selectors that assume a certain hierarchy without using this specific class name?

I was thinking about adding the ID directly to the sizer in this case, instead of the preceding element.
So we'd end up with this instead:
```html
'#__ID__{height:%s}'
```

I can't see why adding an ID to the sizer would cause a problem, so I'll just try it.

sebastianbenz: > I can't see why adding an ID to the sizer would cause a problem, so I'll just try it.

The validator does not allow `id` on sizer elements.

lib/optimizer/src/Transformer/ServerSideRendering.php
westonruter: Makes sense. I wasn't sure if that was the case, or if it was a typo.

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: Are classes allowed then? Or anything else I could target with a selector?

lib/common/src/Dom/CssByteCountCalculator.php
westonruter: OK, I can see that on the docs for [Node](https://developer.mozilla.org/en-US/docs/Web/API/Node).

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: I'm always inserting the sizer as the first child, so I could potentially rely on `#__ID__:first-child{height:%s}` instead.

However, that will not work as expected on already optimized code where this assumption is wrong. Not sure how probable that is, though. I think to remember that all implementations insert the sizer as the first child...

lib/optimizer/src/Transformer/ServerSideRendering.php
westonruter: Yes, my concern/question is if another transformer removed the element after it had been stored. But I suppose that couldn't happen while the transformer is running anyway.

lib/optimizer/src/Transformer/ServerSideRendering.php
westonruter: Here's the spec: https://github.com/ampproject/amphtml/blob/cb57fe236c0e904b2030efcf611332d4c83fcea3/validator/validator-main.protoascii#L4575-L4588

As I understand, the only allowed attribute is `class` and it must be `i-amphtml-sizer` which can't be used in a selector either.

Using `#ID>:first-child` would work as long as the transformer makes sure the `i-amphtml-sizer` is the first element, before any singling `[fallback]`, `[placeholder]`, or `noscript` elements.

schlessera: The `[fallback]`, `[placeholder]` and `noscript` elements would be provided outside of the Optimizer already before the markup even hits the Optimizer.

The Optimizer then does this, so it should always be true:
```php
if ($sizer) {
    $element->insertBefore($sizer, $element->firstChild);
}
```

schlessera: It seems to work, so I've adapted the code to use that for now.
Would it make sense to flag this in `amphtml`? Or are we already deep in "hacking" territory here?

I'm not sure what the long-term goal is for these attributes with regards to SSR.

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +699,271 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+         * pass the validator.
+         * See https://github.com/ampproject/amp-wp/pull/4482#discussion_r421149537
+         */
+        return $this->extractAttributeCss(
+            $document,
+            $element,
+            $attributeName,
+            $attributeValue,
+            '#__ID__:first-child{height:%s}',
+            '@media %s{#__ID__:first-child{height:%s}}'
```
westonruter: This won't work as expected, as the `:first-child` is a pseudo-selector qualifying what it is attached to. So this would match elements that have ID of `__ID__` which are the first-child position. So it needs rather to be:

```suggestion
            '#__ID__>:first-child{height:%s}',
            '@media %s{#__ID__>:first-child{height:%s}}'
```

So it selects the first child of the element that has the ID.

schlessera: Ah, you are right. Will correct.

schlessera: ... I just realized that it is not the `height` that needs to be adapted for the sizer, I think, but the `padding-top`.

lib/optimizer/src/Transformer/ServerSideRendering.php
westonruter: I believe I've opened the relevant issue for flagging this: https://github.com/ampproject/amphtml/issues/27468

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: ... which can't be set as it is inlined by JS already. 🤔

schlessera: So, as the Optimizer is the one that inlined the `padding-top`, I could skip that if I detect that I already had a `heights` attribute that turned into CSS.

But at what point are we just plain "hacking" here?

schlessera: Oh, I could actually store the CSS and properly inline it onto the sizer. That way, I can omit adding it to the size-limited `<style amp-custom>` as well.

schlessera: I'm running in circles here. Inline styles don't allow for media queries. I will just skip adding the `padding-top` in that case for now.

schlessera: Ok, I've got a working version now. Not the most elegant stuff, but it seems to work as expected.

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -814,10 +827,11 @@ private function addCustomCss(Document $document, $customCss)
      *
      * @param Document   $document       Document containing the element to adapt.
      * @param DOMElement $element        Element to adapt.
+     * @param string     $attributeName  Name of the attribute to be extracted.
      * @param string     $attributeValue Value of the attribute to extract the CSS styling from.
      * @return string Extract custom CSS styling.
      */
-    private function extractSizesAttributeCss(Document $document, DOMElement $element, $attributeValue)
+    private function extractSizesAttributeCss(Document $document, DOMElement $element, $attributeName, $attributeValue)
```
westonruter: Minor thing, but instead of passing the attribute name and value separately, this could instead just require a `DOMAttr`.

lib/optimizer/src/Exception/InvalidHtmlAttribute.php
```diff
@@ -0,0 +1,30 @@
...
+{
+
+    /**
+     * Instantiate an InvalidHtmlAttribute exception for an invalid attribute value.
+     *
+     * @param string     $attributeName Name of the attribute.
+     * @param DOMElement $element       Element that contains the invalid attribute.
+     * @return self
+     */
+    public static function fromAttribute($attributeName, DOMElement $element)
```
westonruter, line 24: This could take just one `DOMAttr` as an argument rather than the attribute name and the element the attribute is on.

lib/optimizer/src/Transformer/ServerSideRendering.php
westonruter: Done in a36dc0ff. However, I suppose the `$element` isn't needed either, as this can just get `$attribute->parentNode`, right?

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -46,6 +56,79 @@ final class ServerSideRendering implements Transformer
...
+    /**
+     * Map of attribute names to extraction methods.
+     *
+     * @var string[]
+     */
+    const EXTRACTION_METHOD_MAP = [
+        Attribute::SIZES   => 'extractSizesAttributeCss',
+        Attribute::HEIGHTS => 'extractHeightsAttributeCss',
+        Attribute::MEDIA   => 'extractMediaAttributeCss',
+    ];
```
westonruter: A downside to using a map is that static analysis doesn't recognize that these methods are called:

image

schlessera: I can refactor it into using a switch instead.

schlessera:
I'm currently investigating the visual regression with SSR. What I can already see is that the SSR is somehow changing the images from `layout=responsive` into `layout=fixed`. I suspect that something I'm doing is subsequently misinterpreted by the AMP runtime.

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4482/commits/a7de3cda3b4b55ee5b3f28ad9ce6c14d877c23b7

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: As we already have it, it seems preferable to keep it in here, in case the attribute would have been detached from the element for whatever reason.

schlessera:
The differences were due to the fact that the `sizes`, `heights` and `media` attributes were remove too early and were missing for the layout calculations, which yielded in wrong `layout` attributes to be assigned.

I resolved this by deferring the attribute removal via https://github.com/ampproject/amp-wp/pull/4482/commits/876788264a29731c95a7536b1e898d49351ebcb7

With your post content, on my system it looks identical now. Please double-check on your end.

schlessera:
Based on the above issues I also created a new issue #4671, as I think we're at a point where that could drastically reduce peer review time without impacting general coding too much.

westonruter:
> Based on the above issues I also created a new issue #4671, as I think we're at a point where that could drastically reduce peer review time without impacting general coding too much.

Yes, great idea. That will be very useful.

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +713,267 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+        if (empty($attributeValue)) {
+            return '';
+        }
+
+        if ($attributeValue[0] === '(') {
+            $attributeValue = 'all and ' . $attributeValue;
+        }
+
+        return sprintf(
+            '@media not %s{#%s{display:none}}',
```
sebastianbenz, line 965: What about `not screen and (min-width: 650px)` as input? Won't this result in `not not screen and (min-width: 650px)`?

includes/utils/class-amp-dom-utils.php
```diff
@@ -365,31 +365,22 @@ public static function has_class( DOMElement $element, $class ) {
 	 * If the element does not have an ID, create one first.
 	 *
 	 * @since 1.4.0
+	 * @since 1.5.1 Deprecated for AmpProject\Dom\Document::getElementId()
+	 *
+	 * @deprecated Use AmpProject\Dom\Document::getElementId() instead.
 	 *
 	 * @param DOMElement $element Element to get the ID for.
-	 * @param string     $prefix  Optional. Defaults to '_amp_wp_id_'.
+	 * @param string     $prefix  Optional. Defaults to 'amp-wp-id'.
```
sebastianbenz, line 373: Should we use a common AMP prefix here? e.g.: `i-amp-`?

sebastianbenz: //cc @westonruter

lib/optimizer/src/Transformer/ServerSideRendering.php
sebastianbenz: Another thing: would be nice to combine multiple rules with the same media query, e.g.:

```
@media (min-width: 650px){
    #amp_o_4,#amp_o_5,#amp_o_6{
        display:none
    }
}
```

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -620,4 +708,274 @@ private function isAmpExperimentUsed(DOMElement $element)
...
+        $sourceSizes = explode(',', $attribute->nodeValue);
+        $lastItem    = trim(array_pop($sourceSizes), self::CSS_TRIM_CHARACTERS);
+
+        if (empty($lastItem)) {
+            throw InvalidHtmlAttribute::fromAttribute($attribute->nodeName, $element);
+        }
+
+        $customCss = sprintf($mainStyle, $lastItem);
+
+        foreach ($sourceSizes as $sourceSize) {
```
sebastianbenz, line 914: Do we need to iterate this in reverse order? The spec says:

> The user agent will pick a width from the sizes attribute, using the first item with a <media-condition> (the part in parentheses) that evaluates to true

This means, we have to reverse the order the media queries in CSS to emulate this behavior (the last definition has precedence).

```
<amp-img id="img1" sizes="(min-width: 200px) 200px, (min-width: 320px) 320px, 100vw"></amp-img>
```

Should result in:

```
#elem {
  width: 100vw;
}
@media  (min-width: 320px) {
  #elem {
    width: 320px;
  }
}
@media (min-width: 200px) {
  #elem {
    width: 200px;
  }
}
```

Or am I missing something?

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: Regarding the `not not` case, I did have logic to prevent that in a previous iteration, but I think that got lost in the latest rewrite. I'll have to add that back in. Good catch!

schlessera: Wouldn't it make more sense to do the rule combining in the minifier?

includes/utils/class-amp-dom-utils.php
schlessera: I'd love to use something more official for this, but we'll have to make sure that all logic takes existing Ids into account, then. Otherwise, JS logic might use same Ids than PHP logic, for example.

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: Yes, you're right, the CSS evaluates in the opposite order.

However, considering this, note that the `sizes` example above is actually broken, it would never pick the middle option.

lib/optimizer/src/Transformer/ServerSideRendering.php
sebastianbenz: You're right. CSS optimizer would be the right place. In this particular case it's easy to merge the rules so I did it in the transform, but for example merging the sizes/heights rules would be way more complex to implement hence I didn't implement it either.

westonruter: In the case of the AMP plugin at least at present, the CSS minification runs earlier before the Optimizer. This is because the CSS needs to be minified at a point where we can still strip out CSS that causes it to go over the budget. So I'd recommend outputting pre-minified CSS in the PHP Optimizer and not rely on a minification transformer which would add quite a bit more processing time.

schlessera: The wrong negation and the missing rule concatenation were fixed via https://github.com/ampproject/amp-wp/pull/4799

includes/utils/class-amp-dom-utils.php
schlessera: Default prefix was changed to `i-amp` via https://github.com/ampproject/amp-wp/pull/4799

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: This was missed in a previous follow-up PR. New follow-up PR to fix this: https://github.com/ampproject/amp-wp/pull/5263

schlessera:
**QA passed**

The blocking attributes were correctly removed, the boilerplate was removed as a consequence of that, and none of the resulting images were distorted.

Image 2020-08-25 at 4 12 46 AM

There was one bug that was already discovered here but hasn't been fixed yet. I split that issue out into a new bugfix PR (https://github.com/ampproject/amp-wp/pull/5263) so that this one can pass QA.

PR #4522, opened 2020-04-03
Add button into Site Health to reenable CSS transient caching

### LINKED ISSUE #4523: Add button to reenable CSS transient cache
## Feature description

As the CSS transient cache can be automatically disabled, we should add a button to the Site Health panel for a user to re-enable it.

---------------

## Acceptance criteria

* A user can re-enable the CSS transient caching via a button on the Site Health screen.

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
Add CSS monitoring time series to Site Health debugging info

## Summary

This PR adds a button to the Site Health AMP panel to reenable CSS transient caching when it was disabled.

Button:
Image 2020-04-04 at 9 27 00 AM

Button when the AJAX request failed:
Image 2020-04-04 at 9 26 14 AM

Button when the AJAX request succeeded:
Image 2020-04-04 at 9 29 01 AM

Fixes #4523 

## Checklist


schlessera:
~The colors still need to be adapted to fit the WordPress color palette.~

However, there's currently a problem with the `AMP_Options_Manager`: Within the AJAX request, the `update_option()` call just fails and returns `false`.

@westonruter Are there any requirements to using the `AMP_Options_Manager` ? Do you have an idea of why this would fail within an AJAX request?

pierlon:
> Do you have an idea of why this would fail within an AJAX request?

Perhaps the WP nonce is not being sent along with the request?

schlessera:
The code does execute the call to the `AMP_Options_Manager`, I could verify that.

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
+		ob_start();
+
+		?>
+		<script>
+			( function ( $ ) {
+				$( function () {
+					var admin_ajax_url = '<?php echo addslashes( admin_url( 'admin-ajax.php' ) ); ?>';
```
westonruter, line 186:
```suggestion
					var admin_ajax_url = <?php echo wp_json_encode( admin_url( 'admin-ajax.php' ) ); ?>;
```

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,238 @@
...
+					var selector       = '<?php echo addslashes( $this->selector ); ?>';
```
westonruter:
```suggestion
					var admin_ajax_url = <?php echo wp_json_encode( admin_url( 'admin-ajax.php' ) ); ?>;
					var selector       = <?php echo wp_json_encode( $this->selector ); ?>;
```

src/Admin/AjaxAction.php
westonruter: Actually, this can be removed in favor of using `wp.ajax`

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,238 @@
...
+					setTimeout( function () {
+						( document.querySelectorAll( selector ) || [] )
+							.forEach( ( element ) => {
+								element.addEventListener( 'click', function ( event ) {
+									event.preventDefault();
+									$.ajax( {
+										type: "post",
+										dataType: "json",
+										url: admin_ajax_url,
```
westonruter: This can be replaced with `wp.ajax.post( action, data )`

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,238 @@
...
+										data: { action: '<?php echo addslashes( $this->action ); ?>' }
```
westonruter:
```suggestion
										data: { action: <?php echo wp_json_encode( $this->action ); ?> }
```

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,238 @@
...
+										} )
+										.fail( function () {
+											element.classList.remove( 'ajax-success' );
+											element.classList.add( 'ajax-failure' )
+										} );
+								} );
+							} );
+					}, 1000 );
+				} );
+			} )( jQuery );
```
westonruter:
```suggestion
			} )();
```

Not needed if using `wp.ajax`.

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,238 @@
...
+						( document.querySelectorAll( selector ) || [] )
+							.forEach( ( element ) => {
+								element.addEventListener( 'click', function ( event ) {
+									event.preventDefault();
+									$.ajax( {
+										type: "post",
+										dataType: "json",
+										url: admin_ajax_url,
+										data: { action: '<?php echo addslashes( $this->action ); ?>' }
+									}, { action: '<?php echo addslashes( $this->action ); ?>' } )
```
westonruter: Note using `wp.ajax.post` above

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,238 @@
...
+				} );
+			} )( jQuery );
+		</script>
+		<?php
+
+		$script = ob_get_clean();
+		// phpcs:enable WordPress.Security.EscapeOutput.OutputNotEscaped
+
+		wp_enqueue_script( 'jquery' );
+		wp_add_inline_script( 'jquery', $script );
```
westonruter:
```suggestion
		wp_enqueue_script( 'wp-util' );
		wp_add_inline_script( 'wp-util', $script );
```

schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4522/commits/d1e2aa21ddf73683daa9e92c66d87cfd0056c09d

src/Admin/AjaxAction.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4522/commits/d1e2aa21ddf73683daa9e92c66d87cfd0056c09d

src/Admin/AjaxAction.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4522/commits/d1e2aa21ddf73683daa9e92c66d87cfd0056c09d

src/Admin/AjaxAction.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4522/commits/d1e2aa21ddf73683daa9e92c66d87cfd0056c09d

src/Admin/AjaxAction.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4522/commits/d1e2aa21ddf73683daa9e92c66d87cfd0056c09d

schlessera:
Figured it out. The `AMP_Options_Manager` has an unexpected validation behavior. If an unknown option is passed through, it doesn't allow it to make a change, but also doesn't unset it either, so the option just remains as it was.

schlessera:
Tests failed for unrelated timeout in E2E tests:
https://travis-ci.org/github/ampproject/amp-wp/jobs/671649112#L861-L871
```
 ● Featured Image Notice › should not display a notice, nor suggest cropping, when the image is the expected size
    TimeoutError: Text found "Crop Image"
      waiting for function failed: timeout 500ms exceeded
  ● Featured Image Notice › should display a notice when the image is too small, but not suggest cropping
    TimeoutError: Text found "Crop Image"
      waiting for function failed: timeout 500ms exceeded
```

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,227 @@
...
+                } );
+            } );
+    }, 1000 );
+} )();
+</script>
+JS_SCRIPT;
+
+		wp_enqueue_script( 'wp-util' );
+		wp_add_inline_script( 'wp-util', $script );
+	}
```
pierlon: Using a `script` tag here throws an error:

```
[07-Apr-2020 02:29:32 UTC] PHP Notice:  wp_add_inline_script was called <strong>incorrectly</strong>. Do not pass <code>&lt;script&gt;</code> tags to <code>wp_add_inline_script()</code>. Please see <a href="https://wordpress.org/support/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 4.5.0.) in /app/public/core-dev/src/wp-includes/functions.php on line 5178
[07-Apr-2020 02:29:32 UTC] PHP Stack trace:
[07-Apr-2020 02:29:32 UTC] PHP   1. {main}() /app/public/core-dev/src/wp-admin/plugins.php:0
[07-Apr-2020 02:29:32 UTC] PHP   2. require_once() /app/public/core-dev/src/wp-admin/plugins.php:486
[07-Apr-2020 02:29:32 UTC] PHP   3. do_action() /app/public/core-dev/src/wp-admin/admin-header.php:104
[07-Apr-2020 02:29:32 UTC] PHP   4. WP_Hook->do_action() /app/public/core-dev/src/wp-includes/plugin.php:478
[07-Apr-2020 02:29:32 UTC] PHP   5. WP_Hook->apply_filters() /app/public/core-dev/src/wp-includes/class-wp-hook.php:311
[07-Apr-2020 02:29:32 UTC] PHP   6. AmpProject\AmpWP\Admin\AjaxAction->register_ajax_script() /app/public/core-dev/src/wp-includes/class-wp-hook.php:287
[07-Apr-2020 02:29:32 UTC] PHP   7. wp_add_inline_script() /app/public/content/plugins/amp/src/Admin/AjaxAction.php:208
[07-Apr-2020 02:29:32 UTC] PHP   8. _doing_it_wrong() /app/public/core-dev/src/wp-includes/functions.wp-scripts.php:123
[07-Apr-2020 02:29:32 UTC] PHP   9. trigger_error() /app/public/core-dev/src/wp-includes/functions.php:5178
```

pierlon:
> Tests failed for unrelated timeout in E2E tests

@schlessera these tests are failing because of an error being thrown (see https://github.com/ampproject/amp-wp/pull/4522#discussion_r404500884).

pierlon:
I think this button could be highlighted a little better, given how deep it's hidden in Site Health. I don't see how the user would know about it unless they've been told where it is.

schlessera:
> I don't see how the user would know about it unless they've been told where it is.

I assumed for now that this is mainly for the support team to be able to point them to some place to reset the cache instead of handing them a snippet filter or a custom filter to do so.

I don't think this should be a standard user operation.

But I'm happy to change this if my assumption is incorrect.

@westonruter @amedina Thoughts?

src/Admin/AjaxAction.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4522/commits/3ae8fedd42a8fd9187668bb0bc7bda00a8c4f510

westonruter:
> I assumed for now that this is mainly for the support team to be able to point them to some place to reset the cache instead of handing them a snippet filter or a custom filter to do so.
> 
> I don't think this should be a standard user operation.
> 
> But I'm happy to change this if my assumption is incorrect.

I think it's placement makes sense now. The user will discover it when they see a Site Health warning (which now as of WordPress 5.4 is on the admin dashboard) and then it will be a logical extension to expand the audit for more information and to find the button.

src/Admin/SiteHealth.php
```diff
@@ -497,4 +519,60 @@ public function add_extensions( $extensions ) {
...
+
+                .wp-core-ui .button.reenable-css-transient-caching.ajax-success ~ .success-icon {
+                    color: #46b450;
+                }
+
+                .wp-core-ui .button.reenable-css-transient-caching.ajax-failure ~ .failure-icon {
+                    color: #dc3232;
+                }
+            </style>
+        ';
```
westonruter: It seems spaces are being used here when tabs are expected.

src/Admin/SiteHealth.php
```diff
@@ -497,4 +519,60 @@ public function add_extensions( $extensions ) {
...
+
+	/**
+	 * Re-enable the CSS Transient caching.
+	 *
+	 * This is triggered via an AJAX call from the Site Health panel.
+	 */
+	public function reenable_css_transient_caching() {
+		if ( ! current_user_can( 'manage_options' ) ) {
+			wp_send_json_error( 'Unauthorized.', 401 );
+		}
```
westonruter: What about nonce checking?

src/Admin/SiteHealth.php
```diff
@@ -264,6 +275,17 @@ public function css_transient_caching() {
 			'status'      => $status,
 			'label'       => esc_html( $label ),
 		];
+
+		if ( $disabled ) {
+			$data['actions'] = sprintf(
+				'<p><a class="button reenable-css-transient-caching" href="#">%s</a><span class="dashicons dashicons-yes success-icon"></span><span class="dashicons dashicons-no failure-icon"></span><span class="success-text">%s</span><span class="failure-text">%s</span></p>',
+				__( 'Re-enable transient caching', 'amp' ),
+				__( 'Reload the page to refresh the diagnostic check.', 'amp' ),
+				__( 'The operation failed, please reload the page and try again.', 'amp' )
```
westonruter: Would these warrant `esc_html__()`? I know the others aren't doing it.

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,225 @@
...
+                            element.classList.remove( 'ajax-failure' );
+                            element.classList.add( 'ajax-success' )
+                        } )
+                        .fail( function () {
+                            element.classList.remove( 'ajax-success' );
+                            element.classList.add( 'ajax-failure' )
+                        } );
+                } );
+            } );
+    }, 1000 );
```
westonruter: Why the 1-second timeout? Wouldn't a `DOMContentLoaded` event handler suffice?

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,225 @@
...
+ */
+
+namespace AmpProject\AmpWP\Admin;
+
+/**
+ * Base class to define a new AJAX action.
+ *
+ * @package AmpProject\AmpWP
+ */
+final class AjaxAction {
```
westonruter: I wonder… should the REST API rather be used?

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,225 @@
...
+
+
```
westonruter:
```suggestion
```

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,225 @@
...
+	 * @var int
+	 */
+	const SCOPE_BACKEND = 1;
+
+	/**
+	 * The AJAX handler listens on the frontend only.
+	 *
+	 * @var int
+	 */
+	const SCOPE_FRONTEND = 2;
```
westonruter: We have no use case for frontend scope, do we? Perhaps the scopes should be omitted in favor of leaving this to be exclusively the realm of the admin? If we add actions to the frontend, I'm sure it will look different. 

I think the scopes can be removed.

pierlon:
> The user will discover it when they see a Site Health warning (which now as of WordPress 5.4 is on the admin dashboard) and then it will be a logical extension to expand the audit for more information and to find the button.

Ah right, that makes sense.

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,225 @@
...
+	 * @var int
+	 */
+	const ACCESS_USER = 1;
+
+	/**
+	 * The AJAX action should only be accessible by unauthenticated visitors.
+	 *
+	 * @var int
+	 */
+	const ACCESS_NO_USER = 2;
```
westonruter: I think we can just restrict this to authenticated users, as we don't have a use case for unauthenticated admin actions. These could be supported later if we need to.

src/Admin/AjaxAction.php
westonruter: Also, maybe it is premature to create an abstract `AjaxAction` construct. How about just naming this to be specific for the purpose of the re-enabling transient caching? Once we have another use case, then it would make sense to create an abstraction construct since we'll have another data point to inform it.

src/Admin/AjaxAction.php
```diff
@@ -0,0 +1,225 @@
...
+	 * @var string
+	 */
+	const FRONTEND_ENQUEUE_ACTION = 'wp_enqueue_scripts';
+
+	/**
+	 * Action to use for enqueueing the JS logic at the backend.
+	 *
+	 * @var string
+	 */
+	const BACKEND_ENQUEUE_ACTION = 'admin_enqueue_scripts';
```
westonruter: I think this needs to be more fine-grained. We only want it to fire on the site health screen, not on any other admin page.

src/Admin/SiteHealth.php
schlessera: Yes, we should escape them.

src/Admin/SiteHealth.php
schlessera: Will add.

src/Admin/AjaxAction.php
schlessera: I tried that but it seems that the SiteHealth screen does other magic _after_ the DOM content was loaded, and I also couldn't find an event to check for. Any suggestion what I could use here instead of a dumb timeout?

src/Admin/AjaxAction.php
schlessera: Right now, this only starts the check to see if we're on the Site Health screen. The selector to attach the click event to is the very first thing that is being checked with an early return, so nothing happens when our button is not present.

src/Admin/AjaxAction.php
schlessera: -.-

src/Admin/AjaxAction.php
westonruter: Something is still happening, though: CPU cycles to parse and evaluate the JS. This JS is being enqueued on the page which may not need it, namely `wp-util`. None of this JS should appear on any page other than Site Health.

src/Admin/SiteHealth.php
schlessera: Wouldn't it be better to have spaces in the HTML markup rather than tabs?

Also, things like this would be nicer via views so you can split the logic from the markup. We should discuss at one point whether we want to have a very simple view abstraction (can be a single class if we want to keep it simple) to pull in external (templated) files.

westonruter: Coding standard for HTML is to use tabs: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/html/#indentation

src/Admin/AjaxAction.php
schlessera: ~No, it will only ever be added on the page that contains this "button": `a.reenable-css-transient-caching`. That is the very first early return in the method that is hook to this action:~

~https://github.com/ampproject/amp-wp/pull/4522/files#diff-e97af8c0e681095ef90cd07c4cea405dR166-R168~

~As that button will only be present on the Site Health page, the JS will also only be enqueued there.~

src/Admin/SiteHealth.php
schlessera: Love the code examples on that page, btw... :)

src/Admin/AjaxAction.php
schlessera: I wrote nonsense above, as the code is only checking whether the variable is set, not whether a query against the DOM succeeded.

I'm adapting the code to use the `$hook_suffix` argument for the `admin_enqueue_scripts` action.

src/Admin/AjaxAction.php
schlessera: Hmm, it seems to work now. I must have done something else wrong when I tried it in the beginning.

schlessera:
As there's no code in place for replacing the nonce once it's been used, the button only works once per page load. However, there's no valid reason to push it multiple times, so I disabled it after the first use, and any further tries will need a reload.

Image 2020-04-09 at 5 56 34 PM

westonruter:
I fleshed out the guidance a bit more, and added link to learn more about object caching:

image

When enabled:

image

westonruter:
Confirmed error condition is shown properly:

image

src/Admin/ReenableCssTransientCachingAjaxAction.php
```diff
@@ -0,0 +1,126 @@
...
+	 * Register the AJAX logic.
+	 *
+	 * @param string $hook_suffix Hook suffix to identify from what admin page the call is coming from.
+	 */
+	public function register_ajax_script( $hook_suffix ) {
+		if ( 'site-health.php' !== $hook_suffix ) {
+			return;
+		}
+
+		$script = <<< JS_SCRIPT
```
schlessera: If we don't use string interpolation, we can use a nowdoc instead of a heredoc:
```suggestion
		$script = <<< 'JS_SCRIPT'
```

kienstra:
**Looks good**

Hi @amedina,
This worked as expected in my local.

After forcing the option to be `true`:

```php
AMP_Options_Manager::update_option( 'amp_css_transient_monitor_disable_caching', true );
```
...the test appeared in Site Health:

looks-good-here


tests-passed

PR #4538, opened 2020-04-06
Restrict doing plugin upgrade routine when not in admin

### LINKED ISSUE #3284: Reduce scenarios where DB writes are performed during frontend requests

westonruter:
## Summary

The changes here in effect restrict the upgrade routine to happen at `admin_init`. This was discussed in depth at #4177. This implements https://github.com/ampproject/amp-wp/issues/3284#issuecomment-601461737:

> The DB upgrade routines don't happen on the frontend either, as you recall the upgrade DB message is only presented once trying to access the admin.

Fixes #3284

## Checklist

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -124,7 +124,7 @@ function amp_init() {
 	 */
 	$options     = get_option( AMP_Options_Manager::OPTION_NAME, [] );
 	$old_version = isset( $options['version'] ) ? $options['version'] : '0.0';
-	if ( AMP__VERSION !== $old_version ) {
+	if ( AMP__VERSION !== $old_version && is_admin() && current_user_can( 'manage_options' ) ) {
 		/**
 		 * Triggers when after amp_init when the plugin version has updated.
 		 *


schlessera:
Before doing this, we'll need to audit what will happen if a new version of the plugin runs with an old DB version, as the upgrade routine will now not run unconditionally anymore and it might be hours, days or weeks between updating the plugin files and updating the DB.

This can lead to unexpected behavior and breakage if the new code doesn't expect the old DB schema.

westonruter:
@schlessera There are only two instances of the `amp_plugin_update` hook being used at present:

https://github.com/ampproject/amp-wp/blob/cfd87a240138ccfe86b2b81bdf0c819f9fcfd116/includes/validation/class-amp-validated-url-post-type.php#L147-L173

https://github.com/ampproject/amp-wp/blob/cfd87a240138ccfe86b2b81bdf0c819f9fcfd116/src/BackgroundTask/MonitorCssTransientCaching.php#L158-L168

I don't see how either of these will cause problems if they are deferred to when the admin user logs in. For the first, it's only for sites running the AMP plugin on versions before 1.0 stable. And it is just renaming a post type for the validated URLs.  There's no impact on that being deferred.

For the second, it's just making sure the transient caching gets re-enabled when upgrading from 1.5.0+ to 1.5.2. This also can wait until the admin user logs in.

So in both cases I see no issue to move forward.

schlessera:
Looks good then. We just need to keep this in mind when we use this hook. It changes from enforcing consistency after upgrades to hopefully end up with eventual consistency instead. So we cannot rely on anything that happens during an update event, it will either need to be optional or trigger user feedback in some way.

schlessera (approved):

pierlon:
Verified in QA. The upgrade routine is only run when an admin logs in that has the permission of managing options.

PR #4542, opened 2020-04-07
Expose fatal errors that occur while validating a URL

### LINKED ISSUE #4580: Fatal errors are not exposed when they occur during validation requests
## Bug Description

Given a plugin like this:

```php
	for ( $i = 0; $i < 8; $i++ ) {
		if ( $i >= 4 && AMP_Validation_Manager::$is_validate_request ) {
			throw new Exception( 'FAIL' );
		}
		echo "<script>// bad: $i</script>";
	}
} );
```

On the frontend of the site, the admin bar will show 8 validation errors:

image

However, when clicking “Validate” the user will only see 4:

image

This is confusing. There was a fatal error that occurred but this is not exposed in any way. In fact, #4533 further hid the fatal error by forcing validation responses to have a 200 status code. Therefore, what is needed is a way to capture any fatal error which occurred during validation and display it to the user. Such a fatal error may indicate a bug in the AMP plugin with how source stacks are obtained, so this is important for the user to be able to relay to the support forums. This PR allows for this by capturing the fatal error and displaying it on the validated URL screen in an admin notice:

image

This came up in the course of helping to debug a [support topic](https://wordpress.org/support/topic/unable-to-validate-urls/). In that particular case, the issue turned out to be due to conflict with the [Widget Options](https://wordpress.org/plugins/widget-options/) plugin (which is actually a [known issue](https://wordpress.org/support/topic/standard-template-mode-internal-error/#post-12547136)). So given a site where the Widget Options plugin is installed as well as this [Old-Style Single Widget with Wide Control](https://gist.github.com/westonruter/7141599) gist plugin with the widget being added to the sidebar, even if the above `wp_footer` plugin code is active to emit validation errors in the footer, they will be completely hidden from the validation results because of the fatal error happening when rendering the sidebar:

image

However, with the changes in this PR checked out, an error is shown to explain why the frontend has validation errors being reported in the admin bar but they aren't showing up on the validated URL screen:

image

## Expected Behaviour

Fatal errors should be indicated in some way.

## Steps to reproduce

See above.

## Screenshots

See above.

## Additional context

 - WordPress version:
 - Plugin version:
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

westonruter:
## Summary

Fixes #4580.

## Checklist

### DIFF
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -2030,7 +2030,7 @@ class AMP_Theme_Support {
 				'php_fatal_error'  => false,
 			];
 			$last_error = error_get_last();
-			if ( $last_error && in_array( $last_error['type'], [ E_ERROR, E_RECOVERABLE_ERROR, E_CORE_ERROR, E_COMPILE_ERROR ], true ) ) {
+			if ( $last_error && in_array( $last_error['type'], [ E_ERROR, E_RECOVERABLE_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_USER_ERROR, E_PARSE ], true ) ) {
 				$data['php_fatal_error'] = $last_error;
 			}
 			$data = array_merge( $data, AMP_Validation_Manager::get_validate_response_data( $sanitization_results ) );
--- a/includes/class-amp-theme-support.php
+++ b/includes/class-amp-theme-support.php
@@ -2025,10 +2025,16 @@ class AMP_Theme_Support {
 		if ( AMP_Validation_Manager::$is_validate_request ) {
 			status_header( 200 );
 			header( 'Content-Type: application/json; charset=utf-8' );
-			return wp_json_encode(
-				AMP_Validation_Manager::get_validate_response_data( $sanitization_results ),
-				JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES
-			);
+			$data       = [
+				'http_status_code' => $status_code,
+				'php_fatal_error'  => false,
+			];
+			$last_error = error_get_last();
+			if ( $last_error && in_array( $last_error['type'], [ E_ERROR, E_RECOVERABLE_ERROR, E_CORE_ERROR, E_COMPILE_ERROR ], true ) ) {
+				$data['php_fatal_error'] = $last_error;
+			}
+			$data = array_merge( $data, AMP_Validation_Manager::get_validate_response_data( $sanitization_results ) );
+			return wp_json_encode( $data, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES );
 		}
 
 		// Determine what the validation errors are.
--- a/includes/validation/class-amp-validated-url-post-type.php
+++ b/includes/validation/class-amp-validated-url-post-type.php
@@ -848,6 +848,14 @@ class AMP_Validated_URL_Post_Type {
 			// Note that json_encode() is being used here because wp_slash() will coerce scalar values to strings.
 			update_post_meta( $post_id, '_amp_stylesheets', wp_slash( wp_json_encode( $args['stylesheets'] ) ) );
 		}
+		if ( isset( $args['php_fatal_error'] ) ) {
+			if ( empty( $args['php_fatal_error'] ) ) {
+				delete_post_meta( $post_id, '_amp_php_fatal_error' );
+			} else {
+				// Note that json_encode() is being used here because wp_slash() will coerce scalar values to strings.
+				update_post_meta( $post_id, '_amp_php_fatal_error', wp_slash( wp_json_encode( $args['php_fatal_error'] ) ) );
+			}
+		}
 
 		delete_transient( static::NEW_VALIDATION_ERROR_URLS_COUNT_TRANSIENT );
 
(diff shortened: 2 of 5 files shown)


includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -848,6 +848,14 @@ public static function store_validation_errors( $validation_errors, $url, $args
...
 			update_post_meta( $post_id, '_amp_stylesheets', wp_slash( wp_json_encode( $args['stylesheets'] ) ) );
 		}
+		if ( isset( $args['php_fatal_error'] ) ) {
+			if ( empty( $args['php_fatal_error'] ) ) {
+				delete_post_meta( $post_id, '_amp_php_fatal_error' );
+			} else {
+				// Note that json_encode() is being used here because wp_slash() will coerce scalar values to strings.
+				update_post_meta( $post_id, '_amp_php_fatal_error', wp_slash( wp_json_encode( $args['php_fatal_error'] ) ) );
+			}
+		}
```
westonruter, line 858: Note the data model for `amp_validated_url` posts needs to be revisited, but now's not the time. For this reason, I've kept the existing scheme of adding more data to postmeta, and making the changes in various places when in the future it should be centralized in a single model.

includes/class-amp-theme-support.php
```diff
@@ -2025,10 +2025,16 @@ public static function prepare_response( $response, $args = [] ) {
...
-			return wp_json_encode(
-				AMP_Validation_Manager::get_validate_response_data( $sanitization_results ),
-				JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES
-			);
+			$data       = [
+				'http_status_code' => $status_code,
+				'php_fatal_error'  => false,
+			];
+			$last_error = error_get_last();
+			if ( $last_error && in_array( $last_error['type'], [ E_ERROR, E_RECOVERABLE_ERROR, E_CORE_ERROR, E_COMPILE_ERROR ], true ) ) {
```
schlessera: I'd add `E_USER_ERROR` (someone used `trigger_error()` in their code somewhere) and `E_PARSE` (some PHP file could not be successfully parsed, for example because of a PHP version mismatch).
```suggestion
			if ( $last_error && in_array( $last_error['type'], [ E_ERROR, E_RECOVERABLE_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_USER_ERROR, E_PARSE ], true ) ) {
```

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -1401,6 +1409,11 @@ public static function print_admin_notice() {
 			);
 		}
 
+		// Add notice for PHP fatal error during validation request.
+		if ( 'post' === get_current_screen()->base && self::POST_TYPE_SLUG === get_current_screen()->post_type && get_post() ) {
+			self::render_php_fatal_error_admin_notice( get_post() );
+		}
```
schlessera: `get_post()` is being called twice which is bad for performance and could potentially cause weird bugs where the second call returns something else than the first call.
```suggestion
		// Add notice for PHP fatal error during validation request.
		$post = get_post();
		if ( $post instanceof WP_Post && 'post' === get_current_screen()->base && self::POST_TYPE_SLUG === get_current_screen()->post_type ) {
			self::render_php_fatal_error_admin_notice( $post );
		}
```

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -1502,6 +1515,34 @@ public static function print_admin_notice() {
 		}
 	}
 
+	/**
+	 * Render PHP fatal error admin notice.
+	 *
+	 * @param WP_Post $post Post.
+	 */
+	private static function render_php_fatal_error_admin_notice( WP_Post $post ) {
```
schlessera, line 1524: I'd recommend importing the class `WP_Post` here ( `use WP_Post;` at the top ) because:
- letting PHP guess where the class comes from incurs a performance hit
- using an import makes it clear that the code depends on this class
- using an explicit import makes sure this doesn't break if we refactor the class to be within a namespace (as the relative `WP_Post` would then take on the namespace as well).

schlessera (changes requested):

includes/validation/class-amp-validated-url-post-type.php
westonruter: Does this matter right now since WordPress core isn't using an autoloader? The `WP_Post` class will be a global already.

Also, IntelliJ doesn't like it:

image

westonruter: Actually, this is also emitting a PHP warning:

> Warning: The use statement with non-compound name 'WP_Post' has no effect in /app/public/content/plugins/amp/includes/validation/class-amp-validated-url-post-type.php on line 9

westonruter:
This is important for users to be able to find out what what the fatal error was without having to go to the error log.

includes/validation/class-amp-validated-url-post-type.php
schlessera: Ah, right, because the `AMP_Validated_URL_Post_Type ` class itself does not use a namespace right now.

Nevermind, then.

schlessera (approved):

PR #4547, opened 2020-04-07
Restore unification of multi-page post content in Reader mode

westonruter:
## Summary

Originally reported on a [support forum topic](https://wordpress.org/support/topic/pagination-error-22/):

<blockquote>Since the last update to your AMP plugin (which we’ve been using with the Zox News WordPress theme for over a year), we’ve encountered this issue. We use <code>&lt;!--nextpage--&gt;</code> to paginate our big features on desktop/mobile (like <a href="https://www.slantmagazine.com/features/the-100-best-westerns-of-all-time/" rel="nofollow ugc">this one</a>). Your plugin has always ignored this code and presented those features as one big page. Now, though, our AMP pages are killing all content after it hits the <code>&lt;!--nextpage--&gt;</code> directive, meaning that our readers aren’t seeing our full content. Click <a href="https://www.slantmagazine.com/features/the-100-best-westerns-of-all-time/amp/" rel="nofollow ugc">here</a> to see how the AMP version of the list above is ending at number 91. We want to avoid having to un-paginate our features in order to correct this issue. Thanks.</blockquote>

The regression was introduced in #3781 via this change: https://github.com/ampproject/amp-wp/pull/3781/files#r405130526

The switch to `get_the_content()` caused pagination to be introduced which was not present before, when `post_content` was directly passed into `the_content` filters.

## Checklist

### DIFF
--- a/includes/templates/class-amp-post-template.php
+++ b/includes/templates/class-amp-post-template.php
@@ -342,8 +342,14 @@ class AMP_Post_Template {
 	 * Build post content.
 	 */
 	private function build_post_content() {
-		/** This filter is documented in wp-includes/post-template.php */
-		$content = apply_filters( 'the_content', get_the_content( null, false, $this->post ) );
+		if ( post_password_required( $this->post ) ) {
+			$content = get_the_password_form( $this->post );
+		} else {
+			// Note: This is intentionally not using get_the_content() because the legacy behavior of posts in Reader
+			// mode is to display multi-page posts as a single page without any pagination links.
+			/** This filter is documented in wp-includes/post-template.php */
+			$content = apply_filters( 'the_content', $this->post->post_content );
+		}
 
 		$this->add_data_by_key( 'post_amp_content', $content );
 	}
--- a/includes/templates/class-amp-post-template.php
+++ b/includes/templates/class-amp-post-template.php
@@ -347,7 +347,7 @@ class AMP_Post_Template {
 		} else {
 			/**
 			 * This filter is documented in wp-includes/post-template.php.
-			 * 
+			 *
 			 * Note: This is intentionally not using get_the_content() because the legacy behavior of posts in
 			 * Reader mode is to display multi-page posts as a single page without any pagination links.
 			 */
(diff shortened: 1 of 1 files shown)


includes/templates/class-amp-post-template.php
```diff
@@ -342,8 +342,14 @@ private function build_post_comments_data() {
 	 * Build post content.
 	 */
 	private function build_post_content() {
-		/** This filter is documented in wp-includes/post-template.php */
-		$content = apply_filters( 'the_content', get_the_content( null, false, $this->post ) );
```
westonruter, line 346: Note that this did not need the `post_password_required()` check before because `get_the_content()` has that logic inside of it.

westonruter:
Workaround plugin which implements this while waiting for 1.5.3: https://gist.github.com/westonruter/559c7a9cbb9460d1099c3d5d9bfa5ad1

pierlon (approved):

includes/templates/class-amp-post-template.php
```diff
@@ -342,8 +342,14 @@ private function build_post_comments_data() {
...
+		if ( post_password_required( $this->post ) ) {
+			$content = get_the_password_form( $this->post );
+		} else {
+			// Note: This is intentionally not using get_the_content() because the legacy behavior of posts in Reader
+			// mode is to display multi-page posts as a single page without any pagination links.
+			/** This filter is documented in wp-includes/post-template.php */
```
schlessera: Attaching two different comments together seems weird here. How about this:
```suggestion
			/**
			 * This filter is documented in wp-includes/post-template.php.
			 * 
			 * Note: This is intentionally not using get_the_content() because the legacy behavior of posts in
			 * Reader mode is to display multi-page posts as a single page without any pagination links.
			 */
```

schlessera (changes requested):

includes/templates/class-amp-post-template.php
westonruter: I originally intended it to be separate, as the inline comment is talking about the parameter and the PhpDoc comment is talking about the filter as a whole. But I'm fine with combining.

schlessera (approved):

PR #4574, opened 2020-04-12
Refactor conditional checks in is_amp_endpoint() and improve guidance when _doing_it_wrong()

### LINKED ISSUE #4525: Error caused by using is_embed() and is_feed() in is_amp_endpoint() before condition (! $did_parse_query)
## Bug Description

When calling is_amp_endpoints() at an early stage in a plugin, the conditionals is_embed() en is_feed() issue an error notice, saying it can't be used before the query is finished. It doesn't reach the code that is used to test if the function is started too early, so the cause is unclear at first.

## Expected Behaviour

The function tests if it is started before the 'parse_query' hook and starts doing_it_wrong saying is_amp_endpoint() it is started to early in the process. This would be the intended behaviour when it is started too early. it would be solved by moving the if(! $did_parse_query) block up so it runs and before testing the template tags. now it doesn't reach that test depending on how early it is called.

## Steps to reproduce

1. Add a if statement with the condition is_amp_embed() at the global level of a plugin
2. activate the plugin
3. refresh a page at the website's frontend.
4. See error in the log (or whoops)

## Screenshots

the notice is:
is_embed was called <strong>incorrectly</strong>. Conditional query tags do not work before the query is run. Before then, they always return false. Please see <a href="https://wordpress.org/support/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 3.1.0.)

## Additional context

 - WordPress version: 5.4
 - Plugin version:  1.5.2
 - PHP version: 7.3.2

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

* Improve the `_doing_it_wrong()` message for when `is_amp_endpoint()` is called before the `wp` action.
* Account for additional template conditionals: `is_comment_feed()`, `is_trackback()`, `is_robots()`, and `is_favicon()`.
* Remove special condition to allow `is_amp_endpoint()` to not emit `_doing_it_wrong()` when called before `wp` when in Reader mode.
* Prevent emitting `_doing_it_wrong()` errors when by calling `is_embed()` or `is_feed()` before they are able to be called.
* Prevent emitting duplicated warnings.
* Remove separate condition for doing redirects to non-AMP when AMP is not available in Reader mode.
* Remove unnecessary `exit` logic.
* Fix ability to access the `page_for_posts` if designated to be enabled in AMP Reader mode.

hansschuijff:
Moved checking against is_embed() and is_feed() after
checking if parse_query has run and $wp_query has been filled, since
both these functions will fire _doing_it_wrong when $wp_query
is not yet set.

## Summary

Fixes #4525

## Checklist

### DIFF
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -603,7 +603,7 @@ function is_amp_endpoint() {
 		);
 	}
 
-	// note: is_embed() and is_feed() need $wp_query, so above checks must go first
+	// note: is_embed() and is_feed() need $wp_query, so above checks must go first.
 	if ( is_embed() || is_feed() ) {
 		return false;
 	}
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -603,7 +603,7 @@ function is_amp_endpoint() {
 		);
 	}
 
-	// note: is_embed() and is_feed() need $wp_query, so above checks must go first.
+	// Note: is_embed() and is_feed() need $wp_query, so above checks must go first.
 	if ( is_embed() || is_feed() ) {
 		return false;
 	}
--- a/includes/amp-helper-functions.php
+++ b/includes/amp-helper-functions.php
@@ -566,45 +566,45 @@ function post_supports_amp( $post ) {
 function is_amp_endpoint() {
 	global $pagenow, $wp_query;
 
-	if ( is_admin() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) || in_array( $pagenow, [ 'wp-login.php', 'wp-signup.php', 'wp-activate.php' ], true ) ) {
+	// Short-circuit for admin requests or requests to non-frontend pages.
+	if ( is_admin() || in_array( $pagenow, [ 'wp-login.php', 'wp-signup.php', 'wp-activate.php' ], true ) ) {
 		return false;
+	}
+
+	$warned        = false;
+	$error_message = sprintf(
+		/* translators: %1$s: is_amp_endpoint(), %2$s: the current action, %3$s: the wp action, %4$s: the parse_request action, %5$s: the WP_Query class, %6$s: the amp_skip_post() function */
+		__( '%1$s was called too early and so it will not work properly. WordPress is currently doing the "%2$s" action. Calling this function before the "%3$s" action means it will not have access to %4$s and the queried object to determine if it is an AMP response, thus neither the %5$s filter nor the AMP enabled toggle will be considered.', 'amp' ),
+		__FUNCTION__ . '()',
+		current_action(),
+		'wp',
+		'WP_Query',
+		'amp_skip_post()'
+	);
+
+	// Make sure the parse_request action has triggered before trying to read from the REST_REQUEST constant, which is set during rest_api_loaded().
+	if ( ! did_action( 'parse_request' ) ) {
+		_doing_it_wrong( __FUNCTION__, esc_html( $error_message ), '1.5.3' );
+		$warned = true;
+	} elseif ( defined( 'REST_REQUEST' ) && REST_REQUEST ) {
+		return false;
+	}
+
+	// Make sure that the parse_query action has triggered, as this is required to initially populate the global WP_Query.
+	if ( ! $warned && ! ( did_action( 'parse_query' ) || $wp_query instanceof WP_Query ) ) {
+		_doing_it_wrong( __FUNCTION__, esc_html( $error_message ), '0.4.2' );
+		$warned = true;
 	}
 
 	// Always return false when requesting service worker.
-	if ( class_exists( 'WP_Service_Workers' ) && ! empty( $wp_query ) && defined( 'WP_Service_Workers::QUERY_VAR' ) && $wp_query->get( WP_Service_Workers::QUERY_VAR ) ) {
+	// Note this is no longer required because AMP_Theme_Support::prepare_response() will abort for non-HTML responses.
+	if ( class_exists( 'WP_Service_Workers' ) && $wp_query instanceof WP_Query && defined( 'WP_Service_Workers::QUERY_VAR' ) && $wp_query->get( WP_Service_Workers::QUERY_VAR ) ) {
 		return false;
 	}
 
-	$did_parse_query = did_action( 'parse_query' );
-
-	if ( ! $did_parse_query ) {
-		_doing_it_wrong(
-			__FUNCTION__,
-			sprintf(
-				/* translators: 1: is_amp_endpoint(), 2: parse_query */
-				esc_html__( '%1$s was called before the %2$s hook was called.', 'amp' ),
-				'is_amp_endpoint()',
-				'parse_query'
-			),
-			'0.4.2'
-		);
-	}
-
-	if ( empty( $wp_query ) || ! ( $wp_query instanceof WP_Query ) ) {
-		_doing_it_wrong(
-			__FUNCTION__,
-			sprintf(
-				/* translators: 1: is_amp_endpoint(), 2: WP_Query */
-				esc_html__( '%1$s was called before the %2$s was instantiated.', 'amp' ),
-				'is_amp_endpoint()',
-				'WP_Query'
-			),
-			'1.1'
-		);
-	}
-
-	// Note: is_embed() and is_feed() need $wp_query, so above checks must go first.
-	if ( is_embed() || is_feed() ) {
+	// Short-circuit for embed or feed requests since they will never be AMP responses.
+	// Note that these conditionals only require the parse_query action to have been run. They don't depend on the wp action having been fired.
+	if ( $wp_query instanceof WP_Query && ( $wp_query->is_embed() || $wp_query->is_feed() ) ) {
 		return false;
 	}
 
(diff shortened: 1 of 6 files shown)


includes/amp-helper-functions.php
```diff
@@ -603,6 +603,11 @@ function is_amp_endpoint() {
 		);
 	}
 
+	// note: is_embed() and is_feed() need $wp_query, so above checks must go first
```
schlessera: The Travis CI tests are currently on the following issue:
```
FILE: includes/amp-helper-functions.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 606 | ERROR | Inline comments must end in full-stops, exclamation marks, or
     |       | question marks
     |       | (Squiz.Commenting.InlineComment.InvalidEndChar)
--------------------------------------------------------------------------------
```
(see https://travis-ci.org/github/ampproject/amp-wp/jobs/674004902#L907-L914)

Until we fix this, we won't see whether the rest of the tests are passing.

Here's how we can fix this:
```suggestion
	// Note: is_embed() and is_feed() need $wp_query, so above checks must go first.
```

You can just choose to "Commit suggestion" to include this change, just make sure you give it an appropriate commit message instead of the GitHub default.

schlessera (changes requested):
Thanks so much for the pull request @hansschuijff !

There's a code style issue right now that prevents the actual tests from running. See my review comment on how to fix it.

includes/amp-helper-functions.php
hansschuijff: I changed it and committed the change, is that enough? It's my first time at this. :)

schlessera: Yes. As you can see, the tests are passing now: https://travis-ci.org/github/ampproject/amp-wp/builds/674015488

hansschuijff: Great, thanks.

schlessera:
@hansschuijff It looks like there's no obvious way right now to add a regression test for the above fix. I've opened an issue to that regard: https://github.com/ampproject/amp-wp/issues/4576

includes/amp-helper-functions.php
```diff
@@ -603,6 +603,11 @@ function is_amp_endpoint() {
 		);
 	}
 
+	// note: is_embed() and is_feed() need $wp_query, so above checks must go first.
```
```suggestion
	// Note: is_embed() and is_feed() need $wp_query, so above checks must go first.
```

includes/amp-helper-functions.php
```diff
@@ -566,7 +566,7 @@ function post_supports_amp( $post ) {
 function is_amp_endpoint() {
 	global $pagenow, $wp_query;
 
-	if ( is_admin() || is_embed() || is_feed() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) || in_array( $pagenow, [ 'wp-login.php', 'wp-signup.php', 'wp-activate.php' ], true ) ) {
+	if ( is_admin() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) || in_array( $pagenow, [ 'wp-login.php', 'wp-signup.php', 'wp-activate.php' ], true ) ) {
```
westonruter: I just realized that `REST_REQUEST` maybe being checked too early here. It is defined in the `rest_api_loaded()` function which is called at the `parse_request` action. So perhaps we need to add another check for whether the `parse_request` action was done.

hansschuijff: If I understand it correctly, that seems to not lead to problems, since than REST_REQUEST just won't be defined and it will pass through the conditional and just fire _doing_it_wrong until the parse_request action is performed and the $wp_query is filled. Is_embed and is_feed just needed to move since otherwise the error reported had no obvious relation with is_amp_endpoint. It might be right to let the function report an error anyway if the function is called too early in the process, so that the plugin developers get to see where what they did wrong?

westonruter: I think the issue is just that during REST API requests, the `wp` action is never triggered so it's somewhat misleading. In order to check the constant we need to check if the `parse_request` action was fired, or else it's not good to check this constant since it will never be defined. I'm going to push up a change to this to account for this and you can see what I mean.

westonruter: OK, I've rearranged the checks in 0bdf639 which I think will produce better results. This change will also ensure that a call to `is_amp_endpoint()` won't cause a warning in addition to warnings about calling `is_embed()` and `is_feed()`.

westonruter:
One more thing that comes to mind to improve this is to look at the call stack to find out which theme/plugin is actually responsible for calling `is_amp_endpoint()` prematurely. While this is out of scope of this PR, it's something that we can either add here or look at in another PR.

includes/amp-helper-functions.php
```diff
@@ -566,41 +568,62 @@ function post_supports_amp( $post ) {
...
-			__FUNCTION__,
-			sprintf(
-				/* translators: 1: is_amp_endpoint(), 2: parse_query */
-				esc_html__( '%1$s was called before the %2$s hook was called.', 'amp' ),
-				'is_amp_endpoint()',
-				'parse_query'
-			),
-			'0.4.2'
-		);
+	// Always return false when requesting service worker.
```
schlessera:
```suggestion
	// Always return false when requesting the service worker.
```

includes/amp-helper-functions.php
```diff
@@ -566,41 +568,62 @@ function post_supports_amp( $post ) {
...
-			__FUNCTION__,
-			sprintf(
-				/* translators: 1: is_amp_endpoint(), 2: WP_Query */
-				esc_html__( '%1$s was called before the %2$s was instantiated.', 'amp' ),
-				'is_amp_endpoint()',
-				'WP_Query'
-			),
-			'1.1'
-		);
+	// Short-circuit queries that can never AMP responses (e.g. post embeds and feeds).
```
schlessera:
```suggestion
	// Short-circuit queries that can never have AMP responses (e.g. post embeds and feeds).
```

includes/amp-helper-functions.php
```diff
@@ -623,34 +646,31 @@ function is_amp_endpoint() {
...
-			'1.0.2'
-		);
-		$supported = true;
+	if ( did_action( 'wp' ) && $wp_query instanceof WP_Query ) {
+		if ( current_theme_supports( AMP_Theme_Support::SLUG ) ) {
+			$availability = AMP_Theme_Support::get_template_availability( $wp_query );
+			$supported    = $availability['supported'];
+		} else {
+			$queried_object = get_queried_object();
+			$supported      = ( $wp_query->is_singular() || $wp_query->is_posts_page ) && $queried_object instanceof WP_Post && post_supports_amp( $queried_object );
```
schlessera: Conditions should always be done with the quickest checks first if there are obvious differences.
```suggestion
			$supported      = $queried_object instanceof WP_Post && ( $wp_query->is_singular() || $wp_query->is_posts_page ) && post_supports_amp( $queried_object );
```

tests/php/test-amp-render-post.php
```diff
@@ -58,8 +58,7 @@ public function test__is_amp_endpoint() {
 		);
 
 		// Set global for WP<5.2 where get_the_content() doesn't take the $post parameter.
```
schlessera: No global is being set here anymore, so the comment doesn't apply anymore either.
```suggestion
```

includes/class-amp-theme-support.php
```diff
@@ -506,24 +494,22 @@ public static function ensure_proper_amp_location( $exit = true ) {
 	 * @since 1.0 Added $exit param.
 	 * @since 1.0 Renamed from redirect_canonical_amp().
```
schlessera, line 496:
```suggestion
	 * @since 1.0 Renamed from redirect_canonical_amp().
	 * @since 1.6 Removed $exit param.
```

includes/class-amp-theme-support.php
```diff
@@ -444,10 +433,9 @@ static function() {
 	 *
 	 * @since 1.0
```
schlessera, line 434:
```suggestion
	 * @since 1.0
	 * @since 1.6 Removed $exit param.
```

includes/amp-helper-functions.php
```diff
@@ -623,34 +646,31 @@ function is_amp_endpoint() {
...
+		}
```
schlessera: Early bails should be preferred over if/else constructs. This is easier to reason about and keeps the level of indentation down.
```suggestion
		if ( current_theme_supports( AMP_Theme_Support::SLUG ) ) {
			$availability = AMP_Theme_Support::get_template_availability( $wp_query );
			return $availability['supported'];
		}

		$queried_object = get_queried_object();
		return ( $wp_query->is_singular() || $wp_query->is_posts_page ) && $queried_object instanceof WP_Post && post_supports_amp( $queried_object );
```

schlessera: I can't provide a suggestion for the rest of this piece of code, as suggestions don't seem to work across deleted lines.
But the change should basically be from this:
```php
if () {
	if () {
		$supported = ...
	else {
		$supported = ...
	}
} else {
	$supported = ...
}
return $supported;
```
...to this:
```php
if () {
	if () {
		return ...;
	}
	return ...;
}
return ...;
```

includes/amp-helper-functions.php
```diff
@@ -566,41 +568,62 @@ function post_supports_amp( $post ) {
...
+	if ( ! did_action( 'parse_request' ) ) {
+		_doing_it_wrong( __FUNCTION__, esc_html( $error_message ), '1.6.0' );
+		$warned = true;
+	} elseif ( defined( 'REST_REQUEST' ) && REST_REQUEST ) {
 		return false;
 	}
 
-	$did_parse_query = did_action( 'parse_query' );
+	// Make sure that the parse_query action has triggered, as this is required to initially populate the global WP_Query.
+	if ( ! $warned && ! ( did_action( 'parse_query' ) || $wp_query instanceof WP_Query ) ) {
```
schlessera: Faster condition first:
```suggestion
	if ( ! $warned && ! ( $wp_query instanceof WP_Query || did_action( 'parse_query' ) ) ) {
```

includes/class-amp-theme-support.php
```diff
@@ -506,24 +494,22 @@ public static function ensure_proper_amp_location( $exit = true ) {
...
 		$non_amp_url = amp_remove_endpoint( $current_url );
 		if ( $non_amp_url === $current_url ) {
 			return false;
 		}
 
-		wp_safe_redirect( $non_amp_url, $status );
-		// @codeCoverageIgnoreStart
-		if ( $exit ) {
+		if ( wp_safe_redirect( $non_amp_url, $status ) ) {
+			// @codeCoverageIgnoreStart
```
schlessera, line 510: Why is code coverage being ignored here? This would warrant an explanatory comment.

schlessera (dismissed):

includes/class-amp-theme-support.php
westonruter: Because `exit` cannot be tested.

includes/amp-helper-functions.php
westonruter: Addressed in 751c645681bc674d3d0640614319fc16e8936b19

westonruter:
@hansschuijff Sorry for increasing the scope of the PR so greatly. Your original change was very needed, and then it identified a whole bunch of other improvements that also needed to be made.

hansschuijff:
@westonruter No worries. I'm glad I could help, even if it was just in a small way. Seeing you guys work is educational too and having done my first pr. Sometimes just initiating something is enough and glad it inspired to enhance amp even more.

westonruter (approved):

Issue #4489, opened 2020-03-31
Bad usage of `tax_query` param deletes a lot more than it should


aidvu:
## Bug Description

Wrong usage of tax_query. Please check: https://developer.wordpress.org/reference/classes/wp_query/#taxonomy-parameters. This ended up deleting all reusable blocks for our customers. A hotfix release would be good.

## Expected Behaviour

The query should only fetch the taxonomy specified. Instead, it completely skips the `tax_query` param because it is used incorrectly.

## Steps to reproduce

1. Install an old AMP version. e.g. 1.4.4
2. Create a few reusable blocks on your site
3. Add them to pages, confirm they work
4. Updated to 1.5
5. Reusable blocks gone...

## Additional context

 - WordPress version: 5.3
 - Plugin version: upgrade to 1.5.0
 - PHP version: 7.*
 - OS: Linux

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

westonruter:
@aidvu Hotfix 1.5.1 is now published on WordPress.org. 

Release notes: https://github.com/ampproject/amp-wp/releases/tag/1.5.1

Thank you _very much_ for reporting and opening a fix! 🙇

aidvu:
> Thank you very much for reporting and opening a fix! bowing_man

Thank you for handling it. Based on the [graphs](https://wordpress.org/plugins/amp/advanced/), it was timely.

westonruter:
Indeed:

![image](https://user-images.githubusercontent.com/134745/78056796-8717c200-733a-11ea-8e66-0869e8be8f54.png)

PR #4614, opened 2020-04-21
Tweak optimizer for consistency

### LINKED ISSUE #4612: Tweak Optimizer interfaces for consistency
While writing the documentation, I'm noticing some minor consistency issues that should be straightened out at the earliest possible convenience.

I'm collecting these here to make a grouped PR for this as a dependency for the docs PR.

- [x] The `TransformationEngine`'s first argument `Configuration $configuration` should be made optional via a poka-yoke, just as the second argument is. This would then fall back to the default out-of-the-box configuration.

- [x] The required order of the arguments when a transformer implements both `Configurable` and `MakesRemoteRequests` should be swapped, so that it matches the order of the `Configuration` and `RemoteGetRequest` arguments in the `TransformationEngine`'s constructor.

- [x] Docblocks in `FilesystemRemoteGetRequest` are copypasta from `StubbedRemoteGetRequest` and need to be adapted.

schlessera:
## Summary

Consistency fixes that are assumed by the new Optimizer documentation:

- [x] The `TransformationEngine`'s first argument `Configuration $configuration` should be made optional via a poka-yoke, just as the second argument is. This would then fall back to the default out-of-the-box configuration.

- [x] The transformation now uses a smart dependency resolution using reflection so that the order of constructor arguments can be arbitrarily swapped or extended.

- [x] The interfaces `Configurable` and `MakesRemoteRequests` are removed in favor of the dependency resolution mentioned above.

- [x] Tests are added to ensure dependency resolution works as expected.

- [x] The static variable to store transformers within the transformation engine was removed in favor of using a property per instance.

- [x] Docblocks in `FilesystemRemoteGetRequest` are copypasta from `StubbedRemoteGetRequest` and need to be adapted.

Fixes #4612 

## Checklist


lib/optimizer/src/TransformationEngine.php
```diff
@@ -31,13 +31,13 @@ final class TransformationEngine
     /**
      * Instantiate a TransformationEngine object.
      *
-     * @param Configuration    $configuration Configuration data to use for setting up the transformers.
+     * @param Configuration    $configuration Optional. Configuration data to use for setting up the transformers.
      * @param RemoteGetRequest $remoteRequest Optional. Transport to use for remote requests. Defaults to the
```
westonruter: Is it useful to indicate `null` can be supplied here?

```suggestion
     * @param Configuration|null    $configuration Optional. Configuration data to use for setting up the transformers.
     * @param RemoteGetRequest|null $remoteRequest Optional. Transport to use for remote requests. Defaults to the
```

lib/optimizer/src/TransformationEngine.php
```diff
@@ -84,14 +84,14 @@ private function getTransformers()
...
-                }
-
                 if (is_a($transformerClass, Configurable::class, true)) {
                     $arguments[] = $this->configuration->getTransformerConfiguration($transformerClass);
                 }
 
+                if (is_a($transformerClass, MakesRemoteRequests::class, true)) {
+                    $arguments[] = $this->remoteRequest;
+                }
+
```
westonruter: Aside: instead of relying on positional arguments, would it be preferable to pass the arguments as an associative array of args so the order does not matter?

westonruter (approved):
LGTM. Left a couple comments.

westonruter (changes requested):
Actually, there are PHPUnit failures for the Optimizer now.

lib/optimizer/src/TransformationEngine.php
schlessera: I remember we had this discussion already. I would really try to stay away from the WordPress' "args pattern", as that makes dependency injection extremely difficult for anyone using it within a modern PHP framework, and makes it very liable to use for everyone else.

Just as an example for why I loathe it, this "args pattern" leads to bugs like this: https://github.com/ampproject/amp-wp/issues/4489

I'll change the code so that it uses reflection to detect the needed arguments instead. This is not something that happens thousands of times per request, so the performance impact of using reflection will be negligible.

schlessera:
I've changed the transformation engine now so that it uses reflection to resolve the correct type and order of the dependencies.

This makes the code more robust now, easier for consumers to build transformers, and straight-forward to extend to many more types of dependencies as needed.

schlessera:
I think this also fixed a latent bug that now appeared through a broken test (see last commit).

westonruter (approved):
A thing of beauty.

Issue #4560, opened 2020-04-10
Allow Reader Mode users to chose a theme from a curated list


amedina:
Upon activating the plugin and/or selecting Reader Mode, the user will be provided with a set of pre-defined themes to chose from (a subset of themes in ecosystem page)

westonruter:
@johnwatkins0 Assigning this to you since this is what you built as part of the wizard (#4702) and also what you'll be adding to the admin settings screen (#4704).

The plumbing of actually loading the selected Reader theme is what I'm working on in #4478.

westonruter:
As noted in #4984, when selecting a non-legacy theme in Reader mode, the same UI to “show all templates as AMP” and the ability to select individual templates needs to be offered. These options should be disabled when the Legacy theme is selected.

PR #4644, opened 2020-05-02
Prototype Reader themes

westonruter:
## Summary

This PR introduces the ability to select from among AMP-compatible themes to be used when in Reader mode, in addition to a “Classic” theme which behaves as the current Reader mode. If not using Classic, then selecting a theme allows for all templates on a site to be used as AMP, including singular, homepage, archives, etc.

* Being able to select a Reader theme is dependent on a site not customizing the AMP endpoint (query var). So if a site changes the query var from `amp` to something else like `lite`, then a warning is displayed instead of the theme list.
* The listed themes exclude the active theme.
* When the active theme is switched to be the same as the theme that is selected for the Reader theme, then the template mode is automatically switched from Reader to Transitional.
* Enable developer tools in Reader mode, along with a “Dev Tools” toggle to the admin settings screen. Note that this is at an intermediary stage, as eventually the Dev Tools toggle should be user-specific.

 ### Customizer

* At the moment, the Customizer for the selected Reader theme can be accessed by clicking “Customize” in the admin bar when viewing an AMP page on the frontend. (There will need to be a Customize button added to the AMP admin settings screen.)
* The AMP panel in the Customizer which is normally shown in Reader mode is omitted entirely when a Reader theme has been selected, since the Customizer is loaded with that Reader theme loaded in its entirety. The AMP toggle appearing next to the device switcher is not shown when a Reader theme is selected; this is only applicable to classic Reader mode. When a Reader theme is selected, the Customizer loaded up-front with _either_ the active theme or the reader theme applied.
* Theme-specific customizations (e.g. color scheme, custom header, background) are only applied to the selected theme, leaving the active theme's settings intact.
* Nav menus can be fully managed in AMP. Nav menu locations are specific to the Reader theme or active theme, but the menus assigned to those locations are shared between the AMP and non-AMP versions of the sites.
* Widgets are disabled in Reader themes; this includes hiding the Widgets panel in the Customizer. Since WordPress lacks the concept of “sidebar locations” as it has with “nav menu locations”, being able to manage widgets in Reader mode would be very complicated. Since widgets are not deemed necessary for Reader mode (especially given the mobile viewport), they are disabled.
* The Themes panel is removed when opening AMP Customizer for a Reader theme. Selecting a different Reader theme must be done currently from the AMP settings screen.
* When the Customizer is opened for a Reader theme, then the previewed device will default to mobile.

Addresses issue #4560 

## Screenshots

When a Reader theme is selected, you can choose from among the installed themes which are AMP-compatible. Currently this includes the core themes:

image

When the Classic theme is selected, templates are not available to be selected and a notice is explained why.

image

Opening the Customizer from an AMP page will hide the themes panel, widgets panel (since widgets are disabled), default to the mobile device, and include a note indicating the AMP version of the site is being customized:

image

## Checklist

- [ ] My pull request is addressing an open issue (please create one otherwise).


src/ReaderThemeLoader.php
```diff
@@ -0,0 +1,157 @@
...
+
+		$get_template   = static function () use ( $theme ) {
+			return $theme->get_template();
+		};
+		$get_stylesheet = static function () use ( $theme ) {
+			return $theme->get_stylesheet();
+		};
+
+		add_filter( 'stylesheet', $get_stylesheet );
+		add_filter( 'template', $get_template );
```
westonruter, line 120:
```suggestion
		add_filter( 'stylesheet', $get_stylesheet );
		add_filter( 'template', $get_template );
		add_filter( 'pre_option_current_theme', static function () use ( $theme ) {
			return $theme->display( 'Name );
		} );
```

westonruter: Also may need to do this which the Customizer also does in `\WP_Customize_Manager::start_previewing_theme()`:

```php
// Handle custom theme roots.
add_filter( 'pre_option_stylesheet_root', array( $this, 'get_stylesheet_root' ) );
add_filter( 'pre_option_template_root', array( $this, 'get_template_root' ) );
```

amedina:
Comments on the flow of switching modes:

1. I was in Reader mode with TwentyNineteen for main site, TwentyTwenty for Reader mode  site.
2. Switched to Transitional, and TwentyTwenty became the theme for both (AMP, non-AMP)
3. Then I switched back to Reader, and TwentyTwenty stayed as the theme for the main Site
4. The Classic was the selected for Reader Mode
5. And the customizer settings I had made in Reader mode (i.e. for the TwentyTwenty theme) were gone

All these points are related to the main observed behavior in point (2).

The expected behavior would be for the plugin to have "memory" related to mode-selection and customizer state.

Update: 

If I manually switch the main site back to the original theme, I m able to select TwentyTwenty again for Reader mode (as expected), and the customizer settings made before the switch are there.

westonruter:
> 5\. And the customizer settings I had made in Reader mode (i.e. for the TwentyTwenty theme) were gone
> If I manually switch the main site back to the original theme, I m able to select TwentyTwenty again for Reader mode (as expected), and the customizer settings made before the switch are there.

Yes, this is expected. The theme customizations are stored with the theme itself. WordPress doesn't share theme customizations across themes, not even between parent and child themes.

westonruter:
Add some indication of the Customizer controls for settings that are shared between the AMP and non-AMP sites (i.e. options) and then for settings which are theme-specific, provide a way to import those specific settings from the active theme.

src/ReaderThemeLoader.php
westonruter: Done in 1991f5bd7.

westonruter:
Closing in favor of #4984.

PR #4650, opened 2020-05-04
Unwrap all embeds with paragraph tags

### LINKED ISSUE #4450: amp-instagram and amp-youtube wrapped in unnecessary p tag
## Bug Description

This is a continuation of #4358 where the same problem applied to `amp-facebook` and was fixed. 

The premise is that these `amp-*` objects should be in the root of the document, and not be wrapped in `<p>` tags. Twitter already worked this way, and in #4358 `amp-facebook` was fixed to match. 

I've now noticed that `amp-instagram` and `amp-youtube` face the same problem:

Screen Shot 2020-03-25 at 7 19 09 PM

Screen Shot 2020-03-25 at 7 21 08 PM

As with the previous ticket, it's important to me because I need to style something that immediately follows the object (`amp-instagram+.translation`), but also generally it just seems like good practice all around. 

Instagram embed code:

```
[embed]https://www.instagram.com/p/Buq8vU6Hcxn/[/embed]
<blockquote class="translation">This is a translation of the Instagram post above.</blockquote>
```

YouTube embed code

```
[embed]https://www.youtube.com/watch?v=vXPJVwwEmiM[/embed]
<blockquote class="translation">This is a translation of the YouTube video above.</blockquote>
```

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

### LINKED ISSUE #4697: Imgur post embed gives 404 error
## Bug Description

The Imgur post embed generates AMP tag, it uses post id as data-imgur-id, and it creates incorrect embed URL: `https://imgur.com/2q6AS6i/embed?pub` so it’s not displayed on the AMP page.

```html
<amp-imgur
    data-imgur-id="2q6AS6i"
    layout="responsive"
    width="540"
    height="633"
  >
  </amp-imgur>

```
In the official AMP documentation, there is no information available for Imgur post embed.
https://amp.dev/documentation/components/amp-imgur/

And there is a merged PR to add the support for Imgur posts:
https://github.com/ampproject/amphtml/pull/23829

## Expected Behaviour

The Imgur post embed should generate `data-imgur-id` as `a/<post_id>`. It generates the correct embed URL `https://imgur.com/a/2q6AS6i/embed?pub`.

```html
<amp-imgur
    data-imgur-id="a/2q6AS6i"
    layout="responsive"
    width="540"
    height="633"
  >
  </amp-imgur>

```

## Steps to reproduce

1. Edit/Create WordPress post.
2. Add an Imgur post URL. eg. https://imgur.com/gallery/2q6AS6i
3. Submit the post.
4. View it on AMP mode

## Screenshots

C3r0URGdKN

## Additional context

 - WordPress version: 5.4
 - Plugin version: 1.5.3
 - AMP plugin template mode: Standad
 - PHP version: 7.4
 - OS: Windows 10
 - Browser: Google Chrome

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

### LINKED ISSUE #4754: Embedding VideoPress video causes validation error
## Bug Description

When embedding a VideoPress video such as https://videopress.com/v/kUJmAcSf in a VideoPress block, the result is a validation error:

image

The fix for this was partially implemented in #3542 because WordPress.tv also uses VideoPress architecture. So it just needs to be extended.

## Expected Behaviour

No validation error should be raised when embedding videos from VideoPress.

## Steps to reproduce

1. Enable Standard mode.
2. Embed a VideoPress block with the URL https://videopress.com/v/kUJmAcSf
3. Save draft
4. See validation error

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

### LINKED ISSUE #4757: Tumblr embeds cause validation errors and fail to embed
## Bug Description

When embedding a Tumblr post like https://teded.tumblr.com/post/184736320764/how-do-vaccines-work

I get a validation error for an invalid script:

image

And the post fails to embed:

image

## Expected Behaviour

The AMP version should look the same as the non-AMP verison.

## Steps to reproduce

Add a Tumblr block for https://teded.tumblr.com/post/184736320764/how-do-vaccines-work

---------------

## Acceptance criteria

* 

## Implementation brief

I think all that is needed is fixing the pattern for the `data-href` attribute:

```diff
- if ( preg_match( '#data-href="(?P<href>https://embed.tumblr.com/embed/post/\w+/\w+)"#', $cache, $matches ) ) {
+ if ( preg_match( '#data-href="(?P<href>https://embed.tumblr.com/embed/post/[^"]+?)"#', $cache, $matches ) ) {
```

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

There are inconsistencies with some embeds being wrapped with a `<p>` tag, and some not. We should strive towards having a uniform markup across all embeds.

Fixes #4450.
Fixes #4697.
Fixes #4754.
Fixes #4757.

## Embeds to be unwrapped:
- [x] crowdsignal
- [x] dailymotion
- [x] facebook
- [x] gfycat
- [x] hulu
- [x] imgur
- [x] instagram
- [x] issuu
- [x] meetup
- [x] pinterest
- [x] reddit
- [x] scribd
- [x] soundcloud
- [x] tiktok
- [x] tumblr
- [x] twitter
- [x] vimeo
- [x] vine
- [x] wordpress-tv
- [x] youtube

## Checklist


includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -107,4 +107,21 @@ function ( $attr_name ) {
...
+	 * This usually happens while `wpautop()` processes the element.
+	 *
+	 * @param DOMElement $node Node.
+	 */
+	protected function maybe_unwrap_p_element( DOMElement $node ) {
+		$parent_element = AMP_DOM_Utils::get_parent_element( $node );
+
+		if ( $parent_element && 'p' === $parent_element->nodeName && false === $parent_element->hasAttributes() ) {
+			$children = $parent_element->getElementsByTagName( '*' );
+			if ( 1 === $children->length ) {
```
westonruter: This won't return the expected NodeList if the child node itself has children (grandchildren). Therefore, maybe something like this is needed given that PHP's `dom` extension doesn't support [`childElementCount`](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/childElementCount):

```suggestion
			$child_element_count = array_sum(
				array_map(
					function ( DOMNode $child ) {
						return $child instanceof DOMElement ? 1 : 0;
					},
					$node->childNodes
				)
			);
			if ( 1 === $child_element_count ) {
```

includes/embeds/class-amp-dailymotion-embed-handler.php
```diff
@@ -48,86 +49,69 @@ public function __construct( $args = [] ) {
...
 	 */
-	public function oembed( $matches, $attr, $url, $rawattr ) {
-		$video_id = $this->get_video_id_from_url( $url );
-		return $this->render(
-			[
-				'video_id' => $video_id,
-			]
-		);
+	public function sanitize_raw_embeds( Document $dom ) {
+		$nodes = $dom->xpath->query( '//iframe[ starts-with( @src, "https://www.dailymotion.com/embed/video/" ) ]' );
```
westonruter: I recommend extracting this URL to be a class constant.

includes/embeds/class-amp-dailymotion-embed-handler.php
```diff
@@ -48,86 +49,69 @@ public function __construct( $args = [] ) {
...
 	 */
-	private function get_video_id_from_url( $url ) {
-		$parsed_url = wp_parse_url( $url );
-		parse_str( $parsed_url['path'], $path );
-		$tok = explode( '/', $parsed_url['path'] );
-		$tok = explode( '_', $tok[2] );
+	private function sanitize_raw_embed( DOMElement $iframe_node ) {
+		$iframe_src = $iframe_node->getAttribute( 'src' );
+
+		if ( preg_match( '#video/(?P<video_id>[a-zA-Z0-9]+)#', $iframe_src, $matches ) ) {
```
westonruter: Use the above class constant here to remove the URL prefix. For example:

```php
$video_id = strtok( substr( $iframe_src, strlen( self::BASE_URL ) ), '/?#' );
```

includes/embeds/class-amp-dailymotion-embed-handler.php
```diff
@@ -48,86 +49,69 @@ public function __construct( $args = [] ) {
...
+			$video_id = $matches['video_id'];
+
+			$amp_node = AMP_DOM_Utils::create_node(
+				Document::fromNode( $iframe_node ),
+				'amp-dailymotion',
+				[
+					'data-videoid' => $video_id,
+					'layout'       => 'responsive',
+					'width'        => $this->args['width'],
+					'height'       => $this->args['height'],
```
westonruter: Can the `width` and `height` from the `$iframe_node` not be used?

includes/embeds/class-amp-gfycat-embed-handler.php
```diff
@@ -21,61 +23,80 @@ class AMP_Gfycat_Embed_Handler extends AMP_Base_Embed_Handler {
...
-			$pieces = explode( '/detail/', $parsed_url['path'] );
-			if ( ! isset( $pieces[1] ) ) {
-				if ( ! preg_match( '/\/([A-Za-z0-9]+)/', $parsed_url['path'], $matches ) ) {
-					return $return;
-				}
-				$attributes['data-gfyid'] = $matches[1];
-			} else {
-				$attributes['data-gfyid'] = $pieces[1];
-			}
+		if ( ! preg_match( '#/ifr/([A-Za-z0-9]+)#', $iframe_src, $matches ) ) {
```
westonruter: See note above in DailyMotion for using a constant here with `strtok()`.

includes/embeds/class-amp-gfycat-embed-handler.php
```diff
@@ -21,61 +23,80 @@ class AMP_Gfycat_Embed_Handler extends AMP_Base_Embed_Handler {
...
-		$parsed_url = wp_parse_url( $url );
-		if ( false !== strpos( $parsed_url['host'], 'gfycat.com' ) ) {
-			if ( preg_match( '/width=["\']?(\d+)/', $return, $matches ) ) {
-				$attr['width'] = $matches[1];
-			}
-			if ( preg_match( '/height=["\']?(\d+)/', $return, $matches ) ) {
-				$attr['height'] = $matches[1];
-			}
+	public function sanitize_raw_embeds( Document $dom ) {
+		$nodes = $dom->xpath->query( '//iframe[ starts-with( @src, "https://gfycat.com/ifr/" ) ]' );
```
westonruter: Put the URL in a constant?

includes/embeds/class-amp-gfycat-embed-handler.php
```diff
@@ -21,61 +23,80 @@ class AMP_Gfycat_Embed_Handler extends AMP_Base_Embed_Handler {
...
+		if ( ! preg_match( '#/ifr/([A-Za-z0-9]+)#', $iframe_src, $matches ) ) {
+			// Nothing to do if video ID could not be found.
+			return;
+		}
+
+		$new_attributes = [
+			'data-gfyid' => $matches[1],
+			'layout'     => 'responsive',
+			'height'     => $iframe_node->getAttribute( 'height' ),
+			'width'      => $iframe_node->getAttribute( 'width' ),
```
westonruter: As with DailyMotion, can the `width` and `height` from the `$iframe_node` be used?

includes/embeds/class-amp-hulu-embed-handler.php
```diff
@@ -28,60 +30,92 @@ class AMP_Hulu_Embed_Handler extends AMP_Base_Embed_Handler {
...
-		$parsed_url = wp_parse_url( $url );
-		if ( false !== strpos( $parsed_url['host'], 'hulu.com' ) ) {
-			if ( preg_match( '/width=["\']?(\d+)/', $return, $matches ) ) {
-				$attr['width'] = $matches[1];
-			}
-			if ( preg_match( '/height=["\']?(\d+)/', $return, $matches ) ) {
-				$attr['height'] = $matches[1];
-			}
+	public function sanitize_raw_embeds( Document $dom ) {
+		$nodes = $dom->xpath->query( '//iframe[ contains( @src, "www.hulu.com/embed.html" ) ]' );
```
westonruter: See note above about using constant.

includes/embeds/class-amp-hulu-embed-handler.php
```diff
@@ -28,60 +30,92 @@ class AMP_Hulu_Embed_Handler extends AMP_Base_Embed_Handler {
...
+	}
+
+	/**
+	 * Get video from URL.
+	 *
+	 * @param string $url Video URL.
+	 * @return string|null Video ID, or null if it was not found.
+	 */
+	private function get_video_id( $url ) {
+		if ( preg_match( '#hulu\.com/embed\.html\?eid=(?P<id>[A-Za-z0-9_-]+)#', $url, $matches ) ) {
```
westonruter: Instead of using a regex here, this can use `wp_parse_url( $url, PHP_URL_QUERY )`  and then pass the result into `parse_str()` to then obtain the `eid` parameter. This ensure the parameter can be extracted even if it is not the first in the URL's query params.  There is no need to examine the host or path because those are checked above.

includes/embeds/class-amp-scribd-embed-handler.php
```diff
@@ -15,63 +17,66 @@ class AMP_Scribd_Embed_Handler extends AMP_Base_Embed_Handler {
...
-	 * @param string $cache Cache for oEmbed.
-	 * @param string $url   Embed URL.
-	 * @return string Embed.
+	 * @param Document $dom DOM.
 	 */
-	public function filter_embed_oembed_html( $cache, $url ) {
-		if ( ! in_array( wp_parse_url( $url, PHP_URL_HOST ), [ 'scribd.com', 'www.scribd.com' ], true ) ) {
-			return $cache;
+	public function sanitize_raw_embeds( Document $dom ) {
+		$nodes = $dom->xpath->query( '//iframe[ starts-with( @src, "https://www.scribd.com/embeds/" ) ]' );
```
westonruter: Pre above, let's put the URL in a constant.

includes/embeds/class-amp-scribd-embed-handler.php
```diff
@@ -15,63 +17,66 @@ class AMP_Scribd_Embed_Handler extends AMP_Base_Embed_Handler {
...
-	private function sanitize_iframe( $html ) {
-		return preg_replace_callback(
-			'#^.*<iframe(?P<iframe_attributes>[^>]+?)></iframe>.*$#s',
-			function ( $matches ) {
-				$attrs = $matches['iframe_attributes'];
+	private function sanitize_raw_embed( DOMElement $iframe_node ) {
+		$required_sandbox_permissions = 'allow-popups allow-scripts';
+		$iframe_node->setAttribute(
+			'sandbox',
+			sprintf( '%s %s', $iframe_node->getAttribute( 'sandbox' ), $required_sandbox_permissions )
```
westonruter: Seems simpler:

```suggestion
			$iframe_node->getAttribute( 'sandbox' ) . ' ' . $required_sandbox_permissions
```

includes/embeds/class-amp-soundcloud-embed-handler.php
```diff
@@ -23,163 +25,101 @@ class AMP_SoundCloud_Embed_Handler extends AMP_Base_Embed_Handler {
...
 	 */
-	private function parse_amp_component_from_iframe( $html, $url = null ) {
-		$props = $this->match_element_attributes( $html, 'iframe', [ 'src', 'title', 'width', 'height' ] );
-		if ( ! isset( $props ) || empty( $props['src'] ) ) {
-			return $html;
-		}
-
-		$src   = html_entity_decode( $props['src'], ENT_QUOTES );
+	private function sanitize_raw_embed( DOMElement $iframe_node ) {
+		$src   = html_entity_decode( $iframe_node->getAttribute( 'src' ), ENT_QUOTES );
```
westonruter: The decoding of entities shouldn't be needed here, as this will be done by the HTML parser.

includes/embeds/class-amp-soundcloud-embed-handler.php
```diff
@@ -23,163 +25,101 @@ class AMP_SoundCloud_Embed_Handler extends AMP_Base_Embed_Handler {
...
+	 * Get embed ID from Soundcloud iframe src.
 	 *
 	 * @param string $url URL.
-	 * @return array Params extracted from URL.
+	 * @return array|null Array containing ID name and value if found.
 	 */
-	private function extract_params_from_iframe_src( $url ) {
+	private function parse_embed_id_from_url( $url ) {
 		$parsed_url = wp_parse_url( $url );
 		if ( preg_match( '#tracks/(?P<track_id>\d+)#', $parsed_url['path'], $matches ) ) {
```
westonruter: it seems the matching could be simplified to be something like:

```php
if ( ! preg_match( '#^/player/(?P<type>tracks|playlists)/(?P<id>\d+)'#, $parsed_url['path'], $matches ) ) {
    return [];
}
if ( 'tracks' === $matches['type'] ) {
    return [ 'track_id' => $matches['id'] ];
} elseif ( 'playlists' === $matches['type] ) {
    return [ 'playlist_id' => $matches['id'] ];
}
```

includes/embeds/class-amp-twitter-embed-handler.php
```diff
@@ -223,13 +225,54 @@ private function get_tweet_id( $node ) {
...
+
+		// Remove any <br> siblings.
+		while ( $next_element_sibling && $next_element_sibling instanceof DOMElement && 'br' === $next_element_sibling->nodeName ) {
+			$next_element_sibling->parentNode->removeChild( $next_element_sibling );
+			$next_element_sibling = $node->nextSibling;
+		}
+
+		if ( $next_element_sibling instanceof DOMElement && 'br' === $next_element_sibling->nodeName ) {
+			$next_element_sibling->parentNode->removeChild( $next_element_sibling );
+		}
```
westonruter: Is the `if` statement redundant? Is it not handled in the `while` loop above?

includes/embeds/class-amp-vimeo-embed-handler.php
```diff
@@ -59,75 +54,69 @@ public function __construct( $args = [] ) {
...
-	 * @param array  $attr    Shortcode attribues.
-	 * @param string $url     URL.
-	 * @param string $rawattr Unmodified shortcode attributes.
-	 * @return string Rendered oEmbed.
+	 * @param Document $dom DOM.
 	 */
-	public function oembed( $matches, $attr, $url, $rawattr ) {
-		$video_id = $this->get_video_id_from_url( $url );
+	public function sanitize_raw_embeds( Document $dom ) {
+		$nodes = $dom->xpath->query( '//iframe[ starts-with( @src, "https://player.vimeo.com/video/" ) ]' );
```
westonruter: Let's use a constant for this.

includes/embeds/class-amp-vimeo-embed-handler.php
```diff
@@ -137,42 +126,12 @@ public function render( $args ) {
 	 * @return int Video ID.
 	 */
 	private function get_video_id_from_url( $url ) {
-		$path = wp_parse_url( $url, PHP_URL_PATH );
-
 		// @todo This will not get the private key for unlisted videos (which look like https://vimeo.com/123456789/abcdef0123), but amp-vimeo doesn't support them currently anyway.
-		$video_id = '';
-		if ( $path && preg_match( ':/(\d+):', $path, $matches ) ) {
+		$video_id = null;
+		if ( preg_match( ':/video/(\d+):', $url, $matches ) ) {
```
westonruter: See note above about using constant and `strtok()`. It will be more robust than looking _anywhere_ in the URL for `/video/...` as it could erroneously match in a query parameter, for example.

includes/embeds/class-amp-wordpress-tv-embed-handler.php
```diff
@@ -24,14 +26,56 @@ class AMP_WordPress_TV_Embed_Handler extends AMP_Base_Embed_Handler {
...
+
+	/**
+	 * Make WordPress TV embed AMP compatible.
+	 *
+	 * @param DOMElement $iframe_node The node to make AMP compatible.
+	 */
+	private function sanitize_raw_embed( DOMElement $iframe_node ) {
+		$next_sibling = $iframe_node->nextSibling;
+
+		if ( $next_sibling && 'script' === $next_sibling->nodeName ) {
```
westonruter: Should this verify the script is the expected one? Namely, one with the `src` of `https://videopress.com/videopress-iframe.js`?

westonruter: Come to mention it, what does this script do? It appears to be handling resizing. Is resizing something that we're missing? Or is resizing not needed in an AMP context? If it is, we should contact WordPress.com about emitting the AMP resizing messages.

includes/embeds/class-amp-youtube-embed-handler.php
```diff
@@ -63,116 +55,123 @@ public function __construct( $args = [] ) {
...
-	private function parse_props( $html, $url, $video_id ) {
-		$props = $this->match_element_attributes( $html, 'iframe', [ 'title', 'height', 'width' ] );
-		if ( ! isset( $props ) ) {
-			return null;
+	private function sanitize_raw_embed( DOMElement $iframe_node, $video_id ) {
+		$attributes = [
+			'data-videoid' => $video_id,
+			'layout'       => 'responsive',
+			'width'        => $this->args['width'],
+			'height'       => $this->args['height'],
```
westonruter, line 98: Note the width/height are overridden below if defined on the iframe.

includes/utils/class-amp-dom-utils.php
```diff
@@ -499,4 +499,19 @@ public static function copy_attributes( $attributes, DOMElement $from, DOMElemen
...
 		}
 	}
+
+	/**
+	 * Get closest parent element for specified DOM node.
+	 *
+	 * @param DOMNode $node DOMNode.
+	 * @return DOMElement|null
+	 */
+	public static function get_parent_element( DOMNode $node ) {
```
westonruter: Is this needed? It's only being used to get the parent node of `DOMElement` instances, which can only have other `DOMElement` instances as parents, except in the case of the root element, right? (It won't ever be a `DocumentFragment` in our usage.) So you could then just check if `$node->parentNode` instance of `DOMElement` as opposed to checking if truthy.

includes/embeds/class-amp-scribd-embed-handler.php
```diff
@@ -15,63 +17,66 @@ class AMP_Scribd_Embed_Handler extends AMP_Base_Embed_Handler {
...
-					1,
-					$replaced
-				);
+		// Remove the accompanied script tag so that the iframe can be later unwrapped.
+		if ( 'script' === $iframe_node->nextSibling->nodeName ) {
+			$parent_element = AMP_DOM_Utils::get_parent_element( $iframe_node );
+			if ( $parent_element ) {
+				$parent_element->removeChild( $iframe_node->nextSibling );
+			}
+		}
```
westonruter:
```suggestion
		if ( 'script' === $iframe_node->nextSibling->nodeName && $iframe_node->parentNode instanceof DOMElement ) {
			$iframe_node->parentNode->removeChild( $iframe_node->nextSibling );
		}
```

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -107,4 +107,21 @@ function ( $attr_name ) {
...
+	 * Replace the node's parent with itself if the parent is a <p> tag, has no attributes and has no other children.
+	 * This usually happens while `wpautop()` processes the element.
+	 *
+	 * @param DOMElement $node Node.
+	 */
+	protected function maybe_unwrap_p_element( DOMElement $node ) {
+		$parent_element = AMP_DOM_Utils::get_parent_element( $node );
+
+		if ( $parent_element && 'p' === $parent_element->nodeName && false === $parent_element->hasAttributes() ) {
+			$children = $parent_element->getElementsByTagName( '*' );
```
westonruter:
```suggestion
		if ( $node->parentNode instanceof DOMElement && 'p' === $node->parentNode->nodeName && false === $node->parentNode->hasAttributes() ) {
			$children = $node->parentNode->getElementsByTagName( '*' );
```

westonruter (changes requested):
Sorry for the delay in reviewing!

includes/embeds/class-amp-dailymotion-embed-handler.php
```diff
@@ -48,86 +49,69 @@ public function __construct( $args = [] ) {
...
-			[
-				'data-videoid' => $args['video_id'],
-				'layout'       => 'responsive',
-				'width'        => $this->args['width'],
-				'height'       => $this->args['height'],
-			]
-		);
+	protected function is_raw_embed( DOMElement $node ) {
+		return $node->parentNode && 'amp-dailymotion' !== $node->parentNode->nodeName;
 	}
```
schlessera: I'd extract the node name string into a constant or method and move the `is_raw_embed()` method into the base class.

includes/embeds/class-amp-dailymotion-embed-handler.php
```diff
@@ -48,86 +49,69 @@ public function __construct( $args = [] ) {
...
+	 * @param Document $dom DOM.
 	 */
-	public function oembed( $matches, $attr, $url, $rawattr ) {
-		$video_id = $this->get_video_id_from_url( $url );
-		return $this->render(
-			[
-				'video_id' => $video_id,
-			]
-		);
+	public function sanitize_raw_embeds( Document $dom ) {
```
schlessera: I'd extract the Xpath query into a constant or method and put the `sanitizer_raw_embeds()` method into the base class.

includes/embeds/class-amp-dailymotion-embed-handler.php
```diff
@@ -48,86 +49,69 @@ public function __construct( $args = [] ) {
...
-	 * @param string $url URL.
-	 * @return integer Video ID.
+	 * @param DOMElement $iframe_node The node to make AMP compatible.
 	 */
-	private function get_video_id_from_url( $url ) {
-		$parsed_url = wp_parse_url( $url );
-		parse_str( $parsed_url['path'], $path );
-		$tok = explode( '/', $parsed_url['path'] );
-		$tok = explode( '_', $tok[2] );
+	private function sanitize_raw_embed( DOMElement $iframe_node ) {
```
schlessera: I'd add an abstract or empty `sanitize_raw_embed()` method into the base class so that `sanitize_raw_embeds()` can be safely put into the base class as well.

This should then be overridden/implemented on a per embed basis as needed.

schlessera (changes requested):
Before you going, I suggest refactoring the code to put more recurring logic into the base embed handler.

Basically, what this does is provide more of ["Template Method Pattern"](https://sourcemaking.com/design_patterns/template_method) approach to dealing with embeds, where the base embed handler lays down a framework of how to deal with a typical embed. The individual embeds can still override any step that doesn't directly fit within the general framework, but overall, there'll be less code, needing less testing, as a lot of them will be perfectly happy with this default framework if it is well chosen.

So, if you look at my comments, I'd suggest using `sanitize_raw_embeds()` as one of the steps in this template, with its default implementation (using `is_raw_embed()` and `sanitize_raw_embed()`, but minus the hard-coded strings) just being implemented once within the base embed handler.

includes/embeds/class-amp-instagram-embed-handler.php
```diff
@@ -190,6 +117,22 @@ private function create_amp_instagram_and_replace_node( $dom, $node ) {
 		$this->did_convert_elements = true;
```
westonruter: As uncovered in https://github.com/rtCamp/themes/pull/3/files#r423249922, it may make sense in this `create_amp_instagram_and_replace_node()` method to preserve  some inline styles from the `$node` to copy onto `$new_node`.  In particular, `border-radius: 3px; border: 1px solid rgb(219, 219, 219);`. 

Nevertheless, this is is probably not something we should be concerned with. In fact, it's probably actually an AMP issue where `amp-instagram` should be applying those styles. So nevermind!

pierlon:
Still a few more embeds outstanding, but I believe the PR is ready for another round of review.

pierlon:
Marking this as ready for review although one of the tests is failing.

includes/class-amp-theme-support.php
```diff
@@ -1179,8 +1180,9 @@ public static function register_content_embed_handlers() {
 				);
 				continue;
 			}
-
-			$embed_handler->register_embed();
+			if ( $embed_handler instanceof Registerable ) {
```
westonruter, line 1183: Will this have a backwards-incompatibility problem? If someone has existing subclasses of `AMP_Base_Embed_Handler` with `register_embed` defined, should it not do something like this:

```suggestion
			if ( $embed_handler instanceof Registerable || method_exists( $embed_handler, 'register_embed' ) ) {
```

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -40,14 +42,18 @@ abstract class AMP_Base_Embed_Handler {
 	protected $did_convert_elements = false;
```
westonruter, line 42: This property can be marked as `@deprecated` and any usages in the subclasses can be eliminated (all of which are write-only currently), as it is not serving any purpose anymore.

In the past it was used to determine whether or not the AMP component script needed to be added to the page. But this is obsolete.

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -40,14 +42,18 @@ abstract class AMP_Base_Embed_Handler {
...
+	protected $amp_tag = 'amp-iframe';
 
 	/**
-	 * Unregisters embed.
+	 * Base URL used for identifying embeds.
+	 *
+	 * @var string
 	 */
-	abstract public function unregister_embed();
+	protected $base_embed_url = '';
```
westonruter, line 56: I don't see this property as being used anywhere. Can it be removed?

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -64,6 +70,56 @@ public function __construct( $args = [] ) {
...
+		return $node->parentNode && $this->amp_tag !== $node->parentNode->nodeName;
+	}
+
+	/**
+	 * Get all raw embeds from the DOM.
+	 *
+	 * @param Document $dom Document.
+	 * @return DOMNodeList|null A list of DOMElement nodes, or null if not implemented.
+	 */
+	abstract protected function get_raw_embed_nodes( Document $dom );
```
westonruter, line 114: If not implemented? Doesn't an `abstract` method always have to be implemented by a subclass? Shouldn't this method just be changed to be non-`abstract` and then return `null` here? Then only the subclassed methods would return `DOMNodeList`. This would avoid needing to define the method to return `null` as it stands today for Twitter, Playlist, Gallery, and Blocks.

includes/embeds/class-amp-core-block-handler.php
```diff
@@ -355,4 +356,22 @@ private function process_text_widgets( Document $dom ) {
...
+
+	/**
+	 * Get all raw embeds from the DOM.
+	 *
+	 * @param Document $dom Document.
+	 * @return DOMNodeList|null A list of DOMElement nodes, or null if not implemented.
+	 */
+	protected function get_raw_embed_nodes( Document $dom ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+		return null;
+	}
```
westonruter, line 368: I just making this the method in the base class.

includes/embeds/class-amp-gallery-embed-handler.php
```diff
@@ -285,4 +286,22 @@ public function print_styles() {
...
+
+	/**
+	 * Get all raw embeds from the DOM.
+	 *
+	 * @param Document $dom Document.
+	 * @return DOMNodeList|null A list of DOMElement nodes, or null if not implemented.
+	 */
+	protected function get_raw_embed_nodes( Document $dom ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+		return null;
+	}
```
westonruter, line 298: See comments above.

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -64,6 +70,56 @@ public function __construct( $args = [] ) {
...
+	}
+
+	/**
+	 * Determine if the node is indeed a raw embed.
+	 *
+	 * @param DOMElement $node DOM element.
+	 * @return bool True if it is a raw embed, false otherwise.
+	 */
+	protected function is_raw_embed( DOMElement $node ) {
+		return $node->parentNode && $this->amp_tag !== $node->parentNode->nodeName;
```
westonruter, line 105: To facilitate using this to loop over a list of arbitrary nodes (e.g. `childNodes`) probably safer to accept a `DOMNode` instead:

```suggestion
	 * @param DOMNode $node DOM node.
	 * @return bool True if it is a raw embed, false otherwise.
	 */
	protected function is_raw_embed( DOMNode $node ) {
		return $node instanceof DOMElement && $node->parentNode && $this->amp_tag !== $node->parentNode->nodeName;
```

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -107,4 +163,90 @@ function ( $attr_name ) {
...
+			) {
+				$element_sibling->parentNode->removeChild( $element_sibling );
+				return;
+			}
+		}
+
+		// Handle case where script is immediately following.
+		$is_embed_script = (
+			$element_sibling instanceof DOMElement &&
+			'script' === strtolower( $element_sibling->nodeName ) &&
```
westonruter, line 242: Is `strtolower()` needed here? It's not used above:

```suggestion
			'script' === $element_sibling->nodeName &&
```

I found that in all PHP versions the value is normalized to lower case already: https://3v4l.org/oDCO6

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -107,4 +163,90 @@ function ( $attr_name ) {
...
+		$sibling_location = $is_next_to ? 'nextSibling' : 'previousSibling';
+		$element_sibling  = $node->{$sibling_location};
+
+		while ( $element_sibling && ! ( $element_sibling instanceof DOMElement ) ) {
+			$element_sibling = $element_sibling->{$sibling_location};
+		}
+
+		// Handle case where script is wrapped in paragraph by wpautop.
+		if ( $element_sibling instanceof DOMElement && 'p' === $element_sibling->nodeName ) {
+			$children_elements = array_values( $this->get_child_elements( $element_sibling ) );
```
westonruter, line 224: Using `array_values()` isn't needed because `get_child_elements` already uses `iterator_to_array()`:

```suggestion
			$children_elements = $this->get_child_elements( $element_sibling );
```

includes/embeds/class-amp-core-block-handler.php
```diff
@@ -355,4 +356,22 @@ private function process_text_widgets( Document $dom ) {
...
+
+	/**
+	 * Make embed AMP compatible.
+	 *
+	 * @param DOMElement $node DOM element.
+	 */
+	protected function sanitize_raw_embed( DOMElement $node ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+	}
```
westonruter, line 376: Maybe this should be in the base class as well.

includes/templates/class-amp-content.php
```diff
@@ -184,7 +186,9 @@ private function register_embed_handlers( $embed_handler_classes ) {
 				continue;
 			}
 
-			$embed_handler->register_embed();
+			if ( $embed_handler instanceof Registerable ) {
```
westonruter, line 189: Needed for backwards-compatibility:

```suggestion
			if ( $embed_handler instanceof Registerable || method_exists( $embed_handler, 'register_embed' ) ) {
```

includes/templates/class-amp-content.php
```diff
@@ -198,8 +202,9 @@ private function register_embed_handlers( $embed_handler_classes ) {
 	 */
 	private function unregister_embed_handlers( $embed_handlers ) {
 		foreach ( $embed_handlers as $embed_handler ) {
-			$this->add_scripts( $embed_handler->get_scripts() ); // @todo Why add_scripts here? Shouldn't it be array_diff()?
-			$embed_handler->unregister_embed();
+			if ( $embed_handler instanceof Registerable ) {
```
westonruter, line 205: Needed for backwards-compatibility:

```suggestion
			if ( $embed_handler instanceof Registerable || method_exists( $embed_handler, 'unregister_embed' ) ) {
```

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -107,4 +163,90 @@ function ( $attr_name ) {
...
+	 * Removes the node's nearest <script> sibling with a `src` attribute containing the base `src` URL provided.
+	 *
+	 * @since 1.6
+	 *
+	 * @param DOMElement $node         The DOMNode to whose sibling is the script to be removed.
+	 * @param string     $base_src_url Script URL to match against.
+	 * @param string     $content      Text content of node to match against.
+	 * @param bool       $is_next_to   Whether the script sibling is next or preceding the specified element.
+	 */
+	protected function remove_script_sibling( DOMElement $node, $base_src_url, $content = '', $is_next_to = true ) {
```
westonruter, line 214: In seeing the usages of `remove_script_sibling` in the subclasses, it's somewhat hard to read in regards to the `$content` and `$is_next_to` parameters. Can helper methods be used to make it more intuitive? Like `remove_inline_script_sibling()` when seeking to pass `$content` for example.

includes/embeds/class-amp-base-embed-handler.php
```diff
@@ -107,4 +163,90 @@ function ( $attr_name ) {
...
+		}
+	}
+
+	/**
+	 * Removes the node's nearest <script> sibling with a `src` attribute containing the base `src` URL provided.
+	 *
+	 * @since 1.6
+	 *
+	 * @param DOMElement $node         The DOMNode to whose sibling is the script to be removed.
+	 * @param string     $base_src_url Script URL to match against.
```
westonruter, line 210: The `$base_src_url` is not strictly checked as a base. It is getting passed without any protocol. While maybe this is overkill, perhaps a full URL should be passed as `$base_src_url` and then the initial `/https?:(?=\/\/)/` can be stripped from it and the `src` being checked. Then it could look for `0 === strpos(...)` as opposed to `false !== strpos(...)`. That would be more robust.

includes/embeds/class-amp-crowdsignal-embed-handler.php
```diff
@@ -8,59 +8,128 @@
...
-			}
-			$output = sprintf( '<a href="%s" target="_blank">%s</a>', esc_url( $url ), esc_html( $name ) );
+	/**
+	 * Sanitize poll embed.
+	 *
+	 * @param DOMElement $node Poll embed.
+	 */
+	private function sanitize_poll_embed( DOMElement $node ) {
+		// Replace the `noscript` parent element with the iframe.
+		$node->parentNode->parentNode->replaceChild( $node, $node->parentNode );
```
westonruter, line 89: Should this confirm that the `$node->parentNode->parentNode` is actually a `noscript`? This could also be done done at the XPath level.

includes/embeds/class-amp-crowdsignal-embed-handler.php
```diff
@@ -8,59 +8,128 @@
...
 	 */
-	public function filter_embed_oembed_html( $cache, $url, $attr ) {
-		$parsed_url = wp_parse_url( $url );
-		if ( empty( $parsed_url['host'] ) || empty( $parsed_url['path'] ) || ! preg_match( '#(^|\.)(?P<host>polldaddy\.com|crowdsignal\.com|survey\.fm|poll\.fm)#', $parsed_url['host'], $matches ) ) {
-			return $cache;
-		}
+	protected function get_raw_embed_nodes( Document $dom ) {
+		$queries = [
+			// For poll embeds.
+			'//iframe[ @class="cs-iframe-embed" and starts-with( @src, "https://poll.fm/" ) ]',
```
westonruter, line 59:
```suggestion
			'//noscript/iframe[ @class="cs-iframe-embed" and starts-with( @src, "https://poll.fm/" ) ]',
```

includes/embeds/class-amp-crowdsignal-embed-handler.php
```diff
@@ -8,59 +8,128 @@
...
+	/**
+	 * Sanitize survey embed.
+	 *
+	 * @param DOMElement $node Survey embed.
+	 */
+	private function sanitize_survey_embed( DOMElement $node ) {
+		$settings = json_decode( $node->getAttribute( 'data-settings' ), false );
+
+		// We can't form the iframe URL without a domain and survey ID.
+		if ( ! ( property_exists( $settings, 'domain' ) || property_exists( $settings, 'id' ) ) ) {
```
westonruter, line 104:
```suggestion
		if ( ! ( is_object( $settings ) && ( property_exists( $settings, 'domain' ) && property_exists( $settings, 'id' ) ) ) {
```

includes/embeds/class-amp-crowdsignal-embed-handler.php
```diff
@@ -8,59 +8,128 @@
...
+			return;
 		}
 
-		return $output;
+		// Logic for building the iframe `src` can be found in https://polldaddy.com/survey.js.
+		$iframe_src = sprintf(
+			'https://%s/%s?%s',
+			$settings->domain,
+			$settings->id,
+			property_exists( $settings, 'auto' ) && $settings->auto ? 'ft=1&iframe=' . amp_get_current_url() : 'iframe=1'
```
westonruter, line 113:
```suggestion
			( property_exists( $settings, 'auto' ) && $settings->auto ) ? 'ft=1&iframe=' . amp_get_current_url() : 'iframe=1'
```

includes/embeds/class-amp-gallery-embed-handler.php
```diff
@@ -285,4 +286,22 @@ public function print_styles() {
...
+
+	/**
+	 * Make embed AMP compatible.
+	 *
+	 * @param DOMElement $node DOM element.
+	 */
+	protected function sanitize_raw_embed( DOMElement $node ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+	}
```
westonruter, line 306: See comment in base method.

includes/embeds/class-amp-pinterest-embed-handler.php
```diff
@@ -87,4 +95,23 @@ public function render( $args ) {
...
 	}
+
+	/**
+	 * Get all raw embeds from the DOM.
+	 *
+	 * @param Document $dom Document.
+	 * @return DOMNodeList|null A list of DOMElement nodes, or null if not implemented.
+	 */
+	protected function get_raw_embed_nodes( Document $dom ) {
+		return $dom->xpath->query( sprintf( '//%s', $this->amp_tag ) );
```
westonruter, line 106: Why would a _raw_ embed be already an `amp-pinterest`?

includes/embeds/class-amp-playlist-embed-handler.php
```diff
@@ -330,4 +333,22 @@ public function get_title( $track ) {
...
+		return null;
+	}
+
+	/**
+	 * Make embed AMP compatible.
+	 *
+	 * @param DOMElement $node DOM element.
+	 */
+	protected function sanitize_raw_embed( DOMElement $node ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+	}
```
westonruter, line 353: See comments in base class about moving these methods there.

includes/embeds/class-amp-reddit-embed-handler.php
```diff
@@ -6,70 +6,115 @@
...
-		$args = wp_parse_args(
-			$args,
-			[
-				'url' => false,
-			]
-		);
-
-		if ( empty( $args['url'] ) ) {
-			return '';
+	private function get_first_child_element( DOMElement $node ) {
```
westonruter, line 111: This seems useful to be in the base class as a `protected` method. It could be called `get_first_element_child()` to reflect the DOM: https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/firstElementChild

includes/embeds/class-amp-tumblr-embed-handler.php
```diff
@@ -6,54 +6,89 @@
...
+			'div',
+			[
+				'overflow'   => '',
+				'tabindex'   => 0,
+				'role'       => 'button',
+				'aria-label' => esc_attr__( 'See more', 'amp' ),
+			]
+		);
+
+		$overflow_node->textContent = esc_html__( 'See more', 'amp' );
```
westonruter, line 79: No need to escape when manipulating the DOM:

```suggestion
		$overflow_node->textContent = __( 'See more', 'amp' );
```

includes/embeds/class-amp-tumblr-embed-handler.php
```diff
@@ -6,54 +6,89 @@
...
-			);
+		// Add an overflow node to allow the amp-iframe to resize.
+		$overflow_node = AMP_DOM_Utils::create_node(
+			$dom,
+			'div',
+			[
+				'overflow'   => '',
+				'tabindex'   => 0,
+				'role'       => 'button',
+				'aria-label' => esc_attr__( 'See more', 'amp' ),
```
westonruter, line 75: No need to escape when manipulating the DOM:

```suggestion
				'aria-label' => __( 'See more', 'amp' ),
```

includes/embeds/class-amp-tumblr-embed-handler.php
```diff
@@ -6,54 +6,89 @@
...
+			]
+		);
+
+		$overflow_node->textContent = esc_html__( 'See more', 'amp' );
+		$amp_node->appendChild( $overflow_node );
+
+		// Append the original link as a placeholder node.
+		if ( $node->firstChild instanceof DOMElement && 'a' === $node->firstChild->nodeName ) {
```
westonruter, line 83: Perhaps this could make use of `get_first_element_child` mentioned above.

westonruter (changes requested):
Good to see much of the duplicated code being factored away.

There are a couple areas that need to be addressed before this is ready:

It is relatively common for developers to create their own embed handlers, slightly less common I'd say to creating custom sanitizers. Sanitizers and embed handlers are two key ways the plugin is extended. Therefore, it is critical that backwards-compatibility be maintained for existing embed handlers (and sanitizers).

Additionally, it is important that any APIs introduced on the base method be ergonomic for developers to use when they extend the embed base handler. For example, some of the methods take a lot of positional parameters which make it hard to read.

westonruter:
As discussed over chat, let's pause work on this until after beta1 or else 1.7 worst case. This will allow us to focus on the RME issues.

thelovekesh:
Closing in favor of #7564.
cc @westonruter

PR #4657, opened 2020-05-05
Add ability to suppress plugins from AMP responses

### LINKED ISSUE #3474: AMP mode and plugin/theme versions need to be included in validated environment setting to determine whether stale
## Bug Description

If you switch AMP mode, the validation cache is not invalidated. However, some validation results depend on whether the AMP page is the canonical URL, and therefore change between modes.

An example of this would be the change introduced in #3346, where a change from `Transitional` to `Standard` should change the excessive CSS validation errors from `New - Accepted` to `New - Rejected`.

## Expected Behaviour

When I switch the AMP mode, I expect the changes in validation to be immediately reflected.

## Steps to reproduce

1. Set mode to `Transitional`.
2. Add enough CSS to the page to hit the excessive CSS validation error on the AMP url of a page.
3. The validation screen will show that the excessive CSS validation error is marked as `New - Accepted`.
4. Change mode to `Standard`.
5. The validation screen for that same page will still show the excessive CSS validation error marked as `New - Accepted` - it should be `New - Rejected` instead.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

### LINKED ISSUE #4477: Gracefully handle plugins that cause validation errors
In either Transitional mode or Reader mode, when a given plugin causes a validation error, the user will be presented with the option to prevent the plugin from running on AMP pages:

- A new section in the AMP plugin admin screen enabling users to specify which of the installed plugins can be deactivated during amp responses when in transitional or reader mode (plugin suppression)
- Alternatively, the UI to deactivate plugins in AMP responses could be integrated into the plugin list table.
- In Reader Mode, the AMP plugin will automatically remove invalid markup and will notify the user there is an issue with the plugin, and the administrator of the site highlighting the offending markup and plugin (same to be done for Transitional mode and Standard mode).
- Provide assistance when encountering validation errors

westonruter:
## Summary

Fixes #4477.
Fixes #3474.

This PR introduces a new Suppressed Plugins (Plugin Suppression) section in the AMP settings screen which is shown when:

1. There are plugins which have been detected to cause validation errors (in non-stale validated URLs).
2. Such plugins have been suppressed.

image

<details>
<summary><small>Old screenshot</small></summary>

image

</details>

The Suppressed Plugins section is omitted entirely if no validation errors are being caused by plugins.

When a plugin is causing validation errors or it has been suppressed, it will appear in the list of initially-visible plugins with select dropdowns.  The other plugins are hidden, and the entire section will appear only if there are any plugins that have been suppressed or if there are other plugins which are causing validation errors. When there are validation errors associated with a given plugin, the errors will be listed with each plugin to provide you with a quick way to evaluate whether the errors are serious enough to warrant turning off the plugin on AMP responses. Validation errors which are bolded are those which are unreviewed. Validation errors with markup removed have a green bullet, and validation errors with red markup have a red bullet.

image
<details>
<summary><small>Old screenshot</small></summary>

image

</details>

The validation errors are only accounted for when the recent validated URLs are not stale. In other words, when a plugin has been updated and the validated URLs were last checked with an older version of the plugin, then the plugins won't be listed among those which are candidates for suppression. Similarly, once a plugin has been suppressed and then the user updates the plugin, a message will be added next to the plugin to indicate that a new version has been installed and that it should perhaps be re-checked for validity since the issue may not be present anymore:

image
<details>
<summary><small>Old screenshot</small></summary>

image

</details>

Plugins are suppressed by:

* Unhooking filters and actions whose callbacks are defined in the plugin. 
* Overriding shortcodes with plugin-defined callbacks to just `__return_empty_string`.
* Overriding widgets with plugin-defined callbacks to just `__return_null`. In addition to this, the display is also blocked with the `widget_display_callback` filter to ensure widgets rendered with `the_widget()` are accounted for.
* Overriding dynamic blocks with plugin-defined callbacks to just `__return_empty_string` and to also nullify the frontend `style` and `script`.

Some plugins to help with testing (which have also been adapted to be included as part of the tests):

* [Bad Shortcode](https://gist.github.com/westonruter/6880918d45deb39c89e17f3b76583521)
* [Bad Widget](https://gist.github.com/westonruter/43c6e09354a4bcac008cc14d248e804c)
* [Bad Block](https://gist.github.com/westonruter/e4331b3f6d03aac075b86b226acd3ba5)
* [Bad Hooks](https://gist.github.com/westonruter/f451337351e803044413d942f307229b)

When all of those plugins are active and a post is validated which has the shortcode, widget, and block all present on its template, the validation errors will show as:

image

When all of these are suppressed, no validation errors should be raised by any of them:

image

## Checklist


westonruter:
This isn't ready for review yet.

src/PluginSuppression.php
```diff
@@ -0,0 +1,140 @@
...
+	 * @see \AMP_Validation_Manager::wrap_widget_callbacks() Which needs to run after this.
+	 *
+	 * @param string[] $suppressed_plugins Suppressed plugins.
+	 * @global array $wp_registered_widgets
+	 */
+	private function suppress_widgets( $suppressed_plugins ) {
+		global $wp_registered_widgets;
+		foreach ( $wp_registered_widgets as $widget_id => &$registered_widget ) {
+			if ( $this->is_callback_plugin_suppressed( $registered_widget['callback'], $suppressed_plugins ) ) {
+				$registered_widget['callback'] = '__return_empty_string';
```
westonruter:
```suggestion
				$registered_widget['callback'] = '__return_null';
```

includes/options/class-amp-options-menu.php
```diff
@@ -423,6 +434,227 @@ private function list_template_conditional_options( $options, $parent = null ) {
...
+				$other_plugin_count = count( $plugins ) - count( $already_listed_plugins );
+				?>
+				<?php if ( $other_plugin_count > 0 ) : ?>
+					<details>
+						<summary>
+							<?php
+							echo esc_html(
+								sprintf(
+									/* translators: %s is the count of plugins */
+									_n(
```
amedina: Let's find an alternative copy for `Other %d plugins`. When there are no plugins with validation issues `other` does not apply. Perhaps `Compliant plugins` or something like that?

westonruter: Good catch. Addressed now that the Other section is removed entirely in d3e772e.

westonruter:
I still have to add tests, but this is ready for review otherwise.

includes/options/class-amp-options-manager.php
```diff
@@ -272,12 +273,85 @@ public static function validate_options( $new_options ) {
...
+			}
+		}
+
+		// When the suppressed plugins changed, re-check the most recently validated URL so validation errors can be
+		// re-computed with the plugins newly-suppressed or un-suppressed.
+		// @todo Instead of checking the most recently-validated URL, instead check the URL(s) that the plugin's validation errors were known to occur on.
+		if ( ! empty( $deleted_suppressed_plugins ) || ! empty( $inserted_suppressed_plugins ) ) {
+			add_action(
+				'update_option_' . self::OPTION_NAME,
+				function () {
```
pierlon:
```suggestion
				static function () {
```

includes/options/class-amp-options-menu.php
```diff
@@ -423,6 +436,191 @@ private function list_template_conditional_options( $options, $parent = null ) {
...
+					array_keys( AMP_Options_Manager::get_option( Option::SUPPRESSED_PLUGINS ) )
+				),
+				array_map(
+					function ( $plugin_file ) {
+						return strtok( $plugin_file, '/' );
+					},
+					get_option( 'active_plugins', [] )
+				)
+			)
+		);
```
pierlon: Perhaps this could be refactored so that it could be easier to understand what's going on here.

includes/options/class-amp-options-menu.php
```diff
@@ -423,6 +436,191 @@ private function list_template_conditional_options( $options, $parent = null ) {
...
+	 */
+	private static function get_suppressible_plugin_sources() {
+		return array_unique(
+			array_intersect(
+				array_merge(
+					array_keys( self::get_plugin_errors_by_sources() ),
+					array_keys( AMP_Options_Manager::get_option( Option::SUPPRESSED_PLUGINS ) )
+				),
+				array_map(
+					function ( $plugin_file ) {
```
pierlon:
```suggestion
					static function ( $plugin_file ) {
```

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -890,7 +953,7 @@ public static function get_validated_environment() {
 	 *     Staleness of the validation results. An empty array if the results are fresh.
 	 *
 	 *     @type string $theme   The theme that was active but is no longer. Absent if theme is the same.
-	 *     @type array  $plugins Plugins that used to be active but are no longer, or which are active now but weren't. Absent if the plugins were the same.
+	 *     @type array  $plugins Plugins that used to be active but are no longer, or which are active now but weren't. Also includes plugins that have version updates. Absent if the plugins were the same.
 	 *     @type array  $options Options that are now different. Absent if the options were the same.
```
pierlon: Is an array of stale options still supposed to be returned here?

includes/validation/class-amp-validated-url-post-type.php
```diff
@@ -865,16 +865,79 @@ public static function store_validation_errors( $validation_errors, $url, $args
...
+			if ( empty( $validation_errors ) ) {
+				continue;
+			}
+
+			foreach ( $validation_errors as $validation_error ) {
+				if ( empty( $validation_error['data']['sources'] ) || ! $validation_error['term'] instanceof WP_Term ) {
+					continue;
+				}
+				foreach ( $validation_error['data']['sources'] as $source ) {
+					if ( ! isset( $source['type'], $source['name'] ) || ! $validation_error['term'] instanceof WP_Term ) {
```
pierlon: Does `$validation_error['term']` needs to be checked again if it's an instance of `WP_Term`?

```suggestion
					if ( ! isset( $source['type'], $source['name'] ) {
```

includes/options/class-amp-options-menu.php
```diff
@@ -423,6 +436,191 @@ private function list_template_conditional_options( $options, $parent = null ) {
...
+
+			<?php
+			$element_name = AMP_Options_Manager::OPTION_NAME . '[' . Option::SUPPRESSED_PLUGINS . ']';
+
+			$errors_by_sources = self::get_plugin_errors_by_sources()
+			?>
+			<ul>
+				<?php foreach ( $plugins as $plugin_file => $plugin ) : ?>
+					<?php
+					$plugin_slug = strtok( $plugin_file, '/' );
```
pierlon: What about in (rare) cases where multiple plugins can be found in the same directory?

includes/options/class-amp-options-menu.php
```diff
@@ -423,6 +436,191 @@ private function list_template_conditional_options( $options, $parent = null ) {
...
+							}
+							echo esc_html( $plugin['Name'] );
+							if ( ! $is_suppressed ) {
+								echo '</strong>';
+							}
+							?>
+						</label>
+						<?php if ( $is_suppressed && version_compare( $suppressed_plugins[ $plugin_slug ][ Option::SUPPRESSED_PLUGINS_LAST_VERSION ], $plugins[ $plugin_file ]['Version'], '!=' ) ) : ?>
+							<small>
+								<?php if ( $suppressed_plugins[ $plugin_slug ][ Option::SUPPRESSED_PLUGINS_LAST_VERSION ] && $plugins[ $plugin_file ]['Version'] ) : ?>
```
pierlon: Should this condition be checked before comparing versions above?

pierlon (changes requested):

pierlon:
This should also address #3474.

includes/options/class-amp-options-menu.php
westonruter: Yeah, it's a good point. The reality is that `AMP_Validation_Manager::get_source()` is not able to detect such nested plugins anyway. The source slug extracted from a plugin will always be just the directory name if the plugin is not a single file:

https://github.com/ampproject/amp-wp/blob/9d662d07da228c6cc8dedfab5abc724feaf39177/includes/validation/class-amp-validation-manager.php#L1636-L1639

So that is why the the directory alone is being considered (or else just the file name if the plugin is not in a directory).

includes/options/class-amp-options-menu.php
westonruter: Yes, good. Done in d24b29c6bff37e676196dbe8dd89bad95615e4b9.

includes/options/class-amp-options-menu.php
westonruter: This is here in the rare situation where a plugin was updated to add a version which previously lacked it, or it was updated to remove the version where it previously had one.

includes/validation/class-amp-validated-url-post-type.php
westonruter: Actually, you identified some code that needed to be resurrected to address #3474. Resurrection implemented in 4e1c478.

westonruter:
Thoughts on the use of strike-through styling when a plugin is checked? I realize that may not be so helpful.

amedina:
I personally don't like the strike through approach much. IIUC the goal is to make sure users understand clearly that the given plugin has been disabled on AMP. I would like to see that clearly conveyed. Perhaps a label next to the plugin name?

westonruter:
> I personally don't like the strike through approach much. IIUC the goal is to make sure users understand clearly that the given plugin has been disabled on AMP. I would like to see that clearly conveyed. Perhaps a label next to the plugin name?

I think perhaps the strike-through is entirely redundant. The checked state indicates what is suppressed. I'll just remove the styling.

pierlon:
> I think perhaps the strike-through is entirely redundant. The checked state indicates what is suppressed.

Agreed. I also think the updated suppressed plugins could be more indicative. Maybe having them highlighted, or even put into a separate group, for example.

schlessera:
I think the checkboxes might visually indicate the opposite of what they should. Normally, a checked checkbox indicates "enabled", "on", "active", ...

How about using similar comboboxes than with the validation error lists here, to keep the same language? Instead of `Removed`/`Kept`, they could say `Suppressed`/`Active` or so.

westonruter:
@schlessera @pierlon What do you think of this:

image

westonruter:
I'm not entirely sure of the title of the “Details” column. I added a datetime at which the plugin was suppressed in order to avoid the cell from being blank. I've also added the plugin URL, description and author so that the user can get more context for the plugin to decide whether it is safe to suppress:

image

Aside: Someone should really reach out to that Weston Ruter guy to educate him on how to write plugins so they don't cause validation errors!

amedina:
I am not too keen on putting the details of the offending plugin functionality on this screen. The amount of real estate is very limited and the details may not convey the right information.

amedina:
The description of the `Suppressed plugins` section is too verbose. We should simplify it and link to associated documentation if needed.

amedina:
Consider using `Blocked plugins` in lieu of `Suppressed plugins`

westonruter:
> Consider using `Blocked plugins` in lieu of `Suppressed plugins`

I still prefer “suppressed” in favor of “blocked”. The term “block” is getting overly used in WordPress now, and I think “suppress” precisely depicts what is going on: “prevent the development, action, or expression of (a feeling, impulse, idea, etc.); restrain.”

> I am not too keen on putting the details of the offending plugin functionality on this screen. The amount of real estate is very limited and the details may not convey the right information.

I'm not sure. Where else would the information be displayed? The `details` elements are collapsed by default. How would someone discover understand what validation errors are being caused by the plugin, or to what extent a plugin is causing validation errors?

It's important for there to be as much information here as possible to help the user make a judgement as to whether the plugin can indeed be suppressed. This is why I added links off to the plugin homepage, the author name, and the plugin description.

In general, it could feel somewhat familiar to the plugins list table on the Plugins screen. That's kinda what I was aiming for.

> The description of the Suppressed plugins section is too verbose. We should simplify it and link to associated documentation if needed.

Agreed. It should be made more concise. But a benefit of it being in the plugin is that it will get translated. If we link off to docs on amp-wp.org then it won't get the benefit of translations (at least via translate.wordpress.org).

Nevertheless, it's important to note that this section only appears when there are plugins that are causing validation errors. Otherwise, the section is not shown at all. So in an ideal world, the appearance of this section should an be “exception”.

westonruter:
> In general, it could feel somewhat familiar to the plugins list table on the Plugins screen. That's kinda what I was aiming for.

In general, a table is able to present more more information in a structured way. My initial design of displaying the information in a linear list was inadequate for that purpose.

amedina:
> I still prefer “suppressed” in favor of “blocked”. The term “block” is getting overly used in WordPress now, and I think “suppress” precisely depicts what is going on: “prevent the development, action, or expression of (a feeling, impulse, idea, etc.); restrain.”

Makes sense. 
  
> I'm not sure. Where else would the information be displayed? The `details` elements are collapsed by default. How would someone discover understand what validation errors are being caused by the plugin, or to what extent a plugin is causing validation errors?

Is the goal to provide information to the user about the functionality of the plugin? And then the assumption is that users will use that information to get a better understanding of the potential issues being reported?

If that is the goal, my concern is that the functionality of a plugin is hard to capture in one sentence or so. For example, how to define the functionality of Jetpack? And having either a truncated description may not serve the purpose we want, and if we have a full description the UI can be cluttered. 
 
> It's important for there to be as much information here as possible to help the user make a judgement as to whether the plugin can indeed be suppressed. This is why I added links off to the plugin homepage, the author name, and the plugin description.

Agree. We should strike the right balance. See my concern above. 

> Nevertheless, it's important to note that this section only appears when there are plugins that are causing validation errors. Otherwise, the section is not shown at all. So in an ideal world, the appearance of this section should an be “exception”.

I agree with the goal. We may convey the same information more concisely.

westonruter:
> Is the goal to provide information to the user about the functionality of the plugin? And then the assumption is that users will use that information to get a better understanding of the potential issues being reported?

Yes.

> If that is the goal, my concern is that the functionality of a plugin is hard to capture in one sentence or so. For example, how to define the functionality of Jetpack? And having either a truncated description may not serve the purpose we want, and if we have a full description the UI can be cluttered.

The description being added here is the full description as provided by the plugin. If a user still doesn't understand what the plugin is doing from the description, they can click on the plugin title to visit the plugin homepage to get more information. You're right that the description can't encapsulate all of the information, but it can at least provide better context.

@amedina How about hiding the author name and description inside of `details` like so:

image

This will hide the extra context unless the user needs it. With all details expanded:

image

amedina:
> The description being added here is the full description as provided by the plugin. If a user still doesn't understand what the plugin is doing from the description, they can click on the plugin title to visit the plugin homepage to get more information. You're right that the description can't encapsulate all of the information, but it can at least provide better context.

Ok; that make sense to me. You are right. 
> 
> @amedina How about hiding the author name and description inside of `details` like so:

That would be good. Once the user suppresses a plugin they can collapse it.

westonruter:
Update based on our meeting: the list of validation errors should include the status of each, whether removed/kept and reviewed. Plugins that have all the validation errors reviewed would be less likely needing to be suppressed.

westonruter:
In 5667e67e9a63edb25b8eb35388e9af3b7e83c6cc I've added _very_ abbreviated indications for the statuses of validation errors:

* Orange glow means unreviewed.
* Green bullet is removed invalid markup.
* Red bullet indicates kept invalid markup.

Disclaimer: I'm not a designer. But it feels pretty good, capturing the information concisely. Here's plugin expanded with validation errors in all 4 states:

image

Each validation error link has a tooltip, in 4 possible values (for accessibility and discoverability):

1. Invalid markup causing the validation error is removed and reviewed. See all validated URL(s) with this validation error.
2. Invalid markup causing the validation error is kept and reviewed. See all validated URL(s) with this validation error.
3. Invalid markup causing the validation error is removed and unreviewed. See all validated URL(s) with this validation error.
4. Invalid markup causing the validation error is kept and unreviewed. See all validated URL(s) with this validation error.

schlessera:
⚠ ^^^ Force push happened here ^^^ ⚠

schlessera:
@googlebot I consent.

pierlon:
@googlebot I consent.

schlessera:
I don't think the Travis tests ran here.

schlessera:
Hmm, does Travis only run when the `develop` branch is targeted?

westonruter:
> Hmm, does Travis only run when the `develop` branch is targeted?

Yes. See this comment: https://github.com/ampproject/amp-wp/pull/4828#issuecomment-642447198

westonruter:
I'm working on adding tests for the new functionality.

includes/options/class-amp-options-manager.php
```diff
@@ -286,6 +287,17 @@ public static function validate_options( $new_options ) {
...
+		/**
+		 * Filter the options being updated, so services can handle the sanitization and validation of
+		 * their respective options.
+		 *
+		 * @internal
+		 *
+		 * @param array $options     Existing options with already-sanitized values for updating.
+		 * @param array $new_options Unsanitized options being submitted for updating.
+		 */
+		$options = apply_filters( 'amp_options_updating', $options, $new_options );
```
westonruter, line 299: @schlessera I've added this to allow the option updating logic to be removed from the `AMP_Options_Manager` and over to the respective services.

schlessera: I thought about turning the individual options into objects to inject as dependencies already. But the filter will do the trick for now.

BTW, this is what I referred to when I meant that we might need to add hooks so we can reverse the flow of the logic. So we seem to be on the same page wrt this.

includes/validation/class-amp-validation-manager.php
```diff
@@ -2135,6 +2124,45 @@ public static function validate_url( $url ) {
...
+
+		$r = AMP_Validated_URL_Post_Type::store_validation_errors(
+			wp_list_pluck( $validity['results'], 'error' ),
+			$validity['url'],
+			$args
+		);
+		if ( $r instanceof WP_Error ) {
+			return $r;
+		}
+		$validity['post_id'] = $r;
```
schlessera: Abbreviations like this should be avoided.
```suggestion
		$result = AMP_Validated_URL_Post_Type::store_validation_errors(
			wp_list_pluck( $validity['results'], 'error' ),
			$validity['url'],
			$args
		);
		if ( $result instanceof WP_Error ) {
			return $result;
		}
		$validity['post_id'] = $result;
```

src/PluginRegistry.php
```diff
@@ -0,0 +1,141 @@
...
+			 * done because when the Reflection API is being used to determine which plugin a given piece of markup is
+			 * coming from, it cannot absolutely determine which plugin file was responsible for including the PHP file
+			 * that the function was defined inside of.
+			 */
+			if ( isset( $plugins[ $plugin_slug ] ) && basename( $plugins[ $plugin_slug ]['File'] ) === "{$plugin_slug}.php" ) {
+				continue;
+			}
+
+			$plugins[ $plugin_slug ] = array_merge(
+				[ 'File' => $plugin_file ], // Camel case is used for consistency with the other keys.
```
schlessera: This is not camelCase, this is PascalCase (or StudlyCaps).
```suggestion
				[ 'File' => $plugin_file ], // PascalCase is used for consistency with the other keys.
```

schlessera (changes requested):
Looks good, only 2 nitpicks!

schlessera (approved):

PR #4588, opened 2020-04-15
Stub request based on test scenario

pierlon:
## Summary

Fixes the [failing spec test](https://travis-ci.org/github/ampproject/amp-wp/jobs/675380098#L929-L953) in the Optimizer library for the data set `AmpRuntimeCss - always_inlines_v0css`.

## Checklist

- [ ] My pull request is addressing an [open issue](https://github.com/ampproject/amp-wp/contributing/project-management.md#life-of-an-issue) (please create one otherwise).
- [ ] My code is tested and passes existing [tests](https://github.com/ampproject/amp-wp/contributing/engineering.md#tests).
- [ ] My code follows the [Engineering Guidelines](https://github.com/ampproject/amp-wp/contributing/engineering.md) (updates are often made to the guidelines, check it out periodically).

PR #4662, opened 2020-05-06
Add input file configuration parsing for spec tests

### LINKED ISSUE #4654: Add configuration parsing to transformer spec test runner
## Feature description

The language agnostic spec tests in the `amp-toolbox` package use a leading HTML comment to configure individual tests: https://github.com/ampproject/amp-toolbox/blob/bb242b62266a71c777437d199161ba850ddab53b/packages/optimizer/spec/transformers/valid/AmpBoilerplateTransformer/always_inlines_v0css/input.html#L1-L6

This is now being used for a few tests that we are running as well, so we need to implement the same thing.

---------------

## Acceptance criteria

* Leading HTML comments with arguments should properly configure the transformer spec tests.

## Implementation brief

* Check if the input document starts with an HTML comment.
* If yes, fetch the HTML comment and strip it from the input file.
* Use `json_decode()` on the text content of the comment.
* The resulting JSON document represents the options that should be passed into the transformer(s).

Note: This might require mapping from Node.js options to PHP options if these are not the same. Alternatively, we might consider adapting the options to keep them in sync with the Node.js version.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
## Summary

This PR includes the following changes:
- Updates spec files to latest `amp-toolbox` version.
- Adapts `SpecTest.php` to extract the HTML comment with configuration arguments from the input HTML files of the spec test suite.
- Adds a `STYLES` configuration key to the `AmpRuntimeCss` configuration.
- Overrides stylesheet fetching with the stylesheet provided via the `STYLES` config argument.
- Removes the empty `<style amp-runtime>` element if the stylesheet ended up being linked to.
- Adds additional tests for `AmpRuntimeCss` to ensure configuration arguments behave as expected.
- Removes temporary `StubbedRequests` filtering from https://github.com/ampproject/amp-wp/pull/4588
- Adds `.gitattributes` file to denote generated files, so that these are collapsed by default in code review.

Fixes #4654 

## Checklist


.gitattributes
```diff
@@ -0,0 +1,2 @@
+lib/optimizer/resources/local_fallback/* linguist-generated=true
```
pierlon, line 1: TIL about using `.gitattributes` to suppress diffs for generated code ([docs](https://github.com/github/linguist#generated-code)) :smile:.

Would the path `lib/optimiser/` be a better place for this file?

westonruter: Yes, that's right. Placing it into `lib/optimizer` will work, [per Git docs](https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#:~:text=in%20one%20of%20your%20directories):

image

westonruter: As long as GitHub recognizes this, as [their docs](https://help.github.com/en/github/administering-a-repository/customizing-how-changed-files-appear-on-github) only mention the repo root.

lib/optimizer/tests/SpecTest.php
```diff
@@ -108,33 +108,99 @@ public function testTransformerSpecFiles($scenario, $transformerClass, $source,
...
+     */
+    public function extractConfigurationData(&$source)
+    {
+        $matches = [];
+        if (!preg_match(self::LEADING_HTML_COMMENT_REGEX_PATTERN, $source, $matches)) {
+            return [];
+        }
+
+        $json   = trim($matches['json']);
+        $source = str_replace($matches[0], '', $source);
```
westonruter: Minor thing, but this could alternatively be:

```suggestion
        $source = substr($source, strlen($matches[0]));
```

Since the match is at the beginning of the string. This could potentially prevent a scenario where a replacement is done later in the source than desired.

lib/optimizer/tests/SpecTest.php
```diff
@@ -108,33 +108,99 @@ public function testTransformerSpecFiles($scenario, $transformerClass, $source,
...
+
+        if (empty($json)) {
+            return [];
+        }
+
+        $configurationData = (array)json_decode($json, true, 512, JSON_OBJECT_AS_ARRAY);
```
westonruter: Why is `JSON_OBJECT_AS_ARRAY` needed if `assoc=true` is the second argument?

.gitattributes
schlessera: I assumed it would only work in the project root, but I can try moving it around to see what happens.

lib/optimizer/tests/SpecTest.php
schlessera: Resolved via https://github.com/ampproject/amp-wp/pull/4662/commits/96c7991cfe4808f97d87623b7c07e6c1d03df10c

.gitattributes
schlessera: No, it doesn't work when the `.gitattributes` file is in a subfolder. I'll revert that change again and leave it in the root for now. This will need to be adapted once we extract the libraries into separate repositories.

Image 2020-05-07 at 8 29 08 AM

westonruter (approved):

Issue #4656, opened 2020-05-05
Preview doesn't work in reader mode on a published post


ktmn:
I've got Twenty Seventeen in reader mode, editing the default "Hello world!" post in the block editor, hitting "Preview" button will display the post with it's changes, but if I hit the new AMP icon next to the preview button, then it will navigate correctly to `?preview_id=7&preview_nonce=51e7746d07&preview=true&amp=1` but the page doesn't contain the updated changes in the post content. 

If I revert the post to draft then the preview works correctly.

This wasn't an issue before AMP plugin v1.5 afaik.

Anyone able to reproduce?

westonruter:
Yes, I can reproduce this issue. The issue seems to be that when the post is published, clicking the AMP Preview button doesn't force the post revision to be saved before the post is navigated to.

This logic seems to be incorrect:

https://github.com/ampproject/amp-wp/blob/4880f0f58daaf07685854be8574ff25d76ff583e/assets/src/block-editor/components/amp-preview.js#L206-L211

But I'm not sure why because this is the same code that core button is using:

https://github.com/WordPress/gutenberg/blob/2beb59861039c43c71f03c25256479b30c148387/packages/editor/src/components/post-preview-button/index.js#L144-L180

ktmn:
This `$content` here doesn't contain the preview edits (while `get_the_content()` does):

https://github.com/ampproject/amp-wp/blob/90b306d5a35631a08cd7a1ec2798217bde0d6f4b/includes/templates/class-amp-post-template.php#L348-L354

The post is retrieved using `get_post()`:

https://github.com/ampproject/amp-wp/blob/90b306d5a35631a08cd7a1ec2798217bde0d6f4b/includes/templates/class-amp-post-template.php#L162-L168

but the preview content seems to be only injected for the [main query](https://github.com/WordPress/WordPress/blob/3ee9488292318153390d5631a08e2f7713ef7ac1/wp-includes/class-wp-query.php#L3125).

A fix that seems to work for me is to adapt [this core code](https://github.com/WordPress/WordPress/blob/3a333ef030345c459b9ba131b24735a3f8f635f3/wp-includes/revision.php#L553-L600) to load the preview content into the post object from `get_post()` (at [class-amp-post-template.php#L164](https://github.com/ampproject/amp-wp/blob/90b306d5a35631a08cd7a1ec2798217bde0d6f4b/includes/templates/class-amp-post-template.php#L164)):

```php
	public function __get( $name ) {
[...]

PR #4663, opened 2020-05-06
Fix preview of published post in reader mode

ktmn:
Addresses #4656

### DIFF
--- a/includes/templates/class-amp-post-template.php
+++ b/includes/templates/class-amp-post-template.php
@@ -162,7 +162,14 @@ class AMP_Post_Template {
 	public function __get( $name ) {
 		switch ( $name ) {
 			case 'post':
-				return get_post( $this->ID );
+				$post = get_post( $this->ID );
+				if ( isset( $_GET['preview_id'] ) && isset( $_GET['preview_nonce'] ) ) {
+					$id = (int) $_GET['preview_id'];
+					if( wp_verify_nonce( $_GET['preview_nonce'], 'post_preview_' . $id ) ) {
+						return _set_preview( $post );
+					}
+				}
+				return $post;
 		}
 		return null;
 	}


includes/templates/class-amp-post-template.php
```diff
@@ -162,7 +162,14 @@ private function set_data() {
 	public function __get( $name ) {
 		switch ( $name ) {
 			case 'post':
-				return get_post( $this->ID );
+				$post = get_post( $this->ID );
+				if ( isset( $_GET['preview_id'] ) && isset( $_GET['preview_nonce'] ) ) {
```
pierlon, line 166:
```suggestion
				if ( isset( $_GET['preview_id'], $_GET['preview_nonce'] ) ) {
```

includes/templates/class-amp-post-template.php
```diff
@@ -162,7 +162,14 @@ private function set_data() {
...
+					$id = (int) $_GET['preview_id'];
+					if( wp_verify_nonce( $_GET['preview_nonce'], 'post_preview_' . $id ) ) {
```
pierlon, line 168:
```suggestion
					if ( wp_verify_nonce( $_GET['preview_nonce'], 'post_preview_' . $id ) ) {
```

pierlon (changes requested):
Great work here, @ktmn. Thanks for opening this PR!

A few lint errors to be resolved and we should be able to get this merged :smile:.

pierlon:
@ktmn also, don't forget to sign the CLA, as outlined in https://github.com/ampproject/amp-wp/pull/4663#issuecomment-624776452.

westonruter (changes requested):
After digging into this, I found an alternative solution that I think will be more robust. All we need to do is pass in the initial `$post` when constructing the `AMP_Post_Template` and use it from then on. It will have `the_preview` applied already, so we don't need to call it manually:

```diff
--- a/includes/templates/class-amp-post-template.php
+++ b/includes/templates/class-amp-post-template.php
@@ -66,6 +66,13 @@ class AMP_Post_Template {
 	 */
 	public $ID;
 
+	/**
+	 * Post
+	 *
+	 * @var WP_Post
+	 */
+	public $post;
+
 	/**
 	 * AMP_Post_Template constructor.
 	 *
@@ -73,9 +80,11 @@ class AMP_Post_Template {
 	 */
 	public function __construct( $post ) {
 		if ( is_int( $post ) ) {
-			$this->ID = $post;
+			$this->ID   = $post;
+			$this->post = get_post( $post );
 		} elseif ( $post instanceof WP_Post ) {
-			$this->ID = $post->ID;
+			$this->ID   = $post->ID;
+			$this->post = $post;
 		}
 	}
 
@@ -151,22 +160,6 @@ class AMP_Post_Template {
 		$this->data = apply_filters( 'amp_post_template_data', $this->data, $this->post );
 	}
 
-	/**
-	 * Getter.
-	 *
-	 * @since 1.5
-	 *
-	 * @param string $name Property name.
-	 * @return mixed
-	 */
-	public function __get( $name ) {
-		switch ( $name ) {
-			case 'post':
-				return get_post( $this->ID );
-		}
-		return null;
-	}
-
 	/**
 	 * Get template directory for Reader mode.
 	 *
diff --git a/includes/templates/reader-template-loader.php b/includes/templates/reader-template-loader.php
index d518343dc..5833bc9ba 100644
--- a/includes/templates/reader-template-loader.php
+++ b/includes/templates/reader-template-loader.php
@@ -32,5 +32,5 @@ do_action( 'pre_amp_render_post', get_queried_object_id() );
 require_once AMP__DIR__ . '/includes/amp-post-template-functions.php';
 amp_post_template_init_hooks();
 
-$amp_post_template = new AMP_Post_Template( get_queried_object_id() );
+$amp_post_template = new AMP_Post_Template( $post );
 $amp_post_template->load();
```

ktmn:
@westonruter Yeah that makes sense, go with that.

westonruter:
@ktmn cool. Please test the changes in #4665.

PR #4665, opened 2020-05-07
Fix post preview in Reader mode

### LINKED ISSUE #4656: Preview doesn't work in reader mode on a published post

westonruter:
## Summary

Fixes #4656
Supersedes #4663

This fixes a regression which was introduced in c74ba974a3121b0fef195f4e7b64724760bc1be9 (as part of #3781) when moving to lazy-load properties of `AMP_Post_Template`. As I recall, switching to load the post data with a getter when accessing `post` was to avoid keeping a copy of that post around necessarily, as that would introduce a possibility of the copy in `AMP_Post_Template` becoming out of sync with what is in the DB. Nevertheless, this broke the ability to preview changes to the queried object, which gets populated by the `the_preview` filter.

This PR reverts part of c74ba974a3121b0fef195f4e7b64724760bc1be9.

## Checklist

- [ ] My pull request is addressing an open issue (please create one otherwise).


pierlon (approved):

ktmn:
Works for me

westonruter:
I found something that was missed: previewing changes to the title. I've addressed this in 7d81390. Please review.

pierlon (approved):
Verified that the post title and excerpt are updated when previewed :+1:.

PR #4674, opened 2020-05-08
Use <amp-script> for comment form handling

### LINKED ISSUE #4389: Use amp-script instead of amp-bind to implement comment form interactivity with 100% feature parity
## Feature description

When support for WordPress comments was introduced, amp-bind was used to implement support for inline replies. Using amp-bind got us most of the way there, but there are some pieces that don't work.

Clicking the Reply link brings the user down to the comment form instead of bringing the comment form _up_ to appear right after the comment they are replying to.:

image

Instead of bringing the form up inline the comment being replied to, the AMP implementation changes the reply-title heading to indicate the name of the commenter being replied to. This is not ideal. Additionally, when `comment_form()` is called with arguments to override the reply-title, we cannot use amp-bind to override heading in the first place (see #2489).

The reason why amp-bind was originally used to implement interactivity for WordPress comments is that it is what we had available at the time. Now, however, we have amp-script. We should be able to achieve 100% AMP parity in functionality by replacing amp-bind with amp-script. We may even be able to core's [`comment-reply.js`](https://github.com/WordPress/wordpress-develop/blob/5.3.2/src/js/_enqueues/lib/comment-reply.js) as is (or rather `comment-reply.min.js`)! The main thing we'd need to do is use a sanitizer to rewrite the use of `onclick` attributes as used by `get_post_reply_link()`.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
## Summary

This is a broken work-in-progress PR just meant to store the current state of the work while we put the corresponding issue into the icebox.

As discussed during a plugin sync, the current implementation of the `<amp-script>` element and how it executes the JS within a worker DOM seems to be still very rough. The initial idea of just loading the WordPress built-in `comment-reply.js` file into the `<amp-script>` element and execute it from there is nowhere near what that component is currently capable of. A lot of JS APIs are either missing, or - more deceivingly - execute just fine, but don't work as expected because they are run within the worker DOM.

The most prominent example that the work was currently stuck on was the support for `event.preventDefault()`. The documentation claimed this was supported ([fixed in the meantime](https://github.com/ampproject/worker-dom/pull/858)), but even though it executes just fine within the worker DOM, it doesn't have any repercussions on the actual browser window and document. Therefore, a link will still navigate to a different page, for example.

Right now, this issue seems like a big time sink with unpredictable time estimates or results. Therefore we're putting it on hold for now, in the hopes that the `<amp-script>` will continue to mature and provide a more efficient experience the next time around.

Addresses issue #4389 

## Checklist


andrei0x309:
I don't know if it helps but I implemented a minimalist custom commenting system using `amp-script`, the comments are submitted and displayed by custom `fetch` call from the `amp-script`.

Also, comments are fetched only if someone presses _show comments_ and it has support for fetch pagination.
The comment form moves like in default WordPress. 

I put all the relevant files [in this gist](https://gist.github.com/andrei0x309/496c328a34e2a2c4bada00c439ef629b).

Also in case, the `amp-wp` plugin will migrate to `amp-script` for the comments section it would be a very good idea to add a filter for disabling the `amp-script` feature on comments in case someone has made a custom `amp-script`.

Here's the demo.gif from the gist.

Demo Image

westonruter:
@andrei0x309 Nicely done! Thanks for sharing. I like how you've improved the UX by showing the posted without requiring a reload of the page. Something to be wary of there is the possibility that there could be some markup the user ads to a comment that which wasn't previously on the page (like maybe an `em`). If you have styling for `em` and it wasn't previously on the page then the CSS for that would have been tree-shaken. To guard against that, you could have CSS selectors that include `amp-script` as an ancestor, like `amp-script em` as tree shaking in that case would disregard looking to see if `em` is in the page it's understood to be inside of a dynamically-updated element.

PR #4679, opened 2020-05-10
Strip leading BOM+whitespace and trailing HTML comment before parsing validation response JSON

### LINKED ISSUE #4678: Byte order mark (BOM) breaks ability to do validation
## Bug Description

As noted in a [support topic](https://wordpress.org/support/topic/url-validation-failed-error/#post-12811452), when a byte order mark (BOM) is present in a theme/plugin PHP file, it is prepended to the JSON that is output which then causes the AMP plugin to throw a `response_not_json` error since the response does not start with `{`. The user sees this error message:

URL validation failed due to the AMP validation request not returning JSON data. This may be due to a PHP fatal error occurring. Please check your server PHP error logs; to do this you may need to enable WP_DEBUG_LOG. If you are stuck, please search the support forum for possible related topics, or otherwise start a new support topic including the error message, the URL to your site, and your active theme/plugins. Please include your Site Health Info.

The plugin should be more forgiving about the presence of the BOM as well as other leading whitespace.

Additionally, some caching plugins append an HTML comment to the end of the response, for example LiteSpeed Cache (https://github.com/litespeedtech/lscache_wp/issues/212). This also breaks JSON parsing. So any HTML comments that are appended to the response should also be stripped prior to JSON parsing.

## Expected Behaviour

Validation responses that have a leading BOM or whitespace, or trailing HTML comment, should all be stripped prior to JSON parsing.

## Steps to reproduce

Activate this plugin:

```php
<?php
/**
 * Plugin Name: Add BOM and HTML Comment to End
 */

add_action(
	'init',
	function () {
		echo "\xEF\xBB\xBF ";
	}
);

add_action( 'shutdown', function () {
	if ( did_action( 'wp_footer' ) ) {
		echo "<!-- Generated by me! -->";
	}
} );
```

Then try to validate a page.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

westonruter:
## Summary

Fixes #4678

## Checklist


westonruter:
Build for testing: [amp.zip](https://github.com/ampproject/amp-wp/files/4606599/amp.zip) (1.5.4-alpha-20200510T191400Z-314ee33d9)

amedina (approved):
Neat. LGTM.

pierlon (approved):

includes/validation/class-amp-validation-manager.php
```diff
@@ -2114,6 +2114,17 @@ public static function validate_url( $url ) {
 			$validation_url
 		);
 
+		// Strip byte order mark (BOM).
+		if ( "\xEF\xBB\xBF" === substr( $response, 0, 3 ) ) {
```
schlessera, line 2118: I would have turned this into a `while` loop, as under certain circumstances, some systems prepend the BO without checking if there is one already. So you can end up with multiple ones as well.

westonruter: Good point. Done in #4683.

PR #4685, opened 2020-05-12
Raise PHPStan level to 4 in lib/optimizer

### LINKED ISSUE #4380: Raise PHPStan level for the `amp/optimizer` package to 4
## Feature description

The Optimizer package (`ampproject/optimizer`) is currently sitting at and enforcing level 3 of PHPStan static analysis.

There are 3 errors that need to be fixed so that we can raise the level to 4 and enforce this new quality level in Travis CI.

---------------

## Acceptance criteria

* PHPStan does not show any detected errors when run on level 4.

## Implementation brief

* Change the level to `4` in the file `phpstan.neon.dist`.
* Run `composer analyze`.
* Fix detected errors until no more are found.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
## Summary

This PR fixes the following issues that PHPStan detected at level 4:

- [AmpRuntimeCss.php:173](https://github.com/ampproject/amp-wp/tree/9a8928b00d272cb6457bad3df2b9a70ec4c8174b/lib/optimizer/src/Transformer/AmpRuntimeCss.php#L173) - **Comparison operation ">=" between int<min, 200> and 300 is always false.** => actual bug, fixed via https://github.com/ampproject/amp-wp/commit/83018f2168f2fa48ac21724f11ac01ea94d85855
- [ReorderHead.php:85](https://github.com/ampproject/amp-wp/tree/9a8928b00d272cb6457bad3df2b9a70ec4c8174b/lib/optimizer/src/Transformer/ReorderHead.php#L85) - **Negated boolean expression is always false.** => actual bug, fixed via https://github.com/ampproject/amp-wp/commit/6c272288dc515c33721a26458703acbed908890d
- [ServerSideRendering.php:553](https://github.com/ampproject/amp-wp/tree/9a8928b00d272cb6457bad3df2b9a70ec4c8174b/lib/optimizer/src/Transformer/ServerSideRendering.php#L553) - **Strict comparison using === between float and 0 will always evaluate to false.** => actual bug, fixed via https://github.com/ampproject/amp-wp/commit/14084492b0a71fd6b24121140dbbba8ed0868aa6
- [ServerSideRendering.php:728](https://github.com/ampproject/amp-wp/tree/9a8928b00d272cb6457bad3df2b9a70ec4c8174b/lib/optimizer/src/Transformer/ServerSideRendering.php#L728) - **Strict comparison using === between DOMNamedNodeMap and null will always evaluate to false.** => redundant code, removed via https://github.com/ampproject/amp-wp/commit/63e99a63ca0624d5cff8f6e8d8d1ac85b89e337b

Fixes #4380

## Checklist


lib/optimizer/src/Transformer/AmpRuntimeCss.php
```diff
@@ -170,7 +170,7 @@ private function inlineCss(DOMElement $ampRuntimeStyle, $version)
             $response   = $this->remoteRequest->get($v0CssUrl);
             $statusCode = $response->getStatusCode();
 
-            if (200 < $statusCode || $statusCode >= 300) {
+            if ($statusCode < 200 || $statusCode >= 300) {
```
westonruter, line 173: cf https://github.com/ampproject/amp-wp/pull/4686#discussion_r424647168

lib/optimizer/src/Transformer/ServerSideRendering.php
```diff
@@ -978,4 +982,26 @@ private function isExcludedElement(DOMElement $element)
...
+    {
+        if (is_int($number)) {
+            return $number === 0;
+        }
+
+        if (!is_float($number)) {
+            throw InvalidArgument::forNumericComparison($number);
+        }
+
+        return abs($number - 0.0) < self::FLOATING_POINT_EPSILON;
```
westonruter: Why subtract `0.0`? Is this for PhpStan?

lib/optimizer/src/Transformer/AmpRuntimeCss.php
schlessera: Yes again, a good candidate for backporting.

lib/optimizer/src/Transformer/ServerSideRendering.php
schlessera: It is the blind application of an algorithm for comparing floating point numbers: `abs( $a - $b ) < $epsilon`. I was not 100% sure whether subtracting a float `0.0` would have an actual effect on the result or not.

However, we're probably safe removing the subtraction, I can't think of a way it would break without it, as we're already checking whether `$number` is a `float` to begin with.

pierlon (approved):

lib/optimizer/src/Transformer/ServerSideRendering.php
pierlon: Right, subtracting `0.0` would convert it into a float, which it already is.

westonruter: OK, then to prevent confusion later:

```suggestion
        return abs($number) < self::FLOATING_POINT_EPSILON;
```

lib/optimizer/src/Transformer/AmpRuntimeCss.php
schlessera: Note, the changes are all in separate commits, so backporting the above should be a straight-forward cherry-pick.

westonruter:
Travis build does pass but it is not being reported here. See https://travis-ci.org/github/ampproject/amp-wp/builds/687604455

lib/optimizer/src/Transformer/AmpRuntimeCss.php
westonruter: The branch doesn't merge cleanly into 1.5, and 83018f2168f2fa48ac21724f11ac01ea94d85855 doesn't merge cleanly either. But I'll apply this one fix manually, and leave the other changes to 1.6 alone.

westonruter: See f9f10d121edc059ab55530d0e0ccf21d25ffd65b.

amedina:
I am assuming that this (and related) PRs were tested in CI properly. No further QA needed.

PR #4689, opened 2020-05-13
Expose `amp_enabled` as an attribute in the REST API and always protect the postmeta counterpart

### LINKED ISSUE #4659: Enable AMP toggle doesn't work when custom fields advanced panel is enabled
## Bug Description

The [Enable AMP](https://jumpshare.com/v/lcjZ4P1kp8dqmfVuo20g) toggle in the editor does not retain the state chosen when the custom fields advanced panel option is enabled:

custom fields advanced panels option

After updating the post, the value for  _amp_status_  does not change in the custom fields panel: 

amp_status value does not change

## Expected Behaviour

The toggle option should retain the state chosen after updating the post/ page in the editor.

## Steps to reproduce

1. Go to the editor for a page/ post with the custom fields panel enabled
2. Toggle [Enable AMP](https://jumpshare.com/v/lcjZ4P1kp8dqmfVuo20g) either on or off
3. Update the page/ post
4. View the page/ post and notice that it is not in the state chosen (AMP enabled or disabled) 
5. Go back to the editor and see the toggle has not retained the state chosen 

## Screenshots

Video: https://recordit.co/A46veHk7wZ

## Additional context

 - WordPress version: 5.4.1
 - Plugin version: 1.5.3
 - AMP plugin template mode: All
 - PHP version: 7.3.5
 - OS: MacOS
 - Browser: Chrome, Firefox
 - Device: MacBook Air

Workaround mini plugin: https://gist.github.com/westonruter/392c3042ede361dd3cd5089733356649

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

This PR adds a new REST API attribute to AMP supported post types called `amp_enabled`, which is a boolean representation of the `amp_status` post meta. It's counterpart, the `amp_status` post meta, is now permanently protected and cannot be modified via an API request nor updated via the Custom Fields meta box, for example.

The AMP status can now only be modified via the REST API, given that the user has the right permissions to do so.

Fixes #4659.

## Checklist


includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +423,22 @@ public function preview_post_link( $link ) {
...
+	 * @param bool   $protected Whether the key is considered protected.
+	 * @param string $meta_key Metadata key.
+	 * @param string $meta_type Type of object metadata is for.
+	 * @return bool
+	 */
+	public function hide_amp_status_meta( $protected, $meta_key, $meta_type ) {
+		global $pagenow;
+
+		if ( is_admin() && 'post.php' === $pagenow && self::STATUS_POST_META_KEY === $meta_key && 'post' === $meta_type ) {
+			$protected = true;
```
westonruter: This is just the hacky way I was able to quickly make this work. However, it seems there should be a more elegant way to expose a setting in the REST API but hide it from the Custom Fields?

pierlon: In my testing the field was still visible in the REST API, unless I'm missing something?

westonruter: Yes, it is still visible. But I mean for postmeta in general. How is this handled for other plugins that expose postmeta that manipulate them via REST API but want to avoid them to be shown in the Custom Fields metabox? There's gotta be a more elegant way to do this.

pierlon: The only way to not render this field would be to mark it as protected ([ref](https://github.com/WordPress/wordpress-develop/blob/0b4e2c46044759e665f7eb516cd4bc5eb37d04d4/src/wp-admin/includes/meta-boxes.php#L755)).

schlessera: @westonruter You are probably thinking about `register_meta()`, but while that can decide whether a meta value should be shown in a REST API response or not (`'show_in_rest'`), it does not have any effect on the post meta box.

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +423,22 @@ public function preview_post_link( $link ) {
 
 		return $link;
 	}
+
+	/**
+	 * Hide the `amp_status` custom field when edition a post.
+	 *
+	 * @param bool   $protected Whether the key is considered protected.
+	 * @param string $meta_key Metadata key.
+	 * @param string $meta_type Type of object metadata is for.
```
schlessera:
```suggestion
	 * @param bool   $protected Whether the key is considered protected.
	 * @param string $meta_key  Metadata key.
	 * @param string $meta_type Type of object metadata is for.
```

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +423,22 @@ public function preview_post_link( $link ) {
 
 		return $link;
 	}
+
+	/**
+	 * Hide the `amp_status` custom field when edition a post.
```
schlessera:
```suggestion
	 * Hide the `amp_status` custom field when editing a post.
```

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +423,22 @@ public function preview_post_link( $link ) {
...
+	 *
+	 * @param bool   $protected Whether the key is considered protected.
+	 * @param string $meta_key Metadata key.
+	 * @param string $meta_type Type of object metadata is for.
+	 * @return bool
```
schlessera:
```suggestion
	 * @return bool Filtered value for whether the key should be considered protected.
```

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +423,22 @@ public function preview_post_link( $link ) {
...
+	 */
+	public function hide_amp_status_meta( $protected, $meta_key, $meta_type ) {
+		global $pagenow;
+
+		if ( is_admin() && 'post.php' === $pagenow && self::STATUS_POST_META_KEY === $meta_key && 'post' === $meta_type ) {
```
schlessera: I don't think we should filter this only on a single admin screen. For example, if someone uses  [`the_meta()`](https://developer.wordpress.org/reference/functions/the_meta/) in their theme to  output the movie attributes from their movie collection post type, we shouldn't show the AMP status either.
```suggestion
		if ( self::STATUS_POST_META_KEY === $meta_key && 'post' === $meta_type ) {
```

schlessera (changes requested):

includes/admin/class-amp-post-meta-box.php
pierlon: @schlessera I've decided to go for a more robust way of solving this issue. I've added a summary of the changes above in the OP.

includes/class-amp-post-type-support.php
```diff
@@ -49,6 +49,28 @@ public static function get_eligible_post_types() {
...
 	}
 
+	/**
+	 * Get post types that can be shown in the REST API and supports AMP.
+	 *
+	 * @return string[] Post types.
+	 */
+	public static function get_post_types_for_rest_api() {
+		if ( amp_is_canonical() ) {
+			$post_types = get_post_types_by_support( 'editor' ); // @todo Shouldn't this actually only be those with 'amp' support, or if if all_templates_supported?
```
pierlon: @westonruter Does this `todo` still apply?

westonruter: Yes, I think so. Some old code that needs to be re-evaluated.

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +429,79 @@ public function preview_post_link( $link ) {
...
 	}
+
+	/**
+	 * Add a REST API field to display the AMP status on AMP supported post types.
+	 *
+	 * @return void
+	 */
+	public function add_rest_api_fields() {
+		register_rest_field(
+			AMP_Post_Type_Support::get_post_types_for_rest_api(),
```
westonruter, line 442: Shouldn't this rather be just simply:

```suggestion
			'post',
```

I'm not seeing a way to register a field for specific post types.

includes/admin/class-amp-post-meta-box.php
```diff
@@ -88,9 +96,10 @@ public function init() {
 			self::STATUS_POST_META_KEY,
 			[
 				'sanitize_callback' => [ $this, 'sanitize_status' ],
+				'auth_callback'     => '__return_false',
```
westonruter, line 99: This has the effect of just hiding from the Custom Fields, correct?

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +429,79 @@ public function preview_post_link( $link ) {
...
+	 * @param bool    $is_enabled Whether AMP is enabled.
+	 * @param WP_Post $post       Post being updated.
+	 * @return void|WP_Error Returns an instance of `WP_Error` if the post failed to be updated.
+	 */
+	public function amp_status_update_callback( $is_enabled, $post ) {
+		if ( ! current_user_can( 'edit_post', $post->ID ) ) {
+			return new WP_Error(
+				'rest_insufficient_permission',
+				__( 'Insufficient permissions met to update the AMP status for this post', 'amp' ),
+				[ 'status' => 500 ]
```
westonruter:
```suggestion
				[ 'status' => 403 ]
```

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +429,79 @@ public function preview_post_link( $link ) {
...
+			);
+		}
+
+		if ( wp_is_post_revision( $post->ID ) || wp_is_post_autosave( $post->ID ) ) {
+			return new WP_Error(
+				'rest_invalid_post',
+				__( 'Failed to update the AMP status for this post', 'amp' ),
+				[ 'status' => 500 ]
```
westonruter:
```suggestion
				[ 'status' => 400 ]
```

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +429,79 @@ public function preview_post_link( $link ) {
...
+	public function amp_status_update_callback( $is_enabled, $post ) {
+		if ( ! current_user_can( 'edit_post', $post->ID ) ) {
+			return new WP_Error(
+				'rest_insufficient_permission',
+				__( 'Insufficient permissions met to update the AMP status for this post', 'amp' ),
+				[ 'status' => 500 ]
+			);
+		}
+
+		if ( wp_is_post_revision( $post->ID ) || wp_is_post_autosave( $post->ID ) ) {
```
westonruter: Why specifically is this condition here?

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +429,79 @@ public function preview_post_link( $link ) {
...
+	}
+
+	/**
+	 * Update the AMP status of a post via the REST API.
+	 *
+	 * @param bool    $is_enabled Whether AMP is enabled.
+	 * @param WP_Post $post       Post being updated.
+	 * @return void|WP_Error Returns an instance of `WP_Error` if the post failed to be updated.
+	 */
+	public function amp_status_update_callback( $is_enabled, $post ) {
```
westonruter: Should this have a check to see if `$post->post_type` is among `AMP_Post_Type_Support::get_post_types_for_rest_api()`?

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +429,79 @@ public function preview_post_link( $link ) {
...
+		}
+
+		$status = $is_enabled ? self::ENABLED_STATUS : self::DISABLED_STATUS;
+
+		// Note: The sanitize_callback has been supplied in the register_meta() call above.
+		update_post_meta(
+			$post->ID,
+			self::STATUS_POST_META_KEY,
+			$status
+		);
```
westonruter: Should the method return a `WP_Error` if this method returns `false`?

includes/validation/class-amp-validation-manager.php
```diff
@@ -755,21 +755,8 @@ function( $post ) {
...
-					[
-						'show_in_rest' => true,
-					]
-				)
-			);
-		}
-
 		register_rest_field(
-			$object_types,
+			AMP_Post_Type_Support::get_post_types_for_rest_api(),
```
westonruter, line 759: Is this not supposed to be just `'post'`? I don't see the post types as being allowed for this function: https://developer.wordpress.org/reference/functions/register_rest_field/

includes/admin/class-amp-post-meta-box.php
westonruter: Also asked the same below.

includes/admin/class-amp-post-meta-box.php
pierlon: Yes, when the `auth_callback` argument is supplied it [adds a filter](https://github.com/WordPress/wordpress-develop/blob/3659b767f1f8a4d3ca08f6d8310963b9d3a73b08/src/wp-includes/meta.php#L1240) called `auth_post_meta_amp_status`. That filter is later applied in the [call to `map_meta_cap()`](https://github.com/WordPress/wordpress-develop/blob/0b4e2c46044759e665f7eb516cd4bc5eb37d04d4/src/wp-includes/class-wp-user.php#L749) when `post_custom_meta_box()` checks if the user [has the `edit_post_meta` capability](https://github.com/WordPress/wordpress-develop/blob/0b4e2c46044759e665f7eb516cd4bc5eb37d04d4/src/wp-admin/includes/meta-boxes.php#L755).

Returning false for `auth_callback` means that the current user does not have the `edit_post_meta`, `add_post_meta`, nor `delete_post_meta` capability for this meta key.

includes/admin/class-amp-post-meta-box.php
pierlon: Setting this to `post` means that the field will always be displayed on the `post` object. My intention here was to not register the REST field if the `post` content type was not marked as supported on the AMP settings page. I'll update this  to reflect that intention.

includes/admin/class-amp-post-meta-box.php
pierlon: I'll be limiting the `amp_enabled` field to only  `posts`, so yes this should be checking to see if `$post->post_type` is `post`.

includes/admin/class-amp-post-meta-box.php
pierlon: I wanted to ensure post revisions nor autosaves were to be updated. Limiting the REST field to the `post` post type should also prevent this.

includes/admin/class-amp-post-meta-box.php
pierlon: Yes, I'll be adding that.

includes/validation/class-amp-validation-manager.php
pierlon: They are actually, the field will be applied as an additional property [for each object type](https://github.com/WordPress/wordpress-develop/blob/86cbde382a0d323480a14b53c86f274599880ee2/src/wp-includes/rest-api.php#L130) supplied.

includes/admin/class-amp-post-meta-box.php
pierlon: Addressed in https://github.com/ampproject/amp-wp/pull/4689/commits/49992b009c6da80bc13fc40d6410231d3e3eedbb.

includes/admin/class-amp-post-meta-box.php
pierlon: Addressed in https://github.com/ampproject/amp-wp/pull/4689/commits/49992b009c6da80bc13fc40d6410231d3e3eedbb.

includes/admin/class-amp-post-meta-box.php
pierlon: Addressed in https://github.com/ampproject/amp-wp/pull/4689/commits/02625fb09eea381c65a4d1611cdbff77fef71bb7.

includes/admin/class-amp-post-meta-box.php
```diff
@@ -422,4 +429,101 @@ public function preview_post_link( $link ) {
...
+
+	/**
+	 * Update the AMP status of a post via the REST API.
+	 *
+	 * @param bool    $is_enabled Whether AMP is enabled.
+	 * @param WP_Post $post       Post being updated.
+	 * @return void|WP_Error Returns an instance of `WP_Error` if the post failed to be updated.
+	 */
+	public function amp_status_update_callback( $is_enabled, $post ) {
+		if ( 'post' !== $post->post_type ) {
```
westonruter: Is this right? Shouldn't rather be checking if the post type is among `get_post_types_for_rest_api`? Otherwise, wouldn't this fail for `page` posts?

pierlon: Is this only supposed to register the REST field for `page` and `post` post types?

In https://github.com/ampproject/amp-wp/pull/4689/commits/91a5eb8d6014c8e85cf3047ed44dc4ff82bdb782 I've reverted to using `AMP_Post_Type_Support::get_post_types_for_rest_api()` to get the supported post types.

westonruter: My understanding is that the parameter for `register_rest_field()` is the object type not the post type (object subtype?). In other words, it could be `post`, `user`, `comment`, etc as opposed to types of post (i.e. the post types). It's confusing because `post` is overloaded being both an object type and a post type. Nevertheless it seemed to work before? So I'm confused.

pierlon: Oh, I see what you mean now. Yes the `object_type` parameter is overloaded and the name is indeed confusing. In fact, the issue has been raised before the REST API was merged into WordPress (see https://github.com/WP-API/WP-API/issues/2692).

westonruter: I shall make sure to never to `register_post_type( 'comment' )`!

westonruter:
I think I'll merge this into `develop` and then just cherry-pick the first commit adb2b7d57b76f9588e655c97c2b53af698fd7166 into the 1.5 branch for 1.5.4 since it's less drastic of a change.

includes/class-amp-post-type-support.php
westonruter: Actually, it seems to be entirely incorrect. Consider the case where you're using Standard mode but disable AMP for the pages post type:

image

The REST API currently then for pages is still including these fields:

image

This is wrong. They should be omitted entirely. They don't need to be there.

I think this condition was put in place at a time where I thought that canonical should imply that AMP cannot be disabled for any part of the site. But this is not correct.

westonruter: The method `AMP_Post_Type_Support::add_post_type_support()` will ensure that all post types have `amp` post type supported added when the “serve all templates” checkbox is checked.

westonruter:
I tried disabling AMP support for the Page post type. Then when editing a page I saw errors in the console:

* Warning: Failed prop type: The prop `isEnabled` is marked as required in `AMPToggle`, but its value is `undefined`.
* Uncaught Invariant Violation: `AMPPreview(...)`: Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.

I fixed this via 0e9d4ad.

westonruter (approved):

pierlon:
> I think I'll merge this into develop and then just cherry-pick the first commit adb2b7d into the 1.5 branch for 1.5.4 since it's less drastic of a change.

If that is still the case, I would recommend editing that commit to apply a suggestion @schlessera recommended in https://github.com/ampproject/amp-wp/pull/4689#discussion_r424930349. Actually, a separate PR may be the best way to make this change.

includes/admin/class-amp-post-meta-box.php
westonruter: Maybe instead of `is_admin()` then the check should be for `! REST_REQUEST`, since the point is to only expose it in the REST API.

westonruter:
Going a head and merging this. @schlessera can follow up with anything that needs to be amended.

@pierlon Let's skip the 1.5 branch merge for now. This can be fixed with the workaround plugin for anyone that has the issue.

PR #4698, opened 2020-05-13
Suppress Site Health ICU test if site or home URL is not an IDN

### LINKED ISSUE #4616: Site Health test for ICU should be suppressed if site/home URLs are not in IDN (punycode)
## Feature description

The only place that PHP's ICU extension is used is inside of `\AMP_HTTP::get_amp_cache_hosts()` to use the [`idn_to_utf8()`](https://www.php.net/manual/en/function.idn-to-utf8.php) function to convert the site's domain name from IDNA ASCII to Unicode:

https://github.com/ampproject/amp-wp/blob/07b638e1f931bbcdfbc9162940763da1cb301dd9/includes/class-amp-http.php#L223-L234

Nevertheless, Site Health has a test for this extension to be configured properly:

https://github.com/ampproject/amp-wp/blob/07b638e1f931bbcdfbc9162940763da1cb301dd9/src/Admin/SiteHealth.php#L194-L247

But if the `site_url()` and `home_url()` already only contain ASCII characters, then it is pointless for this test to complain about ICU not being active since it won't make any difference.

Therefore, the test should be omitted if `idn_to_utf8()` would not make any change to the domain, or else the test should be changed from `recommend` to `good`.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

With the WordPress Address or Site Address **not** being an IDN:
1. Go to the Site Health screen
2. Verify that the ICU status test is **not** shown on the page

With the WordPress Address or Site Address being an IDN (this can be tested by setting the Site Address ~to `https://⚡.com` or~ `https://xn--57h.com`, for example):
1. Go to the Site Health screen
2. Verify that the ICU status test **is shown** on the page. For example:

    

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

Fixes #4616.

## Checklist

### DIFF
--- a/src/Admin/SiteHealth.php
+++ b/src/Admin/SiteHealth.php
@@ -590,8 +590,6 @@ final class SiteHealth {
 	 * @return bool True if the `intl` extension is needed, otherwise false.
 	 */
 	public function is_intl_extension_needed() {
-		$has_idn = false;
-
 		// Publisher's own origins.
 		$domains = array_unique(
 			[
@@ -602,11 +600,10 @@ final class SiteHealth {
 
 		foreach ( $domains as $domain ) {
 			if ( preg_match( '/(^|\.)xn--/', $domain ) ) {
-				$has_idn = true;
-				break;
+				return true;
 			}
 		}
 
-		return $has_idn;
+		return false;
 	}
 }
--- a/src/Admin/SiteHealth.php
+++ b/src/Admin/SiteHealth.php
@@ -50,34 +50,11 @@ final class SiteHealth {
 			'test'  => [ $this, 'curl_multi_functions' ],
 		];
 
-		if ( function_exists( 'idn_to_utf8' ) ) {
-			$has_idn = false;
-
-			// Publisher’s own origins.
-			$domains = array_unique(
-				[
-					wp_parse_url( site_url(), PHP_URL_HOST ),
-					wp_parse_url( home_url(), PHP_URL_HOST ),
-				]
-			);
-
-			foreach ( $domains as $domain ) {
-				// The third parameter is set explicitly to prevent issues with newer PHP versions compiled with an old ICU version.
-				// phpcs:ignore PHPCompatibility.Constants.RemovedConstants.intl_idna_variant_2003Deprecated
-				$unicode_domain = idn_to_utf8( $domain, IDNA_DEFAULT, defined( 'INTL_IDNA_VARIANT_UTS46' ) ? INTL_IDNA_VARIANT_UTS46 : INTL_IDNA_VARIANT_2003 );
-
-				if ( $unicode_domain && $domain !== $unicode_domain ) {
-					$has_idn = true;
-					break;
-				}
-			}
-
-			if ( $has_idn ) {
-				$tests['direct']['amp_icu_version'] = [
-					'label' => esc_html__( 'ICU version', 'amp' ),
-					'test'  => [ $this, 'icu_version' ],
-				];
-			}
+		if ( $this->is_intl_extension_needed() ) {
+			$tests['direct']['amp_icu_version'] = [
+				'label' => esc_html__( 'ICU version', 'amp' ),
+				'test'  => [ $this, 'icu_version' ],
+			];
 		}
 
 		$tests['direct']['amp_css_transient_caching'] = [
@@ -539,33 +516,35 @@ final class SiteHealth {
 	/**
 	 * Adds suggested PHP extensions to those that Core depends on.
 	 *
-	 * @param array $extensions The existing extensions from Core.
+	 * @param array $core_extensions The existing extensions from Core.
 	 * @return array The extensions, including those for AMP.
 	 */
-	public function add_extensions( $extensions ) {
-		return array_merge(
-			$extensions,
-			[
-				'intl'     => [
-					'extension' => 'intl',
-					'function'  => 'idn_to_utf8',
-					'required'  => false,
-				],
-				'json'     => [
-					'extension' => 'json',
-					'function'  => 'json_encode',
-					'required'  => false,
-				],
-				'mbstring' => [
-					'extension' => 'mbstring',
-					'required'  => false,
-				],
-				'zip'      => [
-					'extension' => 'zip',
-					'required'  => false,
-				],
-			]
-		);
+	public function add_extensions( $core_extensions ) {
+		$extensions = [
+			'json'     => [
+				'extension' => 'json',
+				'function'  => 'json_encode',
+				'required'  => false,
+			],
+			'mbstring' => [
+				'extension' => 'mbstring',
+				'required'  => false,
+			],
+			'zip'      => [
+				'extension' => 'zip',
+				'required'  => false,
+			],
+		];
+
+		if ( $this->is_intl_extension_needed() ) {
+			$extensions['intl'] = [
+				'extension' => 'intl',
+				'function'  => 'idn_to_utf8',
+				'required'  => false,
+			];
+		}
+
+		return array_merge( $core_extensions, $extensions );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


tests/php/test-class-site-health.php
```diff
@@ -67,8 +67,21 @@ public function test_add_tests() {
...
+		// Test that the the ICU version test is added only when site URL is an IDN.
+		add_filter( 'site_url', [ self::class, 'get_idn' ], 10, 4 );
+
+		$tests = $this->instance->add_tests( [] );
+		$this->assertArrayHasKey( 'amp_icu_version', $tests['direct'] );
+
+		remove_filter( 'site_url', [ self::class, 'get_idn' ] );
+	}
+
+	public static function get_idn() {
```
westonruter: Minor thing, but this could just as well be defined as a `static` closure function stored in a `$get_idn` variable. Maybe not any better, but just an idea.

src/Admin/SiteHealth.php
```diff
@@ -49,15 +49,42 @@ public function add_tests( $tests ) {
 			'label' => esc_html__( 'cURL multi functions', 'amp' ),
 			'test'  => [ $this, 'curl_multi_functions' ],
 		];
-		$tests['direct']['amp_icu_version']             = [
-			'label' => esc_html__( 'ICU version', 'amp' ),
-			'test'  => [ $this, 'icu_version' ],
-		];
-		$tests['direct']['amp_css_transient_caching']   = [
+
+		if ( function_exists( 'idn_to_utf8' ) ) {
```
westonruter: I suppose this makes sense because if `idn_to_utf8()` is not defined then the `intl` extension won't be installed, and for sure the `INTL_ICU_VERSION` constant won't be defined. 

However, if the site doesn't have an IDN, then they don't need `idn_to_utf8()` either and shouldn't be warned about it.

Therefore, I recommend not using `idn_to_utf8()` to determine whether they need the `INTL_ICU_VERSION` constant. Instead, I suggest adding a new method like: `is_intl_extension_needed()`. Then use this method in the `SiteHealth::add_extensions` method to determine whether to add the `intl` array key. Then also use this new `is_intl_extension_needed()` method here for whether to warn about the 'INTL_ICU_VERSION` constant being absent. 

So then can `is_intl_extension_needed()` checking for the the need of `intl` be done without `idn_to_utf8()`? Can it not be done simply by checking if the domain (or subdomain too?) has `xn--`?

src/Admin/SiteHealth.php
```diff
@@ -49,15 +49,42 @@ public function add_tests( $tests ) {
...
+			$has_idn = false;
+
+			// Publisher’s own origins.
```
westonruter:
```suggestion
			// Publisher's own origins.
```

westonruter (changes requested):

tests/php/test-class-site-health.php
pierlon: If I remember correctly closures could not be removed with `remove_filter()`. I'll need to re-verify that.

westonruter: It shouldn't be a problem. Here's the logic for restoring the hooks: https://github.com/WordPress/wordpress-develop/blob/e72fff9ceffb41d22c48febfc5f97c0cd46f5884/tests/phpunit/includes/abstract-testcase.php#L293-L335

pierlon: Thanks for confirming.

src/Admin/SiteHealth.php
pierlon: > Can it not be done simply by checking if the domain (or subdomain too?) has xn--?

That would not work If I we're to set the site URL to `https://⚡.com`, for example. The domain would have to be converted to IDNA ASCII first before being compared.

pierlon: I suppose we could check if the domain begins with `xn--` or if it contains unicode, then the `intl` extension is needed.

westonruter: It looks like the `xn--` prefix can be part of a domain or subdomain: https://serverfault.com/a/142805/127453

So we'd need to split the domain by `.`, remove the TLD (e.g. `com`), and then check if any of the segments begin with `xn--`.

In regards to determining whether we need `idn_to_utf8()` in the first place, for my own edification since I'm not an expert on IDN:

* `idn_to_utf8('xn--57h.com', IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46) === '⚡.com`'
* `idn_to_ascii('⚡.com', IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46) === 'xn--57h.com`'

According to the [AMP docs](https://amp.dev/documentation/guides-and-tutorials/learn/amp-caches-and-cors/amp-cache-urls/):

> When possible, the Google AMP Cache will create a subdomain for each AMP document's domain by first converting it from IDN (punycode) to UTF-8. The caches replaces every - (dash) with -- (2 dashes) and replace every . (dot) with - (dash). For example, pub.com will map to pub-com.cdn.ampproject.org.

This being the case, if the domain is domain is already in UTF-8, then we don't have to do anything to transform it. So that means we are only concerned with hostnames that have a the IDN prefix `xn--` in the domain or a subdomain. In other words, IDN is required if and only if this is true:

```php
preg_match( '/(^|\.)xn--/', $host )
```

tests/php/test-class-site-health.php
```diff
@@ -391,5 +394,30 @@ public function test_add_extensions() {
...
+
+		remove_filter( 'site_url', [ self::class, 'get_idn' ] );
+	}
+
+	/**
+	 * Get a an IDN for testing purposes.
+	 *
+	 * @return string
+	 */
+	public static function get_idn() {
```
pierlon, line 420: Kept the static function since it's being used in multiple places now.

src/Admin/SiteHealth.php
pierlon: See https://github.com/ampproject/amp-wp/pull/4698/commits/b97582b56d53792c22caee39a4de3bf31dca2587 for changes discussed above.

src/Admin/SiteHealth.php
```diff
@@ -577,4 +583,30 @@ public function add_styles() {
...
+		);
+
+		foreach ( $domains as $domain ) {
+			if ( preg_match( '/(^|\.)xn--/', $domain ) ) {
+				$has_idn = true;
+				break;
+			}
+		}
+
+		return $has_idn;
```
westonruter:
```suggestion
		foreach ( $domains as $domain ) {
			if ( preg_match( '/(^|\.)xn--/', $domain ) ) {
				return true;
			}
		}

		return false;
```

src/Admin/SiteHealth.php
```diff
@@ -577,4 +583,30 @@ public function add_styles() {
...
 		';
 	}
+
+	/**
+	 * Determine if the `intl` extension is needed.
+	 *
+	 * @return bool True if the `intl` extension is needed, otherwise false.
+	 */
+	public function is_intl_extension_needed() {
+		$has_idn = false;
```
westonruter:
```suggestion
```

src/Admin/SiteHealth.php
pierlon: Resolved via https://github.com/ampproject/amp-wp/pull/4698/commits/da6bd14b20544ed15c6dba52eb97005f0726869e.

src/Admin/SiteHealth.php
```diff
@@ -577,4 +583,27 @@ public function add_styles() {
 			</style>
 		';
 	}
+
+	/**
+	 * Determine if the `intl` extension is needed.
+	 *
+	 * @return bool True if the `intl` extension is needed, otherwise false.
+	 */
+	public function is_intl_extension_needed() {
```
westonruter: Perhaps make this private?

```suggestion
	private function is_intl_extension_needed() {
```

westonruter (approved):

pierlon:
Looks like Travis is broken, merging the `develop` branch to give it a push.

westonruter:
Something seems off here:

> image

Why is there no merge commit?

pierlon:
Made an oopsie there, forgot to pull before merging, reverting those commits and pushing the merge commit.

pierlon:
Some Travis [jobs](https://travis-ci.org/github/ampproject/amp-wp) are still running so I'll wait for those to complete first.

amedina:
@pierlon Please specify QA/Testing instructions for this PR/issue.

pierlon:
@amedina I've updated the addressed issue with testing instructions.

PR #4741, opened 2020-05-18
Scaffolding for new onboarding wizard

### LINKED ISSUE #4718: Scaffolding for AMP Setup Wizard
## Feature description

An admin page will be created containing an empty div to serve as the onboarding application root. On that page, the minimum required core packages will be loaded, and a piece of test text will be rendered into the application root via React.

---------------

## Acceptance criteria

- `/wp-admin/admin.php?page=amp-setup` loads in all WP versions back to 4.9.
- Page contains test text loaded into the app root via React

## Implementation brief

- **PHP:** Onboarding code will consist of a single entry point wrapped in a check for a constant, `AMP_NEW_ONBOARDING`. When this constant is set to `true`, the onboarding PHP classes will be loaded. Outside this entry point, existing PHP code will not need to be modified. (The constant will be removed at the completion of UX Phase 1.) We'll also allow the screen to be loaded if a specific URL flag exists.
- **JS:** The JavaScript for the onboarding page will also be self-contained. A new `setup` directory will be created in the JS source files.
~- **Polyfilling:** A target WP version (currently 5.4) will be set in the main onboarding PHP class. Existing Webpack config will be used to build required core packages from the most up-to-date versions available from NPM and compatible with the target WP version. Where the WP version is lower than the target version, the plugin-provided versions of the packages will load in place of those provided by core.~

Note: as discussed in Slack, WP 4.9 will not be supported. 

## QA testing instructions

1. If you don't have a local dev environment, follow the steps in the “Getting Started” and “Local Environment” sections here: https://github.com/ampproject/amp-wp/wiki/Engineering-Guidelines You'll need to run `composer install && npm install && npm run build` at the end of those steps to get the plugin working.
2. Navigate to http://localhost:8890/wp-admin/admin.php?page=amp-setup&amp-new-onboarding=1 and confirm the test text loads on the screen.
3. Navigate to http://localhost:8890/wp-admin/admin.php?page=amp-options&amp-new-onboarding=1 and confirm you see "Setup Wizard" in the AMP submenu. Note that this menu item will only show if `amp-new-onboarding=1` is in the URL.

## Demo

* 

## Changelog entry

*

johnwatkins0:
### "Setup Wizard" submenu page

Sets up a new Submenu page called "Setup Wizard." This name can easily be changed. The markup on the page consists of an empty `div` serving as the root for the future onboarding wizard.

### JS directory

Creates a JS directory called `/setup`. Inside the directory is an entry file that renders test text ("Setup Wizard") into the app root for testing purposes.

### Refactor polyfills/new polyfills

~Updates the existing `polyfills` webpack config and removes some files from the `js/polyfills` directory. What those files were doing can be accomplished via Webpack only. For reference, I referred to the `output` setting in [the Gutenberg plugin webpack config](https://github.com/WordPress/gutenberg/blob/master/webpack.config.js#L54).~

~I moved some of the WP core packages from `devDependencies` to `dependencies` to allow Webpack to determine which packages need to be polyfilled.~

~The core packages compiled by the Webpack config will be loaded only if the WP version is below a target version, currently set to WP 5.4. In that version and later, the copies provided by core will be loaded instead.~

**Edit note:** The above is mostly no longer included. I did keep in the change to how `@wordpress/dom-ready`, `@wordpress/i18n`, and `@wordpress/url` were polyfilled. 

### Tests

Adds unit tests to cover the change and a small e2e test to be expanded on in future tasks.

Fixes #4718

## Checklist


johnwatkins0:
@googlebot I signed it!

johnwatkins0:
Marking this as ready for review @westonruter

assets/src/setup/index.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * WordPress dependencies
+ */
+import { render } from '@wordpress/element';
+import domReady from '@wordpress/dom-ready';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * External dependencies
```
schlessera, line 9: This is an internal dependency, not a third-party package.
```suggestion
 * Internal dependencies
```

bin/local-env/install-wordpress.sh
```diff
@@ -119,3 +119,10 @@ if [ "$SCRIPT_DEBUG" != $SCRIPT_DEBUG_CURRENT ]; then
 	SCRIPT_DEBUG_RESULT=$(wp config get --type=constant --format=json SCRIPT_DEBUG | tr -d '\r')
 	echo -e $(status_message "SCRIPT_DEBUG: $SCRIPT_DEBUG_RESULT...")
 fi
+
+AMP_NEW_ONBOARDING_CURRENT=$(wp config get --type=constant --format=json SCRIPT_DEBUG | tr -d '\r')
```
schlessera:
```suggestion
AMP_NEW_ONBOARDING_CURRENT=$(wp config get --type=constant --format=json AMP_NEW_ONBOARDING | tr -d '\r')
```

includes/admin/functions.php
```diff
@@ -194,3 +194,22 @@ function amp_bootstrap_admin() {
 	$site_health = new SiteHealth();
 	$site_health->init();
 }
+
+/**
+ * Whether to activate the new onboarding feature.
+ *
+ * @todo Remove this function on release.
+ *
+ * @return boolean
```
schlessera: WPCS requires `bool` instead of `boolean`.
```suggestion
 * @return bool
```

includes/options/class-amp-setup-wizard-submenu.php
```diff
@@ -0,0 +1,60 @@
...
+ * @package AMP
+ * @since @todo NEW_ONBOARDING_RELEASE_VERSION
+ */
+
+/**
+ * AMP setup wizard class.
+ *
+ * @since @todo NEW_ONBOARDING_RELEASE_VERSION
+ */
+class AMP_Setup_Wizard_Submenu {
```
schlessera: Classes should default to `final`, unless we have a specific use case we want to open up extensibility for.
```suggestion
final class AMP_Setup_Wizard_Submenu {
```

includes/options/class-amp-setup-wizard-submenu.php
```diff
@@ -0,0 +1,60 @@
...
+	/**
+	 * Sets up hooks.
+	 */
+	public function init() {
+		add_submenu_page(
+			$this->parent_menu_slug,
+			__( 'Setup Wizard', 'amp' ),
+			__( 'Setup Wizard', 'amp' ),
+			'manage_options',
+			static::SCREEN_ID,
```
schlessera: As the class should be `final` (and it already uses `private` and would thus probably break when being extended), we should omit late static binding when not needed, to avoid to performance hit it has.
```suggestion
			self::SCREEN_ID,
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+ * @package AMP
+ * @since @todo NEW_ONBOARDING_RELEASE_VERSION
+ */
+
+/**
+ * AMP setup wizard submenu page class.
+ *
+ * @since @todo NEW_ONBOARDING_RELEASE_VERSION
+ */
+class AMP_Setup_Wizard_Submenu_Page {
```
schlessera: Classes should default to `final`, unless we have a specific use case we want to open up extensibility for.
```suggestion
final class AMP_Setup_Wizard_Submenu_Page {
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+	 * @var string
+	 */
+	const APP_ROOT_ID = 'amp-setup';
+
+	/**
+	 * The minimum core WP version whose core assets are compatibile with the setup feature.
+	 *
+	 * @var float
+	 */
+	const TARGET_WP_VERSION = 5.4;
```
schlessera: Using floating point numbers to make exact comparisons is prone to error and should be avoided.

The version, as interpreted by floating point math depending on the architecture it runs on, could turn something like `5.4` into `5.3999999..`, throwing everything off.

This should use a string version and use `version_compare()` instead.
```suggestion
	/**
	 * The minimum core WP version whose core assets are compatibile with the setup feature.
	 *
	 * @var string
	 */
	const TARGET_WP_VERSION = '5.4';
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+	 * @return array Associative array containing the asset's dependencies and version.
+	 */
+	public function get_asset( $handle, $path = null ) {
+		/**
+		 * Filters AMP asset details.
+		 *
+		 * @param null|array $asset  Null or, to override, an array containing the asset's dependencies and version string.
+		 * @param string     $handle The asset handle.
+		 */
+		$asset = apply_filters( 'amp_setup_asset', null, $handle );
```
schlessera: I assume you're using the `null` with override here to avoid doing the file access when it would be overridden anyway?

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	/**
+	 * Registers single script.
+	 *
+	 * @param string  $handle  Script handle.
+	 * @param boolean $enqueue Whether to enqueue the asset immediately. Default false.
+	 */
+	public function add_setup_script( $handle, $enqueue = false ) {
+		$asset = $this->get_asset( $handle, sprintf( '%s/assets/js/%s.asset.php', AMP__DIR__, $handle ) );
```
schlessera: I'd prefer the following notation here, to avoid `sprintf()` and make it more readable:
```suggestion
		$asset = $this->get_asset( $handle, AMP__DIR__ . "/assets/js/{$handle}.asset.php" );
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+
+		wp_register_script(
+			$handle,
+			amp_get_asset_url( sprintf( 'js/%s.js', $handle ) ),
```
schlessera: Again, `sprintf()` can easily be avoided here for both performance and readability:
```suggestion
			amp_get_asset_url( "js/{$handle}.js" ) ),
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+	 * Enqueues setup assets.
+	 *
+	 * @param string $hook_suffix The current admin page.
+	 */
+	public function enqueue_assets( $hook_suffix ) {
+		if ( $this->screen_handle() !== $hook_suffix ) {
+			return;
+		}
+
+		$this->add_setup_script( static::JS_HANDLE, true );
```
schlessera: Avoiding late static binding as not needed when class is made `final`:
```suggestion
		$this->add_setup_script( self::JS_HANDLE, true );
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+
+		wp_localize_script(
+			static::JS_HANDLE,
```
schlessera: Avoiding late static binding as not needed when class is made `final`:
```suggestion
			self::JS_HANDLE,
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+			'ampSetup',
+			[
+				'APP_ROOT_ID' => static::APP_ROOT_ID,
```
schlessera: Avoiding late static binding as not needed when class is made `final`:
```suggestion
				'APP_ROOT_ID' => self::APP_ROOT_ID,
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+	 * @param string $hook_suffix The current admin page.
+	 */
+	public function override_scripts( $hook_suffix ) {
+		global $wp_version;
+
+		if ( $this->screen_handle() !== $hook_suffix ) {
+			return;
+		}
+
+		if ( floatval( $wp_version ) >= static::TARGET_WP_VERSION && wp_script_is( 'react', 'registered' ) ) {
```
schlessera: Avoiding late static binding as not needed when class is made `final`:
```suggestion
		if ( floatval( $wp_version ) >= self::TARGET_WP_VERSION && wp_script_is( 'react', 'registered' ) ) {
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+			return;
+		}
+
+		$scripts = wp_scripts();
+
+		$main_asset_dependencies = $this->get_asset( static::JS_HANDLE, sprintf( '%s/assets/js/%s.asset.php', AMP__DIR__, static::JS_HANDLE ) )['dependencies'];
```
schlessera: Avoiding late static binding as not needed when class is made `final`:
```suggestion
		$main_asset_dependencies = $this->get_asset( self::JS_HANDLE, sprintf( '%s/assets/js/%s.asset.php', AMP__DIR__, self::JS_HANDLE ) )['dependencies'];
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+			if ( 0 === strpos( $dependency, 'wp-' ) ) {
+				$wp_script_dependencies[] = $dependency;
+			} else {
+				$external_dependencies[] = $dependency;
+			}
+		}
+
+		foreach ( $wp_script_dependencies as $package ) {
+			if ( array_key_exists( $package, $scripts->registered ) ) {
+				$scripts->registered[ $package ]->src = amp_get_asset_url( sprintf( 'js/%s.js', $package ) );
```
schlessera:
```suggestion
				$scripts->registered[ $package ]->src = amp_get_asset_url( "js/{$package}.js" ) );
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+
+				$asset = $this->get_asset( $package, sprintf( '%s/assets/js/%s.asset.php', AMP__DIR__, $package ) );
```
schlessera:
```suggestion
				$asset = $this->get_asset( $package, AMP__DIR__ . "/assets/js/{$package}.asset.php" );
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,223 @@
...
+								]
+							),
+							'after'
+						);
+				}
+			}
+		}
+
+		foreach ( $external_dependencies as $library ) {
+			$src = amp_get_asset_url( sprintf( 'js/vendor/%s.js', $library ) );
```
schlessera:
```suggestion
			$src = amp_get_asset_url( "js/vendor/{$library}.js" );
```

tests/php/test-class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,194 @@
...
+	 *
+	 * @covers AMP_Setup_Wizard_Submenu_Page::add_setup_script
+	 *
+	 * @dataProvider get_test_setup_scripts
+	 *
+	 * @param string  $handle   Script handle
+	 * @param boolean $enqueued Whether to enqueue the script.
+	 */
+	public function test_add_setup_script( $handle, $enqueued ) {
+		$filter_asset = function( $asset, $asset_handle ) use ( $handle ) {
```
schlessera: Closure should be static as it doesn't need instanced data:
```suggestion
		$filter_asset = static function( $asset, $asset_handle ) use ( $handle ) {
```

tests/php/test-class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,194 @@
...
+	 *
+	 * @covers AMP_Setup_Wizard_Submenu_Page::get_asset
+	 */
+	public function test_get_asset() {
+		$test_data = [
+			'dependencies' => [],
+			'version'      => '1.0',
+		];
+
+		$filter_asset = function() use ( $test_data ) {
```
schlessera: Closure should be static as it doesn't need instanced data:
```suggestion
		$filter_asset = static function() use ( $test_data ) {
```

tests/php/test-class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,194 @@
...
+		$this->assertTrue( wp_script_is( $handle ) );
+	}
+
+	/**
+	 * Tests AMP_Setup_Wizard_Submenu_Page::override_scripts
+	 *
+	 * @covers AMP_Setup_Wizard_Submenu_Page::override_scripts
+	 */
+	public function test_override_scripts() {
+		$filter_asset = function( $asset, $handle ) {
```
schlessera: Closure should be static as it doesn't need instanced data:
```suggestion
		$filter_asset = static function( $asset, $handle ) {
```

schlessera (changes requested):
PR looks good, I mostly just flagged nitpicks.

Great job on streamlining the webpack configuration, love that!

As a general note regarding classes in PHP, we default to `final` & `private`, unless a use case specifically requires extensibility. This minimizes the backward compatibility implications in case we need to deprecate something later on.

schlessera:
@johnwatkins0 I'll adapt the description so that it doesn't state it "resolves" the issue anymore. This would close it immediately when merged, and leads to problem within ZenHub for the follow-up processes.

We uses `Addresses issue #...` for that reason.

includes/options/views/class-amp-setup-wizard-submenu-page.php
johnwatkins0: Yes, the immediate use case for this filter was in PHPUnit, where I'm using it across a few tests, so that the Webpack-generated files aren't required to exist. I also see it as potentially useful in future tasks where we may need a core package but not some of its dependencies. It may make sense at some point to change this filter to `amp_pre_setup_asset` and use `amp_setup_asset` to filter after the file is retrieved.

assets/src/setup/index.js
johnwatkins0: Unfortunately ESLint won't allow this change because the import is interpreted as external, but I added a comment to the import to make sure it's clear where it's coming from.

johnwatkins0:
Thank you for the process note, @schlessera, and great suggestions. I've addressed all the comments. Let me know if you see anything else.

johnwatkins0:
In discussions today it was decided we would not support 4.9 with the setup wizard. I've pushed some updates to remove the polyfilling I'd had in place for 4.9. This made the diff on the PHP side a bit smaller and less complex.

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,197 @@
...
+
+	/**
+	 * HTML ID for the app root element.
+	 *
+	 * @var string
+	 */
+	const APP_ROOT_ID = 'amp-setup';
+
+	/**
+	 * The minimum core WP version whose core assets are compatibile with the setup feature.
```
pierlon:
```suggestion
	 * The minimum core WP version whose core assets are compatible with the setup feature.
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,197 @@
...
+
+		if ( $this->screen_handle() !== $hook_suffix ) {
+			return;
+		}
+
+		if ( ! $this->should_override_scripts() ) {
+			return;
+		}
+
+		$bundled_core_packages = [
```
pierlon: The paired browsing client and app also depends on some of these polyfilled assets, and is currently supported up to WP 4.9. We will need to disable it for WP 5.0, possibly in a subsequent PR.

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,197 @@
...
+
+	/**
+	 * Whether the plugin should override scripts for the onboarding screen.
+	 *
+	 * @return boolean
+	 */
+	public function should_override_scripts() {
+		global $wp_version;
+
+		$should_override_scripts = version_compare( $wp_version, self::TARGET_WP_VERSION, '<' ) && version_compare( $wp_version, '5.0', '>=' );
```
pierlon: One concern here is overriding an asset that may have a higher version than the one it's replacing. That asset may have breaking changes and may inadvertently break WP or plugins on that page.

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,197 @@
...
+
+			$wp_scripts->registered[ $package ]->ver = $asset['version'];
+		}
+
+		foreach ( $bundled_external_libraries as $library ) {
+			if ( ! array_key_exists( $library, $wp_scripts->registered ) ) {
+				continue;
+			}
+
+			$wp_scripts->registered[ $library ]->ver = AMP__VERSION;
```
pierlon: Any reason why you chose to set these libraries to the current version of the plugin? I know the alternative would be having to manually keep the versions up to date, but I'm wondering if this would have any adverse effects in WP.

tests/php/test-class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,236 @@
...
+
+	/**
+	 * Setup.
+	 *
+	 * @inheritdoc
+	 */
+	public function setUp() {
+		parent::setUp();
+
+		$this->old_wp_scripts = isset( $GLOBALS['wp_scripts'] ) ? $GLOBALS['wp_scripts'] : null;
```
pierlon: Properties `old_wp_scripts` and `page` are being declared dynamically; they should be declared as a class property.

tests/php/test-class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,236 @@
...
+
+		$this->page = new AMP_Setup_Wizard_Submenu_Page( 'amp-options' );
```
pierlon:
```suggestion
		$this->page = new AMP_Setup_Wizard_Submenu_Page();
```

tests/php/test-class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,236 @@
...
+	public function tearDown() {
+		parent::tearDown();
+
+		add_action( 'wp_default_scripts', 'wp_default_scripts' );
+
+		if ( function_exists( 'wp_default_packages' ) ) {
+			add_action( 'wp_default_scripts', 'wp_default_packages' );
+		}
+
+		$GLOBALS['wp_scripts'] = new WP_Scripts();
```
pierlon: Why not set it back to `$this->old_wp_scripts`?

```suggestion
		$GLOBALS['wp_scripts'] = $this->old_wp_scripts;
```

tests/php/test-class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,236 @@
...
+	 * @inheritdoc
+	 */
+	public function tearDown() {
+		parent::tearDown();
+
+		add_action( 'wp_default_scripts', 'wp_default_scripts' );
+
+		if ( function_exists( 'wp_default_packages' ) ) {
+			add_action( 'wp_default_scripts', 'wp_default_packages' );
+		}
```
pierlon: Not sure what's going on here.

tests/php/test-class-amp-setup-wizard-submenu.php
```diff
@@ -0,0 +1,44 @@
...
+
+	/**
+	 * Setup.
+	 *
+	 * @inheritdoc
+	 */
+	public function setUp() {
+		parent::setUp();
+
+		$this->wizard = new AMP_Setup_Wizard_Submenu( 'amp-options' );
```
pierlon, line 34: Dynamically created property should be declared first as a class property.

.gitignore
```diff
@@ -7,6 +7,7 @@ amp.zip
 /assets/css/*
 !/assets/css/src/
 assets/js/*.js
+assets/js/vendor
```
pierlon:
```suggestion
assets/js/**/*.js
```

pierlon (changes requested):
Great work so far!

includes/options/views/class-amp-setup-wizard-submenu-page.php
johnwatkins0: @pierlon In an email discussion with some Google folks who worked on Site Kit it was determined all this polyfilling wasn't worth the effort. Reach out if you want additional details.

johnwatkins0:
@pierlon @schlessera This has changed a bit since your earlier reviews. Essentially we're not going to do any of the script juggling on the backend due to the complication that would be involved in maintaining it (exactly as Alain was saying yesterday). Instead, we're just going to bundle everything with Webpack, and if there are any bundle optimizations needed on these screens we can handle it with Webpack/JS.

Re-requesting review for when you have a chance.

pierlon (approved):
Looks good! LGTM.

includes/admin/functions.php
```diff
@@ -194,3 +194,28 @@ function amp_bootstrap_admin() {
...
+
+/**
+ * Whether to activate the new onboarding feature.
+ *
+ * @return bool
+ */
+function amp_should_use_new_onboarding() {
+	global $wp_version;
+
+	if ( version_compare( $wp_version, '5.0', '<' ) ) {
```
westonruter: Minor thing, but instead of getting the global variable you could rather do:

```suggestion
	if ( version_compare( get_bloginfo( 'version' ), '5.0', '<' ) ) {
```

includes/admin/functions.php
```diff
@@ -194,3 +194,28 @@ function amp_bootstrap_admin() {
...
+		return false;
+	}
+
+	// @todo Remove this check when the onboarding feature is released.
+	if ( '1' === filter_input( INPUT_GET, 'amp-new-onboarding', FILTER_SANITIZE_NUMBER_INT ) ) {
```
westonruter: Note that I generally don't recommend using `filter_input()` because (1) core doesn't use it, and (2) as far as I know the input values can't be mutated during unit testing.

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,80 @@
...
+			true
+		);
+
+		wp_localize_script(
+			self::JS_HANDLE,
+			'ampSetup',
+			[
+				'APP_ROOT_ID' => self::APP_ROOT_ID,
+			]
+		);
```
westonruter: I recommend using `wp_add_inline_script()` and manually export data via `wp_json_encode()`. The reason for this is that `wp_localize_script()` casts all array values to strings, which can be problematic when exporting integers and booleans. This isn't an issue here, but it's something to keep in mind.

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,80 @@
...
+	public function enqueue_assets( $hook_suffix ) {
+		if ( $this->screen_handle() !== $hook_suffix ) {
+			return;
+		}
+
+		wp_enqueue_script(
+			self::JS_HANDLE,
+			amp_get_asset_url( 'js/' . self::JS_HANDLE . '.js' ),
+			[],
+			AMP__VERSION,
```
westonruter: Shouldn't this be using the `version` as exported into `amp-setup.asset.php`?

includes/options/views/class-amp-setup-wizard-submenu-page.php
```diff
@@ -0,0 +1,80 @@
...
+	 */
+	public function enqueue_assets( $hook_suffix ) {
+		if ( $this->screen_handle() !== $hook_suffix ) {
+			return;
+		}
+
+		wp_enqueue_script(
+			self::JS_HANDLE,
+			amp_get_asset_url( 'js/' . self::JS_HANDLE . '.js' ),
+			[],
```
westonruter: Shouldn't this be using the `dependencies` as exported into `amp-setup.asset.php`?

includes/admin/functions.php
westonruter: Granted this is just a feature flag, so no big deal.

includes/options/views/class-amp-setup-wizard-submenu-page.php
westonruter: Actually I'm not seeing an `amp-setup.asset.php` being generated like for the other scripts. Why?

webpack.config.js
```diff
@@ -210,12 +236,23 @@ const wpPolyfills = {
...
-		'wp-dom-ready': './assets/src/polyfills/wp-dom-ready.js',
-		'wp-polyfill': './assets/src/polyfills/wp-polyfill.js',
-		'wp-url': './assets/src/polyfills/wp-url.js',
+		'amp-setup': './assets/src/setup',
+	},
+	externals: {
+		'amp-setup': 'ampSetup',
 	},
+	plugins: [
+		new WebpackBar( {
```
pierlon, line 261:
```suggestion
		...defaultConfig.plugins,
		new WebpackBar( {
```

includes/options/views/class-amp-setup-wizard-submenu-page.php
pierlon: Good catch. The `DependencyExtractionWebpackPlugin` plugin is not being applied in the Webpack config. I'll make a suggestion.

johnwatkins0: If we use the dependencies from the dependency extraction plugin here, this will by default end up enqueuing all the core packages that we'll now be bundling into our JS. So we need the dependency array to be empty. We can definitely use the generated version instead of AMP__VERSION here, though, so I'll make that update with @pierlon's suggestion below.

westonruter (approved):

PR #4746, opened 2020-05-20
Suppress display of fatal errors during validate request

### LINKED ISSUE #4747: Suppress validation errors that may occur from fatal errors being displayed
## Bug Description

While performing QA on #4580, @schlessera brought to light an issue with the number of validation errors that should be displayed when validating a page with fatal errors. It was found that the number of validation errors being reported for a page with fatal errors being displayed can differ, depending on the environment the plugin is in (see https://github.com/ampproject/amp-wp/issues/4580#issuecomment-630835233).

To prevent this discrepancy, we can set the PHP configuration value `display_errors` to `false` before the validation occurs, as suggested by @westonruter in https://github.com/ampproject/amp-wp/issues/4580#issuecomment-630976083.

Related: #4580.

## Expected Behaviour

Validation errors should not be reported for fatal errors being displayed on a page.

## Steps to reproduce

1. Set the `display_errors` PHP configuration value to `true`
2. Download, install, and activate the following gist plugin: https://gist.github.com/pierlon/cb3dc7e78ae99ba48e377d8256de5cb5
3. On the frontend of the site, the admin bar should be showing 8 validation errors:
image
4. When viewing the validation errors for the screen, more than 4 validation errors should be reported, instead of the 4 actual errors on the page:
image

## Additional context

 - WordPress version:
 - Plugin version:
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

pierlon:
## Summary

This PR suppresses the display of any fatal errors that may arise during validation. When these errors are displayed, they may be flagged as validation errors and can can give a false count of the actual state of validation errors for the page. 

Fixes #4747.

## Demo

With this [gist plugin](https://gist.github.com/pierlon/cb3dc7e78ae99ba48e377d8256de5cb5) activated:

- Additional validation errors should be reported when fatal errors are displayed:

- Only relevant errors are reported when the display of fatal errors are suppressed:

## Checklist

### DIFF
--- a/includes/validation/class-amp-validation-manager.php
+++ b/includes/validation/class-amp-validation-manager.php
@@ -601,12 +601,12 @@ class AMP_Validation_Manager {
 			self::add_validation_error_sourcing();
 			self::$is_validate_request = true;
 
-			$display_errors = ini_get( 'display_errors' );
+			$display_errors = (string) ini_get( 'display_errors' );
 
-			if ( ! empty( $display_errors ) || 'stderr' !== $display_errors ) {
+			if ( '1' === $display_errors ) {
 				// Suppress the display of fatal errors that may arise during validation so that they will not be counted
 				// as actual validation errors.
-				ini_set( 'display_errors', false );
+				ini_set( 'display_errors', 0 ); // phpcs:ignore WordPress.PHP.IniSet.display_errors_Blacklisted
 			}
 		} else {
 			self::$is_validate_request = false;
--- a/includes/validation/class-amp-validation-manager.php
+++ b/includes/validation/class-amp-validation-manager.php
@@ -601,9 +601,7 @@ class AMP_Validation_Manager {
 			self::add_validation_error_sourcing();
 			self::$is_validate_request = true;
 
-			$display_errors = (string) ini_get( 'display_errors' );
-
-			if ( '1' === $display_errors ) {
+			if ( '1' === (string) ini_get( 'display_errors' ) ) {
 				// Suppress the display of fatal errors that may arise during validation so that they will not be counted
 				// as actual validation errors.
 				ini_set( 'display_errors', 0 ); // phpcs:ignore WordPress.PHP.IniSet.display_errors_Blacklisted
--- a/includes/validation/class-amp-validation-manager.php
+++ b/includes/validation/class-amp-validation-manager.php
@@ -600,6 +600,14 @@ class AMP_Validation_Manager {
 		if ( true === $should_validate_response ) {
 			self::add_validation_error_sourcing();
 			self::$is_validate_request = true;
+
+			$display_errors = ini_get( 'display_errors' );
+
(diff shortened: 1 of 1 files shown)


includes/validation/class-amp-validation-manager.php
```diff
@@ -600,6 +600,14 @@ public static function init_validate_request() {
 		if ( true === $should_validate_response ) {
 			self::add_validation_error_sourcing();
 			self::$is_validate_request = true;
+
+			$display_errors = ini_get( 'display_errors' );
+
+			if ( ! empty( $display_errors ) || 'stderr' !== $display_errors ) {
```
pierlon: If the value of `display_errors` is set to `stderr` it will not be send to `stdout`, as documented in the [manual](https://www.php.net/manual/en/errorfunc.configuration.php#ini.display-errors).

includes/validation/class-amp-validation-manager.php
```diff
@@ -600,6 +600,14 @@ public static function init_validate_request() {
...
+				// Suppress the display of fatal errors that may arise during validation so that they will not be counted
+				// as actual validation errors.
+				ini_set( 'display_errors', false );
```
westonruter: WordPress core is using `0` here:

```suggestion
				ini_set( 'display_errors', 0 );
```

includes/validation/class-amp-validation-manager.php
westonruter: Should we just check for the value of `'1'`?

```suggestion
			$display_errors = (string) ini_get( 'display_errors' );

			if ( '1' === $display_errors ) {
```

pierlon: Good point.

includes/validation/class-amp-validation-manager.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/4746/commits/0614c80de60edf4aff4859250ce063fee0874af4.

includes/validation/class-amp-validation-manager.php
pierlon: Done in https://github.com/ampproject/amp-wp/pull/4746/commits/0614c80de60edf4aff4859250ce063fee0874af4.

westonruter (approved):

includes/validation/class-amp-validation-manager.php
```diff
@@ -600,6 +600,14 @@ public static function init_validate_request() {
 		if ( true === $should_validate_response ) {
 			self::add_validation_error_sourcing();
 			self::$is_validate_request = true;
+
+			$display_errors = (string) ini_get( 'display_errors' );
+
+			if ( '1' === $display_errors ) {
```
westonruter: Minor improvement:

```suggestion
			if ( '1' === (string) ini_get( 'display_errors' ) ) {
```

PR #4759, opened 2020-05-22
Omit non-standard type attribute during <amp-iframe> normalization

### LINKED ISSUE #4758: Post with Amazon Kindle block producing validation error
## Bug Description

In the WordPress Block Editor, if the **Amazon Kindle** block is added then a validation error occurs.

As there is a `type="text/html"` attribute-value combination present in the iframe generated from the **Amazon Kindle** embed and `type` is not a standard `<iframe>` attribute, that is why the validation error occurs. Though `type` is a non-standard attribute, it is fairly common in some embeds.

## Expected Behaviour

Embeds with fairly common non-standard attributes like `type` should not produce validation errors.

## Steps to reproduce
1. Create a post with **Amazon Kindle** block.
2. Validate the post.
3. See the validation error in back-end.

## Screenshots

**Validation error with Amazon Kindle block**

image

## Additional context

 - WordPress version:
 - Plugin version:
 - AMP plugin template mode:
 - PHP version:
 - OS:

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

itowhid06:
## Summary

- Omit non-standard `type` attribute  during `<amp-iframe>` normalization

Fixes #4758

## Checklist

### DIFF
--- a/includes/sanitizers/class-amp-iframe-sanitizer.php
+++ b/includes/sanitizers/class-amp-iframe-sanitizer.php
@@ -221,7 +221,7 @@ class AMP_Iframe_Sanitizer extends AMP_Base_Sanitizer {
 	 *      @type int $frameborder <iframe> `frameborder` attribute - Filter to '0' or '1'; default to '0'
 	 *      @type bool $allowfullscreen <iframe> `allowfullscreen` attribute - Convert 'false' to empty string ''
 	 *      @type bool $allowtransparency <iframe> `allowtransparency` attribute - Convert 'false' to empty string ''
-	 *      @type string $type <iframe> `type` attribute - Pass along if found
+	 *      @type string $type <iframe> `type` attribute - Pass along if value is not `text/html`
 	 * }
 	 * @return array Returns HTML attributes; normalizes src, dimensions, frameborder, sandbox, allowtransparency and allowfullscreen
 	 */
@@ -307,9 +307,12 @@ class AMP_Iframe_Sanitizer extends AMP_Base_Sanitizer {
 
 				case 'type':
 					/*
-					 * Though type is non-standard attribute in <iframe> tag, some embeds (i.e Amazon Kindle Embed) uses
-					 * it, therefore we need to omit this.
+					 * Omit the `type` attribute if its value is `text/html`. Popular embed providers such as YouTube
+					 * and Amazon use this non-standard attribute.
 					 */
+					if ( 'text/html' !== strtolower( $value ) ) {
+						$out[ $name ] = $value;
+					}
 					break;
 
 				default:
--- a/includes/sanitizers/class-amp-iframe-sanitizer.php
+++ b/includes/sanitizers/class-amp-iframe-sanitizer.php
@@ -221,6 +221,7 @@ class AMP_Iframe_Sanitizer extends AMP_Base_Sanitizer {
 	 *      @type int $frameborder <iframe> `frameborder` attribute - Filter to '0' or '1'; default to '0'
 	 *      @type bool $allowfullscreen <iframe> `allowfullscreen` attribute - Convert 'false' to empty string ''
 	 *      @type bool $allowtransparency <iframe> `allowtransparency` attribute - Convert 'false' to empty string ''
+	 *      @type string $type <iframe> `type` attribute - Pass along if found
 	 * }
 	 * @return array Returns HTML attributes; normalizes src, dimensions, frameborder, sandbox, allowtransparency and allowfullscreen
 	 */
@@ -302,6 +303,13 @@ class AMP_Iframe_Sanitizer extends AMP_Base_Sanitizer {
 
 				case 'data-amp-overflow-text':
 					// No need to copy.
+					break;
+
+				case 'type':
+					/*
+					 * Though type is non-standard attribute in <iframe> tag, some embeds (i.e Amazon Kindle Embed) uses
+					 * it, therefore we need to omit this.
+					 */
 					break;
 
 				default:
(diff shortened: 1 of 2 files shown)


westonruter:
I think the better approach would be to handle this at the embed level, rather than the sanitizer level, if this is primarily an issue with Kindle embeds. /cc @pierlon

Otherwise, that validation error should probably be raised so that authors know they have a non-standard attribute that should be removed. See also https://github.com/Automattic/jetpack/pull/15225.

itowhid06:
Hello @westonruter,

Yes, we can handle this at embed level, but then again this issue may occur from some other embed code in the future.

westonruter:
Yeah, it depends on how common this bogus attribute is.

pierlon:
> I think the better approach would be to handle this at the embed level, rather than the sanitizer level, if this is primarily an issue with Kindle embeds.

Agreed. I haven't been able to find any resources on how this nonstandard attribute came about, but for the meantime I think it's best to not ignore this since it is not valid HTML.

westonruter:
@pierlon So you think it _should_ be done in the sanitizer for all iframes and not in an embed handler for just Kindle?

Perhaps it should only be ignored when it has a value of text/html?

pierlon:
Ignoring the `type` attribute if it's value is `text/html` does seem like a good compromise. Otherwise from that, a validation error should be reported.

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -221,6 +221,7 @@ public function sanitize() {
 	 *      @type int $frameborder <iframe> `frameborder` attribute - Filter to '0' or '1'; default to '0'
 	 *      @type bool $allowfullscreen <iframe> `allowfullscreen` attribute - Convert 'false' to empty string ''
 	 *      @type bool $allowtransparency <iframe> `allowtransparency` attribute - Convert 'false' to empty string ''
+	 *      @type string $type <iframe> `type` attribute - Pass along if found
```
pierlon:
```suggestion
	 *      @type string $type <iframe> `type` attribute - Pass along if value is not `text/html`
```

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -304,6 +305,13 @@ private function normalize_attributes( $attributes ) {
 					// No need to copy.
 					break;
 
+				case 'type':
+					/*
+					 * Though type is non-standard attribute in <iframe> tag, some embeds (i.e Amazon Kindle Embed) uses
+					 * it, therefore we need to omit this.
+					 */
+					break;
+
```
pierlon:
```suggestion
				case 'type':
					/*
					 * Omit the `type` attribute if its value is `text/html`. Popular embed providers such as YouTube
					 * and Amazon use this non-standard attribute.
					 */
					if ( 'text/html' !== strtolower( $value ) ) {
						$out[ $name ] = $value;
					}
					break;
```

pierlon (changes requested):
An accompanying test will be needed for this change. Here's an example of how the `security` attribute is tested:

https://github.com/ampproject/amp-wp/blob/5d97c5fd5ca481b0e35a311c7a28b7b9e15e0177/tests/php/test-amp-iframe-sanitizer.php#L469-L477

includes/sanitizers/class-amp-iframe-sanitizer.php
westonruter: Is this a `bool`?

pierlon: Ah no :sweat_smile:. Fixed it.

includes/sanitizers/class-amp-iframe-sanitizer.php
westonruter: Also, the YouTube embed code doesn't use this `type` attribute as far as I know. It's only the Jetpack embed handler for YouTube that used to use it (but now does no longer).

pierlon: Oh I see. I dug further into this, and I believe the `type` attribute may be a remnant of when the YouTube API used to return `<object>` and `<embed>` tags, before the switch to [iframes in 2011](https://youtube-eng.googleblog.com/2011/05/youtube-google-io-iframe-player-api_31.html). The `<object>` tag must have either a `data` or `type` attribute defined, and the `<embed>` would have a `type` attribute as well. The `type` attribute would specify the MIME type of the resource being embedded.

[This issue](https://bugzilla.mozilla.org/show_bug.cgi?id=769117) reported on Bugzilla does refer to the `<object>` embed having the `type` attribute set to `text/html`, for example.

Fun fact: you can still get `<object>` embeds by setting the `iframe` query argument to `0` in an oEmbed URL.

itowhid06:
Hello @westonruter & @pierlon,

I have implemented the suggested changes and the test case. Please let me know if I am missing something or anything else needs to be done.

Thank you.

includes/sanitizers/class-amp-iframe-sanitizer.php
```diff
@@ -304,6 +305,16 @@ private function normalize_attributes( $attributes ) {
 					// No need to copy.
 					break;
 
+				case 'type':
+					/*
+					 * Omit the `type` attribute if its value is `text/html`. Popular embed providers such as YouTube
+					 * and Amazon use this non-standard attribute.
+					 */
```
westonruter:
```suggestion
					/*
					 * Omit the `type` attribute if its value is `text/html`. Popular embed providers such as Amazon
					 * Kindle use this non-standard attribute, which is apparently a vestige from usage on <object>.
					 */
```

itowhid06: @westonruter, I have updated the reasoning behind omitting `type=“text/html”` attribute as you suggested.

westonruter (approved):

pierlon (approved):

